PR #5405, opened 2015-07-23
fix for underdamped avatar rotation momentum

AndrewMeadows:
I dunno why this broke.  The code in MyAvatar::updateOrientation() is wrong, but hasn't been changed in a while.  My guess is that this code was not being called -- something else was updating orientation.   That other path was removed and now we were falling into this old and buggy path.  In any case, here is a fix.  If anyone things it needs to be tuned (turn faster, ramp up or down quicker) just let me know.

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -54,7 +54,7 @@
 using namespace std;
 
 const glm::vec3 DEFAULT_UP_DIRECTION(0.0f, 1.0f, 0.0f);
-const float YAW_SPEED = 500.0f;   // degrees/sec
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: b26f4b0c8d034f298bc317db6fbb157cc66992d1 available at: https://gist.github.com/e2fab4dd28829af1363b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5405.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5405.dmg

PR #5408, opened 2015-07-23
CR for Job #20626 - edit.js - entities list html - multi select using shift key / working delete key

thoys:
Code Review for Job #20626 - Workitem available at https://www.worklist.net/20626


sethalves:
I've tested this on Linux and it looks good.

PR #5412, opened 2015-07-24
decouple FPS from rotation rate during mouse drag

AndrewMeadows:
Seth noticed that the speed at which the avatar's view rotated depended on the FPS: during times of low frame rates the avatar would spin faster.  This is a theoretical fix for the problem.  I call it "theoretical" because I was having trouble testing because of render problems on linux (the skybox (?) was causing everything to be a solid soft pink color... but only most of the time and not in the mirror window).


hifi-gustavo:
Build console output for commit: 2fc158246dd8504bc524182f708286d2a2529993 available at: https://gist.github.com/88dc8fb8861f8cdf6b6b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5412.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5412.dmg

PR #5414, opened 2015-07-24
allow the QEvent queue in Application to get behind

birarda:
This brings back the previous idleTimer restart we had that allowed the Application to stay mostly in time for UI events but with touch events definitely getting backed up. For now this is better than having the window lock up.

Tomorrow I'll see if I can find which events specifically are taking lots of time to process on the main thread.


hifi-gustavo:
Build console output for commit: cdcab299b65c9ca400881e1e25c13be104279b5f available at: https://gist.github.com/7d63113cce79696a7d6b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5414.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5414.dmg

PR #5415, opened 2015-07-24
Overlay Fade Feature

hyperlogic:
NOTE: This a PR into the plugins branch **NOT** master.

Overlay can be shown/hidden using the middle-mouse button.  In standing VR mode, the overlay will fade away automatically as you walk or look away from it.  2d and seated VR overlay, should behave the same, except that it also has the ability to be hidden with the middle mouse.

This PR also addresses recent changes to paintGL, which affected HTC Vive compatibility.

A new class was introduced, OverlayConductor, a member variable of Application.  It is updated every-frame and keeps the both the AnimationCompositor and Overlays object in sync with the users desire to show or hide the overlay as a whole.

### DIFF
--- a/interface/src/ui/ApplicationCompositor.cpp
+++ b/interface/src/ui/ApplicationCompositor.cpp
@@ -283,14 +283,11 @@ void ApplicationCompositor::displayOverlayTextureHmd(RenderArgs* renderArgs, int
     batch.setResourceTexture(0, overlayFramebuffer->getRenderBuffer(0));
 
     mat4 camMat;
-    _cameraTransform.getMatrix(camMat);
-    camMat = camMat * qApp->getEyeOffset(eye);
+    _cameraBaseTransform.getMatrix(camMat);
+    camMat = camMat * qApp->getEyePose(eye);
     batch.setViewTransform(camMat);
 
     batch.setProjectionTransform(qApp->getEyeProjection(eye));
-
-    mat4 eyePose = qApp->getEyePose(eye);
-    glm::mat4 overlayXfm = glm::inverse(eyePose);
 
 #ifdef DEBUG_OVERLAY
     {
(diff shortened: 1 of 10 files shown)


hifi-gustavo:
Build console output for commit: 87a0e48d306e4912c609e6fb1d94a6c8243443e2 available at: https://gist.github.com/d48e2d9728dfbfac61a8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5415.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5415.dmg

interface/src/ui/ApplicationCompositor.cpp
```diff
@@ -619,6 +627,19 @@ void ApplicationCompositor::drawSphereSection(gpu::Batch& batch) {
     batch.setInputFormat(streamFormat);
 
     static const int VERTEX_STRIDE = sizeof(vec3) + sizeof(vec2) + sizeof(vec4);
+
+    if (_prevAlpha != _alpha) {
```
jherico: It seems like it would be a lot more efficient to have an alpha uniform you could pass into the shader rendering the overlay.

interface/src/ui/ApplicationCompositor.cpp
```diff
@@ -708,3 +729,13 @@ void ApplicationCompositor::updateTooltips() {
         }
     }
 }
+
+void ApplicationCompositor::update(float dt) {
+    const int ALPHA_FADE_RATE = 2.0f;
```
jherico: Don't write this logic yourself.  use QEasingCurve

jherico:
Looks pretty good, just a couple minor comments.  If the alpha thing is a pain, we can defer it, but updating an entire VBO just to change the alpha value is pretty inefficient.

hyperlogic:
Yeah, the alpha thing is a bit of a pain.  It would require creating a new render::pipeline and new shaders for the hemisphere. How about I create an asana task for this, so we don't forget it.

PR #5420, opened 2015-07-24
Example script solarsystem.js with orbiting satellite game

bwent:
The start to a project to build a virtual physics classroom to simulate the solar system, gravity, and orbital physics. 
      - A sun with oribiting planets is created in front of the user
      - UI elements allow for adjusting the period, gravity, trails, and energy recalculations
      - Click "PAUSE" to pause the animation and show planet labels
      - In this mode, double-click a planet label to zoom in on that planet
             -Double-clicking on earth label initiates satellite orbiter game
             -Press "TAB" to toggle back to solar system view

### DIFF
--- a/examples/example/games/satellite.js
+++ b/examples/example/games/satellite.js
@@ -3,13 +3,13 @@
 //  games
 //  
 //  Created by Bridget Went 7/1/2015. 
-//	Copyright 2015 High Fidelity, Inc.
+//  Copyright 2015 High Fidelity, Inc.
 //  
 //  A game to bring a satellite model into orbit around an animated earth model . 
-//  	- Double click to create a new satellite
-//		- Click on the satellite, drag a vector arrow to specify initial velocity
-//		- Release mouse to launch the active satellite
-//		- Orbital movement is calculated using equations of gravitational physics
+//      - Double click to create a new satellite
+//      - Click on the satellite, drag a vector arrow to specify initial velocity
+//      - Release mouse to launch the active satellite
+//      - Orbital movement is calculated using equations of gravitational physics
 // 
 //  Distributed under the Apache License, Version 2.0.
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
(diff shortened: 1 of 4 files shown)


cozza13:
ok to test

cozza13:
test this please

examples/example/games/satellite.js
```diff
@@ -0,0 +1,278 @@
...
+	Controller.mouseDoublePressEvent.connect(mouseDoublePressEvent);
+	Controller.mouseMoveEvent.connect(mouseMoveEvent);
+	Controller.mouseReleaseEvent.connect(mouseReleaseEvent);
+	Script.update.connect(update);
+	Script.scriptEnding.connect(this.endGame);
+
+}
+	
+
+
```
AndrewMeadows: This file appears to be tab-indented but should be using spaces instead.

examples/example/games/satellite.js
```diff
@@ -0,0 +1,278 @@
...
+		if(!activeSatellite) {
+			return;
+		}
+		// Update all satellites
+		for (var i = 0; i < satellites.length; i++) {
+			if (!satellites[i].launched) {
+				return;
+			}
+			satellites[i].update(deltaTime);	
+		}
```
AndrewMeadows: This **for** loop will update each satellite until it finds one that has not been **launched** at which point it **returns** and bypasses the code below.  Is this intentional?  Maybe the loop should **break** at or **continue** over unlaunched satellites?

examples/example/games/satellite.js
```diff
@@ -0,0 +1,278 @@
...
+		if(!activeSatellite || !activeSatellite.arrow || !activeSatellite.arrow.isDragging) {
+			return;
+		}
+		activeSatellite.arrow.onMouseReleaseEvent(event);
+	  	activeSatellite.launch();
+	  	activeSatellite.arrow.cleanup();
+	}
+
+	var counter = 0.0;
+	var TIME = 500;
```
AndrewMeadows: Does not block the merge: From how it is used I would guess that a better name for **TIME** would be something like:  **PROPERTIES_UPDATE_PERIOD**.

examples/example/games/satellite.js
```diff
@@ -0,0 +1,278 @@
...
+SatelliteGame = function() {
+	var MAX_RANGE = 50.0;
+	var Y_AXIS = {
+		x: 0,
+		y: 1,
+		z: 0
+	}
+	var LIFETIME = 6000;
+
+	var v0;
```
AndrewMeadows: **v0** is declared here as global to the **SatelliteGame** context, but as far as I can tell this instance is not used: **v0** is redeclared in later contexts that actually use it.  I don't know JavaScript very well so it isn't quite clear to me if this declaration is unused or just recycled.  I'd be in favor of removing this one and keeping those that are declared and used in local context.

BTW, the name **v0** isn't very descriptive, nor are the variables below (T, M, m).  While this is technically against coding policy I'm OK with it when the variables are used in a physics formula where their meaning may be clear in context.  I think these fall into that category, although I'd prefer to see them declared closer to where they are actually used (e.g. in calcEnergyError()).

examples/utilities/tools/vector.js
```diff
@@ -0,0 +1,197 @@
...
+        position: statsPosition,
+        dimensions: {
+          x: 4.0,
+          y: 1.5, 
+          z: 0.1
+        },
+        lineHeight: 0.3,
+        faceCamera: true
+      });
+    }
```
AndrewMeadows: There are some magic constants here when showing the stats.  Maybe they should be pulled out with some of the other global constants (such as **LINE_DIMENSIONS** and friends)?

examples/example/games/satellite.js
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	Controller.mousePressEvent.connect(mousePressEvent);
+	Controller.mouseDoublePressEvent.connect(mouseDoublePressEvent);
+	Controller.mouseMoveEvent.connect(mouseMoveEvent);
+	Controller.mouseReleaseEvent.connect(mouseReleaseEvent);
+	Script.update.connect(update);
+	Script.scriptEnding.connect(this.endGame);
+
+}
```
AndrewMeadows: The indentation of this file is still using tabs on most (but not all) lines.

PR #5431, opened 2015-07-25
Limit the amount of time consumed by QML rendering

jherico:
QML rendering can consume 10ms per call and occurs on the main thread currently.  Until we move it to a distinct thread, this modification should limit QML updates to no more than 10 times per second.  This introduces some choppiness in the QML items (most noticeable during scrolling or dragging operations) but should stop the gorging of runtime by QML during touch events.

### DIFF
--- a/libraries/render-utils/src/OffscreenQmlSurface.cpp
+++ b/libraries/render-utils/src/OffscreenQmlSurface.cpp
@@ -19,6 +19,7 @@
 
 #include "FboCache.h"
 #include <PerfStat.h>
+#include <NumericalConstants.h>
 
 class QMyQuickRenderControl : public QQuickRenderControl {
 protected:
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 7be33dcb5845908254950206ce6cff794ff1edde available at: https://gist.github.com/5e64a0ce94482f6c256c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5431.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5431.dmg

PhilipRosedale:
Great that we found this.  Is there a structural problem with QML though that might make it unsuitable for rendering larger amounts of UI because it's CPU cost is too high?  how does rendering/updating something with QML compare to doing it with native controls or HTML + JavaScript, etc?

jherico:
The long term fix is to push UI rendering off of the main thread.  We'd have to do some work and investigation to determine an apples to apples comparison of rendering cost of HTML vs QML rendering.

PR #5437, opened 2015-07-27
DO NOT MERGE - Include and pass version data for all nodes

murillodigital:
All applications now use a cmake macro for including version number in the build. The NodeList class now keeps a version string for each node. DS web ui displays version data for all nodes as well as its own (DS) version.

This merge will add data to the node packets and will break things for out of date nodes, do not merge without proper coordination.


hifi-gustavo:
Build console output for commit: 4ea1fb8700222b1ed15bf442c7663b9673da0958 available at: https://gist.github.com/98d166a5e46df6d3e3e6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5437.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5437.dmg

murillodigital:
@PhilipRosedale @birarda opening up a new PR due to conflicts from network refactor

murillodigital:
@PhilipRosedale @birarda can I get this reviewed/merged? Consider this will force everybody to update to latest due to packet data changes.

hifi-gustavo:
Build console output for commit: bfc73594bb5853cda8cdb85fa8415560bf38247a available at: https://gist.github.com/98d166a5e46df6d3e3e6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5437.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5437.dmg

birarda:
I will take a look today

assignment-client/CMakeLists.txt
```diff
@@ -17,4 +17,6 @@ if (UNIX)
   target_link_libraries(${TARGET_NAME} ${CMAKE_DL_LIBS})
 endif (UNIX)
 
-copy_dlls_beside_windows_executable()
\ No newline at end of file
+include_application_version()
+include_directories("${PROJECT_SOURCE_DIR}/src" "${PROJECT_BINARY_DIR}/includes")
```
birarda: What is this needed to include? Can we include that file specifically instead of this blanket include (with the redundant `PROJECT_SOURCE_DIR` include?

domain-server/CMakeLists.txt
```diff
@@ -31,4 +31,6 @@ include_directories(SYSTEM "${OPENSSL_INCLUDE_DIR}")
 # append OpenSSL to our list of libraries to link
 target_link_libraries(${TARGET_NAME} ${OPENSSL_LIBRARIES})
 
+include_application_version()
+include_directories("${PROJECT_SOURCE_DIR}/src" "${PROJECT_BINARY_DIR}/includes")
```
birarda: same comment as above

assignment-client/src/AssignmentClient.cpp
```diff
@@ -79,7 +79,7 @@ AssignmentClient::AssignmentClient(Assignment::Type requestAssignmentType, QStri
     LogHandler::getInstance().setShouldOutputPID(true);
 
     // setup our _requestAssignment member variable from the passed arguments
-    _requestAssignment = Assignment(Assignment::RequestCommand, requestAssignmentType, assignmentPool);
+    _requestAssignment = Assignment(Assignment::RequestCommand, requestAssignmentType, QCoreApplication::applicationVersion(), assignmentPool);
```
birarda: why do we need to store `QCoreApplication::applicationVersion` with these objects if it is available via that static

assignment-client/src/AssignmentClientMonitor.cpp
```diff
@@ -220,7 +220,7 @@ void AssignmentClientMonitor::handleChildStatusPacket(QSharedPointer<NLPacket> p
             if (!senderID.isNull()) {
                 // We don't have this node yet - we should add it
                 matchingNode = DependencyManager::get<LimitedNodeList>()->addOrUpdateNode
-                    (senderID, NodeType::Unassigned, senderSockAddr, senderSockAddr, false, false);
+                (senderID, NodeType::Unassigned, QCoreApplication::applicationVersion(), senderSockAddr, senderSockAddr, false, false);
```
birarda: this should be indented over

domain-server/resources/web/js/tables.js
```diff
@@ -43,8 +50,10 @@ $(document).ready(function(){
   
   // do the first GET on page load
   getNodesAndAssignments();
+  getDomainServerVersion();
   // grab the new assignments JSON every two seconds
   var getNodesAndAssignmentsInterval = setInterval(getNodesAndAssignments, 2000);
+  var getDomainServerVersionInterval = setInterval(getDomainServerVersion, 2000);
```
birarda: Do we expect this to change while you're on the page? Why do we need to keep reloading it

domain-server/src/DomainServer.cpp
```diff
@@ -567,7 +570,7 @@ void DomainServer::populateDefaultStaticAssignmentsExcludingTypes(const QSet<Ass
             && defaultedType != Assignment::AgentType) {
             // type has not been set from a command line or config file config, use the default
             // by clearing whatever exists and writing a single default assignment with no payload
-            Assignment* newAssignment = new Assignment(Assignment::CreateCommand, (Assignment::Type) defaultedType);
+            Assignment* newAssignment = new Assignment(Assignment::CreateCommand, (Assignment::Type) defaultedType, applicationVersion());
```
birarda: Why would the two `Assignment` above need to know anything about the domain-server application version

domain-server/resources/web/js/tables.js
murillodigital: If you have the page open, and there's a deploy (which we can't say won't happen), you'd have to manually refresh.

domain-server/src/DomainServer.cpp
```diff
@@ -739,11 +743,12 @@ void DomainServer::processConnectRequestPacket(QSharedPointer<NLPacket> packet)
 void DomainServer::processListRequestPacket(QSharedPointer<NLPacket> packet, SharedNodePointer sendingNode) {
     
     NodeType_t throwawayNodeType;
+    QString throwawayNodeVersion;
     HifiSockAddr nodePublicAddress, nodeLocalAddress;
 
     QDataStream packetStream(packet.data());
 
-    parseNodeData(packetStream, throwawayNodeType, nodePublicAddress, nodeLocalAddress, packet->getSenderSockAddr());
+    parseNodeData(packetStream, throwawayNodeType, throwawayNodeVersion, nodePublicAddress, nodeLocalAddress, packet->getSenderSockAddr());
```
birarda: I don't need the node version as general purpose for the entire system at this point in time - I think it would make more sense to store this in DomainServerNodeData

libraries/networking/src/Assignment.h
```diff
@@ -93,6 +94,7 @@ class Assignment : public NodeData {
     QUuid _uuid; /// the 16 byte UUID for this assignment
     Assignment::Command _command; /// the command for this assignment (Create, Deploy, Request)
     Assignment::Type _type; /// the type of the assignment, defines what the assignee will do
+    QString _version;
```
birarda: As mentioned before - not clear why the version needs to be stored and passed with the assignment

libraries/networking/src/LimitedNodeList.cpp
```diff
@@ -423,11 +423,12 @@ SharedNodePointer LimitedNodeList::addOrUpdateNode(const QUuid& uuid, NodeType_t
         matchingNode->setCanAdjustLocks(canAdjustLocks);
         matchingNode->setCanRez(canRez);
         matchingNode->setConnectionSecret(connectionSecret);
+        matchingNode->setVersion(nodeVersion);
```
birarda: Don't think this should be here - should be in DomainServerNodeData

libraries/networking/src/Node.cpp
```diff
@@ -89,6 +91,7 @@ QDataStream& operator<<(QDataStream& out, const Node& node) {
 
 QDataStream& operator>>(QDataStream& in, Node& node) {
     in >> node._type;
+    in >> node._version;
     in >> node._uuid;
     in >> node._publicSocket;
     in >> node._localSocket;
```
birarda: this is causing us to write the node version everytime we put the node on the wire - I don't think we need to do this more than once on application connect

libraries/networking/src/NodeList.cpp
```diff
@@ -273,7 +274,7 @@ void NodeList::sendDomainServerCheckIn() {
         }
 
         // pack our data to send to the domain-server
-        packetStream << _ownerType << _publicSockAddr << _localSockAddr << _nodeTypesOfInterest.toList();
+        packetStream << _ownerType << _version << _publicSockAddr << _localSockAddr << _nodeTypesOfInterest.toList();
```
birarda: I would only send this on connect, not a regular check in

murillodigital:
@birarda yeah using DomainServerNodeData is indeed a better alternative, will change accordingly.

hifi-gustavo:
Build console output for commit: b95eacdb16be2c38ffa76090816bb3f8989e6537 available at: https://gist.github.com/98d166a5e46df6d3e3e6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5437.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5437.dmg

PR #5441, opened 2015-07-28
if OBJ data isn't from a url, don't dereference null QUrl pointers

sethalves:
Fix reading of OBJ data that isn't over the network by avoid dereference of null QUrl pointers.  This fixes a crash in the vhacd tool.

### DIFF
--- a/libraries/fbx/src/OBJReader.h
+++ b/libraries/fbx/src/OBJReader.h
@@ -69,12 +69,13 @@ public:
     QVector<FaceGroup> faceGroups;
     QString currentMaterialName;
     QHash<QString, OBJMaterial> materials;
-    QUrl* url;
-    
+
     QNetworkReply* request(QUrl& url, bool isTest);
     FBXGeometry readOBJ(const QByteArray& model, const QVariantHash& mapping);
     FBXGeometry readOBJ(QIODevice* device, const QVariantHash& mapping, QUrl* url);
 private:
+    QUrl* _url = nullptr;
+
     QHash<QByteArray, bool> librariesSeen;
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: f4a23065b4adbd86ec5a674cf5c65b886bad8a6a available at: https://gist.github.com/47fd8125b6b672cbed4c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5441.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5441.dmg

libraries/fbx/src/OBJReader.h
```diff
@@ -69,12 +69,13 @@ class OBJReader: public QObject { // QObject so we can make network requests.
...
     QString currentMaterialName;
     QHash<QString, OBJMaterial> materials;
-    QUrl* url;
-    
+
     QNetworkReply* request(QUrl& url, bool isTest);
     FBXGeometry readOBJ(const QByteArray& model, const QVariantHash& mapping);
     FBXGeometry readOBJ(QIODevice* device, const QVariantHash& mapping, QUrl* url);
 private:
+    QUrl* _url = nullptr;
```
howard-stearns: FWIW, and not to hold the merge, I don't understand our trend towards private _iVarname (with or without accessor methods) instead of public iVarname. Why is that better?

sethalves: well, the _ is dictated by the coding standard.  In this case, there were some methods which took an argument called "url", and it was easier for me to change the name and have the compiler find all the relevant places.  I made it private to be more sure I wasn't breaking something, some place else.

PR #5444, opened 2015-07-28
Fix issue with skyboxes not being importable from marketplace

huffman:


examples/edit.js
```diff
@@ -1064,7 +1064,7 @@ function importSVO(importURL) {
 
     if (success) {
         var VERY_LARGE = 10000;
-        var position = { x: 0, y: 0, z: 0};
+        var position = { x: 0.01, y: 0.01, z: 0.01};
```
ZappoMan: this seems like a weird fix. We should be able to import to 0,0,0 shouldn't we?

PhilipRosedale:
@ozanserim can you test this please?

PR #5446, opened 2015-07-28
Add cookies.js functionality for collapsable subpanels, key navigation

bwent:
Allows scripts to use the UI library script cookies.js to now create collapsable subpanels containing sliders and checkboxes. Subpanels are collapsable by clicking the icon on the left(TODO: change to a more descriptive icon). TAB key allows for navigation down the panel; SHIFT + TAB allows for navigation up the panel. Highlighted widgets are adjustable using left and right arrow keys for precision.

### DIFF
--- a/examples/utilities/tools/cookies.js
+++ b/examples/utilities/tools/cookies.js
@@ -30,204 +30,230 @@ var CHECK_MARK_COLOR = {
 };
 
 // The Slider class
-(function () {
-    var Slider = function(x,y,width,thumbSize) {
-       
+(function() {
+    var Slider = function(x, y, width, thumbSize) {
+
         this.background = Overlays.addOverlay("text", {
-                        backgroundColor: { red: 200, green: 200, blue: 255 },
-                        x: x,
-                        y: y,
-                        width: width,
-                        height: thumbSize,
-                        alpha: 1.0,
-                        backgroundAlpha: 0.5,
-                        visible: true
-                    });
+            backgroundColor: {
+                red: 200,
+                green: 200,
+                blue: 255
+            },
+            x: x,
+            y: y,
+            width: width,
+            height: thumbSize,
+            alpha: 1.0,
+            backgroundAlpha: 0.5,
+            visible: true
+        });
         this.thumb = Overlays.addOverlay("text", {
-                        backgroundColor: THUMB_COLOR,
-                        x: x,
-                        y: y,
-                        width: thumbSize,
-                        height: thumbSize,
-                        alpha: 1.0,
-                        backgroundAlpha: 1.0,
-                        visible: true
-                    });
-    
+            backgroundColor: THUMB_COLOR,
+            x: x,
+            y: y,
+            width: thumbSize,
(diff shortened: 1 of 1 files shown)


examples/utilities/tools/cookies.js
```diff
@@ -9,72 +9,91 @@
 //  Distributed under the Apache License, Version 2.0.
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
+
+HIFI_PUBLIC_BUCKET = "http://s3.amazonaws.com/hifi-public/";
+
+var SCALE = 1000;
```
samcake: A more precise name would be better for this var, something like "SLIDER_RANGE_INCREMENT"
And since it s used as a division, just put 1/1000 directly so you do a multiply where you use it
Not a show stopper,

examples/utilities/tools/cookies.js
```diff
@@ -289,82 +457,171 @@ ColorBox = function(x,y,width,thumbSize) {
...
+        });
+        this.green.setThumbColor({
+            x: 0,
+            y: color.y,
+            z: 0
+        });
+        this.blue.setThumbColor({
+            x: 0,
+            y: 0,
+            z: color.z
```
samcake: Not a problem, more a comment on the style
is there a technical reason why you choose to go for a new line for every members ?
I feel like it s a lot of space for not much, that s why i prefer the one liner, but maybe it s against the coding guidelines ?

bwent: I used the new line for each member formatting since that was what was used in the original version of the script, but I agree that these can all be on one line...Will change across the whole script

howard-stearns: Also, should I be confused about using x/y/z for color instead of red/green/blue?
(On the C++ side, glm::vec3 uses unions so that vectors can be accessed in a number of ways, but I don't feel we should make use of that in scripts such that we switch around, if indeed that's what is going on here.)

examples/utilities/tools/cookies.js
```diff
@@ -372,21 +629,43 @@ DirectionBox = function(x,y,width,thumbSize) {
...
+    DirectionBox.prototype.onMouseReleaseEvent = function(event) {
         this.yaw.onMouseReleaseEvent(event);
         this.pitch.onMouseReleaseEvent(event);
     };
 
-    this.setterFromWidget = function(value) {
-        var yawPitch = self.getValue();
-        self.onValueChanged(yawPitch);
+    DirectionBox.prototype.updateWithKeys = function(direction) {
+        this.yaw.updateWithKeys(direction);
```
samcake: This is going to be a problem because you cannot have several sliders highlighted and dispatch the key event to both of them or did i miss something ?
I think we are going to merge and we will repair next

bwent: Yeah I wasn't going to worry about DirectionBox and ColorBox widgets for now since cascading events will be a lot more complicated. Will work on fixing this in the next pass

PR #5448, opened 2015-07-28
test script to load test with lots of entities

PhilipRosedale:
Using this script you can make a grid of entities in a domain to test load on entities:  rendering, octree, physics, networking, etc.   Change the TYPE, ROWS, SIZE, etc to control what and how many you are making.

### DIFF
--- a/examples/gridTest.js
+++ b/examples/gridTest.js
@@ -7,17 +7,22 @@
 //
 //  Creates a rectangular grid of objects, starting at the origin and proceeding along the X/Z plane.
 //  Useful for testing the rendering, LOD, and octree storage aspects of the system.  
-//   
+//
+//  Note that when creating things quickly, the entity server will ignore data if we send updates too quickly.
+//  like Internet MTU, these rates are set by th domain operator, so in this script there is a RATE_PER_SECOND 
+//  variable letting you set this speed.  If entities are missing from the grid after a relog, this number 
+//  being too high may be the reason. 
 
 var SIZE = 10.0; 
 var SEPARATION = 20.0;
 var ROWS_X = 30; 
 var ROWS_Z = 30;
-var TYPE = "Sphere";    //   Right now this can be "Box" or "Model" or "Sphere"
+var TYPE = "Sphere";            //   Right now this can be "Box" or "Model" or "Sphere"
 var MODEL_URL = "https://hifi-public.s3.amazonaws.com/models/props/LowPolyIsland/CypressTreeGroup.fbx";
 var MODEL_DIMENSION = { x: 33, y: 16, z: 49 };
-var RATE_PER_SECOND = 1000; 
-var SCRIPT_INTERVAL = 100; 
+var RATE_PER_SECOND = 1000;    //    The entity server will drop data if we create things too fast.
+var SCRIPT_INTERVAL = 100;
+var LIFETIME = 600;            //     By default, these entities will live in the server for 10 minutes 
 
 var addRandom = false; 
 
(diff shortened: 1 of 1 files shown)


examples/gridTest.js
```diff
@@ -0,0 +1,71 @@
...
+//   
+
+var SIZE = 10.0; 
+var SEPARATION = 20.0;
+var ROWS_X = 30; 
+var ROWS_Z = 30;
+var TYPE = "Sphere";    //   Right now this can be "Box" or "Model" or "Sphere"
+var MODEL_URL = "https://hifi-public.s3.amazonaws.com/models/props/LowPolyIsland/CypressTreeGroup.fbx";
+var MODEL_DIMENSION = { x: 33, y: 16, z: 49 };
+var RATE_PER_SECOND = 1000;
```
howard-stearns: Why do we need to rate-limit it?  Should we say why in the script?

examples/gridTest.js
```diff
@@ -0,0 +1,71 @@
...
+var x = 0;
+var z = 0;
+var totalCreated = 0;
+
+Script.setInterval(function () {
+    if (!Entities.serversExist() || !Entities.canRez()) {
+        return;
+    } 
+    
+    var numToCreate = RATE_PER_SECOND * (SCRIPT_INTERVAL / 1000.0);
```
howard-stearns: FWIW, var numToCreate = ROWS_X * ROWS_Z; worked fine for me with the default ROWS_X and ROWS_Z. (Of course, then you can change setInterval to setTimeout.)

examples/gridTest.js
```diff
@@ -0,0 +1,71 @@
+//
+//  Created by Philip Rosedale on July 28, 2015
+//  Copyright 2015 High Fidelity, Inc.
+//  
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+//  Creates a rectangular grid of objects, starting at the origin and proceeding along the X/Z plane.
+//  Useful for testing the rendering, LOD, and octree storage aspects of the system.  
+//
```
howard-stearns: Should we give a warning that this creates the given number of entities and does not remove them?
Alternatively, we could keep the script running after the entities are created, and then delete them when the script is ended.

jherico:
You may want to consider tagging the objects somehow.  I have a similar script here:  https://github.com/highfidelity/hifi/blob/master/examples/cubePerfTest.js which sets all of the entities to a given name, and deletes any previous entities with the same name on execution, preventing objects from piling up on subsequent runs.  

Additionally, we may want to consider a mechanism for marking scripts as single use, so that the application doesn't reload the script on the next run, creating more and more objects.

PhilipRosedale:
@jherico if the viewer can't see all the objects at login (which is typically the case for these), the attempt to delete by name will be unsuccessful.   So, for this script, I don't think any sort of 'cleanup' will be useful - we don't yet have a way to do that domain-wide, and I think it is possible that we actually never may want to, because of the danger of sending such a command to a server.

examples/gridTest.js
PhilipRosedale: Added a comment to that effect.   Rate limiting is needed else the server will drop the requests for a large number of entities.

examples/gridTest.js
PhilipRosedale: Can't find them all to delete.  Instead, I will add a default lifetime that will make them disappear after a while.

howard-stearns:
Maybe not important/worth-it, but we _could_ have a cleanup button in the script that removed all objects tagged by the script, and on startup the script could avoid adding more items if there already exists any such tagged objects.

PR #5449, opened 2015-07-28
Plugins Branch - Fix a crash in pendingChanges.updateItem() for the Particle Entity

ZappoMan:
The basic issue is that the Particle Entity is using the updateItem() interface on Scene and when the scene is being cleaned up (due to changing domains, or quitting) the scene.removeItem() would get processed before the updateItem() and as a result we'd have a dangling pointer to an object that has been deleted.


libraries/entities-renderer/src/RenderableParticleEffectEntityItem.cpp
```diff
@@ -212,8 +211,7 @@ void RenderableParticleEffectEntityItem::updateRenderItem() {
     }
 
     render::PendingChanges pendingChanges;
-    pendingChanges.updateItem<ParticlePayload>(_renderItemId, [&](ParticlePayload& payload) {
-
+    pendingChanges.updateItem<ParticlePayload>(_renderItemId, [this](ParticlePayload& payload) {
```
ZappoMan: Capture "this" since that's all we're using, and we wouldn't want to capture by reference for variables on this stack frame, since the lambda will be called much later when this stack frame is not valid. This wasn't actually causing a crash since the only thing we used in the lambda was the this pointer and other members... but this is a good preventative fix.

libraries/render/src/render/Scene.cpp
```diff
@@ -184,8 +184,8 @@ void Scene::processPendingChangesQueue() {
         // Now we know for sure that we have enough items in the array to
         // capture anything coming from the pendingChanges
         resetItems(consolidatedPendingChanges._resetItems, consolidatedPendingChanges._resetPayloads);
-        removeItems(consolidatedPendingChanges._removedItems);
         updateItems(consolidatedPendingChanges._updatedItems, consolidatedPendingChanges._updateFunctors);
+        removeItems(consolidatedPendingChanges._removedItems);
```
ZappoMan: this is the actual fix for the crash, but doing remove items after updates, we don't attempt to run an update on a previously removed item.

hifi-gustavo:
Build console output for commit: c15161b71afa8aff8e662df55965bf4cd4450e80 available at: https://gist.github.com/0c6f64e2bfb579a1b503<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5449.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5449.dmg

jherico:
Why wasn't this affecting master?

ZappoMan:
I don't know why it wasn't affecting master... and actually, I'm not sure my fix is a 100% guaranteed fix. I might have just seen a similar crash.

PR #5454, opened 2015-07-29
Fix for edit.js Camera.getPosition(), getOrientation() and getRotation()

hyperlogic:
NOTE: this is a pull request into the **plugins** branch, not master.

JavaScript has problems with const refs, apparently.
Camera.getPosition() was returning undefined.

### DIFF
--- a/interface/src/Camera.h
+++ b/interface/src/Camera.h
@@ -53,13 +53,13 @@ public slots:
     QString getModeString() const;
     void setModeString(const QString& mode);
 
-    const glm::quat& getRotation() const { return _rotation; }
+    glm::quat getRotation() const { return _rotation; }
     void setRotation(const glm::quat& rotation);
 
-    const glm::vec3& getPosition() const { return _position; }
+    glm::vec3 getPosition() const { return _position; }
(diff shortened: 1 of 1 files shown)


birarda:
It's probably preferable to keep the return of the const refs for C++ callers though right?

Surprised registering `glm::vec3` wouldn't also have Qt handle `const glm::vec3&` for conversion to/from scripts.

hifi-gustavo:
Build console output for commit: 608898848f9b841e8baa818f0e10c1fcb4ecb6e3 available at: https://gist.github.com/b3270b5641c34fd79ed8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5454.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5454.dmg

ZappoMan:
I think this is a reasonable fix and I'm glad we merged it. I wonder if to Stephen's point there isn't a solution to generally add a conversion for const glm::xxx& to the Qt/JS conversion layer.

PR #5457, opened 2015-07-29
Remove avatar->shift hips for idle animations.

howard-stearns:
This was experimental code for keeping the feet from sliding around while doing idle animations, accomplished by determining how the feet were moving and translating the hip position accordingly.

It had a number of problems, including:
* Not turning off correctly. (When you restart, it acts as as though it is on, even if the checkbox is off. You have to manually toggle on and then off again.)
* Precessing the avatar when there are repeated animations of the legs back and forth (or side to side).

See https://app.asana.com/0/32622044445063/43127903156601


hifi-gustavo:
Build console output for commit: 838711e5261e1eccf73936f5c80242dff35e73e3 available at: https://gist.github.com/62f33ec420b26fa6fb45<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5457.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5457.dmg

PR #5460, opened 2015-07-30
treat a "o" in an obj file as if it were a "g"

sethalves:
Models in Wavefront OBJ files can be divided into mesh-parts with "g" tokens.  The file can be divided into separate objects with "o" tokens.  We don't support multiple objects per OBJ file, so treat an "o" like a "g" rather than ignore it.


hifi-gustavo:
Build console output for commit: b7ecffa0beba9a47969e97f69063ab761a613781 available at: https://gist.github.com/69fa3fde99d8a58862d1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5460.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5460.dmg

PR #5462, opened 2015-07-30
CR for 20574 - Use single file robot model as fallback for full avatar models

ctrlaltdavid:
Fixes having only a skeleton with no head if an invalid full model file
is configured.


hifi-gustavo:
Build console output for commit: f943dc8492ed768977e7cbd96e9700b6f2e82650 available at: https://gist.github.com/b15a1a13e2b826f456c1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5462.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5462.dmg

BingShearer:
I tried to test this on Windows and followed the following recipe:

(1) delete Interface.ini
(2) login
(3) check preferences to see what the default avatar is set to

According to the Preferences sub-window I still end up with a two-part robot avatar.  When I examine the resulting Interface.ini file it is confirmed.

AndrewMeadows:
Oops, I accidentally logged in as Bing.  Trying again:

I tried to test this on Windows and followed the following recipe:

(1) delete Interface.ini
(2) login
(3) check preferences to see what the default avatar is set to

According to the Preferences sub-window I still end up with a two-part robot avatar. When I examine the resulting Interface.ini file it is confirmed.

ctrlaltdavid:
This is expected behaviour. This PR adds a single model robot avatar in addition to the head and skeleton models. ... The point of this PR is to handle the case that a user selects an invalid single model avatar: previously, they ended up with just the default robot skeleton displayed (no head), whereas now they end up with the default head + skeleton from the single model robot avatar displaying. ... And if the user is using separate head and skeleton models, the default robot head or skeleton model is displayed appropriately if either of the user-specified models is invalid.

ctrlaltdavid:
Note: Chris has tested this PR.

PR #5466, opened 2015-07-31
vhacd tool -- make the height of the tetrahedrons in 'fatten' mode less tall.

sethalves:
The tool for creating collision hulls has a "fatten" mode which pulls renderable triangles into collidable tetrahedrons.  Reduce the height of the resulting tetrahedrons.

### DIFF
--- a/tools/vhacd-util/src/VHACDUtil.cpp
+++ b/tools/vhacd-util/src/VHACDUtil.cpp
@@ -125,12 +125,14 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
             continue;
         }
 
+        // from the middle of the triangle, pull a point down to form a tetrahedron.
         float dropAmount = 0;
         dropAmount = glm::max(glm::length(p1 - p0), dropAmount);
         dropAmount = glm::max(glm::length(p2 - p1), dropAmount);
         dropAmount = glm::max(glm::length(p0 - p2), dropAmount);
+        dropAmount *= 0.25f;
 
-        glm::vec3 p3 = av - glm::vec3(0, dropAmount, 0);  // a point 1 meter below the average of this triangle's points
+        glm::vec3 p3 = av - glm::vec3(0, dropAmount, 0);
         int index3 = result.vertices.size();
         result.vertices << p3; // add the new point to the result mesh
 
--- a/tools/vhacd-util/src/VHACDUtil.cpp
+++ b/tools/vhacd-util/src/VHACDUtil.cpp
@@ -12,7 +12,7 @@
 #include <QVector>
 #include "VHACDUtil.h"
 
-const float collisionTetrahedronScale = 0.25f;
+const float COLLISION_TETRAHEDRON_SCALE = 0.25f;
 
 
 // FBXReader jumbles the order of the meshes by reading them back out of a hashtable.  This will put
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: c8f398024eb21f49959ab2fc92d2e3a4f95a1fbb available at: https://gist.github.com/96c40cbf94da6509cc3c<br/><br/>No installer was built for this commit

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -12,6 +12,8 @@
 #include <QVector>
 #include "VHACDUtil.h"
 
+const float collisionTetrahedronScale = 0.25f;
```
howard-stearns: https://docs.highfidelity.com/docs/coding-standard#section-1-1-4-constants-must-be-all-uppercase-using-underscore-to-separate-words-

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -125,12 +127,14 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
...
 
+        // from the middle of the triangle, pull a point down to form a tetrahedron.
         float dropAmount = 0;
         dropAmount = glm::max(glm::length(p1 - p0), dropAmount);
         dropAmount = glm::max(glm::length(p2 - p1), dropAmount);
         dropAmount = glm::max(glm::length(p0 - p2), dropAmount);
+        dropAmount *= collisionTetrahedronScale;
 
-        glm::vec3 p3 = av - glm::vec3(0, dropAmount, 0);  // a point 1 meter below the average of this triangle's points
+        glm::vec3 p3 = av - glm::vec3(0, dropAmount, 0);
```
AndrewMeadows: I know the logic in the line above didn't actually change in this PR, but there are some implicit casts happening here:  int 0 --> float 0.0f.  If you take the time to fix the ALL_CAPS_RULE_FOR_CONSTANTS then you might consider updating this line as well.

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -125,12 +127,14 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
             continue;
         }
 
+        // from the middle of the triangle, pull a point down to form a tetrahedron.
         float dropAmount = 0;
         dropAmount = glm::max(glm::length(p1 - p0), dropAmount);
         dropAmount = glm::max(glm::length(p2 - p1), dropAmount);
         dropAmount = glm::max(glm::length(p0 - p2), dropAmount);
+        dropAmount *= collisionTetrahedronScale;
```
AndrewMeadows: Using the max length of the triangle sides is one way to measure the "size" of the triangle when computing how far to pull it down.  I can think of two other ways that might work better for more extreme triangles:

(1) Compute the average side length, or even the root-mean-square of the sides.

(2) Compute the cross product of any two triangle sides.  The magnitude of the resulting vector is twice the area of the triangle.  Given the area a reasonable dropAmount would be the some fraction of its square root.

hifi-gustavo:
Build console output for commit: 5bae9843f5ee7fe3220d54bbf343490fa4c6ef5c available at: https://gist.github.com/96c40cbf94da6509cc3c<br/><br/>No installer was built for this commit

PR #5465, opened 2015-07-30
Trying to fix/clean up 3DConnexion stuff

SamGondelman:
Mostly style fixes, trying to fix up crash on exit on windows.

This is just a first pass mostly meant to fix up some linking/crashing issues.  There are lots of style changes that still need to be made but that will be later.

PR #5480, opened 2015-08-03
Update crash 3Dconnexion

MarcelEdward:
On windows when exiting interface crashes when the 3Dconnexion device was not enabled in the menu.

Added a check to see if it is enabled in the menu, to be sure to only try to do the removeNativeEventFilter when it is initialized.


hifi-gustavo:
Build console output for commit: 0e1ccb585cfe904d22e0d6fb7917ee5e13f40d1d available at: https://gist.github.com/cc511c445a2ee8da224d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5480.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5480.dmg

ctrlaltdavid:
Tested OK on Windows. ... No SpaceNavigator connected. ... Other builds crash on exit but this one doesn't.

SamGondelman:
Isn't the event filter already removed in ConnexionClient::destroy?  In that case, the destructor doesn't need to do anything.  This PR handles that and a number of other issues: https://github.com/highfidelity/hifi/pull/5465

MarcelEdward:
True the event filter will clear automatic when interface quits, the remove should not be called when not initialized but will be removed when the object is set to null.  
Missed it cause I clicked the stop button in the debug build visual studio, that appears to be different than a normal quit. 

void ConnexionClient::destroy() is called when the enable 3Dconnexion is disabled in the menu. It should clean up the event filter there.

SamGondelman:
ConnexionClient::destroy() is also called from within ~Application(), so the event filter will get removed there.

MarcelEdward:
yes the destructor ConnexionClient::~ConnexionClient() can be removed completely

SamGondelman:
Great, you can close this PR and I'll fix it up in the other one :D

PR #5481, opened 2015-08-03
GL CORE PROFILE:Fixing a few shader and enableing the INput core profile that breaks on mac

samcake:


hifi-gustavo:
Build console output for commit: d3fc05e99fafc21a294e86fe6b76006c4fd8f658 available at: https://gist.github.com/26fd9b926a4c3d2719e3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5481.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5481.dmg

libraries/render-utils/src/simple_textured_emisive.slf
```diff
@@ -19,8 +19,8 @@ uniform sampler2D originalTexture;
```
jherico: I don't see this shader referenced in the shaders-test project, so it must be something I missed. As we find and modify shaders we should make sure that they have coverage in that test to ensure that they at least build.  Could you add this to the tests?

jherico: Never mind other Sam already found the shaders missing in the test and added them in a different PR.

PR #5483, opened 2015-08-03
remove legacy shapes and friends

AndrewMeadows:
We stopped using the legacy Shape (and the related ShapeCollider) classes several months ago... for the most part, but there were a few places where they were still being used.  I'm fixing the last dependencies so that all of this cruft can be removed.


libraries/entities/src/TextEntityItem.cpp
```diff
@@ -128,46 +128,13 @@ void TextEntityItem::appendSubclassData(OctreePacketData* packetData, EncodeBits
...
-        distance = rayInfo._hitDistance;
-    }
-    return intersects;
+    glm::vec3 dimensions = getDimensions();
+    glm::vec2 xyDimensions(dimensions.x, dimensions.y);
+    glm::quat rotation = getRotation();
+    glm::vec3 position = getPosition() + rotation * 
+            (dimensions * (getRegistrationPoint() - ENTITY_ITEM_DEFAULT_REGISTRATION_POINT));
+    return findRayRectangleIntersection(origin, direction, rotation, position, xyDimensions, distance);
 }
```
AndrewMeadows: We were using a PlaneShape to compute the results of they ray-cast against a TextEntityItem.  Since all legacy Shapes are going away I had to re-implement the logic using findRayRectangleIntersection().

libraries/entities/src/WebEntityItem.cpp
```diff
@@ -98,50 +98,17 @@ void WebEntityItem::appendSubclassData(OctreePacketData* packetData, EncodeBitst
...
-        distance = rayInfo._hitDistance;
-    }
-    return intersects;
+    glm::vec3 dimensions = getDimensions();
+    glm::vec2 xyDimensions(dimensions.x, dimensions.y);
+    glm::quat rotation = getRotation();
+    glm::vec3 position = getPosition() + rotation * 
+            (dimensions * (getRegistrationPoint() - ENTITY_ITEM_DEFAULT_REGISTRATION_POINT));
+    return findRayRectangleIntersection(origin, direction, rotation, position, xyDimensions, distance);
 }
```
AndrewMeadows: We were using a PlaneShape to compute the results of they ray-cast against a WebEntityItem.  Since all legacy Shapes are going away I had to re-implement the logic using findRayRectangleIntersection().

libraries/shared/src/GeometryUtil.cpp
```diff
@@ -496,31 +495,45 @@ void PolygonClip::copyCleanArray(int& lengthA, glm::vec2* vertexArrayA, int& len
...
+        if (fabsf(localHitPosition.x) < halfDimensions.x && fabsf(localHitPosition.y) < halfDimensions.y) {
+            // only update distance on intersection
+            distance = d;
+            return true;
+        }
     }
 
-    return intersects;
+    return false;
 }
```
AndrewMeadows: findRayRectangleIntersection() was using RayIntersectionInfo which is one of the legacy cruft classes so it has been rewritten.  Unit tests for findRayRectangleIntersection() have been added.

hifi-gustavo:
Build console output for commit: b00d47d62e4148e0685e8c17af2cae5a2613a117 available at: https://gist.github.com/2733ed991d4718246f83<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5483.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5483.dmg

libraries/shared/src/GeometryUtil.cpp
AndrewMeadows: Another subtle behavior change is that co-planer rays would be considered to be hitting the rectangle no matter where they are on the plane -- even if they did not hit the rectangle in a 2D sense.  For the co-planar case the new code will compute a "closest approach" of the line to the rectangle's center and then consider that point the "hit" if it lies inside the rectangle.

tests/shared/src/GeometryUtilTests.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+        glm::vec3 rayEnd = rectCenter + rectRotation * (-rectDimensions.x * xAxis + (yFraction * rectDimensions.y) * yAxis);
+        glm::vec3 rayHitDirection = glm::normalize(rayEnd - rayStart);
+
+        float distance = FLT_MAX;
+        bool hit = findRayRectangleIntersection(rayStart, rayHitDirection, rectRotation, rectCenter, rectDimensions, distance);
+        QCOMPARE(hit, false);
+        QCOMPARE(distance, FLT_MAX); // distance should be unchanged
+    }
+}
```
AndrewMeadows: Yay!  Unit tests that verify the code actually works.

libraries/shared/src/GeometryUtil.cpp
ZappoMan: note: I believe RayIntersectionInfo is used to send the details of the ray intersection to JavaScript. If so, we need to make sure we still support this if/when we change anything related to RayIntersectionInfo.

PR #5485, opened 2015-08-03
Reduce the speed of movement when in HMD view

ctrlaltdavid:
When using hmdControls.js.


ctrlaltdavid:
@PhilipRosedale As you requested, this reduces the speed of avatar movement when in HMD view and using hmdControls.js.

PR #5486, opened 2015-08-03
GL CORE PROFILE: Update BackendINput to work with 4.1 and remove glColor4f

samcake:
- INtroduce a separate Define for the "GPU_INPUT_PROFILE" and set it all over for GPU_CORE_41 which is what is supported by mac.
 We can think about having a more efficient path for windows and linux

- Replaced the alias "_glCOlor4f" to acutally do a glVertexAttrib4f under the hood (in GLBackend) on the dedicated COLOR slot

THis is rendering meshes almost correctly
apparently skinned objects are wrong


hifi-gustavo:
Build console output for commit: 4b1d2f00fb459b485888fb841dcbd66edf0988cd available at: https://gist.github.com/dca5d41c2bcfdc7d675c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5486.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5486.dmg

jherico:
Looks like the build error might be spurious.  I'm merging anyway and will test locally.

PR #5493, opened 2015-08-04
Fix crash for ModelOverlays, caused by Rig changes

thoys:
ModelOverlay causes a crash in the Model class. This is because ModelOverlay doesn't give a valid rig to the Model.

### DIFF
--- a/interface/src/ui/overlays/ModelOverlay.cpp
+++ b/interface/src/ui/overlays/ModelOverlay.cpp
@@ -10,6 +10,7 @@
 //
 
 #include "ModelOverlay.h"
+#include "EntityRig.h"
 
 #include "Application.h"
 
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 4bf6f8ed7798a056eaa788535c261c0e8449f837 available at: https://gist.github.com/4de0755e475eef6ecce8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5493.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5493.dmg

thoys:
This fixes the lobby.js script. Not sure if we want to be hooked on the EntityRig with the OverlayModel though. But it fixes the crash for now.

howard-stearns:
We should probably make sure that either Rig or a new SimpleRig is functional without being an EntityRig, in case we want to have special functionality for interacting with either animation vs none or entity server vs local.  But we don't have to hold up this fix for that.

Thanks Thoys!

thoys:
You're welcome, thanks for the fast response. Yeh SimpleRig sounds like a good solution.

PR #5496, opened 2015-08-04
CORE PROFILE: Removing the line width completely from gpu::lib

samcake:
Because it s not supporting wide lines correctly and universally anymore on core profile and we would rather have a specialized pipeline for this

So the idea is too remove the lineWidth from the Batch interface

Also turn the state "antialisedLineEnable" true by default

### DIFF
--- a/libraries/gpu/src/gpu/GLBackend.h
+++ b/libraries/gpu/src/gpu/GLBackend.h
@@ -445,7 +445,6 @@ protected:
     void do_glUniformMatrix4fv(Batch& batch, uint32 paramOffset);
 
     void do_glColor4f(Batch& batch, uint32 paramOffset);
-    void do_glLineWidth(Batch& batch, uint32 paramOffset);
 
     typedef void (GLBackend::*CommandCall)(Batch&, uint32);
     static CommandCall _commandCalls[Batch::NUM_COMMANDS];
--- a/libraries/gpu/src/gpu/State.h
+++ b/libraries/gpu/src/gpu/State.h
@@ -260,7 +260,7 @@ public:
             depthClampEnable(false),
             scissorEnable(false),
             multisampleEnable(false),
-            antialisedLineEnable(false),
+            antialisedLineEnable(true),
             alphaToCoverageEnable(false)
         {}
     };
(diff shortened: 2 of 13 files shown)


hifi-gustavo:
Build console output for commit: d07ab353fc69f815011814d6c5a463af2feaeb17 available at: https://gist.github.com/b8d13b04a3f82021f170<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5496.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5496.dmg

libraries/gpu/src/gpu/State.h
```diff
@@ -260,7 +260,7 @@ class State {
             depthClampEnable(false),
             scissorEnable(false),
             multisampleEnable(false),
-            antialisedLineEnable(false),
+            antialisedLineEnable(true),
```
jherico: don't we need a glEnable(GL_LINE_SMOOTH) in the backendstate setup?

libraries/gpu/src/gpu/GLBackend.h
```diff
@@ -445,7 +445,6 @@ class GLBackend : public Backend {
     void do_glUniformMatrix4fv(Batch& batch, uint32 paramOffset);
 
     void do_glColor4f(Batch& batch, uint32 paramOffset);
```
jherico: why is this still here?

jherico:
for places where we've removed the line width of something other than 1.0 we should add a FIXME to restore the current rendering width.

libraries/gpu/src/gpu/State.h
samcake: THis is done as part of the regular state cached values line 587 of GLBackendState.cpp
void GLBackend::do_setStateAntialiasedLineEnable(bool enable) {

hifi-gustavo:
Build console output for commit: 34183b9ddf54541e84eea4fbad6b2cf902a00dc0 available at: https://gist.github.com/b8d13b04a3f82021f170<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5496.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5496.dmg

PR #5500, opened 2015-08-04
Possible fix for infrequent inability to move on Windows startup

SamGondelman:
It seems that _actionsCaptured wasn't being initialized so on Windows it was just garbage and would sometimes be true, leaving you unable to move or rotate the camera.  Starting default scripts often fixed it, likely because one of the default scripts probably sets it to false.

However, there might be a separate issue relating to the Keyboard/Scripted Motor Controls menu options on startup, which could prevent you from moving but would NOT prevent you from rotating the camera.

### DIFF
--- a/interface/src/scripting/ControllerScriptingInterface.cpp
+++ b/interface/src/scripting/ControllerScriptingInterface.cpp
@@ -24,7 +24,8 @@
 ControllerScriptingInterface::ControllerScriptingInterface() :
     _mouseCaptured(false),
     _touchCaptured(false),
-    _wheelCaptured(false)
+    _wheelCaptured(false),
+    _actionsCaptured(false)
 {
 
 }


interface/src/scripting/ControllerScriptingInterface.cpp
```diff
@@ -24,7 +24,8 @@
 ControllerScriptingInterface::ControllerScriptingInterface() :
     _mouseCaptured(false),
     _touchCaptured(false),
-    _wheelCaptured(false)
+    _wheelCaptured(false),
+    _actionsCaptured(false)
```
jherico: So, this is more of a pet peeve than anything else, but I really wish we would stop doing this.  Having a constructor just to initialize a bunch of values is so 2010.  You're better off declaring the defaults at declaration time....

```
class ControllerScriptingInterface {
   bool _actionsCaptured{ false };
   ... etc etc ...
}
```

hifi-gustavo:
Build console output for commit: 52be44d6215693c88b4737cdc9de2067709e246d available at: https://gist.github.com/5455a78aeb34fbd5fee3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5500.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5500.dmg

PR #5508, opened 2015-08-05
Removing integrated glew

jherico:


hifi-gustavo:
Build console output for commit: 04d579be7a0b1678727742cdeb74ad683ee5fc56 available at: https://gist.github.com/dc8f9970f9dd5a9981ee<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5508.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5508.dmg

OmegaHeron:
Prior to this merge cmake on Windows worked - now it throws can't find ZLib.

sethalves:
you may need to install from

  http://gnuwin32.sourceforge.net/packages/zlib.htm

and you may need to fix a header file, as described here:

  http://sourceforge.net/p/gnuwin32/bugs/169/

OmegaHeron:
Fair enough, but, it was compiling without issue before pulling after this commit.  I'll check into the other.

OmegaHeron:
Thanks for the pointer to installer - if anyone else gets stuck you now need to download the installer from ref listed, intsall zlib along with devel files and cmake will pick up.  Prior to today this wasn't necessary so an updatte to windows build guide may be in order.

birarda:
@OmegaHeron this is actually do to the inclusion of Quazip, not the GLEW change.

We will make a change to make sure that zlib is also downloaded for windows users if not present on machine.

OmegaHeron:
Thanks for the heads up @birarda - much appreciated.

PR #5522, opened 2015-08-06
Fix BillboardOverlay breaking when changing url

huffman:
@Atlante45  Can you take a look at this? There was a change in https://github.com/highfidelity/hifi/commit/5cfa1605dfaf27604e78c52eee04172c9e8f21d6 that caused BillboardOverlays to stop working if the url was changed after a texture was already created. This reverts the changes in that commit, so I want to check with you to make sure those changes weren't important. The change to use refresh isn't reverted here.


interface/src/ui/overlays/BillboardOverlay.cpp
```diff
@@ -53,12 +53,12 @@ void BillboardOverlay::update(float deltatime) {
 }
 
 void BillboardOverlay::render(RenderArgs* args) {
-    if (!_texture) {
+    if (!_isLoaded) {
```
ZappoMan: I'm curious about this logic... couple questions:

1) when is _isLoaded set to false?
2) what does _isLoaded mean?

interface/src/ui/overlays/BillboardOverlay.cpp
```diff
@@ -53,12 +53,12 @@ void BillboardOverlay::update(float deltatime) {
...
         _isLoaded = true;
         _texture = DependencyManager::get<TextureCache>()->getTexture(_url);
     }
 
-    if (!_visible || !_texture || !_texture->isLoaded()) {
```
ZappoMan: how do we make sure _texture is not null? Is it ever null?

hifi-gustavo:
Build console output for commit: 8ebf34e93fed02bb5771cb0ec2d5e137eec3a828 available at: https://gist.github.com/c003e3d48e8f00a012a3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5522.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5522.dmg

interface/src/ui/overlays/BillboardOverlay.cpp
huffman: From what I can gather: _isLoaded is set to false in the ctor and when the url changes. It means the texture needs to be (re)loaded.  FWIW, I put the PR in response to an older change - there's a note in the PR description for @Atlante45 - I was hoping he could respond as well with the motivation for the change.

interface/src/ui/overlays/BillboardOverlay.cpp
huffman: I can put it back in - I was reverting the changes in the referenced commit, but this doesn't affect it.

hifi-gustavo:
Build console output for commit: dd2c0be2d292a6a78875d180ae538beefdf21063 available at: https://gist.github.com/c003e3d48e8f00a012a3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5522.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5522.dmg

hifi-gustavo:
Build console output for commit: fec460886b1300f9d85c8f714c61b546ac795781 available at: https://gist.github.com/c003e3d48e8f00a012a3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5522.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5522.dmg

PR #5523, opened 2015-08-06
Worklist 20624 My mixamo avatar crashing interface.exe

MarcelEdward:
Added a check if the fileName() checked is the same as the file in the
cache.
Move the refresh into the if, so the cache only gets refreshed when
needed.

The fst file is cached ok, and loaded almost instantly when in the
cache. However when the fst is loaded in, it causes the maybeRefresh
being called a second time with the fbx file, this should not trigger a
refresh of the fst file.

### DIFF
--- a/libraries/networking/src/ResourceCache.cpp
+++ b/libraries/networking/src/ResourceCache.cpp
@@ -365,9 +365,9 @@ void Resource::maybeRefresh() {
         QVariant variant = reply->header(QNetworkRequest::LastModifiedHeader);
         QNetworkCacheMetaData metaData = NetworkAccessManager::getInstance().cache()->metaData(_url);
 
-        QUrl url(reply->url().toString());
+        QUrl replyUrl(reply->url().toString());
         // check if the reply is the same url as the file from the cache
-        if (_url.fileName() == url.fileName()) {
+        if (_url.fileName() == replyUrl.fileName()) {
             if (variant.isValid() && variant.canConvert<QDateTime>() && metaData.isValid()) {
                 QDateTime lastModified = variant.value<QDateTime>();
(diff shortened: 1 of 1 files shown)


MarcelEdward:
The cache for the avatar fst was already checked for the avatar and found oke. The avatar is loaded and displayed. It moves, so all is oke.

Then it does a second call and tries to check the avatar.fst against the avatar.fbx file. It should never refresh if different files are checked in the maybeRefresh. That sounds like refresh file x with the contents of file y, update the cache of file x with file y and refresh the interface content of file x with file y .... 

When someone updates an fbx model file, then it will get a newer timestamp and it will refresh the cache of that specific file.

hifi-gustavo:
Build console output for commit: 0dff47c004f9e9555843c166e785330d42156b9e available at: https://gist.github.com/65c1eb5ac974d215fbc8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5523.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5523.dmg

hyperlogic:
Hey, we were able to work around the crash in [PR #5524](https://github.com/highfidelity/hifi/pull/5524) by copying the data that was being freed when the Avatar's skeletalModel FBXGeometry was loaded a second time (due to the caching issue in NetworkGeometry).  We have some work in progress to refactor NetworkGeometry to use two separate resource objects to load fst and fbx files separately.  This should provide correct caching behavior, and be easier to maintain.

Thank you for your submission, but I'm closing this PR, I will cc you on the upcoming PR for the NetworkGeometry fix in the coming days.

PR #5526, opened 2015-08-07
Fix the multi part per mesh with various opacity level issue

samcake:
- THe filtering of transparent / opacity parts to create the payload with the proper key should be per PART
not per mesh

- Trying to use the texture bound to the field  "TransparentColor" as diffuse & opacity map together...


hifi-gustavo:
Build console output for commit: e4ce1002ea9e42e8cac90b0cb660c91894f14ff3 available at: https://gist.github.com/ab6e95c844be1a41ee57<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5526.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5526.dmg

PR #5530, opened 2015-08-07
optimizations to RenderableModelEntityItem::render()

ZappoMan:
* **Optimize how we check to see if the entity's modelURL has changed** - We were checking that the model URL hadn't changed but comparing a constructed QUrl to the _model->getURL()... this was slow because constructing a QUrl from a string required url parsing logic. So to fix this we keep a copy of the string for the url in the model and compare against that. **On cell science, this dropped the RenderableModelEntityItem::render() from 9% cpu to 5% cpu.**

* **Optimize ModelEntityItem::getAnimationFrame()** - we now cache the last known frame state so we aren't recalculating it in case the actual frame hasn't change (consider a low animation FPS case where the frame doesn't change every rendered frame). Also using some better functions that access frame animations through references to prevent unneeded copies of vectors. **In cell science this drops getAnimationFrame() from 1.3% CPU to ~0.4% CPU.**

### DIFF
--- a/libraries/animation/src/AnimationCache.cpp
+++ b/libraries/animation/src/AnimationCache.cpp
@@ -91,6 +91,10 @@ QVector<FBXAnimationFrame> Animation::getFrames() const {
     return _geometry.animationFrames;
 }
 
+const QVector<FBXAnimationFrame>& Animation::getFramesReference() const {
+    return _geometry.animationFrames;
+}
+
 void Animation::setGeometry(const FBXGeometry& geometry) {
     _geometry = geometry;
     finishedLoading(true);
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -275,7 +275,7 @@ void RenderableModelEntityItem::render(RenderArgs* args) {
                     }
 
                     if (jointsMapped()) {
-                        QVector<glm::quat> frameData = getAnimationFrame();
+                        auto frameData = getAnimationFrame();
                         for (int i = 0; i < frameData.size(); i++) {
                             _model->setJointState(i, true, frameData[i]);
                         }
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -227,24 +227,30 @@ void RenderableModelEntityItem::render(RenderArgs* args) {
 
     if (hasModel()) {
         if (_model) {
-            if (QUrl(getModelURL()) != _model->getURL()) {
+            if (getModelURL() != _model->getURLAsString()) {
                 qDebug() << "Updating model URL: " << getModelURL();
                 _model->setURL(getModelURL());
             }
 
+            render::ScenePointer scene = AbstractViewStateInterface::instance()->getMain3DScene();
+
             // check to see if when we added our models to the scene they were ready, if they were not ready, then
             // fix them up in the scene
-            render::ScenePointer scene = AbstractViewStateInterface::instance()->getMain3DScene();
-            render::PendingChanges pendingChanges;
             if (_model->needsFixupInScene()) {
+                render::PendingChanges pendingChanges;
+
                 _model->removeFromScene(scene, pendingChanges);
 
                 render::Item::Status::Getters statusGetters;
                 makeEntityItemStatusGetters(this, statusGetters);
                 _model->addToScene(scene, pendingChanges, statusGetters);
-            }
-            scene->enqueuePendingChanges(pendingChanges);
 
+                scene->enqueuePendingChanges(pendingChanges);
+            }
+
+            // FIXME: this seems like it could be optimized if we tracked our last known visible state in
+            //        the renderable item. As it stands now the model checks it's visible/invisible state
+            //        so most of the time we don't do anything in this function.
             _model->setVisibleInScene(getVisible(), scene);
         }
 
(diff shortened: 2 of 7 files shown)


libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -227,24 +227,30 @@ void RenderableModelEntityItem::render(RenderArgs* args) {
 
     if (hasModel()) {
         if (_model) {
-            if (QUrl(getModelURL()) != _model->getURL()) {
+            if (getModelURL() != _model->getURLAsString()) {
```
ZappoMan: this is the main performance fix

hifi-gustavo:
Build console output for commit: f4c3c30f6a70a49323370860754daca2a451f59e available at: https://gist.github.com/7fe133ffd396709e30a7<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5530.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5530.dmg

libraries/animation/src/AnimationCache.cpp
```diff
@@ -91,6 +91,10 @@ QVector<FBXAnimationFrame> Animation::getFrames() const {
     return _geometry.animationFrames;
 }
 
+const QVector<FBXAnimationFrame>& Animation::getFramesReference() const {
```
ZappoMan: allow access to the _geometry.animationFrames by reference to prevent unneeded copying of the QVector<>

libraries/entities/src/ModelEntityItem.h
```diff
@@ -106,7 +106,7 @@ class ModelEntityItem : public EntityItem {
     float getAnimationLastFrame() const { return _animationLoop.getLastFrame(); }
     
     void mapJoints(const QStringList& modelJointNames);
-    QVector<glm::quat> getAnimationFrame();
+    const QVector<glm::quat>& getAnimationFrame();
```
ZappoMan: allow access to the animation frame by reference to prevent unneeded copying of the QVector<>

libraries/entities/src/ModelEntityItem.h
```diff
@@ -123,6 +123,9 @@ class ModelEntityItem : public EntityItem {
     static void cleanupLoadedAnimations();
     
 protected:
+    QVector<glm::quat> _lastKnownFrameData;
```
ZappoMan: used to cache the last know animation frame data

libraries/entities/src/ModelEntityItem.cpp
```diff
@@ -233,18 +236,22 @@ QVector<glm::quat> ModelEntityItem::getAnimationFrame() {
...
             }
 
-            QVector<glm::quat> rotations = frames[animationFrameIndex].rotations;
-
-            frameData.resize(_jointMapping.size());
-            for (int j = 0; j < _jointMapping.size(); j++) {
-                int rotationIndex = _jointMapping[j];
-                if (rotationIndex != -1 && rotationIndex < rotations.size()) {
-                    frameData[j] = rotations[rotationIndex];
+            if (animationFrameIndex != _lastKnownFrameIndex) {
```
ZappoMan: only recalculate the animation frame if the frame index has actually moved forward. Consider the case of a 30fps animation (typical) - the animation index only moves forward every other rendered frame at 60fps.. this will cause us to cache these values so we aren't recalculating the exact same thing.

libraries/entities/src/ModelEntityItem.cpp
```diff
@@ -217,15 +218,17 @@ void ModelEntityItem::mapJoints(const QStringList& modelJointNames) {
...
-        return frameData;
+        return _lastKnownFrameData;
     }
     
-    AnimationPointer myAnimation = getAnimation(_animationURL);
+    AnimationPointer myAnimation = getAnimation(_animationURL); // FIXME: this could be optimized
     if (myAnimation && myAnimation->isLoaded()) {
-        QVector<FBXAnimationFrame> frames = myAnimation->getFrames();
+    
+        const QVector<FBXAnimationFrame>&  frames = myAnimation->getFramesReference(); // NOTE: getFrames() is too heavy
```
ZappoMan: use the reference to the frames... since unlike script users we don't need to convert to JS objects (which requires access to the QVector<>) we can use a const reference and save on copying

libraries/entities/src/ModelEntityItem.cpp
```diff
@@ -233,18 +236,22 @@ QVector<glm::quat> ModelEntityItem::getAnimationFrame() {
...
-
-            frameData.resize(_jointMapping.size());
-            for (int j = 0; j < _jointMapping.size(); j++) {
-                int rotationIndex = _jointMapping[j];
-                if (rotationIndex != -1 && rotationIndex < rotations.size()) {
-                    frameData[j] = rotations[rotationIndex];
+            if (animationFrameIndex != _lastKnownFrameIndex) {
+                _lastKnownFrameIndex = animationFrameIndex;
+                
+                const QVector<glm::quat>& rotations = frames[animationFrameIndex].rotations;
```
ZappoMan: changed this to const reference also to prevent unneeded copying

hifi-gustavo:
Build console output for commit: cec0f8ed1da5e2884da5494510b03a7d02b2d41a available at: https://gist.github.com/7fe133ffd396709e30a7<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5530.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5530.dmg

PR #5531, opened 2015-08-07
Add temporary debug info for tracking down invalid scale value assert

ctrlaltdavid:
In debug builds, extra information will be written to the program log when an assert on an invalid scale value is encountered trying to render a display name.


hifi-gustavo:
Build console output for commit: 3995c4476d498d44607224b8382b29f059488497 available at: https://gist.github.com/0554f4190787f29c5d50<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5531.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5531.dmg

PR #5540, opened 2015-08-10
DO NOT MERGE: Texture improvements (Bump maps)

AlessandroSigna:
FBXReader.cpp - Added conversion from bump map to true normal map


hifi-gustavo:

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1532,6 +1534,41 @@ QByteArray fileOnUrl(const QByteArray& filenameString, const QString& url) {
...
+
+// determine intensity of pixel, from 0 to 1
+const double intensity(const QRgb& qRgb) {
+    const double r = (double)qRed(qRgb);
+    const double g = (double)qGreen(qRgb);
+    const double b = (double)qBlue(qRgb);
+
+    const double average = (r + g + b) / 3.0;
+
+    return average / 255.0;
```
samcake: Coding Standard: You need to use a declared const for the magic number

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1532,6 +1534,41 @@ QByteArray fileOnUrl(const QByteArray& filenameString, const QString& url) {
...
+    }
+}
+
+// determine intensity of pixel, from 0 to 1
+const double intensity(const QRgb& qRgb) {
+    const double r = (double)qRed(qRgb);
+    const double g = (double)qGreen(qRgb);
+    const double b = (double)qBlue(qRgb);
+
+    const double average = (r + g + b) / 3.0;
```
samcake: I don't consider the 3 a magic number but since you are doing this function for every pixel, you should have a const 1.0/3.0 that you can multiply with the sum

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1532,6 +1534,41 @@ QByteArray fileOnUrl(const QByteArray& filenameString, const QString& url) {
...
+
+    return average / 255.0;
+}
+
+// transform -1 - 1 to 0 - 255 (from sobel value to rgb)
+const double mapComponent(double sobelValue) {
+    return (sobelValue + 1.0) * (255.0 / 2.0);
```
samcake: Magic number

ZappoMan:
Ok to test

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1532,6 +1536,43 @@ QByteArray fileOnUrl(const QByteArray& filenameString, const QString& url) {
...
+// determine intensity of pixel, from 0 to 1
+const double intensity(const QRgb& qRgb) {
+    const double factor = 1.0/3.0;
+    const double r = (double)qRed(qRgb);
+    const double g = (double)qGreen(qRgb);
+    const double b = (double)qBlue(qRgb);
+
+    const double average = (r + g + b) * factor;
+
+    return average / RGBA_MAX;
```
samcake: Because your constant is a int then when you divide here you will only get 1 or 0
So you chnage the constant rgba_max to be float

jherico: No.  When combining fixed and floating point values in an arithmetic operation, the fixed point values will be promoted to floating point.  as long as 'average' is a double, RGBA_MAX will automatically be promoted to a double as well.

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1532,6 +1536,31 @@ QByteArray fileOnUrl(const QByteArray& filenameString, const QString& url) {
...
+// returns a value performing bounds checking
+const int clamp(int pixelCoordinate, int maxCoordinate) {
+    if (pixelCoordinate > maxCoordinate) {
+        return maxCoordinate;
+    } else if (pixelCoordinate < 0) {
+        return 0;
+    } else {
+        return pixelCoordinate;
+    }
+}
```
AndrewMeadows: Your clamp has much branching (if/else logic pivots) and branching can suck performance.  Not always a problem, but something to consider when doing an operation inside a double-for loop over an MxN texture.  For more info : https://software.intel.com/en-us/articles/avoiding-the-cost-of-branch-misprediction

There is a branchless way to achieve clamping that looks like more work but is probably faster:

    int clampPixelCoordinate(int coordinate, int maxCoordinate) {
        return coordinate - ((int)(coordinate < 0) * coordinate) + ((int)(coordinate > maxCoordinate) * (maxCoordinate - coordianate));
    }

Meanwhile, when I look at the code where you're using the clamping I note that you're performing the same clamp operation many times.  You could cache the various clamped indices as local variables and then use those in your RGB extraction code.  Note the **i** clamps could be computed outside the **j** loop.

samcake:
Integrated in PR 5681

PR #5541, opened 2015-08-10
DO NOT MERGE: Texture improvements (Opacity maps)

EdgarPironti:
FBXReader.cpp - Added functionalities for the combination of diffuse and opacity maps.

### DIFF
--- a/libraries/fbx/src/FBXReader.cpp
+++ b/libraries/fbx/src/FBXReader.cpp
@@ -1544,27 +1544,28 @@ string getFileExtension(const string& filename) {
 }
 
 // This function checks whether the alpha channel of the image is used or not.
-// (If the alpha channel is not present, the function still works fine: hasAlphaChannel() would be redundant)
-bool checkAlpha(QImage image) {
+bool checkAlpha(const QImage& image) {
+
+    if (!image.hasAlphaChannel()) {
+        return false;
+    }
+
     int width = image.width();
     int height = image.height();
-    bool isUsed = false;
 
-    for (int i = 0; i < width && !isUsed; i++) {
-        for (int j = 0; j < height && !isUsed; j++) {
-            QRgb color = image.pixel(i, j);
-            int alpha = qAlpha(color);
-            if (alpha != RGBA_MAX){
-                isUsed = true;
+    for (int i = 0; i < width; i++) {
+        for (int j = 0; j < height; j++) {
+            if (qAlpha(image.pixel(i, j)) != RGBA_MAX) {
+                return true;
             }
         }
     }
-    return isUsed;
+    return false;
 }
 
 // This function combines the diffuse map and the opacity map.
 // It returns the resulting image with the computed alpha channel.
-QImage combineDiffuseOpacity(QImage imageDiff, QImage imageOpa) {
+void addOpacity(QImage& imageDiff, const QImage& imageOpa) {
 
     if (imageDiff.format() != QImage::Format_ARGB32) {
         imageDiff = imageDiff.convertToFormat(QImage::Format_ARGB32);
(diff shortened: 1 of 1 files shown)


hifi-gustavo:

ZappoMan:
Ok to test

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1532,6 +1533,66 @@ QByteArray fileOnUrl(const QByteArray& filenameString, const QString& url) {
...
+bool checkAlpha(QImage image) {
+    int width = image.width();
+    int height = image.height();
+    bool isUsed = false;
+
+    for (int i = 0; i < width && !isUsed; i++) {
+        for (int j = 0; j < height && !isUsed; j++) {
+            QRgb color = image.pixel(i, j);
+            int alpha = qAlpha(color);
+            if (alpha != 255){
```
samcake: Coding standard: 255 is a Magic number, declare a const Var that you will use instead

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1532,6 +1535,66 @@ QByteArray fileOnUrl(const QByteArray& filenameString, const QString& url) {
...
+    int width = image.width();
+    int height = image.height();
+    bool isUsed = false;
+
+    for (int i = 0; i < width && !isUsed; i++) {
+        for (int j = 0; j < height && !isUsed; j++) {
+            QRgb color = image.pixel(i, j);
+            int alpha = qAlpha(color);
+            if (alpha != RGBA_MAX){
+                isUsed = true;
```
AndrewMeadows: Don't bother to declare **isUsed**.  Instead, return **true** at this point (no need to check further pixels as soon as the first one is found) and return **false** at the bottom of the function.

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1532,6 +1535,66 @@ QByteArray fileOnUrl(const QByteArray& filenameString, const QString& url) {
...
+    int i = filename.rfind('.', filename.length());
+    if (i != string::npos) {
+        return (filename.substr(i + 1, filename.length() - i));
+    }
+    return ("");
+}
+
+// This function checks whether the alpha channel of the image is used or not.
+// (If the alpha channel is not present, the function still works fine: hasAlphaChannel() would be redundant)
+bool checkAlpha(QImage image) {
```
AndrewMeadows: There is a method called **QImage::hasAlphaChannel()**.  You can check that first and return **false** right away it it doesn't have one -- no need to continue and check one pixel at a time.

AndrewMeadows: Yes, the **QImage** class supports **implicit sharing** however there is no loss of generality, and the code is more "clear" to a normal C++ programmer, if you use **const QImage& image** as the argument.

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1532,6 +1535,66 @@ QByteArray fileOnUrl(const QByteArray& filenameString, const QString& url) {
...
+                isUsed = true;
+            }
+        }
+    }
+    return isUsed;
+}
+
+// This function combines the diffuse map and the opacity map.
+// It returns the resulting image with the computed alpha channel.
+QImage combineDiffuseOpacity(QImage imageDiff, QImage imageOpa) {
```
AndrewMeadows: Similarly here the arguments should be passed as **const QImage&** to make things clear.  Meanwhile it is our good fortune that the return value will use **implicit sharing** to avoid a big fat copy.

ericrius1:
quick comment test- will remove soon

ericrius1:
dasdassda

PR #5551, opened 2015-08-11
Quiet compiler

sethalves:
Quiet compiler

### DIFF
--- a/libraries/fbx/src/FBXReader.cpp
+++ b/libraries/fbx/src/FBXReader.cpp
@@ -976,7 +976,7 @@ ExtractedMesh extractMesh(const FBXNode& object, unsigned int& meshIndex) {
     data.extracted.mesh.meshIndex = meshIndex++;
     QVector<int> materials;
     QVector<int> textures;
-    // bool isMaterialPerPolygon = false;
+    bool isMaterialPerPolygon = false;
 
     foreach (const FBXNode& child, object.children) {
         if (child.name == "Vertices") {
@@ -1107,13 +1107,12 @@ ExtractedMesh extractMesh(const FBXNode& object, unsigned int& meshIndex) {
             foreach (const FBXNode& subdata, child.children) {
                 if (subdata.name == "Materials") {
                     materials = getIntVector(subdata);
-                } // else if (subdata.name == "MappingInformationType") {
-                //     if (subdata.properties.at(0) == "ByPolygon")
-                //         isMaterialPerPolygon = true;
-                //     } else {
-                //         isMaterialPerPolygon = false;
-                //     }
-                // }
+                } else if (subdata.name == "MappingInformationType") {
+                    if (subdata.properties.at(0) == "ByPolygon")
+                        isMaterialPerPolygon = true;
+                } else {
+                    isMaterialPerPolygon = false;
+                }
             }
 
 
@@ -1124,6 +1123,11 @@ ExtractedMesh extractMesh(const FBXNode& object, unsigned int& meshIndex) {
                 }
             }
         }
+    }
+
+    bool isMultiMaterial = false;
+    if (isMaterialPerPolygon) {
+        isMultiMaterial = true;
     }
 
     // convert the polygons to quads and triangles
--- a/libraries/fbx/src/FBXReader.cpp
+++ b/libraries/fbx/src/FBXReader.cpp
@@ -976,7 +976,7 @@ ExtractedMesh extractMesh(const FBXNode& object, unsigned int& meshIndex) {
     data.extracted.mesh.meshIndex = meshIndex++;
     QVector<int> materials;
     QVector<int> textures;
-    bool isMaterialPerPolygon = false;
+    // bool isMaterialPerPolygon = false;
 
     foreach (const FBXNode& child, object.children) {
         if (child.name == "Vertices") {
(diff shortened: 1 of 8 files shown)


sethalves:
consider adding ?w=1 to the end of the url while reviewing -- I fixed some line endings.

hifi-gustavo:
Build console output for commit: 804c1ecf8e2523545553b62da2d6587b883e63d0 available at: https://gist.github.com/8e3fb7256593f61017a2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5551.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5551.dmg

hifi-gustavo:
Build console output for commit: fedbfe88815d25b11e68e0963162549a398d38c0 available at: https://gist.github.com/8e3fb7256593f61017a2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5551.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5551.dmg

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1107,13 +1107,13 @@ ExtractedMesh extractMesh(const FBXNode& object, unsigned int& meshIndex) {
...
                 if (subdata.name == "Materials") {
                     materials = getIntVector(subdata);
-                } else if (subdata.name == "MappingInformationType") {
-                    if (subdata.properties.at(0) == "ByPolygon") {   
-                        isMaterialPerPolygon = true;
-                    } else {
-                        isMaterialPerPolygon = false;
-                    }
-                }
+                } // else if (subdata.name == "MappingInformationType") {
```
ZappoMan: ????

sethalves: isMaterialPerPolygon isn't used, so the entire block does nothing.  Should I get rid of it?

libraries/shared/src/PhysicsCollisionGroups.h
```diff
@@ -44,7 +44,7 @@ const int16_t COLLISION_GROUP_OTHER_AVATAR = 1 << 6;
 const int16_t COLLISION_GROUP_MY_ATTACHMENT = 1 << 7;
 const int16_t COLLISION_GROUP_OTHER_ATTACHMENT = 1 << 8;
 // ...
-const int16_t COLLISION_GROUP_COLLISIONLESS = 1 << 15;
+const int16_t COLLISION_GROUP_COLLISIONLESS = 1 << 14;
```
howard-stearns: OK, but why? It's not obvious to me why this change is important/useful. Comment?

AndrewMeadows: The compiler was warning about shifting a 16-bit int all the way into a negative number.  By only shifting 14 bits the warning goes away.  I would have used **uint16_t** when I first wrote this, but Bullet really does use a 16-bit int and I wanted to agree with what they were using.

howard-stearns: Fair enough. (Could maybe add a comment // 14 to avoid changing sign on signed int used by Bullet. But not a blocker.)

libraries/fbx/src/FBXReader.cpp
AndrewMeadows: In fact **isMaterialPerPolygon** IS used later, but used to set **isMultiMaterial** which is itself unused, so it does lead to a dead end.

Keep this logic I think.  Maybe someday we'll want to handle multi-material objects?  Of course, it would be nice to get rid of a warning about unused variable -- maybe set **isMaterialPerPolygon** equal to **isMultiMaterial** inside that check and add a comment about why that is being done (e.g. to remove a warning while still keeping the multi-material scan intact).

libraries/render-utils/src/Model.cpp
```diff
@@ -1807,18 +1807,6 @@ void Model::segregateMeshGroups() {
...
-        bool translucentMesh = networkMesh.getTranslucentPartCount(mesh) == networkMesh.parts.size();
-        bool hasTangents = !mesh.tangents.isEmpty();
-        bool hasSpecular = mesh.hasSpecularTexture();
-        bool hasLightmap = mesh.hasEmissiveTexture();
-        bool isSkinned = state.clusterMatrices.size() > 1;
-        bool wireframe = isWireframe();
-        
-        if (wireframe) {
-            translucentMesh = hasTangents = hasSpecular = hasLightmap = isSkinned = false;
-        }
```
AndrewMeadows: I think @samcake should look this over and decide if it should be removed, or if we are computing some values that are then incorrectly ignored and should actually be used.

libraries/fbx/src/FBXReader.cpp
ZappoMan: I just don't like to see this commented out code here... it seems like this code is supposed to do something...

hifi-gustavo:
Build console output for commit: 1c8030472ae6a06f85ec0d9f71aa61278fd78d1b available at: https://gist.github.com/8e3fb7256593f61017a2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5551.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5551.dmg

PR #5553, opened 2015-08-12
Warning: protocol change - Remove obsolete AvatarMixer wire data

howard-stearns:
Changes avatar mixer data.
Was duplicated in JointStates that are also sent.

### DIFF
--- a/libraries/networking/src/udt/PacketHeaders.h
+++ b/libraries/networking/src/udt/PacketHeaders.h
@@ -142,5 +142,6 @@ const PacketVersion VERSION_ENTITIES_HAVE_SIMULATION_OWNER_AND_ACTIONS_OVER_WIRE
 const PacketVersion VERSION_ENTITIES_NEW_PROTOCOL_LAYER = 35;
 const PacketVersion VERSION_POLYVOX_TEXTURES = 36;
 const PacketVersion VERSION_ENTITIES_POLYLINE = 37;
+const PacketVersion VERSION_AVATAR_NO_LEAN = 38;
 
 #endif // hifi_PacketHeaders_h
--- a/libraries/avatars/src/AvatarData.cpp
+++ b/libraries/avatars/src/AvatarData.cpp
@@ -161,16 +161,6 @@ QByteArray AvatarData::toByteArray() {
     // Body scale
     destinationBuffer += packFloatRatioToTwoByte(destinationBuffer, _targetScale);
 
-    // Head rotation
-    destinationBuffer += packFloatAngleToTwoByte(destinationBuffer, _headData->getFinalPitch());
-    destinationBuffer += packFloatAngleToTwoByte(destinationBuffer, _headData->getFinalYaw());
-    destinationBuffer += packFloatAngleToTwoByte(destinationBuffer, _headData->getFinalRoll());
-
-    // Body lean
-    destinationBuffer += packFloatAngleToTwoByte(destinationBuffer, _headData->_leanForward);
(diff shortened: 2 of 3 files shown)


hifi-gustavo:
Build console output for commit: a5d8125aa784c5e690565ec56a8ba415cbc6cabb available at: https://gist.github.com/c1390c1bc1324abb8037<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5553.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5553.dmg

libraries/networking/src/udt/PacketHeaders.h
```diff
@@ -142,5 +142,6 @@ const PacketVersion VERSION_ENTITIES_HAVE_SIMULATION_OWNER_AND_ACTIONS_OVER_WIRE
 const PacketVersion VERSION_ENTITIES_NEW_PROTOCOL_LAYER = 35;
 const PacketVersion VERSION_POLYVOX_TEXTURES = 36;
 const PacketVersion VERSION_ENTITIES_POLYLINE = 37;
+const PacketVersion VERSION_AVATAR_NO_LEAN = 38;
```
sethalves: This suggests that the entity packets and the avatar packets are related.  Maybe move this away from the end of this list and make it 12 rather than 38?

ZappoMan: Packet versions should be unique for different packet types.

howard-stearns: We were just discussing this. We concluded that neither is strictly required (i.e., they are functionally different namespaces, but it doesn't matter if they get collapsed to one), and that there are weak pros and cons to keeping them separate (run out of numbers faster (rotate?), vs the possibility of getting confused). Since the existing set already used a single namespace with non-entity members, we concluded that it wasn't time to change the practice. Were we wrong? Should we undo the merge? Should we make a second merge to separate this?  (And if so, which items should be in which others namespaces?)

ZappoMan: Im not sure why you're saying the current set had a single namespace for different packet types. Afaik these are all entity or Octree related.

howard-stearns: Yeah, it's not apparent from the code. (I think you're right, but as you say, "afaik".) Presumably VERSION_OCTREE_HAS_FILE_BREAKS, VERSION_ROOT_ELEMENT_HAS_DATA, VERSION_POLYVOX_TEXTURES are for "entity packets", but you'd have to trace their use to know. (Of course, there are no "entity packets", but rather EntityAdd, EntityEdit, and EntityData.) And what namespace is 11 in? (See PacketVersion::versionForPacketType).  Anyway, from the simultaneous conversation in slack, I gather you agree that it doesn't break either way, but that going forward you'd like to see something more systematic?

sethalves: That's not what he meant.  It came down to this -- we could make a different list for avatar packet data, but someone making a change in the code is more likely to get it wrong if there are two lists.  Having them in one big list wastes some number-space, but is safer for the unwary code-changer.

ZappoMan: All of these versions are entity and Octree packet related. Period. 

You have for the first time mixed these versions with avatar data. 

This is a change to how versioning has worked in the code base to date. Period.

PR #5554, opened 2015-08-12
cache calculation of aabox and aacubes

ZappoMan:
Rumor has it, we spend a lot of time in this function. so cache the results and only recalculate when the parameters change.

### DIFF
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1185,74 +1185,84 @@ void EntityItem::setDimensions(const glm::vec3& value) {
         return;
     }
     _transform.setScale(value);
+    requiresRecalcBoxes();
 }
 
 /// The maximum bounding cube for the entity, independent of it's rotation.
 /// This accounts for the registration point (upon which rotation occurs around).
 ///
-AACube EntityItem::getMaximumAACube() const {
-    // * we know that the position is the center of rotation
-    glm::vec3 centerOfRotation = getPosition(); // also where _registration point is
+const AACube& EntityItem::getMaximumAACube() const {
+    if (_recalcMaxAACube) {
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 0990229c64b9c9418e9e1120c21f7907cae7fcaf available at: https://gist.github.com/5f607d90a8bba1ef879b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5554.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5554.dmg

libraries/entities/src/EntityItem.cpp
```diff
@@ -1185,74 +1185,84 @@ void EntityItem::setDimensions(const glm::vec3& value) {
...
 }
 
 /// The maximum bounding cube for the entity, independent of it's rotation.
 /// This accounts for the registration point (upon which rotation occurs around).
 ///
-AACube EntityItem::getMaximumAACube() const {
-    // * we know that the position is the center of rotation
-    glm::vec3 centerOfRotation = getPosition(); // also where _registration point is
+const AACube& EntityItem::getMaximumAACube() const {
+    if (_recalcMaxAACube) {
```
howard-stearns: Not a blocker, and not a big deal here, but as good a time as any to bring it up:
In general, the pattern:
if (!something) {
  return whatever;
}
sometimes has a couple of advantages over:
if (something) {
  ... do all the work...
}
1) As conditionals are changed, it doesn't cause blocks of program text to change only in their indentation, which makes it much easier to see what has actually changed.
2) It tends to reduce deeply nested conditionals, which can be hard to follow. (Which is not the case here.)

Just food for thought.

PR #5557, opened 2015-08-12
Fixing crash in vive controllers

jherico:
The OpenVR software appears to be a little over-eager and is causing me to crash because of inaccurate detection.  This adds some more checking before we consider the controller to be 'supported' and matches the behavior to the Vive display plugin.


hifi-gustavo:
Build console output for commit: 0ff283d09e68afad49cdc10088c823e8fe528545 available at: https://gist.github.com/52d058e688c161d416df<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5557.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5557.dmg

PR #5563, opened 2015-08-13
Fix rotation of image3d overlays

huffman:
This rotation was added with the panel overlay work, which causes inconsistent orientations when compared to the old code. Looking down the -z axis (forward in hifi coordinate system), the image shows correctly without this rotation. With this rotation it is flipped horizontally. I think the former behavior is correct.

### DIFF
--- a/interface/src/ui/overlays/Image3DOverlay.cpp
+++ b/interface/src/ui/overlays/Image3DOverlay.cpp
@@ -90,7 +90,6 @@ void Image3DOverlay::render(RenderArgs* args) {
     applyTransformTo(_transform, true);
     Transform transform = _transform;
     transform.postScale(glm::vec3(getDimensions(), 1.0f));
-    transform.postRotate(glm::angleAxis(glm::pi<float>(), IDENTITY_UP));
 
     batch->setModelTransform(transform);
     batch->setResourceTexture(0, _texture->getGPUTexture());


hifi-gustavo:
Build console output for commit: 0daa9d4871336866d6e5b7dbbf2a9b2ea559a205 available at: https://gist.github.com/63a7b9663922e6200f6c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5563.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5563.dmg

AlexanderOtavka:
I flipped the rotation to be +z so it would be consistent with text3d overlays. If you are going to go -z instead, you should make it consistent across types of planar overlays.

huffman:
Hmm, I think we might be saying the same thing. I tried testing a text3d with default values and it is consistent with image3d _with_ these changes.

AlexanderOtavka:
Really?  In that case, there is nothing to worry about.  However, could you check out `controlPanel.js` and `example/ui/overlayPanel.js` to make sure everything is facing the right way?  I designed them to face +z.  If they are backwards, I can fix them.

huffman:
@AlexanderOtavka I think they are okay, but I'm not sure what images it shows other than the circles. image3d isn't backface culled so for circles (and other symmetric images) the change is not noticeable.

AlexanderOtavka:
Ok, great.  All is well then.

PR #5566, opened 2015-08-13
Get rid of separate head/body avatar

howard-stearns:
The purpose of this is to simplify some code paths, but there are some direct benefits as well:
- The choosing of avatars from preferences is less fidley, with one less dialog.
- An empty url string now consistently produces the default avatar. (It used to produce the mech in some circumstances.)
- The default avatar is taken from resources rather than downloading, for a better new-user experience.
  ==> N.B.: I did not remove the (just under 1MB) of unused old default avatar info from resources, in case that would create build-chain problems.
- While the avatar shows up immediately upon choosing one (as before) the choice is not persisted until you press "Save all changes", and your avatar is reverted to the previous choice if you press "Cancel". (Actually, it's persisted on change, and then undone if cancelled. Sigh...)
- Head/neck gap in some situations is gone/moot. (https://app.asana.com/0/32622044445063/43174289852667)
- Typo fixed in wiring MenuOption::EnableCharacterController to MyAvatar::updateMotionBehaviorFromMenuupdateMotionBehaviorFromMenu. ("No such slot" in logs.)
- Spelling in the source code has been fixed re: "apperance" and "avartar".


howard-stearns:
Testing notes: It should be the case that:
- All old full body avatars set using a previous version should work unchanged.
- All old separate head/body avatars set using a previous version should be silently (except in logs) changed to the default.
- Choosing any full-body avatar from the marketplace should work.
   ==> N.B.: The marketplace page https://metaverse.highfidelity.com/marketplace?category=avatars has NOT been updated with this PR to only show full body avatars.
   However, picking a head-only or body-only does produce a message explaining that these are not supported, and the old avatar is retained.
- Entering a valid .fst url directly in the preferences Appearance box will load and demo that avatar immediately.
- Entering an invalid .fst url produces the default avatar.
- walk.js, facial animation and lean, hydra, and rig animations should work.

hifi-gustavo:
Build console output for commit: 0f6c0a3117cb05e5a1a2fb09045b7156730412c9 available at: https://gist.github.com/a9efb2c21a51ab2e5597<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5566.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5566.dmg

hifi-gustavo:
Build console output for commit: f00d3c037d99584892dcc468cd3b2509ae45205a available at: https://gist.github.com/a9efb2c21a51ab2e5597<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5566.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5566.dmg

interface/src/Menu.cpp
```diff
@@ -258,7 +258,7 @@ Menu::Menu() {
     addCheckableActionToQMenuAndActionHash(avatarMenu, MenuOption::NamesAboveHeads, 0, true);
     addCheckableActionToQMenuAndActionHash(avatarMenu, MenuOption::BlueSpeechSphere, 0, true);
     addCheckableActionToQMenuAndActionHash(avatarMenu, MenuOption::EnableCharacterController, 0, true,
-            avatar, SLOT(updateMotionBehavior()));
+            avatar, SLOT(updateMotionBehaviorFromMenu()));
```
jherico: I'd suggest that whenever we're changing something to do with signals and slots we change from runtime resolution to compile time resolution, so we can't have these invisible bugs in the connections.  However, I don't know if we have a compile time version of `addCheckableActionToQMenuAndActionHash`, so not really a critical issue on this PR.

howard-stearns: I don't know what this particular code is meant to do. I just saw that it wasn't compiling properly and was obviously a typo.

ZappoMan:
Before we merge this we need to make sure the following things have been done:

* Remove all 2 part avatars from the market place.
* Test and verify that if you have a 2 part avatar setting and you start the new interface, that your 2 part avatar is properly replaced with a meaningful 1 part avatar (the default?)
* Verify that our default 1 part avatar is the 1 part avatar we want... (last I checked it was the MechBot... but I think Ozan has a 1 part "droid/robot" we probably want to use instead.
* Make sure a blog post is written about this change.

cozza13:
To @ZappoMan points:

 - Remove all 2 part avatars from the market place. (Done. The marketplace does still show the sections for Head and Body. Yet all content is gone. I will put up an asana to remove those options. 
 - Test and verify that if you have a 2 part avatar setting and you start the new interface, that your 2 part avatar is properly replaced with a meaningful 1 part avatar (the default?). (MINE DID)
Verify that our default 1 part avatar is the 1 part avatar we want... (last I checked it was the MechBot... but I think Ozan has a 1 part "droid/robot" we probably want to use instead. (I TESTED THIS. It is currently the one part robot. I think this is fine and should not be a blocker.)
Make sure a blog post is written about this change. (Done : https://alphas.highfidelity.io/t/updates-domains-and-avatars/7028 )

howard-stearns:
FYI: with this PR, the default avatar is the full body robot, not mechsuit. However, if your last setting before the PR is a two part avatar, then with this PR you will get your previous full body avatar as your avatar (or the default, if you've never had a full body avatar). If the previous code eagerly set mechsuit as a full body avatar (even if you weren't using it, I think), that will be used.

PR #5585, opened 2015-08-17
always persist entities as gzipped json

sethalves:
- remove entity-server setting which allowed a choice of entity-persist-file types.  gzipped json becomes the only (non)choice.

### DIFF
--- a/assignment-client/src/octree/OctreeServer.cpp
+++ b/assignment-client/src/octree/OctreeServer.cpp
@@ -971,12 +971,7 @@ void OctreeServer::readConfiguration() {
         strcpy(_persistFilename, qPrintable(persistFilename));
         qDebug("persistFilename=%s", _persistFilename);
 
-        QString persistAsFileType;
-        if (!readOptionString(QString("persistAsFileType"), settingsSectionObject, persistAsFileType)) {
-            persistAsFileType = "svo";
-        }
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 4a456488a1dede351aaf1c43312f025134ad076a available at: https://gist.github.com/ab1619ebea4a9256b43a<br/><br/>No installer was built for this commit

murillodigital:
test this please

hifi-gustavo:
Build console output for commit: 4a456488a1dede351aaf1c43312f025134ad076a available at: https://gist.github.com/ab1619ebea4a9256b43a<br/><br/>No installer was built for this commit

OmegaHeron:
This will convert any existing content using binary SVO format then re-save as gzipped json once it goes mainstream?  I can see there being a period where it could get tricky if you wanted to restore a saved previous version that's binary SVO and without changing extension names it might get fairly confusing.

sethalves:
The code always reads the newest save, regardless of the file extension.  To restore an old svo, renaming it to models.svo and "touching" it to update the timestamp should be enough.

AndrewMeadows:
Yup, I tested with a touched **models.svo** and it loaded fine, then when I modified it the new saved file was **models.json.gz**.

PR #5587, opened 2015-08-18
Decouple Avatar Update

howard-stearns:
Decouples update/simulation of my avatar and other avatars versus main idle update (and render).
Report "Avatar Simrate" in stats. (This is for my avatar and update of others, but doesn't indicate the rate we're receiving from the mixer.)

This is toggleable with Develop->Avatar->Enable Avatar Update Threading (default is off)


hifi-gustavo:
Build console output for commit: 9ed418f7b612030cf6d2b825f043456319b99af1 available at: https://gist.github.com/1ff0751ca001f1dcd9fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.dmg

interface/src/Application.cpp
```diff
@@ -2415,6 +2421,16 @@ void Application::updateMouseRay() {
     }
 }
 
+void Application::avatarUpdate() {
+    PerformanceTimer perfTimer("myAvatar");
+    qint64 now = usecTimestampNow();
+    float deltaTime = (now - _lastAvatarUpdate) / (1000.0f * 1000.0f);
```
jherico: there's a USECS_PER_SECOND, which you can cast to a float.

howard-stearns: Yup. Thanks.

ZappoMan:
I realize this is a "proof of concept"... and it might make a small difference, but honestly, this isn't moving us forward in where we need to go. This will still be serialized with all the other work happening intra-frame on the main loop. It won't do anything to parallelize the work done by updateAvatar with other tasks in idle() it just moves this work a different location in the main thread between renders.

What specifically were you attempting to test?

howard-stearns:
Just a baseline in which,
  I can dial the frequency up and down, just to see what pops out.
  The next step be in a separate thread, such that that would be the only difference vs the previous version.

ZappoMan:
Ok - but you won't be able to put this in a different thread until you address the contention that comes from Avatars use of Model and rendering's use of Model... You can't just put your new Appliction::avatarUpdate() into a tight loop in a thread, you'll get tons of crashes.

hifi-gustavo:
Build console output for commit: dc75b238a8b6b4323248e3a0d8daced9878b09e0 available at: https://gist.github.com/1ff0751ca001f1dcd9fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.dmg

sethalves:
Settings.setValue("AvatarUpdateType", 3);

interface/src/avatar/AvatarUpdate.cpp
```diff
@@ -0,0 +1,98 @@
...
+    if (_timer) {
+        return;
+    }
+    avatarUpdate();
+}
+
+void AvatarUpdate::avatarUpdate() {
+    PerformanceTimer perfTimer("AvatarUpdate");
+    quint64 now = usecTimestampNow();
+    float deltaTime = (now - _lastAvatarUpdate) / (1000.0f * 1000.0f);
```
sethalves: USECS_PER_SEC

interface/src/avatar/AvatarUpdate.cpp
```diff
@@ -0,0 +1,98 @@
...
+    Application::getInstance()->setAvatarSimrateSample(1.0f / deltaTime);
+    _lastAvatarUpdate = now;
+    
+    //loop through all the other avatars and simulate them...
+    //gets current lookat data, removes missing avatars, etc.
+    DependencyManager::get<AvatarManager>()->updateOtherAvatars(deltaTime);
```
sethalves: should updateOtherAvatars still be mixed in, here?

howard-stearns: Haven't decided yet.
I figure the idea is to remove all avatar updates to this other thread, so that's what it's doing here.
If that's happening twice (e.g., once here and once within in idle => update,) then I've got something wrong.

hifi-gustavo:
Build console output for commit: 2819f3ac1bfae2a2e97327663b7e60a78233b178 available at: https://gist.github.com/1ff0751ca001f1dcd9fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.dmg

interface/src/avatar/AvatarUpdate.cpp
PhilipRosedale: I would think that other avatars would probably make sense here.

hifi-gustavo:
Build console output for commit: e84cc03bcce9e91475c797f7af9ad025415cb845 available at: https://gist.github.com/1ff0751ca001f1dcd9fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.dmg

hifi-gustavo:
Build console output for commit: 828509e8ebe3de421bbcdd584bcc51c1a3cefa5d available at: https://gist.github.com/1ff0751ca001f1dcd9fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.dmg

hifi-gustavo:
Build console output for commit: 0e18395fd05acd528b0f7f318bd92739b1791dba available at: https://gist.github.com/1ff0751ca001f1dcd9fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.dmg

hifi-gustavo:
Build console output for commit: c0fec44f3253b8770725b02bd439a085485120ae available at: https://gist.github.com/1ff0751ca001f1dcd9fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.dmg

hifi-gustavo:
Build console output for commit: 27f4bca0a483316e72fdc3a4915455fc860c8921 available at: https://gist.github.com/1ff0751ca001f1dcd9fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.dmg

hifi-gustavo:
Build console output for commit: ad411d1bae43af3913002231f9ebaa3ec55d0fa8 available at: https://gist.github.com/1ff0751ca001f1dcd9fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.dmg

howard-stearns:
There are four areas to know about with this PR:

* A new AvatarUpdate class manages the thread.
  - It can operate either on its own thread or in the main thread. (It is a sublcass of GenericThread.)
  - It handles one's own avatar (Application::updateMyAvatarLookAtPosition() and AvatarManager::updateMyAvatar()),
    and your down-mixer version of everyone else's avatar (AvatarManager::updateOtherAvatars()).
    The calls were moved from Application.
  - The Application instantiates this singleton, controlled by Developer->Avatar->Enable Avatar Update Threading (default off).
  - The class updates an avatar-simulations/second counter in Application, deplayed in stats. When run by the main update thread, the counter should match Simulations/second (modulo window-averaging differences). When run in its own thread, it should be close to 60fps (controlled by a secret AvatarUpdateTargetSimrate .ini setting), and the other simrate should shoot way up.

* AvatarData has some methods that mark the start and end of regions of code execution: When someone captures avatar data (e.g., the Application figuring out where the camera is); when the avatar is updated; when it is rendered. The methods apply the appropriate locks and maintain data consistency that would otherwise cause judder.

* Some subtle shifts of responsibility, in which three new methods have been separated out from existing inline code:
  - Model::updateClusterMatrices() is now called within Model::render rather than Model::update (simulate).
  - SkeletonModel::updateAttitude() is used by the AvatarData consistency code in addition to being used within simulate() as before.
  - MyAvatar::doUpdateBillboard() is done in the main synchronous update regardless of AvatarUpdate threading, rather than update/simulate.

* Some other issues of wider scope have been worked around, rather than fixing at the source (which would enlarge the scope of this PR):
  - Application::getActiveDisplayPlugin() is not thread safe. (Used by isHMDMode and getHeadPose.) For now AvatarUpdate maintains its own data, updated during the synchronous main-thread update tick.
  - GenericThread can't handle signals, because GenericThread::threadRoutine spins rather than letting run()/exec() do their thing. For now, AvatarUpdate uses sleep rather than a QTimer, and animations are always loaded via the main thread rather than the avatar update thread.
  - Models still use the old school rendering code, rather than Sam's new batch mechanism. Also, the camera position is still updated with code that reaches into the MyAvatar structure, rather than being on some "camera dolly" part maintained by the Rig. With these two holdovers, there ends up being a cap between the camera setup and the rendering, during which sensors and physics can set new values. When the camera is based on one position and my avatar's model on another, we get judder. So for now, the bone animations are based on the latest data, but rendering is done by bashing in the position/orientation captured for the camera calculations, and the setting back to the latest values.

interface/src/Application.h
```diff
@@ -551,6 +558,10 @@ private slots:
 
     KeyboardMouseDevice* _keyboardMouseDevice{ nullptr };   // Default input device, the good old keyboard mouse and maybe touchpad
     MyAvatar* _myAvatar;                         // TODO: move this and relevant code to AvatarManager (or MyAvatar as the case may be)
+    AvatarUpdate* _avatarUpdate = nullptr;
```
ZappoMan: our coding standard is currently silent on the use of c++11 brace initialization, but more and more we're starting to see it creep into our code. The following 4 lines of code are an example of mixing brace initialization with old style assignment. Why are we mixing these? In the case of SimpleMoveAvarage the old style parenthesis initialization is identical to the brace initialization.

We should discuss as a team what we want to do in these cases. The mixing seems wrong, unless we have clear rules for when and why we should use one form or the other.

interface/src/avatar/AvatarUpdate.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "AvatarUpdate.h"
+#include <display-plugins/DisplayPlugin.h>
+#include "InterfaceLogging.h"
+
+// GenericThread accepts an optional "parent" argument, defaulting to nullptr.
+// This is odd, because the moveToThread() in GenericThread::initialize() will
+// become a no-op if the instance ever inits QObject(someNonNullParentQObject).
+// (The only clue will be a log message "QObject::moveToThread: Cannot move objects with a parent",
+//  and things will end up in the same thread that created this instance. Hillarity ensues.)
+// As it turns out, all the other subclasses of GenericThread (at this time) do not init
```
ZappoMan: Why don't you remove this parent param from the GenericThread class... no point in keeping it around since as you point out, it's not used, and would be problematic if it was used.

howard-stearns: Wilco.

interface/src/Application.h
howard-stearns: Agreed. I'll change the ones I added to be consistent.
For now, I'll vote "with my feet" for brace initialization, but we can change that later.

interface/src/avatar/AvatarUpdate.cpp
```diff
@@ -0,0 +1,83 @@
...
+        process();
+    }
+}
+
+bool AvatarUpdate::process() {
+    PerformanceTimer perfTimer("AvatarUpdate");
+    quint64 start = usecTimestampNow();
+    quint64 deltaMicroseconds = start - _lastAvatarUpdate;
+    _lastAvatarUpdate = start;
+    float deltaSeconds = deltaMicroseconds / (float) USECS_PER_SECOND;
```
AndrewMeadows: There is an implicit cast happening here.  You've got to cast both the numerator and the denominator to **float**.

interface/src/avatar/AvatarUpdate.cpp
```diff
@@ -0,0 +1,83 @@
...
+        return true;
+    }
+    int elapsed = (usecTimestampNow() - start);
+    int usecToSleep =  _targetInterval - elapsed;
+    if (usecToSleep < 0) {
+        usecToSleep = 1; // always yield
+    }
+    usleep(usecToSleep);
+    return true;
+}
```
AndrewMeadows: This problem is **NOT** new to this branch, but I noticed it when looking reviewing this project:

AvatarManager::updateMyAvatar() calls MyAvatar::update() calls ModelReferential::update() (the only valid Referential class derivation of Referential is ModelReferential) calls EntityTree::findEntityByID()... which expects the tree to be locked higher up... but it isn't locked by this code path.  We probably haven't noticed this yet because no one uses Referentials.

My conclusion: I think we should rip out the Referential feature and reimplement it later.  But not as part of this project.

libraries/avatars/src/AvatarData.h
```diff
@@ -313,6 +324,8 @@ class AvatarData : public QObject {
 
     bool shouldDie() const { return _owningAvatarMixer.isNull() || getUsecsSinceLastUpdate() > AVATAR_SILENCE_THRESHOLD_USECS; }
 
+    QMutex avatarLock; // Name is redundant, but it aids searches.
```
AndrewMeadows: Why is **avatarLock** public?

interface/src/avatar/AvatarUpdate.cpp
howard-stearns: I agree.

libraries/avatars/src/AvatarData.h
howard-stearns: It was exposed/used in earlier development, before the start/end stuff. I'll change it now.

AndrewMeadows:
I'm a deadlock when I test this code on linux.  The two abbreviated contending callstacks appear to be:

    #0  syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
    #1  0x00007fffefc3242c in QBasicMutex::lockInternal() () from Qt/5.4/gcc_64/lib/libQt5Core.so.5
    #2  0x0000000000fc0e66 in AvatarData::startRenderRun (this=0x1d881d8) at libraries/avatars/src/AvatarData.cpp:139
    #3  0x0000000000be4f35 in Application::displaySide (this=0x7fffffffaa60, renderArgs=0x7fffffff9290, theCamera=..., selfAvatarOnly=true,
        billboard=false) at interface/src/Application.cpp:3615
    #4  0x0000000000be5c36 in Application::renderRearViewMirror (this=0x7fffffffaa60, renderArgs=0x7fffffff9290, region=..., billboard=false)
        at interface/src/Application.cpp:3696
    #5  0x0000000000bd7246 in Application::paintGL (this=0x7fffffffaa60) at interface/src/Application.cpp:1077
    #6  0x0000000000b50877 in GLCanvas::paintGL (this=0x1effad0) at interface/src/GLCanvas.cpp:66


    #0  0x00007fffef94ee1e in std::string::_Rep::_M_clone(std::allocator<char> const&, unsigned long) ()
       from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
    #1  0x00007fffef94ee64 in std::string::reserve(unsigned long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
    #2  0x00007fffef94f0af in std::string::append(char const*, unsigned long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
    #3  0x000000000102b4e6 in std::operator+<char, std::char_traits<char>, std::allocator<char> > (__lhs=..., __rhs=0x133bff8 "OnLoop")
        at /usr/include/c++/4.8/bits/basic_string.h:2406
    #4  0x000000000102aae1 in AnimClip::accumulateTime (this=0x8360218, frame=0, dt=1.4418231e+09, triggersOut=...)
        at libraries/animation/src/AnimClip.cpp:98
    #5  0x000000000102a607 in AnimClip::evaluate (this=0x8360218, animVars=..., dt=1.4418231e+09, triggersOut=...)
        at libraries/animation/src/AnimClip.cpp:38
    #6  0x0000000001022c8f in AnimStateMachine::evaluate (this=0x8360988, animVars=..., dt=1.4418231e+09, triggersOut=...)
        at libraries/animation/src/AnimStateMachine.cpp:65
    #7  0x00000000010065a4 in Rig::updateAnimations (this=0x1d4f238, deltaTime=1.4418231e+09, rootTransform=...)
        at libraries/animation/src/Rig.cpp:578
    #8  0x0000000001099d99 in Model::updateRig (this=0x1d88340, deltaTime=1.4418231e+09, parentTransform=...)
        at libraries/render-utils/src/Model.cpp:1286
    #9  0x0000000000bbc409 in SkeletonModel::updateRig (this=0x1d88340, deltaTime=1.4418231e+09, parentTransform=...)
        at interface/src/avatar/SkeletonModel.cpp:105
    #10 0x0000000001099ea2 in Model::simulateInternal (this=0x1d88340, deltaTime=1.4418231e+09)
        at libraries/render-utils/src/Model.cpp:1293
    #11 0x0000000001099d3f in Model::simulate (this=0x1d88340, deltaTime=1.4418231e+09, fullUpdate=true)
        at libraries/render-utils/src/Model.cpp:1280
    #12 0x0000000000bbcb8d in SkeletonModel::simulate (this=0x1d88340, deltaTime=1.4418231e+09, fullUpdate=true)
        at interface/src/avatar/SkeletonModel.cpp:163
    #13 0x0000000000b8b82d in MyAvatar::simulate (this=0x1d881d8, deltaTime=1.4418231e+09)
        at interface/src/avatar/MyAvatar.cpp:205
    #14 0x0000000000b8b559 in MyAvatar::update (this=0x1d881d8, deltaTime=1.4418231e+09)
        at interface/src/avatar/MyAvatar.cpp:175
    #15 0x0000000000b7b010 in AvatarManager::updateMyAvatar (this=0x1db8470, deltaTime=1.4418231e+09)
        at interface/src/avatar/AvatarManager.cpp:96
    #16 0x0000000000b77b4e in AvatarUpdate::process (this=0xa868140) at interface/src/avatar/AvatarUpdate.cpp:70
    #17 0x0000000000dacab9 in GenericThread::threadRoutine (this=0xa868140)
        at libraries/shared/src/GenericThread.cpp:70

hifi-gustavo:
Build console output for commit: 83d14d33809a0702d50b938a9d1ab8e496eb6291 available at: https://gist.github.com/1ff0751ca001f1dcd9fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.dmg

howard-stearns:
I think I've got the pseudo-deadlock fixed.

hifi-gustavo:
Build console output for commit: ed2ed525b2958b8507071001c0c59efd4eb05c8f available at: https://gist.github.com/1ff0751ca001f1dcd9fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5587.dmg

interface/src/avatar/AvatarUpdate.cpp
ZappoMan: AFAICT - Referentials are required for sitting - I agree they are rarely used and probably broken... but we should be aware of what we lose by "ripping it out"

AndrewMeadows:
Yes, the "deadlock" problem is gone.

libraries/avatars/src/AvatarData.h
```diff
@@ -200,6 +200,17 @@ class AvatarData : public QObject {
...
+    void nextAttitude(glm::vec3 position, glm::quat orientation); // Can be safely called at any time.
+    void startCapture();    // start/end of the period in which the latest values are about to be captured for camera, etc.
+    void endCapture();
+    void startUpdate();     // start/end of update iteration
+    void endUpdate();
+    void startRender();     // start/end of rendering of this object
+    void startRenderRun();  // start/end of entire scene.
+    void endRenderRun();
+    void endRender();
+    virtual void updateAttitude() {} // Tell skeleton mesh about changes
```
AndrewMeadows: As far as I can tell most of these methods don't need to be declared at the **AvatarData** class.  They could be declared at the **Avatar** or even the **MyAvatar** class and the code would probably still work.  This would also allow us to move some of the data members up (down?) the derivation ladder (e.g. _nextPosition and friends).

An exception to this would be **nextAttitude()* which is currently called inside the **DynamicCharacterController** which lives in the **physics** lib.  The fact of the matter is I've been meaning to move **DynamicCharacterController** out of the **physics** lib and into the **interface** codebase but I haven't been able to get to it yet.

howard-stearns: Right- I think all of these could be moved to Avatar IFF DynamicCharacterController::postSimulation() were able to use Avatar instead of AvatarData. 

(I don't want to break up the set into two places (e.g., Avatar and AvatarData) because the implementation of the methods change together. For example, different locking, bashing, or memoizing strategies can be experimented with, but only as a set.)

PR #5589, opened 2015-08-18
Simulations per second readout

sethalves:
- in on-screen stats, show how many times per second Application::idle is called

### DIFF
--- a/interface/src/ui/Stats.cpp
+++ b/interface/src/ui/Stats.cpp
@@ -114,6 +114,7 @@ void Stats::updateStats() {
     STAT_UPDATE(avatarCount, avatarManager->size() - 1);
     STAT_UPDATE(serverCount, nodeList->size());
     STAT_UPDATE(framerate, (int)qApp->getFps());
+    STAT_UPDATE(simrate, (int)Application::getInstance()->getAverageSimsPerSecond());
 
     auto bandwidthRecorder = DependencyManager::get<BandwidthRecorder>();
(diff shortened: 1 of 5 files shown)


hifi-gustavo:
Build console output for commit: e29bb9183b230bce810617c2f75a1aa911072ba1 available at: https://gist.github.com/06eeea3a61d0f2e394a8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5589.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5589.dmg

interface/src/ui/Stats.cpp
```diff
@@ -114,6 +114,7 @@ void Stats::updateStats() {
     STAT_UPDATE(avatarCount, avatarManager->size() - 1);
     STAT_UPDATE(serverCount, nodeList->size());
     STAT_UPDATE(framerate, (int)qApp->getFps());
+    STAT_UPDATE(simrate, (int)Application::getInstance()->getAverageSimsPerSecond());
```
howard-stearns: Why not qApp instead of Application::getInstance()?
I don't know that one is better than the other, but the difference distracts the reader.

ZappoMan: I would suggest we always use Application::getInstance(), and we downplay the direct use of Qt'isms wherever possible.

PR #5590, opened 2015-08-18
Add further temporary debug for tracking down invalid scale value assert

ctrlaltdavid:


hifi-gustavo:
Build console output for commit: 00aa336a2bda14ee9eb8279640f13c6ba53331f2 available at: https://gist.github.com/08a134b86840ecccdb8b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5590.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5590.dmg

interface/src/avatar/Avatar.cpp
```diff
@@ -688,6 +688,23 @@ glm::vec3 Avatar::getDisplayNamePosition() const {
...
+            qDebug() << "getHeadHeight() =" << getHeadHeight();
+        } else {
+            qDebug() << "_position =" << _position;
+            qDebug() << "getBodyUpDirection() =" << getBodyUpDirection();
+            qDebug() << "getBillboardSize() =" << getBillboardSize();
+        }
+    }
+#endif
     return namePosition;
 }
```
AndrewMeadows: As I mentioned in the asana task, the problem is in the line that looks like this:

    // Projected point are between -1.0f and 1.0f, hence 0.5f * windowSizeY
    float pixelHeight = 0.5f * windowSizeY * glm::abs((p1.y / p1.w) - (p0.y / p0.w)); //

In short: **p0.w** is zero.

PR #5594, opened 2015-08-18
CR for Job #20638 - Scripting: WebSocket Implementation

thoys:
Code Review for Job #20638 - Workitem available at https://www.worklist.net/20638


cozza13:
ok to test

hifi-gustavo:
Build console output for commit: 29c857f1fa1d95913e7b3a0fc949c8e7fe5a21e6 available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

cozza13:
ok to test

hifi-gustavo:
Build console output for commit: 1b6538cc59fa154fa9302247cb5770b003a71072 available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

hifi-gustavo:
Build console output for commit: 7f8506acc0c4851a090dc74986d994c79f3415d1 available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

hifi-gustavo:
Build console output for commit: 07ae00f20715b8a5c0ab814c31c50c9b49d5edca available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

hifi-gustavo:
Build console output for commit: ecd368fac35172a5de2c52259366c4934e4bdde5 available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

jherico:
I'm concerned about what impact this will have in terms of security.  Does this mean that someone will be able to, for instance, run a Domain Server with a JS script in an assignment client which will in turn automatically scan my hard drive for financial information and passwords and then send it to some arbitrary server?  Is the Qt script engine sandboxed in such a way to ensure this isn't possible?  If not I would want to make sure it is before we allow scripts to send arbitrary information to arbitrary locations.

huffman:
@jherico I do think we need to spend more time thinking about this. Qt exposes a lot of methods by default, so I have some concerns about exploits making their way through without being explicitly visible to us.

That said, we do already provide a way to communicate to external servers via XMLHttpRequest. I can't think of any specific interface where we provide access to a user's file system - XMLHttpRequest did previously allow requests to `file:` urls, but that was removed for (obvious) security reasons.

thoys:
@jherico @huffman , I don't think this addition adds to the data that Qt might expose to its engine. If the Qt Script Engine can read files from the local disk then they could be exposed through the XMLHttpRequest as @huffman  mentioned, but also through for example the userdata of multiple entities. WebSockets are now supported by most modern browsers. The WebSockets themselves should only be able to connect to other WebSockets. You could for example use your LeapMotion by script using WebSockets: https://developer.leapmotion.com/documentation/javascript/supplements/Leap_JSON.html .

jherico:
As long as we're not actively making things worse, I'm fine with this then.  We may want to create some backlog items to ensure that we're not vulnerable to people grabbing content with malicious scripts.

examples/libraries/unitTest.js
```diff
@@ -101,4 +171,11 @@ UnitTest.prototype.raises = function(func, message) {
     }
     
     throw new UnthrownException(message);
-}
\ No newline at end of file
+}
+
+UnitTest.prototype.ready = function() {
```
huffman: What do you think about naming this something like `finished` or `done`? I think that might be more clear.

libraries/script-engine/src/WebSocketClass.cpp
```diff
@@ -0,0 +1,118 @@
...
+void WebSocketClass::initialize() {
+    connect(_webSocket, &QWebSocket::disconnected, this, &WebSocketClass::handleOnClose);
+    connect(_webSocket, &QWebSocket::textMessageReceived, this, &WebSocketClass::handleOnMessage);
+    connect(_webSocket, &QWebSocket::connected, this, &WebSocketClass::handleOnOpen);
+    _binaryType = "blob";
+}
+
+QScriptValue WebSocketClass::constructor(QScriptContext* context, QScriptEngine* engine) {
+    QString url;
+    QScriptValue callee = context->callee();
```
huffman: this looks unused?

libraries/script-engine/src/WebSocketClass.cpp
```diff
@@ -0,0 +1,118 @@
...
+    if (context->argumentCount() == 1) {
```
huffman: What do you think about making this `> 1`? This would be more in line with how you would expect a javascript function to handle its arguments.

examples/libraries/unitTest.js
thoys: That makes sense :+1:

libraries/script-engine/src/WebSocketClass.h
```diff
@@ -0,0 +1,132 @@
...
+        Q_PROPERTY(ulong bufferedAmount READ getBufferedAmount)
+        Q_PROPERTY(QString extensions READ getExtensions)
+
+        Q_PROPERTY(QScriptValue onclose READ getOnClose WRITE setOnClose)
+        Q_PROPERTY(QScriptValue onerror READ getOnError WRITE setOnError)
+        Q_PROPERTY(QScriptValue onmessage READ getOnMessage WRITE setOnMessage)
+        Q_PROPERTY(QScriptValue onopen READ getOnOpen WRITE setOnOpen)
+
+        Q_PROPERTY(QString protocol READ getProtocol)
+        Q_PROPERTY(uint readyState READ getReadyState)
```
huffman: Can this return `ReadyState`?

libraries/script-engine/src/WebSocketClass.h
```diff
@@ -0,0 +1,132 @@
...
+private:
+    QWebSocket* _webSocket;
+    QScriptEngine* _engine;
+
+    QScriptValue _onCloseEvent;
+    QScriptValue _onErrorEvent;
+    QScriptValue _onMessageEvent;
+    QScriptValue _onOpenEvent;
+
+    QString _binaryType;
```
huffman: What is this for - did you want to use it in send? I couldn't figure out where it was actually used.

libraries/script-engine/src/WebSocketClass.cpp
thoys: > 0 it is :+1:

libraries/script-engine/src/WebSocketClass.h
thoys: binaryType is one of the properties of WebSockets, at the moment its kind of a dummy property, so you could implement existing WebSocket based libraries and they wont miss it.

libraries/script-engine/src/WebSocketClass.h
thoys: you mean the type ReadyState?

libraries/script-engine/src/WebSocketClass.cpp
```diff
@@ -0,0 +1,118 @@
...
+    this->close(closeCode, QStringLiteral(""));
+}
+
+void WebSocketClass::close(QWebSocketProtocol::CloseCode closeCode, QString reason) {
+    _webSocket->close(closeCode, reason);
+}
+
+void WebSocketClass::handleOnClose() {
+    bool hasError = false;
+    if (_webSocket->error() != QAbstractSocket::SocketError::UnknownSocketError) {
```
huffman: I think it would be better to handle errors in a connection to the `QWebSocket::error` signal - this is seems like a reasonable error (UnknownSocketError) that we wouldn't want to suppress.

libraries/script-engine/src/WebSocketClass.h
```diff
@@ -0,0 +1,132 @@
...
+    void setBinaryType(QString binaryType) { _binaryType = binaryType; }
+    QString getBinaryType() { return _binaryType; }
+
+    // extensions is a empty string until supported in QT WebSockets
+    QString getExtensions() { return QString(); }
+
+    // protocol is a empty string until supported in QT WebSockets
+    QString getProtocol() { return QString(); }
+
+    ulong getBufferedAmount() { return 0; }
```
huffman: Is this a stub for future use?

hifi-gustavo:
Build console output for commit: 55fd847636cbc99a438ecaef7cb5b0d4ee69ed0e available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

libraries/script-engine/src/WebSocketClass.h
huffman: I see it's here for symmetry with the WebSocket API - I think this is fine to leave.

libraries/script-engine/src/WebSocketClass.h
thoys: Yes

libraries/script-engine/src/WebSocketClass.cpp
thoys: This doesn't work. The QWebSocket also has a function named error. I cannot connect the error signal.

hifi-gustavo:
Build console output for commit: e581a57320495e3a5fc6b04ab5671428eafda9cc available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

libraries/script-engine/src/WebSocketClass.cpp
```diff
@@ -0,0 +1,118 @@
...
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include "ScriptEngine.h"
+#include "WebSocketClass.h"
+
+WebSocketClass::WebSocketClass(QScriptEngine* engine, QString url) :
+    _engine(engine),
+    _webSocket(new QWebSocket())
```
huffman: This looks like a leak to me right now - we should make `this` the parent of the `QWebSocket`

libraries/script-engine/src/WebSocketClass.cpp
```diff
@@ -0,0 +1,118 @@
...
+{
+    initialize();
+    _webSocket->open(url);
+}
+
+WebSocketClass::WebSocketClass(QScriptEngine* engine, QWebSocket* qWebSocket) :
```
huffman: We'd need to handle cleanup of this web socket as well - instead of dealing with parenting in either of these cases you could delete in the dtor.

libraries/script-engine/src/WebSocketServerClass.cpp
```diff
@@ -0,0 +1,67 @@
...
+WebSocketServerClass::~WebSocketServerClass() {
+    if (_webSocketServer.isListening()) {
+        close();
+    }
+}
+
+void WebSocketServerClass::onNewConnection() {
+    WebSocketClass* newClient = new WebSocketClass(_engine, _webSocketServer.nextPendingConnection());
+    _clients << newClient;
+    emit newConnection(newClient);
```
huffman: I think the case where a client disconnects also needs to be handled, possibly by handling the disconnect or close signal.

libraries/script-engine/src/WebSocketClass.cpp
```diff
@@ -0,0 +1,118 @@
...
+    _binaryType = "blob";
+}
+
+QScriptValue WebSocketClass::constructor(QScriptContext* context, QScriptEngine* engine) {
+    QString url;
+    QScriptValue callee = context->callee();
+    if (context->argumentCount() == 1) {
+        url = context->argument(0).toString();
+    }
+    return engine->newQObject(new WebSocketClass(engine, url));
```
huffman: I would consider making this and the server classes use `QScriptEngine:::ScriptOwnership` so they are automatically cleaned up with the script.

libraries/script-engine/src/WebSocketClass.h
huffman: Yeah

libraries/script-engine/src/WebSocketClass.cpp
huffman: Good catch, yes `> 1` would have been wrong :D

libraries/script-engine/src/WebSocketClass.h
huffman: Oops, I probably made a few other commetns without refreshing and seeing that you responded.

libraries/script-engine/src/WebSocketClass.cpp
huffman: Hmm, what if you use the `SIGNAL` syntax which includes the parameters?

huffman:
This is awesome btw, this will be super useful in general

libraries/script-engine/src/WebSocketServerClass.cpp
thoys: The separate WebSocket objects all have their onclose event. Should WebSocketServer have it too?

libraries/script-engine/src/WebSocketClass.cpp
thoys: I will give it another try

libraries/script-engine/src/WebSocketServerClass.cpp
huffman: Nope, I think you had it right. I thought the server class was holding on to a list of the connections.

hifi-gustavo:
Build console output for commit: c7346209fb40aba7902d76c73e4bc503e47aa69b available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -343,6 +345,12 @@ void ScriptEngine::init() {
     QScriptValue xmlHttpRequestConstructorValue = newFunction(XMLHttpRequestClass::constructor);
     globalObject().setProperty("XMLHttpRequest", xmlHttpRequestConstructorValue);
 
+    QScriptValue webSocketConstructorValue = newFunction(WebSocketClass::constructor);
+    globalObject().setProperty("WebSocket", webSocketConstructorValue);
+
+    QScriptValue webSocketServerConstructorValue = newFunction(WebSocketServerClass::constructor);
+    globalObject().setProperty("WebSocketServer", webSocketServerConstructorValue);
```
huffman: Do you need the server to be available from interface for the work you are doing? I would prefer we put this on the Agent assignment client only for now.

huffman:
I'm done commenting now.

libraries/script-engine/src/ScriptEngine.cpp
thoys: Im fine with that, let me find out if I get the unit tests to run on the AssignmentClient aswell

libraries/script-engine/src/WebSocketServerClass.cpp
huffman: Oops, I guess it does (or it was added later). My concern is sockets being removed from `_clients` after their connections are closed. The list stands to grow indefinitely.

hifi-gustavo:
Build console output for commit: d89ecc1e420cd01b0bf501bbb38d22b6b0686f3c available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

hifi-gustavo:
Build console output for commit: 176d7372adf72fbc3c7d61af52534ee52e9bc297 available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

libraries/script-engine/src/WebSocketClass.cpp
thoys: Leaks should be fixed, by leaving the QScriptEngine handle the deletion of the objects.

hifi-gustavo:
Build console output for commit: 9edba451eb989f68d9190887566dc610a0524027 available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

libraries/script-engine/src/ScriptEngine.cpp
ctrlaltdavid: I'll probably want WebSocket access from Interface script at some stage.

hifi-gustavo:
Build console output for commit: 0e255f4f3ee785b6c35799423e0dad447e2291a1 available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

libraries/script-engine/src/ScriptEngine.cpp
huffman: Out of curiosity, what is the use-case?

ctrlaltdavid: Assuming Interface displays on Magic Leap / HoloLens at some stage, have a native app that implements UI feature for user authorized on the domain.

thoys:
@huffman I believe I have handed all of your comments. Feel free to re-review :+1: . And have fun using the WebSockets aswell :)

hifi-gustavo:
Build console output for commit: d84cf1859c25e9e94d2613da9367f0d53ffd38a8 available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

libraries/script-engine/src/ScriptEngine.cpp
huffman: Ok, when the time comes we can definitely discuss making this available on interface.

huffman:
@thoys Looks good but it looks like the tests are failing because of the change in error handling:

```
[DEBUG] [08/21 17:36:49] script:print()<< "Skipping WebSocketServer tests."
[DEBUG] [08/21 17:36:49] script:print()<< "Running test: Test default WebSocket values"
[DEBUG] [08/21 17:36:50] script:print()<< "  Success: 9 assertions passed"
[DEBUG] [08/21 17:36:50] script:print()<< "Running test: Test WebSocket invalid URL"
[DEBUG] [08/21 17:36:50] script:print()<< "Creating exception"
[DEBUG] [08/21 17:36:50] script:print()<< "  Failure: AssertionException hadError should be true
: false != true"
[DEBUG] [08/21 17:36:50] script:print()<< "Completed all UnitTests"
```

thoys:
let me fix that one

thoys:
Hmm, the problem lies with QWebSocket, usually the error event gets called before the close event gets called in a socket. QWebSocket does it in opposite direction.

hifi-gustavo:
Build console output for commit: 67e945b8847a833f8ad9ac0800984ecafc40ee98 available at: https://gist.github.com/ff121109ca6c3d5a5fa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5594.dmg

huffman:
@thoys Ah, that is odd behavior, I don't see an easy way to fix that.

Great work!

PR #5598, opened 2015-08-19
kill render assert

howard-stearns:
This assert causes dev builds to fail when starting interface on osx.

I'm happy to withdraw this when this is fixed more generally/correctly, but until then, let's have dev builds work without having to keep track of the-changes-you-want vs the-changes-to-make-things-run.

### DIFF
--- a/libraries/render-utils/src/OffscreenQmlSurface.cpp
+++ b/libraries/render-utils/src/OffscreenQmlSurface.cpp
@@ -239,7 +239,7 @@ private:
             return;
         }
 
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 0a8c1846e9539e5320cefde79d8416bbd0cb1124 available at: https://gist.github.com/f71b30bc4a3883240d2d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5598.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5598.dmg

PR #5603, opened 2015-08-20
memory debugging on linux

sethalves:
- on Linux builds, if environment variable HIFI_MEMORY_DEBUGGING is defined when cmake is run, enable gcc's memory debugging guards.


hifi-gustavo:
Build console output for commit: 2056f588e3575e0dae35b66c288f0147fe7bbe39 available at: https://gist.github.com/0a1d67045bf86167d115<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5603.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5603.dmg

jherico:
Looking at this it looks like the memory debugging will only be enabled for the actual executables, not all the libraries we build.

So, two questions.  First, why not make the code here part of setup_hifi_project and setup_hifi_library, or just put it in the top level CMakeLists.txt.  Second, why make it depend on an environment variable instead of a CMake parameter?

sethalves:
Hm... I wonder if you're right about the libraries.  I've been doing this for a while, and it does find problems (in the libraries).  I'll try putting the blob in the toplevel one.  I used the environment variable because I'm worried I'll accidentally enable this on the production servers.  A better "if" suggestion is welcome -- I not overly familiar with cmake.

jherico:
when you run a cmake command you can provide defines, like `cmake .. -DHIFI_MEMORY_DEBUGGING=ON` and then in the CMake file just test for `if (HIFI_MEMORY_DEBUGGING) ...`

jherico:
> I've been doing this for a while, and it does find problems (in the libraries)

Not sure then.  It's certainly possible that template classes and inlined functions might get the benefit even if you're only applying the flag to the executables, giving the impression of memory testing in the libs.  Or I could be wrong about how the cflags are affecting the compilation of the program, but I don't see how a compiler flag in the executable only could affect the library compilation, and this feature seems unlikely to be entirely linker based.

sethalves:
You're at least partially right -- I think I've mostly been getting the benefit of the link-enabled portion of this.  I'm trying to figure out if I can use the compile-enable part on the rest.  Putting the cmake blob in the top-level CMakeLists.txt doesn't seem to do it.  I'll try putting it in all of them.

hifi-gustavo:
Build console output for commit: c25082d86f45c5562750ddd81e6afb77f159b903 available at: https://gist.github.com/0a1d67045bf86167d115<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5603.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5603.dmg

hifi-gustavo:
Build console output for commit: 0ac885da80ab10197608c5c7f902d6259deffc63 available at: https://gist.github.com/0a1d67045bf86167d115<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5603.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5603.dmg

sethalves:
I like being able to set this with an environment variable so I can set it in my .bashrc and not have to remember to enable it.  I think what I have now will allow either the env variable or the cmake -D.

jherico:
I still think putting the `setup_memory_debugging` would be simpler if it was in both the setup_hifi_library and setup_hifi_project macros.  You'd only need to modify two cmake files, but every library or exe has to call one of those two.

PR #5622, opened 2015-08-21
ResourceCache, NetworkGeometry and Model refactoring and optimizations.

hyperlogic:
ResourceCache

* Removed validation logic from Resource class, Qt does this internally and is more
  standards compliant.  This should result in more accurate caching and faster resource
  fetching when cache is stale and validation fails.
* Added loaded and failed slots to Resource class, so it does not have to be polled.

NetworkGeometry

* NetworkGeometry now uses multiple Resource objects to download
  the fst/mapping file and the fbx/obj models.
* NetworkGeometry is no longer a subclass of Resource
* NetworkGeometry now has signals for success and failure, you no longer
  have to poll it to determine when loading is complete (except for textures *sigh*)
* NetworkGeometry no longer has a fallback
* NetworkGeometry no longer loads LODs or has lod logic.

FBXReader

* The number of FBXGeometry copies is greatly reduced.

Model

* Model::setURL no longer supports fallback URL, delayLoad or retainCurrent option.
  This can result in a pop when switching avatars, and there's no longer a default
  if avatar loading fails.

### DIFF
--- a/tests/networking/src/ResourceTests.cpp
+++ b/tests/networking/src/ResourceTests.cpp
@@ -1,0 +1,95 @@
+//
+//  ResoruceTests.cpp
+//
+//  Copyright 2015 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include <QNetworkDiskCache>
+
+#include "ResourceCache.h"
+#include "NetworkAccessManager.h"
+#include "DependencyManager.h"
+
+#include "ResourceTests.h"
+
+QTEST_MAIN(ResourceTests)
+
+void ResourceTests::initTestCase() {
+
+    auto resourceCacheSharedItems = DependencyManager::set<ResourceCacheSharedItems>();
+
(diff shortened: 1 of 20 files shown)


hifi-gustavo:
Build console output for commit: 1b3d7fabc8a4ac6021441d671d249885ee4c741d available at: https://gist.github.com/2718a46cc0b333a09d72<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5622.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5622.dmg

tests/networking/src/ResourceTests.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+#include "ResourceTests.h"
+
+QTEST_MAIN(ResourceTests)
+
+void ResourceTests::initTestCase() {
+
+    auto resourceCacheSharedItems = DependencyManager::set<ResourceCacheSharedItems>();
+
+    const qint64 MAXIMUM_CACHE_SIZE = 1024 * 1024 * 1024; // 1GB
```
PhilipRosedale: We should let users set their own cache size at some point, yes?

PR #5639, opened 2015-08-22
Stereo rendering work

jherico:
This includes a number of changes and fixed to support stereo rendering and plugins in general.

* Don't make `Application` derive from `PluginContainer`.  `PluginContainer` will probably need to be changed frequently in the short term so having to rebuild all of interface every time it changes is a pain.  For now, we instead create a proxy class which is mutually friends with `Application` 
* Fix support for exclusive menu groups when creating menu items from plugins (needed for the fullscreen targets for stereo displays)
* Fix bug that was causing the projection to slowly slide to the right when in a stereo mode
* Add support for passive 3D displays requiring polarized glasses
* Automatically enter fullscreen mode when activating a stereo display
* Add menu items for targeting the various screens in stereo mode so you can place the screen on the correct display

Known bugs
==========

* Switching from one 3D mode to the other (SBS, Interleaved) drops you out of fullscreen mode due to order of operations
* The projection matrix is incorrect in stereo modes due to incorrect calculation of the aspect ratio (the aspect ratio of the offscreen buffer is NOT the aspect ratio of the presented image).  Projection matrix handling should be moved entirely into the display plugins, as different types of plugins will have different requirements and options
* Interleaved display of the overlay leaves it somewhat blurry.  Need to refactor display plugins so that composition of the overlay and the scene occurs in the plugin, rather than in the app.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -58,6 +58,8 @@
 #include <AutoUpdater.h>
 #include <DeferredLightingEffect.h>
 #include <DependencyManager.h>
+#include <plugins/PluginContainer.h>
+#include <plugins/PluginManager.h>
 #include <display-plugins/DisplayPlugin.h>
 
 #include <EntityScriptingInterface.h>
@@ -147,6 +149,8 @@
 #include "ui/UpdateDialog.h"
 
 #include "ui/overlays/Cube3DOverlay.h"
+
+#include "PluginContainerProxy.h"
 
 // ON WIndows PC, NVidia Optimus laptop, we want to enable NVIDIA GPU
 // FIXME seems to be broken.
@@ -301,7 +305,7 @@ bool setupEssentials(int& argc, char** argv) {
 // continuing to overburden Application.cpp
 Cube3DOverlay* _keyboardFocusHighlight{ nullptr };
 int _keyboardFocusHighlightID{ -1 };
-
+PluginContainer* _pluginContainer;
 
 Application::Application(int& argc, char** argv, QElapsedTimer &startup_time) :
         QApplication(argc, argv),
@@ -351,7 +355,9 @@ Application::Application(int& argc, char** argv, QElapsedTimer &startup_time) :
         _applicationOverlay()
 {
     setInstance(this);
-    Plugin::setContainer(this);
+
+    _pluginContainer = new PluginContainerProxy();
+    Plugin::setContainer(_pluginContainer);
 #ifdef Q_OS_WIN
     installNativeEventFilter(&MyNativeEventFilter::getInstance());
 #endif
@@ -1262,6 +1268,7 @@ void Application::resizeGL() {
         // Possible change in aspect ratio
         loadViewFrustum(_myCamera, _viewFrustum);
         float fov = glm::radians(DEFAULT_FIELD_OF_VIEW_DEGREES);
+        // FIXME the aspect ratio for stereo displays is incorrect based on this.
         float aspectRatio = aspect(_renderResolution);
         _myCamera.setProjection(glm::perspective(fov, aspectRatio, DEFAULT_NEAR_CLIP, DEFAULT_FAR_CLIP));
     }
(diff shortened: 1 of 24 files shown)


hifi-gustavo:
Build console output for commit: 2e2ef43da0429a3649f1caf6af58fef2ab50cc79 available at: https://gist.github.com/58894217ac8836543fa0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5639.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5639.dmg

interface/src/Application.cpp
```diff
@@ -302,6 +304,28 @@ bool setupEssentials(int& argc, char** argv) {
 Cube3DOverlay* _keyboardFocusHighlight{ nullptr };
 int _keyboardFocusHighlightID{ -1 };
 
+class PluginContainerProxy : public QObject, PluginContainer {
```
ZappoMan: according to the coding standard, this should be in a different CPP file

libraries/plugins/src/plugins/PluginManager.h
```diff
@@ -1,17 +1,13 @@
...
-#include <QSharedPointer>
-
-class DisplayPlugin;
-class InputPlugin;
-
-using DisplayPluginPointer = QSharedPointer<DisplayPlugin>;
-using DisplayPluginList = QVector<DisplayPluginPointer>;
-using InputPluginPointer = QSharedPointer<InputPlugin>;
-using InputPluginList = QVector<InputPluginPointer>;
+#include "PluginFwd.h"
```
ZappoMan: I'm not sure about this name... Maybe PluginDependecies.h or PluginHelpers.h if you really want forward declares than maybe call it that...

hifi-gustavo:
Build console output for commit: ba0b4bfb3329238f18dfc59eda3d8b2a5e9a4531 available at: https://gist.github.com/58894217ac8836543fa0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5639.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5639.dmg

hifi-gustavo:
Build console output for commit: ba8382dcf1e1d168edeccf8c13d08fe9146bfdf3 available at: https://gist.github.com/58894217ac8836543fa0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5639.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5639.dmg

interface/src/Application.cpp
jherico: fixed

libraries/plugins/src/plugins/PluginManager.h
jherico: changed to `Forward.h` to match gpu forward declaration file name.

hifi-gustavo:
Build console output for commit: b88d946235ac4fb5ec84b9a3390db7212832bed6 available at: https://gist.github.com/58894217ac8836543fa0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5639.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5639.dmg

hifi-gustavo:
Build console output for commit: d3135dcfa6167052e6a34520e0f5e64db7a4ba0f available at: https://gist.github.com/58894217ac8836543fa0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5639.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5639.dmg

ZappoMan:
Something seems broken about v-sync in this build. When I switch between different modes, I got into a state where vsync menu was checked, but my fps was 80+ on my desktop display.

jherico:
Are you sure that's specific to this build?  I didn't change anything related to v-sync that I know of.

ZappoMan:
you are correct Austin. I will report another bug...

PR #5644, opened 2015-08-24
eye correction correction

howard-stearns:
set head rotation of other avatars from joints transmitted from mixer

### DIFF
--- a/interface/src/avatar/SkeletonModel.cpp
+++ b/interface/src/avatar/SkeletonModel.cpp
@@ -41,7 +41,7 @@ SkeletonModel::~SkeletonModel() {
 
 void SkeletonModel::initJointStates(QVector<JointState> states) {
     const FBXGeometry& geometry = _geometry->getFBXGeometry();
-    glm::mat4 parentTransform = glm::scale(_scale) * glm::translate(_offset) * geometry.offset/* * glm::mat4_cast(getRotation())*/;
+    glm::mat4 parentTransform = glm::scale(_scale) * glm::translate(_offset) * geometry.offset;
 
     int rootJointIndex = geometry.rootJointIndex;
     int leftHandJointIndex = geometry.leftHandJointIndex;
@@ -123,10 +123,15 @@ void SkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
         params.rightEyeJointIndex = geometry.rightEyeJointIndex;
 
         _rig->updateFromHeadParameters(params);
-    } else if (true/*_owningAvatar->getHead()->isLookingAtMe()*/) {
+    } else {
+        // This is a little more work than we really want.
+        //
         // Other avatars joint, including their eyes, should already be set just like any other joints
         // from the wire data. But when looking at me, we want the eyes to use the corrected lookAt.
         //
+        // Thus this should really only be ... else if (_owningAvatar->getHead()->isLookingAtMe()) {...
+        // However, in the !isLookingAtMe case, the eyes aren't rotating the way they should right now.
+        // We will revisit that as priorities allow, and particularly after the new rig/animation/joints.
         const FBXGeometry& geometry = _geometry->getFBXGeometry();
         Head* head = _owningAvatar->getHead();
         // If the head is not positioned, updateEyeJoints won't get the math right
--- a/interface/src/avatar/SkeletonModel.cpp
+++ b/interface/src/avatar/SkeletonModel.cpp
@@ -137,15 +137,14 @@ void SkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
         // If the head is not positioned, updateEyeJoints won't get the math right
         glm::quat headOrientation;
         _rig->getJointRotation(geometry.headJointIndex, headOrientation);
-        glm::quat modelOrientation = getRotation();
-        glm::quat adjust = modelOrientation;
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 484db87abcda00d98c63f5354ee0776f21bb4bc4 available at: https://gist.github.com/28ac534b61513453204b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5644.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5644.dmg

howard-stearns:
There's at least two problems here. One is that things were being computed based on a head that was not yet pivoted. This fixes that, and it's a big step forward.

The other problem is, I think, related to confusion about what space things are computed in. (World v avatar v model v head, etc.) In this PR, I've dealt with the above "by hand", without relying on methods that depend on a consistent frame. I'd like to look at that issue on the new branch. (Making such big changes on master would complicate the merge.)

hifi-gustavo:
Build console output for commit: 57e4ff182d98800d5f596160bcbe718d89a971c1 available at: https://gist.github.com/28ac534b61513453204b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5644.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5644.dmg

hifi-gustavo:
Build console output for commit: 476ca364607b877b02a5a241db63cac45aa50be7 available at: https://gist.github.com/28ac534b61513453204b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5644.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5644.dmg

interface/src/avatar/SkeletonModel.cpp
```diff
@@ -131,12 +131,20 @@ void SkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
...
+        glm::quat modelOrientation = getRotation();
+        glm::quat adjust = modelOrientation;
+        glm::quat headOrientationInAvatar = adjust * headOrientation;
+        glm::vec3 eulers = safeEulerAngles(headOrientationInAvatar);
+        head->setBasePitch(glm::degrees(-eulers.x));
+        _rig->updateEyeJoints(geometry.leftEyeJointIndex, geometry.rightEyeJointIndex,
+                              getTranslation(), modelOrientation,
+                              head->getFinalOrientationInWorldFrame(), head->getCorrectedLookAtPosition());
     }
 }
```
AndrewMeadows: The **adjust** variable is unnecessary -- it appears to be a temporary copy of **modelOrientation** and is not used again.

**headOrientationInAvatar** is an odd name.  As far as I can tell it contains the "rotation of the head in the world-frame" but from the name I'd expect "head rotation in the avatar's frame (i.e. model-frame)".

howard-stearns: Indeed, and you've given me an idea...

hifi-gustavo:
Build console output for commit: 2ed6900c7e2f717da5728fb9f6edd1a07bbb7dc4 available at: https://gist.github.com/28ac534b61513453204b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5644.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5644.dmg

interface/src/avatar/SkeletonModel.cpp
```diff
@@ -131,13 +131,20 @@ void SkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
...
+        glm::quat headOrientation;
+        _rig->getJointRotation(geometry.headJointIndex, headOrientation);
+        glm::vec3 eulers = safeEulerAngles(headOrientation);
+        head->setBasePitch(glm::degrees(-eulers.x));
+        head->setBaseYaw(glm::degrees(eulers.y));
+        head->setBaseRoll(glm::degrees(-eulers.z));
+        _rig->updateEyeJoints(geometry.leftEyeJointIndex, geometry.rightEyeJointIndex,
+                              getTranslation(), getRotation(),
+                              head->getFinalOrientationInWorldFrame(), head->getCorrectedLookAtPosition());
+     }
```
AndrewMeadows: This looks more correct when I read the code.  The orienation of the head joint in the model-frame is used to compute the "base" orientation of the head model.  When I look at the implementations of **setBaseFoo()** methods and how they are used in the code I'm convinced that "base" means "in the model frame".  Finally the arguments in Rig::updateEyeJoints() look correct.

That said, this code is logically different from the previous submit.  It needs to be tested and I'm not in a good position to do it since I can't get an HMD running on my machines.

howard-stearns: All correct. FWIW, you can test to some extent by using the 2D camera. 
i.e., the pitch (which was the only part being set before) can be done with shift-w/shift-s, but all three axes can be tested (e.g., for the +/- of the eulers) by rotating/tilting/pitching your physical head in front of the camera.

PR #5646, opened 2015-08-24
de-bloat the domain-server by adding a gatekeeper

birarda:
I was starting to get embarrassed at how large the `DomainServer` class was becoming - this is an initial pass at cleanup to that class by moving the logic that handles connecting nodes to a separate class `DomainGatekeeper`.


hifi-gustavo:
Build console output for commit: d57c1fc2d52740d768f7286edb9711dc0e27e590 available at: https://gist.github.com/4c7b87ca017783909deb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.dmg

hifi-gustavo:
Build console output for commit: fe591f61ccd52c3818f6b8cad809a62020af34d2 available at: https://gist.github.com/4c7b87ca017783909deb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.dmg

hifi-gustavo:
Build console output for commit: c6f1567361dadd561a7cad4ab84f6a12940a1f25 available at: https://gist.github.com/4c7b87ca017783909deb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.dmg

hifi-gustavo:
Build console output for commit: a9c04e5116a7f8b58ceea2789b6041db314c7128 available at: https://gist.github.com/4c7b87ca017783909deb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.dmg

hifi-gustavo:
Build console output for commit: d02e5b2450c6ea7d880dccdd4b3195569f2f6cf1 available at: https://gist.github.com/4c7b87ca017783909deb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.dmg

hifi-gustavo:
Build console output for commit: c8ba24817f37459f1c7c7934184d5d95db11daa2 available at: https://gist.github.com/4c7b87ca017783909deb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.dmg

domain-server/src/DomainGatekeeper.cpp
```diff
@@ -0,0 +1,609 @@
...
+    
+    bool onlyEditorsAreRezzers = false;
+    if (editorsAreRezzersVariant) {
+        onlyEditorsAreRezzers = editorsAreRezzersVariant->toBool();
+    }
+    
+    bool canRez = true;
+    if (onlyEditorsAreRezzers) {
+        canRez = canAdjustLocks;
+    }
```
AndrewMeadows: **canRez** set but never used --> cruft?

domain-server/src/DomainGatekeeper.cpp
```diff
@@ -0,0 +1,609 @@
...
+    
+    // send off the packet unreliably
+    DependencyManager::get<LimitedNodeList>()->sendUnreliablePacket(*connectionTokenPacket, senderSockAddr);
+}
+
+const int NUM_PEER_PINGS_BEFORE_DELETE = 2000 / UDP_PUNCH_PING_INTERVAL_MS;
+
+void DomainGatekeeper::pingPunchForConnectingPeer(const SharedNetworkPeer& peer) {
+    
+    if (peer->getConnectionAttempts() > 0 && peer->getConnectionAttempts() % NUM_PEER_PINGS_BEFORE_DELETE == 0) {
```
AndrewMeadows: The double check against peer->getConnectionAttempts() seems complicated.  Why not use something simpler?

    if (peer->getConnectionAttempts() >= NUM_PEER_PINGS_BEFORE_DELETE) {

domain-server/src/DomainGatekeeper.cpp
birarda: missed the set, thanks - about to add below

domain-server/src/DomainGatekeeper.cpp
```diff
@@ -0,0 +1,609 @@
...
+        // immediately ping the new peer, and start a timer to continue pinging it until we connect to it
+        newPeer->startPingTimer();
+        
+        qDebug() << "Sending ping packets to establish connectivity with ICE peer with ID"
+            << newPeer->getUUID();
+        
+        pingPunchForConnectingPeer(newPeer);
+    } else {
+        delete receivedPeer;
+    }
```
AndrewMeadows: Does not block the PR but it makes me sad: if only there were a way to extract the peers UUID from a iceResponseStream then we wouldn't need to speculatively **new** and subsequently **delete** already existing peers.

hifi-gustavo:
Build console output for commit: 9940f375ea3e368dae626d2d5dad3b1b002113b0 available at: https://gist.github.com/4c7b87ca017783909deb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.dmg

domain-server/src/DomainGatekeeper.cpp
birarda: I'm not sure there's a much better way to do this - even if I did just peek the UUID for the conditional only the implementation of the `NetworkPeer` class knows the size of the data being read, so it would be difficult to skip past the peer you wanted to skip.

AndrewMeadows:
How best to test this?  Login sure, setting edit capabilities in the domain... but what else?  Got some test instructions I should follow?

birarda:
Here are a couple of cases I tested - note that it's hard to test from the same machine as DS since that by default always lets you into your domain. You can either remove that exemption temporarily or test from another machine.

* connect to non-restricted DS (with full auto networking and without)
* connect to restricted DS (with full auto networking and without)
* denied connection for max capacity
* exemption for denied connection if in allowed editors
* allowed editing + rezzing of entities (in a restricted and non-restricted domain)

birarda:
@AndrewMeadows note that I pushed a commit to change behaviour here - originally I made it so that if the DS wasn't verifying users to let them in but was verifying them for the allowed editors list they would be let in even if they failed the verification and just not given edit rights. Now their client is told they are not allowed to connect so that we cover the case where the public key has just been refreshed.

hifi-gustavo:
Build console output for commit: c776f04ab2ce4eabcf74d459a8270cd4985bd398 available at: https://gist.github.com/4c7b87ca017783909deb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.dmg

AndrewMeadows:
I found two bugs.  Please opine on whether these should block the PR or not:

(1) The "exemption for denied connection if in allowed editors" does not appear to work -- I cannot login if I'm only on the "Allowed Editors" list but not on "Allowed Users" list.

(2) (I'm guessing this problem exists in highfidelity/master -- I need to test) When NOT allowed to edit (restricted editing is enabled and the list is not empty) I can still edit an existing entity calling Entities.editEntity().  Not by using edit.js but as follows:

(a) using my circularMotion.js script:
(b) create an orbiting box (stand near /100,100,100 and left-click the mouse once and wait for it to get started)
(c) remove edit rights and restart the domain out from under the avatar
(d) --> when the entity server comes back the orbiting object still moves via script edits

Do we care about (2)?

birarda:
(1) That exemption is specifically to let you in if you're both an allowed user AND and allowed editor but the domain is at its set capacity of users. It does not let you in if you're only in the allowed editors list.

(2) I believe 2 is an ongoing issue. I think we check that specifically in edit.js and then disallow you but we don't have the same checks for manual scripts currently.

birarda:
@AndrewMeadows before we merge this I want to revert the change I made and explained yesterday.
It can cause a bad issue where you can connect to the domain but we're not allowing you because we can't verify your edit rights. In the interim until I fix the public key re-creation process (to not happen on every interface start) if you're in the allowed editors list I will simply allow you to edit without checking the sig.

hifi-gustavo:
Build console output for commit: cd7d0f383b05c5af811ae2d257d12dcef2d0e71a available at: https://gist.github.com/4c7b87ca017783909deb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5646.dmg

PR #5651, opened 2015-08-25
fix hand IK rotation for all models

AndrewMeadows:
It used to be the case that the IK rotation of the hand would work for some models but not all.  This PR fixes their rotation for all models (at least all models that use the correct local orientation of the hand joints).

Also, this fixes the sixense support issues on linux.


cmake/externals/sixense/CMakeLists.txt
```diff
@@ -47,13 +47,13 @@ if (WIN32)
 elseif(APPLE)
 
   # FIXME need to account for different architectures 
-  set(${EXTERNAL_NAME_UPPER}_LIBRARIES ${SOURCE_DIR}/lib/osx32/libopenvr_api.dylib CACHE TYPE INTERNAL)
+  set(${EXTERNAL_NAME_UPPER}_LIBRARIES ${SOURCE_DIR}/lib/osx_x64/release_dll/libsixense_x64.dylib CACHE TYPE INTERNAL)
   add_paths_to_fixup_libs(${SOURCE_DIR}/bin/osx32)
```
AndrewMeadows: In theory Sixense support was broken for Mac since it was trying to include **libopenvr_api.dylib** instead of **libsixense_64.dylib**.  This was likely a copy-n-paste error from the input-plugins project.

Similarly for linux below.

interface/src/avatar/SkeletonModel.cpp
```diff
@@ -241,40 +241,24 @@ void SkeletonModel::applyHandPosition(int jointIndex, const glm::vec3& position)
...
-        _rig->setJointRotationInConstrainedFrame(jointIndex, glm::quat(), PALM_PRIORITY);
-    } else {
-        inverseKinematics(jointIndex, palmPosition, palmRotation, PALM_PRIORITY);
-    }
+
+    // the palm's "raw" rotation is already in the model-frame
+    glm::quat palmRotation = palm.getRawRotation();
+
+    inverseKinematics(jointIndex, palmPosition, palmRotation, PALM_PRIORITY);
 }
```
AndrewMeadows: Rig::inverseKinematics() has been modified to expect the rotation of the joint in the "model-frame" rather than relative to the "bind-frame".

libraries/input-plugins/src/input-plugins/SixenseManager.cpp
```diff
@@ -436,16 +434,66 @@ void SixenseManager::handleButtonEvent(unsigned int buttons, int index) {
...
+
+    // The total rotation of the hand uses the formula:
+    //
+    //     rotation = postOffset * Qsh^ * (measuredRotation * preOffset) * Qsh
+    //
+    rotation = postOffset * glm::inverse(sixenseToHand) * rotation * preOffset * sixenseToHand;
+
     _poseStateMap[makeInput(JointChannel(index)).getChannel()] = UserInputMapper::PoseValue(position, rotation);
 #endif // HAVE_SIXENSE
 }
```
AndrewMeadows: After much blood sweat and tears I finally figured out how to really get the sixense data into the right frame.  I put documentation in comments so that the next developer may have an easier time of it.

AndrewMeadows:
Oops, I forgot that this work probably breaks the Vive support for hand rotations since the IK system now expects their rotation in the model-frame rather than relative to the bind-frame.  I'll have to test and fix.

hifi-gustavo:
Build console output for commit: aea6d3b87354a0acee9a9d835a53b953e718250d available at: https://gist.github.com/78a2443d6cc9e2eecfc4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5651.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5651.dmg

hifi-gustavo:
Build console output for commit: cdba5428099496147197e7bcead0de9fdac72ffa available at: https://gist.github.com/78a2443d6cc9e2eecfc4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5651.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5651.dmg

AndrewMeadows:
Ok this should be ready to go for Vive.

hifi-gustavo:
Build console output for commit: c3b6e69731d660e610d7c2bf549c13b8548a7300 available at: https://gist.github.com/78a2443d6cc9e2eecfc4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5651.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5651.dmg

libraries/animation/src/Rig.cpp
```diff
@@ -666,11 +681,10 @@ void Rig::inverseKinematics(int endIndex, glm::vec3 targetPosition, const glm::q
...
             }
 
-            // Apply the rotation, but use mixRotationDelta() which blends a bit of the default pose
-            // in the process.  This provides stability to the IK solution for most models.
+            // Apply the rotation delta.
             glm::quat oldNextRotation = nextState.getRotation();
-            float mixFactor = 0.03f;
-            nextState.mixRotationDelta(deltaRotation, mixFactor, priority);
+            float mixFactor = 0.05f;
+            nextState.applyRotationDelta(deltaRotation, mixFactor, priority);
```
AndrewMeadows: The mix of the "default pose" which was inside the **mixRotationDelta()** method has been moved out and is done explicitly above, before this **for** loop --> we only "relax" once before the IK iterations rather than doing it inside each iteration.

libraries/animation/src/JointState.cpp
```diff
@@ -153,6 +153,24 @@ void JointState::setRotationInBindFrame(const glm::quat& rotation, float priorit
...
+        // r' = (Rp * Rpre)^ * R' * Rpost^
+        glm::quat targetRotation = glm::inverse(parentRotation * _preRotation) * rotationInModelFrame * glm::inverse(_postRotation);
+        if (constrain && _constraint) {
+            _constraint->softClamp(targetRotation, _rotationInConstrainedFrame, 0.5f);
+        }
+        _rotationInConstrainedFrame = glm::normalize(targetRotation);
+        _transformChanged = true;
+        _animationPriority = priority;
+    }
+}
```
AndrewMeadows: Setting rotation in the model-frame is the important change in this PR (used to be specified relative to the bind-frame).  The future IK stuff will expect to be given positions and rotations in the model-frame, or something very similar.

libraries/animation/src/Rig.cpp
```diff
@@ -188,7 +188,7 @@ void Rig::deleteAnimations() {
     _animationHandles.clear();
 }
 
-float Rig::initJointStates(QVector<JointState> states, glm::mat4 parentTransform,
+float Rig::initJointStates(QVector<JointState> states, glm::mat4 rootTransform,
```
AndrewMeadows: I'm changing the name from **parentFrame** to **rootFrame** so that the variable name is more correct.

libraries/animation/src/Rig.cpp
```diff
@@ -606,12 +606,27 @@ void Rig::inverseKinematics(int endIndex, glm::vec3 targetPosition, const glm::q
...
+        JointState& nextState = _jointStates[nextIndex];
+        if (! nextState.getIsFree()) {
+            continue;
+        }
+
+        // Apply the zero rotationDelta, but use mixRotationDelta() which blends a bit of the default pose
+        // in the process.  This provides stability to the IK solution for most models.
+        float mixFactor = 0.08f;
+        nextState.mixRotationDelta(glm::quat(), mixFactor, priority);
+    }
```
AndrewMeadows: This is the relaxation toward the default animation pose that has been explicitly pulled out of the **mixRotationDelta()** that was used further down.

libraries/animation/src/Rig.cpp
```diff
@@ -689,10 +703,10 @@ void Rig::inverseKinematics(int endIndex, glm::vec3 targetPosition, const glm::q
         // measure our success
         endPosition = endState.getPosition();
         distanceToGo = glm::distance(targetPosition, endPosition);
-    } while (numIterations < MAX_ITERATION_COUNT && distanceToGo < ACCEPTABLE_IK_ERROR);
+    } while (numIterations < MAX_ITERATION_COUNT && distanceToGo > ACCEPTABLE_IK_ERROR);
```
AndrewMeadows: The inequality sign here was a bug: inverted from what it should have been it would cause the IK system to only iterate at most once every frame.

hifi-gustavo:
Build console output for commit: 47c18fa18a1906f988ed25ce708924e28fc71375 available at: https://gist.github.com/78a2443d6cc9e2eecfc4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5651.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5651.dmg

hifi-gustavo:
Build console output for commit: 7410baedf0a3499eb7ef02085a8c3d8cc1eaf5cd available at: https://gist.github.com/78a2443d6cc9e2eecfc4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5651.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5651.dmg

PR #5659, opened 2015-08-25
Fix UserInputMapper event generation

jherico:
Currently the userinputmapper sends action events for controllers for any non-zero value on every update cycle.  This is unhelpful for two reasons.

* It's no better than polling the value during your script's update method
* It's actually worse than polling because you never get an event telling you the button was released

This PR modifies the event sending logic.  Instead of sending an event whenever a value is non-zero, it stores the last sent state of every action, and sends an event if the current value is different from the one last sent.

This means that instead of getting a series of "1" values as long as the button is held down, you instead get a single "1" value when the button is pressed.  Additionally you now get a "0" value when the button is released.  This will allow for much more natural event based programming patterns inside JS scripts that wish to use events rather than polling for their behavior.

### DIFF
--- a/libraries/input-plugins/src/input-plugins/UserInputMapper.cpp
+++ b/libraries/input-plugins/src/input-plugins/UserInputMapper.cpp
@@ -226,9 +226,12 @@ void UserInputMapper::update(float deltaTime) {
     }
 
     // Scale all the channel step with the scale
+    static const float EPSILON =  0.01f;
     for (auto i = 0; i < NUM_ACTIONS; i++) {
         _actionStates[i] *= _actionScales[i];
-        if (_actionStates[i] > 0) {
+        // Emit only on change, and emit when moving back to 0
+        if (fabs(_actionStates[i] - _lastActionStates[i]) > EPSILON) {
+            _lastActionStates[i] = _actionStates[i];
             emit actionEvent(i, _actionStates[i]);
         }
         // TODO: emit signal for pose changes
--- a/libraries/input-plugins/src/input-plugins/UserInputMapper.h
+++ b/libraries/input-plugins/src/input-plugins/UserInputMapper.h
@@ -248,6 +248,7 @@ protected:
  
     std::vector<float> _actionStates = std::vector<float>(NUM_ACTIONS, 0.0f);
     std::vector<float> _actionScales = std::vector<float>(NUM_ACTIONS, 1.0f);
+    std::vector<float> _lastActionStates = std::vector<float>(NUM_ACTIONS, 0.0f);
     std::vector<PoseValue> _poseStates = std::vector<PoseValue>(NUM_ACTIONS);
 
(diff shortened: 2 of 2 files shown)


hifi-gustavo:
Build console output for commit: 1f5c0fe7a5480f20c178382b68a040d6b1367d02 available at: https://gist.github.com/c0d6743ab2643afcd747<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5659.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5659.dmg

libraries/input-plugins/src/input-plugins/UserInputMapper.cpp
```diff
@@ -226,9 +226,13 @@ void UserInputMapper::update(float deltaTime) {
     }
 
     // Scale all the channel step with the scale
+    static std::vector<float> _lastActionStates( NUM_ACTIONS, 0.0f );
```
ZappoMan: why not make this a member variable?

libraries/input-plugins/src/input-plugins/UserInputMapper.cpp
```diff
@@ -226,9 +226,13 @@ void UserInputMapper::update(float deltaTime) {
...
+    static const float EPSILON{ 0.01f };
```
ZappoMan: personally I would prefer 
```
static const float EPSILON = 0.01f;
```
to match all the other code for PODT

libraries/input-plugins/src/input-plugins/UserInputMapper.cpp
jherico: lazy, didn't want to modify the header since I was in the middle of my dev cycle.  will fix.

libraries/input-plugins/src/input-plugins/UserInputMapper.cpp
jherico: fixed.

hifi-gustavo:
Build console output for commit: 854b2ab786096701789b968f0f003e70c5d40b92 available at: https://gist.github.com/c0d6743ab2643afcd747<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5659.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5659.dmg

hifi-gustavo:
Build console output for commit: 073567558ea3728ed691f6491a9e0c37610cb6e3 available at: https://gist.github.com/c0d6743ab2643afcd747<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5659.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5659.dmg

PR #5660, opened 2015-08-26
Polyvox improvements

sethalves:
- When edit.js creates a polyvox entity, turn on some of the voxel so the user can see it.
- voxel.js has a toolbar.
- Editing a polyvox entity is less likely to cause interface to "hiccup".
- By default polyvox entities now have the less confusing "edged-cubic" surface style.


hifi-gustavo:
Build console output for commit: 6f4683206936b124d092a4d5ed75e355f4a95a71 available at: https://gist.github.com/61152495a8946b45bad6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.dmg

hifi-gustavo:
Build console output for commit: 23dbbcc7bb982aa8f6c73991d9b2effddf7e242c available at: https://gist.github.com/61152495a8946b45bad6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.dmg

hifi-gustavo:
Build console output for commit: b09de4ff653a7f50d25a1f4f25e59d89265ab851 available at: https://gist.github.com/61152495a8946b45bad6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.dmg

hifi-gustavo:
Build console output for commit: e97e7de19c1cd052a4eb2b1f506ac9f5e691ce43 available at: https://gist.github.com/61152495a8946b45bad6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.dmg

HeadClot:
Wait as in the Volumes of Fun Polyvox?

Just curious here.

sethalves:
Yes, this PR involves improvements to integration of the PolyVox library from http://www.volumesoffun.com/

examples/edit.js
```diff
@@ -448,12 +448,31 @@ var toolBar = (function () {
...
+                Entities.setVoxel(polyVoxId, {x: 14, y: y, z: 1}, 255);
+                Entities.setVoxel(polyVoxId, {x: 1, y: y, z: 14}, 255);
+                Entities.setVoxel(polyVoxId, {x: 14, y: y, z: 14}, 255);
+            }
+            for (var z = 2; z <= 13; z++) {
+                Entities.setVoxel(polyVoxId, {x: 1, y: 1, z: z}, 255);
+                Entities.setVoxel(polyVoxId, {x: 14, y: 1, z: z}, 255);
+                Entities.setVoxel(polyVoxId, {x: 1, y: 14, z: z}, 255);
+                Entities.setVoxel(polyVoxId, {x: 14, y: 14, z: z}, 255);
+            }
```
AndrewMeadows: It looks like this code creates a 14 (or maybe 15) unit wide "wireframe" cube in voxel space.  Is that right?

sethalves: That's right.  I still haven't worked out what to do if a polyvox becomes empty -- without some scripting, there's no longer a way to interact with it.  This fills in a default shape when a polyvox is created with edit.js -- one that's easy to see but not too hard to trim down.  Also, if the avatar is in the middle, there's an easy way to escape.

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -188,15 +188,23 @@ void PhysicsEngine::addObjects(VectorOfMotionStates& objects) {
...
         } else if (flags & EASY_DIRTY_PHYSICS_FLAGS) {
-            object->handleEasyChanges(flags, this);
+            if (object->handleEasyChanges(flags, this)) {
+                object->clearIncomingDirtyFlags();
+                stillNeedChange.push_back(object);
+            }
         }
     }
+    return stillNeedChange;
 }
```
AndrewMeadows: When **MotionState::handleHardAndEasyChanges()** returns **true** then we **(a)** clear dirty flags and **(b)** push the object onto the **stillNeedChange** list.  These two operations seem contradictory to each other:  why clear dirty flags if it still needs change?

Looking at the implementation of **handleHardAndEasyChanges()** I see that it returns **false** when the changes are **not** handled (e.g. when the shape is not yet known).

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
```diff
@@ -500,159 +370,571 @@ bool RenderablePolyVoxEntityItem::findDetailedRayIntersection(const glm::vec3& o
...
+    EntityTreeElement* element = getElement();
+    EntityTree* tree = element ? element->getTree() : nullptr;
+    EntitySimulation* simulation = tree ? tree->getSimulation() : nullptr;
+    PhysicalEntitySimulation* peSimulation = static_cast<PhysicalEntitySimulation*>(simulation);
+    EntityEditPacketSender* packetSender = peSimulation ? peSimulation->getPacketSender() : nullptr;
+    if (packetSender) {
+        packetSender->queueEditEntityMessage(PacketType::EntityEdit, _id, properties);
+    }
+    _threadRunning.release();
+}
```
AndrewMeadows: This does not block the merge but let me reveal some of my long distance plans:  I want to move **PhysicalEntitySimulation** out of the **physics** lib and into some other group -- the  ultimate goal is to make **entities** lib depend on **physics** rather than the other way around.  I guess exposing **PhysicalEntitySimulation** here narrows my options: it can't go into **interface** like I was planning --> it will have to go into **entities** lib.

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
```diff
@@ -33,15 +35,17 @@
...
 #include "model/Geometry.h"
-#include "gpu/Context.h"
 #include "EntityTreeRenderer.h"
 #include "polyvox_vert.h"
 #include "polyvox_frag.h"
 #include "RenderablePolyVoxEntityItem.h"
+#include "EntityEditPacketSender.h"
+#include "PhysicalEntitySimulation.h"
+
+#define THREAD_POLYVOX 1
```
AndrewMeadows: Is this an unused custom **define** or is **THREAD_POLYVOX** used by the polyvox library?

hifi-gustavo:
Build console output for commit: 92b5fe34577ad4729f69cd462ba935269d94c71c available at: https://gist.github.com/61152495a8946b45bad6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.dmg

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
```diff
@@ -500,159 +370,571 @@ bool RenderablePolyVoxEntityItem::findDetailedRayIntersection(const glm::vec3& o
...
+                 << getName() << getID();
+        _threadRunning.release();
+        return;
+    }
+
+    _volDataLock.lockForWrite();
+    if (!_volData) {
+        _volDataLock.unlock();
+        return;
+    }
```
AndrewMeadows: This return mode doesn't release **_threadRunning**.  Should it?

hifi-gustavo:
Build console output for commit: a5cd3ff04603e0c6ee5653d695c046764f0fba0f available at: https://gist.github.com/61152495a8946b45bad6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.dmg

hifi-gustavo:
Build console output for commit: 540af4afcbdcbc24000a19e642da0e60e35d27b5 available at: https://gist.github.com/61152495a8946b45bad6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.dmg

hifi-gustavo:
Build console output for commit: 752c787f0b61ee2021ef15c14dd16abea53b02b3 available at: https://gist.github.com/61152495a8946b45bad6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.dmg

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.h
```diff
@@ -38,7 +43,7 @@ namespace render {
 
 
 class RenderablePolyVoxEntityItem : public PolyVoxEntityItem {
-public:
+ public:
```
AndrewMeadows: Accidental indentation.

hifi-gustavo:
Build console output for commit: 7614cababb950c87199a67762f7fd933733604e8 available at: https://gist.github.com/61152495a8946b45bad6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.dmg

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
```diff
@@ -500,159 +368,572 @@ bool RenderablePolyVoxEntityItem::findDetailedRayIntersection(const glm::vec3& o
...
-
-    _dirtyFlags |= EntityItem::DIRTY_SHAPE | EntityItem::DIRTY_MASS;
-    _needsModelReload = true;
-
-    #ifdef WANT_DEBUG
-    qDebug() << "RenderablePolyVoxEntityItem::compressVolumeData" << (usecTimestampNow() - startTime) << getName();
-    #endif
+    _meshLock.unlock();
+    return true;
 }
```
AndrewMeadows: At first I was squinting at this code thinking: "There is already a EntityItem::_dirtyFlags bit for when the shape is dirty, so why do we need this new _meshDirty data member?"  However the EntityItem::_dirtyFlags is really about whether the data in the **EntityItem** is dirty relative to what the **PhysicsEngine** knows about the corresponding **RigidBody**, and if there is some other context of "dirty" about whether the polyvox's mesh is dirty with respect to the voxel states then: "OK, maybe there really is a new "dirty" bit that needs to be tracked".

But then I started to audit where **_meshDirty** is changed...

Pivoting above on the **true** value of **_meshDirty**... each call to EntityMotionState::isReadyToComputeShape() on a Polyvox entity item eventually gets to the RenderablePolyVoxEntityItem::isReadyToComputeShape() method above, which calls computeShapeInfoWorker() which calls computeShapeInfoWorkerAsync()... on another thread right?

Meanwhile **_meshDirty** is not set **false** until the first computeShapeInfoWorkerAsync() completes.  What I'm wondering is... how do we prevent multiple calls to computeShapeInfoWorkerAsync() from queuing up in the QtConcurrent thread pool if the first call takes a while to complete?  In other words... maybe **_meshDirty**  should be set **false** as soon a computeShapeInfoWorkerAsync() is queued up?

Or put another way: if we changed the name of **_meshDirty** to be **_needToKickOffComputeShapeInfoWorkerAsync()** would that more correctly represent what it is used for? and if so, would we really wait until the computeShapeInfoWorkerAsync() completes to set it **false** or should it be done when it is first kicked off?

What is to prevent

sethalves: they do queue up... or, rather, because the semaphore is of size 1, when the 2nd thread tries to start, it blocks.  I had attempted to allow more than one side-thread, but the code got a lot more complicated.  It seems that for the use case I'm trying to fix, one side-thread is enough to keep the mouse from momentarily freezing.

This should probably be revisited, but what's there now is a big improvement in usability.

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
```diff
@@ -500,159 +368,572 @@ bool RenderablePolyVoxEntityItem::findDetailedRayIntersection(const glm::vec3& o
...
+        case PolyVoxEntityItem::SURFACE_EDGED_MARCHING_CUBES:
+            if (x < 0 || y < 0 || z < 0 ||
+                x >= vol->getWidth() - 2 || y >= vol->getHeight() - 2 || z >= vol->getDepth() - 2) {
+                return false;
+            }
+            return true;
+    }
+
+    return false;
+}
```
AndrewMeadows: I only mention this because it looks like maybe RenderablePolyVoxEntityItem::inUserBounds() should be performant.  It could be implemented without a **switch** and even without the **if**:

    bool RenderablePolyVoxEntityItem::inUserBounds(const PolyVox::SimpleVolume<uint8_t>* vol,
            PolyVoxEntityItem::PolyVoxSurfaceStyle surfaceStyle,
            int x, int y, int z) {
        int offset = (surfaceStyle == SURFACE_MARCHING_CUBES || surfaceStyle == SURFACE_CUBIC) ? 0 : 2;
        return !((x < 0 || y < 0 || z < 0 ||
+                x >= vol->getWidth() - offset || y >= vol->getHeight() - offset || z >= vol->getDepth() - offset));
    }

sethalves: Well... yes but I find this much harder to understand.  I don't think this code needs to be especially optimized.

libraries/entities/src/PolyVoxEntityItem.cpp
```diff
@@ -185,9 +186,17 @@ void PolyVoxEntityItem::debugDump() const {
...
-        return;
-    }
-    updateVoxelSurfaceStyle(voxelSurfaceStyle);
+void PolyVoxEntityItem::setVoxelData(QByteArray voxelData)
+{
+    _voxelDataLock.lockForWrite();
+    _voxelData = voxelData;
+    _voxelDataDirty = true;
+    _voxelDataLock.unlock();
+}
```
AndrewMeadows: Incorrect K&R indentation.

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
AndrewMeadows: I'm sure the threading makes for improved usability but the question remains: should **_meshDirty** be set false **right before** the thread is queued rather than at the end of the asynch operation?  Even if only one thread is ever allowed to run at a time it seems to me that if it didn't complete quickly then this code would get called again (unecessarily!) before its done (and would then block?) because **_meshDirty** is still true -- it won't be false until the first async operation is done.

hifi-gustavo:
Build console output for commit: f6fbf4f3242aac0253c8960cc71c2f5f712476c3 available at: https://gist.github.com/61152495a8946b45bad6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5660.dmg

PR #5669, opened 2015-08-27
first cut at breakdance toy

ZappoMan:
this is a little script that demonstrates making a "puppet" that is controlled by using the hydras (or vive) controller and moving your hands in different poses


imgntn:
looks good to me -- some similar code around creating poses locations, combining them, and tying them to animations that could probably be abstracted if we wanted to help devs make their own pose scenarios, but for purposes of toybox & demoing it is A++ --- super fun to play with, great job!

PR #5671, opened 2015-08-27
don't read ice-server packets that are too small

birarda:
This is an interim fix for our crashing ice-server. We will need to add guarding for header reading when the header isn't actually that large to networking layer.


hifi-gustavo:
Build console output for commit: b4e624a64d369e8c37c2cc18f27cb4633c5f4698 available at: https://gist.github.com/1b6a13092dc4757f9379<br/><br/>No installer was built for this commit

PR #5675, opened 2015-08-28
remove PhysicsEntity class

AndrewMeadows:
The PhysicsEntity class predates EntityItem and our usage of Bullet for physics.  It was the beginning of an effort to split out some functionality out of the Model class but it no longer makes sense.  What little useful code therein moves back into Model and the PhysicsEntity class is removed.


interface/src/avatar/SkeletonModel.cpp
```diff
@@ -51,14 +51,14 @@ void SkeletonModel::initJointStates(QVector<JointState> states) {
...
-                                            rightElbowJointIndex,
-                                            rightShoulderJointIndex);
+    _rig->initJointStates(states, parentTransform,
+                          rootJointIndex,
+                          leftHandJointIndex,
+                          leftElbowJointIndex,
+                          leftShoulderJointIndex,
+                          rightHandJointIndex,
+                          rightElbowJointIndex,
+                          rightShoulderJointIndex);
```
AndrewMeadows: The _boundingRadius wasn't being used anywhere so it is removed entirely.

hifi-gustavo:
Build console output for commit: d3f06af892145b897507f0b0b46b4775267c5ddb available at: https://gist.github.com/59875c39558e80d9ce3f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5675.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5675.dmg

PR #5690, opened 2015-09-01
Entity crash bug

SeijiEmery:
Fixed a bug where scripted entity edits could randomly crash the client

### DIFF
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -1017,8 +1017,10 @@ void EntityTreeRenderer::deletingEntity(const EntityItemID& entityID) {
 }
 
 void EntityTreeRenderer::addingEntity(const EntityItemID& entityID) {
+    _tree->lockForRead();
     checkAndCallPreload(entityID);
     auto entity = static_cast<EntityTree*>(_tree)->findEntityByID(entityID);
+    _tree->unlock();
     if (entity) {
         addEntityToScene(entity);
     }
--- a/libraries/script-engine/src/ScriptEngine.h
+++ b/libraries/script-engine/src/ScriptEngine.h
@@ -35,7 +35,7 @@
 
 const QString NO_SCRIPT("");
 
-const unsigned int SCRIPT_DATA_CALLBACK_USECS = floor(((1.0f / 60.0f) * 1000 * 1000) + 0.5f);
+const unsigned int SCRIPT_DATA_CALLBACK_USECS = (unsigned int)floor(((1.0f / 60.0f) * 1000 * 1000) + 0.5f);
 
 typedef QHash<QString, QScriptValueList> RegisteredEventHandlers;
 
--- a/examples/tests/entities/entityEditStressTest.js
+++ b/examples/tests/entities/entityEditStressTest.js
@@ -1,0 +1,160 @@
+// entityEditStressTest.js
+//
+// Created by Seiji Emery on 8/31/15
+// Copyright 2015 High Fidelity, Inc.
+//
+// Stress tests the client + server-side entity trees by spawning huge numbers of entities in
+// close proximity to your avatar and updating them continuously (ie. applying position edits), 
+// with the intent of discovering crashes and other bugs related to the entity, scripting, 
+// rendering, networking, and/or physics subsystems.
+//
+// This script was originally created to find + diagnose an a clientside crash caused by improper
+// locking of the entity tree, but can be reused for other purposes.
+//
+// Distributed under the Apache License, Version 2.0.
+// See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+var NUM_ENTITIES = 20000;           // number of entities to spawn
+var ENTITY_SPAWN_LIMIT = 1000;
+var ENTITY_SPAWN_INTERVAL = 0.1;
+
+var UPDATE_INTERVAL = 0.05;  // Re-randomize the entity's position every x seconds / ms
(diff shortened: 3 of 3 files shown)


libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -1017,11 +1017,14 @@ void EntityTreeRenderer::deletingEntity(const EntityItemID& entityID) {
...
 
 void EntityTreeRenderer::addingEntity(const EntityItemID& entityID) {
+    _tree->lockForRead();
+    
     checkAndCallPreload(entityID);
     auto entity = static_cast<EntityTree*>(_tree)->findEntityByID(entityID);
     if (entity) {
         addEntityToScene(entity);
     }
+    _tree->unlock();
```
jherico: Why do you wait until after the `addEntityToScene` call to unlock?

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -1017,11 +1017,14 @@ void EntityTreeRenderer::deletingEntity(const EntityItemID& entityID) {
 }
 
 void EntityTreeRenderer::addingEntity(const EntityItemID& entityID) {
+    _tree->lockForRead();
```
jherico: Are you sure that we shouldn't be putting this lock deeper inside the call stack, perhaps *inside* the `findEntityItemByID`?  Is there any case where a caller of `findEntityItemByID` wouldn't need the same locking?

jherico:
None of these commit comments are informative as to what this PR is supposed to do.  You may want to consider using `git rebase -i` to compress the PR to a single commit with descriptive comment

hifi-gustavo:
Build console output for commit: f7a39fb4419c73c15375d6655b4ca56717b8573d available at: https://gist.github.com/bf5a314ef78ca671de97<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5690.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5690.dmg

libraries/entities-renderer/src/EntityTreeRenderer.cpp
SeijiEmery: good point; I'll move that up

libraries/entities-renderer/src/EntityTreeRenderer.cpp
SeijiEmery: In all other cases, the tree is locked when findEntityItemByID gets called (I think). Calling unlock there would remove lock protection from the code that called it (right?) and cause more bugs.

libraries/script-engine/src/ScriptEngine.h
```diff
@@ -35,7 +35,7 @@
 
 const QString NO_SCRIPT("");
 
-const unsigned int SCRIPT_DATA_CALLBACK_USECS = floor(((1.0f / 60.0f) * 1000 * 1000) + 0.5f);
+const unsigned int SCRIPT_DATA_CALLBACK_USECS = (unsigned int)floor(((1.0f / 60.0f) * 1000 * 1000) + 0.5f);
```
SeijiEmery: Not relevant to the rest of the commit; just one less warning we have to worry about

hifi-gustavo:
Build console output for commit: af7430dbb6ae8f332216b67dca81b590929c378f available at: https://gist.github.com/bf5a314ef78ca671de97<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5690.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5690.dmg

libraries/entities-renderer/src/EntityTreeRenderer.cpp
jherico: We may want to investigate changing the lock semantics to be recursive so that nested calls to lock / unlock aren't problematic.  However that's beyond the scope of this PR.

imgntn:
if someone could take a look at this, i'd love this to get merged, my entity scripts are crashing all time time... :+1:

hifi-gustavo:
Build console output for commit: d52c3017d1b7f6a492d686fdea3536fca48f29c5 available at: https://gist.github.com/bf5a314ef78ca671de97<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5690.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5690.dmg

hifi-gustavo:
Build console output for commit: 6e6d92172315f3c2c83aaac504bbeb8e803ce79b available at: https://gist.github.com/bf5a314ef78ca671de97<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5690.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5690.dmg

libraries/entities-renderer/src/EntityTreeRenderer.cpp
ZappoMan: Why don't we have a separate lock for the the entityID map? There really are two different pieces of data we want to protect, one would be the tree structure itself, the other is the entityID dictionary.

Also, I agree with Austin that this lock should be in the EntityTree... not the EntityTreeRenderer.

ZappoMan:
This needs to be re-implemented to use a fine grain lock around just the ID list, and not the entire tree. The locking mechanism should be inside the EntityTree and not require callers to lock the entire tree.

PR #5693, opened 2015-09-01
Asset Transfer Protocol and UDT in networking layer

birarda:
Summary of the features added by this PR:
- UDT layer in networking library for reliable congestion and rate controlled sending
- new assignment-client type "asset-server" (type 3) - disabled in all domains by default
- AssetClient/AssetResourceRequest to download/upload ATP files from Interface
- reworking of resource request system to seamlessly handle HTTP/file/ATP requests

### Asset Transfer Protocol (ATP)

ATP is a new protocol High Fidelity clients and servers can use to send assets (meshes, sounds, images, scripts) to each other over their existing hole-punched UDP connections. It is meant to eventually replace the serving of existing HTTP/TCP assets because of the following advantages :

- no requirement for content creator to own a separate HTTP webserver or S3 bucket to serve assets
- fairer than TCP to other concurrent connections
- less aggresive than TCP at backing off when loss occurs (and quicker to recover from a one-off loss event)
- easy direct LAN serving of files to locally connected clients
- client downstream saturation by concurrent requesting of file blocks from multiple asset-servers (future improvement)

ATP URLS are of the form `atp:$HASH.$EXT`. Here's an example:

`atp:89fcfdc5a74f3c78fa277ec4edf8fef367224bdc102dbe9794eb4788f537eac7.fbx`

Currently, when Interface is given an ATP URL (ex: for scripts, entities, or avatars) it confirms with the `asset-server` in its current domain that a file with the given hash and extension exists, and then it sends a request for the `asset-server` to send it the entire file. The `asset-server` dispatches a message that contains the data of the file via a UDT connection.

For clients, much of the ATP request/response is hidden in the `AssetResourceRequest` and `AssetClient` classes and networking layer. Callers that currently use HTTP downloads should expect an almost identical interface for success and failure for ATP downloads.

The `asset-server` will initially not be a default assigned type in domains. This is to allow for further testing of the new system in High Fidelity domains with many remote clients before suggesting widespread adoption in place of HTTP assets.

One major shortcoming of ATP in its current state is that it cannot handle files that contain links to other files via relative paths (meshes are the biggest offender here). This is due to the fact that there is currently no ability to reference a file in ATP by name. We have some potential solutions to this that we will explore in the coming weeks.

Preliminary support for asset upload is enabled in interface via a Menu option at Developer->Assets. If the upload finishes successfully, Interface presents a dialog with a copiable ATP URL. If it fails, it shows an error dialog that explains why the upload failed. Ideally this would have been done via QML but a QML-friendly file dialog is not readily available. We will likely need to add one for global use in Interface, at which point the asset upload flow should be moved over as well.  

### UDT

The reliable ordered sending required by the Asset Transfer Protocol is acheived by the addition of the [UDP-based data transfer (UDT) protocol](http://udt.sourceforge.net). We have only very slightly modified the behaviour of the fourth official version of the UDT protocol, created by Dr. Yunhong Gu and Dr. Robert L. Grossman.

Nodelist/LimitedNodeList UDP datagrams now flow through `udt::Socket` instead of directly through `QUdpSocket`. This allows the `udt::Socket` to determine if it needs simply to send the packet off, or hand it off to a matching `udt::Connection`. 

The entry point for our implemenation of the UDT protocol is `udt::Connection`. At any time that a node is receiving or sending data to a specific socket (address and port), they will have a single connection open to that socket. While sending, the `udt::Connection` will spawn a `udt::SendQueue` on a separate thread (this is required so the thread can sleep for the packet send period and maintain the correct send rate). *It is likely in the future that it will make more sense to create one `udt::Connection` per request-response, as this will allow concurrent data flows between the same client/server that share the same available bandwidth.*

The reciever in a UDT connection sends NAK packets to ask the sender to re-send anything that was lost. In order to reduce the amount of packet loss and therefore re-sent packets, UDT uses a number of techniques to estimate the available bandwidth and current congestion for each connection. It attempts to use as much of the link as possible while being very fair to other protocols also using the same network connection. 

The bulk of the UDT magic can be found in the `udt::DefaultCC` class. `udt::DefaultCC` is a subclass of `udt::CongestionControl`, a base class that UDT connections can leverage (via its subclasses) to adopt using any number of congestion control styles. The default implementation `udt::DefaultCC` is our slightly modified version of the congestion control technique presented in Dr. Gu's and Dr. Grossman's UDT papers.

Currently UDT can handle the following types of packets:
- unreliable, unordered
    - this goes direct from `udt::Socket` to the wire
- reliable, unordered
    - sent via the `udt::SendQueue` of a `udt::Connection`
- reliable, ordered
    - sent via the `udt::SendQueue` of a `udt::Connection`, re-constructed by the `udt::Connection` on receiver

The obvious missing combination is unreliable, but ordered. Support for this will be added shortly - the sender will send a block of data that the receiver knows it should drop as lost as soon as it loses any of the packets in the message.

### WWCD

With @Atlante45 in France we truly couldn't have completed this work without the help of Clerb. 

clerb

Much of our design is driven by the simple question "What would Clerb do?" - the answer is usually just to move it.


hifi-gustavo:
Build console output for commit: 33f69335443413c13365c773b7beaafdb5f8869a available at: https://gist.github.com/bfeb12c0376fe6de3549<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5693.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5693.dmg

PR #5694, opened 2015-09-01
Asset Transfer Protocol and UDT in networking layer

huffman:
Summary of the features added by this PR:
- UDT layer in networking library for reliable congestion and rate controlled sending
- new assignment-client type "asset-server" (type 3) - disabled in all domains by default
- AssetClient/AssetResourceRequest to download/upload ATP files from Interface
- reworking of resource request system to seamlessly handle HTTP/file/ATP requests

### Asset Transfer Protocol (ATP)

ATP is a new protocol High Fidelity clients and servers can use to send assets (meshes, sounds, images, scripts) to each other over their existing hole-punched UDP connections. It is meant to eventually replace the serving of existing HTTP/TCP assets because of the following advantages :

- no requirement for content creator to own a separate HTTP webserver or S3 bucket to serve assets
- fairer than TCP to other concurrent connections
- less aggresive than TCP at backing off when loss occurs (and quicker to recover from a one-off loss event)
- easy direct LAN serving of files to locally connected clients
- client downstream saturation by concurrent requesting of file blocks from multiple asset-servers (future improvement)

ATP URLS are of the form `atp:$HASH.$EXT`. Here's an example:

`atp:89fcfdc5a74f3c78fa277ec4edf8fef367224bdc102dbe9794eb4788f537eac7.fbx`

Currently, when Interface is given an ATP URL (ex: for scripts, entities, or avatars) it confirms with the `asset-server` in its current domain that a file with the given hash and extension exists, and then it sends a request for the `asset-server` to send it the entire file. The `asset-server` dispatches a message that contains the data of the file via a UDT connection.

For clients, much of the ATP request/response is hidden in the `AssetResourceRequest` and `AssetClient` classes and networking layer. Callers that currently use HTTP downloads should expect an almost identical interface for success and failure for ATP downloads.

The `asset-server` will initially not be a default assigned type in domains. This is to allow for further testing of the new system in High Fidelity domains with many remote clients before suggesting widespread adoption in place of HTTP assets.

One major shortcoming of ATP in its current state is that it cannot handle files that contain links to other files via relative paths (meshes are the biggest offender here). This is due to the fact that there is currently no ability to reference a file in ATP by name. We have some potential solutions to this that we will explore in the coming weeks.

Preliminary support for asset upload is enabled in interface via a Menu option at Developer->Assets. If the upload finishes successfully, Interface presents a dialog with a copiable ATP URL. If it fails, it shows an error dialog that explains why the upload failed. Ideally this would have been done via QML but a QML-friendly file dialog is not readily available. We will likely need to add one for global use in Interface, at which point the asset upload flow should be moved over as well.  

### UDT

The reliable ordered sending required by the Asset Transfer Protocol is acheived by the addition of the [UDP-based data transfer (UDT) protocol](http://udt.sourceforge.net). We have only very slightly modified the behaviour of the fourth official version of the UDT protocol, created by Dr. Yunhong Gu and Dr. Robert L. Grossman.

Nodelist/LimitedNodeList UDP datagrams now flow through `udt::Socket` instead of directly through `QUdpSocket`. This allows the `udt::Socket` to determine if it needs simply to send the packet off, or hand it off to a matching `udt::Connection`. 

The entry point for our implemenation of the UDT protocol is `udt::Connection`. At any time that a node is receiving or sending data to a specific socket (address and port), they will have a single connection open to that socket. While sending, the `udt::Connection` will spawn a `udt::SendQueue` on a separate thread (this is required so the thread can sleep for the packet send period and maintain the correct send rate). *It is likely in the future that it will make more sense to create one `udt::Connection` per request-response, as this will allow concurrent data flows between the same client/server that share the same available bandwidth.*

The reciever in a UDT connection sends NAK packets to ask the sender to re-send anything that was lost. In order to reduce the amount of packet loss and therefore re-sent packets, UDT uses a number of techniques to estimate the available bandwidth and current congestion for each connection. It attempts to use as much of the link as possible while being very fair to other protocols also using the same network connection. 

The bulk of the UDT magic can be found in the `udt::DefaultCC` class. `udt::DefaultCC` is a subclass of `udt::CongestionControl`, a base class that UDT connections can leverage (via its subclasses) to adopt using any number of congestion control styles. The default implementation `udt::DefaultCC` is our slightly modified version of the congestion control technique presented in Dr. Gu's and Dr. Grossman's UDT papers.

Currently UDT can handle the following types of packets:
- unreliable, unordered
    - this goes direct from `udt::Socket` to the wire
- reliable, unordered
    - sent via the `udt::SendQueue` of a `udt::Connection`
- reliable, ordered
    - sent via the `udt::SendQueue` of a `udt::Connection`, re-constructed by the `udt::Connection` on receiver

The obvious missing combination is unreliable, but ordered. Support for this will be added shortly - the sender will send a block of data that the receiver knows it should drop as lost as soon as it loses any of the packets in the message.

### WWCD

With @Atlante45 in France we truly couldn't have completed this work without the help of Clerb. 

clerb

Much of our design is driven by the simple question "What would Clerb do?" - the answer is usually just to move it.


hifi-gustavo:
Build console output for commit: 33f69335443413c13365c773b7beaafdb5f8869a available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

hifi-gustavo:
Build console output for commit: f24a267d83779d7906650692fa66acf1b20d7a29 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

hifi-gustavo:
Build console output for commit: d29449b43a45ca816ba235b0c7dbcaedd7d31e67 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

hifi-gustavo:
Build console output for commit: 231ea3b240f2c212349ccd4f5457d9b1a2314482 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

hifi-gustavo:
Build console output for commit: e5a10e3174dacd1ae1f3317d64ad439bde5c34e7 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

Atlante45:
test this please

hifi-gustavo:
Build console output for commit: e5a10e3174dacd1ae1f3317d64ad439bde5c34e7 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

cozza13:
I got this error on the assignment client when loading the stack manager: 

dyld: Library not loaded: @executable_path/QtWebSockets.framework/Versions/5/QtWebSockets

birarda:
@cozza13 that is unrelated to this PR, was introduced with the addition of web sockets for scripting. @murillodigital needs to add the required shared library to be distributed with the Stack Manager.

murillodigital:
Yeah Huffman reported that bug yesterday, that missing framework is happening on PR builds of the stack manager, I will look into that today @birarda @cozza13

hifi-gustavo:
Build console output for commit: ed19987b7dab557f449c69d9a0133e35b3b7edb3 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

hifi-gustavo:
Build console output for commit: 3875a7d8811575dabab0270bd6843537a1aff143 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

hifi-gustavo:
Build console output for commit: 875d10575f50575bf0f5eccd64b130afcb78e591 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

assignment-client/src/AssignmentClientApp.h
```diff
@@ -17,15 +17,15 @@
 
 const QString ASSIGNMENT_TYPE_OVERRIDE_OPTION = "t";
 const QString ASSIGNMENT_POOL_OPTION = "pool";
+const QString ASSIGNMENT_CLIENT_LISTEN_PORT_OPTION = "p";
 const QString ASSIGNMENT_WALLET_DESTINATION_ID_OPTION = "wallet";
-const QString CUSTOM_ASSIGNMENT_SERVER_HOSTNAME_OPTION = "a";
-const QString CUSTOM_ASSIGNMENT_SERVER_PORT_OPTION = "p";
+const QString CUSTOM_ASSIGNMENT_SERVER_HOSTNAME_OPTION = "i";
```
ZappoMan: why are we changing this option from "a" to "i"?

assignment-client/src/assets/AssetServer.cpp
```diff
@@ -0,0 +1,167 @@
...
+        auto filename = fileInfo.fileName();
+        if (!filenameRegex.exactMatch(filename)) {
+            qDebug() << "Found file: " << filename;
+            if (!fileInfo.isReadable()) {
+                qDebug() << "\tCan't open file for reading: " << filename;
+                continue;
+            }
+
+            // Read file
+            QFile file { fileInfo.absoluteFilePath() };
```
ZappoMan: not a merge blocker. But this seems like it might be kind of expensive for a very large file like a >100MB mesh. Should we consider:
* doing this re-indexing in an asynchronous manner
* implement a mechanism that reads the file in chunks and calculates the hash in chunks

assignment-client/src/assets/SendAssetTask.cpp
```diff
@@ -0,0 +1,85 @@
...
+    qDebug() << "Received a request for the file (" << messageID << "): " << hexHash << " from " << start << " to " << end;
+    
+    qDebug() << "Starting task to send asset: " << hexHash << " for messageID " << messageID;
+    auto replyPacketList = std::unique_ptr<NLPacketList>(new NLPacketList(PacketType::AssetGetReply, QByteArray(), true, true));
+
+    replyPacketList->write(assetHash);
+
+    replyPacketList->writePrimitive(messageID);
+
+    if (end <= start) {
```
ZappoMan: Why <=? Shouldn't I be able to request a single byte from an asset? I want byte 100 to 100...

assignment-client/src/assets/UploadAssetTask.cpp
```diff
@@ -0,0 +1,80 @@
...
+        auto hash = hashData(fileData);
+        auto hexHash = hash.toHex();
+        
+        qDebug() << "Hash for uploaded file from" << uuidStringWithoutCurlyBraces(_senderNode->getUUID())
+            << "is: (" << hexHash << ") ";
+        
+        QFile file { _resourcesDir.filePath(QString(hexHash)) + "." + QString(extension) };
+        
+        if (file.exists()) {
+            qDebug() << "[WARNING] This file already exists: " << hexHash;
```
ZappoMan: shouldn't we return an error in this case?

```
 replyPacket->writePrimitive(AssetServerError::AssetAlreadyExists);
```

assignment-client/src/assets/UploadAssetTask.cpp
```diff
@@ -0,0 +1,80 @@
...
+    
+    auto replyPacket = NLPacket::create(PacketType::AssetUploadReply);
+    replyPacket->writePrimitive(messageID);
+    
+    if (fileSize > MAX_UPLOAD_SIZE) {
+        replyPacket->writePrimitive(AssetServerError::AssetTooLarge);
+    } else {
+        QByteArray fileData = buffer.read(fileSize);
+        
+        auto hash = hashData(fileData);
```
ZappoMan: same issue as I mentioned before on re-indexing when starting the server. This approach of reading into a QByteArray and then calling hash can be improved to let the QCryptographicHash read directly from a QIODevice

assignment-client/src/audio/AudioMixer.cpp
```diff
@@ -96,14 +96,11 @@ AudioMixer::AudioMixer(NLPacket& packet) :
...
-        PacketType::MicrophoneAudioNoEcho, PacketType::MicrophoneAudioWithEcho,
-        PacketType::InjectAudio, PacketType::SilentAudioFrame,
-        PacketType::AudioStreamStats
-    };
 
-    packetReceiver.registerListenerForTypes(nodeAudioPackets, this, "handleNodeAudioPacket");
+    packetReceiver.registerListenerForTypes({ PacketType::MicrophoneAudioNoEcho, PacketType::MicrophoneAudioWithEcho,
+                                              PacketType::InjectAudio, PacketType::SilentAudioFrame,
+                                              PacketType::AudioStreamStats },
+                                            this, "handleNodeAudioPacket");
```
ZappoMan: indent looks weird here

libraries/networking/src/udt/PacketHeaders.h
```diff
@@ -21,88 +21,80 @@
...
 };
 
 const int NUM_BYTES_MD5_HASH = 16;
 
-const int MAX_PACKET_SIZE = 1450;
-const int MAX_PACKET_HEADER_BYTES = 4 + NUM_BYTES_RFC4122_UUID + NUM_BYTES_MD5_HASH;
-
 typedef char PacketVersion;
 
-typedef uint16_t PacketSequenceNumber;
```
ZappoMan: why get rid of this typedef, isn't using PacketSequenceNumber more informative to the reader than uint16_t?

assignment-client/src/audio/AudioMixer.cpp
```diff
@@ -652,9 +649,6 @@ void AudioMixer::run() {
 
     auto nodeList = DependencyManager::get<NodeList>();
 
-    // we do not want this event loop to be the handler for UDP datagrams, so disconnect
-    disconnect(&nodeList->getNodeSocket(), 0, this, 0);
```
ZappoMan: what is this change for? do we no longer need to unhook to prevent UDP handling?

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -259,8 +259,8 @@ void AvatarMixer::broadcastAvatarData() {
                         return;
                     }
 
-                    PacketSequenceNumber lastSeqToReceiver = nodeData->getLastBroadcastSequenceNumber(otherNode->getUUID());
-                    PacketSequenceNumber lastSeqFromSender = otherNode->getLastSequenceNumberForPacketType(PacketType::AvatarData);
+                    uint16_t lastSeqToReceiver = nodeData->getLastBroadcastSequenceNumber(otherNode->getUUID());
```
ZappoMan: why remove the typedef for PacketSequenceNumber?

Nex-Pro:
Clearly the baguette was photoshopped, I found the un-photoshopped image. Here it is. :)
clement_hifi

assignment-client/src/assets/UploadAssetTask.cpp
huffman: We could maybe return this information, but I'm not sure it matters in terms of the API - you upload a file and get the hash back, regardless of whether the file already existed on the server.

assignment-client/src/audio/AudioMixer.cpp
birarda: an initializer list, followed by `this, "handleNodeAudioPacket"` aligned with the first character of the first arg (the curly bracket)

assignment-client/src/audio/AudioMixer.cpp
birarda: correct

assignment-client/src/assets/AssetServer.cpp
huffman: This only ends up happening for new files that it finds in the `resources/assets` directory that don't look like ATP files - they end up getting renamed so they won't be reprocessed later.

I think using QCryptographicHash with the QIODevice is a good idea, I will look into that.

assignment-client/src/assets/SendAssetTask.cpp
huffman: `start` is inclusive, `end` is exclusive. For byte 100 you would request `start = 100` `end = 101`

assignment-client/src/AssignmentClientApp.h
birarda: I believe that was an accident, flipping back.

hifi-gustavo:
Build console output for commit: c49088846f45c9fb5bf36c0aaf6357bab7e47a58 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

assignment-client/src/assets/AssetServer.cpp
huffman: If we do this we should also do it when handling uploads, but that would require updating PacketList to support reads via the QIODevice interface. Because this isn't a blocker we will put it into the next PR.

hifi-gustavo:
Build console output for commit: b2778bff09542209094184a4e9fecdea5c7f97aa available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

huffman:
@Nex-Pro Haha, apparently we're a Python shop now?

I can't quite make out what is on that screen :P.

hifi-gustavo:
Build console output for commit: 25aeda2e1827cb5d8613bcf5b01d6c2e3973337f available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

assignment-client/src/assets/AssetServer.cpp
ZappoMan: you could also just use the QIODevice for these file based hashes. I'm not suggesting that QIODevice should be the only way you hash files.

assignment-client/src/assets/SendAssetTask.cpp
ZappoMan: hmmm... if end is exclusive, then maybe it's not a good variable name. It's certainly not clear from the comments. Are these 0 or 1 based values? How does a caller know how to ask for the entire asset?

hifi-gustavo:
Build console output for commit: 987814366149d79e5e98c1315cb1d1d193218c02 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

assignment-client/src/assets/SendAssetTask.cpp
huffman: I think there's a precedent for using [inclusive, exclusive) index-based ranges, but I can't think of any c++ examples off the top of my head. Python, Javascript, Java, all have examples of this. Practically speaking it removes the need to constantly be adding/subtracting 1. An alternative is using a start index and a length.

These are 0-based. A caller requests information about the asset before requesting the data itself.

At the minimum I can add comments clarifying the interface. If you feel strongly about the interface itself I would be open to discussing and changing it.

ZappoMan: I assumed you'd be following the same pattern set by other internet standards like HTTP byte-ranges, which are inclusive. 

https://tools.ietf.org/html/rfc7233#section-2.1

ZappoMan: although not a full standard, the new FTP working group has also suggested adding range support to FTP, and their byte ranges are also inclusive. 

https://tools.ietf.org/html/draft-bryan-ftp-range-08

assignment-client/src/assets/UploadAssetTask.cpp
ZappoMan: I'm just saying this seems strange. The operation appears to be a no-op... the file doesn't actually get copied with the new content (AFAICT), and the protocol returns NoError... I would assume the file got copied, but my new content won't actually be there... Am I misunderstanding the code?

birarda: Why would you expect the file to have new content if there's a hash match? File doesn't need to be copied, content is already there.

ZappoMan: Ok - wait... I see now, the hash will only be the same if the content is the same... so this is a no-error case. I get it now.

ZappoMan:
I don't understand why ATP uses hashes for file names. It seems like if I have some resource I will often want to make changes to that resource that would change it's hash. Consider for example something like edit.js... we change it regularly... but we want people to get the latest edit.js. It seems like the use of hashes will make many asset management tasks that people regularly assume exist much more difficult.

assignment-client/src/assets/SendAssetTask.cpp
huffman: Fair point, I think it's important to look at other protocols for guidance but I don't think this is a case where this particular detail is super important. I'd suggest we leave it as-is. If we really want to change this I would suggest we use an offset + length approach.

birarda:
@ZappoMan I don't think a design discussion of that nature belongs in this code review. I'd invite you to raise concerns via email once we send out design doc (today/tomorrow).

ZappoMan:
@birarda - this was the only forum available. since as you pointed out, there's been no design document sent out at this point.

hifi-gustavo:
Build console output for commit: b3f1b1a8e15e7326e6f61d374dd1b80d63e56292 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

hifi-gustavo:
Build console output for commit: b8ac2ff48e0c6253ca6c2b99c0bef9f7a63780b1 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

interface/src/Menu.cpp
```diff
@@ -33,13 +32,15 @@
...
-#endif
+#include "ui/AssetUploadDialogFactory.h"
 #include "ui/DialogsManager.h"
 #include "ui/StandAloneJSConsole.h"
 #include "InterfaceLogging.h"
 
+#if defined(Q_OS_MAC) || defined(Q_OS_WIN)
+#include "SpeechRecognizer.h"
+#endif
+
```
sethalves: Does this belong here?

sethalves: nevermind, I see it's just moved from above.

hifi-gustavo:
Build console output for commit: b8bf9a1104d1112f39eae8a632d8c90e1cff8295 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

hifi-gustavo:
Build console output for commit: e1c395f5ea62bd5911cb0e87063cdeadded3fe62 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

hifi-gustavo:
Build console output for commit: a3959109f1687cecd6c4ed2e0e9f4c7e825d7af5 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

hifi-gustavo:
Build console output for commit: 255f896f7be9e97d3009b0deb51fc36070f23b32 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

hifi-gustavo:
Build console output for commit: e1a31f4dbe61ea6892b20a6bb2783b30e3feb271 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

hifi-gustavo:
Build console output for commit: c670c9a0d55af846826eed5017f0699ed75e5c27 available at: https://gist.github.com/53ce42bedc62a868e074<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5694.dmg

assignment-client/src/avatars/AvatarMixerClientData.cpp
```diff
@@ -24,13 +27,13 @@ bool AvatarMixerClientData::checkAndSetHasReceivedFirstPackets() {
     return oldValue;
 }
 
-PacketSequenceNumber AvatarMixerClientData::getLastBroadcastSequenceNumber(const QUuid& nodeUUID) const {
+uint16_t AvatarMixerClientData::getLastBroadcastSequenceNumber(const QUuid& nodeUUID) const {
```
ZappoMan: elsewhere in this PR - it looks like you use the type AvatarDataSequenceNumber - but here you're using uint16_t... which is it?

libraries/avatars/src/AvatarData.h
```diff
@@ -59,9 +59,11 @@ typedef unsigned long long quint64;
 #include "Recorder.h"
 #include "Referential.h"
 
-typedef std::shared_ptr<AvatarData> AvatarSharedPointer;
-typedef std::weak_ptr<AvatarData> AvatarWeakPointer;
-typedef QHash<QUuid, AvatarSharedPointer> AvatarHash;
+using AvatarSharedPointer = std::shared_ptr<AvatarData>;
```
ZappoMan: we should decide as a team what our coding standard should be with respect to using "using" vs typedef. I can get behind using the new C++ style, since it is required when using templates. But if we're going to do this for simply typedef cases, we should agree to standardize on it, and we should make an effort to fix other places in the code.

assignment-client/src/avatars/AvatarMixerClientData.h
```diff
@@ -36,10 +36,12 @@ class AvatarMixerClientData : public NodeData {
     
     bool checkAndSetHasReceivedFirstPackets();
 
-    PacketSequenceNumber getLastBroadcastSequenceNumber(const QUuid& nodeUUID) const;
-    void setLastBroadcastSequenceNumber(const QUuid& nodeUUID, PacketSequenceNumber sequenceNumber)
+    uint16_t getLastBroadcastSequenceNumber(const QUuid& nodeUUID) const;
```
ZappoMan: same thing here... elsewhere in this PR - it looks like you use the type AvatarDataSequenceNumber - but here you're using uint16_t... we should clean this up.

Nex-Pro:
@huffman  Lego digital designer of course.

PR #5695, opened 2015-09-01
avoid bad joints when computing avatar bounding capsule

AndrewMeadows:
**The bug:** the collision shape approximation for some avatars was too large.

**The problems:**

(1) We would measure the bounding box of the avatar when it was in its "default pose".  It used to be that the default pose was a relaxed stand position, however as we move toward supporting animations for the avatar the new avatar models are coming in with the T-pose as both their "bind" pose and their "default" pose.

(2) There is a bug in our FBXReader code that interprets meshes that are connected to one of the joint bones as their own joints.  Sometimes the transforms for these meshes get messed up in the JointState data and then the measurement of the bounding capsule of the avatar goes bad (radius is too large).

**The fixes:**

(1) We use the IK code to move the hands and feet closer to the body before we measure its bounding shape.

(2) Unfortunately I was unable to find the source of this bug, and had to cutoff my efforts hunting for it, so I've introduced a **HACK** / **WORKAROUND**: when we find "meshes treated as joints" we just ignore them for the bounding collision shape measurement.  This should work well enough to move forward and maybe we can fix the underlying issue when we start providing per-bone collision shapes rather than one big bounding capsule.

### DIFF
--- a/interface/src/avatar/SkeletonModel.cpp
+++ b/interface/src/avatar/SkeletonModel.cpp
@@ -347,9 +347,9 @@ void SkeletonModel::renderOrientationDirections(gpu::Batch& batch, int jointInde
     }
     OrientationLineIDs& jointLineIDs = _jointOrientationLines[jointIndex];
 
-    glm::vec3 pRight	= position + orientation * IDENTITY_RIGHT * size;
-    glm::vec3 pUp		= position + orientation * IDENTITY_UP    * size;
-    glm::vec3 pFront	= position + orientation * IDENTITY_FRONT * size;
+    glm::vec3 pRight = position + orientation * IDENTITY_RIGHT * size;
+    glm::vec3 pU     = position + orientation * IDENTITY_UP    * size;
+    glm::vec3 pFront = position + orientation * IDENTITY_FRONT * size;
 
     glm::vec3 red(1.0f, 0.0f, 0.0f);
     geometryCache->renderLine(batch, position, pRight, red, jointLineIDs._right);
@@ -478,7 +478,7 @@ void SkeletonModel::computeBoundingShape() {
         return;
     }
 
-    // BOUNDING SHAPE HACK: before we measure the bounds of the joints we use IK to put the 
+    // BOUNDING SHAPE HACK: before we measure the bounds of the joints we use IK to put the
     // hands and feet into positions that are more correct than the default pose.
 
     // Measure limb lengths so we can specify IK targets that will pull hands and feet tight to body
--- a/interface/src/avatar/SkeletonModel.cpp
+++ b/interface/src/avatar/SkeletonModel.cpp
@@ -348,7 +348,7 @@ void SkeletonModel::renderOrientationDirections(gpu::Batch& batch, int jointInde
     OrientationLineIDs& jointLineIDs = _jointOrientationLines[jointIndex];
 
     glm::vec3 pRight = position + orientation * IDENTITY_RIGHT * size;
-    glm::vec3 pU     = position + orientation * IDENTITY_UP    * size;
+    glm::vec3 pUp    = position + orientation * IDENTITY_UP    * size;
     glm::vec3 pFront = position + orientation * IDENTITY_FRONT * size;
 
     glm::vec3 red(1.0f, 0.0f, 0.0f);
@@ -523,7 +523,6 @@ void SkeletonModel::computeBoundingShape() {
                             freeLineage,
                             glm::mat4());
                 }
-                const JointState& movedState = _rig->getJointState(tipIndex);
                 break;
             }
             limbLength += limbJoint.getDistanceToParent();
--- a/interface/src/avatar/SkeletonModel.cpp
+++ b/interface/src/avatar/SkeletonModel.cpp
@@ -86,7 +86,7 @@ void SkeletonModel::initJointStates(QVector<JointState> states) {
         _rig->updateJointState(i, rootTransform);
     }
 
-    buildShapes();
+    computeBoundingShape();
 
     Extents meshExtents = getMeshExtents();
     _headClipDistance = -(meshExtents.minimum.z / _scale.z - _defaultEyeModelPosition.z);
@@ -248,6 +248,7 @@ void SkeletonModel::applyHandPosition(int jointIndex, const glm::vec3& position)
                                   rotationBetween(handRotation * glm::vec3(-sign, 0.0f, 0.0f), forearmVector),
                                   true, PALM_PRIORITY);
 }
+
 void SkeletonModel::applyPalmData(int jointIndex, PalmData& palm) {
     if (jointIndex == -1 || jointIndex >= _rig->getJointStateCount()) {
         return;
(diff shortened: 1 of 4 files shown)


interface/src/avatar/SkeletonModel.cpp
```diff
@@ -476,36 +477,87 @@ void SkeletonModel::buildShapes() {
...
+        // since this IK target is totally bogus we restore the tip's relative rotation
+        _rig->setJointRotationInConstrainedFrame(tipIndex, tipRotation, priority);
+    }
+
+    // recompute all joint model-frame transforms
+    glm::mat4 rootTransform = glm::scale(_scale) * glm::translate(_offset) * geometry.offset;
+    for (int i = 0; i < _rig->getJointStateCount(); i++) {
+        _rig->updateJointState(i, rootTransform);
+    }
+    // END BOUNDING SHAPE HACK
```
AndrewMeadows: This is the IK hack that pulls the skeleton into a more correct pose for measuring the bounding capsule.

interface/src/avatar/SkeletonModel.cpp
```diff
@@ -476,36 +477,87 @@ void SkeletonModel::buildShapes() {
...
-        totalExtents.addPoint(jointPosition + axis);
-        totalExtents.addPoint(jointPosition - axis);
+        // HACK WORKAROUND: ignore joints that may have bad translation (e.g. have been flagged as such with zero radius)
+        if (state.getBoneRadius() > 0.0f) {
+            // Each joint contributes a sphere at its position
+            glm::vec3 axis(state.getBoneRadius());
+            glm::vec3 jointPosition = state.getPosition();
+            totalExtents.addPoint(jointPosition + axis);
+            totalExtents.addPoint(jointPosition - axis);
+        }
```
AndrewMeadows: This is where we ignore joints that may have bad transforms.

libraries/animation/src/Rig.cpp
```diff
@@ -9,30 +9,31 @@
...
-    qCDebug(animation, "    modelTranslation = (%.5f, %.5f, %.5f)", modelTranslation.x, modelTranslation.y, modelTranslation.z);
-    qCDebug(animation, "    eyeLookAt = (%.5f, %.5f, %.5f)", eyeLookAt.x, eyeLookAt.y, eyeLookAt.z);
-    qCDebug(animation, "    eyeSaccade = (%.5f, %.5f, %.5f)", eyeSaccade.x, eyeSaccade.y, eyeSaccade.z);
+    qCDebug(animation, "    modelRotation axis = (%.5f, %.5f, %.5f), theta = %0.5f", (double)axis.x, (double)axis.y, (double)axis.z, (double)theta);
+    qCDebug(animation, "    modelTranslation = (%.5f, %.5f, %.5f)", (double)modelTranslation.x, (double)modelTranslation.y, (double)modelTranslation.z);
+    qCDebug(animation, "    eyeLookAt = (%.5f, %.5f, %.5f)", (double)eyeLookAt.x, (double)eyeLookAt.y, (double)eyeLookAt.z);
+    qCDebug(animation, "    eyeSaccade = (%.5f, %.5f, %.5f)", (double)eyeSaccade.x, (double)eyeSaccade.y, (double)eyeSaccade.z);
     qCDebug(animation, "    leanJointIndex = %.d", leanJointIndex);
     qCDebug(animation, "    neckJointIndex = %.d", neckJointIndex);
     qCDebug(animation, "    leftEyeJointIndex = %.d", leftEyeJointIndex);
```
AndrewMeadows: qCDebug only supports formatting **double**, so there was an implicit cast on the **float** arguments here.

libraries/fbx/src/FBXReader.cpp
```diff
@@ -2677,8 +2677,21 @@ FBXGeometry* extractFBXGeometry(const FBXNode& node, const QVariantHash& mapping
...
+            // in the same frame is just BAD.
+            //
+            // HACK WORKAROUND: prevent these shapes from contributing to the collision capsule by setting
+            // some key members of jointShapeInfo to zero:
+            jointShapeInfo.numVertices = 0;
+            jointShapeInfo.sumVertexWeights = 0.0f;
+            jointShapeInfo.numVertexWeights = 0;
+            jointShapeInfo.boneBegin = glm::vec3(0.0f);
+            jointShapeInfo.averageRadius = 0.0f;
         }
```
AndrewMeadows: This is where we detect joints that may have a bad transform.  Really they are meshes being interpeted as joints (because they are "Nodes" in the FBX format with parents... or something -- I don't fully understand it.)

libraries/fbx/src/FBXReader.cpp
```diff
@@ -2728,7 +2741,6 @@ FBXGeometry* extractFBXGeometry(const FBXNode& node, const QVariantHash& mapping
             // the average radius to the average point.
             if (jointShapeInfo.numVertexWeights == 0
                    && jointShapeInfo.numVertices > 0) {
-                jointShapeInfo.averageRadius /= (float)jointShapeInfo.numVertices;
                 joint.boneRadius = jointShapeInfo.averageRadius;
```
AndrewMeadows: This renormalization of the averageRadius is now done earlier.  See code changes above.

hifi-gustavo:
Build console output for commit: 4a7a384c650f811cb5066470846da3b8a1da6026 available at: https://gist.github.com/b237040633387d6f7eaa<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5695.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5695.dmg

hifi-gustavo:
Build console output for commit: a7d57d7c60d0d865df2ae60ac055a32c11c9785c available at: https://gist.github.com/c417d477f30cf8d009ee<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5695.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5695.dmg

sethalves:
test this please

hifi-gustavo:
Build console output for commit: a7d57d7c60d0d865df2ae60ac055a32c11c9785c available at: https://gist.github.com/b237040633387d6f7eaa<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5695.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5695.dmg

hifi-gustavo:
Build console output for commit: eb78b39bea5de68b060188283d0e2b9391c8110e available at: https://gist.github.com/b237040633387d6f7eaa<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5695.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5695.dmg

libraries/fbx/src/FBXReader.cpp
howard-stearns: See also the discussion at https://github.com/highfidelity/hifi/pull/5700#discussion_r38569583 (and the code above it). Looks like you and Anthony are dealing with the same issue in different ways.
(I _think_ you're talking here about the extra bones that are inserted at every extremity (head, fingers, and toe,) so that the parent of the extra bone has a length for visualization.)

PR #5701, opened 2015-09-02
fix sixense and vive rotation readings for palms

AndrewMeadows:
In a previous PR I broke the "palmTip" and "palmRotation" values as used by scripts and some render paths.  This PR fixes those problems.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -4919,7 +4919,7 @@ void Application::setPalmData(Hand* hand, UserInputMapper::PoseValue pose, float
 
     // Store the one fingertip in the palm structure so we can track velocity
     const float FINGER_LENGTH = 0.3f;   //  meters
-    const glm::vec3 FINGER_VECTOR(0.0f, 0.0f, FINGER_LENGTH);
+    const glm::vec3 FINGER_VECTOR(0.0f, FINGER_LENGTH, 0.0f);
     const glm::vec3 newTipPosition = position + rotation * FINGER_VECTOR;
     glm::vec3 oldTipPosition = palm->getTipRawPosition();
     if (deltaTime > 0.0f) {
--- a/interface/src/avatar/SkeletonModel.cpp
+++ b/interface/src/avatar/SkeletonModel.cpp
@@ -261,9 +261,7 @@ void SkeletonModel::applyPalmData(int jointIndex, PalmData& palm) {
     // the palm's position must be transformed into the model-frame
     glm::quat inverseRotation = glm::inverse(_rotation);
     glm::vec3 palmPosition = inverseRotation * (palm.getPosition() - _translation);
-
-    // the palm's "raw" rotation is already in the model-frame
-    glm::quat palmRotation = palm.getRawRotation();
+    glm::quat palmRotation = inverseRotation * palm.getRotation();
 
     inverseKinematics(jointIndex, palmPosition, palmRotation, PALM_PRIORITY);
 }
--- a/libraries/avatars/src/HandData.h
+++ b/libraries/avatars/src/HandData.h
@@ -64,12 +64,13 @@ public:
     bool findSpherePenetration(const glm::vec3& penetratorCenter, float penetratorRadius, glm::vec3& penetration, 
         const PalmData*& collidingPalm) const;
 
+    glm::quat getBaseOrientation() const;
+
     friend class AvatarData;
 protected:
     AvatarData* _owningAvatarData;
     std::vector<PalmData> _palms;
     
-    glm::quat getBaseOrientation() const;
     glm::vec3 getBasePosition() const;
     float getBaseScale() const;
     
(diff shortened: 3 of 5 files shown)


interface/src/avatar/SkeletonModel.cpp
```diff
@@ -261,9 +261,7 @@ void SkeletonModel::applyPalmData(int jointIndex, PalmData& palm) {
     // the palm's position must be transformed into the model-frame
     glm::quat inverseRotation = glm::inverse(_rotation);
     glm::vec3 palmPosition = inverseRotation * (palm.getPosition() - _translation);
-
-    // the palm's "raw" rotation is already in the model-frame
-    glm::quat palmRotation = palm.getRawRotation();
+    glm::quat palmRotation = inverseRotation * palm.getRotation();
```
AndrewMeadows: Part 1: in SkeletonModel::applyPalmData() we get the palm rotation in the **world-frame** and then transform into the skeleton's frame before setting the IK targets.

libraries/avatars/src/HandData.h
```diff
@@ -95,6 +96,7 @@ class PalmData {
 
     void setRawRotation(const glm::quat rawRotation) { _rawRotation = rawRotation; };
     glm::quat getRawRotation() const { return _rawRotation; }
+    glm::quat getRotation() const { return _owningHandData->getBaseOrientation() * _rawRotation; }
```
AndrewMeadows: Add a method for getting the rotation of the palm in the world-frame.

libraries/input-plugins/src/input-plugins/SixenseManager.cpp
```diff
@@ -491,13 +492,15 @@ void SixenseManager::handlePoseEvent(glm::vec3 position, glm::quat rotation, int
     const glm::quat preOffset = glm::angleAxis(sign * PI / 2.0f, zAxis);
 
     // Finally, there is a post-offset (same for both hands) to get the hand's rest orientation
-    // (fingers forward, palm down) aligned properly in the avatar's model-frame.
-    const glm::quat postOffset = glm::angleAxis(PI / 2.0f, xAxis);
+    // (fingers forward, palm down) aligned properly in the avatar's model-frame,
+    // and then a flip about the yAxis to get into model-frame.
+    const glm::quat postOffset = glm::angleAxis(PI, yAxis) * glm::angleAxis(PI / 2.0f, xAxis);
```
AndrewMeadows: Part 2: Add a flip about y-axis for the sixense measurement.

libraries/input-plugins/src/input-plugins/ViveControllerManager.cpp
```diff
@@ -381,7 +380,7 @@ void ViveControllerManager::handlePoseEvent(const mat4& mat, int index) {
     const glm::quat signedQuaterZ = glm::angleAxis(sign * PI / 2.0f, glm::vec3(0.0f, 0.0f, 1.0f)); 
     const glm::quat eighthX = glm::angleAxis(PI / 4.0f, glm::vec3(1.0f, 0.0f, 0.0f));
     const glm::quat offset = glm::inverse(signedQuaterZ * eighthX);
-    rotation = yFlip * rotation * offset * yFlip * quarterX;
+    rotation = rotation * offset * yFlip * quarterX;
```
AndrewMeadows: Part 3: since a flip about the y-axis is its own inverse we remove a flip in the Vive measurement.

interface/src/Application.cpp
```diff
@@ -4919,7 +4919,7 @@ void Application::setPalmData(Hand* hand, UserInputMapper::PoseValue pose, float
 
     // Store the one fingertip in the palm structure so we can track velocity
     const float FINGER_LENGTH = 0.3f;   //  meters
-    const glm::vec3 FINGER_VECTOR(0.0f, 0.0f, FINGER_LENGTH);
+    const glm::vec3 FINGER_VECTOR(0.0f, FINGER_LENGTH, 0.0f);
     const glm::vec3 newTipPosition = position + rotation * FINGER_VECTOR;
     glm::vec3 oldTipPosition = palm->getTipRawPosition();
     if (deltaTime > 0.0f) {
```
AndrewMeadows: Part 4: the finger tip is on the y-axis in the hand's frame, rather than along the z-axis.

hifi-gustavo:
Build console output for commit: 992bd5c9d20e578df833fe7ec406e00e7da0e22b available at: https://gist.github.com/06fa9ba5a131b31792a0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5701.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5701.dmg

jherico:
Fixes my script rotations.

ericrius1:
@AndrewMeadows - the data from Controller.getSpatialControlNormal(tip) still seems to be incorrect. It seems as though it is now off by the same amount, but in the opppsite direction as before

AndrewMeadows:
The Controller.getSpatialControlNormal() problem is fixed in #5704

PR #5705, opened 2015-09-02
OmniTool first pass

jherico:
This is a new library that will allow for modular behavior using hand controllers based on information stored in entity userdata.

Full documentation pending implementation of some modules.


ZappoMan:
Why did you rename these constants.js and utilities.js to austinConstants.js, etc??? Seems like we should keep the original names.

examples/libraries/omniTool.js
```diff
@@ -0,0 +1,304 @@
+//
+//  Created by Bradley Austin Davis on 2015/09/01
+//  Copyright 2015 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+Script.include("austinConstants.js");
```
ZappoMan: why have these labeled "austinXXX.js"?

jherico: I didn't feel confident about putting something called constants or utils into the main libraries, but I'll update.

examples/libraries/omniTool.js
```diff
@@ -0,0 +1,308 @@
...
+//
+
+Script.include("constants.js");
+Script.include("utils.js");
+Script.include("highlighter.js");
+Script.include("omniTool/models/modelBase.js");
+Script.include("omniTool/models/wand.js");
+
+// FIXME brad's requests
+// callback for unloading
```
ZappoMan: looks like you have this in the latest code... you can remove this fixme comment

PR #5706, opened 2015-09-03
Polyvox improvements

sethalves:
- PolyVox marching-cube entites can be stitched together so they make a continuous surface.
- PolyVox setCuboid allows setting a rectangular region.
- voxel.js supports setting / clearing "spheres" of voxels again.
- Normals on PolyVoxs are now more correct.
- Various PolyVox related bug-fixes.


hifi-gustavo:
Build console output for commit: f20ce12091e2d014cd55d041b759d86708da4f24 available at: https://gist.github.com/644223c0962234823f73<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5706.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5706.dmg

sethalves:
test this please

hifi-gustavo:
Build console output for commit: f20ce12091e2d014cd55d041b759d86708da4f24 available at: https://gist.github.com/644223c0962234823f73<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5706.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5706.dmg

examples/voxels.js
```diff
@@ -12,16 +12,19 @@ var toolWidth = 50;
...
+var onAlpha = 0.9;
+var editSphereRadius = 4;
 
 function floorVector(v) {
     return {x: Math.floor(v.x), y: Math.floor(v.y), z: Math.floor(v.z)};
 }
 
 function vectorToString(v){
-    return "{" + v.x + ", " + v.x + ", " + v.x + "}";
+    return "{" + v.x + ", " + v.y + ", " + v.z + "}";
```
jherico: there is now a `vec3toStr` method in libraries/utils.js that does the same thing as this, but takes an optional second parameter for how many digits of precisition to include:

```
vec3toStr = function (v, digits) {
    if (!digits) { digits = 3; }
    return "{ " + v.x.toFixed(digits) + ", " + v.y.toFixed(digits) + ", " + v.z.toFixed(digits)+ " }";
}
```

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
```diff
@@ -543,23 +587,21 @@ namespace render {
 
 
 glm::vec3 RenderablePolyVoxEntityItem::voxelCoordsToWorldCoords(glm::vec3& voxelCoords) const {
-    return glm::vec3(voxelToWorldMatrix() * glm::vec4(voxelCoords, 1.0f));
+    glm::vec3 adjustedCoords;
+    if (isEdged(_voxelSurfaceStyle)) {
+        adjustedCoords = voxelCoords + glm::vec3(0.5f, 0.5f, 0.5f);
```
jherico: `vec3(a)` is equivalent to `vec3(a, a, a)`

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
```diff
@@ -615,33 +658,25 @@ bool RenderablePolyVoxEntityItem::inUserBounds(const PolyVox::SimpleVolume<uint8
...
-
-        case PolyVoxEntityItem::SURFACE_EDGED_CUBIC:
-        case PolyVoxEntityItem::SURFACE_EDGED_MARCHING_CUBES:
-            if (x < 0 || y < 0 || z < 0 ||
-                x >= vol->getWidth() - 2 || y >= vol->getHeight() - 2 || z >= vol->getDepth() - 2) {
-                return false;
-            }
-            return true;
+    if (isEdged(surfaceStyle)) {
+        if (x < 0 || y < 0 || z < 0 ||
```
jherico: you might consider using vec3's here and the glm functions `all` `any` `lt` and `gt`

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
```diff
@@ -615,33 +658,25 @@ bool RenderablePolyVoxEntityItem::inUserBounds(const PolyVox::SimpleVolume<uint8
                                                PolyVoxEntityItem::PolyVoxSurfaceStyle surfaceStyle,
                                                int x, int y, int z) {
```
jherico: method should be const?

hifi-gustavo:
Build console output for commit: 07b241c778cceec658add5aaeb0e10846463ca27 available at: https://gist.github.com/644223c0962234823f73<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5706.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5706.dmg

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
```diff
@@ -402,21 +435,29 @@ bool RenderablePolyVoxEntityItem::isReadyToComputeShape() {
 }
 
 void RenderablePolyVoxEntityItem::computeShapeInfo(ShapeInfo& info) {
-    _shapeInfoLock.lockForRead();
+    QReadLocker(&this->_shapeInfoLock);
     info = _shapeInfo;
-    _shapeInfoLock.unlock();
 }
```
AndrewMeadows: Ooooh, I like the **QReadLocker** pattern.  I notice that it could be used in other places in this PR if you were so inclined.

sethalves: the other places release the lock before the end of the function.  maybe it could be done inside a block...

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
sethalves: does glm have an integer vec3?  I'd rather not be casting back and forth.

hifi-gustavo:
Build console output for commit: e12b7807ad8c2b8731cc6cefdc600ce33a3de896 available at: https://gist.github.com/644223c0962234823f73<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5706.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5706.dmg

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
```diff
@@ -304,67 +371,33 @@ bool RenderablePolyVoxEntityItem::findDetailedRayIntersection(const glm::vec3& o
...
-    bool ok = false;
-    for (int i = 0; i < 6; i ++) {
-        if (hit[ i ] && dx[ i ] < bestDx) {
-            face = (BoxFace)i;
-            distance = dx[ i ];
-            ok = true;
-            bestDx = dx[ i ];
-        }
-    }
+    glm::vec3 result3 = vec3(result);
```
AndrewMeadows: Not really a blocker, and I guess it compiles but: usually we've been using "glm::vec3(result)" rather than "vec3(result)".

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
```diff
@@ -850,23 +874,157 @@ void RenderablePolyVoxEntityItem::getMesh() {
...
+        if (currentYPNeighbor && currentYPNeighbor->getID() != _yPNeighborID) {
+            _yPNeighbor.reset();
+        }
+    }
+    if (_zPNeighborID != UNKNOWN_ENTITY_ID) {
+        EntityItemPointer currentZPNeighbor = _zPNeighbor.lock();
+        if (currentZPNeighbor && currentZPNeighbor->getID() != _zPNeighborID) {
+            _zPNeighbor.reset();
+        }
+    }
```
AndrewMeadows: There appears to be an unhandled **else** case here... when **currentZPNeighbor** evaluates to **false** --> the weak pointer is invalid and the _xxNeighborID should be cleared.  However, this might not actually matter for how the code is written.  That is, is it ok for _xxNeighborID to be a non-null ID corresponding to a non-existant (deleted) entity?

sethalves:
test this please

hifi-gustavo:
Build console output for commit: 1b92ae95d7428bc7e4655aea4acf71fbdb2a6cf4 available at: https://gist.github.com/644223c0962234823f73<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5706.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5706.dmg

hifi-gustavo:
Build console output for commit: 1b92ae95d7428bc7e4655aea4acf71fbdb2a6cf4 available at: https://gist.github.com/644223c0962234823f73<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5706.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5706.dmg

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
jherico: `ivec3`  the gt/lt/any/all functions will all accept any kind of vector `[(b|i|d)]vec(2|3|4)` and will return (for things like gt/lt) a `bvec(2|3|4)` while any/all just return a bool.

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
jherico: I concur with Andrew.  The advantage to this kind of pattern is that the lock still gets released in the event of early return (which is useful for locks that have a lot of code between the lock and unlock) or if an exception is thrown.  Explicit scoping it handy for using it when you want to unlock early.

In instances where you might want to break out of a long code block early in several different places you can create a scope of 

```
do { 
   QReadLocker locker(lock)
   // do stuff
   if (condition) {
      break;
   } 
} while (false)
```

PR #5707, opened 2015-09-03
added userData manipulations to grab script

ericrius1:
This PR adds logic to the hydraGrab script which sets a field called "activate" on the grabbed entities userData. This is useful because an entity with an attached entity script can now query for this field, and if it's set, can run whatever logic it wants. For example, a gun that is "grabbed" can now be fired, but if it is let go of, this is no longer the case.

This also sets the sessionUUID in the user data field, so an entity script can run a check and only execute code on the client that is actually holding the object.

### DIFF
--- a/examples/controllers/hydra/hydraGrab.js
+++ b/examples/controllers/hydra/hydraGrab.js
@@ -11,8 +11,12 @@
 //
 
 
+Script.include("https://hifi-public.s3.amazonaws.com/scripts/libraries/utils.js");
+
 var RIGHT_HAND_CLICK = Controller.findAction("RIGHT_HAND_CLICK");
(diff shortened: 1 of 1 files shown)


examples/controllers/hydra/hydraGrab.js
```diff
@@ -11,9 +11,13 @@
 //
 
 
+Script.include("https://hifi-public.s3.amazonaws.com/scripts/libraries/utils.js");
```
ZappoMan: you should make this relative to the script instead of hard coding the S3 reference.

ZappoMan: try this instead...
```
Script.include("../../libraries/utils.js");
```

PR #5709, opened 2015-09-03
next wave of breakdance work

ZappoMan:
* fixes a bug in omniTool that wasn't unloading modules on script shutdown
* moves the core behavior of breakdance game into a reusable library
* uses the breakdanceCore.js library in a standalone script called breakdanceToy.js
* uses the breakdanceCore.js library in an omniTool module
* adds some helper routines for finding sphere/point penetrations to util.js

### DIFF
--- a/examples/libraries/utils.js
+++ b/examples/libraries/utils.js
@@ -102,3 +102,28 @@ logInfo = function(str) {
 logDebug = function(str) {
     print(str);
 }
+
+// Computes the penetration between a point and a sphere (centered at the origin)
+// if point is inside sphere: returns true and stores the result in 'penetration'
+// (the vector that would move the point outside the sphere)
+// otherwise returns false
+findSphereHit = function(point, sphereRadius) {
+    var EPSILON = 0.000001;	//smallish positive number - used as margin of error for some computations
+    var vectorLength = Vec3.length(point);
+    if (vectorLength < EPSILON) {
+        return true;
+    }
+    var distance = vectorLength - sphereRadius;
+    if (distance < 0.0) {
+        return true;
+    }
+    return false;
+}
+
+findSpherePointHit = function(sphereCenter, sphereRadius, point) {
+    return findSphereHit(Vec3.subtract(point,sphereCenter), sphereRadius);
+}
+
+findSphereSphereHit = function(firstCenter, firstRadius, secondCenter, secondRadius) {
+    return findSpherePointHit(firstCenter, firstRadius + secondRadius, secondCenter);
+}
--- a/examples/toys/breakdanceCore.js
+++ b/examples/toys/breakdanceCore.js
@@ -1,7 +1,8 @@
 //
-// breakdanceToy.js
-// examples
+//  breakdanceCore.js
+//  examples/toys
 //
+//  This is the core breakdance game library, it can be used as part of an entity script, or an omniTool module, or bootstapped on it's own
 //  Created by Brad Hefta-Gaub on August 24, 2015
 //  Copyright 2015 High Fidelity, Inc.
 //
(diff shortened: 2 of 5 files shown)


examples/libraries/utils.js
```diff
@@ -101,4 +101,29 @@ logInfo = function(str) {
...
\ No newline at end of file
+}
+
+// Computes the penetration between a point and a sphere (centered at the origin)
+// if point is inside sphere: returns true and stores the result in 'penetration'
+// (the vector that would move the point outside the sphere)
+// otherwise returns false
+findSphereHit = function(point, sphereRadius) {
+    var EPSILON = 0.000001;	//smallish positive number - used as margin of error for some computations
+    var vectorLength = Vec3.length(point);
```
jherico: My vec3 PR includes a new function `withinEpsilon` which takes two vectors and determines if the length between them is less than epsilon.  However it does so by comparing the distance *squared* with epsilon *squared*, so no square root is required.  So you could do this by calling `Vec3.withinEpsilon(ZERO_VECTOR, point, EPSILON)`

However, this wouldn't help you with the vector length / sphere radius comparison.  However, if we added a Vec3::length2 that returned the length squared you could simplify this code (and compare against `sphereRadius * sphereRadius`) never incurring the hit of doing a square root.

examples/toys/breakdanceCore.js
```diff
@@ -9,34 +10,8 @@
...
-
-function findSpherePointHit(sphereCenter, sphereRadius, point) {
-    return findSphereHit(Vec3.subtract(point,sphereCenter), sphereRadius);
-}
-
-function findSphereSphereHit(firstCenter, firstRadius, secondCenter, secondRadius) {
-    return findSpherePointHit(firstCenter, firstRadius + secondRadius, secondCenter);
-}
 
+Script.include("../libraries/utils.js");
```
jherico: the omni tool includes this so it's technically redundant.  also bear in mind all paths in these omni tool scripts are relative to the omniTool.js loader, so this could just as easily be `Script.include("utils.js");`

examples/toys/breakdanceCore.js
```diff
@@ -245,13 +220,18 @@ function getPositionRightOnBase() {
...
-// We will also demonstrate some 3D overlays. We will create a couple of cubes, spheres, and lines
-// our 3D cube that moves around...
-var handSize = 0.25;
-var leftCubePosition = MyAvatar.getLeftPalmPosition();
-var rightCubePosition = MyAvatar.getRightPalmPosition();
+// some globals we will need access to
+var HAND_SIZE = 0.25;
+var TARGET_SIZE = 0.3;
+var TARGET_COLOR = { red: 128, green: 128, blue: 128};
+var TARGET_COLOR_HIT = { red: 0, green: 255, blue: 0};
```
jherico: these are equivalent to `"Gray"` and `"Lime"` (?!) now that overlays parse HTML colors.

examples/toys/breakdanceCore.js
```diff
@@ -536,47 +509,48 @@ Script.update.connect(function(deltaTime) {
...
     }
 
-    lastPoseValue = poseValue;
-
-    Overlays.editOverlay(leftOnBase,   { position: getPositionLeftOnBase(),   color: targetColorLeftOnBase   } );
-    Overlays.editOverlay(leftOverhead, { position: getPositionLeftOverhead(), color: targetColorLeftOverhead } );
-    Overlays.editOverlay(leftLowered,  { position: getPositionLeftLowered(),  color: targetColorLeftLowered  } );
-    Overlays.editOverlay(leftSide,     { position: getPositionLeftSide()    , color: targetColorLeftSide     } );
-    Overlays.editOverlay(leftFront,    { position: getPositionLeftFront()   , color: targetColorLeftFront    } );
+    Overlays.editOverlay(leftOnBaseOverlay,   { position: getPositionLeftOnBase(),   color: targetColorLeftOnBase   } );
```
jherico: I'm currently create a class that lets you create overlays relative to the current avatar position and orientation and automatically re-positions them as the avatar moves and turns, you could could drop this once that's added.

examples/toys/breakdanceCore.js
jherico: Ah I see you've got two places this can be loaded from, hence the need for the relative path.

examples/toys/breakdanceCore.js
ZappoMan: but you didn't add support in the C++ for color names... I'd need to include the JS with names... right?

jherico: No.  `QColor` [supports](http://doc.qt.io/qt-5/qcolor.html#setNamedColor) both HTML hex colors as well as the standard SVG color keyword names, which is what the htmlColor.js defines.  So you can use it either way.

PR #5711, opened 2015-09-03
fix a bug in entity script includes and file based includes on windows

ZappoMan:


libraries/script-engine/src/BatchLoader.cpp
```diff
@@ -58,11 +58,7 @@ void BatchLoader::start() {
             connect(this, &QObject::destroyed, reply, &QObject::deleteLater);
 
         } else {
-#ifdef _WIN32
-            QString fileName = url.toString();
```
ZappoMan: I'm not quite sure why this code was previously special cased... maybe it was an old version of Qt thing. But in 5.4 it definitely doesn't work to use the "toString" version of the URL and later use it with a QFile

libraries/shared/src/RegisteredMetaTypes.cpp
```diff
@@ -26,8 +26,7 @@ static int quatMetaTypeId = qRegisterMetaType<glm::quat>();
 static int xColorMetaTypeId = qRegisterMetaType<xColor>();
 static int pickRayMetaTypeId = qRegisterMetaType<PickRay>();
 static int collisionMetaTypeId = qRegisterMetaType<Collision>();
-
-
+static int qMapURLStringMetaTypeId = qRegisterMetaType<QMap<QUrl,QString>>();
```
ZappoMan: the metatype is required as the the underlying include code does cross thread marshaling of parameters of this type.

hifi-gustavo:
Build console output for commit: 063320771ca7d19b916f8ce373d4bf3f01c5d0fa available at: https://gist.github.com/be231d5e63983e450faa<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5711.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5711.dmg

ZappoMan:
this build might cause other problems with local files and includes... checking now

hifi-gustavo:
Build console output for commit: b0f239b89dca2799f7d3707caf3133877c56716a available at: https://gist.github.com/be231d5e63983e450faa<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5711.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5711.dmg

libraries/shared/src/RegisteredMetaTypes.cpp
SeijiEmery: wait, that's it? I figured out that line was missing a few weeks ago, but thought it was only a small part of the bug -_-

What API call is using QMap<QUrl, QString>, anyways?

PR #5713, opened 2015-09-04
fix collision shape of avatars

AndrewMeadows:
Yet another fix for avatar collision shapes -- fixes too-small collision shape for Mery.

We now know what the main real problem is: we're using the child offsets to compute the bone dimensions, however we really need to do is get build convex hulls of the **mesh** in the joint's frame, ignoring the **begin** and **end** points of the bone -- these offsets may be completely arbitrary relative to the mesh itself.  This is still a problem after this PR, but at least we know what needs to be done later when we want to actually build per-joint collision shapes.

### DIFF
--- a/interface/src/avatar/SkeletonModel.cpp
+++ b/interface/src/avatar/SkeletonModel.cpp
@@ -547,14 +547,11 @@ void SkeletonModel::computeBoundingShape() {
         // compute the default transform of this joint
         const JointState& state = _rig->getJointState(i);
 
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 16dfc39f5f7fb6351dba84b39feaaf97beb9cd48 available at: https://gist.github.com/946d78928fbd3dd71bad<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5713.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5713.dmg

PR #5714, opened 2015-09-04
Magballs polish

jherico:
* add omni tool the default scripts
* modify omni tool to only display the wand when there is an active module
* add UI to switch modes for the magballs... add and delete mode supported
* modify `web3d` overlays to support changing the URL after creation
* add AvatarRelativeOverlays helper class to allow 3D overlays to be created (and kept) at a position relative to the default eye position


hifi-gustavo:
Build console output for commit: c8069ba73ce3ee80cf63d96cf4c49dc4dde38b95 available at: https://gist.github.com/e9154daf0ad5b39eedbe<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5714.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5714.dmg

PR #5723, opened 2015-09-04
Toggle mute while holding the hydras

jherico:
Right now to toggle mute while in the HMD and using hydra navigation I have to either tilt my head to peer down my nose at my keyboard, or I have to remove my headset.  Since I'm using the hydras it looks very odd to anyone else.  I explicitly got called out about fiddling my headset in the alpha meeting.

As a quick path this PR adds a new toggle mute action to the user input mapper and binds it to the 0 button on the right controller, allow people in HMD/hand controllers to easily switch their mic on and off.    

Longer term I think we need to evaluate whether the SixenseManager is the right place to bind between controls and semantic meanings (I don't think it is...  the input plugins should expose abstracted spatial / joystick / keyboard mouse controllers as events and then the application / scripts should be free to modify the bindings to semantic meanings.  For instance, I also wanted to add an easy way to bring up the menu from the hydras, but there's no easy way for me to navigate the menu from there.  Rather some higher level like the application would need to be getting events like up / down / left / right from the controller and mapping it on to either avatar movement or to menu selection, depending on whether the menu was currently active.  However, such a change is outside the scope of this PR.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -612,6 +612,15 @@ Application::Application(int& argc, char** argv, QElapsedTimer &startup_time) :
     // Setup the userInputMapper with the actions
     auto userInputMapper = DependencyManager::get<UserInputMapper>();
     connect(userInputMapper.data(), &UserInputMapper::actionEvent, &_controllerScriptingInterface, &AbstractControllerScriptingInterface::actionEvent);
+    connect(userInputMapper.data(), &UserInputMapper::actionEvent, [this](int action, float state) {
+        if (state) {
+            switch (action) {
+            case UserInputMapper::Action::TOGGLE_MUTE:
+                DependencyManager::get<AudioClient>()->toggleMute();
+                break;
+            }
+        }
+    });
 
     // Setup the keyboardMouseDevice and the user input mapper with the default bindings
     _keyboardMouseDevice->registerToUserInputMapper(*userInputMapper);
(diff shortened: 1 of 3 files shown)


hifi-gustavo:
Build console output for commit: 8e28b33d44329be140fa29c2bd31c46c6d49e67f available at: https://gist.github.com/0986ba7829a615376084<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5723.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5723.dmg

ZappoMan:
"the input plugins should expose abstracted spatial / joystick / keyboard mouse controllers as events and then the application / scripts should be free to modify the bindings to semantic meanings" -- **YES YES a MILLION TIMES YES.**

hifi-gustavo:
Build console output for commit: 644e81e20b4c8b15cd53e7be0a9910668ac592a0 available at: https://gist.github.com/0986ba7829a615376084<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5723.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5723.dmg

imgntn:
+1 for customizable controller bindings. tested on mac, quick mute is nice.

PR #5724, opened 2015-09-04
AnimDebugDraw Improvements

hyperlogic:
* Now always works, regardless of whether or not Rig or AnimGraph animations
  are enabled.
* Changed joint radius to 1 cm.
* Changed xyz axis length to 4 cm.

Also, minor refactoring of AnimSkeleton constructor.

### DIFF
--- a/libraries/animation/src/AnimSkeleton.cpp
+++ b/libraries/animation/src/AnimSkeleton.cpp
@@ -87,11 +87,11 @@ int AnimSkeleton::getNumJoints() const {
     return _joints.size();
 }
 
-AnimPose AnimSkeleton::getAbsoluteBindPose(int jointIndex) const {
+const AnimPose& AnimSkeleton::getAbsoluteBindPose(int jointIndex) const {
     return _absoluteBindPoses[jointIndex];
 }
 
-AnimPose AnimSkeleton::getRelativeBindPose(int jointIndex) const {
+const AnimPose& AnimSkeleton::getRelativeBindPose(int jointIndex) const {
     return _relativeBindPoses[jointIndex];
 }
 
--- a/libraries/animation/src/AnimSkeleton.cpp
+++ b/libraries/animation/src/AnimSkeleton.cpp
@@ -58,7 +58,52 @@ AnimPose::operator glm::mat4() const {
                      glm::vec4(zAxis, 0.0f), glm::vec4(trans, 1.0f));
 }
 
+AnimSkeleton::AnimSkeleton(const FBXGeometry& fbxGeometry) {
+    // convert to std::vector of joints
+    std::vector<FBXJoint> joints;
+    joints.reserve(fbxGeometry.joints.size());
+    for (auto& joint : fbxGeometry.joints) {
+        joints.push_back(joint);
+    }
+
+    AnimPose geometryOffset(fbxGeometry.offset);
+    buildSkeletonFromJoints(joints, geometryOffset);
+}
+
 AnimSkeleton::AnimSkeleton(const std::vector<FBXJoint>& joints, const AnimPose& geometryOffset) {
+    buildSkeletonFromJoints(joints, geometryOffset);
+}
+
+int AnimSkeleton::nameToJointIndex(const QString& jointName) const {
+    for (size_t i = 0; i < _joints.size(); i++) {
+        if (_joints[i].name == jointName) {
+            return i;
+        }
+    }
+    return -1;
+}
+
+int AnimSkeleton::getNumJoints() const {
+    return _joints.size();
+}
+
+AnimPose AnimSkeleton::getAbsoluteBindPose(int jointIndex) const {
+    return _absoluteBindPoses[jointIndex];
+}
+
+AnimPose AnimSkeleton::getRelativeBindPose(int jointIndex) const {
+    return _relativeBindPoses[jointIndex];
+}
+
+int AnimSkeleton::getParentIndex(int jointIndex) const {
+    return _joints[jointIndex].parentIndex;
+}
+
+const QString& AnimSkeleton::getJointName(int jointIndex) const {
+    return _joints[jointIndex].name;
+}
+
+void AnimSkeleton::buildSkeletonFromJoints(const std::vector<FBXJoint>& joints, const AnimPose& geometryOffset) {
     _joints = joints;
 
     // build a cache of bind poses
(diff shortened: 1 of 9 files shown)


hifi-gustavo:
Build console output for commit: 69e463bbfdcfbacc2e185e6059d8e980be9c42a5 available at: https://gist.github.com/b6641f6a5cf3d78210b2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5724.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5724.dmg

hifi-gustavo:
Build console output for commit: bb5d061d782fbfc8db9c541f9a0176ad0b39927e available at: https://gist.github.com/b6641f6a5cf3d78210b2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5724.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5724.dmg

libraries/animation/src/AnimSkeleton.cpp
```diff
@@ -58,7 +58,52 @@ AnimPose::operator glm::mat4() const {
...
+
+int AnimSkeleton::getNumJoints() const {
+    return _joints.size();
+}
+
+AnimPose AnimSkeleton::getAbsoluteBindPose(int jointIndex) const {
+    return _absoluteBindPoses[jointIndex];
+}
+
+AnimPose AnimSkeleton::getRelativeBindPose(int jointIndex) const {
```
Atlante45: Could we return `const AnimPose&` here instead?

libraries/render-utils/src/AnimDebugDraw.cpp
```diff
@@ -350,6 +360,17 @@ void AnimDebugDraw::update() {
             }
         }
 
+        for (auto&& iter : _poses) {
```
Atlante45: why `auto&&`, you don't need an rvalue ref here.
`const auto&` is enough.

libraries/render-utils/src/AnimDebugDraw.cpp
```diff
@@ -418,6 +439,43 @@ void AnimDebugDraw::update() {
...
```
Atlante45: why `auto&&`, you don't need an rvalue ref here.
`const auto&` is enough.

libraries/render-utils/src/AnimDebugDraw.cpp
```diff
@@ -418,6 +439,43 @@ void AnimDebugDraw::update() {
...
+            if (hipsIndex >= 0) {
+                rootPose.trans -= skeleton->getRelativeBindPose(hipsIndex).trans;
+            }
+            glm::vec4 color = std::get<3>(iter.second);
+
+            std::vector<AnimPose> absAnimPose;
+            absAnimPose.resize(skeleton->getNumJoints());
+
+            for (int i = 0; i < skeleton->getNumJoints(); i++) {
+                AnimPose pose = poses[i];
```
Atlante45: You could take a ref here

libraries/render-utils/src/AnimDebugDraw.cpp
```diff
@@ -418,6 +439,43 @@ void AnimDebugDraw::update() {
...
+
+                const float radius = BONE_RADIUS / (pose.scale.x * rootPose.scale.x);
+
+                auto parentIndex = skeleton->getParentIndex(i);
+                if (parentIndex >= 0) {
+                    absAnimPose[i] = absAnimPose[parentIndex] * poses[i];
```
Atlante45: Use `pose`

libraries/render-utils/src/AnimDebugDraw.cpp
```diff
@@ -418,6 +439,43 @@ void AnimDebugDraw::update() {
...
+                } else {
+                    absAnimPose[i] = poses[i];
```
Atlante45: Use `pose`

libraries/render-utils/src/AnimDebugDraw.cpp
```diff
@@ -418,6 +439,43 @@ void AnimDebugDraw::update() {
...
+                }
+
+                // draw bone
+                addBone(rootPose, absAnimPose[i], radius, v);
+
+                // draw link to parent
+                if (parentIndex >= 0) {
+                    assert(parentIndex < skeleton->getNumJoints());
+                    AnimPose parentPose = poses[parentIndex];
```
Atlante45: Unused

libraries/animation/src/AnimSkeleton.cpp
hyperlogic: Good catch.  It would be more efficient to return a const ref here.

libraries/render-utils/src/AnimDebugDraw.cpp
hyperlogic: So, auto&& is a "universal reference" which could be an r-value or an l-value ref.  I'm toying with the idiom of using always using universal references in range base for loops.  Not sure how I feel about it though, it is confusing and it only has limited benefits.   I think I'll change this back to an auto& instead.

libraries/render-utils/src/AnimDebugDraw.cpp
hyperlogic: good catch

libraries/render-utils/src/AnimDebugDraw.cpp
hyperlogic: good catch.

hifi-gustavo:
Build console output for commit: 6d5927c0c9ddbc69780c49bf358a94d2877dcf1a available at: https://gist.github.com/b6641f6a5cf3d78210b2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5724.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5724.dmg

libraries/render-utils/src/AnimDebugDraw.cpp
ZappoMan: Food for thought - https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers

PR #5726, opened 2015-09-05
Javascript benchmarking

SeijiEmery:
Enables us to test existing scripts + the HF api for performance and optimization opportunities; this adds a small declarative benchmarking library / utility that makes writing these really easy.

### DIFF
--- a/examples/utilities/tests/allPerfTests.js
+++ b/examples/utilities/tests/allPerfTests.js
@@ -1,0 +1,11 @@
+//
+//  Created by Seiji Emery on 9/4/15
+//  Copyright 2015 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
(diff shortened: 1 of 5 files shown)


SeijiEmery:
entityPerfTest => discovered that: 
-- Entities.getEntityProperties() is quite slow
-- Entities have a hard limit of ~5000 api calls / sec (after which the thread running the script will spend all its time doing api calls and will start lagging). This is 5000 API calls TOTAL -- if you call getProperties 5 times for each entity and check it 60 frames / second, that's 300 api calls (or ~60 ms per entity). This limit is non-critical, but could be improved drastically w/ improvements to entity tree lookup (which could be O(1) but is not)

Converting to/from glm::vec3 is also an order of magnitude slower than a pure JS version (due to value translation between two languages; the same goes for quaternions + matrix operations), but it's still plenty fast for our purposes. Will shut up about reimplementing it in js now.

imgntn:
tested on mac (2996).  can verify some JS methods are faster.  interested re: hard limit -- are these things that we should formalize into recommendations for content creators?

PR #5735, opened 2015-09-08
Fix for RenderableParticleffectEntityItem assert.

hyperlogic:
When a particle system had no living particles, PendingChanges::updateItem
would assert, due to the access to _vertices[0].  In release this might be
harmless as this memory is never accessed, but different implementations of
std::vector might do different things here.  So, let's be safe and early
return when the number of particles is 0.


hifi-gustavo:
Build console output for commit: c728cd4c4c3fa026aa5af087ed5f105211816125 available at: https://gist.github.com/b88d285548246767b97b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5735.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5735.dmg

PR #5737, opened 2015-09-08
hydraGrab is now called handControllerGrab.js now grabs entities based on how they want to be grabbed (th…

ericrius1:
…rough setting of userData field with a key called spatialKey

the handControllerGrab script will look for two properties:
relativePosition and relativeRotation. If it finds either or both of these, it will correctly set the hold action fields to be synced with the data from the userData

### DIFF
--- a/examples/controllers/handControllerGrab.js
+++ b/examples/controllers/handControllerGrab.js
@@ -11,7 +11,8 @@
 //
 
 
-Script.include("../../libraries/utils.js");
+Script.include("../libraries/utils.js");
+
 
 var RIGHT_HAND_CLICK = Controller.findAction("RIGHT_HAND_CLICK");
 var rightTriggerAction = RIGHT_HAND_CLICK;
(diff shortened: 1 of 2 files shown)


examples/controllers/handControllerGrab.js
```diff
@@ -135,7 +136,9 @@ controller.prototype.checkForIntersections = function(origin, direction) {
 
     var intersection = Entities.findRayIntersection(pickRay, true);
     if (intersection.intersects && intersection.properties.collisionsWillMove === 1) {
-        this.distanceToEntity = Vec3.distance(origin, intersection.properties.position);
+        var handPosition = Controller.getSpatialControlPosition(this.palm);
+        this.distanceToEntity = Vec3.distance(handPosition, intersection.properties.position);
+        print("distance to entity " + JSON.stringify(this.distanceToEntity));
```
ZappoMan: did you mean to leave this debugging in? It looks like it will be spammy.

examples/controllers/handControllerGrab.js
```diff
@@ -188,6 +191,7 @@ controller.prototype.hidePointer = function() {
 
 controller.prototype.letGo = function() {
     if (this.grabbedEntity && this.actionID) {
+        print("DELETE ACTION")
```
ZappoMan: did you mean to leave this debugging in?

PR #5738, opened 2015-09-08
added an example script that detects being touched by avatars hand

ZappoMan:


imgntn:
Tested on Mac -- looks good to me.

PR #5739, opened 2015-09-08
CR for 20669 - Add additional particle properties controlling size, colour, and opacity

ctrlaltdavid:
For particle entities the following new proprties are added:
- radiusSpread
- radiusStart, radiusFinish
- colorSpread
- colorStart, colorFinish
- alpha
- alphaSpread
- alphaStart, alphaFinish

These are demonstrated in the particlesTest.js script in \examples\example\entities.


hifi-gustavo:
Build console output for commit: 2325cb8ee0700a3913d1bae67f9e22d0521f532a available at: https://gist.github.com/3529147180f99fa5d2fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5739.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5739.dmg

cozza13:
I just QA'ed this and it works well.

libraries/entities/src/ParticleEffectEntityItem.cpp
```diff
@@ -587,3 +760,42 @@ quint32 ParticleEffectEntityItem::getLivingParticleCount() const {
         return (_maxParticles - _particleHeadIndex) + _particleTailIndex;
     }
 }
+
+float ParticleEffectEntityItem::cubicInterpolate(float y0, float y1, float y2, float y3, float u) {
```
ZappoMan: these should be moved to shared utilities... is it possible glm already has this? Also, coding standard says these variable names need to be more descriptive.

libraries/entities/src/ParticleEffectEntityItem.cpp
```diff
@@ -587,3 +760,42 @@ quint32 ParticleEffectEntityItem::getLivingParticleCount() const {
...
+    uu = u * u;
+    a0 = y3 - y2 - y0 + y1;
+    a1 = y0 - y1 - a0;
+    a2 = y2 - y0;
+    a3 = y1;
+
+    return (a0 * u * uu + a1 * uu + a2 * u + a3);
+}
+
+float ParticleEffectEntityItem::interpolate(float start, float middle, float finish, float age) {
```
ZappoMan: these should be moved to shared utilities... is it possible glm already has this? Also, coding standard says these variable names need to be more descriptive.

libraries/networking/src/udt/PacketHeaders.h
```diff
@@ -145,5 +145,7 @@ const PacketVersion VERSION_ENTITIES_POLYLINE = 37;
 const PacketVersion VERSION_OCTREE_CENTERED_ORIGIN = 38;
 const PacketVersion VERSION_ENTITIES_PARTICLE_MODIFICATIONS = 39;
 const PacketVersion VERSION_ENTITIES_POLYVOX_NEIGHBORS = 40;
+const PacketVersion VERSION_ENTITIES_PARTICLE_RADIUS_PROPERTIES = 41;
+const PacketVersion VERSION_ENTITIES_PARTICLE_COLOR_PROPERTIES = 42;
```
ZappoMan: why did you add two versions? We don't need to go back and change this now (we can't) but why when you made one PR would you bump the protocol version twice?

PR #5740, opened 2015-09-08
add a boom box entity script

ZappoMan:


imgntn:
@ZappoMan 
- header file needs editing (from touched example)
- maybe bring song constant up to the top -->  or read from user data first, then fall back to this song
 this.song = SoundCache.getSound("http://hifi-public.s3.amazonaws.com/ryan/freaks7.wav")

PR #5746, opened 2015-09-09
Run cleanup when createWand.js ends; Don't delete bubbles on collision (for now)

imgntn:
overlays were getting stranded.  a future PR will have the debug overlays disabled by default.


hifi-gustavo:

ZappoMan:
add to whitelist

PR #5748, opened 2015-09-09
DO NOT MERGE: initial AnimInverseKinematics implementation

AndrewMeadows:
This introduces the AnimInverseKinematics class for perusal and perhaps merging -- it builds and enforces constraints (in theory) but needs to be hooked up so we can see what it **actually** does, and then verify that the constraints do what we expect.

### DIFF
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -26,17 +26,6 @@ AnimInverseKinematics::~AnimInverseKinematics() {
 void AnimInverseKinematics::loadDefaultPoses(const AnimPoseVec& poses) {
     _defaultRelativePoses = poses;
     assert(_skeleton && _skeleton->getNumJoints() == (int)poses.size());
-    /*
-    // BUG: sometimes poses are in centimeters but we expect meters.
-    // HACK WORKAROUND: check for very large translation of hips and scale down as necessary
-    int hipsIndex = _skeleton->nameToJointIndex("Hips");
-    if (hipsIndex > -1 && 
-            glm::length(_defaultRelativePoses[hipsIndex].trans) > 10.0f && 
-            glm::length(_defaultRelativePoses[hipsIndex].scale) > 0.1f) {
-        _defaultRelativePoses[hipsIndex].scale = glm::vec3(0.01f);
-        _defaultRelativePoses[hipsIndex].trans *= 0.01f;
-    }
-    */
 }
 
 void AnimInverseKinematics::loadPoses(const AnimPoseVec& poses) {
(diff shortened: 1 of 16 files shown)


libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -0,0 +1,566 @@
...
+        delete constraintItr->second;
+        ++constraintItr;
+    }
+    _constraints.clear();
+}
+
+const glm::vec3 xAxis(1.0f, 0.0f, 0.0f);
+const glm::vec3 yAxis(0.0f, 1.0f, 0.0f);
+const glm::vec3 zAxis(0.0f, 0.0f, 1.0f);
+
```
AndrewMeadows: The code below here is some hard-coded constraints for the shoulder, elbow, upper legs, and knees.  For testing.

libraries/animation/src/SwingTwistConstraint.cpp
```diff
@@ -69,6 +75,85 @@ void SwingTwistConstraint::setSwingLimits(std::vector<float> minDots) {
...
+                if (leftTheta > theta) {
+                    leftTheta -= TWO_PI;
+                } else {
+                    rightTheta += TWO_PI;
+                }
+            }
+
+            // blend between the left and right minDots to get the value that corresponds to this theta
+            float rightWeight = (theta - leftTheta) / (rightTheta - leftTheta);
+            minDots.push_back((1.0f - rightWeight) * limits[leftIndex]._minDot + rightWeight * limits[rightIndex]._minDot);
```
AndrewMeadows: I'm using a linear blend here between discrete function values.  Ozan expressed concern that maybe we need a curved fit rather than a linear fit.  If the linear fit proves a problem (sudden transitions at clamped boundary) then we can fix it later.

hifi-gustavo:
Build console output for commit: e86e7601188b89403a7ad5d0562144753d7c80a0 available at: https://gist.github.com/12c66371931f5f415e95<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5748.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5748.dmg

tests/animation/src/AnimInverseKinematicsTests.cpp
```diff
@@ -0,0 +1,269 @@
...
+    glm::quat rot = identity;
+    glm::vec3 scale = glm::vec3(1.0f);
+
+    AnimPose poseA(scale, rot, transA);
+    AnimPose poseB(scale, rot, transB);
+    AnimPose poseC = poseA * poseB;
+
+    glm::vec3 expectedTransC = transA + transB;
+    QCOMPARE_WITH_ABS_ERROR(expectedTransC, poseC.trans, EPSILON);
+}
```
AndrewMeadows: **testBar()** is a stub method for a test I was thinking about writing.  I'll definitely be adding more tests and will eventually clean this up.

ZappoMan:
How do you recommend we test this? And or play with it to provide feedback? Is it wired into the avatar? Can we test it just by picking up hydras and seeing what it does?

AndrewMeadows:
No, this is not testable yet.  I put it up so that **hyperlogic** could look it over.  We're probably going to close this PR and use a different path: merge this with some work that he's doing and then submit a new PR.

AndrewMeadows:
Closing this PR.  hyperlogic will pull this into his own branch, fix some things, and submit a new one.

PR #5753, opened 2015-09-09
Terminate DDE earlier during shut-down and always at start-up

ctrlaltdavid:
To reduce the instances of the Installer failing because DDE is left running from a crash
 - Terminate DDE earlier on in the Interface shutdown process (i.e., before any shutdown crashes if possible).
 - Terminate DDE at Interface start-up even if DDE isn't enabled (e.g., user may have selected one of the post -crash options at start-up to wipe settings).

### DIFF
--- a/interface/src/devices/DdeFaceTracker.cpp
+++ b/interface/src/devices/DdeFaceTracker.cpp
@@ -254,9 +254,8 @@ void DdeFaceTracker::setEnabled(bool enabled) {
         _ddeProcess = new QProcess(qApp);
         connect(_ddeProcess, SIGNAL(finished(int, QProcess::ExitStatus)), SLOT(processFinished(int, QProcess::ExitStatus)));
         _ddeProcess->start(QCoreApplication::applicationDirPath() + DDE_PROGRAM_PATH, DDE_ARGUMENTS);
-    }
-
-    if (!enabled && _ddeProcess) {
+    } else {
+        // Send a command to stop face tracker evening if no _ddeProcess in case DDE has been left running after a crash
         _ddeStopping = true;
         _udpSocket.writeDatagram(DDE_EXIT_COMMAND, DDE_SERVER_ADDR, _controlPort);
         qCDebug(interfaceapp) << "DDE Face Tracker: Stopping";
--- a/interface/src/devices/DdeFaceTracker.cpp
+++ b/interface/src/devices/DdeFaceTracker.cpp
@@ -236,28 +236,25 @@ void DdeFaceTracker::setEnabled(bool enabled) {
         cancelCalibration();
     }
 
-
     // isOpen() does not work as one might expect on QUdpSocket; don't test isOpen() before closing socket.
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 01d7e68dbb8d7689c0d31e76e9ea57d0604f7fdc available at: https://gist.github.com/e565ba0e543efa222f54<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5753.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5753.dmg

ctrlaltdavid:
Tested, ready for CR.

interface/src/devices/DdeFaceTracker.cpp
```diff
@@ -254,9 +254,8 @@ void DdeFaceTracker::setEnabled(bool enabled) {
...
         connect(_ddeProcess, SIGNAL(finished(int, QProcess::ExitStatus)), SLOT(processFinished(int, QProcess::ExitStatus)));
         _ddeProcess->start(QCoreApplication::applicationDirPath() + DDE_PROGRAM_PATH, DDE_ARGUMENTS);
-    }
-
-    if (!enabled && _ddeProcess) {
+    } else {
+        // Send a command to stop face tracker evening if no _ddeProcess in case DDE has been left running after a crash
         _ddeStopping = true;
         _udpSocket.writeDatagram(DDE_EXIT_COMMAND, DDE_SERVER_ADDR, _controlPort);
         qCDebug(interfaceapp) << "DDE Face Tracker: Stopping";
```
AndrewMeadows: Note: the logic flow has changed here since **!(!enabled && _ddeProcess)** is actually **(!enabled || _ddeProcess)** which is not the logic in the **if** statement higher up.

The new effective logic that makes it fall into the **else** case is **!(enabled && !_ddeProcess)** which is equivalent to **(!enabled || _ddeProcess)**.  It isn't clear to me that this is really what you want... it sets **_ddeStopping** to **true**...  when it wasn't enabled.  Is this ok?

If the above note does not require any changes then I note that this line is in both the **if** and **else** blocks, so maybe it should be pulled out and put above:
    _udpSocket.writeDatagram(DDE_EXIT_COMMAND, DDE_SERVER_ADDR, _controlPort);

ctrlaltdavid: Good point. I'll reinstate the logic and pull out the exit command to make things clearer.

hifi-gustavo:
Build console output for commit: fdcfa9657c2523333d63264c18b072a79ca8a52a available at: https://gist.github.com/e565ba0e543efa222f54<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5753.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5753.dmg

PR #5756, opened 2015-09-09
added lifetime to pointer, default grab behavior now honors objects i…

ericrius1:
…initial position and rotation

If you wish your entity to exhibit a different behavior on grab, you can set relativePosition and relativeRotation on the grabFrame key of the userData.

### DIFF
--- a/examples/controllers/handControllerGrab.js
+++ b/examples/controllers/handControllerGrab.js
@@ -11,8 +11,7 @@
 //
 
 
-Script.include("../libraries/utils.js");
-
+Script.include("https://hifi-public.s3.amazonaws.com/scripts/libraries/utils.js");
 
 var RIGHT_HAND_CLICK = Controller.findAction("RIGHT_HAND_CLICK");
 var rightTriggerAction = RIGHT_HAND_CLICK;
@@ -21,6 +20,10 @@ var GRAB_USER_DATA_KEY = "grabKey";
 
 var LEFT_HAND_CLICK = Controller.findAction("LEFT_HAND_CLICK");
 var leftTriggerAction = LEFT_HAND_CLICK;
+
+var LIFETIME = 10;
+var currentLife = 0;
+var POINTER_CHECK_TIME = 5000;
 
 var ZERO_VEC = {
     x: 0,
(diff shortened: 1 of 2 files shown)


examples/controllers/handControllerGrab.js
```diff
@@ -11,8 +11,7 @@
 //
 
 
-Script.include("../libraries/utils.js");
-
+Script.include("https://hifi-public.s3.amazonaws.com/scripts/libraries/utils.js");
```
ZappoMan: why did you change this? this really should be relative like it previously was... please change this back.

examples/controllers/handControllerGrab.js
```diff
@@ -127,6 +139,18 @@ controller.prototype.updateLine = function() {
 }
 
 
+controller.prototype.checkPointer = function() {
+    var self = this;
+    Script.setTimeout(function() {
+        var props = Entities.getEntityProperties(self.pointer);
+        var currentLife = LIFETIME + POINTER_CHECK_TIME + currentLife;
```
ZappoMan: this doesn't make sense... currentLife is undefined when you execute this... did you mean props.age??

ZappoMan:
Eric,

I have a couple of pieces of general feedback about this script in general, I really think you should fix these issues... as you have this script written now, it's very inefficient and is sending way more updates than it needs to.

* Fix **controller.prototype.updateLine** to only make one call to Entities.editEntity() as it is written now this function makes multiple calls to editEntity() for the same entityID. this increases the network traffic to the server and all the viewers. Instead the logic should determine the new properties and only make one call to editEntity()

* Fix **controller.prototype.update** to make only one call to Entities.editEntity() the same issue that exists with updateLine is even more pronounced in update.. basically you call several different helper functions which each make multiple calls to editEntity(). Every call to editEntity() creates a full roundtrip to the server and all the viewers. As it's written now we're increasing the traffic to the server and back to clients by at least 10x.

* Fix **controller.prototype.checkPointer** to not use a timer, and make extra calls to editEntities() when it could just adjust the lifetime in the other calls to editEntity. Like the other issues mentioned above, these extra calls to editEntity() are causing way to much traffic. In this case, you only need to extend the lifetime during any one of your other editEntity() calls.

PR #5760, opened 2015-09-10
fix degenerative url scheme case

ZappoMan:


libraries/networking/src/ResourceManager.cpp
```diff
@@ -25,6 +25,15 @@ ResourceRequest* ResourceManager::createResourceRequest(QObject* parent, const Q
         return new HTTPResourceRequest(parent, url);
     } else if (scheme == URL_SCHEME_ATP) {
         return new AssetResourceRequest(parent, url);
+    } else {
+        // check the degenerative file case: on windows we can often have urls of the form c:/filename
+        // this checks for and works around that case.
+        QString urlWithFileSchemeName = URL_SCHEME_FILE + ":///" + url.toString();
+        QUrl urlWithFileScheme = urlWithFileSchemeName;
```
birarda: `QUrl urlWithFileScheme { urlWithFileSchemeName };`

birarda: That's an explicit call to the constructor that takes a QString instead of calling the assignment with a QString

libraries/networking/src/ResourceManager.cpp
```diff
@@ -25,6 +25,15 @@ ResourceRequest* ResourceManager::createResourceRequest(QObject* parent, const Q
...
+        QString fileName = urlWithFileScheme.toLocalFile();
+        if (!fileName.isEmpty()) {
```
birarda: if this fileName is never empty I suppose that means we can remove the guard in `FileResourceRequest`

birarda: and since fileName is not used I would think this should be `!urlWithFileScheme.toLocalFile().isEmpty()`

libraries/networking/src/ResourceManager.cpp
ZappoMan: funny - I had that and changed it back

libraries/networking/src/ResourceManager.cpp
ZappoMan: let's not change FileResourceRequest since other callers could still use that directly

hifi-gustavo:
Build console output for commit: d62945c6bf72ed8483a4f238937ea50824bdc78c available at: https://gist.github.com/b058af7a22a17bdc5861<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5760.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5760.dmg

hifi-gustavo:
Build console output for commit: f710afbfd1a3ed6bca47f90aa79972cbb0e8151c available at: https://gist.github.com/b058af7a22a17bdc5861<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5760.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5760.dmg

hifi-gustavo:
Build console output for commit: 626fcb867b88cf8e102eb78aa0354209e87375a4 available at: https://gist.github.com/b058af7a22a17bdc5861<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5760.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5760.dmg

PR #5762, opened 2015-09-10
Adding the flashlight toy script

samcake:
- flashlight.js is an entity script that can be assigned to the flashlight model https://hifi-public.s3.amazonaws.com/models/props/flashlight.fbx. it will create and manage a spotlight entity while being grabbed by the hand controller

- createFlashlight.js is a simple client side script to run and create a flashlight entity and the script together, when leaving the domain it will be deleted.

### DIFF
--- a/examples/toys/flashlight/flashlight.js
+++ b/examples/toys/flashlight/flashlight.js
@@ -19,7 +19,7 @@
       //  print(message);
     }
 
-    Script.include("https://hifi-public.s3.amazonaws.com/scripts/libraries/utils.js");
+    Script.include("../libraries/utils.js");
 
     var _this;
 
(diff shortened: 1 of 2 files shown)


examples/toys/flashlight/flashlight.js
```diff
@@ -0,0 +1,154 @@
...
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+(function() {
+    
+    function debugPrint(message) {
+      //  print(message);
+    }
+
+    Script.include("https://hifi-public.s3.amazonaws.com/scripts/libraries/utils.js");
```
ZappoMan: this should be done as a relative path

PR #5768, opened 2015-09-10
AnimInverseKinematics and AnimController node

hyperlogic:
* When AnimGraph is enabled the root IK node is ALWAYS enabled.  Even when there are no hand controllers.
* The IK solution is full body, no constraints are in play.  This results in very flexible spine and elbow joints.


hifi-gustavo:
Build console output for commit: 3149baeefc65a03fc048a86c746e3a72c95e6776 available at: https://gist.github.com/77607a47c6403d3b7479<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5768.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5768.dmg

tests/animation/src/AnimInverseKinematicsTests.cpp
```diff
@@ -0,0 +1,269 @@
+//
+//  AnimInverseKinematicsTests.cpp
```
howard-stearns: These don't compile yet.

hifi-gustavo:
Build console output for commit: 9a9838fd0d97cdf349399f9c7d9849aa595d5729 available at: https://gist.github.com/77607a47c6403d3b7479<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5768.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5768.dmg

libraries/animation/src/AnimationCache.h
```diff
@@ -12,9 +12,9 @@
 #ifndef hifi_AnimationCache_h
 #define hifi_AnimationCache_h
 
-#include <QRunnable>
-#include <QScriptEngine>
-#include <QScriptValue>
+#include <QtCore/QRunnable>
```
howard-stearns: Why these changes (and for ResourceCache.h)? Does this PR actually need changes to resource headers? Merging issue? gcc compiler fixes?

libraries/networking/src/ResourceCache.h
```diff
@@ -12,17 +12,17 @@
...
-#include <QNetworkReply>
-#include <QNetworkRequest>
-#include <QObject>
-#include <QPointer>
-#include <QSharedPointer>
-#include <QUrl>
-#include <QWeakPointer>
-#include <QReadWriteLock>
-#include <QQueue>
+#include <QtCore/QHash>
```
howard-stearns: See comments for AnimationCache.h.

howard-stearns:
Great integration! It doesn't look like anything broke. I've checked the cartesian product of:
- with and without rig animation
- with and without avatar update thread
- mac and windows
- hydra
Plus HMD on mac only, in both standing and sitting mode. (The standing HMD was a bit wonky, but I think it was before.)
Plus no crashes and expected behavior for anim-graph in main thread. (Might be a separate thread anim-graph conflict that I'll have to poke at, but I wouldn't let that stop a merge.)

libraries/animation/src/AnimationCache.h
hyperlogic: These changes were made by @AndrewMeadows, not sure perhaps they were made for Linux?

tests/animation/src/AnimInverseKinematicsTests.cpp
hyperlogic: Good catch, I haven't updated the tests.

PR #5775, opened 2015-09-10
Paint can

ericrius1:
Spray Paint can now resets back onto table if user drops it

sprayPaintSpawner.js spawns a sprayPaint can in correct location on table

### DIFF
--- a/examples/entityScripts/sprayPaintCan.js
+++ b/examples/entityScripts/sprayPaintCan.js
@@ -25,6 +25,10 @@
         running: true
     });
 
+    var totalTime = 0;
+    var RESET_TIME_THRESHOLD = 2;
+    var RESET_POSITION = {x: 549.12, y: 495.555, z: 503.77};
+
     this.getUserData = function() {
 
 
@@ -40,8 +44,21 @@
     }
 
     this.update = function(deltaTime) {
-        self.properties = Entities.getEntityProperties(self.entityId);
         self.getUserData();
+        self.properties = Entities.getEntityProperties(self.entityId);
+        totalTime += deltaTime;
+        if(totalTime > RESET_TIME_THRESHOLD) {
+            if(!self.activated) {
+                print("RESET");
+                Entities.editEntity(self.entityId, {
+                    position: RESET_POSITION,
+                    rotation: Quat.fromPitchYawRollDegrees(0, 0, 0),
+                    angularVelocity: ZERO_VEC,
+                    velocity: ZERO_VEC
+                });
+            }
+            totalTime = 0;
+        }
         if (self.userData.grabKey && self.userData.grabKey.activated === true) {
             if (self.activated !== true) {
                 Entities.editEntity(self.paintStream, {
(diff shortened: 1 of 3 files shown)


examples/entityScripts/sprayPaintCan.js
```diff
@@ -1,5 +1,7 @@
 (function() {
-    Script.include("../libraries/utils.js");
+    // Script.include("../libraries/utils.js");
+    //Need absolute path for now, for testing before PR merge and s3 cloning. Will change post-merge
+    Script.include("https://hifi-public.s3.amazonaws.com/scripts/libraries/utils.js");
```
sethalves: why not a relative url, here?

ericrius1: Using absolute paths is necessary for testing until a merge occurs- once that does, then the scripts are copied into S3, and I can change it

examples/entityScripts/sprayPaintCan.js
```diff
@@ -25,6 +27,14 @@
         running: true
     });
 
+    var timeSinceLastMoved = 0;
+    var RESET_TIME_THRESHOLD = 5;
+    var DISTANCE_FROM_HOME_THRESHOLD = .5;
```
sethalves: zero before the decimal

https://docs.highfidelity.com/docs/coding-standard#section-3-5-2-floating-point-constants-should-always-be-written-with-decimal-point-and-at-least-one-decimal-

(I had to look it up)

examples/sprayPaintSpawner.js
```diff
@@ -8,67 +8,33 @@
 //  Distributed under the Apache License, Version 2.0.
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 
-var scriptURL = "entityScripts/sprayPaintCan.js";
+//Just temporarily using my own bucket here so others can test the entity. Once PR is tested and merged, then the entity script will appear in its proper place in S3, and I wil switch it
+var scriptURL = "https://hifi-public.s3.amazonaws.com/eric/scripts/sprayPaintCan.js?=v1" + Math.random();
```
sethalves: relative url... the random number will break http caching, which I suppose is the point, but let's remove that bit now that testing is done.

examples/sprayPaintSpawner.js
```diff
@@ -8,67 +8,33 @@
...
+// var scriptURL = "https://hifi-public.s3.amazonaws.com/scripts/entityScripts/sprayPaintCan.js";
+// var scriptURL = "file:////Users/ericlevin/hifi/examples/entityScripts/sprayPaintCan.js?=v1" + Math.random();
```
sethalves: left-over commented out lines

examples/entityScripts/sprayPaintCan.js
sethalves: check out the top of edit.js for a nice way to handle it

PR #5777, opened 2015-09-11
Revert "edit.js adjustment"

sethalves:
Reverts highfidelity/hifi#5769


sethalves:
I'm not actually sure if this should be reverted -- after #5769 went in I had a strange edit.js experience in sandbox, and haven't had time to determine if this was the cause.

PR #5778, opened 2015-09-11
Fixing some entity race condition crashes

jherico:
This removes much of the explicit use of QReadWriteLock, using lambda functions instead to mediate the creation and releasing of locks.  

Additionally it provides some additional locking by ensuring that the locks used are recursive.

Ultimately I'd like to remove the need for explicit locking by pushing the locks into the actual functions that modify or read the object.


hifi-gustavo:
Build console output for commit: 4d96be6a55570d46efc2cc32843ae7e6c7f742f2 available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

assignment-client/src/octree/OctreeInboundPacketProcessor.cpp
```diff
@@ -156,20 +156,20 @@ void OctreeInboundPacketProcessor::processPacket(QSharedPointer<NLPacket> packet
                         packet->pos(), maxSize);
             }
 
-            quint64 startLock = usecTimestampNow();
-            _myServer->getOctree()->lockForWrite();
-            quint64 startProcess = usecTimestampNow();
-            int editDataBytesRead =
-                _myServer->getOctree()->processEditPacketData(*packet, editData, maxSize, sendingNode);
+            quint64 startProcess, startLock = usecTimestampNow();
```
ZappoMan: this change breaks the timing of how long it takes to get the lock. There are server stats visible in the server admin page that shows how long the server is waiting on tree locks vs processing inbound packets. I don't want to lose that important debug info.

jherico: breaks it how?  while I'm declaring startProcess here (I have to because if I declare inside the lock scope it won't be visible to the code that actually calculates the wait time) I don't initialize it.  I initialize it in exactly the same place, the first call after the lock has been acquired.

ZappoMan: ok wait - I was confused by the diff... I see now that the startProcess is being set inside.

interface/src/Application.cpp
```diff
@@ -3764,13 +3751,13 @@ void Application::clearDomainOctreeDetails() {
     // reset the environment so that we don't erroneously end up with multiple
 
     // reset our node to stats and node to jurisdiction maps... since these must be changing...
-    _entityServerJurisdictions.lockForWrite();
-    _entityServerJurisdictions.clear();
-    _entityServerJurisdictions.unlock();
+    _entityServerJurisdictions.withWriteLock([&] {
```
ZappoMan: why not just put this mutex inside the _entityServerJurisdictions object?

interface/src/Application.cpp
```diff
@@ -3764,13 +3751,13 @@ void Application::clearDomainOctreeDetails() {
...
+        _entityServerJurisdictions.clear();
+    });
 
-    _octreeSceneStatsLock.lockForWrite();
-    _octreeServerSceneStats.clear();
-    _octreeSceneStatsLock.unlock();
+    _octreeServerSceneStats.withWriteLock([&] {
```
ZappoMan: why not just put this mutex inside the _octreeServerSceneStats object?

interface/src/Application.cpp
```diff
@@ -3899,43 +3887,42 @@ int Application::processOctreeStats(NLPacket& packet, SharedNodePointer sendingN
     const QUuid& nodeUUID = sendingNode->getUUID();
     
     // now that we know the node ID, let's add these stats to the stats for that node...
-    _octreeSceneStatsLock.lockForWrite();
-    
-    OctreeSceneStats& octreeStats = _octreeServerSceneStats[nodeUUID];
-    statsMessageLength = octreeStats.unpackFromPacket(packet);
-    
-    _octreeSceneStatsLock.unlock();
+    _octreeServerSceneStats.withWriteLock([&] {
```
jherico: This change deserves a bit of explanation.  Previously the `OctreeSceneStats& octreeStats` value was initialized inside a lock, but then used in a number of spots outside the lock.  This is not valid behavior, since as soon as you leave the lock some other thread is free to modify the map and destroy the validity of your reference.  As such the write lock scope here is now significantly bigger, enclosing all use of the reference.

hifi-gustavo:
Build console output for commit: 41e4a6be2ba0f20e621e3135f21d498a1e4a90c4 available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

interface/src/Application.cpp
jherico: Because `_entityServerJurisdictions` is a `NodeToJurisdictionMap`, which in turn is derived from `QMap<QUuid, JurisdictionMap>`

there's no way to embed a lock into the QMap cleanly.  I could make the parent QMap private, and then expose all the map functions we actually use as public functions on the new class, but that would be a ton of work with not a lot of applicability.

interface/src/Application.cpp
jherico: Similar to above.  `NodeToOctreeSceneStats` is a class derived `std::map`.  There's no way to make the parent class operations inherently thread safe without hiding them and re-writing them.

interface/src/avatar/AvatarActionHold.cpp
```diff
@@ -44,43 +44,42 @@ void AvatarActionHold::updateActionWorker(float deltaTimeStep) {
...
-    }
-
-    auto rotation = palmRotation * _relativeRotation;
-    auto offset = rotation * _relativePosition;
-    auto position = palmPosition + offset;
-    unlock();
+    glm::quat rotation;
+    glm::vec3 position;
+    glm::vec3 offset;
+    bool result = withTryReadLock([&]{
```
ZappoMan: Not a merge blocker, but please fix -- I think this variable could be way better named, something like "gotLock" or "lockWasSuccesful" would make the code easier to read.

libraries/entities/src/EntityItem.cpp
```diff
@@ -1543,33 +1544,33 @@ bool EntityItem::addActionInternal(EntitySimulation* simulation, EntityActionPoi
...
     return success;
 }
 
 bool EntityItem::removeAction(EntitySimulation* simulation, const QUuid& actionID) {
-    lockForWrite();
-    checkWaitingToRemove(simulation);
-
-    bool success = removeActionInternal(actionID);
-    unlock();
+    bool success;
```
ZappoMan: you're not initializing this... don't you need `bool success = false;`

libraries/shared/src/shared/ReadWriteLockable.h
```diff
@@ -0,0 +1,71 @@
...
+        f();
+        return true;
+    }
+
+    template <typename F>
+    bool withTryReadLock(F f) const {
+        return withReadLock(f, false);
+    }
+
+//    void lockForRead() const;
```
ZappoMan: do you want this code in here? or get rid of it?

libraries/shared/src/shared/ReadWriteLockable.cpp
```diff
@@ -0,0 +1,93 @@
...
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include "ReadWriteLockable.h"
+
+#include <QtCore/QThread>
+//
+//bool ReadWriteLockable::isLocked() const {
```
ZappoMan: do we need any of this code?

libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -121,8 +121,9 @@ void PhysicalEntitySimulation::clearEntitiesInternal() {
 // end EntitySimulation overrides
 
 
-VectorOfMotionStates& PhysicalEntitySimulation::getObjectsToDelete() {
-    _tempVector.clear();
+void PhysicalEntitySimulation::getObjectsToDelete(VectorOfMotionStates& result) {
+    result.clear();
```
ZappoMan: why do you have the result.clear() outside of the lock? It probably doesn't matter, but this order seems unusual to me.

libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -244,29 +246,28 @@ void PhysicalEntitySimulation::handleCollisionEvents(CollisionEvents& collisionE
 
 void PhysicalEntitySimulation::addAction(EntityActionPointer action) {
     if (_physicsEngine) {
-        lock();
+        // FIXME put fine grain locking into _physicsEngine
+        QMutexLocker lock(&_mutex);
```
ZappoMan: this change is definitely changing the scope of a lock... was this intentional? The old version of the code would unlock the PhysicalEntitySimulation before calling EntitySimulation::addAction() - this new version keeps like lock on PhysicalEntitySimulation while calling EntitySimulation::addAction() -- are we sure this change won't introduce deadlock?

libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -244,29 +246,28 @@ void PhysicalEntitySimulation::handleCollisionEvents(CollisionEvents& collisionE
...
 void PhysicalEntitySimulation::applyActionChanges() {
     if (_physicsEngine) {
-        lock();
-        foreach (QUuid actionToRemove, _actionsToRemove) {
+        // FIXME put fine grain locking into _physicsEngine
+        QMutexLocker lock(&_mutex);
+        foreach(QUuid actionToRemove, _actionsToRemove) {
             _physicsEngine->removeAction(actionToRemove);
         }
-        _actionsToRemove.clear();
```
ZappoMan: did you mean to remove this? It looks like the _actionsToRemove set runs the risk of not getting cleared, so we might end up repeatedly calling _physicsEngine->removeAction() on the same action.

libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -244,29 +246,28 @@ void PhysicalEntitySimulation::handleCollisionEvents(CollisionEvents& collisionE
...
         foreach (EntityActionPointer actionToAdd, _actionsToAdd) {
             if (!_actionsToRemove.contains(actionToAdd->getID())) {
                 _physicsEngine->addAction(actionToAdd);
             }
         }
-        _actionsToAdd.clear();
```
ZappoMan: did you mean to remove this? It looks like the _actionsToAdd set runs the risk of not getting cleared, so we might end up repeatedly calling _physicsEngine->addAction() on the same action.

interface/src/Application.cpp
ZappoMan: ok

interface/src/Application.cpp
ZappoMan: ok

ZappoMan:
I crashed when I ran this script - https://raw.githubusercontent.com/jherico/hifi/homer/examples/tests/entityEditStressTest.js

interface/src/avatar/AvatarActionHold.cpp
jherico: fixed

examples/tests/timerStressTest.js
```diff
@@ -0,0 +1,68 @@
...
+
+var boxes=[];
+var TEST_ENTITY_NAME = "TimerScript";
+
+
+var TOTAL_ENTITIES = 100;
+for (var i = 0; i < TOTAL_ENTITIES; i++) {
+    var box = Entities.addEntity({
+        type: "Box",
+        name: ""
```
ZappoMan: this script has a syntax error right here... should be a comma after name.

```
[DEBUG] [09/10 20:30:54] Uncaught exception at ( "https://raw.githubusercontent.com/jherico/hifi/homer/examples/tests/timerStressTest.js" : "" ) line 35 : "SyntaxError: Parse error"
```

libraries/entities/src/EntityItem.cpp
jherico: technically I don't need to initialize it since there's no code path where it doesn't pass through the body of the lock lambda.  However, I will go ahead and init it to false for style sake.

libraries/shared/src/shared/ReadWriteLockable.h
jherico: It was there initially so that I could retain the same interface as the previous code, but have it centralized.  However, I decided it was better if there was only one consistent mechanism for lock management.  Will remove.

libraries/shared/src/shared/ReadWriteLockable.cpp
jherico: Will remove.

libraries/physics/src/PhysicalEntitySimulation.cpp
jherico: result is passed in by the caller.  if it's a structure being used in multiple threads it's definitely the callers responsibility to manage access.

libraries/physics/src/PhysicalEntitySimulation.cpp
jherico: The lock is now declared recursive, so it can't deadlock by being called twice.  I believe the code is safer by not releasing the lock before calling the parent class addAction (which immediately reacquires it anyway).

libraries/physics/src/PhysicalEntitySimulation.cpp
jherico: Yes.  Ideally I'd like to remove the ability for PhysicalEntitySimulation to even see _actionsToRemove and _actionsToAdd (as opposed to having a protected member in the parent class that gives access to a const reference.  

The new code calls the parent class `EntitySimulation::applyActionChanges()` to clear the values.  In fact, it makes no sense whatsoever to call `_actionsToRemove.clear()` and then in the next for loop be checking to ensure that an added action isn't in the actions to remove as well.  That was probably a bug.

libraries/physics/src/PhysicalEntitySimulation.cpp
jherico: as above.  `PhysicalEntitySimulation` doesn't actually own these containers... they belong to the parent class which should be responsible for clearing them, and does in `EntitySimulation::applyActionChanges()` which is now called before the function returns.

libraries/physics/src/PhysicalEntitySimulation.cpp
ZappoMan: My concern is that they will deadlock on two different locks... I know that at some point @AndrewMeadows was doing a fair bit of juggling between different locks in different classes.

hifi-gustavo:
Build console output for commit: 3bc9e8c98de63c7063ad65285e41a58fdf890472 available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

libraries/entities/src/EntityItem.cpp
ZappoMan: gak - I see... I misread and thought this was a withTryWriteLock

examples/tests/timerStressTest.js
```diff
@@ -0,0 +1,68 @@
...
+Script.include("https://raw.githubusercontent.com/highfidelity/hifi/master/examples/libraries/utils.js");
+
+var bubbleModel = 'http://hifi-public.s3.amazonaws.com/james/bubblewand/models/bubble/bubble.fbx?' + randInt(0, 10000);;
+//var scriptURL'http://hifi-public.s3.amazonaws.com/james/bubblewand/scripts/wand.js?'+randInt(0,10000);
+
+//for local testing 
+//var scriptURL = "http://localhost:8080/scripts/setRecurringTimeout.js?" + randInt(0, 10000);
+
+
+var scriptURL='http://hifi-public.s3.amazonaws.com/james/debug/timeouts/setRecurringTimeout.js?'+ randInt(0, 10000);
```
ZappoMan: btw- the script you link to here, maybe we should add it to the PR as well - since as you have it now, it references a file that is not under version control... But... not a big deal.That being said, this script setRecurringTimeout.js is definitely some crazy scripting that I would never expect... but I guess we need to guard against it...

It's hooking an update signal (which is called at ~60hz) and in the update signal, it creates a timer that fires later, then once the timer fires it creates another timer in the next update signal. 

I'm not sure why it would crash... and we should fix whatever is causing the crash... but... still... an unexpected pattern for sure.

hifi-gustavo:
Build console output for commit: 91d4cf18a26ed1c697f342df602a280379b849f9 available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

hifi-gustavo:
Build console output for commit: 92eeb564fe4a74c6aa9cad5b4a13d47fa583f808 available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

libraries/physics/src/PhysicalEntitySimulation.cpp
jherico: I'm confident it isn't an issue, but I've added explicit scoping to the QMutexLocker to make it match the previous lock scope.

hifi-gustavo:
Build console output for commit: 2a15726aafc2f4c27ee814cc571c54170d85ae0e available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

jherico:
I'm getting some weird behavior switching domain.  Doing some testing.

hifi-gustavo:
Build console output for commit: e9edbb9dd1aca8159f575f45d7421a2d26d7be48 available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

ZappoMan:
test this please

hifi-gustavo:
Build console output for commit: e9edbb9dd1aca8159f575f45d7421a2d26d7be48 available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

hifi-gustavo:
Build console output for commit: 244795b629e2c3c3c8cf1cf8a19673d374483df5 available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

hifi-gustavo:
Build console output for commit: 7fecd3261da3d725dc058d49081147d66e7cb819 available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

hifi-gustavo:
Build console output for commit: fd1cfadc473003a71e607192977e960f6e21c9d8 available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

hifi-gustavo:
Build console output for commit: 19806c16a710897fa345038dd82fd8d99186e5d7 available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -543,36 +546,39 @@ bool EntityScriptingInterface::actionWorker(const QUuid& entityID,
         return false;
     }
 
-    _entityTree->lockForWrite();
+    EntityItemPointer entity;
+    bool success;
+    _entityTree->withWriteLock([&] {
+        EntitySimulation* simulation = _entityTree->getSimulation();
+        EntityItemPointer entity = _entityTree->findEntityByEntityItemID(entityID);
```
sethalves: this declaration shadows the outer one and causes a crash

hifi-gustavo:
Build console output for commit: 9b818428503ca2481a4d07795af3e7f102949c96 available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

libraries/entities/src/EntityScriptingInterface.cpp
jherico: fixed

hifi-gustavo:
Build console output for commit: 2daa5c3bf63506c30d0b67e1821dece467b086e2 available at: https://gist.github.com/f0b2c5688f43158220c8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5778.dmg

PR #5779, opened 2015-09-11
Make sure all entity script callbacks happen on the Entities script engine thread

ZappoMan:
The callbacks for things like preload, unload, mouse events and others were happening on the main thread. They really should happen on the entities script engine thread. This fix will make sure all these calls are invoked in the write thread.


hifi-gustavo:
Build console output for commit: 92a5a999c6a2478fb46de91bae65fabcf57e95af available at: https://gist.github.com/83abe3aeff392a0534f7<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5779.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5779.dmg

PR #5780, opened 2015-09-11
prevent sending signals in the entities script engine

ZappoMan:
this will disable the emitting of the global signals from the script engine. This should stop the crashes we are seeing in the current toy content.

Notes: 
1) this will break many entity scripts
2) this doesn't stop the emitting of other global signals (like the mouse messages) so scripts that use those might also be problematic
3) this doesn't yet create an alternative

Known content that this will break:
1) cell science - the overlays won't be clickable, other possible issues
2) toys that attempt to detect grab: breakdance, bubble want, flashlight

### DIFF
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -113,10 +113,10 @@ void EntityTreeRenderer::init() {
 
     if (_wantScripts) {
         _entitiesScriptEngine = new ScriptEngine(NO_SCRIPT, "Entities",
-                                        _scriptingServices->getControllerScriptingInterface());
+                                        _scriptingServices->getControllerScriptingInterface(), false);
         _scriptingServices->registerScriptEngineWithApplicationServices(_entitiesScriptEngine);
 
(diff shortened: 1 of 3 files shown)


hifi-gustavo:
Build console output for commit: b9d5412aad8a6b23cb2b45ec2cde96f9581b1362 available at: https://gist.github.com/6c781e8a6fff173c13b8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5780.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5780.dmg

thoys:
@ZappoMan this also breaks our script, that loads another script when you enter the domain. Is there a solution for this coming?

ZappoMan:
Yes. We are working on a fix. I hope to have something early next week. Can you share your scripts with me? So I can consider  them in my design for a fix?

thoys:
Its basicly the same script as this one, with a different Script URL: https://github.com/highfidelity/hifi/blob/master/examples/example/entities/jsstreamplayerdomain-zone-entity.js , thanks for the fast reply!

ZappoMan:
Hmmm that's interesting.  I wouldn't expect this script to stop working from this change. I will look into it.

PR #5782, opened 2015-09-11
quick hack to get scripts evaluating on their own thread

ZappoMan:
This is a quick hack that will move Script evaluation onto the ScriptEngine's thread. 

The goal is to fix these crashes we're seeing with entity scripts. The logic behind why this might fix the issue is as follows: The entity scripts are running in a QScriptEngine on their own thread.But when new scripts are downloaded for new entities, they were getting evaluated on the main application thread, We know that QScriptEngine is not thread safe, and so the evaluate() could be corrupting the engine.

This is definitely not an elegant and correct long term solution. We should be moving all ScriptEngine related work onto the thread of the script. Some of the code that currently lives in EntityTreeRenderer (which is mostly running on the main thread) should be moved into ScriptEngine.


hifi-gustavo:
Build console output for commit: aa1c9e23397ee0c8a9ba725cc45621ad9efc65f4 available at: https://gist.github.com/13ea20e0a37f570c6206<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5782.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5782.dmg

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -541,6 +546,17 @@ void ScriptEngine::evaluate() {
 }
 
 QScriptValue ScriptEngine::evaluate(const QString& program, const QString& fileName, int lineNumber) {
+    if (QThread::currentThread() != thread()) {
+        QScriptValue result;
+        QMetaObject::invokeMethod(this,
```
jherico: Do we not need to use `Qt::BlockingQueuedConnection` in order to ensure that the return value is valid?

ZappoMan: ok - will fix

ZappoMan: the fix isn't so easy. I can work on it... but it's not as simple as adding Qt::BlockingQueuedConnection

ZappoMan:
reworking ScriptEngine to be thread safe in a different PR

PR #5783, opened 2015-09-11
Revert "Make sure all entity script callbacks happen on the Entities …

ZappoMan:
This may seem counter-intuitive, but this will force all entity script stuff BACK ONTO the MAIN THREAD... which now that we've disabled update signals is the easiest way to make all entity scripts activity on a single thread.


hifi-gustavo:
Build console output for commit: 5c6d8da97b26c3ffb4a0fbe7b5593077d028fba2 available at: https://gist.github.com/40a7e7fa9da2a681bb0b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5783.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5783.dmg

PR #5784, opened 2015-09-11
Rationalize rig settings

howard-stearns:
1. You can now have at most one of rig animations and graph animations.
2. Force changes to these or changes of avatar to occur while single threaded.
3. Force changes of single/multi-threaded avatar updates to occur while no animations are on.

This gets us more consistent and rational behavior. It eliminates most but not all of the multi-threaded crashes (https://app.asana.com/0/32622044445063/49519400151430). 

We're currently prioritizing movement to the new anim graph, so sweating the rest of these will wait until we've simplified.


hifi-gustavo:
Build console output for commit: 02fba12a488656b3048b198f8cec691df270d5d7 available at: https://gist.github.com/5b004567c0d22e57f67c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5784.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5784.dmg

interface/src/Application.cpp
```diff
@@ -1,4 +1,4 @@
-//
+		//
```
sethalves: whitespace

hifi-gustavo:
Build console output for commit: 841e23e29d38382700315adb865040c2037290d3 available at: https://gist.github.com/5b004567c0d22e57f67c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5784.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5784.dmg

hifi-gustavo:
Build console output for commit: 5314e0c744752ce5d78c1cbef40553d22815348d available at: https://gist.github.com/5b004567c0d22e57f67c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5784.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5784.dmg

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -29,7 +29,7 @@ void AnimInverseKinematics::loadDefaultPoses(const AnimPoseVec& poses) {
 }
 
 void AnimInverseKinematics::loadPoses(const AnimPoseVec& poses) {
-    assert(_skeleton && _skeleton->getNumJoints() == (int)poses.size());
+    assert(_skeleton && (!poses.size() || (_skeleton->getNumJoints() == (int)poses.size())));
     if (_skeleton->getNumJoints() == (int)poses.size()) {
```
AndrewMeadows: So... it is OK if poses.size() is not the same as _skeleton->getNumJoints() as long as it is zero.. in which case _relativePoses will be cleared.  Seth mentioned that he crashed on that assert once in a recent build so I guess this is to fix some race condition where things are not quite setup correctly by the time the program reaches this spot.

Nevertheless, the conversion to bool of the integer value of poses.size() is verboten according to coding standard section 3.2.5 (Implicit test for 0 should not be used other than for boolean variables or non-NULL pointers).

howard-stearns: Thanks. I'll change !poses.size() to (poses.size() == 0)

hifi-gustavo:
Build console output for commit: 140ecc8e2c0a012375319fb91f3671b04afedfc3 available at: https://gist.github.com/5b004567c0d22e57f67c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5784.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5784.dmg

AndrewMeadows:
I tested this.  My FPS has recovered (IK happening on other thread, right?) and I could switch back and forth between old and new IK without problems.

PR #5786, opened 2015-09-11
inverse kinematics works with rotation constraints

AndrewMeadows:
This PR fixes some bugs when using IK with constraints at the joints.

Also modifies how the hand targets are rendered -- the XYZ axes of the targets are rendered and the "finger tip" is shown as a green sphere on the y-axis.

### DIFF
--- a/interface/src/avatar/Hand.cpp
+++ b/interface/src/avatar/Hand.cpp
@@ -8,21 +8,20 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 
-#include <QImage>
+#include "Hand.h"
+
+#include <glm/glm.hpp>
 
 #include <GeometryUtil.h>
-#include <NodeList.h>
+#include <RenderArgs.h>
 
+#include "Avatar.h"
 #include "AvatarManager.h"
-#include "Hand.h"
-#include "Menu.h"
 #include "MyAvatar.h"
 #include "Util.h"
+#include "world.h"
 
 using namespace std;
-
-const float PALM_COLLISION_RADIUS = 0.03f;
-
 
 Hand::Hand(Avatar* owningAvatar) :
     HandData((AvatarData*)owningAvatar),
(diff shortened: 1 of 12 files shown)


libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -203,7 +203,13 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
...
                                 if (constraint) {
-                                    constraint->apply(newRot);
+                                    bool constrained = constraint->apply(newRot);
+                                    if (constrained) {
+                                        // the constraint will modify the movement of the tip so we have to compute the modified
+                                        // model-frame deltaRotation
+                                        // Q' = Qp^ * dQ * Q  -->  dQ =   Qp * Q' * Q^
+                                        deltaRotation = absolutePoses[parentIndex].rot * newRot * glm::inverse(absolutePoses[index].rot);
+                                    }
                                 }
```
AndrewMeadows: This was an important fix: when the constraint modifies the rotation we must compute the effective delta-rotation in the model frame so we can correctly update the "tip" position as we iterate down (up?) the skeleton joints.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -258,7 +264,18 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
...
+        // relax toward underpose
+        const float RELAXATION_TIMESCALE = 0.25f;
+        const float alpha = glm::clamp(dt / RELAXATION_TIMESCALE, 0.0f, 1.0f);
+        int numJoints = (int)_relativePoses.size();
+        for (int i = 0; i < numJoints; ++i) {
+            float dotSign = copysignf(1.0f, glm::dot(_relativePoses[i].rot, _defaultRelativePoses[i].rot));
+            _relativePoses[i].rot = glm::normalize(glm::lerp(_relativePoses[i].rot, dotSign * underPoses[i].rot, alpha));
+        }
+    }
     return evaluate(animVars, dt, triggersOut);
```
AndrewMeadows: The IK is much more stable if we always start at the last solution but relax toward the underpose (e.g. the target animation state before IK).

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -558,14 +604,11 @@ void AnimInverseKinematics::setSkeletonInternal(AnimSkeleton::ConstPointer skele
...
 
-    // No constraints!
-    /*
     if (skeleton) {
         initConstraints();
     } else {
         clearConstraints();
     }
-    */
 }
```
AndrewMeadows: We now create some default constraints for the arm and spine joints.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -516,7 +562,7 @@ void AnimInverseKinematics::initConstraints() {
             eConstraint->setAngleLimits(minAngle, maxAngle);
 
             constraint = static_cast<RotationConstraint*>(eConstraint);
-        } else if (0 == name.compare("Foot", Qt::CaseInsensitive)) {
+        } else if (0 == baseName.compare("FootXXX", Qt::CaseInsensitive)) {
```
AndrewMeadows: The "XXX" is to disable default constraints for these joints until later testing.

libraries/animation/src/SwingTwistConstraint.cpp
```diff
@@ -201,6 +241,7 @@ bool SwingTwistConstraint::apply(glm::quat& rotation) const {
     }
 
     if (swingWasClamped || twistWasClamped) {
+        // update the rotation
         twistRotation = glm::angleAxis(clampedTwistAngle, yAxis);
         rotation = swingRotation * twistRotation * _referenceRotation;
         return true;
```
AndrewMeadows: These changes reduce the "popping" when greatly exceeding twist limits and rolling around toward the far limits on the other side.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -422,15 +455,28 @@ void AnimInverseKinematics::initConstraints() {
...
             stConstraint->setSwingLimits(swungDirections);
+            */
 
-            /*
+            // simple cone
             std::vector<float> minDots;
-            const float MAX_HAND_SWING = PI / 3.0f;
+            const float MAX_HAND_SWING = PI / 2.0f;
             minDots.push_back(cosf(MAX_HAND_SWING));
             stConstraint->setSwingLimits(minDots);
```
AndrewMeadows: Although PI/2 is a bit far for the wrist to bend, especially about its z-axis (normal to palm), this simple cone limit works well enough.

hifi-gustavo:
Build console output for commit: c589f0e58e2a01a62d1404ea321b4dcce9199eb6 available at: https://gist.github.com/5f29d6ddf89ad4a3d799<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5786.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5786.dmg

AndrewMeadows:
Don't merge -- there is a crash bug I'm hunting.

AndrewMeadows:
Ok, this is good to go.

I've noticed that my FPS has gone way down after adding the IK computations (taking up to 8 msec to iterate 16 times over three targets) for the debug build.  This makes me sad, however these calculations are 10x faster in an optimized release build (on linux).

hifi-gustavo:
Build console output for commit: d01c5d1ae5ea9102af84165c9690a07ba52991a7 available at: https://gist.github.com/5f29d6ddf89ad4a3d799<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5786.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5786.dmg

interface/src/avatar/Hand.cpp
```diff
@@ -40,76 +39,64 @@ void Hand::simulate(float deltaTime, bool isMine) {
     }
 }
 
-void Hand::render(RenderArgs* renderArgs, bool isMine) {
+void Hand::renderHandTargets(RenderArgs* renderArgs, bool isMine) {
+    const float avatarScale = DependencyManager::get<AvatarManager>()->getMyAvatar()->getScale();
```
howard-stearns: _owningAvatar->getScale() instead?  (Avoid use of DependencyManger pseudo-globals, which will cause problems for threading.)

hifi-gustavo:
Build console output for commit: e85cdad8c7a2520c7b6a53895690fb8994c41ce3 available at: https://gist.github.com/5f29d6ddf89ad4a3d799<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5786.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5786.dmg

PR #5787, opened 2015-09-11
DO NOT MERGE: Cleanup asset requests on node kill

huffman:


hifi-gustavo:
Build console output for commit: a5ba86514c9851bacd229b3a61652175a4669d89 available at: https://gist.github.com/0048ba860b58414d010d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5787.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5787.dmg

huffman:
@birarda The leak should be fixed now

libraries/networking/src/AssetClient.cpp
```diff
@@ -244,8 +265,55 @@ void AssetClient::handleAssetUploadReply(QSharedPointer<NLPacket> packet, Shared
...
+        auto requestIt = messageCallbackMap.find(messageID);
+        if (requestIt != messageCallbackMap.end()) {
+            auto callback = requestIt->second;
+            callback(true, error, hashString);
+            messageCallbackMap.erase(requestIt);
+        }
+
+        // Although the messageCallbackMap may now be empty, we won't delete the node until we have disconnected from
+        // it to avoid constantly creating/deleting the map on subsequent requests.
+    }
```
birarda: I don't love how redundant the above three sections are (checking for the node and then the messageID) but I get that they're touching different containers - this is probably a place where sharing a method between the three would make it a bit more difficult to see what is going on.

libraries/networking/src/AssetClient.cpp
```diff
@@ -244,8 +265,55 @@ void AssetClient::handleAssetUploadReply(QSharedPointer<NLPacket> packet, Shared
...
+
+    {
+        auto messageMapIt = _pendingUploads.find(node);
+        if (messageMapIt != _pendingUploads.end()) {
+            for (const auto& value : messageMapIt->second) {
+                value.second(false, AssetServerError::NoError, "");
+            }
+            messageMapIt->second.clear();
+        }
     }
```
birarda: The scoping above is nice to separate the three checks, but I think the same could be accomplished in a way that doesn't make me wonder why we're scoping by leading each with a comment.

libraries/networking/src/AssetClient.cpp
```diff
@@ -244,8 +265,55 @@ void AssetClient::handleAssetUploadReply(QSharedPointer<NLPacket> packet, Shared
...
+            messageCallbackMap.erase(requestIt);
+        }
+
+        // Although the messageCallbackMap may now be empty, we won't delete the node until we have disconnected from
+        // it to avoid constantly creating/deleting the map on subsequent requests.
+    }
+}
+
+void AssetClient::handleNodeKilled(SharedNodePointer node) {
+    {
```
birarda: should check here if the node that was killed is even an AssetClient before we start to check the data structures

libraries/networking/src/AssetClient.cpp
huffman: Do you have a suggestion for how you'd like to see that done? When I wrote this I was somewhat inclined to combine the data structures, store the packet type expected, and verify on receive. This would reduce a lot of the code.

libraries/networking/src/AssetClient.h
```diff
@@ -46,15 +47,17 @@ private slots:
...
     bool getAsset(const QString& hash, const QString& extension, DataOffset start, DataOffset end, ReceivedAssetCallback callback);
     bool uploadAsset(const QByteArray& data, const QString& extension, UploadResultCallback callback);
 
     static MessageID _currentID;
-    QHash<MessageID, ReceivedAssetCallback> _pendingRequests;
-    QHash<MessageID, GetInfoCallback> _pendingInfoRequests;
-    QHash<MessageID, UploadResultCallback> _pendingUploads;
+    std::unordered_map<SharedNodePointer, std::unordered_map<MessageID, ReceivedAssetCallback>> _pendingRequests;
+    std::unordered_map<SharedNodePointer, std::unordered_map<MessageID, GetInfoCallback>> _pendingInfoRequests;
+    std::unordered_map<SharedNodePointer, std::unordered_map<MessageID, UploadResultCallback>> _pendingUploads;
```
birarda: Something feels weird to me about having these be shared pointers to the nodes instead of just their UUID (which we already have a hasher for) or a weak pointer, but I also can't specifically think of a case where the AssetClient wouldn't receive the nodeKilled signal (and therefore keep a node around that we're trying to clean up) unless it was also being torn down at the same time.

hifi-gustavo:
Build console output for commit: b8c1c279c62e1e371f519e4b18e8c55b0809de2f available at: https://gist.github.com/0048ba860b58414d010d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5787.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5787.dmg

hifi-gustavo:
Build console output for commit: e87fcb78cd3b7775c8da93014bce04ab5a32c439 available at: https://gist.github.com/0048ba860b58414d010d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5787.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5787.dmg

libraries/networking/src/AssetClient.cpp
birarda: I think something like that is best - it would be great to have a way to check the sig of the callback and just call it if you have one that matches what you'd want to call for the packet type that just arrived.

birarda: https://functionalcpp.wordpress.com/2013/08/05/function-traits/ something like this could be an option to do the right thing depending on the type of third param of callback

libraries/networking/src/AssetClient.h
huffman: The reason to store a pointer rather than a uuid is that we are concerned about the connection to the server going away, of which there can be multiple for a single UUID (although not simultaneously AFAIK). I can imagine a race condition where you have a "nodeKilled" message queued up but you don't handle it before a new connection is made and a new message is sent.

libraries/networking/src/AssetClient.cpp
huffman: I scoped it so I didn't have to come up with unique variable names for each. I can switch it though.

birarda: I suppose this would also be cleaned up by putting everything into one `unordered_map`

hifi-gustavo:
Build console output for commit: 42d21e4dc1eec4af2eff094ce3335434923af0b0 available at: https://gist.github.com/0048ba860b58414d010d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5787.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5787.dmg

libraries/networking/src/AssetClient.h
birarda: Okay, makes sense. Protects the case where you kill a node the DS hasn't killed that it gives back to you with the same UUID.

libraries/networking/src/AssetClient.cpp
huffman: Thanks, I'll check that out

hifi-gustavo:
Build console output for commit: 29d26a1f9a349de957e20cf621603dd5e9fed9e2 available at: https://gist.github.com/0048ba860b58414d010d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5787.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5787.dmg

PR #5789, opened 2015-09-12
edit.js adjustment

sethalves:
- don't move camera when edit.js is activated
- don't highlight entities that are moused over


howard-stearns:
Works for me. I guess my machine is fast enough to not mind the highlight box, but that changed perspective thing had been driving me nuts.

PR #5795, opened 2015-09-14
fix high resolution timing for VS2012 in networking

birarda:
Below VS2015, the `high_resolution_clock` in `std::chrono` is typedef'ed to `system_clock`. On windows this gives a resolution of around 1ms, which is not sufficient for some of the timing code in the networking library.

Added a class `win_high_resolution_clock` that uses `QueryPerformanceFrequency` for properly high resolution timing if using VS below 2015.


hifi-gustavo:
Build console output for commit: 302fd923b851aa9110576459107c4e065a6aac76 available at: https://gist.github.com/df501ba29a3d06f8599a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5795.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5795.dmg

PR #5796, opened 2015-09-14
Fixing differing highlights between left and right eyes in stereo modes

jherico:
This adds a stereo flag to the deferred rendering so that the highlights calculations are compensating for the stereo framebuffer when calculating the position and subsequently the specular highlights of an object.  

This is an incomplete fix because it doesn't provide the proper left / right projection matrices for unprojection back into eye space, but the difference should be slight enough to avoid making people ill.

### DIFF
--- a/libraries/gpu/src/gpu/Context.cpp
+++ b/libraries/gpu/src/gpu/Context.cpp
@@ -44,6 +44,10 @@ void Context::enableStereo(bool enable) {
     _backend->enableStereo(enable);
 }
 
+bool Context::isStereo() {
+    return _backend->isStereo();
+}
+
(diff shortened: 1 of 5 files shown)


hifi-gustavo:
Build console output for commit: 1bb15e28d67fc45de0b96337a2aa8a075e123abf available at: https://gist.github.com/ebccc39b0ce813f76b5c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5796.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5796.dmg

PR #5800, opened 2015-09-14
Fix over-calculation of cluster matrices

jherico:
Currently the model parts of an FBX call updateClusterMatrices on the model each time any part is rendered.  If you have an FBX mesh with 400 parts and they're all visible, then you're updating the cluster matrices for 400 parts 400 times, i.e. 159,999 more times more than you need to.  Since the update involves multiple matrix compositions, this impacts framerates.

This adds an update hook to the model parts so that they reset a flag.  In the above situation we'll technically reset the flag 399 more times than we need to, but that's pretty cheap.  We'll update the cluster matrices only once and trip the flag, so that subsequent parts don't repeat the calculation needlessly.

### DIFF
--- a/libraries/render-utils/src/Model.cpp
+++ b/libraries/render-utils/src/Model.cpp
@@ -798,7 +798,7 @@ namespace render {
             return payload->model->renderPart(args, payload->meshIndex, payload->partIndex, payload->transparent);
         }
     }
-
+    
    /* template <> const model::MaterialKey& shapeGetMaterialKey(const MeshPartPayload::Pointer& payload) {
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 35e4440c4f6941ceffdbaff4634500522e79bac1 available at: https://gist.github.com/09f32535a8adf112b37f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5800.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5800.dmg

PR #5801, opened 2015-09-14
fix animation unit tests

AndrewMeadows:
The **animation-tests** build target was failing to compile and some tests were failing.  Now fixed.

### DIFF
--- a/libraries/animation/src/SwingTwistConstraint.cpp
+++ b/libraries/animation/src/SwingTwistConstraint.cpp
@@ -72,7 +72,7 @@ SwingTwistConstraint::SwingTwistConstraint() :
         _swingLimitFunction(),
         _minTwist(-PI),
         _maxTwist(PI),
-        _lastTwistBoundary(0)
+        _lastTwistBoundary(LAST_CLAMP_NO_BOUNDARY)
 {
 }
 
@@ -249,3 +249,6 @@ bool SwingTwistConstraint::apply(glm::quat& rotation) const {
     return false;
 }
 
+void SwingTwistConstraint::clearHistory() {
+    _lastTwistBoundary = LAST_CLAMP_NO_BOUNDARY;
+}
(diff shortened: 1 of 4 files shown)


libraries/animation/src/SwingTwistConstraint.cpp
```diff
@@ -249,3 +249,6 @@ bool SwingTwistConstraint::apply(glm::quat& rotation) const {
     return false;
 }
 
+void SwingTwistConstraint::clearHistory() {
+    _lastTwistBoundary = LAST_CLAMP_NO_BOUNDARY;
+}
```
AndrewMeadows: The behavior of the SwingTwistConstraint had changed to remember what boundary it had last clamped against, in an effort to prevent joints from popping about when their input rotation had wrapped all the way around.  We need to clear the history of the last clamp when doing unit tests.

tests/animation/src/AnimInverseKinematicsTests.cpp
```diff
@@ -130,10 +129,13 @@ void AnimInverseKinematicsTests::testSingleChain() {
...
         //
-        int indexD = 3;
         glm::vec3 targetPosition(2.0f, 1.0f, 0.0f);
         glm::quat targetRotation = glm::angleAxis(PI / 2.0f, zAxis);
-        ikDoll.updateTarget(indexD, targetPosition, targetRotation);
+        AnimVariantMap varMap;
+        varMap.set("positionD", targetPosition);
+        varMap.set("rotationD", targetRotation);
+        ikDoll.setTargetVars("D", "positionD", "rotationD");
+        AnimNode::Triggers triggers;
```
AndrewMeadows: The API of **AnimNode::evaluate()** had changed but this unit test was not maintained... because we don't yet automatically build unit tests for PR's.

hifi-gustavo:
Build console output for commit: 75ec1428275dee7fa39ac2c347d559b7cc79ef64 available at: https://gist.github.com/312e6208874409fa1d6a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5801.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5801.dmg

PR #5808, opened 2015-09-15
Audio fixes

kencooke:
Bug fix and optimization of int16_t to float conversions

### DIFF
--- a/libraries/audio/src/AudioFilterBank.h
+++ b/libraries/audio/src/AudioFilterBank.h
@@ -130,7 +130,7 @@ public:
         // de-interleave and convert int16_t to float32 (normalized to -1. ... 1.)
         for (uint32_t i = 0; i < frameCount; ++i) {
             for (uint32_t j = 0; j < _channelCount; ++j) {
-                _buffer[j][i] = ((float)(*in++)) / scale;
+                _buffer[j][i] = ((float)(*in++)) * (1.0f / scale);
             }
         }
 
--- a/libraries/audio/src/AudioFilterBank.h
+++ b/libraries/audio/src/AudioFilterBank.h
@@ -125,7 +125,7 @@ public:
             return;
         }
 
-        const int scale = (2 << ((8 * sizeof(int16_t)) - 1));
+        const int scale = (1 << ((8 * sizeof(int16_t)) - 1));
 
         // de-interleave and convert int16_t to float32 (normalized to -1. ... 1.)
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: ffe3fcf4cea8fa88bc172a64ce5b219ae09de774 available at: https://gist.github.com/3ef5903c8fd1dc0fc502<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5808.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5808.dmg

libraries/audio/src/AudioFilterBank.h
```diff
@@ -125,12 +125,12 @@ class AudioFilterBank {
...
         }
 
-        const int scale = (2 << ((8 * sizeof(int16_t)) - 1));
+        const int scale = (1 << ((8 * sizeof(int16_t)) - 1));
 
         // de-interleave and convert int16_t to float32 (normalized to -1. ... 1.)
         for (uint32_t i = 0; i < frameCount; ++i) {
             for (uint32_t j = 0; j < _channelCount; ++j) {
-                _buffer[j][i] = ((float)(*in++)) / scale;
+                _buffer[j][i] = ((float)(*in++)) * (1.0f / scale);
```
AndrewMeadows: I don't understand how this saves a division since you still need to divide **1.0f** by **scale**.  Are you saying that you expect the compiler to cache the value (1.0f / scale) in a register?  If I had been trying to optimize this I would have created a **const float invScale = 1.0f / scale** outside of the loop and then crossed my fingers that the compiler would cache it.  I'm just curious: how smart are the optimization tricks of compilers these days?

jherico: I'm pretty sure any modern compiler would be able to optimize (`const_value_expr1` / `const_value_expr2`) to (`const_value_expr3`)

kencooke: Yes, a modern compiler should hoist a simple loop-invariant constant outside the loop, where the expensive division is only done once.  You can do it manually, if you feel that it does not clutter the code.  The manual vs. compiler tradeoff is tricky enough that for anything important, I always look at the assembly output and make sure the compiler is doing what I want/expect.

The problem with the original code is the MSVC compiler defaults to /fp:precise and will not strength-reduce a division into multiply-by-inverse.

PR #5810, opened 2015-09-15
Script-accessible animation-state

howard-stearns:
Create a thread-safe copy of anim vars exposed to javascript.

This is a minimal version in which javascript can only write and read new vars (not the ones maintained by the Rig), and only bool and float of those.

### DIFF
--- a/libraries/animation/src/Rig.h
+++ b/libraries/animation/src/Rig.h
@@ -42,27 +42,13 @@
 
 #include "AnimNode.h"
 #include "AnimNodeLoader.h"
+#include "ScriptableRigStateVariables.h"
 
 class AnimationHandle;
 typedef std::shared_ptr<AnimationHandle> AnimationHandlePointer;
 
 class Rig;
 typedef std::shared_ptr<Rig> RigPointer;
-
-class SharedAnimationState : public QObject {
-    Q_OBJECT
-public:
-    // locked so that multiple scripts (and the rig) can access from different threads.
-    Q_INVOKABLE bool get(const QString& key, const bool& defaultValue) const { QMutexLocker locker(&_mutex); return hash.contains(key) ? hash.value(key).getBool() : defaultValue; }
-    Q_INVOKABLE float get(const QString& key, const float& defaultValue) const { QMutexLocker locker(&_mutex); return hash.contains(key) ? hash.value(key).getFloat() : defaultValue; }
-    Q_INVOKABLE void set(const QString& key, bool value) { QMutexLocker locker(&_mutex);  hash.insert(key, AnimVariant(value)); }
-    Q_INVOKABLE void set(const QString& key, float value) { QMutexLocker locker(&_mutex);  hash.insert(key, AnimVariant(value)); }
-    // Note that Javascript [] notation affects QObject properties, not any operator[] that might be defined here.
-    QHash<QString, AnimVariant> hash {};
-private:
-    mutable QMutex _mutex;
-};
-Q_DECLARE_METATYPE(SharedAnimationState*)
 
 class Rig : public QObject, public std::enable_shared_from_this<Rig> {
 public:
@@ -194,7 +180,7 @@ public:
 
     AnimNode::ConstPointer getAnimNode() const { return _animNode; }
     AnimSkeleton::ConstPointer getAnimSkeleton() const { return _animSkeleton; }
-    SharedAnimationState* getAnimationState() { return &_sharedVars; }
+    ScriptableRigStateVariables* getAnimationState() { return &_sharedVars; }
     AnimVariantMap& getAnimVars() { return _animVars; } // for testing
 
  protected:
@@ -226,7 +212,7 @@ public:
     std::shared_ptr<AnimSkeleton> _animSkeleton;
     std::unique_ptr<AnimNodeLoader> _animLoader;
     AnimVariantMap _animVars;
-    SharedAnimationState _sharedVars {}; // Exposed to all scripts through MyAvatar
+    ScriptableRigStateVariables _sharedVars {}; // Exposed to all scripts through MyAvatar
     enum class RigRole {
         Idle = 0,
         Turn,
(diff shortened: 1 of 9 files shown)


hifi-gustavo:
Build console output for commit: 5ae5e772d53d3b3b482da2f262d7a67553bc4569 available at: https://gist.github.com/337dce93ed34e6caf6dd<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5810.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5810.dmg

libraries/animation/src/Rig.h
```diff
@@ -179,6 +180,8 @@ class Rig : public QObject, public std::enable_shared_from_this<Rig> {
 
     AnimNode::ConstPointer getAnimNode() const { return _animNode; }
     AnimSkeleton::ConstPointer getAnimSkeleton() const { return _animSkeleton; }
+    ScriptableRigStateVariables* getAnimationState() { return &_sharedVars; }
+    AnimVariantMap& getAnimVars() { return _animVars; } // for testing
```
hyperlogic: If for testing only, please remove, otherwise make it a const method that returns a const ref.

libraries/animation/src/ScriptableRigStateVariables.h
```diff
@@ -0,0 +1,35 @@
...
+    Q_OBJECT
+public:
+    // Note that Javascript [] notation affects QObject properties, not any operator[] that might be defined here.
+    Q_INVOKABLE bool get(const QString& key, const bool& defaultValue) const;
+    Q_INVOKABLE float get(const QString& key, const float& defaultValue) const;
+    Q_INVOKABLE void set(const QString& key, bool value);
+    Q_INVOKABLE void set(const QString& key, float value);
+    void doHash(std::function<void(const QHash<QString, AnimVariant>& hash)> functionOfHash); // Safe access of hash, using lambda.
+private:
+    mutable QMutex _mutex;
```
hyperlogic: Is it worth while using a read-write lock?

libraries/animation/src/ScriptableRigStateVariables.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include <QObject>
+#include <QMutex>
+#include <QString>
+#include "AnimVariant.h"
+
+class ScriptableRigStateVariables : public QObject {
+    Q_OBJECT
+public:
+    // Note that Javascript [] notation affects QObject properties, not any operator[] that might be defined here.
```
hyperlogic: What is the JavaScript interface look like?  Does can we use the . operator?  i.e. 

    MyAvatar.getAnimationState().isLeftGrab = true;

libraries/animation/src/ScriptableRigStateVariables.cpp
```diff
@@ -0,0 +1,37 @@
...
+    return _hash.contains(key) ? _hash.value(key).getFloat() : defaultValue;
+}
+void ScriptableRigStateVariables::set(const QString& key, bool value) {
+    QMutexLocker locker(&_mutex);
+    _hash.insert(key, AnimVariant(value));
+}
+void ScriptableRigStateVariables::set(const QString& key, float value) {
+    QMutexLocker locker(&_mutex);
+    _hash.insert(key, AnimVariant(value));
+}
```
hyperlogic: Support for glm::vec3, glm::quat and glm::mat4?

libraries/animation/src/ScriptableRigStateVariables.h
howard-stearns: As I understand it, read-write locks are useful when there's a known asymmetry between the frequency of reads and writes. Alas, there's no way for us to know that about all scripts. I imagine the common use case is more symmetric, in which case we would not be winning with the (slight) extra complexity.

hifi-gustavo:
Build console output for commit: 5c8f603d50ebc68ada4b4402f25ee8ad356f01e8 available at: https://gist.github.com/337dce93ed34e6caf6dd<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5810.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5810.dmg

libraries/animation/src/ScriptableRigStateVariables.cpp
howard-stearns: Correct. See line 15.

hyperlogic:
So, there's a pretty big problem with this interface with script.  The main thread can read the QHash while a script update function is in the middle of manipulating some values.  So the animation graph can read some values from script update t=0, and some values from script update t=1.  Instead we should have an interface that only allows JavaScript to set all values simultaneously.  Similar to the way Entity.setProperties works.

libraries/animation/src/ScriptableRigStateVariables.h
howard-stearns: Alas, no. MyAvatar.getAnimationState().set('isLeftGrab', true). The dot notation (and bracket notation) refers to properties, and here we have implemented each AnimVariant value as a value in a hash table, rather than directly as a property.  The latter would be nice for javascript programmers, but I don't know (yet) how to implement that with the appropriate locks. (Presumably, it would involve not only dynamically creating a property for each 'set' value, but also dynamically creating a (locking) getter/setter for each value.)

howard-stearns:
Agreed re atomic script access. We might want to change to that -- after we figure out what other requirements are important.  For example, a script might also "need" to ensure that the values are changed consistently with physics updates or entity-properties dribbling in on the wire, and we don't have any way to coordinate with that.  (Of course, I don't want us to fall victim to the Xanadu Fallacy, so we may indeed change this as we get more experience.)

hyperlogic:
I can almost guarantee that we'll need the ability to set many fields in an AnimState as an atomic operation.  A common assumption is that a set of boolean animVars are mutually exclusive, i.e. like radio buttons.  Not having atomic script updates makes it impossible to guarantee this.  this could lead to difficult to diagnose, happens every once and a while bugs.

howard-stearns:
Closing for now, until we figure out what we want/need to expose.

PR #5814, opened 2015-09-15
Fixing HMD and stereo lighting bug

samcake:
All the deferred passes (for the main global directional lights and all the local lights) were broken in stereo since our last improvment on the performance front. We now explictely do the 2 sides rendering for the deferred lighting if in stereo.

- Introduce the explicit 2 side rendering if in stereo and pass the correct transformation stack information for each side

- Fix and maybe simplify the evaluation of the eye position from the Z buffer value and fragment position using the actual projection matrix of the side

- Gather all the different parameters used for deferred in a single uniforma buffer, make one version for each side, and basically evaluate the value once and bind the Uniform Buffer for all the different deferred shaders in on place


hifi-gustavo:
Build console output for commit: 7b8e11e01ce4b38002719713844f5d4970872e5f available at: https://gist.github.com/86a088889070a156fa77<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5814.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5814.dmg

libraries/render-utils/src/DeferredLightingEffect.cpp
```diff
@@ -289,296 +291,287 @@ void DeferredLightingEffect::render(RenderArgs* args) {
...
+            // Combine the side projection with the side View offset (the view matrix is same for all side)
+            projMats[i] = projMats[i] * eyeViews[i];
 
-        glm::vec4 color(1.0f, 1.0f, 1.0f, 1.0f);
-        glm::vec2 topLeft(-1.0f, -1.0f);
-        glm::vec2 bottomRight(1.0f, 1.0f);
-        glm::vec2 texCoordTopLeft(sMin, tMin);
-        glm::vec2 texCoordBottomRight(sMin + sWidth, tMin + tHeight);
+            deferredTransforms[i].projection = projMats[i];
+            deferredTransforms[i].viewInverse = monoViewMat;
```
jherico: Why monoViewMat here instead of eyeViews[i]?

libraries/render-utils/src/DeferredLightingEffect.cpp
```diff
@@ -289,296 +291,287 @@ void DeferredLightingEffect::render(RenderArgs* args) {
...
+        float halfWidth = 0.5 * sWidth;
 
-    {
-        Transform model;
-        model.setTranslation(glm::vec3(sMin, tMin, 0.0));
-        model.setScale(glm::vec3(sWidth, tHeight, 1.0));
-        batch.setModelTransform(model);
+        for (int i = 0; i < numPasses; i++) {
+            // In stereo, the 2 sides are layout side by side in the mono viewport and their width is half
+            int sideWidth = monoViewport.z * 0.5;
```
jherico: monoViewport and sideWidth are both ints.  Should use `monoViewport.z >> 1`

hifi-gustavo:
Build console output for commit: f00e3829f1b6c6c7b1493f798f6cc7d0a7e2861c available at: https://gist.github.com/86a088889070a156fa77<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5814.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5814.dmg

libraries/render-utils/src/DeferredLightingEffect.cpp
samcake: Because instead of having 2 viewMatrices, i can simply combine the eye View offset in the side dependent projection matrices
So only the proj Matrix is different for each side

hifi-gustavo:
Build console output for commit: 8ab96fe9072a4884ff243c0ad68e2a0bbfbc376f available at: https://gist.github.com/86a088889070a156fa77<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5814.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5814.dmg

libraries/render-utils/src/DeferredLightingEffect.cpp
jherico: How does that work if the user is leaning their head to the left or right?  In that case the translation of the projection matrix doesn't align with the translation of the view matrix.  The view matrix is only translated on the X axis if the user's head is perfectly level and they're facing in negative Z.

samcake: That's a good point, does the eyeView offset matrices contains more than just the eye lateral offset ?
Does that also contains the head movement (orientation and translation) coming from the head sensor ?
If so yes that doesn;t work.
I would think that the head movement should be integrated in the official viewTransform, and the eyeViewOffset be just the lateral inter eye along X translation

jherico: that's what eyeViews[i] is.  The original head transform composed with the rotation and orientation obtained from the Hmd sensors.  I think this should just be eyeViews[i] and eliminate whatever code you're using to combine the eye offset with the projection matrix.

hifi-gustavo:
Build console output for commit: 5410bc0945145772d9101cc4fbc45d89ad2d2d36 available at: https://gist.github.com/86a088889070a156fa77<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5814.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5814.dmg

libraries/render-utils/src/DeferredLightingEffect.cpp
samcake: OK so just reverted to the separated 2 offset matrices

hifi-gustavo:
Build console output for commit: 80d25a68b639b1d93867635f642ab90b53b6228d available at: https://gist.github.com/86a088889070a156fa77<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5814.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5814.dmg

PR #5815, opened 2015-09-15
couple of entity server corner cases

ZappoMan:
* fix for crash in OctreeServer::sendStatsPacket() for timer firing before the inbound packet processor is setup
* fix for crash in empty settings object
* fix for debug assert in glm::perspective() if query ever has near and far clip set to same value

### DIFF
--- a/assignment-client/src/octree/OctreeQueryNode.cpp
+++ b/assignment-client/src/octree/OctreeQueryNode.cpp
@@ -242,7 +242,8 @@ bool OctreeQueryNode::updateCurrentViewFrustum() {
 
     if (0.0f != getCameraAspectRatio() &&
         0.0f != getCameraNearClip() &&
-        0.0f != getCameraFarClip()) {
+        0.0f != getCameraFarClip() &&
+        getCameraNearClip() != getCameraFarClip()) {
         newestViewFrustum.setProjection(glm::perspective(
             glm::radians(wideFOV), // hack
             getCameraAspectRatio(),
--- a/assignment-client/src/octree/OctreeServer.cpp
+++ b/assignment-client/src/octree/OctreeServer.cpp
@@ -908,11 +908,11 @@ void OctreeServer::readConfiguration() {
 
     qDebug() << "Got domain settings from domain-server.";
 
-    if (domainHandler.getSettingsObject().isEmpty()) {
-        qDebug() << "No settings object from domain-server.";
-    }
     QJsonObject settingsObject { domainHandler.getSettingsObject() };
 
+    if (settingsObject.isEmpty()) {
+        qDebug() << "No settings object from domain-server.";
+    }
 
     QString settingsKey = getMyDomainSettingsKey();
     QJsonObject settingsSectionObject = settingsObject[settingsKey].toObject();
--- a/assignment-client/src/octree/OctreeServer.cpp
+++ b/assignment-client/src/octree/OctreeServer.cpp
@@ -887,7 +887,7 @@ bool OctreeServer::readOptionString(const QString& optionName, const QJsonObject
     return optionAvailable;
 }
 
-void OctreeServer::readConfiguration() {
+bool OctreeServer::readConfiguration() {
     // if the assignment had a payload, read and parse that
     if (getPayload().size() > 0) {
         parsePayload();
@@ -906,14 +906,13 @@ void OctreeServer::readConfiguration() {
     domainHandler.requestDomainSettings();
     loop.exec();
 
-    qDebug() << "Got domain settings from domain-server.";
-
-    QJsonObject settingsObject { domainHandler.getSettingsObject() };
-
-    if (settingsObject.isEmpty()) {
-        qDebug() << "No settings object from domain-server.";
+    if (domainHandler.getSettingsObject().isEmpty()) {
+        qDebug() << "Failed to retreive settings object from domain-server. Bailing on assignment.";
+        setFinished(true);
+        return false;
     }
 
+    const QJsonObject& settingsObject = domainHandler.getSettingsObject();
     QString settingsKey = getMyDomainSettingsKey();
     QJsonObject settingsSectionObject = settingsObject[settingsKey].toObject();
     _settings = settingsSectionObject; // keep this for later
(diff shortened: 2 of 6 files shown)


assignment-client/src/octree/OctreeServer.cpp
```diff
@@ -906,10 +906,14 @@ void OctreeServer::readConfiguration() {
     domainHandler.requestDomainSettings();
     loop.exec();
 
+    qDebug() << "Got domain settings from domain-server.";
+
     if (domainHandler.getSettingsObject().isEmpty()) {
         qDebug() << "No settings object from domain-server.";
     }
-    const QJsonObject& settingsObject = domainHandler.getSettingsObject();
+    QJsonObject settingsObject { domainHandler.getSettingsObject() };
```
ZappoMan: To be honest, this doesn't make that much sense to me. But I crashed and `settingsObject` was nullptr... I'm not sure how that could happen, since it appears to return a reference to a member variable, and how is it that the line above didn't crash?

assignment-client/src/octree/OctreeServer.cpp
```diff
@@ -1333,19 +1337,22 @@ void OctreeServer::sendStatsPacket() {
...
-    
     QJsonObject timingArray2;
-    timingArray2["1. avgTransitTimePerPacket"] = (double)_octreeInboundPacketProcessor->getAverageTransitTimePerPacket();
-    timingArray2["2. avgProcessTimePerPacket"] = (double)_octreeInboundPacketProcessor->getAverageProcessTimePerPacket();
-    timingArray2["3. avgLockWaitTimePerPacket"] = (double)_octreeInboundPacketProcessor->getAverageLockWaitTimePerPacket();
-    timingArray2["4. avgProcessTimePerElement"] = (double)_octreeInboundPacketProcessor->getAverageProcessTimePerElement();
-    timingArray2["5. avgLockWaitTimePerElement"] = (double)_octreeInboundPacketProcessor->getAverageLockWaitTimePerElement();
+
+    // Stats Object 3
+    if (_octreeInboundPacketProcessor) {
```
ZappoMan: it's possible for the stats timer to fire before the `_octreeInboundPacketProcessor` has been set up, this guards against that case and simply skips those stats until it's available.

hifi-gustavo:
Build console output for commit: 22b979a663e735ecb15700c3afbb510c49e94e6b available at: https://gist.github.com/7e9c77b7e0d349f67ba1<br/><br/>No installer was built for this commit

assignment-client/src/octree/OctreeServer.cpp
```diff
@@ -906,10 +906,14 @@ void OctreeServer::readConfiguration() {
     domainHandler.requestDomainSettings();
     loop.exec();
 
-    if (domainHandler.getSettingsObject().isEmpty()) {
+    qDebug() << "Got domain settings from domain-server.";
+
+    QJsonObject settingsObject { domainHandler.getSettingsObject() };
```
ZappoMan: To be honest, this doesn't make that much sense to me. But I crashed and settingsObject was null a couple times, and a couple of other times it was some bogus pointer... I'm not sure how that could happen, since it appears to return a reference to a member variable. But this change works around the crash.

birarda: The reference to the member variable isn't the issue - it's that the QJsonObject itself is empty and then used below `settingsObject[settingsKey]` even if we saw it was empty.

I believe the right fix is the following, from the AudioMixer startup:

```
if (domainHandler.getSettingsObject().isEmpty()) {
    qDebug() << "Failed to retreive settings object from domain-server. Bailing on assignment.";
    setFinished(true);
    return;
}
```

ZappoMan: ok - that's reasonable.. you think that we shouldn't run at all if we got no settings?

birarda: Right - it's rare enough and I'm not sure it would make sense to try and default to anything. What we could eventually add is a retry, but stopping the run for now effectively does that - at least from the domain-server's perspective.

hifi-gustavo:
Build console output for commit: 4a24a3767b4dba69c2376f498e32832438009f1f available at: https://gist.github.com/7e9c77b7e0d349f67ba1<br/><br/>No installer was built for this commit

assignment-client/src/octree/OctreeServer.cpp
ZappoMan: interesting... now I crashed in setFinished(true) because the ThreadedAssginment's _domainServerTimer was 0xfeeefeee

birarda: commonInit is called before readConfiguration, right?

assignment-client/src/octree/OctreeQueryNode.cpp
```diff
@@ -242,7 +242,8 @@ bool OctreeQueryNode::updateCurrentViewFrustum() {
 
     if (0.0f != getCameraAspectRatio() &&
         0.0f != getCameraNearClip() &&
-        0.0f != getCameraFarClip()) {
+        0.0f != getCameraFarClip() &&
+        getCameraNearClip() != getCameraFarClip()) {
```
ZappoMan: note - glm::prespective() will assert() if near and far are the same

libraries/networking/src/ThreadedAssignment.cpp
```diff
@@ -65,7 +65,7 @@ void ThreadedAssignment::commonInit(const QString& targetName, NodeType_t nodeTy
     auto nodeList = DependencyManager::get<NodeList>();
     nodeList->setOwnerType(nodeType);
 
-    _domainServerTimer = new QTimer();
+    _domainServerTimer = new QTimer(this);
```
ZappoMan: make sure _domainServerTimer is owned so it doesn't leak

hifi-gustavo:
Build console output for commit: 54972515dc65542f57573da9881fca6088021f8d available at: https://gist.github.com/7e9c77b7e0d349f67ba1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5815.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5815.dmg

PR #5818, opened 2015-09-16
Updated left hand point and grab State Machine and animations

hyperlogic:
These are only visible when Developer > Use Anim Graph is set to true, and you are using the trigger via a RazorHydra or vive controller.


hifi-gustavo:
Build console output for commit: 70f58352786ce74bc986fbfb9021d6313b32cb9d available at: https://gist.github.com/e69c73ec39dc73653664<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5818.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5818.dmg

howard-stearns:
While I don't have hydras at home to test this, I can see by inspection that this shouldn't break anything much, so let's merge and test in production!

PR #5819, opened 2015-09-16
Fix warning and possible leak in EntitySimulation

Atlante45:
- Fix warning on OS X
- Fix some cleanup not being done when changing the entity tree


libraries/entities/src/EntitySimulation.cpp
```diff
@@ -17,11 +17,7 @@
 
 void EntitySimulation::setEntityTree(EntityTreePointer tree) {
     if (_entityTree && _entityTree != tree) {
-        _mortalEntities.clear();
-        _nextExpiry = quint64(-1);
-        _entitiesToUpdate.clear();
-        _entitiesToSort.clear();
-        _simpleKinematicEntities.clear();
```
Atlante45: I noticed some code was in `clearEntities()` but not here:
```
    _entitiesToDelete.clear();
    clearEntitiesInternal();
    for (auto entityItr : _allEntities) {
        entityItr->_simulated = false;
    }
    _allEntities.clear();
```

Was it on purpose?

libraries/entities/src/EntitySimulation.cpp
```diff
@@ -65,7 +62,7 @@ void EntitySimulation::changeEntityInternal(EntityItemPointer entity) {
 void EntitySimulation::expireMortalEntities(const quint64& now) {
     if (now > _nextExpiry) {
         // only search for expired entities if we expect to find one
-        _nextExpiry = quint64(-1);
```
Atlante45: `-1` to unsigned?
This seems dangerous, since it's a timestamp since epoch, we might as well use 0.

hifi-gustavo:
Build console output for commit: 99cfba3ba6951fde7573b51cceb2874a2070bf34 available at: https://gist.github.com/c5a10064aadc13cd39a8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5819.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5819.dmg

PR #5822, opened 2015-09-16
add back the missing JurisdictionListener to headless Agent

birarda:


assignment-client/src/Agent.cpp
```diff
@@ -191,22 +191,21 @@ void Agent::run() {
...
 
     // wire up our additional agent related processing to the update signal
-    QObject::connect(_scriptEngine, &ScriptEngine::update, this, &Agent::processAgentAvatarAndAudio);
+    QObject::connect(_scriptEngine.get(), &ScriptEngine::update, this, &Agent::processAgentAvatarAndAudio);
 
     _scriptEngine->run();
     setFinished(true);
-
-    // kill the avatar identity timer
-    delete _avatarIdentityTimer;
```
birarda: _avatarIdentityTimer is parented by the Agent, it'll go away automagically

assignment-client/src/Agent.cpp
```diff
@@ -191,22 +191,21 @@ void Agent::run() {
...
-
-    // delete the script engine
-    delete _scriptEngine;
```
birarda: don't need this with the unique_ptr

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -263,7 +263,7 @@ void ScriptEngine::init() {
     }
 
     _isInitialized = true;
-
+
```
ZappoMan: ??

libraries/octree/src/OctreeHeadlessViewer.cpp
```diff
@@ -34,8 +31,9 @@ void OctreeHeadlessViewer::init() {
 void OctreeHeadlessViewer::queryOctree() {
     char serverType = getMyNodeType();
     PacketType packetType = getMyQueryMessageType();
+    
     NodeToJurisdictionMap& jurisdictions = *_jurisdictionListener->getJurisdictions();
-
+
```
ZappoMan: ?

sethalves:
I've tested this on linux, it fixes the crash and AC scripts work again.

libraries/octree/src/OctreeHeadlessViewer.cpp
sethalves: !?

ZappoMan: why are we adding a blank line with 4 spaces on it???

libraries/octree/src/OctreeHeadlessViewer.cpp
```diff
@@ -34,8 +31,9 @@ void OctreeHeadlessViewer::init() {
 void OctreeHeadlessViewer::queryOctree() {
     char serverType = getMyNodeType();
     PacketType packetType = getMyQueryMessageType();
+
```
ZappoMan: same here - why a blank line with 4 spaces at front.

hifi-gustavo:
Build console output for commit: 5df7ce7424b8b4855a7e43f3df573726b91772a8 available at: https://gist.github.com/163da545eb2af3879515<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5822.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5822.dmg

ZappoMan:
the blank line with 4 spaces isn't a merge blocker... but it seems like it's something we want to avoid... I assume it's an editor settings issue.

hifi-gustavo:
Build console output for commit: bca00db4a8f4d0d74cc3b8e3d014333c847d1642 available at: https://gist.github.com/163da545eb2af3879515<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5822.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5822.dmg

hifi-gustavo:
Build console output for commit: 229e73f88c4444bf3f8d9812ceb061192ea96477 available at: https://gist.github.com/163da545eb2af3879515<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5822.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5822.dmg

libraries/script-engine/src/ScriptEngine.cpp
birarda: When new lines are added in Xcode they get added with the correct tabbing over - must have removed this line and added it back.

birarda:
Don't think there's been any consensus on tabbing for blank lines matching the lines surrounding them. I have noticed that some editors add these automatically (Xcode) and some don't handle indentation when adding a line properly without them. Can't see why this would be an issue.

ZappoMan:
note: I didn't think the blank line space was a merge blocker... I still don't.. But the reason I mention the white space and tabs on blank lines is because I've seen this showing up more often in many different commits... and it seems like it's polluting our diffs and merges... my gut says this is the same as line endings and tabs vs spaces and we should standardize and make sure our editors are doing the right thing...

PR #5825, opened 2015-09-16
CR for Job #20655 - Add ability to close one script from another script

thoys:
Code Review for Job #20655 - Workitem available at https://www.worklist.net/20655


hifi-gustavo:
Build console output for commit: 9992f4b5c5bbb557a05a0c7e09c35bf8127e533f available at: https://gist.github.com/dbf45325d34eb881a9ee<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5825.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5825.dmg

cozza13:
I have qa'ed this and it works

interface/src/Application.h
```diff
@@ -292,7 +292,7 @@ class Application : public QApplication, public AbstractViewStateInterface, publ
     NodeToJurisdictionMap& getEntityServerJurisdictions() { return _entityServerJurisdictions; }
 
     QStringList getRunningScripts() { return _scriptEnginesHash.keys(); }
-    ScriptEngine* getScriptEngine(QString scriptHash) { return _scriptEnginesHash.contains(scriptHash) ? _scriptEnginesHash[scriptHash] : NULL; }
+    ScriptEngine* getScriptEngine(const QString& scriptHash) { return _scriptEnginesHash.contains(scriptHash) ? _scriptEnginesHash[scriptHash] : NULL; }
```
AndrewMeadows: Not a blocker but **getScriptEngine()** does a double lookup for existing **scriptHash**.  There is a way to implement this function that only uses one lookup.

interface/src/ui/RunningScriptsWidget.cpp
```diff
@@ -57,16 +57,15 @@ RunningScriptsWidget::RunningScriptsWidget(QWidget* parent) :
...
-    connect(&_stopSignalMapper, SIGNAL(mapped(QString)), Application::getInstance(), SLOT(stopScript(const QString&)));
+    connect(ui->reloadAllButton, &QPushButton::clicked, Application::getInstance(), &Application::reloadAllScripts);
+    connect(ui->stopAllButton, &QPushButton::clicked, this, &RunningScriptsWidget::allScriptsStopped);
+    connect(ui->loadScriptFromDiskButton, &QPushButton::clicked, Application::getInstance(), &Application::loadDialog);
+    connect(ui->loadScriptFromURLButton, &QPushButton::clicked, Application::getInstance(), &Application::loadScriptURLDialog);
+    connect(&_reloadSignalMapper, static_cast<void(QSignalMapper::*)(const QString&)>(&QSignalMapper::mapped),
+            Application::getInstance(), &Application::reloadOneScript);
+
+    connect(&_stopSignalMapper, static_cast<void(QSignalMapper::*)(const QString&)>(&QSignalMapper::mapped),
+            [](const QString& script) { Application::getInstance()->stopScript(script); });
```
AndrewMeadows: I'm seeing two patterns here between **_reloadSignalMapper** and **_stopSignalMapper**.  One explicitly calls Application::getInstance() and connects to a real member of the instance whereas the other connects to a lambda which eventually calls Application::getInstance().  I'm wondering... why two patterns?  Does one of the patterns not work in the other case?

interface/src/ui/RunningScriptsWidget.cpp
```diff
@@ -294,3 +291,16 @@ QVariantList RunningScriptsWidget::getLocal() {
...
 }
+
+bool RunningScriptsWidget::stopScriptByName(const QString& name) {
+    foreach (const QString& runningScript, Application::getInstance()->getRunningScripts()) {
+        if (QUrl(runningScript).fileName().toLower() == name.trimmed().toLower()) {
+            return Application::getInstance()->stopScript(runningScript, false);
+        }
+    }
+    return false;
+}
```
AndrewMeadows: **stopScriptByName()** is a new function but is never called by the C++ code.  I take it that, since it is a **public slot** its purpose is to be invoked roundabout by JS rather than by the C++ code directly.

interface/src/ui/RunningScriptsWidget.cpp
thoys: Yes, the lambda one I didn't get to work as a straight connected signal (in the new way of signaling) , in the old style signaling it would work IIRC. That's why I took that approach.

interface/src/ui/RunningScriptsWidget.cpp
thoys: Yes, this function is exposed to JavaScript.

interface/src/Application.h
thoys: Thanks will fix

hifi-gustavo:
Build console output for commit: f38bb77d0a9fc2bb46d717e0e86f7cf8e069c86b available at: https://gist.github.com/dbf45325d34eb881a9ee<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5825.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5825.dmg

PR #5826, opened 2015-09-16
prevent initial bad IK targets for hands 

AndrewMeadows:
This PR fixes two bugs:

(1)  On login: the avatar's IK targets are above the head but get corrected as soon as the hand controllers are actually used.

(2) Sometimes the IK solution will "flutter" with small glitches (these small glitches actually corresponded to momentarily bad FPS).

The bug below is NOT yet fixed:

(1) The hand IK targets should be disabled when the hydra controllers are placed on their orb station.

### DIFF
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -211,8 +211,8 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
                                 // compute joint's new parent-relative rotation
                                 // Q' = dQ * Q   and   Q = Qp * q   -->   q' = Qp^ * dQ * Q
                                 glm::quat newRot = glm::normalize(glm::inverse(
-                                        absolutePoses[parentIndex].rot) * 
-                                        deltaRotation * 
+                                        absolutePoses[parentIndex].rot) *
+                                        deltaRotation *
                                         absolutePoses[pivotIndex].rot);
                                 RotationConstraint* constraint = getConstraint(pivotIndex);
                                 if (constraint) {
@@ -221,8 +221,8 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
                                         // the constraint will modify the movement of the tip so we have to compute the modified
                                         // model-frame deltaRotation
                                         // Q' = Qp^ * dQ * Q  -->  dQ =   Qp * Q' * Q^
-                                        deltaRotation = absolutePoses[parentIndex].rot * 
-                                            newRot * 
+                                        deltaRotation = absolutePoses[parentIndex].rot *
+                                            newRot *
                                             glm::inverse(absolutePoses[pivotIndex].rot);
                                     }
                                 }
@@ -319,7 +319,7 @@ void AnimInverseKinematics::initConstraints() {
     // We create constraints for the joints shown here
     // (and their Left counterparts if applicable).
     //
-    //                           
+    //
     //                                    O RightHand
     //                      Head         /
     //                          O       /
@@ -473,7 +473,7 @@ void AnimInverseKinematics::initConstraints() {
             const float MAX_HAND_SWING = PI / 2.0f;
             minDots.push_back(cosf(MAX_HAND_SWING));
             stConstraint->setSwingLimits(minDots);
-            
+
             constraint = static_cast<RotationConstraint*>(stConstraint);
         } else if (baseName.startsWith("Shoulder", Qt::CaseInsensitive)) {
             SwingTwistConstraint* stConstraint = new SwingTwistConstraint();
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -283,12 +283,16 @@ const AnimPoseVec& AnimInverseKinematics::overlay(const AnimVariantMap& animVars
         loadPoses(underPoses);
     } else {
         // relax toward underpose
-        const float RELAXATION_TIMESCALE = 0.125f;
-        const float alpha = glm::clamp(dt / RELAXATION_TIMESCALE, 0.0f, 1.0f);
+        // HACK: this relaxation needs to be constant per-frame rather than per-realtime
+        // in order to prevent IK "flutter" for bad FPS.  The bad news is that the good parts
+        // of this relaxation will be FPS dependent (low FPS will make the limbs align slower 
+        // in real-time), however most people will not notice this and this problem is less
+        // annoying than the flutter.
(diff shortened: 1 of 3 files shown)


libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -68,6 +82,12 @@ static int findRootJointInSkeleton(AnimSkeleton::ConstPointer skeleton, int inde
     return rootIndex;
 }
 
+struct IKTarget {
+    AnimPose pose;
+    int index;
+    int rootIndex;
+};
```
AndrewMeadows: This struct was moved here from the header file since the class no longer maintains a member list of targets -- a local list is rebuilt for each call to **evaluate()**.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -76,43 +96,53 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
...
+                IKTarget target;
                 AnimPose defaultPose = _skeleton->getAbsolutePose(targetVar.jointIndex, _relativePoses);
-                iter->second.pose.trans = animVars.lookup(targetVar.positionVar, defaultPose.trans);
-                iter->second.pose.rot = animVars.lookup(targetVar.rotationVar, defaultPose.rot);
+                target.pose.trans = animVars.lookup(targetVar.positionVar, defaultPose.trans);
+                target.pose.rot = animVars.lookup(targetVar.rotationVar, defaultPose.rot);
+                target.rootIndex = targetVar.rootIndex;
+                target.index = targetVar.jointIndex;
+                targets.push_back(target);
             }
```
AndrewMeadows: We only create a **target** if the right animVars are found.  What this means is that the target's effect is disabled if **animVars** lacks information about it.  Note this causes a rather rapid **snap** into place (or back again) whenever an IK target is is created/removed between successive frames.  Smoothing out such transitions will be part of some future PR's.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -468,6 +499,18 @@ void AnimInverseKinematics::initConstraints() {
...
+            const float MAX_SPINE_TWIST = PI / 8.0f;
+            stConstraint->setTwistLimits(-MAX_SPINE_TWIST, MAX_SPINE_TWIST);
+
+            std::vector<float> minDots;
+            const float MAX_SPINE_SWING = PI / 14.0f;
+            minDots.push_back(cosf(MAX_SPINE_SWING));
+            stConstraint->setSwingLimits(minDots);
+
+            constraint = static_cast<RotationConstraint*>(stConstraint);
         } else if (0 == baseName.compare("Neck", Qt::CaseInsensitive)) {
```
AndrewMeadows: I started adding support for the "Hips2" joint experiment, but the support here is not finished.  We'll finish it in a later PR.

hifi-gustavo:
Build console output for commit: c379093444edb51e383f3e492e6edf2d08cdf58d available at: https://gist.github.com/508eee66b05069a32f76<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5826.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5826.dmg

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -248,12 +283,16 @@ const AnimPoseVec& AnimInverseKinematics::overlay(const AnimVariantMap& animVars
...
+        // of this relaxation will be FPS dependent (low FPS will make the limbs align slower 
+        // in real-time), however most people will not notice this and this problem is less
+        // annoying than the flutter.
+        const float blend = (1.0f / 60.0f) / (0.25f); // effectively: dt / RELAXATION_TIMESCALE
         int numJoints = (int)_relativePoses.size();
         for (int i = 0; i < numJoints; ++i) {
             float dotSign = copysignf(1.0f, glm::dot(_relativePoses[i].rot, underPoses[i].rot));
-            _relativePoses[i].rot = glm::normalize(glm::lerp(_relativePoses[i].rot, dotSign * underPoses[i].rot, alpha));
+            _relativePoses[i].rot = glm::normalize(glm::lerp(_relativePoses[i].rot, dotSign * underPoses[i].rot, blend));
         }
```
AndrewMeadows: These changes prevent the small glitchy behavior when the IK solution varies about simple steady-state solutions.  It was triggered by momentary bad FPS (variations in **dt** between successive calls to this codepath).

hifi-gustavo:
Build console output for commit: c3be4b97fa986fedecb1b795d5f76aaf09b12320 available at: https://gist.github.com/508eee66b05069a32f76<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5826.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5826.dmg

hifi-gustavo:
Build console output for commit: e1fc1900ab08af7418417c50fb2931a02558c6ed available at: https://gist.github.com/508eee66b05069a32f76<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5826.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5826.dmg

libraries/animation/src/AnimVariant.h
```diff
@@ -154,6 +154,8 @@ class AnimVariantMap {
     void setTrigger(const std::string& key) { _triggers.insert(key); }
     void clearTriggers() { _triggers.clear(); }
 
+    bool hasKey(const std::string& key) const { return _map.find(key) != _map.end(); }
```
howard-stearns: Why not just _map.contains(key) ?

DaveDubUK:
@AndrewMeadows - 'momentarily bad FPS' - possibly related to: https://worklist.net/20680 ?
(very short worklist description of MyAvatar.getJointPosition calls taking more than 16mS)

libraries/animation/src/AnimVariant.h
AndrewMeadows: Because I had forgotten that aspect of the std::map API.  I'll fix it in another PR.

AndrewMeadows:
@DaveDubUK - I could believe that.  **MyAvatar.getJointPosition()** must get its answer from a separate and very busy thread -- the script thread will block.  We'll investigate and try to come up with a way to speed things up.  Nevertheless for the record: MyAvatar.getJointPosition() feature has little to do with the work in this PR, except that the IK calculations will influence the final data that is obtained.

PR #5832, opened 2015-09-17
Fix the lighting issue when the Rear View mirror is enabled

samcake:
- We have the issue that the uniform buffer when bound to the COntext cannot be updated correctly by the sync / version mechanism
 I will need to look at that into more details

- For now just unbind the "deferredTransformBuffer" at the end of the deferred pass and we are good

- added a simple check on the BufferView so we can initialize it with nullptr for the buffer

### DIFF
--- a/libraries/gpu/src/gpu/Resource.h
+++ b/libraries/gpu/src/gpu/Resource.h
@@ -160,15 +160,22 @@ typedef std::vector< BufferPointer > Buffers;
 
 
 class BufferView {
+protected: 
+    void initFromBuffer(const BufferPointer& buffer) {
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: b207d97f72c31c46c395021dcdefcdddcdfd0954 available at: https://gist.github.com/ff7476f3e91462bbc432<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5832.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5832.dmg

PR #5837, opened 2015-09-18
blend rotations between distinct IK end effector solutions

AndrewMeadows:
This PR fixes most of the "locked shoulders" problem where movement of the avatar hands has zero effect on the rotation of the shoulders.  As we iterate over the skeleton for each IK target we solve the rotations without moving the skeleton and accumulate all of the rotation changes for each joint.  Then once we've made a pass for all of the constraints we average the final results.  Wash, rinse, repeat.  The final effect is to give the IK targets equal influence over the joints further up the hierarchy.

Note: the head still has a translation target for screener mode so the spine doesn't move much for this case, however once we remove the translation aspect of that target the head and shoulders should move around fine.  I tested this disabling the third (head) target but didn't want to commit that hack since I suspect it the code is still necessary for passing the head translation for HMD mode.

### DIFF
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -262,12 +262,10 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
             }
 
             // only update the absolutePoses that need it: those between lowestMovedIndex and _maxTargetIndex
-            if (lowestMovedIndex < _maxTargetIndex) {
-                for (int i = lowestMovedIndex; i < _maxTargetIndex; ++i) {
-                    int parentIndex = _skeleton->getParentIndex(i);
-                    if (parentIndex != -1) {
-                        absolutePoses[i] = absolutePoses[parentIndex] * _relativePoses[i];
-                    }
+            for (int i = lowestMovedIndex; i <= _maxTargetIndex; ++i) {
+                int parentIndex = _skeleton->getParentIndex(i);
+                if (parentIndex != -1) {
+                    absolutePoses[i] = absolutePoses[parentIndex] * _relativePoses[i];
                 }
             }
         } while (largestError > ACCEPTABLE_RELATIVE_ERROR && numLoops < MAX_IK_LOOPS && usecTimestampNow() < expiry);
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -153,6 +153,11 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
             ++constraintItr;
         }
     } else {
+        // clear the accumulators before we start the IK solver
+        for (auto& accumulatorPair: _accumulators) {
+            accumulatorPair.second.clear();
+        }
+
         // compute absolute poses that correspond to relative target poses
         AnimPoseVec absolutePoses;
         computeAbsolutePoses(absolutePoses);
@@ -165,8 +170,8 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
         quint64 expiry = usecTimestampNow() + MAX_IK_TIME;
         do {
             largestError = 0.0f;
+            int lowestMovedIndex = _relativePoses.size();
             for (auto& target: targets) {
-                int lowestMovedIndex = _relativePoses.size() - 1;
                 int tipIndex = target.index;
                 AnimPose targetPose = target.pose;
                 int rootIndex = target.rootIndex;
@@ -226,7 +231,8 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
                                             glm::inverse(absolutePoses[pivotIndex].rot);
                                     }
                                 }
-                                _relativePoses[pivotIndex].rot = newRot;
+                                // store the rotation change in the accumulator
+                                _accumulators[pivotIndex].add(newRot);
                             }
                             // this joint has been changed so we check to see if it has the lowest index
                             if (pivotIndex < lowestMovedIndex) {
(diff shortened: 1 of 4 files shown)


libraries/animation/src/RotationAccumulator.h
```diff
@@ -0,0 +1,34 @@
...
+
+    void add(const glm::quat& rotation) { _rotations.push_back(rotation); }
+
+    glm::quat getAverage();
+
+    void clear() { _rotations.clear(); }
+
+private:
+    std::vector<glm::quat> _rotations;
+};
```
AndrewMeadows: The **RotationAccumulator** stores a vector of rotations and then can supply their average.

hifi-gustavo:
Build console output for commit: 420acde72004d7299dd9a2c6739dbe54871b237a available at: https://gist.github.com/f95546a2982e0277aa3c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5837.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5837.dmg

libraries/animation/src/AnimInverseKinematics.h
```diff
@@ -60,6 +64,7 @@ class AnimInverseKinematics : public AnimNode {
     };
 
     std::map<int, RotationConstraint*> _constraints;
+    std::map<int, RotationAccumulator> _accumulators;
```
AndrewMeadows: **AnimInverseKinematics** gets a map of RotationAccumulators that it uses to remember the rotation changes from each target.  After the partial solution of all of the targets the skeleton is adjusted according to the average rotations and the iterations start anew.

hifi-gustavo:
Build console output for commit: b6a153d92690a9a2b19f1adfd23ffb61d85b83c1 available at: https://gist.github.com/f95546a2982e0277aa3c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5837.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5837.dmg

hifi-gustavo:
Build console output for commit: 9e94e7f1d0f6f96952eb1c98afde7db9dc6e9223 available at: https://gist.github.com/f95546a2982e0277aa3c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5837.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5837.dmg

libraries/animation/src/RotationAccumulator.cpp
```diff
@@ -0,0 +1,28 @@
...
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include "RotationAccumulator.h"
+
+#include <glm/gtx/quaternion.hpp>
+
+void RotationAccumulator::add(glm::quat rotation) {
+    if (_numRotations == 0) {
```
howard-stearns: If we initialize and clear to zero, we don't need this separate branch here (i.e., always += rotation).

libraries/animation/src/RotationAccumulator.cpp
```diff
@@ -0,0 +1,28 @@
...
+        _rotationSum = rotation;
+    } else {
+        if (glm::dot(_rotationSum, rotation) < 0.0f) {
```
howard-stearns: I don't understand what this is doing. Naively, it seems like if one side said rotate 30 degrees and the other said -30, we want the average to be zero, not 60.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -243,36 +249,49 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
...
+                AnimPose targetPose = target.pose;
+                // compute tip's new parent-relative rotation
+                // Q = Qp * q   -->   q' = Qp^ * Q
+                glm::quat newRelativeRotation = glm::inverse(absolutePoses[parentIndex].rot) * targetPose.rot;
+                RotationConstraint* constraint = getConstraint(tipIndex);
+                if (constraint) {
+                    constraint->apply(newRelativeRotation);
+                    // TODO: ATM the final rotation target just fails but we need to provide
+                    // feedback to the IK system so that it can adjust the bones up the skeleton
+                    // to help this rotation target get met.
```
howard-stearns: I realize this is not a changed in this PR, but asking here just because I noticed:
Does all this mean that the IK iteration is only about meeting the positionVar? I.,e that the the rotationVar is just matched (subject to constraints) at the end, without actually influencing the iterations? I assume the consequence then is a big bend at the wrist or neck would not influence the solution further up the chain, and so, for example, rotating the head target will not rotate the shoulders, etc.?

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -153,6 +153,11 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
             ++constraintItr;
         }
     } else {
+        // clear the accumulators before we start the IK solver
+        for (auto& accumulatorPair: _accumulators) {
+            accumulatorPair.second.clear();
+        }
+
```
howard-stearns: Opinion outside the scope of just this PR: AnimInverseKinematics::evaluate() is a nice elegant piece of code, will become a key part of our company value, and will have stuff added to it by multiple people over time. To that end, it would be cool to work towards separating out some of the pieces into separate tiny functions, allowing the main functionality to stand out more. (Trying to channel K&R here...)

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -165,8 +170,8 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
         quint64 expiry = usecTimestampNow() + MAX_IK_TIME;
         do {
             largestError = 0.0f;
+            int lowestMovedIndex = _relativePoses.size();
```
howard-stearns: I didn't follow the fencepost difference here vs the -1 in previous line 169.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -243,36 +249,49 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
...
-                if (lowestMovedIndex <= _maxTargetIndex && lowestMovedIndex < tipIndex) {
-                    // only update the absolutePoses that matter: those between lowestMovedIndex and _maxTargetIndex
-                    for (int i = lowestMovedIndex; i <= _maxTargetIndex; ++i) {
-                        int parentIndex = _skeleton->getParentIndex(i);
-                        if (parentIndex != -1) {
-                            absolutePoses[i] = absolutePoses[parentIndex] * _relativePoses[i];
-                        }
-                    }
+            // harvest accumulated rotations and apply the average
+            for (auto& accumulatorPair: _accumulators) {
```
howard-stearns: (Ack out of time on the train, so rushing this...) Are we using the average for all the moved joints? I thought we would use the average only for the joints in common between the chains associated with multiple targets. In other words, the arms and head would not use a rotation averaged with zero-rotation influences from the other side, but rather the arms and head would get the full rotation based on their target.

libraries/animation/src/AnimInverseKinematics.cpp
AndrewMeadows: That change was a very minor optimization -- it avoids a case where **lowestMovedIndex == _maxTargetIndex**, which doesn't need any absolute transform computed because each **targetIndex** will have its absolute rotation set at the very end -- where there is a comment that looks like this:

    // finally set the relative rotation of each tip to agree with absolute target rotation

libraries/animation/src/AnimInverseKinematics.cpp
AndrewMeadows: Yes, we're using the average for all joints because the algorithm is simpler if we don't have to distinguish between the two varieties.  Any joint that gets its rotation set only once will have an average that is equal to that rotation so it all works out.

libraries/animation/src/AnimInverseKinematics.cpp
AndrewMeadows: Yes, that is correct: when the final target's rotationVar is not achieved at this point (because of constraint) there is no consequence further up the chain, which is what the **TODO** comment here is about.  That up-chain rotation consequence seemed complicated so I'm punting that work for later.

libraries/animation/src/RotationAccumulator.cpp
AndrewMeadows: Hrm... yes you're right.  I guess that didn't occur to me because an all-zero quaternion is invalid, but it works because of  how the **getAverage()** method is implemented.

libraries/animation/src/RotationAccumulator.cpp
AndrewMeadows: What's happening here is that quaternions have a discrepancy of two for all rotations: each rotation has two valid corresponding quaterions: Q and -Q.  When linearly adding quaternions we always want to make sure they are both on the same "hyper-hemisphere".  Which means: when the dot-product between the two quaternions is negative we want to negate one of them (to be on the same hyper-hemisphere) before we add.

howard-stearns: Cool. I did not know that! 
(Add a comment IFF you make other changes?)

AndrewMeadows:
BTW, these last few changes were made without testing (I'm in a coffee shop and didn't bring a hydra with me).  The final result should be tested before it is merged.

hifi-gustavo:
Build console output for commit: 4cb2249cda69af25bd8121e33a8043776ddeebc6 available at: https://gist.github.com/f95546a2982e0277aa3c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5837.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5837.dmg

howard-stearns:
Something has gone very wrong. The hydra rotation is tracked, but not the position. (T-Rex hands.)

hifi-gustavo:
Build console output for commit: 07f3abfc911a83a5f2562f8019ab5690835194f9 available at: https://gist.github.com/f95546a2982e0277aa3c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5837.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5837.dmg

howard-stearns:
yay!

PR #5838, opened 2015-09-18
rework handControllerGrab.js

sethalves:
- rework handControllerGrab.js

### DIFF
--- a/examples/controllers/handControllerGrab.js
+++ b/examples/controllers/handControllerGrab.js
@@ -58,9 +58,6 @@ var ZERO_VEC = {x: 0, y: 0, z: 0};
 var NULL_ACTION_ID = "{00000000-0000-0000-000000000000}";
 var MSEC_PER_SEC = 1000.0;
 
-var rightController = new controller(RIGHT_HAND, Controller.findAction("RIGHT_HAND_CLICK"));
-var leftController = new controller(LEFT_HAND, Controller.findAction("LEFT_HAND_CLICK"));
-
 // these control how long an abandoned pointer line will hang around
 var startTime = Date.now();
 var LIFETIME = 10;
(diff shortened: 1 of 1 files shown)


ZappoMan:
Coding Standard: Magic Numbers

examples/controllers/handControllerGrab.js
```diff
@@ -83,309 +83,284 @@ function controller(side, triggerAction, pullAction, hand) {
...
+            self.state = STATE_DISTANCE_HOLDING;
+            lineOn(self, pickRay.origin, Vec3.multiply(pickRay.direction, LINE_LENGTH), NO_INTERSECT_COLOR);
+        }
+    } else {
+        // forward ray test failed, try sphere test.
+        var nearbyEntities = Entities.findEntities(handPosition, GRAB_RADIUS);
+        var minDistance = GRAB_RADIUS;
+        var grabbedEntity = null;
+        for (var i = 0; i < nearbyEntities.length; i++) {
+            var props = Entities.getEntityProperties(nearbyEntities[i]);
```
ZappoMan: use the new faster Entities.getEntityProperties(); only ask for the properties you want, it will be faster by about 80%.

```
 var props = Entities.getEntityProperties(nearbyEntities[i], ["position","name","locked"]);
```

examples/controllers/handControllerGrab.js
```diff
@@ -83,309 +83,284 @@ function controller(side, triggerAction, pullAction, hand) {
...
-
-
-            Entities.updateAction(this.grabbedEntity, this.actionID, {
-                targetPosition: this.currentObjectPosition, linearTimeScale: .1,
-                targetRotation: this.currentObjectRotation, angularTimeScale: .1
-            });
+    var handPosition = self.getHandPosition();
+    var handControllerPosition = Controller.getSpatialControlPosition(self.palm);
+    var handRotation = Quat.multiply(MyAvatar.orientation, Controller.getSpatialControlRawRotation(self.palm));
+    var grabbedProperties = Entities.getEntityProperties(self.grabbedEntity);
```
ZappoMan: use the new faster Entities.getEntityProperties(); only ask for the properties you want, it will be faster by about 80%.

```
var grabbedProperties = Entities.getEntityProperties(self.grabbedEntity, ["position","rotation"]);
```

examples/controllers/handControllerGrab.js
```diff
@@ -83,309 +83,284 @@ function controller(side, triggerAction, pullAction, hand) {
...
+    if (!triggerSmoothedSqueezed(self)) {
+        self.state = STATE_RELEASE;
+        return;
+    }
 
-    var objectPosition = Entities.getEntityProperties(this.grabbedEntity).position;
-    var offset = Vec3.subtract(objectPosition, handPosition);
-    var offsetPosition = Vec3.multiplyQbyV(Quat.inverse(Quat.multiply(handRotation, offsetRotation)), offset);
+    // keep track of the measured velocity of the held object
+    var grabbedProperties = Entities.getEntityProperties(self.grabbedEntity);
```
ZappoMan: use the new faster Entities.getEntityProperties(); only ask for the properties you want, it will be faster by about 80%.

```
var grabbedProperties = Entities.getEntityProperties(self.grabbedEntity, "position");
```

ZappoMan:
I don't understand why this code constantly uses the following construct:

```
function foo(self) {
    spam(self);
    self.bah = 10;
    self.balz = 1;
}
```

Instead the foo function should be part of the original controller prototype, and then the function would have a legitimate `this` and would be able to just access items directly.

This should be cleaned up.

ZappoMan:
I also think we should use the following style for defining classes in JavaScript....

```
   // this is the "constructor"
    controller = controller(hand, triggerAction) { 
        this.hand = hand;
        ...
    };

    controiller.prototype = {

        update: function() { ... },
        lineOn: function(closePoint, farPoint, color) { ... },
        ...
    };
```

This provides a cleaner description of the class nature of the code and hides the functions from misuse.

PR #5844, opened 2015-09-18
Cruft removal machine

ZappoMan:
Clean up how Application::displaySide() works:
* remove the EntityTreeRenderer::render() since all rendering is now done as true render items.
* move the best zone logic out of render and into update where the other check enter/leave entities was happening.
* remove the old renderingInWorldInterface signal and one user of it
* change GlobalServicesScriptingInterface to use a timer instead of renderingInWorldInterface

### DIFF
--- a/interface/src/scripting/GlobalServicesScriptingInterface.cpp
+++ b/interface/src/scripting/GlobalServicesScriptingInterface.cpp
@@ -22,8 +22,10 @@ GlobalServicesScriptingInterface::GlobalServicesScriptingInterface() {
     connect(&accountManager, &AccountManager::logoutComplete, this, &GlobalServicesScriptingInterface::loggedOut);
 
     _downloading = false;
-    connect(Application::getInstance(), &Application::renderingInWorldInterface, 
-            this, &GlobalServicesScriptingInterface::checkDownloadInfo);
+    QTimer* checkDownloadTimer = new QTimer(this);
+    connect(checkDownloadTimer, &QTimer::timeout, this, &GlobalServicesScriptingInterface::checkDownloadInfo);
+    const int CHECK_DOWNLOAD_INTERVAL = MSECS_PER_SECOND / 2;
+    checkDownloadTimer->start(CHECK_DOWNLOAD_INTERVAL);
 
     auto discoverabilityManager = DependencyManager::get<DiscoverabilityManager>();
     connect(discoverabilityManager.data(), &DiscoverabilityManager::discoverabilityModeChanged,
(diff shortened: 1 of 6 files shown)


hifi-gustavo:
Build console output for commit: 416acb1d4ad171bf2edb84d7f08fce09795b9ef2 available at: https://gist.github.com/a6e745b970003ada68fd<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5844.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5844.dmg

interface/src/scripting/GlobalServicesScriptingInterface.cpp
```diff
@@ -22,8 +22,10 @@ GlobalServicesScriptingInterface::GlobalServicesScriptingInterface() {
     connect(&accountManager, &AccountManager::logoutComplete, this, &GlobalServicesScriptingInterface::loggedOut);
 
     _downloading = false;
-    connect(Application::getInstance(), &Application::renderingInWorldInterface, 
-            this, &GlobalServicesScriptingInterface::checkDownloadInfo);
+    QTimer* checkDownloadTimer = new QTimer(this);
```
jherico: previously this logic would only get called when not rendering `!selfAvatarOnly`.  Is it important that this is no longer the case?

libraries/entities-renderer/src/EntityTreeRenderer.h
```diff
@@ -40,7 +40,7 @@ class EntityTreeRenderer : public OctreeRenderer, public EntityItemFBXService {
     virtual char getMyNodeType() const { return NodeType::EntityServer; }
     virtual PacketType getMyQueryMessageType() const { return PacketType::EntityQuery; }
     virtual PacketType getExpectedPacketType() const { return PacketType::EntityData; }
-    virtual void renderElement(OctreeElementPointer element, RenderArgs* args);
+    virtual void renderElement(OctreeElementPointer element, RenderArgs* args) { }
```
jherico: Why retain the function?  Why not remove the pure virtual from the parent class?

interface/src/scripting/GlobalServicesScriptingInterface.cpp
ZappoMan: the old code was decidedly poorly done... it was basically calling this function every render frame... 60hz??? why... the downloads can't possibly change that frequently... instead this new method is checked every half second.

libraries/entities-renderer/src/EntityTreeRenderer.h
ZappoMan: There's a more extensive work item to come back in here and rename these classes since they are not actually renderers anymore... but I'm not prepared to totally do that work right now. I will make the renderElement() not PURE in the base class so that we don't have to implement here.

hifi-gustavo:
Build console output for commit: 020fb25ace3dfb179fbdddd8aced64d5521a15eb available at: https://gist.github.com/a6e745b970003ada68fd<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5844.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5844.dmg

PR #5846, opened 2015-09-18
Display the face blend coefficients and update the value in real time.

bob-long:
Display the face blend coefficients and update the value in real time.

- Expose the face model blend coefficients from Interface.

- A sample js script is added to read the coefficient data from Interface. Currently the total number of available coefficients is not being used by this script. A slider is used to select which coefficient to display. the coefficient is being periodically updated by an interval timer. A menu item is added under Developer> Avatar to control the show/hide the coefficients in the scene.

### DIFF
--- a/libraries/render-utils/src/Model.h
+++ b/libraries/render-utils/src/Model.h
@@ -191,6 +191,11 @@ public:
     const std::unordered_set<int>& getCauterizeBoneSet() const { return _cauterizeBoneSet; }
     void setCauterizeBoneSet(const std::unordered_set<int>& boneSet) { _cauterizeBoneSet = boneSet; }
 
+    int getBlendshapeCoefficientsNum() const { return _blendshapeCoefficients.size(); }
+    float getBlendshapeCoefficient(int index) const {
+        return index >= _blendshapeCoefficients.size() || index < 0 ?
+               0.0f : _blendshapeCoefficients.at(index); }
+
 protected:
 
     void setPupilDilation(float dilation) { _pupilDilation = dilation; }
--- a/libraries/render-utils/src/Model.h
+++ b/libraries/render-utils/src/Model.h
@@ -192,9 +192,9 @@ public:
     void setCauterizeBoneSet(const std::unordered_set<int>& boneSet) { _cauterizeBoneSet = boneSet; }
 
     int getBlendshapeCoefficientsNum() const { return _blendshapeCoefficients.size(); }
-    float getBlendshapeCoefficient(int index) const {
-        return index >= _blendshapeCoefficients.size() || index < 0 ?
-               0.0f : _blendshapeCoefficients.at(index); }
+    float getBlendshapeCoefficient(unsigned int index) const {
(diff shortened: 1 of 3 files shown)


hifi-gustavo:

ZappoMan:
ok to test

hifi-gustavo:
Build console output for commit: efc321bc1053d5bd53cdc48ee199b61d778bd701 available at: https://gist.github.com/d440b3dfcea4947f50b9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5846.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5846.dmg

libraries/render-utils/src/Model.h
```diff
@@ -191,6 +191,11 @@ class Model : public QObject {
     const std::unordered_set<int>& getCauterizeBoneSet() const { return _cauterizeBoneSet; }
     void setCauterizeBoneSet(const std::unordered_set<int>& boneSet) { _cauterizeBoneSet = boneSet; }
 
+    int getBlendshapeCoefficientsNum() const { return _blendshapeCoefficients.size(); }
+    float getBlendshapeCoefficient(int index) const {
```
ZappoMan: coding standard: in general we discourage including this much code in an inline function in a header, if the code can fit on one line (likely in this case) then it's acceptable. So if you could reformat this to fit on a single line that would be great. (note: max line length is 128 char)

bob-long: Thanks for the comments.

The code was initially just 1 line and I placed the check in the the script. However, later during testing I realized that was a bad idea for two reasons:

1) You cannot expect the script to always enforce this for the Interface, the Interface should be the line of defense.
2) The number of valid coefficients do change underneath and it's not easy/practical for script to synchronously query the total number of coefficients and enforce it for the Interface when querying an individual one.

So I ended up moving the check inside the Interface and got what we see right now. I will see if I can reformat it to a single line, otherwise I will move the implementation to the source file instead of inline it in the header.

ZappoMan: I agree the C++ should have a guard against bad indices... I was just commenting on the code matching the coding standard.

hifi-gustavo:
Build console output for commit: a23a90bf5f99dbf4b1d7c2e4ed108b97b667e4a8 available at: https://gist.github.com/d440b3dfcea4947f50b9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5846.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5846.dmg

libraries/render-utils/src/Model.h
bob-long: agreed of both.

I made some changes to the code so the line is shorter and can be properly inlined. I also fixed a typo in the file header name. the change has been committed to my forked version. You should be able to see it down below.

PR #5851, opened 2015-09-18
Flashlight updates

imgntn:
This commit updates the flashlight:
- Don't automatically orient in hand
- Don't automatically reset position
- If user releases trigger below threshold, turn light off
- If user pulls trigger above threshold, turn light on
- Adds a bulb to the end of the flashlight


imgntn:
Please test! :)

PhilipRosedale:
Sometimes when I grab it with my hand, it still shifts it's position a little bit.

PhilipRosedale:
I am noticing that the error when grabbing (I think) results from the object drifting to a stop and being at a different place (I think) on client versus server.  @AndrewMeadows this is something we should look at together and chat about tomorrow.

PR #5863, opened 2015-09-21
fix a couple of properties

ZappoMan:
this fixes an issue with particle effects losing their texture

### DIFF
--- a/libraries/entities/src/EntityItemProperties.cpp
+++ b/libraries/entities/src/EntityItemProperties.cpp
@@ -498,18 +498,21 @@ QScriptValue EntityItemProperties::copyToScriptValue(QScriptEngine* engine, bool
     // Models only
     if (_type == EntityTypes::Model) {
         COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_MODEL_URL, modelURL);
-        COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_COMPOUND_SHAPE_URL, compoundShapeURL);
         COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_ANIMATION_URL, animationURL);
-        COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_TEXTURES, textures);
     }
-
 
     if (_type == EntityTypes::Model || _type == EntityTypes::Zone || _type == EntityTypes::ParticleEffect) {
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: b88c8e507b56985d34da33774839c88691563d3f available at: https://gist.github.com/ac8b1f2c5ba1909669cc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5863.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5863.dmg

ZappoMan:
test this please

hifi-gustavo:
Build console output for commit: b88c8e507b56985d34da33774839c88691563d3f available at: https://gist.github.com/ac8b1f2c5ba1909669cc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5863.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5863.dmg

libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -498,18 +498,21 @@ QScriptValue EntityItemProperties::copyToScriptValue(QScriptEngine* engine, bool
...
-        COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_TEXTURES, textures);
     }
 
-
     if (_type == EntityTypes::Model || _type == EntityTypes::Zone || _type == EntityTypes::ParticleEffect) {
         COPY_PROPERTY_TO_QSCRIPTVALUE_GETTER(PROP_SHAPE_TYPE, shapeType, getShapeTypeAsString());
     }
 
+    // FIXME - it seems like ParticleEffect should also support this
+    if (_type == EntityTypes::Model || _type == EntityTypes::Zone) {
```
jherico: this approach will potentially lead to an N! number of different tests as more entities are created that share properties.  

In the longer term it would be wiser to either have each entity type copy its available properties, regardless of whether this results in duplicated code.  Alternatively, we could have a master map of entity type to set of properties, so that we'd just fetch the appropriate set and execute the copy property operation for the listed properties.  Not suggesting we do that now though.

ZappoMan: I've been planning on moving to a map of types to properties...I think it would significantly simplify the code... but yes... that's for a more complex PR.

PR #5869, opened 2015-09-22
cleanup IK implementation

AndrewMeadows:
(1) Split **AnimInverseKinematics::evaluate()** into several protected methods.

(2) Put IK targets in **model-frame** instead of **root-frame**.

(3) Split **AnimNode** implementation into two files.

### DIFF
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -82,22 +82,8 @@ static int findRootJointInSkeleton(AnimSkeleton::ConstPointer skeleton, int inde
     return rootIndex;
 }
 
-struct IKTarget {
-    AnimPose pose;
-    int index;
-    int rootIndex;
-};
-
-//virtual
-const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVars, float dt, AnimNode::Triggers& triggersOut) {
-
-    // NOTE: we assume that _relativePoses are up to date (e.g. loadPoses() was just called)
-    if (_relativePoses.empty()) {
-        return _relativePoses;
-    }
-
-    // build a list of targets from _targetVarVec
-    std::vector<IKTarget> targets;
+void AnimInverseKinematics::computeTargets(const AnimVariantMap& animVars, std::vector<IKTarget>& targets) {
+    // build a list of valid targets from _targetVarVec and animVars
     bool removeUnfoundJoints = false;
     for (auto& targetVar : _targetVarVec) {
         if (targetVar.jointIndex == -1) {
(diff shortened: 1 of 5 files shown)


libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -82,22 +82,8 @@ static int findRootJointInSkeleton(AnimSkeleton::ConstPointer skeleton, int inde
     return rootIndex;
 }
 
-struct IKTarget {
-    AnimPose pose;
-    int index;
-    int rootIndex;
-};
```
AndrewMeadows: **IKTarget** struct moved to the **AnimInverseKinematics.h** file

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -141,155 +127,161 @@ const AnimPoseVec& AnimInverseKinematics::evaluate(const AnimVariantMap& animVar
...
-            for (auto& target: targets) {
-                int tipIndex = target.index;
-                AnimPose targetPose = target.pose;
-                int rootIndex = target.rootIndex;
-                if (rootIndex != -1) {
-                    // transform targetPose into skeleton's absolute frame
-                    AnimPose& rootPose = _relativePoses[rootIndex];
-                    targetPose.trans = rootPose.trans + rootPose.rot * targetPose.trans;
-                    targetPose.rot = rootPose.rot * targetPose.rot;
-                }
```
AndrewMeadows: This code moved here from **evaluate()** mostly unchanged except for one thing... the shifting of **targetPose** into the **model-frame** is no longer done here.  Instead it is done before the **animVars** are submitted to **evaluate()**.

libraries/animation/src/AnimNode.h
```diff
@@ -87,19 +75,14 @@ class AnimNode {
...
+    void setCurrentFrame(float frame);
 
     virtual void setCurrentFrameInternal(float frame) {}
-    virtual void setSkeletonInternal(AnimSkeleton::ConstPointer skeleton) {
-        _skeleton = skeleton;
-    }
+    virtual void setSkeletonInternal(AnimSkeleton::ConstPointer skeleton) { _skeleton = skeleton; }
 
     // for AnimDebugDraw rendering
     virtual const AnimPoseVec& getPosesInternal() const = 0;
```
AndrewMeadows: **AnimNode.h** is mostly unchanged except the implementation of its multi-line methods moved to its **AnimNode.cpp** file.  However, the **getRootPose()** method is **NEW**!  It is used to shift things into the model-frame in **Rig::updateNeckJoint()** and **Rig::updateFromHandParameters()** below.

hifi-gustavo:
Build console output for commit: 3869887610e10e69ecfb7d2368448905f9b4776b available at: https://gist.github.com/da07a626bc750ee6b87e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5869.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5869.dmg

howard-stearns:
I'll want to check with hydra tomorrow, but this appears to work correctly. (We'll also want to give @hyperlogic  the opportunity to review.)

When the new AnimInverseKinematics::computeTargets is modified to unequivocally keep all targets, the animation does appear to do a nice job of following the hands (without being in the air) and also the feet (when I add them as targets using, e.g., PR#5862). Yay!

The bad news, though, is that we still have the damping, so there's some other problem (not with this PR, but with our theory that the damping can be fixed by tracking unset positionVars).

PhilipRosedale:
I'll give it a try

howard-stearns:
@PhilipRosedale Hopefully, you won't see any difference from what you're used to. (The key is what it allows us to do next.)

howard-stearns:
@AndrewMeadows it looks like AnimInverseKinematics::relaxTowardDefaults is still unused, correct?

PhilipRosedale:
Ah, I see.  I had thought from your comments that the positioning of the body was going to be different.  
Yes it seems to work the same for me.

AndrewMeadows:
Yes, **relaxTowardDefaults()** is unused.  I originally used it for IK testing.  I can remove it now.

Yes, the relaxation (toward underpose) step in the IK is still happening on all joints (even those that are not affected by IK) and is what is causing the damping of the underpose.  I can fix that today without too much trouble.

libraries/animation/src/Rig.cpp
```diff
@@ -995,7 +995,9 @@ void Rig::updateNeckJoint(int index, const HeadParameters& params) {
 
             // There's a theory that when not in hmd, we should _animVars.unset("headPosition").
             // However, until that works well, let's always request head be positioned where requested by hmd, camera, or default.
-            _animVars.set("headPosition", params.localHeadPosition);
+            int headIndex = _animSkeleton->nameToJointIndex("Head");
+            AnimPose rootPose = _animNode->getRootPose(headIndex);
+            _animVars.set("headPosition", rootPose.trans + rootPose.rot * params.localHeadPosition);
```
howard-stearns: I'm not sure why, but experimentally, it looks like the (rootPose.rot *) part is wrong.
As is, the head gets tiled forward and down (when not in hmd), because we are forcing it to a slightly bad position. (Separately, I'll restore the code that unsets this when not in hmd.)

However, it also tilts wildly when walking forward, and always tilts to the left when rotating in either direction. If you remove the (rootPose.rot *), then it still tilts forward as expected, but the wild sidewise tilt goes away. I suspect that the head data is already adjusting for the rotation, and multiplying again gives us too much.

howard-stearns:
This is ready to merge, but there are a couple of minor things that we would like to do at the same time.
@AndrewMeadows and I discussed, and we decided that rather than coordinating the merge of two separate co-dependent PRs, we would close this one and produce one combined PR: https://github.com/highfidelity/hifi/pull/5874

PR #5872, opened 2015-09-22
New low-latency sample rate conversion library

kencooke:
added new low-latency sample rate conversion library, intended to replace SOXR


hifi-gustavo:
Build console output for commit: 2d1150115f2dee25c0ad5b57e389a9168f0a8d6b available at: https://gist.github.com/df7e07f94f3fb91f9e15<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5872.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5872.dmg

hifi-gustavo:
Build console output for commit: b983ba7056908af746e5fa9e8fd711aacc450245 available at: https://gist.github.com/df7e07f94f3fb91f9e15<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5872.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5872.dmg

ZappoMan:
I couldn't leave a comment inline because github decided not to show the CPP file, so I'll leave some comments here...

* coding standard has spaces around operators, couple places to fix would be: line 251: `float f = (float)input[_numChannels*i + j];`, line 281, 252, 271, 375
* you're using `32768.0f`, `65536.0f` and other constants in several locations, make those named static constants.

AndrewMeadows:
You're using K&R formatting inside the function but not for its outer block like this:
[code]void doStuff() 
{
    if (true) {
        // do stuff
    }
}[/code]
The coding standard section 4.1.4 says that method functions should use K&R formatting, so it should look like:
[code]void doStuff() {
    if (true) {
        // do stuff
    }
}[/code]
https://docs.highfidelity.com/docs/coding-standard

The coding standard is trying to encourage us to use human-readable names which is a good idea most of the time.  However, anyone who can make sense of the rest of this code would recognize numbers that relate to 2^N (or 2^N - 1) so I'm not bothered by 32768.0f (or 32767.0f).

A question about optimization: Will most compilers introduce a branch for this line or will they optimize it out?
[code]f += (f < 0.0f ? -0.5f : +0.5f);[/code]

kencooke:
I prefer obvious constants like 32768.0f over complicated expressions like (float)((1 << ((8 * sizeof(int16_t)) - 1))) because I like to keep simple things simple.  Funny enough, the audio code is full of bugs where someone thought 2^15 and coded as (2 << ((8 * sizeof(int16_t)) - 1)) which is actually wrong!

MSVS will sadly produce branchy code for f += (f < 0.0f ? -0.5f : +0.5f).  The Intel compiler is more likely to recognize idioms like this and produce fast code, which in this case is to generate a CVTSS2SI instruction which rounds instead of truncates to int.

Since we use MSVC, I will probably have to manually optimize this library with intrinsics.

ZappoMan:
@kencooke - wrt to the use of 32768.0f in this code... I believe the intent here is a float version of max of a signed 16bit integer (int16_t) right? If so, I think that's more readable than the number... but I agree it's certainly far more readable than some complicated expression...

I guess what I'd prefer to see would be something like....

```
static const float MAX_INT16 = 32768.0f;
```

Admittedly, a float named INT16 is odd.. I'm open to a better name,

hifi-gustavo:
Build console output for commit: c4bc5233b1f0f5b068561fa1a7e8335691a7a0f7 available at: https://gist.github.com/df7e07f94f3fb91f9e15<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5872.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5872.dmg

hifi-gustavo:
Build console output for commit: 478eab38432b3ad0b4a4d25700127ed174cba498 available at: https://gist.github.com/df7e07f94f3fb91f9e15<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5872.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5872.dmg

PhilipRosedale:
Tested on both Mac and Windows.

murillodigital:
test this please

ZappoMan:
couple of questions:

1) Do we need to test with with WAV files with different sample sizes? I thought we'd added support at one point for injecting from different source formats @PhilipRosedale ?
2) Does this mean we are no longer dependent on soxr? And if so shouldn't we remove it from the cmake dependencies, etc? @murillodigital?

kencooke:
@ZappoMan Yes, this version is completely replaces SOXR.  Once it has been validated, we can go ahead and remove the SOXR package from cmake.

hifi-gustavo:
Build console output for commit: 478eab38432b3ad0b4a4d25700127ed174cba498 available at: https://gist.github.com/df7e07f94f3fb91f9e15<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5872.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5872.dmg

PhilipRosedale:
Scope testing confirms 100msecs reduction in latency on Windows when 48->24 conversion required at both input and output.  Nice!

PR #5874, opened 2015-09-22
cleanup IK implementation plusplus

howard-stearns:
(1) Split AnimInverseKinematics::evaluate() into several protected methods.
(2) Put IK targets in model-frame instead of root-frame.
(3) Split AnimNode implementation into two files.
  - This was all done by  @AndrewMeadows in "cleanup IK implementation" #5869.

(4)  Always keep targets, even when both position and rotation are unset. (Get from underpose. AnimInverseKinematics::computeTargets)
(5)  For head position, don't include the root rot, and restore the hmd conditional. (Rig::updateNeckJoint)


hifi-gustavo:
Build console output for commit: 7e52d38870d296ead5bf9b3053cfbcaf6686c0fa available at: https://gist.github.com/0f5fa46d19a9baa8cebe<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5874.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5874.dmg

PR #5881, opened 2015-09-23
remove underpose attenuation from IK relaxation step

AndrewMeadows:
This PR fixes a bug where the IK relaxation step was impeding the underpose animation orientations for joints that were not being moved by the IK logic.

### DIFF
--- a/libraries/animation/src/AnimInverseKinematics.h
+++ b/libraries/animation/src/AnimInverseKinematics.h
@@ -50,8 +50,6 @@ protected:
     // for AnimDebugDraw rendering
     virtual const AnimPoseVec& getPosesInternal() const override { return _relativePoses; }
 
-    void relaxTowardDefaults(float dt);
-
     RotationConstraint* getConstraint(int index);
     void clearConstraints();
     void initConstraints();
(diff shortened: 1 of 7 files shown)


libraries/animation/src/AnimInverseKinematics.h
```diff
@@ -72,7 +70,7 @@ class AnimInverseKinematics : public AnimNode {
     };
 
     std::map<int, RotationConstraint*> _constraints;
-    std::map<int, RotationAccumulator> _accumulators; // class-member to exploit temporal coherency
+    std::vector<RotationAccumulator> _accumulators;
```
AndrewMeadows: **_accumulators** is now a **std::vector** rather than a **std::map** -- every joint has one.  When an accumulator tracks a rotation it is marked as **dirty** and that dirty-flag is used during the relaxation step to decide whether its corresponding joint needs to be relaxed toward its underpose, or if it should just be slammed to it.

libraries/render-utils/src/Model.cpp
```diff
@@ -1727,9 +1727,7 @@ void Model::segregateMeshGroups() {
     // Run through all of the meshes, and place them into their segregated, but unsorted buckets
     int shapeID = 0;
     for (int i = 0; i < (int)networkMeshes.size(); i++) {
-        const NetworkMesh& networkMesh = *(networkMeshes.at(i).get());
         const FBXMesh& mesh = geometry.meshes.at(i);
-        const MeshState& state = _meshStates.at(i);
```
AndrewMeadows: Unused cruft.

hifi-gustavo:
Build console output for commit: d25ba8946e31b3cf8d22183196a6df335535db8c available at: https://gist.github.com/d47a73651b4d57fc722f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5881.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5881.dmg

howard-stearns:
nice!

PR #5887, opened 2015-09-23
[Scripts] createFlashlight.js - Close script after running

thoys:
Close the createFlashlight.js script after running once to prevent a flashlight from being created each time you start Interface

### DIFF
--- a/examples/toys/flashlight/createFlashlight.js
+++ b/examples/toys/flashlight/createFlashlight.js
@@ -1,5 +1,5 @@
 //
-//  createFlashligh.js
+//  createFlashlight.js
 //  examples/entityScripts
 //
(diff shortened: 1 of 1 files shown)


imgntn:
@ZappoMan suggested that we leave it persistent -- the spawner script is mainly a convenience.  interested in your thoughts about it -- createFlashLight shouldn't be in the default scripts, so if you remove it you shouldn't get one on restart.  Cheers.

ZappoMan:
@imgntn - this PR from Thoys does not change the persist behavior of the flashlight at all...

ZappoMan:
What it does is what it says in the description, namely after running the script, the script automatically shuts itself down... this causes the script to not stick around in the running scripts dialog box, so that when you restart the app it's not running creating more flashlights.

PR #5895, opened 2015-09-24
allow head translation when using hand IK in 3rd person screenie mode

AndrewMeadows:
The IK now supports a target that adjusts rotation and not position.  This allows us to orient the head during 3rd person screenie mode but allowing the head to translate when the hand targets pull the torso and neck around.

Also cleaned up the IK solver and fixed a bug that was preventing the solver from actually taking advantage of multiple iterations.

### DIFF
--- a/interface/resources/meshes/defaultAvatar_full/avatar-animation.json
+++ b/interface/resources/meshes/defaultAvatar_full/avatar-animation.json
@@ -31,7 +31,8 @@
                         {
                             "jointName": "Head",
                             "positionVar": "headPosition",
-                            "rotationVar": "headRotation"
+                            "rotationVar": "headRotation",
+                            "typeVar": "headType"
                         }
                     ]
                 },
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -141,6 +141,7 @@ QByteArray MyAvatar::toByteArray(bool cullSmallChanges, bool sendAll) {
 }
 
 void MyAvatar::reset() {
+    _skeletonModel.reset();
     getHead()->reset();
 
     _targetVelocity = glm::vec3(0.0f);
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -141,7 +141,6 @@ QByteArray MyAvatar::toByteArray(bool cullSmallChanges, bool sendAll) {
 }
 
 void MyAvatar::reset() {
-    _skeletonModel.reset();
     getHead()->reset();
 
     _targetVelocity = glm::vec3(0.0f);
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -42,9 +42,8 @@ void AnimInverseKinematics::loadPoses(const AnimPoseVec& poses) {
 
 void AnimInverseKinematics::computeAbsolutePoses(AnimPoseVec& absolutePoses) const {
     int numJoints = (int)_relativePoses.size();
-    absolutePoses.clear();
-    absolutePoses.resize(numJoints);
     assert(numJoints <= _skeleton->getNumJoints());
+    assert(numJoints == (int)absolutePoses.size());
     for (int i = 0; i < numJoints; ++i) {
         int parentIndex = _skeleton->getParentIndex(i);
         if (parentIndex < 0) {
@@ -55,7 +54,11 @@ void AnimInverseKinematics::computeAbsolutePoses(AnimPoseVec& absolutePoses) con
     }
 }
 
-void AnimInverseKinematics::setTargetVars(const QString& jointName, const QString& positionVar, const QString& rotationVar) {
+void AnimInverseKinematics::setTargetVars(
+        const QString& jointName, 
+        const QString& positionVar, 
+        const QString& rotationVar, 
+        const QString& typeVar) {
     // if there are dups, last one wins.
     bool found = false;
     for (auto& targetVar: _targetVarVec) {
@@ -63,13 +66,14 @@ void AnimInverseKinematics::setTargetVars(const QString& jointName, const QStrin
             // update existing targetVar
             targetVar.positionVar = positionVar;
             targetVar.rotationVar = rotationVar;
+            targetVar.typeVar = typeVar;
             found = true;
             break;
         }
     }
     if (!found) {
         // create a new entry
-        _targetVarVec.push_back(IKTargetVar(jointName, positionVar, rotationVar));
+        _targetVarVec.push_back(IKTargetVar(jointName, positionVar, rotationVar, typeVar));
     }
 }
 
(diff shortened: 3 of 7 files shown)


interface/resources/meshes/defaultAvatar_full/avatar-animation.json
```diff
@@ -31,7 +31,8 @@
                         {
                             "jointName": "Head",
                             "positionVar": "headPosition",
-                            "rotationVar": "headRotation"
+                            "rotationVar": "headRotation",
+                            "typeVar": "headType"
                         }
```
AndrewMeadows: IK targets get a new data member: type

interface/src/avatar/MyAvatar.cpp
```diff
@@ -141,7 +141,6 @@ QByteArray MyAvatar::toByteArray(bool cullSmallChanges, bool sendAll) {
 }
 
 void MyAvatar::reset() {
-    _skeletonModel.reset();
```
AndrewMeadows: Resetting the Model (and subsequently its Rig) is not helpful here in the new IK paradigm.  Removing this line removes an animation glitch triggered by pressing the apostrophe key.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -42,9 +42,8 @@ void AnimInverseKinematics::loadPoses(const AnimPoseVec& poses) {
 
 void AnimInverseKinematics::computeAbsolutePoses(AnimPoseVec& absolutePoses) const {
     int numJoints = (int)_relativePoses.size();
-    absolutePoses.clear();
-    absolutePoses.resize(numJoints);
     assert(numJoints <= _skeleton->getNumJoints());
+    assert(numJoints == (int)absolutePoses.size());
     for (int i = 0; i < numJoints; ++i) {
```
AndrewMeadows: We now assume that absolutePoses was correctly resized **before** this function is called.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -139,25 +148,23 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(std::vector<IKTarge
         accumulator.clearAndClean();
     }
 
-    float largestError = 0.0f;
-    const float ACCEPTABLE_RELATIVE_ERROR = 1.0e-3f;
     int numLoops = 0;
-    const int MAX_IK_LOOPS = 16;
-    const quint64 MAX_IK_TIME = 10 * USECS_PER_MSEC;
-    quint64 expiry = usecTimestampNow() + MAX_IK_TIME;
+    const int MAX_IK_LOOPS = 4;
```
AndrewMeadows: We don't want to limit the IK iterations on "time spent" or "accuracy of solution" since these checks can cause small glitches in IK solutions.  Instead we always make the same number of iterations.  When the initial conditions are close to the final solution (and this is typically true for our IK system) then making a relatively small number of iterations will rapidly converge the IK solver, hence we only need a few iterations (4, down from 16) each frame.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -166,59 +173,60 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(std::vector<IKTarge
...
-                            _accumulators[pivotIndex].add(newRot);
-                        }
-                        // this joint has been changed so we check to see if it has the lowest index
-                        if (pivotIndex < lowestMovedIndex) {
-                            lowestMovedIndex = pivotIndex;
+                        angle /= fractionDenominator;
+                        deltaRotation = glm::angleAxis(angle, axis);
+                    }
+                }
+                fractionDenominator++;
```
AndrewMeadows: We're using **fractionDenominator** to reduce the weight of further joints to the IK solution.  In other words, closer joints contribute more toward the solution than further.

hifi-gustavo:
Build console output for commit: 17e3e9394fbae1af48e93eefc8acb4ef45fc3604 available at: https://gist.github.com/cf975ed80d1c00bfcaa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5895.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5895.dmg

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -166,59 +173,60 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(std::vector<IKTarge
...
+                    // store the rotation change in the accumulator
+                    _accumulators[pivotIndex].add(newRot);
+                }
+                // this joint has been changed so we check to see if it has the lowest index
+                if (pivotIndex < lowestMovedIndex) {
+                    lowestMovedIndex = pivotIndex;
                 }
+
+                // keep track of tip's new position as we descend towards root
+                tip = jointPosition + deltaRotation * leverArm;
```
AndrewMeadows: This code used to be skipped if the deltaRotation from this particular joint was near identity, however we now include it because we are averaging the contribution of each iteration across common joints (e.g. the spine joints) and whenever an iteration contributes zero-rotation it still contributes to the average rotation of the joints involved.  To not do this causes fluctuations for some end-effectors, depending on whether small rotation contributions down the joint chain were larger than some threshold or not.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -86,6 +90,7 @@ static int findRootJointInSkeleton(AnimSkeleton::ConstPointer skeleton, int inde
 
 void AnimInverseKinematics::computeTargets(const AnimVariantMap& animVars, std::vector<IKTarget>& targets) {
     // build a list of valid targets from _targetVarVec and animVars
+    _maxTargetIndex = -1;
```
AndrewMeadows: Resetting **maxTargetIndex** to -1 here is important!  The fact that we weren't doing this would cause the IK to effectively only iterate once each frame -- in fact it would iterate several times but each successive iteration after the first would contribute nothing toward the final solution!

libraries/animation/src/AnimInverseKinematics.h
```diff
@@ -31,20 +31,33 @@ class AnimInverseKinematics : public AnimNode {
...
         int rootIndex;
+        Type type = Type::RotationAndPosition;
+
+        void setType(const QString& typeVar) {
+            if (typeVar == "RotationOnly") {
+                type = Type::RotationOnly;
+            } else {
+                type = Type::RotationAndPosition;
+            }
+        }
```
AndrewMeadows: Hrm... this should be changed to be a one-liner so we can keep it in the header file.

hifi-gustavo:
Build console output for commit: 59da684eb94dc59e82f95d66cee6d8457a659407 available at: https://gist.github.com/cf975ed80d1c00bfcaa5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5895.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5895.dmg

PR #5901, opened 2015-09-25
Threshold based walking/leaning while in HMD. 

hyperlogic:
This is a blend of the previous 'sitting' and 'standing' HMD modes.
Basically, when you move your head within a small range (20cm) your
avatar will lean appropriately, however when you cross that threshold
your body will move underneath you, re-centering your head above the body.

While this occurs the avatar should play the appropriate walking animations.

The View > Standing HMD Sensor Mode menu has been removed.


hifi-gustavo:
Build console output for commit: 474c847ef1da4538f99afd707c4a826ec66b91d8 available at: https://gist.github.com/a21c459a5505556467fb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5901.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5901.dmg

ZappoMan:
Unused variable warning: OverlayConductor.cpp line 71 - 

/data/jenkins/Jenkins/workspace/hifi-pull-request-ubuntu/interface/src/ui/OverlayConductor.cpp:71:19: warning: unused variable ‘myAvatar’ [-Wunused-variable]

ZappoMan:
test this please

hifi-gustavo:
Build console output for commit: 474c847ef1da4538f99afd707c4a826ec66b91d8 available at: https://gist.github.com/a21c459a5505556467fb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5901.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5901.dmg

interface/src/ui/OverlayConductor.cpp
```diff
@@ -69,11 +69,7 @@ void OverlayConductor::updateMode() {
     Mode newMode;
     if (qApp->isHMDMode()) {
         MyAvatar* myAvatar = DependencyManager::get<AvatarManager>()->getMyAvatar();
```
ZappoMan: this line is causing an unused variable warning

interface/src/avatar/MyAvatar.h
```diff
@@ -362,6 +358,9 @@ public slots:
     AudioListenerMode _audioListenerMode;
     glm::vec3 _customListenPosition;
     glm::quat _customListenOrientation;
+
+    bool _straightingLean = false;
+    float _straightingLeanAlpha = 0.0f;
```
AndrewMeadows: I don't think **_straigtingLeanAlpha** should be a class data member.  It is only ever used in **MyAvatar::updateFromHMDSensorMatrix()** and could be a local variable there.

AndrewMeadows: Oh I see, you're actually incrementing _straigtingLeanAlpha each frame, and then eventually just slamming the position when it exceeds 1.0, so it does need to be a class data member.  Hrm... this means that the lerp will accelerate over time.  If you were to keep **straigtingLeanAlpha** a constant ratio (dt / timescale) then you'll get exponential decay toward the target position (it will move fast at the beginning, but will slow down as it gets closer, getting to within about 1/e of where it was 1.0 timescale ago) and you could just slam it (and stop lerping) when you get close enough.

So... I guess incrementing the alpha will result in something that is a bit more "linear" than exponential decay.  Meh, leave it.  If the behavior doesn't bother anyone then its good enough.

PhilipRosedale: good observation though - we should always use exponential (by multiplying by a constant < 1 each dt) for anything in motion, not linear.

hifi-gustavo:
Build console output for commit: 402573dbf15d193a42b67735611ff9304203e900 available at: https://gist.github.com/a21c459a5505556467fb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5901.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5901.dmg

PR #5903, opened 2015-09-25
enforce target rotation during IK calculations

AndrewMeadows:
The IK system was not trying very hard to enforce the final orientation of the end-effector's target rotation.  This PR makes it apply extra twist to parent joints down the chain if the end-effector's rotation violates the target orientation.

In short: the upper arm will rotate harder to get the hand where it is supposed to be.

### DIFF
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -9,6 +9,7 @@
 
 #include "AnimInverseKinematics.h"
 
+#include <GeometryUtil.h>
 #include <GLMHelpers.h>
 #include <NumericalConstants.h>
 #include <SharedUtil.h>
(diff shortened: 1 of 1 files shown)


libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -464,7 +504,7 @@ void AnimInverseKinematics::initConstraints() {
         } else if (0 == baseName.compare("Hand", Qt::CaseInsensitive)) {
             SwingTwistConstraint* stConstraint = new SwingTwistConstraint();
             stConstraint->setReferenceRotation(_defaultRelativePoses[i].rot);
-            const float MAX_HAND_TWIST = PI;
+            const float MAX_HAND_TWIST = 3.0f * PI / 5.0f;
```
AndrewMeadows: These minor adjustments to hand constraints make the IK upper arms start to twist closer to when the user's real-arms start to twist.

hifi-gustavo:
Build console output for commit: a97f5569584371878752c8a18f5dfbcbd4e2da58 available at: https://gist.github.com/85dcc8fce6fd26634800<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5903.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5903.dmg

PR #5906, opened 2015-09-26
Algorithmic optimization of new resampler

kencooke:
All common sample rates now use a rational mode that does direct FIR filtering instead of coefficient interpolation.  Complexity reduced by 2x for mono, 1.5x for stereo.


hifi-gustavo:
Build console output for commit: 4a1458e152518947674fd86c2492655cf01638b5 available at: https://gist.github.com/85934dd5003cb58f55d4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5906.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5906.dmg

hifi-gustavo:
Build console output for commit: 956d83c87277b1c94bbc1b297fbe6f13750c6785 available at: https://gist.github.com/85934dd5003cb58f55d4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5906.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5906.dmg

hifi-gustavo:
Build console output for commit: 4c1263045873ddbb49496b65bdc2042c8086a91c available at: https://gist.github.com/85934dd5003cb58f55d4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5906.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5906.dmg

AndrewMeadows:
Github refuses to show the diff for AudioSRC.cpp after the latest commit so it is hard to browse and comment on the code differences but I noticed one weirdness...

What caught my eye is that **AudioSRC::createIrrationalFilter()** sets **_stepTable** to **nullptr** but doesn't delete whatever it points to.  I was wondering: "Is that a memory leak?"

Looking further I notice that **_stepTable** is not initialized explicitly in the ctor -- the only thing that initializes it is **AudioSRC::createRationalFilter()** which doesn't bother to check its value before it does --> the two **createFooFilter()** methods are **incompatible** and cannot be called together.  Meanwhile the dtor is guaranteed to delete **_stepTable** which means one or the other MUST be called for every instance of AudioSRC or there could be memory corruption whenever an AudioSRC is destroyed.

Turns out that condition is met -- one or the other is called by the ctor and they are properly classified as **private** methods.

So the code works but it isn't clear without a detailed read and that "somewhat out of context" initialization inside **AudioSRC::createIrrationalFilter()** poked me in the eye.  I think **_stepTable** should be explicitly initialized to **nullptr** by the ctor, even if only to reduce the WTF/m metric:

http://images.cryhavok.org/v/WTFs+per+Minute.jpg.html

hifi-gustavo:
Build console output for commit: c000c635b18ca06e8629598105b9f21083f0cddb available at: https://gist.github.com/85934dd5003cb58f55d4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5906.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5906.dmg

hifi-gustavo:
Build console output for commit: 7faada24a886da7537b6825191ab9d037889ded3 available at: https://gist.github.com/85934dd5003cb58f55d4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5906.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5906.dmg

ZappoMan:
```
/Users/jenkins/Jenkins/workspace/pull-request-osx/libraries/audio/src/AudioSRC.cpp:772:23: warning: unused variable 'coef' [-Wunused-variable]
                float coef = c0[j];
                      ^
/Users/jenkins/Jenkins/workspace/pull-request-osx/libraries/audio/src/AudioSRC.cpp:836:23: warning: unused variable 'coef' [-Wunused-variable]
                float coef = c0[j];
                      ^
2 warnings generated.
```

ZappoMan:
btw - I think the reason GitHub doesn't show the diff is because it has some hard limit on the number of lines (or characters) in a change that it will display, in this case, since the commit includes deleting 4,166 lines, I suspect that that's the source of the problem. @kencooke - I assume the lines you deleted were related to the prototypeFilter.. is there anyway we could make that be a different file? would we make it an include? Just so that in the future, we can see changes to the actual algorithm.

Your thoughts? If you think that will complicate the code then I'd understand.

kencooke:
Good catch on the compiler warnings.

I considered putting the table in a separate .cpp, but that requires another header file too and I didn't think it was worth the clutter.  The table should be final now, so hopefully it will stay out of the diffs.

hifi-gustavo:
Build console output for commit: aa01dca0a87367054aedaec45b8186cf341f2e32 available at: https://gist.github.com/85934dd5003cb58f55d4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5906.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5906.dmg

PR #5917, opened 2015-09-28
Fix for bad textures when using _glActiveBindTexture

jherico:
Currently we have some renderables that provide texture ID's directly, disallowing the use of the gpu::Texture interface.  This requires us to use `_glActiveBindTexture` to bind the texture during rendering.  However, previously this command didn't update the GLBackend cached state regarding textures, so as far as GLBackend was concerned, whatever texture was bound to a given slot before the call to `_glActiveBindTexture` was still bound.  This meant that if something used texture X, then a web entity was rendered, then something else which used texture X was rendered, the last thing would instead be rendered with the texture from the web entity.

This PR modifies the `_glActiveBindTexture` call in Batch to explicitly unbind corresponding texture slot, so the next render will be aware it needs to rebind whatever it needs.

### DIFF
--- a/libraries/gpu/src/gpu/GLBackend.cpp
+++ b/libraries/gpu/src/gpu/GLBackend.cpp
@@ -418,6 +418,8 @@ void GLBackend::resetStages() {
 #define DO_IT_NOW(call, offset) 
 
 void Batch::_glActiveBindTexture(GLenum unit, GLenum target, GLuint texture) {
+    setResourceTexture(unit - GL_TEXTURE0, nullptr);
+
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 203c601e95fcd1c1c320afaef62cc6e64b985222 available at: https://gist.github.com/8277b5345ca4335f795e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5917.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5917.dmg

PR #5918, opened 2015-09-28
Throw held objects with velocity from fingertip

PhilipRosedale:


examples/controllers/handControllerGrab.js
```diff
@@ -274,11 +277,44 @@ function MyController(hand, triggerAction) {
...
+        var currentPosition = MyAvatar.position;
+        var avatarDeltaPosition = Vec3.subtract(currentPosition, this.currentAvatarPosition); 
+        this.currentAvatarPosition = currentPosition;
+        
+        // How far did the avatar turn this timestep?
+        // Note:  The following code is too long because we need a Quat.quatBetween() function
+        // that returns the minimum quaternion between two quaternions. 
+        var currentOrientation = MyAvatar.orientation;
+        if (Quat.dot(currentOrientation, this.currentAvatarOrientation) < 0.0) {
+            var negativeCurrentOrientation = { x: -currentOrientation.x,
```
ZappoMan: according to the newly agreed to coding standard this should be formatted as:

```
        var negativeCurrentOrientation = { 
            x: -currentOrientation.x, 
            y: -currentOrientation.y, 
            z: -currentOrientation.z, 
            w: -currentOrientation.w 
    };
```

examples/controllers/handControllerGrab.js
```diff
@@ -86,7 +86,7 @@ function MyController(hand, triggerAction) {
 
     this.triggerAction = triggerAction;
     this.palm = 2 * hand;
-    // this.tip = 2 * hand + 1; // unused, but I'm leaving this here for fear it will be needed
+    this.tip = 2 * hand + 1;
```
ZappoMan: coding standard: magic numbers? Can you document what 2 and 1 are and/or just add a comment explaining this? I believe the 2 is SPATIAL_CONTROLLERS_PER_PALM and the 1 is TIP_CONTROLLER_OFFSET or something like that.

PR #5919, opened 2015-09-28
keep interface from printing message about changing avatar url over and over

sethalves:
- don't log a message repeatedly if an avatar or scripted agent has an empty string for avatar url


hifi-gustavo:
Build console output for commit: a8ceb6e9d7a3d95b6ebf66bcf8f67949269c925a available at: https://gist.github.com/350a458d530d4c99d0f8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5919.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5919.dmg

howard-stearns:
Sorry about that. My bad. It looks like when I wrote the default avatar stuff, I never tested what the logs look like when someone ELSE has the default avatar. Is that the issue you are trying to fix here?

sethalves:
I *think* it's caused by the scripted agent in my domain.  The issue isn't that someone has the default, it's that some avatar has an avatar url which is an empty string.

howard-stearns:
I think it is that the someone has the default. I deliberately made AvatarData::identityByteArray() send an empty url when using the default, because the default is to a local file url that is different on different machines. Then on the receiving side, the empty string is expanded to the default.  

I think the following would give us cleaner logs with less change:
@@ -989,8 +989,11 @@ void AvatarData::setFaceModelURL(const QUrl& faceModelURL) {
 }
 
 void AvatarData::setSkeletonModelURL(const QUrl& skeletonModelURL) {
-    _skeletonModelURL = skeletonModelURL.isEmpty() ? AvatarData::defaultFullAvatarModelUrl() : skeletonModelURL;
-
+    const QUrl& expanded = skeletonModelURL.isEmpty() ? AvatarData::defaultFullAvatarModelUrl() : skeletonModelURL;
+    if (expanded == _skeletonModelURL) {
+        return;
+    }
+    _skeletonModelURL = expanded;
     qCDebug(avatars) << "Changing skeleton model for avatar to" << _skeletonModelURL.toString();
 
     updateJointMappings();

What do you think?

sethalves:
closing in favor of #5933

PR #5927, opened 2015-09-29
fix skybox not always loading when you haven't moved or if the texture is slow in loading

ZappoMan:
Two bug fixes:
* If you are stationary and not moving, and a zone loaded around you, we weren't detecting enter/leave entity and therefore weren't setting zone properties
* If we did apply zone properties for a skybox, but the texture wasn't loaded, it would set the texture to black, and never reset the texture when the texture was actually eventually downloaded

This fixes both of those issues.


hifi-gustavo:
Build console output for commit: 2040fb4507c0aae989163df75cd5eac902171c8d available at: https://gist.github.com/3e7596304d7fb8951e34<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5927.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5927.dmg

PR #5930, opened 2015-09-29
Setting hand data for close grab of non physical objects

ericrius1:


imgntn:
LGTM

examples/controllers/handControllerGrab.js
```diff
@@ -29,9 +29,21 @@ var TRIGGER_ON_VALUE = 0.2;
 var DISTANCE_HOLDING_RADIUS_FACTOR = 5; // multiplied by distance between hand and object
 var DISTANCE_HOLDING_ACTION_TIMEFRAME = 0.1; // how quickly objects move to their new position
 var DISTANCE_HOLDING_ROTATION_EXAGGERATION_FACTOR = 2.0; // object rotates this much more than hand did
-var NO_INTERSECT_COLOR = { red: 10, green: 10, blue: 255}; // line color when pick misses
-var INTERSECT_COLOR = { red: 250, green: 10, blue: 10}; // line color when pick hits
-var LINE_ENTITY_DIMENSIONS = { x: 1000, y: 1000, z: 1000};
+var NO_INTERSECT_COLOR = {
```
ZappoMan: do you REALLY have to format these to 5 lines each? I realize that with the new coding standard this is allowed... but do you really need it this way? these are short lines initializing single objects colors, and positions.

imgntn: @ericrius1 the object spacing doesn't bother me as much (although it IS very readable short-form in this case), but i think brad is right it's kind of hard to read, the comments come after the object they reference --- maybe switch them to before

ericrius1: @ZappoMan @imgntn 
You guys have good points- it was a reflex autoformat usage. Will fix

PR #5936, opened 2015-09-29
fix a timer leak, line width in PacketList

birarda:


libraries/networking/src/ThreadedAssignment.cpp
```diff
@@ -105,7 +106,7 @@ void ThreadedAssignment::sendStatsPacket() {
 void ThreadedAssignment::startSendingStats() {
     // send the stats packet every 1s
     if (!_statsTimer) {
-        _statsTimer = new QTimer();
+        _statsTimer = new QTimer;
```
ZappoMan: not a merge blocker, but why did you change this?

birarda: because it's redundant to show the parentheses

hifi-gustavo:
Build console output for commit: f4bc3f8e388a6b38456f880dd7a3904150dae072 available at: https://gist.github.com/5d088dfa31d702b366b8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5936.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5936.dmg

libraries/networking/src/ThreadedAssignment.cpp
ZappoMan: so IYO should we be removing all empty parenthesis from all constructors? This isn't explicitly in our out of our coding standard, but in pretty much 99% of our code we do include the () on all new calls to classes. There's 6 other occurrences of `new QTimer();` in our codebase. This is the only occurrence of `new QTimer`. I personally don't mind one way or the other on this... I can get behind either approach, but we should be consistent.

birarda: I would say that parentheses should not be required when newing an object
with default ctor.

This is in fact not good habit since in other places that syntax can
actually do the wrong thing (operator()).

ZappoMan: if it can sometimes do the wrong thing... then we should prohibit it.

Although, I'm a little confused as to how the operator() would execute in this case... I'm not doubting you, I just haven't thought about it or researched it.

Like I said, I'm happy to agree to that as a standard... but the current random nature is a little confusing, I see a similar pattern with `new QThread()` vs `new QThread`... I haven't checked for others those were just the first two that came to mind when I thought about things we often "new" with no params.

birarda: It's called the "most vexing parse". More info here:
https://en.m.wikipedia.org/wiki/Most_vexing_parse.

ZappoMan: this sounds like a really good argument for why we should actually switch to requiring 	"Uniform initialization syntax" our current coding standard actually prohibits use of Uniform Initialization syntax in a few places. Maybe it's time to reevaluate this issue?

birarda: Yes, I prefer that syntax and would support the change.

howard-stearns: ditto.

huffman: Also agree. There are actually quite a few places where we've used that syntax on the new networking code where it would probably be prohibited according to the style guide, ie initialization/defaults of properties in class declarations.

PR #5938, opened 2015-09-29
Fixing the lighting view matrix for stereo again

samcake:
This randomly chosen combination of head and eye matrix seems to finally work in stereo... :)

Just kidding we actually know what we are doing here, we are professionals!
The eyeView we get from the Context is the matrix going from the mono eye space to the stereo eye space. to create the correct "ViewTransform" for a given side is the matrix going from the stereo eye space back to the world space. So that s why we need monoView * inv(stereoView).
Long term we should put the eye offset in the projection matrix and not fool around with different view matrices for stereo.

### DIFF
--- a/libraries/render-utils/src/DeferredLightingEffect.cpp
+++ b/libraries/render-utils/src/DeferredLightingEffect.cpp
@@ -442,7 +442,7 @@ void DeferredLightingEffect::render(RenderArgs* args) {
 
                 deferredTransforms[i].projection = projMats[i];
 
-                auto sideViewMat =  eyeViews[i] * monoViewMat;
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 1bd07b58c3371df3f1e1e548bdd145f22330fa5d available at: https://gist.github.com/7d92722df8e375728f33<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5938.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5938.dmg

RyanDowne:
Tested on Win in toybox and works good

PR #5941, opened 2015-09-29
Fix particle radius, color, and alpha starts/finishes

ctrlaltdavid:
Particle radius, color, and alpha start and finish values now stay in sync with the base particle radius, color, and alpha values until such time as start/finish values are explicitly set.

### DIFF
--- a/libraries/entities/src/EntityItemProperties.cpp
+++ b/libraries/entities/src/EntityItemProperties.cpp
@@ -1573,14 +1573,17 @@ void EntityItemProperties::markAllChanged() {
     _accelerationSpreadChanged = true;
     _particleRadiusChanged = true;
     _radiusSpreadChanged = true;
-    _radiusStartChanged = true;
-    _radiusFinishChanged = true;
     _colorSpreadChanged = true;
-    _colorStartChanged = true;
-    _colorFinishChanged = true;
     _alphaSpreadChanged = true;
-    _alphaStartChanged = true;
-    _alphaFinishChanged = true;
+
+    // Only mark the following as changed if their values are specified in the properties when the particle is created. If their
+    // values are specified then they are marked as changed in getChangedProperties().
+    //_radiusStartChanged = true;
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: e2bee8ec08e441ce16a47ff70be7f1d2cf10b4b9 available at: https://gist.github.com/6184fb683f216a14260b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5941.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5941.dmg

imgntn:
Tested w/ @ctrlaltdavid on Mac, good to go for that platform

libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -1573,14 +1573,17 @@ void EntityItemProperties::markAllChanged() {
...
-    _alphaFinishChanged = true;
+
+    // Only mark the following as changed if their values are specified in the properties when the particle is created. If their
+    // values are specified then they are marked as changed in getChangedProperties().
+    //_radiusStartChanged = true;
+    //_radiusFinishChanged = true;
+    //_colorStartChanged = true;
+    //_colorFinishChanged = true;
+    //_alphaStartChanged = true;
+    //_alphaFinishChanged = true;
```
Atlante45: Why did you comment those ones, I'm not sure I understand the comment?
If we really don't need them to be reset then delete the code and maybe add a list to your comment.

ctrlaltdavid: I am trying to prevent someone coming along and thinking that lines setting _radiusStartChanged = true; are missing and so adding them in and breaking these properties' behaviours. I think having the statements commented out like they are makes it more obvious that they have been considered but intentionally not done. (Who reads comments?) ... But I'll change to a list if you think still that's better.

PR #5942, opened 2015-09-29
Change the update rate of the flickering light to 30hz

ZappoMan:
this will reduce the load on the entity server

### DIFF
--- a/examples/acScripts/flickeringLight.js
+++ b/examples/acScripts/flickeringLight.js
@@ -78,10 +78,10 @@ function update(deltaTime) {
         intensity += randFloat(-LIGHT_INTENSITY_RANDOMNESS, LIGHT_INTENSITY_RANDOMNESS);
         var properties = Entities.getEntityProperties(LightMaker.light, ["age", "lifetime"]);
         //print("props:" +JSON.stringify(properties));
-        //print("age:" + properties.age);
-        //print("lifetime:" + properties.lifetime);
+        print("age:" + properties.age);
+        print("lifetime:" + properties.lifetime);
         var newLifetime = properties.age + EPHEMERAL_LIFETIME;
-        //print("newLifetime:" + newLifetime);
+        print("newLifetime:" + newLifetime);
         Entities.editEntity(LightMaker.light, { type: "Light", intensity : intensity, lifetime: newLifetime });
     }
 }
--- a/examples/acScripts/flickeringLight.js
+++ b/examples/acScripts/flickeringLight.js
@@ -43,7 +43,7 @@ var totalTime = 0;
 var MINIMUM_LIGHT_INTENSITY = 0.75;
 var MAXIMUM_LIGHT_INTENSITY = 2.75;
 var LIGHT_INTENSITY_RANDOMNESS = 0.3;
-var EPHEMERAL_LIFETIME = 10; // ephemeral entities will live for 10 seconds after script stops running
+var EPHEMERAL_LIFETIME = 60; // ephemeral entities will live for 60 seconds after script stops running
 
 var LightMaker = {
     light: null,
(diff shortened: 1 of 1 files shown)


examples/acScripts/flickeringLight.js
```diff
@@ -74,10 +76,17 @@ function update(deltaTime) {
...
-        Entities.editEntity(LightMaker.light, { intensity: intensity, lifetime: properties.age + EPHEMERAL_LIFETIME });
+
+        // We don't want to edit the entity EVERY update cycle, because that's just a lot
+        // of wasted bandwidth and extra effort on the server for very little visual gain
+        if (totalTime - lastUpdate > UPDATE_INTERVAL) {
+            var intensity = (MINIMUM_LIGHT_INTENSITY + (MAXIMUM_LIGHT_INTENSITY + (Math.sin(totalTime) * MAXIMUM_LIGHT_INTENSITY)));
+            intensity += randFloat(-LIGHT_INTENSITY_RANDOMNESS, LIGHT_INTENSITY_RANDOMNESS);
+            var properties = Entities.getEntityProperties(LightMaker.light, "age");
+            var newLifetime = properties.age + EPHEMERAL_LIFETIME;
+            Entities.editEntity(LightMaker.light, { type: "Light", intensity: intensity, lifetime: newLifetime });
```
jherico: what's the point of specifying the type here?

ZappoMan: It allows the encode edit packet code to better know which properties are valid. it's best practice.

examples/acScripts/flickeringLight.js
```diff
@@ -74,10 +76,17 @@ function update(deltaTime) {
...
         totalTime += deltaTime;
-        var intensity = (MINIMUM_LIGHT_INTENSITY + (MAXIMUM_LIGHT_INTENSITY + (Math.sin(totalTime) * MAXIMUM_LIGHT_INTENSITY)));
-        intensity += randFloat(-LIGHT_INTENSITY_RANDOMNESS, LIGHT_INTENSITY_RANDOMNESS);
-        var properties = Entities.getEntityProperties(LightMaker.light, "age");
-        Entities.editEntity(LightMaker.light, { intensity: intensity, lifetime: properties.age + EPHEMERAL_LIFETIME });
+
+        // We don't want to edit the entity EVERY update cycle, because that's just a lot
+        // of wasted bandwidth and extra effort on the server for very little visual gain
+        if (totalTime - lastUpdate > UPDATE_INTERVAL) {
+            var intensity = (MINIMUM_LIGHT_INTENSITY + (MAXIMUM_LIGHT_INTENSITY + (Math.sin(totalTime) * MAXIMUM_LIGHT_INTENSITY)));
```
jherico: This will produce a light level of up to `MINIMUM_LIGHT_INTENSITY + 2 * MAXIMUM_LIGHT_INTENSITY`, which is  counter-intuitive from the variable names.  

If we want the light to vary between MIN and MAX then the intensity should be 

```
var LIGHT_INTENSITY_RANGE = MAXIMUM_LIGHT_INTENSITY  - MINIMUM_LIGHT_INTENSITY;
var HALF_LIGHT_INTENSITY_RANGE = LIGHT_INTENSITY_RANGE / 2.0;
var LIGHT_MEDIAN_INTENSITY = MINIMUM_LIGHT_INTENSITY + HALF_LIGHT_INTENSITY_RANGE;

var intensity = LIGHT_MEDIAN_INTENSITY + Math.sin(totalTime) * HALF_LIGHT_INTENSITY_RANGE;
```

examples/acScripts/flickeringLight.js
```diff
@@ -74,10 +76,17 @@ function update(deltaTime) {
...
+            intensity += randFloat(-LIGHT_INTENSITY_RANDOMNESS, LIGHT_INTENSITY_RANDOMNESS);
```
jherico: This line will still cause a random variation with a total range of 0.6 at 30 Hz.  Rather than use a random value here, you might want to replace the entire intensity calculation with a noise function with several octaves.  This would cause the light to change gradually over time but without a (possibly perceptible) sine wave, yet still retain occasional highlights.

examples/acScripts/flickeringLight.js
ZappoMan: this is the behavior that the original fire did, this is the behavior Ryan is happy with

examples/acScripts/flickeringLight.js
ZappoMan: correct - this is the behavior that the original fire did, this is the behavior Ryan is happy with

PR #5944, opened 2015-09-29
fix bug: IK attenuates fast underpose animations

AndrewMeadows:
The IK was using its current local pose for the default targets when the various target variables were not properly set externally.  This would cause the resulting animation to be the result of the relaxation phase, which attenuates rapid oscillations in motion.  The relaxation has a timescale of about 0.25 seconds so any motion faster than a couple multiples of the timescale would get blunted.

### DIFF
--- a/interface/resources/meshes/defaultAvatar_full/avatar-animation.json
+++ b/interface/resources/meshes/defaultAvatar_full/avatar-animation.json
@@ -26,13 +26,14 @@
                         {
                             "jointName": "Neck",
                             "positionVar": "neckPosition",
-                            "rotationVar": "neckRotation"
+                            "rotationVar": "neckRotation",
+                            "typeVar": "headAndNeckType"
                         },
                         {
                             "jointName": "Head",
                             "positionVar": "headPosition",
                             "rotationVar": "headRotation",
-                            "typeVar": "headType"
+                            "typeVar": "headAndNeckType"
                         }
                     ]
                 },
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -89,7 +89,7 @@ static int findRootJointInSkeleton(AnimSkeleton::ConstPointer skeleton, int inde
     return rootIndex;
 }
 
-void AnimInverseKinematics::computeTargets(const AnimVariantMap& animVars, std::vector<IKTarget>& targets) {
+void AnimInverseKinematics::computeTargets(const AnimVariantMap& animVars, std::vector<IKTarget>& targets, const AnimPoseVec& underPoses) {
     // build a list of valid targets from _targetVarVec and animVars
     _maxTargetIndex = -1;
     bool removeUnfoundJoints = false;
@@ -107,7 +107,7 @@ void AnimInverseKinematics::computeTargets(const AnimVariantMap& animVars, std::
             }
         } else {
             IKTarget target;
-            AnimPose defaultPose = _skeleton->getAbsolutePose(targetVar.jointIndex, _relativePoses);
+            AnimPose defaultPose = _skeleton->getAbsolutePose(targetVar.jointIndex, underPoses);
             target.pose.trans = animVars.lookup(targetVar.positionVar, defaultPose.trans);
             target.pose.rot = animVars.lookup(targetVar.rotationVar, defaultPose.rot);
             target.setType(animVars.lookup(targetVar.typeVar, QString("")));
@@ -154,7 +154,6 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
     do {
         int lowestMovedIndex = _relativePoses.size();
         for (auto& target: targets) {
-            int tipIndex = target.index;
             if (target.type == IKTarget::Type::RotationOnly) {
                 // the final rotation will be enforced after the iterations
                 continue;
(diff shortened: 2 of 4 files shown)


libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -107,7 +107,7 @@ void AnimInverseKinematics::computeTargets(const AnimVariantMap& animVars, std::
             }
         } else {
             IKTarget target;
-            AnimPose defaultPose = _skeleton->getAbsolutePose(targetVar.jointIndex, _relativePoses);
+            AnimPose defaultPose = _skeleton->getAbsolutePose(targetVar.jointIndex, underPoses);
```
AndrewMeadows: Using the underPoses at this line fixes the bug, but gettingthe underposes passed into this context required the rest of the changes.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -288,6 +288,17 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
...
+    /* KEEP: example code for measuring endeffector error of IK solution
+    for (uint32_t i = 0; i < targets.size(); ++i) {
+        auto& target = targets[i];
+        if (target.type == IKTarget::Type::RotationOnly) {
+            continue;
+        }
+        glm::vec3 tipPosition = absolutePoses[target.index].trans;
+        std::cout << i << " IK error = " << glm::distance(tipPosition, target.pose.trans) << std::endl;  // adebug
+    }
+    */
```
AndrewMeadows: I'm leaving this debug code in case anyone else wants to examine the effectiveness of the IK algorithm.  It tends to get within a few tenths of a millimeter after its 4 iteration loops.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -312,25 +323,8 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
...
-                ++constraintItr;
-            }
-        } else {
-            solveWithCyclicCoordinateDescent(targets);
-        }
-    }
+    // don't call this function, call overlay() instead
+    assert(false);
     return _relativePoses;
 }
```
AndrewMeadows: The interesting code inside **execute()** has been moved to **overlay()**, which was the only thing calling **execute()**.

AndrewMeadows:
BTW: a side effect of this change is that the underpose-driven-IK-hands will add a little bit of extra motion to the shoulders.  It isn't terribly wacky but is noticeable in a side-by-side comparison with the old Rig animation code and is worth judging before merging this.

hifi-gustavo:
Build console output for commit: cdae16e07bd2e8649d310b7c5e0e2a49e5075d20 available at: https://gist.github.com/f319f375fab22136b36d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5944.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5944.dmg

hifi-gustavo:
Build console output for commit: a052f5e125f7d75fb4d36929c05b2ba2cc6e28df available at: https://gist.github.com/f319f375fab22136b36d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5944.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5944.dmg

interface/resources/meshes/defaultAvatar_full/avatar-animation.json
```diff
@@ -26,13 +26,14 @@
...
+                            "rotationVar": "neckRotation",
+                            "typeVar": "headAndNeckType"
                         },
                         {
                             "jointName": "Head",
                             "positionVar": "headPosition",
                             "rotationVar": "headRotation",
-                            "typeVar": "headType"
+                            "typeVar": "headAndNeckType"
                         }
```
AndrewMeadows: This change fixes a bug (slight sideways buckle in avatar spine) in 3rd person non-HMD mode.

What I think was going on is that the neck's position requires the spine to buckle a little bit, but with the IK pulling both hands toward the underpose the spine will decide to buckle to a little bit to one side rather than forward and back.  This change (with supporting changes in **Rig::updateNeckJoint()**) allows the neck to move linearly during IK and allows the spine to relax toward its proper default underpose.

hifi-gustavo:
Build console output for commit: b192d0a9cdfdada443fc3957804287cfff0d08ab available at: https://gist.github.com/f319f375fab22136b36d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5944.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5944.dmg

interface/resources/meshes/defaultAvatar_full/avatar-animation.json
howard-stearns: What does this do for hmd? Do we still have the spine compression/curve?

howard-stearns:
Here's the plan from some discussion:
* We're holding merges because there are some unrelated fixes pending for demo.
* When we unfreeze, we'll merge this, which is completely awesome for non-HMD. In HMD, it is cooler than current master, but can occassionally be a bit "too interesting" in the shoulders due to the compressed spine.
* We won't try to solve those cases until we get the moving hips. (A combination of root translation in a ready-to-go PR from Seth, and some new development by Andrew.)  We think the moving hips will allow us to distribute the compression between curved spine/shoulders and lowered hips with bent knees. We'll want to see the results before deciding how to proceed.
* All of this will be considered from the point of view of "small" head motions leading to small hip motions, as larger motions will be handled by whatever we come up with for "taking a seat" vs "standing up" state transitions.

PR #5950, opened 2015-09-30
guard for insert/get in SentPacketHistory

birarda:


libraries/networking/src/SentPacketHistory.cpp
```diff
@@ -35,10 +35,11 @@ void SentPacketHistory::packetSent(uint16_t sequenceNumber, const NLPacket& pack
     }
     _newestSequenceNumber = sequenceNumber;
 
+    QWriteLocker locker(&_packetsLock);
     _sentPackets.insert(NLPacket::createCopy(packet));
 }
 
-const NLPacket* SentPacketHistory::getPacket(uint16_t sequenceNumber) const {
```
ZappoMan: you could keep this const but make your mutex mutable

birarda: It's no longer truly const, it modifies the mutex - what's the benefit to the other approach if we don't have callers that need this to be const?

hifi-gustavo:
Build console output for commit: f7e7b07441e014f3875cb1db0440252cee354342 available at: https://gist.github.com/af2f91643a63a044aaf3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5950.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5950.dmg

hifi-gustavo:
Build console output for commit: 82ac0b1a27da3600dfca483c39ef495e877ce384 available at: https://gist.github.com/af2f91643a63a044aaf3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5950.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5950.dmg

ZappoMan:
smoke test passed on windows, mac, interface and stack manager

PR #5951, opened 2015-09-30
Fix infinite recursion error with CC rng

huffman:
On some std::random implementations tryin get a range [low, high) where
high < low will cause infinite recursion.


hifi-gustavo:
Build console output for commit: 1e9593029c260c5c346479972e13be7417c3a038 available at: https://gist.github.com/3f4fcd9e08823042e8c5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5951.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5951.dmg

PR #5960, opened 2015-09-30
fix possible crash in agent

ZappoMan:


assignment-client/src/Agent.cpp
```diff
@@ -364,10 +364,14 @@ void Agent::processAgentAvatarAndAudio(float deltaTime) {
 }
 
 void Agent::aboutToFinish() {
-    _scriptEngine->stop();
+    if (_scriptEngine) {
```
ZappoMan: it's theoretically possible that the ThreadedAssignment could stop before the Agent::run() completed and created a _scriptEngine or _pingTimer

assignment-client/src/Agent.cpp
```diff
@@ -364,10 +364,14 @@ void Agent::processAgentAvatarAndAudio(float deltaTime) {
...
+        _scriptEngine->stop();
+    }
 
-    _pingTimer->stop();
-    delete _pingTimer;
+    if (_pingTimer) {
+        _pingTimer->stop();
+        delete _pingTimer;
```
huffman: Should `_pingTimer` be set to `nullptr` after deleting? I doubt it would cause a problem either way.

ZappoMan: I think it's safe as is. this is essentially the last thing that gets done before shutting down.

hifi-gustavo:
Build console output for commit: 34dc8bdbcb8de07f6e6726efefd9462c67f241d5 available at: https://gist.github.com/a1b5e6104780d3272aea<br/><br/>No installer was built for this commit

PR #5965, opened 2015-09-30
Bug fix for deriveBodyFromHMDSensor calculations

hyperlogic:
The calculation that determined where the body position relative to the HMD
was incorrect, one of the components was in the wrong coordinate frame.

Now use the skeleton's bind pose to compute the proper avatar offsets for the eyes, neck and hips.
Use these offsets to calculate where the hips should be given a specific hmd position and orientation.

Also, added a bug fix for Rig, which would cause a -1 index deference when an avatar was missing
certain named joints, such as, "LeftEye", "Neck" and "Head".

Also, a the check for determining when the desired body position is too far from the current, has been changed from a sphere check to a capsule check.  Even though at the moment the capsule is tuned to be more or less spherical.  We might want to tune this capsule in the future.

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -271,6 +271,25 @@ glm::mat4 MyAvatar::getSensorToWorldMatrix() const {
     return _sensorToWorldMatrix;
 }
 
+// returns true if pos is OUTSIDE of the vertical capsule
+// where the middle cylinder length is defined by capsuleLen and the radius by capsuleRad.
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: d993d397184a24d3d7b742fcadee6821cc96d10c available at: https://gist.github.com/17e041d94d978f080b53<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5965.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5965.dmg

RyanDowne:
Tested on Windows and it works well. Nice result.

PR #5967, opened 2015-10-01
allow physical objects to be set collisionless

AndrewMeadows:
This PR makes it so that the **ignoreForCollisions** entity property means: use COLLISION_GROUP_COLLISIONLESS for such objects -- they are still in the physics engine but have a collisionless group.

The idea is that the grabbed object can be set collisionless (set **ignoreForCollisions** true) while grabbed and then set collidable again (set **ignoreForCollisions** false) when done grabbing.

### DIFF
--- a/libraries/entities/src/EntityItem.h
+++ b/libraries/entities/src/EntityItem.h
@@ -334,7 +334,7 @@ public:
     bool getCollisionsWillMove() const { return _collisionsWillMove; }
     void setCollisionsWillMove(bool value) { _collisionsWillMove = value; }
 
-    virtual bool shouldBePhysical() const { return !_ignoreForCollisions; }
+    virtual bool shouldBePhysical() const { return false; }
 
     bool getLocked() const { return _locked; }
     void setLocked(bool value) { _locked = value; }
--- a/libraries/entities/src/ModelEntityItem.cpp
+++ b/libraries/entities/src/ModelEntityItem.cpp
@@ -457,5 +457,5 @@ QString ModelEntityItem::getAnimationSettings() const {
 
 // virtual
 bool ModelEntityItem::shouldBePhysical() const {
-    return EntityItem::shouldBePhysical() && getShapeType() != SHAPE_TYPE_NONE;
+    return getShapeType() != SHAPE_TYPE_NONE;
 }
--- a/libraries/physics/src/EntityMotionState.cpp
+++ b/libraries/physics/src/EntityMotionState.cpp
@@ -596,6 +596,12 @@ QString EntityMotionState::getName() {
 
 // virtual
 int16_t EntityMotionState::computeCollisionGroup() {
+    if (!_entity) {
+        return COLLISION_GROUP_STATIC;
+    }
+    if (_entity->getIgnoreForCollisions()) {
+        return COLLISION_GROUP_COLLISIONLESS;
+    }
     switch (computeObjectMotionType()){
         case MOTION_TYPE_STATIC:
             return COLLISION_GROUP_STATIC;
--- a/libraries/physics/src/ObjectMotionState.h
+++ b/libraries/physics/src/ObjectMotionState.h
@@ -37,11 +37,11 @@ enum MotionStateType {
 
 // The update flags trigger two varieties of updates: "hard" which require the body to be pulled 
 // and re-added to the physics engine and "easy" which just updates the body properties.
-const uint32_t HARD_DIRTY_PHYSICS_FLAGS = (uint32_t)(EntityItem::DIRTY_MOTION_TYPE | EntityItem::DIRTY_SHAPE);
+const uint32_t HARD_DIRTY_PHYSICS_FLAGS = (uint32_t)(EntityItem::DIRTY_MOTION_TYPE | EntityItem::DIRTY_SHAPE | 
+                                                     EntityItem::DIRTY_COLLISION_GROUP);
 const uint32_t EASY_DIRTY_PHYSICS_FLAGS = (uint32_t)(EntityItem::DIRTY_TRANSFORM | EntityItem::DIRTY_VELOCITIES |
-                                                     EntityItem::DIRTY_MASS | EntityItem::DIRTY_COLLISION_GROUP |
-                                                     EntityItem::DIRTY_MATERIAL | EntityItem::DIRTY_SIMULATOR_ID | 
-                                                     EntityItem::DIRTY_SIMULATOR_OWNERSHIP);
+                                                     EntityItem::DIRTY_MASS | EntityItem::DIRTY_MATERIAL | 
+                                                     EntityItem::DIRTY_SIMULATOR_ID | EntityItem::DIRTY_SIMULATOR_OWNERSHIP);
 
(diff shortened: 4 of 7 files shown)


libraries/entities/src/EntityItem.h
```diff
@@ -334,7 +334,7 @@ class EntityItem : public std::enable_shared_from_this<EntityItem>, public ReadW
     bool getCollisionsWillMove() const { return _collisionsWillMove; }
     void setCollisionsWillMove(bool value) { _collisionsWillMove = value; }
 
-    virtual bool shouldBePhysical() const { return !_ignoreForCollisions; }
+    virtual bool shouldBePhysical() const { return false; }
```
AndrewMeadows: By default all EntityItem types are NOT in physics engine, except for: Box, Sphere, Model, and PolyVox.

Note: ModelEntityItem is special in that it already had a valid **shouldBePhysical()** method that works (did not need to be changed).

libraries/physics/src/EntityMotionState.cpp
```diff
@@ -596,6 +596,12 @@ QString EntityMotionState::getName() {
 
 // virtual
 int16_t EntityMotionState::computeCollisionGroup() {
+    if (!_entity) {
+        return COLLISION_GROUP_STATIC;
+    }
+    if (_entity->getIgnoreForCollisions()) {
+        return COLLISION_GROUP_COLLISIONLESS;
+    }
```
AndrewMeadows: Added special case for when getIgnoreForCollisions() is set --> use collisionless group no matter if it is dynamic or static.

libraries/physics/src/ObjectMotionState.h
```diff
@@ -37,11 +37,11 @@ enum MotionStateType {
...
 // and re-added to the physics engine and "easy" which just updates the body properties.
-const uint32_t HARD_DIRTY_PHYSICS_FLAGS = (uint32_t)(EntityItem::DIRTY_MOTION_TYPE | EntityItem::DIRTY_SHAPE);
+const uint32_t HARD_DIRTY_PHYSICS_FLAGS = (uint32_t)(EntityItem::DIRTY_MOTION_TYPE | EntityItem::DIRTY_SHAPE | 
+                                                     EntityItem::DIRTY_COLLISION_GROUP);
 const uint32_t EASY_DIRTY_PHYSICS_FLAGS = (uint32_t)(EntityItem::DIRTY_TRANSFORM | EntityItem::DIRTY_VELOCITIES |
-                                                     EntityItem::DIRTY_MASS | EntityItem::DIRTY_COLLISION_GROUP |
-                                                     EntityItem::DIRTY_MATERIAL | EntityItem::DIRTY_SIMULATOR_ID | 
-                                                     EntityItem::DIRTY_SIMULATOR_OWNERSHIP);
+                                                     EntityItem::DIRTY_MASS | EntityItem::DIRTY_MATERIAL | 
+                                                     EntityItem::DIRTY_SIMULATOR_ID | EntityItem::DIRTY_SIMULATOR_OWNERSHIP);
```
AndrewMeadows: The DIRTY_COLLISION_GROUP bit was incorrectly categorized as an EASY change -- it is HARD because the object must be removed from the DynamicsWorld and re-added in order for its new collision group to register (the collision group is cached by the broadphase data structure on add and there is no way to update it on the fly except to reinsert).

libraries/entities/src/ModelEntityItem.cpp
```diff
@@ -457,5 +457,5 @@ QString ModelEntityItem::getAnimationSettings() const {
 
 // virtual
 bool ModelEntityItem::shouldBePhysical() const {
-    return EntityItem::shouldBePhysical() && getShapeType() != SHAPE_TYPE_NONE;
+    return getShapeType() != SHAPE_TYPE_NONE;
 }
```
AndrewMeadows: I made an earlier comment saying that ModelEntityItem::shouldBePhysical() didn't need to be changed... but I realized that it did: to not call the method from its base class.

hifi-gustavo:
Build console output for commit: 71fd229b294b6b6b0c8a2e417423298bebd161b6 available at: https://gist.github.com/796824b6cb84f74161c3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5967.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5967.dmg

hifi-gustavo:
Build console output for commit: 1e8643d90fe6098c9fa5005a559be41830c73171 available at: https://gist.github.com/796824b6cb84f74161c3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5967.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5967.dmg

hifi-gustavo:
Build console output for commit: 94d613e3fe672143f3b61a0c39c9f241f73e9d63 available at: https://gist.github.com/796824b6cb84f74161c3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5967.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5967.dmg

hifi-gustavo:
Build console output for commit: 381c98c4fa268b09ec81c85d38ad4a8676a58db0 available at: https://gist.github.com/796824b6cb84f74161c3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5967.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5967.dmg

howard-stearns:
How should we test this?

AndrewMeadows:
One noticeable difference with this PR is that **actions** should work on entities that have **ignoreForCollisions** set **true** -- this was not the case prior.  Otherwise these changes provide no changes visible to the user.  Nevertheless it would be nice to verify that nothing broke:

(1) Make a static collidable entity --> verify that things collide against it
(2) Use edit.js to set **ignoreForCollisions** to be **true** --> verify that nothing collides with it
(3) Set **ignoreForCollisions** false again --> verify that it things collide with it again.
(4) Create a dynamic object near the collisionless one at some known offset (e.g. a few meters along +zAxis), give the collisionless object velocity toward the other (with non-zero linear damping so that it will eventually stop) --> the collisionless object should move kinematically and collide with the dynamic one.
(5) Set the static object collisionless and repeat step (4) --> it should move kinematically but NOT collide with the dynamic object.

(6) Make the collisionless object dynamic but keep its gravity at zero.  Use one of the grab example scripts that use actions on dynamic objects --> the object should be movable by actions (this should not have been true prior, although it may be that the object was movable using the non-dynamic code paths of the grab script (if any)).

howard-stearns:
The current version of this PR is not behaving quite right with wrt to #6 in the test plan, above. Andrew is on it...

howard-stearns:
Cannot reproduced the failure. Pilot error? Anyway, we'll merge...

PR #5979, opened 2015-10-02
[Script] Updates to Ping Pong Gun

imgntn:
This PR reduces the force of the shot so that the ball goes through fewer objects.  It also tries to respect only the initial hand when the gun is released.


examples/toys/ping_pong_gun/pingPongGun.js
```diff
@@ -82,9 +82,14 @@
 
         releaseGrab: function() {
             var _t = this;
```
ZappoMan: coding standard: variable names should be descriptive... make this `_this` instead of `_t`

imgntn:
bump for review :)

PR #5983, opened 2015-10-02
Light fixes in toybox

ericrius1:
Light switches is toybox now work without adding any extra lights, no matter how many users there are in toybox.


examples/toys/lightSwitch.js
```diff
@@ -0,0 +1,92 @@
+//
+//  lightSwitch.js.js
```
ZappoMan: .js.js?

examples/toys/lightSwitch.js
```diff
@@ -0,0 +1,92 @@
...
+/*global  LightSwitch */
+
+(function () {
+    var _this;
+    var utilitiesScript = Script.resolvePath("../libraries/utils.js");
+    Script.include(utilitiesScript);
+    LightSwitch = function () {
+        _this = this;
+        this.switchSound = SoundCache.getSound("https://hifi-public.s3.amazonaws.com/sounds/Switches%20and%20sliders/lamp_switch_2.wav");
+    }
```
ZappoMan: I think we want a ; here

PR #5991, opened 2015-10-05
[Scripts] Test Block Spawner for Blockers Game

imgntn:
This PR adds a script that creates a set of blocks in a 'Stonehenge' formation which are used to to test different block property combinations to find the best fit for the game.  

https://rawgit.com/imgntn/hifi/knockover_blocks/examples/toys/blockers/createTestBlocks.js


examples/toys/blockers/createTestBlocks.js
```diff
@@ -0,0 +1,156 @@
...
+});
+
+function cleanUp() {
+    Entities.deleteEntity(ground);
+    Entities.deleteEntity(topBlock);
+    Entities.deleteEntity(sideBlock1);
+    Entities.deleteEntity(sideBlock2);
+}
+
+// Script.scriptEnding.connect(cleanUp);
```
Atlante45: Should that be commented?

imgntn: no, probably just deleted along with the block above it-- we decided that for scripts that spawn entities, as a rule we aren't cleaning them up when the script ends.  good find!

PR #5994, opened 2015-10-05
Display name in terms of arc size not pixels

Atlante45:
This makes the display names size only dependent on a vertical arc size (ie. 2 degrees)

### DIFF
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -696,34 +696,51 @@ void Avatar::renderBillboard(RenderArgs* renderArgs) {
 }
 
 float Avatar::getBillboardSize() const {
-    return _scale * BILLBOARD_DISTANCE * tanf(glm::radians(BILLBOARD_FIELD_OF_VIEW / 2.0f));
+    return _scale * BILLBOARD_DISTANCE * glm::tan(glm::radians(BILLBOARD_FIELD_OF_VIEW / 2.0f));
 }
+
+#ifdef DEBUG
+void debugValue(const QString& str, const glm::vec3& value) {
+    if (glm::any(glm::isnan(value)) || glm::any(glm::isinf(value))) {
+        qCWarning(interfaceapp) << "debugValue() " << str << value;
+    }
+};
+void debugValue(const QString& str, const float& value) {
+    if (glm::isnan(value) || glm::isinf(value)) {
+        qCWarning(interfaceapp) << "debugValue() " << str << value;
+    }
+};
+#define DEBUG_VALUE(str, value) debugValue(str, value)
+#else
+#define DEBUG_VALUE(str, value)
+#endif
 
 glm::vec3 Avatar::getDisplayNamePosition() const {
     glm::vec3 namePosition(0.0f);
+    glm::vec3 bodyUpDirection = getBodyUpDirection();
+    DEBUG_VALUE("bodyUpDirection =", bodyUpDirection);
+    
     if (getSkeletonModel().getNeckPosition(namePosition)) {
-        namePosition += getBodyUpDirection() * getHeadHeight() * 1.1f;
+        float headHeight = getHeadHeight();
+        DEBUG_VALUE("namePosition =", namePosition);
+        DEBUG_VALUE("headHeight =", headHeight);
+        
+        namePosition += bodyUpDirection * headHeight * 1.1f;
     } else {
         const float HEAD_PROPORTION = 0.75f;
-        namePosition = _position + getBodyUpDirection() * (getBillboardSize() * HEAD_PROPORTION);
+        float billboardSize = getBillboardSize();
+        
+        DEBUG_VALUE("_position =", _position);
+        DEBUG_VALUE("billboardSize =", billboardSize);
+        namePosition = _position + bodyUpDirection * (billboardSize * HEAD_PROPORTION);
     }
-#ifdef DEBUG
-    // TODO: Temporary logging to track cause of invalid scale value; remove once cause has been fixed.
-    // See other TODO below.
-    if (glm::isnan(namePosition.x) || glm::isnan(namePosition.y) || glm::isnan(namePosition.z)
-        || glm::isinf(namePosition.x) || glm::isinf(namePosition.y) || glm::isinf(namePosition.z)) {
-        qDebug() << "namePosition =" << namePosition;
-        glm::vec3 tempPosition(0.0f);
-        if (getSkeletonModel().getNeckPosition(tempPosition)) {
-            qDebug() << "getBodyUpDirection() =" << getBodyUpDirection();
-            qDebug() << "getHeadHeight() =" << getHeadHeight();
-        } else {
-            qDebug() << "_position =" << _position;
-            qDebug() << "getBodyUpDirection() =" << getBodyUpDirection();
-            qDebug() << "getBillboardSize() =" << getBillboardSize();
-        }
+    
+    if (glm::any(glm::isnan(namePosition)) || glm::any(glm::isinf(namePosition))) {
+        qCWarning(interfaceapp) << "Invalid display name position" << namePosition
+                                << ", setting is to (0.0f, 0.5f, 0.0f)";
+        namePosition = glm::vec3(0.0f, 0.5f, 0.0f);
     }
-#endif
+    
     return namePosition;
 }
 
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -552,13 +552,7 @@ void Avatar::render(RenderArgs* renderArgs, const glm::vec3& cameraPosition) {
         auto& frustum = *renderArgs->_viewFrustum;
         auto textPosition = getDisplayNamePosition();
         
-        // In the view frustum
-        auto keyhole = frustum.getKeyholeRadius();
-        frustum.setKeyholeRadius(-1.0f);
-        auto inFrustum = frustum.pointInFrustum(textPosition) == ViewFrustum::INSIDE;
-        frustum.setKeyholeRadius(keyhole);
-        
-        if (inFrustum) {
+        if (frustum.pointInFrustum(textPosition, true) == ViewFrustum::INSIDE) {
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: 1be702f02466a22fc5ea32dc1cb49237472e162c available at: https://gist.github.com/0ff288db4f9b1a4f65b8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5994.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5994.dmg

hifi-gustavo:
Build console output for commit: 6d162d81d0865eae398abc33891dbf1719a41880 available at: https://gist.github.com/0ff288db4f9b1a4f65b8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5994.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5994.dmg

hifi-gustavo:
Build console output for commit: 11ab7280f9758f969d0fdfbd57135365eca7d2c7 available at: https://gist.github.com/0ff288db4f9b1a4f65b8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5994.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5994.dmg

interface/src/avatar/Avatar.cpp
```diff
@@ -541,7 +541,18 @@ void Avatar::render(RenderArgs* renderArgs, const glm::vec3& cameraPosition) {
 
     auto cameraMode = Application::getInstance()->getCamera()->getMode();
     if (!isMyAvatar() || cameraMode != CAMERA_MODE_FIRST_PERSON) {
-        renderDisplayName(batch, *renderArgs->_viewFrustum, renderArgs->_viewport);
+        auto& frustum = *renderArgs->_viewFrustum;
```
ZappoMan: this seems like a dangerous approach here... instead of modifying the actual `renderArgs->_viewFrustum` why don't you make a copy of it, modify that copy, use it, and discard it.

interface/src/avatar/Avatar.cpp
```diff
@@ -677,116 +688,77 @@ void Avatar::renderBillboard(RenderArgs* renderArgs) {
...
+    glm::vec3 bodyUpDirection = getBodyUpDirection();
+    DEBUG_VALUE("bodyUpDirection =", bodyUpDirection);
+    
     if (getSkeletonModel().getNeckPosition(namePosition)) {
-        namePosition += getBodyUpDirection() * getHeadHeight() * 1.1f;
+        float headHeight = getHeadHeight();
+        DEBUG_VALUE("namePosition =", namePosition);
+        DEBUG_VALUE("headHeight =", headHeight);
+        
+        namePosition += bodyUpDirection * headHeight * 1.1f;
```
ZappoMan: coding standard: magic number

interface/src/avatar/Avatar.cpp
```diff
@@ -677,116 +688,77 @@ void Avatar::renderBillboard(RenderArgs* renderArgs) {
...
-    glm::vec3 dPosition = frustum.getPosition() - getPosition();
-    // If x and z are 0, atan(x, z) is undefined, so default to 0 degrees
-    float yawRotation = dPosition.x == 0.0f && dPosition.z == 0.0f ? 0.0f : glm::atan(dPosition.x, dPosition.z);
+    // If x and z are 0, atan(x, z) adais undefined, so default to 0 degrees
+    const float yawRotation = (dPosition.x == 0.0f && dPosition.z == 0.0f) ? 0.0f : glm::atan(dPosition.x, dPosition.z);
     glm::quat orientation = glm::quat(glm::vec3(0.0f, yawRotation, 0.0f));
     
-    // Set transform (The order IS important)
+    // Compute correct scale to apply
+    static const float DESIRED_HIGHT_RAD = glm::radians(2.0f);
```
ZappoMan: spelling? HIGHT?

interface/src/avatar/Avatar.cpp
```diff
@@ -677,116 +688,77 @@ void Avatar::renderBillboard(RenderArgs* renderArgs) {
...
-    
-    // Compute pixel alignment offset
-    float clipToPix = 0.5f * windowSizeY / p1.w; // Got from clip to pixel coordinates
-    glm::vec4 screenPos = clipToPix * p1; // in pixels coords
-    glm::vec4 screenOffset = (glm::round(screenPos) - screenPos) / clipToPix; // in clip coords
-    glm::vec3 worldOffset = glm::vec3(screenOffset.x, screenOffset.y, 0.0f) / (float)pixelHeight;
+Transform Avatar::calculateDisplayNameTransform(const ViewFrustum& frustum, const glm::vec3& textPosition) const {
+    Q_ASSERT_X(frustum.pointInFrustum(textPosition) == ViewFrustum::INSIDE,
+               "Avatar::calculateDisplayNameTransform", "Text not in viewfrustum.");
+    glm::vec3 dPosition = frustum.getPosition() - textPosition;
```
ZappoMan: coding standard: variable names should be descriptive... what is "d"-Position?

interface/src/avatar/Avatar.cpp
```diff
@@ -810,39 +782,43 @@ void Avatar::renderDisplayName(gpu::Batch& batch, const ViewFrustum& frustum, co
...
-    renderer->draw(batch, text_x, -text_y, nameUTF8.data(), textColor);
+    const glm::vec2 extent = renderer->computeExtent(renderedDisplayName);
+    if (!glm::any(glm::isCompNull(extent, EPSILON))) {
+        const QRect nameDynamicRect = QRect(0, 0, (int)extent.x, (int)extent.y);
+        const int text_x = -nameDynamicRect.width() / 2;
+        const int text_y = -nameDynamicRect.height() / 2;
+        
+        // Compute background position/size
+        static const float SLIGHTLY_IN_FRONT = 0.1f;
+        const int border = 0.1f * nameDynamicRect.height();
```
ZappoMan: coding standard: magic number... what is 0.1f???

interface/src/avatar/Avatar.cpp
```diff
@@ -810,39 +782,43 @@ void Avatar::renderDisplayName(gpu::Batch& batch, const ViewFrustum& frustum, co
...
+        const int left = text_x - border;
+        const int bottom = text_y - border;
+        const int width = nameDynamicRect.width() + 2.0f * border;
+        const int height = nameDynamicRect.height() + 2.0f * border;
+        const int bevelDistance = 0.1f * height;
```
ZappoMan: coding standard: magic number... what is 0.1f???

interface/src/avatar/Avatar.cpp
Atlante45: It's the difference of position.
Like you would write `dx`

hifi-gustavo:
Build console output for commit: 90957912a0505c60eac4c4e6456e7bd214f95332 available at: https://gist.github.com/0ff288db4f9b1a4f65b8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5994.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5994.dmg

hifi-gustavo:
Build console output for commit: 454dbfd3cbcf2cd59bf6c3b6fdfaaf9316b9649a available at: https://gist.github.com/0ff288db4f9b1a4f65b8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5994.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5994.dmg

ZappoMan:
I tested this PR last night, and it seems to create some strange effects that I don't think are very desirable... here's a list:

1) The names are sunken down into the heads of avatars, so they are actually cut off by the mesh
2) The names appear to be very very small when you close to an avatar
3) The names appear to be very very very large when you are far away from an avatar... in fact, the names almost fill the entire screen.

ZappoMan:
also - can you pull from master, because the ADT packet format change, this build can no longer connect to domains

hifi-gustavo:
Build console output for commit: fb284448b5f07edc05e03cb2ecda7c2820f4b2d1 available at: https://gist.github.com/0ff288db4f9b1a4f65b8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5994.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5994.dmg

Atlante45:
1) This happens on master to, it mainly depends on your avatar, and also how far you are.
[Reading your comment made me think of a fix so though.](https://github.com/highfidelity/hifi/commit/fb284448b5f07edc05e03cb2ecda7c2820f4b2d1)

2/3) The display names always have the exact same size on screen no matter how close/far the avatar is. This is what was happening before and from what I understand the desired behavior.
If we don't want that then all this scaling code is unnecessary and it would be much easier to compute the transform.
 I'm not opposed to that approach but I do think it's nice to have the names all have the same size. One improvement we could make though is borders to make them more readable when a few overlap.

The main difference with that PR, is that the actual size of display name is computed from a vertical angular size, meaning it'll be the same no matter the resolution of the screen and we also don't need the projection matrix to compute it.
That also means that the size of the display names will change with the height of the window like the rest of the scene.

ZappoMan:
re: 2/3 - ok - sounds reasonable... I thought you were trying to fix this behavior with this PR... you're right it seems to behave similar to master... so after I smoke test, I'll merge.

ZappoMan:
I will merge... but now it seems like the names are MUCH higher than they used to be.... don't be surprised if someone says they don't like that. :)

Atlante45:
@ZappoMan If people really don't like that, it's easily revert-able. One line to delete ^^

PR #5997, opened 2015-10-05
DO NOT MERGE - QA for 20699 - Sanitize particle property values and usage

ctrlaltdavid:


libraries/entities/src/ParticleEffectEntityItem.cpp
```diff
@@ -798,9 +875,12 @@ void ParticleEffectEntityItem::stepSimulation(float deltaTime) {
...
 }
 
-void ParticleEffectEntityItem::setMaxParticles(quint32 maxParticles) {
-    if (_maxParticles != maxParticles) {
-        _maxParticles = maxParticles;
+void ParticleEffectEntityItem::setMaxParticles(float maxParticles) {
+    // Sanitize JavaScript signed float value to quint32 value
+    maxParticles = glm::clamp(maxParticles, (float)MINIMUM_MAX_PARTICLES, (float)MAXIMUM_MAX_PARTICLES);
+
+    if (_maxParticles != (quint32)maxParticles) {
```
ZappoMan: ??? Is this really the right fix?

libraries/entities/src/ParticleEffectEntityItem.h
```diff
@@ -117,18 +121,26 @@ class ParticleEffectEntityItem : public EntityItem {
     float getAnimationLastFrame() const { return _animationLoop.getLastFrame(); }
 
     static const quint32 DEFAULT_MAX_PARTICLES;
-    void setMaxParticles(quint32 maxParticles);
-    quint32 getMaxParticles() const { return _maxParticles; }
+    static const quint32 MINIMUM_MAX_PARTICLES;
+    static const quint32 MAXIMUM_MAX_PARTICLES;
+    void setMaxParticles(float maxParticles);
+    float getMaxParticles() const { return (float)_maxParticles; }
```
ZappoMan: I don't understand why you're storing maxParticicles internally as quint32 and exposing them to callers as floats???

libraries/entities/src/ParticleEffectEntityItem.cpp
```diff
@@ -285,7 +351,7 @@ int ParticleEffectEntityItem::readEntitySubclassDataFromBuffer(const unsigned ch
     }
     READ_ENTITY_PROPERTY(PROP_ANIMATION_SETTINGS, QString, setAnimationSettings);
     READ_ENTITY_PROPERTY(PROP_SHAPE_TYPE, ShapeType, updateShapeType);
-    READ_ENTITY_PROPERTY(PROP_MAX_PARTICLES, quint32, setMaxParticles);
+    READ_ENTITY_PROPERTY(PROP_MAX_PARTICLES, float, setMaxParticles);
```
ZappoMan: I believe this will cause problems with older streams...

your older streams think these are quint32s, and they were written as such... but you're parsing them as a float... that's going to create wild random numbers.

hifi-gustavo:
Build console output for commit: efd1688b405c540ea8abe0480278410db4d79c2c available at: https://gist.github.com/48b19a0faa5f82ee2bf3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5997.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5997.dmg

ctrlaltdavid:
@ZappoMan In C++, the _maxParticles property is used to size a bunch of QVectors. But all JavaScript numbers are signed floats so we need to handle them as such (e.g., handle -ve values).

Previously, there was a mixture of quint32s and floats being used in the various property macro calls for this property, which caused problems (e.g., not actually being able to change the maximum number of particles in JavaScript). Now they are all floats.

Both quint32s and floats are 4 bytes but, as you say, it could possibly cause problems mixing old and new Interfaces / assignment clients. Perhaps I should bump the PacketVersion?

ZappoMan:
you need to do more than just bump the packet version, you need to detect the old version of the stream and it accordingly. I guess I'm saying that seems like more work than just solving this on the JS side of things. it doesn't make sense to have a negative or fractional "max packets" on JS... so this seems like the wrong solution.

ZappoMan:
In other words: make them all uint32s and I think you've got an easier path without a broken/changed binary stream...

ctrlaltdavid:
I guess I didn't figure that this change to the binary stream was that big a deal (i.e., no need to process old streams) given that we're alpha, it's only particles, the number of bytes is the same, and any transient problem encountered will be fixed by updating to the latest Interface and assignment client. Besides, I doubt that anyone's actually been using the maxParticles property to date because it didn't actually work properly.

Nevertheless, I've updated to use quint32 instead of float. The resulting JavaScript behaviour, I think, is not as nice because if you set a value < 0 or > 2^32 - 1, the value is left unchanged rather than being clamped to the valid maxParticles range or 1 - 10,000.

(Alternatively, could make it so that values < 0 or > 2^32 - 1 result in setting the property to 10,000.)

hifi-gustavo:
Build console output for commit: 53d899cb7d29bbd714d00e08ba18fc2ef1532b85 available at: https://gist.github.com/48b19a0faa5f82ee2bf3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5997.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5997.dmg

ZappoMan:
Why can't you clamp it to 0 to max uint32?

ctrlaltdavid:
After messing about I've concluded that the current stream behaviour using quint32 is "correct" when compared with similar conversions from JavaScript values for other entity properties; namely, if the JavaScript value is not valid or out of range the stream value is not updated.

Therefore, I should probably change particle properties to not update if the value is out of range rather than clamp to valid values. I'll suggest this to Chris etc.

murillodigital:
test this please

hifi-gustavo:
Build console output for commit: 53d899cb7d29bbd714d00e08ba18fc2ef1532b85 available at: https://gist.github.com/48b19a0faa5f82ee2bf3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5997.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-5997.dmg

libraries/entities/src/ParticleEffectEntityItem.cpp
```diff
@@ -114,37 +135,82 @@ ParticleEffectEntityItem::~ParticleEffectEntityItem() {
...
 }
 
 
 void ParticleEffectEntityItem::setEmitDimensions(const glm::vec3& emitDimensions) {
-    _emitDimensions = emitDimensions;
+    _emitDimensions = glm::vec3(
+        glm::clamp(emitDimensions.x, MINIMUM_EMIT_DIMENSION, MAXIMUM_EMIT_DIMENSION),
+        glm::clamp(emitDimensions.y, MINIMUM_EMIT_DIMENSION, MAXIMUM_EMIT_DIMENSION),
+        glm::clamp(emitDimensions.z, MINIMUM_EMIT_DIMENSION, MAXIMUM_EMIT_DIMENSION)
+    );
```
ZappoMan: coding standard: 4.2.8.... "When the multiple lines are bound by parentheses (as in arguments to a function call), the prefered style has no whitespace after the opening parenthesis or before the closing parenthesis. The n+1 lines are generally indented to the column immediately after the opening parenthesis (following the style for split expressions in 2.1.6)."

So... you should use:

```
_emitDimensions = glm::vec3(glm::clamp(emitDimensions.x, MINIMUM_EMIT_DIMENSION, MAXIMUM_EMIT_DIMENSION),
        glm::clamp(emitDimensions.y, MINIMUM_EMIT_DIMENSION, MAXIMUM_EMIT_DIMENSION),
        glm::clamp(emitDimensions.z, MINIMUM_EMIT_DIMENSION, MAXIMUM_EMIT_DIMENSION));
```

libraries/entities/src/ParticleEffectEntityItem.cpp
```diff
@@ -114,37 +135,82 @@ ParticleEffectEntityItem::~ParticleEffectEntityItem() {
...
     computeAndUpdateDimensions();
 }
 
+void ParticleEffectEntityItem::setEmitRadiusStart(float emitRadiusStart) {
+    _emitRadiusStart = glm::clamp(emitRadiusStart, MINIMUM_EMIT_RADIUS_START, MAXIMUM_EMIT_RADIUS_START); 
+}
+
 void ParticleEffectEntityItem::setEmitAcceleration(const glm::vec3& emitAcceleration) {
-    _emitAcceleration = emitAcceleration;
+    _emitAcceleration = glm::vec3(
```
ZappoMan: coding standard: 4.2.8.... "When the multiple lines are bound by parentheses (as in arguments to a function call), the prefered style has no whitespace after the opening parenthesis or before the closing parenthesis. The n+1 lines are generally indented to the column immediately after the opening parenthesis (following the style for split expressions in 2.1.6)."

libraries/entities/src/ParticleEffectEntityItem.cpp
```diff
@@ -114,37 +135,82 @@ ParticleEffectEntityItem::~ParticleEffectEntityItem() {
...
+        glm::clamp(emitAcceleration.x, MINIMUM_EMIT_ACCELERATION, MAXIMUM_EMIT_ACCELERATION),
+        glm::clamp(emitAcceleration.x, MINIMUM_EMIT_ACCELERATION, MAXIMUM_EMIT_ACCELERATION),
+        glm::clamp(emitAcceleration.y, MINIMUM_EMIT_ACCELERATION, MAXIMUM_EMIT_ACCELERATION)
+    );
     computeAndUpdateDimensions();
 }
 
 void ParticleEffectEntityItem::setAccelerationSpread(const glm::vec3& accelerationSpread){
-    _accelerationSpread = accelerationSpread;
+    _accelerationSpread = glm::vec3(
```
ZappoMan: coding standard: 4.2.8.... "When the multiple lines are bound by parentheses (as in arguments to a function call), the prefered style has no whitespace after the opening parenthesis or before the closing parenthesis. The n+1 lines are generally indented to the column immediately after the opening parenthesis (following the style for split expressions in 2.1.6)."

libraries/entities/src/ParticleEffectEntityItem.cpp
```diff
@@ -672,10 +738,18 @@ void ParticleEffectEntityItem::stepSimulation(float deltaTime) {
                 _radiusMiddles[i] =_particleRadius;
                 _radiusFinishes[i] = getRadiusFinish();
             } else {
-                float spreadMultiplier = 1.0f + (2.0f * randFloat() - 1.0f) * _radiusSpread / _particleRadius;
-                _radiusStarts[i] = spreadMultiplier * getRadiusStart();
-                _radiusMiddles[i] = spreadMultiplier * _particleRadius;
-                _radiusFinishes[i] = spreadMultiplier * getRadiusFinish();
+                float spreadMultiplier;
+                if (_particleRadius > 0.0f) {
+                    spreadMultiplier = 1.0f + (2.0f * randFloat() - 1.0f) * _radiusSpread / _particleRadius;
```
ZappoMan: coding standard: magic numbers... maybe used named constants like max and min or something

libraries/entities/src/ParticleEffectEntityItem.cpp
```diff
@@ -753,9 +827,12 @@ void ParticleEffectEntityItem::stepSimulation(float deltaTime) {
                 xColor finishColor = getColorFinish();
 
                 float spread = 2.0f * randFloat() - 1.0f;
```
ZappoMan: coding standard: magic numbers... maybe used named constants like max and min or something

libraries/entities/src/ParticleEffectEntityItem.cpp
```diff
@@ -753,9 +827,12 @@ void ParticleEffectEntityItem::stepSimulation(float deltaTime) {
...
-                float spreadMultiplierRed = 1.0f + spread * (float)_colorSpread.red / (float)middleColor.red;
-                float spreadMultiplierGreen = 1.0f + spread * (float)_colorSpread.green / (float)middleColor.green;
-                float spreadMultiplierBlue = 1.0f + spread * (float)_colorSpread.blue / (float)middleColor.blue;
+                float spreadMultiplierRed = 
+                    middleColor.red > 0 ? 1.0f + spread * (float)_colorSpread.red / (float)middleColor.red : 1.0f;
+                float spreadMultiplierGreen = 
+                    middleColor.green > 0 ? 1.0f + spread * (float)_colorSpread.green / (float)middleColor.green :1.0f;
```
ZappoMan: space after :

ZappoMan:
@ctrlaltdavid - are you reopening this PR? If so... please note the comments on coding standard issues I left.

PR #6005, opened 2015-10-06
Fix for bug where Script.resolvePath returned an incorrect URL in some cases

ericrius1:
There was a bug when an entity script which was attached to an entity from a parent entity script and called Script.resolvePath, it would have the incorrect path returned. This is now fixed


libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -974,9 +974,6 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
     QScriptValue entityScriptObject = entityScriptConstructor.construct();
     EntityScriptDetails newDetails = { scriptOrURL, entityScriptObject };
     _entityScripts[entityID] = newDetails;
-    if (isURL) {
```
ZappoMan: hmmmm - not sure this is the right fix... maybe we need a stack of parent urls?

ZappoMan: I think this will cause other bugs... but it does highlight the problem you found.

I think the correct fix is to add a `QStack<QString> _parentScriptStack` and at line 958 you should do something like `_parentScriptStack.push(getParentURL());` then the call to `setParentURL(scriptOrURL);`

Then here you want to do something like...

```
 if (isURL) {
    QString lastParent = _parentScriptStack.pop();
    setParentURL(lastParent);
}
```

hifi-gustavo:
Build console output for commit: 713499d6d05a418bdc5378d591dbf61fdfc51bb9 available at: https://gist.github.com/8ecfab35bc97617ac88f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6005.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6005.dmg

libraries/script-engine/src/ScriptEngine.cpp
ericrius1: @ZappoMan - On line 958 I didn't have success with pushing the parentURL onto the stack, but I did pushing scriptOrURL. Tested with all of the entities in toybox without issue.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -974,10 +975,12 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
     QScriptValue entityScriptObject = entityScriptConstructor.construct();
     EntityScriptDetails newDetails = { scriptOrURL, entityScriptObject };
     _entityScripts[entityID] = newDetails;
+
     if (isURL) {
-        setParentURL("");
+        QString lastParent = _parentScriptStack.pop();
+        setParentURL(lastParent);
```
ZappoMan: you should print out the values you're setting to... it looks to me like this is basically going to set the value to the same value above... so it won't reset it... so basically you're getting the same broken behavior as your first PR

hifi-gustavo:
Build console output for commit: cd3de3175794315cf54550413956e33b677bd48b available at: https://gist.github.com/8ecfab35bc97617ac88f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6005.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6005.dmg

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -955,6 +955,7 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
     }
 
     if (isURL) {
+        _parentScriptStack.push(scriptOrURL);
```
ZappoMan: this really should be the getParentURL()

ericrius1:
After speaking with @huffman, I realized I was not resolving the paths in the correct location in the script. So this wasn't a bug in the C++ after all.

PR #6006, opened 2015-10-06
Animation group settings

ZappoMan:
* fixes bugs in animations skipping when you start/stop or change propeties
* gets rid of the animationSettings as a JSON string, and makes it a true property group like stage, zone, or atmosphere settings
* removes "animation" settings from particle effects and makes a simplified "isEmitting" property instead since the particle effect wasn't actually using any of the other animation properties
* makes the individual animation properties editable in the Entity Properties UI instead of having to know how to properly format the animationSettings string.


hifi-gustavo:
Build console output for commit: 016bf4011d4a7e6d1600c7ffa72a973f8afc0ed0 available at: https://gist.github.com/1447c8cc7b63c5812d08<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6006.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6006.dmg

examples/html/entityProperties.html
```diff
@@ -37,10 +46,11 @@
             };
         }
         function createEmitGroupCheckedPropertyUpdateFunction(group, propertyName) {
-            return function() {
+            return function () {
                 var properties = {};
                 properties[group] = {};
                 properties[group][propertyName] = this.checked;
+                //debugPrint("   about to emitWebEvent() properties:" + JSON.stringify(properties));
```
AndrewMeadows: Was it intentional to leave this commented out debug code?

ZappoMan: Yes. But I will delete it in future PR if you want. I was verifying that the update was actually firing correctly.

examples/example/entities/butterflies.js
```diff
@@ -85,8 +85,17 @@ function addButterfly() {
...
+            url: "http://public.highfidelity.io/models/content/butterfly/butterfly.fbx",
+            firstFrame: 0,
+            fps: newFrameRate,
+            frameIndex: 0,
+            hold: false,
+            lastFrame: 10000,
+            loop: true,
+            running: true,
+            startAutomatically:false
+        },
```
AndrewMeadows: I know that the word **frameIndex** is not new in these changes but it makes me a little sad and this is a good opportunity to clean things up a little bit.  If we want to take advantage of this opportunity I would suggest three possible combinations of names:

frame, firstFrame, lastFrame
currentFrame, firstFrame, lastFrame
currentFrameIndex, firstFrameIndex, lastFrameIndex

ZappoMan: I'm fine with any of those names. It would be easy to fix now before we merge, so let's change it now and not revisit it later. 

I'll let you seek input from the team as to which names are best.

libraries/entities/src/ModelEntityItem.cpp
```diff
@@ -35,9 +30,11 @@ EntityItemPointer ModelEntityItem::factory(const EntityItemID& entityID, const E
 ModelEntityItem::ModelEntityItem(const EntityItemID& entityItemID, const EntityItemProperties& properties) :
         EntityItem(entityItemID)
 {
+    _animationProperties.associateWithAnimationLoop(&_animationLoop);
+    _animationLoop.setResetOnRunning(false);
```
AndrewMeadows: **AnimationLoop** class is initialized with **_resetOnRunning** set to false so setting it false here is redundant.

libraries/animation/src/AnimationLoop.cpp
```diff
@@ -89,9 +102,14 @@ void AnimationLoop::setRunning(bool running) {
         _running = running;
         
         // If we just set running to true, then also reset the frame to the first frame
-        if (running) {
+        if (running && (_resetOnRunning)) {
             // move back to the beginning
             _frameIndex = _firstFrame;
         }
```
AndrewMeadows: I can't find any code that ever sets **_resetOnRunning** to be **true** so the **if** case above will never be true.

libraries/animation/src/AnimationLoop.h
```diff
@@ -41,7 +41,12 @@ class AnimationLoop {
     void setLastFrame(float lastFrame) { _lastFrame = glm::clamp(lastFrame, 0.0f, MAXIMUM_POSSIBLE_FRAME); }
     float getLastFrame() const { return _lastFrame; }
     
+    /// by default the AnimationLoop will always reset to the first frame on any call to setRunning
+    /// this is not desirable in the case of entities with animations, if you don't want the reset
+    /// to happen then call this method with false;
+    void setResetOnRunning(bool resetOnRunning) { _resetOnRunning = resetOnRunning; }
     void setRunning(bool running);
+    bool getRunning() const { return _running; }
     bool isRunning() const { return _running; }
```
AndrewMeadows: **getRunning()** does the same thing as **isRunning()**.  Pick one and get rid of the other.

libraries/animation/src/AnimationLoop.cpp
ZappoMan: Good catch. Default should be true. To match old behavior. Which may or may not be depended on by avatar animations. I didn't want to change that behavior. 

Will fix when I'm at a keyboard.

examples/entityScripts/portal.js
```diff
@@ -41,8 +40,7 @@
     
     this.hoverEnterEntity = function(entityID) {
       Entities.editEntity(entityID, {
-        animationURL: animationURL,
-        animationSettings: '{ "fps": 24, "firstFrame": 1, "lastFrame": 25, "frameIndex": 1, "running": true, "hold": true }'
+        animation: { url: animationURL, fps: 24, firstFrame: 1, lastFrame: 25, frameIndex: 1, running: true, hold: true }
       });
     };
 })
```
AndrewMeadows: I know the **hold** property is not new, but I'm still trying to figure out what it for.  Is it like "pause"?

ZappoMan: Presumably if hold is true, when you stop the animation loop, it holds the last frame, if hold is false and you stop, it resets to first frame. 

Come to think of it. I'm not sure I tested that behavior. I just assumed it worked, since I didn't change anything about hold.

examples/html/entityProperties.html
AndrewMeadows: Not a blocker.  I was just curious and pointing it out in case you didn't want to keep it.  I usually delete debug lines but occasionally I leave them in.

libraries/entities/src/AnimationPropertyGroup.cpp
```diff
@@ -0,0 +1,434 @@
...
+                        newFirstFrame != firstFrame ||
+                        newLastFrame != lastFrame ||
+                        newLoop != loop ||
+                        newHold != hold ||
+                        newStartAutomatically != startAutomatically;
+
+
+
+    return bytesRead;
+}
```
AndrewMeadows: Not a issue with the code but just some thoughts that came to mind while read it.  I figured I would write them down since they occurred to me at this point in the code review:

(1) Writing so much boiler plate is a big disincentive for me to add new properties to Entities -- I've developed an aversion to the process and will privately grumble the next time I've got to go through it.

(2) How much of the code above is pure boiler plate and how much requires custom attention?  I see that we have some large logic pivots on whether **_animationLoop** is a valid pointer or not which obviously requires custom code.  I wonder if there is a way to abstract such a pattern out to make it even easier to add new properties.  Dunno what such a framework would look like exactly...

(3) The EntityServer is like a 3D database: property lists with spatial bounds.  There must be a way to implement it so that it is ignorant of most the actual types/data stored within the instances (not ignorant of the spatial info and whatever else is absolutely necessary to sort/serve the data).  Any abstraction at (2) would help simplify the EntityServer implementation but the interface would still have to know how to unpack all the details of the properties in order to "reveal" all of the features.  Hrm...

libraries/animation/src/AnimationLoop.cpp
ZappoMan: I've fixed this issue in latest commit.

libraries/entities/src/ModelEntityItem.cpp
ZappoMan: fixed in latest commit, _resetOnRunning is default to true... which was old behavior.

libraries/entities/src/AnimationPropertyGroup.cpp
ZappoMan: your opinions on this topic are noted and well understood.

Our current design has served us well to date and we've been able to have many developers efficiently add new entity types and new properties to existing entity types in this design.

I agree that reducing boiler place code would be nice.

The next logical step would be to implement a system that allows you to register property names, their type, and the entity types they should be supported for. I've roughed out some ideas for how to do that, but since there's nothing breaking the current design, and we've got so many other areas where the architecture is actually failing, this refactoring work is rightfully lower priority.

We did have a "generic 3d database" which you may recall no one could figure out... it was called metavoxels... it had all the grand vision of completely typelessness and utterly generic... and it never got any traction.

I admit, this current design has the problem of repeated boiler plate code... but it has the advantage that people seem to ACTUALLY UNDERSTAND what it does.

ZappoMan: in the particular case of the _animationLoop helper/associated object... I considered deriving AnimationGroupProperties from AnimationLoop since you almost always want both of them anyway... but that would have required more rewiring in several places and so I opted to take this more direct approach.

hifi-gustavo:
Build console output for commit: ef054ffc6d456575a6aa0556f780d0a05be4ebb1 available at: https://gist.github.com/1447c8cc7b63c5812d08<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6006.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6006.dmg

libraries/entities/src/ModelEntityItem.cpp
```diff
@@ -100,39 +94,40 @@ int ModelEntityItem::readEntitySubclassDataFromBuffer(const unsigned char* data,
...
+
+        bytesRead += bytesFromAnimation;
+        dataAt += bytesFromAnimation;
+    }
+
     READ_ENTITY_PROPERTY(PROP_SHAPE_TYPE, ShapeType, updateShapeType);
 
+    if (_animationProperties.somethingChanged()) {
+        _dirtyFlags |= EntityItem::DIRTY_UPDATEABLE;
+    }
```
AndrewMeadows: **PropertyGroup** picked up a data member called **_somethingChanged** which is queried here and is updated when unpacking data.  I think there are two problems...

(1) _somethingChanged should be cleared to **false** here where its value is "harvested".

(2) When unpacking properties _somethingChanges should only be allowed to shift from **false** to **true**, not from **true** to **false**.  That is, if it was already **true** before the properties were handled in then it should remain true (even if no properties were changed)... until it is harvested here.  I'll point the particular spots out with other comments.

libraries/entities/src/AnimationPropertyGroup.cpp
```diff
@@ -0,0 +1,434 @@
...
+    bool newStartAutomatically = _animationLoop ? _animationLoop->getStartAutomatically() : getStartAutomatically();
+
+     // NOTE: we don't check frameIndex because that is assumed to always be changing.
+    _somethingChanged = newFPS != fps ||
+                        newRunning != running ||
+                        newFirstFrame != firstFrame ||
+                        newLastFrame != lastFrame ||
+                        newLoop != loop ||
+                        newHold != hold ||
+                        newStartAutomatically != startAutomatically;
```
AndrewMeadows: **_somethingChanged** should only be allowed to change from **false** to **true** at this point.

libraries/entities/src/AnimationPropertyGroup.cpp
```diff
@@ -0,0 +1,434 @@
...
+        SET_ENTITY_GROUP_PROPERTY_FROM_PROPERTIES(Animation, FrameIndex, frameIndex, setFrameIndex);
+        SET_ENTITY_GROUP_PROPERTY_FROM_PROPERTIES(Animation, Running, running, setRunning);
+        SET_ENTITY_GROUP_PROPERTY_FROM_PROPERTIES(Animation, Loop, loop, setLoop);
+        SET_ENTITY_GROUP_PROPERTY_FROM_PROPERTIES(Animation, FirstFrame, firstFrame, setFirstFrame);
+        SET_ENTITY_GROUP_PROPERTY_FROM_PROPERTIES(Animation, LastFrame, lastFrame, setLastFrame);
+        SET_ENTITY_GROUP_PROPERTY_FROM_PROPERTIES(Animation, Hold, hold, setHold);
+        SET_ENTITY_GROUP_PROPERTY_FROM_PROPERTIES(Animation, StartAutomatically, startAutomatically, setStartAutomatically);
+    }
+
+    _somethingChanged = somethingChanged;
```
AndrewMeadows: **_somethingChanged** should only be allowed to change from **false** to **true** at this point.

libraries/entities/src/ModelEntityItem.cpp
ZappoMan: I'm pretty sure the code does that.

libraries/networking/src/udt/PacketHeaders.cpp
```diff
@@ -38,7 +38,7 @@ PacketVersion versionForPacketType(PacketType packetType) {
         case PacketType::EntityAdd:
         case PacketType::EntityEdit:
         case PacketType::EntityData:
-            return VERSION_ENTITIES_PARTICLE_ELLIPSOID_EMITTER;
+            return VERSION_ENTITIES_ANIMATION_PROPERTIES_GROUP;
         case PacketType::AvatarData:
```
AndrewMeadows: I verify that you are indeed changing the version number for a PR that adds new properties.

libraries/entities/src/ModelEntityItem.cpp
ZappoMan: The something changed is meaningful only after setProperties() or readEntitySubclassDataFromBuffer() in both cases the output will always be true iff something in those functions caused a meaningful change. The state of somethingChanged is only meaningful immediately after that case. Really this is a side effect of the readEntitySubclassDataFromBuffer() not having the bandwidth (API) to inform the caller that something changed... that's why I added this.

I could change that API... and the change would be more significant. (it would touch all entities)

ZappoMan: Actually, I will change it to do that... and remove this concept...

AndrewMeadows: I see... you have two methods that might set it.  One returns whether something changed and the other returns bytesRead, so you have to store the result in the **_animationProperties** instance.  This works because whether something changed or not only matters within the context of this **ModelEntityItem::setProperties()** call.

hifi-gustavo:
Build console output for commit: 57cc7adbfe8ac4f2c1abc2c24ef5b9ca573bd322 available at: https://gist.github.com/1447c8cc7b63c5812d08<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6006.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6006.dmg

examples/example/entities/butterflies.js
ZappoMan: frameIndex has been renamed to currentFrame throughout the code base

hifi-gustavo:
Build console output for commit: 7a85e5b7a7215203cd3d25111d7b1fb54daf9643 available at: https://gist.github.com/1447c8cc7b63c5812d08<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6006.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6006.dmg

libraries/entities/src/EntityItemPropertiesMacros.h
```diff
@@ -131,6 +132,18 @@ inline QScriptValue convertScriptValue(QScriptEngine* e, const EntityItemID& v)
...
+    if ((desiredProperties.isEmpty() || desiredProperties.getHasProperty(X)) && \
+        (!skipDefaults || defaultEntityProperties.get##G().get##P() != get##P())) { \
+        QScriptValue groupProperties = properties.property(#g); \
+        if (!groupProperties.isValid()) { \
+            groupProperties = engine->newObject(); \
+                } \
+        QScriptValue V = convertScriptValue(engine, M()); \
+        groupProperties.setProperty(#p, V); \
+        properties.setProperty(#g, groupProperties); \
+        }
```
AndrewMeadows: The indentation looks a little messed up here.  OTOH we all know it's really a on-line macro anyway...

ZappoMan: yeah - that's totally messed up... let me fix it

hifi-gustavo:
Build console output for commit: b09ae1755e50041ad7ea7614c02fcb213812a064 available at: https://gist.github.com/1447c8cc7b63c5812d08<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6006.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6006.dmg

PR #6008, opened 2015-10-06
IK targets specified in the skeleton's model-frame

AndrewMeadows:
There was a bug where the hands would wobble when walking despite the fact that the hydra sensors were at rest in the real-world -- animation at the hips would affect the (sorta) hip-relative end-effector target transforms.  This PR fixes the problem so that the hand targets are not influenced by animations.  You avatar can now walk with a cup of coffee without spilling it.

The test goes as follows:

(1) Login
(2) Pick up the hyrdras
(3) Walk forward
--> Using the old code the hands wiggle as the avatar walks.
--> Using the new code the hands do not wiggle.

### DIFF
--- a/libraries/animation/src/Rig.cpp
+++ b/libraries/animation/src/Rig.cpp
@@ -1036,7 +1036,7 @@ void Rig::updateNeckJoint(int index, const HeadParameters& params) {
                 glm::vec3 headPos, neckPos;
                 glm::quat headRot, neckRot;
 
-                AnimPose avatarHMDPose(glm::vec3(1), params.localHeadOrientation, params.localHeadPosition);
+                AnimPose avatarHMDPose(glm::vec3(1.0f), params.localHeadOrientation, params.localHeadPosition);
                 computeHeadNeckAnimVars(_animSkeleton, avatarHMDPose, headPos, headRot, neckPos, neckRot);
 
                 // debug rendering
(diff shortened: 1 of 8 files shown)


libraries/animation/src/Rig.cpp
```diff
@@ -1126,18 +1126,17 @@ void Rig::updateFromHandParameters(const HandParameters& params, float dt) {
...
+            _animVars.set("leftHandRotation", rootBindPose.rot * yFlipHACK * params.leftOrientation);
         } else {
             _animVars.unset("leftHandPosition");
             _animVars.unset("leftHandRotation");
         }
         if (params.isRightEnabled) {
-            _animVars.set("rightHandPosition", rootPose.trans + rootPose.rot * yFlipHACK * params.rightPosition);
-            _animVars.set("rightHandRotation", rootPose.rot * yFlipHACK * params.rightOrientation);
+            _animVars.set("rightHandPosition", rootBindPose.trans + rootBindPose.rot * yFlipHACK * params.rightPosition);
+            _animVars.set("rightHandRotation", rootBindPose.rot * yFlipHACK * params.rightOrientation);
```
AndrewMeadows: The trick is to use the root's bind pose when transforming into the model-frame (the bind-pose is constant!).

hifi-gustavo:
Build console output for commit: 224fc550339d910b4e29439fa62f1268dcb71bbb available at: https://gist.github.com/9c7420044cd688a8f1fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6008.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6008.dmg

ZappoMan:
I still see some wiggle, but it's definitely way less pronounced. Merging.

PR #6010, opened 2015-10-07
Disable hand mouse by default

ZappoMan:
this feature doesn't really work, it breaks a lot of things... but we're not quite ready to remove it or fix it... so we'll disable it by default.


hifi-gustavo:
Build console output for commit: 6e5847bb1c773c7f790a90b3bd65b9b6cbab04e5 available at: https://gist.github.com/3654a197ce278bc99201<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6010.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6010.dmg

PR #6013, opened 2015-10-07
Picking fixes in the HMD

jherico:
The 2D overlay picking was still sort of broken if the window wasn't very close to 1920x1080 (the UI size when in the HMD).  However, this should correct that as well as the 3D picking.

However, I need to check where else in the code base we're calling the picking code and ensure that it's all working as well.  I suspect that right now entities that hook mouse hover and mouse press events might still be broken.  I'm going to do some testing and if necessary correct them in a subsequent PR.

### DIFF
--- a/examples/example/hmd/pickerTest.js
+++ b/examples/example/hmd/pickerTest.js
@@ -1,0 +1,103 @@
+Script.include("../../libraries/utils.js");
+
+
+PickerTest = function() {
+    // Switch every 5 seconds between normal IPD and 0 IPD (in seconds)
+    this.UPDATE_INTERVAL = 10.0;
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: eff5a4c395958027b5d686337a9cf0cc4d7db8f4 available at: https://gist.github.com/674741d335ff1bc6280c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6013.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6013.dmg

PR #6016, opened 2015-10-07
Fix the cauterized Cluster Matrices issue when rear view mirror is enabled

samcake:
- Introduce a second gpu::Buffer in the Cluster state to deal with the cauterizedClusterMatrices case.
  Make sure to allocate both clusterBuffer and  or cauterizedClusteBuffer only if needed in the updateClusterMAtrices call.

- Clarify (improve maybe) the format definition for depth/stencil texture. 
 IN particular found the correct gl format for depth_stencil Textures and not RenderBUffer


hifi-gustavo:
Build console output for commit: f0b7622ef044ebe06fbdeee7158fa8085186138b available at: https://gist.github.com/146218d6418f5009094f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6016.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6016.dmg

hyperlogic:
:+1:

PR #6017, opened 2015-10-07
fix a bunch of warnings

ZappoMan:


interface/src/Application.cpp
```diff
@@ -4750,24 +4750,6 @@ mat4 Application::getEyeProjection(int eye) const {
     return _viewFrustum.getProjection();
 }
 
-mat4 Application::getEyePose(int eye) const {
```
ZappoMan: this function isn't used anywhere and really should be accessed from the display plugin. removing it.

hifi-gustavo:
Build console output for commit: bfd014203e796d807c32b1b1403d739456dd6b1c available at: https://gist.github.com/58480bc7dce810cc6e52<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6017.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6017.dmg

PR #6020, opened 2015-10-08
Added grab far events for non physical objects

ericrius1:
This is useful for being able to trigger items from afar that are static- for example, triggering the reset switch in toybox from across the room, without having to walk up and grab it in front of a bunch of demo-ees

### DIFF
--- a/examples/controllers/handControllerGrab.js
+++ b/examples/controllers/handControllerGrab.js
@@ -77,7 +77,9 @@ var STATE_NEAR_GRABBING = 4;
 var STATE_CONTINUE_NEAR_GRABBING = 5;
 var STATE_NEAR_GRABBING_NON_COLLIDING = 6;
 var STATE_CONTINUE_NEAR_GRABBING_NON_COLLIDING = 7;
-var STATE_RELEASE = 8;
+var STATE_FAR_GRABBING_NON_COLLIDING = 8;
+var STATE_CONTINUE_FAR_GRABBING_NON_COLLIDING = 9;
+var STATE_RELEASE = 10;
 
 var GRABBABLE_DATA_KEY = "grabbableKey"; // shared with grab.js
 var GRAB_USER_DATA_KEY = "grabKey"; // shared with grab.js
@@ -120,7 +122,7 @@ function MyController(hand, triggerAction) {
     var TIP_CONTROLLER_OFFSET = 1;
     this.triggerAction = triggerAction;
     this.palm = SPATIAL_CONTROLLERS_PER_PALM * hand;
-    this.tip = SPATIAL_CONTROLLERS_PER_PALM * hand + TIP_CONTROLLER_OFFSET; 
+    this.tip = SPATIAL_CONTROLLERS_PER_PALM * hand + TIP_CONTROLLER_OFFSET;
 
     this.actionID = null; // action this script created...
     this.grabbedEntity = null; // on this entity.
@@ -131,49 +133,55 @@ function MyController(hand, triggerAction) {
 
     var _this = this;
 
-    this.update = function() {
+    this.update = function () {
 
         this.updateSmoothedTrigger();
 
         switch (this.state) {
-            case STATE_OFF:
-                this.off();
-                this.touchTest();
-                break;
-            case STATE_SEARCHING:
-                this.search();
-                break;
-            case STATE_DISTANCE_HOLDING:
-                this.distanceHolding();
-                break;
-            case STATE_CONTINUE_DISTANCE_HOLDING:
-                this.continueDistanceHolding();
-                break;
-            case STATE_NEAR_GRABBING:
-                this.nearGrabbing();
-                break;
-            case STATE_CONTINUE_NEAR_GRABBING:
-                this.continueNearGrabbing();
-                break;
-            case STATE_NEAR_GRABBING_NON_COLLIDING:
-                this.nearGrabbingNonColliding();
-                break;
-            case STATE_CONTINUE_NEAR_GRABBING_NON_COLLIDING:
-                this.continueNearGrabbingNonColliding();
-                break;
-            case STATE_RELEASE:
-                this.release();
-                break;
+        case STATE_OFF:
+            this.off();
+            this.touchTest();
+            break;
+        case STATE_SEARCHING:
+            this.search();
+            break;
+        case STATE_DISTANCE_HOLDING:
+            this.distanceHolding();
+            break;
+        case STATE_CONTINUE_DISTANCE_HOLDING:
+            this.continueDistanceHolding();
+            break;
+        case STATE_NEAR_GRABBING:
+            this.nearGrabbing();
+            break;
+        case STATE_CONTINUE_NEAR_GRABBING:
+            this.continueNearGrabbing();
+            break;
+        case STATE_NEAR_GRABBING_NON_COLLIDING:
+            this.nearGrabbingNonColliding();
+            break;
+        case STATE_CONTINUE_NEAR_GRABBING_NON_COLLIDING:
+            this.continueNearGrabbingNonColliding();
+            break;
+        case STATE_FAR_GRABBING_NON_COLLIDING:
+            this.farGrabbingNonColliding();
+            break;
+        case STATE_CONTINUE_FAR_GRABBING_NON_COLLIDING:
+            this.continueFarGrabbingNonColliding();
+            break;
+        case STATE_RELEASE:
+            this.release();
+            break;
         }
     };
 
-    this.setState = function(newState) {
+    this.setState = function (newState) {
         // print("STATE: " + this.state + " --> " + newState);
         this.state = newState;
     }
 
 
-    this.lineOn = function(closePoint, farPoint, color) {
+    this.lineOn = function (closePoint, farPoint, color) {
         // draw a line
         if (this.pointer === null) {
             this.pointer = Entities.addEntity({
@@ -197,41 +205,41 @@ function MyController(hand, triggerAction) {
 
     };
 
-    this.lineOff = function() {
+    this.lineOff = function () {
         if (this.pointer !== null) {
             Entities.deleteEntity(this.pointer);
         }
         this.pointer = null;
     };
 
-    this.updateSmoothedTrigger = function() {
+    this.updateSmoothedTrigger = function () {
         var triggerValue = Controller.getActionValue(this.triggerAction);
         // smooth out trigger value
         this.triggerValue = (this.triggerValue * TRIGGER_SMOOTH_RATIO) +
             (triggerValue * (1.0 - TRIGGER_SMOOTH_RATIO));
     }
-    
-    this.triggerSmoothedSqueezed = function() {
+
+    this.triggerSmoothedSqueezed = function () {
         return this.triggerValue > TRIGGER_ON_VALUE;
     };
 
-    this.triggerSmoothedReleased = function() {
+    this.triggerSmoothedReleased = function () {
         return this.triggerValue < TRIGGER_OFF_VALUE;
     };
 
-    this.triggerSqueezed = function() {
+    this.triggerSqueezed = function () {
         var triggerValue = Controller.getActionValue(this.triggerAction);
         return triggerValue > TRIGGER_ON_VALUE;
     };
 
-    this.off = function() {
+    this.off = function () {
         if (this.triggerSmoothedSqueezed()) {
             this.setState(STATE_SEARCHING);
             return;
         }
     }
 
-    this.search = function() {
+    this.search = function () {
         if (this.triggerSmoothedReleased()) {
             this.setState(STATE_RELEASE);
             return;
@@ -251,9 +259,7 @@ function MyController(hand, triggerAction) {
         };
 
         var intersection = Entities.findRayIntersection(pickRay, true);
-        if (intersection.intersects &&
-            intersection.properties.collisionsWillMove === 1 &&
-            intersection.properties.locked === 0) {
+        if (intersection.intersects && intersection.properties.locked === 0) {
             // the ray is intersecting something we can move.
             var handControllerPosition = Controller.getSpatialControlPosition(this.palm);
             var intersectionDistance = Vec3.distance(handControllerPosition, intersection.intersection);
@@ -266,14 +272,22 @@ function MyController(hand, triggerAction) {
             }
             if (intersectionDistance < NEAR_PICK_MAX_DISTANCE) {
                 // the hand is very close to the intersected object.  go into close-grabbing mode.
-                this.setState(STATE_NEAR_GRABBING);
+                if (intersection.properties.collisionsWillMove === 1) {
+                    this.setState(STATE_NEAR_GRABBING);
+                } else {
+                    this.setState(STATE_NEAR_GRABBING_NON_COLLIDING);
+                }
             } else {
                 // don't allow two people to distance grab the same object
                 if (entityIsGrabbedByOther(intersection.entityID)) {
                     this.grabbedEntity = null;
                 } else {
                     // the hand is far from the intersected object.  go into distance-holding mode
-                    this.setState(STATE_DISTANCE_HOLDING);
+                    if (intersection.properties.collisionsWillMove === 1) {
+                        this.setState(STATE_DISTANCE_HOLDING);
+                    } else {
+                        this.setState(STATE_FAR_GRABBING_NON_COLLIDING);
+                    }
                 }
             }
         } else {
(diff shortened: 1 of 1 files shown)


examples/controllers/handControllerGrab.js
```diff
@@ -131,49 +133,55 @@ function MyController(hand, triggerAction) {
...
-            case STATE_NEAR_GRABBING_NON_COLLIDING:
-                this.nearGrabbingNonColliding();
-                break;
-            case STATE_CONTINUE_NEAR_GRABBING_NON_COLLIDING:
-                this.continueNearGrabbingNonColliding();
-                break;
-            case STATE_RELEASE:
-                this.release();
-                break;
+        case STATE_OFF:
```
ZappoMan: coding standard: the coding standard clearly states how switch statements should be formatted... you appear to be changing correctly formatted code to violate the coding standard. Please fix, and please fix your auto-formatter to not make this mistake.

examples/controllers/handControllerGrab.js
```diff
@@ -131,49 +133,55 @@ function MyController(hand, triggerAction) {
...
 
-    this.setState = function(newState) {
+    this.setState = function (newState) {
         // print("STATE: " + this.state + " --> " + newState);
         this.state = newState;
     }
 
 
-    this.lineOn = function(closePoint, farPoint, color) {
+    this.lineOn = function (closePoint, farPoint, color) {
```
ZappoMan: coding standard: no space here -- again, I think your code beautifier is reformatting correct coding standard compliant code. Please fix your auto-formatter.

examples/controllers/handControllerGrab.js
```diff
@@ -197,41 +205,41 @@ function MyController(hand, triggerAction) {
...
-    this.lineOff = function() {
+    this.lineOff = function () {
         if (this.pointer !== null) {
             Entities.deleteEntity(this.pointer);
         }
         this.pointer = null;
     };
 
-    this.updateSmoothedTrigger = function() {
+    this.updateSmoothedTrigger = function () {
```
ZappoMan: coding standard: no space here

examples/controllers/handControllerGrab.js
```diff
@@ -197,41 +205,41 @@ function MyController(hand, triggerAction) {
...
         var triggerValue = Controller.getActionValue(this.triggerAction);
         // smooth out trigger value
         this.triggerValue = (this.triggerValue * TRIGGER_SMOOTH_RATIO) +
             (triggerValue * (1.0 - TRIGGER_SMOOTH_RATIO));
     }
-    
-    this.triggerSmoothedSqueezed = function() {
+
+    this.triggerSmoothedSqueezed = function () {
```
ZappoMan: coding standard: no space here

examples/controllers/handControllerGrab.js
```diff
@@ -197,41 +205,41 @@ function MyController(hand, triggerAction) {
...
         return this.triggerValue > TRIGGER_ON_VALUE;
     };
 
-    this.triggerSmoothedReleased = function() {
+    this.triggerSmoothedReleased = function () {
```
ZappoMan: coding standard: no space here

examples/controllers/handControllerGrab.js
```diff
@@ -197,41 +205,41 @@ function MyController(hand, triggerAction) {
...
         return this.triggerValue < TRIGGER_OFF_VALUE;
     };
 
-    this.triggerSqueezed = function() {
+    this.triggerSqueezed = function () {
```
ZappoMan: coding standard: no space here

examples/controllers/handControllerGrab.js
```diff
@@ -197,41 +205,41 @@ function MyController(hand, triggerAction) {
...
         var triggerValue = Controller.getActionValue(this.triggerAction);
         return triggerValue > TRIGGER_ON_VALUE;
     };
 
-    this.off = function() {
+    this.off = function () {
```
ZappoMan: coding standard: no space here

examples/controllers/handControllerGrab.js
```diff
@@ -308,12 +322,13 @@ function MyController(hand, triggerAction) {
         }
     };
 
-    this.distanceHolding = function() {
+    this.distanceHolding = function () {
```
ZappoMan: coding standard: no space here

examples/controllers/handControllerGrab.js
```diff
@@ -308,12 +322,13 @@ function MyController(hand, triggerAction) {
...
 
         var handControllerPosition = Controller.getSpatialControlPosition(this.palm);
         var handRotation = Quat.multiply(MyAvatar.orientation, Controller.getSpatialControlRawRotation(this.palm));
         var grabbedProperties = Entities.getEntityProperties(this.grabbedEntity, ["position", "rotation",
-                                                                                  "gravity", "ignoreForCollisions"]);
```
ZappoMan: coding standard: "When the multiple lines are bound by parentheses (as in arguments to a function call), the prefered style has no whitespace after the opening parenthesis or before the closing parenthesis. The n+1 lines are generally indented to the column immediately after the opening parenthesis (following the style for split expressions in 2.1.6)."

examples/controllers/handControllerGrab.js
```diff
@@ -351,7 +366,7 @@ function MyController(hand, triggerAction) {
 
     };
 
-    this.continueDistanceHolding = function() {
+    this.continueDistanceHolding = function () {
```
ZappoMan: coding standard: no space here

examples/controllers/handControllerGrab.js
```diff
@@ -431,7 +446,7 @@ function MyController(hand, triggerAction) {
         });
     };
 
-    this.nearGrabbing = function() {
+    this.nearGrabbing = function () {
```
ZappoMan: coding standard: no space here

examples/controllers/handControllerGrab.js
```diff
@@ -521,16 +537,48 @@ function MyController(hand, triggerAction) {
         this.setState(STATE_CONTINUE_NEAR_GRABBING_NON_COLLIDING);
     };
 
-    this.continueNearGrabbingNonColliding = function() {
+    this.farGrabbingNonColliding = function () {
```
ZappoMan: coding standard: no space here

examples/controllers/handControllerGrab.js
```diff
@@ -521,16 +537,48 @@ function MyController(hand, triggerAction) {
...
+        if (this.hand === RIGHT_HAND) {
+            Entities.callEntityMethod(this.grabbedEntity, "setRightHand");
+        } else {
+            Entities.callEntityMethod(this.grabbedEntity, "setLeftHand");
+        }
+        Entities.callEntityMethod(this.grabbedEntity, "startFarGrabNonColliding");
+        this.setState(STATE_CONTINUE_FAR_GRABBING_NON_COLLIDING);
+    };
+
+    this.continueNearGrabbingNonColliding = function () {
```
ZappoMan: coding standard: no space here

examples/controllers/handControllerGrab.js
```diff
@@ -521,16 +537,48 @@ function MyController(hand, triggerAction) {
...
+            direction: Quat.getUp(this.getHandRotation())
+        };
+
+        this.lineOn(pickRay.origin, Vec3.multiply(pickRay.direction, LINE_LENGTH), NO_INTERSECT_COLOR);
+        Entities.callEntityMethod(this.grabbedEntity, "continueFarGrabbingNonColliding");
+    };
+
     _this.allTouchedIDs = {};
-    this.touchTest = function() {
+    this.touchTest = function () {
```
ZappoMan: coding standard: no space here

examples/controllers/handControllerGrab.js
```diff
@@ -521,16 +537,48 @@ function MyController(hand, triggerAction) {
...
+    this.continueNearGrabbingNonColliding = function () {
+        if (this.triggerSmoothedReleased()) {
+            this.setState(STATE_RELEASE);
+            return;
+        }
+
         Entities.callEntityMethod(this.grabbedEntity, "continueNearGrabbingNonColliding");
     };
 
+    this.continueFarGrabbingNonColliding = function () {
```
ZappoMan: coding standard: no space between function and ()

examples/controllers/handControllerGrab.js
```diff
@@ -539,19 +587,19 @@ function MyController(hand, triggerAction) {
         var ids = [];
 
         if (leftEntities.length !== 0) {
-            leftEntities.forEach(function(entity) {
+            leftEntities.forEach(function (entity) {
```
ZappoMan: coding standard: no space between function and ()

examples/controllers/handControllerGrab.js
```diff
@@ -539,19 +587,19 @@ function MyController(hand, triggerAction) {
...
                 ids.push(entity);
             });
 
         }
 
         if (rightEntities.length !== 0) {
-            rightEntities.forEach(function(entity) {
+            rightEntities.forEach(function (entity) {
```
ZappoMan: coding standard: no space between function and ()

examples/controllers/handControllerGrab.js
```diff
@@ -584,19 +632,19 @@ function MyController(hand, triggerAction) {
 
     };
 
-    this.startTouch = function(entityID) {
+    this.startTouch = function (entityID) {
```
ZappoMan: coding standard: no space between function and ()

examples/controllers/handControllerGrab.js
```diff
@@ -584,19 +632,19 @@ function MyController(hand, triggerAction) {
...
         Entities.callEntityMethod(entityID, "startTouch");
     };
 
-    this.continueTouch = function(entityID) {
+    this.continueTouch = function (entityID) {
```
ZappoMan: coding standard: no space between function and ()

examples/controllers/handControllerGrab.js
```diff
@@ -584,19 +632,19 @@ function MyController(hand, triggerAction) {
...
         Entities.callEntityMethod(entityID, "continueTouch");
     };
 
-    this.stopTouch = function(entityID) {
+    this.stopTouch = function (entityID) {
```
ZappoMan: coding standard: no space between function and ()

examples/controllers/handControllerGrab.js
```diff
@@ -584,19 +632,19 @@ function MyController(hand, triggerAction) {
...
         Entities.callEntityMethod(entityID, "stopTouch");
     };
 
-    this.computeReleaseVelocity = function(deltaPosition, deltaTime, useMultiplier) {
+    this.computeReleaseVelocity = function (deltaPosition, deltaTime, useMultiplier) {
```
ZappoMan: coding standard: no space between function and ()

examples/controllers/handControllerGrab.js
```diff
@@ -637,11 +685,11 @@ function MyController(hand, triggerAction) {
         this.setState(STATE_OFF);
     };
 
-    this.cleanup = function() {
+    this.cleanup = function () {
```
ZappoMan: coding standard: no space between function and ()

examples/controllers/handControllerGrab.js
```diff
@@ -637,11 +685,11 @@ function MyController(hand, triggerAction) {
...
         this.release();
     };
 
-    this.activateEntity = function(entityID, grabbedProperties) {
+    this.activateEntity = function (entityID, grabbedProperties) {
```
ZappoMan: coding standard: no space between function and ()

examples/controllers/handControllerGrab.js
```diff
@@ -650,12 +698,19 @@ function MyController(hand, triggerAction) {
...
+                    z: 0
+                },
+                ignoreForCollisions: true
+            });
         }
         setEntityCustomData(GRAB_USER_DATA_KEY, entityID, data);
     };
 
-    this.deactivateEntity = function(entityID) {
+    this.deactivateEntity = function (entityID) {
```
ZappoMan: coding standard: no space between function and ()

examples/controllers/handControllerGrab.js
huffman: If we are trying to follow jslint, we may want to revise this rule for anonymous JS functions - AFAIK jslint considers it correct to have a space after `function` here.

`function updateSmoothedTrigger()`
`updateSmoothedTrigger = function ()`

ZappoMan: Are you saying jslint considers no space to be wrong? I doubt that. If jslint says our coding standard is correct, then we should not change the coding standard.... 

If EVERYONE wants to change the coding standard to say we want spaces before () of function calls then that rule should apply to C++ and JS... I've seen C++ developers make this same "mistake" (according to the standard). If that's what we want to do in all our code, great, lets change the coding standard.

ZappoMan: jslint.com says the following code is fine...

```
"use strict";
function spam() {
    console.log("spam");
}
```

huffman: > Are you saying jslint considers no space to be wrong?
Yes, you can find an explanation here: https://jslinterrors.com/expected-exactly-one-space-between-a-and-b

> If jslint says our coding standard is correct, then we should not change the coding standard....
I agree, but in this case it's saying it is incorrect.

> spaces before () of function calls
This isn't a function call, it's very specifically the case where you are creating an anonymous function. The explanation above explains it as "Between the function keyword and its associated opening parenthesis, when the function is parsed as an expression (but not when it's parsed as a function declaration)"

I don't necessarily agree with the general notion that our C++ and JS standards should be exactly the same. We should try to keep them consistent, but let them diverge where it makes sense.

FWIW, I don't actually like having the space there, but jslint considers it correct.

huffman: re: your jslint.com example

In my provided examples I showed the difference, and specifically said **anonymous** functions - I agree your example is correct. Try this on jslint:

```javascript
"use strict";

var badFunc = function() {
    return 0;
};

var goodFunc = function () {
    return 0;
};
```

ZappoMan: so digging deeper it does look like JSLint has sort of conflicting style rules... apparently it *does match our style guide* and explicitly *DOESN'T want a space between* the function name and the parenthesis when it's this form:
```
function foo() {
   ...
}
```

But when it's the assignment to a variable form, *it doesn't match our style guide* and *insists on a space*...

```
var spam = function () {
    ...
}
```

ZappoMan:
in other news: Douglas Crawford slams the door on any debate about TAB vs SPACE -- "You can't explain it well because there is not a clear advantage to using tabs over spaces or vice versa, except for one: There is a universally accepted standard for how many spaces a space occupies. There isn't such a standard for tabs. We should pick one because having to choose between two is a waste of time, and there can be interoperability hazards. So we go with spaces. Get used to it."

examples/controllers/handControllerGrab.js
ZappoMan: It's really simple to me guys -- I'm going to document all cases I see that don't match our coding standard. This is not about what I prefer or don't prefer... it's about following our coding standard. 

If you want to change the coding standard, there is a well understood process for changing the coding standard: start a conversation in the "developers" email list, reach consensus on the intent of the rule, reach consensus on the WORDING of the rule, and change the document.

examples/controllers/handControllerGrab.js
```diff
@@ -440,8 +455,7 @@ function MyController(hand, triggerAction) {
 
         this.lineOff();
 
-        var grabbedProperties = Entities.getEntityProperties(this.grabbedEntity,
-                                                             ["position", "rotation", "gravity", "ignoreForCollisions"]);
+        var grabbedProperties = Entities.getEntityProperties(this.grabbedEntity, ["position", "rotation", "gravity", "ignoreForCollisions"]);
```
sethalves: coding standard -- max line length 128

examples/controllers/handControllerGrab.js
sethalves: this sort of makes sense -- the parens should be tight against the function name, which is (in the anonymous case) empty.

ZappoMan: Let's also be clear about this... **"I don't necessarily agree with the general notion that our C++ and JS standards should be exactly the same. We should try to keep them consistent, but let them diverge where it makes sense."** -- Right now, there are no exceptions in the coding standard for JS vs C++... I've heard opinions that "in some cases it may make sense to diverge" -- yet somehow... no one has ever ACTUALLY recommended an example where divergence makes sense... Because if they had... we would surely have adjusted the coding standard to reflect that.

Even in our recent discussion of "multi-line formatting" the new wording DOES NOT MAKE a language exception.

If we believe that anonymous function assignments (something that exists in both C++ and JS) should have a different formatting than other function declarations... then GREAT! Let's agree to that... if it makes sense in JS... that new formatting also probably makes sense in C++ also...

If the answer is "no-- I want C++ to stay the same, but I don't care about how JS is formatted"... that's not a very credible argument for what you want a particular formatting... you should be able to make an affirmative case for why the formatting is better.

examples/controllers/handControllerGrab.js
ZappoMan: if it makes @ericrius1 feel any better, I just found this exact same coding standard violation in my own PR... and I commented on it as a coding standard violation, and fixed it. ;)

PR #6021, opened 2015-10-08
Support reload-on-change for local scripts

jherico:
Similar to work done in the Procedural.cpp file, detect if a script is a local file URL, if so, keep an eye on the modified time and reload the script if it changes.

### DIFF
--- a/libraries/script-engine/src/ScriptEngine.cpp
+++ b/libraries/script-engine/src/ScriptEngine.cpp
@@ -11,12 +11,13 @@
 
 #include <QtCore/QCoreApplication>
 #include <QtCore/QEventLoop>
+#include <QtCore/QFileInfo>
 #include <QtCore/QTimer>
 #include <QtCore/QThread>
 #include <QtNetwork/QNetworkRequest>
 #include <QtNetwork/QNetworkReply>
-#include <QScriptEngine>
-#include <QScriptValue>
+#include <QtScript/QScriptEngine>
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 99d1e17ae8a08fd41c340cbd23ede360da0e117f available at: https://gist.github.com/ba6df46292f275f85dc4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6021.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6021.dmg

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1022,6 +1033,41 @@ void ScriptEngine::unloadAllEntityScripts() {
     _entityScripts.clear();
 }
 
+void ScriptEngine::refreshFileScript(const EntityItemID& entityID) {
```
ZappoMan: are you calling unload()/preload() on these? it doesn't look like you are... that will break the paradigm

ZappoMan: I see unload() gets called from unloadEntityScript() - so that's good.

ZappoMan:
this doesn't seem to work on Windows 8.1... I've set an entity script to reference a local file, and I'm editing and saving the file, and it's not running the unload/preload methods

PR #6025, opened 2015-10-08
fix obj model meshes

sethalves:
- set up obj model meshes so renderer doesn't crash when an obj model is in the domain.


hifi-gustavo:
Build console output for commit: 3db148f3dcbd5882877dc924a793f03414783f70 available at: https://gist.github.com/e958e8b7b65557026992<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6025.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6025.dmg

PR #6028, opened 2015-10-08
[Scripts] [Design Review] Javascript Build Tool Chain + Task That Adds Linter Globals to all .js files

imgntn:
**While the tool chain stuff should probably be included in any case, I'm looking for some feedback on this way of doing linter globals.  There definitely other ways to do this (i.e. a per editor config file) and advantages and disadvantages to each method (i.e. i've only ever seen a sublime config).  This is a quick implementation of one possibility, to get discussion started.  Please see my second comment below for alternatives**

**General Info**
This PR adds a Javascript build tool chain to Hifi in the javascript_build folder.   The tool chain uses node.js and gulp.js to manipulate script files.  Nothing is installed by default, to keep the Interface download slim.  A user will need to manually install the packages and run the task.

As a script writer, I would write my script and then run the gulp task before pushing to github.

Prerequisites: Installed and working node.js  and gulp.js 

from 'javascript_build' folder:
-npm install
-gulp [taskname]

current valid tasknames: 
-[updateLinterGlobals]

**Update Linter Globals Info**
Why do we need this?  Code quality.  Linting all of our Javascript code will prevent errors and ensure readability. Right now, we're all doing setup on a per-user basis.  The linter settings often differ from user to user, as well as from what our C++ coding standards say.  We should stop repeating ourselves, and get on the same page for our benefit and that of our users.  We can use a build chain to maintain templates, settings, etc.


imgntn:
Other ways to do this could include:

- A linter settings file for each of the editors (sublime, emacs, visual studio, etc?) that includes the globals.  There could be a gulp task to read from a list of globals and update the settings file to be run whenever a new global is added.  Users would need to point their editor at the settings file inside of the repo.

- A gulp task that runs the linter on files.  A gulp-watch task + a gulp-jslint task would mean that files got linted as you saved -- much the same as running JSLint on save inside of an editor, with the benefit of being cross editor and the drawback of requiring a running gulp task during development.  Which isn't really that much of a drawback, since most web devs will be doing something similar already.

hifi-gustavo:
Build console output for commit: 6ac1171eb2f07a96a114fce1938c1102b995195a available at: https://gist.github.com/2e93ec47227145822b2a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6028.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6028.dmg

PR #6029, opened 2015-10-08
Low level support for URL overrides when loading content

jherico:
This change is to support a content creation workflow.  Imagine that you're a developer with a bunch Hifi content in a local folder which is also mirrored in S3.  

In order to tweak your content, you have to go into Interface, modify the entity to change the URL to point to the local file, make changes, test them, push the content and then edit the entity to change the URL back to the server version.  While the entities point to local files, they can't be viewed by others.  Alternatively, you have to make changes, push to S3 and then zone out and back to see them.  Both workflows are annoying if you want to iterate on content like scripts or shaders quickly especially since both scripts and shaders support automatic reloading if the content is a local file and the modification date changes.

This change starts to enable a new workflow.  Instead of editing the entities at all, you can create a mapping... for instance:  `    ResourceManager::setUrlPrefixOverride("https://s3.amazonaws.com/DreamingContent/", "file:///C:/Users/bdavis/Git/dreaming/");`

When you use ResourceManager::normalizeURL, this mapping will be applied to the URL.  This means that a client running with this mapping can work directly against their local content without disrupting the view of other clients connecting to the server.  When you're happy with the changes, you can push your changes to S3 and everyone else can see them.  This drastically reduces the effort required to make modifications to shader and script content.

This functionality is incomplete right now because there's no mechanism to create or change a mapping right now, short of adding a line of code in the app startup as above.  A future PR will support CRUD operations on mappings through Javascript.

### DIFF
--- a/libraries/networking/src/ResourceManager.cpp
+++ b/libraries/networking/src/ResourceManager.cpp
@@ -17,31 +17,24 @@
 
 #include <SharedUtil.h>
 
+ResourceManager::PrefixMap ResourceManager::_prefixMap;
+QMutex ResourceManager::_prefixMapLock;
 
-using PrefixMap = std::map<QString, QString>;
-static PrefixMap PREFIX_MAP;
-static QMutex PREFIX_MAP_LOCK;
 
 void ResourceManager::setUrlPrefixOverride(const QString& prefix, const QString& replacement) {
-    QMutexLocker locker(&PREFIX_MAP_LOCK);
-    PREFIX_MAP[prefix] = replacement;
+    QMutexLocker locker(&_prefixMapLock);
+    _prefixMap[prefix] = replacement;
 }
 
 QString ResourceManager::normalizeURL(const QString& urlString) {
     QString result = urlString;
-    QMutexLocker locker(&PREFIX_MAP_LOCK);
-    bool modified{ false };
-    foreach(const auto& entry, PREFIX_MAP) {
+    QMutexLocker locker(&_prefixMapLock);
+    foreach(const auto& entry, _prefixMap) {
         const auto& prefix = entry.first;
         const auto& replacement = entry.second;
         if (result.startsWith(prefix)) {
-            qDebug() << prefix;
             result.replace(0, prefix.size(), replacement);
-            modified = true;
         }
-    }
-    if (modified) {
-        qDebug() << result;
     }
     return result;
 }
--- a/libraries/networking/src/ResourceManager.cpp
+++ b/libraries/networking/src/ResourceManager.cpp
@@ -17,7 +17,37 @@
 
 #include <SharedUtil.h>
 
-QUrl ResourceManager::normalizeURL(const QUrl& url) {
+
+using PrefixMap = std::map<QString, QString>;
+static PrefixMap PREFIX_MAP;
+static QMutex PREFIX_MAP_LOCK;
+
+void ResourceManager::setUrlPrefixOverride(const QString& prefix, const QString& replacement) {
+    QMutexLocker locker(&PREFIX_MAP_LOCK);
+    PREFIX_MAP[prefix] = replacement;
+}
+
+QString ResourceManager::normalizeURL(const QString& urlString) {
+    QString result = urlString;
+    QMutexLocker locker(&PREFIX_MAP_LOCK);
+    bool modified{ false };
+    foreach(const auto& entry, PREFIX_MAP) {
+        const auto& prefix = entry.first;
+        const auto& replacement = entry.second;
+        if (result.startsWith(prefix)) {
+            qDebug() << prefix;
+            result.replace(0, prefix.size(), replacement);
+            modified = true;
+        }
+    }
+    if (modified) {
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: 17d0a70a52749a24092f1a94b777a4a7d61b894b available at: https://gist.github.com/b3c2af7f9951ae7abb29<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6029.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6029.dmg

libraries/networking/src/ResourceManager.cpp
```diff
@@ -17,7 +17,37 @@
 
 #include <SharedUtil.h>
 
-QUrl ResourceManager::normalizeURL(const QUrl& url) {
+
+using PrefixMap = std::map<QString, QString>;
+static PrefixMap PREFIX_MAP;
```
ZappoMan: these should be private or protected members of the class...

ZappoMan: this is also a strange use of ALL_CAPS which are generally reserved for constants

libraries/networking/src/ResourceManager.cpp
```diff
@@ -17,7 +17,37 @@
...
+
+QString ResourceManager::normalizeURL(const QString& urlString) {
+    QString result = urlString;
+    QMutexLocker locker(&PREFIX_MAP_LOCK);
+    bool modified{ false };
+    foreach(const auto& entry, PREFIX_MAP) {
+        const auto& prefix = entry.first;
+        const auto& replacement = entry.second;
+        if (result.startsWith(prefix)) {
+            qDebug() << prefix;
```
ZappoMan: did you mean to leave this debug code in here?

libraries/networking/src/ResourceManager.cpp
```diff
@@ -17,7 +17,37 @@
...
+            result.replace(0, prefix.size(), replacement);
+            modified = true;
+        }
+    }
+    if (modified) {
+        qDebug() << result;
```
ZappoMan: dead code??

libraries/networking/src/ResourceManager.cpp
```diff
@@ -17,7 +17,37 @@
...
+
+void ResourceManager::setUrlPrefixOverride(const QString& prefix, const QString& replacement) {
+    QMutexLocker locker(&PREFIX_MAP_LOCK);
+    PREFIX_MAP[prefix] = replacement;
+}
+
+QString ResourceManager::normalizeURL(const QString& urlString) {
+    QString result = urlString;
+    QMutexLocker locker(&PREFIX_MAP_LOCK);
+    bool modified{ false };
```
ZappoMan: coding standard/style: although the question of how to use universal initializers is potentially worth discussing, this usage for a PODT is certainly not common in our code and looks out of place... especially considering you had a QString above that you didn't use {} for.

libraries/networking/src/ResourceManager.cpp
jherico: fixed

libraries/networking/src/ResourceManager.cpp
jherico: fixed

libraries/networking/src/ResourceManager.cpp
jherico: fixed

libraries/networking/src/ResourceManager.cpp
jherico: removed

hifi-gustavo:
Build console output for commit: e5fc2e552550dc03fa933d8a291c5fd06490326d available at: https://gist.github.com/b3c2af7f9951ae7abb29<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6029.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6029.dmg

hifi-gustavo:
Build console output for commit: 85c565e41784e2297f1f8e1c7aaca1c2837328b8 available at: https://gist.github.com/b3c2af7f9951ae7abb29<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6029.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6029.dmg

hifi-gustavo:
Build console output for commit: cd1b7585348a6468ae4f138502b9f6fbad378fdb available at: https://gist.github.com/b3c2af7f9951ae7abb29<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6029.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6029.dmg

PR #6033, opened 2015-10-09
FIxing the backgroung leaking in the rear view mirror and moving from Float32 to 24 bits Depth buffer

samcake:
- Grouped the depth buffer and stencil buffer into a single buffer format 24_8 so we do loose accuracy compared to the Float 32 bits we were using but that didn't show up so far. In a the future we should improve the depth buffer accuracy by using it flipped ( from 1 to 0) to get a better distribution the possible values.

- Using a single DepthStencil buffer instead of 2 separated ones fixed the problem with the rear view mirror leaking the background.

- Expose the DetphRange parameters near and far mappings into the depthBuffer.


hifi-gustavo:
Build console output for commit: 3897716ff5730c0f3ffb21c963599857be7366ff available at: https://gist.github.com/35d6b5c366384bcbcd3a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6033.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6033.dmg

ZappoMan:
This PR doesn't seem to fix the problem.
broken mirror

samcake:
wtf

samcake:
i think it works

ZappoMan:
hmmmm.... testing again this morning, and it DOES work??!!!! Maybe I had a bad install?

PR #6035, opened 2015-10-09
Added targets to entity switch reset

ericrius1:
This PR adds the automatically resetting targets to the Toybox reset scripts.

### DIFF
--- a/unpublishedScripts/hiddenEntityReset.js
+++ b/unpublishedScripts/hiddenEntityReset.js
@@ -273,7 +273,7 @@
             resetMe: true
           },
           grabbableKey: {
-            grabbable: true
+            grabbable: false
           }
 
         })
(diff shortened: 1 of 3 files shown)


unpublishedScripts/masterReset.js
```diff
@@ -450,8 +458,12 @@ MasterReset = function() {
                         rotation: rotation,
                         script: targetsScriptURL
                     };
+                    var target = Entities.addEntity(targetProperties);
+                    targets[index] = target;
+                    setEntityCustomData(resetKey, target, {
```
ZappoMan: this construct doesn't make that much sense and is definitely wasteful from a performance and network perspective. You just created the entity "target" with the call ` var target = Entities.addEntity(targetProperties);` and then 2 lines later you set a property on that exact entity... you should really set this property in the original construction of the entity, and not spend the extra time setting a property on something you just created.

The better approach would be....

```
var targetProperties = {
    name: 'Target',
    type: 'Model',
    modelURL: MODEL_URL,
    shapeType: 'compound',
    collisionsWillMove: true,
    dimensions: TARGET_DIMENSIONS,
    compoundShapeURL: COLLISION_HULL_URL,
    position: originalPositions[index],
    rotation: rotation,
    script: targetsScriptURL,
    userData: JSON.stringify({
        resetMe: {
            resetMe: true
        }
    });
};
var target = Entities.addEntity(targetProperties);
```

We should really repair this in all these new object creation cases.

unpublishedScripts/masterReset.js
```diff
@@ -421,8 +426,11 @@ MasterReset = function() {
                     rotation: rotation,
                     script: targetsScriptURL
                 };
-
-                targets.push(Entities.addEntity(targetProperties));
+                var target = Entities.addEntity(targetProperties);
+                targets.push(target);
+                setEntityCustomData(resetKey, target, {
```
ZappoMan: see my note below about calling `setEntityCustomData()` immediately after `Entities.addEntity()`

unpublishedScripts/masterReset.js
```diff
@@ -388,6 +388,11 @@ MasterReset = function() {
             ignoreForCollisions: true,
         });
 
+
+        setEntityCustomData(resetKey, targetIntervalClearer, {
```
ZappoMan: see my note below about calling `setEntityCustomData()` immediately after `Entities.addEntity()`

unpublishedScripts/hiddenEntityReset.js
```diff
@@ -368,6 +371,153 @@
...
+                        collisionsWillMove: true,
+                        dimensions: TARGET_DIMENSIONS,
+                        compoundShapeURL: COLLISION_HULL_URL,
+                        position: position,
+                        rotation: rotation,
+                        script: targetsScriptURL
+                    };
+                    var target = Entities.addEntity(targetProperties);
+                    targets.push(target);
+                    setEntityCustomData(resetKey, target, {
```
ZappoMan: see my note below about calling `setEntityCustomData()` immediately after `Entities.addEntity()`

unpublishedScripts/hiddenEntityReset.js
```diff
@@ -368,6 +371,153 @@
...
+                            collisionsWillMove: true,
+                            dimensions: TARGET_DIMENSIONS,
+                            compoundShapeURL: COLLISION_HULL_URL,
+                            position: originalPositions[index],
+                            rotation: rotation,
+                            script: targetsScriptURL
+                        };
+                        var target = Entities.addEntity(targetProperties);
+                        targets[index] = target;
+                        setEntityCustomData(resetKey, target, {
```
ZappoMan: see my note below about calling `setEntityCustomData()` immediately after `Entities.addEntity()`

PR #6037, opened 2015-10-09
fix bug in AvatarData::toByteArray related code

sethalves:
- fixes a bug which caused unneeded sending of a few avatar joint translations.

### DIFF
--- a/libraries/avatars/src/AvatarData.cpp
+++ b/libraries/avatars/src/AvatarData.cpp
@@ -287,6 +287,7 @@ QByteArray AvatarData::toByteArray(bool cullSmallChanges, bool sendAll) {
 
     #ifdef WANT_DEBUG
     int rotationSentCount = 0;
+    unsigned char* beforeRotations = destinationBuffer;
     #endif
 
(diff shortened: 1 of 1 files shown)


libraries/avatars/src/AvatarData.cpp
```diff
@@ -409,7 +416,8 @@ void AvatarData::doneEncoding(bool cullSmallChanges) {
                     _lastSentJointData[i].rotation = data.rotation;
                 }
             }
-
+        }
+        if (_lastSentJointData[i].translation != data.translation) {
```
sethalves: this is the fix.  the rest is just debugging adjustments.

hifi-gustavo:
Build console output for commit: 14ad1ed6be6749266a4620c9286e74da104cd73e available at: https://gist.github.com/72471b95e2ba5dafdc49<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6037.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6037.dmg

PR #6038, opened 2015-10-09
Asset file drop

Atlante45:
- Adds drag-and-drop capabilities for FBX and OBJ file.
- Automatically sends dropped models to the asset server.
- Automatically spawns models in front of avatar with their correct size.


Atlante45:
@ZappoMan Could you checkout [this commit](https://github.com/Atlante45/hifi/commit/b708c9208f78cfcf79f46c98baa65c1d0e59d8b2).
My reasoning is that since we are creating the entity, we don't need to mark all the properties as changed since the newly created entity object should have those values already set to their default.

hifi-gustavo:
Build console output for commit: 83160dfa81f4ff04778c3ae2f8b6c623c9c59cda available at: https://gist.github.com/cefe6dbfa441636d24bb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6038.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6038.dmg

libraries/entities/src/EntityTypes.cpp
```diff
@@ -88,9 +88,7 @@ EntityItemPointer EntityTypes::constructEntityItem(EntityType entityType, const
         factory = _factories[entityType];
     }
     if (factory) {
-        EntityItemProperties mutableProperties = properties;
-        mutableProperties.markAllChanged();
-        newEntityItem = factory(entityID, mutableProperties);
+        newEntityItem = factory(entityID, properties);
```
ZappoMan: this change feels a little risky... it assumes that all entity factories initialize their entity properties to the same defaults as the "properties" do... Although I'd like to think that's true, I'm not sure that is unless you've confirmed it.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -290,19 +297,28 @@ void RenderableModelEntityItem::render(RenderArgs* args) {
...
                 bool movingOrAnimating = isMoving() || isAnimatingSomething();
                 if ((movingOrAnimating || _needsInitialSimulation) && _model->isActive()) {
-                    _model->setScaleToFit(true, dimensions);
+                    if (!_dimensionsInitialized) {
+                        EntityItemProperties properties;
+                        auto extents = _model->getMeshExtents();
+                        properties.setDimensions(extents.maximum - extents.minimum);
+                        
+                        qCDebug(entitiesrenderer) << "Autoresizing:" << (!getName().isEmpty() ? getName() : getModelURL());
+                        DependencyManager::get<EntityScriptingInterface>()->editEntity(getEntityItemID(), properties);
```
ZappoMan: something about this feels a little sleazy... I see what you're doing here, and I do like the idea that there's some auto-resizing going on, but something doesn't sit right with me about how we're doing this. I can't quite put my finger on it.

libraries/entities/src/EntityTypes.cpp
ZappoMan: Ok, so I think I know why this approach is gnawing at me... I don't think it works in the "load the scene from a previous creation" case... Have you tested what happens when a second viewer comes along and looks at one of these auto-resized models?

The code path I'd expect to see in that case is....

The entity server is streaming the scene to the client... and the viewer will get a stream of octree data that eventually lands on `EntityTreeElement::readElementDataFromBuffer()`...

```
int EntityTreeElement::readElementDataFromBuffer(const unsigned char* data, int bytesLeftToRead,
            ReadBitstreamToTreeParams& args) {
...
                // we hit this section of code, read out the EntityItemID from the stream...
                // it's for a "new" to us, ID... because we just started viewing the scene, so no entities
                // are in our local tree... so they are all new...
                if (args.bitstreamVersion >= VERSION_ENTITIES_SUPPORT_SPLIT_MTU) {
                    entityItemID = EntityItemID::readEntityItemIDFromBuffer(dataAt, bytesLeftToRead);
                    entityItem = _myTree->findEntityByEntityItemID(entityItemID);
                }
                ...
                if (entityItem) {
                    ...
                } else {
                    // this is where we will land....
                    entityItem = EntityTypes::constructEntityItem(dataAt, bytesLeftToRead, args);
                    if (entityItem) {
                        bytesForThisEntity = entityItem->readEntityDataFromBuffer(dataAt, bytesLeftToRead, args);
                        ...
                        // now you might ask yourself... why are we calling construct, and then turning 
                        // around and reading the data into our item... well, the construct call, creates essentially
                        // an empty entity... with NO properties set... so the properties used to construct it
                        // are basically set as all "not changed"...
                    }
                }
```

Since I don't see you also checking/setting `_dimensionsInitialized` in the readEntityDataFromBuffer() you're not going to properly detect this case... and as a result EVERY VIEWER is going to turn around and do an edit to the model... This is not what we want at all...

It might visually work... but it's going to FLOOD the entity server with a TON of edits for every viewer.

ZappoMan: hmmm - ok wait... testing it. it doesn't seem to do that... so...maybe this is ok

ZappoMan: I guess that the `EntityItem::readEntityDataFromBuffer()` does end up calling `EntityItem::updateDimensions()` which calls `EntityItem::setDimensions()` -- so maybe this does work

ZappoMan: let's double check that all Entities properly construct themselves with the same default values that they would get from the default EntityItemProperties values. since that's the most specific change you've made here. If we can confirm that, then I think this is ok.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
ZappoMan: Maybe the reason I feel like this is sleazy is because it's happening in the `render()` method... if this was happening in `update()` it would seem less sleazy.

libraries/entities/src/EntityTypes.cpp
Atlante45: Ok, I'm gonna double check all of them.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
Atlante45: Yes, I was a bit uneasy adding it here too.
Update is a much better place for it, I'm gonna move the entire model initialization there.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -35,6 +41,11 @@ RenderableModelEntityItem::~RenderableModelEntityItem() {
     }
 }
 
+void RenderableModelEntityItem::setDimensions(const glm::vec3& value) {
+    _dimensionsInitialized = true;
```
huffman: What happens if the dimensions are the default? What if I change the dimensions back to the default, after it has been resized? I'm asking because I'm wondering about the following 2 cases:

 * Does the server send dimensions if they are the default? I know this was recognized as an optimization we could do awhile back.
 * Is there a possible race here, where we have received some information about an entity, but not all of its properties?

Atlante45: No because everything coming from the server is using markAllChanged so everything is updated the first time.
That means any entity coming/going to the server has `_dimensionsInitialized` set to false on construction.

The only case where it stays true is for local created entities.
There is on case where the resize will not happen, which is if you create an entity and leave the domain before the model is loaded.
But I think it's a minor case.

huffman: Ah, got it. Sounds good!

Atlante45:
@ZappoMan Since I was in there, I removed `CONSTRUCT_PROPERTY` and modified `DEFINE_POPERTY*` to take a default value.
That way, we know for sure all properties are initialized.

hifi-gustavo:
Build console output for commit: ae83dfe041be5e6140e9a9fea664542e81bb5420 available at: https://gist.github.com/cefe6dbfa441636d24bb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6038.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6038.dmg

murillodigital:
test this please

hifi-gustavo:
Build console output for commit: ae83dfe041be5e6140e9a9fea664542e81bb5420 available at: https://gist.github.com/cefe6dbfa441636d24bb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6038.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6038.dmg

libraries/entities/src/AtmospherePropertyGroup.h
```diff
@@ -49,8 +49,8 @@ class ReadBitstreamToTreeParams;
 
 class AtmospherePropertyGroup : public PropertyGroup {
 public:
-    AtmospherePropertyGroup();
-    virtual ~AtmospherePropertyGroup() {}
+    AtmospherePropertyGroup() = default;
+    virtual ~AtmospherePropertyGroup() = default;
```
ZappoMan: maybe we can just drop this?

libraries/entities/src/AnimationPropertyGroup.h
```diff
@@ -13,26 +13,26 @@
...
-#include <stdint.h>
-#include <glm/glm.hpp>
-
 
 class AnimationPropertyGroup : public PropertyGroup {
 public:
-    AnimationPropertyGroup();
-    virtual ~AnimationPropertyGroup() {}
+    AnimationPropertyGroup() = default;
+    virtual ~AnimationPropertyGroup() = default;
```
ZappoMan: maybe we can just drop this?

libraries/entities/src/EntityItemProperties.h
```diff
@@ -60,7 +66,7 @@ class EntityItemProperties {
     friend class PolyLineEntityItem; // TODO: consider removing this friend relationship and use public methods
 public:
     EntityItemProperties(EntityPropertyFlags desiredProperties = EntityPropertyFlags());
-    virtual ~EntityItemProperties();
+    virtual ~EntityItemProperties() = default;
```
ZappoMan: maybe we can just drop this?

libraries/entities/src/StagePropertyGroup.h
```diff
@@ -29,8 +29,8 @@ class ReadBitstreamToTreeParams;
 
 class StagePropertyGroup : public PropertyGroup {
 public:
-    StagePropertyGroup();
-    virtual ~StagePropertyGroup() {}
+    StagePropertyGroup() = default;
+    virtual ~StagePropertyGroup() = default;
```
ZappoMan: maybe we can just drop this?

libraries/entities/src/EntityItemProperties.h
Atlante45: No we need this one to declare the destructor virtual.
Otherwise it might not run in some polymorphism cases.

hifi-gustavo:
Build console output for commit: 3e71c4cc4e9f15bc80955fea7f239ce9255de6fb available at: https://gist.github.com/cefe6dbfa441636d24bb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6038.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6038.dmg

libraries/entities/src/EntityItemProperties.h
ZappoMan: ok

PR #6040, opened 2015-10-09
Fix case where we deref a null pointer in OctreeInboundPacketProcessor

huffman:


libraries/networking/src/ReceivedPacketProcessor.cpp
```diff
@@ -88,6 +88,7 @@ bool ReceivedPacketProcessor::process() {
 
 void ReceivedPacketProcessor::nodeKilled(SharedNodePointer node) {
     lock();
+    _packets.remove_if([&](const NodeSharedPacketPair& pair) { return pair.first == node; });
```
huffman: @ZappoMan It didn't look like we were cleaning up old packets when a node was killed - would you have any input on this?

hifi-gustavo:
Build console output for commit: 4dddb508a8917796ae058d03cd0f4bdf4712e19e available at: https://gist.github.com/86cb6ba4a8d307f157f9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6040.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6040.dmg

libraries/networking/src/ReceivedPacketProcessor.cpp
ZappoMan: well - I don't think we want to just "drop" packets because the client disconnected... we should still process those packets...

Consider the case where a client connects, sends a bunch of edits to the server and disconnects... just because the client disconnects we don't want to ignore those edits... we still want to process those edits. So... no, I don't think we want to remove or skip inbound packets just because the node disconnected.

assignment-client/src/octree/OctreeInboundPacketProcessor.cpp
```diff
@@ -261,6 +261,12 @@ int OctreeInboundPacketProcessor::sendNackPackets() {
         }
 
         const SharedNodePointer& destinationNode = DependencyManager::get<NodeList>()->nodeWithUUID(nodeUUID);
+        // If the node no longer exists, wait until the ReceivedPacketProcessor has cleaned up the node
+        // to remove it from our stats list.
+        // FIXME Is it safe to clean it up here before ReceivedPacketProcess has?
+        if (!destinationNode) {
```
ZappoMan: it is reasonable to not send NACKed packets to a client that has disconnected... so long as we still process any inbound edit packets from that client

libraries/networking/src/ReceivedPacketProcessor.cpp
huffman: Ah, good point, I didn't realize they were still being processed after the node was killed. Thanks!

assignment-client/src/octree/OctreeInboundPacketProcessor.cpp
huffman: I reverted the other change, so I think that will satisfy this.

hifi-gustavo:
Build console output for commit: 21e9d30bbe282f3971cd0bd5bd4b9758479ccfa8 available at: https://gist.github.com/5e3b49801b96274b4952<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6040.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6040.dmg

hifi-gustavo:
Build console output for commit: 36aaffce43891cd50b490d9ead43c86e2089cb80 available at: https://gist.github.com/86cb6ba4a8d307f157f9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6040.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6040.dmg

murillodigital:
test this please

hifi-gustavo:
Build console output for commit: 36aaffce43891cd50b490d9ead43c86e2089cb80 available at: https://gist.github.com/86cb6ba4a8d307f157f9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6040.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6040.dmg

PR #6042, opened 2015-10-09
Fix not stopping emitting particles when set isEmitting false

ctrlaltdavid:


libraries/entities/src/ParticleEffectEntityItem.cpp
```diff
@@ -663,7 +663,7 @@ void ParticleEffectEntityItem::stepSimulation(float deltaTime) {
     }
 
     // emit new particles, but only if we are emmitting
-    if (isEmittingParticles() && _emitRate > 0.0f && _lifespan > 0.0f && _polarStart <= _polarFinish) {
+    if (getIsEmitting() && _emitRate > 0.0f && _lifespan > 0.0f && _polarStart <= _polarFinish) {
```
ZappoMan: Can we double check this is what you really want? I ask because originally I copied the old code that didn't check the "isAnimationPlaying" because it wanted to keep animating EXISTING (previously emitted) particles. Does that make sense?

ctrlaltdavid: isEmittingParticles() is true if there are any particles in existence to animate, and this guard is the test to determine whether or not to create new particles. So once there is one particle, more particles will always be created.
Tested by confirming that particlesTest.js successfully starts emitting, stops emitting, and restarts emitting. (It didn't before, which is why I noticed the bug.)

ZappoMan: ahhhhh ok - cool! so the original original code was buggy... cool.

ctrlaltdavid: It used to work correctly, perhaps by mistake ; )

hifi-gustavo:
Build console output for commit: 09e5b30c160a1f368c9e9832b919252443c93993 available at: https://gist.github.com/fd82ff7699024d60387c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6042.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6042.dmg

murillodigital:
test this please

hifi-gustavo:
Build console output for commit: 09e5b30c160a1f368c9e9832b919252443c93993 available at: https://gist.github.com/13b3c3234b88930490ad<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6042.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6042.dmg

PR #6046, opened 2015-10-09
add a userdata flag to toggle ignoreForCollisions while holding an entity

sethalves:
- add a userdata flag to toggle ignoreForCollisions while holding an entity


sethalves:
to test, set userData to {"grabbableKey":{"invertSolidWhileHeld":true}}

PR #6049, opened 2015-10-10
grab script picks less

sethalves:
- in it's "search" state, the grab script picks about 5x per second rather than about 60x


ZappoMan:
looks like your two different grab PRs conflict with each other... @sethalves

sethalves:
man.  you try to do the right thing.

PR #6063, opened 2015-10-12
[Scripts] Wait until toybox gun targets stop moving to reset

imgntn:
This PR adds some logic so that the targets in Toybox don't reset until they've stopped moving.


imgntn:
closing and rolling this into another commit

PR #6069, opened 2015-10-13
Hack in NewControllers into the JS layer of interface to allow testing

ZappoMan:


hifi-gustavo:
Build console output for commit: 9a9bdbbc4462d78769f08ce2f58241e5a98b7b65 available at: https://gist.github.com/ed0ff8cc225e6b7a01bd<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6069.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6069.dmg

examples/example/scripts/controllerScriptingExamples.js
```diff
@@ -12,6 +12,20 @@
 // Assumes you only have the default keyboard connected
 
 
+
+var hydra = Controller.Hardware.Hydra2;
```
jherico: We have to figure out a better way of doing this.  Right now if you have a joystick and a hydra connected, then it becomes Controller.Hardware.Hydra3.

Either we should design for the idea that you're just mapping to a generic hydra:  `Controllers.Hardware.Hydra` or that you're mapping to a specific indexed hydra `Controllers.Hardware.Hydra[0]`.

jherico: Alternatively, there's an example of finding the first XBox Controller under the current scheme in this commit:  https://github.com/jherico/hifi/commit/8a0540234cf0979ef7654ccf91962f68cdecf676

PR #6072, opened 2015-10-13
Don't reload/recenter some avatar data on startup

howard-stearns:
because we don't necessarily have all the required data initialized.
This should fix another cause of being in a garbage place on startup.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1559,7 +1559,7 @@ void Application::keyPressEvent(QKeyEvent* event) {
                         cursor->setIcon(Cursor::Icon::DEFAULT);
                     }
                 } else {
-                    resetSensors();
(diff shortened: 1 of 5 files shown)


hifi-gustavo:
Build console output for commit: 93e908a592093671fbd6c1e19c97c2a58a5e54f6 available at: https://gist.github.com/2dacf6ac44092f47456a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6072.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6072.dmg

AndrewMeadows:
I tested local builds of hifi/master and this branch -- the changes here appear to fix the "arrive way out near infinity on login" problem I've been experiencing.

PR #6073, opened 2015-10-13
implement comfort mode hack

ZappoMan:


hifi-gustavo:
Build console output for commit: ad96d7692191bd959f7760cb6b740d8f19d26493 available at: https://gist.github.com/619bedf6f80000a259a8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6073.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6073.dmg

AndrewMeadows:
This works.  However it is **enabled** by default (is this what we want)?  It persists in settings and only affects HMD mode.

Note: the timer around the right/left comport turns is not reset when I let go of the control keys.  Perhaps this is what we really want but as a user I expected to be able to tap it faster and get a turn every tap.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1514,33 +1514,68 @@ bool MyAvatar::shouldRenderHead(const RenderArgs* renderArgs) const {
...
+            _bodyYawDelta = (1.0f - blend) * _bodyYawDelta + blend * targetSpeed;
         }
-        _bodyYawDelta = (1.0f - blend) * _bodyYawDelta + blend * targetSpeed;
-    } else if (_bodyYawDelta != 0.0f) {
-        // attenuate body rotation speed
-        const float ROTATION_DECAY_TIMESCALE = 0.05f;
-        float attenuation = 1.0f - deltaTime / ROTATION_DECAY_TIMESCALE;
-        if (attenuation < 0.0f) {
-            attenuation = 0.0f;
+        else if (_bodyYawDelta != 0.0f) {
```
howard-stearns: Should be combined with the previous line per coding standard 3.4.1?

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1514,33 +1514,68 @@ bool MyAvatar::shouldRenderHead(const RenderArgs* renderArgs) const {
...
+        setOrientation(getOrientation() *
+            glm::quat(glm::radians(glm::vec3(0.0f, _bodyYawDelta * deltaTime, 0.0f))));
+
+    } else {
+        // comfort mode....
+        _bodyYawDelta = 0.0f;
+
+        static quint64 lastPulse = 0;
+        quint64 now = usecTimestampNow();
+        quint64 COMFORT_MODE_PULSE_TIMING = USECS_PER_SECOND / 2; // turn once per second
```
howard-stearns: Is it once per second or once per half second. I'm confused.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1514,33 +1514,68 @@ bool MyAvatar::shouldRenderHead(const RenderArgs* renderArgs) const {
...
         }
-        _bodyYawDelta *= attenuation;
 
-        float MINIMUM_ROTATION_RATE = 2.0f;
-        if (fabsf(_bodyYawDelta) < MINIMUM_ROTATION_RATE) {
-            _bodyYawDelta = 0.0f;
+        // update body orientation by movement inputs
+        setOrientation(getOrientation() *
+            glm::quat(glm::radians(glm::vec3(0.0f, _bodyYawDelta * deltaTime, 0.0f))));
+
```
AndrewMeadows: The "rotate every tap" behavior I'd like to see could be achieved by moving **lastPulse** up outside of this **if/else** block for expanded context and then clearing it to zero in the **if** case.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1514,33 +1514,68 @@ bool MyAvatar::shouldRenderHead(const RenderArgs* renderArgs) const {
...
+        _bodyYawDelta = 0.0f;
+
+        static quint64 lastPulse = 0;
+        quint64 now = usecTimestampNow();
+        quint64 COMFORT_MODE_PULSE_TIMING = USECS_PER_SECOND / 2; // turn once per second
+
+        float driveLeft = _driveKeys[ROT_LEFT];
+        float driveRight= _driveKeys[ROT_RIGHT];
+
+        if ((driveLeft != 0.0f || driveRight != 0.0f) && (now - lastPulse > COMFORT_MODE_PULSE_TIMING)) {
```
howard-stearns: It feels hard to explain this timing behavior in words to a user?

interface/src/avatar/MyAvatar.cpp
AndrewMeadows: Yeah, I noticed that too.  The comment is wrong, but since it is a HACK that will hopefully be removed I decided to not worry about it.

interface/src/avatar/MyAvatar.cpp
ZappoMan: damn you Visual Studio!!

interface/src/avatar/MyAvatar.cpp
ZappoMan: fixing in a moment

interface/src/avatar/MyAvatar.cpp
ZappoMan: fixing in a moment

interface/src/avatar/MyAvatar.cpp
ZappoMan: fixing in a moment

ZappoMan:
@AndrewMeadows - re enabled by default... YES... we've been told this is the desired novice behavior

interface/src/avatar/MyAvatar.cpp
ZappoMan: added a block comment explaining the behavior

hifi-gustavo:
Build console output for commit: 8a2e23cae4909057bb1bf2c7c9505b704dcaebe6 available at: https://gist.github.com/619bedf6f80000a259a8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6073.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6073.dmg

PR #6084, opened 2015-10-15
move DIRTY_FLAGS out of EntityItem namespace

AndrewMeadows:
I have a dream of unraveling the spaghetti between the **physics** and **entities** libs.  This PR is a step in that direction.

The **DIRTY_FOO** flags are used to communicate what kind of simulation properties have been changed and need to be relayed to the physics simulation.  They were originally flags for entities but they are necessary for all objects that happen to be in the physics simulation (e.g. avatars).  So I'm pulling them out of the **EntityItem** namespace and putting them in a new one called **Simulation**.  Eventually the **Simulation** namespace will be moved out of the **entities** lib and into the **physics** lib.

### DIFF
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -1034,7 +1034,7 @@ int Avatar::parseDataFromBuffer(const QByteArray& buffer) {
     const float MOVE_DISTANCE_THRESHOLD = 0.001f;
     _moving = glm::distance(oldPosition, _position) > MOVE_DISTANCE_THRESHOLD;
     if (_moving && _motionState) {
-        _motionState->addDirtyFlags(EntityItem::DIRTY_POSITION);
+        _motionState->addDirtyFlags(Simulation::DIRTY_POSITION);
     }
     endUpdate();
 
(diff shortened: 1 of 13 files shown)


interface/src/avatar/Avatar.cpp
```diff
@@ -1034,7 +1034,7 @@ int Avatar::parseDataFromBuffer(const QByteArray& buffer) {
     const float MOVE_DISTANCE_THRESHOLD = 0.001f;
     _moving = glm::distance(oldPosition, _position) > MOVE_DISTANCE_THRESHOLD;
     if (_moving && _motionState) {
-        _motionState->addDirtyFlags(EntityItem::DIRTY_POSITION);
+        _motionState->addDirtyFlags(Simulation::DIRTY_POSITION);
     }
```
AndrewMeadows: Clearly the avatar's MotionState shouldn't be using **EntityItem** flags.

hifi-gustavo:
Build console output for commit: fa17b77d70a565265e92611da0f887372a03cc25 available at: https://gist.github.com/d33f6cdaad82c35f246c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6084.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6084.dmg

PR #6097, opened 2015-10-16
Expose animation vars to Javascript

howard-stearns:
AvatarList.getAvatars() => list of avatar identifiers
AvatarList.getAvatar(avatarIdentifier) => an Avatar object (with a subset of MyAvatar capabilities)

MyAvatar.addAnimationStateHandler(handler, propertiesList)
MyAvatar.removeAnimationStateHandler(handler)
Adds or removes handler(animStateDictionaryIn) => animStateDictionaryOut, which will be invoked just before each animGraph state update.

The handler will be called with an animStateDictionaryIn that has all those properties specified by the (possibly empty) propertiesList argument. However for debugging, if the properties argument is null, all internal animGraph state is provided.

The animStateDictionaryOut can be a different object than animStateDictionaryIn. Any properties set in animStateDictionaryOut will override those of the internal animation machinery.

The animStateDictionaryIn may be shared among multiple handlers, and thus may contain additional properties specified when adding one of the other handlers. While any handler may change a value in animStateDictionaryIn (or supply different values in animStateDictionaryOut) a handler must not remove properties from animStateDictionaryIn, nor change property values that it does not intend to change. It is not specified in what order multiple handlers are called.


hifi-gustavo:
Build console output for commit: f25cc93936229efc139bd4e5b318cb22643353b5 available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

hifi-gustavo:
Build console output for commit: 770282ad8849afdf0b5e7f87acaf89ce1f5ae54d available at: https://gist.github.com/af6fc076c60a555298a3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

hifi-gustavo:
Build console output for commit: 38a967ac5493ba53940f8b768a0808d22f77b0f2 available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

interface/src/avatar/AvatarManager.h
```diff
@@ -52,6 +52,10 @@ class AvatarManager : public AvatarHashMap {
     
     Q_INVOKABLE void setLocalLights(const QVector<AvatarManager::LocalLight>& localLights);
     Q_INVOKABLE QVector<AvatarManager::LocalLight> getLocalLights() const;
+    // Currently, your own avatar will be included as the null avatar id.
+    Q_INVOKABLE QVector<QUuid> getAvatars() const { return _avatarHash.keys().toVector(); }
```
huffman: Given that `getAvatar` returns `AvatarData`, I would expect this to return an array of AvatarData. Given that, I think we should name this something like `getAvatarIDs`

interface/src/avatar/AvatarManager.h
```diff
@@ -52,6 +52,10 @@ class AvatarManager : public AvatarHashMap {
...
+    // Minor Bug: A bogus avatarID answers your own avatar.
+    Q_INVOKABLE AvatarData* getAvatar(QUuid avatarID) const { return _avatarHash[avatarID].get(); }
```
huffman: For this and `getAvatars`, are these thread-safe? It doesn't look like they are, and this method will be invoked on a script thread if being called from a script.

ZappoMan: we definitely need to protect _avatarHash since as Huffman points out scripts are always run on separate threads, and _avatarHash is also manipulated by the main thread.

hifi-gustavo:
Build console output for commit: 2213a4bb021962ae38aa90f77d2dc0311aed2972 available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

interface/src/avatar/AvatarManager.h
howard-stearns: I'm not sure what updateAvatarRenderStatus(), and isAvatarInRange() are meant to accomplish, but it looks like regardless of the status of this PR, we'll have to make those two thread safe for the same reasons.

hifi-gustavo:
Build console output for commit: 4c6867bb9cd5eac827bb46d99368123fc7bb514e available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

hifi-gustavo:
Build console output for commit: 574a51e83140e7784e7ee1e7ab37d6e23bd4df5f available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

hyperlogic:
One suggestion on the addAnimationStateHandler function, if you return an id, you can later pass to removeAnimationStateHandler.  This follows the style of the setTimeout functions that JavaScript programmers should be familiar with.

libraries/animation/src/Rig.cpp
```diff
@@ -604,6 +605,42 @@ void Rig::computeMotionAnimationState(float deltaTime, const glm::vec3& worldPos
     _lastPosition = worldPosition;
 }
 
+void Rig::addAnimationStateHandler(QScriptValue handler, QScriptValue propertiesList) {
```
hyperlogic: probably obvious, but probably need a lock here, QScriptValue assignment is likely not atomic, and this will become a list in the future, which we wont't want to mutate while the main thread is iterating over it.

libraries/animation/src/Rig.cpp
```diff
@@ -604,6 +605,42 @@ void Rig::computeMotionAnimationState(float deltaTime, const glm::vec3& worldPos
...
+    _stateHandlers = handler;
+}
+void Rig::removeAnimationStateHandler(QScriptValue handler) {
```
hyperlogic: Same here, need a lock

libraries/animation/src/Rig.cpp
```diff
@@ -604,6 +605,42 @@ void Rig::computeMotionAnimationState(float deltaTime, const glm::vec3& worldPos
...
+    // TODO: iterate multiple handlers, but with one shared arg.
+    // TODO: fill the properties based on the union of requested properties. (Keep all properties objs and compute new union when add/remove handler.)
+    // TODO: check QScriptEngine::hasUncaughtException()
+    // TODO: call asynchronously (through a signal on script), so that each script is single threaded, and so we never block here.
+    //       This will require inboundMaps to be kept in the list of per-handler data.
+    QScriptEngine* engine = _stateHandlers.engine();
+    QScriptValue outboundMap = _animVars.animVariantMapToScriptValue(engine);
+    QScriptValueList args;
+    args << outboundMap;
+    _stateHandlersResults = _stateHandlers.call(QScriptValue(), args);
```
hyperlogic: Dumb question, does this block the main thread? if the Script thread happens to be busy doing something?

libraries/animation/src/Rig.cpp
```diff
@@ -604,6 +605,42 @@ void Rig::computeMotionAnimationState(float deltaTime, const glm::vec3& worldPos
...
 }
 
+void Rig::addAnimationStateHandler(QScriptValue handler, QScriptValue propertiesList) {
+    _stateHandlers = handler;
+}
+void Rig::removeAnimationStateHandler(QScriptValue handler) {
+    _stateHandlersResultsToRemove = _stateHandlersResults;
+    _stateHandlers = _stateHandlersResults = QScriptValue();
+}
+void Rig::cleanupAnimationStateHandler() {
```
hyperlogic: Should this be invoked by Rig::removeAnimationStateHandler?

hifi-gustavo:
Build console output for commit: 9fd61907f51aa79d136f5cbf1cb6a7032b3654d8 available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

libraries/animation/src/AnimVariantMap.cpp
```diff
@@ -0,0 +1,96 @@
...
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include <QScriptEngine>
+#include <QScriptValueIterator>
+#include <RegisteredMetaTypes.h>
+#include "AnimVariant.h"
+
+QScriptValue AnimVariantMap::animVariantMapToScriptValue(QScriptEngine* engine) const {
+    QScriptValue target = engine->newObject();
```
ZappoMan: QScriptEngine is not thread safe, so you need to make sure that you only call things like engine->newObject() on the thread of the QScriptEngine.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -898,14 +923,20 @@ void ScriptEngine::load(const QString& loadFile) {
     }
 }
 
-// Look up the handler associated with eventName and entityID. If found, evalute the argGenerator thunk and call the handler with those args
-void ScriptEngine::generalHandler(const EntityItemID& entityID, const QString& eventName, std::function<QScriptValueList()> argGenerator) {
+bool ScriptEngine::checkThread() const {
```
ZappoMan: this naming convention seems backward "checkThread()" returns true for failure? I think the opposite would read better.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -898,14 +923,20 @@ void ScriptEngine::load(const QString& loadFile) {
...
         assert(false);
-        return;
+        return true;
     }
+    return false;
+}
 
+// Look up the handler associated with eventName and entityID. If found, evalute the argGenerator thunk and call the handler with those args
+void ScriptEngine::generalHandler(const EntityItemID& entityID, const QString& eventName, std::function<QScriptValueList()> argGenerator) {
+    if (checkThread()) {
```
ZappoMan: this naming convention seems backward "checkThread()" returns true for failure? I think the opposite would read better. -- also it seems odd that this silently fails.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -718,6 +728,21 @@ void ScriptEngine::stop() {
     }
 }
 
+// Other threads can invoke this through invokeMethod, which causes the callback to be asynchronously executed in this script's thread.
+void ScriptEngine::invokeAnimationCallback(QScriptValue callback, AnimVariantMap parameters) {
```
ZappoMan: honestly I'd prefer we use the other pattern used in ScriptEngine where the method self-invokes if called on the wrong thread. Since the parameters are bassed by value, and there's no return value, the self-invoke can cleanly be asynchronous

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -718,6 +728,21 @@ void ScriptEngine::stop() {
...
+    checkThread();
+    QScriptValue javascriptParametgers = parameters.animVariantMapToScriptValue(this);
+    QScriptValueList callingArguments;
+    callingArguments << javascriptParametgers;
+    QScriptValue result = callback.call(QScriptValue(), callingArguments);
+    // We want to give the result back to the rig, but we don't have the rig or the avatar. But the global does.
+    // This is sort of like going through DependencyManager.get.
+    QScriptValue resultHandler = globalObject().property("MyAvatar").property("animationStateHandlerResult");
```
ZappoMan: this is a strange way to get back to the MyAvatar interface... more importantly... ScriptEngine objects also appear in AC agents, and there they don't have a MyAvatar with a animationStateHandlerResult property. This feels fragile.

libraries/animation/src/Rig.cpp
```diff
@@ -604,6 +605,34 @@ void Rig::computeMotionAnimationState(float deltaTime, const glm::vec3& worldPos
...
+void Rig::animationStateHandlerResult(QScriptValue handler, QScriptValue result) { // called synchronously from script
+    // handler is currently ignored but might be used in storing individual results
+    QMutexLocker locker(&_stateMutex);
+    _stateHandlersResults.animVariantMapFromScriptValue(result); // Into our own copy.
+}
+void Rig::updateAnimationStateHandlers() { // called on avatar update thread (which may be main thread)
+    if (_stateHandlers.isValid()) {
+        // invokeMethod makes a copy of the args, and copies of AnimVariantMap do copy the underlying map, so this will correctly capture
+        // the state of _animVars and allow continued changes to _animVars in this thread without conflict.
+        QMetaObject::invokeMethod(_stateHandlers.engine(), "invokeAnimationCallback",  Qt::QueuedConnection,
```
ZappoMan: see my other notes on this... I'm not a fan of this pattern. all the other public methods on ScriptEngine have been made thread safe by embedding the self-invocation... this approach breaks that.Just call engine->animationCallback() and put the thread safety check inside of ScriptEngine

interface/src/avatar/AvatarManager.h
ZappoMan: this doesn't look thread safe... since _avatarHash isn't guaranteed to be thread safe

hifi-gustavo:
Build console output for commit: a58539be8249c91a451dee57a17846e47c074b7a available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

hifi-gustavo:
Build console output for commit: 30429e8138776e3f56d2effc82972fb43dd26b3b available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

interface/src/avatar/AvatarManager.h
howard-stearns: Thanks. Proposed solution (for existing case) in https://github.com/highfidelity/hifi/pull/6154
(Looks to me like updateAvatarRenderStatus is ok.)

birarda:
test this please

hifi-gustavo:
Build console output for commit: 30429e8138776e3f56d2effc82972fb43dd26b3b available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

hifi-gustavo:
Build console output for commit: dbf50728142828c2d640aa118e6ac582440d582d available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

hifi-gustavo:
Build console output for commit: 773e4014cae504b5fe7c879f319a48ee1b6ed78d available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

hifi-gustavo:
Build console output for commit: 92ddedd44b4d1487ea2e802351617136ab38cf0d available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

hifi-gustavo:
Build console output for commit: 84cfeaec13f79615e788815b14ef937dc76a10d5 available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

hifi-gustavo:
Build console output for commit: e11b0add9a63d5497dbb30c184a18489cacb331d available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

hifi-gustavo:
Build console output for commit: 4b4907c9ef8bb4398d1db9cb7c8285c823c758e5 available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

howard-stearns:
One limitation of this API is that scripts don't have a way to remove an animvar that they add. Not a big deal.
Also, it's clumsier than I'd like for a script to set an animvar once, to remain indefinitely until set differently by the script, without leaving a handler in place in between. (It can be done, but it's clunky.)

Neither are tombstone issues, and both can be addressed later if we want to.

hifi-gustavo:
Build console output for commit: 3c6d4f9c221ac19c8f3b12070ee396c52f72bec1 available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

libraries/animation/src/AnimVariantMap.cpp
```diff
@@ -0,0 +1,118 @@
...
+//
+
+#include <QScriptEngine>
+#include <QScriptValueIterator>
+#include <QThread>
+#include <RegisteredMetaTypes.h>
+#include "AnimVariant.h"
+
+QScriptValue AnimVariantMap::animVariantMapToScriptValue(QScriptEngine* engine, const QStringList& names, bool useNames) const {
+    if (QThread::currentThread() != engine->thread()) {
```
hyperlogic: Perhaps we should make this an assert as well?

libraries/animation/src/AnimVariantMap.cpp
```diff
@@ -0,0 +1,118 @@
...
+        qCWarning(animation) << "Cannot create Javacript object from non-script thread" << QThread::currentThread();
+        return QScriptValue();
+    }
+    QScriptValue target = engine->newObject();
+    auto setOne = [&] (QString name, AnimVariant value) {
```
hyperlogic: Consider making the `value` argument a const AnimVariant ref, to avoid a copy.  I would suggest the same for `name` as well, even though under the hood, it apparently does not make a copy of the string bytes.

hifi-gustavo:
Build console output for commit: 502cc7f580c90483b65d10638f2e622a40c8c702 available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

libraries/animation/src/Rig.cpp
```diff
@@ -576,6 +577,70 @@ void Rig::computeMotionAnimationState(float deltaTime, const glm::vec3& worldPos
     _lastPosition = worldPosition;
 }
 
+// Allow script to add/remove handlers and report results, from within their thread.
+QScriptValue Rig::addAnimationStateHandler(QScriptValue handler, QScriptValue propertiesList) { // called in script thread
+    QMutexLocker locker(&_stateMutex);
+    while (!_nextStateHandlerId || _stateHandlers.contains(_nextStateHandlerId)) { // 0 is unused, and don't reuse existing after wrap.
```
hyperlogic: Just a suggestion, start _nextStateHandlerId at 1.  Then use it as your index then increment it before the return.  Then we can get rid of this while loop, which is kind of klunky.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -743,6 +766,27 @@ void ScriptEngine::stop() {
     }
 }
 
+// Other threads can invoke this through invokeMethod, which causes the callback to be asynchronously executed in this script's thread.
+void ScriptEngine::callAnimationStateHandler(QScriptValue callback, AnimVariantMap parameters, QStringList names, bool useNames, AnimVariantResultHandler resultHandler) {
+    if (QThread::currentThread() != thread()) {
+#ifdef THREAD_DEBUGGING
+        qDebug() << "*** WARNING *** ScriptEngine::callAnimationStateHandler() called on wrong thread [" << QThread::currentThread() << "], invoking on correct thread [" << thread() << "]  name:" << name;
```
hyperlogic: Maybe assert here as well?

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -743,6 +766,27 @@ void ScriptEngine::stop() {
...
+#endif
+        QMetaObject::invokeMethod(this, "callAnimationStateHandler",
+                                  Q_ARG(QScriptValue, callback),
+                                  Q_ARG(AnimVariantMap, parameters),
+                                  Q_ARG(QStringList, names),
+                                  Q_ARG(bool, useNames),
+                                  Q_ARG(AnimVariantResultHandler, resultHandler));
+        return;
+    }
+    QScriptValue javascriptParametgers = parameters.animVariantMapToScriptValue(this, names, useNames);
```
hyperlogic: Misspelling of parameters? Parametgers

libraries/animation/src/Rig.cpp
howard-stearns: The while loop is in case scripts add so many handlers that we wrap our int32. (Imagine if several scripts add in an update, removing after an indeterminate number of cycles.) I want the normal case to be pretty efficient in add and remove, and the silly case to handle lots even if inefficient in adding them.

libraries/animation/src/AnimVariantMap.cpp
howard-stearns: I guess the comment overlapped my checkin. Line 21.

libraries/animation/src/AnimVariantMap.cpp
```diff
@@ -0,0 +1,118 @@
...
+            continue;
+        } else if (value.isNumber()) {
+            int asInteger = value.toInt32();
+            float asFloat = value.toNumber();
+            if (asInteger == asFloat) {
+                set(property.name(), asInteger);
+            } else {
+                set(property.name(), asFloat);
+            }
+            continue;
```
hyperlogic: The above code is somewhat problematic. When animating float values, like say blend values, we might convert between floats and integers as we hit integer boundaries.  This would cause unexpected values for non 0 values, because of the current implementation of AnimVariant.getFloat() will cast the integer bits to a float.

We can fix this by adding a case for converting between integers and floats in the AnimVariant class.  If you'd like I can make a patch.  But it would look something like this:

    int getInt() const { assert(_type == Type::Int || _type == Type::Float); return _type == Type::Float ? (int)_val.floatVal : _val.intVal; }
    float getFloat() const { assert(_type == Type::Float || _type == Type::Int); return _type == Type::Int ? (float)_val.intVal : _val.floats[0]; }

libraries/animation/src/AnimVariantMap.cpp
```diff
@@ -0,0 +1,118 @@
...
+    return target;
+}
+void AnimVariantMap::copyVariantsFrom(const AnimVariantMap& other) {
+    for (auto& pair : other._map) {
+        _map[pair.first] = pair.second;
+    }
+}
+void AnimVariantMap::animVariantMapFromScriptValue(const QScriptValue& source) {
+    if (QThread::currentThread() != source.engine()->thread()) {
+        qCWarning(animation) << "Cannot examine Javacript object from non-script thread" << QThread::currentThread();
```
hyperlogic: assert just in case?

libraries/animation/src/AnimVariantMap.cpp
```diff
@@ -0,0 +1,118 @@
...
+    // we would enter it into the dictionary. Then switch on that type here, with the code that follow being executed only if
+    // the type is not known. One problem with that is that there is no checking that two different script use the same name differently.
+    QScriptValueIterator property(source);
+    // Note: QScriptValueIterator iterates only over source's own properties. It does not follow the prototype chain.
+    while (property.hasNext()) {
+        property.next();
+        QScriptValue value = property.value();
+        if (value.isBool()) {
+            set(property.name(), value.toBool());
+            continue;
```
hyperlogic: I don't think these continues are strictly necessary, I'd rather have duplicate the warning messages then have continues, they make the loop harder to follow.

libraries/animation/src/Rig.cpp
```diff
@@ -576,6 +577,70 @@ void Rig::computeMotionAnimationState(float deltaTime, const glm::vec3& worldPos
...
+    // rather than having each callAnimationStateHandler invocation make its own copy.
+    // However, that copying is done on the script's own time rather than ours, so even if it's less cpu, it would be more
+    // work on the avatar update thread (which is possibly the main thread).
+    for (auto data = _stateHandlers.begin(); data != _stateHandlers.end(); data++) {
+        // call out:
+        int identifier = data.key();
+        StateHandler& value = data.value();
+        QScriptValue& function = value.function;
+        auto handleResult = [this, identifier](QScriptValue result) {
+            animationStateHandlerResult(identifier, result);
```
hyperlogic: A warning comment here saying that this lambda will be invoked on a different thread would be helpful, I think

libraries/script-engine/src/ScriptEngine.cpp
howard-stearns: This one is different. The other's are an exceptional case that is a sign of misuse. (I added the asserts.) 

This one however, is a legitimate use, albeit one that is not actually used within this PR. The warning is not normally issued, but only in the special circumstance where someone compiles in such a way as to try to trace through what's happening. (Brad has done so recently.) The wording is perhaps confusing, but it is the wording used in the existing code throughout this file, for the same purpose.

libraries/animation/src/Rig.cpp
hyperlogic: true, but they would have to add a script handler once a frame at 90hz for 17 years before it flipped.  I guess a comment about the reason for the while loop would have helped me understand why it was there.

hyperlogic:
It's kind of strange that we have a AnimVariant.h file with no cpp file, and we have a AnimVariantMap.cpp file with no .h file.   Either we should break AnimVariantMap into a seperate class with it's own .h and .cpp file.  Or we should rename AnimVariantMap.cpp to AnimVariant.cpp

libraries/animation/src/Rig.cpp
howard-stearns: Fair enough re comment. But hey, I said the word "wrap". By my math, btw, 100 scripts at 60z would only run 8 days before wrapping, but maybe my math is wrong. What was it that early Windows runtime clock could run for before wrapping? They never though Windows could stay running for more than something like 50 days.

hifi-gustavo:
Build console output for commit: dcc173c93a5da52364666e78f85081a884055a4b available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

hifi-gustavo:
Build console output for commit: 1918f1835cfc44ee995ff2d0f64274d7cae66f5e available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

howard-stearns:
ok, changes for all the comments through now should be included.

hifi-gustavo:
Build console output for commit: eb9e54de41eb7add77d5cc91cf85eb0164375106 available at: https://gist.github.com/aad1c93a47e41841c7db<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6097.dmg

hyperlogic:
:+1:

PR #6102, opened 2015-10-17
Fix particle bug by having lifetimes increase

Atlante45:
Fix particle bug causing a crash in debug mode

### DIFF
--- a/libraries/entities/src/ParticleEffectEntityItem.cpp
+++ b/libraries/entities/src/ParticleEffectEntityItem.cpp
@@ -644,14 +644,13 @@ void ParticleEffectEntityItem::stepSimulation(float deltaTime) {
 
     // update particles between head and tail
     for (quint32 i = _particleHeadIndex; i != _particleTailIndex; i = (i + 1) % _maxParticles) {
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 1da68e02146056ce922d853db8a68c6ea7a86820 available at: https://gist.github.com/23d1b8010d59c6c007d0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6102.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6102.dmg

ZappoMan:
What's the repro case for the bug this fixes?

Atlante45:
@ZappoMan Just seen that. Repro was (in debug mode) start an emiter then set the lifespan property to a smaller value than it is. Lifetime was not guaranteed to be smaller than lifespan anymore which was causing an assert to fail.

ericrius1:
shnurp

ericrius1:
bdurp

PR #6110, opened 2015-10-19
ifdef out buggy texture map code

ZappoMan:
this code is causing a massive flood of QImage.setPixel() warnings on my machine in some domains.. it looking at the code, it appears to not actually do anything because the `QImage result` is not actually used after the removed section.

I've left it in here since Alessandro will be arriving in SF soon, and we can repair it correctly at that point


hifi-gustavo:
Build console output for commit: bddb0e1e7b279995b5dca2dba3ac42dbe5c9519c available at: https://gist.github.com/ce977dbe4bd0d1abb5ea<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6110.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6110.dmg

PR #6113, opened 2015-10-19
Joint rotation fix for FBX files exported from Blender.

hyperlogic:
This isn't a great fix.  But it will at least allow people to export avatars from Blender.


hyperlogic:
There is an issue with avatar's exported with blender and the AnimInverseKinematics system.  Blender will introduce a new root joint named "Armature" with zero rotation and translation.  This confuses the IK system which makes the assumption that the root joint are the hips of the Avatar.  @AndrewMeadows has proposed a possible fix for this.

hifi-gustavo:
Build console output for commit: 62c89eaf7d60df820bc1e799150366ed21dbd17c available at: https://gist.github.com/91aef77076f281a42253<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6113.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6113.dmg

PR #6115, opened 2015-10-19
fix model LOD behavior

ZappoMan:
A recent refactor was breaking the calculation of "mesh part" bounds... as a result, mesh parts were not being culled for either in/out of view and/or LOD size. This fixes that, and also cleans up some erroneous stats.

### DIFF
--- a/interface/resources/qml/Stats.qml
+++ b/interface/resources/qml/Stats.qml
@@ -181,15 +181,33 @@ Item {
                         color: root.fontColor;
                         font.pixelSize: root.fontSize
                         visible: root.expanded;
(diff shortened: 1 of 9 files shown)


hifi-gustavo:
Build console output for commit: 0b2bf8bda8725f445b8fc59e90862d542c424e19 available at: https://gist.github.com/756b7c7f3ab06db87bfa<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6115.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6115.dmg

PR #6116, opened 2015-10-19
hack to reduce hand influence of hips in HMD mode

AndrewMeadows:
There is a bug where during HMD+hydra sessions the hydra hand motion can have too strong of an influence on the hip positions.  This PR reduces the problem but does not completely fix it.  A more correct fix will have to be done later.

### DIFF
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -255,7 +255,7 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
                 }
 
                 // store the rotation change in the accumulator
-                _accumulators[pivotIndex].add(newRot);
+                _accumulators[pivotIndex].add(newRot, target.getWeight());
 
                 // this joint has been changed so we check to see if it has the lowest index
                 if (pivotIndex < lowestMovedIndex) {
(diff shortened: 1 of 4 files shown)


libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -255,7 +255,7 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
                 }
 
                 // store the rotation change in the accumulator
-                _accumulators[pivotIndex].add(newRot);
+                _accumulators[pivotIndex].add(newRot, target.getWeight());
```
AndrewMeadows: While solving each IK target the **RotationAccumulator** remembers all of the rotation contributions and can then be used to compute the "average" rotation.  The "fix" here is to give **IKTarget**s of type **HmdHead** more influence over the IK solution than the other target types.  It helps, and is an easy implementation so we'll try it out.

hifi-gustavo:
Build console output for commit: 917bfbf64e4a1dc81f0693ab597ada3f7e76f328 available at: https://gist.github.com/b8e8eb6955c20c003c39<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6116.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6116.dmg

PR #6127, opened 2015-10-20
gitattributes take 2

AndrewMeadows:
This is a second attempt to add a **.gitattributes** file.  The trick is to update the line endings of some JS scripts that somehow missed earlier sweeps for bad files.

Also adding CSS and CFG files as "text" in .gitattributes.


examples/html/walkStyle.css
```diff
@@ -1,48 +1,48 @@
...
-input[type=button] {
-    cursor: pointer;
-    background-color: #608e96;
-    border-color: #608e96;
-    border-radius: 3.75pt;
-    padding: 3.75pt 7.5pt;
-    border: 0;
-    color: #fff;
-    font-weight: bold;
-}
```
AndrewMeadows: This was another one of those files with mixed DOS/unix line endings.

hyperlogic:
AFAIK, the only way for clients to properly reflect these changes is to re-fresh all the files in their working directory, ether by re-cloning. or.

    $> git rm -r . --cached
    $> git reset --hard

https://help.github.com/articles/dealing-with-line-endings/

hifi-gustavo:
Build console output for commit: d02fe784a95771fbfa98449f40eb3bcbc1652eb8 available at: https://gist.github.com/f091e3e9278908ba9af5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6127.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6127.dmg

jherico:
Before this is merged I'd suggest that we have someone who has been having trouble with the line ending changes walk through merging it to their repo's master and then walk through merging it to their other branches as well.  Clearly part of the problem is that transitioning between the old line endings situation and the new is migrating everything you have locally, including any work branches you might have.

If we can get a solid guide on how to do this, or a confirmation that it's not problematic, then I'm good to go on this.

jherico:
Process for getting things working on your local repositories...

* Make if you're on windows that you have `autocrlf` set to true.  

`git config --global core.autocrlf true`

*Fetch from the upstream repository

`git fetch upstream` (or whatever you have for the main hifi repository)

Next, merge the upstream master into your own master branch

`git checkout master`
`git status` <- should return no local changes
`git merge upstream/master`
`git status`

If the second `git status` suddenly report a bunch of changed files with only whitespace changes, you need to clear and refresh the working tree.

`git rm --cached -r .`
`git reset --hard`

Now, for any local branches you have, you'll need to merge the changes into them...

`git checkout -f branchX`
`git rm --cached -r .`
`git reset --hard`
`git merge master`
`git rm --cached -r .`
`git reset --hard`

At this point `git status` should return a clean result with no local changes and you should be able to switch between `master` and `branchX` without a bunch of changed files suddenly appearing.  Repeat for each working branch you have.

PR #6130, opened 2015-10-21
grab script changes

sethalves:
- grab script: possible fix for line disappearing
- grab script: go from 3 pickrays back to 1
- grab script: put "sphere" test back in, but with fixed distance test
- grab script: require user-data to have wantsTrigger before non-colliding calls will happen

### DIFF
--- a/examples/controllers/handControllerGrab.js
+++ b/examples/controllers/handControllerGrab.js
@@ -49,6 +49,7 @@ var PICK_MAX_DISTANCE = 500; // max length of pick-ray
 // near grabbing
 //
 
+var GRAB_RADIUS = 0.3; // if the ray misses but an object is this close, it will still be selected
 var NEAR_GRABBING_ACTION_TIMEFRAME = 0.05; // how quickly objects move to their new position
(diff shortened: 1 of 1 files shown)


examples/controllers/handControllerGrab.js
```diff
@@ -226,14 +228,13 @@ function MyController(hand, triggerAction) {
             });
         } else {
             var age = Entities.getEntityProperties(this.pointer, "age").age;
-            Entities.editEntity(this.pointer, {
+            this.pointer = Entities.editEntity(this.pointer, {
```
ZappoMan: hmmm - not sure you should need this... the ID shouldn't change on you.

sethalves: I did this because I noticed that editEntity can return a null uuid as an error indicator.  In this case (if this ever actually happens), I would want to recreate the line.

ZappoMan: ok. Sounds reasonable - I guess it would be possible for your lifetime to expire between edits and/or someone else to just delete your line... honestly, I've never been a fan of these laser beams as visible entities in world... overlays make more sense to me... but I guess since this is what we seem to want... this is the best way to do that.

PR #6134, opened 2015-10-21
implement improvement in sphere test

ZappoMan:
This allows Entities.findEntity(sphere) to properly pick against proper spheres.

### DIFF
--- a/libraries/entities/src/EntityItemPropertiesDefaults.h
+++ b/libraries/entities/src/EntityItemPropertiesDefaults.h
@@ -21,8 +21,8 @@
 // There is a minor performance gain when comparing/copying an existing glm::vec3 rather than 
 // creating a new one on the stack so we declare the ZERO_VEC3 constant as an optimization.
 const glm::vec3 ENTITY_ITEM_ZERO_VEC3 = glm::vec3(0.0f);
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 246e46b69a8e6556aea29cdcaf4b678962f481a7 available at: https://gist.github.com/cecb21b7414f652ce0e8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6134.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6134.dmg

imgntn:
@ZappoMan if i wanted to write a script to test this, i would expect findEntities to find entities that the sphere only partially intersects, without hitting the center of the entity?

ZappoMan:
@imgntn - here's your test case, call Entities.findEntities({x:0, y:0, z:0}, 1) - so you're searching from origin for radius of 1.... In the current production code, the following sphere will "be found" even though it is not actually in that search sphere [dimension 0.2,0.2,0.2; position: 0.2, 1.07, 0.2]... if you make two spheres in world to visualize this, you'll see your smaller sphere is not in fact touching the search sphere, but the old code would say it was, the new code gets the answer correct.

imgntn:
cool, testing now

imgntn:
awesome, this works great.  thanks @ZappoMan

PR #6139, opened 2015-10-21
Controllers : Fixing the pose transmission

samcake:
The pose route is communicating the pose value if valid
this fixes the fact that the pose was not tagged as valid when it was
now the LeftHand and RIghtHand Actions can be used correctly


hifi-gustavo:
Build console output for commit: feeb4b68a6d1d238c06d4539d0635687ca8b007d available at: https://gist.github.com/4b606137566defd499d5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6139.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6139.dmg

PR #6145, opened 2015-10-21
DO NOTE MERGE[Scripts] Toybox AC Resetter updates

imgntn:
This PR makes the AC script for toybox to reset the balls and targets work.

### DIFF
--- a/examples/toybox/AC_scripts/originalPositionResetter.js
+++ b/examples/toybox/AC_scripts/originalPositionResetter.js
@@ -61,6 +61,7 @@ var Resetter = {
             y: 494.84,
             z: 475.06
         };
+
         var i;
         var ballPosition;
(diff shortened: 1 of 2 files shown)


imgntn:
test

imgntn:
test2

imgntn:
test3

imgntn:
@ZappoMan it's finding entities now (awesome!), but i'm trying to work through a bug where deleting entities from the client side doesn't delete them from the AC -- that is, i press the reset button in-world and from client side it shows correct number of entities (4 balls, 6 targets), but on the server the number of basketballs and targets keeps increasing.  any thoughts?

imgntn:
this bug still exists, took a different route to do the resetting but will try to get this working.  closing for now

PR #6147, opened 2015-10-22
Controllers Branch - Wiring up step yaw

jherico:
This PR also adds locking to the UserInputMapper to make access from the scripting interface more safe (hopefully), and changes the way the routes are processed.

This only exposes step yaw in the actions, it's not yet wired up to the avatar.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -2732,7 +2732,7 @@ void Application::update(float deltaTime) {
                 // controllers to provide a delta_per_second value rather than a raw delta.)
                 const float EXPECTED_FRAME_RATE = 60.0f;
                 float timeFactor = EXPECTED_FRAME_RATE * deltaTime;
-                myAvatar->setDriveKeys(PITCH, userInputMapper->getActionState(controller::Action::PITCH) / timeFactor);
+                myAvatar->setDriveKeys(PITCH, -1.0f * userInputMapper->getActionState(controller::Action::PITCH) / timeFactor);
                 myAvatar->setDriveKeys(YAW, -1.0f * userInputMapper->getActionState(controller::Action::YAW) / timeFactor);
                 myAvatar->setDriveKeys(STEP_YAW, -1.0f * userInputMapper->getActionState(controller::Action::STEP_YAW) / timeFactor);
             }
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -2718,15 +2718,13 @@ void Application::update(float deltaTime) {
     }
 
     // Transfer the user inputs to the driveKeys
+    // FIXME can we drop drive keys and just have the avatar read the action states directly?
     myAvatar->clearDriveKeys();
     if (_myCamera.getMode() != CAMERA_MODE_INDEPENDENT) {
         if (!_controllerScriptingInterface->areActionsCaptured()) {
-            myAvatar->setDriveKeys(FWD, userInputMapper->getActionState(controller::Action::LONGITUDINAL_FORWARD));
-            myAvatar->setDriveKeys(BACK, userInputMapper->getActionState(controller::Action::LONGITUDINAL_BACKWARD));
-            myAvatar->setDriveKeys(UP, userInputMapper->getActionState(controller::Action::VERTICAL_UP));
-            myAvatar->setDriveKeys(DOWN, userInputMapper->getActionState(controller::Action::VERTICAL_DOWN));
-            myAvatar->setDriveKeys(LEFT, userInputMapper->getActionState(controller::Action::LATERAL_LEFT));
-            myAvatar->setDriveKeys(RIGHT, userInputMapper->getActionState(controller::Action::LATERAL_RIGHT));
+            myAvatar->setDriveKeys(TRANSLATE_Z, -1.0f * userInputMapper->getActionState(controller::Action::TRANSLATE_Z));
+            myAvatar->setDriveKeys(TRANSLATE_Y, userInputMapper->getActionState(controller::Action::TRANSLATE_Y));
+            myAvatar->setDriveKeys(TRANSLATE_X, userInputMapper->getActionState(controller::Action::TRANSLATE_X));
             if (deltaTime > FLT_EPSILON) {
                 // For rotations what we really want are meausures of "angles per second" (in order to prevent 
                 // fps-dependent spin rates) so we need to scale the units of the controller contribution.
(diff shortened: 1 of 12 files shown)


hifi-gustavo:
Build console output for commit: c20fcafb52feea5d9e0b4184e77889d618e9600a available at: https://gist.github.com/8f6231dad4a7c14dc7f5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6147.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6147.dmg

hifi-gustavo:
Build console output for commit: 637654adeacce63cdd0da4495ec5b609818c7a8f available at: https://gist.github.com/8f6231dad4a7c14dc7f5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6147.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6147.dmg

hifi-gustavo:
Build console output for commit: 044a28212d6b2430715c5d68243a6891b0a70b84 available at: https://gist.github.com/8f6231dad4a7c14dc7f5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6147.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6147.dmg

ZappoMan:
Gak - I started adding a read/write locker to this class before seeing this... oh well...

hifi-gustavo:
Build console output for commit: afcec347ffab06a3ac9ce7a5367d2393bdde93f9 available at: https://gist.github.com/8f6231dad4a7c14dc7f5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6147.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6147.dmg

interface/src/Application.cpp
```diff
@@ -2718,15 +2718,13 @@ void Application::update(float deltaTime) {
...
     myAvatar->clearDriveKeys();
     if (_myCamera.getMode() != CAMERA_MODE_INDEPENDENT) {
         if (!_controllerScriptingInterface->areActionsCaptured()) {
-            myAvatar->setDriveKeys(FWD, userInputMapper->getActionState(controller::Action::LONGITUDINAL_FORWARD));
-            myAvatar->setDriveKeys(BACK, userInputMapper->getActionState(controller::Action::LONGITUDINAL_BACKWARD));
-            myAvatar->setDriveKeys(UP, userInputMapper->getActionState(controller::Action::VERTICAL_UP));
-            myAvatar->setDriveKeys(DOWN, userInputMapper->getActionState(controller::Action::VERTICAL_DOWN));
-            myAvatar->setDriveKeys(LEFT, userInputMapper->getActionState(controller::Action::LATERAL_LEFT));
-            myAvatar->setDriveKeys(RIGHT, userInputMapper->getActionState(controller::Action::LATERAL_RIGHT));
+            myAvatar->setDriveKeys(TRANSLATE_Z, -1.0f * userInputMapper->getActionState(controller::Action::TRANSLATE_Z));
```
ZappoMan: this seems dangerous and weird... why is the Action::TRANSLATE_Z a flipped axis from the Drive key?

interface/src/Application.cpp
```diff
@@ -2734,14 +2732,12 @@ void Application::update(float deltaTime) {
                 // controllers to provide a delta_per_second value rather than a raw delta.)
                 const float EXPECTED_FRAME_RATE = 60.0f;
                 float timeFactor = EXPECTED_FRAME_RATE * deltaTime;
-                myAvatar->setDriveKeys(ROT_UP, userInputMapper->getActionState(controller::Action::PITCH_UP) / timeFactor);
-                myAvatar->setDriveKeys(ROT_DOWN, userInputMapper->getActionState(controller::Action::PITCH_DOWN) / timeFactor);
-                myAvatar->setDriveKeys(ROT_LEFT, userInputMapper->getActionState(controller::Action::YAW_LEFT) / timeFactor);
-                myAvatar->setDriveKeys(ROT_RIGHT, userInputMapper->getActionState(controller::Action::YAW_RIGHT) / timeFactor);
+                myAvatar->setDriveKeys(PITCH, -1.0f * userInputMapper->getActionState(controller::Action::PITCH) / timeFactor);
```
ZappoMan: same here? why are these reversed axes?

interface/src/avatar/MyAvatar.h
```diff
@@ -21,6 +21,20 @@
 
 class ModelItemID;
 
+enum DriveKeys {
+    TRANSLATE_X = 0,
+    TRANSLATE_Y,
+    TRANSLATE_Z,
+    YAW,
+    STEP_TRANSLATE_X,
```
ZappoMan: why add STEP_TRANSLATE_* but not add STEP_PITCH

libraries/controllers/src/controllers/Actions.h
```diff
@@ -27,6 +27,16 @@ enum class Action {
     ROTATE_Y, YAW = ROTATE_Y,
     ROTATE_Z, ROLL = ROTATE_Z,
 
+    STEP_YAW,
+    // FIXME does this have a use case?
```
ZappoMan: I can imagine us implementing STEP for these

interface/src/Application.cpp
jherico: because the input mapper follows the OpenGL convention of -Z being 'forward', but the drive keys treat negative values as backward (if you look they multiply the drive key by `vec3(0, 0, -1)`)

interface/src/Application.cpp
jherico: again, different opinions on what yaw pitch and roll mean when looking down the positive axis.  I'd like for user input mapper to be consistent with either clockwise or anti-clockwise, though I'm not certain it is.

interface/src/avatar/MyAvatar.h
jherico: the only use case right now for stepping is if you're in the HMD, in which case pitch and roll are directly controlled by the HMD.  I suppose we could add the possibility of step pitch, but I'd argue that step pitch and step roll should be applied to the avatar, not the head.

PR #6150, opened 2015-10-22
final fixes required for successful Qt 5.5.1 build

birarda:


libraries/render-utils/src/GlWindow.cpp
```diff
@@ -43,11 +43,9 @@ bool GlWindow::makeCurrent() {
         qDebug() << "GL Renderer: " << QString((const char*) glGetString(GL_RENDERER));
     });
     
-    #ifndef QT_NO_DEBUG
-    QOpenGLContext * currentContext =
-    #endif
-        QOpenGLContext::currentContext();
+    QOpenGLContext * currentContext = QOpenGLContext::currentContext();
     Q_ASSERT(_context == currentContext);
+
```
ZappoMan: this is probably going to generate a warning in release builds... might want to add a Q_UNUSED(currentContext) to suppress that.

libraries/ui/src/VrMenu.cpp
```diff
@@ -108,13 +108,10 @@ void VrMenu::addMenu(QMenu* menu) {
...
     }
     QVariant returnedValue;
-    #ifndef QT_NO_DEBUG
-    bool invokeResult =
-    #endif
-    QMetaObject::invokeMethod(this, "addMenu", Qt::DirectConnection,
-        Q_RETURN_ARG(QVariant, returnedValue),
-        Q_ARG(QVariant, QVariant::fromValue(qmlParent)),
-        Q_ARG(QVariant, QVariant::fromValue(menu->title())));
+    bool invokeResult = QMetaObject::invokeMethod(this, "addMenu", Qt::DirectConnection,
```
ZappoMan: this is probably going to generate a warning in release builds... might want to add a Q_UNUSED(invokeResult) to suppress that.

birarda:
This has changed for Qt 5.5 - the Q_ASSERT macro now expands the condition when asserts are disabled (without actually checking the condition). This means we should not get the unused warnings.

hifi-gustavo:
Build console output for commit: a8d591166e7b6fbacf6cb928ad165e8f01a92f3e available at: https://gist.github.com/119077a62f3280693dc5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6150.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6150.dmg

ZappoMan:
@birarda - cool - that's handy

ZappoMan:
hmmm - actually that might be bad... that means that code people previously thought would only run in a debug build, will now run in a release build... if the condition of the assert is an expensive function, we will take a performance hit... We need to audit our use of Q_ASSERT() if this is the case.

birarda:
No, it does not run - look at the Q_ASSERT macro. From the Qt 5.5 release notes:

*Q_ASSERT will now expand the condition even in release mode when asserts are disabled, albeit in an unreachable code path. This solves compiler warnings about variables and functions that were unused in release mode because they were only used in assertions. Unfortunately, codebases that hid those functions and variables via #ifndef will need to remove the conditionals to compile with Qt 5.5.*

jherico:
sounds like an ideal solution as long as the compiler can't detect that the code path is unreachable.  

When can I update to 5.5.1 locally?

birarda:
@jherico with this merged you should be good to go

PR #6152, opened 2015-10-22
Parents

sethalves:
- allow avatars and entities to have locations which are relative to something else


jherico:
YES!

sethalves:
This doesn't work yet, but I think it will soon.  I'm sort of extracting the relevant parts from another branch that did somewhat work, but also had local physics stuff mixed in.

hifi-gustavo:
Build console output for commit: 778dc8dc15ddb1831d33c05350fbbc07c0bf62b9 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: f6fe503c515d636bf4ddc0f66b58c111576a6e03 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: d86d69aeba4b72f69d2d0d3f626801bfa4ee0fb6 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 7b08d047b19598784cfa03c67d67329a6351724a available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

libraries/shared/src/SpatiallyNestable.cpp
```diff
@@ -0,0 +1,169 @@
...
+}
+
+void SpatiallyNestable::setTransform(const Transform& transform) {
+    Transform parentTransform = getParentTransform();
+    Transform::inverseMult(_transform, parentTransform, transform);
+}
+
+const glm::vec3& SpatiallyNestable::getScale() const {
+    return _transform.getScale();
+}
```
howard-stearns: This and setScale() appear to be the same as get/setLocalScale, below.
I'm not sure what global/world scale would mean. However, "dimensions" or "extents" (in global space) might be a perfectly reasonable concept, as is axisAlignedDimensions or axisAlignedExtents.

libraries/shared/src/SpatiallyNestable.cpp
```diff
@@ -0,0 +1,169 @@
...
+    return _transform;
+}
+
+void SpatiallyNestable::setLocalTransform(const Transform& transform) {
+}
+
+const glm::vec3& SpatiallyNestable::getLocalPosition() const {
+    return _transform.getTranslation();
+}
+
```
howard-stearns: Both our avatar coordinate system definitions, and Sam's Transform class, avoid using "position" and "orientation" for local quantities, but instead use "translation" and "rotation" (or localTranslation/localRotation, if you prefer).

libraries/render-utils/src/Model.cpp
```diff
@@ -667,17 +685,17 @@ Extents Model::calculateScaledOffsetExtents(const Extents& extents) const {
...
+                                  rotatedExtents.maximum + modelPosition };
 
     return translatedExtents;
 }
 
 /// Returns the world space equivalent of some box in model space.
-AABox Model::calculateScaledOffsetAABox(const AABox& box) const {
-    return AABox(calculateScaledOffsetExtents(Extents(box)));
+AABox Model::calculateScaledOffsetAABox(const AABox& box, glm::vec3 modelPosition, glm::quat modelOrientation) const {
+    return AABox(calculateScaledOffsetExtents(Extents(box), modelPosition, modelOrientation));
```
howard-stearns: It's hard for me to tell what is model space and what is world space (throughout this file, this method is just an example). That was true before this pr, but it's such a good opportunity to get some consistent naming and/or code comments...

libraries/shared/src/SpatiallyNestable.cpp
sethalves: Hmm.  I have been deciding like this:  If it's a translation from an origin, it's a position.  If it's a translation from an arbitrary point, it's a translation.  These calls deal with a translation from the origin of the parent's frame.

libraries/shared/src/SpatiallyNestable.cpp
sethalves: Yeah... I didn't actually have to bring scale down into this class, but I think it would have ended up here eventually.  I think we need a new property.  Scale is tangled up with an entity's dimensions.  If I don't set scale to one in a parent's transform, the *size* of the parent entity affects the offset of the child, which isn't what we want.  I think we may also need a scale property that does affect the children, so that we can have miniature scenes.

hifi-gustavo:
Build console output for commit: 7f6ac0207da8aea918360190753d7656d8de04dd available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: bf290c35be4ce5ff2c188d601785ad26ee27f24e available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 61269c3ce837b30cfcd389849832045a9d55e05f available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: f3c61a823e1064d7bfec23f6f56a3605875fb617 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 27c44aa22cb36107545e642e8c358c5e3eb9f491 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

thoys:
What do you guys think, will this feature allow us to attach the camera to entities with more ease?

hifi-gustavo:
Build console output for commit: 74ade02bc631629465637a123f5151a28b3a48da available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hyperlogic:
You deleted myAvatar->startCapture(); and end capture. This is kind of fudamental to how avatar threading works. Are we getting rid of it? If so we should remove the menu item.

sethalves:
I deleted it because I pulled out the _nextAllowed _nextPosition stuff, which is all it did.  I pulled that out because it didn't fully solve the other-avatar jitter problem (thought it does help it quite a bit).  I've tried to find a better solution to that problem, but I haven't yet succeeded.

libraries/shared/src/SpatiallyNestable.cpp
jherico: Why wouldn't we want scale to be propagated?  If I want to turn a house into a dollhouse by scaling the top level item, wouldn't I need that?

sethalves: Yeah, but I think it needs to be a new property.  The current "scale" is derived from the (parent) entity's dimensions.  If I don't squash the scale in the parent's trasform, then the child gets transitioned by an unexpected amount -- meaning that if something is a child of a 1x1x1 box, it will be in one place, but will move when the box is scaled.  I'm not sure what we would call the new property.  childScale?  childrenScaler?

hifi-gustavo:
Build console output for commit: 3ac0fef1ea1479b3a5db57c1a2e8e41fe6ae7ad1 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: dba9eb70466d19b15bec5afdf2f74a50ef5e08c9 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: f60b1d9bd7c7f64fc9a23ca958667906c0d7d47a available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 60824a1fb013dc3a2eb839aef5272d8456a7b670 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: ffb0317a5044cfa218847ae13a4d76478dfd3fe5 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 9b3708ea66a9ebd2cdf6e910039eb4b443776688 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: f2be71d1a500d9fb513d849a9f6482e0845ffa26 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 4ab727c47c046477fadb0dcab36e06340938c8d8 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 4ffaa7db8e573ee941b4b60b9ec0217151eea852 available at: https://gist.github.com/8b9c201d86f94ca5149b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 4646c0c1030b8c67ca4768d443fb39097a446067 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: a9669531ae324c90d34bd1caa3c9c931dccca893 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: abfc3e994a534ac82235be74cc95611faa8f663d available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: a0c1f9a1f9d8944fd33bb0f8f6d736cf0acd2666 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: feec3fdcd3c684420e3d7fb0331e12de0be47d90 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 171f6da264500be247aa24a2f1e0a51fc2fba372 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: ba30e016644a61c2d2a4a2b3697011303511bcf4 available at: https://gist.github.com/8b9c201d86f94ca5149b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

sethalves:
This isn't done, but I think it's in a stable place.  There are some remaining problems to solve, esp around bounding boxes used in octree-server queries and what "scale" means.

hifi-gustavo:
Build console output for commit: 5fe41662d802ef852852a5db0d5c37cfefa0a837 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: ca75802b9c2c058c4175097d4f106ed26aa2a8ab available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 10cf85bad975090e01cb95da7edbf2b815af7c0f available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 6062691c2a58f42486a1e28d975818582eb8e144 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: facf91faacf7cf6f9e5ec85ed2b91f4b7fcd8191 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 204c3d839e487775542ffa3adf1d1b4aa6fb9cf9 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: cf39cac7fcfb24b38a47228ad8cc46d280d68970 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: bbc9d5d1317ad655aa9ff5c3209770537ad9d62a available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 9548b0e138e0b9ec840c3fa4f8268655ff028420 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

libraries/shared/src/SpatialParentFinder.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <QUuid>
+
+#include "DependencyManager.h"
+
+class SpatiallyNestable;
+using SpatiallyNestableWeakPointer = std::weak_ptr<SpatiallyNestable>;
+using SpatiallyNestablePointer = std::shared_ptr<SpatiallyNestable>;
+class SpatialParentFinder;
+using SpatialParentFinderPointer = std::shared_ptr<SpatialParentFinder>;
```
AndrewMeadows: **SpatialParentFinderPointer** is unused.

We're using **std::shared_ptr** for the **SpatiallyNestable**'s while using **QSharedPointer** for the **ParentFinder** stuff.  Mixing these two smart pointer implementations so close to each other seems dangerous, but as long as we don't actually cross the beams we should be fine.

imgntn:
@sethalves this is the script i've been using to test this PR -- if you look through it you'll see a couple of easily reproducible crash conditions (mostly when unsetting the parentId or setting it to something strange) and some lag when changing a child's position dynamically.  let me know if i can clarify / help fix / test anything else.  thanks!!

https://rawgit.com/imgntn/hifi/parentsExamples/examples/entityScripts/parentChildRelationship.js

interface/src/InterfaceParentFinder.h
```diff
@@ -0,0 +1,27 @@
...
+#include <SpatialParentFinder.h>
+
+class InterfaceParentFinder : public SpatialParentFinder {
+public:
+    InterfaceParentFinder() { }
+    virtual ~InterfaceParentFinder() { }
+    virtual SpatiallyNestableWeakPointer find(QUuid parentID) const;
+};
+
+#endif // hifi_InterfaceParentFinder_h
```
AndrewMeadows: I was trying to understand "Why do we need the **ParentFinder** stuff?  It seems... complicated."

I guess the reason is that although **SpatiallyNestable** is a base class for most "things" in the virtual world, we don't have a single generic container for those things inside the **shared** lib.  Instead we have some unknown container API (unknown at the **shared** lib context) for each type derived from **SpatiallyNestable** --  only **EntityItem** and **AvatarData** so far, but we'll probably realize later that others should also derive from SpatiallyNestable, such as: ParticleSystem, and Attachments.  The containers (i.e. EntityTree and AvatarManager) of the derived classes live in higher level projects (**entities** lib , **interface**) that depend on **shared**... so what we have is the **shared** API performing minor convolutions to accommodate existing API's in dependent libs.

This work feels like **technical debt** to me but I'm not sure what it would look like with the deb all paid.  I guess we could ask: "If we knew from the very start of the project that 'things' would exist in a spatial hierarchy relative to other 'things'... how would we have implemented things differently?"  The debt will be fully paid when our implementation looks very much like whatever we would have written with great foresight.

This isn't necessarily a blocker for this PR to move forward, but we should be aware of it.  Technical debt piles up and eventually we'll have to pay it... or get stuck.

sethalves: The short answer is: "Because we store entities and avatars in different data structures, and we need to look them up by their ID without knowing which type we'll get back".

libraries/shared/src/SpatiallyNestable.h
```diff
@@ -0,0 +1,121 @@
...
+class SpatiallyNestable : public std::enable_shared_from_this<SpatiallyNestable> {
+public:
+    SpatiallyNestable(NestableTypes::NestableType nestableType, QUuid id);
+    virtual ~SpatiallyNestable() { }
+
+    virtual const QUuid& getID() const { return _id; }
+    virtual void setID(const QUuid& id) { _id = id; }
+
+    virtual const QUuid getParentID() const { return _parentID; }
+    virtual void setParentID(const QUuid parentID);
```
AndrewMeadows: It looks like these *etParentID() methods should use **const QUuid&** rather than **const QUuid**.  Was it intentional to leave out the reference designator?

libraries/shared/src/SpatiallyNestable.h
```diff
@@ -0,0 +1,121 @@
...
+
+    virtual quint16 getParentJointIndex() const { return _parentJointIndex; }
+    virtual void setParentJointIndex(quint16 parentJointIndex) { _parentJointIndex = parentJointIndex; }
+
+    glm::vec3 worldToLocal(const glm::vec3 position);
+    glm::quat worldToLocal(const glm::quat orientation);
```
AndrewMeadows: The input arguments should probably be passed by const reference rather than const  copy.

sethalves: Yeah -- I left the reference for ID because that never changes.  For the members that do change, I've made them all return-by-copy rather than reference.  ParentID can change.

sethalves: I got rid of the pass-by-reference parts of the interface in order to make the code thread-safe.

sethalves: As above -- I made them pass-by-value on purpose.  I think the pass-by-reference stuff may be premature-optimization and it certainly makes threading more treacherous.

sethalves: though -- I guess I should have removed the const.

libraries/shared/src/SpatiallyNestable.h
```diff
@@ -0,0 +1,121 @@
...
+    virtual void setLocalTransform(const Transform transform);
+
+    virtual const glm::vec3 getLocalPosition() const;
+    virtual void setLocalPosition(const glm::vec3 position);
+
+    virtual const glm::quat getLocalOrientation() const;
+    virtual void setLocalOrientation(const glm::quat orientation);
+
+    virtual const glm::vec3 getLocalScale() const;
+    virtual void setLocalScale(const glm::vec3 scale);
```
AndrewMeadows: Return values or input arguments that have complex types and are declared **const** should either be by reference or be non-const (if copies really should be made) it almost never makes sense to return const copies.

libraries/shared/src/SpatiallyNestable.h
```diff
@@ -0,0 +1,121 @@
...
+    // this object's frame
+    virtual const Transform getJointTransformInObjectFrame(int jointIndex) const;
+    virtual glm::quat getJointRotation(int index) const = 0;
+    virtual glm::vec3 getJointTranslation(int index) const = 0;
+
+protected:
+    NestableTypes::NestableType _nestableType; // EntityItem or an AvatarData
+    QUuid _id;
+    QUuid _parentID; // what is this thing's transform relative to?
+    quint16 _parentJointIndex; // which joint of the parent is this relative to?
```
AndrewMeadows: **_parentJointIndex** is never initialized here (default ctor initialization) or explicitly in the ctor.  It is only ever set explicitly via **setParentJointIndex()**.  It is not clear to me that **setParentJointIndex()** is guaranteed to be called for all **SpatiallyNestable** instances.  Is it?

sethalves: You're correct -- I'll fix it.

hifi-gustavo:
Build console output for commit: 4b87e0984806e7d2cb1eb1d4346defd4ddfe2c16 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 941dfe5bd2dcb09f4a32878f0743432acc74be01 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

interface/src/Application.h
```diff
@@ -185,7 +185,7 @@ class Application : public QApplication, public AbstractViewStateInterface, publ
     virtual float getSizeScale() const;
     virtual int getBoundaryLevelAdjust() const;
     virtual PickRay computePickRay(float x, float y) const;
-    virtual const glm::vec3& getAvatarPosition() const;
+    virtual const glm::vec3 getAvatarPosition() const;
```
AndrewMeadows: Return value should not be const.

hifi-gustavo:
Build console output for commit: 89b78986d5b219bec205348840aa6e43f6306645 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 5c327edd5a8e1825e19a68a7cd5f81b3ca1cd0d4 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: aa383455f5967dfdcdf542fcf2fa29f4111f10ad available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 02c4730388fbcd5d317dce12366c76f8078afe26 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

libraries/shared/src/SpatiallyNestable.cpp
```diff
@@ -0,0 +1,340 @@
...
+    if (parent) {
+        parentTransform = parent->getTransform(parentJointIndex);
+        parentTransform.setScale(1.0f);
+    }
+    Transform orientationTransform;
+    orientationTransform.setRotation(orientation);
+    Transform result;
+    Transform::mult(result, parentTransform, orientationTransform);
+    return result.getRotation();
+}
```
AndrewMeadows: **worldToLocal()** and **localToWorld()** appear to be misnamed.  What they actually provide are **localToChild()** and **localToParent()** respectively.  They will work fine as long as the hierarchy is never deeper than two (one parent with children).

libraries/shared/src/SpatiallyNestable.cpp
```diff
@@ -0,0 +1,340 @@
...
+    Transform::mult(jointInWorldFrame, worldTransform, jointInObjectFrame);
+    return jointInWorldFrame;
+}
+
+void SpatiallyNestable::setTransform(const Transform transform) {
+    Transform parentTransform = getParentTransform();
+    _transformLock.withWriteLock([&] {
+        Transform::inverseMult(_transform, parentTransform, transform);
+    });
+}
```
AndrewMeadows: Similarly, **set/getPosition/Orientation/Transform()** are misnamed.  What they actually do is:  **set/getPosition/Orientation/TransformInParentFrame()**.

AndrewMeadows: I take that back.

getParentTransform() calls getTransform() calls getParentTransform() etc... it is indeed recursive.

libraries/shared/src/SpatiallyNestable.cpp
```diff
@@ -0,0 +1,340 @@
...
+    _childrenLock.withWriteLock([&] {
+        _children[newChild->getID()] = newChild;
+    });
+}
+
+void SpatiallyNestable::forgetChild(SpatiallyNestablePointer newChild) const {
+    _childrenLock.withWriteLock([&] {
+        _children.remove(newChild->getID());
+    });
+}
```
AndrewMeadows: There is no support yet for physics of joined objects: each parent will have its own btRigidBody with MotionState as will each child.  This is only a blocker if anyone intends to use dynamic or kinematic parents/children.  Static objects should work ok.

It occurs to me that when communicating with the physics engine we're going to need the notion of an "Object" (or something) where an Object has: one heirarchy of SpatiallyNestable's and one RigidBody.

nbq:
Just ran a test on this using the HiFi Server and the pr interface and with the Hydras, the left hand cannot grab or be passed items. I tested this and was able to reproduce this between the release interface and the pr interface.

sethalves:
@nbq -- I can't reproduce the left-hand thing.  This is with handControllerGrab.js ?  Anything unusual about the test item?

nbq:
@sethalves just updated my comment:

After restart both hands work but after a bit of grabbing a moving using the Hydras, one hand will stop working on grabs and with show IK constraints, you can see it no longer emits a red/blue line.

hifi-gustavo:
Build console output for commit: 9b54924524783b3731c2a09f562705902f6ad2a5 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

libraries/avatars/src/AvatarData.cpp
```diff
@@ -515,11 +476,12 @@ int AvatarData::parseDataFromBuffer(const QByteArray& buffer) {
...
+
+        // TODO is this safe? will the floats not exactly match?
+        if (getBodyYaw() != yaw || getBodyPitch() != pitch || getBodyRoll() != roll) {
             _hasNewJointRotations = true;
-            _bodyYaw = yaw;
-            _bodyPitch = pitch;
-            _bodyRoll = roll;
+            glm::vec3 eulerAngles(pitch, yaw, roll);
+            setLocalOrientation(glm::quat(glm::radians(eulerAngles)));
         }
```
AndrewMeadows: Yes, there is a possibility that the transmitted <pitch, yaw, roll> will not quite match the extracted <pitch, yaw, roll> despite being originally extracted from the exact same quaternion.  I followed the code through and it appears the risk is that the avatar's SkeletonModel might fall into the CPU expensive part of Model::updateClusterMatrices() when otherwise it would not have required it.

However, we know we can update many simultaneously animating avatars, and most avatars will be moving constantly anyway, so I don't think we need to worry.

sethalves:
@nbq it sounds like the grab script is crashing.  do you see anything in the interface log?

libraries/avatars/src/AvatarData.cpp
```diff
@@ -90,49 +87,16 @@ const QUrl& AvatarData::defaultFullAvatarModelUrl() {
...
-    }
-}
-
 // There are a number of possible strategies for this set of tools through endRender, below.
 void AvatarData::nextAttitude(glm::vec3 position, glm::quat orientation) {
     avatarLock.lock();
-    setPosition(position, true);
-    setOrientation(orientation, true);
+    SpatiallyNestable::setPosition(position);
+    SpatiallyNestable::setOrientation(orientation);
```
AndrewMeadows: Would be much cheaper to use SpatiallyNestable::setTransform() here rather than calling position and orientation individually.

nbq:
@sethalves I have reproduced it, its with the "hifi-bow" from toybox. Once you grab it and let go, one of your hands no longer works.

imgntn:
@nbq nice find.  we changed the bow API for doing hand stuff over the weekend and the name of the message didn't get updated on release

hifi-gustavo:
Build console output for commit: 3402585d1ac2b053c4708025765afe9d40c5eb26 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: b7ebe5b92bb2ffa07696d0cbd11aec2fc106f3ed available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

nbq:
Between using the original HFS compile, the newer clients give messages about the server version being incompatible with the client. I have had to recompile the AC/DS multiple times for different clients. I am doing another PR pull and compile now for the newest client

hifi-gustavo:
Build console output for commit: b8d800f5b8257f4e8319bba2a905dafcd240b8f6 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 0595ef09384909b41918aed0f7dbaebf2f213631 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: f18fc84c2df8140cbe5afbe10437b4563549456d available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

imgntn:
@sethalves have been testing this running the following scripts and see no difference in fps btw this PR and master: (37fps)

-playaSpawner.js
-cubePerfTest.js
-particleDance.js
-createPingPongGun.js (and shooting ping pong balls)
-flockingBirds.js

imgntn:
since performance seems the same, and i'm not able to explicitly crash anymore with my other parenting example script, i think this is a good candidate for merging soon -- v.1 entity parenting, v.2 avatar joint parenting or w/e.  maybe not today though because of tonight's demo

hifi-gustavo:
Build console output for commit: 6eae98fb288ae940d14beb5282135fa0e05c98fd available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: 4e74c98ae787af64e97d5c66eb59559a9a38329a available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

hifi-gustavo:
Build console output for commit: e5b1cfa920d5a09b2e2b1375a0f797f4d21cf9a9 available at: https://gist.github.com/6c25512b8ba2d0ea7c30<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6152.dmg

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -26,13 +26,14 @@
...
 }
-RenderableModelEntityItem::RenderableModelEntityItem(const EntityItemID& entityItemID,
-                                                     const EntityItemProperties& properties) :
-    ModelEntityItem(entityItemID, properties),
-    _dimensionsInitialized(properties.getDimensionsInitialized())
-{
+
+RenderableModelEntityItem::RenderableModelEntityItem(const EntityItemID& entityItemID, bool dimensionsInitialized) :
+    ModelEntityItem(entityItemID),
+    _dimensionsInitialized(dimensionsInitialized) {
```
ZappoMan: not merge blocker, but technically the coding standard says to place this open brace `{` on the next line.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -43,7 +44,7 @@ RenderableModelEntityItem::~RenderableModelEntityItem() {
     }
 }
 
-void RenderableModelEntityItem::setDimensions(const glm::vec3& value) {
+void RenderableModelEntityItem::setDimensions(const glm::vec3 value) {
```
ZappoMan: not a merge blocker, but it seems odd to have removed the reference on this. there doesn't appear to be any thread safety/value semantics issue here, since you end up assigning it a couple lines later... would be better to keep the reference.

libraries/entities-renderer/src/RenderableModelEntityItem.h
```diff
@@ -24,11 +24,11 @@ class RenderableModelEntityItem : public ModelEntityItem {
 public:
     static EntityItemPointer factory(const EntityItemID& entityID, const EntityItemProperties& properties);
 
-    RenderableModelEntityItem(const EntityItemID& entityItemID, const EntityItemProperties& properties);
+    RenderableModelEntityItem(const EntityItemID& entityItemID, bool dimensionsInitialized);
```
ZappoMan: not a merge blocker, but it's unclear to me why you changed the constructor signature to not take properties. Essentially the code ends up doing the same thing, but in this one case you have the precise problem of needing to expose the fact that for RenderableModelEntityItems they need the details about the dimensions state to properly construct themselves... but removing the properties as a parameter to constructor, you've exposed this private detail outside of the class.

libraries/entities/src/EntityItem.h
```diff
@@ -337,6 +306,10 @@ class EntityItem : public std::enable_shared_from_this<EntityItem>, public ReadW
     /// return preferred shape type (actual physical shape may differ)
     virtual ShapeType getShapeType() const { return SHAPE_TYPE_NONE; }
 
+    // these are only needed because the names don't match
+    virtual const glm::quat getRotation() const { return getOrientation(); }
+    virtual void setRotation(glm::quat orientation) { setOrientation(orientation); }
```
ZappoMan: not a merge blocker: but I don't see why these are `const glm::quat&` that would allow better performance and is not a thread safety or value semantics issue

libraries/shared/src/SpatiallyNestable.h
```diff
@@ -0,0 +1,128 @@
...
+    virtual const QUuid& getID() const { return _id; }
+    virtual void setID(const QUuid& id) { _id = id; }
+
+    virtual const QUuid getParentID() const { return _parentID; }
+    virtual void setParentID(const QUuid parentID);
+
+    virtual quint16 getParentJointIndex() const { return _parentJointIndex; }
+    virtual void setParentJointIndex(quint16 parentJointIndex) { _parentJointIndex = parentJointIndex; }
+
+    static glm::vec3 worldToLocal(glm::vec3 position, QUuid parentID, int parentJointIndex);
```
ZappoMan: not a merge blocker: but I don't see why these are not `const glm::vec3&`  and `const glm::quat&` that would allow better performance and is not a thread safety or value semantics issue

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -64,8 +64,56 @@ void EntityScriptingInterface::setEntityTree(EntityTreePointer elementTree) {
     }
 }
 
+EntityItemProperties convertLocationToScriptSemantics(EntityItemProperties entitySideProperties) {
```
ZappoMan: not a merge blocker, but this could definitely be a `const EntityItemProperties&` to improve performance

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -64,8 +64,56 @@ void EntityScriptingInterface::setEntityTree(EntityTreePointer elementTree) {
...
+                                                              entitySideProperties.getParentID(),
+                                                              entitySideProperties.getParentJointIndex());
+    scriptSideProperties.setPosition(worldPosition);
+    scriptSideProperties.setRotation(worldRotation);
+
+    return scriptSideProperties;
+}
+
+
+EntityItemProperties convertLocationFromScriptSemantics(EntityItemProperties scriptSideProperties) {
```
ZappoMan: not a merge blocker, but this could definitely be a `const EntityItemProperties&` to improve performance

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -130,10 +187,11 @@ EntityItemProperties EntityScriptingInterface::getEntityProperties(QUuid identit
         });
     }
 
-    return results;
+    return convertLocationToScriptSemantics(results);
 }
 
-QUuid EntityScriptingInterface::editEntity(QUuid id, EntityItemProperties properties) {
+QUuid EntityScriptingInterface::editEntity(QUuid id, EntityItemProperties scriptSideProperties) {
+    EntityItemProperties properties = scriptSideProperties;
```
ZappoMan: not a merge blocker, why make a copy here?

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -143,10 +201,23 @@ QUuid EntityScriptingInterface::editEntity(QUuid id, EntityItemProperties proper
 
     bool updatedEntity = false;
     _entityTree->withWriteLock([&] {
+        if (scriptSideProperties.parentDependentPropertyChanged()) {
```
ZappoMan: not a merge blocker, I don't see why you introduced this extra copy of the properties with the name `scriptSideProperties` -- you end up making copies of it, and using the original and the copy as if they were the same object, which you could have done, since the original was a value copied into your local scope to begin with.

libraries/entities/src/EntityTree.cpp
```diff
@@ -140,8 +140,11 @@ bool EntityTree::updateEntityWithElement(EntityItemPointer entity, const EntityI
             if (!wantsLocked) {
                 EntityItemProperties tempProperties;
                 tempProperties.setLocked(wantsLocked);
-                UpdateEntityOperator theOperator(getThisPointer(), containingElement, entity, tempProperties);
+
+                BoundingBoxRelatedProperties newBBRelProperties(entity, tempProperties);
```
ZappoMan: not a merge blocker, but I think this class could have been named `BoundingBoxProperties` and it would be equally clear and more convenient to read and write

libraries/entities/src/EntityTree.cpp
```diff
@@ -206,8 +209,34 @@ bool EntityTree::updateEntityWithElement(EntityItemPointer entity, const EntityI
...
+        foreach (SpatiallyNestablePointer child, entity->getChildren()) {
+            if (child && child->getNestableType() == NestableTypes::Entity) {
+                toProcess.enqueue(child);
+            }
+        }
+
+        while (!toProcess.empty()) {
+            EntityItemPointer childEntity = std::static_pointer_cast<EntityItem>(toProcess.dequeue());
+            BoundingBoxRelatedProperties newChildBBRelProperties(childEntity);
+            UpdateEntityOperator theChildOperator(getThisPointer(),
```
ZappoMan: not a merge blocker... but -- we will want to make a version of the `UpdateEntityOperator` that handles updating multiple entities in a single pass... this will improve performance significantly.

libraries/entities/src/UpdateEntityOperator.cpp
```diff
@@ -44,83 +44,32 @@ UpdateEntityOperator::UpdateEntityOperator(EntityTreePointer tree,
     _oldEntityCube = _existingEntity->getMaximumAACube();
     _oldEntityBox = _oldEntityCube.clamp((float)-HALF_TREE_SCALE, (float)HALF_TREE_SCALE); // clamp to domain bounds
 
-    // If the old properties doesn't contain the properties required to calculate a bounding box,
```
ZappoMan: ok - these changes make me a little nervous... you've tested the hell out of this???

libraries/entities/src/EntityTree.cpp
```diff
@@ -140,8 +140,11 @@ bool EntityTree::updateEntityWithElement(EntityItemPointer entity, const EntityI
             if (!wantsLocked) {
                 EntityItemProperties tempProperties;
                 tempProperties.setLocked(wantsLocked);
-                UpdateEntityOperator theOperator(getThisPointer(), containingElement, entity, tempProperties);
+
+                BoundingBoxRelatedProperties newBBRelProperties(entity, tempProperties);
+                UpdateEntityOperator theOperator(getThisPointer(), containingElement, entity, newBBRelProperties);
                 recurseTreeWithOperator(&theOperator);
+                entity->setProperties(tempProperties);
```
ZappoMan: I'm not really a fan about you moving the call to `setProperties()` outside of the `UpdateEntityOperator`... why did you do this? The point of the `UEO` was that it fully updated the tree and the entity.

libraries/entities/src/UpdateEntityOperator.cpp
```diff
@@ -44,83 +44,32 @@ UpdateEntityOperator::UpdateEntityOperator(EntityTreePointer tree,
...
 
         if (_wantDebug) {
-            qCDebug(entities) << "    **** TYPICAL NO MOVE CASE ****";
-            qCDebug(entities) << "        _properties.containsBoundsProperties():" << _properties.containsBoundsProperties();
-            qCDebug(entities) << "                             oldElementBestFit:" << oldElementBestFit;
-        }
-
-    } else {
-        _newEntityCube = _properties.getMaximumAACube();
-        _removeOld = true; // our properties are going to move us, so remember this for later processing
```
ZappoMan: this is interesting... did you find a bug related to this change? I am looking at the code you're deleting and **the old code** looks wrong... but unless you are removing it intentionally it seems wrong to be removing it... what's the story with this change?

libraries/render-utils/src/Model.cpp
ZappoMan: Welcome to the wonders of `Model.cpp`

model-cpp

libraries/entities/src/UpdateEntityOperator.cpp
sethalves: Yeah I've run this code for quite a while and I haven't noticed a problem.  I wanted a way to reuse the relevant parts of the properties on all the children of the updated entity.

libraries/shared/src/SpatiallyNestable.cpp
```diff
@@ -0,0 +1,371 @@
...
+
+glm::vec3 SpatiallyNestable::getPosition() const {
+    return getTransform().getTranslation();
+}
+
+glm::vec3 SpatiallyNestable::getPosition(int jointIndex) const {
+    return getTransform(jointIndex).getTranslation();
+}
+
+void SpatiallyNestable::setPosition(glm::vec3 position) {
```
ZappoMan: not a merge blocker: but like I've said on other lines, these input parameters might as well be `const type&` there's no benefit in copying by value here

libraries/shared/src/SpatiallyNestable.h
ZappoMan: I'm with Andrew on this one... you might as well pass by reference, you don't need value semantics here.

plugins/oculusLegacy/src/OculusLegacyDisplayPlugin.cpp
```diff
@@ -167,8 +167,11 @@ void OculusLegacyDisplayPlugin::activate() {
         }
     });
 
-    ovrBool result = ovrHmd_ConfigureRendering(_hmd, &config.Config, distortionCaps, _eyeFovs, _eyeRenderDescs);
-    Q_ASSERT(result);
+    #ifndef NDEBUG
+    ovrBool result =
+    #endif
+        ovrHmd_ConfigureRendering(_hmd, &config.Config, distortionCaps, _eyeFovs, _eyeRenderDescs);
+    assert(result);
```
ZappoMan: can't we use `Q_ASSERT` and be done with it?

sethalves: It was generating a warning for non-debug builds.

libraries/shared/src/SpatiallyNestable.cpp
sethalves: Okay, I agree that the setters might as well be pass-by-reference.  I'll fix this in a future PR.

libraries/entities/src/EntityTree.cpp
sethalves: As mentioned on slack -- I needed to be able to call UpdateEntityOperator on the children without calling setProperties on them.

libraries/entities-renderer/src/RenderableModelEntityItem.h
sethalves: This was done because something inside the properties code ends up calling shared_from_this, which doesn't work until after the constructor has completed.

ZappoMan: C++ can go pound sand!

PR #6154, opened 2015-10-22
Safe avatar list access

howard-stearns:
Most of the access to AvatarHashMap's _avatarHash member is done on the main thread, but the instance is exposed to scripts. Discussion for an R&D PR (#6097) shows that scripted access to _avatarHash is not thread safe.

Currently the only scriptable access I can find is isAvatarInRange(). This PR makes it safe, independently of any other more complicated work, laying the groundwork for any future access. See line 55 of AvatarHashMap.h.


hifi-gustavo:
Build console output for commit: 624ed7c71128cd413f577ca26963c648f7eb9fd8 available at: https://gist.github.com/30630d2b1501fa2e8a35<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6154.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6154.dmg

hyperlogic:
Looks good to me. :+1: 

One observation though, I noticed that the locking is slightly inconsistent.  In that we only take out a read lock for methods that are called by the script thread.  But when we read the _avatarHash in the main thread, we don't.   Which, makes perfect sense from a performance point of view, because we know that information goes only one way, from main thread to script threads.  However, if we we add script methods that mutate the _avatarHash, we will have to add read locks in the main thread.

howard-stearns:
Quite right. The only thing preventing someone from adding a from-scripts mutator is the comment on line 56 of AvatarHashMap.h. Both before and after this PR, there is no compiler-enforced prevention. This PR is only for making it possible to be read-safe from scripts (in the manner of isAvatarInRange).

ZappoMan:
I don't understand why we would only make it partly thread safe. AFAIR - the avatar hash is written to by the NodeList thread not the main thread. And so we do want the main thread reads to be protected too.

howard-stearns:
My initial guess was also that stuff would happen on the NodeList thread. But then we would be crashing all the time. It turns out that Birarda has arranged for packet messages to be registered to be queued on various objects in their respective threads. (See the packetReceiver.registerListener code in the AvatarManager constructor.) Thus, e.g., processAvatarDataPacket and processAvatarIdentityPacket are happening on the same thread as all the other non-script-driven AvatarManger/AvatarHashMap activity (which happens to be the main thread). That's why we're not constantly crashing from this.

Or are your referring to some other activity that I missed, that is not going through this mechanism and is thus causing crashes?

So, as far as I can tell, there's still the option of sprinkling read locks throughout the code "just in case", but I don't think that NodeList activity is reason that we _must_ do so to avoid an existing problem.

So, is there a specific existing problem that I'm not seeing? (It's ok to have a discussion about safety just-in-case, but I want to make sure I understand which discussion we're having.)

ZappoMan:
ahh... so that's likely a recent change. I guess my argument for thread safety is... we might change it in the future and in general this should be a relatively inexpensive safety measure.

howard-stearns:
OK, so on the "just in case" front:
My conclusion: whatever the consensus is.
My opinion: don't do the "just in case" locks at this time.

That opinion is based on the idea that adding bulk to the code (with an admittedly slight CPU performance hit, and an admittedly slight Cognitive Developer Load hit) is worth it IF there's a clear benefit, but I am (personally) not yet convinced that sprinkling locks everywhere provides the safety that we imagine that it does. Rather, I feel it gives us a false sense of security. But again, let's do whatever the consensus is (or whoever argues the loudest, which won't be me on this particular issue).

ZappoMan:
"sprinkling locks everywhere" is a mischaracterization of what I'm suggesting. - I don't condone that either. :)

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1018,71 +1018,72 @@ void MyAvatar::updateLookAtTargetAvatar() {
...
-                gazeOffset = humanOrientation * gazeOffsetLocalToHead;
-                glm::vec3 corrected = humanSystem->getPosition() + gazeOffset;
-               */
-
-                // And now we can finally add that offset to the camera.
-                glm::vec3 corrected = qApp->getViewFrustum()->getPosition() + gazeOffset;
-
-                avatar->getHead()->setCorrectedLookAtPosition(corrected);
-
+    DependencyManager::get<AvatarManager>()->withAvatarHash([&] (const AvatarHash& hash) {
```
jherico: I'd be worried that this might keep the lock for an excessive amount of time, especially since it increases linearly with the number of avatars.  Additionally, since it's making calls into other functions I'd be concerned about the potential for race conditions leading to a deadlock.

Wouldn't it be simpler to use the read lock to make a copy of the avatar hash and then do the for loop across the copy?

hifi-gustavo:
Build console output for commit: 7b0b77f4d11eb8b7f40631b111d0dafabfc56b68 available at: https://gist.github.com/30630d2b1501fa2e8a35<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6154.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6154.dmg

interface/src/avatar/MyAvatar.cpp
howard-stearns: I have no preference between:
1. as is
2. leave it up to the caller of withAvatarHash, such that in this case, this call site would make a copy to iterate over, and then iterate outside the call to withAvatarHash
3. create the copy in withAvatarHash, releasing the lock before calling the callback, such that users of withAvatarHash have no rope with which to hang themselves (but which may be wasteful, as in the other caller of withAvatarHash).

Which do you prefer? And which do you prefer, @ZappoMan ?

interface/src/avatar/AvatarManager.cpp
```diff
@@ -349,5 +355,9 @@ AvatarSharedPointer AvatarManager::getAvatarBySessionID(const QUuid& sessionID)
     if (sessionID == _myAvatar->getSessionUUID()) {
         return std::static_pointer_cast<Avatar>(_myAvatar);
     }
-    return getAvatarHash()[sessionID];
+    AvatarSharedPointer avatar;
+    withAvatarHash([&avatar, &sessionID] (const AvatarHash& hash) {
```
ZappoMan: we should probably just move this into the AvatarHash class

interface/src/avatar/MyAvatar.cpp
ZappoMan: Austin has a good point... I guess since the avatar hash has smart pointers to the avatars it's safe enough to copy it.. right now with a max of about 20 or so avatars in a domain,I'm sure it's fine.... but in the long run, I wonder if we aren't introducing a performance issue when 100+ avatars in a domain...

I feel like this is 6 one way half dozen the other... 

I'm ok with either solution since both seem temporary

interface/src/avatar/AvatarManager.cpp
howard-stearns: AvatarHash doesn't know about _myAvatar.

hifi-gustavo:
Build console output for commit: 367175b8a64e6ef9e5953861047479fb53a07241 available at: https://gist.github.com/30630d2b1501fa2e8a35<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6154.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6154.dmg

PR #6158, opened 2015-10-23
Synced blending for avatar movement animations

hyperlogic:
This adds separate animations for fast and slow avatar movement, which are dynamically blended and time-scaled at run-time to minimize foot slipping.  A new AnimBlendLinearMove node was introduced to the AnimGraph to support this.  This node maintains a list of characteristic speeds for each of its child animations which it uses for reference.  At run-time the Rig will set a desiredSpeed and alpha animVar.  The implementation of the AnimBlendLinearMove node will calculate the proper playback speed necessary to achieve the speed desired.

For the forward state, there are 3 animations, short walk, normal walk and run.  Backward and lateral movement only has two animations for blending, short and normal.

It also includes some other misc animation changes.  Including:

   * AnimClip::accumulateTime is moved into AnimUtil as a pure function for reuse and unitTesting.
   * The AnimPose class was moved out of AnimSkeleton into its own compilation unit.
   * Unit Tests for AnimUtil::accumulateTime were created


hifi-gustavo:
Build console output for commit: 86891704159f0154d48070a66a2ad3c8c55a84e6 available at: https://gist.github.com/73dd7fbcf82c765fcceb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.dmg

libraries/animation/src/AnimBlendLinearMove.cpp
```diff
@@ -0,0 +1,126 @@
...
+//
+//  Created by Anthony J. Thibault on 10/22/15.
+//  Copyright (c) 2015 High Fidelity, Inc. All rights reserved.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include "AnimBlendLinearMove.h"
+#include "GLMHelpers.h"
```
AndrewMeadows: GLMHelpers.h comes from the **shared** lib and so it should be:

    #include <GLMHelpers.h>

libraries/animation/src/AnimBlendLinearMove.cpp
```diff
@@ -0,0 +1,126 @@
...
+        evaluateAndBlendChildren(animVars, triggersOut, alpha, prevPoseIndex, nextPoseIndex, prevDeltaTime, nextDeltaTime);
+    } else {
+
+        float clampedAlpha = glm::clamp(_alpha, 0.0f, (float)(_children.size() - 1));
+        size_t prevPoseIndex = glm::floor(clampedAlpha);
+        size_t nextPoseIndex = glm::ceil(clampedAlpha);
+        float alpha = glm::fract(clampedAlpha);
+        float prevDeltaTime, nextDeltaTime;
+        setFrameAndPhase(dt, alpha, prevPoseIndex, nextPoseIndex, &prevDeltaTime, &nextDeltaTime, triggersOut);
+        evaluateAndBlendChildren(animVars, triggersOut, alpha, prevPoseIndex, nextPoseIndex, prevDeltaTime, nextDeltaTime);
```
AndrewMeadows: I don't understand why the range of **alpha** would be **[0, numChildren-1]** rather than just **[0, 1]**.

In particular, the math in **setFrameAndPhase()** does not work for **alpha** > 1.0.

Maybe I'm missing something?  I'll keep reading...

AndrewMeadows: Ok, I guess it is ok to **lerp** with **alpha** larger than 1.0... it just overshoots the **next** value in a linear way. This means that... the **setFrameAndPhase()** math might also work fine for blend factors larger than unity?  I'm still trying to wrap my brain around it.

tests/animation/src/AnimTests.cpp
```diff
@@ -238,3 +239,87 @@ void AnimTests::testVariant() {
...
+    QVERIFY(resultFrame == endFrame + 0.5f);
+    QVERIFY(triggers.empty());
+    triggers.clear();
+
+    // now we should hit loop trigger
+    resultFrame = accumulateTime(startFrame, endFrame, timeScale, resultFrame, dt, loopFlag, id, triggers);
+    QVERIFY(resultFrame == startFrame + 0.5f);
+    QVERIFY(!triggers.empty() && triggers[0] == "testNodeOnLoop");
+    triggers.clear();
+}
```
AndrewMeadows: Yay! Unit tests!

libraries/animation/src/AnimBlendLinearMove.cpp
hyperlogic: The AnimBlendLinear and AnimBlendLinearMove can have more than two children, in this case if you wish to blend between halfway between the second and third child, you would use an alpha value of 1.5.

Internally in setFrameAndPhase and evaulateAndBlendChildren() we call glm::fract to convert to the relevant [0,1) range.

hifi-gustavo:
Build console output for commit: e639d5313944f7b4689c4454f1a9b3f652cc897c available at: https://gist.github.com/73dd7fbcf82c765fcceb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.dmg

libraries/animation/src/AnimBlendLinearMove.cpp
hyperlogic: Yeah, I it should be fine in the integer boundary case, prevPoseIndex and nextPoseIndex will be the same, because of the floor and ceil functions, in that case the fract(clampedAlpha) will be 0.

hifi-gustavo:
Build console output for commit: 06de0878020e1be466acef54dfd52b20219cbf89 available at: https://gist.github.com/73dd7fbcf82c765fcceb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.dmg

libraries/animation/src/Rig.cpp
```diff
@@ -426,19 +459,41 @@ void Rig::computeMotionAnimationState(float deltaTime, const glm::vec3& worldPos
     }
 #endif
 
+    _lastVelocity = workingVelocity;
+
```
AndrewMeadows: It seems to me that this update of _lastVelocity should also NOT happen when deltaTime is small. It should go inside the if block higher up or else get its own if block.

hyperlogic: Good catch, this was working ok for me due to the fact that the updates with small dt values were infrequent and isolated.

hifi-gustavo:
Build console output for commit: f9dc05c98919f4287cefc247711ec9d572af797c available at: https://gist.github.com/73dd7fbcf82c765fcceb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.dmg

howard-stearns:
I'm seeing a lot of judder on Mac, where the simrate varies quite a bit, sometimes quite a bit faster than the framerate.  I don't see as much on Windows (stable simrate of 35-40/s), but I then I never really see any running?

AndrewMeadows:
Is this blocked for judder on Mac?  Has that been confirmed as new in this PR?  Should we merge and fix later or fix now?

howard-stearns:
@hyperlogic is looking at whether https://github.com/highfidelity/hifi/pull/6200 provides enough to get good velocity, and in what cases. If so, then we can incorporate that in this PR and merge. Otherwise, I think there are a couple of different options we could discuss.

hifi-gustavo:
Build console output for commit: f7b6fab0c33afe1fdbdcc2dde873beecbf98a5eb available at: https://gist.github.com/73dd7fbcf82c765fcceb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.dmg

hifi-gustavo:
Build console output for commit: e3e759248ffa59fa4b955dd7e05d15a58977dc4d available at: https://gist.github.com/73dd7fbcf82c765fcceb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.dmg

hifi-gustavo:
Build console output for commit: 2236573a9dc335a5b366a90070154b37d0d68c6f available at: https://gist.github.com/73dd7fbcf82c765fcceb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6158.dmg

howard-stearns:
I still some judder, mostly mac, but let's deal with that (or not) after merge.

PR #6159, opened 2015-10-23
Fix controllerExample.js

EdgarPironti:


hifi-gustavo:

ZappoMan:
There are a lot of files being moved around here that SHOULD NOT be moved... many files are being moved into an "entityScripts" directory that are not entity scripts.

jherico:
You need to repair this branch in order to make it sane.  I suggest the following:

`git checkout controllers`
`git reset upstream/controllers --hard` 
`git cherry-pick eaaec516c27b5578f09c28d9093806c4796d1342` 


That should eliminate the merge commits listed here which are creating confusion.  This assumes your remote name for the hifi repository is `upstream`, but if it's something else replace `upstream/controllers` with that.


In the future I would suggest avoiding merging from upstream/controllers, and instead rebasing.

PR #6160, opened 2015-10-23
Fix paddleBall.js

AlessandroSigna:


hifi-gustavo:

examples/controllers/hydra/paddleBall.js
```diff
@@ -63,7 +63,7 @@ var ball, paddle, paddleModel, line;
 function createEntities() {
     ball = Entities.addEntity(
                 { type: "Sphere",
-                position: Controller.getSpatialControlPosition(controllerID),   
+                position: Controller.getPoseValue(hand).translation,
```
ZappoMan: I don't think this is going to work. The old code was trying to create a ball at the hand position in world space, because that's what `getSpatialControlPosition()` would return. The new code `getPoseValue(hand).translation` is going to return a control relative position (which will basically be near 0,0,0)... so this isn't going to work. You need to use `MyAvatar.leftHandPosition` and/or `MyAvatar.rightHandPosition`

examples/controllers/hydra/paddleBall.js
```diff
@@ -73,27 +73,27 @@ function createEntities() {
...
                 visible: false,
-                rotation: Quat.multiply(MyAvatar.orientation, Controller.getSpatialControlRawRotation(controllerID)),
+                rotation: Quat.multiply(MyAvatar.orientation, Controller.getPoseValue(hand).rotation,
                 collisionsWillMove: false });
 
     modelURL = "http://public.highfidelity.io/models/attachments/pong_paddle.fbx";
     paddleModel = Entities.addEntity(
                 { type: "Model",
-                position: Vec3.sum(Controller.getSpatialControlPosition(controllerID), PADDLE_BOX_OFFSET),   
+                position: Vec3.sum(Controller.getPoseValue(hand).translation, PADDLE_BOX_OFFSET),
```
ZappoMan: same issue as noted above

examples/controllers/hydra/paddleBall.js
```diff
@@ -118,7 +118,7 @@ function deleteEntities() {
 }
 
 function update(deltaTime) {
-    var palmPosition = Controller.getSpatialControlPosition(controllerID);
+    var palmPosition = Controller.getPoseValue(hand).translation;
```
ZappoMan: same issue as noted above

examples/controllers/hydra/paddleBall.js
```diff
@@ -146,10 +146,10 @@ function update(deltaTime) {
         Entities.editEntity(ball, { velocity: ballVelocity });
         Overlays.editOverlay(line, { start: props.position, end: holdPosition });
         Entities.editEntity(paddle, { position: holdPosition, 
-                                      velocity: Controller.getSpatialControlVelocity(controllerID),
+                                      velocity: Controller.getPoseValue(hand).velocity,
```
ZappoMan: this also won't work... the old velocity was returned in world frame... the new one is returned in controller frame. You need to either calculate the velocity from the sampled values of the avatar hand, or we need to find a new solution

jherico: should we not just expose MyAvatar.leftPose / MyAvatar.rightPose with all the rotations and translations and velocities in world space?

ZappoMan: hmmm - that might not be a bad Idea... one second. let me see if that's readily available...

ZappoMan: yes - let's do that... I will add a card and do it.

PR #6170, opened 2015-10-23
various controller script updates

ZappoMan:
* make handGrab treat shoulder bumbers and triggers as merged control
* fix hydraGrabHockey.js to use new API
* fix hydraPoint to use new API

### DIFF
--- a/examples/controllers/handControllerGrab.js
+++ b/examples/controllers/handControllerGrab.js
@@ -250,7 +250,6 @@ function MyController(hand, triggerAction) {
     };
 
     this.updateSmoothedTrigger = function () {
-        //var triggerValue = Controller.getValue(this.triggerAction); // this.rawTriggerValue; // 
         var triggerValue = this.rawTriggerValue;
         // smooth out trigger value
         this.triggerValue = (this.triggerValue * TRIGGER_SMOOTH_RATIO) +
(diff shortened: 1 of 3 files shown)


examples/controllers/handControllerGrab.js
```diff
@@ -861,6 +869,12 @@ function MyController(hand, triggerAction) {
 var rightController = new MyController(RIGHT_HAND, Controller.Standard.RT);
 var leftController = new MyController(LEFT_HAND, Controller.Standard.LT);
 
+var mapping = Controller.newMapping("handGrab");
```
jherico: if using a named mapping, I would suggest a more distinct name, like 'com.highfidelity.handGrab'.  Additionally, this should be a constant, rather than repeated in multiple places.  

However, you mya prefer to use an anonymous mapping instead.

ZappoMan:
test this please

examples/controllers/handControllerGrab.js
```diff
@@ -861,6 +867,12 @@ function MyController(hand, triggerAction) {
 var rightController = new MyController(RIGHT_HAND, Controller.Standard.RT);
 var leftController = new MyController(LEFT_HAND, Controller.Standard.LT);
 
+var mapping = Controller.newMapping("com.highfidelity.handControllerGrab");
```
jherico: still using a magic string in 3 places instead of a constant

ZappoMan: duh - will fix

PR #6171, opened 2015-10-24
Add Bow & Arrow

imgntn:
This PR adds a Bow and Arrow toy to /toybox.  Pick it up with one hand and shoot with the other.

To test:
 https://rawgit.com/imgntn/hifi/bow/examples/toybox/bow/createBow.js


examples/controllers/handControllerGrab.js
```diff
@@ -37,6 +37,7 @@ var BUMPER_ON_VALUE = 0.5;
 var DISTANCE_HOLDING_RADIUS_FACTOR = 5; // multiplied by distance between hand and object
 var DISTANCE_HOLDING_ACTION_TIMEFRAME = 0.1; // how quickly objects move to their new position
 var DISTANCE_HOLDING_ROTATION_EXAGGERATION_FACTOR = 2.0; // object rotates this much more than hand did
+
```
sethalves: are the changes in this file requires for the bow-and-arrow to work?

examples/playa/playaSpawner.js
```diff
@@ -127,16 +155,14 @@ spawnGround();
...
-        Entities.deleteEntity(box);
-    });
-    dustSystems.forEach(function(dustEffect) {
-        Entities.deleteEntity(dustEffect);
-    })
-}
-
-Script.scriptEnding.connect(cleanup);
-
+// function cleanup() {
```
sethalves: why is cleanup commented out?

examples/controllers/handControllerGrab.js
imgntn: yep -- the part at the bottom listens for a message to disable the opposite hand of what you use to pick up the bow.  otherwise you'll far-grab stuff with your beams while you're pulling back, so i was calling it the 'beam disabler', but in reality it disables the whole hand.

examples/playa/playaSpawner.js
imgntn: good question, this file shouldn't really be changed, i was using it during testing.  will update

examples/controllers/handControllerGrab.js
sethalves: Would you mind pulling out the handControllerGrab.js changes into a separate PR?

imgntn: how would i do that?

imgntn:
@ericrius1 can you please test and merge this?

PR #6176, opened 2015-10-24
fix frisbee.js to work with new API

ZappoMan:


examples/controllers/hydra/frisbee.js
```diff
@@ -130,29 +130,38 @@ function Hand(name, palm, tip, forwardButton, button3, trigger) {
     this.trigger = trigger;
     this.holdingFrisbee = false;
     this.entity = false;
-    this.palmPosition = function() { return Controller.getSpatialControlPosition(this.palm); }
+    this.palmPosition = function () {
+        return this.palm == LEFT_PALM ? MyAvatar.leftHandPose.translation : MyAvatar.rightHandPose.translation;
```
jherico: I'm seeing and having to do this pattern in a lot of places.  

I think we should consider exposing arrays of symmetrical controls (in addition to the left / right versions).

Hence, Standard.LT == Standard.Triggers[0], Standard.RT == Standard.Triggers[1], MyAvatar.leftHandPose[0] == MyAvatar.handPoses[0], etc...

examples/controllers/hydra/frisbee.js
```diff
@@ -130,29 +130,38 @@ function Hand(name, palm, tip, forwardButton, button3, trigger) {
...
+    };
 }
 
 function MouseControl(button) {
     this.button = button;
 }
 
-var leftHand = new Hand("LEFT", LEFT_PALM, LEFT_TIP, LEFT_BUTTON_FWD, LEFT_BUTTON_3, 0);
-var rightHand = new Hand("RIGHT", RIGHT_PALM, RIGHT_TIP, RIGHT_BUTTON_FWD, RIGHT_BUTTON_3, 1);
+var leftHand = new Hand("LEFT", LEFT_PALM, LEFT_TIP, Controller.Standard.LB, Controller.Standard.LeftPrimaryThumb, Controller.Standard.LT);
```
jherico: I'd suggest Controller.Standard.LeftPrimaryIndex here for constency with LeftPrimaryThumb.  Don't think it's wired up yet though.

PR #6177, opened 2015-10-24
Controller Branch - Fix gun.js

jherico:


libraries/controllers/src/controllers/impl/endpoints/ScriptEndpoint.cpp
```diff
@@ -23,14 +23,18 @@ void ScriptEndpoint::updateValue() {
         return;
     }
 
-    _lastValue = (float)_callable.call().toNumber();
+    _lastValueRead = (float)_callable.call().toNumber();
 }
 
 void ScriptEndpoint::apply(float newValue, float oldValue, const Pointer& source) {
+    if (newValue == _lastValueWritten) {
```
jherico: This code removes the need for storing state on the JS side just so you can only react to changes in state.

examples/controllers/hydra/gun.js
```diff
@@ -356,137 +450,70 @@ function keyPressEvent(event) {
...
-    // check for trigger press
-
-    var numberOfTriggers = 2; 
-    var controllersPerTrigger = 2;
-
-    if (numberOfTriggers == 2 && controllersPerTrigger == 2) {
-        for (var t = 0; t < 2; t++) {
-            var shootABullet = false;
-            var triggerValue = Controller.getTriggerValue(t);
-            if (triggerPulled[t]) {
```
jherico: Note that the logic below may be valuable, but we don't have a filter type for it.  It's similar to a dead zone but has a minimum activation value higher than the deactivation value.  This prevents the possibility of rapidly spamming simply because the control is hovering right at some critical value.

hifi-gustavo:
Build console output for commit: 3e50174114b36f8e7bd77aaf664f1091fed78b26 available at: https://gist.github.com/3a6883d10c5ed90bf8bf<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6177.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6177.dmg

examples/controllers/hydra/gun.js
ZappoMan: Philip often calls this type of thing hysteresis, the idea that the threshold in one direction is different than the other direction so you don't get constantly flipping pops if you sit on the boundary. 

Other possible names might be: 

threshold(x,y) 
boundary(x,y)
gate(x,y)

examples/controllers/hydra/gun.js
```diff
@@ -15,24 +15,61 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 
+Script.include([ "../../libraries/utils.js" ]);
```
ZappoMan: the "array" version of these includes will cause the include call to be asynchronous, I don't think you want that... if you do want that, then why not actually include all 3 includes in the same array.

examples/controllers/hydra/gun.js
```diff
@@ -188,41 +209,45 @@ function shootBullet(position, velocity, grenade) {
...
-    audioOptions.position = position;   
+    // Play firing sounds
+    audioOptions.position = position;
     Audio.playSound(fireSound, audioOptions);
     shotsFired++;
     if ((shotsFired % RELOAD_INTERVAL) == 0) {
         Audio.playSound(loadSound, audioOptions);
     }
-
-    // Kickback the arm
```
ZappoMan: why did you remove this? does MyAvatar.setJointData() no longer work? Maybe we should leave this in with a FIXME... I'm sure that Philip would like kickback to work still... I think he's the one that added that feature.

examples/controllers/hydra/gun.js
```diff
@@ -356,137 +450,70 @@ function keyPressEvent(event) {
...
 }
 
 function deleteBulletAndTarget() {
     Entities.deleteEntity(bulletID);
     Entities.deleteEntity(targetID);
-    bulletID = false; 
-    targetID = false; 
-}
-
-function takeFiringPose() {
```
ZappoMan: same here.. why are we removing this?

examples/controllers/hydra/gun.js
```diff
@@ -356,137 +450,70 @@ function keyPressEvent(event) {
...
-function shootFromMouse(grenade) {
-    var DISTANCE_FROM_CAMERA = 1.0;
-    var camera = Camera.getPosition();
-    var forwardVector = Quat.getFront(Camera.getOrientation());
-    var newPosition = Vec3.sum(camera, Vec3.multiply(forwardVector, DISTANCE_FROM_CAMERA));
-    var velocity = Vec3.multiply(forwardVector, BULLET_VELOCITY);
-    shootBullet(newPosition, velocity, grenade);
+        // Need to adjust the laser
+        var tipPose = tipPoses[side];
+        var handRotation = MyAvatar.getJointCombinedRotation(side == 0 ? "LeftHand" : "RightHand");
```
ZappoMan: shouldn't this come from MyAvatar.leftHandPose.rotation and MyAvatar.rightHandPose.rotation?

examples/controllers/hydra/gun.js
```diff
@@ -99,78 +107,91 @@ var NUM_BUTTONS = 3;
...
-Script.include(["../../libraries/toolBars.js"]);
-var toolBar = new ToolBar(0, 0, ToolBar.HORIZONTAL, "highfidelity.gun.toolbar", function (screenSize) {
+
+var toolBar = new ToolBar(0, 0, ToolBar.HORIZONTAL, "highfidelity.gun.toolbar", function(screenSize) {
     return {
         x: startX,
         y: (screenSize.y - (BUTTON_SIZE + PADDING)),
     };
 });
-var reticle = Overlays.addOverlay("image", {
```
ZappoMan: why are you removing the reticle?

examples/controllers/hydra/gun.js
huffman: Actually the include call should only be asynchronous if you pass a function as the second parameter. So, this should be fine, but I agree these three lines should all be in one Script.include call.

examples/controllers/hydra/gun.js
jherico: Yeah, all of the join setting calls failed because they weren't setting position AND rotation.  I'll update.

examples/controllers/hydra/gun.js
jherico: I was playing with this for quite a while.  In theory, I should be able to use the left / right hand pose to fetch everything I need.  In practice, for some reason the laser vector doesn't line up with my gun barrel consistently as I rotate the gun.  Since it's attached to the left hand joint, I thought using that as the basis for all the calculations would improve the situation.

examples/controllers/hydra/gun.js
jherico: The reticle doesn't move and is only there for mouse functionality.  I removed the mouse based input in order to make this a demo strictly of spatial controllers.

examples/controllers/hydra/gun.js
jherico: none of these names really lend themselves to an obvious interpretation of what is happening, at least to me.  However hysteresis at least suggests something non-obvious is going on, so I'll use that.

examples/controllers/hydra/gun.js
jherico: The `ScriptEngine::include` method for a single file is actually implemented in terms of the version for multiple files.  

For clarity I've removed the brackets, but I've retained three separate include calls.

examples/controllers/hydra/gun.js
jherico: added fixme to document.

examples/controllers/hydra/gun.js
ZappoMan: I don't think we want to do that... I mean... the sample is "gun.js" not "handControllerGun.js"

examples/controllers/hydra/gun.js
jherico: same reason, join manipulation as it was in the code didn't work, and caused exceptions.  Fixing that should be out of scope for the sprint, IMO.

examples/controllers/hydra/gun.js
jherico: it's gun.js in `examples/controllers/hydra`.  Really it should be in `examples/controllers/spatial` now...

examples/controllers/hydra/gun.js
jherico: fixed

hifi-gustavo:
Build console output for commit: 93fc1e3765867c7fe97615be6421aecbecab2055 available at: https://gist.github.com/3a6883d10c5ed90bf8bf<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6177.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6177.dmg

hifi-gustavo:
Build console output for commit: 75b2ee94b243f34f7c7fab4c8bfe6f6dce627816 available at: https://gist.github.com/3a6883d10c5ed90bf8bf<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6177.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6177.dmg

PR #6190, opened 2015-10-27
Bug fix for hand IK when using the Owen avatar.

hyperlogic:
The IK was assiming that the "Hips" bone index was always 0.
This was not the case for Owen.  Now, we lookup the Hips index
and cache it for use during the hipsOffset computation.


hifi-gustavo:
Build console output for commit: 485e36d824c5d843beec47cde01c525ba272ddca available at: https://gist.github.com/e3768275c09da9ae13b7<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6190.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6190.dmg

howard-stearns:
great catch!

PR #6199, opened 2015-10-27
thread safety clean up of use of PalmData

ZappoMan:
This implements various thread safety fixes for HandData and PalmData
* changes HandData to guard it's _palms info with a mutex
* changes most "read only" users of palm data to get a copy of the PalmData and/or a copy of the current Palms vector for their use
* Removes a bunch of old APIs for accessing palm info that are deprecated in favor of the Pose apis
* removes the notion of SixenseID's from Palm data and their users

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -2741,13 +2741,13 @@ void Application::update(float deltaTime) {
     controller::Pose leftHand = userInputMapper->getPoseState(controller::Action::LEFT_HAND);
     controller::Pose rightHand = userInputMapper->getPoseState(controller::Action::RIGHT_HAND);
     Hand* hand = DependencyManager::get<AvatarManager>()->getMyAvatar()->getHand();
-    setPalmData(hand, leftHand, deltaTime, LEFT_HAND_INDEX, userInputMapper->getActionState(controller::Action::LEFT_HAND_CLICK));
-    setPalmData(hand, rightHand, deltaTime, RIGHT_HAND_INDEX, userInputMapper->getActionState(controller::Action::RIGHT_HAND_CLICK));
+    setPalmData(hand, leftHand, deltaTime, HandData::LeftHand, userInputMapper->getActionState(controller::Action::LEFT_HAND_CLICK));
+    setPalmData(hand, rightHand, deltaTime, HandData::RightHand, userInputMapper->getActionState(controller::Action::RIGHT_HAND_CLICK));
     if (Menu::getInstance()->isOptionChecked(MenuOption::EnableHandMouseInput)) {
         emulateMouse(hand, userInputMapper->getActionState(controller::Action::LEFT_HAND_CLICK),
-            userInputMapper->getActionState(controller::Action::SHIFT), LEFT_HAND_INDEX);
+            userInputMapper->getActionState(controller::Action::SHIFT), HandData::LeftHand);
         emulateMouse(hand, userInputMapper->getActionState(controller::Action::RIGHT_HAND_CLICK),
-            userInputMapper->getActionState(controller::Action::SHIFT), RIGHT_HAND_INDEX);
+            userInputMapper->getActionState(controller::Action::SHIFT), HandData::RightHand);
     }
 
     updateThreads(deltaTime); // If running non-threaded, then give the threads some time to process...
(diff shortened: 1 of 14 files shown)


hifi-gustavo:
Build console output for commit: aa07a6382c86c7e72d00d512870d4e9c327968f6 available at: https://gist.github.com/d60be2fb6d11c7af5c56<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.dmg

hifi-gustavo:
Build console output for commit: 4e40599ffcf5f30a6d5607afad81223616e851e8 available at: https://gist.github.com/d60be2fb6d11c7af5c56<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.dmg

hifi-gustavo:
Build console output for commit: 48e9b45220a200306d126099eb28d53df8f4eb77 available at: https://gist.github.com/d60be2fb6d11c7af5c56<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.dmg

hifi-gustavo:
Build console output for commit: cbea94489437e530905f9af15c842e3e2a77e3f5 available at: https://gist.github.com/d60be2fb6d11c7af5c56<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.dmg

hifi-gustavo:
Build console output for commit: 09288cd12222ee444a964eede7c0e0ec95c39a43 available at: https://gist.github.com/d60be2fb6d11c7af5c56<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.dmg

hifi-gustavo:
Build console output for commit: 5a42991e0ccd075100c2da985f7d67925fd076c5 available at: https://gist.github.com/d60be2fb6d11c7af5c56<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.dmg

libraries/avatars/src/HandData.h
```diff
@@ -12,26 +12,30 @@
...
 
-const int LEFT_HAND_INDEX = 0;
-const int RIGHT_HAND_INDEX = 1;
-const int NUM_HANDS = 2;
-
-const int SIXENSEID_INVALID = -1;
-
 class HandData {
 public:
+    enum Hand {
```
jherico: What's the use case for 'UnknownHand'?  The only time palms are created is in the HandData constructor, where it adds two.  I would suggest that we make Hand a required parameter of PalmData's constructor and just set left or right as appropriate.

libraries/avatars/src/HandData.h
```diff
@@ -12,26 +12,30 @@
...
+        UnknownHand,
+        RightHand,
```
jherico: virtually everywhere else uses the convention Left / Right, usally as aliases for 0 & 1 but the order is reversed here, leaving Right as 1, but making left 2.  

I'd also suggest we put an enum in shared called `Bilateral` with `Left` and `Right` members and just use that, rather than repeatedly making enums in multiple locations.

libraries/avatars/src/HandData.h
```diff
@@ -163,21 +153,14 @@ class PalmData {
...
-    unsigned int _controllerButtons;
-    unsigned int _lastControllerButtons;
     float _trigger;
-    float _joystickX, _joystickY;
     
-    bool      _isActive;             // This has current valid data
-    int       _sixenseID;            // Sixense controller ID for this palm
-    int       _numFramesWithoutData; // after too many frames without data, this tracked object assumed lost.
+    bool _isActive; /// This has current valid data
+    HandData::Hand _hand = HandData::UnknownHand;
```
jherico: if we set this via the ctor to left or right we can make it const.

interface/src/Application.cpp
```diff
@@ -4805,86 +4805,101 @@ mat4 Application::getHMDSensorPose() const {
...
+    // reading or writing to the Palms. This is definitely not the best way of handling this, and I'd like to see more
+    // of this palm manipulation in the Hand class itself. But unfortunately the Hand and Palm don't knbow about
+    // controller::Pose. More work is needed to clean this up.
+    hand->modifyPalms([&](std::vector<PalmData>& palms) {
+        auto myAvatar = DependencyManager::get<AvatarManager>()->getMyAvatar();
+        PalmData* palm;
+        bool foundHand = false;
+
+        // FIXME - this little chuck of code is a hot mess. It's basically searching the palms
+        // for the one that matches the "sixenseID" that is the "index" parameter. If it
```
jherico: update comment, since sixenseID isn't a thing anymore.

ZappoMan: damn you and your eye for detail!

libraries/avatars/src/HandData.h
```diff
@@ -67,10 +70,15 @@ class HandData {
 
     glm::quat getBaseOrientation() const;
 
+    /// Allows a lamda function write access to the palms for this Hand
+    void modifyPalms(std::function<void(std::vector<PalmData>& palms)> callback)
```
Atlante45: Use a `template<typename Callable>` here instead. That'll allow the compiler to inline it if it deems it useful.

libraries/avatars/src/HandData.h
```diff
@@ -90,10 +98,12 @@ class PalmData {
...
-    int getSixenseID() const { return _sixenseID; }
+    bool isValid() const { return _owningHandData; }
 
     void setActive(bool active) { _isActive = active; }
-    void setSixenseID(int id) { _sixenseID = id; }
+
+    HandData::Hand whichHand() const { return _hand; }
+    void setHand(HandData::Hand hand) { _hand = hand; }
 
     void setRawRotation(const glm::quat rawRotation) { _rawRotation = rawRotation; };
```
jherico: missing &

libraries/avatars/src/HandData.h
Atlante45: Also means you function can have a different signature and still work (ie. bool(const std::vector<PalmData>&) would be acceptable)

interface/src/avatar/MyAvatar.cpp
```diff
@@ -554,74 +554,76 @@ void MyAvatar::updateFromTrackers(float deltaTime) {
...
             if (numberOfActivePalms == palmIndex) {
-                return &handData->getPalms()[i];
+                return palm;
             }
             numberOfActivePalms++;
         }
     }
-    return NULL;
+    PalmData noData(nullptr);
+    return noData;
```
Atlante45: Why not `return PalmData(nullptr)`?

libraries/avatars/src/HandData.h
ZappoMan: I sometimes need to return PalmData that is invalid... what Hand will an invalid PalmData have?

libraries/avatars/src/HandData.h
ZappoMan: ?

libraries/avatars/src/HandData.h
jherico: I'd suggest we should stop tossing around PalmData references and pointers and instead use std::shared_pointer<PalmData>, so invalid PalmData would be an empty pointer.  However, that's out of scope for now.

libraries/avatars/src/HandData.h
jherico: `const glm::quat rawRotation` makes no sense.  it should be either `glm::quat` or `const glm::quat&`

ZappoMan: yeah - that's not code I changed... but I will fix it...

hifi-gustavo:
Build console output for commit: eb1936412954d0eabf2152f9fc759b5a2c9405c0 available at: https://gist.github.com/d60be2fb6d11c7af5c56<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.dmg

libraries/avatars/src/HandData.h
```diff
@@ -12,26 +12,30 @@
...
 #include <glm/glm.hpp>
 #include <glm/gtc/quaternion.hpp>
 
+#include <QReadWriteLock>
+
 #include <NumericalConstants.h>
 #include <SharedUtil.h>
 
 class AvatarData;
 class PalmData;
```
Atlante45: Hmmm I wonder why it doesn't fail on windows.
PalmData needs be fully defined for the template, I wonder what they do to grab the definition.

ZappoMan: windows is magix

hifi-gustavo:
Build console output for commit: 19743c1f39c17743634d3b7894a57a882f8d3b55 available at: https://gist.github.com/d60be2fb6d11c7af5c56<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.dmg

libraries/avatars/src/HandData.h
ZappoMan: Let me clarify something here... because I think we're missing the larger goal of this PR... I want to not pass around pointers to data at all... because in doing so, even if they're shared pointers, the data can get changed on me from some other thread while I access it. So I've changed these to pass around PalmData AS VALUES... so that you get the last good copy of the data, and can be guaranteed that you won't have to worry about another thread changing or corrupting it on you while you use it. AFAICT - shared_pointer doesn't help us at all in this case.

hifi-gustavo:
Build console output for commit: d86f1b50a7c8a68c1d261879520eb0d915dc19ca available at: https://gist.github.com/d60be2fb6d11c7af5c56<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.dmg

hifi-gustavo:
Build console output for commit: 7f313e7f86bd93696b8b6d4fde40b15d6cceed7d available at: https://gist.github.com/d60be2fb6d11c7af5c56<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.dmg

hifi-gustavo:
Build console output for commit: 65eae3543329ac8b2585f7a24fac18cc40752c2e available at: https://gist.github.com/d60be2fb6d11c7af5c56<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6199.dmg

libraries/avatars/src/HandData.h
jherico: Sorry, in my head I was also thinking *and make the types themselves thread safe*, but apparently didn't write that down.

PR #6200, opened 2015-10-27
Move DynamicCharacterController to MyAvatarController

AndrewMeadows:
This work was part of an effort to keep the avatar's feet on the ground in HMD mode, however I got bogged down in that work so I decided to split out this part of it into its own PR.  This is all cleanup and preparation for future work -- there are no feature changes.

What to test: 

Does your avatar still move around correctly when walking, jumping, and flying?

Does your avatar still move around and collide correctly after changing to a different avatar model?


hifi-gustavo:
Build console output for commit: 8b285fd2287a3a6a02567208303f6f3e34693580 available at: https://gist.github.com/c1f2bd51bfa63c50e70d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6200.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6200.dmg

interface/src/avatar/MyAvatar.cpp
```diff
@@ -334,11 +339,6 @@ void MyAvatar::updateFromHMDSensorMatrix(const glm::mat4& hmdSensorMatrix) {
     const float BIGGEST_DELTA_TIME_SECS = 0.25f;
     float deltaTime = glm::clamp((float)actualDeltaTime, 0.0f, BIGGEST_DELTA_TIME_SECS);
 
-    // update the sensorMatrices based on the new hmd pose
-    _hmdSensorMatrix = hmdSensorMatrix;
-    _hmdSensorPosition = extractTranslation(hmdSensorMatrix);
-    _hmdSensorOrientation = glm::quat_cast(hmdSensorMatrix);
```
AndrewMeadows: These lines moved up to top of function.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -154,7 +154,8 @@ void MyAvatar::reset(bool andReload) {
     }
 
     // Reset dynamic state.
-    _wasPushing = _isPushing = _isBraking = _billboardValid = _straighteningLean = false;
+    _wasPushing = _isPushing = _isBraking = _billboardValid = false;
+    _isFollowingHMD = false;
```
AndrewMeadows: _straighteningLean is changing name to _isFollowingHMD.  I pulled it out into its own line because I'll be deleting is a future PR.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -360,58 +360,60 @@ void MyAvatar::updateFromHMDSensorMatrix(const glm::mat4& hmdSensorMatrix) {
...
 
-    processStraighteningLean(deltaTime);
+    followHMD(deltaTime);
 }
 
-void MyAvatar::beginStraighteningLean() {
+glm::vec3 MyAvatar::getHMDCorrectionVelocity() const {
+    // TODO: impelement this
+    return Vectors::ZERO;
+}
```
AndrewMeadows: This HMDVelocity stuff is in preparation for future work.  It doesn't do anything yet, but this PR calls it in a couple of the right spots for when it does.

interface/src/avatar/MyCharacterController.cpp
```diff
@@ -1,13 +1,26 @@
+//
+//  MyCharacterController.h
+//  interface/src/avatar
+//
+//  Created by AndrewMeadows 2015.10.21
+//  Copyright 2015 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
```
AndrewMeadows: **DynamicCharacterController** has been broken up into two parts:

**CharacterController** is the pure virtual base class and lives in the **physics** lib.

**MyCharacterController** is the real class and lives in the **interface** project.

hifi-gustavo:
Build console output for commit: e35596428c7e9445036131326122a5efc3d59cd4 available at: https://gist.github.com/c1f2bd51bfa63c50e70d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6200.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6200.dmg

interface/src/avatar/MyCharacterController.cpp
```diff
@@ -158,9 +160,20 @@ void DynamicCharacterController::playerStep(btCollisionWorld* dynaWorld, btScala
...
+    // Rather than add _hmdVelocity to the velocity of the RigidBody, we explicitly teleport 
+    // the RigidBody forward according to the formula: distance = rate * time
+    if (_hmdVelocity.length2() > 0.0f) {
+        btTransform bodyTransform = _rigidBody->getWorldTransform();
+        bodyTransform.setOrigin(bodyTransform.getOrigin() + dt * _hmdVelocity);
+        _rigidBody->setWorldTransform(bodyTransform);
+    }
+    // MyAvatar will ask us how far we stepped for HMD motion, which will depend on how 
+    // much time has accumulated in _lastStepDuration.
+    _lastStepDuration += dt;
```
AndrewMeadows: This code is not used yet but will be when we start making the avatar follow the HMD sensor data more closely.  Basically the idea is as follows:

(1) interface code will know how much MyAvatar needs to move the avatar's RigidBody to agree with the current state of animation/hips-shift and also how fast it should be closing the gap.

(2) interface code tells the CharacterController its intended velocity it will use to close the gap.

(3) CharacterController will move according to the stated velocity, independently of whatever outside velocity it has (this code above teleports the avatar's RigidBody transform accordingly).

(4) After a simulation step the CharacterController can be queried to see how far it actually moved -- which will be a function of how many substeps were taken.  This shift will then be used to update the sensor transforms and the interface's logic about how much discrepancy between the hips and MyAvatar's center remains to be closed.

None of this works yet, I'm just laying the groundwork here.  The **_hmdVelocity** variable is guaranteed to be zero right now.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -360,58 +360,60 @@ void MyAvatar::updateFromHMDSensorMatrix(const glm::mat4& hmdSensorMatrix) {
...
+    if (shouldFollowHMD() || hmdIsAtRest || justStartedMoving) {
+        beginFollowingHMD();
     }
 
-    processStraighteningLean(deltaTime);
+    followHMD(deltaTime);
+}
+
+glm::vec3 MyAvatar::getHMDCorrectionVelocity() const {
+    // TODO: impelement this
```
Atlante45: No sure how often this is called, but I found out recently that Qt has a nice `Q_UNIMPLEMENTED()` macro that prints out a warning when an unimplemented piece of code is getting called.

hifi-gustavo:
Build console output for commit: c9cd3364005f29368c5412560adb86fea0fc381d available at: https://gist.github.com/c1f2bd51bfa63c50e70d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6200.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6200.dmg

interface/src/avatar/MyAvatar.h
```diff
@@ -154,8 +159,10 @@ class MyAvatar : public Avatar {
 
     virtual void setAttachmentData(const QVector<AttachmentData>& attachmentData) override;
 
-    DynamicCharacterController* getCharacterController() { return &_characterController; }
+    MyCharacterController* getCharacterController() { return &_characterController; }
```
Atlante45: I know this is unrelated to that PR, but it feels like we should return a ref here.

interface/src/avatar/MyAvatar.h
```diff
@@ -348,21 +355,21 @@ public slots:
...
     glm::vec3 _customListenPosition;
     glm::quat _customListenOrientation;
 
-    bool _straighteningLean = false;
-    float _straighteningLeanAlpha = 0.0f;
+    bool _isFollowingHMD { false };
+    float _followHMDAlpha{0.0f};
 
-    quint64 _lastUpdateFromHMDTime = usecTimestampNow();
+    quint64 _lastUpdateFromHMDTime{usecTimestampNow()};
```
Atlante45: Spacing consistency, everything is spaced except those two.

interface/src/avatar/MyCharacterController.cpp
```diff
@@ -310,19 +278,27 @@ void DynamicCharacterController::updateShapeIfNecessary() {
...
+            if (!_rigidBody) {
+                btCollisionShape* shape = new btCapsuleShape(_radius, 2.0f * _halfHeight);
+                _rigidBody = new btRigidBody(mass, nullptr, shape, inertia);
+            } else {
+                btCollisionShape* shape = _rigidBody->getCollisionShape();
+                if (shape) {
+                    delete shape;
+                }
+                shape = new btCapsuleShape(_radius, 2.0f * _halfHeight);
+                _rigidBody->setCollisionShape(shape);
```
Atlante45: Lots of raw pointers lying around here, it might be good to consider using `std::unique_ptr` for those that are private owning member.
```
btCollisionShape* shape = _rigidBody->getCollisionShape();
if (shape) {
    delete shape;
}
shape = new btCapsuleShape(_radius, 2.0f * _halfHeight);
_rigidBody->setCollisionShape(shape);
```
would become
` _rigidBody->setCollisionShape(new btCapsuleShape(_radius, 2.0f * _halfHeight));`

libraries/physics/src/CharacterController.cpp
```diff
@@ -0,0 +1,52 @@
...
+    }
+    if (_dynamicsWorld) {
+        if (_pendingFlags & PENDING_FLAG_UPDATE_SHAPE) {
+            // shouldn't fall in here, but if we do make sure both ADD and REMOVE bits are still set
+            _pendingFlags |= PENDING_FLAG_ADD_TO_SIMULATION | PENDING_FLAG_REMOVE_FROM_SIMULATION;
+        } else {
+            _pendingFlags &= ~PENDING_FLAG_ADD_TO_SIMULATION;
+        }
+    } else {
+        _pendingFlags &= ~ PENDING_FLAG_REMOVE_FROM_SIMULATION;
```
Atlante45: Extra space after `~`

interface/src/avatar/MyAvatar.h
AndrewMeadows: No not yet; I want to clean this up a different way.  This method is only ever used to give the **PhysicsEngine** a pointer to a **CharacterController** (the base class) and this is, IMO, a short-term hack.  Eventually we'll give **MyAvatar** its own **MotionState** which will manage the **CharacterController** instance.

interface/src/avatar/MyCharacterController.cpp
AndrewMeadows: No.  Again I want to clean this up a different way:  the CharacterController stuff should eventually use the **ShapeManager** to get its shapes which does ref counts, shape sharing, and garbage collection of unused shapes.

In the short term the ownership of this raw pointer is clear: the btRigidBody owns it until it needs another.

libraries/physics/src/CharacterController.cpp
AndrewMeadows: I can't find the coding standard that forbids the space after the **~**.  Could you identify it for me?

hifi-gustavo:
Build console output for commit: afec2a9771cadd729f21aca42020fada86e898cd available at: https://gist.github.com/c1f2bd51bfa63c50e70d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6200.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6200.dmg

howard-stearns:
tests ok

libraries/physics/src/CharacterController.cpp
Atlante45: I stand corrected, it's not in there. I could have sworn `~bits` and `-value` were the only exception to space all the operators, before we relaxed it to allow `x*y`.

PR #6203, opened 2015-10-28
adjust locking in EntityItem::getActionDataInternal

sethalves:
- possible fix for a crash in EntityItem::getActionDataInternal

### DIFF
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1792,8 +1792,7 @@ const QByteArray EntityItem::getActionData() const {
     assertUnlocked();
 
     if (_actionDataDirty) {
-        EntityItem* unconstThis = const_cast<EntityItem*>(this);
-        unconstThis->withWriteLock([&] {
+        withWriteLock([&] {
             getActionDataInternal();
             result = _allActionsDataCache;
         });
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1795,7 +1795,6 @@ const QByteArray EntityItem::getActionData() const {
         EntityItem* unconstThis = const_cast<EntityItem*>(this);
         unconstThis->withWriteLock([&] {
             getActionDataInternal();
-            return _allActionsDataCache;
         });
     }
 
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1795,6 +1795,7 @@ const QByteArray EntityItem::getActionData() const {
         EntityItem* unconstThis = const_cast<EntityItem*>(this);
         unconstThis->withWriteLock([&] {
             getActionDataInternal();
+            return _allActionsDataCache;
         });
     }
 
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1776,7 +1776,7 @@ void EntityItem::serializeActions(bool& success, QByteArray& result) const {
     return;
 }
 
-QByteArray EntityItem::getActionDataInternal() const {
+const QByteArray EntityItem::getActionDataInternal() const {
     if (_actionDataDirty) {
         bool success;
         serializeActions(success, _allActionsDataCache);
@@ -1787,7 +1787,7 @@ QByteArray EntityItem::getActionDataInternal() const {
     return _allActionsDataCache;
 }
 
-QByteArray EntityItem::getActionData() const {
+const QByteArray EntityItem::getActionData() const {
     QByteArray result;
     assertUnlocked();
 
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1776,7 +1776,7 @@ void EntityItem::serializeActions(bool& success, QByteArray& result) const {
     return;
 }
 
-const QByteArray EntityItem::getActionDataInternal() const {
+QByteArray EntityItem::getActionDataInternal() const {
     if (_actionDataDirty) {
         bool success;
         serializeActions(success, _allActionsDataCache);
@@ -1787,7 +1787,7 @@ const QByteArray EntityItem::getActionDataInternal() const {
     return _allActionsDataCache;
 }
 
-const QByteArray EntityItem::getActionData() const {
+QByteArray EntityItem::getActionData() const {
     QByteArray result;
     assertUnlocked();
 
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1795,12 +1795,13 @@ const QByteArray EntityItem::getActionData() const {
         EntityItem* unconstThis = const_cast<EntityItem*>(this);
         unconstThis->withWriteLock([&] {
             getActionDataInternal();
+            result = _allActionsDataCache;
+        });
+    } else {
+        withReadLock([&] {
+            result = _allActionsDataCache;
         });
     }
-
-    withReadLock([&] {
-        result = _allActionsDataCache;
-    });
     return result;
 }
 
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1790,8 +1790,16 @@ const QByteArray EntityItem::getActionDataInternal() const {
 const QByteArray EntityItem::getActionData() const {
     QByteArray result;
     assertUnlocked();
+
+    if (_actionDataDirty) {
+        EntityItem* unconstThis = const_cast<EntityItem*>(this);
+        unconstThis->withWriteLock([&] {
+            getActionDataInternal();
+        });
+    }
+
     withReadLock([&] {
-        result = getActionDataInternal();
+        result = _allActionsDataCache;
     });
     return result;
 }
--- a/libraries/entities/src/EntityItem.h
+++ b/libraries/entities/src/EntityItem.h
@@ -401,7 +401,7 @@ public:
     bool removeAction(EntitySimulation* simulation, const QUuid& actionID);
     bool clearActions(EntitySimulation* simulation);
     void setActionData(QByteArray actionData);
-    QByteArray getActionData() const;
+    const QByteArray getActionData() const;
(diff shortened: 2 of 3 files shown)


libraries/entities/src/EntityItem.cpp
```diff
@@ -1790,8 +1790,16 @@ const QByteArray EntityItem::getActionDataInternal() const {
 const QByteArray EntityItem::getActionData() const {
```
ZappoMan: why is this `const`?

libraries/entities/src/EntityItem.cpp
```diff
@@ -1790,8 +1790,16 @@ const QByteArray EntityItem::getActionDataInternal() const {
...
+    if (_actionDataDirty) {
+        EntityItem* unconstThis = const_cast<EntityItem*>(this);
+        unconstThis->withWriteLock([&] {
+            getActionDataInternal();
+        });
+    }
+
     withReadLock([&] {
-        result = getActionDataInternal();
+        result = _allActionsDataCache;
```
ZappoMan: why do you need a read lock around this?

libraries/entities/src/EntityItem.cpp
```diff
@@ -1790,8 +1790,16 @@ const QByteArray EntityItem::getActionDataInternal() const {
 const QByteArray EntityItem::getActionData() const {
     QByteArray result;
     assertUnlocked();
+
+    if (_actionDataDirty) {
+        EntityItem* unconstThis = const_cast<EntityItem*>(this);
+        unconstThis->withWriteLock([&] {
+            getActionDataInternal();
```
ZappoMan: wouldn't it make more sense to make this line be...  `result =  getActionDataInternal();` since `getActionDataInternal()` returns the QByteArray as a result.

hifi-gustavo:
Build console output for commit: f91dc93620a331a72953978d6d6b78e7a3964396 available at: https://gist.github.com/928cf941a04f3cec6207<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.dmg

libraries/entities/src/EntityItem.cpp
ZappoMan: I guess you need the readLock because you want to guard against multiple threads accessing the byte array.

libraries/entities/src/EntityItem.cpp
sethalves: It's a getter with a hidden cache and getters are generally const.  Maybe no better reason than that.

libraries/entities/src/EntityItem.cpp
sethalves: Yes, I think we generally get away with not locking our entity properties because they are mostly atomic.  The QStrings and QByteArrays may need careful locking.

libraries/entities/src/EntityItem.cpp
sethalves: Yes, probably.  In the past there was more than one caller to getActionDataInternal, and I tried to minimize the diff in this PR.

libraries/entities/src/EntityItem.cpp
ZappoMan: but it's not returning a const& so it's kinda a moot const... see what I mean? You should probably either just return a `QByteArray` in which case you're returning a value which may result in a copy (same result as what you have here)... or you should return a `const QByteArray&` which then means you're returning a reference to the member variable... which of course is consistent with our coding standard on getters, but also potentially not thread safe.

ZappoMan: also note: I see the same issue in a couple other methods in this file.

libraries/entities/src/EntityItem.cpp
ZappoMan: my main observation is that you can get the result with just this write lock for this case, and so you should set your result value and return without dropping down below to do the read lock. are you going to change that?

libraries/entities/src/EntityItem.cpp
sethalves: My (possibly incomplete) understanding is that QByteArray and QString do copy-on-write tricks, so when you return a non-reference, you are actually returning a reference.  I've also read that these data-structures do not guard against access from multiple threads at once.

libraries/entities/src/EntityItem.cpp
sethalves: okay, something like this?

hifi-gustavo:
Build console output for commit: 2b5b4f111823c4858c78be41aa371144c70fc1c3 available at: https://gist.github.com/928cf941a04f3cec6207<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.dmg

hifi-gustavo:
Build console output for commit: d504f449e4754b3d88030888d1c48d30e5dc019d available at: https://gist.github.com/928cf941a04f3cec6207<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.dmg

hifi-gustavo:
Build console output for commit: 08a0bf33a49833eb080c42b531634c61ed4e05b7 available at: https://gist.github.com/928cf941a04f3cec6207<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.dmg

libraries/entities/src/EntityItem.cpp
Atlante45: Yes this is what they call [implicit sharing](http://doc.qt.io/qt-4.8/implicit-sharing.html) and it is thread safe.
This is the only case I know of where returning a const value makes sense.
Since it's const you can't modify it and you won't have to pay the price of a copy, but if the QByteArray is deleted by an other thread at it's source, your object is still valid. It wouldn't be with a ref.

Of course you have to take into account here the cost of Qt's atomic ref count. But unless the platform doesn't support atomic int operations it doesn't cost much.

ZappoMan: In the event that anyone comes along and looks at these comments later... here's the answer to the QByteArray as value semantics with implicit sharing and thread safety question... 

http://doc.qt.io/qt-5/threads-modules.html#threads-and-implicitly-shared-classes

*To sum it up, implicitly shared classes in Qt 4 are really implicitly shared. Even in multithreaded applications, you can safely use them as if they were plain, non-shared, reentrant value-based classes.*

hifi-gustavo:
Build console output for commit: f6a0004f26f73ec7c506ec3d7a287fbd6799f29b available at: https://gist.github.com/928cf941a04f3cec6207<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.dmg

libraries/entities/src/EntityItem.cpp
```diff
@@ -1790,9 +1790,18 @@ const QByteArray EntityItem::getActionDataInternal() const {
 const QByteArray EntityItem::getActionData() const {
     QByteArray result;
     assertUnlocked();
-    withReadLock([&] {
-        result = getActionDataInternal();
-    });
+
+    if (_actionDataDirty) {
+        EntityItem* unconstThis = const_cast<EntityItem*>(this);
+        unconstThis->withWriteLock([&] {
```
Atlante45: Casting out const is usually unwise. Since `getActionDataInternal` is const, couldn't we add a const version of `withWriteLock` and avoid the cast?

sethalves: I would have thought so, but apparently compilers don't do any optimization based on const -- it's just a signal for devs.  In this case, it's doing lazy evaluation and caching the value.  This is the example people usually use when talking about the "mutable" modifier.

hifi-gustavo:
Build console output for commit: a1096510e80dc566ec9be49411099cc11c1584a2 available at: https://gist.github.com/928cf941a04f3cec6207<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.dmg

libraries/entities/src/EntityItem.cpp
ZappoMan: I checked the callers to getActionData() and we basically have the following:
* one that takes the value and uses it as a parameter to a function taking a `const QByteArray& bytes` -- in this case, I agree the const could theoretically give the compiler a hint that it's not changing
* 4 other cases are assignments into another different QByteArray - so no benefit there.
* one other case where it's compared to another QByteArray - so maybe some benefit there.

I guess we can say that 2 out of 6 uses might result in slight optimization over the already built in implicit sharing... and it doesn't hurt... so sure... I guess we can leave in the const.

libraries/entities/src/EntityItem.cpp
ZappoMan: honestly - we should make withWriteLock() work with const instances by making the mutex mutable.

ZappoMan: maybe add something like withConstWriteLock() or something... if we feel that it's too confusing to see withWriteLock() in a const method.

hifi-gustavo:
Build console output for commit: 22d6b6df34526da4be267bdafbeeae188c91fdb0 available at: https://gist.github.com/928cf941a04f3cec6207<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6203.dmg

PR #6204, opened 2015-10-28
emit hardwareChanged signal to scripts

ZappoMan:


hifi-gustavo:
Build console output for commit: 30f7c44f611dc644922673f1423cde1fe063995e available at: https://gist.github.com/bdf0fc39e7994ea82905<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6204.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6204.dmg

libraries/controllers/src/controllers/ScriptingInterface.cpp
```diff
@@ -50,6 +50,7 @@ controller::ScriptingInterface::ScriptingInterface() {
     // FIXME make this thread safe
     connect(userInputMapper.data(), &UserInputMapper::hardwareChanged, [=] {
```
Atlante45: I know you didn't add that code, but you should add a context object here:
`connect(userInputMapper.data(), &UserInputMapper::hardwareChanged, this, [=] {`
Otherwise the connection won't be cleaned up if `ScriptingInterface` is deleted and the lambda might get called with and deleted this pointer.

The capture parameter doesn't need to be `[=]` too, we just need `[this]` here.

ZappoMan: great catch - will fix

ZappoMan: wrt to the implicit capture, I'm going to leave it as is... there's no performance difference, and it makes it easy for future developers if they need more than just the this.

http://stackoverflow.com/questions/19018100/is-it-a-bad-practice-to-always-capture-all-in-a-lambda-expression

```
The standard guarantees that if you do a default capture, the only variables that will be captured by that default capture from the surrounding environment are those that you actually use inside the lambda.

As such, specifying individual variables to capture acts as documentation of what you expect to use, but should never affect performance.

For anybody who cares, the exact wording from the standard is (§5.1.2/11, 12):

11 If a lambda-expression has an associated capture-default and its compound-statement odr-uses (3.2) this or a variable with automatic storage duration and the odr-used entity is not explicitly captured, then the odr-used entity is said to be implicitly captured; such entities shall be declared within the reaching scope of the lambda expression. [Note elided]

12 An entity is captured if it is captured explicitly or implicitly. [...]
Summary: an implicit capture specification ([=] or [&]) will only capture variables that are used in the lambda.
```

hifi-gustavo:
Build console output for commit: 457ec76d3d85f69e6153b66b1d63afa0c73ab432 available at: https://gist.github.com/bdf0fc39e7994ea82905<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6204.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6204.dmg

PR #6212, opened 2015-10-28
Don't play other avatar's collision sounds.

howard-stearns:
https://app.asana.com/0/32622044445063/61104546898252


hifi-gustavo:
Build console output for commit: 8e83943efd4840f25166c2d12ffa482f38938d12 available at: https://gist.github.com/dc6901f5270fe8d7ee49<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6212.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6212.dmg

AndrewMeadows:
It occurs to me now that these changes will funnel **entity-avatar** and **avatar-avatar** collision events into the EntitySimulation API (specifically PhysicalEntitySimulation::handleCollisionEvents()) and will be interpreted as if they were **entity-entity** collision events.

I think we'll have to somehow flag the collision events by objType-objType (or maybe flag the CollisionEvent::idA and idB data members by object type) and then filter them accordingly inside PhysicalEntitySimulation::handleCollisionEvents().

howard-stearns:
Sounds right. Let me poke at this some more and see if I can dig up (or add) enough info to the avatar motion state.

AndrewMeadows:
No, I suspect you want to add information to the **Collision** class rather than the **AvatarMotionState**.  The **Colliison** class currently has a **type** data member (BEGIN, CONTINUE, END), but the objects involved probably need their own **objectType** (ENTITY, AVATAR, and eventually MY_AVATAR).

hifi-gustavo:
Build console output for commit: c2873ef7204f4c0cd2fabf199c73d77c9a884b0a available at: https://gist.github.com/dc6901f5270fe8d7ee49<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6212.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6212.dmg

hifi-gustavo:
Build console output for commit: ce1d0fe07c1d89e3bff075a2e31455d5e58688a1 available at: https://gist.github.com/dc6901f5270fe8d7ee49<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6212.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6212.dmg

hifi-gustavo:
Build console output for commit: fc63fa6f7b963acc1c6c481a35009c8a5b9e939b available at: https://gist.github.com/dc6901f5270fe8d7ee49<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6212.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6212.dmg

howard-stearns:
After discussion:

- This fixes the bug in which everyone separately injected collisions sounds for not only their avatars, but also for others. 

- This does not address any new strategy for sorting through different pairs of collision types. As it stands:
  -- Other people's avatars get an avatar motion state (and now with a valid type). 
  -- The collision stores non-null ids for entities and (now) for other people's avatars. (Still null for one's own avatar.)  
  -- Collision sounds get injected (only by the simulation owner) for every entity involved in a collision (including entity on entity, or entity on avatar), using that entity's collision sound, and also find their way to any script that has registered for a collision event on that entity.
  -- Collision sounds get injected for any collision with my avatar (whether with an entity or another avatar, and regardless of entity simulation owner), using my avatar's collision sound, and also find their way to any script that has connected to MyAvatar's collision signal.

AndrewMeadows:
I don't see the "only send collisions sounds if I own the simulation" changes.  Where did they go?

Also, it looks like entity-avatar collisions are still arriving at the EntityTreeRenderer::entityCollisionWithEntity() context as an entity-entity collision ==> scripts will get these events and they will appear to be a collision between one findable and one unfindable entityID.

howard-stearns:
re owning the simulation: Not a change. It has long been the case that EntityTreeRenderer::playEntityCollisionSound does not play sounds when not the owner (beginning line 720). Scripts get the raw events for those entities they have registered for, and have to decide for themselves if they should further filter by simulation owner.

re entity collision with avatar: EntityTreeRenderer::playEntityCollisionSound does not process when the id is not found in the entity tree (line 717).  (And as above, the script has to registered for the entity, so no scripts will be registered for an "entity" that has an avatar id instead.)

PR #6216, opened 2015-10-29
FBXReader: More deterministic loading behavior

hyperlogic:
This makes iteration over meshes and connections between them deterministic.
[QHash](http://doc.qt.io/qt-5/qhash.html#details) and QMultiHash do not guarantee consistent iteration order.  This is problematic for the FBXReader because it could result in different behavior each time a model was loaded.

Specifically, This was causing a bug with some avatars that contained multiple-bind poses.
The bind pose returned to the application from the FBXReader would be different on different runs.
This PR doesn't add support for multiple-bind poses, but it does make the choice of which
bind pose is chosen deterministic.

This non-determinism was the cause of the Mery avatar having "bug-eyes" 1/12 times.

Also, dangling whitespace was removed

### DIFF
--- a/libraries/fbx/src/FBXReader.cpp
+++ b/libraries/fbx/src/FBXReader.cpp
@@ -1401,11 +1401,6 @@ FBXGeometry* FBXReader::extractFBXGeometry(const QVariantHash& mapping, const QS
                 joint.bindTransform = cluster.transformLink;
                 joint.bindTransformFoundInCluster = true;
 
-                if (fbxCluster.jointIndex == 63) { // Head
-                    qCDebug(modelformat) << "AJT: Head found in cluster, id = " << clusterID;
-                    qCDebug(modelformat) << "AJT:     trans = " << extractTranslation(cluster.transformLink);
-                }
-
                 // update the bind pose extents
                 glm::vec3 bindTranslation = extractTranslation(geometry.offset * joint.bindTransform);
                 geometry.bindExtents.addPoint(bindTranslation);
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 5acb088c4644d521fa9a66f28e2ecdacdbc4b271 available at: https://gist.github.com/08e3da9b3b60d323594e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6216.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6216.dmg

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1401,6 +1401,11 @@ FBXGeometry* FBXReader::extractFBXGeometry(const QVariantHash& mapping, const QS
                 joint.bindTransform = cluster.transformLink;
                 joint.bindTransformFoundInCluster = true;
 
+                if (fbxCluster.jointIndex == 63) { // Head
+                    qCDebug(modelformat) << "AJT: Head found in cluster, id = " << clusterID;
+                    qCDebug(modelformat) << "AJT:     trans = " << extractTranslation(cluster.transformLink);
+                }
+
```
howard-stearns: Remove debugging code?

hyperlogic: DOH!

hifi-gustavo:
Build console output for commit: 7367f2a28cd375a064141c586403dc4514998c27 available at: https://gist.github.com/08e3da9b3b60d323594e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6216.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6216.dmg

PR #6221, opened 2015-10-29
Physics substep fix

huffman:


libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -251,7 +251,12 @@ void PhysicsEngine::stepSimulation() {
         _characterController->preSimulation(timeStep);
     }
 
-    int numSubsteps = _dynamicsWorld->stepSimulation(timeStep, PHYSICS_ENGINE_MAX_NUM_SUBSTEPS, PHYSICS_ENGINE_FIXED_SUBSTEP);
+    auto onSubStep = [this]() {
+        updateContactMap();
+        _hasOutgoingChanges = true;
+    };
+
+    int numSubsteps = _dynamicsWorld->stepSimulation(timeStep, PHYSICS_ENGINE_MAX_NUM_SUBSTEPS, PHYSICS_ENGINE_FIXED_SUBSTEP, onSubStep);
```
huffman: @AndrewMeadows Can you take a look at this? What I was seeing was that if a collision and ended before the last substep in `stepSimulation`, the collision would not show up in the collision dispatcher. Unfortunately I wasn't able to find much documentation on bullet's collision dispatcher, so this was all based on observation. This change makes sure we update the contact map after each substep.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -774,7 +774,7 @@ void EntityTreeRenderer::entityCollisionWithEntity(const EntityItemID& idA, cons
     }
     // Don't respond to small continuous contacts.
     const float COLLISION_MINUMUM_PENETRATION = 0.002f;
-    if ((collision.type != CONTACT_EVENT_TYPE_START) && (glm::length(collision.penetration) < COLLISION_MINUMUM_PENETRATION)) {
+    if ((collision.type == CONTACT_EVENT_TYPE_CONTINUE) && (glm::length(collision.penetration) < COLLISION_MINUMUM_PENETRATION)) {
```
huffman: If a collision starts and ends within a single `stepSimulation`, the only dispatched collision will be for `END`.

hifi-gustavo:
Build console output for commit: 913026280c082a0d7df6311a1136ca95e45afac1 available at: https://gist.github.com/51a00f44f11759c340f4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6221.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6221.dmg

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -260,8 +265,6 @@ void PhysicsEngine::stepSimulation() {
         if (_characterController) {
             _characterController->postSimulation();
         }
-        updateContactMap();
-        _hasOutgoingChanges = true;
     }
```
AndrewMeadows: The **_hasOutgoingChanges** line should remain here.

libraries/physics/src/PhysicsEngine.cpp
AndrewMeadows: Yes that makes sense.  The physics engine will take multiple substeps if given a long enough **timeStep** and collisions that begin and end within multiple substeps taken together would never get noticed.  Now if we could get to the point where we call **PhysicsEngine::stepSimulation()** much more often we wouldn't need this change, but this makes it more resilient to plodding frame rates which will probably always plague us as the world's content gets more interesting.

hifi-gustavo:
Build console output for commit: 49877c2dffba35b6902e3c92dfd9568615dcab78 available at: https://gist.github.com/51a00f44f11759c340f4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6221.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6221.dmg

libraries/physics/src/PhysicsEngine.cpp
huffman: Ok, I moved it out of the callback and back here.

PR #6228, opened 2015-10-29
removed old metavoxel example, since the api is no longer available

ZappoMan:


AndrewMeadows:
If we need to recover this code we can always just download it from the internet.

PR #6229, opened 2015-10-29
guard timeElapsed for simulateKinematicMotion to max 1s

birarda:
This is a fix for having your machine go to sleep and then getting stuck looping on `while (dt > PHYSICS_ENGINE_FIXED_SUBSTEP)` if `dt` is very large.


birarda:
@AndrewMeadows please double check to make sure this is what we would want (not allowing `simulateKinematicMotion` to be called with a `timeElapsed` larger than 1s.

hifi-gustavo:
Build console output for commit: fb55f0becbbb1a0ae589224f4f26906d939ed191 available at: https://gist.github.com/7363444f8545ee4550ad<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6229.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6229.dmg

PR #6253, opened 2015-10-30
Crash fix for AvatarManager when iterating over _avatarHash

hyperlogic:
The main problem is that a null shared pointer was inserted into the _avatarHash
via the AvatarManager::getAvatarBySessionID().  When the sessionID is not present
in the _avatarHash, [QHash](http://doc.qt.io/qt-5/qhash.html#operator-5b-5d) will *insert* an empty smart_ptr into the hash.


hifi-gustavo:
Build console output for commit: 085282db4f4cf1ed69ec4489c3906a3ab9f32bcf available at: https://gist.github.com/1332c234fc345d2abcf9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6253.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6253.dmg

PR #6264, opened 2015-11-01
tweaks to how VrMenu is shown

ZappoMan:
* Show the context menu even on non-stereo displays
* wire up the Actions.ContextMenu action to show the menu

### DIFF
--- a/interface/resources/controllers/keyboardMouse.json
+++ b/interface/resources/controllers/keyboardMouse.json
@@ -81,6 +81,7 @@
 
         { "from": "Keyboard.Space", "to": "Actions.SHIFT" },
         { "from": "Keyboard.R", "to": "Actions.ACTION1" },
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: e899229cdeb0f78d274ad4b83c91585b778d37c1 available at: https://gist.github.com/c163e6b7f9bf7597cc4e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6264.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6264.dmg

jherico:
I think if were going to add this we need to have the VR menu own a mapping which allows you to navigate the menu using a controller.  we'll need to think about how to map it to something like a hydra or touch controller, but at the very least Dpad direction keys should navigate the menu and something like right primary thumb should select a menu item while the menu is visible.

PR #6269, opened 2015-11-02
proper avatar velocity for HMD motion

AndrewMeadows:
This PR routes the avatar motion for HMD shifts through the physics engine.  The velocity of the HMD shift is added to other velocities before the animation system queries it for walk speeds.

The main thing to pay attention to is the time it takes for the HMD shift to be corrected.  Although the tuned the period of time has not changed the time it actually takes has increased.  The main reason for this (I think) is that the velocity of completion is computed as soon as the shift threshold is hit, however the user typically is not done shifting so the shift endpoint moves out while the shift is in progress.  The velocity of the shift is locked in as soon as it starts (for the most part, details details) and so the early velocity (for the short distance) is used for the whole shift (longer distance).  This can be tuned before or after the PR is merged.  Please test it out.

Also, there are some behavior changes to how the hips are allowed to shift when wearing the HMD:

(1) The avatar no longer has a shift threshold when moving DOWN, which means it is easier to squat.  However...

(2) If the avatar shifts laterally while squatting then the shift correction logic will kick in, the HMD shift will be corrected,  and the avatar will stand up again.  What is really happening is that the avatar's RigidBody is begin pushed into the floor and the physics pushes it out.  Since the shift correction is moving down to remove the squat the effect is that the avatar stands up.

(3) The avatar's feet can still rise off the ground when the user stands up from a sitting position while wearing the HMD... until they exceed the threshold, at which point the avatar will fall to the ground.  This will be fixed in another PR.

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -365,15 +365,16 @@ void MyAvatar::updateHMDFollowVelocity() {
     // compute offset to body's target position (in sensor-frame)
     auto sensorBodyMatrix = deriveBodyFromHMDSensor();
     _hmdFollowOffset = extractTranslation(sensorBodyMatrix) - extractTranslation(_bodySensorMatrix);
-    if (_hmdFollowOffset.y < 0.0f) {
+    glm::vec3 truncatedOffset = _hmdFollowOffset;
+    if (truncatedOffset.y < 0.0f) {
         // don't pull the body DOWN to match the target (allow animation system to squat)
-        _hmdFollowOffset.y = 0.0f;
+        truncatedOffset.y = 0.0f;
     }
 
     bool needNewFollowSpeed = (_isFollowingHMD && _hmdFollowSpeed == 0.0f);
     if (!needNewFollowSpeed) {
         // check to see if offset has exceeded its threshold
-        float distance = glm::length(_hmdFollowOffset);
+        float distance = glm::length(truncatedOffset);
         const float MAX_HMD_HIP_SHIFT = 0.2f;
         if (distance > MAX_HMD_HIP_SHIFT) {
             _isFollowingHMD = true;
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -250,7 +250,7 @@ void MyAvatar::simulate(float deltaTime) {
     {
         PerformanceTimer perfTimer("transform");
         bool stepAction = false;
-        // When there are no step values, we zero out the last step pulse. 
+        // When there are no step values, we zero out the last step pulse.
         // This allows a user to do faster snapping by tapping a control
         for (int i = STEP_TRANSLATE_X; !stepAction && i <= STEP_YAW; ++i) {
             if (_driveKeys[i] != 0.0f) {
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: 3b11fd92852f41d54ec55e100556fa693a1575a5 available at: https://gist.github.com/c8bd9ea416a24bf1becb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6269.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6269.dmg

interface/src/avatar/MyAvatar.cpp
```diff
@@ -359,101 +342,57 @@ void MyAvatar::updateFromHMDSensorMatrix(const glm::mat4& hmdSensorMatrix) {
...
-            glm::vec3 nextBodyPos = extractTranslation(newBodySensorMatrix);
-            glm::quat nextBodyRot = glm::normalize(glm::quat_cast(newBodySensorMatrix));
-            glm::vec3 prevBodyPos = extractTranslation(_bodySensorMatrix);
-            glm::quat prevBodyRot = glm::normalize(glm::quat_cast(_bodySensorMatrix));
-            pos = lerp(prevBodyPos, nextBodyPos, _followHMDAlpha);
-            rot = glm::normalize(safeMix(prevBodyRot, nextBodyRot, _followHMDAlpha));
-            _bodySensorMatrix = createMatFromQuatAndPos(rot, pos);
-        }
+        _hmdFollowVelocity = (_hmdFollowOffset / FOLLOW_HMD_DURATION);
+        _hmdFollowSpeed = glm::length(_hmdFollowVelocity);
```
AndrewMeadows: Hrm... the follow speed is computed based on the truncated offset (negative y value set to zero) but when the resolution of the offset kicks in it resolves the non-truncated distance... which means it resolves slower than it should. I'll fix this.

AndrewMeadows:
This is busted: avatar cannot move in 2D mode.

hifi-gustavo:
Build console output for commit: 35b1e5d06144cdda3bc72543cdd0787a09c40c9b available at: https://gist.github.com/c8bd9ea416a24bf1becb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6269.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6269.dmg

hifi-gustavo:
Build console output for commit: a545d770d888f12b42dbde32df275783afb02236 available at: https://gist.github.com/c8bd9ea416a24bf1becb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6269.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6269.dmg

hifi-gustavo:
Build console output for commit: a77f888f1574589f802daa62b373499858e5d7cb available at: https://gist.github.com/c8bd9ea416a24bf1becb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6269.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6269.dmg

PR #6270, opened 2015-11-02
Removing bball hoop from reset scripts- will be saved in SVO instead

ericrius1:
Added name to basketball hoop in reset scripts


imgntn:
@ericrius1 when you add the hoop to the SVO, be sure to set the userData to prevent it from being grabbable      
```grabbableKey: {grabbable: false  }```

imgntn:
lgtm w/SVO updates

PR #6274, opened 2015-11-02
add example of overriding default rightClick behavior

ZappoMan:


hifi-gustavo:
Build console output for commit: b94258f9536dab302db52651d54b8e94b5ddfb59 available at: https://gist.github.com/8be51c5356170f2f4ccb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6274.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6274.dmg

imgntn:
@ZappoMan hey -- the functionality works but i'm a little confused by our key events -- i see a "Keyboard.RightMouseClicked" event when i mousedown and when i mouseup, unless i move the mouse between those two events.  so if i click, move, and then release, i only see the first "Keyboard.RightMouseClicked" event.  is this expected?   i'd think you would see the event on down & up.  then listen to different events if you want specifically down / up, and then get a series of drag events during the drag. seems a little strange to me (maybe i can help document our keyboard events or something).

ZappoMan:
@imgntn - these aren't "events" they are input signals... they are different than events... the are more like a stream of input states. and yes, what you described is what's expected.

imgntn:
@ZappoMan interesting.  how do i find out how signals work? is there at least a list or something?  are they QML or QT?   the weirdest one is not getting the mouseclick signal when you release the mouse after moving it at all with the button down.  do you get a drag up signal or something instead?  it's kind of non-intuitive coming from web events.  maybe the mouse click is a special case, but entities have mousePressOnEntity and mouseReleaseOnEntity

CrackerHax:
So how do I disable this to use my own right click menu in my domain?

ZappoMan:
@crackerhax this script shows you exactly how to disable right click menus. If you've already got code to show your own menu, then just add this code to your script and this will disable the default menu.

CrackerHax:
Excellent, thanks!

PR #6282, opened 2015-11-03
Runtime plugins fixes

jherico:
This is work to correct two issues with our runtime plugins.  

First, if a runtime plugin attempts to call `Menu::instance()` or `DependencyManager::get()` right now it will fail to return the correct value.  This is because both of these objects use a singleton stored as a static.  However, a runtime plugin operating in its own memory space will have a separate copy of the static.  `Menu::instance()` will return a brand new menu object, while `DependencyManager::get()` will return a null pointer, because it's DLL local instance hasn't had any calls to `DependencyManager::set()`.  

This is corrected by having the QtCoreApplication object store the value as a property.  The property is looked up by name and instantiated it if doesn't exist.  However, the function that does so also has a function local static that will be populated, so subsequent attempts to fetch the value won't suffer any performance hit from the name lookup.  Currently I've applied this to these two singletons, Menu and DependencyManager, but there are probably others we will need to fix.

The second fix is to the build process.  Currently our only plugin is for the Oculus HMD, which functions because it looks up the Oculus DLL which is installed with the runtime.  However many of our plugins will rely on DLLs that have to be packaged with the executable, like SDL.  Right now the plugin is copied because the interface.exe app links to it, but once the functionality is in a plugin, interface.exe no longer has a link to the DLL and the CMake fixup_bundle command won't grab the DLL.  

This is corrected by using the second parameter to fixup_bundle, which takes a list of shared libraries to also scan for dependencies, and appears to be designed for precisely this situation.

### DIFF
--- a/libraries/shared/src/DependencyManager.cpp
+++ b/libraries/shared/src/DependencyManager.cpp
@@ -12,12 +12,13 @@
 #include "DependencyManager.h"
 
 #include "SharedUtil.h"
+#include "Finally.h"
 
 static const char* const DEPENDENCY_PROPERTY_NAME = "com.highfidelity.DependencyMananger";
 
 DependencyManager& DependencyManager::manager() {
-    static DependencyManager& instance = globalInstace<DependencyManager>(DEPENDENCY_PROPERTY_NAME);
-    return instance;
+    static DependencyManager* instance = globalInstace<DependencyManager>(DEPENDENCY_PROPERTY_NAME);
+    return *instance;
 }
 
 QSharedPointer<Dependency>& DependencyManager::safeGet(size_t hashCode) {
--- a/libraries/shared/src/DependencyManager.cpp
+++ b/libraries/shared/src/DependencyManager.cpp
@@ -17,7 +17,7 @@
 static const char* const DEPENDENCY_PROPERTY_NAME = "com.highfidelity.DependencyMananger";
 
 DependencyManager& DependencyManager::manager() {
-    static DependencyManager* instance = globalInstace<DependencyManager>(DEPENDENCY_PROPERTY_NAME);
+    static DependencyManager* instance = globalInstance<DependencyManager>(DEPENDENCY_PROPERTY_NAME);
     return *instance;
 }
 
--- a/libraries/shared/src/DependencyManager.cpp
+++ b/libraries/shared/src/DependencyManager.cpp
@@ -11,7 +11,14 @@
 
 #include "DependencyManager.h"
 
-DependencyManager DependencyManager::_manager;
+#include "SharedUtil.h"
+
+static const char* const DEPENDENCY_PROPERTY_NAME = "com.highfidelity.DependencyMananger";
+
+DependencyManager& DependencyManager::manager() {
+    static DependencyManager& instance = globalInstace<DependencyManager>(DEPENDENCY_PROPERTY_NAME);
+    return instance;
+}
 
 QSharedPointer<Dependency>& DependencyManager::safeGet(size_t hashCode) {
     return _instanceHash[hashCode];
--- a/libraries/shared/src/SharedUtil.h
+++ b/libraries/shared/src/SharedUtil.h
@@ -27,10 +27,10 @@
 // Provides efficient access to a named global type.  By storing the value
 // in the QApplication by name we can implement the singleton pattern and 
 // have the single instance function across DLL boundaries.  
-template <typename T, typename ...Args>
-T* globalInstace(const char* propertyName, Args&&... args) {
-    static std::shared_ptr<T> instancePtr;
-    static T *resultInstance { nullptr };
+template <typename T, typename... Args>
+T* globalInstance(const char* propertyName, Args&&... args) {
+    static std::unique_ptr<T> instancePtr;
+    static T* resultInstance { nullptr };
     static std::mutex mutex;
     if (!resultInstance) {
         std::unique_lock<std::mutex> lock(mutex);
@@ -39,7 +39,8 @@ T* globalInstace(const char* propertyName, Args&&... args) {
             if (variant.isNull()) {
                 // Since we're building the object, store it in a shared_ptr so it's 
                 // destroyed by the destructor of the static instancePtr
-                instancePtr = std::make_shared<T>(args...);
+                instancePtr = std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+
                 void* voidInstance = &(*instancePtr);
                 variant = QVariant::fromValue(voidInstance);
                 qApp->setProperty(propertyName, variant);
--- a/libraries/shared/src/SharedUtil.h
+++ b/libraries/shared/src/SharedUtil.h
@@ -28,24 +28,27 @@
 // in the QApplication by name we can implement the singleton pattern and 
 // have the single instance function across DLL boundaries.  
 template <typename T, typename ...Args>
-T& globalInstace(const char* propertyName, Args&&... args) {
-    static T *instance { nullptr };
+T* globalInstace(const char* propertyName, Args&&... args) {
+    static std::shared_ptr<T> instancePtr;
+    static T *resultInstance { nullptr };
     static std::mutex mutex;
-    if (!instance) {
+    if (!resultInstance) {
         std::unique_lock<std::mutex> lock(mutex);
-        if (!instance) {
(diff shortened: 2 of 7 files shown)


hifi-gustavo:
Build console output for commit: b77759956c9fab142dcd8e86cf3a5db9eba5d272 available at: https://gist.github.com/1041be63ad3542751a53<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6282.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6282.dmg

hifi-gustavo:
Build console output for commit: 025c4e45dcc1d21c3e0cf7ddfe28c2cec8a78385 available at: https://gist.github.com/1041be63ad3542751a53<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6282.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6282.dmg

libraries/shared/src/SharedUtil.h
```diff
@@ -20,7 +21,32 @@
...
+template <typename T, typename ...Args>
+T& globalInstace(const char* propertyName, Args&&... args) {
+    static T *instance { nullptr };
+    static std::mutex mutex;
+    if (!instance) {
+        std::unique_lock<std::mutex> lock(mutex);
+        if (!instance) {
+            auto variant = qApp->property(propertyName);
+            if (variant.isNull()) {
+                auto* typedInstance = new T(args...);
```
Atlante45: Those instances never get destructed.
The DependencyManager heavily relies on static destruction to cleanup everything in it's hash.

Atlante45: We could require T to be a defined QMetaType and store a `QSharedPointer<T>` in the variant, that way they would get cleaned up with the app.

jherico: Thanks for the suggestion, I'll look into that.

hifi-gustavo:
Build console output for commit: 1c19f5df84d94674cde508ba7cb48c34856f725a available at: https://gist.github.com/1041be63ad3542751a53<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6282.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6282.dmg

libraries/shared/src/SharedUtil.h
jherico: I tried to use Q_DECLARE_METATYPE for the dependency manager but could not get it to work.  I finally realized I could resolve this by having the template function declare not only a static pointer, but a static `shared_ptr` as well.  The latter only gets populated if the function constructs the object in question, ensuring that it will be destroyed when the static `shared_ptr` is.  If the function finds the existing instance in the `qApp`, it never initializes the `shared_ptr`.

hifi-gustavo:
Build console output for commit: 2dd84a49c4705fe92219c177b0ac5dce09ade241 available at: https://gist.github.com/1041be63ad3542751a53<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6282.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6282.dmg

libraries/shared/src/SharedUtil.h
```diff
@@ -20,7 +21,35 @@
...
 
-#include <QDebug>
+#include <QtCore/QDebug>
+#include <QtCore/QCoreApplication>
+
+// Provides efficient access to a named global type.  By storing the value
+// in the QApplication by name we can implement the singleton pattern and 
+// have the single instance function across DLL boundaries.  
+template <typename T, typename ...Args>
+T* globalInstace(const char* propertyName, Args&&... args) {
```
samcake: Little typo ?
globalInstance instead of globalInstace no ?

libraries/shared/src/DependencyManager.cpp
```diff
@@ -11,7 +11,15 @@
...
 #include "DependencyManager.h"
 
-DependencyManager DependencyManager::_manager;
+#include "SharedUtil.h"
+#include "Finally.h"
+
+static const char* const DEPENDENCY_PROPERTY_NAME = "com.highfidelity.DependencyMananger";
+
+DependencyManager& DependencyManager::manager() {
+    static DependencyManager* instance = globalInstace<DependencyManager>(DEPENDENCY_PROPERTY_NAME);
```
samcake: typo on the name "globalInstace" see below comment in SharedUtils.h

libraries/shared/src/SharedUtil.h
```diff
@@ -20,7 +21,35 @@
...
+    static std::shared_ptr<T> instancePtr;
```
Atlante45: Since it's only stored once, we should probably use a `std::unique_ptr` instead.

libraries/shared/src/SharedUtil.h
```diff
@@ -20,7 +21,35 @@
...
+    static T *resultInstance { nullptr };
```
Atlante45: Coding Standard: `T* resultInstance`

libraries/shared/src/SharedUtil.h
```diff
@@ -20,7 +21,35 @@
...
 #endif
 
-#include <QDebug>
+#include <QtCore/QDebug>
+#include <QtCore/QCoreApplication>
+
+// Provides efficient access to a named global type.  By storing the value
+// in the QApplication by name we can implement the singleton pattern and 
+// have the single instance function across DLL boundaries.  
+template <typename T, typename ...Args>
```
Atlante45: I'm using the same form in the DependencyManager but I just realized that based on our coding standard regarding `*` and `&`, maybe `typename... Args` would make more sense.
What do you think?

libraries/shared/src/SharedUtil.h
```diff
@@ -20,7 +21,35 @@
...
+    static T *resultInstance { nullptr };
+    static std::mutex mutex;
+    if (!resultInstance) {
+        std::unique_lock<std::mutex> lock(mutex);
+        if (!resultInstance) {
+            auto variant = qApp->property(propertyName);
+            if (variant.isNull()) {
+                // Since we're building the object, store it in a shared_ptr so it's 
+                // destroyed by the destructor of the static instancePtr
+                instancePtr = std::make_shared<T>(args...);
```
Atlante45: Since you're grabbing rvalue refs, you can perfect forward here: `std::make_shared<T>(std::forward<Args>(args)...)`

libraries/shared/src/SharedUtil.h
jherico: concur, will fix.

libraries/shared/src/SharedUtil.h
jherico: fixed

libraries/shared/src/DependencyManager.cpp
jherico: fixed

libraries/shared/src/SharedUtil.h
jherico: fixed

libraries/shared/src/SharedUtil.h
jherico: fixed

hifi-gustavo:
Build console output for commit: 2bbae0eab668e0734dd004b7427aeef257c5353b available at: https://gist.github.com/65418651f283efe6bdd8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6282.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6282.dmg

hifi-gustavo:
Build console output for commit: 6122fa145db9beadb66581223c80a48cc11901fc available at: https://gist.github.com/1041be63ad3542751a53<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6282.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6282.dmg

PR #6283, opened 2015-11-03
Remove idleTimer and call idle directly from paintGL.

hyperlogic:
Synchronize Application::idle with Application::paintGL.  At high frame rates, this should improve the quality of MyAvatar animations and the motion of locally simulated physics objects.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1041,8 +1041,25 @@ void Application::initializeUi() {
 }
 
 void Application::paintGL() {
+    _frameCount++;
 
-    idle();
+    // update fps moving average
+    uint64_t now = usecTimestampNow();
+    static uint64_t lastPaintBegin{ now };
+    uint64_t diff = now - lastPaintBegin;
+    if (diff != 0) {
+        _framesPerSecond.updateAverage((float)USECS_PER_SECOND / (float)diff);
+    }
+
+    lastPaintBegin = now;
+
+    // update fps once a second
+    if (now - _lastFramesPerSecondUpdate > USECS_PER_SECOND) {
+        _fps = _framesPerSecond.getAverage();
+        _lastFramesPerSecondUpdate = now;
+    }
+
+    idle(now);
 
     PROFILE_RANGE(__FUNCTION__);
     PerformanceTimer perfTimer("paintGL");
@@ -1317,24 +1334,6 @@ void Application::paintGL() {
         gpu::doInBatch(renderArgs._context, [=](gpu::Batch& batch) {
             batch.resetStages();
         });
-    }
-
-    // update fps moving average
-    {
-        uint64_t now = usecTimestampNow();
-        static uint64_t lastPaintEnd{ now };
-        uint64_t diff = now - lastPaintEnd;
-        if (diff != 0) {
-            _framesPerSecond.updateAverage((float)USECS_PER_SECOND / (float)diff);
-        }
-        lastPaintEnd = now;
-
-        // update fps once a second
-        if (now - _lastFramesPerSecondUpdate > USECS_PER_SECOND) {
-            _fps = _framesPerSecond.getAverage();
-            _lastFramesPerSecondUpdate = now;
-        }
-        _frameCount++;
     }
 }
 
@@ -2107,7 +2106,7 @@ void Application::ping() {
     }
 }
 
-void Application::idle() {
+void Application::idle(uint64_t now) {
     if (_aboutToQuit) {
         return; // bail early, nothing to do here.
     }
@@ -2130,7 +2129,6 @@ void Application::idle() {
 
     {
         PROFILE_RANGE(__FUNCTION__);
-        uint64_t now = usecTimestampNow();
         static uint64_t lastIdleStart{ now };
         uint64_t idleStartToStartDuration = now - lastIdleStart;
         if (idleStartToStartDuration != 0) {
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -160,7 +160,6 @@ static QTimer balanceUpdateTimer;
 static QTimer identityPacketTimer;
 static QTimer billboardPacketTimer;
 static QTimer checkFPStimer;
-static QTimer idleTimer;
 
 static const QString SNAPSHOT_EXTENSION  = ".jpg";
 static const QString SVO_EXTENSION  = ".svo";
@@ -844,7 +843,6 @@ void Application::cleanupBeforeQuit() {
     identityPacketTimer.stop();
     billboardPacketTimer.stop();
     checkFPStimer.stop();
-    idleTimer.stop();
     QMetaObject::invokeMethod(&_settingsTimer, "stop", Qt::BlockingQueuedConnection);
 
     // save state
@@ -982,9 +980,6 @@ void Application::initializeGL() {
     connect(&checkFPStimer, &QTimer::timeout, this, &Application::checkFPS);
     checkFPStimer.start(1000);
 
-    // call our idle function whenever we can
-    connect(&idleTimer, &QTimer::timeout, this, &Application::idle);
-    idleTimer.start(TARGET_SIM_FRAME_PERIOD_MS);
     _idleLoopStdev.reset();
 
     // update before the first render
@@ -1048,6 +1043,9 @@ void Application::initializeUi() {
 }
 
 void Application::paintGL() {
+
+    idle();
+
     PROFILE_RANGE(__FUNCTION__);
     PerformanceTimer perfTimer("paintGL");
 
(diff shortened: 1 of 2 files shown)


interface/src/Application.cpp
```diff
@@ -982,9 +980,6 @@ void Application::initializeGL() {
     connect(&checkFPStimer, &QTimer::timeout, this, &Application::checkFPS);
     checkFPStimer.start(1000);
 
-    // call our idle function whenever we can
-    connect(&idleTimer, &QTimer::timeout, this, &Application::idle);
-    idleTimer.start(TARGET_SIM_FRAME_PERIOD_MS);
```
Atlante45: I think the is the only place we use TARGET_SIM_FRAME_PERIOD_MS, can you check and remove it if it is?

hifi-gustavo:
Build console output for commit: 0c12bb5a4ea613697599d77ad3970cdb0390596f available at: https://gist.github.com/7f2b50e9ec54522187be<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6283.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6283.dmg

ZappoMan:
Here's my feedback on this PR. I shared most of this with Tony 1:1.

--------------
 I think there are really several different issues going on here... and I think this change improves on 2 issues, while opening up the risk to some other issues

the two areas this improves:
1) it gets away from the whims of Qt's timer processing
2) it runs the simulation at "faster than 60hz" in case of faster than 60hz displays (75hz for oculus, and 90hz for vive)


I think both those things are improvements. and I think that's actually the positive thing that you're all experiencing... and I think we should merge this for precisely those improvements.

however - I think that we're conflating those two improvements with a side effect of this implementation and that is "synching the simulation with the rendering".

I think we will discover that... if we can't render + simulate at 75hz/90hz -- we will not like the new synched behavior, and we will prefer a "render at vsynch & simulate as fast as possible".

The old (current) code tried to implement "render at vsynch & simulate as fast as possible with whatever time is left". However, it has two issues with get in it's way:

1) due to Qt's timers, the simulates often got skipped, so we didn't really get to simulate as often as possible.
2) due to the present being on the same thread as the simulation thread, if we missed a vsynch even by just a fraction - we end up waiting (blocking) until the next vsynch... which of course eats into our potential simulation time.

You'll notice that with your change, if we can keep up, we do have "render at vsynch & simulate as fast as possible"... but where this code will likely fail is the case of when we fall behind. In that case, the old code degraded simulation (and because of the present blocking on vsynch it slowed it down unfairly) but kept the rendering up.

I think we will discover that rendering at 90hz and simulating at 45hz will be better than rendering + simulating at 45hz. This code will likely produce those less desirable effects in the case where we can't keep up.

There is an even better solution to that case... but it's not as simple as this fix.

So... we should merge this... and we should all expect to have some really really bad frame rates when we get overloaded (like our company meetings).

jherico:
I concur that we should merge this, but I think we should add a `FIXME` at the idle call that says to move back to the timer based mechanism once we move GL present off the main thread.

ZappoMan:
Let's do some testing on this tomorrow in demo/toybox... if we can get to 4-5 people in the room and frame rate is not impacted, then we should merge... If toybox fails at 4-5 people then we need to come up with a temporary solution to not let simulate get in the way of the render in that use case.

ZappoMan:
On my windows machine, while in "demo" and holding a box in my hand and moving it around quickly - I see a frame rate drop from 75hz down to ~65hz fluctuating. As expected, simulation is ~ in synch with frame rate. (I think they are on slightly different running averages, so they aren't perfectly in synch).

On the production build 3473 - doing the same thing, my frame rate stays at 75hz, while simulation is about 60hz.

ZappoMan:
Very interesting... after doing a little more testing... It seems like I can run pretty smoothly at 75hz render/simulate for a while and if I'm gentle... But... if I hold a block, and shake my hand very vigorously -- after about 10 seconds of doing that, I get into a state of a precipitous FPS drop... which quickly drops me down to ~50hz render/simulation -- and things get very sickening... I then am stuck in this state even if I put the block down. In fact I am stuck in this mode in "Oculus" apparently permanently - but if I switch into 2D (with n-vsynch) or Oculus test mode, I can render/simulate at >100hz. - So this appears to be some kind of a problem specific to oculus.

interface/src/Application.cpp
```diff
@@ -982,9 +980,6 @@ void Application::initializeGL() {
     connect(&checkFPStimer, &QTimer::timeout, this, &Application::checkFPS);
```
ZappoMan: we should consider getting rid of this "checkFPStimer" as well and just put the FPS calculations into idle() - this being on a Qt timer would explain why we see a disparity between FPS and simulation hz even when we know they should be in synch with this PR.

PhilipRosedale:
Interesting the dropping FPS... agree @ZappoMan we should figure out what is going on there.  @birarda can you guys test this PR on the Vive?

birarda:
This is identical to the change we made to test this approach last week. Good to go on Vive.

I agree that there are cases where this will produce a more nauseating experience in the short term - I think the benefits we've seen for Baseball and what this will allow outweigh those cases in the short term (and it seems like you're saying the same @ZappoMan if we can have a 4-5 person experience hold up).

PhilipRosedale:
I don't see the changes here that would cause the physics engine to step at 75Hz or 90Hz ... isn't it tied to 60Hz max right now?

birarda:
Yes @PhilipRosedale is correct that needs to be added to this PR.

birarda:
@Atlante45 anything else we changed that is missing here?

Atlante45:
@hyperlogic @birarda Yes, we set `PHYSICS_ENGINE_FIXED_SUBSTEP` to `1.0f / 90.0f` instead of `1.0f / 60.0f` in PhysicsHelpers.h

ZappoMan:
this simulates at vsynch and so on vive it will be 90hz

Atlante45:
Also we removed the bail on throttle at 15 FPS, but I think we should keep that.

Atlante45:
@ZappoMan idle will run at 90 FPS but the PhysicsEngine doesn't do anything if a subset hasn't passed yet, so we're actually simulating at 90 FPS.

ZappoMan:
Let's make the following changes before merge:
* remove the unneeded constant (Clement's suggestion)
* fix fps calculation to not be on a timer
* vive display plugin shouldn't have "throttle" this is a plugin issue. so we don't need to change application for that
* check on the PHYSICS_ENGINE_FIXED_SUBSTEP issue with Andrew

Atlante45:
@ZappoMan Talked with Andrew about the substep, he said that it might become a problem when we get to really big scene (A `Physics Bomb` in Second Life terms).
But for now it's good and it's a step foreward. There are ways to deal with those physic bombs that we can implement when we get there.

hifi-gustavo:
Build console output for commit: c139b62217742ce96ee43c712db6e5c40f0cc012 available at: https://gist.github.com/7f2b50e9ec54522187be<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6283.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6283.dmg

hifi-gustavo:
Build console output for commit: 31f361d49ca445ce84ca7c0a3ca8875554f42a7a available at: https://gist.github.com/7f2b50e9ec54522187be<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6283.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6283.dmg

interface/src/Application.cpp
```diff
@@ -1048,6 +1041,26 @@ void Application::initializeUi() {
 }
 
 void Application::paintGL() {
+    _frameCount++;
+
+    // update fps moving average
+    uint64_t now = usecTimestampNow();
+    static uint64_t lastPaintBegin{ now };
```
ZappoMan: I realize the coding standard is silent on this issue, you can use whichever initializer you want in code. But it seems strange to me to use an "=" on line 1047, and then turn around and use { } on the very next line. Since these are both plain old data types, the compiler is going to produce the same code either way... for consistency why don't we use `=` since it's more readable, and there's no impact. If you really really really want to use `{  }` then why not use it on the line before as well.

interface/src/Application.h
```diff
@@ -541,6 +541,8 @@ private slots:
     EntityItemID _keyboardFocusedItem;
     quint64 _lastAcceptedKeyPress = 0;
 
+    SimpleMovingAverage _framesPerSecond{10};
+    quint64 _lastFramesPerSecondUpdate = 0;
```
ZappoMan: not a merge blocker -- technically the coding standard says to prefer a `{ }` here

PR #6287, opened 2015-11-03
Removing iviewhmd due to broken production builds

jherico:
The `iviewhmd.dll` we package in the installer depends on MSVCR110.DLL and MSVCP110.DLL, but these are not getting packaged into the installer.  I'm disabling this to prevent users from getting builds that can't run.

### DIFF
--- a/interface/CMakeLists.txt
+++ b/interface/CMakeLists.txt
@@ -2,7 +2,7 @@ set(TARGET_NAME interface)
 project(${TARGET_NAME})
 
 # set a default root dir for each of our optional externals if it was not passed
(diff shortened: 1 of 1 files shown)


ZappoMan:
You should merge this. I am on mobile or else I would merge it.

hifi-gustavo:
Build console output for commit: eeabc9b093384731431df63c8efaaf30ef8ade57 available at: https://gist.github.com/e88f56fe5b6ed63edc61<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6287.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6287.dmg

PR #6289, opened 2015-11-03
Fix: Vec3.UP was pointed to Vectors::UNIT_X

thoys:
This fixes the scripting property Vec3.UP to point to Vectors::UP which is equal to {X: 0, Y: 1, Z: 0}.

PR #6290, opened 2015-11-03
Add growth script to examples

zzmp:
<s>@PhilipRosedale the spheres added with `Vec3.UP` do not show up. Do you know why this is?</s>

### DIFF
--- a/examples/growth.js
+++ b/examples/growth.js
@@ -8,32 +8,79 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 var RADIUS = 5;
-var LIFETIME = 6000;
+var LIFETIME = 60; // only keep these entities around for a minute
 
-var size = 1;
+var SIZE = 1; // starting size
+var SIZE_MAX = 3;
+var SIZE_MIN = 0.5;
+
 var center = MyAvatar.position;
 var positions = [
     Vec3.sum(center, Vec3.multiply(RADIUS, Vec3.FRONT)),
     Vec3.sum(center, Vec3.multiply(-RADIUS, Vec3.FRONT)),
     Vec3.sum(center, Vec3.multiply(RADIUS, Vec3.RIGHT)),
     Vec3.sum(center, Vec3.multiply(-RADIUS, Vec3.RIGHT)),
-    Vec3.sum(center, Vec3.multiply(RADIUS, Vec3.UP)),
-    Vec3.sum(center, Vec3.multiply(-RADIUS, Vec3.UP)),
 ];
 
+// Create spheres around the avatar
 var spheres = map(positions, getSphere);
 
-Script.update.connect(function(delta) {
-    size += delta; // grow by 1 unit/s
-    each(spheres, function(sphere) {
-        Entities.editEntity(sphere, { dimensions: getDimensions(size) });
-    });
+// Toggle sphere growth
+each(spheres, function(sphere) {
+    var state = {
+        size: SIZE,
+        grow: true, // grow/shrink
+        state: false, // mutate/pause
+    };
+
+    // Grow or shrink on click
+    Script.addEventHandler(sphere, 'clickReleaseOnEntity', toggleGrowth);
+    // TODO: Toggle on collisions as well
+
+    function toggleGrowth() {
+        if (state.state) {
+            stop();
+        } else {
+            start();
+        }
+    }
+
+    function start() {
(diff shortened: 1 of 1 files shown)


hifi-gustavo:

thoys:
@zzmp I made a Fix for this an half hour ago. Vec3.UP was connected to the X positive value before.

thoys:
see https://github.com/highfidelity/hifi/pull/6289

zzmp:
Cool! I've updated the code to take out the question. Thanks @thoys.

examples/growth.js
```diff
@@ -0,0 +1,76 @@
...
+    Vec3.sum(center, Vec3.multiply(-RADIUS, Vec3.FRONT)),
+    Vec3.sum(center, Vec3.multiply(RADIUS, Vec3.RIGHT)),
+    Vec3.sum(center, Vec3.multiply(-RADIUS, Vec3.RIGHT)),
+    Vec3.sum(center, Vec3.multiply(RADIUS, Vec3.UP)),
+    Vec3.sum(center, Vec3.multiply(-RADIUS, Vec3.UP)),
+];
+
+var spheres = map(positions, getSphere);
+
+Script.update.connect(function() {
```
jherico: this will cause the spheres to grow at an uneven rate, depending on how often this is called.  The function you define here will receive a deltaTime value that contains the interval since the last time the update function was called, allowing you to tie the speed of growth to actual time.

zzmp:
@jherico good idea. I've updated it to use the `delta` given to `Script.update.connect`'s callback such that it grows by 1 unit/s.

imgntn:
@zzmp i think this would be more interesting if the spheres started shrinking once they touch your avatar.  being inside of the spheres is not a very good experience... you don't really see anything.  additionally, it would show people how to toggle the directionality of that growth

zzmp:
@imgntn, I agree that the experience would be better, and it would be more interesting as an interactive example. I can rework it with that in mind.

What did you mean by "directionality of that growth"?

imgntn:
just that its interesting to see it flip from growing to shrinking at a certain point.  the cyclicality would be nice for people looking to just copy / paste a script and see some stuff happen

zzmp:
Is there an easy way to detect a collision event/signal, for the sake of adding this interactivity? @ericrius1 do you know of one? Is there something I can `Entities.???.connect` to?

I looked through the documentation and nothing seemed to fit the bill.

huffman:
@zzmp You can do something like this:

```
Script.addEventHandler(entityID, "collisionWithEntity", function(entityA, entityB, collision) {
    // handle collision
});
```

The collision object has these properties:
```
    ContactEventType type;
    QUuid idA;
    QUuid idB;
    glm::vec3 contactPoint;
    glm::vec3 penetration;
    glm::vec3 velocityChange;
```

zzmp:
OK, I've implemented `clickReleaseOnEntity`, but could not get it to work with `collisionWithEntity`. It will now transition from a growing->static->shrinking->static->... state with each click.

**However, I noticed a bug**, which I was able to reproduce, where running the script subsequent times causes the `collisionWithEntity` event to fire off N times, where N is the number of times I've run the script (having waited at lease an entity lifetime between each run). I included instructions to reproduce it in the commit message for 33851ba.
Is this bug known at all? It seems to be related more to Interface than to the `examples/growth.js` script. I'll investigate it more later if I don't hear back from someone with prior knowledge.

imgntn:
@zzmp i think the reason collisionWithEntity is not registering is because the avatar doesn't count as an entity, for the moment.  so if you were trying to run into a sphere with your avatar and produce a collision, for example, it wouldn't work.  should work when the spheres hit each other though.  still checking out your other bug

imgntn:
@zzmp  i can't reproduce this below.  i set a short lifetime (2 sec) and clicked on one, then ctrl-r to refresh scripts, and it works the same.  when you say 'rerun a second time', what do you mean, specifically?  thanks!  

*There is a bug in here, but it may be in Interface.*
The sphere entities have a lifetime of 60s.
The first time the script is run, each click fires one clickReleaseOnEntity event.
If the entities expire (lifetime is reached), and the script is rerun a second time,
each click will fire two clickReleaseOnEntity events.
If this is repeated (waiting for expiration each time), then
the N-th repetition will fire N clickReleaseOnEntity events.
Note that each entity, across runs, has a unique uuid, as expected.


p.s. you can also manually remove event handlers by calling  Script.removeEventHandler(sphere, 'clickReleaseOnEntity', toggleGrowth);

PR #6293, opened 2015-11-03
Disabling billboard rendering, source of the evil black rectangles

jherico:
The billboard logic seems to be messed up somehow and is triggering when it shouldn't.  Disabling for now until we can resolve billboarding properly.


hifi-gustavo:
Build console output for commit: 4c076d8b128d1984dca520ec5c8bba28ca56a333 available at: https://gist.github.com/ce7f49bf36f3a92bd27b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6293.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6293.dmg

PR #6294, opened 2015-11-03
fix hydra sensor matrix when no HMD is attached

AndrewMeadows:
This is the right fix in the short (initialize _bodySensorMatrix to a reasonable default value) but  the long term solution is to compute a default _bodySensorMatrix when we load the avatar model.

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -122,11 +122,7 @@ MyAvatar::MyAvatar(RigPointer rig) :
             this, &MyAvatar::goToLocation);
     _characterController.setEnabled(true);
 
-    _hmdSensorMatrix = glm::mat4();
-    _hmdSensorPosition = extractTranslation(_hmdSensorMatrix);
-    _hmdSensorOrientation = glm::quat_cast(_hmdSensorMatrix);
     _bodySensorMatrix = deriveBodyFromHMDSensor();
-    updateSensorToWorldMatrix();
 }
 
 MyAvatar::~MyAvatar() {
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 12ff9734e05391cfe3124cd3f24c6fe4a1ffbf8c available at: https://gist.github.com/88e37f6a84aefaeed3d2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6294.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6294.dmg

hifi-gustavo:
Build console output for commit: e160e2a7ae88a1d7ca632fa6c15ad69196811987 available at: https://gist.github.com/88e37f6a84aefaeed3d2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6294.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6294.dmg

AndrewMeadows:
Seth tested this branch and found that there is still a problem when he uses the Albert model for his avatar.  I realize now that the default _bodySensorMatrix is jut not going to work for all avatars.  To fix it correctly I'm going to have to update the matrix to be more correct whenever the avatar is loaded.

ZappoMan:
did you mean to merge this? after your last comment it seemed like you had a different strategy

AndrewMeadows:
No, merging this was an accident, but it isn't a terrible one.  I'll have to fix the "avatars come in many shapes" problem in another PR.

PR #6296, opened 2015-11-03
fix bug preventing avatar from walking when HMD is at rest

AndrewMeadows:
When the HMD is "at rest" then the "resolve hips offset" code will kick in constantly and prevent the avatar from walking or turning.  This is a quick fix for that bug.

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -347,23 +347,6 @@ void MyAvatar::updateFromHMDSensorMatrix(const glm::mat4& hmdSensorMatrix) {
 }
 
 void MyAvatar::updateHMDFollowVelocity() {
-    bool isMoving;
-    if (_lastIsMoving) {
-        const float MOVE_EXIT_SPEED_THRESHOLD = 0.07f;  // m/sec
-        isMoving = glm::length(_velocity) >= MOVE_EXIT_SPEED_THRESHOLD;
-    } else {
-        const float MOVE_ENTER_SPEED_THRESHOLD = 0.2f; // m/sec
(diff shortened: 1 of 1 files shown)


interface/src/avatar/MyAvatar.cpp
```diff
@@ -347,6 +347,16 @@ void MyAvatar::updateFromHMDSensorMatrix(const glm::mat4& hmdSensorMatrix) {
...
 void MyAvatar::updateHMDFollowVelocity() {
+    // compute offset to body's target position (in sensor-frame)
+    auto sensorBodyMatrix = deriveBodyFromHMDSensor();
+    _hmdFollowOffset = extractTranslation(sensorBodyMatrix) - extractTranslation(_bodySensorMatrix);
+    glm::vec3 truncatedOffset = _hmdFollowOffset;
+    if (truncatedOffset.y < 0.0f) {
+        // don't pull the body DOWN to match the target (allow animation system to squat)
+        truncatedOffset.y = 0.0f;
+    }
+    float truncatedOffsetDistance = glm::length(truncatedOffset);
```
AndrewMeadows: This code move up from below so that we would have the **truncatedOffset** variable in context when trying to decide if we want to start resolving the hips offset for HMD at rest.  This prevents the offset resolution code from constantly kicking in.

hifi-gustavo:
Build console output for commit: d544b7a645a779d95caed858b36b1440a5f9dc5a available at: https://gist.github.com/8fe4744cf00010ed66a7<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6296.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6296.dmg

PR #6304, opened 2015-11-04
Controllers: Fixing the step pulse behavior and changes to standard mapping

samcake:
- Remove the pulsing behavior hardocded in MyAvatar for the stepYaw and instead rely purely on the pulsing of the controller route to the StepYaw axis.

- Updated the standard mapping:
  strafe is back on the left x joystick
  removed vertical moves completely

### DIFF
--- a/interface/resources/controllers/standard.json
+++ b/interface/resources/controllers/standard.json
@@ -2,8 +2,9 @@
     "name": "Standard to Action",
     "channels": [
         { "from": "Standard.LY", "to": "Actions.TranslateZ" },
+        { "from": "Standard.LX", "to": "Actions.TranslateX" },
 
-        { "from": "Standard.LX",
+        { "from": "Standard.RX",
           "when": [ "Application.InHMD", "Application.ComfortMode" ],
           "to": "Actions.StepYaw",
           "filters":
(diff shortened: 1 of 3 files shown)


interface/resources/controllers/standard.json
```diff
@@ -2,21 +2,19 @@
...
 
-        { "from": "Standard.LX",
+        { "from": "Standard.RX",
           "when": [ "Application.InHMD", "Application.ComfortMode" ],
           "to": "Actions.StepYaw",
           "filters":
             [
                 { "type": "pulse", "interval": 0.5 },
-                { "type": "scale", "scale": 15 }
+                { "type": "scale", "scale": 22.5 }
```
samcake: Adjusted to 22.5 degrees after confirming with Philip

jherico: make sure phillip understands that because RX is an analog joystick, the actual degrees turned will be scaled by how far he pushes the joystick.

If he wants any push on the joystick to always snap to exactly 22.5 degrees no matter how hard you push, we need a constraint to integer filter, and we should probably have a small deadzone as well.

samcake: ok, let s test that, i can fool around with the values from the mapping with him validating

hifi-gustavo:
Build console output for commit: d4259931b23f3afedb49a0f9313095754fda78dc available at: https://gist.github.com/9adcb36c36e9d4c4ed23<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6304.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6304.dmg

PR #6309, opened 2015-11-04
CR for Job #20724 - Camera entity mode

thoys:
Adds ability to have camera follow an entity.

Code Review for Job #20724 - Workitem available at https://www.worklist.net/20724


hifi-gustavo:
Build console output for commit: 42cbccc725191bdf922e8ccb8ca7d6b3fad621ca available at: https://gist.github.com/98f0a80fd06fc8fe6c31<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6309.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6309.dmg

cozza13:
this is great.

hifi-gustavo:
Build console output for commit: de98a17ada0c21f57bfffe92a45bd67326f38399 available at: https://gist.github.com/98f0a80fd06fc8fe6c31<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6309.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6309.dmg

ZappoMan:
@thoys - this is really interesting... but I'm unclear on why this is needed when we already have independent camera mode? What are the specific limitations of independent camera mode that prevent you from just having a camera follow along? Is it because you can't keep the camera in-synch? (that would make sense) Is it because it's hard to move/animate a camera in a script and easier to just give velocity to an entity as a way of moving things? (that would also make sense to me)...

libraries/entities/src/EntityTypes.cpp
```diff
@@ -48,7 +49,8 @@ REGISTER_ENTITY_TYPE(ParticleEffect)
 REGISTER_ENTITY_TYPE(Zone)
 REGISTER_ENTITY_TYPE(Line)
 REGISTER_ENTITY_TYPE(PolyVox)
-REGISTER_ENTITY_TYPE(PolyLine);
+REGISTER_ENTITY_TYPE(PolyLine)
+REGISTER_ENTITY_TYPE(Anchor);
```
ZappoMan: Is it a Camera or an Anchor? I think Anchor is not a good name, since it can be given velocity, etc... which I believe is the point. So it seems to me that Camera is a better name.

ZappoMan:
Since you're introducing a new entity type, you need to make sure to bump the protocol version so that clients who don't know how to instantiate Cameras will be upgraded before connecting to domains that might have cameras.

ZappoMan:
I don't see any code here to support what happens if the Camera mode has been changed to "entity camera" and then one of the following things happen:
- The entity is deleted locally
- The entity is deleted by some remote editor
- The entity's lifetime expires
- The user changes domains

These cases need to be handled (the first three are all rooted in the entity being deleted, and can be solved with one change to handle all three cases).

libraries/entities/src/EntityTypes.cpp
sethalves: I asked him to rename it, because I expect to use it for "groups" as well.  There's nothing camera specific about it -- it' s just a non-rendering entity with a location and velocity.  I guess it's more of an AnchorTarget.

ZappoMan: Ok... I can see that. It still seems weird to call it an "Anchor" when the main feature it provides is physical moving behavior .... no?

thoys:
@ZappoMan I am not sure if we should switch mode when a entity that is being followed by the camera is deleted. To which mode would you fall back? What happens if it was only a  restart and the entity stays in front of you after restarting.

ZappoMan:
@thoys - my point is we need SOME behavior... I'm not sure what the right behavior is... but I know we need to do something.

thoys:
What happens now is that the camera stays where it was last moved to. But I see that it will freeze HMD's maybe to just fix that? The script which enables the camera is expected to return to a normal camera mode after the scripted camera functionality is stopped.

hifi-gustavo:
Build console output for commit: 7f6a20af99b3a118475c2c4dc87a9c507e55b533 available at: https://gist.github.com/98f0a80fd06fc8fe6c31<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6309.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6309.dmg

thoys:
@ZappoMan ^ ?

ZappoMan:
@thoys - after talking more with @sethalves about this - we would like you to remove the Anchor entity type from this PR... it's not really needed for your entity following camera (since you can actually follow any type of entity with the application/camera mode code you have written)... this way you don't need to add the new type, you don't need to bump the protocol number, etc.

Can you remove the Anchor stuff from this PR please? Thanks.

thoys:
@ZappoMan no problem. I understand. I firstly created the entity Camera specific because we might some more properties regarding to the camera and make it more friendly. But was being told to replace it to Anchor for the parenting PR. I will just use an invisible box entity for the example.

examples/libraries/entityPropertyDialogBox.js
```diff
@@ -260,6 +260,10 @@ EntityPropertyDialogBox = (function () {
             array.push({ label: "Cutoff (in degrees):", value: properties.cutoff });
             index++;
         }
+
+        array.push({ label: "", type: "inlineButton", buttonLabel: "Preview Camera", name: "previewCamera" });
```
ZappoMan: do you still need this?

thoys:
@ZappoMan , I don't seem to find your "do you need this?" for the preview button in the edit.js comment which I got in an e-mail. Did you remove it? In case not, the button will allow you to see what the camera will see when its attached to that entity. It might be handy, but its not a requirement.

hifi-gustavo:
Build console output for commit: 4c2c37acce3909a125f935c6f555cb0544fba95c available at: https://gist.github.com/98f0a80fd06fc8fe6c31<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6309.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6309.dmg

ZappoMan:
Yes @thoys - I realized after I left that comment, what that was doing, and so I removed the comment. I will test the PR some later today and merge.

thoys:
Thanks for renaming the PR, forgot to do that

PR #6314, opened 2015-11-04
Fixing step yaw to move by consistent amounts, be 'tappable'

jherico:
This lets you tap the keyboard to turn faster than the normal interval of 0.5 seconds.  It also fixes the angle turned which was previously somewhat random due to incorrectly being divided by a timeslice (which isn't a valid thing to do for pulsed inputs).

### DIFF
--- a/interface/resources/controllers/keyboardMouse.json
+++ b/interface/resources/controllers/keyboardMouse.json
@@ -12,28 +12,49 @@
         { "from": "Keyboard.W", "when": "Keyboard.Shift", "to": "Actions.PITCH_UP" },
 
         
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: 20d1035821b83b5d4d29bbf3ca94d4db09a15498 available at: https://gist.github.com/d691d494f9794d5829a2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6314.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6314.dmg

PR #6317, opened 2015-11-05
Added some documentation and examples for the new controller mapping

AlessandroSigna:
This JS provides some examples to better understand the new controller mapping system


imgntn:
this is great; however it's a little confusing next to rightClickExample.js

we should probably just provide one example with both mapping strategies, and explain why you'd want to use one over the other in different situations.  

here's what i see:  

rightclickexample uses Controller.newMapping and then chains .from and .to calls w/ enableMapping and disableMapping

while this example uses a function that returns a JSON structure with to and from, then calls parseMapping and then enable directly on the mapping that this returns.

also, is the header correct?  this doesn't seem to be from 6/2/2015

AlessandroSigna:
I updated the old controllerMapping.js because it didn't work anymore and was removed. I didn't touch its header, do you think I have to modify it?
 
Both the example basically do the same things but i think that the rightClickExample is good for one or two inputs/actions modifications. 
This one instead is very good and quick for a lot of new mappings which you can write in the JSON body and then enable.

However I can include also that example in this file as you suggested, it's a good idea.

imgntn:
@AlessandroSigna yeah -- in your file, why don't you do one the first mapping the way it is now, and do the second one in the rightClickExample style?  then we can delete the rightClickExample and it will all be in one place.  i think that would be nice.  and you can add yourself to the header after sam, since its a pretty substantial rewrite!

PR #6318, opened 2015-11-05
grab changes

sethalves:
- rename non-colliding-grab to trigger
- fix whiteboard and some other scripts

### DIFF
--- a/examples/toybox/lights/lightSwitch.js
+++ b/examples/toybox/lights/lightSwitch.js
@@ -33,7 +33,7 @@
             this.toggleLights();
         },
 
-        startNearGrabNonColliding: function () {
+        startNearTrigger: function () {
             this.toggleLights();
         },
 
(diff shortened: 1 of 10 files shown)


examples/toybox/lights/lightSwitch.js
```diff
@@ -61,7 +61,8 @@
             setEntityCustomData(this.resetKey, this.entityID, {
```
imgntn: I don't think this is going to work, since it's setting the grabbableKey on the reset Key.  should be
```
setEntityCustomData(this.resetKey, this.entityID, {
                on: on,
                type: lightType,
                resetMe: true
});

setEntityCustomData('grabbableKey', this.entityID, {wantsTrigger:true});
```

sethalves: we sort of merged this to test it, I probably made serveral mistakes.  make a PR and I'll merge it

PR #6320, opened 2015-11-05
IK fix for avatars exported from Blender

hyperlogic:
This should fix the issue with the hips moving erratically when arm IK
is enabled.  The main issue is that the IK system assumed that the "Hips"
joint was the root of the skeleton.  For Blender avatar this is not the case
as it inserts an "Armature" node at the root instead.


hifi-gustavo:
Build console output for commit: 20d95080f1c0a6a252e1ee7e4364451c93db672c available at: https://gist.github.com/c20b207558cffedfe091<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6320.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6320.dmg

PR #6326, opened 2015-11-06
Add Ability to Pass Parameters When Calling Methods on Entities

imgntn:
This PR adds some useful functionality to entities.  Specifically, when using Entities.callEntityMethod users can now pass an array of parameters with the call that can be accessed from the callee.  

``` Entities.callEntityMethod('entityID','someAwesomeMethod',['array','of','parameters'] ```

and on the entity...

``` someAwesomeMethod:function(myID,parameters){print('parameters are:' + parameters};```

### DIFF
--- a/libraries/entities/src/EntitiesScriptEngineProvider.h
+++ b/libraries/entities/src/EntitiesScriptEngineProvider.h
@@ -20,6 +20,7 @@
 class EntitiesScriptEngineProvider {
 public:
     virtual void callEntityScriptMethod(const EntityItemID& entityID, const QString& methodName) = 0;
+    virtual void callEntityScriptMethod(const EntityItemID& entityID, const QString& methodName, const QStringList& params) = 0;
 };
 
 #endif // hifi_EntitiesScriptEngineProvider_h
--- a/libraries/entities/src/EntitiesScriptEngineProvider.h
+++ b/libraries/entities/src/EntitiesScriptEngineProvider.h
@@ -19,8 +19,7 @@
 
 class EntitiesScriptEngineProvider {
 public:
-    virtual void callEntityScriptMethod(const EntityItemID& entityID, const QString& methodName) = 0;
-    virtual void callEntityScriptMethod(const EntityItemID& entityID, const QString& methodName, const QStringList& params) = 0;
+    virtual void callEntityScriptMethod(const EntityItemID& entityID, const QString& methodName, const QStringList& params = QStringList()) = 0;
 };
 
 #endif // hifi_EntitiesScriptEngineProvider_h
--- a/libraries/script-engine/src/ScriptEngine.cpp
+++ b/libraries/script-engine/src/ScriptEngine.cpp
@@ -1200,6 +1200,37 @@ void ScriptEngine::callEntityScriptMethod(const EntityItemID& entityID, const QS
     }
 }
 
+void ScriptEngine::callEntityScriptMethod(const EntityItemID& entityID, const QString& methodName, const QStringList& params) {
+    if (QThread::currentThread() != thread()) {
+        #ifdef THREAD_DEBUGGING
+        qDebug() << "*** WARNING *** ScriptEngine::callEntityScriptMethod() called on wrong thread [" << QThread::currentThread() << "], invoking on correct thread [" << thread() << "]  "
+            "entityID:" << entityID << "methodName:" << methodName;
+        #endif
+
+        QMetaObject::invokeMethod(this, "callEntityScriptMethod",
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by hifi-gustavo** [comment]
Build console output for commit: 3cee569baefd46d9ed8c44fc005f0e558f40a63b available at: https://gist.github.com/1f024be31fd60a5c8334<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6326.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6326.dmg

File: libraries/entities/src/EntitiesScriptEngineProvider.h [hunk 1]
```diff
@@ -20,6 +20,7 @@
 class EntitiesScriptEngineProvider {
 public:
     virtual void callEntityScriptMethod(const EntityItemID& entityID, const QString& methodName) = 0;
+    virtual void callEntityScriptMethod(const EntityItemID& entityID, const QString& methodName, const QStringList& params) = 0;
```
jherico: is there any reason to have both these functions, instead of just one with a default empty list for params?

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1200,6 +1201,38 @@ void ScriptEngine::callEntityScriptMethod(const EntityItemID& entityID, const QS
     }
 }
 
+void ScriptEngine::callEntityScriptMethod(const EntityItemID& entityID, const QString& methodName, const QStringList& params) {
```
jherico: could we add some test scripts for demonstrating this functionality?  I'm a little concerned because I've seen issues in the past with overloaded functions exposed to JS.

jherico:
test this please

hifi-gustavo:
Build console output for commit: 3cee569baefd46d9ed8c44fc005f0e558f40a63b available at: https://gist.github.com/1f024be31fd60a5c8334<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6326.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6326.dmg

libraries/script-engine/src/ScriptEngine.cpp
imgntn: you mean a .js example script, right?  not a unit test?  if so, i got it :)

libraries/entities/src/EntitiesScriptEngineProvider.h
imgntn: i'm not very familiar with C++ but i was getting 'too many parameters for function' until i added this -- seems like theres a lot of repeating functions so i just copied that pattern.  happy to combine them if there's a better way though!  what would that look like @jherico

imgntn:
@jherico got a little help from @ericrius1 and overloaded the function like you suggested.  added some examples too.

hifi-gustavo:
Build console output for commit: e928c8278b2acfbb968a661b6558e63d2eaba734 available at: https://gist.github.com/1f024be31fd60a5c8334<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6326.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6326.dmg

hifi-gustavo:
Build console output for commit: cf86d27417716ca48a3dd7805730870e920f30e8 available at: https://gist.github.com/1f024be31fd60a5c8334<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6326.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6326.dmg

imgntn:
@ZappoMan is this one good to go or something else i should do here / a better pattern?

ZappoMan:
@imgntn - I think this pattern is reasonable. I was wondering if there was a way to more generically support other data types than just strings (like for example passing a vec3, quat, etc), and it probably could be done similar to what you have here. But if "just strings" is sufficient for now, then this approach seems reasonable and is a stepwise enhancement.

I didn't test it, so assuming someone else has tested and verified that other uses of calling entity methods are still working properly than yes, I think this is good to go.

imgntn:
@ZappoMan yeah it would be great to natively support some other types / do the object serialization on the C++ side so you don't have to stringify / parse stuff out, but i think strings are good enough for now. i'll iterate.  thanks for the input!

PR #6327, opened 2015-11-06
Examples of  the new input mapping system

AlessandroSigna:
This JS presents two ways to use the new mapping system and some useful comments regarding to it


imgntn:
nice, i like these updates.

PR #6331, opened 2015-11-06
Changes to Standard.RY

AlessandroSigna:
Now Y axis of the right hand controller controls the jump/fly action


hifi-gustavo:
Build console output for commit: fdd4a550571d3cf7527787a371600a50eeea10bc available at: https://gist.github.com/8478a810170e51c5cb11<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6331.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6331.dmg

jherico:
In the future you may want to consider rebasing your branch off master rather than merging, so as to avoid creating so many merge commits.

interface/src/Application.cpp
```diff
@@ -646,6 +646,9 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
     _applicationStateDevice->addInputVariant(QString("ComfortMode"), controller::StateController::ReadLambda([]() -> float {
         return (float)Menu::getInstance()->isOptionChecked(MenuOption::ComfortMode);
     }));
+	_applicationStateDevice->addInputVariant(QString("Grounded"), controller::StateController::ReadLambda([]() -> float {
```
ZappoMan: note: the indent looks broken here... is your editor set to TABs?

PR #6347, opened 2015-11-07
Set clip for ControlledAC

EdgarPironti:


hifi-gustavo:

ZappoMan:
add to whitelist

examples/acScripts/ControlledAC.js
```diff
@@ -141,6 +145,12 @@ function update(event) {
             }
             Agent.isAvatar = false;
             break;
+        case LOAD:
+            print("Load");            
+            if(filename !== null) {
+                Avatar.loadRecording(filename);
```
ZappoMan: I don't see how this is going to work. The "AC script" e.g. `examples/acScripts/ControlledAC.js` is likely going to be running on a totally different machine, and so the "file name" will not be useful to it. Maybe we should just let people access a "URL" and/or use the Asset server to store the script.

EdgarPironti: The ControlACs.js passes information to the ControlledAC.js through an entity that is considered the "channel" between the two (it must have the same size and position). Until now, the only thing that was passed was an integer "hidden" in the color information, now I pass the filename/url through the name of the entity. The same mechanism can be used to set a url manually I think.

ZappoMan: My point is that if you tell the AC script to load 'C:\path\to\filename.ext' that that file won't exist in the file system on the AC and so it won't work. You will need to make sure that the "file" is available via a url.

ZappoMan: Also: you should use the userData property to pass this info. That's better than the name property.

examples/acScripts/ControlledAC.js
```diff
@@ -21,8 +20,6 @@ var useAvatarModel = true;
 // ID of the agent. Two agents can't have the same ID.
 var id = 0;
 
-// Set avatar model URL
-Avatar.skeletonModelURL = "https://hifi-public.s3.amazonaws.com/marketplace/contents/e21c0b95-e502-4d15-8c41-ea2fc40f1125/3585ddf674869a67d31d5964f7b52de1.fst?1427169998";
```
jherico: why are we no longer setting the model?  are we planning on relying on the recorded model?

samcake: Yes that is the plan, to load the model from the clip
or maybe from ajust a simple  "HEader frame" describing the avatar (model
and starting position
what clement had with the  the "Recording context" i think

Does that works for you ?

PR #6351, opened 2015-11-09
Update script engine path resolution behavior to behave consistently

huffman:
Do not merge yet, feedback welcome.

The path resolution will now be relative to the script currently
being evaluated *on its initial evaluation.* The previous behavior
was that all paths would be resolved relative to the root script
for client scripts, and inconsistent for entity scripts depending
on the order that scripts were loaded. The entity script situation
was particularly bad because including more than 1 level deep produced
inconsistent results.

### DIFF
--- a/examples/example/games/satellite.js
+++ b/examples/example/games/satellite.js
@@ -15,7 +15,7 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 
-Script.include('../utilities/tools/vector.js');
+Script.include('../../utilities/tools/vector.js');
 
 var URL = "https://s3.amazonaws.com/hifi-public/marketplace/hificontent/Scripts/planets/";
 
--- a/examples/libraries/entityCameraTool.js
+++ b/examples/libraries/entityCameraTool.js
@@ -9,7 +9,7 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 
-Script.include("libraries/overlayUtils.js");
+Script.include("overlayUtils.js");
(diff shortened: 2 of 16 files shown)


hifi-gustavo:
Build console output for commit: b8357112660a0955077056d138acbe1e9d5e9d81 available at: https://gist.github.com/0679dc861f27934bef6e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6351.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6351.dmg

ZappoMan:
I haven't tested this... but this is similar to an idea I suggested we do in the past. The one test case to consider is includes inside of includes inside of includes. I think your fix should support that case, if so I'm all in favor of this change.

huffman:
Yep, the " includes inside of includes inside of include" is actually what I was originally trying to solve, so I will be testing that. I will remove the do-not-merge once I do more testing. We definitely shouldn't merge this before the demos today.

imgntn:
there are likely some scripts that reference things relative to the root script that will need to be updated as well -- off the top of my head i think magBalls does this but i'm not sure about other

hifi-gustavo:
Build console output for commit: 08c27599472978cce6310e9ea4275efc94f1d203 available at: https://gist.github.com/e3e18b6bd2b377df42e4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6351.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6351.dmg

PR #6355, opened 2015-11-09
don't attempt to near-grab things that aren't physical

sethalves:
- fix a bug that made the grab script sometimes attempt to near-grab unlocked non-physical things


imgntn:
tested win10 lgtm

PR #6359, opened 2015-11-10
fix render item leak on Model::reset()

ZappoMan:


libraries/render-utils/src/Model.cpp
```diff
@@ -135,9 +135,6 @@ void Model::reset() {
         const FBXGeometry& geometry = _geometry->getFBXGeometry();
         _rig->reset(geometry.joints);
     }
-    _meshGroupsKnown = false;
```
ZappoMan: if `_meshGroupsKnown` is false, then other helper functions will incorrectly re-add the render items to the scene. There's also no reason to invalidate the mesh boxes in this case either.

jherico: It's nice that this fix is so small, but it feels like it's correcting a symptom, but not the root cause.  If the helpers are capable of re-adding the render items to the scene should they not also be able to recognize that they've added items already and clear them out first?  

I'll accept that this may be out of scope, but we may at least want to add a `FIXME` in the appropriate spot.

hifi-gustavo:
Build console output for commit: e47b049200c6679fb266a99777f229d77796611a available at: https://gist.github.com/2edff51e2b4e2800d49f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6359.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6359.dmg

libraries/render-utils/src/Model.cpp
ZappoMan: The helpers are told to re-add based on these values... is all I'm saying... maybe I could explain it better.

ZappoMan: Let me explain this better... 

The member variable `_meshGroupsKnown` should only be initialized to false when a Model is first created (and/or if it's URL) is changed. 

If `_meshGroupsKnown` then `Model::segregateMeshGroups()` will get called to do this extra initialization and calculation work. Some of that work is to create the `MeshPartPayload` render items.

I agree that `Model.cpp` is a mess and this is one example of a fragile piece of code in it... but we really need to junk the whole damn thing... which is clearly out of scope of fixing this memory leak. 

I will add some ASSERTS to the `segregateMeshGroups` that test to make sure it's not being called after items have already been added to the scene (which should not happen)!

hifi-gustavo:
Build console output for commit: 021a6e689d92c26a4d2a84c6f0426bc592d8b4e7 available at: https://gist.github.com/2edff51e2b4e2800d49f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6359.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6359.dmg

PR #6360, opened 2015-11-10
Recording and playback of the Avatar animation update but not quite over

samcake:
- This is an update of the recording and playback of the skeletal animation of the avatar.

- Remove the old code about the "old" rec format

- Remove the separate jointRotations and jointTranslations by the equivalent of the skeleton joint state as exposed by the AvatarData: QVector<JointData> that i call "JointArray"
- Replaced the serializing code to use the JOintArray

- In the Recorder use the newly exposed getRawJointData to capture the full joint information and generate a frame. 
- In the PLayer after trying to use the newly exposed setRawJoitnData, we found a temporary workaround using the deprecated setJointRotations setJointTranslations functions which do the "right" thing for the specialized version of AvatarData which are MyAvatar and ScriptableAvatar. 
This need to be fixed, i m planning to work with the Animation team on this.

THis is not perfect but allow us to get moving on updating the full recording/playback use case.


hifi-gustavo:
Build console output for commit: b47c5dbff2f8d1d20e3378014d99efa6cf3c8fd0 available at: https://gist.github.com/2e5d8c995ae6502f301e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6360.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6360.dmg

hifi-gustavo:
Build console output for commit: 06a87012ad63567d5d1c6b07138d10cad682735a available at: https://gist.github.com/2e5d8c995ae6502f301e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6360.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6360.dmg

ZappoMan:
Please fix this warning....
```
/data/jenkins/Jenkins/workspace/hifi-pull-request-ubuntu/libraries/avatars/src/Recording.cpp:239:29: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
         for (int j = 0; j < numJoints; j++) {
```

hifi-gustavo:
Build console output for commit: 2cf2257783198ab14435347ca098871396258268 available at: https://gist.github.com/2e5d8c995ae6502f301e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6360.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6360.dmg

interface/src/avatar/Avatar.cpp
```diff
@@ -888,6 +888,17 @@ glm::quat Avatar::getJointRotation(int index) const {
     return rotation;
 }
 
+QVector<glm::vec3> Avatar::getJointTranslations() const {
+    if (QThread::currentThread() != thread()) {
+        return AvatarData::getJointTranslations();
```
ZappoMan: wth?? If the current thread is not the thread, just call yourself? That makes no sense... if the goal is to make this thread safe, we need to invoke method to martial across to the other thread,

ZappoMan: ok - I see what you're doing here... but this still seems wrong.

Is the vector from `AvatarData` going to be the same as the vector from `_skeletonModel` if so, why not always use the data from `AvatarData` -- if they are potentially different, then why? and why is this not problematic to sometimes return a different version.

libraries/avatars/src/AvatarData.cpp
```diff
@@ -1146,6 +1146,21 @@ void AvatarData::setJointRotations(QVector<glm::quat> jointRotations) {
...
+    if (QThread::currentThread() != thread()) {
+        QVector<glm::vec3> result;
+        QMetaObject::invokeMethod(const_cast<AvatarData*>(this),
+            "getJointTranslations", Qt::BlockingQueuedConnection,
+            Q_RETURN_ARG(QVector<glm::vec3>, result));
+        return result;
+    }
+    QVector<glm::vec3> jointTranslations(_jointData.size());
+    for (int i = 0; i < _jointData.size(); ++i) {
+        jointTranslations[i] = _jointData[i].translation;
```
ZappoMan: what can change `_jointData`? Why not create this list of joint translations when `_jointData` is set in the first place, and just always have a value copy ready to return when asked for it? That way, we could simply have a mutex around this cache and not have to do a `Qt::BlockingQueuedConnection` to get the result.

This seems like an inefficient approach to getting the current joint translation state.

libraries/avatars/src/Player.cpp
```diff
@@ -246,24 +246,38 @@ void Player::play() {
...
+            joint.translation = glm::mix(prevJoint.translation, nextJoint.translation, _frameInterpolationFactor);
+            jointTranslations[i] = joint.translation; // FIXME: remove once the setRawJointData is fixed
+        }
     }
 
-    _avatar->setJointRotations(jointRotations);
-    _avatar->setJointTranslations(jointTranslations);
+   // _avatar->setRawJointData(jointArray); // FIXME: Enable once the setRawJointData is fixed
+     _avatar->setJointRotations(jointRotations); // FIXME: remove once the setRawJointData is fixed
+   //  _avatar->setJointTranslations(jointTranslations); // FIXME: remove once the setRawJointData is fixed
```
ZappoMan: don't you want to call `_avatar->setJointTranslations()` here? since you don't have `setRawJointData()` seems like you need to set both rotations and translations.

libraries/avatars/src/Recording.cpp
```diff
@@ -229,22 +229,27 @@ void writeRecordingToFile(RecordingPointer recording, const QString& filename) {
...
-            mask.resize(mask.size() + numJoints);
+            // 2 fields per joints
+            mask.resize(mask.size() + numJoints * 2);
         }
-        for (quint32 j = 0; j < numJoints; ++j) {
-            if (i == 0 ||
-                frame._jointRotations[j] != previousFrame._jointRotations[j]) {
-                writeQuat(stream, frame._jointRotations[j]);
-                // TODO -- handle translations
+        for (quint32 j = 0; j < numJoints; j++) {
```
ZappoMan: isn't it easier/cleaner to do....

```
for (const auto& joint : jointArray) {
```

samcake: I wanted to stick to the original style of that code :)

libraries/avatars/src/Player.cpp
samcake: THis is not giving me the correct results but it s not critical to get the animation looking plausible so i d say let s move on with this

libraries/avatars/src/AvatarData.cpp
samcake: This code is deprecated before being added there
i ll remove asap

hifi-gustavo:
Build console output for commit: d3d5d3a19e4970707e5c552585aa336c638beece available at: https://gist.github.com/2e5d8c995ae6502f301e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6360.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6360.dmg

PR #6362, opened 2015-11-10
Added scripts for helicopter

ericrius1:
Scripts to deal with the spawning and destruction of a helicopter in the alley scene


imgntn:
@ericrius1 not sure exactly what i'm looking at here -- the helicopter.js file has a very specific spawning point.  is it part of a 'set', like /toybox?  if so, we should start collecting those scripts together -- you mention an 'alley' set?  is there a reset script for that where we can add the specific instance of the helicopter and then try to make it more generically spawn near you if you just run the spawner script?  thanks

PR #6367, opened 2015-11-11
Dynamic loading of models from specified directory

ericrius1:
This script is an example of a way to dynamically load and place models in a grid from a specified s3 directory on the hifi-public bucket. The directory can be specified by changing the query string variable on line 19 to the desired relative path. 

So, for example, to load all models from 	
https://s3.amazonaws.com/hifi-public/eric/models, you'd change the URL from 
"https://serene-headland-4300.herokuapp.com/?assetDir=ozan/3d_marketplace/sets"
to
https://serene-headland-4300.herokuapp.com/?assetDir=eric/models"


imgntn:
@ericrius1 i think it's probably worth mentioning early in the script that you set up a custom API for this to return the directory listing, and explain what the response from that api is supposed to look like.  otherwise people might assume they can just put a bunch of files in a folder on dropbox or whatever and see what's up.

also, you could try setting up an s3 bucket as static website with directory listing for the bucket, and then parse that.  then you wouldn't need an api

ericrius1:
I updated the script header description to be more specific. The AWS S3 API works great, is simple to use, and is well suited for what Ozan and I are working on, which is an in-world marketplace to showcase various pieces of content. This is one component of the eventual system, and it may make sense in the future to modify/extend it, but for now it works well.

imgntn:
@ericrius1 what do you mean the 'aws s3 api'  ?? you wrote your own API that uses the AWS SDK in node or something?  probably worth saying you're running an express server, etc.  or actually just include that code also.  if we are running  https://serene-headland-4300.herokuapp.com as a utility for everyone to use then we should make sure we're setup correctly for scaling, etc and not just running a free node

ericrius1:
@imgntn This is really just a proof of concept feature right now, meant for Ozan and I's experimentation and use. The server is not set up for scaling, and it's using heroku's free tier. I don't feel that is the best use of my time or money to do scale it right now.  I opened a PR for this in the spirit of having everything we use in production in the repo. I can move this to unpublished scripts if you think that would make more sense.

imgntn:
@ericrius1 why not just include the directory with the files you use to run the heroku app, so anyone can do it?  or don't include this example at all if its not something other people will be able to use

ericrius1:
That's a good idea, I will add the server code, and people can modify it as they wish to use their own S3 bucket

hifi-gustavo:
Build console output for commit: 806d303dcd8b245de1d2093846c6d445d97655e5 available at: https://gist.github.com/4baec9a899ad27a30e6a<br/><br/>No installer was built for this commit

imgntn:
nice additions.

PR #6368, opened 2015-11-11
Refactor of MyAvatar animation JS interface

hyperlogic:
JavaScript changes:
   * removed MyAvatar.playAnimation
   * removed MyAvatar.stopAnimation
   * removed MyAvatar.getGetAnimationDetails
   * removed MyAvatar.startAnimationByRole
   * removed MyAvatar.stopAnimationByRole
   * removed MyAvatar.getGetAnimationDetailsByRole
   * removed MyAvatar.clearJointPriorities
   * added MyAvatar.overrideAnimation(url, fps, loop, firstFrame, lastFrame)
   * added MyAvatar.restoreAnimation()
   * added MyAvatar.getAnimationRoles()
   * added MyAvatar.overrideRoleAnimation(role, url, fps, loop, firstFrame, lastFrame)
   * added MyAvatar.restoreRoleAnimation(role)
   * added MyAvatar.prefetchAnimation(url)
   * added two examples: kneel.js and theBird.js

C++ changes:
   * Added getParent() and replaceChild() to AnimNode
   * Added findByName() and traverse() to AnimNode
   * Changed AnimStateMachine to hold nodes by childIndex instead of smart pointer.
     This allows script to replace nodes dynamically via overrideRoleAnimation


hifi-gustavo:
Build console output for commit: e1d0a97807a89b7bdbb4e118380e9518333ed6f0 available at: https://gist.github.com/9acff0a1050ef5411256<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6368.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6368.dmg

hifi-gustavo:
Build console output for commit: 56e96ed24c3e3af7513b8eebe4af5546e156115e available at: https://gist.github.com/9acff0a1050ef5411256<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6368.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6368.dmg

imgntn:
@tony so far this is awesome -- noticed that when i reset my hydra controllers i lose my role overrides.  that seems reasonable and probably expected.  is there a way / do you think ppl will want to reapply role overrides that we active before the hydra reset?  if there were a hydra reset event (maybe there is already?) script authors could hook into that to reapply, perhaps?  anyhow, just an observation!

hyperlogic:
@imgntn was this fixed when you checked developer - avatar - enable anim graph?

imgntn:
@hyperlogic yeah, i had anim graph enabled when i noticed the hydra reset thing -- actually, it kind of seems like this pattern could be problematic:
1. click the kneel button (now i'm kneeling)
2. do hydra reset  (i'm standing again)
3. no longer have control of my arms!!
4. i regain control of my arms when i hit the kneel button again (still standing)

howard-stearns:
Can we remove Rig Animations now? (If not, we should make them  work again. Selecting Developer->Avatar->Rig Animations doesn't work on my machine with the current version of this PR.)

howard-stearns:
Reset doesn't do the T-pose any more. That's a good thing visually, and may well be the right thing going forward. The reason it did so before was to make certain that everything that can be reset, was indeed reset. I.e., if a demo goes all pear-shaped, a reset is should get things to a known state. I don't know how to access the risk of not doing all the same resetting now.

howard-stearns:
Also re reset: it used to be the case that one could get back to the bind pose by turning off animations and resetting. (Although it wasn't perfect before. I think one had to go to Rig Animations first, and then turn off animations, and then reset.) That is no longer true. I don't think it's important to get to T-pose per se, but I do think it is important to get back to a "freshly loaded" state in which there isn't any consequence from previous history. Instead, with this PR, I don't know how to get rid of the last pose (e.g., idle).

howard-stearns:
walk.js appears to be broken. I don't know if it is from this PR or from Amazon S3 lossage. (I get "[UncaughtException] ReferenceError: Can't find variable: walkAssets in http://s3.amazonaws.com/hifi-public/scripts/walk.js:34".)  

In any case, we should check that there is some way to run walk.js after this scope of changes. It's a reasonable example of stuff that people want to be able to do and that I think we want to support.

howard-stearns:
I think the MyAvatar.overrideAnimation is easy to understand and the right thing (or rather, one of the many right things). Can we put a short comment at the top of the kneel.js example that makes it clear that while override, hand controllers, head controllers, etc. will not effect the animations (i.e., so no one is surprised that the hands don't follow the hydra, or that one can "drive" while kneeling without it combining into any other animations).

hifi-gustavo:
Build console output for commit: 9523dafff18a6aef38c967c9e5b4bce13e767e1f available at: https://gist.github.com/9acff0a1050ef5411256<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6368.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6368.dmg

hyperlogic:
Thanks for the comments @howard-stearns, I'm going to close this PR, the issues that you've raised with regard to rig animations are real.  It would take too much time to fix them in this PR, however, I will roll this work into another PR which removes the old animation system all together.

PR #6369, opened 2015-11-11
handControlledHead.js added

AlessandroSigna:
Drive the head's pitch and yaw with the rotation of the right hand controller

### DIFF
--- a/examples/example/avatarcontrol/handControlledHead.js
+++ b/examples/example/avatarcontrol/handControlledHead.js
@@ -1,0 +1,65 @@
+// handControlledHead.js
+
+//This script allows you to look around, driving the rotation of the avatar's head by the right hand orientation.
+
+const YAW_MULTIPLIER = 20000;
+const PITCH_MULTIPLIER = 15000;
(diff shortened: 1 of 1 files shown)


imgntn:
@AlessandroSigna when i tested this last night, my view moved up and down and my head moved side to side in my mirror but my view did not move side to side.  is that expected?

AlessandroSigna:
Did you tried in first person mode? If this is the case it's expected. At the moment it should work fine in the third person mode.

imgntn:
yeah you're right i was in first person mode... seems to work like expected then --but why no turning in 1st person mode?  not sure i get it.

also, please add a header file!

AlessandroSigna:
In first person the camera seems not following the yaw of the head, I don't know why. 
I can investigate on this later but i think that it's enough for the purpose of this example

imgntn:
Cool -- can you just add that as a to do: in the comments?  then i think we are good to go

```To do: In first person the camera seems not following the yaw of the head.  ```

PR #6370, opened 2015-11-11
QTryRead/WriteLocker

Atlante45:
Makes ReadWriteLocker safe against function that might throw.

The reason I named those classes QTryReadLocker/QTryWriteLocker is because I'm hoping I can get those accepted into Qt, then we'll just have to remove those files.

### DIFF
--- a/libraries/shared/src/shared/QTryReadLocker.h
+++ b/libraries/shared/src/shared/QTryReadLocker.h
@@ -1,0 +1,92 @@
+//
+//  QTryReadLocker.h
+//  shared/src/shared/QTryReadLocker.h
+//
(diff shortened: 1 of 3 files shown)


hifi-gustavo:
Build console output for commit: 65e9dc4be556fc58abdc48bc77983a92465d1b79 available at: https://gist.github.com/f1b766657b1ad0e47280<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6370.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6370.dmg

PR #6375, opened 2015-11-12
Fix models resizing

Atlante45:
When models are MUCH bigger one one axis than the others
and we modify the scale only of one of those small axis,
it wouldn't propagate because the change was relatively
too small conpared to the size of the model


hifi-gustavo:
Build console output for commit: 954391ae3e832090f1ac4643482f759d08ad4473 available at: https://gist.github.com/6421601992072a6e25ef<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6375.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6375.dmg

PR #6393, opened 2015-11-14
Fix for missing avatars on entry

hyperlogic:
This should fix the issue where a persons avatar was occasionally not visible to others for a long period of time.  This was due to several factors:

1. When a new avatar was added to the simulation that identity packet was not broadcast to other avatars on the server. This would rely on random logic to send the identity eventually.  This is fixed in this PR, by sending identity packets to all other clients when a new clients arrive.

2. The random identity logic was not being executed if the sequence number check caused an update to be skipped.  This means the probability of sending a random packet was reduced significantly, especially for clients that were loading geometry on entry.  This was fixed by doing the random check before sequence number check.

3. The 1/300 probably used in the check was too low, this was only a 63% chance of sending a identity packet within 5 seconds.  This was fixed by changing the probability to 1/187, which is a 80% chance to send an identity packet within 5 seconds.

4. The randFloat() implementation slightly reduced the identity packet probability due to quantization errors.  This has been replaced by a C++ std random number generator.

### DIFF
--- a/assignment-client/src/avatars/AvatarMixer.cpp
+++ b/assignment-client/src/avatars/AvatarMixer.cpp
@@ -63,7 +63,9 @@ AvatarMixer::~AvatarMixer() {
     _broadcastThread.wait();
 }
 
-const float BILLBOARD_AND_IDENTITY_SEND_PROBABILITY = 1.0f / 300.0f;
+// An 80% chance of sending a identity packet within a 5 second interval.
+// assuming 60 htz update rate.
+const float BILLBOARD_AND_IDENTITY_SEND_PROBABILITY = 1.0f / 187.0f;
 
 // NOTE: some additional optimizations to consider.
 //    1) use the view frustum to cull those avatars that are out of view. Since avatar data doesn't need to be present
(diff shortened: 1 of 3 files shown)


hyperlogic:
Also, I removed some tabs and some dangling whitespace.

hifi-gustavo:
Build console output for commit: 245150333b949200a11bb40e8b1d840fe8e66353 available at: https://gist.github.com/1c42640803b6eeb6c780<br/><br/>No installer was built for this commit

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -291,53 +334,11 @@ void AvatarMixer::broadcastAvatarData() {
 
                     numAvatarDataBytes += avatarPacketList->write(otherNode->getUUID().toRfc4122());
                     numAvatarDataBytes +=
-                        avatarPacketList->write(otherAvatar.toByteArray(false, randFloat() < AVATAR_SEND_FULL_UPDATE_RATIO));
+                        avatarPacketList->write(otherAvatar.toByteArray(false, distribution(generator) < AVATAR_SEND_FULL_UPDATE_RATIO));
```
PhilipRosedale: where are distribution and generator defined?  Is this something that is global to all our stuff now, and we should use it in place of randFloat()?  How does it compare for speed?

hyperlogic: These are defined at the start of AvatarMixer::broadcastAvatarData

    // setup for distributed random floating point values
    std::random_device randomDevice;
    std::mt19937 generator(randomDevice());
    std::uniform_real_distribution<float> distribution;

This uses the 64-bit [Mersenne Twister](https://en.wikipedia.org/wiki/Mersenne_Twister) random number generator, it is slower then the [LCG](https://en.wikipedia.org/wiki/Linear_congruential_generator) method that it used in the c library, however I don't expect that it will be a performance issue.

hyperlogic: I don't think we should retrofit all code to use this generator.  Only code that needs the additional accuracy.  In my tests, the randFloat() was slightly less likely to send identity packets at the desired rate, by about 5%, if I remember correctly.  So, I decided to change it the new generator to be sure we can properly predict the expected send rates.

nbq:
I compiled this on Ubuntu 15.04 and tested with two avatars. I noticed they both seem to show up quicker overall.

PhilipRosedale:
thanks for the test, @nbq

PR #6411, opened 2015-11-18
A couple fixes for windows 64 bit builds

jherico:
The OpenVR and SDL2 cmake files don't properly switch to the 64 bit binaries on 64 bit windows builds.  This should correct that.


hifi-gustavo:
Build console output for commit: 68c6718d1ca93410594fb8ae5ffc7e1d6edfd7ad available at: https://gist.github.com/46b55a66cb8b2e757779<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6411.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6411.dmg

PR #6413, opened 2015-11-18
Recording: Moving the Recording js interface to the Script engine

samcake:
- Moving the RecordingScriptingENgine to script-engine so it can be used by both the agent and the interface

- set the "controlled avatar" manually on the singleton right after instanciation in Agent and in Application, i don;t like that but that was the easiest

- Disabled the Audio playback in order to be able to move forward

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,7 +1,7 @@
 ###Dependencies
 
 * [cmake](http://www.cmake.org/cmake/resources/software.html) ~> 3.3.2
-* [Qt](http://www.qt.io/download-open-source) ~> 5.4.1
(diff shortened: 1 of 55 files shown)


hifi-gustavo:
Build console output for commit: c1675bba2168b610e389fc7dc01409a01ba12001 available at: https://gist.github.com/2aaad93838ed32d2a30d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6413.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6413.dmg

PR #6433, opened 2015-11-19
cleanup directory + recording bug fix

AlessandroSigna:
The recording master was able to start the recording session even if he never entered the recording area


examples/entityScripts/recordingMaster.js
```diff
@@ -27,7 +27,6 @@ var toolBar = null;
 var recordIcon;
 var isRecording = false;
 var channel = "groupRecordingChannel";
-Messages.subscribe(channel);
```
jherico: You can also remove the unsubscribe in the `cleanup()` function if you're removing this.

PR #6434, opened 2015-11-19
fix avatar reset

AndrewMeadows:
Sometimes resetting the avatar didn't work. The reason was that, despite the fact the HmdSensor was being reset **before** the avatar reset code was reached, the hmd transform used in the reset calculations was stale.

The fix is to update the cached HmdSensor transform immediately after re-centering the sensor.


hifi-gustavo:
Build console output for commit: 28e349f421225fb3ec085f86a29291705b9b1f23 available at: https://gist.github.com/1d27bec38fe1f3467bf6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6434.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6434.dmg

PR #6435, opened 2015-11-19
Load recorded clips from URLs, not file paths

jherico:
The only thing I don't like about this is that there is a period of time after 'loadRecording' where the clip is essentially empty.  I'd like to have the script block until the clip is valid (i.e. downloaded) but I can't seem to find the proper mechanism to do so.

### DIFF
--- a/libraries/recording/CMakeLists.txt
+++ b/libraries/recording/CMakeLists.txt
@@ -4,6 +4,6 @@ set(TARGET_NAME recording)
 setup_hifi_library(Script)
 
 # use setup_hifi_library macro to setup our project and link appropriate Qt modules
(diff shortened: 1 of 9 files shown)


hifi-gustavo:
Build console output for commit: e187aaedcbf60fb72f5d475da2e373d63db6ccf6 available at: https://gist.github.com/ef7e074c8a3cb8ef46b1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6435.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6435.dmg

PR #6443, opened 2015-11-20
fix crash with no asset server

ZappoMan:


hifi-gustavo:
Build console output for commit: 5c6cd9b06c2f653830f7bc51f0d4386e84b0f66e available at: https://gist.github.com/1203cbc595ab8f7e25d9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6443.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6443.dmg

libraries/networking/src/AssetClient.cpp
```diff
@@ -369,16 +369,18 @@ void AssetClient::handleNodeKilled(SharedNodePointer node) {
...
     QByteArray dataByteArray = data.toUtf8();
     auto upload = DependencyManager::get<AssetClient>()->createUpload(dataByteArray, extension);
-    QObject::connect(upload, &AssetUpload::finished, this, [callback, extension](AssetUpload* upload, const QString& hash) mutable {
-        if (callback.isFunction()) {
-            QString url = "atp://" + hash + "." + extension;
-            QScriptValueList args { url };
-            callback.call(QScriptValue(), args);
-        }
-    });
+    if (upload) {
```
jherico: I would suggest that you test for `!upload` and if that's the case do an early return and emit a warning in the log.

libraries/networking/src/AssetClient.cpp
```diff
@@ -405,7 +407,10 @@ void AssetScriptingInterface::downloadData(QString urlString, QScriptValue callb
         return;
     }
 
-    _pendingRequests << assetRequest;
+    {
```
jherico: So I actually initially put in a comment about thread safety for this class, before I realized it lived inside the `ScriptEngine` instance, and so I assumed that it would never be access except on that thread.  Are you seeing issues or is this just paranoia?

hifi-gustavo:
Build console output for commit: f1badc017b3c79b11d70fcc583fe78dd2f2911dd available at: https://gist.github.com/1203cbc595ab8f7e25d9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6443.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6443.dmg

PR #6445, opened 2015-11-20
Group recording - upload performance file on asset

AlessandroSigna:


AlessandroSigna:
missing a way to let the master copy the url of the performance file

PR #6447, opened 2015-11-20
Add texture property to polyline entity type

huffman:


libraries/networking/src/udt/PacketHeaders.cpp
```diff
@@ -41,7 +41,7 @@ PacketVersion versionForPacketType(PacketType packetType) {
         case PacketType::EntityAdd:
         case PacketType::EntityEdit:
         case PacketType::EntityData:
-            return VERSION_ENTITIES_PARTICLES_ADDITIVE_BLENDING;
+            return VERSION_ENTITIES_POLYLINE_TEXTURE;
```
huffman: @ZappoMan I wasn't 100% sure if I need to do anything else w.r.t. the updated version. Feedback appreciated.

hifi-gustavo:
Build console output for commit: 9e31614fce4379adf4e5675d54a7e5c40af174fc available at: https://gist.github.com/1e0aa6152e160cc60265<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6447.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6447.dmg

libraries/networking/src/udt/PacketHeaders.cpp
ZappoMan: this should be sufficient

PR #6425, opened 2015-11-18
fix case where servers could run without settings

birarda:
This fixes an issue the accidentally immediately failing messages-mixer exposed yesterday. It occurred when requesting settings from a domain-server on a hostname/port combination recently used by the kicked AC.

PR #6450, opened 2015-11-20
fix for agent script download fail

birarda:
This is a fix for something I added in #6425. 

Agents now connect to the domain sooner and the previous domain-server script download code assumed that they would not yet be connected to the domain when making their request.

Now I explicitly force them to wait until they have connected to the domain to download the script and execute it. That order of events also lays the groundwork for letting Agents get scripts from the domain-server over UDT, which we will want as we don't intend to handle TCP hole punching.


hifi-gustavo:
Build console output for commit: cf204ad5b07abc3a4db42febb9b861fa24bd68ef available at: https://gist.github.com/35b4265ae9d7d605c176<br/><br/>No installer was built for this commit

PR #6453, opened 2015-11-20
Survive Flash Mobs Without Making Everyone Sick

howard-stearns:
The HighFidelity system can handle connections of many avatars in the same domain. (It currently handles more than 200 before things start falling over with timeouts, etc.) This PR makes sure that when we have such large numbers, we still maintain frame rate (if possible) by skipping render of far-away avatars. This LOD-like render distance is automatically adjusted up and down as required, using a "Proportional-Integral-Derivative Closed Loop Controller" (aka PID). In the best circumstances, my desktop machine can render roughly 40 avatars @75Hz.

This PR does not do anything to try to increase either the number renderable, nor the number connectable. (One exception: with this PR, we do not update avatar joint state when not rendering.) As such improvements are made, this control will automatically adapt, although it may eventually need to be "re-tuned" by adjusting the hardcoded PID parameters.

Intended Effect, and How I tested:

0. Note that this is a forced situation with more avatars closer together than you are likely to encounter in typical use.
1. Fire up a bunch of avatar assignment clients on a stack manager on another (wired) machine. I used 100 instances of examples/acScripts/animatedAvatarAgent.js (updated in this PR). The avatars are scattered within 20m around 500,500,500. Note that it takes a while for each assignment client to spin up in turn. Although the rest of this test will work while the NPCs fire up, it gets confusing to interpret the results. I advise waiting. [I've fired up 100 in the domain 'howard', which I'll leave open for testing over the next couple of days.]
2. In Interface running this PR, open stats to view Framerate, and the extended stats to see "Renderable avatars" (new in this PR) underneath "LOD".
3. Go to the edge of the avatar group and turn away so that you don't render any avatars. You should see "Avatars: 100", "Framerate: 75" (or 60 in desktop), and in the extended part, "Renderable avatars: 100 w/in 32768m".
4. Turn partway back towards the group. You should see the number of renderable avatars and the distance spin down in order to keep the framerate up. Distant avatars will dispapear.
5. Turn all the way to the group and note the continued spinning down, and note the reported distance when it stabilizes.
6. Walk towards and through the group of NPCs. The distance should spin down more as you get closer to them, to keep the other numbers stable. As you swim through the group, avatars should appear and disappear to keep the frame rate at target. You should be able to back out again and have the distance move back up to keep the other numbers stable.

PID does not "optimize" the controlled value -- it is a robust method for getting pretty close much of the time. It can be slow to respond and is subject to hill-climbing traps. The main thing here is that we not oscillate intollerably. Remember that most of the time, the controller will be operating "pinned" (wide open, in this case meaning a renderable distance of 32768m).

The "knobs" we have (which in this PR are all exposed to Javascript so that we can experiment interactively) are:
- The constants for the P, I, and D terms.
- The high and low limits for the controlled value (1 / cuttoff distance, so that higher controlled value gives higher frame rate).

The code:

There are just two interesting things here, and both are very simple code:
1) The PID Controller. See libraries/shared/src/PIDController.h
2) An additional kind of frame rate for the controller to measure. See diff of Application.cpp
The rest is just using the controller (AvatarManager), using the result (Avatar), and reporting (Stats.*).


hifi-gustavo:
Build console output for commit: 9673133bf4d76aade057a5a0d3d50e7cf2d33af7 available at: https://gist.github.com/3940a853b092d597e099<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6453.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6453.dmg

interface/src/Application.cpp
```diff
@@ -1077,8 +1077,10 @@ void Application::paintGL() {
     uint64_t now = usecTimestampNow();
     static uint64_t lastPaintBegin{ now };
     uint64_t diff = now - lastPaintBegin;
+    float instantaneousFps = 0.0f;
     if (diff != 0) {
-        _framesPerSecond.updateAverage((float)USECS_PER_SECOND / (float)diff);
+        instantaneousFps = (float)USECS_PER_SECOND / (float)diff;
+        _framesPerSecond.updateAverage(_lastInstantaneousFps);
```
AndrewMeadows: There is no harm in integrating **instantaneousFps** for **_framesPerSecond** at this point rather than using **_lastInstantaneousFps**.

In fact, if the above change were made then the **_lastInstantaneousFps** data member would become cruft and could be removed.

interface/src/Application.h
```diff
@@ -159,6 +159,11 @@ class Application : public QApplication, public AbstractViewStateInterface, publ
     bool isForeground() const { return _isForeground; }
     
     float getFps() const { return _fps; }
+    float getTargetFramePeriod() { return isHMDMode() ? 1.0f / 75.0f : 1.0f / 60.0f; }
+    float getLastInstanteousFps() const { return _lastInstantaneousFps; }
+    float getLastPaintWait() const { return _lastPaintWait; };
```
AndrewMeadows: **getLastInstanteousFps()** and **getLastPaintWait()** are never called anywhere.  Are they cruft from a debug framework?  Do we anticipate that external code will ever want to query these data members?

interface/src/Application.cpp
howard-stearns: In the current PR, I don't use getLastInstanteousFps() or getLastPaintWait(), but during development, I found it very convenient to be able to compare these vs the getLastDeducedNonVSyncFps() that I am using. I'm pretty confident that they'll be useful to someone going forward. Note that getLastInstanteousFps() is different from getFps() in that the latter is windowed (and therefore lagged).

Do you think I should remove the getters/ivars?

interface/src/Application.cpp
```diff
@@ -1109,6 +1111,29 @@ void Application::paintGL() {
...
+    // Time between previous paintGL call and this one, which can vary not only with vSync misses, but also with QT timing.
+    // We're using this as a proxy for the time between vsync and displayEnd, below. (Not exact, but tends to be the same over time.)
+    // This is not the same as update(deltaTime), because the latter attempts to throttle to 60hz and also clamps to 1/4 second.
+    const float actualPeriod = diff / (float)USECS_PER_SECOND; // same as 1/instantaneousFps but easier for compiler to optimize
+    // Note that _lastPaintWait (stored at end of last call) is for the same paint cycle.
+    float deducedNonVSyncPeriod = actualPeriod - _lastPaintWait + _marginForDeducedFramePeriod; // plus a some non-zero time for machinery we can't measure
+    // We don't know how much time to allow for that, but if we went over the target period, we know it's at least the portion
+    // of paintWait up to the next vSync. This gives us enough of a penalty so that when actualPeriod crosses two cycles,
+    // the key part (and not an exagerated part) of _lastPaintWait is accounted for.
+    const float targetPeriod = getTargetFramePeriod();
```
AndrewMeadows: **getTargetFramePeriod()** returns 1/FPS, and is called three times.

In one call (this one) it is used to get the framePeriod.

In the other two calls it is used to compute the frameRate:  frameRate = 1/targetFramePeriod = 1/ (1/targetFPS).

Maybe it should be **getTargetFrameRate()** and this call should do division to get the framePeriod?

interface/src/avatar/AvatarManager.cpp
```diff
@@ -141,10 +167,14 @@ void AvatarManager::updateOtherAvatars(float deltaTime) {
         } else {
             avatar->startUpdate();
             avatar->simulate(deltaTime);
+            if (!avatar->getShouldSkipRendering()) {
+                renderableCount++;
+            }
```
AndrewMeadows: This is the only place where **getShouldSkipRendering()** is called.

"getShouldSkipRendering" could be translated to "doNotRender", so this code looks kinda like a double negative.  Perhaps it should be changed to **getShouldRender()** instead?

libraries/shared/src/PIDController.h
```diff
@@ -0,0 +1,89 @@
...
+    // In normal operation (where we can easily reach setpoint), controlledValue is typcially pinned at max.
+    // Defaults to [0, max float], but for 1/LODdistance, it might be, say, [0, 0.2 or 0.1]
+    float getControlledValueLowLimit() const { return _controlledValueLowLimit; }
+    float getControlledValueHighLimit() const { return _controlledValueHighLimit; }
+    float getAntiWindupFactor() const { return _antiWindupFactor; } // default 10
+    float getKP() const { return _kp; }
+    float getKI() const { return _ki; }
+    float getKD() const { return _kd; }
+    float getAccumulatedValueHighLimit() const { return getAntiWindupFactor() * getMeasuredValueSetpoint(); }
+    float getAccumulatedValueLowLimit() const { return -getAntiWindupFactor() * getMeasuredValueSetpoint(); }
```
AndrewMeadows: I don't mind simple variable names (such as **p**, **i**, and **d**) when they are used in very standard formula, such as the PID controller, but maybe there should be a helpful comment for future developers that says:

p = proportional term
i = integral term
d = derivative term

libraries/shared/src/PIDController.cpp
```diff
@@ -0,0 +1,78 @@
...
+    Row& next = _history[n];
+    next.measured = measuredValue;
+    next.dt = dt;
+    next.error = error;
+    next.accumulated = accumulatedError;
+    next.changed = changeInError;
+    next.p = p;
+    next.i = i;
+    next.d = d;
+    next.computed = computedValue;
```
AndrewMeadows: Not a blocker, but it occurs to me:

**PIDController::updateHistory()** takes nine arguments.

**PIDController::Row** is a struct with nine corresponding data members.

**PIDController** itself maintains nine corresponding data members.  In a sense, it is a fancy version of a **Row**... with history.

There is probably a way to use OOP to simplify this implementation.

interface/src/Application.cpp
AndrewMeadows: Not a blocker.  If _lastInstantaneousFps is useful for debugging then let's keep it around.

interface/src/Application.h
AndrewMeadows: Nevermind.  Let's keep for future debugging.

interface/src/avatar/AvatarManager.cpp
howard-stearns: I'll change it. The thinking was to avoid text-search-conflicts with shouldRenderLocally and shouldRenderBillboard, but I don't feel strongly about it.

interface/src/Application.cpp
howard-stearns: There's now a separate getTargetFrameRate() per the related line-comment.

hifi-gustavo:
Build console output for commit: f1e0742aeb6aa5e65de04c96b974bd197082e023 available at: https://gist.github.com/3940a853b092d597e099<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6453.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6453.dmg

hifi-gustavo:
Build console output for commit: c94a5389a0af9144182e90659c0807d61a91d22b available at: https://gist.github.com/3940a853b092d597e099<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6453.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6453.dmg

PR #6462, opened 2015-11-23
Blendshapes and attachments in recordings, plus other minor fixes.

jherico:
removing the defunct recording files from the avatars library


hifi-gustavo:
Build console output for commit: 64ba5b4f1488828a4e8b6626b4560b25231c63a3 available at: https://gist.github.com/384b2ace2646b9121bab<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6462.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6462.dmg

hifi-gustavo:
Build console output for commit: 5c2b980a4532dfe3ff95b3590858a1642fc7c7f6 available at: https://gist.github.com/384b2ace2646b9121bab<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6462.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6462.dmg

ZappoMan:
test this please

libraries/shared/src/Transform.cpp
```diff
@@ -128,17 +128,25 @@ QJsonObject Transform::toJson(const Transform& transform) {
     }
 
     QJsonObject result;
-    auto json = toJsonValue(transform.getTranslation());
-    if (!json.isNull()) {
-        result[JSON_TRANSLATION] = json;
+    if (transform.getTranslation() != vec3()) {
```
ZappoMan: so if the translation is (0,0,0) then we don't put anything into the json? Is that really the behavior we want?

hifi-gustavo:
Build console output for commit: 5c2b980a4532dfe3ff95b3590858a1642fc7c7f6 available at: https://gist.github.com/384b2ace2646b9121bab<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6462.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6462.dmg

libraries/shared/src/Transform.cpp
```diff
@@ -128,17 +128,25 @@ QJsonObject Transform::toJson(const Transform& transform) {
...
+        auto json = toJsonValue(transform.getTranslation());
```
ZappoMan: I also just noticed that in `shared` there's `glmToJson` that essentially does this same thing of checking against the default value, why aren't you using those?

```
template <typename T> 
QJsonValue glmToJson(const T& t) {
    static const T DEFAULT_VALUE = T();
    if (t == DEFAULT_VALUE) {
        return QJsonValue();
    }
    QJsonArray result;
    for (auto i = 0; i < t.length(); ++i) {
        result.push_back(t[i]);
    }
    return result;
}
```

libraries/shared/src/Transform.cpp
jherico: the idea is to serialize the minimum amount of information.  so if the rotation is identity (0, 0, 0, 1) we don't populate the rotation field, and if the translation is (0, 0, 0) we don't populate the translation field and if the scale is (1, 1,1) we don't populate the scale field.  so an identity transform would be empty.  so, yes

libraries/shared/src/Transform.cpp
jherico: we are actually.  the `toJsonValue()` is implement in terms of the `glmToJson` function.  We're just doing additional checking which probably isn't necessary, but makes the intent of the transform method clearer.  The only deviation is that for scale we want to test against a `vec3(1)`, not a `vec3()`.

PR #6463, opened 2015-11-23
fix bug where observers don't see a held/equipped object moving

sethalves:
- fix a bug where a held object's motion-state might have a type of "static" so it wouldn't appear to move

### DIFF
--- a/interface/src/avatar/AvatarActionHold.cpp
+++ b/interface/src/avatar/AvatarActionHold.cpp
@@ -328,5 +328,5 @@ void AvatarActionHold::deserialize(QByteArray serializedArguments) {
         _active = true;
     });
 
-    activateBody();
+    forceBodyNonStatic();
 }
--- a/interface/src/avatar/AvatarActionHold.cpp
+++ b/interface/src/avatar/AvatarActionHold.cpp
@@ -326,6 +326,7 @@ void AvatarActionHold::deserialize(QByteArray serializedArguments) {
         #endif
 
         _active = true;
-        activateBody();
     });
+
+    activateBody();
 }
--- a/interface/src/avatar/AvatarActionHold.cpp
+++ b/interface/src/avatar/AvatarActionHold.cpp
@@ -326,5 +326,6 @@ void AvatarActionHold::deserialize(QByteArray serializedArguments) {
         #endif
 
         _active = true;
+        activateBody();
     });
 }
--- a/libraries/physics/src/ObjectAction.cpp
+++ b/libraries/physics/src/ObjectAction.cpp
@@ -244,6 +244,15 @@ void ObjectAction::activateBody() {
     if (rigidBody) {
         rigidBody->activate();
     }
+    auto ownerEntity = _ownerEntity.lock();
+    if (!ownerEntity) {
+        return;
+    }
+    void* physicsInfo = ownerEntity->getPhysicsInfo();
+    ObjectMotionState* motionState = static_cast<ObjectMotionState*>(physicsInfo);
+    if (motionState && motionState->getMotionType() == MOTION_TYPE_STATIC) {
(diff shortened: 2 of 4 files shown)


hifi-gustavo:
Build console output for commit: 6f19093cb9959881b2002393ba6f66aec6c1d498 available at: https://gist.github.com/52a0aaf2f585bf263716<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6463.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6463.dmg

hifi-gustavo:
Build console output for commit: 5612174894b602f818528023dfea37376e1ff658 available at: https://gist.github.com/52a0aaf2f585bf263716<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6463.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6463.dmg

hifi-gustavo:
Build console output for commit: c3821d7202b4acc409e56c6702f2278304612170 available at: https://gist.github.com/52a0aaf2f585bf263716<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6463.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6463.dmg

interface/src/avatar/AvatarActionHold.cpp
```diff
@@ -327,4 +327,6 @@ void AvatarActionHold::deserialize(QByteArray serializedArguments) {
 
         _active = true;
     });
+
+    activateBody();
 }
```
AndrewMeadows: At first glance this looks dangerous: to always activate on deserialization.  If we deserialize often (more than once every two seconds) then the body will be kept active, even if the deserialization didn't actually change anything.

libraries/physics/src/ObjectAction.cpp
```diff
@@ -246,6 +246,18 @@ void ObjectAction::activateBody() {
...
+    auto ownerEntity = _ownerEntity.lock();
+    if (!ownerEntity) {
+        return;
+    }
+    void* physicsInfo = ownerEntity->getPhysicsInfo();
+    ObjectMotionState* motionState = static_cast<ObjectMotionState*>(physicsInfo);
+    if (motionState && motionState->getMotionType() == MOTION_TYPE_STATIC) {
+        ownerEntity->flagForMotionStateChange();
+    }
+}
```
AndrewMeadows: Change this to be called something like: **checkForMotionTypeChange()** and then:

Compare the results between these two calls:

motionState->getMotionType()  =  (current MotionType of the RigidBody, cached in ObjectMotionState::_motionType) and
motionState->computeObjectMotionType()  =  (expected MotionType based on properties of the "object" (in this context, the EntityItem))

Iff they are different then go ahead and call: ownerEntity->flagForMotionStateChange()

hifi-gustavo:
Build console output for commit: d4937071e189121dab2dd0d46b0d33eeb1878771 available at: https://gist.github.com/52a0aaf2f585bf263716<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6463.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6463.dmg

libraries/physics/src/ObjectAction.cpp
AndrewMeadows: Nevermind.  Seth pointed out that the object should pretty much never be STATIC when hold action is in effect.

PR #6465, opened 2015-11-23
moved logic to build vertex buffer to renderable polyline class, recalculati…

ericrius1:
…ng vertices anytime any of the normals, point, or strokeWidths data changes

Testing performance with release build now, so do not merge

### DIFF
--- a/examples/painting/whiteboard/whiteboardSpawner.js
+++ b/examples/painting/whiteboard/whiteboardSpawner.js
@@ -247,4 +247,4 @@ function cleanup() {
 
 
 // Uncomment this line to delete whiteboard and all associated entity on script close
-Script.scriptEnding.connect(cleanup);
+//Script.scriptEnding.connect(cleanup);
--- a/examples/painting/whiteboard/whiteboardSpawner.js
+++ b/examples/painting/whiteboard/whiteboardSpawner.js
@@ -247,4 +247,4 @@ function cleanup() {
 
 
 // Uncomment this line to delete whiteboard and all associated entity on script close
-// Script.scriptEnding.connect(cleanup);
+Script.scriptEnding.connect(cleanup);
--- a/libraries/entities-renderer/src/RenderablePolyLineEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderablePolyLineEntityItem.cpp
@@ -31,6 +31,7 @@ EntityItemPointer RenderablePolyLineEntityItem::factory(const EntityItemID& enti
 RenderablePolyLineEntityItem::RenderablePolyLineEntityItem(const EntityItemID& entityItemID, const EntityItemProperties& properties) :
 PolyLineEntityItem(entityItemID, properties) {
     _numVertices = 0;
+    _vertices = QVector<glm::vec3>(0.0f);
 
 }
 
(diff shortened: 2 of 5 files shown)


hifi-gustavo:
Build console output for commit: 1409ed2d1892a8bd659a937efb24c0a6b9f97933 available at: https://gist.github.com/f0a435db5aab68ef8015<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6465.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6465.dmg

huffman:
I would recommend testing this with a solid texture so any gaps due to regeneration of vertices is obvious. I can give you a test script tomorrow if you are interested.

examples/painting/whiteboard/whiteboardSpawner.js
```diff
@@ -247,4 +247,4 @@ function cleanup() {
 
 
 // Uncomment this line to delete whiteboard and all associated entity on script close
-// Script.scriptEnding.connect(cleanup);
\ No newline at end of file
+Script.scriptEnding.connect(cleanup);
```
ZappoMan: did you mean to leave this in?

ericrius1: Nope- thanks!

ericrius1:
Sounds good @huffman- we can also incorporate your dynamic texture changes

hifi-gustavo:
Build console output for commit: 40be9ae1211409c9c34997c544c73201ab163281 available at: https://gist.github.com/f0a435db5aab68ef8015<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6465.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6465.dmg

ericrius1:
Just tested a bunch with a solid texture- there are no gaps. Seems ready to merge to me.

libraries/entities-renderer/src/RenderablePolyLineEntityItem.cpp
```diff
@@ -31,6 +31,7 @@ EntityItemPointer RenderablePolyLineEntityItem::factory(const EntityItemID& enti
 RenderablePolyLineEntityItem::RenderablePolyLineEntityItem(const EntityItemID& entityItemID, const EntityItemProperties& properties) :
 PolyLineEntityItem(entityItemID, properties) {
     _numVertices = 0;
+    _vertices = QVector<glm::vec3>(0.0f);
```
huffman: These should be in the initializer list or set to a default value in the header.

libraries/entities-renderer/src/RenderablePolyLineEntityItem.h
```diff
@@ -40,8 +40,10 @@ class RenderablePolyLineEntityItem : public PolyLineEntityItem {
     
 protected:
     void updateGeometry();
+    void updateVertices();
     gpu::BufferPointer _verticesBuffer;
     unsigned int _numVertices;
+    QVector<glm::vec3> _vertices;
```
huffman: Do the vertices need to be stored? We only seem to need them when updating the geometry.

libraries/entities-renderer/src/RenderablePolyLineEntityItem.cpp
```diff
@@ -114,13 +115,56 @@ void RenderablePolyLineEntityItem::updateGeometry() {
         _numVertices += 2;
     }
     _pointsChanged = false;
+    _normalsChanged = false;
+    _strokeWidthsChanged = false;
+}
+
+void RenderablePolyLineEntityItem::updateVertices() {
```
Atlante45: @ericrius1 This is an example of somewhere you can get a big perf boost.
You could just copy your points/normal/width buffers to the GPU and compute all of that in the shaders like I do for particles.

We can talk more about it next week when you're back and I can show you how to do it if you want.

ZappoMan:
FWIW - there's currently (in master) a bug, that causes polylines to often not render correctly. This PR fixes that bug. The bug was caused by the fact that _vertices is needed to calculate the geometry, but it's only calculated when the PROP_NORMALS property is set... which doesn't always happen last.

Even though I agree with both @huffman and @Atlante45 feedback about further improvements to this PR... it might be worthwhile to merge this as is, since it does fix an existing problem with master.

libraries/entities-renderer/src/RenderablePolyLineEntityItem.cpp
ericrius1: @Atlante45 - that would be awesome, thanks!

libraries/entities-renderer/src/RenderablePolyLineEntityItem.h
ericrius1: @huffman- that's a great point. I plan on implementing Clement's suggestion to calculate vertices in shader next week, so this won't be an issue

ericrius1:
@ZappoMan - definitely plan on implementing the improvements @huffman and @Atlante45 suggest, but merging this now for the bug fix seems fine to me too.

Atlante45:
@ZappoMan I agree, I wasn't suggesting we block the PR until then.

I'm gonna go over it one last time and merge it.
I'll leave my comments on further improvements here:
In the short term, moving the 2 update calls out of `render()` and into `update()` would be nice.
All that work wouldn't be done during the rendering pass.
Longer term, let's not even have those 2 functions at all and do those computations per vertex on the GPU.

PR #6468, opened 2015-11-23
change the ScriptEngine::waitTillDoneRunning() to wait for the script thread to complete

ZappoMan:


hifi-gustavo:
Build console output for commit: 14b2353bb22873c1514d678b0e3e67cd72abf35e available at: https://gist.github.com/a69ec8149f379bc22c05<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6468.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6468.dmg

birarda:
Wanting to post the same comment here for potential reviewers - I'm not sure why we wouldn't just use a combination of `quit` and `wait` on the ScriptEngine thread where we currently call `waitTillDoneRunning` in `stopAllScripts`

hifi-gustavo:
Build console output for commit: 241603b929edd260e798ec41cbec047038e2c69a available at: https://gist.github.com/a69ec8149f379bc22c05<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6468.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6468.dmg

birarda:
Not being able to `wait` is an issue with having threads needing things from the main thread while we're attempting to tear them down. 

Letting the ScriptEngines do whatever they like on shutdown, especially things that may need resources from the main thread, does not feel very good to me - but I think that's a separate issue that I'm happy tabling on this PR.

birarda:
In reading the Qt docs I don't think it's safe to assume this will do what we want.

**When this signal is emitted, the event loop has already stopped running. No more events will be processed in the thread, except for deferred deletion events. This signal can be connected to QObject::deleteLater(), to free objects in that thread.**

Since the ScriptEngine should process threadDoneRunning in the thread that's now done running I would assume it's possible that threadDoneRunning is never called.

PR #6469, opened 2015-11-23
Group recording - overlay added 

AlessandroSigna:


examples/entityScripts/recordingEntityScript.js
```diff
@@ -60,12 +66,22 @@
...
 
         leaveEntity: function (entityID) {
             print("leaving the recording area");
             _this.stopRecording();
             Messages.unsubscribe(MASTER_TO_CLIENTS_CHANNEL);
+            Overlays.deleteOverlay(overlay);
+            overlay = null;
         },
 
         startRecording: function (entityID) {
```
ZappoMan: no point in having an entityID here, because you're not passing it in your call in leaveEntity and/or receivingMessage

examples/entityScripts/recordingEntityScript.js
```diff
@@ -74,6 +90,7 @@
                 Messages.sendMessage(CLIENTS_TO_MASTER_CHANNEL, PARTICIPATING_MESSAGE);  //tell to master that I'm participating
                 Recording.startRecording();
                 isAvatarRecording = true;
+                Overlays.editOverlay(overlay, {imageURL: RECORDING_ICON_URL});
             }
         },
```
ZappoMan: no point in having an entityID here, because you're not passing it in your call in leaveEntity and/or receivingMessage

PR #6471, opened 2015-11-23
fix unnecessary model URL updating thanks to URL encoding

birarda:


birarda:
Yep @thoys just commented on the forum post!

hifi-gustavo:
Build console output for commit: 6339c2c48f0d75977bdbf844c9ffcfad606dcec7 available at: https://gist.github.com/c25954baad4c9d01d7f2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6471.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6471.dmg

thoys:
:+1:

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -219,7 +219,7 @@ void RenderableModelEntityItem::render(RenderArgs* args) {
 
     if (hasModel()) {
         if (_model) {
-            if (getModelURL() != _model->getURL().toString()) {
+            if (getModelURL() != _model->getURL().toEncoded()) {
```
ZappoMan: this doesn't really make sense. `getModelURL()` returns a `QString`, but `QUrl::toEncoded()` returns a `QByteArray`

http://doc.qt.io/qt-5/qurl.html#toEncoded

birarda: There's specifically an equality operator in `QString` that handles a `QByteArray` that is being leveraged here.

http://doc.qt.io/qt-5/qstring.html#operator-eq-eq-2

ZappoMan: I am well aware of that, but the point is that, but that hides the conversion from Utf8. This should be made more clear. I think there's a cleaner fix here. More importantly, I believe that this doesn't actually handle the other aspects of URL resolution that can happen in the background. For example, local files get further "resolution" treatment in the resource cache that may result in the _model->getURL() not just being an encoding difference. This might fix the case of spaces, but it doesn't fix the underlying differences between the original url property string, and the resolved URL after Model.cpp has processed it.

ZappoMan: A cleaner fix would be to just use a QUrl member variable inside of RenderableModelEntityItem. And compare against that.

ZappoMan: relevant code is....

```
class Model {
   ...
    Q_INVOKABLE void setURL(const QUrl& url);
    const QUrl& getURL() const { return _url; }
```
```
void RenderableModelEntityItem::render(RenderArgs* args) {
    ...
                qDebug() << "Updating model URL: " << getModelURL();
                _model->setURL(getModelURL());
```

Since we call _model->setURL() with a QString, the `QUrl(const QString & url, ParsingMode parsingMode = TolerantMode)` is being called to form the QUrl used by Model. We should not then compare the `QUrl::toEncoded(FormattingOptions options = FullyEncoded)` to that original string since the QString->QUrl->QString loop will not be guaranteed.

hifi-gustavo:
Build console output for commit: 0e19e50047cdeb0ceaf327669d02e5b6ad7bd8a2 available at: https://gist.github.com/c25954baad4c9d01d7f2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6471.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6471.dmg

hifi-gustavo:
Build console output for commit: 0d6b9194486d8bd50adb5f9f18bf032cee02e515 available at: https://gist.github.com/c25954baad4c9d01d7f2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6471.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6471.dmg

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -219,9 +219,13 @@ void RenderableModelEntityItem::render(RenderArgs* args) {
 
     if (hasModel()) {
         if (_model) {
-            if (getModelURL() != _model->getURL().toString()) {
-                qDebug() << "Updating model URL: " << getModelURL();
-                _model->setURL(getModelURL());
+            // convert the QString from getModelURL to a URL
+            _url = getModelURL();
```
ZappoMan: I think this is going to get called on every render frame, even when the property hasn't actually changed.

I think the better place to handle this is in `ModelEntityItem::setModelURL(const QString& url)`. So that it's only actually changing when the property is changed.

Unfortunately, as we discussed with Seth, Tony, and Sam yesterday afternoon, you can't actually ever change the URL of the Model class instance except on the main render thread, so we still need this check to see if there has been a change in this render call, but we shouldn't be reparsing the string every frame.

Also note: `_url` is a poor name since Model Entities can have at least 3 different urls related to their rendering: the mesh/model url, their animation url, and their collision url.

I think ultimately the behavior we want is that someone can set the URL property to some string "http://foo/with a space.fbx" and that if they ask for the property back, it is the same value they used to set it... but internally we convert it once to the QUrl format, and use that to compare against any other cases where we want to compare against the url.

birarda: I can move this to `ModelEntityItem` (although it can't be called `_modelURL` given the confusingly named QString that uses that name). Are there any issues adding member variables to ModelEntityItem that will not be handled as entity properties? I try to stay away from anything that may disturb them.

ZappoMan: There's nothing wrong with adding member variables to EntityItem classes, plenty of things don't need to go over the wire and are just plain old member variables. With respect to the variable names, in general, we've tried to keep the member variables that store properties named the same as the property, which explains why it's called `_modelURL` - however, it's perfectly reasonable to rename that member variable.

If you don't feel comfortable making this fix, I can close this PR and fix it.

hifi-gustavo:
Build console output for commit: 5a7b0dd1adc8b673029f93c0dbcc43891ed53d85 available at: https://gist.github.com/c25954baad4c9d01d7f2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6471.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6471.dmg

PR #6473, opened 2015-11-24
change the ScriptEngine::waitTillDoneRunning() to wait for the script thread to complete

ZappoMan:


libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -122,7 +122,7 @@ void EntityTreeRenderer::init() {
 }
 
 void EntityTreeRenderer::shutdown() {
-    _entitiesScriptEngine->disconnect(); // disconnect all slots/signals from the script engine
+    _entitiesScriptEngine->disconnectNonEssentialSignals(); // disconnect all slots/signals from the script engine, except essential
```
ZappoMan: this was very likely the bug that caused some ScriptEngines to not get deleted, specifically, the _entitiesScriptEngine was getting disconnected from all signals, as a result, it would disconnect the ScriptEngine::doneRunning() from the Thread::quit, which would cause the thread to never quit and subsequently not properly delete

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -124,14 +124,9 @@ static bool hadUncaughtExceptions(QScriptEngine& engine, const QString& fileName
 
 ScriptEngine::ScriptEngine(const QString& scriptContents, const QString& fileNameString, bool wantSignals) :
     _scriptContents(scriptContents),
-    _isFinished(false),
```
ZappoMan: this is just coding standard cleanup

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -140,6 +135,8 @@ ScriptEngine::ScriptEngine(const QString& scriptContents, const QString& fileNam
 }
 
 ScriptEngine::~ScriptEngine() {
+    qDebug() << "Script Engine shutting down (destructor) for script:" << getFilename();
```
ZappoMan: to verify that ScriptEngines really are being destructed

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -150,7 +147,13 @@ ScriptEngine::~ScriptEngine() {
     }
 }
 
+void ScriptEngine::disconnectNonEssentialSignals() {
```
ZappoMan: this method is used specifically for the EntityTreeRenderer and it's _entitiesScriptEngine to disconnect signals from entities before the application is shutdown (but after we've already unloaded the domain content)... but we don't want it to disconnect the signal from `doneRunning` to `QThread::quit` or else we won't shut down the engine properly.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -176,12 +179,13 @@ void ScriptEngine::runInThread() {
 QSet<ScriptEngine*> ScriptEngine::_allKnownScriptEngines;
 QMutex ScriptEngine::_allScriptsMutex;
 bool ScriptEngine::_stoppingAllScripts = false;
-bool ScriptEngine::_doneRunningThisScript = false;
```
ZappoMan: changed the logic to loop and wait on the thread actually being done running, so we didn't need this anymore

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -227,21 +233,19 @@ void ScriptEngine::stopAllScripts(QObject* application) {
...
+    qDebug() << "DONE Stopping all scripts....";
 }
 
 
 void ScriptEngine::waitTillDoneRunning() {
     // If the script never started running or finished running before we got here, we don't need to wait for it
-    if (_isRunning) {
-
-        _doneRunningThisScript = false; // NOTE: this is static, we serialize our waiting for scripts to finish
+    if (_isRunning && _isThreaded) {
```
ZappoMan: note: we don't have non-threaded script engines in the application, only in agent, and that code doesn't actually ever call this. However, if in the future we use non-threaded scripts in Application, we wouldn't want them to be waiting on the thread() to stop running.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -227,21 +233,19 @@ void ScriptEngine::stopAllScripts(QObject* application) {
...
 
         // NOTE: waitTillDoneRunning() will be called on the main Application thread, inside of stopAllScripts()
         // we want the application thread to continue to process events, because the scripts will likely need to
         // marshall messages across to the main thread. For example if they access Settings or Meny in any of their
         // shutdown code.
-        while (!_doneRunningThisScript) {
-
+        while (thread()->isRunning()) {
```
ZappoMan: This is another critical change. In the past we would wait for `_doneRunningThisScript` which happened at the end of `::run()` but we want to make sure we wait until the thread is really done running.

hifi-gustavo:
Build console output for commit: 0148f891e2f03de17b507417baa0642be0c11e00 available at: https://gist.github.com/2bdd55f18a8709f17676<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6473.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6473.dmg

hifi-gustavo:
Build console output for commit: ce46c3064754f940cf3ba6a10609e956a57f55b6 available at: https://gist.github.com/2bdd55f18a8709f17676<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6473.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6473.dmg

libraries/entities-renderer/src/EntityTreeRenderer.cpp
birarda: Can we add a comment where the important connections are made that anything crucial should be reconnected here? Seems like it could be easy to miss this special disconnect and lose an important connection.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -219,7 +223,9 @@ void ScriptEngine::stopAllScripts(QObject* application) {
             // We need to wait for the engine to be done running before we proceed, because we don't
             // want any of the scripts final "scriptEnding()" or pending "update()" methods from accessing
             // any application state after we leave this stopAllScripts() method
+            qDebug() << "waiting on script:" << scriptName;
             scriptEngine->waitTillDoneRunning();
+            qDebug() << "done waiting on script:" << scriptName;
```
birarda: Still want this debug?

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -227,21 +233,19 @@ void ScriptEngine::stopAllScripts(QObject* application) {
     }
     _stoppingAllScripts = false;
     _allScriptsMutex.unlock();
+    qDebug() << "DONE Stopping all scripts....";
```
birarda: Still want this debug?

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: Yes. I left it in on purpose.

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: Yes. I left it in on purpose.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
ZappoMan: Where exactly do you suggest I add this comment? Signals can be connected anywhere. I don't know if a place to add this comment exists.

birarda: I agree people could still shoot themselves in the foot from other places - I think a comment in ScriptEngine runInThread would be a useful thing to add.

ZappoMan: done

hifi-gustavo:
Build console output for commit: 5f88d958ab592f21122eaa388363700bb0052fe4 available at: https://gist.github.com/2bdd55f18a8709f17676<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6473.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6473.dmg

PR #6476, opened 2015-11-24
fix for objects suffering from very small updates to physics properties

AndrewMeadows:
There is a bug when a script that updates an object in small increments the changes would not propagate to the physics representation.  So when the object reverts to physical simulation it could have the wrong position/rotation.

This PR makes all external (script or entity-server) physics changes propagate to the simulation.


AndrewMeadows:
BTW, when testing this with two avatars present I noticed an oddity where the simulation ownership was sometimes not properly passed from one interface to the other.  One owned the simulation but the other would not give it up.  I don't think it was related to the changes I made here but I haven't yet had time to test on the master branch.  So don't merge this one quite yet.

hifi-gustavo:
Build console output for commit: 1eb9ededdcf218eca56d947eff976f1e9f9771be available at: https://gist.github.com/fa8fe6f16bbda522c0d8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6476.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6476.dmg

imgntn:
@AndrewMeadows yeah lets test with multiple avatars tomorrow but in my first tests with the bow the arrow seems to shoot as expected!!!! :+1:

AndrewMeadows:
I did more testing this morning.  I think the "unexpected behavior" I was seeing is really just that the arrow will never go inactive while it is near an avatar... because the avatar object never goes inactive and it will keep all nearby dynamic objects active.  This is wrong (the avatar should be allowed to go inactive) but it is how it currently works in the master branch so the problem is not new to this PR.

imgntn:
@AndrewMeadows yeah i just tested with multiple avatars and the bow and arrow behaves as expected.  thanks!!

hifi-gustavo:
Build console output for commit: e26081e9818a7e635249ebcc8db19b3ab38d8c8f available at: https://gist.github.com/fa8fe6f16bbda522c0d8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6476.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6476.dmg

PhilipRosedale:
Tested and looks great.  Will update for meeting.

PR #6479, opened 2015-11-24
No more right-click-move editing in edit.js

thoys:
ignore other than left press events in edit.js, this makes the rightclick camera more comfortable while in edit.js/edit mode. The no-market info available is still present on right clicking an entity ;).

### DIFF
--- a/examples/edit.js
+++ b/examples/edit.js
@@ -738,6 +738,10 @@ var mousePressStartPosition = {
 var mouseDown = false;
 
 function mousePressEvent(event) {
+    if (!event.isLeftButton) {
+        // If another mouse button than left is pressed ignore it
+        return false;
+    }
     mouseDown = true;
     mousePressStartPosition = {
         x: event.x,
--- a/examples/edit.js
+++ b/examples/edit.js
@@ -392,6 +392,11 @@ var toolBar = (function() {
             url,
             file;
 
+        if (!event.isLeftButton) {
+            // if another mouse button than left is pressed ignore it
+            return false;
+        }
+
         clickedOverlay = Overlays.getOverlayAtPoint({
             x: event.x,
             y: event.y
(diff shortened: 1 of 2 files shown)


imgntn:
@thoys i like what this does generally, but then there's no way to orbit with the mouse once an object is 'focused' (select an object and press "f") -- can you make it conditional?  no right click when nothing is focused, but if something is focused, allow the right click so we can orbit?  thanks

thoys:
Oh, I will have a look into that

huffman:
@thoys The camera should just-work if you make sure to at least pass the event to camera manager **if** `isActive` is `true`:
`cameraManager.mousePressEvent(event)`

thoys:
I think the right-mouse camera is build into the interface (maybe the controllers?), you can use it while you are in the avatars camera mode. In which I am alot while I'm just editing something. I dont think we're talking about the same camera @huffman .

huffman:
@thoys Yes, that is a different camera. The camera I am talking about is the orbit camera that edit.js uses. It uses right-click+drag for orbiting the camera around a focused object. You can try it by selecting something with edit.js on, then pressing `f`.

huffman:
I just talked to James who clarified a few things.

The reason I am talking about the orbit camera is that this camera will break if you disable right-clicks altogether. After thinking about it more, it sounds like what you really want is to disable right-click for translation, scaling, etc. What I would recommend is modifying entitySelectionTool.js, line 2341, and bailing there if it is not a left click. What do you think?

thoys:
Yes @huffman , that would be great already, although you might also accidentally right click the toolbar, so that might also be a point where to check for left clicks only?

huffman:
@thoys Yes, that's true, there may be other places that handle non-left clicks when they shouldn't. If it's not too tedious I would update those specific cases.

thoys:
:+1: Will do that as soon as I got some time to work on this

huffman:
@thoys Awesome, thanks! :D

ZappoMan:
What's the story on this PR? Should we close it or merge it? @thoys @huffman @imgntn

thoys:
@ZappoMan I didn't have time to make a patch last week, it is still waiting for a patch. I am thinking somewhere this weekend.

thoys:
@ZappoMan this is now merge-able, thanks @huffman  and @imgntn for the review

huffman:
Thanks for fixing @thoys !

PR #6480, opened 2015-11-24
Playback: Trying to get the Agent to Master connection more reliable

samcake:


examples/acScripts/playbackAgents.js
```diff
@@ -104,11 +107,21 @@ function getAction(channel, message, senderID) {
             Agent.isAvatar = false;
             break;
         case LOAD:
-            print("Load");            
-            if(clip_url !== null) {
-                Recording.loadRecording(clip_url);
+            print("Load");   
+            if (!Agent.isAvatar) {
+                Agent.isAvatar = true;
```
ZappoMan: btw - I wonder if this is a problem... I don't know how well ACs work to "become" avatars after they've been running for a while.

PR #6482, opened 2015-11-24
Fix for avatar eye tracking

hyperlogic:
When computing a full eye to world matrix, the translations need to be the geometry coordinate frame, not scaled into meters.

### DIFF
--- a/libraries/animation/src/JointState.h
+++ b/libraries/animation/src/JointState.h
@@ -108,6 +108,7 @@ public:
     const glm::quat& getPostRotation() const { return _postRotation; }
     const glm::quat& getDefaultRotation() const { return _defaultRotation; }
     glm::vec3 getDefaultTranslation() const { return _defaultTranslation * _unitsScale; }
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: e07e1c5c9204eb15b3d9e4895817bf7b104ec951 available at: https://gist.github.com/cd6e17e2ea430f2eaacf<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6482.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6482.dmg

howard-stearns:
Already merged, but now that I think of it, how do we know which, if any, of the many other occurrences of getDefaultTranslationInConstrainedFrame() and similar need to be changed? For example, the occurrence of glm::translate(getJointDefaultTranslationInConstrainedFrame(index)) a few lines above.

hyperlogic:
It's possible, that is the old procedural leaning code.  It might be incorrect as well.

PR #6487, opened 2015-11-25
Get the Leap Motion working again

ctrlaltdavid:
This PR makes it so that the Leap can be used again.  The .js scripts need updating due to recent animation API changes; those will come in another PR.

### DIFF
--- a/interface/src/scripting/ControllerScriptingInterface.cpp
+++ b/interface/src/scripting/ControllerScriptingInterface.cpp
@@ -84,11 +84,11 @@ glm::vec2 ControllerScriptingInterface::getViewportDimensions() const {
     return qApp->getUiSize();
 }
 
-controller::InputController::Pointer ControllerScriptingInterface::createInputController(const QString& deviceName, const QString& tracker) {
-    // This is where we retreive the Device Tracker category and then the sub tracker within it
+controller::InputController* ControllerScriptingInterface::createInputController(const QString& deviceName, const QString& tracker) {
+    // This is where we retrieve the Device Tracker category and then the sub tracker within it
     auto icIt = _inputControllers.find(0);
     if (icIt != _inputControllers.end()) {
-        return (*icIt).second;
+        return (*icIt).second.get();
     } 
 
 
@@ -110,15 +110,15 @@ controller::InputController::Pointer ControllerScriptingInterface::createInputCo
                 controller::InputController::Pointer inputController = std::make_shared<InputController>(deviceID, trackerID, this);
                 controller::InputController::Key key = inputController->getKey();
                 _inputControllers.insert(InputControllerMap::value_type(key, inputController));
-                return inputController;
+                return inputController.get();
             }
         }
     }
 
-    return controller::InputController::Pointer();
+    return nullptr;
 }
 
-void ControllerScriptingInterface::releaseInputController(controller::InputController::Pointer input) {
+void ControllerScriptingInterface::releaseInputController(controller::InputController* input) {
     _inputControllers.erase(input->getKey());
 }
 
--- a/interface/src/scripting/ControllerScriptingInterface.cpp
+++ b/interface/src/scripting/ControllerScriptingInterface.cpp
@@ -89,8 +89,7 @@ controller::InputController* ControllerScriptingInterface::createInputController
     auto icIt = _inputControllers.find(0);
     if (icIt != _inputControllers.end()) {
         return (*icIt).second.get();
-    } 
-
+    }
 
     // Look for device
     DeviceTracker::ID deviceID = DeviceTracker::getDeviceID(deviceName.toStdString());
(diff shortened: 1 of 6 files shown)


ctrlaltdavid:
https://worklist.net/20740

hifi-gustavo:
Build console output for commit: 46cbc349d15d7f103e4568abb42fa183f2e9f7d8 available at: https://gist.github.com/508670bed545e4e53cdf<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6487.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6487.dmg

interface/src/scripting/ControllerScriptingInterface.cpp
```diff
@@ -84,13 +84,12 @@ glm::vec2 ControllerScriptingInterface::getViewportDimensions() const {
     return qApp->getUiSize();
 }
 
-controller::InputController::Pointer ControllerScriptingInterface::createInputController(const QString& deviceName, const QString& tracker) {
-    // This is where we retreive the Device Tracker category and then the sub tracker within it
+controller::InputController* ControllerScriptingInterface::createInputController(const QString& deviceName, const QString& tracker) {
```
ZappoMan: what's this change for?

ZappoMan:
why are you changing the smart pointers into naked pointers?

hifi-gustavo:
Build console output for commit: 6e076985f2ececbb2013cb14d7f72e346ce91394 available at: https://gist.github.com/508670bed545e4e53cdf<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6487.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6487.dmg

ctrlaltdavid:
Smart pointers are still used internally in the C++ (stored in _inputControllers, and used in createInputController() and releaseInputController()); the naked pointers are only passed to JavaScript for its reference and communication back to C++. It was changed because createController() wasn't working.

imgntn:
@ctrlaltdavid hey, i've been testing this on windows 10 with a leap that we have in the office.  the two scripts (laserpointer and leaphands) seem to work, although intermittently.  i'm not sure what the previous baseline functionality was that we're trying to get back to.  if it's simply working at all, then we're good to go (it doesn't work in release in my testing but works in this PR).  as far as being useful/robust, i don't think that these controls schemes are there yet, but i'm testing in a pretty bright office.  any tips as to best ways to test?  my arms often end up in really weird configurations  / loops when i use leaphands.  additionally, you have to uncheck the "Draw Anim Graph" option, which i'm not sure is something we can do programatically...?  thanks!

imgntn:
@ctrlaltdavid - I spoke with @cozza13 about where functionality should be and working at all seems to be a major improvement.  the control script will need a lot of work to be usable (so did the hydra one).  I think we should merge this as long as the C++ stuff @ZappoMan was asking about is good to go.  Then anyone with a leap can start working on the control script.

cozza13:
@ctrlaltdavid @murillodigital which Leap SDK are we using on the build machine?

murillodigital:
@cozza13 looks like (2.2.1+24116) but since this has been not maintained for some time I'd have to double check all servers to confirm consistency.

However, please make it so leap is handled as a cmake external and assume no library is present otherwise.

ctrlaltdavid:
@murillodigital Interface installer currently includes Leap.dll version 2.2.1, presumably with the SDK handled per \HiFi\interface\external\leapmotion\readme.txt (i.e., uses particular version of SDK installed on the build machine).
"handled as a cmake external" means move Leap SDK handling into \HiFi\cmake\externals\, i.e., automatically download latest SDK from Leap Motion when building? If so, merge this PR now and change to using a cmake external as a separate PR?

imgntn:
@ctrlaltdavid i think that's a good plan.  let's merge this and then we'll fund a worklist item to do the external https://worklist.net/20745

ctrlaltdavid:
Changed to "do not merge" because since the PR was submitted, MyAvatar.setJointModelPositionAndOrientation() has been removed from the JavaScript API, meaning that leapHands.js no longer works.

imgntn:
@ctrlaltdavid -- are the C++ changes required for the leap to work at all? if so, i think we could isolate those, merge, and then do another worklist item for making the leaphands script better -- i think we are more interested in moving toward using it to grab / equip items, rather than full finger-level control

ctrlaltdavid:
@imgntn Yes, the C++ changes are required to get the Leap Motion working at all: fixes a Leap Motion InputController not being able to be created and fixes the resulting controller not being updated.
Could certainly merge this PR as is and do a subsequent PR to update leapHands.js to do something else instead of the setJointModelPositionAndOrientation() calls.

PR #6489, opened 2015-11-25
a rough draft of a possible generic actor director library for JS

ZappoMan:
this is just an example of a set of classes I'm thinking would be helpful in building remote controllable agents...

this script shows the usage examples for an Actor and Director class.


examples/exampleActorDirectorLibrary.js
```diff
@@ -0,0 +1,79 @@
...
+//  remote control of several agent like actors and a single director controller
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+
+// Actor class used in agents that can do a job
+// note: type is usually an application name, like com.highfidelity.recordingPlayback
+var me = new Actor("type", {
```
ZappoMan: Note: the idea here is that the constructor for the Actor class, would enumerate the second parameter and mix-in those methods to allow easy callbacks for certain common operations. This basically allows you to implement your derived class of an actor type easily.

Maybe there's a more "JavaScript" way of doing this, but this felt easy and obvious to me.

Feedback welcome for how a "java script developer" would want to do this.

imgntn:
@ZappoMan i like the idea, but its unclear where Actor and Director are defined (are they globals?  includes from another script?)

It's a little weird to call a function with a word that isn't a string, unclear where that constant is coming from   ```var hiredActors = director.getActors(HIRED)```

I think a more typical (although verbose) API might be director.getHiredActors or director.getActors('hired'), or some currying if we want to be super clever.

Also I'm not sure I understand passing those functions as a parameter to the constructor? Maybe I'm missing the connection with another file somewhere?
```
 var Actor = function(type) {
    this.type = type || 'mediocre'
    return this
}

Actor.prototype = {
    hired: function(role) {
        print("I have been hired for the role:" + role);
    },

    fired: function(role) {
        // note: this will also happen if the director has abandoned
        print("I have been fired from the role:" + role);
    },

    messageReceived: function(message) {
        if(this.type==='shatner'){
            message = message.replace(/,/g, '')
            message = message.split(" ");
            message = message.join("...")
        }
        console.log("The director sent me the message:" + message);
    },
}

var me = new Actor('bad');
me.messageReceived = function(message){
  console.log("The director sent me the message:" + message.toUpperCase())
}
me.messageReceived('To be, or not to be...');
var william = new Actor('shatner');
william.messageReceived('To be, or not to be...');
Actor.prototype.retire = function(){
console.log("It's been a good run.")
}
me.retire();
william.retire();
```

ZappoMan:
@imgntn - answers below:

1) *its unclear where Actor and Director are defined* - they will be in an included library file, I'm working on that implementation right now.
2) *little weird to call a function with a word that isn't a string* - yeah, I'm changing that, that was more psuedo-code
3) *Also I'm not sure I understand passing those functions as a parameter to the constructor? Maybe I'm missing the connection with another file somewhere?* - As a user of the Actor/Director library, you can choose to pass in a mixin which will automatically extend the prototype or you can new the Actor and extend the prototype after the fact, both will work.

PR #6490, opened 2015-11-25
Particle speedup

Atlante45:
This PR improves our particle rendering pipeline about 100 fold by moving a lot of the computational load to the GPU.
As a side effect, it temporarily disables spread on color, alpha and radius.

There's still a lot of room for improvements, but it will have to be done in a later pass.


hifi-gustavo:
Build console output for commit: 61143ef757023703676f0f514ce1565fc53f5302 available at: https://gist.github.com/d8f4fa4663cb6354635b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6490.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6490.dmg

libraries/entities-renderer/src/textured_particle.slv
```diff
@@ -10,21 +10,84 @@
...
 
-<@include gpu/Inputs.slh@>
-
 <@include gpu/Transform.slh@>
 
 <$declareStandardTransform()$>
 
-out vec4 _color;
-out vec2 _texCoord0;
+struct Radiuses {
```
hyperlogic: The plural of radius is radii.  English is weird.

libraries/entities/src/ParticleEffectEntityItem.h
```diff
@@ -218,16 +217,27 @@ class ParticleEffectEntityItem : public EntityItem {
     virtual bool supportsDetailedRayIntersection() const { return false; }
 
 protected:
+    struct Particle;
+    using Particles = std::deque<Particle>;
```
hyperlogic: I'm slightly concerned that using a deque here is going to really slow down per-particle processing times due to all the cache misses.  One of the benefits of the previous ring-buffer based approach was cache coherency, at the cost of extra complexity.  Did you have a chance to bench mark the cpu time spent here?

libraries/entities/src/ParticleEffectEntityItem.h
```diff
@@ -218,16 +217,27 @@ class ParticleEffectEntityItem : public EntityItem {
...
-    void extendBounds(const glm::vec3& point);
-    void integrateParticle(quint32 index, float deltaTime);
-    quint32 getLivingParticleCount() const;
-    // the properties of this entity
+    void integrateParticle(Particle& particle, float deltaTime);
+    
+    struct Particle {
+        float seed { 0.0f };
+        float lifetime { 0.0f };
+        glm::vec3 position { Vectors::ZERO};
```
hyperlogic: super minor, but there should be like this right?

    glm::vec3 position { Vectors::ZERO };

hyperlogic:
All the smart pointer fixups really should be in another PR.  Can you update the description to include an overview of the changes.

hyperlogic:
Did we drop support for additive blending?

hyperlogic:
Are the particles sorted by depth?

hyperlogic:
Why did you remove the alpha discard?  That was useful to avoid square particle occlusion when two particle systems overlapped.

libraries/entities/src/ParticleEffectEntityItem.h
Atlante45: A deque is mostly cache coherent (vector of vectors) so the worst case scenario is 1 cache miss per chunk and I'm sure the CPU is able to do some good predictions.

I debated using it for a bit but decided the advantages and ease of use of the deque were worth it, unless we can prove it is a big bottleneck.
I didn't do a proper benchmark but I'd be surprised if it wasn't an improvement on the old implementation that had 1 vector per particle property, it's hard to tell after changing so much of the code.

Atlante45: Just did some research, and it looks like the deque behavior is very similar to a vector:
http://baptiste-wicht.com/posts/2012/12/cpp-benchmark-vector-list-deque.html

Atlante45:
No I made additive blending the only render pipeline.
This way sorting is not necessary and neither is alpha discard (The only reason we needed it in the first place is because the sorting was done in the wrong direction).

The problem is that once we move the position simulation to the GPU, we don't have access to it on the CPU anymore and so we can't sort. I tought

hifi-gustavo:
Build console output for commit: f4d1c116b6e9f5b5ae54874615beee44b40ffdc8 available at: https://gist.github.com/d8f4fa4663cb6354635b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6490.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6490.dmg

libraries/shared/src/GLMHelpers.h
```diff
@@ -154,6 +154,7 @@ vec2 toGlm(const QPointF& pt);
 vec3 toGlm(const xColor& color);
 vec4 toGlm(const QColor& color);
 ivec4 toGlm(const QRect& rect);
+vec4 toGlm(const xColor& color, float alpha);
```
jherico: recommend that alpha have a default parameter value of `1.0f`

Atlante45: Since `vec3 toGlm(const xColor& color);` is already defined, I can't or it will be an ambiguous call.

Atlante45:
@hyperlogic I removed the smart pointers changes to put in a different PR as you suggested.
It didn't have it's place here.

hifi-gustavo:
Build console output for commit: c8ab95bb53f12b7f8e03f8b2e81e97ee7ebc8f48 available at: https://gist.github.com/d8f4fa4663cb6354635b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6490.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6490.dmg

hyperlogic:
@Atlante45 can you resolve the merge conflicts?

hifi-gustavo:
Build console output for commit: 1d513cfbbaec177eb7388c690d0c579d4da6c1a3 available at: https://gist.github.com/d8f4fa4663cb6354635b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6490-win32.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6490.dmg

hifi-gustavo:
Build console output for commit: 1d513cfbbaec177eb7388c690d0c579d4da6c1a3 available at: https://gist.github.com/d8f4fa4663cb6354635b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6490-win32.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6490.dmg

hifi-gustavo:
Build console output for commit: 3316c4eb251fcdf2427ec313646f0f0f1d10e665 available at: https://gist.github.com/d8f4fa4663cb6354635b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6490-win32.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6490.dmg

hifi-gustavo:
Build console output for commit: a1df320ffbfd512c381d37466fdcbe2949f6cbb2 available at: https://gist.github.com/d8f4fa4663cb6354635b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6490-win32.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6490.dmg

hyperlogic:
:+1:

PR #6491, opened 2015-11-25
Restore rendering of one's own avatar.

howard-stearns:
At some point in https://github.com/highfidelity/hifi/pull/6453, I caused one's own avatar to not render. This fixes it.

### DIFF
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -197,12 +197,14 @@ void Avatar::simulate(float deltaTime) {
     if (_shouldSkipRender) {
         if (distance < renderDistance * (1.0f - SKIP_HYSTERESIS_PROPORTION)) {
             _shouldSkipRender = false;
+            _skeletonModel.setVisibleInScene(true, qApp->getMain3DScene());
             if (!isControllerLogging) {  // Test for isMyAvatar is prophylactic. Never occurs in current code.
                 qCDebug(interfaceapp) << "Rerendering" << (isMyAvatar() ? "myself" : getSessionUUID()) << "for distance" << renderDistance;
             }
         }
     } else if (distance > renderDistance * (1.0f + SKIP_HYSTERESIS_PROPORTION)) {
         _shouldSkipRender = true;
+        _skeletonModel.setVisibleInScene(false, qApp->getMain3DScene());
         if (!isControllerLogging) {
             qCDebug(interfaceapp) << "Unrendering" << (isMyAvatar() ? "myself" : getSessionUUID()) << "for distance" << renderDistance;
         }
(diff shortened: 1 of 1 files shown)


interface/src/avatar/Avatar.cpp
```diff
@@ -96,7 +96,6 @@ Avatar::Avatar(RigPointer rig) :
     _moving(false),
     _initialized(false),
     _shouldRenderBillboard(true),
-    _shouldSkipRender(true),
```
howard-stearns: Where the ivar is declared, it is correctly initialized to false.  This one just shouldn't be here at all (no need for two initializations), and the value here gives wrong behavior for one's own avatar.

hifi-gustavo:
Build console output for commit: 546f09c0b1248d909ab5e7fd46fdf2bc4fcd1a28 available at: https://gist.github.com/3a579d0e377504c42232<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6491.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6491.dmg

howard-stearns:
On hold while I check out an unexpected interaction.

howard-stearns:
ok, that should do it.

interface/src/avatar/Avatar.cpp
howard-stearns: This was initialized where defined and also in the constructor, with the later having the wrong value. Removed.

interface/src/avatar/Avatar.cpp
```diff
@@ -609,10 +610,6 @@ void Avatar::fixupModelsInScene() {
     // check to see if when we added our models to the scene they were ready, if they were not ready, then
     // fix them up in the scene
     render::ScenePointer scene = qApp->getMain3DScene();
-    _skeletonModel.setVisibleInScene(!_shouldSkipRender, scene);
-    if (_shouldSkipRender) {
-        return;
-    }
```
howard-stearns: This usage conflicted with other things that set this state (such as as the pause script or Developer->Avatar->Draw Mesh). So now we set the state only when there's a reason to change it (lines 200 and 207).

hifi-gustavo:
Build console output for commit: 6c7099a5f7b23a0a1dbdb10cff9653de18b670d1 available at: https://gist.github.com/3a579d0e377504c42232<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6491.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6491.dmg

hyperlogic:
:+1:

PR #6492, opened 2015-11-27
Isolate handControllerGrab.js changes from bow branch

imgntn:
This PR adds a listener for messages to disable a hand when using a two-handed object.  It also makes the beams entities ungrababble and non-colliding.

### DIFF
--- a/examples/controllers/handControllerGrab.js
+++ b/examples/controllers/handControllerGrab.js
@@ -1226,7 +1226,7 @@ function update() {
 
 Messages.subscribe('Hifi-Hand-Disabler');
 
-handleBeamDisablerMessages = function(channel, message, sender) {
+handleHandDisablerMessages = function(channel, message, sender) {
 
     if (sender === MyAvatar.sessionUUID) {
         handToDisable = message;
@@ -1240,7 +1240,7 @@ handleBeamDisablerMessages = function(channel, message, sender) {
 
 }
 
-Messages.messageReceived.connect(handleBeamDisablerMessages);
+Messages.messageReceived.connect(handleHandDisablerMessages);
 
 function cleanup() {
     rightController.cleanup();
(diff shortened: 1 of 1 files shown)


examples/controllers/handControllerGrab.js
```diff
@@ -1218,12 +1213,35 @@ mapping.from([Controller.Standard.LB]).peek().to(leftController.bumperPress);
...
 }
 
+Messages.subscribe('Hifi-Beam-Disabler');
+
+handleBeamDisablerMessages = function(channel, message, sender) {
+
+    if (sender === MyAvatar.sessionUUID) {
+        handToDisable = message;
+        if (message === 'left') {
+            handToDisable = 1;
```
sethalves: I find it weird that the 1 and 0 constants here suggest the opposite of what I would expect, given the constants at the top:
var RIGHT_HAND = 1;
var LEFT_HAND = 0;

imgntn:
@sethalves i updated the bow api to do the hand switching so disabling should be clearer here now as well

examples/controllers/handControllerGrab.js
```diff
@@ -1218,12 +1212,35 @@ mapping.from([Controller.Standard.LB]).peek().to(leftController.bumperPress);
...
-    leftController.update();
+    if (handToDisable !== LEFT_HAND) {
+        leftController.update();
+    }
+    if (handToDisable !== RIGHT_HAND) {
+        rightController.update();
+    }
 }
 
+Messages.subscribe('Hifi-Hand-Disabler');
```
sethalves: I'm not sure how this works, but if you grab a bow with your left hand, is this going to disable the right hand of everyone in the zone?

examples/controllers/handControllerGrab.js
```diff
@@ -1218,12 +1212,35 @@ mapping.from([Controller.Standard.LB]).peek().to(leftController.bumperPress);
...
+
+handleHandDisablerMessages = function(channel, message, sender) {
+
+    if (sender === MyAvatar.sessionUUID) {
```
sethalves: ah, only if the sender is me.

PR #6500, opened 2015-11-30
Octree crash fix

Atlante45:
Fixes an octree crash we experienced during an alpha meeting.
https://app.asana.com/0/26551373415487/66099014260312

### DIFF
--- a/assignment-client/src/octree/OctreeQueryNode.cpp
+++ b/assignment-client/src/octree/OctreeQueryNode.cpp
@@ -59,7 +59,6 @@ OctreeQueryNode::~OctreeQueryNode() {
 
 void OctreeQueryNode::nodeKilled() {
     _isShuttingDown = true;
-    elementBag.unhookNotifications(); // if our node is shutting down, then we no longer need octree element notifications
     if (_octreeSendThread) {
         // just tell our thread we want to shutdown, this is asynchronous, and fast, we don't need or want it to block
         // while the thread actually shuts down
@@ -69,7 +68,6 @@ void OctreeQueryNode::nodeKilled() {
 
 void OctreeQueryNode::forceNodeShutdown() {
     _isShuttingDown = true;
-    elementBag.unhookNotifications(); // if our node is shutting down, then we no longer need octree element notifications
     if (_octreeSendThread) {
         // we really need to force our thread to shutdown, this is synchronous, we will block while the thread actually
         // shuts down because we really need it to shutdown, and it's ok if we wait for it to complete
--- a/assignment-client/src/octree/OctreeQueryNode.cpp
+++ b/assignment-client/src/octree/OctreeQueryNode.cpp
@@ -336,8 +336,7 @@ void OctreeQueryNode::dumpOutOfView() {
     int stillInView = 0;
     int outOfView = 0;
     OctreeElementBag tempBag;
-    while (!elementBag.isEmpty()) {
-        OctreeElementPointer elementToCheck = elementBag.extract();
+    while (OctreeElementPointer elementToCheck = elementBag.extract()) {
         if (elementToCheck->isInView(_currentViewFrustum)) {
             tempBag.insert(elementToCheck);
             stillInView++;
@@ -346,8 +345,7 @@ void OctreeQueryNode::dumpOutOfView() {
         }
     }
     if (stillInView > 0) {
-        while (!tempBag.isEmpty()) {
-            OctreeElementPointer elementToKeepInBag = tempBag.extract();
+        while (OctreeElementPointer elementToKeepInBag = tempBag.extract()) {
             if (elementToKeepInBag->isInView(_currentViewFrustum)) {
                 elementBag.insert(elementToKeepInBag);
             }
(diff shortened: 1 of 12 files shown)


Atlante45:
UpdateHooks were unused, so I removed them as per coding standard.

Crash was caused by DeleteHooks which were used in one occasion but rendered null, by the SmartPointer change we had a while back.
I removed the the class and use weak pointers to track deleted objects.

ZappoMan:
Make sure to test that when an element is removed from the tree it's actually removed from any pending scene element bags. This is an important feature.

hifi-gustavo:
Build console output for commit: a364e85e1dec5285b0350134274c3e32e080ec6d available at: https://gist.github.com/0f7a65261321571955a4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6500.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6500.dmg

hifi-gustavo:
Build console output for commit: 4aaa9ca02f10f6d51c524e2423a5e59fa1889a62 available at: https://gist.github.com/0f7a65261321571955a4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6500.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6500.dmg

Atlante45:
I tested by running a full stack and restarting interface/entity-server, no issues I could notice.

I tried to make to make the class a bit safer by removing some unused method that could be dangerous to use since other threads can modify the data of the container (i.e. by deleting one of their nodes).
At the very least I made sure the calls to `extract()` don't depend on calls to `isEmpty()` and that we always check the pointer returned by `extract()`.

ZappoMan:
This seems like a perfectly reasonable change... but.... How do you know this fixes the crash from the alpha meeting? Have you found an actual repro cases that causes the same crash from the alpha meeting?

I assume that what you believe was happening was that someone was modifying the tree while a sending thread was busy sending a scene to some viewer. This would exercise this old code of the DeleteHooks.

To test this assumption, and verify that your new code handles the case correctly. You should: run a tree modifying stress test, while simultaneously running a viewer stress test. The viewer stress test is available in an AC script. And there are a couple of different scripts out there that would do rapid modification of the tree.

Atlante45:
I just tried with 15 Acs running the stress test script but I couldn't reproduce. It's a concurrency issue so it'll be hard to reproduce.
However I'm 99% sure this change will fix it based on my research of the stack trace and the source code.

I went into Qt's sources to find exactly where it crashed, it was a hash node pointer access. That means it's a concurrent access to the ElementBag's QSet.

After going through all the access to ElementBags, I could determine that all the Bags are properly used on their own thread, except for the `elementDeleted` callback.
Since I replaced the mechanism by an atomic operation (nulling the weak pointer) and left it to the right thread to cleanup the data structure that stores the pointers, this gets us rid of those concurrent accesses.

assignment-client/src/octree/OctreeQueryNode.cpp
```diff
@@ -338,8 +336,7 @@ void OctreeQueryNode::dumpOutOfView() {
     int stillInView = 0;
     int outOfView = 0;
     OctreeElementBag tempBag;
-    while (!elementBag.isEmpty()) {
-        OctreeElementPointer elementToCheck = elementBag.extract();
+    while (OctreeElementPointer elementToCheck = elementBag.extract()) {
```
ZappoMan: how do you know that `elementToCheck` isn't going to be a null pointer?

assignment-client/src/octree/OctreeSendThread.cpp
```diff
@@ -417,7 +417,7 @@ int OctreeSendThread::packetDistributor(OctreeQueryNode* nodeData, bool viewFrus
             quint64 startInside = usecTimestampNow();
 
             bool lastNodeDidntFit = false; // assume each node fits
-            if (!nodeData->elementBag.isEmpty()) {
```
ZappoMan: what we discussed in chat.... try not to get a shared pointer to a node in the bag unless we know we got a read lock on the tree

hifi-gustavo:
Build console output for commit: 0bc914bf1986bb1792456795efe4e57fc6c0e3a9 available at: https://gist.github.com/0f7a65261321571955a4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6500.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6500.dmg

assignment-client/src/octree/OctreeSendThread.cpp
```diff
@@ -421,12 +421,15 @@ int OctreeSendThread::packetDistributor(OctreeQueryNode* nodeData, bool viewFrus
 
                 quint64 lockWaitStart = usecTimestampNow();
                 _myServer->getOctree()->withReadLock([&]{
+                    OctreeElementPointer subTree = nodeData->elementBag.extract();
```
ZappoMan: we want this code to be after the code to get the `lockWaitEnd` and `lockWaitElapsedUsec` because other wise the stats for how long we waited on locks will be wrong.

hifi-gustavo:
Build console output for commit: c3e2d920ee4180fc6997fd661167bba80788c3ec available at: https://gist.github.com/0f7a65261321571955a4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6500.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6500.dmg

PR #6507, opened 2015-12-01
don't persist the octree if it never loaded

birarda:
- fixes a case where a constantly restarting server could overwrite a good content file with empty content


hifi-gustavo:
Build console output for commit: 7dbf29821afbcb3dc973dab8ff802ddb1a0c1e57 available at: https://gist.github.com/0c42ae7cb27f56d0f242<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6507.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6507.dmg

PR #6511, opened 2015-12-01
Avatar Sim Rate Display fix

hyperlogic:
Division by zero on first update was causing the displayed
avatar simulation rate to be incorrect.


howard-stearns:
Should we get rid of the separate avatar sim rate until such time as avatar threading is restored?

hifi-gustavo:
Build console output for commit: 71f1be2b3804261d73c2db2ee60e19bc9e9dd1d8 available at: https://gist.github.com/22028e05b93dbb6e9a62<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6511.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6511.dmg

hyperlogic:
We should leave it for now.  We might also decide to throttle the updates even in single threaded mode.

PR #6512, opened 2015-12-01
remove debug option for disabling physics updates

AndrewMeadows:
We have a technical writeer who is working on some documentation of our JS API and was puzzling over these calls:

Entities.setSendPhysicsUpdates()
Entities.getSendPhysicsUpdates()

They are JS hooks for turning on an old debug feature that was useful when we were debugging the distributed physics system.

I'm pulling this feature out since it isn't very useful anymore.  When we next have to debug the physics system we'll make new temporary tools.

### DIFF
--- a/examples/utilities/tools/developerMenuItems.js
+++ b/examples/utilities/tools/developerMenuItems.js
@@ -19,7 +19,6 @@ var createdStereoInputMenuItem = false;
 var DEVELOPER_MENU = "Developer";
 
 var ENTITIES_MENU = DEVELOPER_MENU + " > Entities";
-var COLLISION_UPDATES_TO_SERVER = "Don't send collision updates to server";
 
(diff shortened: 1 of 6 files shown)


hifi-gustavo:
Build console output for commit: ab6367ff4e6e486643870c5d85a868663739210e available at: https://gist.github.com/a5afd2777db8bb9c9915<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6512.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6512.dmg

huffman:
Maybe not applicable to this PR, but it would be nice to have a DebugScriptingInterface for this sort of thing so we aren't cluttering the official public API and so we don't need to remove them if they are generally useful.

PR #6513, opened 2015-12-01
Fix ScriptEngine setting _isFinished when it shouldn't, causing scripts to accumulate when restarting

huffman:
There is a race condition, where if a script is stopped before its thread and run() method have been called, it will not be stopped.

### DIFF
--- a/libraries/script-engine/src/ScriptEngine.cpp
+++ b/libraries/script-engine/src/ScriptEngine.cpp
@@ -673,7 +673,6 @@ void ScriptEngine::run() {
     }
 
     _isRunning = true;
-    _isFinished = false;
     if (_wantSignals) {
         emit runningStateChanged();
     }


hifi-gustavo:
Build console output for commit: 4d84362757bbeb0fd7f6f86f5f6768f33ea835ac available at: https://gist.github.com/8d35724f252ea272bc4c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6513.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6513.dmg

PhilipRosedale:
could this cause the master reset script in demo to duplicate objects?

huffman:
Is that a client script or the button on the marquee of the arcade machine? If it's a client script... maybe, but unlikely if you aren't restarting the script a few times in a row. If it's the marquee button, there shouldn't be any relation. If there's a bug report I would be happy to check it out, though.

imgntn:
@huffman what's the test for this?  if i try refreshing default scripts a couple of times quickly i get some doubles and then i crash

huffman:
I didn't try with defaultscripts specifically, but I will. What you did
should be sufficient to test. You could have "duplicates" if one is still
shutting down, but not more than 2. I'll test more. I think there are a
handful of issues related to the asana issue.

imgntn:
(anectodal) i see way fewer (2) type duplicates than before but can still get it to crash pretty quickly.

huffman:
Testing:
1. Testing with an empty script I can't create duplicates, but I can eventually cause it to crash.
2. Testing with defaultScripts.js can create duplicates fairly easily, but I also notice other weird behavior. For instance, I had a duplicate of everything, then reloaded all, and ended up with edit.js and defaultScripts.js. Reloading all multiple more times doesn't make a difference, still only edit.js and defaultScripts.js. Then, loading defaultScripts.js from the script list causes a crash from trying to add a menu item.

I think for 1. we have a race condition where a ScriptEngine can be deleted while the Application still has a reference to it. This would require a bit of refactoring to fix and is a separate issue from what this PR is trying to fix. There could be other issues, but I think this is at least one of them.

For 2, I'll need to look a little more... I'm not sure why we are getting duplicates, and I'm not sure if the scripts are even shutting down.

PhilipRosedale:
If this is an improvement thusfar, should we pull this PR?

huffman:
That sounds fine to me. I can further improve this in a subsequent PR if it is a priority.

PR #6515, opened 2015-12-01
Personal lasers in handControllerGrab.js

imgntn:
This PR makes it so that only you see you distance grabbing laser beams, until you've actually grabbed something.  To test: https://rawgit.com/imgntn/hifi/personal_lasers/examples/controllers/handControllerGrab.js

### DIFF
--- a/examples/controllers/handControllerGrab.js
+++ b/examples/controllers/handControllerGrab.js
@@ -318,6 +318,34 @@ function MyController(hand) {
         });
     }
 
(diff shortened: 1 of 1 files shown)


sethalves:
there's another call to lineOn in continueDistanceHolding.  I don't see other's search beam, but I do see their distance grab beam.  Is that desired?

sethalves:
seems to be by design.

PR #6520, opened 2015-12-02
Make pistol shoot differently and not alter controller mappings

imgntn:
This PR updates the script attached to the pistol entity in toybox so that it doesn't disable everyone's hands.  The previous script would override the controller mapping of anyone nearby.  Now the pistol script doesn't use a custom mapping, and reads the trigger along with a canShoot boolean to tell whether its time to fire.

### DIFF
--- a/examples/toybox/pistol/pistol.js
+++ b/examples/toybox/pistol/pistol.js
@@ -22,6 +22,8 @@
         Controller.Standard.LT,
         Controller.Standard.RT,
     ];
+    var RELOAD_THRESHOLD = 0.95;
+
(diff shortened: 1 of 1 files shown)


examples/toybox/pistol/pistol.js
```diff
@@ -101,7 +116,7 @@
 
         preload: function(entityID) {
             this.entityID = entityID;
-            this.initControllerMapping();
+            // this.initControllerMapping();
```
PhilipRosedale: remove debug?

PR #6524, opened 2015-12-02
Removed collision sound for bow and lowered restitution on flashlight and spraycan to avoid multiple bounce sounds

ericrius1:


ericrius1:
Test: https://rawgit.com/ericrius1/hifi/toyboxSounds/unpublishedScripts/immediateClientReset.js

PR #6525, opened 2015-12-02
Add an avatar LOD preference called "Minimum Avatar Display Distance"

howard-stearns:


hifi-gustavo:
Build console output for commit: 861393910a47a367799567e0641769731ecc4e00 available at: https://gist.github.com/42ebcffaa23ab00125af<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6525.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6525.dmg

interface/src/avatar/AvatarManager.cpp
```diff
@@ -75,6 +76,13 @@ AvatarManager::AvatarManager(QObject* parent) :
     packetReceiver.registerListener(PacketType::AvatarBillboard, this, "processAvatarBillboardPacket");
 }
 
+const float SMALLEST_REASONABLE_HORIZON = 5.0f; // meters
+Setting::Handle<float> avatarRenderDistanceHighLimit("avatarRenderDistanceHighLimit", 1.0f / SMALLEST_REASONABLE_HORIZON);
```
AndrewMeadows: I'm confused by the units of the "DistanceHighLimit" and the "ReasonableHorizon"...

ResonableHorizon is in meters.  Which means "avatarRenderDistanceHighLimit" (stored in settings) is in 1/meters.  Its units are: inverse distance, and yet it is called a "distance".  Why do we invert it before we store it.

Meanwhile, the "DistanceHighLimit" is exposed to JS (it is INVOKABLE)... but its units are 1/m.  Is that clear to the JS user?  Why would the JS API use 1/m?

Finally, the UI in the preferences re-inverts the inverted value so that it can display the settings in meters (horizon).  That makes sense.

howard-stearns: I agree that this is confusing. It's made worse by having existing doc that doesn't show up in the diffs because they didn't change.

Control system values generally have a measured value in which bigger means more work by the "plant" (e.g. fps), and a controlled value in which bigger means the measured value will also go up (e.g., 1/render-distance-cutoff). This is documented in general at the top of PIDController.h, and again specifically for this use at line 150 of AvatarManger.cpp.  Alas, my existing code there sometimes uses "distance" to mean "the distance-related controlled value which is in units of 1/m. :-( 

But that's too confusing for a user preference, so the visible user preference is "Minimum Avatar Display Distance" and displayed in meters, but is immediately converted to "the distance-related controlled value which is in units of 1/m.

So... how should we make this clearer?  I'm not opposed in principle to replacing "distance" with "inverseDistance", so that we would have, e.g., avatarRenderInverseDistanceHighLimit. Not sure that that produces clarity, however.

AndrewMeadows: Ok, so I guess what is throwing me is the name:

avatarRenderDistanceHighLimit... sounds like we're talking about the "max limit of avatar rendering distance"...

but we're really talking about the "high limit of the controller value for measured FPS, relating to 'distance at which to render avatars', which happens to have an inverse relationship with the measured value, which means it is actually a minimum distance limit".

How to make that clearer?  And now that I notice it: setRenderDistanceKD() and setRenderDistanceKI() will not make sense to our average user.  I'm wondering if Karen Bryla has asked you about documenting those exposed JS call yet and whether they are something that we'll want to have in our JS API forever.  It's nice to have such controls exposed to JS and preferences, definitely true for debugging and performance tuning.

Unfortunately I'm not coming up with many better name conventions.  Maybe in JS it should be something like: setMinAvatarRenderCutoffDistance() and then we hide the "control value" aspect from the JS user?

howard-stearns: OK, cool. Here's what I'm thinking:
* Stuff within the PIDController is consistent.
* The AvatarManager (exposed to js) has a render distance controller, which controls the distance beyond which we render. That's mostly all consistent with distance in meters -- except for the existing get/setRenderDistanceLow/HighLimit and the setting avatarRenderDistanceHighLimit introduced in this PR, which actually refer to inverse distance.

So I'll change those 5 names. I like to be able to grep for text, so I propose keeping the "renderDiistance" substring, with: get/setRenderDistanceInverseLow/HighLimit and a avatarRenderDistanceInverseHighLimit setting.

hifi-gustavo:
Build console output for commit: 4a8649c202b5b02f49be6ee4f55f6328d4d466f0 available at: https://gist.github.com/42ebcffaa23ab00125af<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6525.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6525.dmg

PR #6536, opened 2015-12-03
Fix uninitialized xColor value when converting script value

huffman:
The color may never be set if the script value isn't a valid color.


ZappoMan:
what's the repro that tickles this? Shouldn't all users of ` xColor_convertFromScriptValue(const QScriptValue& v, bool& isValid)` be checking the `isValid` result?

hifi-gustavo:
Build console output for commit: 667d2b728e6d7e6ce977b20831ec9399040f3c08 available at: https://gist.github.com/faae85c65128537f0ba3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6536.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6536.dmg

huffman:
I get a runtime check failure when running a debug build from VS when using a bad color value from a script.

PR #6544, opened 2015-12-04
distance-hold adjustment

sethalves:
- adjust how hand motions are exaggerated onto distantly held object
- distantly held objects can be moved with head


sethalves:
test this please

PR #6553, opened 2015-12-04
Blocks dont collide when held

ericrius1:
Prevents blocks from bumping into an avatar holding one.

### DIFF
--- a/unpublishedScripts/hiddenEntityReset.js
+++ b/unpublishedScripts/hiddenEntityReset.js
@@ -1366,7 +1366,11 @@
                         userData: JSON.stringify({
                             resetMe: {
                                 resetMe: true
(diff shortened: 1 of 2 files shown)


ericrius1:
test: https://rawgit.com/ericrius1/hifi/noBlockCollisions/unpublishedScripts/immediateClientReset.js

RyanDowne:
This worked

RyanDowne:
But I wonder now that we have "parents" if we can't make the blocks collide with each other but not an avatar?

PhilipRosedale:
no - parenting doesn't affect the physics engine

PhilipRosedale:
we need the 'collision groups' feature

PR #6565, opened 2015-12-05
Make Avatar::getJointRotation and getJointTranslation thread-safe

hyperlogic:
It can be called from script with minimal blocking, because it inspects a copy of the joint values from the Rig, which is updated atomically. This copy occurs in Rig::updateAnimations()

### DIFF
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -857,18 +857,12 @@ QVector<glm::quat> Avatar::getJointRotations() const {
 }
 
 glm::quat Avatar::getJointRotation(int index) const {
-    if (QThread::currentThread() != thread()) {
-        return AvatarData::getJointRotation(index);
-    }
     glm::quat rotation;
     _skeletonModel.getJointRotation(index, rotation);
     return rotation;
 }
 
 glm::vec3 Avatar::getJointTranslation(int index) const {
-    if (QThread::currentThread() != thread()) {
-        return AvatarData::getJointTranslation(index);
-    }
     glm::vec3 translation;
     _skeletonModel.getJointTranslation(index, translation);
     return translation;
--- a/libraries/animation/src/Rig.cpp
+++ b/libraries/animation/src/Rig.cpp
@@ -14,6 +14,8 @@
 #include <glm/gtx/vector_angle.hpp>
 #include <queue>
 #include <QScriptValueIterator>
+#include <QWriteLocker>
+#include <QReadLocker>
 
 #include <NumericalConstants.h>
 #include <DebugDraw.h>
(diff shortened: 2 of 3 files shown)


interface/src/avatar/Avatar.cpp
```diff
@@ -857,18 +857,12 @@ QVector<glm::quat> Avatar::getJointRotations() const {
 }
 
 glm::quat Avatar::getJointRotation(int index) const {
-    if (QThread::currentThread() != thread()) {
-        return AvatarData::getJointRotation(index);
-    }
```
hyperlogic: Because _skeletonModel.getJointRotation is thread safe, we no longer need to call AvatarData::getJointRotation() which would issue a blocking invokeMethod on the main thread.

libraries/animation/src/Rig.cpp
```diff
@@ -708,21 +712,27 @@ void Rig::updateAnimations(float deltaTime, glm::mat4 rootTransform) {
...
 
     applyOverridePoses();
-    buildAbsoluteRigPoses(_relativePoses, _absolutePoses);
+    buildAbsoluteRigPoses(_internalPoseSet._relativePoses, _internalPoseSet._absolutePoses);
+
+    // copy internal poses to external poses
+    {
+        QWriteLocker writeLock(&_externalPoseSetLock);
+        _externalPoseSet = _internalPoseSet;
+    }
```
hyperlogic: This is where we copy the _internalPoseSet to the _externalPoseSet.

libraries/animation/src/Rig.cpp
```diff
@@ -362,8 +364,9 @@ bool Rig::getJointRotationInWorldFrame(int jointIndex, glm::quat& result, const
 }
 
 bool Rig::getJointRotation(int jointIndex, glm::quat& rotation) const {
-    if (isIndexValid(jointIndex)) {
-        rotation = _relativePoses[jointIndex].rot;
+    QReadLocker readLock(&_externalPoseSetLock);
+    if (jointIndex >= 0 && jointIndex < (int)_externalPoseSet._relativePoses.size()) {
+        rotation = _externalPoseSet._relativePoses[jointIndex].rot;
         return true;
```
hyperlogic: In order to make this method thread-safe, we access the _externalPoseSet *copy*, rather than the _internalPoseSet which is mutated by the main thread, during animations.

hifi-gustavo:
Build console output for commit: 7abfc93ff964cc57db629c1549c5971784b4adf4 available at: https://gist.github.com/61c8f706dad32ef1092e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6565.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6565.dmg

PR #6566, opened 2015-12-05
fix avatar-mixer parents-related crash

sethalves:
- fix avatar-mixer parents-related crash


hifi-gustavo:
Build console output for commit: 62b4eac875fa5e76487965ab828e32dec02dfb61 available at: https://gist.github.com/f44b53caa6928c12722c<br/><br/>No installer was built for this commit

Atlante45:
I don't understand how that would fix this crash.

sethalves:
It's a fix for this one:

  http://public.highfidelity.io/crashes/avatar-mixer-production-3-20151205-010154.txt

The problem is that shared_from_this is being called from a "this" that isn't a pointer -- it's just a member of another class.  You can only call shared_from_this *after* something else has already made a shared pointer to "this".

Atlante45:
Gotcha

PR #6570, opened 2015-12-05
Update Leap Motion script to use animation state handlers

ctrlaltdavid:
Known bug: In desktop mode, restarting the script has an incorrect calibration pose. Work around is to restart Interface.

### DIFF
--- a/examples/controllers/leap/leapHands.js
+++ b/examples/controllers/leap/leapHands.js
@@ -40,7 +40,13 @@ var leapHands = (function () {
         avatarScale,
         avatarFaceModelURL,
         avatarSkeletonModelURL,
-        settingsTimer;
+        settingsTimer,
+        HMD_CAMERA_TO_AVATAR_ROTATION = [
+            Quat.angleAxis(180.0, { x: 0, y: 0, z: 1 }),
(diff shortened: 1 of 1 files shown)


ctrlaltdavid:
Known bug mentioned above is fixed in: https://github.com/highfidelity/hifi/pull/6578

imgntn:
@ctrlaltdavid is this one i need to test with the leap attached to a HMD?  get some weird results with the leap on my desk.  thanks! :)

ctrlaltdavid:
@imgntn Set Developer > Hands > Leap Motion > Leap Motion on HMD according to whether you're using the device on your desk or on your HMD. ... Leap on desk: place your two hands above it in a "zero position" to calibrate it (i.e., similar to what your avatar does when Leap Motion data is first detected).

imgntn:
cool, seems to be working a bit better now -- i'm not sure i understand exactly how to calibrate it (does it do it automatically?  i usually have one hand on the mouse or keyboard to reset the script)
capture

ctrlaltdavid:
All it does is set a reference position for the hands so that it can map Leap-measured hand positions to avatar hand positions (just height and forward/aft position, from memory).

PR #6572, opened 2015-12-06
Fix render rate.

howard-stearns:
Depending on how the threads get assigned, the actual render rate (not just the stats) could be as little as a third of what is targetted.

That's because the code was insisting on an empty texture escrow before our update/render cycle could begin, and otherwise just waits for the next whole cycle.
On my MacbookPro, this was regularly giving a render average of 20fps instead of 60 (and that plays havoc with anything that depends on anything that targets a particular frame rate, such as the avatar render controller).

### DIFF
--- a/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
+++ b/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
@@ -157,7 +157,7 @@ OpenGLDisplayPlugin::OpenGLDisplayPlugin() {
     });
 
     connect(&_timer, &QTimer::timeout, this, [&] {
-        if (_active && _sceneTextureEscrow.depth() < 1) {
+        if (_active && _sceneTextureEscrow.depth() <= 1) {
             emit requestRender();
         }
     });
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 188c477f7e32885b716caa99b6a56fbf380f53f3 available at: https://gist.github.com/1b488768deaab038d8ce<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6572.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6572.dmg

howard-stearns:
@jherico is working on a better way based on an idea from @Atlante45 
Assuming that pans out, this PR would get closed.
Other work is blocking on the resolution, so not closing this yet in case there are unexpected issues today.

howard-stearns:
The conclusion is that its not worth exploring the more involved change at this time, given that there will be further pipeline/thread changes soon. So, this PR will makes the change for Mac only (which is the only platform in which we've seen the artificially low update/simulate/render rates).

hifi-gustavo:
Build console output for commit: 72ce46af5916b79cd940f3c5132d7fb2ddd42a4c available at: https://gist.github.com/1b488768deaab038d8ce<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6572.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6572.dmg

PR #6594, opened 2015-12-08
No physics until ready

sethalves:
- Don't simulate physics until enough of the scene is received over the network.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -195,6 +195,8 @@ static const QString INFO_EDIT_ENTITIES_PATH = "html/edit-commands.html";
 static const unsigned int THROTTLED_SIM_FRAMERATE = 15;
 static const int THROTTLED_SIM_FRAME_PERIOD_MS = MSECS_PER_SECOND / THROTTLED_SIM_FRAMERATE;
 
+static const float PHYSICS_READY_RANGE = 3.0f; // how far from avatar to check for entities that aren't ready for simulation
+
 #ifndef __APPLE__
 static const QString DESKTOP_LOCATION = QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
 #else
@@ -2905,7 +2907,7 @@ void Application::update(float deltaTime) {
 
     _avatarUpdate->synchronousProcess();
 
-    {
+    if (_physicsEnabled) {
         PerformanceTimer perfTimer("physics");
 
         static VectorOfMotionStates motionStates;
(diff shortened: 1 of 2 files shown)


sethalves:
https://app.asana.com/0/26225263936266/66264103181929/f

hifi-gustavo:
Build console output for commit: c462bda86cb221aa7b8e5329f93367a42399d3d7 available at: https://gist.github.com/5824aa46a512fbd551f5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6594.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6594.dmg

interface/src/Application.cpp
```diff
@@ -2905,7 +2907,7 @@ void Application::update(float deltaTime) {
 
     _avatarUpdate->synchronousProcess();
 
-    {
+    if (_physicsEnabled) {
         PerformanceTimer perfTimer("physics");
```
AndrewMeadows: Note: the avatar will be unable to move physically until **_physicsEnabled** is **true**.

What happens if a collision shape URL goes 404 or 500?  Will the object ever be ReadyToComputeShape?  Could a griefer set an invalid collision URL at the spawn point and thereby trap the physics of all incoming avatars?

interface/src/Application.cpp
```diff
@@ -3893,6 +3897,31 @@ void Application::trackIncomingOctreePacket(ReceivedMessage& message, SharedNode
...
+        entityTree->findEntities(box, entities);
+    });
+
+    foreach (EntityItemPointer entity, entities) {
+        if (!entity->isReadyToComputeShape()) {
+            return false;
+        }
+    }
+    return true;
+}
```
AndrewMeadows: This is a step in the right direction, but is not foolproof.

Suppose we had a table with 100 BoxEntityItem's on top.  There is no guarantee that the entity-server will transmit the table first, so it is possible for the receiving interface to be in a state where there are 27 out of 100 boxes nearby that are all ready for physics, but the table data hasn't even been put on the wire yet.

A watertight solution to this problem is not easy.  Do we care?

sethalves: This code is only called from the code that processes stats packets.  Stats packets are only generated after a full scene send.

interface/src/Application.cpp
sethalves: While physics is disabled, the old fallback avatar motion is available.  The user would have to move 3 meters away from the broken object before they could start simulating.

PR #6596, opened 2015-12-08
fix bad simulation ownership when using edit.js

AndrewMeadows:
This PR fixes some bad ownership when using edit.js.  It might also fix some of the simulation flickering seen when multiple interfaces simulate tumbling objects.

### DIFF
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -628,71 +628,50 @@ int EntityItem::readEntityDataFromBuffer(const unsigned char* data, int bytesLef
     bool weOwnSimulation = _simulationOwner.matchesValidID(myNodeID);
 
 
-    if (args.bitstreamVersion >= VERSION_ENTITIES_HAVE_SIMULATION_OWNER_AND_ACTIONS_OVER_WIRE) {
-        // pack SimulationOwner and terse update properties near each other
+    // pack SimulationOwner and terse update properties near each other
 
-        // NOTE: the server is authoritative for changes to simOwnerID so we always unpack ownership data
-        // even when we would otherwise ignore the rest of the packet.
+    // NOTE: the server is authoritative for changes to simOwnerID so we always unpack ownership data
+    // even when we would otherwise ignore the rest of the packet.
 
-        if (propertyFlags.getHasProperty(PROP_SIMULATION_OWNER)) {
+    if (propertyFlags.getHasProperty(PROP_SIMULATION_OWNER)) {
 
-            QByteArray simOwnerData;
-            int bytes = OctreePacketData::unpackDataFromBytes(dataAt, simOwnerData);
-            SimulationOwner newSimOwner;
-            newSimOwner.fromByteArray(simOwnerData);
-            dataAt += bytes;
-            bytesRead += bytes;
+        QByteArray simOwnerData;
+        int bytes = OctreePacketData::unpackDataFromBytes(dataAt, simOwnerData);
+        SimulationOwner newSimOwner;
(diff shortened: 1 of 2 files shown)


libraries/entities/src/EntityItem.cpp
```diff
@@ -628,72 +628,51 @@ int EntityItem::readEntityDataFromBuffer(const unsigned char* data, int bytesLef
...
+    READ_ENTITY_PROPERTY(PROP_GRAVITY, glm::vec3, updateGravity);
+
+    READ_ENTITY_PROPERTY(PROP_DAMPING, float, updateDamping);
+    READ_ENTITY_PROPERTY(PROP_RESTITUTION, float, updateRestitution);
+    READ_ENTITY_PROPERTY(PROP_FRICTION, float, updateFriction);
+    READ_ENTITY_PROPERTY(PROP_LIFETIME, float, updateLifetime);
+    READ_ENTITY_PROPERTY(PROP_SCRIPT, QString, setScript);
+    READ_ENTITY_PROPERTY(PROP_SCRIPT_TIMESTAMP, quint64, setScriptTimestamp);
+    READ_ENTITY_PROPERTY(PROP_REGISTRATION_POINT, glm::vec3, setRegistrationPoint);
+
```
AndrewMeadows: This diff looks complicated, but what is really happening is I'm finally getting around to removing a bunch of cruft for backward version compatibility when the simulation ownership stuff was first added.  These are not the interesting changes you are looking for.

libraries/entities/src/EntityItem.cpp
```diff
@@ -1104,7 +1072,7 @@ bool EntityItem::setProperties(const EntityItemProperties& properties) {
     bool somethingChanged = false;
 
     // these affect TerseUpdate properties
-    SET_ENTITY_PROPERTY_FROM_PROPERTIES(simulationOwner, setSimulationOwner);
+    SET_ENTITY_PROPERTY_FROM_PROPERTIES(simulationOwner, updateSimulationOwner);
```
AndrewMeadows: This is bug #2: simulation ownership on the server was not always setting the DIRTY_SIMULATOR_ID flag, which is used by the **SimpleEntitySimulation** to remember to put it on a list of owned objects.  The simulation owner would be properly set, but the server would not remember to clear out the ownership when the owner logged off.

libraries/entities/src/EntityItem.cpp
```diff
@@ -1324,6 +1292,13 @@ void EntityItem::updatePosition(const glm::vec3& value) {
...
     if (getLocalPosition() != value) {
         setLocalPosition(value);
+        _dirtyFlags |= Simulation::DIRTY_POSITION;
+        forEachDescendant([&](SpatiallyNestablePointer object) {
+            if (object->getNestableType() == NestableTypes::Entity) {
+                EntityItemPointer entity = std::static_pointer_cast<EntityItem>(object);
+                entity->_dirtyFlags |= Simulation::DIRTY_POSITION;
+            }
+        });
     }
```
AndrewMeadows: This is the fix for bug number 1: the **DIRTY_POSITION** flag was not being set on objects who had suffered position updates from outside the simulation --> the **PhysicsEngine** would not update the position of the corresponding **RigidBody**.

hifi-gustavo:
Build console output for commit: 987b8d3334a1136dbce95122159a98650f4216b7 available at: https://gist.github.com/6d39237ff744d7569348<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6596.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6596.dmg

PR #6601, opened 2015-12-09
Fix sync period calc, restoring avatar render distance

howard-stearns:
Replaces Application::getLastDeducedNonVSyncFps() with getLastUnsynchronizedFps() -- both name and implementation -- to match the new threaded paint. Also get rid of getLastPaintWait(), which is so low now that it it isn't worth the complexity.

Uses it in avatar render distance controller, restoring its functionality. (No other changes to controller -- not even the tuning parameters.)


howard-stearns:
Testing note: this PR does _not_ fix "render rate cannot catch up" (https://app.asana.com/0/26225263936266/72197876981925). With that bug in place, it is possible to load up Windows or Linux such that the control system knows that there's headroom to operate, but the actual render rate will be halved (or even thirded occasionally).  There are several options for testing this PR:

* Trust me.
* Test on Mac.
* Don't load up Windows or Linux to the point that this happens -- which might mean you don't get to confirm that the avatar load limiter actually cuts in. You only get to confirm that it doesn't go to minimum when there is, in fact, headroom.
* Locally, work around the referenced bug by using the OS_MAC code within OpenGLDisplayPlugin constructor.

hifi-gustavo:
Build console output for commit: 32e4967ee0a66e8b2f9eba8a06e53b5d553b0d19 available at: https://gist.github.com/a6d10fb26fae8eaf6ed4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6601-win32.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6601.dmg

interface/src/Application.cpp
```diff
@@ -1423,6 +1396,14 @@ void Application::paintGL() {
             batch.resetStages();
         });
     }
+
+    // Some LOD-like controls need to know a smoothly varying "potential" frame rate that doesn't
+    // include time waiting for sync, and which can report a number above target if we've got the headroom.
+    // In my tests, the following is mostly less than 0.5ms, and never more than 3ms. I don't think its worth measuring during runtime.
+    static const float paintWaitAndQTTimerAllowance = 0.001; // seconds
```
AndrewMeadows: As per the coding standards the literal value must end with **f**.

Also, I'm curious: why is the **static** keyword used here rather than "const float VARIABLE_NAME" ?

hifi-gustavo:
Build console output for commit: d964032572544a0eeabc3cc6a8622c8a219873ac available at: https://gist.github.com/a6d10fb26fae8eaf6ed4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6601-win32.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6601.dmg

PR #6606, opened 2015-12-10
[PROTOCOL CHANGE] Entities: Removed Start Automatically property from Animation Group

hyperlogic:
There is an issue when startAutomatically was set to true and animationPlaying was set to false.
This would cause clients to play the animation when they connected to the entity server, even though the server was not.

https://app.asana.com/0/26225263936266/66264103181939


hifi-gustavo:
Build console output for commit: e9f529bf12d300050892eae9f16b5de81ca14d9c available at: https://gist.github.com/99436a3207d8fd983665<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6606-win32.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6606.dmg

libraries/networking/src/udt/PacketHeaders.h
```diff
@@ -161,5 +161,6 @@ const PacketVersion VERSION_ENTITIES_KEYLIGHT_PROPERTIES_GROUP_BIS = 48;
 const PacketVersion VERSION_ENTITIES_PARTICLES_ADDITIVE_BLENDING = 49;
 const PacketVersion VERSION_ENTITIES_POLYLINE_TEXTURE = 50;
 const PacketVersion VERSION_ENTITIES_HAVE_PARENTS = 51;
+const PacketVersion VERSION_ENTITITES_REMOVED_START_AUTOMATICALLY_FROM_ANIMATION_PROPERTY_GROUP = 52;
```
ZappoMan: could we make this longer? :) jk

PR #6614, opened 2015-12-10
Fix whiteboard problems

ZappoMan:
The problem was that the logic for handling repacking attempts was broken and we would keep attempting to repack over and over again even if things wouldn't fit. This change properly handle repack attempts.

### DIFF
--- a/assignment-client/src/octree/OctreeSendThread.cpp
+++ b/assignment-client/src/octree/OctreeSendThread.cpp
@@ -416,27 +416,6 @@ int OctreeSendThread::packetDistributor(OctreeQueryNode* nodeData, bool viewFrus
                         return;
                     }
 
-                    /* TODO: Looking for a way to prevent locking and encoding a tree that is not
-                    // going to result in any packets being sent...
-                    //
-                    // If our node is root, and the root hasn't changed, and our view hasn't changed,
-                    // and we've already seen at least one duplicate packet, then we probably don't need
-                    // to lock the tree and encode, because the result should be that no bytes will be
-                    // encoded, and this will be a duplicate packet from the  last one we sent...
-                    OctreeElementPointer root = _myServer->getOctree()->getRoot();
-                    bool skipEncode = false;
-                    if (
-                            (subTree == root)
-                            && (nodeData->getLastRootTimestamp() == root->getLastChanged())
-                            && !viewFrustumChanged
-                            && (nodeData->getDuplicatePacketCount() > 0)
-                    ) {
-                        qDebug() << "is root, root not changed, view not changed, already seen a duplicate!"
-                            << "Can we skip it?";
-                        skipEncode = true;
-                    }
-                    */
-
                     float octreeSizeScale = nodeData->getOctreeSizeScale();
                     int boundaryLevelAdjustClient = nodeData->getBoundaryLevelAdjust();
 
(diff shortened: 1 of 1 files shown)


assignment-client/src/octree/OctreeSendThread.cpp
```diff
@@ -416,27 +416,6 @@ int OctreeSendThread::packetDistributor(OctreeQueryNode* nodeData, bool viewFrus
                         return;
                     }
 
-                    /* TODO: Looking for a way to prevent locking and encoding a tree that is not
```
ZappoMan: dead code removal

assignment-client/src/octree/OctreeSendThread.cpp
```diff
@@ -471,20 +450,9 @@ int OctreeSendThread::packetDistributor(OctreeQueryNode* nodeData, bool viewFrus
                     // the packet and send it
                     completedScene = nodeData->elementBag.isEmpty();
 
-                    // if we're trying to fill a full size packet, then we use this logic to determine if we have a DIDNT_FIT case.
```
ZappoMan: this really is the bug... there was no reason to special case the "full size packet"... basically if the encode came back as "DIDNT_FIT" then we wanted to drop into the `lastNodeDidntFit` logic. you will also notice the old code wasn't properly incrementing the `extraPackingAttempts`... no in all cases of DIDNT_FIT we consider it a `lastNodeDidntFit` and we track out `extraPackingAttempts`

assignment-client/src/octree/OctreeSendThread.cpp
```diff
@@ -526,8 +493,8 @@ int OctreeSendThread::packetDistributor(OctreeQueryNode* nodeData, bool viewFrus
                 // the packet doesn't have enough space to bother attempting to pack more...
                 bool sendNow = true;
 
-                if (nodeData->getAvailable() >= MINIMUM_ATTEMPT_MORE_PACKING &&
-                    extraPackingAttempts <= REASONABLE_NUMBER_OF_PACKING_ATTEMPTS) {
+                if (!completedScene && (nodeData->getAvailable() >= MINIMUM_ATTEMPT_MORE_PACKING &&
```
ZappoMan: In the case of a completed scene we don't want to set the sendNow to false, although this is probably unrelated to the whiteboard problem

assignment-client/src/octree/OctreeSendThread.cpp
```diff
@@ -539,6 +506,7 @@ int OctreeSendThread::packetDistributor(OctreeQueryNode* nodeData, bool viewFrus
                     packetSendingElapsedUsec = (float)(packetSendingEnd - packetSendingStart);
 
                     targetSize = nodeData->getAvailable() - sizeof(OCTREE_PACKET_INTERNAL_SECTION_SIZE);
+                    extraPackingAttempts = 0;
```
ZappoMan: also only reset `extraPackingAttempts` if we've actually sent the packet using `handlePacketSend()`

assignment-client/src/octree/OctreeSendThread.cpp
```diff
@@ -517,7 +485,6 @@ int OctreeSendThread::packetDistributor(OctreeQueryNode* nodeData, bool viewFrus
                     }
 
                     nodeData->writeToPacket(_packetData.getFinalizedData(), _packetData.getFinalizedSize());
-                    extraPackingAttempts = 0;
```
ZappoMan: this is probably the "money shot" -- basically we kept resetting the `extraPackingAttempts` to `0` and that would mean we kept thinking we had more attempts to repack available... we shouldn't be resetting this until we actually send the packet.

hifi-gustavo:
Build console output for commit: 65e41198bf850752ce850128ccfd896a75aa357e available at: https://gist.github.com/d85eaff9c1ee9dd0b7bb<br/><br/>No installer was built for this commit

hifi-gustavo:
Build console output for commit: 2fbabf20db0879a404f8dafbbdb79c92706bdcab available at: https://gist.github.com/d85eaff9c1ee9dd0b7bb<br/><br/>No installer was built for this commit

murillodigital:
test this please

hifi-gustavo:
Build console output for commit: 2fbabf20db0879a404f8dafbbdb79c92706bdcab available at: https://gist.github.com/d85eaff9c1ee9dd0b7bb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6614-win32.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6614.dmg

PR #6617, opened 2015-12-10
Add keyhole radius to OctreeQuery

huffman:
@ZappoMan As far as I can tell this was never working, at least for the headless viewer. How is interface getting entities that I would have expected it to get from the keyhole?


libraries/networking/src/udt/PacketHeaders.cpp
```diff
@@ -41,7 +41,7 @@ PacketVersion versionForPacketType(PacketType packetType) {
         case PacketType::EntityAdd:
         case PacketType::EntityEdit:
         case PacketType::EntityData:
-            return VERSION_ENTITITES_REMOVED_START_AUTOMATICALLY_FROM_ANIMATION_PROPERTY_GROUP;
+            return VERSION_ENTITIES_QUERY_KEYHOLE_RADIUS;
```
ZappoMan: not sure you need this

hifi-gustavo:
Build console output for commit: 46d645376b7f31dff11493382a59e43c24953aa2 available at: https://gist.github.com/637515d290948b610f97<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6617-win32.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6617.dmg

libraries/networking/src/udt/PacketHeaders.cpp
```diff
@@ -41,7 +41,9 @@ PacketVersion versionForPacketType(PacketType packetType) {
         case PacketType::EntityAdd:
         case PacketType::EntityEdit:
         case PacketType::EntityData:
-            return VERSION_ENTITITES_REMOVED_START_AUTOMATICALLY_FROM_ANIMATION_PROPERTY_GROUP;
+            return VERSION_ENTITIES_REMOVED_START_AUTOMATICALLY_FROM_ANIMATION_PROPERTY_GROUP;
+        case PacketType::EntityQuery:
+            return VERSION_ENTITIES_QUERY_KEYHOLE_RADIUS;
```
huffman: @ZappoMan Is this correct?

hifi-gustavo:
Build console output for commit: cfcc07da2e83e2d93ab111f6d017eacb42ee336c available at: https://gist.github.com/637515d290948b610f97<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6617-win32.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6617.dmg

ZappoMan:
I was thinking more about this, and we could fix this in a way that doesn't require a protocol version change... because these packets are only sent from the client to the server, the result is that users won't be warned that they are talking to an incompatible server version.

A better solution is to put this new float that END of the packet, and make the server check the size of the packet before reading the key hole radius. This would allow old clients to talk to new servers, and new clients talk to old servers (although the old servers will just ignore the keyhole radius, and so the feature won't work until people upgrade their servers)...

What do you think? Can we change the PR to use this technique?

hifi-gustavo:
Build console output for commit: 87e30a2157973550cffd47c845136199be4d401d available at: https://gist.github.com/637515d290948b610f97<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6617-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6617.dmg

huffman:
@ZappoMan Sounds good, I think that will be an improvement. I updated the PR.

ZappoMan:
tested:
- new client to old server
- old client to new server
- new client to new server

PR #6625, opened 2015-12-11
Scriptable anim var triggers

howard-stearns:
Report animation triggers to normal javascript animationStateHandlers, and update away.js as an example.

(As in away.js, this is needed for future scripts that use an intro/idle-hold/outro pattern, such as, e.g. sitting.)


examples/away.js
```diff
@@ -14,7 +14,6 @@
 // Goes into "paused" when the '.' key (and automatically when started in HMD), and normal when pressing any key.
 // See MAIN CONTROL, below, for what "paused" actually does.
```
howard-stearns: The changes in this file are getting rid of the workaround need before we had this capability.

libraries/animation/src/AnimVariant.cpp
```diff
@@ -50,9 +50,11 @@ QScriptValue AnimVariantMap::animVariantMapToScriptValue(QScriptEngine* engine,
     if (useNames) { // copy only the requested names
         for (const QString& name : names) {
             auto search = _map.find(name);
-            if (search != _map.end()) { // scripts are allowed to request names that do not exist
+            if (search != _map.end()) {
                 setOne(name, search->second);
-            }
+            } else if (_triggers.count(name) == 1) {
+                target.setProperty(name, true);
```
howard-stearns: These two lines are the whole PR. _triggers is a set of strings that are "on", so the requested name is a trigger, we just set that property in the target copy that, which the javascript state handler receives.

hifi-gustavo:
Build console output for commit: e8ffa598a25b0d668cefe3397e4478d821d9f025 available at: https://gist.github.com/dfe85d0625e7f36984ac<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6625-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6625.dmg

PR #6626, opened 2015-12-11
Correct collision capsule for scaled avatars

AndrewMeadows:
The size collision capsule now agrees with the avatar's scale.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1231,15 +1231,28 @@ void Application::paintGL() {
             }
         } else if (_myCamera.getMode() == CAMERA_MODE_MIRROR) {
             if (isHMDMode()) {
+                auto mirrorBodyOrientation = myAvatar->getWorldAlignedOrientation() * glm::quat(glm::vec3(0.0f, PI + _rotateMirror, 0.0f));
+
                 glm::quat hmdRotation = extractRotation(myAvatar->getHMDSensorMatrix());
-                _myCamera.setRotation(myAvatar->getWorldAlignedOrientation()
-                    * glm::quat(glm::vec3(0.0f, PI + _rotateMirror, 0.0f)) * hmdRotation);
+                // Mirror HMD yaw and roll
+                glm::vec3 mirrorHmdEulers = glm::eulerAngles(hmdRotation);
+                mirrorHmdEulers.y = -mirrorHmdEulers.y;
+                mirrorHmdEulers.z = -mirrorHmdEulers.z;
+                glm::quat mirrorHmdRotation = glm::quat(mirrorHmdEulers);
+
+                glm::quat worldMirrorRotation = mirrorBodyOrientation * mirrorHmdRotation;
+
+                _myCamera.setRotation(worldMirrorRotation);
+
                 glm::vec3 hmdOffset = extractTranslation(myAvatar->getHMDSensorMatrix());
+                // Mirror HMD lateral offsets
+                hmdOffset.x = -hmdOffset.x;
+
                 _myCamera.setPosition(myAvatar->getDefaultEyePosition()
                     + glm::vec3(0, _raiseMirror * myAvatar->getAvatarScale(), 0)
-                    + (myAvatar->getOrientation() * glm::quat(glm::vec3(0.0f, _rotateMirror, 0.0f))) *
-                    glm::vec3(0.0f, 0.0f, -1.0f) * MIRROR_FULLSCREEN_DISTANCE * _scaleMirror
-                    + (myAvatar->getOrientation() * glm::quat(glm::vec3(0.0f, PI + _rotateMirror, 0.0f))) * hmdOffset);
+                   + mirrorBodyOrientation * glm::vec3(0.0f, 0.0f, 1.0f) * MIRROR_FULLSCREEN_DISTANCE * _scaleMirror
+                   + mirrorBodyOrientation * hmdOffset);
+
             } else {
                 _myCamera.setRotation(myAvatar->getWorldAlignedOrientation()
                     * glm::quat(glm::vec3(0.0f, PI + _rotateMirror, 0.0f)));
@@ -1273,6 +1286,10 @@ void Application::paintGL() {
     // Primary rendering pass
     auto framebufferCache = DependencyManager::get<FramebufferCache>();
     const QSize size = framebufferCache->getFrameBufferSize();
+
+    // Final framebuffer that will be handled to the display-plugin
+    auto finalFramebuffer = framebufferCache->getFramebuffer();
+
     {
         PROFILE_RANGE(__FUNCTION__ "/mainRender");
         PerformanceTimer perfTimer("mainRender");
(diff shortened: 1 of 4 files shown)


interface/src/Application.cpp
```diff
@@ -1237,7 +1237,7 @@ void Application::paintGL() {
                     * glm::quat(glm::vec3(0.0f, PI + _rotateMirror, 0.0f)) * hmdRotation);
                 glm::vec3 hmdOffset = extractTranslation(myAvatar->getHMDSensorMatrix());
                 _myCamera.setPosition(myAvatar->getDefaultEyePosition()
-                    + glm::vec3(0, _raiseMirror * myAvatar->getAvatarScale(), 0)
+                    + glm::vec3(0, _raiseMirror * myAvatar->getUniformScale(), 0)
```
AndrewMeadows: Name change from **Avatar::getAvatarScale()** to **Avatar::getUniformScale()**.

interface/src/avatar/Avatar.cpp
```diff
@@ -99,7 +99,8 @@ Avatar::Avatar(RigPointer rig) :
     // we may have been created in the network thread, but we live in the main thread
     moveToThread(qApp->thread());
 
-    setAvatarScale(1.0f);
+    setScale(glm::vec3(1.0f)); // avatar scale is uniform
```
AndrewMeadows: I removed **Avatar::setAvatarScale()** and rather than replace it with **Avatar::setUniformScale()** I just call **setScale()** directly in two places.

interface/src/avatar/Avatar.cpp
```diff
@@ -143,15 +144,32 @@ AABox Avatar::getBounds() const {
...
+        float animatedScale = currentScale + deltaTime * scaleVelocity;
+        const float MIN_SCALE_SPEED = 0.3f;
+        if (fabsf(scaleVelocity) < MIN_SCALE_SPEED) {
+            // close enough
+            animatedScale = _targetScale;
+        }
+        setScale(glm::vec3(animatedScale)); // avatar scale is uniform
+        rebuildCollisionShape();
+    }
 }
```
AndrewMeadows: The scale animation logic moved out of **setAvatarScale()** and into **animateScaleChanges()**.  At the end it calls **reguildCollisionShape()**, which is virtual and overridden in **MyAvatar**.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -284,13 +284,26 @@ void MyAvatar::update(float deltaTime) {
...
+        float blendFactor = deltaTime / SCALE_ANIMATION_TIMESCALE;
+        float animatedScale = (1.0f - blendFactor) * currentScale + blendFactor * _targetScale;
+        const float CLOSE_ENOUGH = 0.05f;
+        if (fabsf(animatedScale - _targetScale) / _targetScale < CLOSE_ENOUGH) {
+            animatedScale = _targetScale;
+        }
+        setScale(glm::vec3(animatedScale));
+        rebuildCollisionShape();
+    }
+}
```
AndrewMeadows: **animateScaleChanges()** is virtual and overridden in **MyAvatar** right now, but someday **MyAvatar** will have its own **MotionState** and we'll be able to get rid of this implementation and use the version in the base **Avatar** class.

hifi-gustavo:
Build console output for commit: 93a21f333c39a308c617b443d56ec0aa63e7ea92 available at: https://gist.github.com/827aab32864d948c9dfb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6626-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6626.dmg

interface/src/avatar/Avatar.cpp
```diff
@@ -143,15 +144,32 @@ AABox Avatar::getBounds() const {
...
 
 void Avatar::simulate(float deltaTime) {
     PerformanceTimer perfTimer("simulate");
 
-    if (getAvatarScale() != _targetScale) {
-        setAvatarScale(_targetScale);
+    if (!isDead() && !_motionState) {
+        DependencyManager::get<AvatarManager>()->updateAvatarPhysicsShape(this);
     }
+    animateScaleChanges(deltaTime);
```
AndrewMeadows: When an **Avatar** is about to be removed we animate its size toward zero and it doesn't get deleted until it is done shrinking.  We need to be able to tell the difference between avatars that are animating toward departure and those who are just changing their physical size, hence I had to add a data member that could be used to distinguish between these two states.

I added the **isDead** stuff to the base class **AvatarData** because it was easier: **AvatarManger** derives from **AvatarHashMap** which maintains a list of **AvatarData**'s.

AndrewMeadows:
Don't merge this yet.  Ryan found a bug in how the collision shape is rendered and I've noticed a bug where the collision shape is the wrong size when logging in with non-unity scale.  I want to track these down and fix them in this PR.

AndrewMeadows:
Ok, both bugs fixed:

(1) avatar's bounding capsule renders at the correct size

(2) avatar's bounding capsule is correct when logging in with non-unity scale

interface/src/avatar/SkeletonModel.cpp
```diff
@@ -338,35 +338,38 @@ void SkeletonModel::computeBoundingShape() {
...
-            _boundingCapsuleRadius,
-            _boundingCapsuleHeight,
-            _boundingCapsuleLocalOffset);
+    float radius, height;
+    glm::vec3 offset;
+    _rig->computeAvatarBoundingCapsule(geometry, radius, height, offset);
+    float invScale = 1.0f / _owningAvatar->getUniformScale();
+    _boundingCapsuleRadius = invScale * radius;
+    _boundingCapsuleHeight = invScale * height;
+    _boundingCapsuleLocalOffset = invScale * offset;
```
AndrewMeadows: The dimensions of the SkeletonModel::_boundingCapsule* members are invariant under scale operations, but the Rig class will provide scaled measurements, so we have to remove the scale here.

hifi-gustavo:
Build console output for commit: 8e4605784e66961a2b4e254e6ef7f2529c1170c5 available at: https://gist.github.com/827aab32864d948c9dfb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6626-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6626.dmg

interface/src/avatar/SkeletonModel.cpp
```diff
@@ -338,35 +338,38 @@ void SkeletonModel::computeBoundingShape() {
...
         glm::vec4(0.8f, 0.8f, 0.6f, alpha));
 
     // draw a green cylinder between the two points
     glm::vec3 origin(0.0f);
     batch.setModelTransform(Transform().setTranslation(bottomPoint));
     deferredLighting->bindSimpleProgram(batch);
-    Avatar::renderJointConnectingCone(batch, origin, axis, _boundingCapsuleRadius, _boundingCapsuleRadius,
+    Avatar::renderJointConnectingCone(batch, origin, axis, scale * _boundingCapsuleRadius, scale * _boundingCapsuleRadius,
                                       glm::vec4(0.6f, 0.8f, 0.6f, alpha));
 }
```
AndrewMeadows: Since the capsule dimensions are scale-invariant we must scale them when rendering.

libraries/avatars/src/AvatarData.h
```diff
@@ -237,7 +237,7 @@ class AvatarData : public QObject, public SpatiallyNestable {
     //  Scale
     float getTargetScale() const;
     void setTargetScale(float targetScale);
-    void setClampedTargetScale(float targetScale);
+    void setTargetScaleVerbose(float targetScale);
```
AndrewMeadows: I changed the name of this function to correctly represent what it actually does.

AndrewMeadows:
Wtf?  How did I accidentally close this?

huffman:
@AndrewMeadows You should be able to reopen it by clicking the button next to the "Comment" button.

PhilipRosedale:
Yeah I was about to ask the same?  Let's get this one done.

AndrewMeadows:
I accidentally force-pushed an old, unrelated branch on top of the one that this PR was using.  I've re-opened this work in another PR:  https://github.com/highfidelity/hifi/pull/6657

PR #6639, opened 2015-12-12
don't simulate model until it has geometry

sethalves:
- possible fix for collision hulls being in the wrong place

### DIFF
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -455,13 +455,6 @@ bool RenderableModelEntityItem::isReadyToComputeShape() {
             return false; // hmm...
         }
 
-        if (_needsInitialSimulation) {
-            // the _model's offset will be wrong until _needsInitialSimulation is false
-            PerformanceTimer perfTimer("_model->simulate");
-            _model->simulate(0.0f);
-            _needsInitialSimulation = false;
-        }
-
         assert(!_model->getCollisionURL().isEmpty());
     
         if (_model->getURL().isEmpty()) {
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 8ec84bdde826edecdc5260abf6bde4239f4d892e available at: https://gist.github.com/e8e376b959d2cf400700<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6639-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6639.dmg

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -475,6 +468,14 @@ bool RenderableModelEntityItem::isReadyToComputeShape() {
...
             (renderNetworkGeometry && renderNetworkGeometry->isLoaded())) {
             // we have both URLs AND both geometries AND they are both fully loaded.
+
+            if (_needsInitialSimulation) {
+                // the _model's offset will be wrong until _needsInitialSimulation is false
+                PerformanceTimer perfTimer("_model->simulate");
+                _model->simulate(0.0f);
+                _needsInitialSimulation = false;
+            }
+
```
AndrewMeadows: I was wondering: Is it guaranteed that all models have both a **collisionNetworkGeometry** and a **renderNetworkGeometry**?  After looking at the code I've decided the answer is **yes**.

Then I was wondering: Is the **collisionNetworkGeometry** required for computing the correct offset?  Maybe this code should only be wrapped by the checks relating to **renderNetworkGeometry**.

sethalves: It could be, but we only need to "hurry-up" with the initialSimulation if we are about to compute the shape.

PR #6657, opened 2015-12-15
Correct collision capsule for scaled avatars

AndrewMeadows:
The size collision capsule now agrees with the avatar's scale.

Somehow I accidentally invalidated the previous PR... maybe by force pushing the master branch onto the branch.  I was able to salvage the work from my other local sandbox and now am opening a replacement PR.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1260,15 +1260,14 @@ void Application::paintGL() {
                 hmdOffset.x = -hmdOffset.x;
 
                 _myCamera.setPosition(myAvatar->getDefaultEyePosition()
-                    + glm::vec3(0, _raiseMirror * myAvatar->getAvatarScale(), 0)
+                    + glm::vec3(0, _raiseMirror * myAvatar->getUniformScale(), 0)
                    + mirrorBodyOrientation * glm::vec3(0.0f, 0.0f, 1.0f) * MIRROR_FULLSCREEN_DISTANCE * _scaleMirror
                    + mirrorBodyOrientation * hmdOffset);
-
             } else {
                 _myCamera.setRotation(myAvatar->getWorldAlignedOrientation()
                     * glm::quat(glm::vec3(0.0f, PI + _rotateMirror, 0.0f)));
                 _myCamera.setPosition(myAvatar->getDefaultEyePosition()
-                    + glm::vec3(0, _raiseMirror * myAvatar->getAvatarScale(), 0)
+                    + glm::vec3(0, _raiseMirror * myAvatar->getUniformScale(), 0)
                     + (myAvatar->getOrientation() * glm::quat(glm::vec3(0.0f, _rotateMirror, 0.0f))) *
                     glm::vec3(0.0f, 0.0f, -1.0f) * MIRROR_FULLSCREEN_DISTANCE * _scaleMirror);
             }
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -454,7 +454,8 @@ void Avatar::render(RenderArgs* renderArgs, const glm::vec3& cameraPosition) {
         bool renderBounding = Menu::getInstance()->isOptionChecked(MenuOption::RenderBoundingCollisionShapes);
         if (renderBounding && shouldRenderHead(renderArgs) && _skeletonModel.isRenderable()) {
             PROFILE_RANGE_BATCH(batch, __FUNCTION__":skeletonBoundingCollisionShapes");
-            _skeletonModel.renderBoundingCollisionShapes(*renderArgs->_batch, 0.7f);
+            const float BOUNDING_SHAPE_ALPHA = 0.7f;
+            _skeletonModel.renderBoundingCollisionShapes(*renderArgs->_batch, getUniformScale(), BOUNDING_SHAPE_ALPHA);
         }
 
         // If this is the avatar being looked at, render a little ball above their head
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -150,14 +150,17 @@ float Avatar::getLODDistance() const {
 void Avatar::animateScaleChanges(float deltaTime) {
     float currentScale = getUniformScale();
     if (currentScale != _targetScale) {
+        // use exponential decay toward _targetScale
         const float SCALE_ANIMATION_TIMESCALE = 0.5f;
-        float scaleVelocity = (_targetScale - currentScale) / SCALE_ANIMATION_TIMESCALE;
-        float animatedScale = currentScale + deltaTime * scaleVelocity;
-        const float MIN_SCALE_SPEED = 0.3f;
-        if (fabsf(scaleVelocity) < MIN_SCALE_SPEED) {
-            // close enough
+        float blendFactor = glm::clamp(deltaTime / SCALE_ANIMATION_TIMESCALE, 0.0f, 1.0f);
+        float animatedScale = (1.0f - blendFactor) * currentScale + blendFactor * _targetScale;
+
+        // snap to the end when we get close enough
+        const float MIN_RELATIVE_SCALE_ERROR = 0.03f;
+        if (fabsf(_targetScale - currentScale) / _targetScale < 0.03f) {
             animatedScale = _targetScale;
         }
+
         setScale(glm::vec3(animatedScale)); // avatar scale is uniform
         rebuildCollisionShape();
     }
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -167,7 +167,7 @@ void Avatar::simulate(float deltaTime) {
     PerformanceTimer perfTimer("simulate");
 
     if (!isDead() && !_motionState) {
-        DependencyManager::get<AvatarManager>()->updateAvatarPhysicsShape(this);
+        DependencyManager::get<AvatarManager>()->addAvatarToSimulation(this);
     }
     animateScaleChanges(deltaTime);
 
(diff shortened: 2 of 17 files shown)


interface/src/Application.cpp
```diff
@@ -1249,15 +1249,14 @@ void Application::paintGL() {
                 hmdOffset.x = -hmdOffset.x;
 
                 _myCamera.setPosition(myAvatar->getDefaultEyePosition()
-                    + glm::vec3(0, _raiseMirror * myAvatar->getAvatarScale(), 0)
+                    + glm::vec3(0, _raiseMirror * myAvatar->getUniformScale(), 0)
```
AndrewMeadows: Name change from **Avatar::getAvatarScale()** to **Avatar::getUniformScale()**.

interface/src/avatar/Avatar.cpp
```diff
@@ -1021,15 +1040,6 @@ void Avatar::renderJointConnectingCone(gpu::Batch& batch, glm::vec3 position1, g
...
 
-void Avatar::setAvatarScale(float scale) {
-    if (_targetScale * (1.0f - RESCALING_TOLERANCE) < scale &&
-        scale < _targetScale * (1.0f + RESCALING_TOLERANCE)) {
-        setScale(glm::vec3(_targetScale));
-    } else {
-        setScale(glm::vec3(scale));
-    }
-}
-
```
AndrewMeadows: I removed **Avatar::setAvatarScale()** and rather than replace it with **Avatar::setUniformScale()** I just call **setScale()** directly with the proper **vec3**.

interface/src/avatar/Avatar.cpp
```diff
@@ -143,15 +144,32 @@ AABox Avatar::getBounds() const {
...
+        float animatedScale = currentScale + deltaTime * scaleVelocity;
+        const float MIN_SCALE_SPEED = 0.3f;
+        if (fabsf(scaleVelocity) < MIN_SCALE_SPEED) {
+            // close enough
+            animatedScale = _targetScale;
+        }
+        setScale(glm::vec3(animatedScale)); // avatar scale is uniform
+        rebuildCollisionShape();
+    }
 }
```
AndrewMeadows: The scale animation logic moved out of **setAvatarScale()** and into **animateScaleChanges()**. At the end it calls **reguildCollisionShape()**, which is virtual and overridden in **MyAvatar**.

hifi-gustavo:
Build console output for commit: 76d5d9d2c4ab03ef84a0f809ec6c972a56424025 available at: https://gist.github.com/7f5eba6b710d1d3f701c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6657-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6657.dmg

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1032,14 +1045,15 @@ glm::vec3 MyAvatar::getSkeletonPosition() const {
...
+    float scale = getUniformScale();
+    float radius = scale * _skeletonModel.getBoundingCapsuleRadius();
+    float height = scale * _skeletonModel.getBoundingCapsuleHeight() + 2.0f * radius;
     glm::vec3 corner(-radius, -0.5f * height, -radius);
-    corner += _skeletonModel.getBoundingCapsuleOffset();
-    glm::vec3 scale(2.0f * radius, height, 2.0f * radius);
-    _characterController.setLocalBoundingBox(corner, scale);
+    corner += scale * _skeletonModel.getBoundingCapsuleOffset();
+    glm::vec3 diagonal(2.0f * radius, height, 2.0f * radius);
+    _characterController.setLocalBoundingBox(corner, diagonal);
```
AndrewMeadows: Since the capsule dimensions are scale-invariant we must scale them when rendering.

libraries/avatars/src/AvatarData.cpp
```diff
@@ -132,10 +132,7 @@ void AvatarData::setTargetScale(float targetScale) {
...
 }
 
-void AvatarData::setClampedTargetScale(float targetScale) {
-
-    targetScale =  glm::clamp(targetScale, MIN_AVATAR_SCALE, MAX_AVATAR_SCALE);
-
+void AvatarData::setTargetScaleVerbose(float targetScale) {
     setTargetScale(targetScale);
     qCDebug(avatars) << "Changed scale to " << _targetScale;
 }
```
AndrewMeadows: I changed the name of this function to correctly represent what it actually does.

interface/src/avatar/SkeletonModel.cpp
```diff
@@ -338,35 +338,38 @@ void SkeletonModel::computeBoundingShape() {
...
-            _boundingCapsuleRadius,
-            _boundingCapsuleHeight,
-            _boundingCapsuleLocalOffset);
+    float radius, height;
+    glm::vec3 offset;
+    _rig->computeAvatarBoundingCapsule(geometry, radius, height, offset);
+    float invScale = 1.0f / _owningAvatar->getUniformScale();
+    _boundingCapsuleRadius = invScale * radius;
+    _boundingCapsuleHeight = invScale * height;
+    _boundingCapsuleLocalOffset = invScale * offset;
```
AndrewMeadows: The dimensions of the **SkeletonModel::_boundingCapsule** members are invariant under scale operations, but the Rig class will provide scaled measurements, so we have to remove the scale here.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -284,13 +284,26 @@ void MyAvatar::update(float deltaTime) {
...
+        float blendFactor = deltaTime / SCALE_ANIMATION_TIMESCALE;
+        float animatedScale = (1.0f - blendFactor) * currentScale + blendFactor * _targetScale;
+        const float CLOSE_ENOUGH = 0.05f;
+        if (fabsf(animatedScale - _targetScale) / _targetScale < CLOSE_ENOUGH) {
+            animatedScale = _targetScale;
+        }
+        setScale(glm::vec3(animatedScale));
+        rebuildCollisionShape();
+    }
+}
```
AndrewMeadows: **animateScaleChanges()** is virtual and overridden in **MyAvatar** right now, but someday** MyAvatar** will have its own **MotionState** and we'll be able to get rid of this implementation and use the version in the base **Avatar** class.

AndrewMeadows:
test this please

hifi-gustavo:
Build console output for commit: 76d5d9d2c4ab03ef84a0f809ec6c972a56424025 available at: https://gist.github.com/7f5eba6b710d1d3f701c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6657-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6657.dmg

RyanDowne:
Tested on PC and works great. Turned on avatar collision hull and bounds are correct.

libraries/avatars/src/AvatarData.cpp
hyperlogic: So this, function is no longer clamping the scale, is this ok?  I think this is called from the preferences dialog, so we might want to prevent users from typing in negative or huge scales here.

interface/src/avatar/AvatarManager.cpp
```diff
@@ -375,19 +376,16 @@ void AvatarManager::handleCollisionEvents(const CollisionEvents& collisionEvents
 }
 
 void AvatarManager::updateAvatarPhysicsShape(Avatar* avatar) {
```
hyperlogic: Why is this method on AvatarManager instead of Avatar?

interface/src/avatar/MyAvatar.cpp
hyperlogic: So, this does implementation does a lerp over one second, but the implementation in MyAvatar does some sort of non-linear xeno's paradox thing.  Can we make these both use the same animation logic?

interface/src/avatar/AvatarManager.cpp
AndrewMeadows: I will change its name to **AvatarManager::addAvatarToSimulation()** so that it makes more sense.

interface/src/avatar/MyAvatar.cpp
AndrewMeadows: Actually they are both non-linear.  I will clean them up and modify them to be safe even when **deltaTime** is larger than the **timescale**.

libraries/avatars/src/AvatarData.cpp
AndrewMeadows: It calls **setTargetScale()** which does a clamp (using std::max() and std::min()).  I'll modify it to use glm::clamp().

hifi-gustavo:
Build console output for commit: 73952e9002c21ee8bf34b022c11ee644b7319384 available at: https://gist.github.com/7f5eba6b710d1d3f701c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6657-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6657.dmg

hyperlogic:
@AndrewMeadows Looks good, but now there's a merge conflict.

hifi-gustavo:
Build console output for commit: 0e1e5db7eb808a71f209e7c219e74a87b29ddf58 available at: https://gist.github.com/7f5eba6b710d1d3f701c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6657-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6657.dmg

AndrewMeadows:
That's what I get for removing dangling whitespace.

PR #6660, opened 2015-12-15
AnimExpression: boolean expression parser and evaluator

hyperlogic:
The AnimExpression class is intended to be within animation graphs to evaluate state machine transition expressions.  It's not hooked up to the anim system yet, but is a single unit tested class in the animation library.

It supports identifiers from AnimVariantMap and boolean operators in expressions: '&&', '||' and '!', including parenthesis.  The tokenizer and evaluator also has support for floating point numbers and more complex conditional expressions, but these are not yet supported by the parser.

Unit tests are provided.


hifi-gustavo:
Build console output for commit: ab85e2967a9bca036abac6b49c5199fba4d4e54c available at: https://gist.github.com/324f4128c3162021ebb1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6660-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6660.dmg

ZappoMan:
should this be in shared and not "Anim" specific -- but instead more generally available for use?

hyperlogic:
@ZappoMan, yes eventually.  Once it is proven to be robust and more full featured I think it can be moved to shared.  But, currently, it only supports boolean expressions and it has a dependency on the AnimVariantMap.

tests/animation/src/AnimTests.cpp
```diff
@@ -394,3 +394,234 @@ void AnimTests::testAnimPose() {
...
+    TEST_BOOL_EXPR(!true && false || true);
+    TEST_BOOL_EXPR(true && !false || !true);
+    TEST_BOOL_EXPR(true && !false || true);
+    TEST_BOOL_EXPR(true && false || !true);
+    TEST_BOOL_EXPR(true && false || true);
+
+    TEST_BOOL_EXPR(!(true && f) || !t);
+    TEST_BOOL_EXPR(!!!(t) && (!!f || true));
+    TEST_BOOL_EXPR(!(true && f) && true);
+}
```
AndrewMeadows: Love @hyperlogic for tokenizer unit tests.

PR #6662, opened 2015-12-15
Exposing emmisive property of the Image3DOverlay

AlessandroSigna:
This PR allows to set the emmisive property of an Image3DOverlay. In this way via JS it's possible to choose if the created Image3DOverlay will be shaded or not by setting its new 'emmisive' property to false or true.

### DIFF
--- a/interface/src/ui/overlays/Image3DOverlay.cpp
+++ b/interface/src/ui/overlays/Image3DOverlay.cpp
@@ -26,14 +26,16 @@
 QString const Image3DOverlay::TYPE = "image3d";
 
 Image3DOverlay::Image3DOverlay() {
-      _isLoaded = false;
+    _isLoaded = false;
+    _emmisive = false;
 }
 
 Image3DOverlay::Image3DOverlay(const Image3DOverlay* image3DOverlay) :
     Billboard3DOverlay(image3DOverlay),
     _url(image3DOverlay->_url),
     _texture(image3DOverlay->_texture),
-    _fromImage(image3DOverlay->_fromImage)
+    _fromImage(image3DOverlay->_fromImage),
+    _emmisive(image3DOverlay->_emmisive)
 {
 }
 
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: 8e7a2ac5362349aab0c0af4a697733b79a6e493e available at: https://gist.github.com/2a48abe3e755edb97178<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6662-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6662.dmg

interface/src/ui/overlays/Image3DOverlay.cpp
```diff
@@ -26,14 +26,16 @@
 QString const Image3DOverlay::TYPE = "image3d";
 
 Image3DOverlay::Image3DOverlay() {
-      _isLoaded = false;
+    _isLoaded = false;
+    _emmisive = false;
```
huffman: My only comment would be to change `emmisive` to `emissive` (1 m, 2 s) here and in other areas of the code. Other than that the PR looks good!

AlessandroSigna: I totally agree but since I read `emmisive` in [DeferredLightingEffect.h](https://github.com/highfidelity/hifi/blob/8da52e2f17fe7e95d917af2c9b60d4d04bc40e0f/libraries/render-utils/src/DeferredLightingEffect.h) I thought I should use the same name

huffman: Oh, that makes sense. I think since this is user-facing we should at least make sure we get the spelling right in this file, otherwise users will try "emissive" and wonder why it isn't working, and possibly report bugs.

hifi-gustavo:
Build console output for commit: 08f75917876e1e50f7a7c1a4c9f051a12355b117 available at: https://gist.github.com/2a48abe3e755edb97178<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6662-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6662.dmg

PR #6664, opened 2015-12-16
Improved idle to walk forward transition

hyperlogic:
Hooked up a transition animation from idle to walk in the animation json.
Also fixed a bug in the AnimBlendLinearMove which was preventing the interpolation
into walk from being correct.

### DIFF
--- a/interface/resources/meshes/defaultAvatar_full/avatar-animation.json
+++ b/interface/resources/meshes/defaultAvatar_full/avatar-animation.json
@@ -415,10 +415,25 @@
                                                     "states": [
                                                         {
                                                             "id": "idle",
-                                                            "interpTarget": 15,
-                                                            "interpDuration": 15,
+                                                            "interpTarget": 10,
+                                                            "interpDuration": 10,
                                                             "transitions": [
-                                                                { "var": "isMovingForward", "state": "walkFwd" },
+                                                                { "var": "isMovingForward", "state": "idleToWalkFwd" },
+                                                                { "var": "isMovingBackward", "state": "walkBwd" },
+                                                                { "var": "isMovingRight", "state": "strafeRight" },
+                                                                { "var": "isMovingLeft", "state": "strafeLeft" },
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 30480f126bf3d407b40e0f85f70234057bf1d588 available at: https://gist.github.com/a513243751bbc8f81e6f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6664-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6664.dmg

interface/resources/meshes/defaultAvatar_full/avatar-animation.json
```diff
@@ -639,6 +654,18 @@
...
+                                                        "type": "clip",
+                                                        "data": {
+                                                            "url": "http://hifi-content.s3.amazonaws.com/ozan/dev/anim/standard_anims/idle_to_walk.fbx",
+                                                            "startFrame": 1.0,
+                                                            "endFrame": 19.0,
+                                                            "timeScale": 1.0,
+                                                            "loopFlag": false
+                                                        },
+                                                        "children": []
+                                                    },
```
AndrewMeadows: There is now a more pronounced "pop" of the shoulders when the avatar changes from idle to walk, but less foot slip.  I assume that is from the transition animation itself rather than the blend logic.  So... since we now have a new state between **idle** and **walk** we should merge this and fix the animation later?

hyperlogic: We could do two things, merge it now and fix it in the animation, or reduce the interpolation time between the animations, in effect smoothing out the pop.

hyperlogic: After discussing it with Ozan, he's ok with the way it looks for now.

PR #6667, opened 2015-12-16
Entity Sliders for Changing Spotlight Properties

imgntn:
This PR demonstrates one way in-world editing of objects might work. 

Running this script will show light overlay icons in-world.  Enter edit mode by running your distance beam through a light overlay.  Exit using the red X.

When you distant grab the sliders, you can move them along their axis to change their values.  You may also rotate / move the block to which the spotlight is attached.  

To test: https://rawgit.com/imgntn/hifi/light_mod/examples/lights/lightLoader.js
To reset, I recommend stopping all scripts then re-loading lightLoader.js

When you run the lightLoader.js script, several scripts will be loaded:
- handControllerGrab.js (will not impart velocity when you move the parent or a slider, will not move sliders with head movement,will constrain movement for a slider to a given axis start and end)
- lightModifier.js (listens for message to create sliders for a given light.  will start with slider set to the light's initial properties)
- lightModifierTestScene.js (creates a light)
- slider.js (attached to each slider entity)
- lightParent.js (attached to a 3d model of a light, to which a light is parented, so you can move it around.  or keep the current parent if a light already has a parent)
- closeButton.js (for closing the ui)
- ../libraries/lightOverlayManager.js (shows 2d overlays for lights in the world)
- ../libraries/entitySelectionTool.js (visualizes volume of the lights)

Current sliders are (top to bottom):
red
green
blue
intensity
cutoff
exponent

capture


RyanDowne:
I tried this in sandbox and found the following...

When I run the scripts, several of the sliders are rezed in below the sandbox floor. When I try to grab the light entity, it applies some velocity to it and the light speeds away, I cannot bring it towards me. 

Very cool though!

imgntn:
@RyanDowne it comes with its own floor so i'd try it somewhere other than sandbox maybe??  the basePosition is 3 meters below your avatar, in order to account for the floor and then the sliders are distributed with 1.5 meters in front of your avatar and go down 0.2 meters per row.  i can remove the floor and make the sliders start a little higher up.  Any other feedback?  I'm also on slack.

imgntn:
Test this please :) @ericrius1 if you get a chance today / tomorrow that would be great

PR #6670, opened 2015-12-16
Fix entity server race

Atlante45:
OctreeQueryNode used to be guaranteed that it would go down before the server, but that changed recently.
Changed this to a better design where the OctreeServer strongly owns the OctreeSendThreads (instead of OctreeQueryNode) and can clean them up accordingly when it goes down.
This way Nodes don't depend on the server still existing to clean themselves up.

### DIFF
--- a/assignment-client/src/AssignmentClientLogging.cpp
+++ b/assignment-client/src/AssignmentClientLogging.cpp
@@ -1,0 +1,14 @@
+//
+//  AssignmentClientLogging.cpp
+//  assignment-client/src
+//
+//  Created by Clement on 12/14/15.
+//  Copyright 2015 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include "AssignmentClientLogging.h"
+
+Q_LOGGING_CATEGORY(assigmnentclient, "hifi.assignmentclient")
--- a/assignment-client/src/AssignmentClientLogging.cpp
+++ b/assignment-client/src/AssignmentClientLogging.cpp
@@ -11,4 +11,4 @@
 
 #include "AssignmentClientLogging.h"
 
-Q_LOGGING_CATEGORY(assigmnentclient, "hifi.assignmentclient")
+Q_LOGGING_CATEGORY(assigmnentclient, "hifi.assignment-client")
--- a/assignment-client/src/octree/OctreeServer.cpp
+++ b/assignment-client/src/octree/OctreeServer.cpp
@@ -868,6 +868,16 @@ void OctreeServer::parsePayload() {
     }
 }
 
+OctreeServer::UniqueSendThread OctreeServer::createSendThread(const SharedNodePointer& node) {
+    auto sendThread = std::unique_ptr<OctreeSendThread>(new OctreeSendThread(this, node));
+    
+    // we want to be notified when the thread finishes
+    connect(sendThread.get(), &GenericThread::finished, this, &OctreeServer::removeSendThread);
+    sendThread->initialize(true);
+
+    return sendThread;
+}
+
 void OctreeServer::removeSendThread() {
     auto sendThread = static_cast<OctreeSendThread*>(sender());
     
@@ -884,12 +894,7 @@ void OctreeServer::handleOctreeQueryPacket(QSharedPointer<ReceivedMessage> messa
         
         auto it = _sendThreads.find(senderNode->getUUID());
         if (it == _sendThreads.end() || it->second->isShuttingDown()) {
-            auto sendThread = std::unique_ptr<OctreeSendThread>(new OctreeSendThread(this, senderNode));
-            
-            // we want to be notified when the thread finishes
-            connect(sendThread.get(), &GenericThread::finished, this, &OctreeServer::removeSendThread);
-            sendThread->initialize(true);
-            _sendThreads.emplace(senderNode->getUUID(), std::move(sendThread));
+            _sendThreads.emplace(senderNode->getUUID(), createSendThread(senderNode));
         }
     }
 }
--- a/assignment-client/src/octree/OctreeServer.cpp
+++ b/assignment-client/src/octree/OctreeServer.cpp
@@ -897,8 +897,6 @@ void OctreeServer::handleOctreeQueryPacket(QSharedPointer<ReceivedMessage> messa
         if (it == _sendThreads.end()) {
             _sendThreads.emplace(senderNode->getUUID(), createSendThread(senderNode));
         } else if (it->second->isShuttingDown()) {
-            auto& sendThread = *it->second;
-            sendThread.setIsShuttingDown();
             _sendThreads.erase(it); // Remove right away and wait on thread to be
             
             _sendThreads.emplace(senderNode->getUUID(), createSendThread(senderNode));
--- a/assignment-client/src/octree/OctreeServer.cpp
+++ b/assignment-client/src/octree/OctreeServer.cpp
@@ -1460,15 +1460,22 @@ void OctreeServer::didCallWriteDatagram(OctreeSendThread* thread) {
 
 
 void OctreeServer::stopTrackingThread(OctreeSendThread* thread) {
-    QMutexLocker lockerA(&_threadsDidProcessMutex);
-    QMutexLocker lockerB(&_threadsDidPacketDistributorMutex);
-    QMutexLocker lockerC(&_threadsDidHandlePacketSendMutex);
-    QMutexLocker lockerD(&_threadsDidCallWriteDatagramMutex);
-
-    _threadsDidProcess.remove(thread);
-    _threadsDidPacketDistributor.remove(thread);
-    _threadsDidHandlePacketSend.remove(thread);
-    _threadsDidCallWriteDatagram.remove(thread);
+    {
+        QMutexLocker locker(&_threadsDidProcessMutex);
+        _threadsDidProcess.remove(thread);
+    }
+    {
+        QMutexLocker locker(&_threadsDidPacketDistributorMutex);
+        _threadsDidPacketDistributor.remove(thread);
+    }
+    {
+        QMutexLocker locker(&_threadsDidHandlePacketSendMutex);
+        _threadsDidHandlePacketSend.remove(thread);
+    }
+    {
+        QMutexLocker locker(&_threadsDidCallWriteDatagramMutex);
+        _threadsDidCallWriteDatagram.remove(thread);
+    }
 }
 
 int howManyThreadsDidSomething(QMutex& mutex, QMap<OctreeSendThread*, quint64>& something, quint64 since) {
--- a/assignment-client/src/octree/OctreeServer.cpp
+++ b/assignment-client/src/octree/OctreeServer.cpp
@@ -868,16 +868,28 @@ void OctreeServer::parsePayload() {
     }
 }
 
+void OctreeServer::removeSendThread() {
+    auto sendThread = static_cast<OctreeSendThread*>(sender());
+    
+    // This deletes the unique_ptr, so sendThread is destructed after that line
+    _sendThreads.erase(sendThread->getNodeUuid());
+}
+
 void OctreeServer::handleOctreeQueryPacket(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode) {
-    if (!_isFinished) {
+    if (!_isFinished && !_isShuttingDown) {
         // If we got a query packet, then we're talking to an agent, and we
         // need to make sure we have it in our nodeList.
         auto nodeList = DependencyManager::get<NodeList>();
         nodeList->updateNodeWithDataFromPacket(message, senderNode);
         
-        OctreeQueryNode* nodeData = dynamic_cast<OctreeQueryNode*>(senderNode->getLinkedData());
-        if (nodeData && !nodeData->isOctreeSendThreadInitalized()) {
-            nodeData->initializeOctreeSendThread(this, senderNode);
+        auto it = _sendThreads.find(senderNode->getUUID());
+        if (it == _sendThreads.end() || it->second->isShuttingDown()) {
+            auto sendThread = std::unique_ptr<OctreeSendThread>(new OctreeSendThread(this, senderNode));
+            
+            // we want to be notified when the thread finishes
+            connect(sendThread.get(), &GenericThread::finished, this, &OctreeServer::removeSendThread);
+            sendThread->initialize(true);
+            _sendThreads.emplace(senderNode->getUUID(), std::move(sendThread));
         }
     }
 }
@@ -1157,6 +1169,12 @@ void OctreeServer::nodeAdded(SharedNodePointer node) {
 
 void OctreeServer::nodeKilled(SharedNodePointer node) {
     quint64 start  = usecTimestampNow();
+    
+    // Shutdown send thread
+    auto it = _sendThreads.find(node->getUUID());
+    if (it != _sendThreads.end()) {
+        it->second->setIsShuttingDown();
+    }
 
     // calling this here since nodeKilled slot in ReceivedPacketProcessor can't be triggered by signals yet!!
     _octreeInboundPacketProcessor->nodeKilled(node);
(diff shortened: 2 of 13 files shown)


hifi-gustavo:
Build console output for commit: 47e12ab1b6bc99c9a7a3a3ca3ac574acb3632b7c available at: https://gist.github.com/f6e01ebcc8d2ecdf489b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6670-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6670.dmg

birarda:
:+1:

assignment-client/src/AssignmentClientLogging.cpp
```diff
@@ -0,0 +1,14 @@
...
+//  Created by Clement on 12/14/15.
+//  Copyright 2015 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include "AssignmentClientLogging.h"
+
+Q_LOGGING_CATEGORY(assigmnentclient, "hifi.assignmentclient")
```
birarda: `assignment-client`

assignment-client/src/octree/OctreeServer.cpp
```diff
@@ -1226,27 +1215,20 @@ void OctreeServer::aboutToFinish() {
...
-    
-    // What follows is a hack to force OctreeSendThreads to cleanup before the OctreeServer is gone.
-    // I would prefer to allow the SharedNodePointer ref count drop to zero to do this automatically
-    // but that isn't possible as long as the OctreeSendThread has an OctreeServer* that it uses.
-    for (auto& node : nodesToShutdown) {
-        qDebug() << qPrintable(_safeServerName) << "server about to finish while node still connected node:" << *node;
-        forceNodeShutdown(node);
+    // Shut down all the send threads
+    for (auto it = _sendThreads.begin(); it != _sendThreads.end(); ++it) {
+        it->second->disconnect(this); // Disconnect so that removeSendThread doesn't get called later
```
ZappoMan: use a named variable that is more obvious

assignment-client/src/octree/OctreeServer.cpp
```diff
@@ -878,16 +868,28 @@ void OctreeServer::parsePayload() {
...
         nodeList->updateNodeWithDataFromPacket(message, senderNode);
         
-        OctreeQueryNode* nodeData = dynamic_cast<OctreeQueryNode*>(senderNode->getLinkedData());
-        if (nodeData && !nodeData->isOctreeSendThreadInitalized()) {
-            nodeData->initializeOctreeSendThread(this, senderNode);
+        auto it = _sendThreads.find(senderNode->getUUID());
+        if (it == _sendThreads.end() || it->second->isShuttingDown()) {
+            auto sendThread = std::unique_ptr<OctreeSendThread>(new OctreeSendThread(this, senderNode));
+            
+            // we want to be notified when the thread finishes
```
ZappoMan: this code is from the original `OctreeQueryNode::initializeOctreeSendThread()` -- move this code into a helper function/method, so that it's obvious that it's the initialization of the send thread.

assignment-client/src/octree/OctreeServer.cpp
```diff
@@ -878,16 +868,28 @@ void OctreeServer::parsePayload() {
...
         // If we got a query packet, then we're talking to an agent, and we
         // need to make sure we have it in our nodeList.
         auto nodeList = DependencyManager::get<NodeList>();
         nodeList->updateNodeWithDataFromPacket(message, senderNode);
         
-        OctreeQueryNode* nodeData = dynamic_cast<OctreeQueryNode*>(senderNode->getLinkedData());
-        if (nodeData && !nodeData->isOctreeSendThreadInitalized()) {
-            nodeData->initializeOctreeSendThread(this, senderNode);
+        auto it = _sendThreads.find(senderNode->getUUID());
+        if (it == _sendThreads.end() || it->second->isShuttingDown()) {
```
ZappoMan: this second clause seems very problematic... did you actually have a case where this occurred? Is it a valid case?

I believe what it's saying is that we had previously been serving to node with ID: "xxx-123-..." and we got a new query message from it, but it was also in the "shuttingDown" case... how can that happen? The only time we should still be in the `_sendThreads` list, but in the `isShuttingDown` state, would be after the `nodeKilled` signal happened... it could be that some query messages are still in flight, in which case we will fire up a new sending thread... do we really want to do that?

assignment-client/src/octree/OctreeServer.cpp
```diff
@@ -1167,6 +1169,12 @@ void OctreeServer::nodeAdded(SharedNodePointer node) {
 
 void OctreeServer::nodeKilled(SharedNodePointer node) {
     quint64 start  = usecTimestampNow();
+    
+    // Shutdown send thread
+    auto it = _sendThreads.find(node->getUUID());
```
ZappoMan: I wish this pattern would be cleaner looking... maybe we should implement an `if_found()` lambda that does the find, checks against `it != x.end()` and passes you "value" which is the `it->second`... it would really clean up this code and other uses of `std::unordered_map<>`

hifi-gustavo:
Build console output for commit: cac800592b9da32906508a05e08368bde53f0528 available at: https://gist.github.com/f6e01ebcc8d2ecdf489b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6670-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6670.dmg

assignment-client/src/octree/OctreeServer.cpp
birarda: Answering since I was helping Clement with this yesterday.

The case you describe will not occur since the `EntityQuery` packet is sourced and would be dropped for a dead node.

The case that you want `it->second->isShuttingDown()` for is the case where you kill the node and immediately re-add it (a blip in connection). In that case, you want a new send thread to start talking to the newly added node again, but it's possible you haven't cleaned up the previous OctreeSendThread. This handles that case and replaces the previously not-finished OctreeSendThread once it is gone.

assignment-client/src/octree/OctreeServer.cpp
```diff
@@ -1226,27 +1215,20 @@ void OctreeServer::aboutToFinish() {
...
+    
+    // Wait on all send threads to be done before continuing
+    _sendThreads.clear();
 
     if (_persistThread) {
         _persistThread->aboutToFinish();
         _persistThread->terminating();
     }
-
+
```
ZappoMan: didn't we agree to no trailing whitespace?

assignment-client/src/octree/OctreeServer.cpp
```diff
@@ -1226,27 +1215,20 @@ void OctreeServer::aboutToFinish() {
...
-    });
-    
-    // What follows is a hack to force OctreeSendThreads to cleanup before the OctreeServer is gone.
-    // I would prefer to allow the SharedNodePointer ref count drop to zero to do this automatically
-    // but that isn't possible as long as the OctreeSendThread has an OctreeServer* that it uses.
-    for (auto& node : nodesToShutdown) {
-        qDebug() << qPrintable(_safeServerName) << "server about to finish while node still connected node:" << *node;
-        forceNodeShutdown(node);
+    // Shut down all the send threads
+    for (auto it = _sendThreads.begin(); it != _sendThreads.end(); ++it) {
```
ZappoMan: this is another example where we should provide `std::unordered_map` with a cleaner `for_each` or something to iterate across the values...

assignment-client/src/octree/OctreeServer.cpp
ZappoMan: but when you "blip" the connection... does the node retain the same ID? Does that happen? If it's really the same ID and the same client, then wouldn't it be preferable to reuse the original send thread? Maybe you can't "resurrect" the send thread easily...

birarda: When the node has stayed connected to the domain but disconnected from the entity-server, it would show back up with the same ID. 

The thinking is that resurrecting the send thread that's already trying to go down would be both dangerous and complex. Cleaner to let it finish its already started shutdown procedure and set up a new one.

assignment-client/src/octree/OctreeServer.cpp
```diff
@@ -1226,27 +1215,20 @@ void OctreeServer::aboutToFinish() {
...
+        it->second->disconnect(this); // Disconnect so that removeSendThread doesn't get called later
+        it->second->setIsShuttingDown();
     }
+    
+    // Wait on all send threads to be done before continuing
```
ZappoMan: can you expand this comment, it's true that calling `.clear()` will delete all the `OctreeSendThreads` which does call `GenericThread::terminate()` that does wait on the `QThread` to finish... but that's not obvious from this one line, and I had to dig 4 functions deep to confirm that. If someone later changes this behavior we might have a case of a dangling pointer back to the Server via the OctreeSendThread's raw pointer.

assignment-client/src/octree/OctreeServer.cpp
Atlante45: Actually I used to have a QHash, which is why I had that setup. Now I can just do `for (auto it : _sendThreads)` I think

assignment-client/src/octree/OctreeServer.cpp
ZappoMan: how does a node "stay connected to a domain" --- but --- "disconnected from the entity-server"? The entity server is trusting "NodeList" to say who's connected to the domain... This is what I don't understand.

Did you guys see this happen in testing? 

If this is a theoretical case that we want to handle to make the code more robust, then cool... I get it... but I'm just trying to understand how does this case of connect to domain but not connect to entity server happen?

assignment-client/src/octree/OctreeServer.cpp
ZappoMan: :+1:

hifi-gustavo:
Build console output for commit: ea5a29c1b082e10b257b8e4ee1cac139c3ab7238 available at: https://gist.github.com/f6e01ebcc8d2ecdf489b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6670-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6670.dmg

assignment-client/src/octree/OctreeServer.cpp
birarda: Given that a domain-server and an entity-server could be on completely different machines and are, at minimum, using different sockets to communicate - it's absolutely possible that a node maintains its connection to the domain-server while momentarily losing it to an entity-server.

This is a theoretical case we knew we'd need to handle.

hifi-gustavo:
Build console output for commit: 2ce0bdc5a6eb89d63cc8f13c472c65692bd6aa23 available at: https://gist.github.com/f6e01ebcc8d2ecdf489b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6670-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6670.dmg

hifi-gustavo:
Build console output for commit: 93443cb09930dcfb2af87eca268b48f35106cc46 available at: https://gist.github.com/f6e01ebcc8d2ecdf489b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6670-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6670.dmg

assignment-client/src/octree/OctreeServer.cpp
ZappoMan: ok - so after walking through this some more, I see how this case can happen... the new design introduces a blocking case on the Server if a client gets into a flapping case like this.

The call below... `_sendThreads.emplace(senderNode->getUUID()...)` will block the server until the OLD send thread is completed. (emplace will cause the map's value dtr to be called which in the case of OctreeSendThread will do a wait on it's thread until complete).

During this time, the server will not be able to service any additional incoming messages (queries, etc). The old design didn't have this blocking case, and handles this node flapping case differently: nodeList would effectively delete the old SendThread (since it was owned by the Node), and would create a new SendThread when the new instance of the node was added.

How does the `removeSendThread` signal get handled in this case? 

```
void OctreeServer::removeSendThread() {
    auto sendThread = static_cast<OctreeSendThread*>(sender());
    
    // This deletes the unique_ptr, so sendThread is destructed after that line
    _sendThreads.erase(sendThread->getNodeUuid());
}
```

How does `std::unordered_map<>::emplace()` handle the dtr of the OLD value vs the storing of the new key/value? 

As I understand the new design... if the OLD send thread is still shutting down, then we will potentially hit the `_sendThreads.emplace()` call before the `OctreeSendThread::finished` signal has fired... during the `emplace()` the OctreeSendThread dtr will run, which will wait for the thread to finish, ultimately causing `finished` signal to fire, and `removeSendThread` to be called, which removes the nodeID from `_sendThreads`

Do we know that the dtr is called before the new (nodeID/newThread) pair is added to the map? Is that guaranteed by the STL?

ZappoMan:
I definitely like this new approach. Let's make sure we can confirm there isn't a race condition in the nodeKilled/nodeAdded flapping case discussed above... if we are certain that `removeSendThread` won't remove the newly added id/sendthread pair, then this is cool.

assignment-client/src/octree/OctreeServer.cpp
ZappoMan: btw - in re-reading the docs on `std::unordered_map<>::emplace()` it's not clear to me that it will "replace" the existing key with the new key/value pair... if that's the case, then... well... does this new code work at all in the case we're discussing?

ZappoMan: quick test of emplace() suggests that it does not replace the value if the key exists....
```
void testEmplace() {
    qDebug() << "------------------ START testEmplace() -------------------------";

    std::unordered_map<int, int> map;
    qDebug() << "map.emplace(1, 100);";
    map.emplace(1, 100);
    auto it = map.find(1);
    if (it != map.end()) {
        qDebug() << "map at 1:" << it->second;
    }
    qDebug() << "map.emplace(1, 200);";
    map.emplace(1, 200);
    it = map.find(1);
    if (it != map.end()) {
        qDebug() << "map at 1:" << it->second;
    }


    qDebug() << "------------------ DONE testEmplace() -------------------------";
}
```

results...

```
[12/16 13:07:56] [DEBUG] ------------------ START testEmplace() -------------------------
[12/16 13:07:56] [DEBUG] map.emplace(1, 100);
[12/16 13:07:56] [DEBUG] map at 1: 100
[12/16 13:07:56] [DEBUG] map.emplace(1, 200);
[12/16 13:07:56] [DEBUG] map at 1: 100
[12/16 13:07:56] [DEBUG] ------------------ DONE testEmplace() -------------------------
```

Atlante45: Nice catch!
Did not know that emplace/insert had this behavior. Got corrupted by Qt's containers.

Atlante45:
@ZappoMan Regarding the race mentioned above, I tweaked the code to take care of it.
`removeSendThread` checks the result of `sender()` (which it should have from the start)
This will return nullptr if the object has already be destructed, so we won't delete newly added threads.

hifi-gustavo:
Build console output for commit: b709c8162c71a05a099273c60db01fba66a5010d available at: https://gist.github.com/f6e01ebcc8d2ecdf489b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6670-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6670.dmg

hifi-gustavo:
Build console output for commit: 2c8be15e6671df5e74f0241d1185d119a504df16 available at: https://gist.github.com/f6e01ebcc8d2ecdf489b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6670-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6670.dmg

ZappoMan:
I try running the SM that comes with this installer on Windows 8.1 and it will not launch Script Assignments from the Domain Web page... Production Stack Manager works fine... I will test another PR to see if this is a problem with PR builds or this particular PR.

PR #6673, opened 2015-12-16
Make Avatar palm position/rotation accessors thread safe

hyperlogic:
These are called directly from script threads and can result in out of bounds array accesses when called while the main thread is initializing the rig.

I was able to reproduce this in debug builds.


hifi-gustavo:
Build console output for commit: 2d4180f06503870304b88bbb6183e6564e66fca9 available at: https://gist.github.com/27add1a595f93a642a2a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6673-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6673.dmg

ZappoMan:
Is there a way to do this without the blocking queued connection?

sethalves:
I've tried this and it passes a smoke-test.

hyperlogic:
@ZappoMan yes, it is possible.  I'll amend this PR, when I get a chance.

jherico:
@hyperlogic @ZappoMan you can't return get a return value from an invokation unless you use a blocking queued connection.  The alternative is to make the Avatar accessors inherently thread safe, which is more danger prone.  Since the avatars live on the main thread, the only thing that would get blocked by this would be a script or other ancillary thread.  Why would it matter if they're blocked for a frame?

hyperlogic:
I introduced a ThreadSafeValueCache template class to abstract the actual lock and state juggling necessary to provide a thread safe access to values in a cache.  I believe this can be a general mechanism for handling values that need to be get/set from script or other threads.

This seems to work well for the palm rotation and position script getter methods.

Unit tests coming soon...

hifi-gustavo:
Build console output for commit: feb63772f39b6ff4c544b5d340ab38d9688f4917 available at: https://gist.github.com/27add1a595f93a642a2a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6673.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6673.dmg

libraries/shared/src/ThreadSafeValueCache.h
```diff
@@ -0,0 +1,73 @@
+//
+//  ThreadSafeValueCache.h
```
ZappoMan: I don't understand why you've implemented this notion of "pending" -- why not just use the mutex mechanism and allow safe sets/gets. As far as I can tell your calls to merge() in the Avatar class always over-write the actual value. It's not clear to me that you really want that behavior (for some period of time the "pending value" is used, until the Avatar is simulated and then the skeleton's joint value is used)... but if you DID want that behavior, it seems you could accomplish the same thing by simply setting the value during the `Avatar::updatePalms()`

Basically, I don't see how the "pending" and `merge()` features are giving you any value, it seems the code would be much simpler and obvious by just having a get/set that does the appropriate mutex locking.

hyperlogic: I created this class to help with a more general case than is actually needed by the palms. For the palms, the merge functionality is "accept-theirs".  i.e. script always overrides the c++.  In the more general case, you might want different behavior, such as:

1. C++ side should take precedence until, script actually sets the value.  In that case the merge lambda would look like this:
    <pre>
    cache.merge(\[&\](const V& value, bool hasPending, const V& pendingValue) {
        if (hasPending) {
            internalCppValue = pendingValue;
            return pendingValue;
        } else {
            return internalCppValue;
        }
    });
    // use internalCppValue for something here
    </pre>
2. There is some actual merge that can be done for example for velocities.
    <pre>
    cache.merge([&](const V& value, bool hasPending, const V& pendingValue) {
        if (hasPending) {
            internalCppValue += pendingValue;
        }
        return internalCppValue;
    });
    </pre>

After thinking about it, I think we should have different templates for the simple case vs the complex case.  I'll go ahead and revise this.

hyperlogic:
I simplified the implementation of ThreadSaveValueCache by removing the pendingValue and pendingFlag.  If we need the more deluxe version of this we can create another template with another name.  Perhaps, as an optimization, this class can perhaps be implemented using std::atomic<T> and compare and swap.  But I'm not an expert on C++11/14 locking primitives...

hifi-gustavo:
Build console output for commit: a21b6047460a28e5def9df3b6fda48aed385b39c available at: https://gist.github.com/27add1a595f93a642a2a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6673.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6673.dmg

ZappoMan:
hmmmm--- quick smoke test of this should an FPS of 10 in demo... I'm checking master now to see if it runs ok. But that's not good.

ZappoMan:
ok - this doesn't look worse than master... I'm going to merge.

PR #6675, opened 2015-12-16
fix crash when visiting new domain with gun in hand

AndrewMeadows:
The bug was: sometimes when leaving one domain for another while holding the ping-pong gun the interface would crash.

The problem was one of the lists (of MotionStates) inside **PhysicalEntitySimulation** was not properly being cleared, leaving a dangling pointer that would be dereferenced later.

### DIFF
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -385,7 +385,7 @@ Model* RenderableModelEntityItem::getModel(EntityTreeRenderer* renderer) {
             _needsInitialSimulation = true;
         }
     }
-    
+
     return result;
 }
 
@@ -398,14 +398,14 @@ void RenderableModelEntityItem::update(const quint64& now) {
         EntityItemProperties properties;
         auto extents = _model->getMeshExtents();
         properties.setDimensions(extents.maximum - extents.minimum);
-        
+
         qCDebug(entitiesrenderer) << "Autoresizing:" << (!getName().isEmpty() ? getName() : getModelURL());
         QMetaObject::invokeMethod(DependencyManager::get<EntityScriptingInterface>().data(), "editEntity",
                                   Qt::QueuedConnection,
                                   Q_ARG(QUuid, getEntityItemID()),
                                   Q_ARG(EntityItemProperties, properties));
     }
-    
+
     ModelEntityItem::update(now);
 }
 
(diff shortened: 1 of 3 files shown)


libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -447,24 +447,22 @@ bool RenderableModelEntityItem::isReadyToComputeShape() {
...
 
     if (type == SHAPE_TYPE_COMPOUND) {
-        if (!_model) {
+        if (!_model || _model->getCollisionURL().isEmpty()) {
             EntityTreePointer tree = getTree();
             if (tree) {
                 QMetaObject::invokeMethod(tree.get(), "callLoader", Qt::QueuedConnection, Q_ARG(EntityItemID, getID()));
             }
-            return false; // hmm...
+            return false;
```
AndrewMeadows: These changes fix a crash bug (assert really) when a model would sometimes be late to inherit its **collisionURL** from its ModelEntityItem.  Seth suggested this fix and it worked for me.

libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -117,6 +117,7 @@ void PhysicalEntitySimulation::clearEntitiesInternal() {
     _pendingRemoves.clear();
     _pendingAdds.clear();
     _pendingChanges.clear();
+    _outgoingChanges.clear();
```
AndrewMeadows: **_outgoingChanges** was not cleared in **clearEntitiesInternal()** which is called when clearing out the entire simulation (when going to a new domain, for example).  Hence when the simulation picked up in the new domain this list had a dangling pointer in it waiting to be dereferenced.

hifi-gustavo:
Build console output for commit: ff83f8fc37081bfb115cb8878ff224235eed4563 available at: https://gist.github.com/babf0b312471fd2aaaed<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6675-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6675.dmg

hifi-gustavo:
Build console output for commit: 45e31ca4422359e6af92ba257fdb20bbaa1469b6 available at: https://gist.github.com/babf0b312471fd2aaaed<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/High+Fidelity-PR-6675-win64.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6675.dmg

PR #6676, opened 2015-12-16
Dejitter hold action

sethalves:
- bug fixes to make held objects not jitter around when walking


hyperlogic:
This change might effect your PR.  https://github.com/highfidelity/hifi/pull/6673

sethalves:
Doesn't seem to.  I think the action calls from the main thread so the cross-thread invoke doesn't happen.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -625,3 +625,11 @@ glm::vec3 RenderableModelEntityItem::getAbsoluteJointTranslationInObjectFrame(in
     }
     return glm::vec3(0.0f);
 }
+
+void RenderableModelEntityItem::locationChanged() {
```
ZappoMan: I'm guessing this is the money shot - eh? Is this the source of the jitter? not updating the model immediately when location changed?

hifi-gustavo:
Build console output for commit: 32fb4decd54f296f97240d75b8b2dc377482f27c available at: https://gist.github.com/73f3dc2bdf1eb8cc57d1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6676.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6676.dmg

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
sethalves: It's... some... of the money shot.  The rest is in the hold action where the entity's position is updated based on the avatar's rigid-body location rather than based on the data in Avatar.h.  Both fixes are required to get rid of all the jitter.

interface/src/avatar/AvatarActionHold.cpp
```diff
@@ -32,6 +33,56 @@ AvatarActionHold::~AvatarActionHold() {
...
+    auto holdingAvatar = std::static_pointer_cast<Avatar>(avatarManager->getAvatarBySessionID(_holderID));
+
+    if (!holdingAvatar) {
+        return;
+    }
+
+    withWriteLock([&]{
+        if (_ignoreIK && holdingAvatar->isMyAvatar()) {
+            return;
+        }
```
AndrewMeadows: Does not block the PR but... the above early exit doesn't require the lock.

interface/src/avatar/AvatarActionHold.cpp
```diff
@@ -32,6 +33,56 @@ AvatarActionHold::~AvatarActionHold() {
...
+            getAvatarRigidBodyLocation(avatarRigidBodyPosition, avatarRigidBodyRotation);
+
+            // determine the difference in translation and rotation between the avatar's
+            // rigid body and the palm position.  The avatar's rigid body will be moved by bullet
+            // between this call and the call to getTarget, below.  A call to get*PalmPosition in
+            // getTarget would get the palm position of the previous location of the avatar (because
+            // bullet has moved the av's rigid body but the rigid body's location has not yet been
+            // copied out into the Avatar class.
+            _palmOffsetFromRigidBody = palmPosition - avatarRigidBodyPosition;
+            _palmRotationFromRigidBody = glm::inverse(avatarRigidBodyRotation) * palmRotation;
```
AndrewMeadows: **_palmOffsetFromRigidBody** is a local offset, but its components are using the unit vectors of the world-frame (it is not rotated into RigidBody local frame).  This is OK for now because the **CharacterController** doesn't change the avatar's rotation during the simulation... yet.  If we ever allow the physics simulation to rotate the avatar (for example, if the avatar were standing on rotating platform) then the "attached to hand" behavior will break.

However, assuming the avatar's angular velocity from physics is small... the breakage will likewise be small.

interface/src/avatar/AvatarActionHold.cpp
sethalves: well... we would still need at least a read-lock before accessing _ignoreIK, right?

interface/src/avatar/AvatarActionHold.cpp
```diff
@@ -55,6 +106,23 @@ std::shared_ptr<Avatar> AvatarActionHold::getTarget(glm::quat& rotation, glm::ve
...
+            // and the data in the Avatar class is stale.  This means that the result of get*PalmPosition will
+            // be stale.  Instead, determine the current palm position with the current avatar's rigid body
+            // location and the saved offsets.
+            palmPosition = avatarRigidBodyPosition + _palmOffsetFromRigidBody;
+            palmRotation = avatarRigidBodyRotation * _palmRotationFromRigidBody;
+            if (isRightHand) {
+                palmRotation = holdingAvatar->getRightPalmRotation();
+            } else {
+                palmRotation = holdingAvatar->getLeftPalmRotation();
+            }
```
AndrewMeadows: The code correctly computes **palmRotation** at line 120, and then overwrites it with queries deep (potentially into another thread in the fucher) into the **holdingAvatar**.  I think lines 121 - 125 should be deleted.

hifi-gustavo:
Build console output for commit: f3adb8a2f70264af4eb50b93b314065ce88d059b available at: https://gist.github.com/73f3dc2bdf1eb8cc57d1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6676.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6676.dmg

hifi-gustavo:
Build console output for commit: b80fa1c8064c23bb3709593dafcef968cb7e1aa3 available at: https://gist.github.com/73f3dc2bdf1eb8cc57d1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6676.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6676.dmg

PR #6688, opened 2015-12-18
Added live custom debug shader support 

Atlante45:
Added support for multiple custom debug shaders the auto-update when the file is saved.

### DIFF
--- a/libraries/render-utils/src/DebugDeferredBuffer.cpp
+++ b/libraries/render-utils/src/DebugDeferredBuffer.cpp
@@ -11,8 +11,6 @@
 
 #include "DebugDeferredBuffer.h"
 
-#include <QString>
-
 #include <gpu/Batch.h>
 #include <gpu/Context.h>
 #include <render/Scene.h>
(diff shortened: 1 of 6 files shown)


libraries/render-utils/src/DebugDeferredBuffer.cpp
```diff
@@ -31,89 +31,160 @@ enum Slots {
     Normal,
     Specular,
     Depth,
-    Lighting,
-    
-    NUM_SLOTS
+    Lighting
+};
+
+static const std::string DEEFAULT_DIFFUSE_SHADER {
```
jherico: editing shaders like this is a pain.  I would suggesting using multiline character strings...

```

static const char * SIMPLE_TEXTURED_FS = R"FS(

uniform sampler2D sampler;
in vec2 vTexCoord;
out vec4 FragColor;

void main() {

    FragColor = texture(sampler, vTexCoord);
}

)FS";
```

hifi-gustavo:
Build console output for commit: 78b805a19a50a286a446f252dc3bd84db1f8e1ee available at: https://gist.github.com/53352be1cb5eb917b1de<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6688.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6688.dmg

PR #6696, opened 2015-12-18
angular velocity is a vector

AndrewMeadows:
AFAICT no one has tried to use the controller's angularVelocity data until this week, and problems have shown up immediately.  The problem is that the angularVelocity was being represented as a **quat** when it should have been a **vec3**(1).

This PR fixes it.

(1) What does it mean to store angularVelocity as a quat?  I believe the quatermnion was the rotation that would exist if the real angular velocity were applied for one unit of time (e.g. one second).  So, the max speed that could be represented without aliasing would be 2PI/second, or one rotation per second.  Higher speeds would alias to mod(2PI).  And if the scripter didn't know that the angularVelocity was stored as a quat then their math would do unexpected things.

More notes: after looking at the code again realized that the "unit of time" was not being employed.  Instead the stored "angularVelocity" was literally the "deltaRotation" of the last measurement.  This means that the true angularVelocity could not be extracted without knowing the deltaTime of the measurement.  This explains why the measured angularVelocity would be so noisy -- when the controller was undergoing uniform angular velocity the measured value would be dependent on frame rate.

### DIFF
--- a/plugins/hifiSixense/src/SixenseManager.cpp
+++ b/plugins/hifiSixense/src/SixenseManager.cpp
@@ -191,7 +191,7 @@ void SixenseManager::InputDevice::update(float deltaTime, bool jointsCaptured) {
                 //  Rotation of Palm
                 glm::quat rotation(data->rot_quat[3], data->rot_quat[0], data->rot_quat[1], data->rot_quat[2]);
                 handlePoseEvent(deltaTime, position, rotation, left);
-                rawPoses[i] = controller::Pose(position, rotation, glm::vec3(0), glm::quat());
+                rawPoses[i] = controller::Pose(position, rotation, Vectors::ZERO, Vectors::ZERO);
             } else {
                 _poseStateMap.clear();
                 _collectedSamples.clear();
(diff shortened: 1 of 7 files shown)


plugins/hifiSixense/src/SixenseManager.cpp
```diff
@@ -457,25 +457,21 @@ void SixenseManager::InputDevice::handlePoseEvent(float deltaTime, glm::vec3 pos
...
-        samples.first.addSample(velocity);
-        velocity = samples.first.average;
-     
-        // FIXME: // Not using quaternion average yet for angular velocity because it s probably wrong but keep the MovingAverage in place
-        //samples.second.addSample(glm::vec4(angularVelocity.x, angularVelocity.y, angularVelocity.z, angularVelocity.w));
-        //angularVelocity = glm::quat(samples.second.average.w, samples.second.average.x, samples.second.average.y, samples.second.average.z);
+        auto speed = glm::angle(deltaRot) / deltaTime;
+        angularVelocity = speed * axis;
+        samples.second.addSample(angularVelocity);
+        angularVelocity = samples.second.average;
```
AndrewMeadows: Since we now store the **angularVelocity** as a **vec3** it is possible to filter it.

hifi-gustavo:
Build console output for commit: d6541e9ee77ce1fdca7ea003b08f4030b3854b4e available at: https://gist.github.com/6e4df460ab9d4dcadd7c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6696.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6696.dmg

ZappoMan:
```
Undefined symbols for architecture x86_64:
  "model::Material::setDiffuse(glm::detail::tvec3<float, (glm::precision)0> const&, bool)", referenced from:
      OBJReader::readOBJ(QByteArray&, QHash<QString, QVariant> const&, QUrl const&) in libfbx.a(OBJReader.o)
  "model::Material::setEmissive(glm::detail::tvec3<float, (glm::precision)0> const&, bool)", referenced from:
      OBJReader::readOBJ(QByteArray&, QHash<QString, QVariant> const&, QUrl const&) in libfbx.a(OBJReader.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

AndrewMeadows:
That error looks unrelated to the changes in this PR.  I suspect a glitch in the automated build.  I'll trigger a new test build to see if it happens again.

AndrewMeadows:
test this please

hifi-gustavo:
Build console output for commit: d6541e9ee77ce1fdca7ea003b08f4030b3854b4e available at: https://gist.github.com/6e4df460ab9d4dcadd7c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6696.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6696.dmg

PR #6720, opened 2015-12-22
Fix particle radius and rgba start, middle, finish interpolation

ctrlaltdavid:
Bezier interpolation of 3 points doesn't intersect the middle value. Instead, use separate Bezier interpolations for start-middle and middle-finish, so that the interpolated values intersect the middle value as a user would expect. (Resurrects particle behaviour prior to particle rendering calcs being moved to the GPU.)


hifi-gustavo:
Build console output for commit: 58f1371ebac48e58eaa0f02b193cdbd93cb2c5fc available at: https://gist.github.com/f6632d42883e75e4e1b6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6720.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6720.dmg

hifi-gustavo:
Build console output for commit: 24f6cfa0183bbf5cbc185d820ff7f8eb858b7bed available at: https://gist.github.com/f6632d42883e75e4e1b6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6720.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6720.dmg

imgntn:
tested win10 with particles.js and particleExplorer -- looks good to me in testing

Atlante45:
The reason I'd removed that is because all those "if" are terrible regarding shaders performances.

We have 2 solution here:
- streamline the function and remove all those "if"
- replace it with something else. I'd started thinking about using cubic splines which would give us room for extension (i.e. n-point interpolations).
But this seem unwise as long as we don't have a compute shader step in place.

 For now, I think this would be sufficient:

```
float interpolate3Points(float y1, float y2, float y3, float u) {
    // Makes the interpolated values intersect the middle value.
    
    if ((u <= 0.5f && y1 == y2) || (u >= 0.5f && y2 == y3)) {
        // Flat line.
        return y2;
    }
    
    // L or inverted and/or mirrored L shape.
    // Make the slope at y2 be the slope between y1 and y3, up to a maximum of double the minimum of the slopes between y1
    // and y2, and y2 and y3. Use this slope to calculate the control points half way between the value points.
    // Note: The maximum ensures that the control points and therefore the interpolated values stay between y1 and y3.
    float slope = y3 - y1;
    float slope12 = 2.0f * (y2 - y1);
    float slope23 = 2.0f * (y3 - y2);
    
    if ((y2 >= y1 && y2 >= y3) || (y2 <= y1 && y2 <= y3)) {
        // U or inverted-U shape.
        // Make the slope at y2 = 0, which means that the control points half way between the value points have the value y2.
        slope = 0.0f;
    } else if (abs(slope) > abs(slope12)) {
        slope = slope12;
    } else if (abs(slope) > abs(slope23)) {
        slope = slope23;
    }
    
    float stepU = step(0.5f, u);
    float slopeSign = 2.0f * stepU - 1.0f; // -1.0 if u < 0.5, 1.0 otherwise
    float start = (1.0f - stepU) * y1 + stepU * y2; // y1 if u < 0.5, y2 otherwise
    float middle = y2 + slopeSign * slope / 2.0f;
    float end = (1.0f - stepU) * y2 + stepU * y3; // y2 if u < 0.5, y3 otherwise
    float v = 2.0f * u - stepU; // 0.0-0.5 -> 0.0-1.0 and 0.5-1.0 -> 0.0-1.0
    
    return bezierInterpolate(start, middle, end, v);
}
```

The first "if" is fine as it return right away.
The other one is not great, but I can't think of an easy way to remove it so it'll have to do for now.

ctrlaltdavid:
@Atlante45 In shaders, are chained ifs - if() {...} else if () {...} else if () { } - like in your suggestion better than nested ifs - if () { } else { if () { } else { } } - or doesn;'t it matter (assuming that there are the same number of "if" statements?

Atlante45:
Depends, what really matters is to have all the threads running execute the same instructions with different data.

Here the real issue is branching depending on "u" since it's the only value that should be different per thread. y1, y2, y3 are taken from the uniforms and should be the same for each particle (at least within one draw call).

I was actually wrong above, the second "if" will be fine here. and even if the first one depends on "u" it bypasses the function completely so that should be alright. @samcake would have a better idea than me about this.

The main thing I did above is remove the "if (u <= 0.5f) {} else {}". The reason I collapsed the englobing "if" is to not duplicate that block:
```
    float stepU = step(0.5f, u);
    float slopeSign = 2.0f * stepU - 1.0f; // -1.0 if u < 0.5, 1.0 otherwise
    float start = (1.0f - stepU) * y1 + stepU * y2; // y1 if u < 0.5, y2 otherwise
    float middle = y2 + slopeSign * slope / 2.0f;
    float end = (1.0f - stepU) * y2 + stepU * y3; // y2 if u < 0.5, y3 otherwise
    float v = 2.0f * u - stepU; // 0.0-0.5 -> 0.0-1.0 and 0.5-1.0 -> 0.0-1.0
``` 
Because it would have been very similar.

hifi-gustavo:
Build console output for commit: 16d4234d559d2a88d9a89e42dc8cf9d3c1eab19c available at: https://gist.github.com/f6632d42883e75e4e1b6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6720.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6720.dmg

libraries/entities-renderer/src/textured_particle.slv
```diff
@@ -56,13 +56,50 @@ float bezierInterpolate(float y1, float y2, float y3, float u) {
...
-                bezierInterpolate(y1.w, y2.w, y3.w, u));
+float interpolate3Points(float y1, float y2, float y3, float u) {
+    // Makes the interpolated values intersect the middle value.
+
+    if ((u <= 0.5f && y1 == y2) || (u >= 0.5f && y2 == y3)) {
+        // Flat line.
+        return y2;
+    }
+
+	float halfSlope;
```
Atlante45: Replace tab with 4 spaces

libraries/entities-renderer/src/textured_particle.slv
```diff
@@ -56,13 +56,50 @@ float bezierInterpolate(float y1, float y2, float y3, float u) {
...
+    if ((y2 >= y1 && y2 >= y3) || (y2 <= y1 && y2 <= y3)) {
+        // U or inverted-U shape.
+        // Make the slope at y2 = 0, which means that the control points half way between the value points have the value y2.
+		halfSlope = 0.0f;
```
Atlante45: Replace tab with 4 spaces

hifi-gustavo:
Build console output for commit: 4d7ed6b68f7506452f1beef3b93a6b9a5b2f402b available at: https://gist.github.com/f6632d42883e75e4e1b6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6720.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6720.dmg

PR #6733, opened 2015-12-23
Simplify renderDebug GUI

zzmp:
- Adds CollapsablePanelItem.prototype.editTitle.


zzmp:
examples/utilities/tools/cookies.js could probably do with some cleanup, considering the subpanel bug with multiple subpanels apparent with multiple sliders, when running examples/utilities/tools/renderEngineDebug.js.

PR #6737, opened 2015-12-24
Move final blit into RenderDeferredTask

zzmp:
- Also changes color buffer clearing, such that
anything run through Blit will now clear BUFFER_COLOR0  to 0,0,1,0.
- Blit will always be added to RenderDeferredTask, but
will only run if renderArgs._blitFramebuffer is set.

@samcake, is it OK to clear the color buffer in this way? Is it *necessary* to clear it at all, if we are doing a blit anyways? Can you explain what the existing calls to clearColorFramebuffer *were* doing?

The only time that Blit will not run is when an avatar billboard is rendered, as it is downloaded and not copied to the screen (AFAICT).

EDIT: Blit will no longer clear BUFFER_COLOR0, as that was redundant with blitting.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1207,23 +1207,12 @@ void Application::paintGL() {
         auto primaryFbo = DependencyManager::get<FramebufferCache>()->getPrimaryFramebuffer();
 
         renderArgs._renderMode = RenderArgs::MIRROR_RENDER_MODE;
+        renderArgs._blitFramebuffer = DependencyManager::get<FramebufferCache>()->getSelfieFramebuffer();
+
         renderRearViewMirror(&renderArgs, _mirrorViewRect);
+
+        renderArgs._blitFramebuffer.reset();
         renderArgs._renderMode = RenderArgs::DEFAULT_RENDER_MODE;
-
-        {
-            float ratio = ((float)QApplication::desktop()->windowHandle()->devicePixelRatio() * getRenderResolutionScale());
-            // Flip the src and destination rect horizontally to do the mirror
-            auto mirrorRect = glm::ivec4(0, 0, _mirrorViewRect.width() * ratio, _mirrorViewRect.height() * ratio);
-            auto mirrorRectDest = glm::ivec4(mirrorRect.z, mirrorRect.y, mirrorRect.x, mirrorRect.w);
-
-            auto selfieFbo = DependencyManager::get<FramebufferCache>()->getSelfieFramebuffer();
-            gpu::doInBatch(renderArgs._context, [=](gpu::Batch& batch) {
-                batch.setFramebuffer(selfieFbo);
-                batch.clearColorFramebuffer(gpu::Framebuffer::BUFFER_COLOR0, glm::vec4(0.0f, 0.0f, 0.0f, 0.0f));
-                batch.blit(primaryFbo, mirrorRect, selfieFbo, mirrorRectDest);
-                batch.setFramebuffer(nullptr);
-            });
-        }
     }
 
     {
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: 687eef37da50f6f79bf44df210b9f15c0b2142e9 available at: https://gist.github.com/c52b8098ae0cbc4d6829<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6737.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6737.dmg

libraries/render-utils/src/RenderDeferredTask.cpp
```diff
@@ -385,6 +387,77 @@ void DrawBackgroundDeferred::run(const SceneContextPointer& sceneContext, const
...
+    int width = renderArgs->_viewport.z;
+    int height = renderArgs->_viewport.w;
+
+    // Blit primary to blit FBO
+    auto framebufferCache = DependencyManager::get<FramebufferCache>();
+    auto primaryFbo = framebufferCache->getPrimaryFramebuffer();
+
+    gpu::doInBatch(renderArgs->_context, [=](gpu::Batch& batch) {
+        batch.setFramebuffer(blitFbo);
+        batch.clearColorFramebuffer(gpu::Framebuffer::BUFFER_COLOR0, glm::vec4(0.0f, 0.0f, 1.0f, 0.0f));
```
samcake: This one is not needed i believe because you are going to overwrite all the pixels through the blit

interface/src/Application.cpp
```diff
@@ -1207,23 +1207,12 @@ void Application::paintGL() {
         auto primaryFbo = DependencyManager::get<FramebufferCache>()->getPrimaryFramebuffer();
 
         renderArgs._renderMode = RenderArgs::MIRROR_RENDER_MODE;
+        renderArgs._blitFramebuffer = DependencyManager::get<FramebufferCache>()->getSelfieFramebuffer();
+
         renderRearViewMirror(&renderArgs, _mirrorViewRect);
```
samcake: I think there is another place where we use rearViewMirror to render an image of your avatar which is potentially used as a billboard when someone look at you from the far distance. it's line 3451 in Applicaiton.cpp
You need to check that the code is doing correctly the blitFramebuffer otherwise you might have surprises

interface/src/Application.cpp
```diff
@@ -1394,65 +1383,11 @@ void Application::paintGL() {
             renderArgs._context->setStereoProjections(eyeProjections);
             renderArgs._context->setStereoViews(eyeOffsets);
         }
+        renderArgs._blitFramebuffer = finalFramebuffer;
```
samcake: Slick!
I like

libraries/render-utils/src/RenderDeferredTask.cpp
```diff
@@ -385,6 +387,77 @@ void DrawBackgroundDeferred::run(const SceneContextPointer& sceneContext, const
...
 }
 
+void Blit::run(const SceneContextPointer& sceneContext, const RenderContextPointer& renderContext) {
+    assert(renderContext->getArgs());
+    assert(renderContext->getArgs()->_context);
+
+    RenderArgs* renderArgs = renderContext->getArgs();
+    auto blitFbo = renderArgs->_blitFramebuffer;
+
+    if (!blitFbo) {
```
samcake: Ok good, that should adrres the case i m ntalking about for the rearViewMirror/billboard

zzmp: I `.reset()` the framebuffer after each use, so this should cover the
billboard case.

hifi-gustavo:
Build console output for commit: dfa2b8d453c06790869fa4a85819f41241d0a3c2 available at: https://gist.github.com/c52b8098ae0cbc4d6829<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6737.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6737.dmg

PR #6741, opened 2015-12-28
Improved hand controller grab and mouse movement

PhilipRosedale:
The initial phase of using the hand controllers to grab is changed to create a sphere marker that starts at the center of the field of view and can then be moved either with the head or the hand controllers.  When the sphere is over an object, squeezing the trigger farther begins the distant grab. 

See how you like this change, in comparison to using the laser pointers out of the hand to select the object to be grabbed.  

This change also demonstrates how easily and accurately the head and hand controllers can position a 'cursor' over a detailed point in a scene, which can be the same behavior for selecting UI elements.  Squeezing the trigger a bit can cause the 'reticle' sphere to appear, and squeezing farther will click/drag.

Separately, for the reticle test script: 

Added velocity filtering of hand controller moved mouse/reticle to allow easier targeting of small features.  Also made switching back and forth between hand pointers smoother... the hand that is moving gets control of the pointer.  To take it with the other hand, just move it a bunch.

### DIFF
--- a/examples/controllers/handControllerGrab.js
+++ b/examples/controllers/handControllerGrab.js
@@ -23,8 +23,9 @@ var WANT_DEBUG = false;
 // these tune time-averaging and "on" value for analog trigger
 //
 
-var TRIGGER_SMOOTH_RATIO = 0.1; // 0.0 disables smoothing of trigger value
-var TRIGGER_ON_VALUE = 0.4;
+var TRIGGER_SMOOTH_RATIO = 0.1;     //  Time averaging of trigger - 0.0 disables smoothing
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 118f919c8b39c2cccbd628b667ba387ee1bc4b8a available at: https://gist.github.com/7821d615d77d89795322<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6741.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6741.dmg

PR #6749, opened 2015-12-30
Allow a script to set a resource override.

jherico:
This allows a script to set a resource override path, allow rapid iteration of content development from locally hosted content without having to change the entity properties of the object(s) being developed.  

For instance, I can load a script with the following code

```
var publicPath = "https://s3.amazonaws.com/DreamingContent/";
var basePath = Script.resolvePath("..");
print("Base path " + basePath);
Resources.overrideUrlPrefix(publicPath, "file:///" + basePath);
```

and restart my client.  Now, all content that would normally be fetched from my S3 bucket will instead be loaded from my local disk, relative the parent folder of the script. This allows me to modify shaders and model files and rapidly see the results before pushing the finished content back to S3 for public consumption.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -47,6 +47,7 @@
 #include <gl/Config.h>
 #include <gl/QOpenGLContextWrapper.h>
 
+#include <ResourceScriptingInterface.h>
 #include <AccountManager.h>
 #include <AddressManager.h>
 #include <ApplicationVersion.h>
@@ -337,6 +338,8 @@ bool setupEssentials(int& argc, char** argv) {
     DependencyManager::set<RecordingScriptingInterface>();
     DependencyManager::set<WindowScriptingInterface>();
     DependencyManager::set<HMDScriptingInterface>();
+    DependencyManager::set<ResourceScriptingInterface>();
+    
 
 #if defined(Q_OS_MAC) || defined(Q_OS_WIN)
     DependencyManager::set<SpeechRecognizer>();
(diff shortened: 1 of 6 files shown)


hifi-gustavo:
Build console output for commit: 07531353e4514c21d170ed618c32758a3c1ee099 available at: https://gist.github.com/c4fe3cf1584ce28e4226<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6749.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6749.dmg

interface/src/Application.cpp
```diff
@@ -337,6 +338,8 @@ bool setupEssentials(int& argc, char** argv) {
     DependencyManager::set<RecordingScriptingInterface>();
     DependencyManager::set<WindowScriptingInterface>();
     DependencyManager::set<HMDScriptingInterface>();
+    DependencyManager::set<ResourceScriptingInterface>();
```
ZappoMan: You probably need to add something similar to the Agent/Assignment client. See `Agent::Agent()` or `AssignmentClient::AssignmentClient()`

jherico: The resource scripting interface is designed to be used by a person while developing content.  It's hard to imagine the use case where it would be valuable to an agent... but OK.

jherico: oh, never mind, I get it... it's referenced in script engine, so it needs to be available to both assignment clients as well as interface.

hifi-gustavo:
Build console output for commit: 3b4f5d36a5cd96ab8d2544e9c15e3f1d23ea34b2 available at: https://gist.github.com/c4fe3cf1584ce28e4226<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6749.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6749.dmg

PR #6755, opened 2015-12-31
Optimize getVertexStream() by using a const&

zzmp:
Tony mentioned that this was a large vector copy when he was looking at performance.

### DIFF
--- a/libraries/model/src/model/Geometry.h
+++ b/libraries/model/src/model/Geometry.h
@@ -58,7 +58,7 @@ public:
     const gpu::Stream::FormatPointer getVertexFormat() const { return _vertexFormat; }
 
     // BufferStream on the mesh vertices and attributes matching the vertex format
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: c6e6aceb8527fb35753c0d603db069916dfb71df available at: https://gist.github.com/722b2ff017b5bb99d228<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6755.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6755.dmg

PhilipRosedale:
cool - where would we expect to see performance gains?   All model renders?

jherico:
@PhilipRosedale this won't have a noticable effect on rendering.  While this is a good change, we're not currently CPU bound on rendering.

AndrewMeadows:
I approve this product or service.

PR #6760, opened 2016-01-04
Fix to pistol.js so muzzle flash particle system doesn't stick around…

ericrius1:
… forever


ericrius1:
Test script: https://rawgit.com/ericrius1/hifi/muzzleFix/unpublishedScripts/immediateClientReset.js

PR #6761, opened 2016-01-04
Add Beam Visualization When Grab Doesn't Intersect

imgntn:
This PR makes it so that your search beams are visible while searching, even if the beam isn't intersecting with anything.  Previously if you tried to grab something in the air and missed, you couldn't tell how you missed and so actually grabbing the object became very difficult.  Now you see beams when you're searching.  

To test: https://rawgit.com/imgntn/hifi/grab_add_beam/examples/controllers/handControllerGrab.js


imgntn:
@ericrius1 can u test when u get a chance, plz.  thanks!

PR #6769, opened 2016-01-05
Interface support for avatar attachments from Marketplace

hyperlogic:
Drag and drop support of ava.json files is also supported and at the moment the only way to test.

To test:

1. Create a cowboyhat.ava.json file on your hard-drive, with the following contents:
<pre>
    {
        "version": "0.1",
        "name": "Cowboy Hat",
        "modelUrl": "https://s3.amazonaws.com/hifi-public/tony/cowboy-hat.fbx",
        "jointName": "Head",
        "transform": {
            "rotation": {"x": 0, "y": 0, "z": 0, "w": 1},
            "translation": {"x": 0, "y": 0, "z": 0},
            "scale": {"x": 1, "y": 1, "z": 1}
        },
        "isSoft": false
    }
</pre>
2. Create a leatherjacket.ava.json file, with the following contents:
<pre>
    {
        "version": "0.1",
        "name": "Leather Jacket",
        "modelUrl": "https://hifi-content.s3.amazonaws.com/ozan/dev/clothes/coat/coat_rig.fbx",
        "isSoft": true
    }
</pre>
3. Run interface
4. Drag-and-Drop cowboyhat.ava.json onto the main Interface window.
5. You should be prompted wear the 'Cowboy Hat' attachment.
6. If you select 'Yes', the cowboy hat should appear on the avatar.
7. Do the same with leatherjacket.ava.json file.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -20,6 +20,8 @@
 #include <glm/gtc/type_ptr.hpp>
 
 #include <QtCore/QDebug>
+#include <QtCore/QJsonDocument>
+#include <QtCore/QJsonObject>
 #include <QtCore/QObject>
 #include <QtCore/QUrl>
 #include <QtCore/QTimer>
@@ -184,6 +186,7 @@ static const QString JS_EXTENSION  = ".js";
 static const QString FST_EXTENSION  = ".fst";
 static const QString FBX_EXTENSION  = ".fbx";
 static const QString OBJ_EXTENSION  = ".obj";
+static const QString AVA_JSON_EXTENSION = ".ava.json";
 
 static const int MIRROR_VIEW_TOP_PADDING = 5;
 static const int MIRROR_VIEW_LEFT_PADDING = 10;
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: bac6570655037ea52577f0f51a6a3c92080f1751 available at: https://gist.github.com/d8a7c397ebead135fdab<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6769.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6769.dmg

howard-stearns:
This works as an illustration of what's possible (so merging), but there are some deep issues that need to be addressed before this is meaningful for non-alpha use:

1) Load failures should happen for the user that initiates the action. Right now, if the modelUrl is not specified, there's no message at all. If there's bad .fbx data in the model, the error will occur on everybody's machine.

2) As we move into attributed content, we will probably want a generic means of identifying metadata for any json serialization (author, author reputation, item reputation, etc.). There's stuff to be worked out about what that means for what properties must be specified in all .json (e.g, perhaps an id and a type), which isn't being checked for here.

3) The hat and jacket didn't fit me. This is just sample data, but it shows that the protocol will need other info such as avatar "size".

PR #6774, opened 2016-01-05
Fix Wall Targets Collision Callback

imgntn:
This PR makes it so that the targets on the wall in toybox only call their collision callback one time per viewer. Previously, the collision callback could be called repeatedly.

### DIFF
--- a/examples/toybox/ping_pong_gun/wallTarget.js
+++ b/examples/toybox/ping_pong_gun/wallTarget.js
@@ -17,6 +17,7 @@
     }
 
     Target.prototype = {
+        hasBecomePhysical: false,
         hasPlayedSound: false,
         preload: function(entityID) {
             this.entityID = entityID;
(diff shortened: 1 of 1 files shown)


examples/toybox/ping_pong_gun/wallTarget.js
```diff
@@ -24,19 +25,22 @@
...
-                    z: 0
-                },
-                velocity: {
-                    x: 0,
-                    y: -0.01,
-                    z: 0
-                }
-            });
+            if (this.hasBecomePhysical === false) {
+                var position = Entities.getEntityProperties(me, "position").position;
```
ZappoMan: do you actually need this?

ZappoMan: maybe you want it in the audio section below around line 42/46

PR #6782, opened 2016-01-06
reduced avatar max fly speed

AndrewMeadows:
The max fly speed used to be 300m/sec, I reduced it to 30m/sec (~67mph).  For comparison, as I recall the max unassisted fly speed in SL was about 50m/sec.

I considered making it a preference so that users could play with it but:

(1) I think maybe someone is currently overhauling the preferences window (?), so I figured we could wait until that is done.

(2) Its hard to tune it with our current animation and position glitches  -- it doesn't look good at any high speed.

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -64,7 +64,7 @@ const float MIN_AVATAR_SPEED = 0.05f; // speed is set to zero below this
 
 // TODO: normalize avatar speed for standard avatar size, then scale all motion logic
 // to properly follow avatar size.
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: b8b113844f33b2b162395687b2c576e0abac0e21 available at: https://gist.github.com/3b803aa8af87d5ef7670<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6782.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6782.dmg

PR #6793, opened 2016-01-08
fix queryAABox bug

sethalves:
- fix a bug that caused queryAABox of some entities to be a unit cube instead of the correct size

### DIFF
--- a/libraries/shared/src/SpatiallyNestable.cpp
+++ b/libraries/shared/src/SpatiallyNestable.cpp
@@ -598,7 +598,8 @@ AACube SpatiallyNestable::getQueryAACube(bool& success) const {
         return _queryAACube;
     }
     success = false;
-    return AACube(getPosition(success) - glm::vec3(defaultAACubeSize / 2.0f), defaultAACubeSize);
+    bool getPositionSuccess;
+    return AACube(getPosition(getPositionSuccess) - glm::vec3(defaultAACubeSize / 2.0f), defaultAACubeSize);
 }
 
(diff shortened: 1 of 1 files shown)


libraries/shared/src/SpatiallyNestable.cpp
```diff
@@ -598,7 +598,8 @@ AACube SpatiallyNestable::getQueryAACube(bool& success) const {
         return _queryAACube;
     }
     success = false;
-    return AACube(getPosition(success) - glm::vec3(defaultAACubeSize / 2.0f), defaultAACubeSize);
+    bool getPositionSuccess;
+    return AACube(getPosition(getPositionSuccess) - glm::vec3(defaultAACubeSize / 2.0f), defaultAACubeSize);
```
ZappoMan: what happens if getPosition() fails?

sethalves: Nothing good.  Maybe I shouldn't bother with this line, but I still need to return *something*.

hifi-gustavo:
Build console output for commit: 31bed934fe82884bdf4b8adeec84fcdcfe72a2dc available at: https://gist.github.com/ed9bb4755f95c46f74ca<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6793.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6793.dmg

PR #6795, opened 2016-01-08
Fix not being able to change model textures - part 1

ctrlaltdavid:
This fixes the "textures" property of model entities not successfully being interpreted. They are now once again loaded through to NetworkGeometry::setTextureWithNameToURL(), as evidenced by program log messages saying that the new texture has been requested and loaded, e.g., "Updating texture named "diffuse" to texture at URL QUrl("http://...")"

### DIFF
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -129,7 +129,11 @@ void RenderableModelEntityItem::remapTextures() {
     QJsonObject currentTexturesAsJsonObject = currentTexturesAsJson.object();
     QVariantMap currentTextureMap = currentTexturesAsJsonObject.toVariantMap();
 
-    QJsonDocument texturesAsJson = QJsonDocument::fromJson(_textures.toUtf8());
+    QJsonParseError error;
+    QJsonDocument texturesAsJson = QJsonDocument::fromJson(_textures.toUtf8(), &error);
+    if (error.error != QJsonParseError::NoError) {
+        qCWarning(entitiesrenderer) << "Could not evaluate textures property:" << _textures;
+    }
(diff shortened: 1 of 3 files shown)


ctrlaltdavid:
https://worklist.net/20763

hifi-gustavo:
Build console output for commit: 6eed6a0bc31de45a28d7cd9e63f9f8913c524efd available at: https://gist.github.com/fe85582e25481421bf91<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6795.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6795.dmg

imgntn:
@ctrlaltdavid sweet -- can you maybe point me at where to 'enable' this so we can fix it?
```changing a named texture on one model will change it for all (this is intentionally disabled in the code at present).```
@AlessandroSigna

ctrlaltdavid:
@imgntn That I unfortunately do not know; I've been told that changing the texture for one model used to change it for all. I may have assumed that it was intentionally disabled, i.e., it may have been unintentionally disabled.

ZappoMan:
I tested this @ctrlaltdavid - and I can't seem to get it to change the textures of any models, can you give me an example of what to set the texture value to to get it to work?

ctrlaltdavid:
@ZappoMan The new textures get loaded and set in the geometry cache, per the program log entries. Field format is identical to that displayed as the "original textures" value. ... However, the new value currently does not affect the model(s) displayed. This may have been intentionally disabled, or it may just have gotten broken sometime; I haven't managed to find either case in the code. @ericrius1 / @imgntn I think were going to look at this as part of making each texture override specified to the particular model (i.e., part 2).

PR #6796, opened 2016-01-08
entity collision masks can be set via script

AndrewMeadows:
We were already using the **collision groups** feature in Bullet but we had not really exposed any "collision filtering" control to JS.  This PR provides some initial basic capabilities.

The C++ code will automatically sort objects (entities and avatars) into the right "collision group".  Each group has a numerical bit value.  The current groups being used are:

bit0 : static = entities that have "collisionsWillMove" = true
bit1 : dynamic = entities that have "collisionsWillMove" = false AND are not moving (zero velocities)
bit2 : kinematic = entities that have "collisionsWillMove" = false BUT happen to be moving
bit3 : my avatar
bit4: other avatars
...
bit15: collisionless = entities that have "ignoreForCollisions" = true

Note: there is no way for JS to specify the collision group used for an entity, except to modify the properties that the C++ uses to sort them.  However, JS can now specify the **collision mask** of an entity.  The collision mask is a single byte of bit-mask that determines which groups the entity collides with.  The default value of the collision mask is as follows:

dynamic objects (entities and avatars) = 0x1f
static entities = 0x1e (static does not collide with static, a simulation optimization)
collisionless entities = 0x00

During a Bullet collision event the objects will collide if (groupA & maskB) && (groupB & maskA) is true.  That is, when both pairs of group/mask comparisons have at least one bit of overlap.

Hence to make a particular entity NOT collide with any avatars we would set its collision mask to 0x03.

Note: although Bullet provides 15 distinct collision groups (two bytes of bitmask) we only expose 8 groups (1 byte of bitmask) and only use 5 of them.


examples/controllers/handControllerGrab.js
```diff
@@ -1927,4 +1929,4 @@ function renewParticleBeamLifetimes(deltaTime) {
     }
     rightController.renewParticleBeamLifetime();
     leftController.renewParticleBeamLifetime();
-}
\ No newline at end of file
+}
```
AndrewMeadows: I've modified handControllerGrab.js to change the **collisionMask** in order to prevent self collision with avatars and also to prevent the grabbed object from colliding with static objects.  They still collide with dynamic objects and other avatars.

libraries/networking/src/udt/PacketHeaders.cpp
```diff
@@ -41,7 +41,7 @@ PacketVersion versionForPacketType(PacketType packetType) {
         case PacketType::EntityAdd:
         case PacketType::EntityEdit:
         case PacketType::EntityData:
-            return VERSION_ENTITITES_HAVE_QUERY_BOX;
+            return VERSION_ENTITITES_HAVE_COLLISION_MASK;
```
AndrewMeadows: Note: this PR changes the packet version.

libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -1653,6 +1660,9 @@ QList<QString> EntityItemProperties::listChangedProperties() {
     if (ignoreForCollisionsChanged()) {
         out += "ignoreForCollisions";
     }
+    if (collisionMaskChanged()) {
+        out += "collisionMask";
+    }
```
AndrewMeadows: There are now two ways to make an entity collisionless:

(1) set **ignoreForCollisions** true

(2) set the **collisionMask** to 0x00

We should probably remove the **ignoreForCollisions** property altogether.

libraries/physics/src/EntityMotionState.cpp
```diff
@@ -523,6 +523,17 @@ uint32_t EntityMotionState::getIncomingDirtyFlags() {
...
+        if (dirtyFlags | Simulation::DIRTY_SIMULATOR_ID) {
+            // when SIMULATOR_ID changes we must check for reinterpretation of asymmetric collision mask
+            // bits for the avatar groups (e.g. MY_AVATAR vs OTHER_AVATAR)
+            uint8_t entityCollisionMask = _entity->getCollisionMask();
+            if ((bool)(entityCollisionMask & USER_COLLISION_GROUP_MY_AVATAR) !=
+                    (bool)(entityCollisionMask & USER_COLLISION_GROUP_OTHER_AVATAR)) {
+                // bits are asymmetric --> flag for reinsertion in physics simulation
+                dirtyFlags |= Simulation::DIRTY_COLLISION_GROUP;
+            }
+        }
```
AndrewMeadows: There is complication with using the **otherAvatar** and **myAvatar** collision groups, since they mean different things depending on whether the simulation is on the local or remote interface.  We solve this problem by swapping the interpretation of the mask bits when we DON'T own the simulation.  So we must flag the object for collision group update whenever its ownership changes AND the two avatar mask bits are not the same.

libraries/physics/src/EntityMotionState.cpp
```diff
@@ -610,19 +621,40 @@ QString EntityMotionState::getName() const {
...
+        if ((bool)(entityCollisionMask & USER_COLLISION_GROUP_MY_AVATAR) !=
+                (bool)(entityCollisionMask & USER_COLLISION_GROUP_OTHER_AVATAR)) {
+            // asymmetric avatar collision mask bits
+            if (!_entity->getSimulatorID().isNull() && _entity->getSimulatorID() != PhysicsEngine::getSessionID()) {
+                // someone else owns the simulation, so we swap the interpretation of the bits
+                entityCollisionMask ^= USER_COLLISION_MASK_AVATARS | ~entityCollisionMask;
+            }
+        }
+        mask &= (int16_t)(entityCollisionMask);
+    }
```
AndrewMeadows: This is where we actually swap the meaning of the two avatar mask bits.

hifi-gustavo:
Build console output for commit: adabf9166181f3413de5cb508c955413b5048a63 available at: https://gist.github.com/9228c3779e826dc98ce4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6796.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6796.dmg

libraries/physics/src/EntityMotionState.h
```diff
@@ -28,8 +28,8 @@ class EntityMotionState : public ObjectMotionState {
     EntityMotionState(btCollisionShape* shape, EntityItemPointer item);
     virtual ~EntityMotionState();
 
-    void updateServerPhysicsVariables(const QUuid& sessionID);
-    virtual bool handleEasyChanges(uint32_t& flags, PhysicsEngine* engine);
+    void updateServerPhysicsVariables();
+    virtual bool handleEasyChanges(uint32_t& flags);
     virtual bool handleHardAndEasyChanges(uint32_t& flags, PhysicsEngine* engine);
```
AndrewMeadows: I took the opportunity to do some cleanup and removed the need for passing a **PhysiceEngine** pointer into the **ObjectMotionState::handleEasyChanges()** interface.

libraries/entities/src/EntityItem.cpp
```diff
@@ -678,6 +681,9 @@ int EntityItem::readEntityDataFromBuffer(const unsigned char* data, int bytesLef
     READ_ENTITY_PROPERTY(PROP_ANGULAR_DAMPING, float, updateAngularDamping);
     READ_ENTITY_PROPERTY(PROP_VISIBLE, bool, setVisible);
     READ_ENTITY_PROPERTY(PROP_IGNORE_FOR_COLLISIONS, bool, updateIgnoreForCollisions);
+    if (args.bitstreamVersion >= VERSION_ENTITITES_HAVE_COLLISION_MASK) {
```
ZappoMan: since you changed the packet version, you don't need to do this version check... this was only for supporting binary files, and we no longer have those in the system and we don't support them... so you can just have this READ_ENTITY_PROPERTY() macro here without the version check wrapper.

ZappoMan:
wouldn't it be preferable to have a string to byte mapping so that JS developers didn't need to know the bits/numbers? This could be accomplished in EntityItemProperties qscriptvalue conversion

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -22,20 +22,43 @@ uint32_t PhysicsEngine::getNumSubsteps() {
...
+
+// static
+void PhysicsEngine::setSessionUUID(const QUuid& sessionID) {
+    _sessionID = sessionID;
+}
+
+// static
+const QUuid& PhysicsEngine::getSessionID() {
+    return _sessionID;
+}
```
AndrewMeadows: This code becomes **static** in the **PhysicsEngine** namespace.

Note: we're currently using a dynamic map between collision groups and their default collision masks.  When I first wrote this code I thought maybe we would dynamically modify this map for special user-defined groups and masks, however now I'm thinking that we won't supply such a feature, so I'll probably replace this map with a **switch** statement soon.

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -139,8 +162,9 @@ void PhysicsEngine::addObjectToDynamicsWorld(ObjectMotionState* motionState) {
     body->setFlags(BT_DISABLE_WORLD_GRAVITY);
     motionState->updateBodyMaterialProperties();
 
-    int16_t group = motionState->computeCollisionGroup();
-    _dynamicsWorld->addRigidBody(body, group, getCollisionMask(group));
+    int16_t group, mask;
+    motionState->computeCollisionGroupAndMask(group, mask);
+    _dynamicsWorld->addRigidBody(body, group, mask);
```
AndrewMeadows: When we add the **RigidBody** to the simulation we get the collision **group** and **mask** from its **MotionState**.

examples/controllers/handControllerGrab.js
```diff
@@ -156,6 +155,12 @@ var STATE_CONTINUE_EQUIP = 14;
 var STATE_WAITING_FOR_BUMPER_RELEASE = 15;
 var STATE_EQUIP_SPRING = 16;
 
+var COLLISION_GROUP_STATIC = 0x01; // bit 0
+var COLLISION_GROUP_DYNAMIC = 0x02; // bit 1
+var COLLISION_GROUP_KINEMATIC = 0x04; // bit 2
+var COLLISION_GROUP_MY_AVATAR = 0x08; // bit 3
+var COLLISION_GROUP_OTHER_AVATAR = 0x10; // bit 4
+var COLLISION_MASK_WHILE_GRABBED = COLLISION_GROUP_DYNAMIC | COLLISION_GROUP_OTHER_AVATAR;
```
AndrewMeadows: James was also asking for some "string" representation for collision group bits.  I didn't quite understand what he was asking for.  Perhaps someone could provide some example API calls here for me to help explain how they would like it to work?

I could understand having these COLLISION_GROUP constants automatically exposed to JS, but how to specify the desired mask as strings?

imgntn: mapping these globals on the c++ side to strings i could pass in as entity properties would be nice --  i'd want to create an entity like...

var properties = {
collisionMask:['static','myAvatar']
}

or 

var properties = {
collisionMask:['grabbed']
}

ZappoMan: sure - give me a bit and I'll post an example.

hifi-gustavo:
Build console output for commit: d3b1a5ea1f6402550eb319931775d64097ab0296 available at: https://gist.github.com/9228c3779e826dc98ce4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6796.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6796.dmg

examples/controllers/handControllerGrab.js
ZappoMan: actually, here's an example: "shape type" -- basically in EntityItemProperties.h/cpp there's the notion of the "ShapeType" property... it uses an enum lookup and allows the JS to use a string "box" and it looks up the enum value... all the rest of the entity system uses the enums and not strings, things sent over the wire are bits/enums not strings, etc... but in "property" and json form it's a string.

The `DEFINE_PROPERTY_REF_ENUM()` macro will handle most of this for you...
```
DEFINE_PROPERTY_REF_ENUM(PROP_SHAPE_TYPE, ShapeType, shapeType, ShapeType, SHAPE_TYPE_NONE);
```

But it assumes you've implemented a `getXxxAsString()` and `setXxxFromString()` method... technically it doesn't need to be an enum, you could still use your bitmask concept... but in your set method you'll want to explode the string on commas or "|" or some delimiter you choose to support.

```
QString EntityItemProperties::getShapeTypeAsString() const {
    if (_shapeType < sizeof(shapeTypeNames) / sizeof(char *))
        return QString(shapeTypeNames[_shapeType]);
    return QString(shapeTypeNames[SHAPE_TYPE_NONE]);
}

void EntityItemProperties::setShapeTypeFromString(const QString& shapeName) {
    if (stringToShapeTypeLookup.empty()) {
        buildStringToShapeTypeLookup();
    }
    auto shapeTypeItr = stringToShapeTypeLookup.find(shapeName.toLower());
    if (shapeTypeItr != stringToShapeTypeLookup.end()) {
        _shapeType = shapeTypeItr.value();
        _shapeTypeChanged = true;
    }
}
```

Does that make sense?

AndrewMeadows: A readable string API to collisionMask would be cool.  Would it obviate this numerical one or would they be compatible?  That is, could it be submitted as a follow up PR, or should it block this one?

AndrewMeadows:
Don't merge yet.  I'll try to modify the JS API to use string names for groups rather than numerical values.

AndrewMeadows:
Ok this is ready for review again.

The way to set the collision mask for an entity now goes something like this:

    Entities.editEntity(id, { "collisionMask" : "groupA,groupB,groupC" });

Where the available groups are:  static, dynamic, kinematic, myAvatar, otherAvatar

examples/controllers/handControllerGrab.js
```diff
@@ -1738,18 +1739,16 @@ function MyController(hand) {
...
+            data["collisionMask"] = grabbedProperties.collisionMask;
             data["collisionsWillMove"] = grabbedProperties.collisionsWillMove;
             var whileHeldProperties = {
                 gravity: {
                     x: 0,
                     y: 0,
                     z: 0
-                }
+                },
+                "collisionMask": COLLISION_MASK_WHILE_GRABBED
```
AndrewMeadows: Previously I was doing some bitwise operations here to preserve any groups that happened to be cleared before the object was grabbed in the first place.  That is, it looked something like:

    "collisionMask" : COLLISION_MASK_WHILE_GRABBED | grabbedProperties.collisionMask

However, with stringy collisionMask arguments that is no longer easy to do so I go ahead and blast over any such cleared bits here while grabbed.

This problem is significant enough that I'll probably make it possible to set the collisionMask by numerical value anyway (by providing supplying a number rather than a listy string) to help preserve the sanity of those of us who think that way.

hifi-gustavo:
Build console output for commit: 752e8ad147de340096b433b6cf1598bf22979ea2 available at: https://gist.github.com/9228c3779e826dc98ce4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6796.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6796.dmg

ZappoMan:
test this please

AndrewMeadows:
test this please

hifi-gustavo:
Build console output for commit: 752e8ad147de340096b433b6cf1598bf22979ea2 available at: https://gist.github.com/9228c3779e826dc98ce4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6796.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6796.dmg

sethalves:
I'm not able to stop flying, with this PR.

hifi-gustavo:
Build console output for commit: 1e9cf0c760a59d7bdf8deee8b1a3937303b83a38 available at: https://gist.github.com/9228c3779e826dc98ce4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6796.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6796.dmg

sethalves:
distance grabbing no longer works.

hifi-gustavo:
Build console output for commit: 037164d936a8dda6ad187fe9ec476b93f5f28b0c available at: https://gist.github.com/9228c3779e826dc98ce4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6796.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6796.dmg

AndrewMeadows:
The script API for the **collisionMask**is really bothering me.  I'm going to change it.

hifi-gustavo:
Build console output for commit: 6a5a74700fa508d03cb8dd9acb1396e1c2dbbb9e available at: https://gist.github.com/9228c3779e826dc98ce4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6796.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6796.dmg

AndrewMeadows:
Belay that "Do Not Merge".  Someone wants this for tomorrow's demo. I will try to fix it before the end of this week before any intrepid JS developers try to use the feature.

libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -116,6 +116,59 @@ void buildStringToShapeTypeLookup() {
     addShapeType(SHAPE_TYPE_CYLINDER_Z);
 }
 
+QString getCollisionGroupAsString(uint8_t group) {
```
ZappoMan: to make this work properly as a "mask" you need to check each bit, and make a string which is a comma separated list of bit states.

libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -116,6 +116,59 @@ void buildStringToShapeTypeLookup() {
...
+            return "kinematic";
+        case USER_COLLISION_GROUP_MY_AVATAR:
+            return "myAvatar";
+        case USER_COLLISION_GROUP_OTHER_AVATAR:
+            return "otherAvatar";
+    };
+    return "";
+}
+
+uint8_t getCollisionGroupAsBitMask(const QStringRef& name) {
```
ZappoMan: to make this work properly as a "mask" you need to explode the QString into a QStringList and then check each named mask type and add the bit to the bit mask accordingly.

ZappoMan: never mind I see you did that

libraries/entities/src/EntityItemProperties.cpp
ZappoMan: never mind I see you did that

PR #6798, opened 2016-01-09
separating out marketplace icon from edit toolbar and making directory icon from directory.js movable and persistent

ericrius1:


ericrius1:
test urls:
https://rawgit.com/ericrius1/hifi/menuTweaks/examples/directory.js
https://rawgit.com/ericrius1/hifi/menuTweaks/examples/marketplace.js

RyanDowne:
Works great. I am assuming that the new edit.js doesn't have the Marketplace in it too. I think we might want to make the drag selection area bigger. I tried to target it with the hand controllers driving the cursor and it was pretty hard.

jherico:
If you're going to do this don't you need to add the new marketplace script to defaultScripts.js?

PR #6801, opened 2016-01-09
Fix AssignmentParentFinder not handling null parentID

huffman:


assignment-client/src/entities/AssignmentParentFinder.cpp
```diff
@@ -13,6 +13,12 @@
 
 SpatiallyNestableWeakPointer AssignmentParentFinder::find(QUuid parentID, bool& success) const {
     SpatiallyNestableWeakPointer parent;
+
+    if (parentID.isNull()) {
+        success = true;
```
ZappoMan: Shouldn't it success be false in this case?

sethalves: He has it right.  If the parentID is null, returning an empty parent pointer is correct and the lookup is successful.

huffman: I didn't make this decision - this is consistent with `InterfaceParentFinder` (https://github.com/highfidelity/hifi/blob/master/interface/src/InterfaceParentFinder.cpp)

I think we should be checking whether parentID is null earlier on, but I think that's a separate change.

sethalves: You're not wrong though -- we probably don't need to be calling this code if parentID is null.

sethalves:
test this please

hifi-gustavo:
Build console output for commit: 46ed5f6479c31937d5f0a62ae763fe18b255aedd available at: https://gist.github.com/7090c83c873b2057b791<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6801.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6801.dmg

PR #6802, opened 2016-01-10
Fix Agents bidding on simulation ownership

huffman:


hifi-gustavo:
Build console output for commit: dda2462315583e6413560d1b450f260dc559c15e available at: https://gist.github.com/e225db270b77b18b1a72<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6802.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6802.dmg

ZappoMan:
Cool, what's the big/repro case on this? Is there a bug report we can test against?

huffman:
To answer Brad's question: no bug report, just something I had come up while working on a project for my domain over the weekend. I was having issues with my local builds so I wanted to get a build from Jenkins. My goal was to make sure Agent's don't own simulations because they don't have a physics engine with which to simulate. The issue I was seeing: when an agent would spawn an entity with physical properties, it would look correct until moving the camera around, in which case the entities would snap back to their original position. I assume it was the case that the simulation was owned by the agent, and when the camera was moved the view frustum was updated on the entity server and all of the entities were resent with their original positions.

@AndrewMeadows 
> The "usefulness" of a scripted agent is limited to scripted actions that are not governed by physics.
I think we can get a little utility out of a scripted agent and physical attributes when spawning physical entities, which is what I was intending to fix. This PR doesn't fully fix that issue (and hasn't been tested yet).

It didn't occur to me that a scripted agent would't be able to affect physical attributes at all, but it makes sense.

I have a question: What happens if you want to apply an impulse to an entity that you are not simulating? Do you have to win simulation ownership with that entity edit? If not, no impulse will be applied? A big use-case for me is being able to apply an impulse from an agent (ex: grenade). Some of this could be done from the client, but for my purposes that would feel like a hack.

ZappoMan:
@huffman - That's an interesting repro case... Is your agent also making "additional edits" to the entity over time? Like changing it's color? If the agent adds and entity with velocity for example, but never "edits it" after that... then it wouldn't be the owner, and the entity should just fly along for all to see.

huffman:
@ZappoMan Interesting, I can make an example AC script for us to try. One other data point: when I made an edit to one of those spawned objects from interface, it no longer jumped back to it's original position.

huffman:
I didn't answer your question - No, the agent shouldn't have been making additional edits after the initial call to `addEntity`. (I mentioned that this PR doesn't fix the issue yet, and that's partially because it's only handling the `editEntity` case)

PhilipRosedale:
where are we on this PR?

huffman:
There isn't an associated bug report so it hasn't been a priority. If we want to add one and make it a priority I would be happy to work on it.

ZappoMan:
test this please

hifi-gustavo:
Build console output for commit: dda2462315583e6413560d1b450f260dc559c15e available at: https://gist.github.com/e225db270b77b18b1a72<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6802.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6802.dmg

AndrewMeadows:
I removed the **DO NOT MERGE** tag.  The work in this PR is progress: we can't have AC's trying to own simulation since they can't do proper simulation (yet).

huffman:
This isn't complete, removing QA Ready label

huffman:
IIRC, this only handles the edit case and not the creation case, I will talk to @AndrewMeadows

AndrewMeadows:
Yes, go ahead and check **_bidOnSimulationOwnership** for the **addEntity()** case: AC's should not bother to try to own the entities that they create with velocity.

hifi-gustavo:
Build console output for commit: eff02d3e3ac4bb480ca6beb1283b44cd5d8d2bd2 available at: https://gist.github.com/e225db270b77b18b1a72<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6802.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6802.dmg

hifi-gustavo:
Build console output for commit: 25632b63b70a4b25c55066a1781a972503fbd81e available at: https://gist.github.com/e225db270b77b18b1a72<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6802.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6802.dmg

huffman:
Added DO NOT MERGE. This doesn't fix the original problem I was trying to solve - I'm not 100% sure why yet, there may be a deeper problem.

hifi-gustavo:
Build console output for commit: 963c71a47698536a979479990f0e712ddfea401b available at: https://gist.github.com/e225db270b77b18b1a72<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6802.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6802.dmg

huffman:
To test:
1. Open interface, go to 0,0,0
2. Go into your domain settings and add this to the persistent script list: https://gist.githubusercontent.com/huffman/0ec1c70ba7af1904af8a/raw/727c20297b452745a736633ebc363a9904a51e36/acSimTest.js - this script will create a box at 0,0,0 and move a sphere inside of it every 4 seconds.
3. From interface, observe the sphere, and after a few seconds, confirm that it moves. Confirm that after another 4 seconds it moves again. Previously the behavior was that the sphere would jump back to the center of the box.

libraries/entities/src/EntityTree.cpp
```diff
@@ -194,7 +194,7 @@ bool EntityTree::updateEntityWithElement(EntityItemPointer entity, const EntityI
                     // the entire update is suspect --> ignore it
                     return false;
                 }
-            } else {
+            } else if (simulationBlocked) {
```
ZappoMan: @AndrewMeadows @sethalves - please review this change.... is this desired? what are the side effects of this

AndrewMeadows:
Yes, these changes are correct and necessary.

AndrewMeadows:
This PR fixes two bugs:

(1) Scripted AC's would state that they were simulation owners even though they did not have the intelligence for actually simulating objects.  Now fixed: AC's will not try to become simulation owners.

(2) It used to be impossible to anonymously change the "physics" properties (position, velocity) of an object that had no simulation owner.  Now AC's can update the position and velocity of an object while it is unowned without trying to take ownership.  Note: if the object becomes active to an observing interface's physics simulation as a consequence of AC property manipulation then the interface will soon "volunteer" to own its simulation, which will prevent the AC from making future anonymous physics updates.

AndrewMeadows:
I tested the example script.  It work in this PR but not yet in hifi/master.

huffman:
Looks like this is good to go whenever someone wants to merge it

PR #6810, opened 2016-01-11
Ensure Oculus session is valid before base class activation

jherico:
Once the base class `activate` method is run, the presentation thread is free to do stuff.  However the presentation thread context customization for oculus depends on the session being available, which isn't guaranteed due to the asynchronicity. The long term solution for all plugins is something similar to what we do with a similar problem with `uncustomizeContext` and `deactivate`... that is, use a conditional variable to ensure that `activate` is done before we proceed with `customizeContext`.  However it's trickier due to the ordering of operations.  

For now, this re-ordering of initializing the session and calling the base class `activate` should suffice.

### DIFF
--- a/plugins/oculus/src/OculusBaseDisplayPlugin.cpp
+++ b/plugins/oculus/src/OculusBaseDisplayPlugin.cpp
@@ -78,7 +78,6 @@ void OculusBaseDisplayPlugin::deinit() {
 }
 
 void OculusBaseDisplayPlugin::activate() {
-    WindowOpenGLDisplayPlugin::activate();
     if (!OVR_SUCCESS(ovr_Initialize(nullptr))) {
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: e58f99c989a3b00219183c88dcb527790494a53e available at: https://gist.github.com/3484d19cea242e3e35ef<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6810.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6810.dmg

ctrlaltdavid:
Tested on my PC and I can now successfully enter HMD view where before I was almost always crashing.

PR #6826, opened 2016-01-13
Fixed entity script loading functionality for the jsstreamplayer.

kevinmthomas-carpool:


ZappoMan:
we should also figure out why Script.load() isn't working. I don't think we really want two methods that do the same thing... as far as I can tell that's what we have here.

nbq:
Tested and put loader on an entity then left and rejoined the domain and the script was loaded into running scripts as it should.

PR #6830, opened 2016-01-14
Fix for arrows that stop moving too early

AndrewMeadows:
The new collisionMask feature and changes to the grab script caused the bow and arrow toy to stop working: arrows stop as soon as they are shot.

The problem is that the arrows are colliding with the bow itself and some "freeze in place" code is applied to the arrows as soon as they hit anything.

The solution is to make the arrows NOT collide with kinematic(*) objects when they are let loose.

Also, I noticed that the **collisionMask** feature would fight with **ignoreForCollisions** so I made **ignoreForCollisions" win when it comes to the final collision behavior in the physics engine.  

(*) Due to some idiosyncrasies in the code (which may be a bug, but I'll fix it later) things that have actions on them are currently categorized as "kinematic", and the bow falls into this category when it is held.

### DIFF
--- a/examples/toybox/bow/bow.js
+++ b/examples/toybox/bow/bow.js
@@ -540,6 +540,7 @@
                 var arrowProperties = {
                     collisionsWillMove: true,
                     ignoreForCollisions: false,
+                    collisionMask: "static,dynamic,otherAvatar", // workaround: not with kinematic --> no collision with bow
                     velocity: releaseVelocity,
                     gravity: ARROW_GRAVITY,
                     lifetime: 10,
--- a/libraries/entities/src/EntityItem.h
+++ b/libraries/entities/src/EntityItem.h
@@ -275,6 +275,7 @@ public:
     void setIgnoreForCollisions(bool value) { _ignoreForCollisions = value; }
 
     uint8_t getCollisionMask() const { return _collisionMask; }
+    uint8_t getFinalCollisionMask() const { return _ignoreForCollisions ? 0 : _collisionMask; }
     void setCollisionMask(uint8_t value) { _collisionMask = value; }
 
     bool getCollisionsWillMove() const { return _collisionsWillMove; }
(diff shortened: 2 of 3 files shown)


cozza13:
ok to test

hifi-gustavo:
Build console output for commit: 0126684a9623486e9c07dde56162efaf55a23786 available at: https://gist.github.com/44aecaff0afcbcecaa2e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6830.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6830.dmg

hifi-gustavo:
Build console output for commit: 0126684a9623486e9c07dde56162efaf55a23786 available at: https://gist.github.com/44aecaff0afcbcecaa2e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6830.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6830.dmg

PhilipRosedale:
will the arrows still stick in the wall when they hit it?

howard-stearns:
In the long run, don't we _want_ arrows to collide with moving objects -- particularly avatars?

cozza13:
This works for me.

examples/toybox/bow/bow.js
```diff
@@ -540,6 +540,7 @@
                 var arrowProperties = {
                     collisionsWillMove: true,
                     ignoreForCollisions: false,
+                    collisionMask: "static,dynamic,otherAvatar", // workaround: not with kinematic --> no collision with bow
```
AndrewMeadows: The arrow collides with **static,dynamic,otherAvatar**.

Note that there is a distinction between the **myAvatar** and the **otherAvatar** groups. So the arrow will still collide with other avatars... if the shooting avatar owns the arrow's simulation(1).

(1) If the interface owns the simulation then it treats these groups explicitly, but if the simulation is owned by someone else then it swaps the meaning of these two groups for its own collisions and lets the other simulation owners do likewise.

AndrewMeadows:
Yes, the arrows will stick to the walls and other avatars.

The arrows will NOT hit kinematic objects (non-dynamic but moving objects, which happens to include "held" objects ATM), however this is a **temporary workaround**.  We'll solve it correctly when we have proper collisionMask operations (another PR) and more correct collision group assignments (I'm currently working toward this).

PR #6833, opened 2016-01-14
Choice of hand ray pick or head pick

PhilipRosedale:
For testing: in handControllerGrab.js change PICK_WITH_HAND_RAY from true to false to pick only with the head and a target ball rather than the line.


RyanDowne:
I tried copying this script to my local dir and loading from disk but it doesn't work. Do I need a gist link or something?

cozza13:
@RyanDowne you can use this link https://raw.githubusercontent.com/PhilipRosedale/hifi/searchChoice/examples/controllers/handControllerGrab.js

cozza13:
I keep seeing the blocks shoot of when I go to grab them. Does not happened everytime. Watch what happens to the blue block https://i.gyazo.com/593095a4522a102821fee6c837613457.gif

RyanDowne:
Same thing is happening to me with the shooting blocks

RyanDowne:
Actually to make it happen go from searching to selecting. If you have the trigger all the way down and collide with an object it won't shoot off.

PhilipRosedale:
the shooting blocks is some new bug.  Andrew is looking.

imgntn:
@PhilipRosedale i like separating the hand /head options; this PR seems to disable the overlay lines while searching -- just wanted to make sure that's a desired effect.

PhilipRosedale:
@imgntn When picking using the head, not seeing the line is exactly the desired effect.  The reason is that showing the line when picking with the 'head' ray (meaning whatever is exactly under the ball) is confusing because you see the ray pass through things that are not getting selected, which is confusing.  So either we need to decide that we want to pick from the camera (and not show the line), or pick as we used to from the hand, in which case you do want to show the line.

imgntn:
gotcha -- with the current settings below, i don't see the hand ray when searching prior to intersecting an object

```var HAND_HEAD_MIX_RATIO = 0.0; ```

```var PICK_WITH_HAND_RAY = true```

```var USE_OVERLAY_LINES_FOR_SEARCHING =true```

 ```  if ((USE_OVERLAY_LINES_FOR_SEARCHING === true) && PICK_WITH_HAND_RAY) {
                  this.overlayLineOn(handPosition, searchSphereLocation, this.triggerSmoothedGrab() ? INTERSECT_COLOR : NO_INTERSECT_COLOR);```

if we substitute ```searchSphereLocation``` for ```searchVisualizationPickRay``` then you would see the lines while searching, whereas right now you don't see it unless the search sphere has hit an entity.  using the whiteboard or grabbing things out of the air is hard.  maybe there's a way to tie the opacity of the searching hand ray overlay to the head mix ratio.

imgntn:
there may be two 'search' states to separate and make options here:  one pre-intersection and one post-intersection but pre-grab

PhilipRosedale:
Yes OK I see what you mean, will add that change and pull this PR today.

PhilipRosedale:
Added always showing search ray as requested, now looking for how we've regressed and are starting grab when search beam is blue.

PR #6836, opened 2016-01-14
wire up audio show level meter menu

ZappoMan:
also included is some slight juggling of the developer menus to place items in more appropriate groups

### DIFF
--- a/interface/src/ui/AvatarInputs.h
+++ b/interface/src/ui/AvatarInputs.h
@@ -17,7 +17,7 @@
 public: \
     type name() { return _##name; }; \
 private: \
-    type _##name{ initialValue }; 
+    type _##name{ initialValue };
 
 class AvatarInputs : public QQuickItem {
     Q_OBJECT
@@ -31,6 +31,7 @@ class AvatarInputs : public QQuickItem {
     AI_PROPERTY(bool, mirrorVisible, false)
     AI_PROPERTY(bool, mirrorZoomed, true)
     AI_PROPERTY(bool, isHMD, false)
+    AI_PROPERTY(bool, showAudioTools, true)
 
 public:
     static AvatarInputs* getInstance();
@@ -46,6 +47,7 @@ signals:
     void mirrorVisibleChanged();
     void mirrorZoomedChanged();
     void isHMDChanged();
+    void showAudioToolsChanged();
 
 protected:
     Q_INVOKABLE void resetSensors();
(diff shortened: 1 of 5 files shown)


hifi-gustavo:
Build console output for commit: 0ad39feeb60eb154f2b516c2351a9d9a22b74699 available at: https://gist.github.com/e0b6440693050156e53f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6836.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6836.dmg

RyanDowne:
Tested on PC. Works great.

interface/src/ui/AvatarInputs.h
```diff
@@ -31,6 +31,7 @@ class AvatarInputs : public QQuickItem {
     AI_PROPERTY(bool, mirrorVisible, false)
     AI_PROPERTY(bool, mirrorZoomed, true)
     AI_PROPERTY(bool, isHMD, false)
```
jherico: isHMD is no longer used, so we should remove it as a property.

ZappoMan: I figure there's a chance we might use it in the future for some hmd only overlay

interface/src/ui/AvatarInputs.h
```diff
@@ -31,6 +31,7 @@ class AvatarInputs : public QQuickItem {
...
+    AI_PROPERTY(bool, showAudioTools, true)
```
jherico: technically I don't think you should have to explicitly add this.  You should be able to find the corresponding item in the QML copy of the menu...  however since we don't have any example code for this, I'm not going to block the PR on it.  Hopefully in this or another QML cleanup pass I can drop this class entirely.

interface/src/ui/AvatarInputs.h
jherico: Right, but even in this case, we shouldn't be exposing it in this manner, but rather putting an object that any dialog can connect to.  However, I don't really care enough right now, so I'll merge as is.

interface/src/ui/AvatarInputs.h
ZappoMan: ok - I was following the similar model of cameraEnabled that also uses a menu item.

PR #6843, opened 2016-01-15
fix bug causing grab to disappear objects

AndrewMeadows:
This PR fixes a problem that was puzzling me last night but when I woke up this morning I had realized what was wrong: we must be producing a NaN.  Sure enough, and it was all my fault.

### DIFF
--- a/libraries/physics/src/ObjectAction.cpp
+++ b/libraries/physics/src/ObjectAction.cpp
@@ -241,7 +241,6 @@ void ObjectAction::activateBody(bool forceActivation) {
     auto rigidBody = getRigidBody();
     if (rigidBody) {
         rigidBody->activate(forceActivation);
-        assert(rigidBody->isActive());
     } else {
         qDebug() << "ObjectAction::activateBody -- no rigid body" << (void*)rigidBody;
     }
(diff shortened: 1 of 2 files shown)


libraries/physics/src/ObjectAction.cpp
```diff
@@ -241,7 +241,6 @@ void ObjectAction::activateBody(bool forceActivation) {
     auto rigidBody = getRigidBody();
     if (rigidBody) {
         rigidBody->activate(forceActivation);
-        assert(rigidBody->isActive());
```
AndrewMeadows: Now that we're using  specifying **forceActivation** we shouldn't assert here.

hifi-gustavo:
Build console output for commit: b9352d3eaaaa2cf472fa571f7ef3498ae317fae7 available at: https://gist.github.com/e36fdd588b97955e4292<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6843.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6843.dmg

PR #6844, opened 2016-01-15
Fix JS web window crash caused by race condition

jherico:
The mechanism used to execute a URL change from a script could incorrectly be called on the script thread.


hifi-gustavo:
Build console output for commit: bb02a1eda2650d6fa6d75d9edcf23e606d80f15a available at: https://gist.github.com/15d03059ee56e9d63913<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6844.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6844.dmg

cozza13:
I tested this and it fixes the crash

PR #6846, opened 2016-01-16
edit-test script

howard-stearns:
A test script to confirm that the specified edit rate can be achieved for the specified number of objects.


imgntn:
trippy!

PR #6847, opened 2016-01-16
tribbles

howard-stearns:
Drops a bunch of "tribbles" (small spheres) in front of you.
They edit change color at 60hz and jump randomly at an average 1hz for 30 seconds, and then dissapear after 60 seconds.
Useful for testing the load of physical objects.

### DIFF
--- a/examples/tests/performance/tribbles.js
+++ b/examples/tests/performance/tribbles.js
@@ -1,0 +1,96 @@
+"use strict";
+/*jslint nomen: true, plusplus: true, vars: true*/
+var Vec3, Quat, MyAvatar, Entities, Camera, Script, print;
+//
+//  Created by Howard Stearns
+//  Copyright 2016 High Fidelity, Inc.
+//  
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+//  Drops a bunch of physical spheres in front of you, each running a script that will:
+//  * Edit color at EDIT_RATE for EDIT_TIMEOUT.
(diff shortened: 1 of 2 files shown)


howard-stearns:
For testing, you need the main script (tribbles.js) to reference the entityScripts/tribble.js.

You can do this by editing tribble.js to point to a (possibly file) url that has the entity script (such as http://howard-stearns.github.io/models/scripts/load/entityScriptTribble.js), 
or test with the pre-edited copy at http://howard-stearns.github.io/models/scripts/load/tribbles.js

examples/tests/performance/tribbles.js
```diff
@@ -0,0 +1,96 @@
...
+            dimensions: {x: SIZE, y: SIZE, z: SIZE},
+            color: {red: Math.random() * 255, green: Math.random() * 255, blue: Math.random() * 255},
+            velocity: VELOCITY,
+            angularVelocity: Vec3.multiply(Math.random(), ANGULAR_VELOCITY),
+            damping: DAMPING,
+            angularDamping: ANGULAR_DAMPING,
+            gravity: GRAVITY,
+            collisionsWillMove: true,
+            lifetime: LIFETIME,
+            script: "https://s3.amazonaws.com/hifi-public/scripts/entityScripts/tribble.js"
```
imgntn: you could do a relative reference if you want ```Script.resolvePath(../../entityScripts/tribble.js```

imgntn:
works for me windows 8

examples/tests/performance/tribbles.js
howard-stearns: I haven't been confident enough that the definition of Script.resolvePath is stable enough, and a quick search did not reveal any other scripts doing this (for entityScripts).  But if you feel it's stable enough, I could do it either now or next time around...

PR #6848, opened 2016-01-18
Include checking the angular size of entities on the server side for LOD support

ZappoMan:
We only want to send down entities from the server to the client that are "large enough" to see based on the LOD settings in the client. The current (old) code only checked that the octree cell was "large enough" to consider, but didn't double check that the entities were in fact large enough to consider. This means that small entities that happen to live inside of larger octree cells (because they cross a cell boundary for example) would still get sent down, even though they wouldn't be rendered.

This PR fixes that by also checking against the size of the entity before determining if it should be included.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -3507,14 +3507,6 @@ glm::vec3 Application::getSunDirection() {
 // FIXME, preprocessor guard this check to occur only in DEBUG builds
 static QThread * activeRenderingThread = nullptr;
 
-float Application::getSizeScale() const {
-    return DependencyManager::get<LODManager>()->getOctreeSizeScale();
-}
-
-int Application::getBoundaryLevelAdjust() const {
-    return DependencyManager::get<LODManager>()->getBoundaryLevelAdjust();
-}
-
 PickRay Application::computePickRay(float x, float y) const {
     vec2 pickPoint { x, y };
     PickRay result;
--- a/interface/src/Application.h
+++ b/interface/src/Application.h
@@ -170,8 +170,6 @@ public:
 
     virtual ViewFrustum* getCurrentViewFrustum() { return getDisplayViewFrustum(); }
     virtual QThread* getMainThread() { return thread(); }
-    virtual float getSizeScale() const;
-    virtual int getBoundaryLevelAdjust() const;
     virtual PickRay computePickRay(float x, float y) const;
     virtual glm::vec3 getAvatarPosition() const;
     virtual void overrideEnvironmentData(const EnvironmentData& newData) { _environment.override(newData); }
--- a/interface/src/LODManager.cpp
+++ b/interface/src/LODManager.cpp
@@ -25,7 +25,7 @@ Setting::Handle<float> hmdLODDecreaseFPS("hmdLODDecreaseFPS", DEFAULT_HMD_LOD_DO
 // pid: renderDistance is adjusted by a PID such that frame rate targets are met.
 // acuity:  a pseudo-acuity target is held, or adjusted to match minimum frame rates (and a PID controlls avatar rendering distance)
 // If unspecified, acuity is used only if user has specified non-default minumum frame rates.
-Setting::Handle<int> lodPreference("lodPreference", (int)LODManager::LODPreference::unspecified);
+Setting::Handle<int> lodPreference("lodPreference", (int)LODManager::LODPreference::acuity);
 const float SMALLEST_REASONABLE_HORIZON = 50.0f; // meters
 Setting::Handle<float> renderDistanceInverseHighLimit("renderDistanceInverseHighLimit", 1.0f / SMALLEST_REASONABLE_HORIZON);
 void LODManager::setRenderDistanceInverseHighLimit(float newValue) {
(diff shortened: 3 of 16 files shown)


hifi-gustavo:
Build console output for commit: 88af387214546ae4b56186e0c109219f940df040 available at: https://gist.github.com/e2841bac862266cbd354<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6848.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6848.dmg

interface/src/LODManager.cpp
```diff
@@ -274,92 +274,23 @@ void LODManager::updatePIDRenderDistance(float targetFps, float measuredFps, flo
...
-    return distanceToCamera <= visibleDistanceAtClosestScale;
+    // FIXME - eventually we want to use the render accuracy as an indicator for the level of detail
+    // to use in rendering.
+    float renderAccuracy = args->_viewFrustum->calculateRenderAccuracy(bounds, args->_sizeScale, args->_boundaryLevelAdjust);
+    return (renderAccuracy > 0.0f);
 };
 
-// TODO: This is essentially the same logic used to render octree cells, but since models are more detailed then octree cells
-//       I've added a voxelToModelRatio that adjusts how much closer to a model you have to be to see it.
-bool LODManager::shouldRenderMesh(float largestDimension, float distanceToCamera) {
```
ZappoMan: this was dead code, it was not called anywhere in the code base.

interface/src/LODManager.cpp
```diff
@@ -274,92 +274,23 @@ void LODManager::updatePIDRenderDistance(float targetFps, float measuredFps, flo
...
+        float distanceToCamera = glm::length(bounds.calcCenter() - args->_viewFrustum->getPosition());
+        float largestDimension = bounds.getLargestDimension();
         const float scenerySize = 300; // meters
         bool isRendered = (largestDimension > scenerySize) || // render scenery regardless of distance
             (distanceToCamera < renderDistance + largestDimension);
         renderedCount += isRendered ? 1 : 0;
         return isRendered;
     }
-
-    const float maxScale = (float)TREE_SCALE;
```
ZappoMan: I moved this logic to ViewFrustum because I need it in the server, and the LOD manager is not in the server. Plus the ViewFrustum contains the full details about the camera position which is required for this calculation.

libraries/octree/src/ViewFrustum.cpp
```diff
@@ -750,3 +750,48 @@ void ViewFrustum::evalViewTransform(Transform& view) const {
     view.setTranslation(getPosition());
     view.setRotation(getOrientation());
 }
+
+float ViewFrustum::calculateRenderAccuracy(const AABox& bounds, float octreeSizeScale, int boundaryLevelAdjust) const {
```
ZappoMan: this is the same logic/code from the old `LODManager::shouldRender()` I've moved it to ViewFrustum to be accessible to server, and I've renamed it `calculateRenderAccuracy()` and have it return a `float` instead of a `bool` so that we can in the future implement support for "less accurate but still visible"

interface/src/LODManager.cpp
```diff
@@ -25,7 +25,7 @@ Setting::Handle<float> hmdLODDecreaseFPS("hmdLODDecreaseFPS", DEFAULT_HMD_LOD_DO
 // pid: renderDistance is adjusted by a PID such that frame rate targets are met.
 // acuity:  a pseudo-acuity target is held, or adjusted to match minimum frame rates (and a PID controlls avatar rendering distance)
 // If unspecified, acuity is used only if user has specified non-default minumum frame rates.
-Setting::Handle<int> lodPreference("lodPreference", (int)LODManager::LODPreference::unspecified);
+Setting::Handle<int> lodPreference("lodPreference", (int)LODManager::LODPreference::acuity);
```
howard-stearns: Yes. There's a bunch of things wrong about my "non-acuity" code. I'll fix that as soon as the dust settles.
So, thanks, and good to fix things here (such as changing default), and especially to the extent of developing/testing the acuity case here, but don't worry about it.

interface/src/LODManager.cpp
howard-stearns: +1 for removing dead code.
Shouldn't we also remove the (now write-only) references to _shouldRenderTableNeedsRebuilding ?

howard-stearns: While we're at at it, there's a getSizeScale() that appears to be dead code. (Not getOctreeSizeScale(), which is very much alive.) We might consider removing getSizeScale().

ZappoMan: re: _shouldRenderTableNeedsRebuilding - I'm pretty sure that's still needed, I will check.

ZappoMan: hmmm --- I see a problem here... I'm going to add DO NOT MERGE to this... the change I made doesn't handle this table rebuilding correctly.

howard-stearns: For what it’s worth, I _think_ the change in this PR accurately reflects what the release code is doing for table rebuilding — i.e., it doesn’t rebuild it.
I _think_ that only the obsolete shouldRenderMesh was rebuilding. I have no idea what we _should_ have been doing.

ZappoMan: ok - cleaned up

interface/src/Application.cpp
```diff
@@ -3507,14 +3507,6 @@ glm::vec3 Application::getSunDirection() {
 // FIXME, preprocessor guard this check to occur only in DEBUG builds
 static QThread * activeRenderingThread = nullptr;
 
-float Application::getSizeScale() const {
```
ZappoMan: dead code

interface/src/Application.h
```diff
@@ -170,8 +170,6 @@ class Application : public QApplication, public AbstractViewStateInterface, publ
 
     virtual ViewFrustum* getCurrentViewFrustum() { return getDisplayViewFrustum(); }
     virtual QThread* getMainThread() { return thread(); }
-    virtual float getSizeScale() const;
```
ZappoMan: dead code

interface/src/LODManager.cpp
```diff
@@ -274,96 +273,26 @@ void LODManager::updatePIDRenderDistance(float targetFps, float measuredFps, flo
...
-        visibleDistanceAtClosestScale *= 2.0f;
-    }
-    
-    return (distanceToCamera <= visibleDistanceAtClosestScale);
-}
-
 void LODManager::setOctreeSizeScale(float sizeScale) {
     _octreeSizeScale = sizeScale;
     calculateAvatarLODDistanceMultiplier();
-    _shouldRenderTableNeedsRebuilding = true;
```
ZappoMan: dead code

interface/src/LODManager.h
```diff
@@ -136,9 +135,6 @@ class LODManager : public QObject, public Dependency {
     SimpleMovingAverage _fpsAverageDownWindow = DOWN_SHIFT_SAMPLES_OF_FRAMES;
     SimpleMovingAverage _fpsAverageUpWindow = UP_SHIFT_SAMPLES_OF_FRAMES;
     
-    bool _shouldRenderTableNeedsRebuilding = true;
```
ZappoMan: dead code

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -417,15 +417,6 @@ const FBXGeometry* EntityTreeRenderer::getCollisionGeometryForEntity(EntityItemP
     return result;
 }
 
-float EntityTreeRenderer::getSizeScale() const {
```
ZappoMan: dead code

libraries/entities-renderer/src/EntityTreeRenderer.h
```diff
@@ -41,8 +41,6 @@ class EntityTreeRenderer : public OctreeRenderer, public EntityItemFBXService, p
     virtual char getMyNodeType() const { return NodeType::EntityServer; }
     virtual PacketType getMyQueryMessageType() const { return PacketType::EntityQuery; }
     virtual PacketType getExpectedPacketType() const { return PacketType::EntityData; }
-    virtual float getSizeScale() const;
```
ZappoMan: dead code

libraries/octree/src/Octree.cpp
```diff
@@ -50,10 +50,6 @@
 
 QVector<QString> PERSIST_EXTENSIONS = {"svo", "json", "json.gz"};
 
-float boundaryDistanceForRenderLevel(unsigned int renderLevel, float voxelSizeScale) {
```
ZappoMan: moved to ViewFrustum

libraries/octree/src/OctreeHeadlessViewer.cpp
```diff
@@ -14,11 +14,7 @@
 #include "OctreeLogging.h"
 #include "OctreeHeadlessViewer.h"
 
-OctreeHeadlessViewer::OctreeHeadlessViewer() :
-    OctreeRenderer(),
```
ZappoMan: coding standard cleanup

libraries/octree/src/OctreeHeadlessViewer.h
```diff
@@ -59,7 +59,7 @@ public slots:
 
     // getters for LOD and PPS
     float getVoxelSizeScale() const { return _voxelSizeScale; }
-    int getBoundaryLevelAdjust() const override { return _boundaryLevelAdjust; }
+    int getBoundaryLevelAdjust() const { return _boundaryLevelAdjust; }
```
ZappoMan: no longer exposed as part of the abstract base class, so no need to include override

hifi-gustavo:
Build console output for commit: b8747709efe1dc209d51ce335c2bbe99e2384aa4 available at: https://gist.github.com/e2841bac862266cbd354<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6848.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6848.dmg

PR #6854, opened 2016-01-18
Access avatar parent information from js

sethalves:
- scripts can call MyAvatar.setParentID, getParentID, setParentJointIndex, getParentJointIndex


hifi-gustavo:
Build console output for commit: 135f25bde79c895e5304007dced67895a9589bd3 available at: https://gist.github.com/ad67cbcb363e42c27db6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6854.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6854.dmg

interface/src/avatar/Avatar.cpp
```diff
@@ -1209,3 +1209,27 @@ void Avatar::updatePalms() {
...
+    bool success;
+    Transform beforeChangeTransform = getTransform(success);
+    SpatiallyNestable::setParentJointIndex(parentJointIndex);
+    if (success) {
+        setTransform(beforeChangeTransform, success);
+        if (!success) {
+            qDebug() << "Avatar::setParentJointIndex failed to reset avatar's location.";
+        }
+    }
+}
```
AndrewMeadows: These setters work for any Avatar rather than just the MyAvatar instance.  This doesn't actually allow one interface to "update" the parent for another Avatar from that remote interface's perspective... maybe it allows a local script to temporarily update it from the perspective of the local interface?  That is, one could write a script that makes it look (on your end) that someone else is parented?

So maybe these methods should just fail if **isMyAvatar()** returns **false**?

sethalves: This is true, but I think the local change will be reverted when the next relevant avatar-mixer packet is received.  I'll make the suggested change, even so.

interface/src/avatar/Avatar.h
```diff
@@ -167,6 +167,12 @@ class Avatar : public AvatarData {
     using SpatiallyNestable::setOrientation;
     virtual void setOrientation(const glm::quat& orientation) override;
 
+    // these call through to the SpatiallyNestable versions, but they are here to expose these to javascript.
+    Q_INVOKABLE virtual QUuid getParentID() const { return SpatiallyNestable::getParentID(); }
```
hyperlogic: SpacialNestible::getParentID() returns a QUuid without locking.  If this is called while another thread is setting the parentID, it could result in an incorrect QUuid, because copying QUuid is not atomic.

libraries/avatars/src/AvatarData.cpp
```diff
@@ -216,15 +216,13 @@ QByteArray AvatarData::toByteArray(bool cullSmallChanges, bool sendAll) {
...
-    SpatiallyNestablePointer parent = getParentPointer(success);
-    if (parent && success) {
+    if (!getParentID().isNull()) {
         setAtBit(bitItems, HAS_REFERENTIAL);
     }
     *destinationBuffer++ = bitItems;
 
-    if (parent) {
-        QByteArray referentialAsBytes = parent->getID().toRfc4122();
+    if (!getParentID().isNull()) {
```
hyperlogic: What would happen if the first getParentID() was not NULL, but the second was?  I suppose this could happen if script sets the parentID while the AvatarData packet is being built.

libraries/avatars/src/AvatarData.cpp
```diff
@@ -492,10 +490,12 @@ int AvatarData::parseDataFromBuffer(const QByteArray& buffer) {
         // avatar's SkeletonModel might fall into the CPU expensive part of Model::updateClusterMatrices() when otherwise it
         // would not have required it.  However, we know we can update many simultaneously animating avatars, and most
         // avatars will be moving constantly anyway, so I don't think we need to worry.
-        if (getBodyYaw() != yaw || getBodyPitch() != pitch || getBodyRoll() != roll) {
+        glm::quat currentOrientation = getLocalOrientation();
+        glm::vec3 newEulerAngles(pitch, yaw, roll);
+        glm::quat newOrientation = glm::quat(glm::radians(newEulerAngles));
+        if (currentOrientation != newOrientation) {
```
hyperlogic: We might want to do a fuzzy check here, the localOrientation quaternion and the newOrientation quaternion are generated differently, and will probably never be binary equal.

sethalves: I think they are generated in the same way.  It's comparing the current value with the one generated in a previous call.

libraries/avatars/src/AvatarData.cpp
sethalves: good point.  I've adjusted it.

hifi-gustavo:
Build console output for commit: 09dd3e2a563751b38dc2b0e6d151a2ebaa830f92 available at: https://gist.github.com/ad67cbcb363e42c27db6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6854.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6854.dmg

libraries/avatars/src/AvatarData.cpp
hyperlogic: Ok, I see. *NEVERMIND*

hifi-gustavo:
Build console output for commit: 90847dc49c201be9df57ad9d40f9ccea45ac6e2c available at: https://gist.github.com/ad67cbcb363e42c27db6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6854.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6854.dmg

PR #6855, opened 2016-01-19
don't persist entites with invalid parentIDs

sethalves:
- entity server won't persist entities that have a (non-null) parentID which doesn't resolve to an entity.

### DIFF
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -1291,12 +1291,13 @@ void EntityTree::remapIDs() {
     recurseTreeWithOperator(&theOperator);
 }
 
-bool EntityTree::writeToMap(QVariantMap& entityDescription, OctreeElementPointer element, bool skipDefaultValues) {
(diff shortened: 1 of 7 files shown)


hifi-gustavo:
Build console output for commit: e365d93ed909be8595e4b9d28847c8ef4aa1dc4f available at: https://gist.github.com/320ec0a25d8bfa680c8c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6855.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6855.dmg

sethalves:
to test -- create an entity, give it a searchable name, edit it with edit.js.  change some digit of its (currently null) parentID.  The entity will vanish, because it has an invalid parentID.  Wait for the entity server to persist to file.  check the file contents to see that most things are there but the entity that was adjusted is not.

AndrewMeadows:
I tested:

(1) entities with bad ID's are not persisted -- works
(2) entities with good ID's are persisted -- works

PR #6856, opened 2016-01-19
Avatar: Fix for one frame lag in hold action

hyperlogic:
AvatarActionHold now uses the most recent un-cached palm positions/orientations
instead of the cached version from the previous frame.

### DIFF
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -1224,22 +1224,9 @@ void Avatar::setOrientation(const glm::quat& orientation) {
 }
 
 void Avatar::updatePalms() {
-
-    // get palm rotations
-    glm::quat leftPalmRotation, rightPalmRotation;
-    getSkeletonModel().getJointRotationInWorldFrame(getSkeletonModel().getLeftHandJointIndex(), leftPalmRotation);
-    getSkeletonModel().getJointRotationInWorldFrame(getSkeletonModel().getRightHandJointIndex(), rightPalmRotation);
-
-    // get palm positions
-    glm::vec3 leftPalmPosition, rightPalmPosition;
-    getSkeletonModel().getLeftHandPosition(leftPalmPosition);
-    getSkeletonModel().getRightHandPosition(rightPalmPosition);
-    leftPalmPosition += HAND_TO_PALM_OFFSET * glm::inverse(leftPalmRotation);
-    rightPalmPosition += HAND_TO_PALM_OFFSET * glm::inverse(rightPalmRotation);
-
     // update thread-safe caches
-    _leftPalmRotationCache.set(leftPalmRotation);
-    _rightPalmRotationCache.set(rightPalmRotation);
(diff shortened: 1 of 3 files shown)


hifi-gustavo:
Build console output for commit: ec2f8db83a70ff7b6023b92fc656a504933de0f1 available at: https://gist.github.com/12e34d8a25db7d0ebd6e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6856.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6856.dmg

interface/src/avatar/Avatar.cpp
```diff
@@ -1160,25 +1160,59 @@ void Avatar::rebuildCollisionShape() {
...
+    return rightPalmPosition;
+}
+
+glm::quat Avatar::getUncachedRightPalmRotation() const {
+    assert(QThread::currentThread() == thread());  // main thread access only
+    glm::quat rightPalmRotation;
+    getSkeletonModel().getJointRotationInWorldFrame(getSkeletonModel().getRightHandJointIndex(), rightPalmRotation);
+    return rightPalmRotation;
+}
+
```
AndrewMeadows: I was thinking... maybe where we're calling this should be the place where we cache the palm transforms.  But there is no guarantee that there will be a hold action to invoke it.

So, maybe in the context where **Action::prepareForPhysicsSimulation()** would be called we should compute and cache the palm transforms there?

Also, this work now copies and splits up the implementation of the palm transform calculations into new places (it already exists in **Avatar::updatePalms()**).  Would it make sense to just call (maybe re-call?) updatePalms() in the right place?  or would that ruin the idea of caching the result since other things flying in close formation may have pulled from the previous values of PalmPositionCache?

sethalves:
I've tried this -- it does seem to fix the squishiness of held entities.

interface/src/avatar/Avatar.cpp
hyperlogic: So, you're right that this duplicates the code for calculating the palmRotation/Position in two places.  `Avatar::updatePalms()` & `Avatar::getUncachedXPalmY()`.  I'll consolidate them.

I *think* we should keep the computation of the cached version in the same place (for MyAvatar) it is very late in the game loop, after physics, because the methods return the palms in *world* frame, so it is very important that they match the rendered position.

hyperlogic: I just pushed b750128 which consolidates the palm local to world transformations.

hifi-gustavo:
Build console output for commit: b750128c945b4646995cfa36f3b8e6227e3b4241 available at: https://gist.github.com/12e34d8a25db7d0ebd6e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6856.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6856.dmg

PR #6865, opened 2016-01-19
fix premature octree query before preferences are loaded

ZappoMan:
In the case of starting up while looking away from content, we would send queries to the octree server before the preferences were actually loaded, causing us to end up getting some domain content that might not be in the direction we were looking.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -2653,6 +2653,8 @@ void Application::loadSettings() {
 
     Menu::getInstance()->loadSettings();
     getMyAvatar()->loadData();
+
+    _settingsLoaded = true;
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: bcdff7ba4a42c8fe68d3da69233b5f9a5c0fc61c available at: https://gist.github.com/0ac9332088be6063889f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6865.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6865.dmg

PR #6869, opened 2016-01-20
Make Pistol Quieter and Minor Adjustments

imgntn:
Previously, the pistol was sometimes noisy.  This PR changes the collision sound and makes some minor adjustments.  To test: https://rawgit.com/imgntn/hifi/pistolfix/examples/toybox/pistol/createPistol.js -- the sound should not hurt to hear.


nbq:
I notice it is quieter when I drop the gun but nothing I do can make it fire on my end. I am not sure if this is a problem on my end or not so @cozza13 maybe you can check this also and see if it fires for you.

PR #6874, opened 2016-01-20
Remove Particle Search Beams from Grab Script and Create / Destroy Systems on Demand

imgntn:
Previously, each avatar had two hidden  particle systems that were used for search and move beams.  Now, since we're only using particle beams for moving, we create and destroy those beams on grab / release of an object instead of having them stick around.  When you test this script, you should only see a particle system in your entities list while you're actively far grabbing something.

https://rawgit.com/imgntn/hifi/fixgrabparticlesystems/examples/controllers/handControllerGrab.js


nbq:
@cozza13 @imgntn I just tested this and I only found that a "ParticleEffect Particle Beam" showed up when having something directly held (with beams) using the Hydras.

PR #6877, opened 2016-01-21
fix interface crash when deleting lots of objects at once

AndrewMeadows:
There was a crash mode where deleting lots of objects at once (using edit.js) it was possible to crash a debug interface client on an assert (and would probably sometimes crash a release client too).

After looking at it I decided that there is a race condition where the **PhysicalEntitySimulation** briefly holds no references to an **EntityItem** while it disassembles the backpointers between it and its **EntityMotionState**.  If the last reference happens to be deleted inside this short window (via the render thread, for example) then  the  backpointers may not quite be cleaned up yet and an assert will be hit.

The solution is to keep a reference until we're sure the backpointers are properly disassembled.

### DIFF
--- a/libraries/physics/src/PhysicalEntitySimulation.cpp
+++ b/libraries/physics/src/PhysicalEntitySimulation.cpp
@@ -142,6 +142,7 @@ void PhysicalEntitySimulation::clearEntitiesInternal() {
     // finally clear all lists maintained by this class
     _physicalObjects.clear();
     _entitiesToRemoveFromPhysics.clear();
+    _entitiesToRelease.clear();
     _entitiesToAddToPhysics.clear();
     _pendingChanges.clear();
     _outgoingChanges.clear();
@@ -157,6 +158,7 @@ void PhysicalEntitySimulation::prepareEntityForDelete(EntityItemPointer entity) 
 // end EntitySimulation overrides
 
 void PhysicalEntitySimulation::getObjectsToRemoveFromPhysics(VectorOfMotionStates& result) {
+    _entitiesToRelease.clear();
     result.clear();
     QMutexLocker lock(&_mutex);
     for (auto entity: _entitiesToRemoveFromPhysics) {
@@ -171,7 +173,7 @@ void PhysicalEntitySimulation::getObjectsToRemoveFromPhysics(VectorOfMotionState
             _entitiesToDelete.insert(entity);
         }
     }
-    _entitiesToRemoveFromPhysics.clear();
+    _entitiesToRemoveFromPhysics.swap(_entitiesToRelease);
 }
 
 void PhysicalEntitySimulation::getObjectsToAddToPhysics(VectorOfMotionStates& result) {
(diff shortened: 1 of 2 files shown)


libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -171,7 +173,7 @@ void PhysicalEntitySimulation::getObjectsToRemoveFromPhysics(VectorOfMotionState
             _entitiesToDelete.insert(entity);
         }
     }
-    _entitiesToRemoveFromPhysics.clear();
+    _entitiesToRemoveFromPhysics.swap(_entitiesToRelease);
```
AndrewMeadows: **_entitiesToRelease** hangs onto **EntityItemPointer**s until the next frame.

hifi-gustavo:
Build console output for commit: e77189fb1bbfc6324c3ac42daf519be5b9ea2e8a available at: https://gist.github.com/64161a907855dc32882d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/pr-interface-win64-PR-6877.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/dev-interface-6877.dmg

howard-stearns:
See https://app.asana.com/0/search/82509696936664/80772566277365 and https://app.asana.com/0/search/82509696936664/74952139477938

libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -157,6 +158,7 @@ void PhysicalEntitySimulation::prepareEntityForDelete(EntityItemPointer entity)
 // end EntitySimulation overrides
 
 void PhysicalEntitySimulation::getObjectsToRemoveFromPhysics(VectorOfMotionStates& result) {
+    _entitiesToRelease.clear();
```
ZappoMan: does this need to be guarded by the mutex?

AndrewMeadows: Not **_entitiesToRelease** (because it is ever only used in this function (except on shutdown)) but maybe **_entitiesToRemoveFromPhysics** needs some protection.  I will audit.

AndrewMeadows: Oh I see.  The answer is "no".

cozza13:
This has been tested up to 5000 entities and is working

ZappoMan:
test this please

hifi-gustavo:
Build console output for commit: e77189fb1bbfc6324c3ac42daf519be5b9ea2e8a available at: https://gist.github.com/79b5436e66f4a856416e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6877.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6877.dmg

PR #6883, opened 2016-01-21
Test script for lod testing of sphere entities and sphere overlays

ericrius1:
Creates a table with two 1cm spheres next to each other- one is a sphere entity, the other is an overlay entity


ericrius1:
https://rawgit.com/ericrius1/hifi/sphereLOD/examples/tests/sphereLODTest.js

PR #6885, opened 2016-01-21
fix unpacking of lastEdited

sethalves:
- fix bug in client unpacking of lastEdited timestamp that caused some entity updates to be wrongly ignored.


murillodigital:
test this please

PR #6895, opened 2016-01-22
MyAvatar: Recenter when the head turns away from the hips

hyperlogic:
Compute the HMD facing moving average.  When the moving average diverges from the hips by more than 45 degrees, recenter the body.

Also, the follow code has been changed, instead of a follow velocity being passed to the CharacterController a desired target is passed.  The CharacterController homes toward it's target based on the time remaining.  Any follow deltas applied to move the avatar's position closer to it's target is stored and re-applied to the bodySensorMatrix.  This centralizes the moving/homing code to one place, the CharacterContoller.  The displacement each timestep is clamped, to avoid tunneling.

A new FollowHelper class was also introduced, it groups together the data and logic necessary to perform the re-centering/follow procedure.  This "hopefully" makes it easier to maintain.

Also, unit tests were added for new helper functions introduced to physics/BulletUtils.h

To test you must be wearing an HMD:

Once you are wearing an HMD, perform the following steps.

1. reset sensors using the "single quote" key.
2. look down and your avatar's toes/body should be pointing forward.
3. look to the right 90 degrees.  
4. Wait 3-4 seconds.
5. look down while maintaining your current head facing direction.  Your avatar toes should be aligned to your head facing.


hifi-gustavo:
Build console output for commit: ca8a832818231f8bc3ae004953ef4cd357bf782d available at: https://gist.github.com/3c70b17cd5f2dcf6507b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6895.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6895.dmg

interface/src/avatar/MyAvatar.cpp
```diff
@@ -263,9 +261,22 @@ void MyAvatar::reset(bool andReload) {
 
 void MyAvatar::update(float deltaTime) {
 
+    // update moving average of HMD facing in xz plane.
+    const float HMD_FACING_TIMESCALE = 4.0f; // very slow average
+    float tau = deltaTime / HMD_FACING_TIMESCALE;
+    _hmdSensorFacingMovingAverage = lerp(_hmdSensorFacingMovingAverage, _hmdSensorFacing, tau);
```
hyperlogic: This is the moving average code, basically it's lerping between two vector2s, the _hmdSensorFacing vector is normalized but the _hmdSensorFacingMovingAverage is not.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -366,46 +377,7 @@ void MyAvatar::updateFromHMDSensorMatrix(const glm::mat4& hmdSensorMatrix) {
     _hmdSensorMatrix = hmdSensorMatrix;
     _hmdSensorPosition = extractTranslation(hmdSensorMatrix);
     _hmdSensorOrientation = glm::quat_cast(hmdSensorMatrix);
-}
-
-void MyAvatar::updateHMDFollowVelocity() {
```
hyperlogic: The code to detect when to follow has been moved into the FollowHelper class.  See the shouldFollow() method.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1058,63 +1030,21 @@ void MyAvatar::prepareForPhysicsSimulation() {
...
-    } else {
-        setVelocity(_characterController.getLinearVelocity());
-    }
+    _bodySensorMatrix = _follow.postPhysicsUpdate(*this, _bodySensorMatrix);
 
     // now that physics has adjusted our position, we can update attachements.
-    Avatar::simulateAttachments(deltaType);
-}
-
-void MyAvatar::adjustSensorTransform() {
```
hyperlogic: The code to alter the _bodySensorMatrix has been moved into the FollowHelper, see the postPhysicsUpdate method.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1832,3 +1762,80 @@ void audioListenModeFromScriptValue(const QScriptValue& object, AudioListenerMod
...
+}
+
+bool MyAvatar::FollowHelper::shouldActivate(const MyAvatar& myAvatar, const glm::mat4& desiredBodyMatrix, const glm::mat4& currentBodyMatrix) const {
+
+    const float FOLLOW_ROTATION_THRESHOLD = cosf(PI / 4.0f);
+
+    glm::vec2 bodyFacing = getFacingDir2D(currentBodyMatrix);
+    if (glm::dot(myAvatar.getHMDSensorFacingMovingAverage(), bodyFacing) < FOLLOW_ROTATION_THRESHOLD) {
+        return true;
+    }
```
hyperlogic: This is where we detect if the angle between the hips and the moving HMD average is larger than 40 degrees.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1832,3 +1762,80 @@ void audioListenModeFromScriptValue(const QScriptValue& object, AudioListenerMod
...
+
+    const float CYLINDER_TOP = 0.1f;
+    const float CYLINDER_BOTTOM = -0.5f;
+    const float CYLINDER_RADIUS = 0.15f;
+
+    glm::vec3 offset = extractTranslation(desiredBodyMatrix) - extractTranslation(currentBodyMatrix);
+    glm::vec3 truncatedOffset(offset.x, 0.0f, offset.y);
+    float truncatedDistance = glm::length(truncatedOffset);
+
+    return (offset.y > CYLINDER_TOP) || (offset.y < CYLINDER_BOTTOM) || (truncatedDistance > CYLINDER_RADIUS);
```
hyperlogic: This cylinder check isn't quite the same as the old logic, but I think it's simpler while still allowing the avatar to squat.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1832,3 +1762,80 @@ void audioListenModeFromScriptValue(const QScriptValue& object, AudioListenerMod
...
+        activate();
+    }
+
+    if (isActive()) {
+        glm::mat4 desiredWorldMatrix = myAvatar.getSensorToWorldMatrix() * _desiredBodyMatrix;
+        myAvatar.getCharacterController()->setFollowParameters(desiredWorldMatrix, _timeRemaining);
+    } else {
+        glm::mat4 currentWorldMatrix = myAvatar.getSensorToWorldMatrix() * currentBodyMatrix;
+        myAvatar.getCharacterController()->setFollowParameters(currentWorldMatrix, 0.0f);
+    }
```
hyperlogic: Before physics simulation, we pass follow params to the CharacterController.  This is the desired target, in world coordinates, and the time left to reach it.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1832,3 +1762,80 @@ void audioListenModeFromScriptValue(const QScriptValue& object, AudioListenerMod
...
+        glm::quat worldAngularDisplacement = myAvatar.getCharacterController()->getFollowAngularDisplacement();
+        glm::quat sensorToWorld = glmExtractRotation(myAvatar.getSensorToWorldMatrix());
+        glm::quat worldToSensor = glm::inverse(sensorToWorld);
+
+        glm::vec3 sensorLinearDisplacement = worldToSensor * worldLinearDisplacement;
+        glm::quat sensorAngularDisplacement = worldToSensor * worldAngularDisplacement * sensorToWorld;
+
+        glm::mat4 newBodyMat = createMatFromQuatAndPos(sensorAngularDisplacement * glmExtractRotation(currentBodyMatrix),
+                                                       sensorLinearDisplacement + extractTranslation(currentBodyMatrix));
+        return newBodyMat;
```
hyperlogic: After physics simulation, we get the angular and linear displacements it used internally, and apply them to the bodySensorMatrix.  This will keep these two motions in sync, which will come in handy if we decide to make the following non-linear.

libraries/physics/src/CharacterController.cpp
```diff
@@ -189,12 +190,39 @@ void CharacterController::playerStep(btCollisionWorld* dynaWorld, btScalar dt) {
...
         btTransform bodyTransform = _rigidBody->getWorldTransform();
-        bodyTransform.setOrigin(bodyTransform.getOrigin() + dt * _followVelocity);
-        _rigidBody->setWorldTransform(bodyTransform);
+
+        btVector3 startPos = bodyTransform.getOrigin();
+        btVector3 deltaPos = _followDesiredBodyTransform.getOrigin() - startPos;
+        btVector3 vel = deltaPos / _followTimeRemaining;
+        btVector3 linearDisplacement = clampLength(vel * dt, MAX_DISPLACEMENT);  // clamp displacement to prevent tunneling.
+        btVector3 endPos = startPos + linearDisplacement;
+
```
hyperlogic: This is the homing code, we dynamically compute the velocity to always point toward our target, the speed is proportional to the distance to our target and inversely proportional to the time remaining.  We also clamp the displacement to 1/2 the capsule radius, to avoid it getting too large, and cause us to tunnel thru geometry.

libraries/physics/src/CharacterController.cpp
```diff
@@ -189,12 +190,39 @@ void CharacterController::playerStep(btCollisionWorld* dynaWorld, btScalar dt) {
...
+        btQuaternion startRot = bodyTransform.getRotation();
+        glm::vec2 currentFacing = getFacingDir2D(bulletToGLM(startRot));
+        glm::vec2 currentRight(currentFacing.y, -currentFacing.x);
+        glm::vec2 desiredFacing = getFacingDir2D(bulletToGLM(_followDesiredBodyTransform.getRotation()));
+        float deltaAngle = acosf(glm::clamp(glm::dot(currentFacing, desiredFacing), -1.0f, 1.0f));
+        float angularVel = deltaAngle / _followTimeRemaining;
+        float sign = copysignf(1.0f, glm::dot(desiredFacing, currentRight));
+        btQuaternion angularDisplacement = btQuaternion(btVector3(0.0f, 1.0f, 0.0f), sign * angularVel * dt);
+        btQuaternion endRot = angularDisplacement * startRot;
+
```
hyperlogic: This is basically the same thing homing code, except for orientation.  Instead of using quaternions, I use 2D facing vectors, mostly to avoid inadvertently adding roll, due to accumulated precision errors.  The input to acosf() needs to be clamped to avoid nans if the vectors are not quite normalized.  The "currentRight" vector is used to determine if the deltaAngle is positive or negative.

libraries/physics/src/CharacterController.cpp
```diff
@@ -189,12 +190,39 @@ void CharacterController::playerStep(btCollisionWorld* dynaWorld, btScalar dt) {
...
+        // in order to accumulate displacement of avatar position, we need to take _shapeLocalOffset into account.
+        btVector3 shapeLocalOffset = glmToBullet(_shapeLocalOffset);
+        btVector3 swingDisplacement = rotateVector(endRot, -shapeLocalOffset) - rotateVector(startRot, -shapeLocalOffset);
+
+        _followLinearDisplacement = linearDisplacement + swingDisplacement + _followLinearDisplacement;
+        _followAngularDisplacement = angularDisplacement * _followAngularDisplacement;
```
hyperlogic: This part was tricky.  Because the accumulated displacements are applied to the avatar position, which is offset slightly from the rigidBody, we have to account for changes in translation caused by rotation of the rigidBody.  Once computed, this swingDisplacement is added into the rest of the linear displacements.

libraries/physics/src/CharacterController.cpp
```diff
@@ -323,8 +351,17 @@ void CharacterController::setTargetVelocity(const glm::vec3& velocity) {
     _walkVelocity = glmToBullet(velocity);
 }
 
-void CharacterController::setFollowVelocity(const glm::vec3& velocity) {
-    _followVelocity = glmToBullet(velocity);
+void CharacterController::setFollowParameters(const glm::mat4& desiredWorldBodyMatrix, float timeRemaining) {
+    _followTimeRemaining = timeRemaining;
+    _followDesiredBodyTransform = glmToBullet(desiredWorldBodyMatrix) * btTransform(btQuaternion::getIdentity(), glmToBullet(_shapeLocalOffset));
+}
```
hyperlogic: The _followDesiredBodyTransform is for the actual rigidBody, so we need to take the _shapeLocalOffset into account.

interface/src/avatar/MyAvatar.h
```diff
@@ -375,9 +374,19 @@ public slots:
     // used to transform any sensor into world space, including the _hmdSensorMat, or hand controllers.
     glm::mat4 _sensorToWorldMatrix;
 
-    glm::vec3 _followVelocity { Vectors::ZERO };
-    float _followSpeed { 0.0f };
-    float _followOffsetDistance { 0.0f };
+    struct FollowHelper {
+        glm::mat4 _desiredBodyMatrix;
+        float _timeRemaining { 0.0f };
+        float _timeTotal { 0.0f };
```
AndrewMeadows: **MyAvatar::_timeTotal** is a new data member but it is not used for anything.

interface/src/avatar/MyAvatar.cpp
AndrewMeadows: Hrm... I know **truncatedDistance** was the name I gave the variable, but it now occurs to me that it should be called **radialDistance** to be more correct and to make the code more readable.

libraries/physics/src/CharacterController.cpp
```diff
@@ -189,12 +190,39 @@ void CharacterController::playerStep(btCollisionWorld* dynaWorld, btScalar dt) {
...
-    // Rather than add _followVelocity to the velocity of the RigidBody, we explicitly teleport
-    // the RigidBody forward according to the formula: distance = rate * time
-    if (_followVelocity.length2() > 0.0f) {
+    // Dynamicaly compute a follow velocity to move this body toward the _followDesiredBodyTransform.
+    // Rather then add this velocity to velocity the RigidBody, we explicitly teleport the RigidBody towards its goal.
+    // This mirrors the computation done in MyAvatar::FollowHelper::postPhysicsUpdate().
+    // These two computations must be kept in sync.
+    const float MAX_DISPLACEMENT = 0.5f * _radius;
+    _followTimeRemaining -= dt;
+    if (_followTimeRemaining >= 0.005f) {
```
AndrewMeadows: There is a magic number here: 0.005f =  "minimum follow time remaining".

libraries/physics/src/CharacterController.cpp
```diff
@@ -189,12 +190,39 @@ void CharacterController::playerStep(btCollisionWorld* dynaWorld, btScalar dt) {
...
+        btVector3 endPos = startPos + linearDisplacement;
+
+        btQuaternion startRot = bodyTransform.getRotation();
+        glm::vec2 currentFacing = getFacingDir2D(bulletToGLM(startRot));
+        glm::vec2 currentRight(currentFacing.y, -currentFacing.x);
+        glm::vec2 desiredFacing = getFacingDir2D(bulletToGLM(_followDesiredBodyTransform.getRotation()));
+        float deltaAngle = acosf(glm::clamp(glm::dot(currentFacing, desiredFacing), -1.0f, 1.0f));
+        float angularVel = deltaAngle / _followTimeRemaining;
+        float sign = copysignf(1.0f, glm::dot(desiredFacing, currentRight));
+        btQuaternion angularDisplacement = btQuaternion(btVector3(0.0f, 1.0f, 0.0f), sign * angularVel * dt);
```
AndrewMeadows: Nitpick: **angularVel** is misnamed: should be **angularSpeed**.

cozza13:
test this please

libraries/shared/src/GLMHelpers.cpp
```diff
@@ -425,3 +425,23 @@ void generateBasisVectors(const glm::vec3& primaryAxis, const glm::vec3& seconda
...
+    } else {
+        return glm::normalize(facing2D);
+    }
+}
+
+glm::vec2 getFacingDir2D(const glm::mat4& m) {
+    glm::vec3 facing3D = transformVector(m, Vectors::UNIT_NEG_Z);
+    glm::vec2 facing2D(facing3D.x, facing3D.z);
+    if (glm::length(facing2D) <= 0.0001f) {
+        return glm::vec2(1, 0);
```
AndrewMeadows: Here and above: float literals should have the form: **1.0f** rather than **1**.

You've got a magic number:  0.001f = "too small", and for such slop you don't need the <= operator, you can get by with just <.

hifi-gustavo:
Build console output for commit: 61b760038a13a20cb2dcd7d522c906dd5f2e8887 available at: https://gist.github.com/3c70b17cd5f2dcf6507b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6895.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6895.dmg

hifi-gustavo:
Build console output for commit: 8bfa80d5fc196061be8f01070faca51e5ae30fb8 available at: https://gist.github.com/3c70b17cd5f2dcf6507b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6895.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6895.dmg

PR #6897, opened 2016-01-22
Fix asset server not creating resources directory before migrating

huffman:


birarda:
This got jammed creating a package on windows - merging.

PR #6899, opened 2016-01-22
repairs to index path behaviour

birarda:
This makes "Go Home" from the Server Console go to the index path. 
To handle this change for new users with the default content set, the domain-server now returns a default index path of "/0,0,0/0,0,0,1". 
The Interface client was not previously going to the index path if it went to a network address and not a place name - that is now fixed.

### DIFF
--- a/domain-server/resources/describe-settings.json
+++ b/domain-server/resources/describe-settings.json
@@ -65,7 +65,7 @@
             {
               "name": "viewpoint",
               "label": "Viewpoint",
-              "placeholder": "/512,512,512"
+              "placeholder": "/0,0,0"
             }
           ]
         }
--- a/domain-server/src/DomainServer.cpp
+++ b/domain-server/src/DomainServer.cpp
@@ -1872,16 +1872,27 @@ void DomainServer::processPathQueryPacket(QSharedPointer<ReceivedMessage> messag
 
         const QString PATHS_SETTINGS_KEYPATH_FORMAT = "%1.%2";
         const QString PATH_VIEWPOINT_KEY = "viewpoint";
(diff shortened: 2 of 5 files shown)


hifi-gustavo:
Build console output for commit: 16981fce7df93ec0b30ff0a8db1b9f91f9421177 available at: https://gist.github.com/91e25c858dbd015b06b9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6899.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6899.dmg

hifi-gustavo:
Build console output for commit: 9061e33a1f8746d06732bcf3cca26bfe94350254 available at: https://gist.github.com/91e25c858dbd015b06b9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6899.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6899.dmg

ZappoMan:
This isn't working as I expected. If my servers "index" is set to something like `/10,0,10` then when I "Go Home" I would expect to end up there. Instead it seems to always send me to `/0,0,0` - this is almost as bad as sending me to the old random location.

ZappoMan:
Actually - never mind that last comment, now it seems to be working. I think what happened was that when I changed the index value in the domain page settings it didn't save it. Now it seems to be working.

PR #6914, opened 2016-01-23
Close all console popup on first confirmation

Atlante45:
This fixes the problem on windows where you get many popups if you click quit several times.

### DIFF
--- a/server-console/src/main.js
+++ b/server-console/src/main.js
@@ -105,46 +105,49 @@ const ipcMain = electron.ipcMain;
 var isShuttingDown = false;
 function shutdown() {
     if (!isShuttingDown) {
-        var idx = 0;
-
         // if the home server is running, show a prompt before quit to ask if the user is sure
         if (homeServer.state == ProcessGroupStates.STARTED) {
-            idx = dialog.showMessageBox({
+            dialog.showMessageBox({
                 type: 'question',
                 buttons: ['Yes', 'No'],
                 title: 'Are you sure?',
                 message: 'Quitting will stop your Server Console and your Home domain will no longer be running.'
-            });
+            }, shutdownCallback);
+        } else {
+            shutdownCallback(0);
         }
 
-        if (idx == 0) {
-            isShuttingDown = true;
+    }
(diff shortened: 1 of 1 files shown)


server-console/src/main.js
```diff
@@ -103,18 +103,21 @@ const ipcMain = electron.ipcMain;
 
 
 var isShuttingDown = false;
+var shutDownPopupOpen = false;
 function shutdown() {
-    if (!isShuttingDown) {
+    if (!isShuttingDown && !shutDownPopupOpen) {
```
ZappoMan: is there anyway to make the "other" question raise to the top? I fear this could get confusing if you've somehow moved the quit question behind other windows, and then you choose quit and it appears to do nothing.

hifi-gustavo:
Build console output for commit: 234918e52f3bbdbd7757e4e80601c278dbbe48cb available at: https://gist.github.com/0a80a17c15704c9d813d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6914.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6914.dmg

server-console/src/main.js
Atlante45: No, we can't get a handle to the actually message box unless we make a full-blown window.

But @birarda just suggested, still raising the box because the user might have lost it,
but not making you answer them all. Answering any of them will close everything.

hifi-gustavo:
Build console output for commit: d159f8dc38b4f6db0b5368cee470fc37f4dd9cc7 available at: https://gist.github.com/0a80a17c15704c9d813d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6914.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6914.dmg

PR #6917, opened 2016-01-23
fix crash for empty modelUrl in ModelEntityItem

AndrewMeadows:
There was a recipe for crashing the interface:

(1) Create a model entity.  For example you can use this FBX file: http://s3.amazonaws.com/hifi-public/models/props/Dice/goldDie.fbx
(2) Select new entity with edit.js
(3) Set the "Model URL" to empty string
(4) Deselect the model (click on the window somewhere)
(5) Exit edit mode (toggle entity.js off)
(6) Re-enter edit mode (toggle entity.js on)
(7) Click on entity again --> crash

The problem was that the model was being deleted out from under the scene and when it was selected again the model was assumed to be complete with valid geometry.

### DIFF
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -480,12 +480,17 @@ Model* RenderableModelEntityItem::getModel(EntityTreeRenderer* renderer) {
         } else { // we already have the model we want...
             result = _model;
         }
-    } else { // if our desired URL is empty, we may need to delete our existing model
-        if (_model) {
-            _myRenderer->releaseModel(_model);
-            result = _model = NULL;
-            _needsInitialSimulation = true;
-        }
+    } else if (_model) {
+        // remove from scene
+        render::ScenePointer scene = AbstractViewStateInterface::instance()->getMain3DScene();
+        render::PendingChanges pendingChanges;
+        _model->removeFromScene(scene, pendingChanges);
+        scene->enqueuePendingChanges(pendingChanges);
+
+        // release interest
+        _myRenderer->releaseModel(_model);
+        result = _model = NULL;
+        _needsInitialSimulation = true;
     }
 
     return result;


libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -480,12 +480,17 @@ Model* RenderableModelEntityItem::getModel(EntityTreeRenderer* renderer) {
...
+        render::ScenePointer scene = AbstractViewStateInterface::instance()->getMain3DScene();
+        render::PendingChanges pendingChanges;
+        _model->removeFromScene(scene, pendingChanges);
+        scene->enqueuePendingChanges(pendingChanges);
+
+        // release interest
+        _myRenderer->releaseModel(_model);
+        result = _model = NULL;
+        _needsInitialSimulation = true;
     }
```
AndrewMeadows: These changes prevent the crash but the question for @zzmp is whether this is the exact right way to do it?  Should we remove from the scene immediately like this, or should the removal be queued up and dealt with later?

hifi-gustavo:
Build console output for commit: d25f54485ce21d5037fbc8e96354a89947e66d6a available at: https://gist.github.com/819571c00676acc1bec0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6917.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6917.dmg

OmegaHeron:
Windows 10 - Release build 4039 - followed steps above - crashed as expected.  Repeated exactly with this PR, no crash.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
zzmp: If we don't remove from the scene before the next engine run, it will try to render with a model you've just released, so this looks good by me. I'll double-check and merge it in.

PR #6923, opened 2016-01-25
Single assignment-client process fix for: `-n 1`

thoys:
This should fix `assignment-client -n 1` not to create an assignment-client monitor but just a single assignment-client class instead.

Test-Notes:
- Run a domain-server
- Spawn an assignment-client with param `-n 1` to have a single fork
- Verify that only one assignment-client process is running


hifi-gustavo:

birarda:
@thoys - this isn't how we want to handle the `-n 1` parameter.

The `-n` parameter, while poorly named, explicitly specifies that you want an assignment client monitor that starts with a certain number of child assignment-client processes.

We need to rename the `-n` parameter to something that makes this behaviour clearer.

PR #6936, opened 2016-01-26
CharacterController: Improve ground support detection

hyperlogic:
In addition to the existing line probe ground check, also check if the bottom sphere of the capsule is contact with any geometry before going into hover mode.  This should prevent going into the fly animation when standing or walking on collision shapes with small gaps between elements.


hifi-gustavo:
Build console output for commit: bc9621baabdce000adb77181cd22c45b13fd670c available at: https://gist.github.com/eaaaed3e4cb56f1e536f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6936.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6936.dmg

PR #6938, opened 2016-01-26
reduce hips swing driven by head orientation when wearing HMD

AndrewMeadows:
This PR fixes a bug where the avatar hips would move too much when the HMD rotates.  The problem was primarily caused a bug in the IK algorithm (we weren't rotating the end-effector (Head) during the IK solution).  I also tuned a few things for more correct motion.

### DIFF
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -153,7 +153,6 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
                 continue;
             }
 
-            // cache tip absolute transform
             int tipIndex = target.getIndex();
             int pivotIndex = _skeleton->getParentIndex(tipIndex);
             if (pivotIndex == -1 || pivotIndex == _hipsIndex) {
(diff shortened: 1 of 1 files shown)


libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -165,12 +164,30 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
...
+                // enforce tip's constraint
+                RotationConstraint* constraint = getConstraint(tipIndex);
+                if (constraint) {
+                    glm::quat tipRelativeRotation = glm::normalize(tipOrientation * glm::inverse(tipParentOrientation));
+                    bool constrained = constraint->apply(tipRelativeRotation);
+                    if (constrained) {
+                        tipOrientation = glm::normalize(tipRelativeRotation * tipParentOrientation);
+                    }
+                }
+            }
```
AndrewMeadows: For the **HmdHead** IKTarget type we now make sure to rotate the end effector first as much as possible.  The rest of the skeleton chain (neck and spine) will only rotate after the head's constraint is maxed out.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -227,9 +244,9 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
                 } else if (targetType == IKTarget::Type::HmdHead) {
                     // An HmdHead target slaves the orientation of the end-effector by distributing rotation
                     // deltas up the hierarchy.  Its target position is enforced later by shifting the hips.
-                    deltaRotation = target.getRotation() * glm::inverse(tipRotation);
+                    deltaRotation = target.getRotation() * glm::inverse(tipOrientation);
                     float dotSign = copysignf(1.0f, deltaRotation.w);
-                    const float ANGLE_DISTRIBUTION_FACTOR = 0.15f;
+                    const float ANGLE_DISTRIBUTION_FACTOR = 0.35f;
```
AndrewMeadows: Minor tweak here.  Increasing this number toward 1.0 reduces the distribution of the remaining rotation (once the head's constraint kicks in) down the spine --> less influence from the head on farther joints since the IK will solve faster when moving the first few joints.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -641,11 +657,11 @@ void AnimInverseKinematics::initConstraints() {
...
             SwingTwistConstraint* stConstraint = new SwingTwistConstraint();
             stConstraint->setReferenceRotation(_defaultRelativePoses[i].rot);
-            const float MAX_HEAD_TWIST = PI / 4.0f;
+            const float MAX_HEAD_TWIST = PI / 8.0f;
             stConstraint->setTwistLimits(-MAX_HEAD_TWIST, MAX_HEAD_TWIST);
 
             std::vector<float> minDots;
-            const float MAX_HEAD_SWING = PI / 4.0f;
+            const float MAX_HEAD_SWING = PI / 6.0f;
             minDots.push_back(cosf(MAX_HEAD_SWING));
```
AndrewMeadows: I thought I was going to have to significantly modify these constraints but in the end they remain simple conical constraints with modified angles for twist and swing.

hifi-gustavo:
Build console output for commit: 2728290ffbeb6e95ed9db0343a80a42a69d755f8 available at: https://gist.github.com/5f0d90d5e914c3677833<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6938.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6938.dmg

judasshuffle:
With the dk2 on and the hydras my avatar jumps into the distance resetting with @ brings it back 
like this
https://i.gyazo.com/6b5a780b0e4006d339a3c243d340d6e7.gif
without the dk2 activated it works fine

AndrewMeadows:
I confirm that the **judasboots** avatar goes unstable after these changes.  I'm looking into it.

AndrewMeadows:
Hrm... the **judasboots** avatar is unstable for me on hifi/master build 4047 (from yesterday 2016.01.25).  I don't think the problem is new to this PR ,but I'll dig a bit more before I  re-evaluate whether to unblock this PR.

AndrewMeadows:
I haven't figured out why **judasboots** is unstable, but it is definitely not because of this PR.  I get the same behavior when I test with the tip of hifi/master.  I don't want to block this PR any longer so I'll file an internal bug and we'll fix it later.

PR #6944, opened 2016-01-26
Animate iff visible

howard-stearns:
Don't stop animating while an avatar is still visible.
https://app.asana.com/0/32622044445063/75547783717817

We now animate using the same test as for rendering LOD visibility.

### DIFF
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -121,7 +121,7 @@ void Avatar::init() {
     getHead()->init();
     _skeletonModel.init();
     _initialized = true;
-    _shouldRenderBillboard = (getLODDistance() >= BILLBOARD_LOD_DISTANCE);
+    _shouldRenderBillboard = false;
 }
 
 glm::vec3 Avatar::getChestPosition() const {
@@ -141,7 +141,8 @@ glm::quat Avatar::getWorldAlignedOrientation () const {
 }
 
 AABox Avatar::getBounds() const {
-    return AABox();
+    // Our skeleton models are rigged, and this method call safely produces the static bounds of the model.
+    return _skeletonModel.getPartBounds(0, 0, getPosition(), getOrientation());
 }
 
 float Avatar::getLODDistance() const {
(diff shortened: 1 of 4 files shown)


interface/src/avatar/Avatar.cpp
```diff
@@ -121,7 +121,7 @@ void Avatar::init() {
     getHead()->init();
     _skeletonModel.init();
     _initialized = true;
-    _shouldRenderBillboard = (getLODDistance() >= BILLBOARD_LOD_DISTANCE);
+    _shouldRenderBillboard = false;
```
ZappoMan: can't we just remove this completely?

hifi-gustavo:
Build console output for commit: 1d49c8f36868b5b7cbad03e20b5cbd0250b457b7 available at: https://gist.github.com/4230bf46b4936b70aae4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6944.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6944.dmg

interface/src/avatar/Avatar.cpp
howard-stearns: Sure, and change initialization, too. (I opted for fewest change lines.)
I'll make both changes and retest.

hifi-gustavo:
Build console output for commit: e145cf60bea205751b896c72e114014ad7d575b6 available at: https://gist.github.com/4230bf46b4936b70aae4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6944.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6944.dmg

interface/src/avatar/Avatar.h
```diff
@@ -255,7 +255,7 @@ public slots:
 private:
     bool _initialized;
     NetworkTexturePointer _billboardTexture;
-    bool _shouldRenderBillboard;
+    bool _shouldRenderBillboard { true };
```
ZappoMan: I'm still confused here... it seems to me that we don't handle billboards at all... I guess I'm saying we should remove all reference to billboards... I might be wrong, but I think we don't ever generate them... so we can't render them

howard-stearns: The existing code goes right up to the point of rendering the billboard, but then bails early from Avatar::renderBillboard.

I've left all of that in place, on the grounds that we're not supposed to be redesigning things -- just fixing the bugs.
I have no objection to the former, but I'm interpreting the tactical-direction as us being essentially in code-freeze (where possible) until we get through the bugs.

cozza13:
I will QA test

howard-stearns:
DNM'ing this after discussion with Brad. We're going to remove the obsolete code.

PR #6951, opened 2016-01-27
fix for animations stopping after domain restart

ZappoMan:


libraries/animation/src/AnimationLoop.cpp
```diff
@@ -18,7 +18,7 @@ const float AnimationLoop::MAXIMUM_POSSIBLE_FRAME = 100000.0f;
 
 AnimationLoop::AnimationLoop() :
     _fps(30.0f),
-    _loop(false),
+    _loop(true),
```
ZappoMan: AnimationLoop and AnimationProperties had a mismatch in the default value of _loop, and as such, the properties being applied to the loop would check to see if the value changed (true default -> true desired) and if not change, it wouldn't set the value in the loop... so the _loop in AnimationLoop would be false and the server would tell clients it wasn't looping.

A long term fix would be to consolidate AnimationLoop and AnimationProperties... since now the only user of AnimationLoop is also a user of AnimationProperties... we used to have multiple other users of AnimationLoop, but now it's only the models.

ZappoMan:
Note: this fixes this bug report - https://app.asana.com/0/26320649018758/66264103181939

Test Case:
- create a model entity with an animation.
```
        {
            "animation": {
                "currentFrame": 0,
                "lastFrame": 30,
                "running": 1,
                "loop":
                "url": "http://public.highfidelity.io/cozza13/club/dragon/flying.fbx"
            },
            "dimensions": {
                "x": 0.89771169424057007,
                "y": 0.22868512570858002,
                "z": 0.62803679704666138
            },
            "modelURL": "http://public.highfidelity.io/cozza13/club/dragon/dragon.fbx",
            "position": {
                "x": 0,
                "y": 0,
                "z": 0
            },
        }
```

- verify that it saves on your entity server
- restart entity server
- leave and return to domain
- verify that the model is still animating... << **The old code failed this step, and model would not be animated**

hifi-gustavo:
Build console output for commit: f178fc789c089465ce67cdfd3b2e68c119b7bb23 available at: https://gist.github.com/aa1e8a3079e5094158c0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6951.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6951.dmg

hyperlogic:
:+1:

PR #6955, opened 2016-01-27
apply "hips offset" to root of state graph

AndrewMeadows:
An alpha user tested #6938 and reported an instability when using a particular avatar.  This PR fixes the problem.

The avatar in question was modified using Blender and had picked up some joints above the "Hips" in the hierarchy and had some non-trivial parent-relative transforms at the root joint and the hips.  This revealed the fact that we were adding the _hipsOffset to the "Hips" joint directly (and hence in the wrong frame) rather than to the root of the entire animation state.

### DIFF
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -246,7 +246,7 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
                     // deltas up the hierarchy.  Its target position is enforced later by shifting the hips.
                     deltaRotation = target.getRotation() * glm::inverse(tipOrientation);
                     float dotSign = copysignf(1.0f, deltaRotation.w);
-                    const float ANGLE_DISTRIBUTION_FACTOR = 0.35f;
+                    const float ANGLE_DISTRIBUTION_FACTOR = 0.45f;
                     deltaRotation = glm::normalize(glm::lerp(glm::quat(), dotSign * deltaRotation, ANGLE_DISTRIBUTION_FACTOR));
                 }
 
@@ -634,7 +634,7 @@ void AnimInverseKinematics::initConstraints() {
         } else if (baseName.startsWith("Spine", Qt::CaseInsensitive)) {
             SwingTwistConstraint* stConstraint = new SwingTwistConstraint();
             stConstraint->setReferenceRotation(_defaultRelativePoses[i].rot);
-            const float MAX_SPINE_TWIST = PI / 8.0f;
+            const float MAX_SPINE_TWIST = PI / 12.0f;
             stConstraint->setTwistLimits(-MAX_SPINE_TWIST, MAX_SPINE_TWIST);
 
             std::vector<float> minDots;
@@ -658,11 +658,11 @@ void AnimInverseKinematics::initConstraints() {
         } else if (0 == baseName.compare("Neck", Qt::CaseInsensitive)) {
             SwingTwistConstraint* stConstraint = new SwingTwistConstraint();
             stConstraint->setReferenceRotation(_defaultRelativePoses[i].rot);
-            const float MAX_NECK_TWIST = PI / 6.0f;
+            const float MAX_NECK_TWIST = PI / 9.0f;
             stConstraint->setTwistLimits(-MAX_NECK_TWIST, MAX_NECK_TWIST);
 
             std::vector<float> minDots;
-            const float MAX_NECK_SWING = PI / 4.0f;
+            const float MAX_NECK_SWING = PI / 8.0f;
             minDots.push_back(cosf(MAX_NECK_SWING));
             stConstraint->setSwingLimits(minDots);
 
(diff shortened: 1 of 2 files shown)


libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -376,13 +376,13 @@ const AnimPoseVec& AnimInverseKinematics::overlay(const AnimVariantMap& animVars
...
         } else {
-            // shift the hips according to the offset from the previous frame
+            // shift the everything according to the _hipsOffset from the previous frame
             float offsetLength = glm::length(_hipsOffset);
             const float MIN_HIPS_OFFSET_LENGTH = 0.03f;
             if (offsetLength > MIN_HIPS_OFFSET_LENGTH) {
                 // but only if offset is long enough
                 float scaleFactor = ((offsetLength - MIN_HIPS_OFFSET_LENGTH) / offsetLength);
-                _relativePoses[_hipsIndex].trans = underPoses[_hipsIndex].trans + scaleFactor * _hipsOffset;
+                _relativePoses[_rootIndex].trans = underPoses[_rootIndex].trans + scaleFactor * _hipsOffset;
```
AndrewMeadows: We apply the measured offset to the root rather than the hips.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -775,9 +775,18 @@ void AnimInverseKinematics::setSkeletonInternal(AnimSkeleton::ConstPointer skele
...
         _headIndex = _skeleton->nameToJointIndex("Head");
         _hipsIndex = _skeleton->nameToJointIndex("Hips");
+
+        // walk up the hierarchy until we find the root and cache its index
+        _rootIndex = _hipsIndex;
+        int parentIndex = _skeleton->getParentIndex(_hipsIndex);
+        while (parentIndex != -1) {
+            _rootIndex = parentIndex;
+            parentIndex = _skeleton->getParentIndex(_rootIndex);
+        }
```
AndrewMeadows: Cache the rootIndex for later so we don't have to constantly recompute it.

hifi-gustavo:
Build console output for commit: e00d899640679ac9920ea5d44687ab71b0451b99 available at: https://gist.github.com/64723da8c6eb1a59c350<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6955.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6955.dmg

libraries/animation/src/AnimInverseKinematics.cpp
AndrewMeadows: BTW, @hyperlogic disagrees with this.  He says that the offset should be transformed into the frame of the hips parent and applied to the hips translation.

hyperlogic:
I prefer to see this offset applied to the hips instead, doing what ever transform we have to do to get it into the right space.  My main concern is that if we shift the root, we could be inadvertently altering the the translation of ALL children of the root, not just the hips.  One example, might be a statue avatar on a pedestal.  If the hips are a child of the pedestal, or the pedestal and the hips share the same parent.  This offset would move BOTH the hips and the pedestal, when I think we would only want to move the hips.

My 2 cents.

judasshuffle:
Hi I just tested with with my original problem avatar and it now seems to be working correctly which is brilliant, thanks for that. I figured out that the problem with with my avatar was not all the elements of it were rotated x90, blender maya workflow weirdness.

AndrewMeadows:
@hyperlogic, I now apply the **_hipsOffset** at the correct joint.  I also took the liberty to tweak the head, neck and spine constraints to help distribute the rotation down to lower joints.

hifi-gustavo:
Build console output for commit: 1b45f5cc0c21f6997bd3e87aa86e67c4ef38908e available at: https://gist.github.com/64723da8c6eb1a59c350<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6955.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6955.dmg

judasshuffle:
Tested again -with my original problem avatar all seems to work great still :)

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -376,13 +376,26 @@ const AnimPoseVec& AnimInverseKinematics::overlay(const AnimVariantMap& animVars
...
+                    // the hips are the root so _hipsOffset is in the correct frame
+                    _relativePoses[_hipsIndex].trans = underPoses[_hipsIndex].trans + scaleFactor * _hipsOffset;
+                } else {
+                    // the hips are NOT the root so we need to transform _hipsOffset into hips local-frame
+                    glm::quat hipsFrameRotation = _relativePoses[_hipsParentIndex].rot;
+                    int index = _skeleton->getParentIndex(_hipsParentIndex);
+                    while (index != -1) {
+                        hipsFrameRotation *= _relativePoses[index].rot;
+                        index = _skeleton->getParentIndex(index);
+                    }
```
hyperlogic: I think you can replace the loop with the following:

    glm::quat hipsFrameRotation = _skeleton->getAbsolutePose(_relativePoses, _hipsParentIndex).rot;

Admittedly, it's not as efficient.  But I think this is only called once per frame, so it's probably ok.

AndrewMeadows: Yes, I considered this, but since I only need **rotation** I decided to reduce the readability in exchange for more efficiency.

Meanwhile, in my next project I think I'm going to change how/where **_hipsOffset** is fed back into the poses in an effort to reduce how the hydra hands influence the hips, and this logic here will either be moved or completely rewritten.  I'll consider these tradeoffs again soon.

PR #6960, opened 2016-01-27
don't adjust a time of '0 usecs since epoch' with clock skew in edit packets

sethalves:
- fix a bug that was causing some entity edit packets to be ignored by server

### DIFF
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -763,7 +763,7 @@ void EntityItem::adjustEditPacketForClockSkew(QByteArray& buffer, int clockSkew)
     // lastEdited
     quint64 lastEditedInLocalTime;
     memcpy(&lastEditedInLocalTime, dataAt, sizeof(lastEditedInLocalTime));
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: f2ea73d646808084f98060c6770319b75e0a4795 available at: https://gist.github.com/2f5291a25062da905e69<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6960.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6960.dmg

nbq:
Testing this now

nbq:
Wasn't really much that could be tested on this other than to make sure editing worked and it saved it, stopped server, restarted and item was where I edited it to be.

ZappoMan:
I'm not entirely sure when or why we'd have last edited times of 0, but this seems like a benign enough change. I'm merging it... and I will kick myself later if this uncovers a bug. :)

PR #6963, opened 2016-01-27
Make all 2D overlays QML-based

jherico:
This addresses the [bug](https://app.asana.com/0/27650181942747/83769475464639) in overlays that causes the toolbar background.  To do so it does several things...

* Migrate all the Image overlay (used for toolbar buttons) to QML, where the text overlay already is
* Add a new 'rectangle overlay' type, which provides a simple rectangle without the need to use an empty 'text' field for the same purpose
* Update the toolBar.js library to use the new rectangle type

Additionally this code refactors the text overlay to use a much simpler design, pushing most of the logic into the QML code.  The new overlay based image and rectangle use the same design.

# Testing

In order to fully test this PR, you should load the edit.js that's included in this code, as the toolbar changes will not be visible otherwise

https://rawgit.com/jherico/hifi/qmlOverlays/examples/edit.js

# Notes
The code also modifies the overlay support somewhat.  Because the main thread can sometimes block on the overlay creating script thread, it's not possible to safely fetch QML properties, because such an action would have to block on the main thread, causing the potential for both threads to be blocked on the other.  However, fetching overlay properties is not critical, because the overlay creating code can always remember what properties it previously set.  

In this case, there were only two locations where overlay properties were being fetched.

The first is the overlay manager which uses them to create JS wrapper classes around the overlays.    However, while the wrapper classes for 3D overlays get used quite a bit, I found no instances of use of the Image and Text overlay wrappers, so they've been removed.  

The second case is in the toolbar code which used the ability to fetch the alpha during a mouse hover so that it could be restored once the mouse hover hand ended.  However, it was simple to modify the toolbar to simply remember the alpha that was last requested by the calling code and restore from that rather than using getProperties.   To that end, the `Overlays`  class has been modified to check if a given overlay supports property fetching before it tried to fetch a property, removing the only case where the overlay logic could block while waiting on the main thread.  

In general I'm unhappy that the main thread can ever block on a scripting thread as I feel it should always be the other way around.  However, that's a bigger task for another time.


examples/libraries/overlayManager.js
```diff
@@ -236,18 +236,10 @@
         "dimensions"
     ]);
 
-    ImageOverlay = generateOverlayClass(Overlay2D, "image", [
```
ZappoMan: why did you remove these?

jherico: I've updated the PR notes.  They explain this.

hifi-gustavo:
Build console output for commit: c1f0db29b74d77164ed32c6c2931a05a9f275d4c available at: https://gist.github.com/80769597255e90e20d0b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6963.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6963.dmg

ZappoMan:
test this please

hifi-gustavo:
Build console output for commit: 3fd1b11b4aaa1ab16225299c1f789eb518e3d448 available at: https://gist.github.com/80769597255e90e20d0b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6963.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6963.dmg

OmegaHeron:
Tested Win 10 - seems to work as expected using test @jherico outlines.  Only thing notes is text size in Users Online display changes with this.  Looks roughly 2x size in release build.

cozza13:
@jherico , was the 2x size intentional?

hifi-gustavo:
Build console output for commit: 16cdff67e3000fc5b5def56470759be436d66495 available at: https://gist.github.com/80769597255e90e20d0b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6963.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6963.dmg

jherico:
@cozza13 increased size was not intentional.  I will investigate.

jherico:
@cozza13 text overlays weren't properly parsing the font object which sets the size, nor the color (which was coming as 0-255 color values, rather than 0-1).  Fixed in the latest commit.

hifi-gustavo:
Build console output for commit: 8c19b95ea22867a69883e0c9466df668e5edfcad available at: https://gist.github.com/80769597255e90e20d0b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6963.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6963.dmg

PR #6965, opened 2016-01-28
Warn user if OpenGL version is too low (< 4.1)

ctrlaltdavid:
A message box is displayed at start-up prior to the Interface window appearing.

### DIFF
--- a/interface/src/OpenGLInfo.cpp
+++ b/interface/src/OpenGLInfo.cpp
@@ -1,0 +1,60 @@
+//
+//  OpenGLInfo.cpp
+//  interface/src
+//
+//  Created by David Rowe on 28 Jan 2016.
+//  Copyright 2016 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include "OpenGLInfo.h"
+
+#include <QMessageBox>
+#include <QRegularExpression>
+
+#include "GLCanvas.h"
+
+OpenGLInfo::OpenGLInfo(int& argc, char** argv) :
+    QApplication(argc, argv)
+{
+}
+
+bool OpenGLInfo::isValidVersion() {
+    bool valid = true;
+
+    // Retrieve OpenGL version
+    GLCanvas* glCanvas = new GLCanvas();
+    glCanvas->initializeGL();
+    QString glVersion = QString((const char*)glGetString(GL_VERSION));
+    delete glCanvas;
+
+    // Compare against minimum
+    // The GL_VERSION string begins with a version number in one of these forms:
+    // - major_number.minor_number 
+    // - major_number.minor_number.release_number
+    // Reference: https://www.opengl.org/sdk/docs/man/docbook4/xhtml/glGetString.xml
+    const int MINIMUM_OPENGL_MAJOR_VERSION = 4;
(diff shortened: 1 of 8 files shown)


hifi-gustavo:
Build console output for commit: af3d842e6096c3ecd9da8ea2964b5882852cb79e available at: https://gist.github.com/74095939854f8c7316f6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6965.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6965.dmg

ctrlaltdavid:
Screen shot from running PR build on my HD4000-based tablet:
opengl-version-too-low

judasshuffle:
Works fine on my normal pc, no warning on my surface pro2 which I expected to fail(cos hifi doesn't work on it .gl view says it has 4.3 so it would pass your check :)

OmegaHeron:
No warning on PC with NV GTX660M or Mac Mini with HD4000 - looked around and I don't have anything it should throw warn on so all I can say is no false positive.

ctrlaltdavid:
@judasshuffle In theory, Interface should work on your Surface Pro 2 'cos it has a good OpenGL version; I believe it doesn't work on Surface Pro 3s either (cursor distorted in login screen and can't log in etc.); I'll add "not working on Surface Pro 2 either" to the Asana bug report.

judasshuffle:
@ctrlaltdavid  happens on the surface pro 3 https://alphas.highfidelity.io/t/render-weirdness-on-surface-pro-3-win-10/9944 also

jherico:
Could we move the rename the OpenGLInfo class to something like OpenGLVersionChecker?  It's current name is kind of misleading.  Also, I'd prefer if we could put it into the `gl` library, rather than in `interface`.   It would work if you make it depend on GLWidget or GLWindow rather than GLCanvas.

interface/src/OpenGLInfo.cpp
```diff
@@ -0,0 +1,60 @@
...
+    QString glVersion = QString((const char*)glGetString(GL_VERSION));
+    delete glCanvas;
+
+    // Compare against minimum
+    // The GL_VERSION string begins with a version number in one of these forms:
+    // - major_number.minor_number 
+    // - major_number.minor_number.release_number
+    // Reference: https://www.opengl.org/sdk/docs/man/docbook4/xhtml/glGetString.xml
+    const int MINIMUM_OPENGL_MAJOR_VERSION = 4;
+    const int MINIMUM_OPENGL_MINOR_VERSION = 1;
```
jherico: The version you look for should reflect the version specified in `gl/Config.h`.  Can you either update this code to use the `GPU_INPUT_PROFILE` value, or alternatively push these minimum values into `gl/Config.h` as `#define` that match the `GPU_INPUT_PROFILE`?  or possibly construct the `GPU_INPUT_PROFILE` from these values...

sorry, micromanaging....  just fix it however. :)

hifi-gustavo:
Build console output for commit: 6232a83f43e5e46beb2bd0d4467bef831bf635bd available at: https://gist.github.com/74095939854f8c7316f6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6965.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6965.dmg

PR #6988, opened 2016-01-29
Fix marketplace imports failing silently if edit.js not running

huffman:


huffman:
How to test:
Make sure edit.js isn't running
Open marketplace
Try to import an item
Confirm that you get a warning that edit.js needs to be running

huffman:
test this please

huffman:
looked like a gist post error

hifi-gustavo:
Build console output for commit: acf96a401f8aeb8cd05270562f41e72fbf9b90c7 available at: https://gist.github.com/b996a3c9876250196c48<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6988.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6988.dmg

imgntn:
@huffman if i try the 'candle light' i still get a message about 'Would You Like to Run This Script" (and you can run it), which I suppose doesn't make sense if I can't import the entity because of edit.js  -- BUT the modal does appear for me

nbq:
Testing this now

nbq:
I would say I get the same as @imgntn - It just tries to run the js script to load the item from the marketplace and it stays running in the running scripts list. This does not prompt me for edit.js or error about it.

According to the testing parameters this seems to work exactly as it should but the problem is that the import script is dumped into running scripts. I would still say this is a PASS though.

PR #6990, opened 2016-01-30
Better head joint setup

howard-stearns:
Fix "Mini-mirror mode doesn't look at  face of avatar."


howard-stearns:
https://app.asana.com/0/32622044445063/79058431420111

The fix here does three things:
1) Look harder for the head joint. This avatar was using an unusual name. (I have _not_ rewritten the fbx parser to look more methodically for joints.)
2) Log a warning if it isn't found, so that we have a better chance of recognizing when things go wrong.
3) It turns out that we were skipping some initialization when we couldn't find a head, which left us with garbage data. But there's really no reason to skip that.

hifi-gustavo:
Build console output for commit: ceed571c82267d3ba8056a42dff2458f91cb36d8 available at: https://gist.github.com/608c9c601ae367ff06a9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6990.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6990.dmg

libraries/fbx/src/FBXReader.cpp
```diff
@@ -600,7 +600,7 @@ FBXGeometry* FBXReader::extractFBXGeometry(const QVariantHash& mapping, const QS
                     } else if (name == jointLeanName) {
                         jointLeanID = getID(object.properties);
 
-                    } else if (name == jointHeadName) {
+                    } else if (name == jointHeadName || name == "head" || name == "Head" || name == "HEAD" || name == "joint_head") {
```
hyperlogic: This might be problematic. We have avatars with head mesh named "Head".  In that case, this code might mis-identify the head joint.

hyperlogic:
I think we should keep #2 and #3, but remove #1. Because it might break some of our avatars in a much worse manner then meerly having in incorrect camera offset.

howard-stearns:
Good point about mesh names. Maybe @hyperlogic  and @ozanserim can get together and decide on our naming conventions. It seems like we shouldn't fail to recognize an avatar whose head joint is named "Head", which is not only obvious and the case in the ticket, but is also the name our "Avatar Standards and Reference" tells people to use.  I think we want the code and the doc to match.

murillodigital:
test this please

hifi-gustavo:
Build console output for commit: ceed571c82267d3ba8056a42dff2458f91cb36d8 available at: https://gist.github.com/608c9c601ae367ff06a9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6990.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6990.dmg

howard-stearns:
After discussion with @hyperlogic , we're going to not do #1, above. (See separate task https://app.asana.com/0/30233891560551/86025700206151).  Leaving #2 and #3 are enough to handle the immediate "lady cop" ticket.

hifi-gustavo:
Build console output for commit: 24f53492f570e76bf213115a78efa7ef7be086d3 available at: https://gist.github.com/608c9c601ae367ff06a9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6990.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR6990.dmg

PR #7003, opened 2016-02-01
automatically save adjustments to offsets of equipped items

sethalves:
- automatically save adjustments to offsets of equipped items

### DIFF
--- a/examples/attachedEntitiesManager.js
+++ b/examples/attachedEntitiesManager.js
@@ -40,7 +40,7 @@ if (SHOW_TOOL_BAR) {
     var saveButton = toolBar.addOverlay("image", {
         width: BUTTON_SIZE,
         height: BUTTON_SIZE,
-        imageURL: "http://headache.hungry.com/~seth/hifi/save.png",
+        imageURL: ".../save.png",
         color: {
             red: 255,
             green: 255,
@@ -51,7 +51,7 @@ if (SHOW_TOOL_BAR) {
     var loadButton = toolBar.addOverlay("image", {
         width: BUTTON_SIZE,
         height: BUTTON_SIZE,
-        imageURL: "http://headache.hungry.com/~seth/hifi/load.png",
+        imageURL: ".../load.png",
         color: {
             red: 255,
             green: 255,
--- a/examples/attachedEntitiesManager.js
+++ b/examples/attachedEntitiesManager.js
@@ -116,6 +116,7 @@ function AttachedEntitiesManager() {
             manager.checkIfWearable(parsedMessage.grabbedEntity, parsedMessage.joint)
             // manager.saveAttachedEntities();
         } else if (parsedMessage.action === 'shared-release') {
+            manager.updateRelativeOffsets(parsedMessage.grabbedEntity);
             // manager.saveAttachedEntities();
         } else if (parsedMessage.action === 'equip') {
             // manager.saveAttachedEntities();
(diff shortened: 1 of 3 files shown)


sethalves:
to test: run the handControllerGrab.js from this PR as well as attachmentManager.js.  equip something in one hand and adjust its position with the other.  drop the entity and then re-equip it and see that the position snaps back to where it was before it was dropped.

hifi-gustavo:
Build console output for commit: e4ef3ae8697f8431c8f4191a43eda3b3b2f0d289 available at: https://gist.github.com/e78491b53059f8be0143<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7003.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7003.dmg

cozza13:
-default spot for the gun, block etc is off the back end of the wrist.
-can not do the adjustments with the bow and arrow.
- object does not collide with an object when it is picked up. e.g. hold a block and try and put another one on top
- I have had situations where the gun is not going back to the same spot.

It is a nice feeling to adjust

sethalves:
The updated entities in demo should have fixed offsets.  We wont be able to adjust the bow, because it disables the other hand when equipped.

To test wearing things, put something like:

{"grabbableKey":{"grabbable":true},"wearable":{"joints":{"Head":[]}}}

in the userData of an entity.  Pick it up and drop it near your head.

nbq:
When I pick something up with the Hydras, I never get it to lock to a position, even on demo. I have to use the other hand to sort of move it into place a bit then it locks into the spot. That is one of the problems I have with grabbing and the Hydras is that nothing I pick up ever locks into place even if I pick it up with the hand its supposed to lock to

hifi-gustavo:
Build console output for commit: 30e97c2a76bcdafa30d2f2cef9866917de680fc6 available at: https://gist.github.com/e78491b53059f8be0143<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7003.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7003.dmg

cozza13:
I have tested this and the feature is working. You do need to be running both scripts.

sethalves:
I added attachedEntitiesManager.js to default scripts.

examples/controllers/handControllerGrab.js
```diff
@@ -1217,10 +1217,12 @@ function MyController(hand) {
...
-        if (handJointName in allowedJoints) {
-            return true;
+        if ("joints" in wearableData) {
+            var allowedJoints = wearableData.joints;
+            var handJointName = this.hand === RIGHT_HAND ? "RightHand" : "LeftHand";
+            if (handJointName in allowedJoints) {
+                return true;
+            }
         }
     }
```
AndrewMeadows: This function has no **return** statement outside the new **if** statements.  Maybe it should **return false** at the end?

examples/defaultScripts.js
```diff
@@ -21,3 +21,4 @@ Script.load("controllers/squeezeHands.js");
 Script.load("grab.js");
 Script.load("directory.js");
 Script.load("dialTone.js");
+Script.load("attachedEntitiesManager.js");
```
AndrewMeadows: If we keep  _attachedEntitiesManager.js_ on the list of default scripts then we will have to eventually cleanup the assets to NOT use links from @sethalves personal assets (e.g. http://headache.hungry.com/~seth/hifi/save.png).  So the question is:  Do we want to clean these up now or later?

examples/attachedEntitiesManager.js
```diff
@@ -22,41 +22,44 @@ var MINIMUM_DROP_DISTANCE_FROM_JOINT = 0.4;
 var ATTACHED_ENTITY_SEARCH_DISTANCE = 10.0;
 var ATTACHED_ENTITIES_SETTINGS_KEY = "ATTACHED_ENTITIES";
 var DRESSING_ROOM_DISTANCE = 2.0;
+var SHOW_TOOL_BAR = false;
```
AndrewMeadows: **SHOW_TOOL_BAR** is hardcoded as **false**.  Does this mean the toolbar is a 100% debug thing?

sethalves: Yes, for now.  The script wont have a UI once it's finished.

examples/defaultScripts.js
sethalves: I've removed the hungry.com bit.

examples/controllers/handControllerGrab.js
sethalves: hm... I suppose javascript was accidently doing the right thing, here.  fixed

hifi-gustavo:
Build console output for commit: ad2a7bfc3c4048bae461a0de08244ab0108c92fc available at: https://gist.github.com/e78491b53059f8be0143<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7003.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7003.dmg

PR #7012, opened 2016-02-02
base palm position on finger joint rather than an offset from hand joint

sethalves:
- base palm position on finger joint rather than an offset from hand joint
- double size of near-grab sphere

### DIFF
--- a/examples/controllers/handControllerGrab.js
+++ b/examples/controllers/handControllerGrab.js
@@ -69,13 +69,14 @@ var PICK_MAX_DISTANCE = 500; // max length of pick-ray
 // near grabbing
 //
 
-var GRAB_RADIUS = 0.03; // if the ray misses but an object is this close, it will still be selected
+var GRAB_RADIUS = 0.06; // if the ray misses but an object is this close, it will still be selected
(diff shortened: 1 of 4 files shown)


sethalves:
To test -- grab small things with handControllerGrab.js.  See if the small things need to be in front of the avatar's hand to be grabbed.

hifi-gustavo:
Build console output for commit: 1e69c12e87c1eb20d38cf5cdb80200e252bfb889 available at: https://gist.github.com/36d70148472a31ba7758<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7012.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7012.dmg

nbq:
Testing this one

nbq:
I have tested and I say it PASSED. I checked and small things do need to be in front of the hand to be picked up, I tested with multiple small objects. I also put two small items side by side and then used the hand to pick it up, each time it only picked up the object my hand was facing.

PR #6999, opened 2016-02-01
Migrate stencil and overlay jobs to use plumbers

zzmp:
- Replace DrawOverlay3D::getOpaquePipeline with a ShapePlumber.
- Move RenderDeferredTask shader initializations to a separate file for readability.

This refactoring leads up to a fix (not yet in PR) for overlay transparency and layering.

PR #6998, opened 2016-02-01
Avoid setting the pipeline within GeometryCache draw calls

zzmp:
Instead, have consumers of GeometryCache draw calls set their own pipelines, so that the pipeline codepath is easier to follow/change.

PR #7013, opened 2016-02-02
Fix text3d layering and transparency

zzmp:
- Add shapeGetShapeKey call to Overlay base class to prep for fixing all overlay layering/transparency.
- Adds pipelines to DrawOverlay3D to correctly layer.
- Split DrawOverlay3D into opaque and transparent jobs.
- Reimplements text3d using new pipelines.

This lays out the framework to fix layering and transparency for all overlays, but only actually fixes it for text3d. Some overlays will need a CULL flag in the ShapeKey to correctly render within this framework, and so are left for a future PR. Any overlay using its own pipeline (i.e. all besides text3d) will not layer or render transparency correctly.

NOTE: This PR depends on #6998 and #6999. It should only include 3a7a8c4 and on.


hifi-gustavo:
Build console output for commit: bdc20103312cdcbd7763534c00a19e02a6404071 available at: https://gist.github.com/c9b29afe151b1251ba0d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7013.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7013.dmg

zzmp:
@samcake: as we've built ShapeKey, those shapes using it are expected not to alter the batch's pipeline, such that multiple shapes with the same key can be rendered by the same pipeline with minimal GPU reprogramming.
Text3DOverlay breaks this expectation by calling a TextRenderer from within its render call, which sets the pipeline for its text. I'm not sure how to fix this (because render order matters here): for opaques we can just render the last text, but for transparents I think we need to render everything.
Thoughts?

zzmp:
Will remake this PR more appropriately, as dependent commits have changed a lot.

PR #7017, opened 2016-02-02
MyAvatar: Improved Jump / InAir / Fly behavior 

hyperlogic:
There are new animation sets for jump takeoff and in air.

Currently, there is only a single first-pass jump takeoff animation.  There will be a left foot version and a stationary jump version added later.  Also, there is only a single set of 3 first-pass in-air poses, which work well for forward jumps but not so well for stationary jumps, a set of stationary in-air poses will be added later.  At the moment, there is no landing animation support.  This will be added in a future PR.

The physics character controller behavior has been changed:
  * There is a new takeoff state, in between when the jump button has been pressed and the upward jump velocity has been applied.
  * The transition from InAir to Flying has been refined.  It requires a 750 ms jump hold, or a double tap.
  * The transition from Hover to InAir/Ground has been refined.  You won't get pulled into the ground if you are flying fast.

The character controller has been changed to have a single internal state, and preStep and preSimulate has been cleaned up.

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -1314,7 +1314,6 @@ void MyAvatar::preRender(RenderArgs* renderArgs) {
 const float RENDER_HEAD_CUTOFF_DISTANCE = 0.5f;
 
 bool MyAvatar::cameraInsideHead() const {
-    const Head* head = getHead();
     const glm::vec3 cameraPosition = qApp->getCamera()->getPosition();
     return glm::length(cameraPosition - getDefaultEyePosition()) < (RENDER_HEAD_CUTOFF_DISTANCE * getUniformScale());
 }
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -1311,21 +1311,23 @@ void MyAvatar::preRender(RenderArgs* renderArgs) {
     _prevShouldDrawHead = shouldDrawHead;
 }
 
-const float RENDER_HEAD_CUTOFF_DISTANCE = 0.50f;
+const float RENDER_HEAD_CUTOFF_DISTANCE = 0.5f;
 
 bool MyAvatar::cameraInsideHead() const {
     const Head* head = getHead();
     const glm::vec3 cameraPosition = qApp->getCamera()->getPosition();
-    return glm::length(cameraPosition - head->getEyePosition()) < (RENDER_HEAD_CUTOFF_DISTANCE * getUniformScale());
+    return glm::length(cameraPosition - getDefaultEyePosition()) < (RENDER_HEAD_CUTOFF_DISTANCE * getUniformScale());
 }
 
 bool MyAvatar::shouldRenderHead(const RenderArgs* renderArgs) const {
-    return ((renderArgs->_renderMode != RenderArgs::DEFAULT_RENDER_MODE) ||
-            (qApp->getCamera()->getMode() != CAMERA_MODE_FIRST_PERSON) ||
-            !cameraInsideHead());
+    bool defaultMode = renderArgs->_renderMode == RenderArgs::DEFAULT_RENDER_MODE;
+    bool firstPerson = qApp->getCamera()->getMode() == CAMERA_MODE_FIRST_PERSON;
+    bool insideHead = cameraInsideHead();
+    return !defaultMode || !firstPerson || !insideHead;
 }
 
 void MyAvatar::updateOrientation(float deltaTime) {
+
     //  Smoothly rotate body with arrow keys
     float targetSpeed = _driveKeys[YAW] * _yawSpeed;
     if (targetSpeed != 0.0f) {
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -1510,7 +1510,8 @@ void MyAvatar::updatePosition(float deltaTime) {
     // rotate velocity into camera frame
     glm::quat rotation = getHead()->getCameraOrientation();
     glm::vec3 localVelocity = glm::inverse(rotation) * _targetVelocity;
-    glm::vec3 newLocalVelocity = applyKeyboardMotor(deltaTime, localVelocity, isHovering());
+    bool isHovering = _characterController.getState() == CharacterController::State::Hover;
+    glm::vec3 newLocalVelocity = applyKeyboardMotor(deltaTime, localVelocity, isHovering);
     newLocalVelocity = applyScriptedMotor(deltaTime, newLocalVelocity);
 
     // rotate back into world-frame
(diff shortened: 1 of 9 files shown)


interface/src/avatar/MyAvatar.cpp
```diff
@@ -1311,21 +1311,23 @@ void MyAvatar::preRender(RenderArgs* renderArgs) {
...
 }
 
-const float RENDER_HEAD_CUTOFF_DISTANCE = 0.50f;
+const float RENDER_HEAD_CUTOFF_DISTANCE = 0.5f;
 
 bool MyAvatar::cameraInsideHead() const {
     const Head* head = getHead();
     const glm::vec3 cameraPosition = qApp->getCamera()->getPosition();
-    return glm::length(cameraPosition - head->getEyePosition()) < (RENDER_HEAD_CUTOFF_DISTANCE * getUniformScale());
+    return glm::length(cameraPosition - getDefaultEyePosition()) < (RENDER_HEAD_CUTOFF_DISTANCE * getUniformScale());
```
hyperlogic: This is a bug fix for the avatar's head sometimes being visible when flying very fast.

libraries/animation/src/Rig.cpp
```diff
@@ -706,15 +737,61 @@ void Rig::computeMotionAnimationState(float deltaTime, const glm::vec3& worldPos
...
+            _animVars.set("isNotFlying", true);
+            _animVars.set("isTakeoff", false);
+            _animVars.set("isNotTakeoff", true);
+            _animVars.set("isInAir", true);
+            _animVars.set("isNotInAir", false);
+
+            // compute blend based on velocity
+            const float JUMP_SPEED = 3.5f;
+            float alpha = glm::clamp(-worldVelocity.y / JUMP_SPEED, -1.0f, 1.0f) + 1.0f;
+            _animVars.set("inAirAlpha", alpha);
```
hyperlogic: This is how we compute an blend factor for the 3 in-air poses.

hifi-gustavo:
Build console output for commit: 656608e2ffaff0f21effbe461b8f36437addd6d1 available at: https://gist.github.com/66d223889fc6e7a52021<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7017.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7017.dmg

judasshuffle:
likes it, works well for me

hifi-gustavo:
Build console output for commit: 6cc3b2b47f82022ea00925887fa6ffbb9029e200 available at: https://gist.github.com/66d223889fc6e7a52021<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7017.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7017.dmg

interface/src/avatar/SkeletonModel.cpp
```diff
@@ -72,6 +72,20 @@ void SkeletonModel::initJointStates() {
...
+    case CharacterController::State::Ground:
+        return Rig::CharacterControllerState::Ground;
+    case CharacterController::State::Takeoff:
+        return Rig::CharacterControllerState::Takeoff;
+    case CharacterController::State::InAir:
+        return Rig::CharacterControllerState::InAir;
+    case CharacterController::State::Hover:
+        return Rig::CharacterControllerState::Hover;
+    };
+}
```
AndrewMeadows: Why have two distinct state enums and this translation method?

libraries/physics/src/CharacterController.cpp
```diff
@@ -153,72 +162,61 @@ void CharacterController::preStep(btCollisionWorld* collisionWorld) {
...
-            const btScalar HOVER_ACCELERATION_TIMESCALE = 0.1f;
-            btScalar tau = dt / HOVER_ACCELERATION_TIMESCALE;
-            _rigidBody->setLinearVelocity(actualVelocity - tau * (actualVelocity - desiredVelocity));
+    // decompose into horizontal and vertical components.
+    btVector3 actualVertVelocity = actualVelocity.dot(_currentUp) * _currentUp;
+    btVector3 actualHorizVelocity = actualVelocity - actualVertVelocity;
+    btVector3 desiredVertVelocity = desiredVelocity.dot(_currentUp) * _currentUp;
+    btVector3 desiredHorizVelocity = desiredVelocity - desiredVertVelocity;
+
+    btVector3 finalVelocity;
```
AndrewMeadows: **finalVelocity** relies on the default **btVector3** constructor which is not guaranteed to initialize the components of the vector (the SIMD implementation most certainly does NOT) and there is no **default** condition in the switch below, so it may be possible to end up using uninitialized values at line **214**.

libraries/physics/src/CharacterController.cpp
```diff
@@ -424,32 +437,74 @@ void CharacterController::preSimulation() {
...
+                SET_STATE(State::InAir, "near ground");
+            } else if (((_floorDistance < JUMP_PROXIMITY_THRESHOLD) || _hasSupport) && !flyingFast) {
+                SET_STATE(State::Ground, "touching ground");
+            }
+            break;
         }
     }
-    _followTime = 0.0f;
 
+    ///  _walkVelocity.dot(_currentUp) > UPWARD_VELOCITY_THRESHOLD
```
AndrewMeadows: I don't understand this comment.  Is it cruft?

interface/src/avatar/SkeletonModel.cpp
hyperlogic: Because Rig is in the animation library and CharacterController is in the physics library.  I suppose I could move the enum into the shared library, but I prefer to keep the enum in the same namespace as the class that uses it.  How strongly do you feel about this?

libraries/physics/src/CharacterController.cpp
hyperlogic: Yes, but all cases are covered, I can add a default case that asserts.  This will at least catch an error at run time if someone adds an item to the enum and forgets to add a case.

libraries/physics/src/CharacterController.cpp
hyperlogic: Yeah, I'll remove it.

hifi-gustavo:
Build console output for commit: fff603e4e2f5a48567f7fb390cebdcb0c07a019d available at: https://gist.github.com/66d223889fc6e7a52021<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7017.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7017.dmg

PR #7031, opened 2016-02-05
server can clear simulation ownership

AndrewMeadows:
- when a simulation owner stops sending updates the entity-server will clear simulation ownership (but will not zero velocities) and will transmit this change
- when no simulation owners volunteer to own a moving entity the entity-server will eventually zero its velocities and will transmit this change


libraries/entities/src/SimpleEntitySimulation.cpp
```diff
@@ -11,38 +11,58 @@
...
             }
-        } else {
-            ++itemItr;
+            entity->markAsChangedOnServer();
+            // dirty all the tree elements that contain the entity
+            DirtyOctreeElementOperator op(entity->getElement());
+            getEntityTree()->recurseTreeWithOperator(&op);
+        } else if (expiry < _nextSimulationExpiry) {
+            _nextSimulationExpiry = expiry;
         }
```
AndrewMeadows: Above is where the entity-server either: (a) removes simulation ownership or (b) zeros the velocity of an un-owned moving object.

In either case it uses the **DirtyOctreeElementOperator** to flag the relevant **OctreeElements** that something has changed.

libraries/entities/src/SimulationOwner.cpp
```diff
@@ -48,11 +48,6 @@ void SimulationOwner::clear() {
 
 void SimulationOwner::setPriority(quint8 priority) {
     _priority = priority;
-    if (_priority == 0) {
-        // when priority is zero we clear everything
-        _expiry = 0;
-        _id = QUuid();
-    }
 }
```
AndrewMeadows: This "hidden optimization" that would clear the **SimulationOwner::_id** whenever the **_priority** changed to zero was preventing some other code from detecting when the ownership had changed, so I removed it.

libraries/octree/src/DirtyOctreeElementOperator.h
```diff
@@ -0,0 +1,30 @@
...
+    ~DirtyOctreeElementOperator() {}
+
+    virtual bool preRecursion(OctreeElementPointer element);
+    virtual bool postRecursion(OctreeElementPointer element);
+private:
+    glm::vec3 _point;
+    OctreeElementPointer _element;
+};
+
+#endif // hifi_DirtyOctreeElementOperator_h
```
AndrewMeadows: Why make this new **DirtyOctreeElementOperator** and not use the existing **UpdateEntityOperator**?

The answer is that the **UpdateEntityOperator** is way more complicated and calculation intensive than we need.  It is used to apply an arbitrary change to an **EntityItem**'s properties whereas in our case we know the specific changes to be made and really just need to update the tree itself.

hifi-gustavo:
Build console output for commit: 2da46ff26a16d8f812d12745735c121856c5e1f8 available at: https://gist.github.com/83977a2ab39a9be4c3a3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7031.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7031.dmg

libraries/physics/src/EntityMotionState.cpp
```diff
@@ -61,13 +56,16 @@ EntityMotionState::EntityMotionState(btCollisionShape* shape, EntityItemPointer
...
+    _nextOwnershipBid(0),
     _measuredDeltaTime(0.0f),
+    _lastMeasureStep(0),
+    _lastStep(0),
+    _loopsWithoutOwner(0),
     _accelerationNearlyGravityCount(0),
-    _nextOwnershipBid(0),
-    _loopsWithoutOwner(0)
+    _numInactiveUpdates(1),
+    _outgoingPriority(ZERO_SIMULATION_PRIORITY)
```
AndrewMeadows: I sorted the data members of the **EntityMotionState** class, and so I had to also sort them here in the constructor.

libraries/physics/src/EntityMotionState.cpp
```diff
@@ -102,27 +100,35 @@ bool EntityMotionState::handleEasyChanges(uint32_t& flags) {
...
+                _body->setActivationState(WANTS_DEACTIVATION);
+                _outgoingPriority = ZERO_SIMULATION_PRIORITY;
+                _loopsWithoutOwner = 0;
+            } else {
+                // unowned object is still moving --> we should volunteer to own it
+                // TODO? put a delay in here proportional to distance from object?
+                setOutgoingPriority(VOLUNTEER_SIMULATION_PRIORITY);
+                _loopsWithoutOwner = LOOPS_FOR_SIMULATION_ORPHAN;
+                _nextOwnershipBid = 0;
+            }
```
AndrewMeadows: We had special case code that would hint to Bullet to deactivate objects whose ownership had been released.  We now expect _moving_ objects to also be releases so we had to differentiate between these two cases.

libraries/physics/src/EntityMotionState.cpp
```diff
@@ -266,16 +271,21 @@ bool EntityMotionState::remoteSimulationOutOfSync(uint32_t simulationStep) {
...
 
-    const float INACTIVE_UPDATE_PERIOD = 0.5f;
-    if (_sentInactive) {
+    if (_numInactiveUpdates > 0) {
+        const uint8_t MAX_NUM_INACTIVE_UPDATES = 3;
+        if (_numInactiveUpdates > MAX_NUM_INACTIVE_UPDATES) {
+            // clear local ownership (stop sending updates) and let the server clear itself
+            _entity->clearSimulationOwnership();
+            return false;
+        }
```
AndrewMeadows: An important fix is to eventually stop bothering to try to release ownership of an object -- we can now rely on the server to eventually release our ownership if we stop sending updates.

AndrewMeadows:
HOW TO TEST:

(1) login with two interfaces simultaneously, UserA and UserB
(2) enable viewing of physics ownership on both interfaces (Developers --> Physics --> Highlight Simulation Ownership)
(3) create a dynamic Box with weak gravity (something like 0.5 m/sec^)
(4) UserA uses edit.js to move the Box up into the air so that it falls (UserA will own the simulation)
(5) Logout UserA ASAP
(6) From UserB's perspective the Box should continue to fall and will eventually lose ownership, at which point UserB's interface will bid for ownership
(7) The end result being that the Box should be owned by UserB before it hits the ground.
(8) Swing UserB's view frustum about... the Box should remain visible on the ground.
(9) Login UserA again once --> UserA should see the Box where it stopped on the ground.

(Note that the current behavior in hifi/master diverges around step (6)... from UserB's perspective the Box never loses UserA's ownership.  Even though UserB sees the object collide and come to rest on the ground the entity-server's notion of where the object is will be quite different --> when UserB swings their view frustum about (step (8)) the Box will disappear... because the entity-server will send an update with it at a different position.

libraries/octree/src/DirtyOctreeElementOperator.h
ZappoMan: Did you determine this with inspection or with performance tests?

AndrewMeadows: By inspection:

UpdateEntityOperator constructor takes four arguments, is 33 lines (sans comments) and creates more than 100 bytes of data members.
UpdateEntityOperator::preRecursion() ~= 88 lines of code with four **if** branches
UpdateEntityOperator::postRecursion() ~= 12 lines with two **if** branches
UpdateEntityOperator::possiblyCreateChildAt() ~= 10 lines with one **if** branch

DirtyOctreeElementOperator constructor takes one argument and creates 20 bytes of data members.
DirtyOctreeElementOperator::preRecursion() ~= 4 lines with one **if** branch
DirtyOctreeElementOperator::postRecursion() = 2 lines with no **if** branch
DirtyOctreeElementOperator::possiblyCreateChildAt() = one line

nbq:
I will test this one

nbq:
I have tested both the scenario in both the published interface and this PR. I can affirm that in this PR, once you start the drop on UserA, then log out, UserB sees the drop sort of stagger then pick up and continue falling once UserB has taken ownership.

Now on the published interface, doing the same thing, once UserA logs out, UserB sees the box stop in mid air and freeze. It does nothing until UserB edits it somehow to make it fall again.

I would say this PR passes.

libraries/entities/src/SimpleEntitySimulation.cpp
```diff
@@ -11,38 +11,58 @@
...
+                // clear ownership and restart timer, giving nearby simulators time to volunteer
+                qCDebug(entities) << "auto-removing simulation owner " << entity->getSimulatorID();
+                entity->clearSimulationOwnership();
             }
-        } else {
-            ++itemItr;
+            entity->markAsChangedOnServer();
+            // dirty all the tree elements that contain the entity
+            DirtyOctreeElementOperator op(entity->getElement());
+            getEntityTree()->recurseTreeWithOperator(&op);
```
ZappoMan: are you confident that the tree has been locked at this point? you need to make sure the tree is locked for reading while you do this, so you know that the tree doesn't get changed on you while you recurse.

ZappoMan: I did a quick inspection of the code, and it appears this is only called from `EntityTree::update()` which does a `withWriteLock()` on the tree before calling into this. So I think we're good.

PR #7032, opened 2016-02-05
MyAvatar: Standing Takeoff and In-Air Animations

hyperlogic:
Now there are two sets of of jump takeoff and in-air animations.
* Run - Used when the character jumps or falls with a small forward velocity.
* Standing - Used when the character jumps or falls in-place or backward.

**CharacterController changes**

* increased takeoff duration to 250 ms
* increased takeoff to fly duration to 1100 ms
* added standing jump and in-air animations
* added 250 milisecond delay between ground and hover, to prevent going into hover when walking over cracks.
* take-off to in-air transitions now use the new snapshotPrev interp type for a smoother tweening.

**AnimGraph changes**

* AnimInverseKinematics: opened up UpLeg and Leg constraints to improve quality of takeoff and in-air animations.
* AnimStateMachine: added interpType parameter for a new type of interp "SnapshotPrev" which was helpful for smoothing out transitions into In-air blends.

### DIFF
--- a/libraries/physics/src/CharacterController.cpp
+++ b/libraries/physics/src/CharacterController.cpp
@@ -60,9 +60,10 @@ CharacterController::CharacterController() {
     _jumpSpeed = JUMP_SPEED;
     _state = State::Hover;
     _isPushingUp = false;
-    _jumpButtonDownStart = 0;
+    _rayHitStartTime = 0;
+    _takeoffToInAirStartTime = 0;
+    _jumpButtonDownStartTime = 0;
     _jumpButtonDownCount = 0;
-    _takeoffToInAirStart = 0;
     _followTime = 0.0f;
     _followLinearDisplacement = btVector3(0, 0, 0);
     _followAngularDisplacement = btQuaternion::getIdentity();
@@ -417,6 +418,8 @@ glm::vec3 CharacterController::getLinearVelocity() const {
 
 void CharacterController::preSimulation() {
     if (_enabled && _dynamicsWorld) {
+        quint64 now = usecTimestampNow();
+
         // slam body to where it is supposed to be
         _rigidBody->setWorldTransform(_characterBodyTransform);
         btVector3 velocity = _rigidBody->getLinearVelocity();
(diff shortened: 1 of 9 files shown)


hifi-gustavo:
Build console output for commit: 51189cfc503aea5d8646723ad13b971d30fe8728 available at: https://gist.github.com/e27fe80c8230891e9976<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7032.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7032.dmg

hifi-gustavo:
Build console output for commit: bfeace78f71b755970013ee1d5ed7336107ae47f available at: https://gist.github.com/e27fe80c8230891e9976<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7032.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7032.dmg

judasshuffle:
Looks good to me I jumped on and off things forward and backwards.

libraries/physics/src/CharacterController.cpp
```diff
@@ -462,19 +468,22 @@ void CharacterController::preSimulation() {
...
             break;
         case State::Takeoff:
-            if (!rayCallback.hasHit() && !_hasSupport) {
+            if (!rayHasHit && !_hasSupport) {
                 SET_STATE(State::Hover, "no ground");
-            } else if ((now - _takeoffToInAirStart) > TAKE_OFF_TO_IN_AIR_PERIOD) {
+            } else if ((now - _takeoffToInAirStartTime) > TAKE_OFF_TO_IN_AIR_PERIOD) {
                 SET_STATE(State::InAir, "takeoff done");
-                _takeoffToInAirStart = now + USECS_PER_SECOND * 86500.0f;
+                _takeoffToInAirStartTime = now + USECS_PER_SECOND * 86500.0f;
```
AndrewMeadows: I know the **86500.0f** factor is not new but it looks magic and it took me a while to come up with a theory about what is going on.  It looks to me like you're pushing **_takeoffToInAirStartTime** into the distant future (slightly more than 24 hours).  If I were trying to communicate to my fellow devs that I was trying to set **takeoffToInAirStartTime** to some unreasonable future time I would have set it to -1, which for unsigned 64-bit timestamps wraps around to sometime in late 2030's.

Also, there are some implicit casts going on here since a **quint64** is being multiplied by a **float** then added to another **quint64**.  Since the **float** is an integer value there is no need for it to be a float literal.

hyperlogic: Good catch, about the float -> quint64 cast.  I'm surprised that didn't cause a warning. 
Also, subtracting two unsigned numbers where the rhs is larger than the lhs has really weird behavior.  Turns out I don't need to reset the **_takeoffToInAirStartTime** when the transition to InAir is complete, because I'm guaranteed that it will be set in the Ground state.

I will push a fix.

hifi-gustavo:
Build console output for commit: f13e31c7fc34b3ef985aec622d1cfc3e538ae06c available at: https://gist.github.com/e27fe80c8230891e9976<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7032.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7032.dmg

PR #7033, opened 2016-02-05
MyAvatar: Use HMD rotation to rotate while moving as well

hyperlogic:
We recently added turn your head to turn while flying, this PR applies that same logic to walking.
This code is disabled in comfort-mode.

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -1361,36 +1361,18 @@ void MyAvatar::updateOrientation(float deltaTime) {
         totalBodyYaw += _driveKeys[STEP_YAW];
     }
 
-    // use head/HMD orientation to turn while flying
-    if (getCharacterController()->getState() == CharacterController::State::Hover) {
-
-        // This is the direction the user desires to fly in.
-        glm::vec3 desiredFacing = getHead()->getCameraOrientation() * Vectors::UNIT_Z;
-        desiredFacing.y = 0.0f;
-
-        // This is our reference frame, it is captured when the user begins to move.
-        glm::vec3 referenceFacing = transformVector(_sensorToWorldMatrix, _hoverReferenceCameraFacing);
-        referenceFacing.y = 0.0f;
-        referenceFacing = glm::normalize(referenceFacing);
-        glm::vec3 referenceRight(referenceFacing.z, 0.0f, -referenceFacing.x);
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: f5bbbd2e5e9e923349ffa98381e8be9fa3782acd available at: https://gist.github.com/8e8f95c05359b4ebc540<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7033.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7033.dmg

jherico:
Two comments...

Is there any way this could be implemented as a script rather than C++?  The interaction between the avatar transform, the camera transform and the head pose is probably going to be one of the greatest areas of experimentation, so if it's possible to make it something content developers and power users can experiment with via scripting, that would be ideal.

Second, even if we can't completely expose it via scripting without some measure of change to the C++, I would really like to see us stop piling functionality into some of the existing overburdened classes like Application and MyAvatar.  Could we pull some of this out into an abstraction meant to encapsulate and mediate the interactions between the head / avatar / camera transforms?  This would also improve testability.

hyperlogic:
@jherico 

> Is there any way this could be implemented as a script rather than C++?

Yes and no, it can be prototyped in script, but in order to get smooth non-jittery motion it is better to do this in C++.  Because our scripting system is running asynchronous from the main update thread, it is difficult to synchronize motion between the two.  Also, there is missing functionality in script, specifically access to the sensorToWorld matrix, so rather than start with script, I chose to use the C++.

> I would really like to see us stop piling functionality into some of the existing overburdened classes like Application and MyAvatar.

I would disagree, in this particular case, MyAvatar already is doing the yaw rotation via the drive keys.  So it was a natural place to put this particular functionality.

> Could we pull some of this out into an abstraction meant to encapsulate and mediate the interactions between the head / avatar / camera transforms?

Possibly, but I don't immediately see what the abstraction would be.  Unfortunately, the head is tightly coupled to the MyAvatar/Head/SkeletonModel/Rig and the camera is tightly coupled with Application/Display Plugins.  I'm not convinced that adding a third actor into this system would make things any simpler.

hyperlogic:
IMHO, this will make people sick.

PR #7036, opened 2016-02-05
Fix native text rendering by forcing positions to integer values

jherico:
I identified the issue where some text (typically anything with native text rendering) could look really bad.  It turns out that the current code can allow a window to end up with a size or position that is not an integer.  If the vertical position was at fractional then you ended up seeing letters with varying up and down baselines, while if the horizontal position was non-integer then narrow letters or letter pairs with tight kerning would render illegibly.

May also fix the double-click problem.

### DIFF
--- a/interface/resources/qml/hifi/dialogs/RunningScripts.qml
+++ b/interface/resources/qml/hifi/dialogs/RunningScripts.qml
@@ -3,9 +3,9 @@ import QtQuick.Controls 1.4
 import QtQuick.Dialogs 1.2 as OriginalDialogs
 import Qt.labs.settings 1.0
 
-import "../styles" as Hifi
-import "../controls" as HifiControls
-import "../windows"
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: ebf0bf394af46b8a42e34cf1b38810d8a994725c available at: https://gist.github.com/9f7c6c59257c5a726e7b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7036.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7036.dmg

OmegaHeron:
I tested this one since I have seen some odd text rendering - everything looks good with this, though it should be said there was no 100% case where client rendered text oddly reproduced.  What's the double click issue?

PR #7049, opened 2016-02-09
Anti-stutter work

jherico:
This is a first pass at reducing the amount of stutter caused when entering a large scene.  

The stutter appears to be caused by a number of factors.  

- overloading the GPU with an excessive number of simultaneous transfers of large amount of texture / vertex data
- overloading the CPU with an excessive number of parsing threads doing file IO

This PR attempts to address the latter component first, simply because it's easier.  It reduces the size of the global thread pool from the default of `QThread::idealThreadCount()` to `2`.  This means that if you have 16 textures/models to load to render a scene, only 2 of them will be concurrently being loaded from disk to CPU memory at a time.  

Previously it would allow all CPU cores to be consumed performing this task, which could starve the main thread and the present thread, preventing keyboard or mouse input from being handled, and causing the screen to simply freeze (regardless of threaded present).  

Because the number of concurrent file transfers is being limited, the GPU transfers are also partially serialized, reducing their impact as well.

As a result, when teleporting from a nearly empty domain to `hifi://demo/547.046,495.032,509.714/0,-0.29785,0,-0.954613`, the number of missed frames (caused by the main / rendering thread being blocked/starved) is reduced and the amount of complete freezes (caused by the present thread being starved) is almost completely eliminated.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -12,13 +12,13 @@
 #include "Application.h"
 
 #include <gl/Config.h>
-
 #include <glm/glm.hpp>
 #include <glm/gtx/component_wise.hpp>
 #include <glm/gtx/quaternion.hpp>
 #include <glm/gtx/vector_angle.hpp>
 #include <glm/gtc/type_ptr.hpp>
 
+#include <Qt>
 #include <QtCore/QDebug>
 #include <QtCore/QJsonDocument>
 #include <QtCore/QJsonObject>
@@ -27,6 +27,7 @@
 #include <QtCore/QTimer>
 #include <QtCore/QAbstractNativeEventFilter>
 #include <QtCore/QMimeData>
+#include <QtCore/QThreadPool>
 
 #include <QtGui/QScreen>
 #include <QtGui/QImage>
@@ -419,6 +420,14 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
     _maxOctreePPS(maxOctreePacketsPerSecond.get()),
     _lastFaceTrackerUpdate(0)
 {
+    // FIXME this may be excessivly conservative.  On the other hand 
+    // maybe I'm used to having an 8-core machine
+    // Perhaps find the ideal thread count  and subtract 2 or 3 
+    // (main thread, present thread, random OS load)
+    // More threads == faster concurrent loads, but also more concurrent 
+    // load on the GPU until we can serialize GPU transfers (off the main thread)
+    QThreadPool::globalInstance()->setMaxThreadCount(2);
+    thread()->setPriority(QThread::HighPriority);
     thread()->setObjectName("Main Thread");
 
     setInstance(this);
@@ -1287,6 +1296,9 @@ void Application::initializeUi() {
 }
 
 void Application::paintGL() {
+
+
+
     // paintGL uses a queued connection, so we can get messages from the queue even after we've quit
     // and the plugins have shutdown
     if (_aboutToQuit) {
@@ -4762,8 +4774,11 @@ void Application::updateDisplayMode() {
         foreach(auto displayPlugin, standard) {
             addDisplayPluginToMenu(displayPlugin, first);
             // This must be a queued connection to avoid a deadlock
-            QObject::connect(displayPlugin.get(), &DisplayPlugin::requestRender,
-                this, &Application::paintGL, Qt::QueuedConnection);
+            QObject::connect(displayPlugin.get(), &DisplayPlugin::requestRender, [=] {
+                postEvent(this, new LambdaEvent([=] {
+                    paintGL(); 
+                }), Qt::HighEventPriority);
+            });
 
             QObject::connect(displayPlugin.get(), &DisplayPlugin::recommendedFramebufferSizeChanged, [this](const QSize & size) {
                 resizeGL();
--- a/libraries/animation/src/AnimationCache.cpp
+++ b/libraries/animation/src/AnimationCache.cpp
@@ -51,6 +51,7 @@ void AnimationReader::run() {
     if (originalPriority == QThread::InheritPriority) {
         originalPriority = QThread::NormalPriority;
     }
+    QThread::currentThread()->setPriority(QThread::LowPriority);
     try {
         if (_data.isEmpty()) {
             throw QString("Reply is NULL ?!");
--- a/libraries/animation/src/AnimationCache.cpp
+++ b/libraries/animation/src/AnimationCache.cpp
@@ -47,6 +47,10 @@ AnimationReader::AnimationReader(const QUrl& url, const QByteArray& data) :
 }
 
 void AnimationReader::run() {
+    auto originalPriority = QThread::currentThread()->priority();
+    if (originalPriority == QThread::InheritPriority) {
+        originalPriority = QThread::NormalPriority;
+    }
     try {
         if (_data.isEmpty()) {
             throw QString("Reply is NULL ?!");
@@ -73,6 +77,7 @@ void AnimationReader::run() {
     } catch (const QString& error) {
         emit onError(299, error);
     }
+    QThread::currentThread()->setPriority(originalPriority);
 }
 
 bool Animation::isLoaded() const {
--- a/libraries/model-networking/src/model-networking/ModelCache.cpp
+++ b/libraries/model-networking/src/model-networking/ModelCache.cpp
@@ -50,6 +50,11 @@ GeometryReader::GeometryReader(const QUrl& url, const QByteArray& data, const QV
 }
 
 void GeometryReader::run() {
+    auto originalPriority = QThread::currentThread()->priority();
+    if (originalPriority == QThread::InheritPriority) {
+        originalPriority = QThread::NormalPriority;
+    }
+    QThread::currentThread()->setPriority(QThread::LowPriority);
     try {
         if (_data.isEmpty()) {
             throw QString("Reply is NULL ?!");
@@ -82,6 +87,8 @@ void GeometryReader::run() {
         qCDebug(modelnetworking) << "Error reading " << _url << ": " << error;
         emit onError(NetworkGeometry::ModelParseError, error);
     }
+
+    QThread::currentThread()->setPriority(originalPriority);
 }
 
 NetworkGeometry::NetworkGeometry(const QUrl& url, bool delayLoad, const QVariantHash& mapping, const QUrl& textureBaseUrl) :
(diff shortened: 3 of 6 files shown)


jherico:
This is a first pass for work on https://app.asana.com/0/27650181942747/86757725244686

hifi-gustavo:
Build console output for commit: 59838587cf62a400053f5d726da13ca60d5acfc4 available at: https://gist.github.com/6cb4840bbfe695e3b468<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7049.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7049.dmg

hifi-gustavo:
Build console output for commit: 7158a2978edad4f84c3e0dc5f072d988105bb40d available at: https://gist.github.com/6cb4840bbfe695e3b468<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7049.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7049.dmg

interface/src/Application.cpp
```diff
@@ -4762,8 +4774,11 @@ void Application::updateDisplayMode() {
         foreach(auto displayPlugin, standard) {
             addDisplayPluginToMenu(displayPlugin, first);
             // This must be a queued connection to avoid a deadlock
-            QObject::connect(displayPlugin.get(), &DisplayPlugin::requestRender,
-                this, &Application::paintGL, Qt::QueuedConnection);
+            QObject::connect(displayPlugin.get(), &DisplayPlugin::requestRender, [=] {
```
ZappoMan: Why change this from a Qt::QueuedConnection to a direct connection with a postEvent()? What is the effect of that change? Is it intended to address stuttering? How? Does it have no other side effects? How did you verify that?

interface/src/ui/ApplicationCompositor.cpp
```diff
@@ -286,6 +287,24 @@ void ApplicationCompositor::displayOverlayTextureHmd(RenderArgs* renderArgs, int
...
+        // with the window.  This means if it's being driven by the mouse and...
+        // * The mouse is not over the Interface window
+        // * The interface window is not active
+        // However, if we're using something else to drive the pointer (hand controllers)
+        // then we don't care where the mouse cursor is.  However, interacting with the 
+        // QML UI probably still won't work if the interface window is not active/focused
+        // because the QML offscreen rendering uses the actual window as a proxy for detecting
+        // focus.
+        
+        // Don't render the mouse cursor if we aren't the active window
```
ZappoMan: I'm not convinced this is the right thing to do. And we certainly shouldn't be doing it in this PR... this change obviously has nothing to do with stuttering, and violates the "fix reported bugs" rule. Please remove this from this PR. We can file a different bug about mouse cursor/focus and active window, and we can discuss the correct fix in that context. For example, I am convinced we probably need to change our application to FORCE always on top when in HMD mode (or something similar)... but we don't want to do that when we're in normal desktop mode.

The point is... no... this "fix" is not correct. This is a merge blocker.

libraries/animation/src/AnimationCache.cpp
```diff
@@ -47,6 +47,10 @@ AnimationReader::AnimationReader(const QUrl& url, const QByteArray& data) :
 }
 
 void AnimationReader::run() {
+    auto originalPriority = QThread::currentThread()->priority();
+    if (originalPriority == QThread::InheritPriority) {
+        originalPriority = QThread::NormalPriority;
+    }
```
ZappoMan: Did you miss a call to `QThread::currentThread()->setPriority(QThread::LowPriority);`? If you didn't intend to change the priority here, why are you storing the original priority?

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -50,6 +50,11 @@ GeometryReader::GeometryReader(const QUrl& url, const QByteArray& data, const QV
 }
 
 void GeometryReader::run() {
+    auto originalPriority = QThread::currentThread()->priority();
+    if (originalPriority == QThread::InheritPriority) {
```
ZappoMan: it seems as though this code can't/shouldn't ever actually execute. AFAICT... `priority()` will only return `QThread::InheritPriority` if the thread in question is not running... but if that was the case, then `setPriority(x);` also "does nothing". Also, since you're in the `run()` method we pretty much can assume your thread has started. So... I don't think we need to check here, and if anything we probably want to simply `assert(originalPriority !== QThread::InheritPriority)`

libraries/model/src/model/TextureMap.cpp
```diff
@@ -50,88 +50,55 @@ void TextureMap::setLightmapOffsetScale(float offset, float scale) {
 }
 
 
-
-
+// FIXME why is this in the model library?  Move to GPU or GPU_GL
 gpu::Texture* TextureUsage::create2DTextureFromImage(const QImage& srcImage, const std::string& srcImageName) {
     QImage image = srcImage;
- 
-    int imageArea = image.width() * image.height();
```
ZappoMan: why are you removing all this code? This doesn't seem to belong in this PR... if there's some other bug you're fixing the make this a different PR, with a clear asana bug report about what you're fixing.

Please remove this change.

ZappoMan: I now see you named the commit as "don't burn CPU cycles calculating average color we don't use"... it would make my review easier if you had included that comment in the PR so I knew to look for the change... and inline comment here would be useful.

That being said, it looks like you also changed the logic a bit. And/or there's multiple calls to `image = image.convertToFormat(QImage::Format_RGB888);` that likely can be consolidated.

interface/src/Application.cpp
jherico: Both versions of the function do the same thing, asynchronously execute `paintGL`.  However, the old version just puts the triggering signal onto a stack of signals to be processed.  The new version ensures that the asynchronous posted message will have higher priority that items already in the queue with normal priority.  

This should result in less chance of a frame being missed if the event queue is being saturated or the main thread is being starved of CPU usage.

interface/src/ui/ApplicationCompositor.cpp
jherico: well, OK, I'm just really tired of trying to click things in interface and they don't work, and I finally take off my headset and discover I've ordered 2 gross of imitation dutch wooden shoes from china.  I will rebase the commit out of the PR.

OmegaHeron:
Tested on Windows 7.  Compared to current production build this is an improvment - I used hifi://sandbox as my comparison point, teleporting from a low impact scene to it.  With this PR AV is (somewhat) responsive during loading, overall scene rez seems smoother.  With production code AV is totally unresponsive during load and scene builds in a chaotic/jarring fashion.  This seems like a decent start to better things.

libraries/model-networking/src/model-networking/ModelCache.cpp
jherico: Before I put in this clause I kept getting log messages that said `You cannot set thread priority to InheritPriority`

libraries/model/src/model/TextureMap.cpp
jherico: There were always multiple calls to `convertToFormat(QImage::Format_RGB888);`, however I think I can consolidate.

libraries/animation/src/AnimationCache.cpp
jherico: fixed

libraries/model/src/model/TextureMap.cpp
```diff
@@ -50,88 +50,55 @@ void TextureMap::setLightmapOffsetScale(float offset, float scale) {
 }
 
 
-
-
+// FIXME why is this in the model library?  Move to GPU or GPU_GL
```
samcake: THis actually belongs here not in GPU
THe goal is to take care of the different variations on how we expect the textures representing the material properties to be created and recombined
This is on the asset side hence here.

libraries/model/src/model/TextureMap.cpp
```diff
@@ -50,88 +50,55 @@ void TextureMap::setLightmapOffsetScale(float offset, float scale) {
...
 gpu::Texture* TextureUsage::create2DTextureFromImage(const QImage& srcImage, const std::string& srcImageName) {
     QImage image = srcImage;
- 
-    int imageArea = image.width() * image.height();
-
```
samcake: WE don't use it yet but the goal of that code is to differentiate between the mask alpha vs true transparency channel and classify the material accordingly (opaque/texel opaque /transparent)
We can probably consolidate the the 2 branches of the the if, but we need do need this to happen there

libraries/animation/src/AnimationCache.cpp
ZappoMan: you appear to not have yet pushed your fix... at least github doesn't have the change.

libraries/model/src/model/TextureMap.cpp
jherico: Maybe the format conversion code belongs here, but it seems like the the actual grunt work of taking a QImage and turning it into a gpu::Texture should be in GPU.

libraries/model/src/model/TextureMap.cpp
jherico: We can restore the code once we're actually using it, presumably with a better implementation than calling `pixel()` for every actual pixel in the image, which is the absolute worst performing way of doing this.  

Right now this code both poorly optimized *and* produces no output.  I'm not going to optimize it if it's not used, so... poof.

hifi-gustavo:
Build console output for commit: 7d99f9e72f574aee8bc88f86bd98f7dbafec3fbb available at: https://gist.github.com/6cb4840bbfe695e3b468<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7049.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7049.dmg

OmegaHeron:
Retested on Windows 7 with latest build - same eval as before, an improvement over current production code for loading a complex scene.

libraries/model-networking/src/model-networking/ModelCache.cpp
ZappoMan: https://www.youtube.com/watch?v=0tJGk4ofc18

PR #7050, opened 2016-02-09
Fix audio stats

Atlante45:
Fixes upstream audio stats not always being reported by the mixer.

### DIFF
--- a/libraries/networking/src/SequenceNumberStats.cpp
+++ b/libraries/networking/src/SequenceNumberStats.cpp
@@ -14,6 +14,10 @@
 
 #include <limits>
 
+float PacketStreamStats::getLostRate() const {
+    return (_expectedReceived == 0) ? 0.0f : (float)_lost / _expectedReceived;
+}
+
 SequenceNumberStats::SequenceNumberStats(int statsHistoryLength, bool canDetectOutOfSync)
     : _lastReceivedSequence(0),
     _missingSet(),
--- a/libraries/networking/src/SequenceNumberStats.cpp
+++ b/libraries/networking/src/SequenceNumberStats.cpp
@@ -15,7 +15,7 @@
 #include <limits>
 
 float PacketStreamStats::getLostRate() const {
-    return (_expectedReceived == 0) ? 0.0f : (float)_lost / _expectedReceived;
+    return (_expectedReceived == 0) ? 0.0f : (float)_lost / (float)_expectedReceived;
 }
 
 SequenceNumberStats::SequenceNumberStats(int statsHistoryLength, bool canDetectOutOfSync)
--- a/libraries/networking/src/SequenceNumberStats.h
+++ b/libraries/networking/src/SequenceNumberStats.h
@@ -43,7 +43,7 @@ public:
         return diff;
     }
 
-    float getLostRate() const { return _expectedReceived ? (float)_lost / _expectedReceived : 0; }
+    float getLostRate() const;
 
     quint32 _received;
     quint32 _unreasonable;
--- a/libraries/networking/src/SequenceNumberStats.h
+++ b/libraries/networking/src/SequenceNumberStats.h
@@ -43,7 +43,7 @@ public:
         return diff;
     }
 
-    float getLostRate() const { return (float)_lost / _expectedReceived; }
+    float getLostRate() const { return _expectedReceived ? (float)_lost / _expectedReceived : 0; }
 
     quint32 _received;
     quint32 _unreasonable;
--- a/interface/src/ui/AudioStatsDialog.cpp
+++ b/interface/src/ui/AudioStatsDialog.cpp
@@ -133,39 +133,45 @@ void AudioStatsDialog::renderStats() {
         
     double totalLatency = audioInputBufferLatency + inputRingBufferLatency + networkRoundtripLatency + mixerRingBufferLatency
         + outputRingBufferLatency + audioOutputBufferLatency;
-        
-    _audioMixerStats.push_back(QString("Audio input buffer: %1ms").arg(
-                                                                       QString::number(audioInputBufferLatency, 'f', 2)) + QString("   - avg msecs of samples read to the audio input buffer in last 10s"));
-    
-    _audioMixerStats.push_back(QString("Input ring buffer: %1ms").arg(
-                                                                      QString::number(inputRingBufferLatency, 'f', 2)) + QString("  - avg msecs of samples read to the input ring buffer in last 10s"));
-    _audioMixerStats.push_back(QString("Network to mixer: %1ms").arg(
-                                                                     QString::number((networkRoundtripLatency / 2.0), 'f', 2)) + QString("  - half of last ping value calculated by the node list"));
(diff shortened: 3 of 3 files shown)


hifi-gustavo:
Build console output for commit: ccadb5483a4fe84d4f06a680c45842728d3fb3d2 available at: https://gist.github.com/2eb9f91b3734728f4583<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7050.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7050.dmg

Atlante45:
https://app.asana.com/0/26551373415487/80470852444837

For me, the issue only ever happened with Sandbox, I did not manage to reproduce locally.
It appeared as though Sandbox's AM wasn't even sending the stats packet.

In the end, I was able to confirm that it wasn't calling `perSecondActions()` hence not updating some stats being sent to the domain and not sending stats to the clients at all.

This is because `if (now - _lastPerSecondCallbackTime > USECS_PER_SECOND)` always evaluated to false because `_lastPerSecondCallbackTime` is bigger `now`.
The problem is that `usecTimestampNow` is not a steady clock and will sometimes "go back in time".

This is a big problem, that I'm sure affects other part of the codebase, especially in places where we use 2 calls to mesure time.The first call could return something like 1,000,000 and the second 42 because in the mean time `usecTimestampNow` decided to change its epoch.

My personal feeling is that we should replace our current implementation of `usecTimestampNow()` with a steady one. But this is a big change that will need to be tested very well, so I didn't want to do it here.

I'm am not sure just how regularly `usecTimestampNow` changes its epoch, but I don't think it is that often.

assignment-client/src/audio/AudioMixer.cpp
```diff
@@ -751,10 +750,17 @@ void AudioMixer::broadcastMixes() {
             ++framesSinceCutoffEvent;
         }
 
-        quint64 now = usecTimestampNow();
-        if (now - _lastPerSecondCallbackTime > USECS_PER_SECOND) {
+        using namespace std::chrono;
+        static const seconds ONE_SECOND { 1 };
+
+        auto now = Clock::now();
```
ZappoMan: you should keep using `usecTimestampNow()` there's no benefit from mixing the timing paradigms, it will only introduce confusion later down the road.

assignment-client/src/audio/AudioMixer.h
```diff
@@ -101,7 +101,9 @@ private slots:
     static bool _printStreamStats;
     static bool _enableFilter;
 
-    quint64 _lastPerSecondCallbackTime;
+    using Clock = std::chrono::steady_clock;
```
ZappoMan: ibid -- we should fix the logic bug that caused the old code to fail. Swithcing to std::chrono is syntactic sugar which will only further confuse the code base. We ought not mix timing paradigms for something like this that has no obvious benefit from using `std::chrono` - **keep your changes minimal and to the point**.

libraries/networking/src/SequenceNumberStats.h
```diff
@@ -43,7 +43,7 @@ class PacketStreamStats {
         return diff;
     }
 
-    float getLostRate() const { return (float)_lost / _expectedReceived; }
+    float getLostRate() const { return _expectedReceived ? (float)_lost / _expectedReceived : 0; }
```
ZappoMan: coding standard says that boolean evaluation of number values (int, float, etc) should be compared against 0... 

> 3.2.5. Implicit test for 0 should not be used

assignment-client/src/audio/AudioMixer.cpp
Atlante45: I actually believe `usecTimestampNow` is what's causing the issue, hence why I removed it.

assignment-client/src/audio/AudioMixer.h
Atlante45: The logic is correct, but appears to fail anyway.
And I believe the reason is `usecTimeStampNow`, using the steady clock will allow me to confirm/deny that assumption by using an independent mechanism.

hifi-gustavo:
Build console output for commit: 82cb5bb78742d451b8285163e6908e06fdaa54bb available at: https://gist.github.com/2eb9f91b3734728f4583<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7050.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7050.dmg

Atlante45:
test this please

hifi-gustavo:
Build console output for commit: 82cb5bb78742d451b8285163e6908e06fdaa54bb available at: https://gist.github.com/2eb9f91b3734728f4583<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7050.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7050.dmg

Atlante45:
Let's hold on this for a little bit, I think I know why that test is falling.
`usecTimestampNow` is using a static QElapsedTimer, that's not protected against concurrent access.

hifi-gustavo:
Build console output for commit: f32562b191045d856e52846f16f6b9fb8850ab1d available at: https://gist.github.com/2eb9f91b3734728f4583<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7050.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7050.dmg

Atlante45:
@birarda fixed the core issue in https://github.com/highfidelity/hifi/pull/7074

libraries/networking/src/SequenceNumberStats.cpp
```diff
@@ -14,6 +14,10 @@
 
 #include <limits>
 
+float PacketStreamStats::getLostRate() const {
+    return (_expectedReceived == 0) ? 0.0f : (float)_lost / _expectedReceived;
```
AndrewMeadows: **_expectedReceived** is implicitly cast to a float here.

hifi-gustavo:
Build console output for commit: 51168a8a29aea86c2bcbe7405b477af2a259d597 available at: https://gist.github.com/2eb9f91b3734728f4583<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7050.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7050.dmg

AndrewMeadows:
The test is to:

(1) Bring up the audio stats window.
(2) Switch domains while paying attention to the "packets lost" stat as reported by the server
--> on bugged client this number can be NaN or bogus
--> on fixed client the number should be valid

I tested it and verified that the number is bad on other clients but fixed in this one.

PR #7043, opened 2016-02-08
Revamp the render engine ui

zzmp:
- Replaces `renderEngineDebug.js` with qml ui.
- Fixes loading of JS-defined render configurations with `Render.load(config)`.

PR #7051, opened 2016-02-09
Move graphics settings to a new dialog

zzmp:
- Adds ComboBox and ComboBoxBrowsable qml-integrated preferences.
- Adds persistent render engine configuration JSON file.
  - Allows for render profile presets (not yet included).
  - Allows for customization.

Pending on #7043 for a fix to the JobConfig's JS/qml integration.

### DIFF
--- a/libraries/render/src/render/Engine.cpp
+++ b/libraries/render/src/render/Engine.cpp
@@ -13,75 +13,39 @@
 
 #include <QtCore/QFile>
 
+#include <PathUtils.h>
+
 #include <gpu/Context.h>
 
 
 using namespace render;
 
-const QString ENGINE_CONFIG_DEFAULT = "Default";
-// TODO: Presets (e.g., "Highest Performance", "Highest Quality") will go here
-const QMap<QString, QString> Engine::PRESETS = {};
-
 Engine::Engine() :
-    _namedConfig(QStringList() << "Render" << "Engine"),
     _sceneContext(std::make_shared<SceneContext>()),
     _renderContext(std::make_shared<RenderContext>()) {
 }
 
-QStringList Engine::getNamedConfigList() {
-    QStringList list;
-    list << ENGINE_CONFIG_DEFAULT << PRESETS.keys();
-    auto current = _namedConfig.get();
-    if (!list.contains(current)) {
-        list << current;
-    }
-
-    return list;
-}
-
-QString Engine::getNamedConfig() {
-    return _namedConfig.get();
-}
-
-void Engine::setNamedConfig(const QString& config) {
-    _namedConfig.set(config);
-    loadConfig();
-}
-
-void Engine::loadConfig() {
+void Engine::load() {
     auto config = getConfiguration();
-    QString current = _namedConfig.get();
+    const QString configFile= "config/render.json";
 
-    if (_defaultConfig.isNull()) {
-        // Set the default
-        _defaultConfig =
-            QJsonDocument::fromJson(config->toJSON().toUtf8()).object();
-    }
-
-    if (current == ENGINE_CONFIG_DEFAULT) {
-        // Load the default
-        config->load(_defaultConfig.toObject());
-    } else if (PRESETS.contains(current)) {
-        // Load a preset
-        config->load(QJsonDocument::fromJson(PRESETS[current].toUtf8()).object());
+    QUrl path(PathUtils::resourcesPath() + configFile);
+    QFile file(path.toString());
+    if (!file.exists()) {
+        qWarning() << "Engine configuration file" << path << "does not exist";
+    } else if (!file.open(QIODevice::ReadOnly | QIODevice::Text)) {
+        qWarning() << "Engine configuration file" << path << "cannot be opened";
     } else {
-        QFile file(current);
-        if (!file.exists()) {
-            qWarning() << "Engine configuration file" << current << "does not exist";
-        } else if (!file.open(QIODevice::ReadOnly | QIODevice::Text)) {
-            qWarning() << "Engine configuration file" << current << "cannot be opened";
+        QString data = file.readAll();
+        file.close();
+        QJsonParseError error;
+        QJsonDocument doc = QJsonDocument::fromJson(data.toUtf8(), &error);
+        if (error.error == error.NoError) {
+            config->setPresetList(doc.object());
+            qDebug() << "Engine configuration file" << path << "loaded";
         } else {
-            QString data = file.readAll();
-            file.close();
-            QJsonParseError error;
-            QJsonDocument doc = QJsonDocument::fromJson(data.toUtf8(), &error);
-            if (error.error == error.NoError) {
-                config->load(doc.object());
-                qDebug() << "Engine configuration file" << current << "loaded";
-            } else {
-                qWarning() << "Engine configuration file" << current << "failed to load:" <<
-                    error.errorString() << "at offset" << error.offset;
-            }
+            qWarning() << "Engine configuration file" << path << "failed to load:" <<
+                error.errorString() << "at offset" << error.offset;
         }
     }
 }
--- a/libraries/render/src/render/Engine.cpp
+++ b/libraries/render/src/render/Engine.cpp
@@ -9,15 +9,81 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 
+#include "Engine.h"
+
+#include <QtCore/QFile>
+
 #include <gpu/Context.h>
 
-#include "Engine.h"
 
 using namespace render;
 
+const QString ENGINE_CONFIG_DEFAULT = "Default";
+// TODO: Presets (e.g., "Highest Performance", "Highest Quality") will go here
+const QMap<QString, QString> Engine::PRESETS = {};
+
 Engine::Engine() :
+    _namedConfig(QStringList() << "Render" << "Engine"),
     _sceneContext(std::make_shared<SceneContext>()),
     _renderContext(std::make_shared<RenderContext>()) {
+}
+
+QStringList Engine::getNamedConfigList() {
+    QStringList list;
+    list << ENGINE_CONFIG_DEFAULT << PRESETS.keys();
+    auto current = _namedConfig.get();
+    if (!list.contains(current)) {
+        list << current;
+    }
+
+    return list;
+}
+
+QString Engine::getNamedConfig() {
+    return _namedConfig.get();
+}
+
+void Engine::setNamedConfig(const QString& config) {
+    _namedConfig.set(config);
+    loadConfig();
+}
+
+void Engine::loadConfig() {
+    auto config = getConfiguration();
+    QString current = _namedConfig.get();
+
+    if (_defaultConfig.isNull()) {
+        // Set the default
+        _defaultConfig =
+            QJsonDocument::fromJson(config->toJSON().toUtf8()).object();
+    }
+
+    if (current == ENGINE_CONFIG_DEFAULT) {
+        // Load the default
+        config->load(_defaultConfig.toObject());
+    } else if (PRESETS.contains(current)) {
+        // Load a preset
+        config->load(QJsonDocument::fromJson(PRESETS[current].toUtf8()).object());
+    } else {
+        QFile file(current);
+        if (!file.exists()) {
+            qWarning() << "Engine configuration file" << current << "does not exist";
+        } else if (!file.open(QIODevice::ReadOnly | QIODevice::Text)) {
+            qWarning() << "Engine configuration file" << current << "cannot be opened";
+        } else {
+            QString data = file.readAll();
+            file.close();
+            QJsonParseError error;
+            QJsonDocument doc = QJsonDocument::fromJson(data.toUtf8(), &error);
+            if (error.error == error.NoError) {
+                config->load(doc.object());
+                qDebug() << "Engine configuration file" << current << "loaded";
+            } else {
+                qWarning() << "Engine configuration file" << current << "failed to load:" <<
+                    error.errorString() << "at offset" << error.offset;
+            }
+        }
+    }
 }
 
 void Engine::run() {
--- a/libraries/render/src/render/Engine.h
+++ b/libraries/render/src/render/Engine.h
@@ -12,6 +12,8 @@
 #ifndef hifi_render_Engine_h
 #define hifi_render_Engine_h
 
+#include <SettingHandle.h>
+
 #include "Context.h"
 #include "Task.h"
 
@@ -25,6 +27,19 @@ public:
     Engine();
     ~Engine() = default;
 
+    // Get the configurations
+    QStringList getNamedConfigList();
+
+    // Get the current named configuration
+    QString getNamedConfig();
+
+    // Set a named configuration
+    void setNamedConfig(const QString& config);
+
+    // Load the current named config
+    // The first time this is run, it will also set the current configuration as Default
+    void loadConfig();
+
     // Register the scene
     void registerScene(const ScenePointer& scene) { _sceneContext->_scene = scene; }
 
@@ -37,8 +52,14 @@ public:
     void run();
 
 protected:
+    Setting::Handle<QString> _namedConfig;
+    QJsonValue _defaultConfig;
+
     SceneContextPointer _sceneContext;
     RenderContextPointer _renderContext;
+
+private:
+    static const QMap<QString, QString> PRESETS;
 };
 using EnginePointer = std::shared_ptr<Engine>;
 
--- a/libraries/render/src/render/Engine.h
+++ b/libraries/render/src/render/Engine.h
@@ -27,18 +27,9 @@ public:
     Engine();
     ~Engine() = default;
 
-    // Get the configurations
-    QStringList getNamedConfigList();
-
-    // Get the current named configuration
-    QString getNamedConfig();
-
-    // Set a named configuration
-    void setNamedConfig(const QString& config);
-
-    // Load the current named config
-    // The first time this is run, it will also set the current configuration as Default
-    void loadConfig();
+    // Load any persisted settings, and set up the presets
+    // This should be run after adding all jobs, and before building ui
+    void load();
 
     // Register the scene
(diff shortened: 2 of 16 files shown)


zzmp:
@jherico I'm marking this as DNM because 98a3111..414042d are all part of #7043.
Feel free to look at e10a515..154ea33. I'll ping you again when it's no longer waiting on the other PR.

zzmp:
#### Intent
- Move the graphics settings to their own dialog (to declutter/allow them to grow?).
- Expand the graphics settings dialog to allow JSON loading and presets, to make future performance profiles easier to implement.

#### Testing
The graphics dialog is separate, so you'll need to test loading a different JSON configuration and reloading the default. To see how to format the JSON, use `Render.toJSON()`. Note that if there is *any* error, it will not load, and the warning should come up in your log. Here's what to test for, specifically:
##### ComboBox
@jherico do you know a way to test this, as it is not in use?
##### ComboBoxBrowsable
The combobox in the graphics menu should behave as expected. It should be formatted as expected. It is expected to display `Default` and any persisted file. As more custom files are selected, it is expected to append them to its list.
##### Loading configurations
- Custom JSON configurations load, and persist across sessions.
- Custom JSON configurations specifying enabling/disabling AO, shadows, or antialiasing should be overridden by the checkboxes in the dialog.
- Bad JSON configurations should still persist, despite not loading.
- Correctly formed JSON configurations should load, and not do anything about extra fields.
- The default configuration should always be selectable and sane. To see the default, use `Render.toJSON()` before beginning testing.
##### Logging
- Good JSON configurations should show a debug message in the log when they load.
- Bad JSON configurations (i.e. malformed JSON) should persist as a configuration, but show warnings in the log.

hifi-gustavo:
Build console output for commit: 154ea33b64d024b0430d718db27f377482a1512c available at: https://gist.github.com/c464c7ac8b779f1b8854<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7051.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7051.dmg

zzmp:
I neglected to make the checkbox configurations persist. Will be adding another commit.

zzmp:
Whole thing is getting revamped after discussion with @samcake. Stay tuned!

examples/utilities/tools/render/AO.qml
```diff
@@ -0,0 +1,102 @@
...
+    // }
+    // Label { id: timer; x: 140 }
+
+    Label { text: qsTr("Ambient Occlusion") }
+
+    CheckBox {
+        y: 1 * 25
+        text: qsTr("Dithering")
+        partiallyCheckedEnabled: false
+        checked: parent.config.ditheringEnabled
```
jherico: Avoid referring to things like this via `parent`.  Instead give the root QML object an id (this is almost universally going to be `root` and then refer to `root.config.ditheringEnabled`.  This makes it easy to re-organize your controls without having to re-write everything to account for the new layout.  The use of `parent` should probably be limited to things related to positioning and sizing of a control.

examples/utilities/tools/render/AO.qml
```diff
@@ -0,0 +1,102 @@
...
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or https://www.apache.org/licenses/LICENSE-2.0.html
+//
+import QtQuick 2.5
+import QtQuick.Controls 1.4
+
+Item {
+    width: 200
+    height: 200
+    property QtObject config: Render.getConfig("AmbientOcclusion")
```
jherico: unless the result of `Render.getConfig` is actually a QObject (and not a derived type) I would suggest using `property var config`

examples/utilities/tools/render/AO.qml
```diff
@@ -0,0 +1,102 @@
...
+//  Created by Zach Pomerantz on 2/8/2016
+//  Copyright 2016 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or https://www.apache.org/licenses/LICENSE-2.0.html
+//
+import QtQuick 2.5
+import QtQuick.Controls 1.4
+
+Item {
```
jherico: should be a column

examples/utilities/tools/render/AO.qml
```diff
@@ -0,0 +1,102 @@
...
+    width: 200
+    height: 200
```
jherico: don't need a height if you use a column

examples/utilities/tools/render/AO.qml
```diff
@@ -0,0 +1,102 @@
...
+    //     running: true
+    //     repeat: true
+    //     onTriggered: { parent.timer.text = config.gpuTime.toFixed(2) }
+    // }
+    // Label { id: timer; x: 140 }
+
+    Label { text: qsTr("Ambient Occlusion") }
+
+    CheckBox {
+        y: 1 * 25
```
jherico: don't need explicit Y positions if you use a column

examples/utilities/tools/render/AO.qml
```diff
@@ -0,0 +1,102 @@
...
+        text: qsTr("Dithering")
+        partiallyCheckedEnabled: false
+        checked: parent.config.ditheringEnabled
+        onCheckedChanged: { parent.config.ditheringEnabled = checked }
```
jherico: I don't think this should be necessary... if it currently is, that might because you're specifying QtObject instead of var above, or because the Q_PROPERTY declaration for `ditheringEnabled` isn't correct.  When you declare `checked: parent.config.ditheringEnabled` you're creating a 'binding' between the two values, the checked value of this CheckBox and the ditheringEnabled value of the config.  This means that a change in either should be propagated to the other.

examples/utilities/tools/render/AO.qml
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    ConfigSlider {
+        y: 2 * 25
+        config: parent.config
+        prop: "resolutionLevel"
```
jherico: you shouldn't need to do this kind of 'two value binding'.  Instead you should be able to replace line 40 and 41 with something like `value: parent.config.resolutionLevel`.  I'll explain more in a comment on `ConfigSlider`

examples/utilities/tools/render/ConfigSlider.qml
```diff
@@ -0,0 +1,58 @@
...
+//
+import QtQuick 2.5
+import QtQuick.Controls 1.4
+
+Item {
+    width: 400
+    height: 24
+    property string label
+    property QtObject config
+    property string prop
```
jherico: line 18 and 19 should be changed to something like `property var value`

examples/utilities/tools/render/ConfigSlider.qml
```diff
@@ -0,0 +1,58 @@
...
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or https://www.apache.org/licenses/LICENSE-2.0.html
+//
+import QtQuick 2.5
+import QtQuick.Controls 1.4
+
+Item {
+    width: 400
+    height: 24
+    property string label
```
jherico: give the `Label` declaration at line 35 an id, something like `labelControl`, and then change this declaration to `property alias label: labelControl.text`

examples/utilities/tools/render/ConfigSlider.qml
```diff
@@ -0,0 +1,58 @@
...
+    property QtObject config
+    property string prop
+    property real min: 0.0
+    property real max: 1.0
```
jherico: slider controls already have a min and max.  Change these to `property alias min: slider.minimumValue` and `property alias max: slider.maximumValue`

examples/utilities/tools/render/ConfigSlider.qml
```diff
@@ -0,0 +1,58 @@
...
+        stat.text = config[prop].toFixed(2);
+        slider.value = (config[prop] - min) / (max - min);
+    }
+    Component.onCompleted: init()
+
+    function update() {
+        var val = min + (max - min) * slider.value;
+        stat.text = val.toFixed(2);
+        config[prop] = val;
+    }
```
jherico: line 23 through 33 are unnecessary, see comments below

examples/utilities/tools/render/ConfigSlider.qml
```diff
@@ -0,0 +1,58 @@
...
+
+    Label {
+        text: parent.label
+        y: 7
+        anchors.left: parent.left
+        anchors.leftMargin: 8
+    }
+
+    Label {
+        id: stat
```
jherico: add a line here of `config[prop].toFixed(2)`.  This creates a binding so that when the value changes the text is automatically updated.

examples/utilities/tools/render/ConfigSlider.qml
```diff
@@ -0,0 +1,58 @@
...
+        anchors.left: parent.left
+        anchors.leftMargin: 140
+    }
+
+    Slider {
+        id: slider
+        y: 3
+        width: 192
+        height: 20
+        onValueChanged: parent.update()
```
jherico: you can bind this directly to the value....  `value: config[prop]`  This eliminates the need for an update or init method and will automatically update the `config[prop]` value when you change the slider control.

examples/utilities/tools/render/ItemsSlider.qml
```diff
@@ -0,0 +1,59 @@
...
+import QtQuick 2.5
+import QtQuick.Controls 1.4
+
+Item {
+    width: 400
+    height: 24
+    property string label
+    property QtObject config
+
+    function update() {
```
jherico: See notes on ConfigSlider.  You pretty much never need this kind of update function.

examples/utilities/tools/render/Buffer.qml
```diff
@@ -0,0 +1,99 @@
...
+    ExclusiveGroup { id: buffer }
+
+    function setDebugMode(mode) {
+        var debug = Render.getConfig("DebugDeferredBuffer");
+        console.log(mode);
+        debug.enabled = (mode != 0);
+        debug.mode = mode;
+    }
+
+    RadioButton {
```
jherico: This whole section could be replaced by

```
Column {
    spacing: 8
    Repeater {
        model: [ "Off", "Diffuse",  ... ]  
        RadioButton { 
            text: qsTr(modelData)
            exclusiveGroup: buffer
            onCheckedChanged: if (checked) setDebugMode(index);
        }
    }
}
```

examples/utilities/tools/render/main.qml
```diff
@@ -0,0 +1,64 @@
...
+        prop: "exposure"
+        label: qsTr("Tone Mapping Exposure")
+        min: -10; max: 10
+    }
+
+    // Ambient occlusion
+    AO { y: 175 }
+
+    // Debug buffer
+    Buffer { y: 475 }
```
jherico: Rather than using these fixed sizes, you should be anchoring things....

```
AO {
  id: aoControls
}
Buffer {
  id: bufferControls
  anchors { top: aoControls.bottom; margins: 8 }
}
```

Of course you'll need to fix AO so that it's controls don't extend past the bottom of it's Item.

interface/resources/qml/dialogs/preferences/ComboBoxBrowsablePreference.qml
```diff
@@ -0,0 +1,68 @@
...
+    Button {
+        id: button
+        anchors { right: parent.right; verticalCenter: dataComboBox.verticalCenter }
+        text: preference.browseLabel
+        onClicked: {
+            var browser = fileBrowserBuilder.createObject(desktop, { filter: preference.browseFilter, folder: fileDialogHelper.pathToUrl(preference.value) });
+            browser.selectedFile.connect(function(fileUrl){
+                console.log(fileUrl);
+                // Clear the file:/// prefix - these are always local files
+                var localPath = fileUrl.toString().replace(/^(file:\/\/\/)/, "");
```
jherico: there is actually a global `fileDialogHelper` type that is exposed to QML in our codebase that has `urlToPath` and `pathToUrl` methods.  It should probably be renamed to `fileHelper` or something, but you should use it instead of this manual string replace.

libraries/render/src/render/Engine.cpp
```diff
@@ -9,17 +9,82 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 
+#include "qfile.h"
```
jherico: `#include <QtCore/QFile>`

libraries/render/src/render/Engine.cpp
```diff
@@ -9,17 +9,82 @@
...
+
 #include <gpu/Context.h>
 
 #include "Engine.h"
```
jherico: The header for a given cpp file should always be the first include in the cpp file.

libraries/render/src/render/Engine.h
```diff
@@ -12,6 +12,8 @@
 #ifndef hifi_render_Engine_h
 #define hifi_render_Engine_h
 
+#include "SettingHandle.h"
```
jherico: includes from other libraries should use the `<>` format.  `""` should be reserved fro files local the this library, i.e. files in `render/src`

libraries/render/src/render/Task.h
```diff
@@ -85,6 +87,9 @@ class TaskConfig : public JobConfig {
 
     void init(Task* task) { _task = task; }
 
+    // getter for qml integration, prefer the templated getter
+    Q_INVOKABLE QObject* getConfig(QString name) { return QObject::findChild<JobConfig*>(name); }
```
jherico: `const QString& name`

libraries/ui/src/QmlWindowClass.cpp
```diff
@@ -281,6 +281,7 @@ void QmlWindowClass::setPosition(const glm::vec2& position) {
     QQuickItem* targetWindow = asQuickItem();
     DependencyManager::get<OffscreenUi>()->executeOnUiThread([=] {
         targetWindow->setPosition(QPointF(position.x, position.y));
+        emit moved(position);
```
jherico: I'm not sure you can (or that it's safe to ) do the emit on the UI thread.

jherico:
I haven't tested this yet, but looking at the QML I have a few notes...

Lots of the items appear to have fixed widths, heights, and positioning.  This may be sufficient for now, but I would recommend familiarizing yourself with the anchoring mechanisms and the layout types.  For instance, the AO.qml is basically just a list of controls each with a custom Y value, you can just do this:

```
Column {
  spacing: 8 
  Label { ... } 
  CheckBox { ... } 
  ConfigSlider { ... } 
  ConfigSlider { ... } 
  ConfigSlider { ... } 
}
```

This will automatically stack the controls, and the height of the Column will automatically be the height of all the controls with the spacing.  This makes it much easier to move controls around and add new controls without having to edit all the Y values.  Also note that because you've given the AO.qml a fixed height of 200, the last 3 controls actually have Y values that push them outside of the `Item`.  This may not be an issue right now because by default QML parent's don't clip their children, but if you were placing other controls below the AO control, you'd end up seeing the AO controls overlay the stuff below it.

hifi-gustavo:
Build console output for commit: 9f82bc114b48d49535cc3d24f0feda5ad41d1204 available at: https://gist.github.com/c464c7ac8b779f1b8854<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7051.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7051.dmg

hifi-gustavo:
Build console output for commit: 6d3cf288f9a4891aa2af255e1ac29b6ee0dafc33 available at: https://gist.github.com/c464c7ac8b779f1b8854<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7051.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7051.dmg

hifi-gustavo:
Build console output for commit: 36c6e9fdfe49e84be3d4bd200f346fc1a1eb5880 available at: https://gist.github.com/c464c7ac8b779f1b8854<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7051.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7051.dmg

hifi-gustavo:
Build console output for commit: d0f3ad75b8f0dbad7f49391a7f69f8580ee5e76d available at: https://gist.github.com/c464c7ac8b779f1b8854<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7051.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7051.dmg

OmegaHeron:
Tested Windows 7

Ambient Occlusions and Antialiasing stick to set vale across runs - shadows **do not*

Triplelexx:
same here

zzmp:
Thanks @OmegaHeron and @Triplelexx. This last commit should fix it.

hifi-gustavo:
Build console output for commit: 232f06e1757bc18fe15c7b2c700e3e362f6d0065 available at: https://gist.github.com/c464c7ac8b779f1b8854<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7051.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7051.dmg

judasshuffle:
Works for me, saves all changes after a relog nice

Triplelexx:
That fix hurts my head to understand. Does indeed do the job. 
Settings are all saving correctly now.

OmegaHeron:
Retested Windows 7 - All settings apply and persist.

PR #7054, opened 2016-02-09
Fix ToolWindow's initial position (e.g., as used by edit.js).

howard-stearns:
Applies when there is no ToolWindow.position in Interface.ini

### DIFF
--- a/interface/resources/qml/ToolWindow.qml
+++ b/interface/resources/qml/ToolWindow.qml
@@ -20,8 +20,8 @@ Windows.Window {
     property string newTabSource
     property alias tabView: tabView
     onParentChanged: {
-        x = desktop.width / 2 - width / 2;
-        y = desktop.height / 2 - height / 2;
+        x = 120;
+        y = 120;
     }
 
     Settings {


howard-stearns:
After first use, there is a setting for this. But without that, the onParentChanged code is used. However, at that time, desktop.width and desktop.height are not initialized yet, so the tool ends up half of the screen in both dimensions. The new code provides a reasonable default without trying to be so fancy.

OmegaHeron:
For anyone else QA testing this - keep in mind it's a fix for case where no pre-existing Interface.ini exists.  On windows this would point to making sure you don't have an Interface.ini in \Users\your_user_name\appData\Roaming\High Fidelity\  -- do not count on PR build using a "blank" config.  I'm not convinced the PR builds actually honor an isolated config - 1) even if you don't tell installer to copy production settings to PR config it knows things from Interface.ini and 2) if you make changes and apply in PR builds they're present in production build via Interface.ini

hifi-gustavo:
Build console output for commit: e2ac48fac0c147a7db7f688cd682735d1bc5ff0b available at: https://gist.github.com/71ca5d77658fff59e366<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7054.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7054.dmg

howard-stearns:
https://app.asana.com/0/32622044445063/87698183154196

Testing/QA instructions:

0) Uninstall High Fidelity / console / stack manager / interface, remove yourDirectory\AppData\Roaming\High Fidelity, and install this pr. 
(Really, it is the removal of AppData\...\Interface.ini that matters, or even just removing the [ToolWindow.Position] section, but let's be complete.)

1)  Start High Fidelity (Interface) and confirm that the default scripts are loaded. (Assuming here that your internet connection is good and that Amazon S3 hasn't space out.)

2) Click the edit button and observe the initial position of the tool window that opens. It should be fully on the screen for any reasonable screen size.  Close and restart, and the tools should be where you left it (even if you move it around before closing).

OmegaHeron:
Alternatively - move out the old appData\Local\High Fidelity and appData\Roaming\High Fidelity folders so you can restore, if desired.

howard-stearns:
Correct @OmegaHeron

OmegaHeron:
Passes on Windows 7.  Edit properties window places well in screen bounds and default scripts load.

interface/resources/qml/ToolWindow.qml
```diff
@@ -20,8 +20,8 @@ Windows.Window {
     property string newTabSource
     property alias tabView: tabView
     onParentChanged: {
-        x = desktop.width / 2 - width / 2;
-        y = desktop.height / 2 - height / 2;
+        x = 120;
```
ZappoMan: I'm not sure this is the best fix.

1) why would the old code not work in the case of an empty ini?
2) what happens if the window starts up with a smaller size (not sure how windows decided to initialize a window the first time an application is launched)

howard-stearns: re 1) See my first comment. desktop.width and .height are zero at that point.

re 2) If the window is narrower than tool.width + 120 or shorter than tool.height + 120, the user is going to have crappy experience. Even if desktop dimensions were known at that point, the old code would still be the wrong answer if the windows were narrower than the tool width or shorter than the tool height. So, I believe the simplest thing is to create a reasonable fixed initial position. (As web pages often do with a offset from top left.)

PR #7056, opened 2016-02-10
Render Scene Octree : A Minute of Arc

samcake:
- Add a spatial octree datastructure to the Scene to store the Items and access them efficiently based on query regions
-The render octree is exactly the same as the domain Entity octree
- When starting a frame, the fetch phase is now achieved by doing a selection of the visible cells in the octree (respecting the viewfrustum AND the lod angle threshold)
- The collected Items after the fetch phase are then tested individually
- Entities now notify their renderItems when they are moving or changing bounds (Primitive...)
- The RenderDeferredTask has been reorganized to use efficiently the octree fetch/culling selection and then filter it into multiple buckets

- The Culling /LOD Test is still NOT EXACTLY the same as the one used by the Network queries. I suggest we go with that for now and refine in future pr


hifi-gustavo:
Build console output for commit: da158bcd85b88d277a3225c6b8175edb2cf32cf2 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: 8a4e2cb0c5c4a077c7fc5e629a3c8593b2ae6be8 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: 5ef7a19d67be3f0e49b4f77a6d56dfd0ff77c4ab available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: 9326aebf20331a37bcb02492b4f0c3ee28e3de6a available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: b295a3732e42a0897f42c4a15a9eab5376fe4544 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: b295a3732e42a0897f42c4a15a9eab5376fe4544 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

howard-stearns:
Looking good!
Here's some results using tests/performance/renderableMatrix.js, in localhost/1000,1000,1000:

Release build 4159:
One instance of script, LOD at max:~17ms in render, of which 7 are in cull opaque
Two instances: 30ms in render, 10 in cull opaque
Two instances, LOD reset to normal after script load: 10ms in render of which 8 in cull opaque
5 instance, LOD reset: 20ms in render of which 14 in cull opaque
     (Sometimes less. Note that the 3rd and 4th case timse in cull opaque is impetus for this PR.)

PR build (the one from b295a37):
One instance of script, LOD at max:12ms, of which 3 are in cull scene selection
Two instances: 20ms in render, 5 in cull scene selection
Two instance, LOD reset to normal after script load: 5ms in render of which 4 in cull scene selection
5 instance, LOD reset: 7ms in render, 5 in cull scene selection -- and then crash during lifetime expiration, but there are separate prs on that.

jherico:
Can we create a set of tests of scene item selection that run on both the old and new code to provide measurable comparisons of performance?

hifi-gustavo:
Build console output for commit: 6922c9adf0b6211f0c7aadb2d20c284e0af35a8f available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

samcake:
Please start reviewing thanks

hifi-gustavo:
Build console output for commit: dedfee3acc1f6012811b6a707f82afabc8236dda available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: b94c2c9b7f9b05dcb65597b4263b2e8c84253620 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: f12030c1a4158b2cd5aa39f1fad7f73cf2513ef5 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

libraries/octree/src/ViewFrustum.h
```diff
@@ -160,7 +165,7 @@ class ViewFrustum {
     float _fieldOfView = DEFAULT_FIELD_OF_VIEW_DEGREES;
     glm::vec4 _corners[8];
     glm::vec3 _cornersWorld[8];
-    enum { TOP_PLANE = 0, BOTTOM_PLANE, LEFT_PLANE, RIGHT_PLANE, NEAR_PLANE, FAR_PLANE };
+ //   enum { TOP_PLANE = 0, BOTTOM_PLANE, LEFT_PLANE, RIGHT_PLANE, NEAR_PLANE, FAR_PLANE };
```
ZappoMan: dead code

samcake:
Test this please

hifi-gustavo:
Build console output for commit: f12030c1a4158b2cd5aa39f1fad7f73cf2513ef5 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: 171acb883f9ea40cbd26efd35bd1afeeb7ab2b38 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: 5acfd5ff53f9f6acf274e537c5311992cf0360b2 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: 40c283cba202394af55cb025142d45ddf5e6614e available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: c248732bc701f61ee815c82f06726d8899197c91 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: 2d33bb83e12d19ad36da15684b2121db31f64129 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

libraries/entities-renderer/src/RenderableEntityItem.h
```diff
@@ -66,8 +68,22 @@ class SimpleRenderableEntityItem {
...
+
+        pendingChanges.updateItem<RenderableEntityItemProxy>(_myItem, [](RenderableEntityItemProxy& data) {
+        });
+
+        scene->enqueuePendingChanges(pendingChanges);
+    }
+
 private:
-    render::ItemID _myItem;
+    render::ItemID _myItem = render::Item::INVALID_ITEM_ID;
```
howard-stearns: Mix of initialization styles (e.g., vs Overlay.h above). Not that it bothers me. Just sayin'.

hifi-gustavo:
Build console output for commit: c4faa091a154ca8543063a37bd5a00ca1439dd71 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: 9bfeb4a1d5cb7ce1c301925a8c16039a60c8f90e available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

OmegaHeron:
For what it's worth - ran this PR on Windows 7 - checked out a few scenes from simple to torturous (sandbox) and couldn't see any difference to production build in impression.

samcake:
Thanks Omega, did you see any positive impact on the framerate in bigger scenes ?

OmegaHeron:
It probably needs some sensors to record performance, but, at Sandbox - this PR seemed to do a little better in that render FPS was closer to Present FPS where with production version render was often slower compare to present rate.  Not sure I understand what Present Rate actually indicates though.

samcake:
@OmegaHeron 
Present rate is basicaly the refresh rate of the screen and we want this to be at exactly what you expect from your typical screen (~60Hz) or you oculus (DK2 at 75Hz and CV1 at 90Hz). Present is basically using the current latest image we produced and "present" it to the final display device

Render rate is actually how frequently we redraw completely a brand new frame (image) from the current point of view that feeds the "present" process. Rendering is a much longer and costly process than Present, that s why the 2 are running in different threads.
Yes the goal is that render rate chase present rate as best as we can so if you see it increasing it s a good sign.

OmegaHeron:
Got it, thanks @samcake - I saw an Asana issue about FPS not living up to Present rate (when it likely should) and can say I see this too in current production code - sometimes render FPS will drop as you pan a scene then recover - other times it drops and doesn't.  The key thing is, if I see render fps = 30 (not 29 or 31 or something else -- 30) it tends to lock at 30 and never recover.  In my brief testing with this PR I'm not seeing that, yet.  This is all non-HMD here.

howard-stearns:
Sam and I looked at a test case together, and found that the culling appears to be working as expected, but that the remaining rendering time is higher than release/expected. Sam is investigating.

libraries/render/src/render/CullTask.cpp
```diff
@@ -0,0 +1,293 @@
...
+#include <algorithm>
+#include <assert.h>
+
+#include <PerfStat.h>
+#include <ViewFrustum.h>
+#include <gpu/Context.h>
+
+using namespace render;
+
+void render::cullItems(const RenderContextPointer& renderContext, const CullFunctor& cullFunctor, RenderDetails::Item& details,
```
ZappoMan: I'm confused by this function... is it still being used? It appears to still be a linear vector approach to culling... shouldn't it be removed completely?

samcake: You are right, we don't use it anymore for the main rendering
it s still used by the shadow pass for now
once we move forward with using the octree , i can clean up the sadow pass
and then remove it

hifi-gustavo:
Build console output for commit: e3307d91ad48195a7d3d3f2b7e852acb6b96aca5 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: bcd7876f6cb8b890b883c98a8ca6df91b7827d0a available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

hifi-gustavo:
Build console output for commit: 21c9b74a6867a01ad6388a2bfaa178350f32b212 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

samcake:
test this please

hifi-gustavo:
Build console output for commit: 21c9b74a6867a01ad6388a2bfaa178350f32b212 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

howard-stearns:
In one try before leaving the office desktop, I was not able to repro the performance variations using a newer version. (Although I had repeated the above-reported performance on most of several restarts earlier.) I don't know what causes the variability. In addition to different version:
  - Maybe different stats? Apparently the stats display can sometimes have a big effect.
  - Octree loading? We've previously observed that even if the leaf count is the same from one run to the next, the non-leaf node count can vary wildly. (Maybe dependent on entity load order?)
  - ?

My feeling is that we should merge and deal with any performance variation fallout later.

nbq:
This seems to work well for me. I tested with the PR and the production Interface and the production Interface seems to lock up when I turn or when its loading data in the scene. The PR seemed to fix that so it felt smoother as a whole package. I say it worked well for me.

ZappoMan:
@samcake - this this statement in your description true? I thought you made them match... "The Culling /LOD Test is still NOT EXACTLY the same as the one used by the Network queries. I suggest we go with that for now and refine in future pr"

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+    };
+}
+
+// CLose the namespace here before including the Item in the picture, maybe Octre could stay pure of it
+namespace render {
+
+    // An octree of Items organizing them efficiently for culling
+    // The octree only cares about the bound & the key of an item to store it a the right cell location
+    class ItemSpatialTree : public Octree {
+        float _size{ 32768.0f };
```
ZappoMan: please use `TREE_SCALE` from `OctreeContstants.h`

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+        float _invSize{ 1.0f / _size };
+        glm::vec3 _origin{ -16384.0f };
```
ZappoMan: please use `HALF_TREE_SCALE` from `OctreeContstants.h`

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+}
+
+// CLose the namespace here before including the Item in the picture, maybe Octre could stay pure of it
+namespace render {
+
+    // An octree of Items organizing them efficiently for culling
+    // The octree only cares about the bound & the key of an item to store it a the right cell location
+    class ItemSpatialTree : public Octree {
+        float _size{ 32768.0f };
+        float _invSize{ 1.0f / _size };
```
ZappoMan: technically coding standard says space between variable name and {

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+        void freeBrick(Index index);
+
+        Cell& editCell(Index index) {
+            assert(checkCellIndex(index));
+            return _cells[index];
+        }
+
+
+        // Octree members
+        Cells _cells = Cells(1, Cell()); // start with only the Cell root
```
ZappoMan: I think the coding standard says this should be a {} style initializer, but maybe we should update the coding standard, because ugh!

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+
+        class FrustumSelector {
+        public:
+            Coord4f frustum[6];
+            Coord3f eyePos;
+            float   angle;
+            float   squareTanAlpha;
+
+            void setAngle(float a) {
+                angle = std::min(glm::radians(45.0f), a); // no worse than 45 degrees
```
ZappoMan: coding standard: magic numbers `MAX_VIEW_ANGLE`

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+                angle = std::max(glm::radians(1.0f/60.0f), a); // no better than 1 minute of degree
```
ZappoMan: coding standard: magic numbers `MIN_VIEW_ANGLE`

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+        // {11, 2048,  8589934592}
+        // {12, 4096,  68719476736}
+        // {13, 8192,  549755813888}
+        // {14, 16384, 4398046511104}
+        // {15, 32768, 35184372088832}
+        // {16, 65536, 281474976710656}
+
+        // Max depth is 15 => 32Km root down to 1m cells
+        using Depth = int8_t;
+        static const Depth ROOT_DEPTH{ 0 };
```
ZappoMan: coding standard says space between variable and {

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+        static const Depth MAX_DEPTH{ 15 };
```
ZappoMan: coding standard says space between variable and {

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+        static const Depth METRIC_COORD_DEPTH{ 15 };
```
ZappoMan: coding standard says space between variable and {

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+        using Coord = int16_t;
+        using Coord3 = glm::i16vec3;
+        using Coord4 = glm::i16vec4;
+        using Coord3f = glm::vec3;
+        using Coord4f = glm::vec4;
+
+        static Coord depthBitmask(Depth depth) { return Coord(1 << (MAX_DEPTH - depth)); }
+
+        static Depth coordToDepth(Coord length) {
+            Depth d = MAX_DEPTH;
```
ZappoMan: coding standard "1.2.12. Abbreviations in names should be avoided."

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+                assert((pos.x < (1 << depth)) && (pos.y < (1 << depth)) && (pos.z < (1 << depth)));
+            }
+        public:
+            using vector = std::vector< Location >;
+
+            Location() {}
+            Location(const Coord3& xyz, Depth d) : pos(xyz), depth(d) { assertValid(); }
+            Location(Depth d) : pos(0), depth(d) { assertValid(); }
+
+            Coord3 pos{ 0 };
```
ZappoMan: coding standard - "3.2.2 Initialization of member variables with default values" says space between variable name and {

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+            uint8_t spare{ 0 };
```
ZappoMan: coding standard - "3.2.2 Initialization of member variables with default values" says space between variable name and {

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+            Depth depth{ ROOT_DEPTH };
```
ZappoMan: ibid

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+
+            Coord3f getCenter() const { 
+                Coord3f center = (Coord3f(pos) + Coord3f(0.5f)) * Coord3f(Octree::getInvDepthDimension(depth));
+                return center;
+            }
+
+            bool operator== (const Location& right) const { return pos == right.pos && depth == right.depth; }
```
ZappoMan: personally I think "other" would be a better name for this. Only because, since we're dealing with "locations" when I first say "right" I thought that the "right" referred to the right part of the location, and I didn't understand how you were comparing only part of the location. I later realized this was an == operator and you used "right" to mean the right side of the ==... anyway, just a suggestion, not a merge blocker.

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+            Location() {}
+            Location(const Coord3& xyz, Depth d) : pos(xyz), depth(d) { assertValid(); }
+            Location(Depth d) : pos(0), depth(d) { assertValid(); }
+
+            Coord3 pos{ 0 };
+            uint8_t spare{ 0 };
+            Depth depth{ ROOT_DEPTH };
+
+            Coord3f getCenter() const { 
+                Coord3f center = (Coord3f(pos) + Coord3f(0.5f)) * Coord3f(Octree::getInvDepthDimension(depth));
```
ZappoMan: I keep looking at this and thinking that this is a confusing way to get the center of a cell...  it may be acurate, but it just reads strange to me... something of the form (position + half my width) makes more sense... so even `Coord3f(pos) * Coord3f(Octree::getInvDepthDimension(depth) * 0.5f);` seems like it would make more sense...

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+        static int getDepthDimension(Depth depth) { return 1 << depth; }
+        static float getDepthDimensionf(Depth depth) { return (float) getDepthDimension(depth); }
+        static float getInvDepthDimension(Depth depth) { return INV_DEPTH_DIM[depth]; }
+        static float getCoordSubcellWidth(Depth depth) { return (1.7320f * getInvDepthDimension(depth) * 0.5f); }
+
+        // Need 16bits integer coordinates on each axes: 32768 cell positions
+        using Coord = int16_t;
+        using Coord3 = glm::i16vec3;
+        using Coord4 = glm::i16vec4;
+        using Coord3f = glm::vec3;
```
ZappoMan: I don't understand why you're using a new type name here, why not use `glm::vec3` directly? I know what a vec3 does and what methods I can use with it... but a Coord3f seems like some class that I'm not familiar with.

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+            void assertValid() {
+                assert((pos.x >= 0) && (pos.y >= 0) && (pos.z >= 0));
+                assert((pos.x < (1 << depth)) && (pos.y < (1 << depth)) && (pos.z < (1 << depth)));
+            }
+        public:
+            using vector = std::vector< Location >;
+
+            Location() {}
+            Location(const Coord3& xyz, Depth d) : pos(xyz), depth(d) { assertValid(); }
+            Location(Depth d) : pos(0), depth(d) { assertValid(); }
```
ZappoMan: variable names not abbreviated

libraries/render/src/render/Octree.h
```diff
@@ -0,0 +1,450 @@
...
+        class Location {
+            void assertValid() {
+                assert((pos.x >= 0) && (pos.y >= 0) && (pos.z >= 0));
+                assert((pos.x < (1 << depth)) && (pos.y < (1 << depth)) && (pos.z < (1 << depth)));
+            }
+        public:
+            using vector = std::vector< Location >;
+
+            Location() {}
+            Location(const Coord3& xyz, Depth d) : pos(xyz), depth(d) { assertValid(); }
```
ZappoMan: variable names not abbreviated

libraries/render/src/render/Octree.h
samcake: THe pos (of the location) is expressed in its Depth level's space where each coordinate is in the range [0 -> 2^Depth] (and also just integer)
In this function we are computing the actual floating point coordinate of the center of the cell in the "Metric_Space" so we go from the cell location ( low left corner) add half and then multiply this by the "InvDepthDimension(depth)" meaning that we scale this coordinate into the metric space.

hifi-gustavo:
Build console output for commit: 545f276fdf61d3d7d0eed3d9b71411375dfeda25 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

libraries/render/src/render/Octree.h
samcake: OctreeConstants.h is in Octree which is not known from the render.lib so i m going to pass these parameters at contstruction ( from Applicaiton.h)
BUt i m going to leave this default values...

ZappoMan: We should at least add a FIXME note. We could move OctreeConstants.h to
shared... and then both can reference them.

libraries/render/src/render/Octree.h
samcake: BY introducing these aliases local to the class i can reason about the unit/type i m expecting in the context of this class and i write the code and the algorithm around that.
WHen i expect a Coord3f i know what it means semantically even though yes it s implemented with a glm::vec3.
I could have another stuff that would be in the same case but is different from COrrd3f and so now i want to make sure i differenciate between the 2 looking at these types
At the end of the day it s really a way to carry the notion of Unit through these types and check that my math is correct

libraries/render/src/render/Octree.h
samcake: what is the rule to name parameters when we have a public memeber that we are assigning from the parameter?

samcake: I do think d is much more appropriate than anything else here no ?

hifi-gustavo:
Build console output for commit: 0c0089f736edd925c8e58c065092683d88c78304 available at: https://gist.github.com/d5c856e5cc60557210f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7056.dmg

libraries/render/src/render/Octree.h
ZappoMan: personally I'd go with depth(depth) but... that's just me... I won't block the merge for this, but the coding standard is pretty clear.

PR #7058, opened 2016-02-10
SpatiallyNestable: bug fixes for kinematic entities

hyperlogic:
* Kinematic entities were being simulated with position in the local frame but velocity in the global frame.
* Also, because getVelocity() was returning non-zero values when the parent was moving, kinematic objects were being simulated even if they should have been at rest.
* Bug fix for calculating velocity direction was incorrect, if the parent frame was rotated.

### DIFF
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -885,83 +885,93 @@ void EntityItem::simulateKinematicMotion(float timeElapsed, bool setFlags) {
         return;
     }
 
-    if (hasAngularVelocity()) {
-        glm::vec3 angularVelocity = getAngularVelocity();
+    if (hasLocalAngularVelocity()) {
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: 727e289ecda21e02434a413b770e2a9aceabc251 available at: https://gist.github.com/443a5f00f4174d85914c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7058.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7058.dmg

sethalves:
to test this:
pick up the bow and shake it.  before this PR, the string would sometimes drift away.

cozza13:
tested this and it works

PR #7063, opened 2016-02-10
Change circle3d innerRadius to fillRatio

zzmp:
The fill ratio calculates the inner radius when rendered,
as outerRadius * fillRatio.
Note that the ratio is a ratio of radii,
*not* the ratio of areas.


zzmp:
#### Intent
Make the behavior of circle3d more intuitive.
Before, `innerRadius` could be set larger than `outerRadius`. Now, `fillRatio` is bounded in [0, 1].

#### Testing
Is this intuitive to you? Is it a problem that fillRatio is a ratio of radii, and not fills?
- Make sure innerRadius has no effect, and fillRatio sets the inner radius on solid circle3d overlays.
- Make sure fillRatio is bounded in [0,1]. Setting it to a value outside of this should not be reflected when checking the property with `Overlays.getProperty`.

examples/libraries/entitySelectionTool.js
```diff
@@ -3259,7 +3259,7 @@ SelectionDisplay = (function() {
             Overlays.editOverlay(rotateOverlayInner, {
                 visible: true,
                 size: innerRadius,
-                innerRadius: 0.9,
+                fillRatio: 0.9,
```
ZappoMan: shouldn't these be ` fillRatio: 0.1,` ??

hifi-gustavo:
Build console output for commit: a638e29b40185b3d54c8321c4abadef4ec6fbc5e available at: https://gist.github.com/58826f7da45cc6cbf290<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7063.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7063.dmg

jherico:
Fill ratio makes setting the inner radius somewhat indirect, and therefore slightly more confusing (as evidenced by @ZappoMan comment).  What's the motivation for this change?

zzmp:
@jherico it was unbounded, such that innerRadius could be larger than outerRadius. It worked, but was semantically off. The motivation is to add bounding and make the properties more intuitive.

If you consider it more confusing, than it fails at that goal, and needs to be abandoned or reworked, even if that means just changing property names (`outerRadius`->`radius`, `fillRatio`->`fill`?).

jherico:
the properties for overlays aren't quite as rigid as for entities, so I think we should be very forgiving of what people do.  For instance, I would like it if I could specify `radius` & `innerRadius` or `radius` & `outerRadius` or `innerRadius` & `outerRadius`, or just `radius` and always have it do the right thing, including switching the inner and outer values if they're in the wrong order.  

We could also include `fillRatio` or `fillPercentage` in there, but I find it a little confusing because I would intuit that it meant the percentage of the raidus that was filled, starting from the center, rather than creating a ring.  So I do feel this approach might be more confusing.

We could also look at the QML mechanism to do something similar.  the QML Rectangle type supports a color, but also a border.color and border.width, and you can create a hollow rectangle by specifying a particular border color and then specifying a color with 0 alpha for the main rectangle color.  We could apply the same thing here and I think it would be a) more intuitive and b) would also happen to allow people to create circles with a colored ring without having to create two circle3d overlays.

zzmp:
Closing - does not meet intented goal.

PR #7061, opened 2016-02-10
Fix crash when deleting many moving dynamic objects.

AndrewMeadows:
- fix a recently added crash mode: when deleting lots of dynamic objects at once

PR #7069, opened 2016-02-10
Fix crash when propagating simulation ownership on deleted objects

AndrewMeadows:
- fix crash when deleting a pile of colliding objects

### DIFF
--- a/libraries/physics/src/PhysicalEntitySimulation.cpp
+++ b/libraries/physics/src/PhysicalEntitySimulation.cpp
@@ -167,12 +167,13 @@ void PhysicalEntitySimulation::getObjectsToRemoveFromPhysics(VectorOfMotionState
         _entitiesToAddToPhysics.remove(entity);
 
         EntityMotionState* motionState = static_cast<EntityMotionState*>(entity->getPhysicsInfo());
-        assert(motionState);
-        _pendingChanges.remove(motionState);
-        _outgoingChanges.remove(motionState);
-        _physicalObjects.remove(motionState);
-        result.push_back(motionState);
-        _entitiesToRelease.insert(entity);
+        if (motionState) {
+            _pendingChanges.remove(motionState);
+            _outgoingChanges.remove(motionState);
+            _physicalObjects.remove(motionState);
+            result.push_back(motionState);
+            _entitiesToRelease.insert(entity);
+        }
 
         if (entity->isSimulated() && entity->isDead()) {
             _entitiesToDelete.insert(entity);
--- a/libraries/physics/src/PhysicsEngine.cpp
+++ b/libraries/physics/src/PhysicsEngine.cpp
@@ -136,23 +136,19 @@ void PhysicsEngine::addObjectToDynamicsWorld(ObjectMotionState* motionState) {
     motionState->clearIncomingDirtyFlags();
 }
 
-// private
-void PhysicsEngine::removeObjectFromDynamicsWorld(ObjectMotionState* object) {
-    // wake up anything touching this object
-    bump(object);
-    removeContacts(object);
-
-    btRigidBody* body = object->getRigidBody();
-    assert(body);
-    _dynamicsWorld->removeRigidBody(body);
-}
-
 void PhysicsEngine::removeObjects(const VectorOfMotionStates& objects) {
+    // first bump and prune contacts for all objects in the list
     for (auto object : objects) {
-        removeObjectFromDynamicsWorld(object);
+        bumpAndPruneContacts(object);
+    }
+
+    // then remove them
+    for (auto object : objects) {
+        btRigidBody* body = object->getRigidBody();
+        assert(body);
+        _dynamicsWorld->removeRigidBody(body);
 
         // NOTE: setRigidBody() modifies body->m_userPointer so we should clear the MotionState's body BEFORE deleting it.
-        btRigidBody* body = object->getRigidBody();
         object->setRigidBody(nullptr);
         body->setMotionState(nullptr);
         delete body;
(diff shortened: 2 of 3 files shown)


AndrewMeadows:
The top of the callstack for this crash mode looks like this:  crash callstack for this mode looks like this:

    interface.exe!PhysicsEngine::bump(ObjectMotionState * motionState) Line 410 C++
    interface.exe!PhysicsEngine::removeObjects(const QVector<ObjectMotionState *> & objects) Line 152 C++
    interface.exe!Application::update(float deltaTime) Line 3110 C++
    interface.exe!Application::idle(unsigned __int64 now) Line 2440 C++
    interface.exe!Application::paintGL() Line 1318 C++

The way to tickle it is to make a pile of dynamic colliding objects and then to delete them all at once.  Sometimes objectB would be deleted before objectA and we would try to "bump" the simulation ownership of invalid objectB.  There is a test script in #7061 that has been able to reproduce this crash at least once, although it appears the success rate is somewhere around one in ten.

Note: this PR depends on PR-7061 since I was hitting that other crash mode before reaching this one.

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -136,23 +136,19 @@ void PhysicsEngine::addObjectToDynamicsWorld(ObjectMotionState* motionState) {
...
-    btRigidBody* body = object->getRigidBody();
-    assert(body);
-    _dynamicsWorld->removeRigidBody(body);
-}
-
 void PhysicsEngine::removeObjects(const VectorOfMotionStates& objects) {
+    // first bump and prune contacts for all objects in the list
+    for (auto object : objects) {
+        bumpAndPruneContacts(object);
+    }
```
AndrewMeadows: The actual fix for the crash is here: first bump and prune contacts for all objects in the list above, then remove them below.

hifi-gustavo:
Build console output for commit: 00e90f6f7d463f93b196c4f99db1eec954dc83a3 available at: https://gist.github.com/af34a185856059296d01<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7069.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7069.dmg

AndrewMeadows:
The internal asana task for this is here: https://app.asana.com/0/26623005675174/87216130762001

bluetoe:
This still crashes interface for me every once in awhile. using the script from PR-7061 I've crashed on 8th run, then on 10th run, then 12th. Running the script a second time before first set has deleted seems to make crash happen quicker, but not always

AndrewMeadows:
@bluetoe, the question I have is: What does the callstack look like?  There is at least one, maybe two, distinct crash modes that could happen from the same test.  I'm trying to track it down however I'm having trouble reproducing it.  Since I've been testing the Debug build so far I'm now moving to Release to see if that helps.

bluetoe:
I was going to get call stack but found that I needed to update VS first and I got lazy.. but ok. I will try to get it

bluetoe:
pr-7069-crash

AndrewMeadows:
Yup, this still needs work.  I was just able to reproduce the crash on windows and linux using the **Release** build.

AndrewMeadows:
Ok, I give up making the cleanup flow airtight... I added a check for valid **MotionState** when I used to assert.  After that I wasn't able to crash on linux despite much creation/deletion of objects, so this might be fixed.  I will test on windows too.

hifi-gustavo:
Build console output for commit: 7aa1d2deeaebec54a229cffbcefc70a070f31b04 available at: https://gist.github.com/af34a185856059296d01<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7069.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7069.dmg

hifi-gustavo:
Build console output for commit: ac60917b546e987e320452a52b742d96db93280e available at: https://gist.github.com/af34a185856059296d01<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7069.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7069.dmg

bluetoe:
That last one seems to have fixed it. I tried 24 runs of the test script and no crash, then tried 12 double click runs, then tried 8 triple click runs and no crash. Although with triple click the physics seem to disappear for some of the blocks. But anyway, 14400 blocks deleted and simulation ownership changed so, much better than before.

AndrewMeadows:
Yup, I did a bunch of testing on Windows and linux simultaneously last night right before I left.  Probably about 15 events, each with 200 entities, and neither crashed.

libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -167,12 +167,13 @@ void PhysicalEntitySimulation::getObjectsToRemoveFromPhysics(VectorOfMotionState
...
-        _physicalObjects.remove(motionState);
-        result.push_back(motionState);
-        _entitiesToRelease.insert(entity);
+        if (motionState) {
+            _pendingChanges.remove(motionState);
+            _outgoingChanges.remove(motionState);
+            _physicalObjects.remove(motionState);
+            result.push_back(motionState);
+            _entitiesToRelease.insert(entity);
+        }
```
Atlante45: Should there be an `else if` here?
I noticed when going through the `_entitiesToRelease`, we then add them to `_entitiesToDelete`.
So they would end up being added twice to `_entitiesToDelete`.

Atlante45: Nvm, I see that's a set.

PR #7070, opened 2016-02-10
don't sink through floor when server restarts

sethalves:
- when the domain restarts, don't sink a ways into the (now missing) floor

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -3877,6 +3877,8 @@ void Application::clearDomainOctreeDetails() {
     qCDebug(interfaceapp) << "Clearing domain octree details...";
     // reset the environment so that we don't erroneously end up with multiple
 
+    _physicsEnabled = false;
+
     // reset our node to stats and node to jurisdiction maps... since these must be changing...
     _entityServerJurisdictions.withWriteLock([&] {
         _entityServerJurisdictions.clear();


sethalves:
https://app.asana.com/0/26485730907942/89355451868995

hifi-gustavo:
Build console output for commit: 11d296a4777e4fe9ffba354ddbf2af67d2888437 available at: https://gist.github.com/72cfe4ba6305c1f4d5ac<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7070.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7070.dmg

sethalves:
to test:

log into a domain and stand on something solid.  restart the domain.  before this change, the avatar would fall for a moment and then start flying.  After this change, the avatar will remain in place until entities are loaded, again.

ZappoMan:
:+1:

cozza13:
tested and works

bluetoe:
I tried it a few times and it seems to work. It also seems to have fixed a problem I had seen where if a solid entity below the intended "ground" loaded before the real "ground", I would fall down to that level and then be stuck underneath the intended landing spot.

PR #7075, opened 2016-02-11
Fix "Add to scene" option while dropping model

Atlante45:
Fixes a bug where the model wouldn't actually be dropped in the scene.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -4333,8 +4333,8 @@ void Application::modelUploadFinished(AssetUpload* upload, const QString& hash) 
     auto filename = QFileInfo(upload->getFilename()).fileName();
 
     if ((upload->getError() == AssetUpload::NoError) &&
-        (upload->getExtension().endsWith(FBX_EXTENSION, Qt::CaseInsensitive) ||
-         upload->getExtension().endsWith(OBJ_EXTENSION, Qt::CaseInsensitive))) {
+        (FBX_EXTENSION.endsWith(upload->getExtension(), Qt::CaseInsensitive) ||
+         OBJ_EXTENSION.endsWith(upload->getExtension(), Qt::CaseInsensitive))) {
 
         auto entities = DependencyManager::get<EntityScriptingInterface>();
 


Atlante45:
https://app.asana.com/0/26551373415487/86331354879296

interface/src/Application.cpp
```diff
@@ -4333,8 +4333,8 @@ void Application::modelUploadFinished(AssetUpload* upload, const QString& hash)
     auto filename = QFileInfo(upload->getFilename()).fileName();
 
     if ((upload->getError() == AssetUpload::NoError) &&
-        (upload->getExtension().endsWith(FBX_EXTENSION, Qt::CaseInsensitive) ||
-         upload->getExtension().endsWith(OBJ_EXTENSION, Qt::CaseInsensitive))) {
```
Atlante45: Constants have dots included but not `getExtension()` so I simply switched them.

hifi-gustavo:
Build console output for commit: 913172b4908e3d67b57e2b7c25e5ab212c8be382 available at: https://gist.github.com/83c7cb1083f9a445ee59<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7075.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7075.dmg

interface/src/Application.cpp
```diff
@@ -4333,8 +4333,8 @@ void Application::modelUploadFinished(AssetUpload* upload, const QString& hash)
...
+        (FBX_EXTENSION.endsWith(upload->getExtension(), Qt::CaseInsensitive) ||
+         OBJ_EXTENSION.endsWith(upload->getExtension(), Qt::CaseInsensitive))) {
```
huffman: This reads like a yoda conditional

OmegaHeron:
Tested windows 7 against production build 4172.  Dragging/dropping a model to this PR works - model is placed in scene (if option to do so is checked).

interface/src/Application.cpp
Atlante45: Indeed, like a yoda conditional it reads.

PR #7077, opened 2016-02-11
Clean up grab signaling + various fixes

sethalves:
- change handControllerGrab entityMethod calls to include hand and avatar
- remove handControllerGrab setHand calls
- fix bug that was causing toybox bow string to drift when bow was equipped and moving
- fix bug that caused entity-server to sometimes send a child-entity but not its parent if the child was larger
- fix a couple bugs related to velocity moving into SpatiallyNestable

### DIFF
--- a/libraries/shared/src/SpatiallyNestable.cpp
+++ b/libraries/shared/src/SpatiallyNestable.cpp
@@ -101,13 +101,11 @@ SpatiallyNestablePointer SpatiallyNestable::getParentPointer(bool& success) cons
     // we have a _parentID but no parent pointer, or our parent pointer was to the wrong thing
     QSharedPointer<SpatialParentFinder> parentFinder = DependencyManager::get<SpatialParentFinder>();
     if (!parentFinder) {
-        qDebug() << "SpatiallyNestable::getParentPointer -- no parentFinder";
         success = false;
         return nullptr;
     }
     _parent = parentFinder->find(parentID, success);
     if (!success) {
-        qDebug() << "parentFinder failed for " << parentID;
         return nullptr;
     }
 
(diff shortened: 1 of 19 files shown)


sethalves:
to test:
The best test for this PR is to use the examples/toybox/bow/createbow.js script (make sure to run the one from this PR, not the one from s3) and play with the bow.  Try equipping in either hand, try tossing it back and forth.  make sure it works and that the string stays in place.

Also try:
(1) grab pistol
(2) equip pistol in other hand
(3) release pistol with first hand
(4) try firing piston with second hand's trigger

sethalves:
https://app.asana.com/0/26485730907942/89451593895214

hifi-gustavo:
Build console output for commit: ffce38769f621596a416635dff58a5fa9c202dba available at: https://gist.github.com/e294dc16b39da5acdfdb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7077.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7077.dmg

libraries/shared/src/SpatiallyNestable.cpp
```diff
@@ -672,6 +690,16 @@ QList<SpatiallyNestablePointer> SpatiallyNestable::getChildren() const {
...
 
+bool SpatiallyNestable::hasChildren() const {
+    bool result = false;
+    _childrenLock.withReadLock([&] {
+        if (_children.size() > 0) {
+            result = true;
+        }
+    });
+    return result;
+}
```
AndrewMeadows: **_children** is a **std::vector** so I would expect its **size()** method to be read-only and fast.  Which means, this lock is probably unnecessary.  It certainly doesn't prevent the truthiness of **hasChildren()** from changing before the **result** can actually be used.

In short: the implementation of **hasChildren()** seems more complex and heavy than it needs to be.

sethalves: I'm going to leave the lock, because even if I convinced myself that this is safe, it might not remain safe in future versions.

AndrewMeadows:
I tested using local scripts and logged onto Seth's eschatology domain.  The bow string looks good.

PR #7087, opened 2016-02-11
yet another attempt to avoid physics crashes

AndrewMeadows:
- prevent crash in physics by checking for nullptr (rather than assert in debug)

### DIFF
--- a/libraries/physics/src/PhysicsEngine.cpp
+++ b/libraries/physics/src/PhysicsEngine.cpp
@@ -145,27 +145,30 @@ void PhysicsEngine::removeObjects(const VectorOfMotionStates& objects) {
     // then remove them
     for (auto object : objects) {
         btRigidBody* body = object->getRigidBody();
-        assert(body);
-        _dynamicsWorld->removeRigidBody(body);
+        if (body) {
+            _dynamicsWorld->removeRigidBody(body);
 
-        // NOTE: setRigidBody() modifies body->m_userPointer so we should clear the MotionState's body BEFORE deleting it.
-        object->setRigidBody(nullptr);
-        body->setMotionState(nullptr);
-        delete body;
+            // NOTE: setRigidBody() modifies body->m_userPointer so we should clear the MotionState's body BEFORE deleting it.
+            object->setRigidBody(nullptr);
+            body->setMotionState(nullptr);
+            delete body;
+        }
     }
 }
 
 // Same as above, but takes a Set instead of a Vector.  Should only be called during teardown.
 void PhysicsEngine::removeObjects(const SetOfMotionStates& objects) {
+    _contactMap.clear();
     for (auto object : objects) {
         btRigidBody* body = object->getRigidBody();
-        assert(body);
-        _dynamicsWorld->removeRigidBody(body);
+        if (body) {
+            _dynamicsWorld->removeRigidBody(body);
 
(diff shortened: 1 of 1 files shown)


libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -145,27 +145,30 @@ void PhysicsEngine::removeObjects(const VectorOfMotionStates& objects) {
...
 
-        // NOTE: setRigidBody() modifies body->m_userPointer so we should clear the MotionState's body BEFORE deleting it.
-        object->setRigidBody(nullptr);
-        body->setMotionState(nullptr);
-        delete body;
+            // NOTE: setRigidBody() modifies body->m_userPointer so we should clear the MotionState's body BEFORE deleting it.
+            object->setRigidBody(nullptr);
+            body->setMotionState(nullptr);
+            delete body;
+        }
```
AndrewMeadows: I thought I knew, or could figure out, how the MotionState's RigidBody could have been set to **nullptr** in this context, hence I was using an assert.  The fact of the matter is there is some path by which it can be null.  So, rather than assert I just check for and explicity avoid null RigidBodies.

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -145,27 +145,30 @@ void PhysicsEngine::removeObjects(const VectorOfMotionStates& objects) {
...
     }
 }
 
 // Same as above, but takes a Set instead of a Vector.  Should only be called during teardown.
 void PhysicsEngine::removeObjects(const SetOfMotionStates& objects) {
+    _contactMap.clear();
```
AndrewMeadows: Clear the contact map on teardown.

AndrewMeadows:
I don't know how to reproduce this crash.  We saw about 30+ physics crashes during our internal meeting today.  Sometimes it was during login, sometimes after hanging around for several minutes.  Unless someone has a semi-reliable crashy experience my advice: let's merge it and let the crash report graph tell us if this makes a difference.

hifi-gustavo:
Build console output for commit: 2b91a3191919eb71861e951fb6f62dcf765c7d24 available at: https://gist.github.com/607c41bc939a0e9c235e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7087.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7087.dmg

AndrewMeadows:
test this please

AndrewMeadows:
The internal asana issue for this is here:  https://app.asana.com/0/26623005675174/89525346205621

hifi-gustavo:
Build console output for commit: 2b91a3191919eb71861e951fb6f62dcf765c7d24 available at: https://gist.github.com/607c41bc939a0e9c235e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7087.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7087.dmg

cozza13:
I tested this and did not crash. I went through the same process with the current release build and crashed.

PR #7090, opened 2016-02-11
move back to unordered_set so that OctreeElementBag will handle uniqueness

ZappoMan:
Restores behavior in Entity Server to ensure proper handling of scene sending.


ZappoMan:
@Atlante45 @AndrewMeadows - please review this... moving back to an unordered_set<> (std, instead of QSet)... I smoke tested this... (ran a server, connected client, saw content, did some editing, etc) -- but haven't done thorough testing.

The note from the stack overflow article does alarm me a little bit... when will unordered_set<> rehash? And what will the result be when it rehashes an expired element... I would think that the result would be you'd basically end up merging expired elements, which would be acceptable.

Maybe we need to construct some unit tests that show how this would work.

hifi-gustavo:
Build console output for commit: 26ab1bdac4ea84726c2726f5f24bb00b31c1bf6f available at: https://gist.github.com/ca5ac454694330cc0673<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7090.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7090.dmg

Atlante45:
@ZappoMan Do we absolutely need the set to be unordered?
It seems to me like it is not necessary, and would remove the danger of rehashing.

According to this post: http://stackoverflow.com/questions/23210092/is-it-safe-to-use-a-weak-ptr-in-a-stdset-or-key-of-stdmap
Using `std::map` and `std::owner_less` is perfectly safe for a set of weak pointers.
Looks like that's even why `owner_less` was created: http://en.cppreference.com/w/cpp/memory/owner_less

ZappoMan:
@Atlante45 set doesn't need to be unorderd... but presumably unordered is faster.

Atlante45:
@ZappoMan If we want to stay with unordered, there's something else we can do that I'd considered before I went with the queue.

It's to use an `std::unordered_map` and store `{ weakPtr.lock().get(), weakPtr }` (obviously we'd check the shared pointer before inserting)

ZappoMan:
I will try that... it seems like `std::unorderd_set<>` should be similar to `std::unordered_map<>`

libraries/octree/src/OctreeElementBag.cpp
```diff
@@ -34,8 +32,9 @@ OctreeElementPointer OctreeElementBag::extract() {
 
     // Find the first element still alive
     while (!result && !_bagElements.empty()) {
-        result = _bagElements.front().lock(); // Grab head's shared_ptr
-        _bagElements.pop();
+        Bag::iterator it = _bagElements.begin();
+        result = (*it).lock(); // Grab head's shared_ptr
+        _bagElements.erase(it);
     }
```
AndrewMeadows: std::unordered_set::erase() returns the next iterator so an alternative implementation would be:

    Bag::iterator it = _bagElements.begin();
    while (it != _bagElements.end() && !result) {
        result = (*it).lock();
        it = _bagElements.erase();
    }

AndrewMeadows:
I like the idea of std::unordered_map<>.  When I considered fixing this that is what I had in mind.

I too don' t fully understand the consequences of rehashing the set after elements had expired.

ZappoMan:
@Atlante45 @AndrewMeadows - I've updated it to use std::unordered_map<> -- I think this is the way to go.

hifi-gustavo:
Build console output for commit: 477dfe1b103fdaa99a3ac9a294ee73fc9e9fb550 available at: https://gist.github.com/ca5ac454694330cc0673<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7090.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7090.dmg

libraries/octree/src/OctreeElementBag.cpp
```diff
@@ -16,26 +16,25 @@ void OctreeElementBag::deleteAll() {
...
 }
 
+/// does the bag contain elements?
+/// if all of the contained elements are expired, they will not report as empty, and
+/// a single last item will be returned by extract as a null pointer
 bool OctreeElementBag::isEmpty() {
-    // Pop all expired front elements
-    while (!_bagElements.empty() && _bagElements.front().expired()) {
-        _bagElements.pop();
-    }
```
Atlante45: This is still needed or we might have a non empty bag we can't extract from.

Atlante45: I should mention too that this bag is by definition not thread safe, and that this case might happen.
It almost feels like we should remove `isEmpty`.

Atlante45: Actually, we NEED to remove that function, it doesn't make any sense to have it.

ZappoMan: 1) it's not needed because the callers are told that they can not assume they won't get an empty item when they extract, I updated the docs, and verified that all callers already check the result. So no... this is wasteful and should not be added back.

2) isEmpty() is needed, and useful, it's used throughout the code. Removing it would require major logic changes, and I'm not doing that.

3) Correct this class is not thread safe... just like all the Qt and std container classes... and in this case, we only ever use it on a single thread... so it's ok.

ZappoMan:
I'm not sure if this handles the case of the weak_ptr expiring, and then the actual memory getting reused in a new weak_ptr, and then having that get inserted before the expiration of the old item is detected... that case might cause problems. It's possible in that case that the insert would fail silently.

ZappoMan:
@Atlante45 says it will work. ;)

Atlante45:
To be fair, it'll only work as long as the std lib used respects the standard and ccpref didn't make a mistake ;)
http://en.cppreference.com/w/cpp/container/unordered_map/operator_at

ZappoMan:
QA Testing Notes: The change will impact the sending behavior of the entity server. Doing various smoke tests to make sure that when users connect and move around the scene they get all the proper content. Do testing of editing with multiple viewers, etc.

AndrewMeadows:
I've got a test case for which I would like to see if these changes make a difference, so I'll checkout this branch...

AndrewMeadows:
The test case was as follows:

(1) In a domain with 4.7k static boxes (all not too far from origin, 100m max)...
(2) Login with two users, each with **Developer --> Physics --> Highlight Simulation Ownership** enabled
(3) UserA creates 200 dynamic boxes in a pile
(4) UserB spins around several times and then stops to watch the dynamic pile...

--> With the the old server UserB would see the boxes sitting there without any updates for several seconds.

--> With the new server the updates on the pile recover pretty fast: noticeably slow for a second or two and then increase to normal quiescent rate.

The work in this PR definitely makes updates flow faster which helps.  I'm still seeing problems in the distributed physics simulation which is not surprising...

In particular UserB will sometimes win simulation ownership of some of the pile because the entity-server will automatically clear it if UserA doesn't update fast enough (2 sec) and this can happen when the UserA has a big pile (lots of updates to send) and many of the objects are "active but not moving" (update rate per object decreases).  If UserA were to suddenly double the size of the pile then things get significantly worse: ownership overlap and inability of the entity-sever to send the full state of the pile makes for some bad physics disagreements between the two owners/observers.

PR #7101, opened 2016-02-12
fix error in preload logic

ZappoMan:
This fixes the mistake where preloads were sometimes not being sent out when they should be.

### DIFF
--- a/libraries/entities/src/EntityItem.h
+++ b/libraries/entities/src/EntityItem.h
@@ -404,18 +404,12 @@ public:
 
     virtual void loader() {} // called indirectly when urls for geometry are updated
 
-    /// Has the entity been added to an entity tree? This is useful in some state decisions
-    /// for example whether or not to send out a scriptChanging signal.
-    void markAsAddedToTree() { _isAddedToTree = true; }
-    bool isAddedToTree() const { return _isAddedToTree; }
-
     /// Should the external entity script mechanism call a preload for this entity.
     /// Due to the asyncronous nature of signals for add entity and script changing
     /// it's possible for two similar signals to cross paths. This method allows the
     /// entity to definitively state if the preload signal should be sent.
-    bool shouldPreloadScript() const { return !_script.isEmpty() &&  
-                                _loadedScript != _script && _loadedScriptTimestamp != _scriptTimestamp; }
-
+    bool shouldPreloadScript() const { return !_script.isEmpty() && 
+                                               _loadedScript != _script && _loadedScriptTimestamp != _scriptTimestamp; }
     void scriptHasPreloaded() { _loadedScript = _script; _loadedScriptTimestamp = _scriptTimestamp; }
 
 protected:
@@ -458,8 +452,15 @@ protected:
     float _restitution;
     float _friction;
     float _lifetime;
-    QString _script;
-    quint64 _scriptTimestamp;
+
+    QString _script; /// the value of the script property
+    QString _loadedScript; /// the value of _script when the last preload signal was sent
+    quint64 _scriptTimestamp{ ENTITY_ITEM_DEFAULT_SCRIPT_TIMESTAMP }; /// the script loaded property used for forced reload
+
+    /// the value of _scriptTimestamp when the last preload signal was sent
+    // NOTE: on construction we want this to be different from _scriptTimestamp so we intentionally bump it
+    quint64 _loadedScriptTimestamp{ ENTITY_ITEM_DEFAULT_SCRIPT_TIMESTAMP + 1 };
+
     QString _collisionSoundURL;
     glm::vec3 _registrationPoint;
     float _angularDamping;
@@ -520,12 +521,6 @@ protected:
     mutable QHash<QUuid, quint64> _previouslyDeletedActions;
 
     QUuid _sourceUUID; /// the server node UUID we came from
-
-    bool _isAddedToTree { false }; /// this is false until the entity has been added to an entity tree
-
-    QString _loadedScript; /// the value of _script when the last preload signal was sent
-    quint64 _loadedScriptTimestamp { 0 };  /// the value of _scriptTimestamp when the last preload signal was sent
-
 };
 
 #endif // hifi_EntityItem_h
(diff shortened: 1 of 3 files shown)


ZappoMan:
Dev Notes:
- fixed the "shouldPreload" logic to handle the case of freshly created entities that have never had a manual "reload" done on their individual script.
- removed the whole added to tree logic, that was not actually making a positive difference since the "shouldPreload" logic handles that case as well.

ZappoMan:
Testing Notes:
- This should fix this issue: https://app.asana.com/0/26320649018758/90090246016012
- AND verify it didn't reintroduce this issue: https://app.asana.com/0/26225263936266/84176462198297

Recommended tests:
- create an an entity with a preload method from a script (like the cell science builder) -- these entities will have a 0 "scriptTimestamp" property -- those were the ones that weren't getting their preloads run.
- create an entity and set it's script from the edit.js ui -- make sure it's preloads work
- create an entity, and use the edit.js UI to "force a reload" for that single script. -- make sure it's preloads still work
- try "reload-all scripts" and make sure preloads still work as expected.

libraries/entities/src/EntityItem.h
```diff
@@ -404,18 +404,12 @@ class EntityItem : public SpatiallyNestable, public ReadWriteLockable {
 
     virtual void loader() {} // called indirectly when urls for geometry are updated
 
-    /// Has the entity been added to an entity tree? This is useful in some state decisions
-    /// for example whether or not to send out a scriptChanging signal.
-    void markAsAddedToTree() { _isAddedToTree = true; }
```
ZappoMan: removed this stuff since it wasn't helping, and only adds complexity

libraries/entities/src/EntityItem.h
```diff
@@ -458,8 +452,15 @@ class EntityItem : public SpatiallyNestable, public ReadWriteLockable {
...
     float _lifetime;
-    QString _script;
-    quint64 _scriptTimestamp;
+
+    QString _script; /// the value of the script property
+    QString _loadedScript; /// the value of _script when the last preload signal was sent
+    quint64 _scriptTimestamp{ ENTITY_ITEM_DEFAULT_SCRIPT_TIMESTAMP }; /// the script loaded property used for forced reload
+
+    /// the value of _scriptTimestamp when the last preload signal was sent
+    // NOTE: on construction we want this to be different from _scriptTimestamp so we intentionally bump it
```
ZappoMan: NOTE: this is the actual bug fix. The problem was that for entities that were created with a "script" property at the time of creation (not changed in edit.js) -- the `_scriptTimestamp` would be 0 and we initialized `_loadedScriptTimestamp` was also 0 and so our `shouldPreloadScript()` logic would never return true

libraries/entities/src/EntityTree.cpp
```diff
@@ -92,9 +92,6 @@ void EntityTree::postAddEntity(EntityItemPointer entity) {
 
     // find and hook up any entities with this entity as a (previously) missing parent
     fixupMissingParents();
-
-    // indicate that the entity has been added to the tree
-    entity->markAsAddedToTree();
```
ZappoMan: removed to reduce complexity

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -927,11 +927,8 @@ int EntityTreeElement::readElementDataFromBuffer(const unsigned char* data, int
                     bool reload = entityScriptTimestampBefore != entityScriptTimestampAfter;
 
                     // If the script value has changed on us, or it's timestamp has changed to force
-                    // a reload then we want to send out a script changing signal... UNLESS this is the
-                    // first time we've read the entity information, in which case, the entity has
-                    // not yet been added to the tree, and we want to wait for the addEntity. In other
-                    // words it's not a "change" in the script, if it's a first time load.
-                    if (entityItem->isAddedToTree() && (entityScriptBefore != entityScriptAfter || reload)) {
```
ZappoMan: removed to reduce complexity

hifi-gustavo:
Build console output for commit: 550b9a98b1c6dedb33645cb6ee4fa8c40b0fc37e available at: https://gist.github.com/a656aae1d10336afa7ae<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7101.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7101.dmg

libraries/entities/src/EntityItem.h
```diff
@@ -404,18 +404,12 @@ class EntityItem : public SpatiallyNestable, public ReadWriteLockable {
...
-
     /// Should the external entity script mechanism call a preload for this entity.
     /// Due to the asyncronous nature of signals for add entity and script changing
     /// it's possible for two similar signals to cross paths. This method allows the
     /// entity to definitively state if the preload signal should be sent.
-    bool shouldPreloadScript() const { return !_script.isEmpty() &&  
-                                _loadedScript != _script && _loadedScriptTimestamp != _scriptTimestamp; }
-
+    bool shouldPreloadScript() const { return !_script.isEmpty() && 
+                                               _loadedScript != _script && _loadedScriptTimestamp != _scriptTimestamp; }
```
ZappoMan: I think this logic is wrong... hold off on testing

ZappoMan: ok - fixed it... good to go!

hifi-gustavo:
Build console output for commit: 9160758af7eb4786bccfc0e35708bf18dfc46a4d available at: https://gist.github.com/a656aae1d10336afa7ae<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7101.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7101.dmg

imgntn:
I've tested all 4 of Brad's test cases for loading entities (via script, script url on entity, force reload on entity, or reload all) and they all pass.  Additionally, I've visited cellscience and things seem to work as expected (buttons appear, scripts that rely on preload work) -- win10

PR #7102, opened 2016-02-12
Revert model texs when unset

zzmp:


zzmp:
Fixes a bug where models with their textures set would not revert to their originalTextures when they were unset.

##### Testing
For a renderable model, change the textures property (available through edit.js) to use a different texture (used textures can be seen in the originalTextures property). The texture should change on that model only.
Remove the texture, and it should revert to those values in originalTextures.

hifi-gustavo:
Build console output for commit: 0903e14b066fce0d2d5830970cc2fbfa39984806 available at: https://gist.github.com/6b0696c824a59a8791a5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7102.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7102.dmg

zzmp:
test this please

hifi-gustavo:
Build console output for commit: 0903e14b066fce0d2d5830970cc2fbfa39984806 available at: https://gist.github.com/6b0696c824a59a8791a5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7102.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7102.dmg

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -101,8 +101,9 @@ int RenderableModelEntityItem::readEntitySubclassDataFromBuffer(const unsigned c
 }
 
 QVariantMap RenderableModelEntityItem::parseTexturesToMap(QString textures) {
+    // If textures are unset, revert to original textures
     if (textures == "") {
-        return QVariantMap();
+        return parseTexturesToMap(_originalTextures.join(",\n"));
```
ZappoMan: is there any possibility that original textures could be "" and this would cause an infinite recursion? Maybe you need a double check that if `_originalTextures == ""` to still return the empty `QVariantMap()`

bluetoe:
I tried this but, when I put a different texture URLs in the textures field the visible texture changed to grey and never seemed to load. But yes, when I deleted the texture URLs from the textures field it reverted back to the original textures from the Original Textures field.

OmegaHeron:
Placed model with embedded texture then copied it - changed copy's texture to another and it changed proper while original maintained its proper texture.  Removed text from field to specify texture and model reverted to original.  I did make an initial mistake leading to no texture displayed then looked at log and saw I had may a typo in URL.  For me, this works proper.  

At some point a more robust notification that a asset couldn't be loaded via http/atp really needs to be added.

OmegaHeron:
I want to do one more test on this - model with material assigned, but, no texture specified for material.  That may exhibit different result.

OmegaHeron:
@zzmp - I found a crash case in this, documenting now - it's a bit complex and will take a few to write up.

OmegaHeron:
Ok - this is something that might only hit one who's modeled for SL/OS too long.  Take this example;

A cube UV with unwrap to apply a single texture with a single material.  Do not specify a texture to material, simply reference material as having a texture.  Place model in scene and it'll be a simple grey cube and show as having no texture assigned.  Apply a reference to a texture and this PR crashes and will crash each time until that model is removed.  Current production build (fortunately) does not so was able to use it to remove model.  This may seem an uncommon thing to do - uv unwrap and not have an initial texture assigned, but, it's pretty much the "standard" way of doing things in SL/OS.  While I consider the model made that leads to this to be an invalid model - it will need to be handled somehow without a crash.

Here's a copy of model that leads to crash and a texture reference you can assign.  Due to filtering rules on server you'll only be able to place model/use textures from Interface.

https://asset-east.o2tech.info/textest/cubeTest3.fbx

You can use any one of following to apply as a single texture like;
Texture:"https://asset-east.o2tech.info/textest/cubeTex1.png"

https://asset-east.o2tech.info/textest/cubeTex1.png
https://asset-east.o2tech.info/textest/cubeTex2.png
https://asset-east.o2tech.info/textest/cubeTex3.png
https://asset-east.o2tech.info/textest/cubeTex4.png

Here's a cube with an initial texture assigned and embedded;
https://asset-east.o2tech.info/textest/cubeTest1.fbx

zzmp:
@OmegaHeron I'm unable to load from https://asset-east.o2tech.info, I get a 403. Can you provide other examples?

OmegaHeron:
I'll make a zip file for you and attach in a moment - that's my server disallowing downloads via anything but Interface.

OmegaHeron:
@zzmp This should contain what's needed - sorry about that.

[testAsset.zip](https://github.com/highfidelity/hifi/files/131297/testAsset.zip)

OmegaHeron:
Sorry - *this* should contain what you need - cubeTest1.fbx has texture assigned proper and works proper not matter what.  cubeTest3.fbx has material/uv assigned but no initial texture and leads to crash if you attempt to apply a texture.

[testAssets.zip](https://github.com/highfidelity/hifi/files/131301/testAssets.zip)

zzmp:
I didn't realize what you'd meant by filtering rules. Thanks for the good repro. Looking at it now.

zzmp:
@OmegaHeron I could not reproduce any crash. cubeTest3.fbx could not have its `Texture` set, because it was not defined to have one.

When a model is first uploaded, its textures are assigned names based on the file uploaded. cubeTest1.fbx has a Texture, so it can be reassigned. cubeTest3.fbx does not, so although the debug log shows that it is trying to set the Texture, it cannot because the material does not associate "Texture" with any of the textures. Is this what you're referring to? Is there other behavior that I'm unable to reproduce?

zzmp:
@OmegaHeron I just realized that I added one more commit to guard against infinite recursion, and it's possible that you triggered that to crash. Let me revert my local commit and see if that was it, and if it is I can push it up to get retested.

OmegaHeron:
In my case if I attempt to assign a texture to cubeTest3.fbx it crashes to desktop with this PR build - it attempts to send a bugsplat report which blocks running VS debug - This is on windows 7.  I'm not prevented from trying to assign texture or otherwise placing invalid data in.  I agree that cubeTest3.fbx is really a screwed up model, but, it will be an issue with people used to getting away with things in another metaverse.  I'll pull this branch and run in debugger to get a stack trace.

OmegaHeron:
Cool - thank you and I'll hold on pulling branch/compiling until we try this.

zzmp:
Yep - this should fix the crash. The behavior that I mentioned still holds.

Do you think that this behavior is wrong?

OmegaHeron:
No, your behavior is probably correct - here's the mechanism of how you get a model like this though.

In SL/OS a model with a UV unwrap space is assumed to get a texture mapped to a flat projection in UV space - you can then arbitrarily assign any texture to that material (or in SL/OS speak [bad], face).  There's no real concept in modeling of assigning an initial texture even though we usually do.  When you import model it begins as a texture free model then you assign textures to each material.  What I was looking at today was something of an attempt to go from that sort of basis.  The cube is uv unwrapped and has a material set to accept a texture but not reference to a texture asset is applied to it.  The main thing is it not crash if it gets a model like this and it seems you've covered that.

hifi-gustavo:
Build console output for commit: 0e9d6968596584007ffbd5682ba6000d0f9e88c3 available at: https://gist.github.com/6b0696c824a59a8791a5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7102.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7102.dmg

OmegaHeron:
This works proper now for "borked" model and, for me, allows a texture to be set and reverted as the point of this PR is.

PR #7106, opened 2016-02-15
tighter bounding capsule for avatars

AndrewMeadows:
- reduced radius for collision capsule of most avatars

### DIFF
--- a/libraries/animation/src/Rig.cpp
+++ b/libraries/animation/src/Rig.cpp
@@ -1274,15 +1274,18 @@ void Rig::computeAvatarBoundingCapsule(
     // even if they do not have legs (default robot)
     totalExtents.addPoint(glm::vec3(0.0f));
 
-    int numPoses = (int)finalPoses.size();
-    for (int i = 0; i < numPoses; i++) {
-        const FBXJointShapeInfo& shapeInfo = geometry.joints.at(i).shapeInfo;
-        AnimPose pose = finalPoses[i];
+    // HACK to reduce the radius of the bounding capsule to be tight with the torso, we only consider joints
+    // from the head to the hips when computing the rest of the bounding capsule.
+    int index = _animSkeleton->nameToJointIndex(QString("Head"));
+    while (index != -1) {
+        const FBXJointShapeInfo& shapeInfo = geometry.joints.at(index).shapeInfo;
+        AnimPose pose = finalPoses[index];
         if (shapeInfo.points.size() > 0) {
             for (int j = 0; j < shapeInfo.points.size(); ++j) {
                 totalExtents.addPoint((pose * shapeInfo.points[j]));
             }
         }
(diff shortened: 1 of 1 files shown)


AndrewMeadows:
To test:

(1) Login.
(2) Enable visibility of collision capsule: **Developer-->Avatar-->Show Bounding Collision Shapes**
(3) Observe the height and radius of the bounding capsule: it should be tight with the avatar's torso (shoulders, arms, and feet may extend outside the capsule) but should still extend from feet to head.

AndrewMeadows:
Internal asana task is:

https://app.asana.com/0/26623005675174/82512045875259

hifi-gustavo:
Build console output for commit: d6a716a1aa5f7619cb0b651be53730fb9bc86df0 available at: https://gist.github.com/8b636c883ee06844cd90<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7106.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7106.dmg

judasshuffle:
https://i.gyazo.com/cc5b92b76dc22e118a81be05739397a2.png  Doesn't seem much different

OmegaHeron:
I see a difference using "Owen" avatar

Production build 4206;

hifi-snap-by-omegaheron-on-2016-02-15_14-43-55

This pr;

hifi-snap-by-omegaheron-on-2016-02-15_14-42-49

It seems somewhat oddly centered in this PR, but, I need to look at it again with different cam settings to be sure.

OmegaHeron:
I'm not going to quibble over what may just be my visual interpretation vs what it's actually doing other than to say it looks a tiny bit offset from apparent visualized center on Owen.  It does - for this AV - present a tighter collision bounds.

hifi-snap-by-omegaheron-on-2016-02-15_14-49-39

AndrewMeadows:
I'm pretty sure that offset is a bug in how we render the shape, not how the shape actually collides.

OmegaHeron:
Yeah, it seems so.  I made a tight corridor and walked AV thru it with current and proposed shapes - it's a bit subjective, but, it seems in both cases the center of visual AV is a bit off from center of physics and it's variant depending upon AV.  Fun fun.  :)

AndrewMeadows:
I checked the code --> we're handling the offset correctly for render so it isn't that.

The tests above were done at <2k, 15k, 2k>.  I traveled that far out and tried it with the same avatar --> didn't see the problem.  It looked a little bit worse than at at <0, 0, 0>, but not as bad as the snapshots.

In any case, I don't think the offset is actually NEW to this PR (is it?).  If so then it shouldn't block this work -- we'll fix it in a different PR.

OmegaHeron:
No, I don't think it's new - it was more just being thorough in noting a perceived thing.

PR #7108, opened 2016-02-15
More precision for edit.js

imgntn:
Before, most of the values in edit.js were fixed at two decimal places. Now, most are fixed at 4 decimal places. Additionally, I've adjusted the step values of certain inputs to allow for slightly more precision when using the up or down keys to change the value (position, rotation, dimensions have a step of 0.1)

To test:

http://rawgit.com/imgntn/hifi/moreprecision/examples/edit.js

### DIFF
--- a/examples/html/entityProperties.html
+++ b/examples/html/entityProperties.html
@@ -64,7 +64,7 @@
         function createEmitNumberPropertyUpdateFunction(propertyName) {
             return function() {
                 EventBridge.emitWebEvent(
-                    '{ "type":"update", "properties":{"' + propertyName + '":' + parseFloat(this.value).toFixed(2) + '}}'
+                    '{ "type":"update", "properties":{"' + propertyName + '":' + Number(this.value.toFixed(4)) + '}}'
                 );
             };
         }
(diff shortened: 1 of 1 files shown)


imgntn:
https://app.asana.com/0/46580934041217/82497611444858

hifi-gustavo:
Build console output for commit: 72186f916c84523072e2b88edae0dd5f73ed65d5 available at: https://gist.github.com/ec227b084f649d5c19e7<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7108.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7108.dmg

OmegaHeron:
Does it not show the added 2 digits of precision right of decimal in edit.js properties window?  It seems to work setting something to 100.0050 in the I see a tiny move and it doesn't change posit to 100.01

imgntn:
@OmegaHeron sorry, wrong testing link.  updated now

OmegaHeron:
@imgntn - Thanks, retesting shortly.

OmegaHeron:
Ok - I see my 4 points of precision now.  I have a really odd behavior though.  I place a cube then cursor in X position, if I use arrow keyes (up/down) I can only move entity 3 units from its initial point - for Y I can only move it 2 units and Z allows me to arrow up/down as much as I like.

OmegaHeron:
If someone else wants to see if they can reproduce my weird issue here's flow.

Load edit.js from link here
Place a cube entity
Place cursor in x position and attempt to move with arrow keys, then y then z

In my really weird case I'm unable to move location in X/Y more than 2 or 3 ticks.

cozza13:
I just tested this and it worked for me

OmegaHeron:
I know this is closed/merged, but, I found source of issue I had/have with this.  It occurs at extremes of range in domain - for example;  within a domain at 0,0,0 this works perfectly -- moving to extremes you see behavior I mentioned (inability to position with incremental keyboard input).  This makes sense in that you begin to run into single precision float issues as value becomes larger.

Go to a domain space and move to 0,0,0 - place a cube and move its position by using arrow keys in X,Y,Z fields, note that it works well.  Now move to 16000,16000,16000 and repeat.  Now move to -16000,-16000,-16000 and repeat.  It remains to be seen how far from 0,0,0 that things begin to get problematic.

OmegaHeron:
If x or y or z position >= 2048 issue manifests.

imgntn:
@OmegaHeron yeah, that's legit -- it's floating point drift that doesn't agree with the step size on those inputs.  i'm going to open a PR that changes it back to an integer step (sigh)

PR #7089, opened 2016-02-11
Update the scene on overlay3d updates

zzmp:
- Add/remove an item from the scene when it's drawOnHUD property is changed.
- Update ItemKey when an overlay edit changes the ItemKey.

Updating the ItemKey without creating a new shared ptr necessitated a new method for the scene, resort (to resort the item based on a new key).

PR #7113, opened 2016-02-15
Remove drawOnHUD property from 3d overlays

zzmp:
This property was only used in entityCameraTool.js (included by edit.js). It was used to show a 3d overlay in screen coordinates, such that it stayed in one spot on the screen.

It has been broken, and is not necessary for HMD editing. With updates to the overlays, overlay pipelining has become more complicated and overlay type-specific for 3d overlays, so it is being removed.

This PR incorporates #7089, and should wait on it to be merged first.

### DIFF
--- a/interface/src/ui/overlays/Text3DOverlay.h
+++ b/interface/src/ui/overlays/Text3DOverlay.h
@@ -41,10 +41,12 @@ public:
     float getRightMargin() const { return _rightMargin; }
     float getBottomMargin() const { return _bottomMargin; }
     xColor getBackgroundColor();
-    float getBackgroundAlpha() const { return _backgroundAlpha; }
+    float getTextAlpha() { return _textAlpha; }
+    float getBackgroundAlpha() { return getAlpha(); }
 
     // setters
     void setText(const QString& text) { _text = text; }
+    void setTextAlpha(float alpha) { _textAlpha = alpha; }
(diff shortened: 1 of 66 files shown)


hifi-gustavo:
Build console output for commit: 0d29f59e398b90cfc92d2c06e9abf50e2a40a1f3 available at: https://gist.github.com/a201f2f5403b187c5d23<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7113.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7113.dmg

interface/src/ui/overlays/Text3DOverlay.h
```diff
@@ -41,10 +41,12 @@ class Text3DOverlay : public Billboard3DOverlay {
     float getRightMargin() const { return _rightMargin; }
     float getBottomMargin() const { return _bottomMargin; }
     xColor getBackgroundColor();
-    float getBackgroundAlpha() const { return _backgroundAlpha; }
+    float getTextAlpha() { return _textAlpha; }
+    float getBackgroundAlpha() { return getAlpha(); }
```
AndrewMeadows: Does not block the PR but: hese **get** methods should be **const**.

PR #7122, opened 2016-02-16
Audio output buffer

howard-stearns:
Revise automatic sizing of audio output buffers as follows:
 * Default number of frames to buffer is 1. (Was 3. 1 works fine on some PC's.)
 * Automatic resizing is on by default. (Was on only for Mac.)
 * Automatic resizing does not persist the value, so restart goes back to explicit user choice (or default).
 * Operating system communication of newly sized results (automatic or otherwise) is done on the audio thread through a signal.


howard-stearns:
https://app.asana.com/0/26225263936266/81223375750105/f
https://app.asana.com/0/26225263936266/90857558486721

hifi-gustavo:
Build console output for commit: d328b82c3cb99d7869e3fe97efde810690b800d8 available at: https://gist.github.com/d00a53e9ddcf307cae2a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.dmg

judasshuffle:
resetting my settings to default I don't notice the buffer size changing in windows (audio buffer set to 1 sounds like a tractor) if you mean the audio output buffer(frames) higher works better for me

howard-stearns:
My bad for not providing testing notes. Sorry about that.

I. Checking explicit selection:
  1. Kill Interface.ini, or remove the line 'audioOutputBufferSize=...' and 'audioOutputStarveDetectionEnabled=...' 
  2. Open Settings->Audio.  "Output buffer Size (frames)" should be 1 and "Output Starve Detection" should be checked. (All platforms.)
  3. The log should report "Output Buffer capacity in frames:  1", and possibly higher values after that.
  4. Manually changing either the Output Buffer Size or Output Starve Detection in the preferences should be sticky, such that the value you set should be appear next time you open the window, even if you stop and restart Interface in between.  Additionally, regardless of whether Output Starve Detection is checked, there should be at least one "Output Buffer capacity in frames:  n" in the log with the value you chose, although this could also be followed by higher values.

II. Checking automatic increases:
1. There's no way to force this. You can try setting the Output Starve Detection Threshold to just 1, and the Output Buffer Size to 1, and see if your machine bumps it up higher. (But only when Output Starve Detection" is checked.)
2. If it does bump it up, you will only see it in the log. In addition to "Output Buffer capacity in frames:   n", there will be a "Starve detection threshold met, increasing buffer size to ..." However, the "Output Buffer Size (frames)" item in preferences will only be the value you explicitly set, not any auto increase.  (The whole idea is that this is the persistent preference, not the current dynamically determined value.)

howard-stearns:
DNM until I change some names:
  - deliberate gratuitous name so that new defaults happen for existing users
  - take that opportunity to supply names that are more perspicuous

howard-stearns:
Change above has been made. Ready to go.
Additional testing note: should now get new behavior without having to do step I.1., above.

hifi-gustavo:
Build console output for commit: 983a0ccb9b7b414bfba08eec8a20c1e1caf96944 available at: https://gist.github.com/d00a53e9ddcf307cae2a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.dmg

judasshuffle:
Not seeing any automatic increases in starve detection in either the log or the prefs dialogue.
or any threshold met messages in the log.
If I change any settings manually these changes are saved
re buffer size a low initial  default setting gives a poor experience, 3 is the minimum I can use, many alphas crank it up as high as 20 to get usable audio.
if I set it to the default 1 I get a crack crack crack sound before the audio appears to disconnect itself rather than the buffer increasing

howard-stearns:
OK, let's try this.

First, let's get a baseline by unchecking the automatic increase, and getting a subjective letter grade for each initial frame count: 1, 2, ... 5.  Grade A for excellent quality, F for fail. Make a note of what it sounds like (eg., pops, no sound at a all, etc.)

Then turn the threshold to just 1 (defaults to 3), the automatic increase on, and see if you can get it to increase (as reported in the log).

Also, what is your hardware and OS?

Thanks so much for your help on this. It's hard to swim in the dark without a test case!

OmegaHeron:
Ok - running production build where output buffer size = 3 I get proper sound, if I set it to 1 I get the well known broken sound, that's a baseline/known response.

Running this PR It begins at 1 as advertised - it does not adjust upward.  Now - thing is, there's none around using audio so all I have is echo local from server, perhaps that doesn't lead to application of dynamic settings.

howard-stearns:
Right. I used loopback sound for my testing. (Results were macbook pro automatically went from 1->3. Windows 10 desktop stayed good at 1.)

So, just confirming - in production you get:
  buffer size 1 - "F" pops
  buffer size 2 - "F"? pops?
  buffer size 3 - ok. "A"? "B"?
and the same results for this PR with Output Starve Detection unchecked?

And then with this PR, Output Starve Detection checked, Output Starve Detection Threshold = 1, and Output Starve detection Period 10000, you get no increase shown in the log from an initial Output Buffer Size of 1 or 2 (it will take 10 seconds to show up), and the same grade sound as above? What hardware/OS?

OmegaHeron:
Windows 7 - SoundBlaster Recon3di - Alienware M18xR2 laptop.

Production build buffer size < 3 F = 3+ A

This PR setup as shown here;

hifi-snap-by-omegaheron-on-2016-02-17_13-08-24

Enable mic, audio set to echo back from server - Audio quality initially = F.  Nothing seen in log about anything changing with audio - left open mic for 10+ minutes.

OmegaHeron:
I'm reluctant to bring this up again, but, I will one last time.  Windows audio is _very_ unforgiving compared to working with such on Mac OS X - it will happily let you do things you shouldn't.  The stuttering sound comes from a well known cause/effect, that being a request to set a playback latency value < what hardware can handle.  All those messages seen about audio device doesn't support rate X - making resampler aren't because a device can't set sample rate to 24k s/s - it's because device can't reliably service playback at the lowest possible latency that Interface requests in its open audio device call and returns an error due to that latency request value.  When Interface gets back an error on open it assumes it's sample rate.

Now that said - if you open a sound device on windows with a playback latency rate < what hardware reports is sane for it you get exactly what we get here - popping/crackling playback as, essentially, you're forcing the DAC to constantly reset - it's trying to convert samples and constantly being reset to maintain an unattainable latency request -- it hasn't played all the incoming samples, but, to maintain latency it MUST jump forward.  As in - plays a little jump pop crackle play a little more repeat.  Why does increasing buffers "fix" this - each buffer increases time from receipt of samples to playback push, in effect it adds latency allowing DAC to have a steady stream of samples to convert.  This is why there's such a wide range of claimed "good" buffer numbers.  My firewire connected Presonus Firebox audio device needs 2, internal soundblaster needs 3, USB attached generic device needs 4 and bluetooth stereo headset needs 7.

howard-stearns:
test this please

howard-stearns:
There's some deactivated code for picking format that "works for me", so I've re-activated it. Also added some diagnostic logging. This is mostly for me to try this out on another machine tonight, but I welcome any other test/log results on the 547d3d6 build.

howard-stearns:
By the way, the comment on the formerly-deactivated code was that it swapped channels. I used utilities/diagnostics/orbitingSound.js to check that left/right channels were correct on the new version.

hifi-gustavo:
Build console output for commit: 547d3d6b9ad87fdd956a8eafc297a96d94e15200 available at: https://gist.github.com/d00a53e9ddcf307cae2a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.dmg

judasshuffle:
Tested again, still no automatic tuning of the settings. I did note that manual changes have to be saved to be applied.
Also I have been monitoring  audio looped back from the server
[02/18 15:02:03] [DEBUG] The audio output device "Speakers (AudioQuest DragonFly)" is available.
[02/18 15:02:03] [DEBUG] The format to be used for audio output is QAudioFormat(24000Hz, 16bit, channelCount=2, sampleType=SignedInt, byteOrder=LittleEndian, codec="audio/pcm")
[02/18 15:02:03] [DEBUG] No resampling required for network output to match actual output format
throws in any audio stats lol
https://i.gyazo.com/9b953f9b6b2ea56199dbe26633159ef9.png

hifi-gustavo:
Build console output for commit: 162be13914f1c44843fbc891e33c44825fc59751 available at: https://gist.github.com/d00a53e9ddcf307cae2a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.dmg

hifi-gustavo:
Build console output for commit: 162be13914f1c44843fbc891e33c44825fc59751 available at: https://gist.github.com/8be6486c3d3c24dc1bd8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.dmg

hifi-gustavo:
Build console output for commit: adb33e7dcaf682d2c3fcb177364e11080ba83b05 available at: https://gist.github.com/d00a53e9ddcf307cae2a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.dmg

howard-stearns:
I've got some extra debugging in right now, so DNM.

howard-stearns:
@judasshuffle and @OmegaHeron - can you guys try this with an initial output buffer size of 3, and send me your logs? Thanks.

OmegaHeron:
Added as my first item for Friday morning.  Time zones suck.  :)

OmegaHeron:
With latest build of this PR - Windows 7 audio setup as;

hifi-snap-by-omegaheron-on-2016-02-19_10-31-38

Log;

[hifi-log.txt](https://github.com/highfidelity/hifi/files/138247/hifi-log.txt)

howard-stearns:
Thanks, @OmegaHeron . I gather the sound was good with those settings (initial buffer size 3)?
(I see in the log that you didn't manually bump it up, and earlier you reported that 3 or higher was good.)

OmegaHeron:
@howard-stearns - For that sound device, yes, sound worked properly at 3.  <3 = Fail.  For other devices it's a different value > 3.

howard-stearns:
@OmegaHeron Cool. Thanks. Can you give me a log for a different device (other than the Sound Blaster Recon3D), at both 3 and when manually bumped up to whatever works?

OmegaHeron:
It should be an interesting log read.  Here's with my bluetooth stereo headset selected.  There's a whole lot of reference to the default SB device that's not selected output device and I can assure you - audio out is coming out the bluetooth headset, not SB internal. :)

[hifi-log.txt](https://github.com/highfidelity/hifi/files/138429/hifi-log.txt)

howard-stearns:
This is a bit of a fail. I haven't seen any auto-increases on Windows. As it stands, this PR:
 - Fixes the crash when auto-increasing (and thus fixes https://app.asana.com/0/26225263936266/90857558486721)
 - Doesn't persist auto-increased values (so each session has the opportunity to be better)
 - Gets good results for each machine I've tried (mac and windows).

But I think actually auto-increasing on Windows is going to have to wait for another PR in the future. Time to move on.

hifi-gustavo:
Build console output for commit: e21d9042d06cadf7c50a9d55ed659664aef23492 available at: https://gist.github.com/d00a53e9ddcf307cae2a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7122.dmg

OmegaHeron:
One last comment - if I set/save a value for buffer, other than a potential automatic increase, it will otherwise leave it alone?

howard-stearns:
That's correct, @OmegaHeron. If you set a preference for "Output Buffer Initial Size (frames)", that value is used at that time and every time you restart after that.  (If auto increase was working it would do so from that initial value.)

OmegaHeron:
@howard-stearns Thanks for the reply, just double checking as that's fairly critical until the real issue with broken audio on Windows is squashed.

PR #7124, opened 2016-02-17
Fixing the disappearing avatars maybe ?

samcake:
Trying tto force a renderItem update when the avatar are updated by Avatar Manager to make sure they are properly updated in the Scene SPatial tree.
Not sure it s the real fix for it yet, no clear repro though.
need QA in crowds makeing sure that avatars show up properly when to avatars meet in a domain

### DIFF
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -322,6 +322,12 @@ void Avatar::removeFromScene(AvatarSharedPointer self, std::shared_ptr<render::S
     }
 }
 
+void Avatar::updateRenderItem(render::PendingChanges& pendingChanges) {
+    if (_renderItemID != render::Item::INVALID_ITEM_ID) {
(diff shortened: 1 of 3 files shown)


howard-stearns:
@sethalves, you had a repro case for this. Does it fix it?

hifi-gustavo:
Build console output for commit: 640bf813397036ecfe64b518aecb5dcb8bf77003 available at: https://gist.github.com/aa0362e9ee238b12d95b<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7124.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7124.dmg

sethalves:
The repro is to just log two avatars into demo, facing each other.  log one or the other in and out a few times and see if everyone sees everyone else, each time.

sethalves:
With this code, I haven't yet had an avatar be invisible.

PR #7126, opened 2016-02-17
cleanup ViewFrustum method names

AndrewMeadows:
- cleanup some method/variable names in ViewFrustum class to be more correct
- render culling now only checks against the "frustum" part of the "keyhole" view
- add more efficient methods for keyhole culling (not used yet, but will be soon)
- add unit tests for some ViewFrustum methods
- add unit tests (incomplete) for AABox and AACube

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -425,11 +425,11 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
     _maxOctreePPS(maxOctreePacketsPerSecond.get()),
     _lastFaceTrackerUpdate(0)
 {
-    // FIXME this may be excessivly conservative.  On the other hand 
+    // FIXME this may be excessivly conservative.  On the other hand
     // maybe I'm used to having an 8-core machine
-    // Perhaps find the ideal thread count  and subtract 2 or 3 
+    // Perhaps find the ideal thread count  and subtract 2 or 3
     // (main thread, present thread, random OS load)
-    // More threads == faster concurrent loads, but also more concurrent 
+    // More threads == faster concurrent loads, but also more concurrent
     // load on the GPU until we can serialize GPU transfers (off the main thread)
     QThreadPool::globalInstance()->setMaxThreadCount(2);
     thread()->setPriority(QThread::HighPriority);
@@ -560,7 +560,7 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
 
     auto discoverabilityManager = DependencyManager::get<DiscoverabilityManager>();
     connect(&locationUpdateTimer, &QTimer::timeout, discoverabilityManager.data(), &DiscoverabilityManager::updateLocation);
-    connect(&locationUpdateTimer, &QTimer::timeout, 
+    connect(&locationUpdateTimer, &QTimer::timeout,
         DependencyManager::get<AddressManager>().data(), &AddressManager::storeCurrentAddress);
     locationUpdateTimer.start(DATA_SERVER_LOCATION_CHANGE_UPDATE_MSECS);
 
@@ -604,7 +604,7 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
 
     connect(addressManager.data(), &AddressManager::hostChanged, this, &Application::updateWindowTitle);
     connect(this, &QCoreApplication::aboutToQuit, addressManager.data(), &AddressManager::storeCurrentAddress);
-    
+
     // Save avatar location immediately after a teleport.
     connect(getMyAvatar(), &MyAvatar::positionGoneTo,
         DependencyManager::get<AddressManager>().data(), &AddressManager::storeCurrentAddress);
@@ -625,7 +625,7 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
         getEntities()->reloadEntityScripts();
     }, Qt::QueuedConnection);
 
-    connect(scriptEngines, &ScriptEngines::scriptLoadError, 
+    connect(scriptEngines, &ScriptEngines::scriptLoadError,
         scriptEngines, [](const QString& filename, const QString& error){
         OffscreenUi::warning(nullptr, "Error Loading Script", filename + " failed to load.");
     }, Qt::QueuedConnection);
@@ -975,7 +975,7 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
         disconnect(_idleTimer);
     });
     // Setting the interval to zero forces this to get called whenever there are no messages
-    // in the queue, which can be pretty damn frequent.  Hence the idle function has a bunch 
+    // in the queue, which can be pretty damn frequent.  Hence the idle function has a bunch
     // of logic to abort early if it's being called too often.
     _idleTimer->start(0);
 }
@@ -1023,7 +1023,7 @@ void Application::cleanupBeforeQuit() {
     getEntities()->shutdown(); // tell the entities system we're shutting down, so it will stop running scripts
     DependencyManager::get<ScriptEngines>()->saveScripts();
     DependencyManager::get<ScriptEngines>()->shutdownScripting(); // stop all currently running global scripts
-    DependencyManager::destroy<ScriptEngines>(); 
+    DependencyManager::destroy<ScriptEngines>();
 
     // first stop all timers directly or by invokeMethod
     // depending on what thread they run in
@@ -1212,10 +1212,10 @@ void Application::initializeUi() {
 
     setupPreferences();
 
-    // For some reason there is already an "Application" object in the QML context, 
+    // For some reason there is already an "Application" object in the QML context,
     // though I can't find it. Hence, "ApplicationInterface"
     rootContext->setContextProperty("SnapshotUploader", new SnapshotUploader());
-    rootContext->setContextProperty("ApplicationInterface", this); 
+    rootContext->setContextProperty("ApplicationInterface", this);
     rootContext->setContextProperty("AnimationCache", DependencyManager::get<AnimationCache>().data());
     rootContext->setContextProperty("Audio", &AudioScriptingInterface::getInstance());
     rootContext->setContextProperty("Controller", DependencyManager::get<controller::ScriptingInterface>().data());
@@ -2435,7 +2435,7 @@ void Application::idle(uint64_t now) {
     if (_aboutToQuit) {
         return; // bail early, nothing to do here.
     }
-    
+
     auto displayPlugin = getActiveDisplayPlugin();
     // depending on whether we're throttling or not.
     // Once rendering is off on another thread we should be able to have Application::idle run at start(0) in
@@ -2459,7 +2459,7 @@ void Application::idle(uint64_t now) {
     // Nested ifs are for clarity in the logic.  Don't collapse them into a giant single if.
     // Don't saturate the main thread with rendering, no paint calls until the last one is complete
     if (!_pendingPaint) {
-        // Also no paint calls until the display plugin has increased by at least one frame 
+        // Also no paint calls until the display plugin has increased by at least one frame
         // (don't render at 90fps if the display plugin only goes at 60)
         if (_renderedFrameIndex == INVALID_FRAME || presentCount > _renderedFrameIndex) {
             // Record what present frame we're on
@@ -2469,14 +2469,14 @@ void Application::idle(uint64_t now) {
             // But when we DO request a paint, get to it as soon as possible: high priority
             postEvent(this, new QEvent(static_cast<QEvent::Type>(Paint)), Qt::HighEventPriority);
         }
-    }  
-    
-    // For the rest of idle, we want to cap at the max sim rate, so we might not call 
+    }
+
+    // For the rest of idle, we want to cap at the max sim rate, so we might not call
     // the remaining idle work every paint frame, or vice versa
     // In theory this means we could call idle processing more often than painting,
     // but in practice, when the paintGL calls aren't keeping up, there's no room left
     // in the main thread to call idle more often than paint.
-    // This check is mostly to keep idle from burning up CPU cycles by running at 
+    // This check is mostly to keep idle from burning up CPU cycles by running at
     // hundreds of idles per second when the rendering is that fast
     if ((timeSinceLastUpdateUs / USECS_PER_MSEC) < CAPPED_SIM_FRAME_PERIOD_MS) {
         // No paint this round, but might be time for a new idle, otherwise return
@@ -3423,7 +3423,7 @@ void Application::queryOctree(NodeType_t serverType, PacketType packetType, Node
                                                   rootDetails.y * TREE_SCALE,
                                                   rootDetails.z * TREE_SCALE) - glm::vec3(HALF_TREE_SCALE),
                                         rootDetails.s * TREE_SCALE);
-                    ViewFrustum::location serverFrustumLocation = _viewFrustum.cubeInFrustum(serverBounds);
+                    ViewFrustum::location serverFrustumLocation = _viewFrustum.computeCubeViewLocation(serverBounds);
 
                     if (serverFrustumLocation != ViewFrustum::OUTSIDE) {
                         inViewServers++;
@@ -3491,7 +3491,7 @@ void Application::queryOctree(NodeType_t serverType, PacketType packetType, Node
                                         rootDetails.s * TREE_SCALE);
 
 
-                    ViewFrustum::location serverFrustumLocation = _viewFrustum.cubeInFrustum(serverBounds);
+                    ViewFrustum::location serverFrustumLocation = _viewFrustum.computeCubeViewLocation(serverBounds);
                     if (serverFrustumLocation != ViewFrustum::OUTSIDE) {
                         inView = true;
                     } else {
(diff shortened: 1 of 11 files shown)


AndrewMeadows:
I'm getting ready to start on the physics LOD work.  Part of that will be to do "in view" checks to see if the interface should be adding objects to the physics simulation or not.  While looking that code over I noticed some methods that were poorly named and did not match our coding standard.  As part of that survey I cleaned up the code as I went and ended up with this PR which is purely name changes and whitespace cleanup.  I figured it would be easier to review this work alone rather than mixed in with the physics simulation changes that will follow.

interface/src/Application.cpp
```diff
@@ -3412,7 +3412,7 @@ void Application::queryOctree(NodeType_t serverType, PacketType packetType, Node
                                                   rootDetails.y * TREE_SCALE,
                                                   rootDetails.z * TREE_SCALE) - glm::vec3(HALF_TREE_SCALE),
                                         rootDetails.s * TREE_SCALE);
-                    ViewFrustum::location serverFrustumLocation = _viewFrustum.cubeInFrustum(serverBounds);
+                    ViewFrustum::location serverFrustumLocation = _viewFrustum.computeCubeViewLocation(serverBounds);
```
AndrewMeadows: As per coding standard section 1.2.2 (The term compute can be used in methods where something is computed) I'm changing the name of **cubeInFrustum()** (and friends) to be **computeCubeViewLocation()**.

Note, I'm using the word "view" to mean "frustum and sphere" of the keyhole.  When the test is just against the frustum then the method name will correctly reflect that fact.

libraries/octree/src/ViewFrustum.cpp
```diff
@@ -241,30 +241,18 @@ ViewFrustum::location ViewFrustum::boxInKeyhole(const AABox& box) const {
...
         float distance = _planes[i].distance(point);
-        if (distance < 0) {
-            return keyholeResult; // escape early will be the value from checking the keyhole
+        if (distance < 0.0f) {
+            return OUTSIDE;
         }
     }
-    return regularResult;
+    return INSIDE;
 }
```
AndrewMeadows: The old **pointInFrustum()** method received special attention.  It was only ever used to test against the **frustum** not the **sphere** (see Avatar.cpp) so I removed the **ignoreKeyhole** argument (which should have been called **ignoreSphere**) and corresponding logic.  Hence this method's new name says **Frustum** rather than **View** (the "frustum" is the truncated cone or pyramid of the "view").

hifi-gustavo:
Build console output for commit: 8cac170db5d59d5e245560ea91f521e359a562f9 available at: https://gist.github.com/0beecd3eccb7d5fee788<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.dmg

libraries/octree/src/ViewFrustum.cpp
ZappoMan: why wouldn't we want avatar.cpp to test against the actual keyhole frustum (sphere+ frustum)?

AndrewMeadows:
We're considering changing the names again and maybe the behavior of the calls inside **CullTask.cpp**, so this is not ready for test or merge:

In particular:

(1) we would change the names:

    computeFooViewLocation() --> fooInKeyhole()

which would return a **ViewFrustum::location** enum with OUTSIDE explicitly defined as 0 (false).

(2) maybe also implement fooInFrustum() if the **CullTask** code could be safely optimized.

interface/src/Application.cpp
```diff
@@ -3420,9 +3420,7 @@ void Application::queryOctree(NodeType_t serverType, PacketType packetType, Node
                                                   rootDetails.y * TREE_SCALE,
                                                   rootDetails.z * TREE_SCALE) - glm::vec3(HALF_TREE_SCALE),
                                         rootDetails.s * TREE_SCALE);
-                    ViewFrustum::location serverFrustumLocation = _viewFrustum.cubeInFrustum(serverBounds);
-
-                    if (serverFrustumLocation != ViewFrustum::OUTSIDE) {
+                    if ((bool)(_viewFrustum.cubeInFrustum(serverBounds))) {
                         inViewServers++;
                     }
```
AndrewMeadows: **ViewFrustum::cubeInFrustum()** now actually checks against only the **frustum** part of the keyhole.  I also made the code more readable by casting the **ViewFrustum::location** return value to **bool**:  **ViewFrustum::OUTSIDE** has a value of **0** which translates to **false** and the other possible values are:

**ViewFrustum::INSIDE**
**ViewFrustum::INTERSECT**

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -304,15 +304,13 @@ OctreeElement::AppendState EntityTreeElement::appendElementData(OctreePacketData
...
-                    // larger octree cell because of its position (for example if it crosses the boundary of a cell it
-                    // pops to the next higher cell. So we want to check to see that the entity is large enough to be seen
-                    // before we consider including it.
-                    if (includeThisEntity) {
+                    } else {
+                        // Check the size of the entity, it's possible that a "too small to see" entity is included in a
+                        // larger octree cell because of its position (for example if it crosses the boundary of a cell it
+                        // pops to the next higher cell. So we want to check to see that the entity is large enough to be seen
+                        // before we consider including it.
                         success = true;
```
AndrewMeadows: I'm simplifying and optimizing the logic here.  **includeThisEntity** MUST be true when entering this context (according to the outer **if** statement not shown in this diff).  The only way it could ever remain **true** is in the **else** case of the **if** you see above.

hifi-gustavo:
Build console output for commit: 73bd4c6ff2431479d608a917d4e86f5a859ab008 available at: https://gist.github.com/0beecd3eccb7d5fee788<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.dmg

hifi-gustavo:
Build console output for commit: 55d11da5caa5d99d0d48b48d3c6e331334ef6259 available at: https://gist.github.com/0beecd3eccb7d5fee788<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.dmg

interface/src/Application.cpp
ZappoMan: This one should be the full keyhole & frustum. Because it is forming the basis for querying the server. So you need it all.

ZappoMan:
I am still not convinced these names are more clear. "InFrustum" vs "TouchesKeyhole" seems like an odd distinction. Does "in" mean fully inside? Because it's not, it also includes intersections. Does "Touches" only refer to intersections but not fully inside? No, because it includes both. Does "Keyhole" refer to just the sphere or the entire sphere & frustum.

interface/src/Application.cpp
AndrewMeadows: Yes, this one and another inside **OctreeElement.cpp** need to be modified.  I noticed last night right before I had to leave.

AndrewMeadows:
In my mind the "keyhole" has two parts: a frustum and a sphere, however the central sphere is called "keyhole" inside the **ViewFrustum** class.  I was going to punt on that cleanup but I'll get it done so that the public API is consistent.

hifi-gustavo:
Build console output for commit: 1722cccc7c0ff2fc6701ff0d354d9416ec9561a4 available at: https://gist.github.com/0beecd3eccb7d5fee788<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.dmg

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -304,15 +304,13 @@ OctreeElement::AppendState EntityTreeElement::appendElementData(OctreePacketData
                     // frustum culling on rendering.
                     bool success;
                     AACube entityCube = entity->getQueryAACube(success);
-                    if (!success || params.viewFrustum->cubeInFrustum(entityCube) == ViewFrustum::OUTSIDE) {
+                    if (!success || params.viewFrustum->cubeTouchesKeyhole(entityCube)) {
```
ZappoMan: WHOA!!! Unless I'm totally misreading this... this appears to totally flip the logic... I think you missed a `!` here...

hifi-gustavo:
Build console output for commit: ac1ea32e8e4feb451dd45ff2d3f1387c908b828e available at: https://gist.github.com/0beecd3eccb7d5fee788<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.dmg

assignment-client/src/octree/OctreeQueryNode.cpp
```diff
@@ -153,7 +153,7 @@ bool OctreeQueryNode::updateCurrentViewFrustum() {
     newestViewFrustum.setPosition(getCameraPosition());
     newestViewFrustum.setOrientation(getCameraOrientation());
 
-    newestViewFrustum.setKeyholeRadius(getKeyholeRadius());
+    newestViewFrustum.setCenterRadius(getCameraCenterRadius());
```
AndrewMeadows: The sphere at the center of the keyhole is no longer called the "keyhole".

hifi-gustavo:
Build console output for commit: 5579549bd1f7f635d0b44fcce786e6f616131cde available at: https://gist.github.com/0beecd3eccb7d5fee788<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.dmg

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -304,15 +304,13 @@ OctreeElement::AppendState EntityTreeElement::appendElementData(OctreePacketData
                     // frustum culling on rendering.
                     bool success;
                     AACube entityCube = entity->getQueryAACube(success);
-                    if (!success || params.viewFrustum->cubeInFrustum(entityCube) == ViewFrustum::OUTSIDE) {
+                    if (!success || !params.viewFrustum->cubeIntersectsKeyhole(entityCube)) {
                         includeThisEntity = false; // out of view, don't include it
-                    }
```
AndrewMeadows: Yes, I had accidentally inverted the logic here, but I caught it while reviewing my results.

hifi-gustavo:
Build console output for commit: 07ea92c4ff5759707c590a0bdd83e6553387a089 available at: https://gist.github.com/0beecd3eccb7d5fee788<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.dmg

libraries/render/src/render/CullTask.cpp
```diff
@@ -39,12 +39,12 @@ void render::cullItems(const RenderContextPointer& renderContext, const CullFunc
 
         // TODO: some entity types (like lights) might want to be rendered even
         // when they are outside of the view frustum...
-        bool outOfView;
+        bool inView;
         {
-            PerformanceTimer perfTimer("boxInFrustum");
-            outOfView = frustum->boxInFrustum(item.bound) == ViewFrustum::OUTSIDE;
+            PerformanceTimer perfTimer("boxIntersectsFrustum");
+            inView = frustum->boxIntersectsFrustum(item.bound);
```
AndrewMeadows: New: These **CullTask** intersection tests really are only testing against the frustum part of the keyhole shape, rather than also the central sphere.

libraries/shared/src/AABox.cpp
```diff
@@ -281,6 +281,12 @@ bool AABox::findRayIntersection(const glm::vec3& origin, const glm::vec3& direct
     return false;
 }
 
+bool AABox::touchesSphere(const glm::vec3& center, float radius) const {
+    // Avro's algorithm from this paper: http://www.mrtc.mdh.se/projects/3Dgraphics/paperF.pdf
+    glm::vec3 e = glm::max(_corner - center, Vectors::ZERO) + glm::max(center - _corner - _scale, Vectors::ZERO);
+    return glm::length2(e) <= radius * radius;
+}
```
AndrewMeadows: This method is new, relatively fast, and unit tested.  We're using it in the keyhole intersection tests.

tests/shared/src/AngularConstraintTests.cpp
```diff
@@ -1,316 +0,0 @@
...
-        glm::quat expectedPitchYaw = glm::angleAxis(- expectedConeAngle, perpAxis);
-        glm::quat expectedRotation = expectedPitchYaw * expectedRoll;
-        
-        QVERIFY2(constrained, "ConeRollerConstraint should clamp()");
-        QVERIFY2(newRotation != rotation, "ConeRollerConstraint should change rotation");
-        QCOMPARE_WITH_ABS_ERROR(newRotation, expectedRotation, EPSILON);
-    }
-    delete c;
-}
-
```
AndrewMeadows: The AngularConstraint classes were purged a while ago.  These unit tests are cruft.

tests/shared/src/AACubeTests.cpp
```diff
@@ -0,0 +1,154 @@
...
+            QCOMPARE(cube.touchesSphere(sphereCenter, sphereRadius), true);
+
+            // outside ik
+            sphereCenter = cubeCenter + edgeOffset + (sphereRadius + delta) * edgeDirection;
+            QCOMPARE(cube.touchesSphere(sphereCenter, sphereRadius), false);
+            sphereCenter = cubeCenter - edgeOffset - (sphereRadius + delta) * edgeDirection;
+            QCOMPARE(cube.touchesSphere(sphereCenter, sphereRadius), false);
+        }
+    }
+}
```
AndrewMeadows: New unit tests for AACube and AABox.  The test coverage is not complete -- there is work to be done here but the NEW stuff added in this PR is tested.

hifi-gustavo:
Build console output for commit: 2509e2d3973d0d81950682473d564d138c29c956 available at: https://gist.github.com/0beecd3eccb7d5fee788<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.dmg

hifi-gustavo:
Build console output for commit: ad035fd2b9b0e8f2eb291f2575596f21b9474c84 available at: https://gist.github.com/0beecd3eccb7d5fee788<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.dmg

AndrewMeadows:
There is no new behavior in this PR, so the test is to make sure that things work as expected.

- When you login everything should be streamed and visible.
- As you move your view around objects do not vanish in correlation to look direction.

ZappoMan:
Testing Notes: 

This PR should have no functional change in behavior. Please test client/server connections old/new new/new new/old between client/server pairs and verify that everything still works.

hifi-gustavo:
Build console output for commit: 9c927ea36bcaca7f913595f0b4ddb1e2e9e5dad1 available at: https://gist.github.com/0beecd3eccb7d5fee788<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7126.dmg

OmegaHeron:
I'm going to divide testing/comments into multiple parts - this round is testing new Interface only with current server side code.

I see no difference in scene between this PR and production for content displayed.  Good.

For culling.  Using a bookmark to insure exact same view between runs.  Production Build 4275 I see 46 rendered of 181 with substantial content behind me.  This PR I see 46 rendered of 181.  I expected a difference there, but, it may be my expectation is incorrect.

Front view is

hifi-snap-by-omegaheron-on-2016-02-25_13-35-57

Turning 180 degrees for what was behind

hifi-snap-by-omegaheron-on-2016-02-25_13-36-56

Moving to testing with server side compiled from this PR next (takes a bit - have to build for linux first).

OmegaHeron:
Testing with server side running on this PR's code all checks pass - scene renders proper with PR Interface or Production.  Can place/remove objects.  Culling is seen on PR (minor but there in my test case).  This passes for me.

PR #7135, opened 2016-02-18
Update renderGrid to use a shader, and fix grid edit tools

zzmp:
- Reimplements GeometryCache::renderGrid to use a frag shader instead of GL_LINES
- Update properties of grid overlay for consistency
- Add back the grid tool to edit.js

### DIFF
--- a/examples/html/gridControls.html
+++ b/examples/html/gridControls.html
@@ -146,7 +146,7 @@
         <div class="property-section">
             <label>Major Grid Every</label>
             <span>
-                <input type='number' id="major-spacing" min="2" step="1", ></input>
+                <input type='number' id="major-spacing" min="1" step="1", ></input>
             </span>
         </div>
 
--- a/examples/html/gridControls.html
+++ b/examples/html/gridControls.html
@@ -139,7 +139,7 @@
         <div class="property-section">
             <label>Minor Grid Size</label>
             <span>
-                <input type='number' id="minor-spacing" min="0.2" step="0.2", ></input>
+                <input type='number' id="minor-spacing" min="0.2" step="0.2"></input>
             </span>
         </div>
 
--- a/examples/html/gridControls.html
+++ b/examples/html/gridControls.html
@@ -132,7 +132,7 @@
         <div id="horizontal-position" class="property-section">
             <label>Position (Y Axis)</label>
             <span>
-                <input type='number' id="horiz-y" class="number" value="-1.0" step="0.1"></input>
+                <input type='number' id="horiz-y" class="number" step="0.1"></input>
             </span>
         </div>
 
--- a/examples/html/gridControls.html
+++ b/examples/html/gridControls.html
@@ -33,8 +33,8 @@
                             elPosY.value = origin.y.toFixed(2);
                         }
     
-                        if (data.minorGridWidth !== undefined) {
-                            elMinorSpacing.value = data.minorGridWidth;
+                        if (data.minorGridEvery !== undefined) {
+                            elMinorSpacing.value = data.minorGridEvery;
                         }
     
                         if (data.majorGridEvery !== undefined) {
@@ -60,7 +60,7 @@
                             origin: {
                                 y: elPosY.value,
                             },
-                            minorGridWidth: elMinorSpacing.value,
+                            minorGridEvery: elMinorSpacing.value,
                             majorGridEvery: elMajorSpacing.value,
                             gridColor: gridColor,
                             colorIndex: gridColorIndex,
@@ -132,14 +132,14 @@
         <div id="horizontal-position" class="property-section">
             <label>Position (Y Axis)</label>
             <span>
-                <input type='number' id="horiz-y" class="number" value="0.0" step="0.1"></input>
+                <input type='number' id="horiz-y" class="number" value="-1.0" step="0.1"></input>
             </span>
         </div>
 
         <div class="property-section">
             <label>Minor Grid Size</label>
             <span>
-                <input type='number' id="minor-spacing" min="0" step="0.01", ></input>
+                <input type='number' id="minor-spacing" min="0.2" step="0.2", ></input>
             </span>
         </div>
 
--- a/examples/edit.js
+++ b/examples/edit.js
@@ -38,6 +38,10 @@ var lightOverlayManager = new LightOverlayManager();
 var cameraManager = new CameraManager();
 
 var grid = Grid();
+gridTool = GridTool({
+    horizontalGrid: grid
+});
+gridTool.setVisible(false);
 
 var entityListTool = EntityListTool();
 
@@ -341,7 +345,7 @@ var toolBar = (function() {
                 isActive = active;
                 if (!isActive) {
                     entityListTool.setVisible(false);
-                    // gridTool.setVisible(false);
+                    gridTool.setVisible(false);
                     grid.setEnabled(false);
                     propertiesTool.setVisible(false);
                     selectionManager.clearSelections();
@@ -349,7 +353,7 @@ var toolBar = (function() {
                 } else {
                     hasShownPropertiesTool = false;
                     entityListTool.setVisible(true);
-                    // gridTool.setVisible(true);
+                    gridTool.setVisible(true);
                     grid.setEnabled(true);
                     propertiesTool.setVisible(true);
                     Window.setFocus();
--- a/examples/libraries/gridTool.js
+++ b/examples/libraries/gridTool.js
@@ -25,7 +25,7 @@ Grid = function(opts) {
     var gridOverlay = Overlays.addOverlay("grid", {
         dimensions: { x: scale, y: scale, z: scale },
         visible: false,
-        drawInFront: true,
+        drawInFront: false,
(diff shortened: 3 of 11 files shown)


examples/html/gridControls.html
```diff
@@ -33,8 +33,8 @@
                             elPosY.value = origin.y.toFixed(2);
                         }
     
-                        if (data.minorGridWidth !== undefined) {
-                            elMinorSpacing.value = data.minorGridWidth;
+                        if (data.minorGridEvery !== undefined) {
+                            elMinorSpacing.value = data.minorGridEvery;
```
huffman: I'm not sure if the meaning changed, but I think the reason it was named different than major grid was that the minor grid size was meant to be in meters, while the major grid was a multiple of the minor grid size.

zzmp:
AFAIK, grid tools were not working before this PR, although I have heard varying opinions.

#### Intents

- Reimplement grids to use shaders instead of GL_LINES for speed.
- Get the grid tools in edit.js working again.

To that end...
#### Testing

##### AudioScope
The audioscope uses a grid in the background. It's lines may look a little thicker, but it should look and operate the same. It can be toggled with `Ctrl+P`.

##### Grid Overlay
The grid overlay's behavior was changed to include new properties:
`majorGridEvery` - Specifies that a thick line should be drawn every `majorGridEvery` meters.
`minorGridEvery` - Specifies a thin line.
`minorGridWidth` was removed.
`followCamera` - Specifies that the grid should follow the camera, drawn on the x/y plane. On by default. *If the grid is not centered on the camera, fail this QA. It* ***should not*** *center on the avatar, but it* ***should*** *have the height of the avatar's feet (or the floor on which it is standing).*
`dimensions` - x and y size the grid, z is not used but must be specified.
`position` - Specifies a position offset. Added to `followCamera`, if it is on.
`color`, `rotation`, `visible`, `drawInFront`.

The grid can be tested with:
```
Overlays.addOverlay('grid', {
    dimensions: { x: 1, y: 1, z: 1 },
    position: { x: 0, y: -1, z: 0 },
    rotation: Quat.fromPitchYawRollDegrees(90, 0, 0),
    drawInFront: true
})
```
- This was my test case, so please fiddle with the variables.
- You may need to increase the dimensions to initially find it.


##### edit.js

There should now be a grid panel. It should work as expected. Make sure that the marks for the grid, when visible, are on the ground. The best way I have for testing this is to turn on `snap to grid` and drag it, and make sure it corresponds to the markings. If not, fail this QA - the grid is elevated for some reason.

Load edit.js locally, so that it uses local gridTool and gridControl files.

examples/html/gridControls.html
zzmp: I changed the meaning. Parts of this used to be static, and I changed it so that mxxorGridEvery always referred to meters, and dimensions refers to the total size. I thought it was more consistent, and easier to understand.

huffman: Cool! Yes, I agree, that makes more sense.

hifi-gustavo:
Build console output for commit: 2d1c676a8b943ff521a084e8e745b4f79a801711 available at: https://gist.github.com/c2f18385b13ccb58770a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7135.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7135.dmg

nbq:
Testing this one now

nbq:
I have tested this up in my local domain and what happens is that the grid does not show on the ground but it shows sort of floating. When I click move to selection or move to avatar it disappears to me.

When position is set to 0, I can see it as I said before so I tried using the snap to grid and the box just jumps around and attaches to its own grid, not the one shown on my screen.

From this I would say it failed on my end.

huffman:
I tried this out, comments:

 * The grid is being drawn in front, it didn't previously do this - this makes it hard to tell where the grid actually is.
 * The follow-camera behavior previously moved in increments such that the grid never actually looked like was moving. ex. a grid line parallel to the x axis at x=1 would stay at x=1, and not appear to move to x=0 or x=2 as you move to the right or left.
 * The grid previously faded out at the edges.
 * The major line doesn't seem to always line up with the minor lines. Screenshot with minor=1, major=6: http://prntscr.com/a5cw6n
 * I'm not certain we don't want the edit.js grid tool to enforce the grid size to be a multiple of the minor grid size. If it's a multiple it always lines up, and if you adjust the minor grid size you don't then have to adjust the major grid size.

zzmp:
@huffman this last push should fix all of your concerns but the last. I agree that it would be good to constrain minor or major grid size so that they match up, but think it can wait for another PR.

@nbq, this should fix your issues as well. Please LMK if they persist, and include the domain and location that they occured on.

zzmp:
I just tested in demo and the grid lines are not visible. My calculation for floor height is wrong. I'm going to ask for help and update this with one more commit.

hifi-gustavo:
Build console output for commit: 0b4e82405f4bde23bbe5b485117af923167cfc3b available at: https://gist.github.com/c2f18385b13ccb58770a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7135.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7135.dmg

huffman:
Ok, tried it out, works a lot better! Couple issues though:

 * The floor height looks incorrect like you mentioned, about 2 meters below where it is supposed to be (if y=0 in edit.js)
 * The grid still moves depending on the size. Examples: 0.75, 2.0
 * I can't seem to change the major grid size. There is an error in the logs when trying to change it:
```
[02/19 14:30:58] [DEBUG] Web Window JS message: http://rawgit.com/zzmp/hifi/fix/grid/examples/html/gridControls.html 33 Uncaught TypeError: undefined is not a function
[02/19 14:30:58] [DEBUG] Web Window JS message: http://rawgit.com/zzmp/hifi/fix/grid/examples/html/gridControls.html 33 Uncaught TypeError: undefined is not a function
[02/19 14:30:58] [DEBUG] Web Window JS message: http://rawgit.com/zzmp/hifi/fix/grid/examples/html/gridControls.html 33 Uncaught TypeError: undefined is not a function
```

zzmp:
@huffman Where are you when you observe the floor height 2m too low? What do you mean when you say the grid still moves?

hifi-gustavo:
Build console output for commit: d74aa4f14d75211d82fd9692115ccdd7b4a940fc available at: https://gist.github.com/c2f18385b13ccb58770a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7135.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7135.dmg

zzmp:
I've logged three new issues related to this. Aside from those, I believe this is all set:
- https://app.asana.com/0/74051501495175/92310886312108
- https://app.asana.com/0/74051501495175/92310886312109
- https://app.asana.com/0/74051501495175/92310886312110

The issue that @huffman raised regarding the major div not being a multiple of the minor will be resolved with https://app.asana.com/0/74051501495175/92310886312108.

hifi-gustavo:
Build console output for commit: c9362a4073a90969eb5bd481c30227ccf24e7b25 available at: https://gist.github.com/c2f18385b13ccb58770a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7135.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7135.dmg

hifi-gustavo:
Build console output for commit: 665ef17fce57c514bfd3e5ec87573254b42635d0 available at: https://gist.github.com/c2f18385b13ccb58770a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7135.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7135.dmg

huffman:
Is the minimum minor grid size 0.2? I can't seem to set it below that. Other than that everything seems to work.

zzmp:
@huffman yes, the minimum major is 1, and the minimum minor is 0.2. I can easily change it if you think it should be smaller.

hifi-gustavo:
Build console output for commit: 5bf9b8bc76315f7b1b268ed2093286f7d230ac5f available at: https://gist.github.com/c2f18385b13ccb58770a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7135.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7135.dmg

PR #7134, opened 2016-02-18
Injector Load Test script

ericrius1:
This script provides a flexible load testing method for the audio injector system. It allows for testing how system handles many sounds playing at once, as well as how it handles many non-playing injectors

PR #7142, opened 2016-02-19
null audio injector

howard-stearns:
When we don't make an audo injector (because we're at a limit), make sure the value returned to scripts is null.

### DIFF
--- a/libraries/script-engine/src/AudioScriptingInterface.cpp
+++ b/libraries/script-engine/src/AudioScriptingInterface.cpp
@@ -45,8 +45,11 @@ ScriptAudioInjector* AudioScriptingInterface::playSound(Sound* sound, const Audi
         // stereo option isn't set from script, this comes from sound metadata or filename
         AudioInjectorOptions optionsCopy = injectorOptions;
         optionsCopy.stereo = sound->isStereo();
-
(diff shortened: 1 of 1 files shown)


howard-stearns:
https://app.asana.com/0/26225263936266/91914017108439

howard-stearns:
Testing note: Try the script from https://github.com/highfidelity/hifi/pull/7134, with numSoundsToPlayPerBatch set to any value greater than 40.  Crashes with release, should work with this build. Should continue to work as you let the script run several minutes. 

See log. The injectors after the first 40 in each batch should print as null.

hifi-gustavo:
Build console output for commit: fcdc6dd6fa34de97552a8d1a03272862832ae0ad available at: https://gist.github.com/a24320db86040613e47a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7142.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7142.dmg

ericrius1:
I ran the test script with over 40 injectors, and did not crash. I also saw the injectors print as null once the limit of 40 had been exceeded

PR #7147, opened 2016-02-19
Too many (or too large) web views crash Interface

jherico:
This addresses a crash found when opening a lot of web views simultaneously.  

* Directly address some crash conditions by catching errors thrown when rendering offscreen QML
* Remove some warnings from the log by ensuring that QML threads are shut down properly before being destroyed.
* Limit the maximum number of webviews that can be concurrently rendered to 100
* Automatically de-allocate the surface of a webview that has not been rendered for more than 30 seconds (the entity still exists, but the rendering resources are destroyed, along with any state in the webview)
* Log warnings if too many web views are attemtping to render concurrently.

### DIFF
--- a/libraries/entities-renderer/src/RenderableWebEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableWebEntityItem.cpp
@@ -28,6 +28,11 @@
 
 const float DPI = 30.47f;
 const float METERS_TO_INCHES = 39.3701f;
+static uint32_t _currentWebCount { 0 };
+// Don't allow more than 100 concurrent web views
+static const uint32_t MAX_CONCURRENT_WEB_VIEWS = 100;
+// If a web-view hasn't been rendered for 30 seconds, de-allocate the framebuffer
+static uint64_t MAX_NO_RENDER_INTERVAL = 30 * USECS_PER_SECOND;
 
 EntityItemPointer RenderableWebEntityItem::factory(const EntityItemID& entityID, const EntityItemProperties& properties) {
     EntityItemPointer entity{ new RenderableWebEntityItem(entityID) };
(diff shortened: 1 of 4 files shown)


jherico:
https://app.asana.com/0/27650181942747/80914271573788

hifi-gustavo:
Build console output for commit: 021b86c3fe0d3497f26ba898e60ba1e6890d5163 available at: https://gist.github.com/67d076eb9af21f68333f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7147.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7147.dmg

howard-stearns:
Hmm. Using this build, I crash immediately when making even just a few Web Entities.

I was using examples/test/performance/renderableMatrix.js:
- set the switch at the top to use: switch ('web').
- set the ROWS_X, _Y, and _Z as desired. The default works in release.
- you might want to set the LIFETIME at the top to use something smaller.
- you might want to turn your LOD up manually to max, so that you can see them all. (The default puts them kind of far away away.)

nbq:
I am testing this now, replicating the test @howard-stearns did earlier

nbq:
Ok here is my input: I noticed the creating goes just fine and if I turn up LOD I can see all of the web windows. Here are the vars I used in the script:

var LIFETIME = 120;
var ROWS_X = 20;
var ROWS_Y = 20;
var ROWS_Z = 20;

I noticed that the only lag/lockup I had is when it deleted the render threads. Normally people would not have this amount.

PhilipRosedale:
@jherico if there are more than 100, are the closest ones rendered?  Or is it random, or the first loaded?  If there are 100 being rendered, how great is the impact on frame rate?  Should we set the limit of such web surfaces lower than 100?

jherico:
@PhilipRosedale it will essentially preferentially render the ones that enter the view frustum first, but otherwise it will be random within the entities that get rendered.  100 entities seems to have little effect on my framerate.

PR #7151, opened 2016-02-20
Improvements to energy system

PhilipRosedale:
The energy system is currently a client-side only prototype of a way to restrict the rate and impact of edits that visitors to a domain can make.  The big idea is to give each visitor an amount of 'energy' which is continually replenished (up to a cap) and debited when changes are made such as creating entities, editing them, deleting them, or moving oneself around. 

There is a script called MyEnergyBar.js which activates this prototype system.  You will see a bar at upper right of screen which will drop as you make changes and then recover.  If you take actions or run scripts which make enough changes that you run out of energy, you will no longer be able to take edit actions.  

to test: https://rawgit.com/PhilipRosedale/hifi/master/examples/example/ui/MyEnergyBar.js

(sorry about the mis-named Alessandro commits, I switched laptops and hadn't changed the user.name) 

In this update: 

Changes to make add/edit/delete entity code cleaner - there were cases where this code if active could result in unmatched client/server entity tree state.  There are separate functions now for computing energy cost of adding, editing, deleting entities.   Right now these functions simply impose a fixed cost for add/delete/edit entities, regardless of the detailed properties of the entity.


hifi-gustavo:
Build console output for commit: e649299ccc1daf6ca90f83969f2fbbd36fc1fc1a available at: https://gist.github.com/56e0c4a7a66eb37fd3a0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7151.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7151.dmg

examples/example/ui/MyEnergyBar.js
```diff
@@ -32,32 +38,29 @@ var bar = Overlays.addOverlay("text", {
...
     Overlays.deleteOverlay(background);
     Overlays.deleteOverlay(bar);
 }
 
 function energyChanged(newValue) {
     Entities.currentAvatarEnergy = newValue;
+    updateDisplay();
 }
 
 function debitAvatarEnergy(value) {
```
PhilipRosedale: The script serves as a bridge between the entityScriptingSystem and myAvatar, which do not know about each other in the C++ code.  If entity edits cause an energy change, it is passed as a signal here, and used along with the energyChanged function.  Once we are happy with the behavior, we'll wire up the direct connections as needed between the lower level C++ classes.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -301,15 +301,15 @@ void MyAvatar::update(float deltaTime) {
     
      simulate(deltaTime);
     
-    currentEnergy += energyChargeRate;
-    currentEnergy -= getAccelerationEnergy();
-    currentEnergy -= getAudioEnergy();
-    
+    _currentEnergy += AVATAR_ENERGY_CHARGE_RATE * deltaTime;
```
PhilipRosedale: Right now, the avatar's energy recharges at a constant rate.  Energy change due to avatar movement is computed in getAccelerationEnergy(), and also speaking loudly into the microphone consumes energy (which will hopefully help with loud background noise users).   Audio and movement are not presently restricted, though, if all energy is used up.

hifi-gustavo:
Build console output for commit: 9c97b38944cc9666bc323f1213d3e5816a0034c2 available at: https://gist.github.com/56e0c4a7a66eb37fd3a0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7151.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7151.dmg

Menithal:
Just wondering, will editing any parameters of an entity also consume same amount of energy?

 Because I can see that impacting as an input restriction of controlling entities such as vehicles. That is, unless the entity is controlled via assignments. I'd think using any of the velocity values consuming less energy than setting positions or rotation manually

sethalves:
test this please

hifi-gustavo:
Build console output for commit: 9c97b38944cc9666bc323f1213d3e5816a0034c2 available at: https://gist.github.com/56e0c4a7a66eb37fd3a0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7151.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7151.dmg

hifi-gustavo:
Build console output for commit: ad140c21d5073bef058947f4716240b57e756739 available at: https://gist.github.com/56e0c4a7a66eb37fd3a0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7151.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7151.dmg

PhilipRosedale:
QA Plan: 

When the myEnergyBar.js script is NOT running, nothing should have changed with respect to editing entities.  Test this by having two people manipulating all the objects in demo repeatedly and verifying that they can see each other and interact the same as in current release version.  

With the myEnergyBar script running, an energy bar will appear at the top right of the screen.  Yelling lots into the microphone will decrease the bar (it will steadily recover).  Grabbing and moving objects as well as running scripts which do so will also decrease the bar (though not by much).  If the bar reaches zero, you will be unable to make more edits.  Currently this may make the far grab behavior unappealing (object will just not move), but this is subject of further work later and is expected.

imgntn:
some initial test results:
- currently no drain on far grabbing a dynamic cube
- tribbles.js  -- the tribbles appear to be moving to me locally, but my edit bar shows that i have no energy
 - createBow.js and pick up -- can aim for about 7 seconds, feels about right
- flockOfFish.js -- instant drain, unusable
- flowArtsSpawner.js -- negligible effect on energy from lightsaber and rave stick
- toybox -- negligible effect from ping pong gun, bubble wand
- earthquakes_live.js  -- big dip at start (it makes a little sphere for each earthquake), but seems like all entities appear as they should.

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -230,70 +224,54 @@ EntityItemProperties EntityScriptingInterface::getEntityProperties(QUuid identit
...
     EntityItemProperties properties = scriptSideProperties;
-
-    auto dimensions = properties.getDimensions();
-    float volume = dimensions.x * dimensions.y * dimensions.z;
-    auto density = properties.getDensity();
-    auto newVelocity = properties.getVelocity().length();
-    float oldVelocity = { 0.0f };
-
     EntityItemID entityID(id);
+    EntityItemPointer entity = _entityTree->findEntityByEntityItemID(entityID);
```
sethalves: We check to see if _entityTree is null, just below this.

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -230,70 +224,54 @@ EntityItemProperties EntityScriptingInterface::getEntityProperties(QUuid identit
...
+                if (!scriptSideProperties.localPositionChanged() && !scriptSideProperties.positionChanged()) {
+                    properties.setPosition(entity->getPosition());
+                }
+                if (!scriptSideProperties.localRotationChanged() && !scriptSideProperties.rotationChanged()) {
+                    properties.setRotation(entity->getOrientation());
+                }
             }
+            properties = convertLocationFromScriptSemantics(properties);
+            updatedEntity = _entityTree->updateEntity(entityID, properties);
+            emit debitEnergySource(energyCost);
```
sethalves: Here, updatedEntity might be false.  If it is, we shouldn't emit

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -170,7 +164,7 @@ QUuid EntityScriptingInterface::addEntity(const EntityItemProperties& properties
 
     // queue the packet
     if (success) {
-        emit debitEnergySource(cost);
+        emit debitEnergySource(energyCost);
```
sethalves: I'm not certain, but I think this setup may allow an avatar to go into negative energy -- the scripts run on their own threads and they are queueing up these debits to be consumed by the Avatar (main) thread.  It seems like scripts will be able to do edits they shouldn't be able to while these signal/slot messages are in flight.

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -357,6 +335,7 @@ QUuid EntityScriptingInterface::editEntity(QUuid id, const EntityItemProperties&
             });
         }
     });
+    // finally, send the edit message to the server
```
sethalves: should moving an entity with children cost more than one without?

PR #7164, opened 2016-02-23
Restyle message box modal dialogs

ctrlaltdavid:
This is the first pass at restyling the message box modal dialogs per the style set out in the UI Toolkit document.

Example test message box: Settings > Avatar... > Browse > Wear

Notes:
- Information, question, and warning message boxes now have their own icons rather than using the critical message box icon.
- The information etc. icons are displayed in the message box's title (if there is a title).
- The screen is no longer whitewashed when a message box is displayed; instead a focus shadow is used the same as the Running Scripts dialog.
- The message box confirming that you want to change your avatar now has a default action (indicated by blue "OK" button).
- "Cancel" buttons now appear on the right hand end of a list of buttons, the same as they do in typical Windows applications.
- Font sizes may be off, same as they are in Running Scripts dialog.
- This PR includes the Running Scripts dialog restyle, so it should not be merged before the Running Scripts dialog PR.


hifi-gustavo:
Build console output for commit: fdb7a9cf7532cf8bab139de9c95b82c8df0d18b5 available at: https://gist.github.com/e9afb99e1850f1214ecb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7164.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7164.dmg

interface/resources/qml/windows-uit/Frame.qml
```diff
@@ -0,0 +1,133 @@
...
+                if (pane.visible) {
+                    pane.visible = false;
+                    frameContent.visible = false
+                    hid = true;
+                }
+                var delta = Qt.vector2d(mouseX, mouseY).minus(pressOrigin);
+                frame.deltaSize(delta.x, delta.y)
+            }
+        }
+        HiFiGlyphs {
```
jherico: I would recommend renaming this to HiFiGlyph (non-plural)

interface/resources/qml/windows-uit/Frame.qml
```diff
@@ -0,0 +1,133 @@
...
+            visible: sizeDrag.enabled
+            x: -11  // Move a little to visually align
+            y: -4   // ""
+            text: "A"
```
jherico: I would create some mechanism to use logical names for these, rather than opaque letters.  Unfortunately you can't create an enum like type from QML.  I would suggest storing them in the Hifi constants type exposing them as named properties mapping to the glyph font values:  `property string fooSymbol: "A"`... etc

interface/resources/qml/windows-uit/Frame.qml
ctrlaltdavid: @jherico It's named HiFiGlyphs (with an 's') to match the font name ("hifi-glyphs.ttf"). If you still think it should be singular, I'll change it.

interface/resources/qml/windows-uit/Frame.qml
ctrlaltdavid: Have made use named properties for HiFi glyphs instead of "naked" character code.

hifi-gustavo:
Build console output for commit: a816d835d3129350eaa7488e52b39c606cfae350 available at: https://gist.github.com/e9afb99e1850f1214ecb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7164.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7164.dmg

RyanDowne:
Tested on PC and things look good. On Mac I am seeing some font size wonkiness. See pic below. 
screen shot 2016-02-24 at 8 27 56 am

PR #7191, opened 2016-02-25
Better Shading Model and support fr Stingray material

samcake:
- Work in progress don't merge yet

- Now detect and load the StingRay materials in a FBX file correctly, load all the maps to produce a Physically Based Shading
- Refactor the DeferredG_Buffer to support more information in the same footprint,
- Added support for Emissive (material value and Texture) and removed the interpretation as lightmap.
- Added support for Occlusion map (StingRay material only)
- Made a simple debug tool to look at what is really produced through the DeferredBuffer allowing to look at the different components individually
- Renamed Diffuse to Albedo
- Renamed Specular to Metallic. Ketpt the color of the specular as Fresnel
- Factorized a bunch of shader code to make it easier to maintain regarding the model draw pass. planning to consolidate in fewer shader variations.
- Passed the materialKey to the shader so we can apply more elaborated logic on how to combine  material value with textures
- In Scribe, added the keyword "_SCRIBE_NULL" allowing to be passed as a Func parameter when instanced to avoid defining a var
- Added the PR work from Zach adding support for the Ambient MAp as Environment map  too

Now working on the Guide  & documentation for content creators.


hifi-gustavo:
Build console output for commit: 287bfeddda0c8c81cab2267bda680dd446c20c61 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

ericrius1:
Tested with this model: https://s3-us-west-1.amazonaws.com/hifi-content/eric/models/rocket.fbx

It looks great!

cozza13:
Avatar face (philip's avatar ) is glossy. 
capture

samcake:
RIght, it s expected that this is breaking some existing content.
THe fixes shouldn't be to hard, we would need to spend time with Ozan fixing some parameters of the model

Menithal:
Yeah stuff seems extra shiny, most likely due to the specular change to metallic. Not sure if that should be the case, shouldn't specular  specular, glossiness glossiness and metallic metallic so that engine wise it is consistent inline with other engines such as Substance Engine, UE4, U5, 3DO, Cycles? 

The Rocket example also looks just shiny instead of metallic.

Could this or a later PR also make more sense to the entity's texture and originalTexture fields to not exclude content creators from non Autodesk suites (As in Blender) as well?

judasshuffle:
Not everything is super shiny for me but i do tend to dial the spec colour right down. i think a little suffering in this case will be worth the gain

bluetoe:
This has strange effects on polyvoxels. The correct textures do not appear and when the viewing angle is changed, the textures and/or colors of the polyvoxels change or sometimes disappear completely.

hifi-gustavo:
Build console output for commit: 96fb9c44662819433a62e400bac362e78d84a3a4 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

hifi-gustavo:
Build console output for commit: 610189ded3de336c171729281363d028b87d0973 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

Menithal:
Heya, Tested the newest build.

I have a Blender exported model with few  texture maps that I am using as a test reference. [Here is a part of the model that may be use for tests.](http://www.norteclabs.com/HF/experiment/nyanta/texture_tester.fbx). 

It has Diffuse, Normal, Specular Color, Specular Intensity which are originally made in Quixel. They are actually from true PBR albeito/gloss/specular values but re-adjusted in Blender through power values prior to export.

Here is how it renders on the current master, on this, and with across section of both
https://youtu.be/NS-eq5B0t9Y

The new values make definitely everything look a tad bit brighter (better for hmds!), which is a bonus, but:

I noticed that Blender Shader's power values are now ignored, thus the texture is read in an as is basis, and most of the values are simply skipped. 

Its either that, or the shinyness is due to no roughness/glossiness being a value we can export to Hifi from here. To tolerate this, shouldn't the this setting be settable or by default set to a smaller value if it is missing?

As a side note, I also noticed that if a normal map is mirrored via UVs, there seems to be some wierd artifact going on within Hifi, which doesn't occur on other engines, But that effects both master and this, and will probably be submitted as a bug. This bug is visible in the video. (See the left side of the video example at the start)

hifi-gustavo:
Build console output for commit: 3ee6f9d6f2848aee57f4d820e2f4735c217f09f9 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

hifi-gustavo:
Build console output for commit: d64b9bb6d95713311fe2b4245336185964728268 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

samcake:
@Menithal THanks much for testing and for the helpful pointers to your asset.
I just tried to see the problem, the gloss map for the coat comes up in the specular map channel.
I also checked in the FBX and that is the issue.
YOu can eventually give it another shot if you could rexport the fbx making sure the map goes to the right channel.
IT might also be another bug that i haven't seen.

I ll try on my own if i can today.

Cheers

samcake:
Also , note that you can use the debugFramebuffer.js tool which is in the scripts/utilities/tools/render folder in the program file folder of the PR once installed in your pc.
THis meeans you need to go load the script from disk.
THis should let you see on the side of the screen all the differrent components coming through the DeferredBuffer and basically diagnose what is coming out from the model file into which channel
menithal-coat-shinny

For example see how i doagnosed that the glossmap was going into the metallic channel of the DeferredBuffer

hifi-gustavo:
Build console output for commit: 540114971392a7d0f8df11e00ae31c3044e9f261 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

hifi-gustavo:
Build console output for commit: 483f9a4ee3cecdcd6faed4ac72580e4bc0abdbdd available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

hifi-gustavo:
Build console output for commit: 99ba0119964d5d0f1479e201481cb57efead11eb available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

hifi-gustavo:
Build console output for commit: cfcde5032ccb9a9f99a98ddf65bad3e9f55643d7 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

Menithal:
OK,
Thanks for the debug tool, I think that is quite handy to have, also for content creators:

My Notes:

- Specular Intensity and Color -> Maps to Metallic ( I guess different color applies to fresnel as described?)
- Specular Hardness -> Maps to Roughness
- Metallic and Roughness do not seem to behave as they should

If the Metallic value is not defined, I suggest making it default to black. Right now it goes full metallic, and it might not be ideal for most objects.

Still awhiles to go though, but looks like a sound direciton, and good that we have that mapped out from blender's side:)

Menithal:
Okay, I have a more simpled testing model

http://www.norteclabs.com/HF/experiment/PBRTester/tester.fbx

 which uses substance designer textures, to nail down assigned to Blender values: 

This is current mapping in Blender to Hifi.

Diffuse Color: Albeido
Specular Hardness: Glossiness
Normal to Normal,
Specular Intensity: Metallicness

The Concrete seems to have way too much highlights, and the light doesnt seem to "spread" on to the texture as it does in Substance Designer . This is the most apparent when looking at the material from other angles such as in the following screenshot

hifi-gustavo:
Build console output for commit: 630c61e61da2009ff1460dc0d286dee18f1cca73 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

hifi-gustavo:
Build console output for commit: e72035ee3d93f549f499800c73e17ca2481ef683 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

Menithal:
Looking great now.

More observations:

- Light by it self seems very aggressive and makes things still quite wet from some angles,


- AmbientLight / reflections seem to stretch when zooming in and out.

libraries/fbx/src/FBXReader_Material.cpp
```diff
@@ -79,11 +79,20 @@ void FBXReader::consolidateFBXMaterials() {
...
+
+
+        FBXTexture transparentTexture;
+        QString transparentTextureID = transparentTextures.value(material.materialID);
+        if (!transparentTextureID.isNull()) {
+            transparentTexture = getTexture(transparentTextureID);
+         
+            material.opacityTexture = transparentTexture;
+            detectDifferentUVs |= (transparentTexture.texcoordSet != 0) || (!transparentTexture.transform.isIdentity());
+        }
```
AndrewMeadows: Not a blocker but... **transparentTexture** is declared outside the **if** context but is only ever used inside.

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -414,3 +503,6 @@ const NetworkMaterial* NetworkGeometry::getShapeMaterial(int shapeID) {
     }
 }
 
+void NetworkGeometry::textureLoaded(const QWeakPointer<NetworkTexture>& networkTexture) {
+    numTextureLoaded++;
+}
```
AndrewMeadows: I note that **numTextureLoaded** is not actually used for anything yet, but I assume it is for debug purposes or will eventually be exposed as a stat somewhere.

libraries/model/src/model/Light.slh
```diff
@@ -109,6 +109,10 @@ SphericalHarmonics getLightAmbientSphere(Light l) {
     return l._ambientSphere;
 }
 
+float getLightAmbientMapNumMips(Light l) {
+    return l._control.x;
+}
```
AndrewMeadows: This does not look right.  **setLightAmbientMapNumMips()** stores the value in the new data member **_ambientMapNumMips** but here we are getting it from **control.x**.

libraries/model/src/model/Light.h
```diff
@@ -120,7 +126,8 @@ class Light {
         Vec4 _spot{0.0f, 0.0f, 0.0f, 0.0f};
         Vec4 _shadow{0.0f};
 
-        Vec4 _control{0.0f, 0.0f, 0.0f, 0.0f};
+        float _ambientMapNumMips{ 0 };
```
AndrewMeadows: **_ambientMapNumMips** is a float and ... (I don't know why the number would have a floating point value but...) it should be initialized to **0.0f** rather than **0**.

libraries/model/src/model/Light.cpp
```diff
@@ -140,3 +140,16 @@ void Light::setAmbientSphere(const gpu::SphericalHarmonics& sphere) {
...
 }
+
+void Light::setAmbientMap(gpu::TexturePointer ambientMap) {
+    _ambientMap = ambientMap;
+    if (ambientMap) {
+        setAmbientMapNumMips(_ambientMap->evalNumMips());
+    } else {
+        setAmbientMapNumMips(0);
+    }
+}
```
AndrewMeadows: Texture::evalNumMips() returns a **uint16** which means either the **uint16** above should be cast to **int** or **setAmbientMapNumMips()** should be changed to accept a **uint16** rather than **int**.

Why is **Light::_ambientMapNumMips** a **float**?

libraries/model/src/model/Material.h
```diff
@@ -223,29 +233,39 @@ class Material {
...
-        
-        glm::vec3 _diffuse{ 0.5f };
-        float _opacity{1.f};
-        glm::vec3 _metallic{ 0.03f };
-        float _gloss{0.1f};
-        glm::vec3 _emissive{ 0.0f };
-        float _spare0{0.0f};
-        glm::vec4  _spareVec4{0.0f}; // for alignment beauty, Material size == Mat4x4
+        glm::vec3 _emissive{ 0.0f }; // No Emissive
+        float _opacity{ 1.f }; // Opacity = 1 => Not Transparent
```
AndrewMeadows: Float literal should be **1.0f**.

libraries/model/src/model/TextureMap.cpp
```diff
@@ -54,22 +54,38 @@ void TextureMap::setLightmapOffsetScale(float offset, float scale) {
...
         }
+
+        // If alpha was meaningfull refine
+        if (validAlpha && (alphaHistogram.size() > 1)) {
+            auto totalNumPixels = image.height() * image.width();
+            auto numOpaques = alphaHistogram[OPAQUE_ALPHA];
+            auto numTranslucents = alphaHistogram[TRANSLUCENT_ALPHA];
+            auto numTransparents = totalNumPixels - numOpaques - numTranslucents;
+
+            alphaAsMask = ((numTransparents / (double)totalNumPixels) < 0.05);
```
AndrewMeadows: Interesting... I would complain about missing an explicit cast to double on **numTransparents** but maybe the compiler's **auto** support is so smart it picked double as the type when it was declared?  Meh... I have no doubt the compiler will get it right since casts can only go up in number of bytes (no hidden truncation here).

libraries/render-utils/src/DeferredBufferWrite.slh
```diff
@@ -40,37 +40,41 @@ float evalOpaqueFinalAlpha(float alpha, float mapAlpha) {
     return mix(alpha, 1.0 - alpha, step(mapAlpha, alphaThreshold));
 }
 
-const vec3 DEFAULT_SPECULAR = vec3(0.1);
+const float DEFAULT_ROUGHNESS = 0.9;
 const float DEFAULT_SHININESS = 10;
+const float DEFAULT_METALLIC = 0;
```
AndrewMeadows: Do we want shader floats literals to be **0.0** or is **0** ok?

libraries/render-utils/src/DeferredGlobalLight.slh
```diff
@@ -11,109 +11,99 @@
...
-    
-    vec3 fragNormal = normalize(vec3(invViewMat * vec4(normal, 0.0)));
-    vec4 fragEyeVector = invViewMat * vec4(-position, 0.0);
-    vec3 fragEyeDir = normalize(fragEyeVector.xyz);
-    
-    vec3 color = diffuse.rgb * evalSphericalLight(getLightAmbientSphere(light), fragNormal).xyz * obscurance * getLightAmbientIntensity(light);
-
-    vec4 shading = evalFragShading(fragNormal, -getLightDirection(light), fragEyeDir, specular, gloss);
+vec3 fresnelSchlickAmbient(vec3 fresnelColor, vec3 lightDir, vec3 halfDir, float gloss) {
+    return fresnelColor + (max(vec3(gloss), fresnelColor) - fresnelColor) * pow(1.0f - clamp(dot(lightDir, halfDir), 0.0, 1.0), 5);
```
AndrewMeadows: I think this is shader code but I see a cpp-style float literal **1.0f** in this line.  Will that interfere with the shader compilation?

libraries/render-utils/src/MaterialTextures.slh
```diff
@@ -0,0 +1,139 @@
...
+<@endfunc@>
+
+<@func evalMaterialAlbedo(fetchedAlbedo, materialAlbedo, matKey, albedo)@>
+{
+    <$albedo$>.xyz = (((<$matKey$> & ALBEDO_VAL_BIT) != 0) ? <$materialAlbedo$> : vec3(1.0));
+
+    if (((<$matKey$> & ALBEDO_MAP_BIT) != 0)) {
+        <$albedo$>.xyz *= <$fetchedAlbedo$>.xyz;
+    }
+}
```
AndrewMeadows: Do we want K&R bracing for shader code too?  That is, do we want:

    <@func foo(bar)@> {
        // do stuff
    }

Or do we are we ok with this?

    <@func foo(bar)@> 
    {
        // do stuff
    }

libraries/render/src/render/ShapePipeline.h
```diff
@@ -206,10 +209,12 @@ class ShapePipeline {
     class Locations {
     public:
         int texcoordMatrices;
-        int diffuseTextureUnit;
+        int albedoTextureUnit;
         int normalTextureUnit;
+        int roughnessTextureUnit;
         int specularTextureUnit;
```
AndrewMeadows: In many places "specular" has been replaced by "roughness".  Was **specularTextureUnit** left here intentionally?

examples/utilities/tools/render/debugFramebuffer.js
```diff
@@ -0,0 +1,51 @@
+//
+//  debug.js
```
zzmp: Should be debugFramBuffer.js

examples/utilities/tools/render/debugFramebuffer.js
```diff
@@ -0,0 +1,51 @@
...
+//  Sam Gateau created on 2/18/2016.
+//  Copyright 2016 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+var DDB = Render.RenderDeferredTask.DebugDeferredBuffer;
+DDB.enabled = true;
+oldConfig = DDB.toJSON();
```
zzmp: You should grab the oldConfig before enabling (swap lines 13 and 14).

examples/utilities/tools/render/debugFramebuffer.js
```diff
@@ -0,0 +1,51 @@
...
+        resizing = true;
+    }
+});
+Controller.mouseReleaseEvent.connect(function() { resizing = false; });
+Controller.mouseMoveEvent.connect(function (e) { resizing && setDebugBufferSize(e.x); });
+
+
+function shouldStartResizing(eventX) {
+    var x = Math.abs(eventX - Window.innerWidth * (1.0 + DDB.size.x) / 2.0);
+    var mode = DDB.mode;
```
zzmp: mode can never be -1. The only toggle is on DDB.enabled, you should check that instead.

examples/utilities/tools/render/framebuffer.qml
```diff
@@ -0,0 +1,52 @@
+//
+//  main.qml
```
zzmp: Should be framebuffer.qml. Also, these files are typically capitalized.

examples/utilities/tools/render/framebuffer.qml
```diff
@@ -0,0 +1,52 @@
...
+//  examples/utilities/tools/render
+//
+//  Created by Zach Pomerantz on 2/8/2016
```
zzmp: :)

examples/utilities/tools/render/framebuffer.qml
```diff
@@ -0,0 +1,52 @@
...
+
+Column {
+    spacing: 8
+    Column {
+        id: debug
+        property var config: Render.getConfig("DebugDeferredBuffer")
+
+        function setDebugMode(mode) {
+            debug.config.enabled = (mode != -1);
+            debug.config.mode = mode;
```
zzmp: The setter for debug.config.mode enforces that mode > 0. It would make more sense to hide this behind a conditional:
```
debug.config.enabled = (mode != -1);
if (mode != -1) debug.config.mode = mode;
```

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -325,6 +325,7 @@ void EntityTreeRenderer::applyZonePropertiesToScene(std::shared_ptr<ZoneEntityIt
             _pendingAmbientTexture = false;
             if (_ambientTexture->getGPUTexture()->getIrradiance()) {
                 sceneKeyLight->setAmbientSphere(_ambientTexture->getGPUTexture()->getIrradiance());
```
zzmp: There is no case where setAmbientMap should not also setAmbientSphere, so why not remove call here and wrap it into setAmbientMap?

setAmbientSphere is used elsewhere, so setAmbientMap could delegate to it as part of its definition.

libraries/model/src/model/Light.cpp
samcake: _ambientMapNumMips is a float because it will come out as a float in the shader, so i  can directly save it in float
Do we really need the cast from uint16 to uint ? is that a warnign on a platform ?
we are not lossing any accuracy here?

libraries/fbx/src/FBXReader.cpp
```diff
@@ -865,6 +865,9 @@ FBXGeometry* FBXReader::extractFBXGeometry(const QVariantHash& mapping, const QS
                     }
                 } else if (object.name == "Material") {
                     FBXMaterial material;
+                    if (object.properties.at(1).toByteArray().contains("StingrayPBS")) {
```
zzmp: This (and other property lookups) use magic numbers.

libraries/fbx/src/FBXReader.cpp
```diff
@@ -879,7 +882,7 @@ FBXGeometry* FBXReader::extractFBXGeometry(const QVariantHash& mapping, const QS
                             propertyName = "P";
                             index = 4;
                         }
-                        if (properties) {
+                        if (properties && !material.isPBSMaterial) {
```
zzmp: Per coding standard, the nominal clause should come first.

libraries/fbx/src/FBXReader.cpp
```diff
@@ -914,6 +917,44 @@ FBXGeometry* FBXReader::extractFBXGeometry(const QVariantHash& mapping, const QS
 #endif
                                 }
                             }
+                        } else if (properties && material.isPBSMaterial) {
+                            std::vector<std::string> unknowns;
+                            foreach(const FBXNode& property, subobject.children) {
+                                if (property.name == propertyName) {
+                                    if (property.properties.at(0) == "Maya|use_normal_map") {
+                                        material.useNormalMap = (bool)property.properties.at(index).value<double>();
```
zzmp: I like that there is an FBXReader::getVec3. Can there be a getBool to avoid all the duplication for getting bools?

libraries/fbx/src/FBXReader.h
```diff
@@ -130,11 +130,10 @@ class FBXMaterial {
 public:
     FBXMaterial() {};
     FBXMaterial(const glm::vec3& diffuseColor, const glm::vec3& specularColor, const glm::vec3& emissiveColor,
-        const glm::vec2& emissiveParams, float shininess, float opacity) :
+         float shininess, float opacity) :
         diffuseColor(diffuseColor),
```
zzmp: Why was this left as diffuseColor, and everywhere else changed to albedo? Should this be albedoColor?

libraries/render-utils/src/MaterialTextures.slh
samcake: THis is preprocessed code, and i d suggest we live with this as it is for now
so keep <@func foo(bar)$>
{
}

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1030,18 +1071,35 @@ FBXGeometry* FBXReader::extractFBXGeometry(const QVariantHash& mapping, const QS
                     if (connection.properties.at(0) == "OP") {
                         int counter = 0;
                         QByteArray type = connection.properties.at(3).toByteArray().toLower();
-                        if (type.contains("diffuse")) {
+                        if ((type.contains("diffuse") && !type.contains("tex_global_diffuse"))) {
+                            diffuseTextures.insert(getID(connection.properties, 2), getID(connection.properties, 1));
```
zzmp: Is this left as diffuse instead of albedo on purpose?

libraries/render/src/render/ShapePipeline.h
samcake: NO, i ll rename
Actually it s specular=> metallic

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1030,18 +1071,35 @@ FBXGeometry* FBXReader::extractFBXGeometry(const QVariantHash& mapping, const QS
...
+                        } else if (type.contains("tex_metallic_map")) {
+                            metallicTextures.insert(getID(connection.properties, 2), getID(connection.properties, 1));
+                        } else if (type.contains("shininess")) {
+                            shininessTextures.insert(getID(connection.properties, 2), getID(connection.properties, 1));
+                        } else if (type.contains("tex_roughness_map")) {
+                            roughnessTextures.insert(getID(connection.properties, 2), getID(connection.properties, 1));
+                        } else if (type.contains("emissive")) {
+                            emissiveTextures.insert(getID(connection.properties, 2), getID(connection.properties, 1));
+                        } else if (type.contains("tex_emissive_map")) {
+                            roughnessTextures.insert(getID(connection.properties, 2), getID(connection.properties, 1));
```
zzmp: roughnessTextures and emissiveTextures are set before this. Should this be emissiveTextures? Should the line above (1097) be a texture?

libraries/fbx/src/FBXReader.h
```diff
@@ -397,11 +417,16 @@ class FBXReader {
 
 
     QHash<QString, QString> diffuseTextures;
```
zzmp: Should this be albedoTextures?

libraries/fbx/src/FBXReader.h
```diff
@@ -144,18 +143,39 @@ class FBXMaterial {
...
+    float emissiveIntensity{ 1.0f };
+
     QString materialID;
     model::MaterialPointer _material;
 
-    FBXTexture diffuseTexture;
-    FBXTexture opacityTexture;
     FBXTexture normalTexture;
+    FBXTexture albedoTexture;
+    FBXTexture opacityTexture;
```
zzmp: Should there be a transparentTexture here, as there is in line 420?

libraries/fbx/src/FBXReader_Material.cpp
```diff
@@ -100,37 +109,66 @@ void FBXReader::consolidateFBXMaterials() {
             material.normalTexture = normalTexture;
             detectDifferentUVs |= (normalTexture.texcoordSet != 0) || (!normalTexture.transform.isIdentity());
```
zzmp: Should this be bumpmapTexture?

Lots of these if-scopes are doing the same thing with different textures. Could you make a helper method or closure to do it to avoid this mistake?

libraries/model/src/model/Light.cpp
AndrewMeadows: There is no code problem here... only a technical (but small) problem contrary to our standard which asks for no implicit casts (Section 3.1.2): 

https://docs.highfidelity.com/docs/coding-standard#section-3-1-2-never-rely-on-implicit-type-conversion-not-floatvalue-intvalue-

Not really a blocker IMO, but I point it out to indicate that I'm reviewing the code carefully.  I wouldn't want someone else to point out that I missed it.

libraries/fbx/src/OBJReader.cpp
```diff
@@ -617,7 +616,7 @@ void fbxDebugDump(const FBXGeometry& fbxgeo) {
             qCDebug(modelformat) << "        specularColor =" << meshPart.specularColor << "mat =" << meshPart._material->getMetallic();
             qCDebug(modelformat) << "        emissiveColor =" << meshPart.emissiveColor << "mat =" << meshPart._material->getEmissive();
             qCDebug(modelformat) << "        emissiveParams =" << meshPart.emissiveParams;
-            qCDebug(modelformat) << "        gloss =" << meshPart.shininess << "mat =" << meshPart._material->getGloss();
+            qCDebug(modelformat) << "        gloss =" << meshPart.shininess << "mat =" << meshPart._material->getRoughness();
```
zzmp: Should this say roughness now, not gloss?

libraries/gpu/src/gpu/GLBackendShader.cpp
```diff
@@ -172,6 +170,11 @@ GLBackend::GLShader* compileShader(const Shader& shader) {
             filestream.close();
         }
         */
+
+        qCWarning(gpulogging) << "GLShader::compileShader - failed to compile the gl shader object:";
```
zzmp: Yes!!!!!

zzmp: This may flood the log. Is it easy to only print the source and errors the first time a shader fails to compile?

libraries/fbx/src/FBXReader_Material.cpp
samcake: I d like to keep it that way so i can check at the end of the function the values which have been collected.
It s easier to debug i argue

libraries/model-networking/src/model-networking/ModelCache.cpp
samcake: THis is correct i added a notifyong mechanism so we can actually revisit the material once ALL the maps bound to it have been loaded.
SO yes unused but will be used

hifi-gustavo:
Build console output for commit: 7120b64a33247d65d68c459e86fb4c735ba73cb2 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

hifi-gustavo:
Build console output for commit: 89920370dbd74f22355a1121c09ac73f8f16f0f2 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

libraries/fbx/src/FBXReader.h
samcake: No because here we are reading what we get from FBX MAterial, and i m trying to stay closer to the names from FBX
But this is totally arbitrary i agree

libraries/fbx/src/FBXReader.cpp
samcake: yes same reason as above

libraries/fbx/src/FBXReader.cpp
samcake: Good catch

libraries/fbx/src/FBXReader.h
samcake: As above, wer are in FBX territory here so it s diffuse

libraries/fbx/src/FBXReader.h
samcake: k

Menithal:
Tested the new PR.

Metallic seems to have been broken or then it is the Framebuffer Debug script showing the wrong layers (Roughness seems to show normal, etc)

I also made an additional test tool to check Ambient Occlusion map and Emit map export from Blender. similar to the previous test model: 

http://www.norteclabs.com/HF/experiment/PBRTester/testerAE.fbx 
You may need to reset cache for some of the textures to update.

It seems like the Emit lighting up the entire model, even if just alpha.
Meanwhile, Ambient occlusion simply isnt bind to anything.

hifi-gustavo:
Build console output for commit: 9fb318570bcb0bcfd2e8392ad1a7c0f649774fd3 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

OmegaHeron:
So - this is what I've been able to do with PR7191 so far.

hifi-snap-by-omegaheron-on-2016-03-01_15-38-52

hifi-snap-by-omegaheron-on-2016-03-01_15-39-10

hifi-snap-by-omegaheron-on-2016-03-01_15-39-23

Nice tone, shine under control and nice bumps from normal map source.

samcake:
I found a problem with the way i convert the gloss map into roughness map which probably explain the wet look.
Give me some time to provide a new version
Cheers

libraries/model/src/model/Material.h
```diff
@@ -223,29 +233,39 @@ class Material {
...
+        glm::vec3 _albedo{ 0.5f }; // Grey albedo => isAlbedo
+        float _roughness{ 1.0f }; // Roughness = 1 => Not Glossy
+
+        glm::vec3 _fresnel{ 0.03f }; // Fresnel value for a default non metallic
+        float _metallic{ 0.0f }; // Not Metallic
+
+
+        glm::vec3 _spare0{ 0.0f };
+
+        uint32_t _key{ 0 }; // a copy of the materialKey
```
zzmp: This key does not match the values set above it.

zzmp: Should this be a float so the schema is correctly passed to the shader? Is it fine as a uint32_t?

libraries/model/src/model/Material.h
```diff
@@ -223,29 +233,39 @@ class Material {
...
+
+        // for alignment beauty, Material size == Mat4x4
```
zzmp: cruft

libraries/model/src/model/Material.slh
```diff
@@ -12,9 +12,9 @@
 <@def MODEL_MATERIAL_SLH@>
 
 struct Material { 
-    vec4 _diffuse;
-    vec4 _specular;
-    vec4 _emissive;
+    vec4 _emissiveOpacity;
```
zzmp: Why do these need to be vec4? It would be clearer as vec3 _emissive float opacity, as is done in Material.h.

libraries/model/src/model/Material.slh
```diff
@@ -12,9 +12,9 @@
...
+    vec4 _albedoRoughness;
+    vec4 _fresnelMetallic;
     vec4 _spare;
```
zzmp: Doesn't this include the key?

libraries/model/src/model/Material.slh
```diff
@@ -26,41 +26,36 @@ Material getMaterial() {
...
-    // A = 1 / 1.055 = 0.94786729857
-    // B = 0.055 * A = 0.05213270142
-    // C = 1 / 12.92 = 0.0773993808
-    // G = 2.4
-    const float T = 0.04045;
-    const float A = 0.947867;
-    const float B = 0.052132;
-    const float C = 0.077399;
-    const float G = 2.4;
+vec3 getMaterialEmissive(Material m) { return m._emissiveOpacity.rgb; }
```
zzmp: Why do we need getters if it is just a struct? Why can't we just name the Material differently (
```
struct Material {
    vec3 emissive;
    float opacity;
    // ...
};
```
) and use it as m.emissive?

libraries/model/src/model/Material.slh
```diff
@@ -26,41 +26,36 @@ Material getMaterial() {
...
+float getMaterialRoughness(Material m) { return m._albedoRoughness.a; }
+
+vec3 getMaterialFresnel(Material m) { return m._fresnelMetallic.rgb; }
+float getMaterialMetallic(Material m) { return m._fresnelMetallic.a; }
+
+float getMaterialShininess(Material m) { return 1.0 - getMaterialRoughness(m); }
+
+int getMaterialKey(Material m) { return floatBitsToInt(m._spare.w); }
+
+const int EMISSIVE_VAL_BIT = 0x00000001;
```
zzmp: This seems prone to error, but I can't think of a more stable way to do it.

libraries/model/src/model/Material.h
```diff
@@ -28,17 +28,18 @@ class MaterialKey {
 public:
    enum FlagBit {
```
zzmp: Can you include a comment here that this bitset is written literally in Material.slh, so that any future editors know not to touch this bitset without updating the shader header?

libraries/model/src/model/TextureMap.cpp
```diff
@@ -54,22 +54,38 @@ void TextureMap::setLightmapOffsetScale(float offset, float scale) {
...
                     break;
                 }
             }
         }
+
+        // If alpha was meaningfull refine
+        if (validAlpha && (alphaHistogram.size() > 1)) {
+            auto totalNumPixels = image.height() * image.width();
+            auto numOpaques = alphaHistogram[OPAQUE_ALPHA];
+            auto numTranslucents = alphaHistogram[TRANSLUCENT_ALPHA];
```
zzmp: You swapped translucents and transparents.

libraries/model/src/model/TextureMap.cpp
```diff
@@ -223,6 +250,99 @@ gpu::Texture* TextureUsage::createNormalTextureFromBumpImage(const QImage& srcIm
     return theTexture;
 }
 
+gpu::Texture* TextureUsage::createRoughnessTextureFromImage(const QImage& srcImage, const std::string& srcImageName) {
```
zzmp: These functions are all nearly the same. Could they all delegate to some helper function to hold most of the logic in one place?

examples/utilities/tools/render/framebuffer.qml
```diff
@@ -0,0 +1,52 @@
...
+                "Normal",
+                "Roughness",
+                "Metallic",     
+                "Emissive",
+                "Occlusion",
+                "Lightmap",
+                "Lighting",
+                "Shadow",
+                "Pyramid Depth",
+                "Ambient Occlusion",
```
zzmp: You are missing "Ambient Occlusion Blurred".

libraries/render-utils/src/DeferredLightingEffect.cpp
```diff
@@ -92,6 +92,11 @@ void DeferredLightingEffect::init() {
     // Add the global light to the light stage (for later shadow rendering)
     _lightStage.addLight(lp);
 
+    lp->setDirection(glm::vec3(-1.0f));
+    lp->setColor(glm::vec3(1.0f));
+    lp->setIntensity(1.0f);
+    lp->setType(model::Light::SUN);
+    lp->setAmbientSpherePreset(gpu::SphericalHarmonics::Preset::OLD_TOWN_SQUARE);
```
zzmp: You'll need to set the ambientMap on the global light here.

libraries/render-utils/src/DeferredLightingEffect.cpp
```diff
@@ -315,10 +320,11 @@ void DeferredLightingEffect::render(const render::RenderContextPointer& renderCo
                 // Setup the global directional pass pipeline
                 {
                     if (_shadowMapEnabled) {
+                        //if (_skyboxTexture) {
```
zzmp: cruft

libraries/render-utils/src/DeferredLightingEffect.cpp
```diff
@@ -564,7 +570,10 @@ void DeferredLightingEffect::setGlobalLight(const model::LightPointer& light, co
     globalLight->setAmbientIntensity(light->getAmbientIntensity());
     globalLight->setAmbientSphere(light->getAmbientSphere());
 
-    _skyboxTexture = skyboxTexture;
```
zzmp: _skyboxTexture should be replaced with globalLight->getAmbientMap().

libraries/render-utils/src/DeferredLightingEffect.cpp
```diff
@@ -564,7 +570,10 @@ void DeferredLightingEffect::setGlobalLight(const model::LightPointer& light, co
     globalLight->setAmbientIntensity(light->getAmbientIntensity());
     globalLight->setAmbientSphere(light->getAmbientSphere());
```
zzmp: Just use globalLight->setAmbientMap(light->getAmbientMap()). no need to restore it in _skyboxTexture, and keeping it around is error-prone.

libraries/render-utils/src/DeferredLightingEffect.cpp
```diff
@@ -564,7 +570,10 @@ void DeferredLightingEffect::setGlobalLight(const model::LightPointer& light, co
...
 
-    _skyboxTexture = skyboxTexture;
+    _skyboxTexture = (light->getAmbientMap() ? light->getAmbientMap() : _skyboxTexture);
```
zzmp: This logic already happens when assigning the AmbientMap.

examples/utilities/tools/render/debugFramebuffer.js
```diff
@@ -0,0 +1,51 @@
+//
+//  ddebugFramBuffer.js
```
zzmp: Still typoed.

hifi-gustavo:
Build console output for commit: f09f28096f1340f2e850361f972d5cef9f3aa270 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

hifi-gustavo:
Build console output for commit: 16ac69315f1002bc1fa72d6de24e40747270a2c0 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

libraries/entities-renderer/src/EntityTreeRenderer.cpp
samcake: I m going to keep the 2 separated for now.
I can foresee the need for that potentially...

libraries/fbx/src/FBXReader.cpp
samcake: ok, let s keep it this way :)

libraries/model/src/model/Material.h
samcake: no it s comment poetry

libraries/model/src/model/Material.slh
samcake: the buffers in gpu memeory tend to be aligned on vec4 of 32bits words so this is why i always group the structure this way in shaders.

libraries/model/src/model/Material.slh
samcake: yeah, the _spare should be _spareKey

libraries/model/src/model/Material.slh
samcake: it s shader language, having a function ( which is inlined by the compiler) just hide whatever we put behind

hifi-gustavo:
Build console output for commit: aed2ad72d13556b7ecffcbcbd43260e8f7581f23 available at: https://gist.github.com/70c1b1b0c88b976e62ff<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7191.dmg

samcake:
Good to go from my tests

OmegaHeron:
I've tested this against my own domain - fed it some models that made current production interface > unhappy, loaded the hell that is sandbox and there's nothing I can really complain about.  I'm sure it'll have things to come that need help, but, for where this is now it passes testing for me on Windows 7.

ZappoMan:
"... there's nothing I can really complain about." - @OmegaHeron -- that's about as good as a QA Approved as I can ever expect to see. :)

OmegaHeron:
;) Well, sometimes I feel like I spend my time here with nothing but a red stamp of NO, but, 'tis the nature of current function.  It's nice to see something like this roll out and be able to say - thanks to the programming team.  Pitchforks and flames later, but - for a few minutes it's all good.

PR #7196, opened 2016-02-25
no expiry for simulation ownership

AndrewMeadows:
- entity-server: remove automatic expiry for simulation ownership
- entity-server: clear all relevant ownership when session ends
- entity-server: track entities that should be owned (dynamic only) on separate list
- interface: increase max number of resends of ownership release, but also increase delay between them

### DIFF
--- a/assignment-client/src/entities/EntityServer.h
+++ b/assignment-client/src/entities/EntityServer.h
@@ -27,6 +27,8 @@ struct ViewerSendingStats {
     quint64 lastEdited;
 };
 
+class SimpleEntitySimulation;
+
 class EntityServer : public OctreeServer, public NewlyCreatedEntityHook {
     Q_OBJECT
 public:
@@ -52,8 +54,8 @@ public:
     virtual void readAdditionalConfiguration(const QJsonObject& settingsSectionObject) override;
     virtual QString serverSubclassStats() override;
 
-    virtual void trackSend(const QUuid& dataID, quint64 dataLastEdited, const QUuid& viewerNode) override;
-    virtual void trackViewerGone(const QUuid& viewerNode) override;
+    virtual void trackSend(const QUuid& dataID, quint64 dataLastEdited, const QUuid& sessionID) override;
+    virtual void trackViewerGone(const QUuid& sessionID) override;
 
 public slots:
     void pruneDeletedEntities();
@@ -65,7 +67,7 @@ private slots:
     void handleEntityPacket(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode);
 
 private:
-    EntitySimulation* _entitySimulation;
+    SimpleEntitySimulation* _entitySimulation;
     QTimer* _pruneDeletedEntitiesTimer = nullptr;
 
     QReadWriteLock _viewerSendingStatsLock;
--- a/libraries/entities/src/SimpleEntitySimulation.cpp
+++ b/libraries/entities/src/SimpleEntitySimulation.cpp
@@ -18,51 +18,69 @@
 #include "EntityItem.h"
 #include "EntitiesLogging.h"
 
-const quint64 MIN_SIMULATION_OWNERSHIP_UPDATE_PERIOD = 2 * USECS_PER_SECOND;
+const quint64 MAX_OWNERLESS_PERIOD = 2 * USECS_PER_SECOND;
 
-void SimpleEntitySimulation::updateEntitiesInternal(const quint64& now) {
-    if (_entitiesWithSimulator.size() == 0) {
-        return;
-    }
-
-    if (now < _nextSimulationExpiry) {
-        // nothing has expired yet
-        return;
-    }
-
-    // If an Entity has a simulation owner but there has been no update for a while: clear the owner.
-    // If an Entity goes ownerless for too long: zero velocity and remove from _entitiesWithSimulator.
-    _nextSimulationExpiry = now + MIN_SIMULATION_OWNERSHIP_UPDATE_PERIOD;
-
+void SimpleEntitySimulation::clearOwnership(const QUuid& ownerID) {
     QMutexLocker lock(&_mutex);
-    SetOfEntities::iterator itemItr = _entitiesWithSimulator.begin();
-    while (itemItr != _entitiesWithSimulator.end()) {
+    SetOfEntities::iterator itemItr = _entitiesWithSimulationOwner.begin();
+    while (itemItr != _entitiesWithSimulationOwner.end()) {
         EntityItemPointer entity = *itemItr;
-        quint64 expiry = entity->getLastChangedOnServer() + MIN_SIMULATION_OWNERSHIP_UPDATE_PERIOD;
-        if (expiry < now) {
-            if (entity->getSimulatorID().isNull()) {
-                // no simulators are volunteering
-                // zero the velocity on this entity so that it doesn't drift far away
-                entity->setVelocity(Vectors::ZERO);
-                entity->setAngularVelocity(Vectors::ZERO);
-                entity->setAcceleration(Vectors::ZERO);
-                // remove from list
-                itemItr = _entitiesWithSimulator.erase(itemItr);
-                continue;
-            } else {
-                // the simulator has stopped updating this object
-                // clear ownership and restart timer, giving nearby simulators time to volunteer
-                qCDebug(entities) << "auto-removing simulation owner " << entity->getSimulatorID();
-                entity->clearSimulationOwnership();
+        if (entity->getSimulatorID() == ownerID) {
+            // the simulator has abandonded this object --> remove from owned list
+            qCDebug(entities) << "auto-removing simulation owner " << entity->getSimulatorID();
+            itemItr = _entitiesWithSimulationOwner.erase(itemItr);
+
+            if (entity->getDynamic() && entity->hasLocalVelocity()) {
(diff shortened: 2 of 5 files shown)


hifi-gustavo:
Build console output for commit: b4edfe2390d5821d13e76aae5e9c77af8bc3324a available at: https://gist.github.com/ca874630c8a3799442c5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7196.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7196.dmg

AndrewMeadows:
Test 1:

(1) Login with two users A and B
(2) Both users enable debug icons for simulation ownership: Developer->Physics->Highlight Simulation Ownership
(3) UserA creates 100 dynamic boxes in a pile (see script below) several meters in front of UserB such that UserB is not touching or near enough to activate the pile (not in bounding box of pile)
--> UserA should own all simulations (blue icon on far right with head and shoulders silhouette)
(4) Eventually the pile will settle down and stop
--> UserB should not get ownership of any portion of the pile


Test 2:

(1) UserB creates a non-dynamic object
(2) UserB sets angular damping to zero
(3) UserB sets angular velocity of the object so that it spins
 --> note that UserB will probably own the object simulation
(4) Log UserB out
--> UserA should not take ownership of the object (because it is kinematic and has been set moving with remote updates)
(5) Log UserB back in
--> UserB should see object spinning, and should not take simulation ownership

AndrewMeadows:
This script creates 100 boxes with a 300 sec lifespan:

    "use strict";
    /*jslint nomen: true, plusplus: true, vars: true*/
    var Vec3, Quat, MyAvatar, Entities, Camera, Script, print;
    var NUMBER_TO_CREATE = 100;
    var LIFETIME = 300; // seconds
    var SIZEX = 0.5;
    var SIZEY = 1.5;
    var SIZEZ = 0.5;
    var TYPE = "Box";
    var RATE_PER_SECOND = 800;
    var SCRIPT_INTERVAL = 100;
    var GRAVITY = { x: 0, y: -9.8, z: 0 };
    var VELOCITY = { x: 0.0, y: 0, z: 0 };
    var ANGULAR_VELOCITY = { x: 1, y: 1, z: 1 };
    var DAMPING = 0.5;
    var ANGULAR_DAMPING = 0.5;
    var RANGE = 1;
    var HOW_FAR_IN_FRONT_OF_ME = 9;
    var HOW_FAR_UP = 3;
    //
    var totalCreated = 0;
    var offset = Vec3.sum(Vec3.multiply(HOW_FAR_UP, Vec3.UNIT_Y),
                          Vec3.multiply(HOW_FAR_IN_FRONT_OF_ME, Quat.getFront(Camera.orientation)));
    var center = Vec3.sum(MyAvatar.position, offset);
    function randomVector(range) {
        return {
            x: (Math.random() -  0.5) * range.x,
            y: (Math.random() -  0.5) * range.y,
            z: (Math.random() -  0.5) * range.z
        };
    }
    //
    Script.setInterval(function () {
        if (!Entities.serversExist() || !Entities.canRez()) {
            return;
        }
        if (totalCreated >= NUMBER_TO_CREATE) {
            print("Created " + totalCreated + " boxes.");
            Script.stop();
        }
        var i, numToCreate = RATE_PER_SECOND * (SCRIPT_INTERVAL / 1000.0);
        for (i = 0; (i < numToCreate) && (totalCreated < NUMBER_TO_CREATE); i++) {
            Entities.addEntity({
                type: TYPE,
                name: "box-" + totalCreated,
                position: Vec3.sum(center, randomVector({ x: RANGE, y: RANGE, z: RANGE })),
                dimensions: {x: SIZEX, y: SIZEY, z: SIZEZ},
                color: {red: Math.random() * 255, green: Math.random() * 255, blue: Math.random() * 255},
                velocity: VELOCITY,
                angularVelocity: Vec3.multiply(Math.random(), ANGULAR_VELOCITY),
                damping: DAMPING,
                angularDamping: ANGULAR_DAMPING,
                gravity: GRAVITY,
                dynamic: true,
                lifetime: LIFETIME
            });
            totalCreated++;
        }
    }, SCRIPT_INTERVAL);

assignment-client/src/entities/EntityServer.h
```diff
@@ -65,7 +67,7 @@ private slots:
     void handleEntityPacket(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode);
 
 private:
-    EntitySimulation* _entitySimulation;
+    SimpleEntitySimulation* _entitySimulation;
```
AndrewMeadows: **EntityServer** gets a pointer to the derived class **SimpleEntitySimulation** so that it can call **SimpleEntitySimulation::clearOwnership()** whenever a sessionID goes away.

libraries/entities/src/SimpleEntitySimulation.h
```diff
@@ -28,8 +30,9 @@ class SimpleEntitySimulation : public EntitySimulation {
     virtual void changeEntityInternal(EntityItemPointer entity) override;
     virtual void clearEntitiesInternal() override;
 
-    SetOfEntities _entitiesWithSimulator;
-    quint64 _nextSimulationExpiry { 0 };
+    SetOfEntities _entitiesWithSimulationOwner;
+    SetOfEntities _entitiesThatNeedSimulationOwner;
+    quint64 _nextOwnerlessExpiry { 0 };
```
AndrewMeadows: **SimpleEntitySimulation** now maintains two lists: (a) entities that have a simulation owner and (b) entities that do not have a simulation owner but should.

List (a) is used to clear ownership whenever an interface's sessionID goes away.

List (b) is used to clear velocities whenever a simulation owner is not found after a few seconds.

NOTE: only moving **dynamic** objects (not kinematic) are on list (b).

OmegaHeron:
Tested with.... one machine running windows, second OS X.  User on Windows machine ran script - observing from mac all the cubes drop, settles and observation of ownership matches what test setup indicates it should.  Test 1 = Good.  Working on Test 2 and will report back.

OmegaHeron:
Test 2 - User B (AV on Mac) created cube, set damping to 0d0 and confirmed it took.  Set yaw velocity to 10d0 and saw object rotating.  User A (AV on Windows machine) sees box rotating.  User B closed Interface, user A continues to see cube rotating and did not take ownership.  User B re-launches Interface and sees cube rotating and does not see ownership indicator.  If I understand Test setup correctly this seems to indicate Test 2 = Good.

libraries/entities/src/SimpleEntitySimulation.cpp
```diff
@@ -70,26 +88,46 @@ void SimpleEntitySimulation::addEntityInternal(EntityItemPointer entity) {
...
+        _entitiesWithSimulationOwner.insert(entity);
+        _entitiesThatNeedSimulationOwner.remove(entity);
     }
     entity->clearDirtyFlags();
 }
 
 void SimpleEntitySimulation::clearEntitiesInternal() {
-    _entitiesWithSimulator.clear();
+    _entitiesWithSimulationOwner.clear();
+    _entitiesThatNeedSimulationOwner.clear();
```
Atlante45: We should probably have a mutex locker here too.

libraries/entities/src/SimpleEntitySimulation.cpp
```diff
@@ -70,26 +88,46 @@ void SimpleEntitySimulation::addEntityInternal(EntityItemPointer entity) {
...
     QMutexLocker lock(&_mutex);
-    _entitiesWithSimulator.remove(entity);
+    _entitiesWithSimulationOwner.remove(entity);
+    _entitiesThatNeedSimulationOwner.remove(entity);
 }
 
 void SimpleEntitySimulation::changeEntityInternal(EntityItemPointer entity) {
     EntitySimulation::changeEntityInternal(entity);
-    if (!entity->getSimulatorID().isNull()) {
+    if (entity->getSimulatorID().isNull()) {
```
ZappoMan: you appear to be flipping the logic here... is that intentional?

Atlante45: The case that was here before is in the else now.

AndrewMeadows: Atlante45 is correct.  I added an **else** case here, then flipped the logic and put the original (insert) stuff in the **else** so this one was intentional.

libraries/entities/src/SimpleEntitySimulation.cpp
AndrewMeadows: I was going to say, "This is only called on shutdown", but I looked at the code and learned: in theory (e.g. the API supports it) if someone were to add a new _different_ simulation to the **EntityTree** then this would be called on the old one NOT during shutdown.  I guess I'll add a mutex locker.

hifi-gustavo:
Build console output for commit: a933935e2d6af70dea252a2707efa2d3ce8d7675 available at: https://gist.github.com/ca874630c8a3799442c5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7196.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7196.dmg

PR #7206, opened 2016-02-26
Application: don't simulate until we are ready to render

hyperlogic:
This should prevent the issue where we would sometimes simulate more often than we were rendering.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -315,9 +315,6 @@ bool setupEssentials(int& argc, char** argv) {
 
     Setting::preInit();
 
-    LogHandler::getInstance().setShouldDisplayMilliseconds(true);
-    LogHandler::getInstance().setShouldOutputThreadID(true);
-
     CrashHandler::checkForAndHandleCrash();
     CrashHandler::writeRunningMarkerFiler();
     qAddPostRoutine(CrashHandler::deleteRunningMarkerFile);
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -2455,16 +2455,16 @@ void Application::idle(uint64_t now) {
         _renderedFrameIndex = INVALID_FRAME;
     }
 
-    // Don't saturate the main thread with rendering, no paint calls until the last one is complete
-    // Also no paint calls until the display plugin has increased by at least one frame
-    // (don't render at 90fps if the display plugin only goes at 60)
+    // Don't saturate the main thread with rendering and simulation,
+    // unless display plugin has increased by at least one frame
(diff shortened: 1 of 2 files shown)


interface/src/Application.cpp
```diff
@@ -1321,10 +1321,6 @@ void Application::paintGL() {
         _lastFramesPerSecondUpdate = now;
     }
 
-    if (_isGLInitialized) {
-        idle(now);
-    }
-
```
hyperlogic: Remove idle from paintGL

interface/src/Application.h
```diff
@@ -509,7 +509,6 @@ private slots:
     int _avatarAttachmentRequest = 0;
 
     bool _settingsLoaded { false };
-    bool _pendingPaint { false };
```
hyperlogic: We no longer need the _pendingPaint flag.

hyperlogic:
https://app.asana.com/0/26225263936266/92650435482459

hifi-gustavo:
Build console output for commit: 08c16beff26e4eb93b10a7d459309606fa4c389a available at: https://gist.github.com/5de3cc9b9c44f1775ed2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7206.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7206.dmg

hifi-gustavo:
Build console output for commit: e5c81f66a54c8c9603fa03bd5e822d8aa67de055 available at: https://gist.github.com/5de3cc9b9c44f1775ed2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7206.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7206.dmg

bluetoe:
Sim rate reported in Stats seems now to closely match render rate. Unless I create a ton of physical stuff, which is to be expected.

bluetoe:
One thing I wonder about is when HiFi is not main focus, i.e. when minimized or other application or window is on top, the render rate drops, which is what is intended, but so does the Stat reported Simrate.

Could this make problems for physical stuff?.. If you click to look at Skype or something will this affect your version of physics in HiFi?

PhilipRosedale:
Tested and confirmed working in multiple locations.  Sim rate goes as high as render/present, no higher.  Release goes up to 120.

PR #7213, opened 2016-02-26
Fix resource caches overlap

Atlante45:
- Put all resource request on the same thread (Formerly only the AssetClient's thread)
- Made sure all resource request have the disk cache accessible.
- Made sure we don't have several disk cache instances sharing a same directory.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1072,10 +1072,8 @@ void Application::cleanupBeforeQuit() {
 }
 
 void Application::emptyLocalCache() {
-    if (auto cache = NetworkAccessManager::getInstance().cache()) {
-        qDebug() << "DiskCacheEditor::clear(): Clearing disk cache.";
-        cache->clear();
-    }
+    auto assetClient = DependencyManager::get<AssetClient>();
+    QMetaObject::invokeMethod(assetClient.data(), "clearCache", Qt::QueuedConnection);
 }
 
 Application::~Application() {
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -370,7 +370,6 @@ bool setupEssentials(int& argc, char** argv) {
     DependencyManager::set<AutoUpdater>();
     DependencyManager::set<PathUtils>();
     DependencyManager::set<InterfaceActionFactory>();
-    DependencyManager::set<AssetClient>();
     DependencyManager::set<AudioInjectorManager>();
     DependencyManager::set<MessagesClient>();
     DependencyManager::set<UserInputMapper>();
@@ -528,13 +527,7 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
 
     audioThread->start();
 
-    // Setup AssetClient
-    auto assetClient = DependencyManager::get<AssetClient>();
-    QThread* assetThread = new QThread;
-    assetThread->setObjectName("Asset Thread");
-    assetClient->moveToThread(assetThread);
-    connect(assetThread, &QThread::started, assetClient.data(), &AssetClient::init);
-    assetThread->start();
+    ResourceManager::init();
 
     // Setup MessagesClient
     auto messagesClient = DependencyManager::get<MessagesClient>();
@@ -643,13 +636,6 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
     // send the identity packet for our avatar each second to our avatar mixer
     connect(&identityPacketTimer, &QTimer::timeout, getMyAvatar(), &MyAvatar::sendIdentityPacket);
     identityPacketTimer.start(AVATAR_IDENTITY_PACKET_SEND_INTERVAL_MSECS);
-
-    QString cachePath = QStandardPaths::writableLocation(QStandardPaths::DataLocation);
-    QNetworkAccessManager& networkAccessManager = NetworkAccessManager::getInstance();
-    QNetworkDiskCache* cache = new QNetworkDiskCache();
-    cache->setMaximumCacheSize(MAXIMUM_CACHE_SIZE);
-    cache->setCacheDirectory(!cachePath.isEmpty() ? cachePath : "interfaceCache");
-    networkAccessManager.setCache(cache);
 
     ResourceCache::setRequestLimit(3);
 
@@ -1129,12 +1115,6 @@ Application::~Application() {
     DependencyManager::destroy<ScriptCache>();
     DependencyManager::destroy<SoundCache>();
 
-    // cleanup the AssetClient thread
-    QThread* assetThread = DependencyManager::get<AssetClient>()->thread();
-    DependencyManager::destroy<AssetClient>();
-    assetThread->quit();
-    assetThread->wait();
-
     QThread* nodeThread = DependencyManager::get<NodeList>()->thread();
 
     // remove the NodeList from the DependencyManager
@@ -1149,6 +1129,8 @@ Application::~Application() {
 #if 0
     ConnexionClient::getInstance().destroy();
 #endif
+
+    ResourceManager::cleanup();
 
     qInstallMessageHandler(NULL); // NOTE: Do this as late as possible so we continue to get our log messages
 }
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1071,11 +1071,6 @@ void Application::cleanupBeforeQuit() {
     DependencyManager::destroy<OffscreenUi>();
 }
 
-void Application::emptyLocalCache() {
-    auto assetClient = DependencyManager::get<AssetClient>();
-    QMetaObject::invokeMethod(assetClient.data(), "clearCache", Qt::QueuedConnection);
-}
-
 Application::~Application() {
     EntityTreePointer tree = getEntities()->getTree();
     tree->setSimulation(NULL);
@@ -3061,7 +3056,7 @@ void Application::reloadResourceCaches() {
     _viewFrustum.setOrientation(glm::quat());
     queryOctree(NodeType::EntityServer, PacketType::EntityQuery, _entityServerJurisdictions);
 
-    emptyLocalCache();
+    DependencyManager::get<AssetClient>()->clearCache();
 
     DependencyManager::get<AnimationCache>()->refreshAll();
     DependencyManager::get<ModelCache>()->refreshAll();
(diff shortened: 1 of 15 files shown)


Atlante45:
Asana task: https://app.asana.com/0/26551373415487/92683736062201

hifi-gustavo:
Build console output for commit: 6f7ef0774ee9fb1f597a78ea0b43aba12d8d8708 available at: https://gist.github.com/624e24f281d070572df6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7213.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7213.dmg

ZappoMan:
We need a solid test plan for this... this is a complicated change. @Atlante45 - please give examples of how to verify that things are working correctly now.

interface/src/Application.cpp
```diff
@@ -1063,10 +1049,8 @@ void Application::cleanupBeforeQuit() {
 }
 
 void Application::emptyLocalCache() {
-    if (auto cache = NetworkAccessManager::getInstance().cache()) {
-        qDebug() << "DiskCacheEditor::clear(): Clearing disk cache.";
-        cache->clear();
-    }
+    auto assetClient = DependencyManager::get<AssetClient>();
+    QMetaObject::invokeMethod(assetClient.data(), "clearCache", Qt::QueuedConnection);
```
ZappoMan: why isn't this encapsulated in the `AssetClient` class. This seems like asking for trouble to make sure every caller knows to call invokeMethod() instead of just calling he method directly and letting the class handle the cross-thread syncronization.

interface/src/ui/DiskCacheEditor.cpp
```diff
@@ -122,30 +123,25 @@ void DiskCacheEditor::refresh() {
...
+    auto buttonClicked = OffscreenUi::question(_dialog, "Clearing disk cache",
+                                               "You are about to erase all the content of the disk cache, "
+                                               "are you sure you want to do that?",
+                                               QMessageBox::Ok | QMessageBox::Cancel);
     if (buttonClicked == QMessageBox::Ok) {
-        if (auto cache = NetworkAccessManager::getInstance().cache()) {
-            qDebug() << "DiskCacheEditor::clear(): Clearing disk cache.";
-            cache->clear();
-        }
+        auto assetClient = DependencyManager::get<AssetClient>();
```
ZappoMan: ibid

interface/src/Application.cpp
Atlante45: That's a good point, I'm gonna do that.

hifi-gustavo:
Build console output for commit: a7f5bd127cdbbda5a403e70c3523014a60e026d5 available at: https://gist.github.com/624e24f281d070572df6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7213.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7213.dmg

Atlante45:
Testing plan:

- ACs can download asset:
I used `ambient.js` with this sound `http://hifi-public.s3.amazonaws.com/birarda/medium-crowd.wav`.

Atlante45:
Testing plan:

- Interface can download/upload assets, disk cache editor works:
    1. Test every type of assets: ATP, HTTP, FILE. (Clear cache manually to test it independently)
    2. I tested the timeout by manually adding code that ignore completion of the download, so that might be tricky to test from the PR build.
    3. Test that reload content/disk cache editor work as expected.
            To do that you can either watch your network traffic or change assets (delete/modify them) in between reloads
    4. Asset uploads. To test this, simply drag and drop an fix on your interface and check the "Add to scene" box. If it shows up in front of you, upload works.
    5. Aside from that, just go to a domain and wait for everything to load. Then look for download failure messages in the log. It'll look like this:
    ```
    [02/26 16:35:55] [DEBUG] [27072] [agent] Requesting sound at "https://s3.amazonaws.com/hifi-public/sounds/08_Funny_Bone.wav"
    [02/26 16:35:55] [DEBUG] [27072] [agent] Starting request for: https://s3.amazonaws.com/hifi-public/sounds/08_Funny_Bone.wav
    [02/26 16:35:55] [DEBUG] [27072] [agent] Error loading  QUrl("https://s3.amazonaws.com/hifi-public/sounds/08_Funny_Bone.wav")
    [02/26 16:35:55] [DEBUG] [27072] [agent] Failed to load: https://s3.amazonaws.com/hifi-public/sounds/08_Funny_Bone.wav
    ```
        
- ACs can download asset:
    I used `ambient.js` with this sound `http://hifi-public.s3.amazonaws.com/birarda/medium-crowd.wav`.

Atlante45:
@ZappoMan That's all I personally tested, I can't really think of anything else to check.
If you see anything off in that list, please let me know.

jherico:
You may wish to load the playa, and then after caching, restart the application and load the playa again to verify there's no network traffic for resources.  I originally stubmled over this problem while investigating slow loading of OBJ file textures in playa.  However, it may be that this issue is orthogonal and can be considered out of scope for this PR.

OmegaHeron:
I'd like to absolutely clear about how HF wants cache to work so I don't consider a result failure that is not. Is it like so;

modelSpecified = a model defined in models.json file

`if modelSpecified not in cache load model from reference and place in cache`
`if modelSpecified in cache and modelExists at source definition load model from cache`
`if modelSpecified in cache and not modelExists at source definition do not load from cache and remove cache model`
`if modelSpecified in cache and not sameAs model at source definition load from source and update cache`

Is that, at least approximately, a valid view of how cache is supposed to work?  I don't want to burn a lot of time documenting what I think I already see as a failure until I'm 100% clear how HF intends cache to function.

OmegaHeron:
Additional question.

What is value for time to live as cached object for assets cached from sources HTTP, ATP and File?

Is HTTP cache life determined from HTTP server response to request for a cached item?

Are ATP and File references considered immortal until cache is manually cleared?

Atlante45:
@jherico Yes I noticed that as well while working on this.
It wasn't my intention to fix this in that PR. One of the other big issues of that code is that it actually locks the current thread while waiting for the resource. Which means we end up with an inactive thread in our global thread pool which only has 2.

@OmegaHeron 
File = Never cached
ATP = Immortal, if it is in the cache, it'll be used, but it might get reclaimed by the caching algorithm when we run out of memory for the cache.
HTTP = Follows HTTP caching rules, based on the server response. Unless it is forced by the app to reload from the network.

Regarding HTTP, I was talking about that with @ZappoMan on Friday night.
There is 2 path we can choose:
- Respect caching rules to the letter. Don't reload from network unless the resource is expired. (current)
- Check HEAD of the files and reload from network if the network resource is more recent.

The problem with the second option being that we would pay the cost of establishing a TCP connection for every HTTP resource we load even if it is in cache and not expired.


Note on ATP assets: ATP asset are identified by there hash, so they will by definition never expire since changing the data would change the hash.

OmegaHeron:
Thanks @Atlante45 

Now to this...
"Note on ATP assets: ATP asset are identified by there hash, so they will by definition never expire since changing the data would change the hash."

Right - it's a sha256 on binary contents of asset so, yes - you can't have a cache collision since any change would generate a new sha256 named file to reference.  But - if a file doesn't exist on ATP server it should not display - same as with how HTTP works.  Why?  One might be providing ATP services for others for a fee - you should not continue seeing assets that can only exist in cache for users who have seen asset(s) before if referenced ATP source no longer has data.  But - that's simply my opinion there.  From here on my comments will be based upon this acting as defined in your comment.

ZappoMan:
http://www.mobify.com/blog/beginners-guide-to-http-cache-headers/

ZappoMan:
@Atlante45 - Although I realize that fixing the honoring of caching rules was not the intent of this PR (maybe you should rename the PR so it's not confusing)...

I don't think we're currently "respecting caching rules to the letter"... as you can see from the article I've linked we are probably not actually honoring many of the official cache header rules...

OmegaHeron:
Issue - cache clearing.

Place a model by drag/drop and have placed in scene.  Verify model displays.  Exit Interface and launch again verifying model still displays.  Find ATP model at ATP server and delete.  Exit Interface and launch again verifying model still displays.

Activate Developer -> Network -> Reload Content
ATP asset that does not exist at ATP sever is still displayed.
Restart Interface, ATP asset that does not exist at ATP sever is still displayed.

Under developer -> Network -> Disk Cache Editor press Clear and confirm.
Quit Interface and start again.
ATP asset that does not exist at ATP server is still displayed.

(Windows specific) Quit Interface and delete \Users\your_user\appData\Local\High Fidelity\
Launch interface and note model that does not exist is not displayed.

It seems that neither action clears cache as an asset that does not exist on ATP server is still displayed.  This may be a bug in how PR Builds work as there's some logic with them to supposedly use PR specific configurations, but, that doesn't actually work.  For example - delete the PR specific configuration and launch PR build - it still has settings.  Delete production build configuration and note PR build starts as expected for a system not having a config for Interface.

OmegaHeron:
Oh - I see now.  I'm digging far too deep into caching for intent of this PR.

For the new view of this - yes, it now caches HTTP, ATP and File references so within that narrow definition this passes.

Atlante45:
Indeed that's odd. Works fine on my dev build, let me test the PR build.

cozza13:
Just reading over the notes from the testing. I believe this passes.

interface/src/ui/DiskCacheEditor.cpp
```diff
@@ -88,17 +85,15 @@ void DiskCacheEditor::makeDialog() {
...
-    QPushButton* refreshCacheButton = new QPushButton(_dialog);
-    Q_CHECK_PTR(refreshCacheButton);
-    refreshCacheButton->setText("Refresh");
-    refreshCacheButton->setToolTip("Reload the cache stats.");
-    connect(refreshCacheButton, SIGNAL(clicked()), SLOT(refresh()));
-    layout->addWidget(refreshCacheButton, 3, 2);
-    
+
+    _refreshTimer = new QTimer(_dialog);
+    _refreshTimer->setInterval(100);
```
ZappoMan: gak magic numbers...

hifi-gustavo:
Build console output for commit: 14ae35e741a17d7f4998810f03fe068edc7c61fe available at: https://gist.github.com/624e24f281d070572df6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7213.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7213.dmg

PR #7214, opened 2016-02-26
handle parenting loops

sethalves:
- when parenting loops are detected, break them rather than crashing


sethalves:
https://app.asana.com/0/26485730907942/93843173735217

sethalves:
to test:
- use edit.js to create a box.  Set the box's parentID to the box's ID.
- Interface will log that a parenting loop was detected and the box will probably end up near the origin with no parentID.

hifi-gustavo:
Build console output for commit: c0c5084149f9724a6dbab7e587e28d99e03bee4a available at: https://gist.github.com/53de89edccf7da239ef0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7214.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7214.dmg

jherico:
Can we not solve this by just disallowing loops in parent / child relationships rather than attempting to patch the API to not crash if it's encountered?  There are well established algorithms for detecting loops in parenting hierarchies.  Granted, this check might have to be done server side in order to guarantee that the loop could be detected.

sethalves:
The trouble is that knowing a parent ID doesn't necessarily mean that we know about the parent (yet).  So... it's tempting to just walk up the chain to the root and see if you encounter yourself, but parts of the chain may still be on their way over the network or out of a json file.

jherico:
That's why I suggested a server side check, although i can see the complexity involved in trying to coordinate the local view of things with a partially visible tree.  Still, couldn't we disallow parenting operations on anything where we don't have a clearly visible path all the way to a root (parentless) node?

sethalves:
Well, setParentID is called by the code that reads json and the code 
that receives entities over the network.  The entities don't come in any 
predictable order -- we'll sometimes get children before parents.  We 
would have to have some sort of pendingParentID that only got copied to 
parentID once we were able to traverse the tree to the root.  We would 
have to keep a list of these and check them continually (though we 
already do something like this).  The entity-server would never reliably 
be able to do this, because it never gets details about avatars.

It's not unreasonable... what I've done here does have the downside of 
artificially limiting the depth of the tree.

OmegaHeron:
Tested on Windows

Placed cube - set its parent ID to self object moved from placed location to 0,0,0 as expected and log complained, as expected, [02/29 12:34:30] [DEBUG] Parenting loop detected.  Parent ID returns to null value.  With current production build Interface crashes.  So - this does solve crash case.

Now - this may be gods of random at play, but, one time I saw entity server crash after setting parent ID to self ID - other times interface log responded as if it couldn't communicate with entity server, but, it remained running and Interface stopped complaining after going through a new handshake with remote.  I can't reproduce entity server crash seen so I'm only documenting it as a thing to watch for in case it's not just a fluke.  Entity server running is on Linux and at level of most current production release.

sethalves:
An entity-server not running this code would still be susceptible to this crash (this is both a client and a server fix).  Was the server running from this branch or from master?

OmegaHeron:
My response said "Entity server running is on Linux and at level of most current production release." so that answers that.  I was testing against current production AC/DS while this branch compiles.  Not sure how common this might be, but, it seems that if it's incompat with existent AC code in this manner it should probably get a version bump.

Nex-Pro:
Just tested this got a crash with the current release. No crash with 7214 parent id reset to default.

sethalves:
It's not really a compatibility change, it's just a fix.  The protocol between them doesn't change -- I don't think we should bump the version.  I should have been more clear in the test procedure.

OmegaHeron:
Tested with AC compiled from this branch.  No issues found.  Passes.

Nex-Pro:
tried this with two boxes 
one moved to 0,0,0 
taking the other with it's relative offset to 1.1360, 0.4504, -0.3999

Nex-Pro:
ok I noticed a side effect.   moving one back to where I was at 3000, 3000, 3000

moved the other to 16384.0 , 6004.5498, 6001.399

this is clearly wrong

sethalves:
Nex-Pro -- I'm not sure I understand what you mean.  Could you list out the steps?

Nex-Pro:
ok made two boxes.  Slightly apart at around 3000, 3000, 3000
parented box1 to box2 
parented box2 to box1
both boxes got moved one to 0,0,0 the other the same relative distance ( about 1 meter from the other ) 
when I selected the one that was  at 0,0,0 and set it's position to be near me at 3000, 3000, 3000

the other ended up at 16384.0 , 6004.5498, 6001.399

sethalves:
okay Nex-Pro -- I think there's a change I can make to improve this.  I'll make a check-in in a few minutes.

Nex-Pro:
8 bit programmers love 16384

sethalves:
Okay @Nex-Pro try it again once this builds.  The boxes should no longer move to the origin (or to some crazy place), but they may move a bit.

Nex-Pro:
will do

hifi-gustavo:
Build console output for commit: bb9d1556c6e148f19449ab0398655da9b8a90085 available at: https://gist.github.com/53de89edccf7da239ef0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7214.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7214.dmg

bluetoe:
I tried this with no child and parented to self = moved to 0,0,0 and set parent to 0
with child and child looped to parent = moved to 0,0,0 and set parent parent to 0
with two childs parented to parent and parent looped to one child = moved and set parent parent to 0
with two childs chained to parent (one child parented to other child which is parented to parent) when parent is parented to second child = moved and  set second parent to 0.

so works as expected, although there was one instance where i ended up with unmovable invisible childs with no parent to be found, at least within 16k. But I was not able to repeat this and all the other times I tried it everything worked fine. I'll try it with the new version now.

Nex-Pro:
Ok tested this with two boxes.   Result of making a parent loop is that the first parenting is cleared.
there's a small movement in the y axis  maybe 50cm.  The movement is up or down dependent on which way the loop parenting  is attempted.

sethalves:
That sounds about right.  When clearing the parentID, there's no way to really know where the thing should be -- I'm falling back on another property which will place the entity reasonably close to (but not exactly at) where it was.

Nex-Pro:
Cool, this stuff explains your hairstyle :)

bluetoe:
i tried all again with the new pr, results were a bit different. each parent was moved slightly up before changing to loop back to the parent so I could get a visual that something had happened.

single entity = move back to original position and parent set to 0
two entities = move back to original position and child's parent set to zero

three entities one parent (a) with two children(b and c) (tree) = moves all back to original position, sets parent's parent (a) to zero but after leaving edit mode sends second child (c) to x,y,z double of original position. Leaving all still linked but at different positions.

three entities one parent(a) with one child (b) that has a child (c) (chained) = moves all back to original position, sets parent's parent (a) and child's child (c) parent to zero but after leaving edit mode linked parent (a) and first child (b) move to 0,0,0 leaving unlinked second child (c) at original position.

So, sort of works but, changes position of one or more when more than two are linked. hope this was helpful

AndrewMeadows:
I'm approving the code for this.  It is too bad we can't detect/sort/send the full hierarchy, but I don't have a better idea on how to do this while maintaining our flexibility to connect anything to anything else.

PR #7219, opened 2016-02-27
More accurate X/Z movement with edit.js

PhilipRosedale:
Plane for colliding pointer was center of object, which could be very different from where the pointer actually was over, meaning that it was easy for object to move wrong relative to pointer. 

Now the point of collision with the selection box is used, meaning objects stay under the cursor.  

Next:   1.  Use actual collision point with the entity itself rather than selection box. 
            2.  Don't allow object to move too far away if the plane passes too close to the pick ray origin.

### DIFF
--- a/examples/libraries/entitySelectionTool.js
+++ b/examples/libraries/entitySelectionTool.js
@@ -2264,7 +2264,6 @@ SelectionDisplay = (function() {
         greatestDimension: 0.0,
         startingDistance: 0.0,
         startingAzimuth: 0.0,
-        lastVector: { x: 0, y: 0, z: 0 },
         onBegin: function(event) {
             SelectionManager.saveProperties();
             startPosition = SelectionManager.worldPosition;
@@ -2337,7 +2336,6 @@ SelectionDisplay = (function() {
             var azimuth = translateXZTool.azimuth(pickRay.origin, pick);
             if ((translateXZTool.startingAzimuth > 0.0 && azimuth < MIN_AZIMUTH) || 
                 (translateXZTool.startingAzimuth < 0.0 && azimuth > MIN_AZIMUTH)) {
-                //vector = translateXZTool.lastVector;
                 if (wantDebug) {
                     print("Azimuth = " + azimuth);
                 }
@@ -2355,8 +2353,6 @@ SelectionDisplay = (function() {
                     return;
                 }
             }
-
-            translateXZTool.lastVector = vector; 
 
             // If shifted, constrain to one axis
             if (event.isShifted) {
@@ -3797,7 +3793,7 @@ SelectionDisplay = (function() {
     };
 
     that.mousePressEvent = function(event) {
-         
+        var wantDebug = false; 
         if (!event.isLeftButton) {
             // if another mouse button than left is pressed ignore it
             return false;
@@ -3823,7 +3819,7 @@ SelectionDisplay = (function() {
 
         if (result.intersects) {
 
-            var wantDebug = false;
+            
             if (wantDebug) {
                 print("something intersects... ");
                 print("   result.overlayID:" + result.overlayID + "[" + overlayNames[result.overlayID] + "]");
@@ -3920,7 +3916,10 @@ SelectionDisplay = (function() {
 
         if (!somethingClicked) {
 
-            print("rotate handle case...");
+            if (wantDebug) {
+                print("rotate handle case...");
+            }
+            
 
             // After testing our stretch handles, then check out rotate handles
             Overlays.editOverlay(yawHandle, {
@@ -3988,15 +3987,17 @@ SelectionDisplay = (function() {
                         break;
 
                     default:
-                        print("mousePressEvent()...... " + overlayNames[result.overlayID]);
+                        if (wantDebug) {
+                            print("mousePressEvent()...... " + overlayNames[result.overlayID]);
+                        }
                         mode = "UNKNOWN";
                         break;
                 }
             }
-
-            print("    somethingClicked:" + somethingClicked);
-            print("                mode:" + mode);
-
+            if (wantDebug) {
+                print("    somethingClicked:" + somethingClicked);
+                print("                mode:" + mode);
+            }
 
             if (somethingClicked) {
 
@@ -4142,17 +4143,22 @@ SelectionDisplay = (function() {
                         translateXZTool.pickPlanePosition = result.intersection;
                         translateXZTool.greatestDimension = Math.max(Math.max(SelectionManager.worldDimensions.x, SelectionManager.worldDimensions.y), 
                             SelectionManager.worldDimensions.z);
-                        print("longest dimension" + translateXZTool.greatestDimension);
-                        translateXZTool.startingDistance = Vec3.distance(pickRay.origin, SelectionManager.position);
-                        print("starting distance" + translateXZTool.startingDistance);
-                        translateXZTool.startingAzimuth = translateXZTool.azimuth(pickRay.origin, translateXZTool.pickPlanePosition);
-                        print("starting azimuth" + translateXZTool.startingAzimuth);
+                        if (wantDebug) {
+                            print("longest dimension: " + translateXZTool.greatestDimension);
+                            translateXZTool.startingDistance = Vec3.distance(pickRay.origin, SelectionManager.position);
+                            print("starting distance: " + translateXZTool.startingDistance);
+                            translateXZTool.startingAzimuth = translateXZTool.azimuth(pickRay.origin, translateXZTool.pickPlanePosition);
+                            print(" starting azimuth: " + translateXZTool.startingAzimuth);
+                        }
+                        
                         mode = translateXZTool.mode;
                         activeTool.onBegin(event);
                         somethingClicked = true;
                         break;
                     default:
-                        print("mousePressEvent()...... " + overlayNames[result.overlayID]);
+                        if (wantDebug) {
+                            print("mousePressEvent()...... " + overlayNames[result.overlayID]);
+                        }
                         mode = "UNKNOWN";
                         break;
                 }
--- a/examples/libraries/entitySelectionTool.js
+++ b/examples/libraries/entitySelectionTool.js
@@ -16,6 +16,11 @@ HIFI_PUBLIC_BUCKET = "http://s3.amazonaws.com/hifi-public/";
 SPACE_LOCAL = "local";
 SPACE_WORLD = "world";
 
+function objectTranslationPlanePoint(position, dimensions) {
+    var newPosition = { x: position.x, y: position.y, z: position.z };
+    newPosition.y -= dimensions.y / 2.0;
+    return newPosition;
+}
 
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 47504a068ee43969869c3e426c2ba2427b7316b3 available at: https://gist.github.com/6ab1a9dcf70fe84a32e1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.dmg

ZappoMan:
test this please

hifi-gustavo:
Build console output for commit: 47504a068ee43969869c3e426c2ba2427b7316b3 available at: https://gist.github.com/6ab1a9dcf70fe84a32e1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.dmg

ZappoMan:
Testing notes: 

1) run using this version of [edit.js](https://rawgit.com/PhilipRosedale/hifi/editFix2/examples/edit.js)
2) Edit some stuff... it should be easier.

ZappoMan:
I don't think this fixes the issues.... here are two examples that perform poorly on production and equally poorly with this version:

1) place an object at eye level vertically. Select it, and then move the cursor slowly UP (away from your hand) you will see the object move TOWARD you... but pushing away should move the object AWAY from you... as it does if the object is below or above eye level. This case fails in production and the new version.

2) with an object placed at eye level vertically - select it and attempt to move the mouse left to right and right to left. You will notice that you see dramatic movements in the "Z" plane (away from or near you)... this does not happen when you select objects that are at eye level...This case fails in production and in the new version.

PhilipRosedale:
Actually, if an object is above the eye level, moving the mouse away from yourself will bring the object toward you - this is (now) the correct behavior for using the mouse to direct motion along a plane.  

The way to imagine this is to imagine turning on the grid tool and setting the height to correspond to the point where the pointer is touching the object.  If the object/grid is above you, moving the mouse 'up' means the pointer is now over a point on the surface that is closer to you.  If the grid is below you, moving the mouse 'up' moves the object away from you, as the pointer is now falling on a point on the grid farther away from you (toward the vanishing point where the grid reaches the horizon). 

So essentially the behavior is now correct to always make the object stay under the mouse pointer, which is the whole idea with these kinds of editing tools... the object follows the point where the pick ray from your eye and going through the pointer intersects with an X/Z plane. 

I would submit that is a lot better than before (where the intersection plane was at the center of the object, not the place where you clicked on the object).  Try using production to move a tall skinny box and you will immediately see what I mean.

PhilipRosedale:
As a further example, try out https://www.tinkercad.com  It is a simple web-based 3D modeler, that lets you drag shapes onto a grid and drag them around.  It was a good reference for our early work.  

Tinkercad works just like this PR now makes edit.js work... if the object is above you, it's motion is toward you for upward mouse movements.   

They do additionally impose a limit on how far away an object can be moved if the angle is close to the vanishing point - I will try that.   I'm not sure how good that is though because moving the object in this way might put it through a wall and then invisible to you.  Possibly just not moving at all if the selected point/plane is close in height to your eyes might also work.

ZappoMan:
Ok - I see that it is behaving as you described and as tinkercad does... I personally think that's confusing... but I guess if that's the behavior you want, then this PR does do it. What about the problem with it still doing dramatic movement in the Z plane when the object is close to eye level?

PhilipRosedale:
More details for testing:  

To explore current behavior versus new behavior, create a box in front of yourself and stretch it into a tall thin shape (make it 2 meters tall or so).   Adjust it so that the bottom is at your avatar's chest height or so.  Try clicking on various points on the box (higher and lower) and dragging it around.  With the production version, you will notice that often the box does not stay in the same position under your cursor, and that the amount that it moves relative to your cursor can vary, and can invert (box moves opposite way of what you expect).   Using this new version, you should see the box stay under your cursor no matter where you drag it. 

There is still the problem of really fast/far movement when your mouse nears the horizon - I am still thinking how best to improve that.

PhilipRosedale:
I'm really quite embarrassed I can't figure out how to commit as myself and not AlessandraSigna.  :(

I've added three new things:  

1.  Drag too close to the XZ horizon and the object will stop moving. 
2.  Drag the object too far away (so it becomes too small) and it will stop moving 
3.  Drag above the horizon (or below if you started above) and object will stop moving. 

Also I have removed some debug spam that was in there from earlier, edit.js should not spam stuff to the logs while being used unless various wantDebug variables are set.

hifi-gustavo:
Build console output for commit: e9aed839205ec314733215c2febf1be85caa508b available at: https://gist.github.com/6ab1a9dcf70fe84a32e1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.dmg

hifi-gustavo:
Build console output for commit: cee79c8274865f2dcd6653798832fe58fae8ae29 available at: https://gist.github.com/6ab1a9dcf70fe84a32e1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.dmg

examples/libraries/entitySelectionTool.js
```diff
@@ -2297,16 +2307,56 @@ SelectionDisplay = (function() {
...
+                return;
+            }
+
             var vector = Vec3.subtract(pick, initialXZPick);
 
+            // If the mouse is too close to the horizon of the pick plane, stop moving
+            var MIN_AZIMUTH = 0.02;   //  Radians 
+            var azimuth = translateXZTool.azimuth(pickRay.origin, pick);
+            if (wantDebug) {
+                    print("Start Azimuth: " + translateXZTool.startingAzimuth + ", Azimuth: " + azimuth);
```
ZappoMan: this formatting is messed up... you should only indent 4 spaces here

ZappoMan:
hmmm... something still seems wrong here. When I try to move things in the X/Z plane, they don't seem to move at all... or they jump wildly... I was in desktop mode, and trying to move a "normal sized" object around. I will add more details.

ZappoMan:
Here's a repo of strange behavior:
1) run interface in desktop mode
2) run your version of edit
3) create box (notice it appears in front of you)
4) click on face of new box and slowly move mouse upward -- which should move it away from you...notice the box doesn't move at all.
5) even if you lower or raise the box... no movement...
6) move mouse right, left, up, or down -- no movement...

examples/libraries/entitySelectionTool.js
```diff
@@ -2297,16 +2307,56 @@ SelectionDisplay = (function() {
                 visible: false
             });
         },
+        azimuth: function(origin, intersection) {
+            return (origin.y - intersection.y) / Vec3.distance(origin, intersection);
```
huffman: Maybe I am misunderstanding, but shouldn't this be altitude?

examples/libraries/entitySelectionTool.js
```diff
@@ -2297,16 +2307,56 @@ SelectionDisplay = (function() {
...
+                if (wantDebug) {
+                    print("Pick ray does not intersect XZ plane.");
+                }
+                return;
+            }
+
             var vector = Vec3.subtract(pick, initialXZPick);
 
+            // If the mouse is too close to the horizon of the pick plane, stop moving
+            var MIN_AZIMUTH = 0.02;   //  Radians
```
huffman: The azimuth function seems to return a ratio, not radians

PhilipRosedale: true.  In the limit of small angles, x/y --> atan(x/y), etc.  

i'll change the comment

examples/libraries/entitySelectionTool.js
PhilipRosedale: duh yes you are right there too.  changing to different name altogether

hifi-gustavo:
Build console output for commit: 08eacc436f3d45251c5abb3a18728838e50be87a available at: https://gist.github.com/6ab1a9dcf70fe84a32e1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.dmg

hifi-gustavo:
Build console output for commit: d1bc31afb64f17001b4c38332c09943f511b50fb available at: https://gist.github.com/6ab1a9dcf70fe84a32e1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.dmg

hifi-gustavo:
Build console output for commit: 42f6ede43cee55399012a097abe585244b401e9c available at: https://gist.github.com/6ab1a9dcf70fe84a32e1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7219.dmg

ZappoMan:
OK @PhilipRosedale - that seems better... can we get someone else to also QA it and see if it works better for them.

bluetoe:
I know this is a bit late to be commenting on this but when I started checking this I got sidetracked by phone that I had to deal with and when I got back, Thoys pointed out that it had already been merged.

I find at least two issues with this change, one being simply that I agree with ZappoMan in that it is confusing as it changes expected behavior depending on where you first click on the object. For instance when trying to edit a zone you have no visible extent of the object being edited, knowing what is above or below eye level is not always apparent. When positioning stuff, habit says I drag it towards me to bring it closer, push away to make it move away.

It also has problems with extent of movement. When moving something away, where it stops depends on how fast you are moving the cursor. Move slowly it stops closer, move fast it can get very tiny. Also the extent is different depending on if you are above or below eye level. If you are below, moving away can cause it to jump from far away to very close as you pass the eye level plane.

I guess it is too late but I felt I should at least say something.

sethalves:
It's not too late -- some of your concerns are shared by ZappoMan and PhilipRosedale.  If we can come up with a consistently better way, we'll switch to it.  The stopping distance may be a remaining bug.

PR #7223, opened 2016-02-29
Avoid losing resource while loading texture images

zzmp:
The TextureCache's NetworkTexture downloads binary image data, marks it as loaded, and then reads it in a separate thread, using the ImageReader. The ImageReader calls back into NetworkTexture::setImage, using a weak ref to its NetworkTexture. If the weak ref is invalid, the texture will not be loaded.

The weak ref may be invalidated if the NetworkTexture is no longer used (e.g. if the user navigates to a new domain). Even though it is no longer used, it is usually kept in the _unusedResources cache, so it will not be reloaded if it was already marked as loaded.

Thus, if a NetworkTexture is invalidated while its ImageReader is processing, the image will never be set, as the ImageReader's weak ref will be invalid and the binary image data will be marked as loaded, and not be reloaded.

This PR fixes this by storing a NetworkTexture's ImageReader as a member on the texture (as the ImageReader stores the binary image data), and retriggering the read (still in a separate thread) every time the NetworkTexture is fetched from the _unusedResources cache if it is not yet fully loaded.


Atlante45:
This is not QA Ready.

hifi-gustavo:
Build console output for commit: 0f18872bb730c7b74ec4dff6caa831e8bad06c7e available at: https://gist.github.com/5c414cd57ff85c348a0e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.dmg

hifi-gustavo:
Build console output for commit: a86cf510be2672cadab6d4d9b0c240ab80bd0c7a available at: https://gist.github.com/5c414cd57ff85c348a0e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.dmg

hifi-gustavo:
Build console output for commit: a2718be32e5f2a4265c2aff1223d20389c48db4d available at: https://gist.github.com/5c414cd57ff85c348a0e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.dmg

hifi-gustavo:
Build console output for commit: 372053c7490f5c075c34ad9af138527bf48964a6 available at: https://gist.github.com/5c414cd57ff85c348a0e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.dmg

zzmp:
https://app.asana.com/0/74051501495175/95025056969423

#### Testing

This is a concurrency issue, and is hard to test :smile_cat:. I reproduced it using a debugger to prevent certain lines from executing.

- Go to a domain and remove *everything* in the entity list in edit.js.
- Add back a single entity with a texture (https://hifi-public.s3.amazonaws.com/sam/models/mitsuba/mitsuba%20reUVmapped_with%20phong.fbx).
- Position yourself to view this entity.
- Close interface.

// On a build of `master`
- Set a breakpoint in TextureCache.cpp before calling `QThreadPool::globalInstance->start`.
- Open interface, and skip all occurences of that call.
- Remove the breakpoint.
- Navigate to a different domain and back. This will move the resource to the _unusedResources cache.
- Observe the entity with a still unloaded texture.

// On a build of this PR
- Repeat.
- Observe the entity with a loaded texture.
- If you want, repeat with the breakpoint back, and notice that it is called every time the model is first added to the used cache.

hifi-gustavo:
Build console output for commit: 0e2dd96507b9b8e1311b5aada3dcbb9a6c59985e available at: https://gist.github.com/5c414cd57ff85c348a0e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.dmg

hifi-gustavo:
Build console output for commit: 0e2dd96507b9b8e1311b5aada3dcbb9a6c59985e available at: https://gist.github.com/5c414cd57ff85c348a0e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.dmg

hifi-gustavo:
Build console output for commit: e93109a9dce1da9692e4ed91cfaa7f0e24f9ad68 available at: https://gist.github.com/5c414cd57ff85c348a0e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.dmg

OmegaHeron:
@zzmp - I;m working on this one and having a little bit of trouble in trying to follow your test process.  Could you copy in code and point to precisely where you set your breakpoint please?  I'm doing this from Windows - VS2k13 community edition.

I see two refs to what I think I'm looking for in TextureCache.cpp

void NetworkTexture::downloadFinished(const QByteArray& data) {
    // send the reader off to the thread pool
    QThreadPool::globalInstance()->start(new ImageReader(_self, data, _url));
}

void NetworkTexture::loadContent(const QByteArray& content) {
    QThreadPool::globalInstance()->start(new ImageReader(_self, content, _url));
}

So far I've not been able to break - skip and end up with model not showing texture after changing domains with production code.  I'm just (barely, maybe) proficient enough with Debugger to do this with a little handholding++

zzmp:
@OmegaHeron you want the second one, but I would just set a break point on both.
You want to have no break points while you load your scene, then set a breakpoint, then load an uncached texture. When it hits the `start` method, *skip that method*. This way the texture does not load, and the NetworkTexture thinks it is already pending the load.

OmegaHeron:
Thanks - I'll give it a go tomorrow if it's still open.  Getting late here :)

zzmp:
Thank *you*! Let me know if you have more questions, I'm happy to answer them.

OmegaHeron:
@zzmp - I was able to repeat your test method with current Interface production build and this PR - forcing fail to load on production leads to it remaining untextured if model leaves view context and is returned to - this PR does load texture after being forced to miss it.  Tested Windows 7 and passes.

libraries/model-networking/src/model-networking/TextureCache.cpp
```diff
@@ -237,37 +237,42 @@ NetworkTexture::TextureLoaderFunc NetworkTexture::getTextureLoader() const {
...
-        const QUrl& url) :
-    _texture(texture),
+ImageReader::ImageReader(const QByteArray& data, const QUrl& url) :
     _url(url),
     _content(data)
 {
+    setAutoDelete(false);
+}
+
+void ImageReader::start(const QWeakPointer<Resource>& resource) {
```
jherico: What is the point of using a weak pointer here instead of a strong pointer?

libraries/model-networking/src/model-networking/TextureCache.h
```diff
@@ -44,36 +45,36 @@ enum TextureType {
     CUSTOM_TEXTURE
 };
 
-/// Stores cached textures, including render-to-texture targets.
```
jherico: triple slash is a mechanism to signal that this comment should be interpreted as documentation by a doxygen compatible parser.  Please do not remove them.

samcake:
I thought that this pr should also fix this bug but unfortunatelly not. In the referenced model, there are 3 materials using textures which have Different relative paths but the SAME filename and only the first version is loaded and used for the 3 different materials. Can you take a look please ?
Cheers

https://app.asana.com/0/26225263936266/97534968960857
"
An FBX with multiple models and multiple materials imports, but only one material is applied to all the models. 
Please see example images.

1. Import this FBX with multiple textures into Maya or Blender.  Notice that it has three different materials for each coin pile: gold, steel and copper.
http://hifi-public.s3.amazonaws.com/blog_content_assets/pbr%20Blog/CoinPile.fbx

2. Upload the same FBX into high fidelity.  
3. Notice that they all have the gold texture.

Expected behavior:  Each coin type would retain its respective material.  
"

libraries/model-networking/src/model-networking/TextureCache.cpp
zzmp: If the texture is loaded, but not read through the ImageReader, it's shared_ptrs may all go out of scope. When this happens, the cache recreates it as a new shared_ptr, but any other pointers to it will be invalidated.

Using a QWeakPointer here allows the texture to be released during loading, and the ImageReader to know about it. This can happen during domain changes. It allows the reading of the image to be delayed until it is used again. The texture holds a ref to the ImageReader, so it can be restarted with a new QWeakPointer later.

zzmp:
@samcake I'll look at that separately from this PR. I don't think they are related.

hifi-gustavo:
Build console output for commit: fda1ed5a5cf55d782754de361481190542e76d0c available at: https://gist.github.com/5c414cd57ff85c348a0e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7223.dmg

libraries/model-networking/src/model-networking/TextureCache.cpp
```diff
@@ -237,37 +237,42 @@ NetworkTexture::TextureLoaderFunc NetworkTexture::getTextureLoader() const {
...
-    QThreadPool::globalInstance()->start(new ImageReader(_self, data, _url));
+    _reader.reset(new ImageReader(data, _url));
+    loadImage();
 }
 
-void NetworkTexture::loadContent(const QByteArray& content) {
-    QThreadPool::globalInstance()->start(new ImageReader(_self, content, _url));
+void NetworkTexture::loadImage() {
+    if (_reader) {
+        static_cast<ImageReader*>(_reader.get())->start(_self);
```
hyperlogic: I don't think you need this static cast, if you change the _reader type to a std::unique_ptr<>, I *think* you can forward declare the ImageReader class if it is not available in the header.  In that case.  This line would just become:

    _reader->start(_self);

libraries/model-networking/src/model-networking/TextureCache.cpp
```diff
@@ -285,9 +290,10 @@ void ImageReader::run() {
     }
     QThread::currentThread()->setPriority(QThread::LowPriority);
 
-    auto texture = _texture.toStrongRef();
-    if (!texture) {
-        qCWarning(modelnetworking) << "Could not get strong ref";
+    auto resource = _resource.toStrongRef();
+    if (!resource) {
+        // The resource may have been released before the reader got to it; just give up
+        qCWarning(modelnetworking) << "Could not get strong ref:" << _url;
```
hyperlogic: I'm concerned that this might cause a memory leak in this case, Should we setAutoDelete() to true here?

libraries/model-networking/src/model-networking/TextureCache.cpp
```diff
@@ -237,37 +237,42 @@ NetworkTexture::TextureLoaderFunc NetworkTexture::getTextureLoader() const {
...
-
+    ImageReader(const QByteArray& data, const QUrl& url = QUrl());
+    void start(const QWeakPointer<Resource>& resource);
     virtual void run();
 
 private:
     static void listSupportedImageFormats();
 
-    QWeakPointer<Resource> _texture;
+    QWeakPointer<Resource> _resource;
```
hyperlogic: Why not make this a QWeakPointer&lt;NetworkTexture&lt;?  Is this ImageReader used by other classes?

libraries/model-networking/src/model-networking/TextureCache.cpp
zzmp: [This cannot be called from within a Runnable](http://doc.qt.io/qt-5.5/qrunnable.html#setAutoDelete):
```
Calling this function after QThreadPool::start() results in undefined behavior.
```

We'd need a different pattern to avoid the leak.
Also, the leak would come from the texture being destructed *twice*, as the first time just moves the memory to an LRU cache.

libraries/model-networking/src/model-networking/TextureCache.cpp
zzmp: It would just require more casting. `Resource` has a method to return a weak pointer to self, which texture inherits.

hyperlogic: Ok, I see.

zzmp:
@hyperlogic's suggested just removing partially loaded textures from the cache, so I'm going to explore that in a new PR.

PR #7230, opened 2016-03-01
Fix inline fbx texture caching

zzmp:
Fixes inline fbx texture caching to cache files under the fbx (`.../*.fbx/cached_texture.png`). These were previously cached in the parent directory (`.../cached_texture.png`), resulting in incorrect cache retrievals when multiple models with inline textures of the same name were uploaded from the same directory.

Also cleaned up some repetition in buildNetworkMaterial while I was in there.

### DIFF
--- a/libraries/model-networking/src/model-networking/ModelCache.cpp
+++ b/libraries/model-networking/src/model-networking/ModelCache.cpp
@@ -309,53 +309,45 @@ static NetworkMaterial* buildNetworkMaterial(const FBXMaterial& material, const 
     auto localBaseUrl = QUrl(textureBaseUrl.url() + "/");
 
     auto textureCache = DependencyManager::get<TextureCache>();
-    NetworkMaterial* networkMaterial = new NetworkMaterial();
 
+    NetworkMaterial* networkMaterial = new NetworkMaterial();
     networkMaterial->_material = material._material;
 
+    auto setupMaterialTexture = [&](const FBXTexture& texture,
+            TextureType type, model::MaterialKey::MapChannel channel,
+            NetworkTexturePointer& networkTexture, QString& networkTextureName){
+        const auto& baseUrl = texture.content.isEmpty() ? textureBaseUrl : localBaseUrl;
+        networkTexture = textureCache->getTexture(baseUrl.resolved(QUrl(texture.filename)), type, texture.content);
+        networkTextureName = texture.name;
+
+        auto map = std::make_shared<model::TextureMap>();
+        map->setTextureSource(networkTexture->_textureSource);
+        material._material->setTextureMap(channel, map);
+        return map;
+    };
+
     if (!material.diffuseTexture.filename.isEmpty()) {
-        const auto& baseUrl = material.diffuseTexture.content.isEmpty() ? textureBaseUrl : localBaseUrl;
-        networkMaterial->diffuseTexture = textureCache->getTexture(baseUrl.resolved(QUrl(material.diffuseTexture.filename)), DEFAULT_TEXTURE, material.diffuseTexture.content);
-        networkMaterial->diffuseTextureName = material.diffuseTexture.name;
-
-        auto diffuseMap = model::TextureMapPointer(new model::TextureMap());
-        diffuseMap->setTextureSource(networkMaterial->diffuseTexture->_textureSource);
+        auto diffuseMap = setupMaterialTexture(material.diffuseTexture,
+            DEFAULT_TEXTURE, model::MaterialKey::DIFFUSE_MAP,
+            networkMaterial->diffuseTexture, networkMaterial->diffuseTextureName);
         diffuseMap->setTextureTransform(material.diffuseTexture.transform);
-
-        material._material->setTextureMap(model::MaterialKey::DIFFUSE_MAP, diffuseMap);
     }
     if (!material.normalTexture.filename.isEmpty()) {
-        const auto& baseUrl = material.normalTexture.content.isEmpty() ? textureBaseUrl : localBaseUrl;
-        networkMaterial->normalTexture = textureCache->getTexture(baseUrl.resolved(QUrl(material.normalTexture.filename)), (material.normalTexture.isBumpmap ? BUMP_TEXTURE : NORMAL_TEXTURE), material.normalTexture.content);
-        networkMaterial->normalTextureName = material.normalTexture.name;
-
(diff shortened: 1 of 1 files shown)


zzmp:
#### Testing

Thanks to @cozza13, who provided this repro:

bugex

>Have a look at the attached picture.
>
What you are seeing there are two different FBX's. Cube and a sphere, each has a texture.jpg (same named texture). The cube is red the sphere is black. Where you see them both black, that is because I uploaded the both to ATP, yet loaded the sphere first (which is black). Where you see them both red, that is because I uploaded them both to the same folder on s3, yet loaded the cube first (which is red). Here they are if you want to try it.
>
Sphere: https://s3.amazonaws.com/hifi-public/cozza13/assetexample/sphere.fbx
Cube:  https://s3.amazonaws.com/hifi-public/cozza13/assetexample/cube.fbx

hifi-gustavo:
Build console output for commit: 19c83e869f41d3d62efc3652e6c4aba94da79546 available at: https://gist.github.com/7d7bdf2cb190cd044b06<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7230.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7230.dmg

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -305,50 +305,49 @@ static NetworkMesh* buildNetworkMesh(const FBXMesh& mesh, const QUrl& textureBas
...
     auto textureCache = DependencyManager::get<TextureCache>();
-    NetworkMaterial* networkMaterial = new NetworkMaterial();
 
+    NetworkMaterial* networkMaterial = new NetworkMaterial();
     networkMaterial->_material = material._material;
 
-    if (!material.diffuseTexture.filename.isEmpty()) {
-        networkMaterial->diffuseTexture = textureCache->getTexture(textureBaseUrl.resolved(QUrl(material.diffuseTexture.filename)), DEFAULT_TEXTURE, material.diffuseTexture.content);
-        networkMaterial->diffuseTextureName = material.diffuseTexture.name;
+    auto setupMaterialTexture = [&](const FBXTexture& texture,
```
ZappoMan: Is there a reason you're making this a lambda? It appears as if you're only reusing `textureBaseUrl ` and `localBaseUrl` from the outer... it feels to me like this would be easier to read if you just made this `setupMaterialTexture()` function a static function in the cpp file scope.

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -305,50 +305,49 @@ static NetworkMesh* buildNetworkMesh(const FBXMesh& mesh, const QUrl& textureBas
...
+        const auto& baseUrl = texture.content.isEmpty() ? textureBaseUrl : localBaseUrl;
+        networkTexture = textureCache->getTexture(baseUrl.resolved(QUrl(texture.filename)), type, texture.content);
+        networkTextureName = texture.name;
 
-        auto diffuseMap = model::TextureMapPointer(new model::TextureMap());
-        diffuseMap->setTextureSource(networkMaterial->diffuseTexture->_textureSource);
-        diffuseMap->setTextureTransform(material.diffuseTexture.transform);
+        auto map = std::make_shared<model::TextureMap>();
+        map->setTextureSource(networkTexture->_textureSource);
+        material._material->setTextureMap(channel, map);
```
ZappoMan: `material` appears to be `const` captured from the outer... but you're calling a setter method on one of it's member variables... that seems odd... is `material._material` mutable?

jherico: It's legal to call a non-const method on a const pointer (or const shared_ptr) because it doesn't actually modify the pointer itself.

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -305,50 +305,49 @@ static NetworkMesh* buildNetworkMesh(const FBXMesh& mesh, const QUrl& textureBas
 }
 
 static NetworkMaterial* buildNetworkMaterial(const FBXMaterial& material, const QUrl& textureBaseUrl) {
+    // Create a local url to cache inline textures
+    auto localBaseUrl = QUrl(textureBaseUrl.url() + "/");
```
ZappoMan: this isn't used in the outer... and is only used in the lambda... along the lines of my comment about making the lambda a first class citizen, it seems like this would be more appropriately contained in that function...

more importantly... it looks like the only place you use this is when you would otherwise use textureBaseUrl... so basically you're either using `textureBaseUrl` or you're using `textureBaseUrl` with `"/"` appended to it... it seems like this could be more clear.. and not require an extra variable.

libraries/model-networking/src/model-networking/ModelCache.cpp
ZappoMan: it also strikes me that you're manipulating `material`

libraries/model-networking/src/model-networking/ModelCache.cpp
zzmp: I made it a lambda so the arguments list wouldn't be too long, as I thought that would be less readable. A static function would also need to be passed the localBaseUrl, textureBaseUrl, and textureCache if they are to be persisted across all textures without recalculating/reloading them.

But you're right, I only use localBaseUrl, textureBaseUrl, and textureCache from the outer. Happy to change it if you still feel it would be more readable.

ZappoMan: I think you'll find that if you reorganize your code a little bit, you'll have the same or fewer parameters to the function... and it will be much more readable...

ZappoMan:
:+1:

hifi-gustavo:
Build console output for commit: d5a5ce8c70e74e51f9e047744bd7475f04c583b5 available at: https://gist.github.com/7d7bdf2cb190cd044b06<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7230.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7230.dmg

cozza13:
tested and this works. Great stuff @zzmp

PR #7233, opened 2016-03-01
Fix the Vive controllers to support menus

jherico:
We map the secondary thumbs to 'cycle camera' and 'context menu', but the vive has no mapping to the secondary thumb.  on the other hand we appear to not map the primary thumb to anything by default, so I'm adding a mapping to the same 'cycle camera' and 'context menu' actions, which should allow them to be triggered from the menu / application button on the Vive hand controllers.

### DIFF
--- a/plugins/openvr/src/ViveControllerManager.cpp
+++ b/plugins/openvr/src/ViveControllerManager.cpp
@@ -311,7 +311,7 @@ void ViveControllerManager::InputDevice::handleButtonEvent(float deltaTime, uint
 
     using namespace controller;
     if (button == vr::k_EButton_ApplicationMenu) {
-        _buttonPressedMap.insert(isLeftHand ? LEFT_PRIMARY_THUMB : RIGHT_PRIMARY_THUMB);
+        _buttonPressedMap.insert(isLeftHand ? BACK : START);
     } else if (button == vr::k_EButton_Grip) {
         _buttonPressedMap.insert(isLeftHand ? LB : RB);
     } else if (button == vr::k_EButton_SteamVR_Trigger) {
@@ -426,10 +426,8 @@ controller::Input::NamedVector ViveControllerManager::InputDevice::getAvailableI
         makePair(LEFT_HAND, "LeftHand"),
         makePair(RIGHT_HAND, "RightHand"),
 
-        makePair(LEFT_PRIMARY_THUMB, "LeftPrimaryThumb"),
-        makePair(LEFT_SECONDARY_THUMB, "LeftSecondaryThumb"),
-        makePair(RIGHT_PRIMARY_THUMB, "RightPrimaryThumb"),
-        makePair(RIGHT_SECONDARY_THUMB, "RightSecondaryThumb"),
+        makePair(BACK, "Back"),
+        makePair(START, "Start"),
     };
 
     //availableInputs.append(Input::NamedPair(makeInput(BUTTON_A, 0), "Left Button A"));
--- a/plugins/openvr/src/ViveControllerManager.cpp
+++ b/plugins/openvr/src/ViveControllerManager.cpp
@@ -426,8 +426,8 @@ controller::Input::NamedVector ViveControllerManager::InputDevice::getAvailableI
         makePair(LEFT_HAND, "LeftHand"),
         makePair(RIGHT_HAND, "RightHand"),
 
-        makePair(BACK, "Back"),
-        makePair(START, "Start"),
+        makePair(BACK, "LeftApplicationMenu"),
+        makePair(START, "RightApplicationMenu"),
     };
 
     //availableInputs.append(Input::NamedPair(makeInput(BUTTON_A, 0), "Left Button A"));
--- a/plugins/openvr/src/ViveControllerManager.cpp
+++ b/plugins/openvr/src/ViveControllerManager.cpp
@@ -303,6 +303,13 @@ void ViveControllerManager::InputDevice::handleAxisEvent(float deltaTime, uint32
     }
 }
 
+// An enum for buttons which do not exist in the StandardControls enum
+enum ViveButtonChannel {
+    LEFT_APP_MENU = controller::StandardButtonChannel::NUM_STANDARD_BUTTONS,
+    RIGHT_APP_MENU
+};
+
+
 // These functions do translation from the Steam IDs to the standard controller IDs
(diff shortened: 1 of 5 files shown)


jherico:
https://app.asana.com/0/27650181942747/95142715746266

hifi-gustavo:
Build console output for commit: 2a03945365927493694c246c6a2dc9a30ba74db4 available at: https://gist.github.com/eb25f09ebf3f7491878e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7233.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7233.dmg

ZappoMan:
I think I know why we didn't do this before... it's because on the Hydra, we map the 1,2,3,4 buttons to PrimaryThumb... and so that means that if you accidentally press any button on the left hydra, you will cycle your camera... I think that will cause confusion...

PhilipRosedale:
I'm more concerned about the Vive working right... testing this there now.

cozza13:
I tried this on the vive and could not get it to work. (it did map the camera cycle back to the primary thumb on the hydra.)

ZappoMan:
Does the Vive have "secondary thumb buttons"? If it does have hardware buttons for secondary thumb then wouldn't the right fix be to map that hardware to the standard secondary thumb?

hifi-gustavo:
Build console output for commit: dbae0caa901a0f9961ed4fb34df2d4d16ea272e9 available at: https://gist.github.com/eb25f09ebf3f7491878e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7233.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7233.dmg

plugins/openvr/src/ViveControllerManager.cpp
```diff
@@ -311,7 +311,7 @@ void ViveControllerManager::InputDevice::handleButtonEvent(float deltaTime, uint
 
     using namespace controller;
     if (button == vr::k_EButton_ApplicationMenu) {
-        _buttonPressedMap.insert(isLeftHand ? LEFT_PRIMARY_THUMB : RIGHT_PRIMARY_THUMB);
+        _buttonPressedMap.insert(isLeftHand ? BACK : START);
```
ZappoMan: wait... don't we also want to signal the primary thumb as well?

cozza13:
I tested this and it works. There is still the issue with the overlay not moving with your head, however this is separate.

plugins/openvr/src/ViveControllerManager.cpp
jherico: if we want to map from Vive.Back to Standard.LeftPrimaryThumb, the place to do that would be in `vive.json`.  The C++ code should only populate the inputs that actually exist for the controller.  I would go so far as to say we should never have a 'primary/secondary thumb/index' control on a device unless that's the official name of that control for the device.

jherico: We should also be adding a mechanism to query what standard controls are actually being fed by something...  like, if you're using Oculus touch and not the xbox controller, you have no d-pad, and scripts should be able to recognize this and act on it.

ZappoMan: but here's the thing... is "Back" and "Start" really the names of these buttons for the vive? Do they say back and start on them? If so, I will merge... if not... this naming seems confusing...

jherico: hoist on my own petard.  will fix.

plugins/openvr/src/ViveControllerManager.cpp
```diff
@@ -426,10 +426,8 @@ controller::Input::NamedVector ViveControllerManager::InputDevice::getAvailableI
         makePair(LEFT_HAND, "LeftHand"),
         makePair(RIGHT_HAND, "RightHand"),
 
-        makePair(LEFT_PRIMARY_THUMB, "LeftPrimaryThumb"),
-        makePair(LEFT_SECONDARY_THUMB, "LeftSecondaryThumb"),
-        makePair(RIGHT_PRIMARY_THUMB, "RightPrimaryThumb"),
-        makePair(RIGHT_SECONDARY_THUMB, "RightSecondaryThumb"),
+        makePair(BACK, "LeftApplicationMenu"),
+        makePair(START, "RightApplicationMenu"),
```
ZappoMan: C'mon now!

hifi-gustavo:
Build console output for commit: f651adbd24c93f4a9c2ef624c287d50518c32861 available at: https://gist.github.com/eb25f09ebf3f7491878e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7233.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7233.dmg

plugins/openvr/src/ViveControllerManager.cpp
jherico: Oh very well.... fixed.

hifi-gustavo:
Build console output for commit: 17852e86d25f1d351f429ff7a01ce0ce28ca58f2 available at: https://gist.github.com/eb25f09ebf3f7491878e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7233.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7233.dmg

jherico:
test this please

hifi-gustavo:
Build console output for commit: 17852e86d25f1d351f429ff7a01ce0ce28ca58f2 available at: https://gist.github.com/eb25f09ebf3f7491878e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7233.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7233.dmg

PR #7236, opened 2016-03-02
entity script safety

howard-stearns:
Entity scripts keep of the which entity they are executing for. When the entity goes away, any created timers do, too. Also, loading interface scripts is not allowed from entity scripts.

### DIFF
--- a/libraries/script-engine/src/ScriptEngine.cpp
+++ b/libraries/script-engine/src/ScriptEngine.cpp
@@ -485,10 +485,10 @@ void ScriptEngine::removeEventHandler(const EntityItemID& entityID, const QStrin
         return;
     }
     RegisteredEventHandlers& handlersOnEntity = _registeredHandlers[entityID];
-    QScriptValueList& handlersForEvent = handlersOnEntity[eventName];
+    CallbackList& handlersForEvent = handlersOnEntity[eventName];
     // QScriptValue does not have operator==(), so we can't use QList::removeOne and friends. So iterate.
     for (int i = 0; i < handlersForEvent.count(); ++i) {
-        if (handlersForEvent[i].equals(handler)) {
+        if (handlersForEvent[i].function.equals(handler)) {
             handlersForEvent.removeAt(i);
             return; // Design choice: since comparison is relatively expensive, just remove the first matching handler.
         }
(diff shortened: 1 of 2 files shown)


howard-stearns:
https://app.asana.com/0/32622044445063/58877925149313
https://app.asana.com/0/32622044445063/78957392999732

howard-stearns:
This is the minimum change that fixes the two bugs cited. It would have been cool to make a special version of Script that knows how to keep track of which timers belong to which scripts, but there would have to be different special version for each entity script (to close over the entity script identifier), and that ends up being the moral equivalent to what I did here. This version is simpler.

I avoided making separate sandboxes for each entity script, as that's an incompatible change (e.g., if folks are doing something evil like communicating between entity scripts through globals). Such a re-arch is not really in the spirit of just fixing the bugs.

howard-stearns:
Testing notes: The entity script I used for testing is attached to https://app.asana.com/0/32622044445063/58877925149313
See the comments and print statements in the test script.

hifi-gustavo:
Build console output for commit: 05943ed35361e3b794822a3b255e1b2260413843 available at: https://gist.github.com/e320db76bb792fcfc939<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7236.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7236.dmg

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1142,6 +1187,32 @@ void ScriptEngine::refreshFileScript(const EntityItemID& entityID) {
...
+    operation();
+    this->globalObject().setProperty("debugEntityID", oldData);
+#else
+    operation();
+#endif
+
+    currentEntityIdentifier = oldIdentifier;
+}
+void ScriptEngine::callWithEnvironment(const EntityItemID& entityID, QScriptValue function, QScriptValue thisObject, QScriptValueList args) {
+    auto operation = [&]() {
```
ZappoMan: why not do this one like this....
```
    doWithEnvironment(entityID, [&]() {
        function.call(thisObject, args);
    });
```

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1055,9 +1095,13 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
         QString file = QUrl(scriptOrURL).toLocalFile();
         lastModified = (quint64)QFileInfo(file).lastModified().toMSecsSinceEpoch();
     }
+    QScriptValue entityScriptConstructor, entityScriptObject;
+    auto initialization = [&]{
+        entityScriptConstructor = evaluate(contents, fileName);
+        entityScriptObject = entityScriptConstructor.construct();
+    };
+    doWithEnvironment(entityID, initialization);
```
ZappoMan: and this one like this...
```
    doWithEnvironment(entityID, [&]{
        entityScriptConstructor = evaluate(contents, fileName);
        entityScriptObject = entityScriptConstructor.construct();
    });
```

libraries/script-engine/src/ScriptEngine.cpp
howard-stearns: No good reason. I started one step at  a time and left it that way. 

(If I hadn't I'd have been tempted to write
  doWithEnvironment(entityID, [&]() { function.call(thisObject, args); });
because I don't like gratuitous newlines. At least with the separate variable there's a reason to break it up.)

I figure the compiler should treat it the same with or without the temporary. (If a compiler can't eliminate temporaries, would good is it?) Am I wrong?

So if it _is_ the same generated code, I don't have an opinion either way as to which is more understandable.

ZappoMan: Personally I feel the version I'm suggesting is more readable... but I won't block the merge for it...

OmegaHeron:
Tested on Windows against production build 4319.  With 4319 running test script, terminating script then deleting green box leads to continual production of red boxes.  With this PR run script, terminate then delete green box stops production of red boxes.  I don't see any negative interactions with anything else, so, it passes for me.

PR #7237, opened 2016-03-02
re-use the node with same sockets during failed domain connection

birarda:
- fixed an issue with duplicates of an unconnected node appearing in domain-server settings and log when attempting to connect

### DIFF
--- a/domain-server/src/DomainGatekeeper.cpp
+++ b/domain-server/src/DomainGatekeeper.cpp
@@ -258,9 +258,26 @@ SharedNodePointer DomainGatekeeper::processAgentConnectRequest(const NodeConnect
     if (onlyEditorsAreRezzers) {
         canRez = isAllowedEditor;
     }
+
+    QUuid hintNodeID;
+
+    // in case this is a node that's failing to connect
+    // double check we don't have a node whose sockets match exactly already in the list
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 35973d4e30d0567096fb10fa1c725b61a5826447 available at: https://gist.github.com/52ae88266cf38f90eec8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7237.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7237.dmg

OmegaHeron:
Tested on Windows (via Interface on same machine -- localhost) and Linux via internet route.  This PR's Interface is able to connect to older server code builds and older Interface can connect to this PR's server side.  Using an older version of Interface known to have packet version < current does not lead to masses of same agent shown as connected, but, it does spam domain-server log with requesting key for user, storing key, could not verify edit rights.

Not sure when Interface quit giving the Interface to Server mismatch warning - but it's missed.  If that's returning to Interface and will stop it from trying to connect to that it can't connect to I'm not overly concerned with log spam as described.

PR #7241, opened 2016-03-02
Add events for grabbing/releasing entity

zzmp:
Adds calls on first grab and release to methods on the entity, if they exist:
`grabEntity(entityID)`
`ungrabEntity(entityID)`

ungrabEntity is called such to mirror grab, although it could be called releaseEntity instead.

### DIFF
--- a/examples/grab.js
+++ b/examples/grab.js
@@ -500,6 +500,9 @@ Grabber.prototype.activateEntity = function(entityID, grabbedProperties) {
     var grabbableData = getEntityCustomData(GRABBABLE_DATA_KEY, entityID, DEFAULT_GRABBABLE_DATA);
     var invertSolidWhileHeld = grabbableData["invertSolidWhileHeld"];
     var data = getEntityCustomData(GRAB_USER_DATA_KEY, entityID, {});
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: bbfe94ba5b680268c55f5d59ad256450b30d59a2 available at: https://gist.github.com/9ee74f6a3539182e3d2c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7241.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7241.dmg

imgntn:
@zzmp are these supposed to be compatible with hand controller grabs or totally different?  if the former, then why not use 'startNearGrab' and 'releaseGrab' events that already exist?  if the latter, perhaps we should name it something like 'mouseGrab' so its not confusing for content creators

sethalves:
the mouse grab.js grab is more like a far grab than a near one... but far grab already has messages, also.

PR #7248, opened 2016-03-04
disable AC child file logging by default

birarda:
- disabled writing of AC child output to log file unless `--log-directory` is passed

### DIFF
--- a/assignment-client/src/AssignmentClientApp.cpp
+++ b/assignment-client/src/AssignmentClientApp.cpp
@@ -96,9 +96,6 @@ AssignmentClientApp::AssignmentClientApp(int argc, char* argv[]) :
     const QCommandLineOption logDirectoryOption(ASSIGNMENT_LOG_DIRECTORY, "directory to store logs", "log-directory");
     parser.addOption(logDirectoryOption);
 
-    const QCommandLineOption noLogFileOption(ASSIGNMENT_NO_CHILD_LOG_FILE_OPTION, "disable writing of child assignment-client output to file");
-    parser.addOption(noLogFileOption);
-
     if (!parser.parse(QCoreApplication::arguments())) {
         qCritical() << parser.errorText() << endl;
         parser.showHelp();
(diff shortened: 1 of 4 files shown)


birarda:
Note: I think we want this to be the default but I need to double check we don't need Server Console changes to handle that.

hifi-gustavo:
Build console output for commit: efa46b8c7b5c5d42a7a48a49cd9311cc0b50b3d8 available at: https://gist.github.com/25c3c96a82ed8afcc989<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7248.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7248.dmg

birarda:
Made it default to not writing logs to file unless `--log-directory` is passed to the AC monitor.

hifi-gustavo:
Build console output for commit: 0b578bf08661b866d548053d535c04e56bce44fc available at: https://gist.github.com/25c3c96a82ed8afcc989<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7248.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7248.dmg

jherico:
In general I'd rather see us expose the Qt logging rules via the command line, with a default set of rules set up to output nothing.  However, that would entail still creating log dirs and log files (just leaving them empty) and this is a reasonable approach.

birarda:
@jherico I think we'd want both options - that categorization can apply to the stdout as well but this specifically enables logging to a file. I assume most direct callers of assignment-client don't need/want that. The Server Console passes a specific log directory for its log handling.

OmegaHeron:
Tested with AC/DS running on Linux - no issues seen, but my setup uses stdout piped to logs vs direct writing (didn't know I could even do so!).  I'll rewrite my startup scripts and better test this later - in so far as it doesn't break existing setup it passes.

OmegaHeron:
Can't find any obvious issue with this - looked at presence of logs from Windows - all seems well.  Tested with my specific Linux setup and all good.  QA Pass.

PR #7256, opened 2016-03-05
Fix the text rendering Bug

samcake:
I broke the text rendering in my previous PR, This is fixing it
Simply make sure the alpha is 1 before going into the packDeferredFragmentLightmap function
 Also fixed the color which was not gamma corrected

### DIFF
--- a/libraries/render-utils/src/sdf_text3D.slf
+++ b/libraries/render-utils/src/sdf_text3D.slf
@@ -15,16 +15,10 @@
 uniform sampler2D Font;
 uniform bool Outline;
 uniform vec4 Color;
-uniform vec4 backgroundColor = vec4(1.0, 0.0, 0.0, 1.0);
 
 // the interpolated normal
 in vec3 _normal;
 in vec2 _texCoord0;
-/*
-layout(location = 0) out vec4 _fragColor0;
-layout(location = 1) out vec4 _fragColor1;
-layout(location = 2) out vec4 _fragColor2;
-*/
 
 const float gamma = 2.2;
 const float smoothing = 256.0;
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 193b98f24fb35580faed499bdcb952712d96ba1a available at: https://gist.github.com/7f642d7603ab11051812<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7256.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7256.dmg

zzmp:
Does this make text always opaque? If so, isn't that a feature degradation?

libraries/render-utils/src/sdf_text3D.slf
```diff
@@ -49,19 +46,15 @@ void main() {
...
     
-    // final color
-   /* _fragColor0 = vec4(Color.rgb, Color.a * a);
-    _fragColor1 = vec4(normalize(_normal.xyz), 0.0) * 0.5 + vec4(0.5, 0.5, 0.5, 0.5);
-    _fragColor2 = vec4(Color.rgb, 10 / 128.0);
-    */
     packDeferredFragmentLightmap(
          normalize(_normal),
-         Color.a * a,
-         Color.rgb,
```
jherico: Won't this break support for colored text?

samcake: NO because the color goes in the emissive still

samcake:
@zzmp No change with that, The fragment is discarded the same way as before when the alpha reaches <= 0.1
It will probably be different in the case this shader is used during the non deferred passes which is another story...
Just going for the quick fix here

bluetoe:
I tried this with text entity on pr domain and on current merged alpha domain with the PR.. looks ok. all text readable now where it was not before, although where I had noticed it most before was when changing graphics setting, which has been changed since then.. but all looks ok now, at least to me, cept the text now is kind of big and thick.. sort of, but yes much better now.

PR #7257, opened 2016-03-07
object outlining effect

toloudis:
Method for drawing a solid outline around an object, for use in selection highlighting.  

Config parameters are line thickness in pixels and a color.

Objects with inverted normals will not behave nicely - the effect will not show up. It tries to push the object extents out along the normal direction.  

Since there is no object selection yet, the code is currently just hard-coded to draw the "first" object with the effect, and the job is disabled for now. (Can flip the config class's ctor flag)


hifi-gustavo:

cozza13:
test this please

hifi-gustavo:
Build console output for commit: 60e9ce74a81f3b1eeed600efc7af6abe1a6c671f available at: https://gist.github.com/5f1caffa2b551a816a68<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7257.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7257.dmg

OmegaHeron:
I don't think we can QA Test this as it's not actually enabled to do anything?  Tested Windows 7 but I don't see any difference for a selected object with it vs production build, but, like I said - I'm not sure it's wired up...

"Since there is no object selection yet, the code is currently just hard-coded to draw the "first" object with the effect, and the job is disabled for now. (Can flip the config class's ctor flag)"

interface/src/Application.cpp
```diff
@@ -343,7 +343,7 @@ bool setupEssentials(int& argc, char** argv) {
     DependencyManager::set<AudioClient>();
     DependencyManager::set<AudioScope>();
     DependencyManager::set<DeferredLightingEffect>();
-    DependencyManager::set<TextureCache>();
+	DependencyManager::set<TextureCache>();
```
ZappoMan: Coding standard: Looks like your indent is set incorrectly. Make sure you have it set for 4 spaces.

toloudis:
Maybe this is premature as a PR, but it is a project I have been discussing with Sam.  It is disabled because there is not yet any mechanism to select the set of objects to which the effect is applied. Note: change HighlightingEffect.h line 29, flip the false to true to re-enable the effect.

Also, the files in gpu/ could technically be eliminated from the PR - they enable the use of glLineWidth in the gpu state objects.  It was part of an early implementation but the current form of the effect does not use it.

samcake:
@toloudis As long as the Job is disabled we are going to be able to merge. so this is good. I agree with you we will only be able to use it once we pipe some renderItems we actually want to highlight in.

I would recommend you remove the code that potentially change the gl line width, this is something that othere APIs wouldn't do so that's why we didn't exposed it either in the gpu lib

Cheers

libraries/render-utils/src/HighlightingEffect.cpp
```diff
@@ -0,0 +1,219 @@
...
+HighlightingEffect::HighlightingEffect() {
+    Parameters parameters;
+    _colorBuffer = gpu::BufferView(std::make_shared<gpu::Buffer>(sizeof(Parameters), (const gpu::Byte*) &parameters));
+    _colorBuffer.edit<Parameters>()._color = glm::vec4(1.0, 0.0, 1.0, 1.0);
+    _lineThickness = 4.0f;
+}
+
+void HighlightingEffect::init() {
+    // prep shader
+    const char Solid_frag[] = R"SCRIBE(#version 410 core
```
samcake: This shader source code should be a separate slf file that gets preprocessed during compilation with the correct targets and defines.

libraries/render-utils/src/HighlightingEffect.h
```diff
@@ -0,0 +1,75 @@
...
+#include <NumericalConstants.h>
+
+#include <gpu/Resource.h>
+#include <gpu/Pipeline.h>
+#include <render/DrawTask.h>
+
+
+class HighlightingEffectConfig : public render::Job::Config {
+    Q_OBJECT
+    Q_PROPERTY(float lineThickness MEMBER lineThickness WRITE setLineThickness)
```
samcake: I would suggest you expose a "enabled" properties too like other debug  jobs and create the Job disabled by default.
THis way we can merge with the Job integrated in the pipeline with no side effects.
At runtime you can enable/disable the job from the JS console by assigning the value to the Property
Render.RenderDeferredTask.Highlighting.enabled = true

samcake:
IT s not ready to be merged yet so  i m going to close for now.
Feel free to ping me directly to reopen

PR #7262, opened 2016-03-07
threadsafe Application::_viewFrustum

AndrewMeadows:
- ViewFrustum class moves out of octree and into shared lib
- Octree-specific code in ViewFrustum now lives in OctreeUtils
- now make copies of ViewFrustum rather than hand out raw pointers to foreign code/threads
- Application::_viewFrustum and friends are protected by lock on mutex
- RenderArgs uses std::stack<ViewFrustum> so views can be pushed/popped
- cleanup and cruft removal


AndrewMeadows:
https://app.asana.com/0/26623005675174/42428945742846

The thread issues indicated in that asana task looked legit.  The solution proved to be bigger than I expected.  

(1) Since the **ViewFrustum** pointers were passed far and wide many files had to be changed.  It seemed to me that rather than sprinkling mutex locks all over the code it would be better to make copies of ViewFrustums for most of that code and then only lock the where the source instances are changed or copied.

(2) In order to pass copies the ViewFrustum implementation had to be moved to the shared lib.  There was some octree-specific code therein which had to remain in the octree library.  **OctreeUtils** was created to house the octree-specific code.

(3) Some cruft fell out of all this work.

hifi-gustavo:
Build console output for commit: 7700c54645a9a560471e3d67687d588858771b12 available at: https://gist.github.com/b1d3842e684296995df6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7262.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7262.dmg

assignment-client/src/octree/OctreeQueryNode.cpp
```diff
@@ -141,6 +141,16 @@ void OctreeQueryNode::writeToPacket(const unsigned char* buffer, unsigned int by
     }
 }
 
+void OctreeQueryNode::copyCurrentViewFrustum(ViewFrustum& viewOut) const {
```
ZappoMan: hmmmm - I'm pretty sure `OctreeQueryNode` only ever gets accessed on a single thread. Did you double check that this change was needed?

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -296,7 +297,7 @@ OctreeElement::AppendState EntityTreeElement::appendElementData(OctreePacketData
                         entityTreeElementExtraEncodeData->entities.contains(entity->getEntityItemID());
                 }
 
-                if (includeThisEntity && params.viewFrustum) {
+                if (includeThisEntity) {
```
ZappoMan: it used to be possible to encode a tree without include a view frustum... this is for things like copy/paste and for save binary files... so... we need to make sure we either assert() or guard against this case. AFAIK we didn't remove the save binary support yet... so this code could be called with a null view.

ZappoMan: maybe you meant to check the newly added "recurse everything".

assignment-client/src/octree/OctreeQueryNode.cpp
AndrewMeadows: No, I didn't verify 100%.  I did verify that `OctreeQueryNode` derives from `NodeData`, and that its `updateLastKnownViewFrustum()` method is called inside `OctreeSendThread::packetDistributor()` which sounded "threaded" to me.

libraries/entities/src/EntityTreeElement.cpp
AndrewMeadows: Yes, it looks like I forgot to check `params.resurseEverything`.

hifi-gustavo:
Build console output for commit: ca58b1b1edc413f1c149dd24a2fc43ebf2d94fa7 available at: https://gist.github.com/b1d3842e684296995df6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7262.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7262.dmg

libraries/octree/src/OctreeHeadlessViewer.h
```diff
@@ -66,7 +66,6 @@ public slots:
     unsigned getOctreeElementsCount() const { return _tree->getOctreeElementsCount(); }
 
 private:
-    ViewFrustum _viewFrustum;
```
AndrewMeadows: `OctreeHeadlessViewer` used to have a `_viewFrustum` data member, but now its base class `OctreeRenderer` (which used to only keep a pointer to one) has a valid `ViewFrustum`, so this one goes away.

libraries/octree/src/Octree.h
```diff
@@ -61,22 +61,22 @@ const int LOW_RES_MOVING_ADJUST  = 1;
...
+    quint64 lastViewFrustumSent;
     int maxEncodeLevel;
     int maxLevelReached;
-    const ViewFrustum* viewFrustum;
     bool includeExistsBits;
     int chopLevels;
-    bool deltaViewFrustum;
-    const ViewFrustum* lastViewFrustum;
+    bool deltaView;
+    bool recurseEverything { false };
```
AndrewMeadows: The `recurseEverything` data member was added to replace the functionality of a NULL `ViewFrustum` pointer used to serve: it was an indication that the view was not important but that we wanted to copy everything (e.g. when saving everything to disk).

libraries/shared/src/RenderArgs.h
```diff
@@ -95,11 +97,22 @@ class RenderArgs {
...
+    const ViewFrustum& getViewFrustum() const { assert(_viewFrustums.size() > 0); return _viewFrustums.top(); }
+    void pushViewFrustum(const ViewFrustum& viewFrustum) { _viewFrustums.push(viewFrustum); }
+    void popViewFrustum() { _viewFrustums.pop(); }
+
     std::shared_ptr<gpu::Context> _context = nullptr;
     std::shared_ptr<gpu::Framebuffer> _blitFramebuffer = nullptr;
     std::shared_ptr<render::ShapePipeline> _pipeline = nullptr;
     OctreeRenderer* _renderer = nullptr;
-    ViewFrustum* _viewFrustum = nullptr;
+    std::stack<ViewFrustum> _viewFrustums;
```
AndrewMeadows: `RenderArgs` now has a `std::stack<ViewFrustum>` as per a request by @samcake.  There are a few places where the current `ViewFrustum` is replaced by a temporary one, and then the old is restored.  To find where this is done look for calls to `RenderArgs::pushViewFrustum()` and `::popViewFrustum()`.

libraries/octree/src/OctreeUtils.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+float boundaryDistanceForRenderLevel(unsigned int renderLevel, float voxelSizeScale) {
+    return voxelSizeScale / powf(2.0f, renderLevel);
+}
+
+float getAccuracyAngle(float octreeSizeScale, int boundaryLevelAdjust) {
+    const float maxScale = (float)TREE_SCALE;
+    float visibleDistanceAtMaxScale = boundaryDistanceForRenderLevel(boundaryLevelAdjust, octreeSizeScale) / OCTREE_TO_MESH_RATIO;
+    return atan(maxScale / visibleDistanceAtMaxScale);
+}
```
AndrewMeadows: This is the octree-specific stuff that used to live inside `ViewFrustum.cpp` before it was moved to the `shared` lib.

libraries/shared/src/ViewFrustum.cpp
```diff
@@ -268,55 +265,6 @@ bool testMatches(float lhs, float rhs, float epsilon = EPSILON) {
...
-        qCDebug(octree, "%s -- compareTo._farClip=%f _farClip=%f",
-                (testMatches(compareTo._farClip, _farClip) ? "MATCHES " : "NO MATCH"),
-                (double)compareTo._farClip, (double)_farClip);
-        qCDebug(octree, "%s -- compareTo._focalLength=%f _focalLength=%f",
-                (testMatches(compareTo._focalLength, _focalLength) ? "MATCHES " : "NO MATCH"),
-                (double)compareTo._focalLength, (double)_focalLength);
-    }
-    return result;
-}
-
```
AndrewMeadows: ViewFrustum::matches() was cruft.

libraries/shared/src/ViewFrustum.cpp
```diff
@@ -662,51 +450,10 @@ void ViewFrustum::evalViewTransform(Transform& view) const {
...
-float boundaryDistanceForRenderLevel(unsigned int renderLevel, float voxelSizeScale) {
-    return voxelSizeScale / powf(2, renderLevel);
-}
-
-float ViewFrustum::getAccuracyAngle(float octreeSizeScale, int boundaryLevelAdjust) const {
-    const float maxScale = (float)TREE_SCALE;
-    float visibleDistanceAtMaxScale = boundaryDistanceForRenderLevel(boundaryLevelAdjust, octreeSizeScale) / OCTREE_TO_MESH_RATIO;
-    return atan(maxScale / visibleDistanceAtMaxScale);
+    _centerSphereRadius = -1.0e6f; // -10^6 should be negative enough
 }
```
AndrewMeadows: `ViewFrustum::invalidate()` is used to produce a `ViewFrustum` that will return false for any reasonable intersection test.  This is used for the rare case where the "lastKnowViewFrustum" was never set but we think we need to use "delta mode" when sending the data in `Octree::encodeTreeBitstreamRecursion()`.

libraries/octree/src/OctreeRenderer.h
```diff
@@ -51,8 +51,8 @@ class OctreeRenderer : public QObject {
     /// render the content of the octree
     virtual void render(RenderArgs* renderArgs);
 
-    ViewFrustum* getViewFrustum() const { return _viewFrustum; }
-    void setViewFrustum(ViewFrustum* viewFrustum) { _viewFrustum = viewFrustum; }
+    const ViewFrustum& getViewFrustum() const { return _viewFrustum; }
+    void setViewFrustum(const ViewFrustum& viewFrustum) { _viewFrustum = viewFrustum; }
```
AndrewMeadows: Note that the ViewFrustums inside OctreeRenderer are NOT guarded by a mutex when they are changed or accessed.

libraries/shared/src/ViewFrustum.h
```diff
@@ -131,52 +126,40 @@ class ViewFrustum {
...
-    glm::vec4 _corners[8];
-    glm::vec3 _cornersWorld[8];
-    ::Plane _planes[6]; // plane normals point inside frustum
+    float _nearClip { DEFAULT_NEAR_CLIP };
+    float _farClip { DEFAULT_FAR_CLIP };
 
     const char* debugPlaneName (int plane) const;
-
-    // Used to project points
-    glm::mat4 _ourModelViewProjectionMatrix;
```
AndrewMeadows: Reordering the data members and deleting some unused variables changed the size of this class from 744 to 616 bytes.

assignment-client/src/octree/OctreeQueryNode.h
```diff
@@ -114,6 +114,8 @@ class OctreeQueryNode : public OctreeQuery {
 
     int _maxSearchLevel { 1 };
     int _maxLevelReachedInLastSearch { 1 };
+
+    mutable QMutex _viewMutex { QMutex::Recursive };
```
AndrewMeadows: `OctreeQueryNode` and `Application` each get a new **mutex** that is used to lock the thread when their`ViewFrustum`'s are changed or queried.

interface/src/avatar/Head.cpp
```diff
@@ -316,9 +316,6 @@ void Head::relaxLean(float deltaTime) {
     _deltaLeanForward *= relaxationFactor;
 }
 
-void Head::render(RenderArgs* renderArgs, float alpha, ViewFrustum* renderFrustum) {
-}
```
AndrewMeadows: Head::render() was cruft.

AndrewMeadows:
This PR has a lot of changes relative to the problem it is trying to fix.  If anyone has an idea about how to get the job done in a less intrusive way then please make some suggestions.  Also, we have to decide if all these changes are worth the risk that new errors will be introduced.  The problem of non-thread-safe ViewFrustums is probably not a crasher type of problem but just a "sometimes the data used to generate the next data update is not quite right, but will probably be right the next time (in a few milliseconds)" kind of problem -- we might not ever notice that it isn't thread-safe.

hifi-gustavo:
Build console output for commit: 7a426ea1d94aaf9863e1a77f06710b718e1c1d4c available at: https://gist.github.com/b1d3842e684296995df6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7262.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7262.dmg

AndrewMeadows:
test this please

hifi-gustavo:
Build console output for commit: e7e34e10c350ef12ef841eaae2b58914aeec2fbd available at: https://gist.github.com/b1d3842e684296995df6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7262.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7262.dmg

libraries/shared/src/ViewFrustum.cpp
ZappoMan: just because it's not currently called - doesn't mean it's cruft... someday someone will want to ask the question... does this ViewFrustum match tat ViewFrustum... and they're going to write some new code that might be buggy... this code at least works.

libraries/shared/src/ViewFrustum.cpp
```diff
@@ -427,177 +370,17 @@ void ViewFrustum::computeOffAxisFrustum(float& left, float& right, float& bottom
...
-    qCDebug(octree, "_right=%f,%f,%f", (double)_right.x, (double)_right.y, (double)_right.z );
-    qCDebug(octree, "_fieldOfView=%f", (double)_fieldOfView);
-    qCDebug(octree, "_aspectRatio=%f", (double)_aspectRatio);
-    qCDebug(octree, "_centerSphereRadius=%f", (double)_centerSphereRadius);
-    qCDebug(octree, "_nearClip=%f", (double)_nearClip);
-    qCDebug(octree, "_farClip=%f", (double)_farClip);
-    qCDebug(octree, "_focalLength=%f", (double)_focalLength);
-}
-
-glm::vec2 ViewFrustum::projectPoint(glm::vec3 point, bool& pointInView) const {
```
ZappoMan: same issue here... not sure we need to be deleting all this stuff..

ZappoMan:
Andrew - as you point out, this fixes and issue that may not be that big of an issue. Let's not merge this at this point, it's too risky at this point. But we can keep you branch and consider re-opening this PR in the future.

libraries/shared/src/ViewFrustum.cpp
AndrewMeadows: I will recover these methods.

PR #7278, opened 2016-03-08
adjust simulation priorities

AndrewMeadows:
- regular entity property changes happen at simulation priority 127 (used to be 128)
- property changes by Actions happen at simulation priority 128

### DIFF
--- a/libraries/entities/src/EntityItem.h
+++ b/libraries/entities/src/EntityItem.h
@@ -368,7 +368,8 @@ public:
 
     void getAllTerseUpdateProperties(EntityItemProperties& properties) const;
 
-    void flagForOwnership() { _dirtyFlags |= Simulation::DIRTY_SIMULATOR_OWNERSHIP; }
+    void pokeSimulationOwnership() { _dirtyFlags |= Simulation::DIRTY_SIMULATION_OWNERSHIP_FOR_POKE; }
+    void grabSimulationOwnership() { _dirtyFlags |= Simulation::DIRTY_SIMULATION_OWNERSHIP_FOR_GRAB; }
     void flagForMotionStateChange() { _dirtyFlags |= Simulation::DIRTY_MOTION_TYPE; }
 
     bool addAction(EntitySimulation* simulation, EntityActionPointer action);
--- a/libraries/physics/src/EntityMotionState.cpp
+++ b/libraries/physics/src/EntityMotionState.cpp
@@ -117,23 +117,22 @@ bool EntityMotionState::handleEasyChanges(uint32_t& flags) {
         } else if (_entity->getSimulatorID() == Physics::getSessionUUID()) {
             // we just inherited ownership, make sure our desired priority matches what we have
             upgradeOutgoingPriority(_entity->getSimulationPriority());
+        } else {
+            _outgoingPriority = 0;
+            _nextOwnershipBid = usecTimestampNow() + USECS_BETWEEN_OWNERSHIP_BIDS;
         }
     }
     if (flags & Simulation::DIRTY_SIMULATION_OWNERSHIP_PRIORITY) {
         // The DIRTY_SIMULATOR_OWNERSHIP_PRIORITY bits really mean "we should bid for ownership because
         // a local script has been changing physics properties, or we should adjust our own ownership priority".
-        if (_outgoingPriority <= VOLUNTEER_SIMULATION_PRIORITY) {
-            // we were previously uninterested in owning this object
-            // which means this is our first frame of interest
-            // therefore we should bid immediately
-            _nextOwnershipBid = 0;
-        }
         // The desired priority is determined by which bits were set.
         if (flags & Simulation::DIRTY_SIMULATION_OWNERSHIP_FOR_GRAB) {
             _outgoingPriority = SCRIPT_GRAB_SIMULATION_PRIORITY;
         } else {
             _outgoingPriority = SCRIPT_POKE_SIMULATION_PRIORITY;
         }
+        // reset bid expiry so that we bid ASAP
+        _nextOwnershipBid = 0;
     }
     if ((flags & Simulation::DIRTY_PHYSICS_ACTIVATION) && !_body->isActive()) {
         _body->activate();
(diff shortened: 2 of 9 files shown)


libraries/physics/src/EntityMotionState.cpp
```diff
@@ -298,7 +302,7 @@ bool EntityMotionState::remoteSimulationOutOfSync(uint32_t simulationStep) {
     }
 
     if (_entity->actionDataNeedsTransmit()) {
-        setOutgoingPriority(SCRIPT_EDIT_SIMULATION_PRIORITY);
+        _outgoingPriority = _entity->hasActions() ? SCRIPT_GRAB_SIMULATION_PRIORITY : SCRIPT_POKE_SIMULATION_PRIORITY;
```
sethalves: I think there are various contants in the PR which have "grab" in then when it maybe should be "action".  Not all actions are grabs.

hifi-gustavo:
Build console output for commit: eaafc1693533d1a89776d3567da2725ddb624dfd available at: https://gist.github.com/5c0d26bb1aadd23ae8b6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7278.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7278.dmg

ZappoMan:
something about the description of this PR feels wrong... it sounds like you're saying that programmatic edits will be lower priority than action based edits... that sounds like it will be the source of many hidden bugs. Are we sure we're approaching this problem from the right angle?

AndrewMeadows:
https://app.asana.com/0/26225263936266/98458250599577

We noticed a problem during the internal hackathon demos:

Caitlyn hit the golf ball out of the field and it started to fall. Seth caught it and put it back on the field. Caitlyn couldn't hit the ball and take ownership of it while Seth still owned it. Since Seth wasn't getting timely avatar updates of Caitlyn's motion the simulation of the golf club would teleport through and the ball would simulate incorrectly.

The problem was:

Seth simulates the ball at priority 128 when he grabbs it. When he dropps it onto the field it was still at 128. Caitlyn simulates the golf club at 128, but when she hits the ball she bids at 128-1 = 127, which loses against Seth's priority of 128.

The proposed solution is as follows:

(1) Adjust the simulation priority rules such that local actions used priority = 128, but all other scripted property changes use 127.

(2) When deleting a local action the priority should drop down to 127.

Note: there is a minor detail of the simulation priority system: when an object collides with the local MyAvatar body the interface will bid for simulation of that object at 127.

So, with the above changes, when UserA holds a sword and whacks UserB with it UserA will retain ownership at priority 128, however when UserA throws the sword at UserB, the priority should drop to 127 when UserA removes its Action and when it hits... UserB will bid for ownership at 127 and should win ownership of it.

libraries/physics/src/EntityMotionState.cpp
AndrewMeadows: Seth, I need two words to distinguish the difference between "a script has changed some random physics property" and "an action is actively updating the physics properties".  The two words I have choses are: "poke" and "grab".  I'm open to suggestions for better terms.

AndrewMeadows:
First test:

(1) Login to a domain with UserA and UserB.
(2) Enable the debug option:  **Developer --> Physics --> Highlight Simulation Ownership**
(3) Create a dynamic Object with gravity that can be picked up and dropped.
(4) UserA: use handControllerGrab.js to pick up Object
--> For both this PR and master branch: UserA should own the simulation (UserA should blue icon with silhouette of head)
(5) UserA: put the object over UserB's head and drop it so that it falls for about one second before it hits UserB.
--> Master branch: UserA should retain ownership of the Object as it hits UserB and falls to the ground
--> This PR: UserB should gain ownership of the Object after it hits.

Second test:

(1) Create three dynamic objects.  Object2 should be a sphere that will roll and not settle too quickly, but zero-gravity objects with zero or very low linear damping should work as well.
(2) UserA: grab Object1 and hit Object2
--> UserA should take ownership of Object1 and Object2
(3) UserB: grab Objec3 and before Object2 settles to a stop, hit Object2 with Object3
--> Master branch: UserA should retain ownership of Object2 after UserB hits it.
--> This PR: UserB should take ownership of Object2 after the hit.

libraries/entities/src/EntityItem.h
```diff
@@ -368,7 +368,8 @@ class EntityItem : public SpatiallyNestable, public ReadWriteLockable {
 
     void getAllTerseUpdateProperties(EntityItemProperties& properties) const;
 
-    void flagForOwnership() { _dirtyFlags |= Simulation::DIRTY_SIMULATOR_OWNERSHIP; }
+    void pokeSimulationOwnership() { _dirtyFlags |= Simulation::DIRTY_SIMULATION_OWNERSHIP_FOR_POKE; }
+    void grabSimulationOwnership() { _dirtyFlags |= Simulation::DIRTY_SIMULATION_OWNERSHIP_FOR_GRAB; }
```
AndrewMeadows: To "poke" an entity means to change any property via script.

To "grab" an entity means to use an action to update its physical properties.

AndrewMeadows:
test this please

hifi-gustavo:
Build console output for commit: 9157ff6eda7d00bc38208f483767c4f19f943a8d available at: https://gist.github.com/5c0d26bb1aadd23ae8b6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7278.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7278.dmg

AndrewMeadows:
Ok, this is ready to go again.  There was a bug where bad priority values could be obtained (e.g. 255) by decrementing a zero priority during a collision, now fixed.

hifi-gustavo:
Build console output for commit: 4ae94c06117a6ac3b87a2ddb98a1021f7bb992a2 available at: https://gist.github.com/5c0d26bb1aadd23ae8b6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7278.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7278.dmg

PhilipRosedale:
Andrew and I tested this thoroughly with 3 machines to verify that far/near grabs working correctly, ownership handoffs for collisions work as desired, everything visually consistent for multiple observers. 

Also reviewed all code changes with Andrew and looks good.

PR #7279, opened 2016-03-08
Bypass baffling OSX GL error

jherico:
Testing showed that the code was failing on setting the modelview matrix uniform, even though it had a valid uniform location ID and even though the projection uniform set immediately before it worked fine.  Investigation showed now immediate reason why this would fail, but since our shader doesn't actually need the projection and modelview as separate matrices, it's simple enough to work around by compressing them into a single matrix.  

Needs testing on HMD devices to ensure the combined matrix is correctly compositing the modelview with the projection.

### DIFF
--- a/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
+++ b/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
@@ -293,14 +293,12 @@ void OpenGLDisplayPlugin::customizeContext() {
     auto uniforms = _program->ActiveUniforms();
     while (!uniforms.Empty()) {
         auto uniform = uniforms.Front();
-        if (uniform.Name() == "Projection") {
-            _projectionUniform = uniform.Index();
-        } else if (uniform.Name() == "ModelView") {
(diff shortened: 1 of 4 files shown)


ZappoMan:
Make sure to test this on Windows, with the HMD.

hifi-gustavo:
Build console output for commit: 7d046c30a9cc95991d013af41612756d0431265a available at: https://gist.github.com/116adfb99b37424b28cf<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7279.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7279.dmg

jherico:
Just tested in the Oculus HMD.  The composition of the projection and modelview seems to work fine.

Atlante45:
Doesn't crash on mac anymore.

On windows:
- Depth Reticle is now working again. - Pass
- Overlay is stready - Pass
- Loading/quick turn behavior seems the same as production - Pass
- Mouse dragging works - Pass

PR #7290, opened 2016-03-09
Add HF username as default username for bugsplat

huffman:


hifi-gustavo:
Build console output for commit: b8df8f170d568df88d301006f8a0963b98c32bda available at: https://gist.github.com/2b3e4ce5d85e5bcbd25f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7290.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7290.dmg

hifi-gustavo:
Build console output for commit: 806bae99aae2d32a51f227f9c0f1e47b8fbcbd3f available at: https://gist.github.com/2b3e4ce5d85e5bcbd25f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7290.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7290.dmg

huffman:
For task: https://app.asana.com/0/40869567426042/99080747796181

interface/src/main.cpp
```diff
@@ -143,6 +143,14 @@ int main(int argc, const char* argv[]) {
 
         QObject::connect(&server, &QLocalServer::newConnection, &app, &Application::handleLocalServerConnection);
 
+#ifdef HAS_BUGSPLAT
+        AccountManager& accountManager = AccountManager::getInstance();
+        mpSender.setDefaultUserName(accountManager.getAccountInfo().getUsername().toLatin1().constData());
```
ZappoMan: `qPrintable(accountManager.getAccountInfo().getUsername())` seems cleaner to me

huffman: Semantically, my understanding is that `qPrintable` is used to create strings that can be used with Qt's printf style printing functions. That seems out of place here, and `.toLatin1().constData()` seems clearer as to what is happening. 

Can you clarify why you think it is cleaner?

OmegaHeron:
Tested Windows 7 - got a crash on close and sent report.  Unfortunately I had already filled in my user name as Omega Heron (notice space) - if this replaced with my HF user ID it'll be omegaheron.

interface/src/main.cpp
ZappoMan: It's not a big deal.. but it's a matter of trying to keep the code base consistent...

`const char * qPrintable(const QString & str) - Returns str as a const char *. This is equivalent to str.toLocal8Bit().constData().`

qPrintable() returns a const char* to the converted 8bit representation of the character... It's not clear from the bugSplat documentation what the expect the character encoding to be for calls to `void MiniDmpSender::setDefaultUserName(LPCSTR szName)` - so they probably don't do anything special with it anyway... My point is just that there are 168 instances of `qPrintable()` and only one instance of `toLatin1().constData()` -- and 12 instances of `toLocal8Bit().constData()` -- we've generally been trying to migrate those callers to use `qPrintable()` -- since they are equivolent.

If bugSplat explicitly says they expect all of their strings in Latin1 then-- sure let's leave it... if not, let's use the more compact form.

Presumably BugSplat copies that string into their own buffer, otherwise all hell will break loose.

cozza13:
I tested this. I removed my name out of the "name" field. I then crashed the client and it left the name field blank.

jherico:
@cozza13 were you looking at the name field in the dialog, or in the corresponding crash report on bugsplat?  According to their support rep, setting the default value sets it for the report, but does not expose it to the user in the dialog.  They can override it by filling out something, but if they don't we see the default value we set.

interface/src/main.cpp
huffman: Ok, I will switch it to `qPrintable`

hifi-gustavo:
Build console output for commit: 4bcd8323f9f3c1396ba6ecc2607f409b78f29dde available at: https://gist.github.com/2b3e4ce5d85e5bcbd25f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7290.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7290.dmg

hifi-gustavo:
Build console output for commit: 55817e6b5af7a0a3f29ff93db2a759133fea7be3 available at: https://gist.github.com/2b3e4ce5d85e5bcbd25f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7290.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7290.dmg

PR #7292, opened 2016-03-09
Mark Resource loaded after postprocessing

zzmp:
This avoids duplicate messages that a Resource is loaded (which occured in Textures), as well as avoiding marking that resource loaded until after it is both downloaded and processed.

The existing Resource base class already had messages for requests starting and finishing, but also called `finishedLoading(true)` if the download worked. Instead, it will now `emit loaded(_data)` with the download, but not call `finishedLoading`. That is expected to be called by the base classes, as part of their implementation of the override of `downloadFinished`.

This was included in the doxygen comments, as well as done in the base class's implementation.


zzmp:
My largest worry with this PR is that future devs will not know to call `finishedLoading` from `downloadFinished`.

zzmp:
#### Intents

Textures are being marked as loaded as soon as they download, when they should not be marked as such until the downloaded data is read into a texture. This occurs because the resource base `Resource` class marks it as loaded as soon as it downloads, and then calls the virtual method `downloadFinished` for further processing (reading into a texture):
https://github.com/highfidelity/hifi/blob/a2eb819bc2a2ad384482b64116602d01e1f68eef/libraries/networking/src/ResourceCache.cpp#L374

Once the data is read into a texture, the `NetworkTexture` also marks it as loaded:
https://github.com/highfidelity/hifi/blob/a2eb819bc2a2ad384482b64116602d01e1f68eef/libraries/model-networking/src/model-networking/TextureCache.cpp#L348

According to the existing comments, the call to mark the `Resource` as loaded should only be done by the base classes, probably in `downloadFinished`: https://github.com/highfidelity/hifi/blob/a2eb819bc2a2ad384482b64116602d01e1f68eef/libraries/networking/src/ResourceCache.h#L201

This PR is to remove that duplicity, as well as make it clear when a `Resource` actually loads.

hifi-gustavo:
Build console output for commit: 54af58834a9038c517c7acf749816d7efbe29fcd available at: https://gist.github.com/698b74ace30bdb001a70<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7292.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7292.dmg

OmegaHeron:
@zzmp This needs a sync to PR7293 so it doesn't crash on start 100%.

zzmp:
Thanks for the heads up @OmegaHeron. Rebuilding now.

hifi-gustavo:
Build console output for commit: d06b99967e3bb59eacf94f97e49ab5d8ddffeb92 available at: https://gist.github.com/698b74ace30bdb001a70<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7292.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7292.dmg

OmegaHeron:
Tested Windows 7 - Did a general comparison test to current production release, visited sandbox, hq, alpha and my own domain. I see no difference between this PR and current release. QA Pass unless there's a more specific test to be done.

zzmp:
@OmegaHeron did you check the logs for an uncached texture loading, and make sure that the three messages mentioned were not being logged simultaneously? That's the best test I could think of for this.

OmegaHeron:
Will double check that after dinner and report back.

OmegaHeron:
Reviewed again and I do not see groups of those messages - repeating this with production build I see (within range of norm) groups of 3, with this I do not indicating it's no longer doing open, fail, stash as with current release code.  I can't find any cases of breakage with sounds, scripts or other here.  Seems reasonably safe to pass on to see if it plays well with others.

PR #7303, opened 2016-03-10
Update the watchdog heartbeat manually during startup

jherico:
Some users are reporting startup times near to the watchdog timeout value.  This change adds several checkpoints in the startup process during which we manually update the heartbeat to avoid the timeout.  Hopefully no one operation (`loadScripts` and `initializeGL`, I'm looking at you) will take more than 10 seconds. 

Testing:  should load and run on systems that have a startup time of more than 10 seconds.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -250,14 +250,18 @@ public:
         setObjectName("Deadlock Watchdog");
         QTimer* heartbeatTimer = new QTimer();
         // Give the heartbeat an initial value
-        _heartbeat = usecTimestampNow();
+        updateHeartbeat();
         connect(heartbeatTimer, &QTimer::timeout, [this] {
-            _heartbeat = usecTimestampNow();
+            updateHeartbeat();
         });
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 9645ed9120314339eee7ade4770ac0e3f5c89514 available at: https://gist.github.com/5c89b746d5bc9cb3b571<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7303.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7303.dmg

ctrlaltdavid:
Works well. My start-up times are typically hover around the 10s mark, as reported in the log file, so previously I was getting false positives where the watchdog timer often 

timed me out even though I was still finishing starting up.
Started up Interface 10 times, some with no other PC activity and others with PC also doing other things. 4 start-up times < 10s and 6 with start-up times between 10 and 14 seconds. All started up OK.
Started up Interface twice with CPU at 100% compiling Interface and both start-ups timed out.

PR #7305, opened 2016-03-10
Adjust edit handle sizes

PhilipRosedale:
When you move a selected object far away, the size and rotate handles will not get bigger than the object (which doesn't work anyway since they overlap).  It is easier therefore to grab for translation a distance object, and it looks better.  

Also reduced the size of the rotate handles somewhat. 

Test plan:  

Create each type of entity within edit.js, test that handles feel like an OK size, and try grabbing handles at different distances from the object.  Should work better than existing.

### DIFF
--- a/examples/libraries/entitySelectionTool.js
+++ b/examples/libraries/entitySelectionTool.js
@@ -894,7 +894,7 @@ SelectionDisplay = (function() {
         alpha: handleAlpha,
         visible: false,
         size: 0.1,
-        scale: 0.025,
+        scale: 0.1,
         isFacingAvatar: false,
         drawInFront: true,
     });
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: c11487735fcafe17c6493c82e4eff6d8accf58cb available at: https://gist.github.com/271f4ad7f677329db628<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7305.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7305.dmg

bluetoe:
I notice that the changes are in examples/libraries/entitySelectionTool.js. Does this require a different version of edit.js? Or is it included in the PR build? I am assuming it is. Quick answer to this PR is yes.

I've tried this with box, sphere, model, text and web entities.  This does seem better than current. The grab points are smaller and easier to deal with. And same with rotation arrows. I am testing this in desktop with a mouse so I can not say how it will be with hand controllers.

With Zone, Light, Particle and Voxel entities it requires double click to focus camera and set independent view. But works the same with them.

But yes. This seems better to me than before.

PR #7308, opened 2016-03-10
Fix link problem with shaders that use transparency and normal maps

jherico:


jherico:
https://app.asana.com/0/26225263936266/99314179876980

hifi-gustavo:
Build console output for commit: aad2aaa7644f32ae412e49d42a335803b05b6d53 available at: https://gist.github.com/c71ce3ebb98c706ae040<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7308.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7308.dmg

samcake:
Sounds good to me, i m surprised we separated color vs alphagood to go on the dev side

OmegaHeron:
Tested Windows 7 - No problems seen.  I think someone who actually sees issue this addresses is testing as well.

Nex-Pro:
Tested on win 10 64bit.  Made a test model in blender with alpha on the normal map and a texture with alpha. Works as expected.  But threw up an unexpected issue.   The model was a flat plane with two maps. Diffuse with alpha and a normal map ( normal map also has alpha. 
https://dl.dropboxusercontent.com/u/96759331/normal_with_trans_test.fbx
The Ycoords for all the verts ( Z axis in blender ) are 0.0 However scaling the mesh vertically which I would have thought would have no effect resulted in the flat plane becoming inclined and thus stretched.

PR #7221, opened 2016-02-27
Migrate composition of the overlay and pointer to the present thread

jherico:

PR #7312, opened 2016-03-10
Better threaded handling of QML overlays

jherico:
QML Overlays have a few issues.  They will call `deleteLater()` multiple times for a single object, and the code to detect if a QML element has been deleted doesn't function properly because a weak pointer isn't initialized, and because an invokation of a QML function is not inside the protected section.

### DIFF
--- a/interface/src/ui/overlays/QmlOverlay.cpp
+++ b/interface/src/ui/overlays/QmlOverlay.cpp
@@ -51,27 +51,24 @@ void QmlOverlay::buildQmlElement(const QUrl& url) {
 }
 
 QmlOverlay::~QmlOverlay() {
-    if (_qmlElement) {
-        _qmlElement->deleteLater();
-        _qmlElement = nullptr;
-    }
+    _qmlElement.reset();
 }
 
 void QmlOverlay::setProperties(const QVariantMap& properties) {
     Overlay2D::setProperties(properties);
     auto bounds = _bounds;
-    std::weak_ptr<QQuickItem> weakQmlElement;
-    DependencyManager::get<OffscreenUi>()->executeOnUiThread([=] {
+    std::weak_ptr<QQuickItem> weakQmlElement = _qmlElement;
+    DependencyManager::get<OffscreenUi>()->executeOnUiThread([weakQmlElement, bounds, properties] {
         // check to see if qmlElement still exists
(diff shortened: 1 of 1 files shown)


jherico:
I'm investigating https://app.asana.com/0/27650181942747/95580481746093 but it's a little hard to say if this is actually the same bug or just another bug I encountered while working on finding that one.

We appear to have at least one bug that causes memory corruption, probably via multiple threads concurrently accessing some container.  In these cases the stack traces are nearly useless, because they often are some random function returned to called by smashing the stack pointer or just causing some other crash when the corrupted memory is used again.  

So right now I'm just trying to stress out my system by running a debug build under various conditions and trying to get crashes.  Since some of the crashes are deep inside the QML DLL and seem to be related to overlays, I was looking into this class and noticed that it was a) always calling `deleteLater()` twice for a given object and b) buggy in it's use of a weak pointer to try to serialize property setting and ensuring that the object was not deleted before the properties were set.

Soooo... I think this is the QtQML DLL stack trace crashes, but it's hard to verify.

hyperlogic:
@jherico I saw the crash mentioned in the asana task, before your overlay refactor PR #7221, but not after.  It was due to qApp->sendEvent() being called via the script thread.  It was clear that this was happening because Qt was asserting in debug builds with a clear error message.

hyperlogic:
test this please

hyperlogic:
These changes look good to me.

hifi-gustavo:
Build console output for commit: e452020126b9759cda97f001fd51f94467799146 available at: https://gist.github.com/a3844b66fa24e3bc36a0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7312.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7312.dmg

ZappoMan:
Smoke test of this on Mac seemed borked... not sure if it's an issue with master... or something with this PR... but basically half the content doesn't seem to render... some models do, some don't... Hell and Playa are basically empty.

OmegaHeron:
Tested Windows 7 - Checked a few places, made sure my scripts ran and I could do my general routines, all good.  QA Pass for the basics.

PR #7314, opened 2016-03-10
Fix a potentially bad access to RenderItem case from Overlay

samcake:
In overlays.cpp, when ever the properties are edited, the Overlay's RenderITem is updated. But this code is trying to access the RenderITem from the scene and there is no guarantee that it has been allocated yet in the scene.
We replaced the logic by adding  some lazy update in the Base3DOverlay::setProperties. THis way no need to go get the Item from the scene. we should actually discourage this use case as this would incur threading issues.
Added a thread safe counter of the number of allocated renderItems which is tracking the true number of allocated renderItems in the scene. It gets updated when the pendingChanges gets processed.


hifi-gustavo:
Build console output for commit: 3944a7f297a822537fa26277bd808a38ac103a06 available at: https://gist.github.com/8b6dca9069f7f14dc1cd<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7314.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7314.dmg

libraries/render/src/render/Scene.h
```diff
@@ -60,19 +60,25 @@ class Scene {
...
     // Process the penging changes equeued
     void processPendingChangesQueue();
 
+    // This next call are  NOT threadsafe, you have to call them from the correct thread to avoid any potential issues
+
     // Access a particular item form its ID
     // WARNING, There is No check on the validity of the ID, so this could return a bad Item
     const Item& getItem(const ItemID& id) const { return _items[id]; }
 
-    size_t getNumItems() const { return _items.size(); }
```
howard-stearns: +1 dead code removal

libraries/render/src/render/Scene.h
```diff
@@ -60,19 +60,25 @@ class Scene {
...
-
-    // Access the spatialized items
+        // Access the spatialized items
```
howard-stearns: stupid msvc indenting... (coding standard whitespace)

libraries/render/src/render/Scene.h
howard-stearns: Oh, wait. It's not removed. just reimplemented above. But is it actually used? Can we kill it?

howard-stearns:
Testing notes: Confirm that this fixes the problem when running Debug code from debugger, and doesn't break anything in Release code.

Confirmed for Windows.
On Mac, I've confirmed that all is well on Release builds. Debug builds get access violations in random places, but they did before this PR as well.

hifi-gustavo:
Build console output for commit: a88f22fa3c33795a3e193fb4294c9ef932f67f41 available at: https://gist.github.com/8b6dca9069f7f14dc1cd<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7314.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7314.dmg

interface/src/ui/overlays/Base3DOverlay.cpp
```diff
@@ -100,6 +106,17 @@ void Base3DOverlay::setProperties(const QVariantMap& properties) {
     if (properties["ignoreRayIntersection"].isValid()) {
         setIgnoreRayIntersection(properties["ignoreRayIntersection"].toBool());
     }
+
+    // Communicate changes to the renderItem if needed
+    if (needRenderItemUpdate) {
+        auto itemID = getRenderItemID();
+        if (render::Item::isValidID(itemID)) {
```
zzmp: Should this have been `isAllocatedID`?

samcake: No because we just push an update to the pending changes
We don t go getItem anymore so we don t care if it s allocated

PR #7316, opened 2016-03-11
Repair Toggle Overlays

ZappoMan:
- Hiding of overlays was recently broken, this fixes it
- This also fixes the previously reported issue where overlays can get stuck at a partial alpha

### DIFF
--- a/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
+++ b/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
@@ -410,27 +410,26 @@ void OpenGLDisplayPlugin::updateFramerate() {
 void OpenGLDisplayPlugin::compositeOverlay() {
     using namespace oglplus;
 
-
-    // set the alpha
     auto compositorHelper = DependencyManager::get<CompositorHelper>();
+
+    // check the alpha
     auto overlayAlpha = compositorHelper->getAlpha();
+    if (overlayAlpha > 0.0f) {
+        // set the alpha
+        Uniform<float>(*_program, _alphaUniform).Set(overlayAlpha);
 
-    if (overlayAlpha <= 0.0f) {
-        return; // don't render the overlay at all.
-    }
-    Uniform<float>(*_program, _alphaUniform).Set(overlayAlpha);
-
-    // Overlay draw
-    if (isStereo()) {
-        Uniform<glm::mat4>(*_program, _mvpUniform).Set(mat4());
-        for_each_eye([&](Eye eye) {
-            eyeViewport(eye);
-            drawUnitQuad();
-        });
-    } else {
         // Overlay draw
-        Uniform<glm::mat4>(*_program, _mvpUniform).Set(mat4());
-        drawUnitQuad();
+        if (isStereo()) {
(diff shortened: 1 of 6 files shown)


ZappoMan:
fixes these bugs:
- https://app.asana.com/0/26320649018758/95492826111228
- https://app.asana.com/0/26320649018758/99682714068198

libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
```diff
@@ -418,11 +432,20 @@ void OpenGLDisplayPlugin::compositeOverlay() {
...
+    Uniform<float>(*_program, _alphaUniform).Set(1.0);
 }
 
 void OpenGLDisplayPlugin::compositePointer() {
     using namespace oglplus;
     auto compositorHelper = DependencyManager::get<CompositorHelper>();
+
+    // set the alpha
+    auto overlayAlpha = compositorHelper->getAlpha();
+    if (overlayAlpha <= 0.0f) {
```
jherico: I would suggest reversing the test and putting the rendering inside the block, so that the last two lines of the function always execute, restoring the alpha and transform to their defaults.

As it stands I believe your app ends up leaving the alpha at (or very close to) 0 when you fade out, but this currently isn't noticeable because when we composite the scene texture to the framebuffer blending is disabled.  If anyone enables blending there for any reason, the app will suddenly break when the overlay fades out.

ZappoMan: But the way I have it coded, we leave the function the same as we enter it... by returning early we never touch the uniforms.

hifi-gustavo:
Build console output for commit: 387c156ce33f305607d01221da71c14ce7de0018 available at: https://gist.github.com/147473812929dab77511<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7316.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7316.dmg

libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
jherico: ok, I wasn't thinking it through enough... I see that you restore it every time you change it, so when it finally hits 0, the program state is left alone and should have '1' from the last call.

hifi-gustavo:
Build console output for commit: 92c47fdc8ba54f8798d97aa62aefdf1b4dae5a4b available at: https://gist.github.com/147473812929dab77511<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7316.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7316.dmg

OmegaHeron:
Tested Windows 7 - Setting overlays off works and persists.  Right click to bring up menu restores as does pressing enter to get location dialog.  Using a script from another PR that creates an overlay and changes its alpha value I saw no case of it getting stuck.  QA Pass

PR #7317, opened 2016-03-11
Expose a property in display plugins to determine if the device is visible to the user

jherico:
Note this only exposes the property.  It does not act on it.  Possible actions would include 

* Switching to a 2D display driver if the device is an HMD
* Ceasing to increment the present count so that we don't render frames that aren't visible to the user

However each of these has tradeoffs that will require further work and design discussion.  For instance, if we switch to a 2D display when the HMD is removed, do we automatically switch back when the HMD is put back on?  What if Interface is running in the background and someone decides to play something else?  

Not incrementing the present count means we remove the rendering load if the HMD is taken off, but the preview also stops updating.  

In other words, we probably want to act on this information, but we need a plan as to how before I go galloping off implementing something?

### DIFF
--- a/libraries/plugins/src/plugins/DisplayPlugin.h
+++ b/libraries/plugins/src/plugins/DisplayPlugin.h
@@ -65,6 +65,12 @@ public:
     virtual bool isThrottled() const { return false; }
     virtual float getTargetFrameRate() { return 0.0f; }
 
+    /// Returns a boolean value indicating whether the display is currently visible 
+    /// to the user.  For monitor displays, false might indicate that a screensaver,
+    /// or power-save mode is active.  For HMDs it may reflect a sensor indicating
+    /// whether the HMD is being worn
+    virtual bool isDisplayVisible() const { return false; }
+
     // Rendering support
 
     // Stop requesting renders, but don't do full deactivation
--- a/interface/src/scripting/HMDScriptingInterface.cpp
+++ b/interface/src/scripting/HMDScriptingInterface.cpp
@@ -92,3 +92,8 @@ glm::quat HMDScriptingInterface::getOrientation() const {
     }
     return glm::quat();
 }
+
+bool HMDScriptingInterface::isMounted() const{
+    auto displayPlugin = qApp->getActiveDisplayPlugin();
+    return (displayPlugin->isHmd() && displayPlugin->isDisplayVisible());
+}
--- a/interface/src/scripting/HMDScriptingInterface.h
+++ b/interface/src/scripting/HMDScriptingInterface.h
@@ -25,6 +25,7 @@ class HMDScriptingInterface : public AbstractHMDScriptingInterface, public Depen
     Q_OBJECT
     Q_PROPERTY(glm::vec3 position READ getPosition)
     Q_PROPERTY(glm::quat orientation READ getOrientation)
+    Q_PROPERTY(bool mounted READ isMounted)
 
 public:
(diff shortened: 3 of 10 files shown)


hifi-gustavo:
Build console output for commit: dc307018b382259f099db90192e756b8e0ce5509 available at: https://gist.github.com/5f57b874b046fa141183<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7317.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7317.dmg

jherico:
https://app.asana.com/0/27650181942747/95978054145798

hifi-gustavo:
Build console output for commit: 87c9961aa8b151eed409cfbae209be9ca14ed23f available at: https://gist.github.com/5f57b874b046fa141183<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7317.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7317.dmg

ZappoMan:
wrt - what to do with this info... I recommend that if the HMD goes from visible to not visible -- in the case of an HMD -- we assume that the user wants to enter the "away state"... I don't think we want to necessarily automatically go back to "here" when you put the HMD back on... but I think taking off the HMD is a guaranteed signal that the person wants to go to "away"

ZappoMan:
Needs to expose the state to JS... preferably through the HMD api.

Also, even though I agree with the idea that we could eventually use this in 2D mode to detect the screen saver... we probably want to be able to distinguish these states from JS... I will want to do something different when someone takes off their HMD than I necessarily would when the screen saver pops up.

hifi-gustavo:
Build console output for commit: 75c3fd18c9c766fe0b15b106dbc954104465c9cd available at: https://gist.github.com/5f57b874b046fa141183<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7317.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7317.dmg

libraries/plugins/src/plugins/DisplayPlugin.h
```diff
@@ -65,6 +65,12 @@ class DisplayPlugin : public Plugin {
     virtual bool isThrottled() const { return false; }
     virtual float getTargetFrameRate() { return 0.0f; }
 
+    /// Returns a boolean value indicating whether the display is currently visible 
+    /// to the user.  For monitor displays, false might indicate that a screensaver,
+    /// or power-save mode is active.  For HMDs it may reflect a sensor indicating
+    /// whether the HMD is being worn
+    virtual bool isDisplayVisible() const { return false; }
```
ZappoMan: should this default to `true`?

ZappoMan:
Test Notes:
- Run this build on the Vive
- open the Console window
- type `print(HMD.mounted);` in the console and press enter
- it should say `true` when the Vive hmd is on someones head, and `false` it's not

cozza13:
Tested and it work. You are working some crazy magic!

jherico:
It's just a sensor in the headset, similar to the kind used to turn off a smartphone screen when it's held to your head.  For the Vive Pre anyway.  The original vive and the DK2 don't have these sensors, so they use statistical analysis of the IMU data.

PR #7318, opened 2016-03-11
Asset Transfer Protocol Mappings

birarda:
# Asset Transfer Protocol Mappings

Working with hashes is no fun. In order to make ATP more usable for content development we are introducing the concept of path mappings. This will allow you to use your Asset Server like you would use a file system.

facf22f9a4c710af1eaaa3102b775f4c

Here is an overview of the features being introduced with this pull request:

- QML Asset Server browser to view and modify path mappings
- Upload of local files to Asset Server at a given mapping
- Rename of exising files/folders in the Asset Server
- Deletion of existing files/folders in the Asset Server
- "Add to World" to add model files from Asset Server in front of your Avatar 
- Persistence of mappings in Asset Server to local map.json file

Internally the Asset Server still has a directory of files whose filenames are the SHA256 hash of their content. The file system you will see via the QML interface in the High Fidelity client does not match how the Asset Server is storing the files on your machine. The client asks the Asset Server for its mappings and produces the displayed tree of files.

It would not be a protocol pull request without a trailing image. We were very happy to have @Atlante45 working with us again so that we could avoid having to make another one of these.

xtln3


hifi-gustavo:
Build console output for commit: 4d2797c25ad04eb4371352235be7935ae76363c0 available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

hifi-gustavo:
Build console output for commit: 843bff1dcf9292ca8dd71e152fe998bea0dc3b1c available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

hifi-gustavo:
Build console output for commit: 4bb3314d872eaf3b4b8fec23e353d988cf9e3590 available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

birarda:
test this please

hifi-gustavo:
Build console output for commit: 4bb3314d872eaf3b4b8fec23e353d988cf9e3590 available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

bluetoe:
This seems like a good idea but, renaming doesn't work for me. I run PR interface, go to PR domain, click on My Asset Server (which might be better being a rendered button like edit, directory, marketplace etc. or maybe better be part of edit) then I click on "Choose" and then ok.. choosing asset from local disk. (the second addy chooser is confusing). Then I click add to world check box. and upload.. or something like that, and it appears. And I tried creating new instance from the QML menu with the (select ATP object and then) "Add to World" button and that did the right thing.

When I quit.. and then log back in the object is there and I can see it in edit entity list and in "My Asset Server" list. So, persistence seems to work.

But when I rename an asset. (what i tested was an .fbx file), and then log off and back in, the object was no longer visible but was still on the "my asset server" with new name and entity list with original name. Other strange issues, but not able to reproduce.

Sorry I cannot be more precise about this, but either it doesn't quite work, or my expectation of what should be happening is wrong.

OmegaHeron:
Will likely take a few hours to fully test this.  But, first thing.  Compiling server side code, **only server side no Interface**, on Linux fails if QtWebChannel isn't installed - prior to this PR you could set a server only build flag and not need QtWebChannel.  Does it now actually need it or was CMake broken in that aspect?  Specifically in past `cmake -DSERVER_ONLY=TRUE path_to_source` allowed Linux to compile without that dependency, now it does not.

birarda:
@bluetoe 

Thanks for the feedback. We're currently discussing if "Add to World" will cause confusion because it may suggest that the added entity would have its modelURL updated should you rename the associated asset in the asset browser.

You shouldn't currently expect the entity modelURL to be renamed along side the file entry in the asset browser. Think of the asset browser like an S3 bucket or FTP server. Once you've put a file there and given it a name any links to that file (from scripts or via entity properties) would need to be updated if you decided to rename that asset.

If you overwrite a file of the same name with a new asset and cause a reload of that entity (go to a different domain and come back, quit interface and come back) you should see it update to the new asset.

Please let us know if any of the functionality that should be expected based on what I just described is not working.

birarda:
@OmegaHeron I'll look into the linking to see what may have caused an additional dependency on QtWebChannel.

OmegaHeron:
Well - I've managed to already bork my setup by renaming a named asset to a new name.  Cleaning up and starting over, but keeping detailed notes.

bluetoe:
Ah makes sense now. Yes, if I rename on S3, I get same result. So then for me it seems to do what is expected, at least for .fbx assets. Needs to check other types probably but I have to go deal with some stuff. Thanks for the explanation.

OmegaHeron:
First thing I notice is...  This is like orders of magnitude slower to load a scene.  I'm coming into a spot where I have one object + its compound shape, both are in cache as I see log say fetched from cache.  Prior to this PR I would enter my domain - and it would take 2-3 seconds for a sky island, zone and one object to appear.  This PR is more like 20 seconds to load only a sky island + compound (that's in disk cache).  Tried the rotate AV trick to make things turn up a bit faster.  No help.  I see Interface report in its log asset has loaded almost instantly, but, its appearance in scene seems to have no bearing on that.

Stack side runs on Linux and I'm not seeing any of this with current production code ATP server - it loads rapidly and consistently as I've come to expect.  There's no good way to see if it's a server or client side issue on rez lag since I can't seem to get current production client to play along with this PR's stack.

Lots more to come.

birarda:
@OmegaHeron that's very helpful, thanks. Could you provide the load portion of your log and give us some sense of when in time you see the scene load?

If the log shows that the mapping has been requested and the asset is in the cache, we must have introduced a bug that causes the actual loading of that asset to be slower.

OmegaHeron:
Will do after a bit of lunch @birarda - my hungries are impacting my logic abilities.  :)

birarda:
Okay @OmegaHeron enjoy! Is it possible that was the experience you had on first load?

We've removed the extension from ATP assets in the cache with this PR, meaning that any previously cached ATP assets would be re-downloaded when testing with this PR. If that were the issue, subsequent visits to the same domain would be the same 2-3 seconds as before.

OmegaHeron:
Yeah - I'm not a fan of extension removal in that it would complicate disaster recovery if mappings were lost, but, that's probably just me.  I'm still documenting things here, but, I can say this.  Repeated visits to same scene where all content is in cache **do** result in long scene load times, but, oh joy, not always.  Will post more once I have evidence chain completed for this first issue.

OmegaHeron:
How about this - look at those time stamps and it should well tell the story - making careful note that there are only 2 objects to load in entire domain (discounting AV) and both indicate fetched from cache.

`[03/11 15:40:59] [DEBUG] stats from new Entity server... [0.000000, 0.000000, 0.000000, 1.000000]`
`[03/11 15:40:59] [DEBUG] Physics disabled until entity loads:  QUuid("{322965ca-94a7-4674-a45d-b45fb3a2a11a}") ""`
`[03/11 15:40:59] [DEBUG] Starting request for: atp:/SkyIsland.fbx`
`[03/11 15:40:59] [DEBUG] Starting request for: atp:/skyisland-physics.obj`
`[03/11 15:40:59] [DEBUG] Got mapping for: "/SkyIsland.fbx" => "c025711042020f7e44a67b4e0a474d0ec8cdaf7bc6e22ad09dc1f99173f72a02"`
`[03/11 15:40:59] [DEBUG] "atp:c025711042020f7e44a67b4e0a474d0ec8cdaf7bc6e22ad09dc1f99173f72a02" loaded from disk cache.`
`[03/11 15:40:59] [DEBUG] Got mapping for: "/skyisland-physics.obj" => "3a3d80b15eaae649c997dd8ca21d7e39b5941217fbc7bcace7a8023cb7707bde"`
`[03/11 15:40:59] [DEBUG] Request finished for atp:/SkyIsland.fbx`
`[03/11 15:40:59] [DEBUG] Finished loading: atp:/SkyIsland.fbx`
`[03/11 15:40:59] [DEBUG] "atp:3a3d80b15eaae649c997dd8ca21d7e39b5941217fbc7bcace7a8023cb7707bde" loaded from disk cache.`
`[03/11 15:40:59] [DEBUG] Request finished for atp:/skyisland-physics.obj`
`[03/11 15:40:59] [DEBUG] Finished loading: atp:/skyisland-physics.obj`
`[03/11 15:41:00] [DEBUG] Physics disabled until entity loads:  QUuid("{322965ca-94a7-4674-a45d-b45fb3a2a11a}") ""`
`[03/11 15:41:05] [DEBUG] Physics disabled until entity loads:  QUuid("{322965ca-94a7-4674-a45d-b45fb3a2a11a}") ""`
`[03/11 15:41:10] [DEBUG] Physics disabled until entity loads:  QUuid("{322965ca-94a7-4674-a45d-b45fb3a2a11a}") ""`
`[03/11 15:41:15] [DEBUG] Physics disabled until entity loads:  QUuid("{322965ca-94a7-4674-a45d-b45fb3a2a11a}") ""`
`[03/11 15:41:20] [DEBUG] Physics disabled until entity loads:  QUuid("{322965ca-94a7-4674-a45d-b45fb3a2a11a}") ""`
`[03/11 15:41:25] [DEBUG] Physics disabled until entity loads:  QUuid("{322965ca-94a7-4674-a45d-b45fb3a2a11a}") ""`
`[03/11 15:41:29] [DEBUG] Finished loading: atp:/SkyIsland.fbx/skyIslandShineReduction2.fbm/noSpecular.png`
`[03/11 15:41:29] [DEBUG] Finished loading: atp:/SkyIsland.fbx/skyIslandShineReduction2.fbm/Rock.jpg`
`[03/11 15:41:29] [DEBUG] Finished loading: atp:/SkyIsland.fbx/skyIslandShineReduction2.fbm/Grass.jpg`
`[03/11 15:41:33] [DEBUG] Present thread shutdown`

So for some reason physics is taking a crazy long time to load and since we want ground to stand on it seems there's a block until we have ground to stand on.  Now moving onto why there's the huge lag in applying compound shape as it's indicated as loaded.

OmegaHeron:
Repeatedly relaunched until I hit a case where it loads as expected and you see;

`[03/11 15:50:37] [DEBUG] stats from new Entity server... [0.000000, 0.000000, 0.000000, 1.000000]`
`[03/11 15:50:37] [DEBUG] Physics disabled until entity loads:  QUuid("{322965ca-94a7-4674-a45d-b45fb3a2a11a}") ""`
`[03/11 15:50:37] [DEBUG] Starting request for: atp:/SkyIsland.fbx`
`[03/11 15:50:37] [DEBUG] Starting request for: atp:/skyisland-physics.obj`
`[03/11 15:50:37] [DEBUG] Got mapping for: "/SkyIsland.fbx" => "c025711042020f7e44a67b4e0a474d0ec8cdaf7bc6e22ad09dc1f99173f72a02"`
`[03/11 15:50:37] [DEBUG] "atp:c025711042020f7e44a67b4e0a474d0ec8cdaf7bc6e22ad09dc1f99173f72a02" loaded from disk cache.`
`[03/11 15:50:37] [DEBUG] Got mapping for: "/skyisland-physics.obj" => "3a3d80b15eaae649c997dd8ca21d7e39b5941217fbc7bcace7a8023cb7707bde"`
`[03/11 15:50:37] [DEBUG] Request finished for atp:/SkyIsland.fbx`
`[03/11 15:50:37] [DEBUG] Finished loading: atp:/SkyIsland.fbx`
`[03/11 15:50:37] [DEBUG] "atp:3a3d80b15eaae649c997dd8ca21d7e39b5941217fbc7bcace7a8023cb7707bde" loaded from disk cache.`
`[03/11 15:50:37] [DEBUG] Request finished for atp:/skyisland-physics.obj`
`[03/11 15:50:37] [DEBUG] Finished loading: atp:/skyisland-physics.obj`
`[03/11 15:50:37] [DEBUG] Finished loading: atp:/SkyIsland.fbx/skyIslandShineReduction2.fbm/Rock.jpg`
`[03/11 15:50:37] [DEBUG] Finished loading: atp:/SkyIsland.fbx/skyIslandShineReduction2.fbm/noSpecular.png`
`[03/11 15:50:37] [DEBUG] Finished loading: atp:/SkyIsland.fbx/skyIslandShineReduction2.fbm/Grass.jpg`

Attaching full log as well that contains this snippet

[hifi-log.txt](https://github.com/highfidelity/hifi/files/169948/hifi-log.txt)

hifi-gustavo:
Build console output for commit: e5376b12296573bd61540fff6fc2bccd35401e48 available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

hifi-gustavo:
Build console output for commit: 1502348731400cc3480d7ec7e2ef7fa8dec90c61 available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

hifi-gustavo:
Build console output for commit: ac7cbbd9600bf5eb0b94d244a0239eece891a196 available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

birarda:
@OmegaHeron okay, thanks. Looks like the initial asset download has completed successfully and it is have issues loading those textures. Are they embedded?

Could you send me `SkyIsland.fbx`? (Feel free to email it to me or send it to me on Slack).

OmegaHeron:
They are embedded - it's nothing special other than my crappy test island build 13...14 months ago.  I'll link a copy in a moment via slack

hifi-gustavo:
Build console output for commit: 66e572f22614e9f0fccf738d18c043c313a0f648 available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

hifi-gustavo:
Build console output for commit: 66e572f22614e9f0fccf738d18c043c313a0f648 available at: https://gist.github.com/b6ee5b87c0c7e06ecda8<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

hifi-gustavo:
Build console output for commit: df6eb8a081157adb7998818be4d76b8a55a47e20 available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

OmegaHeron:
Is there a reason why mouse scroll wheel would not work in this for zooming camera distance in 3rd person mode?

hifi-gustavo:
Build console output for commit: f92793b74886bf06c1c435efee4822829451eb66 available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

OmegaHeron:
Choose file for upload - dialog that open does not allow use of pulldown showing All Files (*.*) [Filter by type] -  Suggest that be hooked up proper, and set for common types like *.fbx *.obj and others possibly image/script formats.  Or - remove it if it can't be used.

When selecting a file its name/path is placed in text box left of pulldown for All Files (*.*) - you can't edit that text - suggest that too be fixed or changed to a non-teaser field type where it doesn't look like you could edit.

OmegaHeron:
Text for Add to world on upload is too light - it almost disappears into dialog's background and resembles text for a control that's inactive/disabled.

OmegaHeron:
When uploading a model and placing in scene it should set Name of placed object to name of model as is case when drag/dropping a file with add to scene enabled.

OmegaHeron:
When deleting a model only its reference is actually deleted, file remains on ATP server which is likely a good thing, but, for purposes of managing an ATP asset store and not ending up with gigabytes in deferenced content I would suggest maintaining a json format list of deleted references.  This would allow an easy culling routine to be developed.

OmegaHeron:
Drag and dropping a file to Interface results in My Asset Server dialog appearing with a file path filled - if I press upload I get a can't read file error.  Drag/drop path below;

`file:///C:/Users/Joe/Documents/HF-mega/hf-mega7c.fbx`

Navigating to same file and choosing via My Asset Server's Choose button works with it picking path of;

`c:/Users/Joe/Documents/HF-mega/hf-mega7c.fbx`

OmegaHeron:
There should be a mechanism for computing SHA256 on client side and doing a check to see if asset already exists at ATP server - there's no need to upload an existing ATP asset over and over - time and bandwidth waster.  It seems that is currently how it's done - no check for its existence - feed it back to ATP server even if a wasted operation.

OmegaHeron:
Could we have a journal containing name-of-asset-file-sent, sha256 of and user_id of sender?  Disaster recovery aide + responsibility tracking.  Sorry if my thinking more of "tomorrow" needs is a bit much for today.

OmegaHeron:
You're allowed to do things that leads to an unusable reference like calling a model MySuperThing (note no extension).  You can't add that to world - and if you decide, OK... I'll copy URL getting atp:/MySuperThing then try placing that.  Nice empty cube.  I think this is why my brain rebelled against dropping extensions on SHA256 file names - if the extension was maintained maybe it could handle such a thing relatively easily - without it looks like you either have to insure you never ever ever name an asset without its proper extension or file will have to be parsed to determine its content type.  It's probably not reasonable to assume people will not want to name things in a more human friendly format -- this looks and feels like an inventory system.

OmegaHeron:
Just in case it looks like I have nothing good to say - !true ;)  I uploaded an Avatar model and its textures such that it could fetch via ATP without embedded textures just to see if it could work.

hifi-snap-by-omegaheron-on-2016-03-12_12-45-06

Given some care in how textures are named and how placed withing virtual file system this could be nice!

assignment-client/src/assets/AssetServer.cpp
```diff
@@ -96,85 +101,209 @@ void AssetServer::completeSetup() {
...
+            qDebug() << "\tMoving" << oldAbsolutePath << "to" << oldAbsolutePath.replace(fullExtension, "");
+
+            bool renamed = oldFile.copy(_filesDirectory.filePath(hash));
+            if (!renamed) {
+                qWarning() << "\tCould not migrate pre-mapping file" << fileInfo.fileName();
+            } else {
+                qDebug() << "\tRenamed pre-mapping file" << fileInfo.fileName();
+
+                // add a new mapping with the old extension and a truncated version of the hash
+                static const int TRUNCATED_HASH_NUM_CHAR = 16;
```
AndrewMeadows: This is not a blocker, but is an unnecessary "wtf"?  http://www.osnews.com/images/comics/wtfm.jpg

This `const int`, which is used to avoid a magic number, should not be static.  In this context `static` means that once the variable has been initialized, it remains in memory until the end of the program.  As such, declaring a **constant** variable as static is an **anti-hint** to the compiler.  It may be that the compiler is smart enough to ignore the hint and instead just replace instances of the variable with a literal const but this is not guaranteed to be true.

I could let the **wtf** at line 60 slide... but I have to speak up when I see it inside a `for` loop.

assignment-client/src/assets/AssetServer.cpp
```diff
@@ -308,3 +437,267 @@ void AssetServer::sendStatsPacket() {
     // send off the stats packets
     ThreadedAssignment::addPacketStatsAndSendStatsPacket(serverStats);
 }
+
+static const QString MAP_FILE_NAME = "map.json";
```
AndrewMeadows: static const

interface/src/Application.cpp
```diff
@@ -4522,6 +4446,14 @@ void Application::toggleRunningScriptsWidget() {
     //}
 }
 
+void Application::toggleAssetServerWidget(QString filePath) {
+    static const QUrl url("AssetServer.qml");
```
AndrewMeadows: static const

libraries/networking/src/AssetResourceRequest.cpp
```diff
@@ -13,37 +13,99 @@
...
+        _assetMappingRequest->deleteLater();
+    }
+    
     if (_assetRequest) {
         _assetRequest->deleteLater();
     }
 }
 
+bool AssetResourceRequest::urlIsAssetHash() const {
+    static const QString ATP_HASH_REGEX_STRING = "^atp:([A-Fa-f0-9]{64})(\\.[\\w]+)?$";
```
AndrewMeadows: static const

assignment-client/src/assets/AssetServer.cpp
```diff
@@ -96,85 +101,209 @@ void AssetServer::completeSetup() {
...
 
-            auto hash = hashData(data);
-            auto hexHash = hash.toHex();
+    auto it = _fileMappings.find(assetPath);
+    if (it != _fileMappings.end()) {
+        auto assetHash = it->toString();
+        replyPacket.writePrimitive(AssetServerError::NoError);
+        replyPacket.write(QByteArray::fromHex(assetHash.toUtf8()));
+    }
+    else {
```
AndrewMeadows: This else should be on the same line as the closing bracket above, as per coding standard 4.1.5: https://docs.highfidelity.com/docs/coding-standard#section-4-1-5-the-if-else-class-of-statements-should-have-the-following-form-

interface/src/scripting/AssetMappingsScriptingInterface.cpp
```diff
@@ -0,0 +1,280 @@
...
+                        if (lastItem) {
+                            lastItem->setChild(lastItem->rowCount(), 0, item);
+                        } else {
+                            appendRow(item);
+                        }
+
+                        lastItem = item;
+                        _pathToItemMap[fullPath] = lastItem;
+                    }
+                    else {
```
AndrewMeadows: `else` should join line above it

assignment-client/src/assets/AssetServer.cpp
```diff
@@ -308,3 +437,267 @@ void AssetServer::sendStatsPacket() {
...
+                } else {
+                    ++it;
+                }
+            }
+
+            auto sizeNow = _fileMappings.size();
+            if (sizeBefore != sizeNow) {
+                qDebug() << "Deleted" << sizeBefore - sizeNow << "mappings in folder: " << path;
+            } else {
+                qDebug() << "Did not find any mappings in folder:" << path;
```
AndrewMeadows: The log message should be "Did not find any mappings **to delete** in folder:"

birarda:
@OmegaHeron wish I could reply directly to each comment - I'll do so in order here.

Appreciate the comments on the file dialog - that's an existing QML dialog and we're not planning on making changes to that with this PR. There is a new design for that dialog in the works so I assume it is to be re-worked shortly.

Will look at the text for Add To World.

Thanks, we'll hook upload to also use the filename of the uploaded file as the entity name. This new dialog replaces the previous drag and drop and add to scene flow.

Not an issue at all to have multiple paths to the same hash. We plan on shortly adding delete for any assets in the directory which do not have a reference. I want to wait on that until we get this feature out into the wild and confirm we don't have any bugs that remove mappings in unexpected cases. Deleting unmapped assets will be strategy in the short term to ensure that Asset Server directories only contain the files people can see they have mappings for. 

Will check on the file path for drag and drop upload. Sounds like an easy repro/fix.

100% agree that we should be avoiding upload when we can already check if the server already has an asset that matches that hash. We discussed that while testing this week. I believe that improvement is outside the scope of this PR (which is already quite large) but is something we'll look at adding soon.

An event log is certainly something that would be useful to provide domain owners. I'd consider that outside the scope of this PR but it's a feature request I'll keep in mind. It's not straight forward because you'd probably like to keep it at the level of the Asset Server, but the Asset Server has no idea what the filename of the file is when the contents are uploaded.

It is completely intentional that you can create filenames without an extension. Even if the hash URL that was returned for a given mapping included the extension that would not solve the issue you are describing. The system uses the original URL (like atp:/chair.fbx) to identify what type the file is. This is meant to work like an FTP server or S3 bucket. If you went into an S3 bucket and dropped the extension from a number of files they would likely be unreadable by clients that are not assuming the file type based on context. It's also possible that you might want two mappings to the same content to use different extensions. I could have `dictionary.json` and `dictionary.txt` point to the same file contents and both are perfectly valid. Perhaps we can add something to the asset browser that makes the user confirm anytime they decide to name a file without an extension.

assignment-client/src/assets/AssetServer.cpp
birarda: @AndrewMeadows I'm doing some more research but I'm fairly certain that until we have `constexpr` available in MSVC this would be the preferred way to have a local unchanging constant variable.

With just `const` you're only saying that the value cannot be changed - I believe it will be initialized every time the code is run. With the addition of `static` you're saying I know I only need this thing initialized one time AND thanks to `const` you know it's also unchanging.

OmegaHeron:
@birarda ty :)  And, to a certain extent, I consider that an avatar via ATP not being transportable a feature, for now.  I imagine ATP will be extended to allow its use across domains and multi-sourced for nearest ATP to end user via rough geo location, but, as that progresses - I would hope it never loses ability to operate much as it does now - a single point system for a single domain.

interface/src/scripting/AssetMappingsScriptingInterface.h
```diff
@@ -0,0 +1,77 @@
...
+//
+//  Created by Ryan Huffman on 2016-03-09.
+//  Copyright 2016 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#pragma once
+
```
AndrewMeadows: This `#pragma once` is new to me.  I looked it up and it appears to be a popular, but not universally supported, pragma for replacing the traditional `#indef hifi_filename_h` macro trick below.

libraries/script-engine/src/AssetScriptingInterface.cpp
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    // Make request to atp
+    auto path = urlString.right(urlString.length() - ATP_SCHEME.length());
+    auto parts = path.split(".", QString::SkipEmptyParts);
+    auto hash = parts.length() > 0 ? parts[0] : "";
+
+    if (hash.length() != SHA256_HASH_HEX_LENGTH) {
+        return;
+    }
```
AndrewMeadows: Should we be logging the two failure modes in this function?  Or are these failures sometimes expected in normal operation?  That is... would logging these failures tend to flood the logs unecessarily?

interface/src/scripting/AssetMappingsScriptingInterface.h
birarda: I have both in my HF C++ template. In the case where it's supported it's quicker and then I fall back to the more universally supported guard below.

hifi-gustavo:
Build console output for commit: 72c64532f5512ba472a59fe6f9af18d84ca15018 available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

hifi-gustavo:
Build console output for commit: bbfbbf65654a008ef2a1c3a30cddc3271e84faae available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

hifi-gustavo:
Build console output for commit: cbed674cba79c185ff59f90c41a23097a16df62a available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

hifi-gustavo:
Build console output for commit: 17e5b441cfe0aa1a9273a760bb31f242fcda1a1a available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

hifi-gustavo:
Build console output for commit: 17e5b441cfe0aa1a9273a760bb31f242fcda1a1a available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

assignment-client/src/assets/AssetServer.cpp
birarda: I think this one is redundant, will fix.

libraries/script-engine/src/AssetScriptingInterface.cpp
birarda: I'm going to let this fall through to AssetRequest since it will correctly set the invalid hash error case. Thanks for highlighting.

hifi-gustavo:
Build console output for commit: 42d7bc4263f4bc7295368a575c16c06928982a43 available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

hifi-gustavo:
Build console output for commit: 253b6e3067628221f90be3f574c398f62f1f7092 available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

OmegaHeron:
@birarda - working on it now.

assignment-client/src/assets/AssetServer.cpp
AndrewMeadows: @birarda from what I can tell: `constexpr` would be used when declaring a const that must be evaluated.  Something like:

constexpr int FOO = BAR * fubar(2);

Where you believe that fubar(2) can be evaluated at compile time.

We assume that the compiler will optimize out a simple local `const int` variable and replace it with a literal int wherever the variable was used.  Nevertheless, it is considered good form to declare the variable in the code for readability and also so the compiler can validate types.

Meanwhile, a static variable may introduce a branch operation in the final compiled instruction set since it will have to check: "if static is not initialized then intitialize it".  As I mentioned above, the compiler may be smart enough to realize that it doesn't need a static here and will use a literal int despite the static hint, but it is not obviously true.  We are less sure of this than we are that it will properly optimize out simple local consts.

birarda: You're right on constexpr, I didn't think it through completely.

I'll cave for a simple `int` but I still think it makes sense for a QString or QUrl to be `static const`. I'd rather be 100% sure initialization will only occur once on those than assume the compiler is going to do the right thing with just the `const`.

assignment-client/src/assets/AssetServer.cpp
birarda: Ended up leaving this one - from @Atlante45 "global variables inside a cpp file should be static so they can't be accessed from a different translation unit"

hifi-gustavo:
Build console output for commit: aa4c0ca15a560ef154a6298b94022bbb9c71a84c available at: https://gist.github.com/cdf9e76d1d7c94ffeb64<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7318.dmg

OmegaHeron:
Tested Windows 7 (Inerface) Linux (Server).  All that I had issue with that's within scope of this PR has been addressed.  Drag & Drop works, models have their ATP name as name of model when placed.  The only thing I'm unclear on is this supposed to be backwards compatible with previous Interface versions and domain construction?  I notice Current Interface can log in to domain running this PR, but, can not see any ATP assets.

birarda:
@OmegaHeron it won't be backwards compatible - we pushed the packet version for the respective Asset Server packets

OmegaHeron:
@birarda - Thanks.  Then I have nothing to block QA from my side.  It should be an interesting deploy and a great first step toward bigger things.  :+1:

assignment-client/src/assets/AssetServer.cpp
AndrewMeadows: Hrm... I guess I was mistaken.  According to this page: http://www.learncpp.com/cpp-tutorial/42-global-variables/

> By default, non-const variables declared outside of a block are assumed to be external. However, const variables declared outside of a block are assumed to be internal.

Which means... `static const int` is exactly the same thing as `const int`.  They are interchangeable.  We just need to have faith that the compilers are as smart as we hear they are.

I will modify how I read the code accordingly.

PR #7323, opened 2016-03-11
Zone visibility

howard-stearns:
Respect zone (skybox) visible property.


howard-stearns:
https://app.asana.com/0/32622044445063/92635876628765

howard-stearns:
There are a lot of different uses / side-effects / copies / special cases involved in zones. (There's even two different copies of the BackgroundMode.)  This change seems to me to be the most conservative way to fix the bug.

hifi-gustavo:
Build console output for commit: e538edba0a267b45d161e819a745c62751a23f19 available at: https://gist.github.com/36d90804e577e89a6216<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.dmg

hifi-gustavo:
Build console output for commit: e538edba0a267b45d161e819a745c62751a23f19 available at: https://gist.github.com/36d90804e577e89a6216<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.dmg

howard-stearns:
test this please

hifi-gustavo:
Build console output for commit: e538edba0a267b45d161e819a745c62751a23f19 available at: https://gist.github.com/36d90804e577e89a6216<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.dmg

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -375,8 +375,9 @@ void EntityTreeRenderer::applyZonePropertiesToScene(std::shared_ptr<ZoneEntityIt
                     qCDebug(entitiesrenderer) << "Failed to load skybox:" << zone->getSkyboxProperties().getURL();
                 }
             }
-
-            skyStage->setBackgroundMode(model::SunSkyStage::SKY_BOX);
+            // Visibility does not effect other side effects such as ambient light or the selection of "best" zone.
```
ZappoMan: hmmm -- interesting...

I think the spec is not clear on this...

But I assumed that if the zone was not visible, it would basically be ignored for "best zone"...

Let's ask for a clarification on the spec.

ZappoMan: chatting with @RyanDowne on this.. we want this zone visibility to have the following behavior:

If zone is not visible then:
- It does not effect lights (key light, ambient light)
- It does not render the skybox
- The "outer" zone's settings should prevail

So... I think the correct fix here, is to actually use the isVisible when picking the best zone... and then you get all the correct behavior.

Zones will still naturally support their other entity features like enter/leave events...

OmegaHeron:
Tested Windows 7 - Perhaps it's only because it's not what I want, but, this is a fail for me.  Yes it toggles visibility of sky texture, but, it leaves lighting in place so it's not really removed from any impact.  If its intent and how HF wants this to be is to leave its lighting in place but not show sky then I have to, begrudgingly, give it a pass.

ZappoMan:
@OmegaHeron - see my note above, I agree this should also toggle the lighting effects.

OmegaHeron:
@ZappoMan :+1:

howard-stearns:
I spoke with Ryan and, between us at least, the ticket being addressed is whether or not the skybox is visible, and Ryan will file a separate ticket to define and implement the other interactions (e.g., whether visible effects the "best" zone determination, or whether non-visible makes the zone not part of the scene for some things (such as ambient lighting) but not others (such as collisions).  That spec can then be prioritized wrt Beta in the usual way.

Is that ok with you @ZappoMan and @OmegaHeron?

OmegaHeron:
I'm fine with it in so far as mine is not to question why, but, do or die.  :)  It passes fine for the narrow interpretation.

ZappoMan:
I am not fine with this implementation... because the fix for doing it right... as I described... is small (in fact you just move where you're checking isVisible())

howard-stearns:
This alternative does it the other way: Only visible zones are considered when picking the best zone.  This is a little higher impact because the best zone was only being done when the avatar moved, so the code has to be moved around a bit to do that test all the time now.

hifi-gustavo:
Build console output for commit: 8799bf2d55484d6c52b7b1cc5523d1552971c176 available at: https://gist.github.com/36d90804e577e89a6216<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.dmg

howard-stearns:
hmm. work for me on Windows and mac. I never know what's going on with the these jensen failures now...

howard-stearns:
test this please

hifi-gustavo:
Build console output for commit: 8799bf2d55484d6c52b7b1cc5523d1552971c176 available at: https://gist.github.com/36d90804e577e89a6216<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.dmg

OmegaHeron:
Tested Windows 7 - Toggle of visibility now seems to make zone really invisible, lighting returns to same as no zone.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -157,61 +157,74 @@ void EntityTreeRenderer::update() {
     deleteReleasedModels();
 }
 
-bool EntityTreeRenderer::checkEnterLeaveEntities() {
+bool EntityTreeRenderer::findBestZoneAndMaybeContainingEntities(glm::vec3& avatarPosition, QVector<EntityItemID>* entitiesContainingAvatar) {
```
AndrewMeadows: When I saw that it was being passed as a regular reference it made me think: "Ah, this method must potentially change the value of `avatarPosition`," but when I looked at the implementation I found this not to be the case.  To make this clear to the reader you could declare it to be const referenence.

ZappoMan: I agree with Andrew on this.

hifi-gustavo:
Build console output for commit: 010aaa55c0ccf1e347f4b5360fe22e7e8b1412bb available at: https://gist.github.com/36d90804e577e89a6216<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.dmg

AndrewMeadows:
test this please

hifi-gustavo:
Build console output for commit: 010aaa55c0ccf1e347f4b5360fe22e7e8b1412bb available at: https://gist.github.com/36d90804e577e89a6216<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7323.dmg

PR #7328, opened 2016-03-11
reduce clamping of animation underpose by IK system

AndrewMeadows:
- reduce motion clamping of the underlying animation from the inverse-kinematics system

### DIFF
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -22,13 +22,9 @@ AnimInverseKinematics::AnimInverseKinematics(const QString& id) : AnimNode(AnimN
 }
 
 AnimInverseKinematics::~AnimInverseKinematics() {
-    std::cout << "adebug dtor" << std::endl;  // adebug
     clearConstraints();
-    std::cout << "adebug dtor 002" << std::endl;  // adebug
     _accumulators.clear();
-    std::cout << "adebug dtor 003 targetVarVec.size() = " << _targetVarVec.size() << std::endl;  // adebug
     _targetVarVec.clear();
-    std::cout << "adebug dtor 004 targetVarVec.size() = " << _targetVarVec.size() << std::endl;  // adebug
 }
 
 void AnimInverseKinematics::loadDefaultPoses(const AnimPoseVec& poses) {
@@ -491,7 +487,6 @@ RotationConstraint* AnimInverseKinematics::getConstraint(int index) {
 }
 
 void AnimInverseKinematics::clearConstraints() {
-    std::cout << "adebug clearConstraints  size = " << _constraints.size() << std::endl;  // adebug
     std::map<int, RotationConstraint*>::iterator constraintItr = _constraints.begin();
     while (constraintItr != _constraints.end()) {
         delete constraintItr->second;
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -22,7 +22,13 @@ AnimInverseKinematics::AnimInverseKinematics(const QString& id) : AnimNode(AnimN
 }
 
 AnimInverseKinematics::~AnimInverseKinematics() {
+    std::cout << "adebug dtor" << std::endl;  // adebug
     clearConstraints();
+    std::cout << "adebug dtor 002" << std::endl;  // adebug
+    _accumulators.clear();
+    std::cout << "adebug dtor 003 targetVarVec.size() = " << _targetVarVec.size() << std::endl;  // adebug
+    _targetVarVec.clear();
+    std::cout << "adebug dtor 004 targetVarVec.size() = " << _targetVarVec.size() << std::endl;  // adebug
 }
 
 void AnimInverseKinematics::loadDefaultPoses(const AnimPoseVec& poses) {
@@ -485,6 +491,7 @@ RotationConstraint* AnimInverseKinematics::getConstraint(int index) {
 }
 
 void AnimInverseKinematics::clearConstraints() {
+    std::cout << "adebug clearConstraints  size = " << _constraints.size() << std::endl;  // adebug
     std::map<int, RotationConstraint*>::iterator constraintItr = _constraints.begin();
     while (constraintItr != _constraints.end()) {
         delete constraintItr->second;
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -396,6 +396,17 @@ const AnimPoseVec& AnimInverseKinematics::overlay(const AnimVariantMap& animVars
             }
             _relativePoses[i].trans = underPoses[i].trans;
         }
+
+        if (!_relativePoses.empty()) {
+            // Sometimes the underpose itself can violate the constraints.  Rather than
+            // clamp the animation we dynamically expand each constraint to accomodate it.
+            std::map<int, RotationConstraint*>::iterator constraintItr = _constraints.begin();
+            while (constraintItr != _constraints.end()) {
+                int index = constraintItr->first;
+                constraintItr->second->dynamicallyAdjustLimits(_relativePoses[index].rot);
+                ++constraintItr;
+            }
+        }
     }
 
     if (!_relativePoses.empty()) {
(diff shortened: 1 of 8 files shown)


AndrewMeadows:
https://app.asana.com/0/26623005675174/97524676647514

There is a bug where the inverse-kinematics (IK) system will "distort" the input animation (aka "the underpose"), slightly altering the final effect.  The problem was that the rotational constraints of the joints (which are currently hard-coded in the C++) are too tight for the input, which causes some of the joints to "clamp" the animation's rotation.  Since the hand and feet locations are driven by IK targets, which themselves are sometimes driven by the animation underpose (depending on whether hand controllers are active or not) the IK system would distribute the missing rotations necessary to achieve the target positions up the joint hierarchy, shifting the joint motion intended for some joints to others.

This PR adds code to expand the constraint limits to "just barely" accommodate the underpose, which reduces (but does not completely fix!) the problem.

AndrewMeadows:
To test:

(1) Login to a location where there is room to walk the avatar
(2) Disable hand controllers so that there is no controller input for hand IK targets.
(3) Make sure inverse kinematics are **disabled**: Developer --> Avatar --> Enable Inverse Kinematics
(4) Switch to "morror mode" by pressing the "h" key.
(5) Press the UP arrow key several times to back the camera away from the avatar until at least the whole avatar torso is visible.
(6) Press and hold the "w" key to start walking forward.
(7) Pay careful attention to how the avatar's chest shifts left to right as the walk cycle animates -- this is what it is supposed to look like (IK disabled).

(8) Enable inverse-kinematics:  Developer --> Avatar --> Enable Inverse Kinematics
(9) Press and hold the "w" to start walking forward.
(10) Pay careful attention to how the avatar's chest shifts left to right as the walk cycle animates -- the ideal is for this animation behavior to look as much as possible to the case when IK is disabled.

(11) Compare animation in (7) to that in (10) for both this PR and the old version.  The IK-enabled results for this PR should be closer to ideal.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -394,6 +396,17 @@ const AnimPoseVec& AnimInverseKinematics::overlay(const AnimVariantMap& animVars
...
+        if (!_relativePoses.empty()) {
+            // Sometimes the underpose itself can violate the constraints.  Rather than
+            // clamp the animation we dynamically expand each constraint to accomodate it.
+            std::map<int, RotationConstraint*>::iterator constraintItr = _constraints.begin();
+            while (constraintItr != _constraints.end()) {
+                int index = constraintItr->first;
+                constraintItr->second->dynamicallyAdjustLimits(_relativePoses[index].rot);
+                ++constraintItr;
+            }
+        }
```
AndrewMeadows: This is where we perform the deed of expanding the joint constraints.  The rest of the code in this PR supports this block of logic.

libraries/animation/src/RotationConstraint.h
```diff
@@ -31,6 +31,10 @@ class RotationConstraint {
     /// \return true if this constraint is part of lower spine
     virtual bool isLowerSpine() const { return false; }
 
+    /// \param rotation rotation to allow
+    /// \brief clear previous adjustment and adjust constraint limits to allow rotation
+    virtual void dynamicallyAdjustLimits(const glm::quat& rotation) {}
```
AndrewMeadows: The **RotationConstraint** base class gets a pure virtual placeholder for this method, but only the **SwingTwistConstraint** has a non-trivial implementation.  We can add real implementation to the **ElbowConstraint** later if necessary.

libraries/animation/src/SwingTwistConstraint.cpp
```diff
@@ -24,32 +25,152 @@ const int LAST_CLAMP_NO_BOUNDARY = 0;
...
+    //   |              |
+    //   +-----+.       V       .+-----+-----+-----+-----+
+    //   |        -           -
+    //   |         ' A--X--B '
+    //  0+------------------------------------------------
+    //   0          pi/2         pi         3pi/2       2pi
+    //
+    // The code below computes the values of A and B such that the line between them
+    // passes through the point X, and we get reasonable interpolation for nearby values
+    // of theta.  The old AB values are saved for later restore.
```
AndrewMeadows: This comment provides an overview of what the "swing limit" adjustment is trying to do.

libraries/animation/src/SwingTwistConstraint.cpp
```diff
@@ -165,51 +285,57 @@ void SwingTwistConstraint::setTwistLimits(float minTwist, float maxTwist) {
...
+                // lower boundary is closer --> phase down one cycle
+                twistAngle -= TWO_PI;
+            } else if (twistAngle < midBoundary - TWO_PI) {
+                // higher boundary is closer --> phase up one cycle
+                twistAngle += TWO_PI;
+            }
+            break;
+    }
+    return twistAngle;
 }
```
AndrewMeadows: The code in **handleTwistBoundaryConditions()** was moved out of **apply()** because I also need to call it in **dynamicallyAdjustLimits()**.

tests/animation/src/AnimInverseKinematicsTests.cpp
```diff
@@ -228,31 +251,33 @@ void AnimInverseKinematicsTests::testSingleChain() {
...
-        QCOMPARE_WITH_ABS_ERROR(relativePoses[0].trans, origin, EPSILON);
-        QCOMPARE_WITH_ABS_ERROR(relativePoses[1].trans, xAxis, EPSILON);
-        QCOMPARE_WITH_ABS_ERROR(relativePoses[2].trans, xAxis, EPSILON);
-        QCOMPARE_WITH_ABS_ERROR(relativePoses[3].trans, xAxis, EPSILON);
+        QCOMPARE_WITH_ABS_ERROR(relativePoses[0].trans, origin, acceptableDistance);
+        QCOMPARE_WITH_ABS_ERROR(relativePoses[1].trans, xAxis, acceptableDistance);
+        QCOMPARE_WITH_ABS_ERROR(relativePoses[2].trans, xAxis, acceptableDistance);
+        QCOMPARE_WITH_ABS_ERROR(relativePoses[3].trans, xAxis, acceptableDistance);
     }
 }
```
AndrewMeadows: The changes in this code was done to fix some bit-rot in the **animation-tests** build target.  The tests are mostly fixed, but not completely.

hifi-gustavo:
Build console output for commit: 48efbba335c93dd6f31f0984ad2b34148d3b3791 available at: https://gist.github.com/5dc51a944b0bb71f6282<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7328.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7328.dmg

tests/animation/src/RotationConstraintTests.cpp
```diff
@@ -308,3 +309,344 @@ void RotationConstraintTests::testSwingTwistConstraint() {
...
+        }
+
+        // clear the adjustment using a null rotation
+        shoulder.dynamicallyAdjustLimits(glm::quat());
+
+        // verify that rotation is constrained again
+        constrainedRotation = totalRotation;
+        QVERIFY(shoulder.apply(constrainedRotation));
+    }
+}
```
AndrewMeadows: Much unit-testing of the "dynamic limit adjustment" logic.  I'm very confident that it is working correctly.

AndrewMeadows:
test this please

hifi-gustavo:
Build console output for commit: 48efbba335c93dd6f31f0984ad2b34148d3b3791 available at: https://gist.github.com/5dc51a944b0bb71f6282<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7328.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7328.dmg

hyperlogic:
Sorry for the delay, I'll review this in more detail on Monday

OmegaHeron:
Tested windows 7

(11) Compare animation in (7) to that in (10) for both this PR and the old version. The IK-enabled results for this PR should be closer to ideal

Compared to current production code I can see this is better - I had to repeat test multiple times to even convince myself the is a difference between IK On/Off with this PR while it was somewhat easy to see in production build.

(12) Make sure nothing IK related broke: hand controllers still work, jump animation is ok, etc.

No hand controllers here, but, jumping, landing, flying all seem fine to me via keyboard control.

hyperlogic:
Great stuff!  The walk/run cycle looks, much more like the source animation, the jump animation is improved as well.

PR #7343, opened 2016-03-13
Fix iDate uniform in procedurals to be continuous

jherico:
Currently the vec4 iDate uniform in procedural entities is supposed to provide the current year, month, day and second.  However, the second value doesn't change smoothly.  Instead it jumps from whole second to whole second.  This PR will allow it to have fractional values allowing for smoothly animated shaders that are based on the shared time for all viewers, rather than the local time since the shader started executing.

### DIFF
--- a/libraries/procedural/src/procedural/Procedural.cpp
+++ b/libraries/procedural/src/procedural/Procedural.cpp
@@ -374,7 +374,8 @@ void Procedural::setupUniforms() {
             v.y = date.month() - 1;
             // But not the day... go figure
             v.z = date.day();
-            v.w = (time.hour() * 3600) + (time.minute() * 60) + time.second();
+            float fractSeconds = (time.msec() / 1000.0f);
+            v.w = (time.hour() * 3600) + (time.minute() * 60) + time.second() + fractSeconds;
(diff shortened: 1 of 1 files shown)


jherico:
You can test the old and new behavior in my domain, which now orbits a Jupiter like gas giant.  If you look at the planet you'll see that it appears to 'tick' across the sky, rather than moving smoothly.  With this PR build it should have continuous smooth motion.

hifi-gustavo:
Build console output for commit: 27ecdffeb304f289a3e3e9c3577997e642484359 available at: https://gist.github.com/48cf522429f47c1eaec2<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7343.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7343.dmg

jherico:
https://app.asana.com/0/26225263936266/100240146725505

imgntn:
jupiter is rad!

imgntn:
capture

PR #7347, opened 2016-03-13
MyAvatar: prevent roll and pitch in avatar via snapshot url.

hyperlogic:


hyperlogic:
https://app.asana.com/0/45663370065758/29510440128830

hifi-gustavo:
Build console output for commit: b921ac77574c58dcc89f58cebdeac7ad5b8e4d03 available at: https://gist.github.com/cf0703a9d72ed0b37c51<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7347.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7347.dmg

ZappoMan:
I feel like the intent of the bug would be to have the orientation in the snapshot also set your head orientation. In the case of HMD, I guess that head orientation would be immediately overridden by whatever your actual head pose is. But in the case of desktop, we can set the head pose so you're looking at what the original snapshot taker was looking at.

jherico:
@ZappoMan we could set the pitch, but setting pitch or yaw on the head would be problematic because when not in HMD mode there's now way for the user to alter the head yaw or roll other than to reset the sensors.

ZappoMan:
Yeah - that's a good point. Well - ok, I think Tony's fix is probably the
best we can do then.

cozza13:
tested, took a photo and I did not get roll or pitch on the snapshot

PR #7362, opened 2016-03-15
Fix gpu timers

zzmp:
The range timer did not used to work, and I am not sure why.
This fixes it so that it will work, and I am not sure how.


hifi-gustavo:
Build console output for commit: c65aab299f42bdf7732f6e09dcff56a65429431e available at: https://gist.github.com/ecb5ab6d7ddd09c71c00<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7362.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7362.dmg

OmegaHeron:
Tested Windows 7 - Basic smoke test, nothing seems broken by this and there's no real test I can see beyond that.  Unless there's something more specific to look for in this, QA Pass.

libraries/gpu/src/gpu/Query.cpp
```diff
@@ -35,37 +22,25 @@ void Query::triggerReturnHandler(uint64_t queryResult) {
...
 }
 
 void RangeTimer::begin(gpu::Batch& batch) {
-    _headIndex++;
-    batch.beginQuery(_timerQueries[rangeIndex(_headIndex)]);
+    // Attempt to get the last query
+    batch.getQuery(_queries[_activeQuery]);
+
+    // Switch to a new query
+    _activeQuery = (_activeQuery == 0) ? 1 : 0;
```
AndrewMeadows: Timers are best when they are **fast** right?  From what I understand the `?:` operator introduces a branch instruction in the compiled code which supposedly has unfavorable performance implications.  A way to do this in a branchless way would be:  `_activeQuery = (_activeQuery + 1) % 2;`

hifi-gustavo:
Build console output for commit: a1713793386b6e46ca875586c946cfdfa4211dc1 available at: https://gist.github.com/ecb5ab6d7ddd09c71c00<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7362.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7362.dmg

libraries/gpu/src/gpu/Query.cpp
```diff
@@ -35,37 +22,25 @@ void Query::triggerReturnHandler(uint64_t queryResult) {
     }
 }
 
-
 RangeTimer::RangeTimer() {
-    for (int i = 0; i < QUERY_QUEUE_SIZE; i++) {
-        _timerQueries.push_back(std::make_shared<gpu::Query>([&, i] (const Query& query) {
-            _tailIndex ++;
+    for (int i = 0; i < 2; i++) {
```
AndrewMeadows: Sorry, I should have mentioned this before but I didn't notice it...

The coding standard says "no magic numbers" which means using `QUERY_QUEUE_SIZE` instead of 2 here (and where it makes sense elsewhere).  I don't mind using magic numbers when they aren't so magic (i.e. looping of the 3 components of a vec3 rather than using NUM_COMPONENTS) but in this case the code would become more readable and easier to maintain if `QUERY_QUEUE_SIZE` were defined and used.

**3.5.1.1** The use of magic numbers in the code should be avoided
https://docs.highfidelity.com/docs/coding-standard#section-3-5-1-1-the-use-of-magic-numbers-in-the-code-should-be-avoided-

samcake:
Adding DNM for now because this is not the right fix for the problem observed

samcake:
I think this is not a correct fix, let s revisit this when needed

PR #7369, opened 2016-03-15
repairs to udt SendQueue packet timing

birarda:
- fixed a bug in `udt::SendQueue` that would cause the SendQueue to not make its desired timing

This was especially bad on windows where `std::this_thread::sleep_for` frequently cannot sleep for less than a millisecond.

### DIFF
--- a/libraries/networking/src/udt/SendQueue.cpp
+++ b/libraries/networking/src/udt/SendQueue.cpp
@@ -280,11 +280,11 @@ void SendQueue::run() {
         // Once we're here we've either received the handshake ACK or it's going to be time to re-send a handshake.
         // Either way let's continue processing - no packets will be sent if no handshake ACK has been received.
     }
-        
+
+    // Keep an HRC to know when the next packet should have been
+    auto nextPacketTimestamp = p_high_resolution_clock::now();
+
     while (_state == State::Running) {
-        // Record how long the loop takes to execute
-        const auto loopStartTimestamp = p_high_resolution_clock::now();
-        
         bool sentAPacket = maybeResendPacket();
         
         // if we didn't find a packet to re-send AND we think we can fit a new packet on the wire
@@ -304,9 +304,15 @@ void SendQueue::run() {
             return;
         }
         
+        // grab the current HRC timestamp
+        const auto now = p_high_resolution_clock::now();
+
+        // push the next packet timestamp forwards by the current packet send period
+        nextPacketTimestamp += std::chrono::microseconds(_packetSendPeriod);
+
         // sleep as long as we need until next packet send, if we can
-        const auto loopEndTimestamp = p_high_resolution_clock::now();
-        const auto timeToSleep = (loopStartTimestamp + std::chrono::microseconds(_packetSendPeriod)) - loopEndTimestamp;
+        const auto timeToSleep = std::chrono::duration_cast<std::chrono::microseconds>(nextPacketTimestamp - now);
+
         std::this_thread::sleep_for(timeToSleep);
     }
 }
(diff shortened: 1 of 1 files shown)


birarda:
This is a fix for https://app.asana.com/0/11984398816645/100831266078435.

You should be able to test by uploading to a remote Asset Server. In the current release, it will not be possible on Windows to saturate your upload pipe. With this fix, it should use the full available bandwidth.

hifi-gustavo:
Build console output for commit: 1cd16da83edbd92105ed89dfb20aafebd3c4d0e0 available at: https://gist.github.com/b81f47d1e8ece01d876c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7369.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7369.dmg

libraries/networking/src/udt/SendQueue.cpp
```diff
@@ -304,9 +304,15 @@ void SendQueue::run() {
...
+        // grab the current HRC timestamp
+        const auto now = p_high_resolution_clock::now();
+
+        // push the next packet timestamp forwards by the current packet send period
+        nextPacketTimestamp += std::chrono::microseconds(_packetSendPeriod);
+
         // sleep as long as we need until next packet send, if we can
-        const auto loopEndTimestamp = p_high_resolution_clock::now();
-        const auto timeToSleep = (loopStartTimestamp + std::chrono::microseconds(_packetSendPeriod)) - loopEndTimestamp;
+        const auto timeToSleep = std::chrono::duration_cast<std::chrono::microseconds>(nextPacketTimestamp - now);
```
AndrewMeadows: Was it intentional to grab now before incrementing `nextPacketTimestamp`?

Since we're using a "high resolution clock" I'd expect the `timeToSleep` calculation to explicitly call `p_high_resolution_clock::now()` rather than use a stale value.

OmegaHeron:
Tested Windows 7 - Used Wireshark to profile network activity.  This PR with a 100MB source file approaches theoretical local port speed while production version is some (highly variable) amount less.  One precaution comes to mind with this.  I'm not sure if **UDT** will manifest this as does **UDP**, but, some home networking gear has rather small buffers for UDP traffic... if hammered by heavy UDP traffic where LAN (for example) = 100 MBit and outbound net = 5Mbit the device will suffer severe buffer overrun and do "bad things".  It may be worth watching for any increases in "I can't upload!!!" messages once this is promoted.  I pushed all my tests via a 50Mb down/5Mb up cable link and had no issue, but, I have gear that's known to have good UDP characteristics.  QA Pass

birarda:
@OmegaHeron thank you for test.

The UDT protocol, like TCP but with more fairness, is designed to back off when it sees any loss (often caused by saturation) on the link. In the cases you are describing I would expect the UDT flow to back off until it is not detecting loss. This should drop the overall throughput on network hardware back to serviceable levels, no matter the capacity.

hifi-gustavo:
Build console output for commit: c94e2bdc8b71dcf1925edf76aaef1fa8095c5456 available at: https://gist.github.com/b81f47d1e8ece01d876c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7369.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7369.dmg

OmegaHeron:
@birarda - Thanks!  I wanted to think that was so from looking at specs, but, that was pre proper caffeine intake.  It passes smoke testing here running Windows 7 client, Linux (with this PR code) Server and tested over net.

PR #7373, opened 2016-03-16
Lazy obj aux file reading

howard-stearns:
Don't read mtl files or default textures until we know that the file actually uses them.


howard-stearns:
This is the lazy evaluation part of https://app.asana.com/0/26225263936266/99290441458989/f
It still does things synchronously when getting or checking resources, but it only does so when it needs to.

hifi-gustavo:
Build console output for commit: 76b5b6aeeb60bb76e2235e49c6cd1f3965d06a3a available at: https://gist.github.com/9d1ec23f0b842f879543<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7373.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7373.dmg

howard-stearns:
I've added a hack to suppress the attempt to get referenced .mtllib files if the .obj file is specified with a url that has a "hifiusemat" query parameter. (E.g., ?hifiusemat=0, although it's not actually looking at the value of the parameter.) 

The idea here is that we have a (hacky) way forward if there we really really can't (or really really don't want) to upload the .mtlllib files referenced by an .obj (e.g., IFF the .obj is being used in a collision hull and we therefore don't need the material).  

In general, we can't have a cache for a single url that might be used with materials in one place and without materials in another. However, differences in url query part will result in different cache values, so the material version and the non-material version can coexist.

hifi-gustavo:
Build console output for commit: 798a225573139b3094159501d83e315ff14ffe70 available at: https://gist.github.com/9d1ec23f0b842f879543<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7373.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7373.dmg

libraries/fbx/src/OBJReader.cpp
```diff
@@ -178,6 +178,13 @@ void OBJFace::addFrom(const OBJFace* face, int index) { // add using data from f
     }
 }
 
+static bool replyOK(QNetworkReply* netReply, QUrl url) { // This will be reworked when we make things asynchronous
+    return netReply->isFinished() &&
+     (url.toString().startsWith("file", Qt::CaseInsensitive) ? // file urls don't have http status codes
```
ZappoMan: coding standard says something about indents for multi-line statements like this... I think it says you need to indent at least 1 full indent (or 4 spaces) for each subsequent line.

OmegaHeron:
I've made a first pass test on this - it would seem that in a case where it's known that an mtl won't be used, even if it were present, Interface should silently add ?hifiusemat=0 to an http request.  We know it's a compound shape - we know it's not going to use a mtl file... why should domain holders have to go through and edit potentially hundreds of models and add parameter?  It's already a rather hacky solution, might as well fill it out.

howard-stearns:
@OmegaHeron - Interesting idea. However:

1. I don't think we can safely do so, because not all Web servers ignore query parameters that they don't know how to deal with. (S3 does, but I think that might even be in violation of the http spec.)

2. I don't personally think we should so even if we can, because people should know that they are being evil. I feel we shouldn't be using .obj that reference other files and then not include those files. Making the change with a query replace in a models.json.gz should take less than a minute with a good editor, and is an explicit act.

But I'm open to discussion.  I'd really rather not have the query hack at all, but I'd rather "win" the argument and pull it out, than to argue without the existence proof of the hack in working code.

By the way, I'm working on some detailed testing notes. _Hope_ to have them within an hour...

OmegaHeron:
It's one thing for > geeks to decompress models.json.gz and search/replace and recompress -- it's another for those who simply want to make pretty and have work.  But, I won't argue the case -- one way or another, compound shape loading must stop making futile fallback requests for non existent assets and this will, at least, offer a way to do that.

howard-stearns:
Testing notes:  These are examples meant to illustrate the issues. You don't have to follow this particular test plan if you have a more practical idea.

1. Basic Operation: http://howard-stearns.github.io/models/monkey/monkey.obj logs that one .obj, one .mtl, and one .jpg are retrieved. The .obj and .jpg should be like all the other model and texture file-cache logging. (The .mtl is not cached in release or in this pr) It should NOT attempt to look for a default texture (monkey.jpg, monkey.png, etc.). (As shown in next item, the PR will log if it is doing for a default texture. The release does looks for the various other files, although it doesn't log so when compiled for release.)

2. Default Texure: http://howard-stearns.github.io/models/cyber/GregOrigSL.obj has no .mtl file, but it finds the texture file GregOrigSL.jpg, and logs that it is doing so. (Note: the head is huge. In order to see it, you have to back up a LOT to not be inside the head. Also, the model looks pretty weird with a skybox, but that's not the issue being tested. You should see a goatee and brown irises over white sclera, and this is easy to see if you remove any skybox, or toggle its visibility off.)

3. Basic Laziness: The default home environment loads simple-chair2.obj, which explicitly uses a material library simple-chair2.mtl, and references a material named "None". However, since all the names are "None" and the shape does not specify texture (UV) coordinates, the simple-chair2.mtl file is not retrieved. (As shown in the next item, the PR will log if it is doing so. The release attempts to get the file, which will fail becuase it hasn't been uploaded, but it does not log so when compiled for release.)

4. Query Parameter Hack: Make a copy of your AppData\Roaming\High Fidelity\assignment-client\entities\models.json.gz, and then download a url obtained from me or Chris to that directory. It will report an attempt to find many .mtllib files that fail because they have not been uploaded (at time of writing this). It should not attempt to find a matching .jpg, .png, etc. (which the release does). Now find one or more of these objects in these edit.js tool -- I hope the entity names are similar -- and change the collision compoundShapeURL (an .obj) to the same thing but with ?hifiusemat=0 appended. Observe that now there is no attempt in the log to load the missing .mttlib (and still not matching image probes). Exit interface and restart, and you you should get the new behavior (no .mtllib or image fetching for this .obj).

5. Overall: Do this url mangling for all the .obj in hq, and see if loads better/faster/more-reliably. (Instead of doing it by hand, I would edit the .model.gz file.)

hifi-gustavo:
Build console output for commit: 07214843e0834aeec81eb8ca250986534be3fa56 available at: https://gist.github.com/9d1ec23f0b842f879543<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7373.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7373.dmg

howard-stearns:
I've done (4) and (5) myself. For the purpose of this PR, "Pass". I'll add more notes in the ticket.

PR #7376, opened 2016-03-17

zzmp:

This results in a deadlock when the scene renderer (`paintGL`) attempts to resize the `OffscreenQmlSurface`, which requires the same `QMutex`. This only occurs when events are posted to execute the following, in order: `Application::paintGL`, `OffscreenQmlRenderer::render`. It will not be easy to reproduce.

The render "loop" in `OffscreenQmlSurface::updateQuick` is no longer necessary, and confusing to understand, so this PR removes it.
It looked like the texture should have been updated after the render (from the presence of a loop), so this PR also moves the `OffscreenQmlSurface::textureEmitted` signal so that it originates with the end of the render, in the renderer.


zzmp:

#### Testing

- This fixes a deadlock that I doubt can be reliably reproduced. Change your display mode a lot and make sure it doesn't affect anything, just in case. In production, this could (currently) possibly trigger a deadlock.

hifi-gustavo:
Build console output for commit: a1e152c8fd48d7bbe5648abdec3202fdf07ef8d6 available at: https://gist.github.com/144e255d90c27d2ca627<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7376.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7376.dmg

hifi-gustavo:
Build console output for commit: c0ca32bd165b78ee8c1eee93c9db0460d10f28ff available at: https://gist.github.com/144e255d90c27d2ca627<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7376.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7376.dmg

PR #7380, opened 2016-03-17

zzmp:

This results in a deadlock when the scene renderer (`paintGL`) attempts to resize the `OffscreenQmlSurface`, which requires the same `QMutex`. This only occurs when events are posted to execute the following, in order: `Application::paintGL`, `OffscreenQmlRenderer::render`; and the multithreading has it occur in that order. It will not be easy to reproduce.

This fixes this bug by only triggering the `RENDER` event, on the `OffscreenQmlRenderer`, and pausing execution on the main thread until that render finishes.


zzmp:

#### Testing

- This fixes a deadlock that I doubt can be reliably reproduced. Change your display mode a lot and make sure it doesn't affect anything, just in case. In production, this could (currently) possibly trigger a deadlock.

hifi-gustavo:
Build console output for commit: 73e97756b63d44115dd707df307b12204c523dcf available at: https://gist.github.com/5a9dd551d6c18fd419cb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7380.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7380.dmg

bluetoe:
I didn't quite know how to go about testing this. I opened up a bunch of qml overlays and then started changing my video card's display mode and resized the window moved stuff around etc. All worked well from 1920x1280 HD 1080p, 720p, 480p  down to 800x600 vga. But then I changed the color depth from 32 bit to 16 bit and crashed.

Not sure if that is what you meant by changing modes and I doubt many will be using 800x600 16 bit.

jherico:
@bluetoe modes here refers to switching between different display plugins, like Desktop, Stereo, Oculus, etc...

libraries/gl/src/gl/OffscreenQmlSurface.cpp
```diff
@@ -481,21 +471,19 @@ void OffscreenQmlSurface::updateQuick() {
...
 
-    if (_polish) {
-        _renderer->_renderControl->polishItems();
-        _polish = false;
-    }
-
-    if (_render) {
+    {
         QMutexLocker lock(&(_renderer->_mutex));
-        _renderer->post(RENDER);
```
jherico: The purpose of this logic derives from the documentation for `QQuickRenderControl::sync()`

> If a dedicated render thread is used, the GUI thread should be blocked for the duration of this call.

The idea here is that if we call sync, and some main thread event does something to the QML at the same time, it could result in a crash, or other undefined behavior.  Since the process events here can lead to deadlocking, then  we obviously need to find another way to resolve this, but we definately need to ensure the main thread does not do anything related to this QML context while sync is running on the other thread. 

The QQuickRenderControl example [here](http://doc.qt.io/qt-5/qtquick-rendercontrol-window-multithreaded-cpp.html) suggests that we might get better performance by *not* using an event loop for the rendering thread, but instead deriving a new QThread type and implementing run, with some kind of custom signaling.  This is more similar to what we're doing in the PresentThread in OpenGLDisplayPlugin.

zzmp: I thought that I'd retained this behavior, except restricted code that could run during the lock to the `QmlOffscreenRenderer`'s `RENDER` event.

@jherico are you saying that this is not a sufficient fix? Or that at some point (this PR or future), the rendering of the `OffscreenQmlSurface` be changed to avoid the event loop? Or something else?

bluetoe:
I tried it again, feeling dumb not knowing which modes, and no crashes. 

I did notice an unrelated quirk: Switching from windowed desktop to either side by side or interleaved stereo switches the display to full screen but switching from side by side to interleaved or from interleaved to side by side sets it to windowed.

libraries/gl/src/gl/OffscreenQmlSurface.cpp
jherico: OK, so I see that you do still have a wait here after issuing the `RENDER` post, but since you've removed the `_cond.wakeOne()` call in the `render()` method, I don't understand how the wait  on line 484 is ever satisfied.

hifi-gustavo:
Build console output for commit: 02926182307bee08231b1316033c076a61c7b9e1 available at: https://gist.github.com/5a9dd551d6c18fd419cb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7380.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7380.dmg

zzmp:
test this please

hifi-gustavo:
Build console output for commit: 95a0a5735796cd94a1f3c792696b76f63f4c0889 available at: https://gist.github.com/5a9dd551d6c18fd419cb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7380.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7380.dmg

hifi-gustavo:
Build console output for commit: 7a0090a0cfc90236c69a9b6c55b377fee5a5fc21 available at: https://gist.github.com/5a9dd551d6c18fd419cb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7380.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7380.dmg

hifi-gustavo:
Build console output for commit: 95a0a5735796cd94a1f3c792696b76f63f4c0889 available at: https://gist.github.com/5a9dd551d6c18fd419cb<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7380.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7380.dmg

libraries/gl/src/gl/OffscreenQmlSurface.cpp
```diff
@@ -481,26 +471,24 @@ void OffscreenQmlSurface::updateQuick() {
...
-            }
-            qApp->processEvents();
+        if (_polish) {
+            _renderer->_renderControl->polishItems();
+            _polish = false;
+        }
+
+        if (_render) {
+            _renderer->post(RENDER);
+            QCoreApplication::sendPostedEvents(_renderer, RENDER);
```
jherico: The documentation for this function says 
> Note: This method must be called from the thread in which its QObject parameter, receiver, lives.

`_renderer` does not live in the main UI thread, where this will be called.

zzmp:
Closing, as this bug will be fixed with a different approach.

PR #7382, opened 2016-03-17
Application: prevent crash on startup when "Crashed" dialog is shown

hyperlogic:
If the application "Interface Crashed" dialog is shown at startup, it can cause Application to receive events before it has been fully constructed.
In this case the QMainWindow _window pointer is uninitialized, causing Menu::getInstance() to de-reference an invalid pointer.

The fix for this is three fold.

1. Construct the MainWindow before calling setupEssentials (which in turn spawns the "Interface Crashed" dialog).
2. Change Menu::getInstance() from a static_cast to a dynamic_cast.  This will cause it to return a nullptr if [QMainWindow::menuBar()](http://doc.qt.io/qt-4.8/qmainwindow.html#menuBar) returns an instance of QMenuBar and not a ui::Menu instance.
3. In Application::event return false if Menu::getInstance() returns a nullptr.  This should prevent any code that relies on having a valid ui::Menu instance from running.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -452,8 +452,8 @@ Q_GUI_EXPORT void qt_gl_set_global_share_context(QOpenGLContext *context);
 
 Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
     QApplication(argc, argv),
-    _dependencyManagerIsSetup(setupEssentials(argc, argv)),
     _window(new MainWindow(desktop())),
+    _dependencyManagerIsSetup(setupEssentials(argc, argv)),
     _undoStackScriptingInterface(&_undoStack),
     _frameCount(0),
     _fps(60.0f),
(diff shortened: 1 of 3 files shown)


hyperlogic:
https://app.asana.com/0/45663370065758/100740284959681

hifi-gustavo:
Build console output for commit: a8bc8a78431dfe58a807992b6a3239a8d606c03f available at: https://gist.github.com/3160f8455c9d276b4a60<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7382.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7382.dmg

cozza13:
I loaded this up and did some testing. I did not crash more, therefore I am going to QA Approved.

PR #7385, opened 2016-03-18

zzmp:

This results in a deadlock when the scene renderer (`paintGL`) attempts to resize the `OffscreenQmlSurface`, which requires the same `QMutex`. This only occurs when events are posted to execute the following, in order: `Application::paintGL`, `OffscreenQmlRenderer::render`; and the multithreading has it occur in that order. It will not be easy to reproduce.

This fixes this bug by making the mutex recursive, and adding a static bool to quickUpdate to make sure it isn't triggered while already running. This allows the main thread to resize while holding a lock from rendering.


zzmp:

#### Testing

- This fixes a deadlock that I doubt can be reliably reproduced. Change your display mode a lot and make sure it doesn't affect anything, just in case. In production, this could (currently) possibly trigger a deadlock. ("display mode": desktop, oculus, 3dtv, etc.)

hifi-gustavo:
Build console output for commit: 2f950bb18f2d0de29b49718b6e348d1ba4d3b31b available at: https://gist.github.com/cb98e6a82324d175d82a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7385.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7385.dmg

OmegaHeron:
Tested Windows 7 - In desktop mode I can't open any menu items like, running scripts - open script - my asset server - any of settings.  The only window I can get an open on is non-qml script editor.  QA Fail

OmegaHeron:
Also, on 5 runs closing GUI has resulted in interface.exe continuing to run as a ghosted process requiring termination via task manager.

zzmp:
`QWaitCondition::wait: Cannot wait on recursive mutexes`.
Foiled by Qt!

hifi-gustavo:
Build console output for commit: 82fad865ffac947e4539dc404f3e915ace3783d2 available at: https://gist.github.com/cb98e6a82324d175d82a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7385.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7385.dmg

OmegaHeron:
For what it's worth as I didn't notice QA Ready removal.  All menus work and repeated attempts to nudge into crash mode playing tricks with window failed, so, as a smoke test it's pretty robust.

zzmp:
This doesn't actually remove the possibility of deadlock. I am working on moving it form an event-processing mechanism to running in a separate thread, with its own queue. Closing this, new PR to come.

PR #7389, opened 2016-03-18
Grab.js: Update Position on Mouse Grab Start

Menithal:
Makes sure that the moment the user starts dragging on a
object (that isn't grabbed by another), the position is immediately updated
instead of waiting for the user to move their mouse.

If this isn't done, the entity would fling it self back to the position grabbed position the moment the mouse if dragged: if the entity one with a lot of mass, this bug can cause unintentional flinging of entities that it collides with when coming back.

Can be tested by making a ball roll down a plane onto a flat horizontal plane and then trying to stop the ball with the grab script.

### DIFF
--- a/examples/grab.js
+++ b/examples/grab.js
@@ -373,6 +373,10 @@ Grabber.prototype.pressEvent = function(event) {
 
     beacon.updatePosition(this.startPosition);
 
+    if(!entityIsGrabbedByOther(this.entityID)){
+      this.moveEvent(event);
+    }
+     
     // TODO: play sounds again when we aren't leaking AudioInjector threads
     //Audio.playSound(grabSound, { position: entityProperties.position, volume: VOLUME });
 }


hifi-gustavo:

cozza13:
ok to test

hifi-gustavo:
Build console output for commit: 066c517dd741c3b58ea5801656800e282fee4705 available at: https://gist.github.com/397e99e2c0a45144200a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7389.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7389.dmg

imgntn:
https://rawgit.com/Menithal/hifi/on-grab-fix/examples/grab.js

TNS-TrueNorth:
Tested PR7389 with new grab.js on Windows 10:

Pass in Oculus First person view, with Leap Motion - it is almost impossible to keep the "mouse/hand" still while grabbing, so it is difficult to isolate the grab from the movement
Pass in Oculus First person and Third person view, with Razer Hydras - user can immediately stop a basketball in motion 
Fail in Oculus First person view and Third person view, with mouse - the basketball does not always stop if the user clicks the mouse without moving the mouse.  The unintentional flinging is much worse in third person view.

Pass in Desktop First person and Third person view, with Razer Hydras
Fail in Desktop First person and Third person view, with mouse

Menithal:
@TNS-TrueNorth

The fix should only impact the grab.js as in mouse grabs. Hand controllers are done through handControllerGrab.js last I checked.

cozza13:
@Menithal , thanks for sending this in. if it "Fail in Desktop First person and Third person view, with mouse" , does that mean it is not working?

Menithal:
@cozza13 unfortunately yes, but I simply cannot get that to not fail (unless i miss the object). I made a video demonstrating the bug, and the current built version that was provided above were I cover both the cases with the current grab script vs the current one. I will upload it shortly.

Menithal:
Here is the video of me demostrating the bug and the fix from both first person and third person view
https://youtu.be/XprZBZTOdds

Can someone replicate @TNS-TrueNorth's result?

TNS-TrueNorth:
Thanks @Menithal for the video.  I noticed that you were loading the .js via URL rather than from local disk.  I had been saving the .js file and loading it locally.

When loading the new grab.js via URL, this passes in Windows 10.
Pass in Oculus First person view and Third person view, with mouse
Pass in Desktop First person and Third person view, with mouse 

@cozza13 the following may be a different bug/issue: 
The user cannot copy the text at https://rawgit.com/Menithal/hifi/on-grab-fix/examples/grab.js then save it as a .js file, and load via local disk.

Menithal:
@TNS-TrueNorth I think if you load the scripts through the Running Scripts window, it actually loads them from the S3, unless you specifically load the script from disk

PhilipRosedale:
Tested for grabbing as described, works well.  Good one, thanks.

PR #7393, opened 2016-03-18
Fix for crash in recalculateMeshBoxes()

ZappoMan:
clean up the recalculateMeshBoxes() calls in findRayIntersectionAgainstSubMeshes() so that it doesn't recalculateMeshBoxes while iterating through the existing vector of submeshes

### DIFF
--- a/libraries/render-utils/src/Model.cpp
+++ b/libraries/render-utils/src/Model.cpp
@@ -252,25 +252,18 @@ bool Model::findRayIntersectionAgainstSubMeshes(const glm::vec3& origin, const g
 
         // If we hit the models box, then consider the submeshes...
         _mutex.lock();
-
-        if (!_calculatedMeshBoxesValid) {
+        if (!_calculatedMeshBoxesValid || (pickAgainstTriangles && !_calculatedMeshTrianglesValid)) {
             recalculateMeshBoxes(pickAgainstTriangles);
         }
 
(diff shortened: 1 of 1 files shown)


ZappoMan:
Possible fix for this bug - https://app.asana.com/0/26320649018758/101905459350866

Testing notes: Smoke test the build, in particular, using HMD, make sure that depth reticle still works properly and that you can still edit/select things with edit.js

Dev Notes: The crash appears to be an out of memory in appending to the end of the QVector<> for the mesh sub boxes. It's not exactly clear why that would happen, but one thing we were doing was recalculating the mesh boxes while already in the process of iterating that list. Qt handles this by doing a copy on write... of course, this is a total waste of time and memory... since we can know if we need to recalc the boxes before we start iterating.

hifi-gustavo:
Build console output for commit: 925d5d36acfb6a51074383ae36ed8f900f712ca9 available at: https://gist.github.com/3e7822c57fad5fe9849e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7393.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7393.dmg

cozza13:
Tested with HMD. I could go into edit and edit things and move them around.

PR #7396, opened 2016-03-18
prevent crash in SwingTwistConstraint

AndrewMeadows:
- SwingTwistConstraint is more resilient when getting bad data (will not crash)

### DIFF
--- a/interface/src/avatar/AvatarUpdate.cpp
+++ b/interface/src/avatar/AvatarUpdate.cpp
@@ -55,6 +55,7 @@ bool AvatarUpdate::process() {
         deltaMicroseconds = 10000; // 10 ms
     }
     float deltaSeconds = (float) deltaMicroseconds / (float) USECS_PER_SECOND;
+    assert(deltaSeconds > 0.0f);
     _lastAvatarUpdate = start;
     qApp->setAvatarSimrateSample(1.0f / deltaSeconds);
 
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -481,7 +481,7 @@ const AnimPoseVec& AnimInverseKinematics::overlay(const AnimVariantMap& animVars
 
             // smooth transitions by relaxing _hipsOffset toward the new value
             const float HIPS_OFFSET_SLAVE_TIMESCALE = 0.15f;
-            float tau = dt > HIPS_OFFSET_SLAVE_TIMESCALE ? 1.0f : dt / HIPS_OFFSET_SLAVE_TIMESCALE;
+            float tau = dt < HIPS_OFFSET_SLAVE_TIMESCALE ?  dt / HIPS_OFFSET_SLAVE_TIMESCALE : 1.0f;
             _hipsOffset += (newHipsOffset - _hipsOffset) * tau;
         }
     }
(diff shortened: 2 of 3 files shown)


AndrewMeadows:
https://app.asana.com/0/26623005675174/102651740414852

There was a crash report that suggested it was possible for the SwingTwistConstraint to get NaN arguments which could cause it to crash when trying to index into a lookup table.  This PR introduces countermeasures against such an attack.

To test:  

(1) avatar inverse kinematics (IK) should still work: make sure hand controllers still move avatar arms around and the feet still track the ground when the hips shift due to IK.

(2) the IK system should not distort the underlying animation:
(2.A) go into mirror mode and back the camera out until the entire avatar torso is visible
(2.B) walk the avatar forward and backward, paying close attention to how much the avatar's torso swings back and forth during the walk-cycle
(2.C) disable IK (uncheck: Developer --> Avatar --> Enable Inverse Kinematics)
(2.D) Repeat step (2.B) --> the motion of the avatar's torso should match very closely to (2.B)

interface/src/avatar/AvatarUpdate.cpp
```diff
@@ -55,6 +55,7 @@ bool AvatarUpdate::process() {
         deltaMicroseconds = 10000; // 10 ms
     }
     float deltaSeconds = (float) deltaMicroseconds / (float) USECS_PER_SECOND;
+    assert(deltaSeconds > 0.0f);
     _lastAvatarUpdate = start;
     qApp->setAvatarSimrateSample(1.0f / deltaSeconds);
```
AndrewMeadows: In the callstack of the crash I examined the value of deltaSeconds appeared to be bogus (either 0.0, very small (< 3e-9) or NaN).  It is likely these were caused by optimization corruption of the Release build executable, but just in case...I'm adding this assert() which will catch negative and NaN values in debug builds.

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -481,7 +481,7 @@ const AnimPoseVec& AnimInverseKinematics::overlay(const AnimVariantMap& animVars
 
             // smooth transitions by relaxing _hipsOffset toward the new value
             const float HIPS_OFFSET_SLAVE_TIMESCALE = 0.15f;
-            float tau = dt > HIPS_OFFSET_SLAVE_TIMESCALE ? 1.0f : dt / HIPS_OFFSET_SLAVE_TIMESCALE;
+            float tau = dt < HIPS_OFFSET_SLAVE_TIMESCALE ?  dt / HIPS_OFFSET_SLAVE_TIMESCALE : 1.0f;
```
AndrewMeadows: Inverting the logic here will catch and filter the case when `dt` happens to be NaN.

hifi-gustavo:
Build console output for commit: e81e49b32e74e05c819a03d0cc5a0af4b64baded available at: https://gist.github.com/e8204871cf8318d9b11f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7396.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7396.dmg

Nex-Pro:
I'll take a look at this one.

OmegaHeron:
Tested Windows 7 - If there's any difference between IK on/off with this it's smaller than I can see.  :+1: 
Unable to test with hand controllers so no comment there.

PR #7400, opened 2016-03-19
Set UserAnimPreAndPostRotations true by Default

Menithal:
If Blender avatars are animated by Blender animations, they have missing
Pre and Post rotations. This step is no longer necessary to have false,
and also makes sure all animations from any suite Work on all avatars (from
other platforms).

This commit make it default to be enabled:

Details available at
https://alphas.highfidelity.io/t/ultimate-blender-compendium-woes-of-the-blender-fbx-export-at-the-mountains-of-documentation-madness/4739/102

-----

Can be tested using the following:

Mixamo Exported Avatar:
http://www.norteclabs.com/HF/labs/tests/skinning_test_mixamo_autorig.fst

Blender Reexported Avatar:
http://www.norteclabs.com/HF/labs/tests/skinning_test_mixamo_autorig_blender_re-exportYX.fst

Run these in the scripting console:
*Control*: MyAvatar.overrideAnimation("http://www.norteclabs.com/HF/labs/tests/skinning_test_mixamo_autorig.fbx", 30, true, 0, 100);

*Blender Export*: MyAvatar.overrideAnimation("http://www.norteclabs.com/HF/labs/tests/skinning_test_mixamo_autorig_blender_re-exportYX.fbx", 30, true, 0, 100);

### DIFF
--- a/interface/src/Menu.cpp
+++ b/interface/src/Menu.cpp
@@ -107,7 +107,7 @@ Menu::Menu() {
 
     auto scriptEngines = DependencyManager::get<ScriptEngines>();
     // Edit > Stop All Scripts... [advanced]
-    addActionToQMenuAndActionHash(editMenu, MenuOption::StopAllScripts, 0, 
+    addActionToQMenuAndActionHash(editMenu, MenuOption::StopAllScripts, 0,
         scriptEngines.data(), SLOT(stopAllScripts()),
         QAction::NoRole, UNSPECIFIED_POSITION, "Advanced");
 
@@ -140,7 +140,7 @@ Menu::Menu() {
     // Edit > Reload All Content [advanced]
     addActionToQMenuAndActionHash(editMenu, MenuOption::ReloadContent, 0, qApp, SLOT(reloadResourceCaches()),
         QAction::NoRole, UNSPECIFIED_POSITION, "Advanced");
-    
+
 
     // Edit > Package Model... [advanced]
     addActionToQMenuAndActionHash(editMenu, MenuOption::PackageModel, 0,
@@ -153,7 +153,7 @@ Menu::Menu() {
     auto audioIO = DependencyManager::get<AudioClient>();
 
     // Audio > Mute
-    addCheckableActionToQMenuAndActionHash(audioMenu, MenuOption::MuteAudio, Qt::CTRL | Qt::Key_M, false, 
+    addCheckableActionToQMenuAndActionHash(audioMenu, MenuOption::MuteAudio, Qt::CTRL | Qt::Key_M, false,
         audioIO.data(), SLOT(toggleMute()));
 
     // Audio > Show Level Meter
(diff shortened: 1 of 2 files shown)


hifi-gustavo:

sethalves:
ok to test

hifi-gustavo:
Build console output for commit: cdbb7b2ad4294796ae007c0d74f0c07df1d5a4dd available at: https://gist.github.com/507f340e341188770e12<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7400.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7400.dmg

ZappoMan:
Why is this tied to a menu item? Is there a way we can detect that this animation should have this setting set automatically? Generally speaking we want to reduce the number of options added to the menus, not add more. Can this be detected automatically?

Menithal:
@ZappoMan Technically yes, but (according to the test gauntlet)[https://alphas.highfidelity.io/t/ultimate-blender-compendium-woes-of-the-blender-fbx-export-at-the-mountains-of-documentation-madness/4739/103] I did, it could be set to true for all animation cases.

When enabled:

- Maya Avatars  works with Maya Animations
- Blender Avatars works with Blender Animations
- Maya Avatars works with Blender Animations
- Blender Avatars works with Maya Animations

If disabled:
All Blender Animations are broken with all Avatars.

-----

Iteratively, the PR is still have the option still there, but if there are any edge cases it could still just be a toggle able thing via script instead. If there are no edge cases for a while (and no one finds an issue) the entire option could be removed.

Menithal:
I tested the current build set and the current release, apparently something changed somewhere and I cant replicate the state on how I remember animations working in builds in February 25th (yes, 200+ builds ago, but which I am unable to start them any longer so I cannot validate and reproduce it there):

They still work in the correct direction, but when the binding pose test animation rexported in Blender reaches back to the T-pose, The pose is off. Neither case has them correct, but having the pre-post true is the closest We can get without starting to look further into the Animation code.

hyperlogic:
@ZappoMan Basically, we've been doing the wrong thing with pre-post rotations for a long time, But our previous avatar set relied on this broken behavior, so we made it a menu option.  Pending @ozanserim updating the avatars skeletons to conform with our standard.  Which was recently completed.

The thing we were doing incorrectly, is that we were applying the pre and post animations from the avatar's fbx file, to the rotations contained in the animation fbx files.  The correct approach, is to apply the pre-and post rotations in the animation fbx, to the rotations in the same animation fbx file.  As a side effect, this allows us to apply animations from maya to avatar's created in blender, and possibly vice versa.

ZappoMan:
So do we need the menu option @hyperlogic ?? or can we just always have this code running?

OmegaHeron:
@ZappoMan - unless I'm misunderstanding, this isn't adding a menu option.  It's changing a default setting to an existing option located in Developer->Avatar.  Though I do agree if it ultimately has no use it should be removed which seems to be an eventual outcome.

ozanserim:
Using the PreAndPostRotations has been very effective so far. Glad it's on by default. 
Perhaps we can just retain the option to turn it off for debugging purposes. 

Ozan

ozanserim:
I also ran tests on this:

ozanserim:
I also ran tests on this:
- When I run the PR build, Interface has the "Use anim pre.." option set to off (perhaps due to my previous settings). 
- When I toggle the option off and on, the animations work correctly with existing hifi avatars and the mixamo/blender avatars. 
- When I run the control override script, the animation is correct. 
- When I run the blender override script, animation is incorrect. The joint angles appear slightly off. See attached. 

screenshot_1

Menithal:
@ozanserim 

It seems it is at least in the correct direction, unlike the folded up avatar without this option being set true. I may have been mistaken with my earlier tests with this case (I may have just missed the binding pose of the start being incorrect at the start). 

Seems to be a deeper issue. I'll leave it to you guys to decide if this should be true by default for now. It is interesting to see both Blender and Maya avatars behave in the same way with just Blender Animations.

PR #7401, opened 2016-03-19
Up the ResourceCache simultaneous request limit from 3 to "cores" - 2.

howard-stearns:


howard-stearns:
This isn't attached to any bug, but it is inpsired by @jherico comment on https://app.asana.com/0/26225263936266/99290441458989.

The idea is that we need main and present, and everthing else can be shared while we burst downloads (which spend most of their time waiting for socket data). On a typical 8 core machine, that gives 6 downloads, or twice what we're allowing before this PR.

I understand that Google periodically analyses the ideal number of parallel GET fo Chrome, and the number has been stable at 10 for many years. However, Web pages are typically smaller than the resources we load, and have a lower percentage of S3 resources (which are unbelievably slow). So a slightly smaller number than 10 might make sense anyway. (Web fetches typically come from the same machine with open connections, though, and that might have the biggest effect on what the ideal number should be for that use case.)

hifi-gustavo:
Build console output for commit: f0552e1d7fb04094b7d05a6d44eeac3564259dde available at: https://gist.github.com/d1424b62cf596b5760d5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7401.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7401.dmg

interface/src/Application.cpp
```diff
@@ -695,7 +695,8 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
     connect(&identityPacketTimer, &QTimer::timeout, getMyAvatar(), &MyAvatar::sendIdentityPacket);
     identityPacketTimer.start(AVATAR_IDENTITY_PACKET_SEND_INTERVAL_MSECS);
 
-    ResourceCache::setRequestLimit(3);
+    int nThreads = QThread::idealThreadCount();
+    ResourceCache::setRequestLimit((nThreads <= 0) ? 3 : nThreads - 2);
```
ZappoMan: If you happen to be on a single or dual core machine (admittedly unlikely, but someday might be true) -- you end up with a negative number. I think you want: `((nThreads <= 4) ? 3 : nThreads - 2)`.

And of course... "Magic" so... really you want something like...

```
const int MINIMUM_NUMBER_OF_CORES = 4;
const int MINIMUM_NUMBER_OF_REQUESTS = 3;
const int THREADS_NOT_USED_FOR_REQUESTS = 2;
ResourceCache::setRequestLimit((nThreads <= MINIMUM_NUMBER_OF_CORES ) 
            ? MINIMUM_NUMBER_OF_REQUESTS  : nThreads - THREADS_NOT_USED_FOR_REQUESTS );
```

ZappoMan:
I can see the merit in this change... But I'm not sure what the right number to reserve for non-request threads really is... How did you pick 2? 

We do have of course many other threads that are running, many of which are often idle waiting on traffic requests... it also seems to me that the resource cache is going to have threads/request that are often happily waiting on network input, and so limiting our requests to our number of cores is probably overkill. Many requests could likely sit happily on a single core without over-working it, since often the thread will be quietly waiting on a I/O interrupt.

Atlante45:
I think there is a confusion here. resource request are all on the same thread and that thread is mostly sitting idle waiting for network data.

What @jherico was talking about is the number of thread available in the thread pool that processes resources and transforms them into an object representation.
That limit is set at the beginning of Application's ctor line 484.

ZappoMan:
@Atlante45 - thanks for clarifying... so basically, this PR doesn't make really make sense, why make the number of pending requests (on a single request thread) be dependent on the number of cores. What we really want is a mechanism to set the number of "parsing" threads. @howard-stearns - is there some other side effect of this PR that you expected to cause the number of pending requests to turn into a similar number of parsing threads, was that your intent of this PR?

There's a separate question as to whether or not 3 is the right number of simultaneous pending requests... how did we come up with that number?

Atlante45:
This number goes back as far as I remember. I think Andrzej might have been
the one to set that limit.

I didn't know that, but if Google found that 10 concurrent downloads is the
way to go, we might want to try that out.

ZappoMan:
This article seems like a worthwhile read... namely... it seems like most modern web browsers set a limit on a PER server basis...

http://sgdev-blog.blogspot.com/2014/01/maximum-concurrent-connection-to-same.html

In particular, it sounds like Chrome for example, will 6 simultaneous downloads PER SERVER.... significantly more than we currently support. 

Of course, this is all based on their empirical testing of HTTP 1.1... and so it is likely not the correct value for a blend of HTTP and ATP.

Atlante45:
Yes, and the files they are loading are also much smaller.
I wouldn't be surprised that the ideal number of concurrent download for model files is different than the one for images and source files.

ZappoMan:
We should do a simple empirical set of tests... 

1) allow this value to be set via an undocumented scripting interface.
2) run several tests of loading our "launch content" with different settings from 3 up to 10 or even 20.
3) measure the total scene load time for each settings.
4) repeat experiment on a handful of different machines and network setups.

Pick MVP value based on results.

So... with that in mind @howard-stearns - can you change this PR to expose setting this value via scripting interface?

howard-stearns:
Sounds good. I take this as empirically optimizing launch scenes loading within the current design.

Thus I'll try spinning:
- the ResourceCache count used for downloading (as shown so far), and
- the thread pool size used for mode/texture/animation parsing after download (as mentioned by Clement),
... but I'll keep everything else out of scope for now. (E.g., no futzing with other counts, nor of priorities, and no dynamic behavior to be different at different times or for different types, etc.)

howard-stearns:
will reopen after the above.

PR #7402, opened 2016-03-20
NetworkGeometry: Fix crash for empty geometry/bad FBX versions

hyperlogic:
Interpret empty geometry as an error.

### DIFF
--- a/libraries/model-networking/src/model-networking/ModelCache.cpp
+++ b/libraries/model-networking/src/model-networking/ModelCache.cpp
@@ -72,11 +72,15 @@ void GeometryReader::run() {
                 const bool grabLightmaps = true;
                 const float lightmapLevel = 1.0f;
                 fbxgeo = readFBX(_data, _mapping, _url.path(), grabLightmaps, lightmapLevel);
+                if (fbxgeo->meshes.size() == 0 && fbxgeo->joints.size() == 0) {
+                    // empty fbx geometry, indicates error
+                    throw QString("empty geometry, possibly due to an unsupported FBX version");
+                }
             } else if (_url.path().toLower().endsWith(".obj")) {
                 fbxgeo = OBJReader().readOBJ(_data, _mapping, _url);
             } else {
                 QString errorStr("unsupported format");
-                emit onError(NetworkGeometry::ModelParseError, errorStr);
+                throw errorStr;
             }
             emit onSuccess(fbxgeo);
         } else {


hyperlogic:
https://app.asana.com/0/26225263936266/102829204238703/f

hifi-gustavo:
Build console output for commit: 6e872edcaeb057126a885ca4da3ff972360f6265 available at: https://gist.github.com/ff34881ac383d40aaa4d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7402.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7402.dmg

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -72,11 +72,15 @@ void GeometryReader::run() {
                 const bool grabLightmaps = true;
                 const float lightmapLevel = 1.0f;
                 fbxgeo = readFBX(_data, _mapping, _url.path(), grabLightmaps, lightmapLevel);
+                if (fbxgeo->meshes.size() == 0 && fbxgeo->joints.size() == 0) {
+                    // empty fbx geometry, indicates error
+                    throw QString("empty geometry, possibly due to an unsupported FBX version");
```
ZappoMan: wait... do we really want to throw an exception here? Isn't that essentially the same as a crash later?

Wouldn't it be better to report an error, and then treat the resource as if it was an empty mesh?

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -72,11 +72,15 @@ void GeometryReader::run() {
...
+                }
             } else if (_url.path().toLower().endsWith(".obj")) {
                 fbxgeo = OBJReader().readOBJ(_data, _mapping, _url);
             } else {
                 QString errorStr("unsupported format");
-                emit onError(NetworkGeometry::ModelParseError, errorStr);
+                throw errorStr;
```
ZappoMan: same here.. it seems like we should report the error, and then fail gracefully.

Do we have some outer code catching this exception and failing more gracefully?

ZappoMan:
disregard my other comments, I see now that we do properly catch the exception, report the error to the user and fail gracefully without crashing.

ZappoMan:
tested with "broken-base-skeleton.fbx" - verified crash in production, verified it didn't crash and logged error with this PR... PASS!

PR #7406, opened 2016-03-21
Model: fix for avatars with late loaded transparent textures.

hyperlogic:
This bug was possibly introduced by this commit:

a89a76dc libraries/render-utils/src/Model.cpp (Zach Pomerantz)      2016-03-11 14:25:39

This was causing avatars with transparent textures to not properly get re-added to the scene for Avatars that had the _needsUpdateTransparentTextures flag set to true.

### DIFF
--- a/libraries/render-utils/src/Model.cpp
+++ b/libraries/render-utils/src/Model.cpp
@@ -571,6 +571,7 @@ void Model::removeFromScene(std::shared_ptr<render::Scene> scene, render::Pendin
     }
     _renderItems.clear();
     _renderItemsSet.clear();
+    _meshGroupsKnown = false;
     _readyWhenAdded = false;
 }
 


hyperlogic:
https://app.asana.com/0/26225263936266/102901243102883

hifi-gustavo:
Build console output for commit: b06f027f2ee9bad5e2f59dca9572130e7bc94494 available at: https://gist.github.com/be3432727c6577c87555<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7406.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7406.dmg

OmegaHeron:
Tested Windows 7 - Using avatar from Asana and current production Interface it flickers then disappears (if you don't blink - otherwise it seems to never show).  Using this PR I see 4 cubes attached to an AV rig - I assume that's correct. :)  My normally used AV works as well.  QA Pass (if the AV displayed with test setup really is what's seen below)

hifi-snap-by-omegaheron-on-2016-03-21_11-55-16

PR #7417, opened 2016-03-21
Default background behavior to stars

zzmp:
The codepath to propagate zone settings to the stage (actually rendered zone) only runs if the zone changes. If there is no zone to begin with, then the default is shown.

This changes the default behavior to SKY_DOME, instead of SKY_BOX, so that it will show stars instead of whatever is in the stage's memory. This was manifesting as a white background with (for me) a thin red border.

### DIFF
--- a/libraries/model/src/model/Stage.h
+++ b/libraries/model/src/model/Stage.h
@@ -173,7 +173,7 @@ public:
     const SkyboxPointer& getSkybox() const { valid(); return _skybox; }
 
 protected:
-    BackgroundMode _backgroundMode = SKY_BOX;
+    BackgroundMode _backgroundMode = SKY_DOME;
 
     LightPointer _sunLight;
     mutable SkyboxPointer _skybox;


zzmp:
https://app.asana.com/0/74051501495175/103307223537137

#### Testing

Repro:
- Navigate to a non-domain (a non-running server)
- Shutdown Interface
- Restart Interface *in the non-domain*
- Skybox is SKY_BOX, not SKY_DOME (stars)

1. Ensure the repro is fixed.
2. Smoke test to make sure that zones still work.

hifi-gustavo:
Build console output for commit: 2b92756c6587c93475f07700a13d51e1a293e7db available at: https://gist.github.com/3418924e58e733e53ae1<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7417.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7417.dmg

ZappoMan:
tested - confirmed it works as expected.

PR #7421, opened 2016-03-22
one more crack at fixing the blender

ZappoMan:
Makes the `_modelsRequiringBlends` thread safe, in case multiple threads attempt to call `noteRequiresBlend()` while blenders are running. Also implements the proper version of iterating and removing items from `std::set<>`

### DIFF
--- a/libraries/render-utils/src/Model.cpp
+++ b/libraries/render-utils/src/Model.cpp
@@ -1277,20 +1277,26 @@ void ModelBlender::noteRequiresBlend(ModelPointer model) {
         }
         return;
     }
-    _modelsRequiringBlends.insert(model);
+
+    {
+        Lock lock(_mutex);
+        _modelsRequiringBlends.insert(model);
+    }
 }
 
 void ModelBlender::setBlendedVertices(const QPointer<Model>& model, int blendNumber,
         const QWeakPointer<NetworkGeometry>& geometry, const QVector<glm::vec3>& vertices, const QVector<glm::vec3>& normals) {
+
     if (!model.isNull()) {
         model->setBlendedVertices(blendNumber, geometry, vertices, normals);
     }
     _pendingBlenders--;
-    while (!_modelsRequiringBlends.empty()) {
-        auto firstItem = _modelsRequiringBlends.begin();
-        if (firstItem != _modelsRequiringBlends.end()) {
-            _modelsRequiringBlends.erase(firstItem);
-            ModelPointer nextModel = firstItem->lock();
+    {
+        Lock lock(_mutex);
+        for (auto i = _modelsRequiringBlends.begin(); i != _modelsRequiringBlends.end();) {
+            auto weakPtr = *i;
+            _modelsRequiringBlends.erase(i++); // remove front of the set
+            ModelPointer nextModel = weakPtr.lock();
             if (nextModel && nextModel->maybeStartBlender()) {
                 _pendingBlenders++;
                 return;
--- a/libraries/render-utils/src/Model.h
+++ b/libraries/render-utils/src/Model.h
@@ -402,11 +402,15 @@ public slots:
         const QVector<glm::vec3>& vertices, const QVector<glm::vec3>& normals);
 
 private:
+    using Mutex = std::mutex;
+    using Lock = std::unique_lock<Mutex>;
+
     ModelBlender();
     virtual ~ModelBlender();
 
     std::set<ModelWeakPointer, std::owner_less<ModelWeakPointer>> _modelsRequiringBlends;
     int _pendingBlenders;
+    Mutex _mutex;
 };
 
 


ZappoMan:
This should fix Seth's linux crash and Tony's debug crash.

https://app.asana.com/0/26320649018758/103400627223272
https://app.asana.com/0/26320649018758/103443039022024

Testing Notes:
1) Run this build, use an avatar with mouth blendshapes (Albert or Will)
2) make sure you have mirror or mini-mirror visible, talk, you should see blendshapes for the mouth
3) Change your avatar, verify it doesn't crash
4) get multiple people to join you in the domain, and make sure you can see their blendshapes properly when they talk.

hifi-gustavo:
Build console output for commit: 513138ca9809cbdef7b263a8ca5c9cb4e733ae15 available at: https://gist.github.com/2707321cf1d1474f72fe<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7421.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7421.dmg

ZappoMan:
test this please

hifi-gustavo:
Build console output for commit: 513138ca9809cbdef7b263a8ca5c9cb4e733ae15 available at: https://gist.github.com/2707321cf1d1474f72fe<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7421.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7421.dmg

judasshuffle:
ok works fine with Albert for me
https://i.gyazo.com/3eef1e3bdc6068fb3b0d00c56bf76bc8.gif
works fine with my avatar 2

ZappoMan:
@judasshuffle - did you do any "multi-user" tests? where you observed someone else speaking and making blendshapes?

libraries/render-utils/src/Model.cpp
```diff
@@ -1277,20 +1277,26 @@ void ModelBlender::noteRequiresBlend(ModelPointer model) {
...
         model->setBlendedVertices(blendNumber, geometry, vertices, normals);
     }
     _pendingBlenders--;
-    while (!_modelsRequiringBlends.empty()) {
-        auto firstItem = _modelsRequiringBlends.begin();
-        if (firstItem != _modelsRequiringBlends.end()) {
-            _modelsRequiringBlends.erase(firstItem);
-            ModelPointer nextModel = firstItem->lock();
+    {
+        Lock lock(_mutex);
```
jherico: This could be a potentially long lock, depending on the size of the set and the cost of the functions within that you call (in particular `maybeStartBlender`).  Additionally, I'm wary of any locked section that calls functions that aren't very obviously trivial, as it can lead to deadlocks.  

From the logic, it looks like you always empty the set, so perhaps you might consider doing this....

```
 std::set<ModelWeakPointer, std::owner_less<ModelWeakPointer>> modelsRequiringBlends;
{ 
    Lock lock(_mutex);
    modelsRequiringBlends.swap(_modelsRequiringBlends);
    for (auto& weakPtr : modelsRequiringBlends) {
       ...
    }
}
```

PhilipRosedale:
Tested as described.

libraries/render-utils/src/Model.h
```diff
@@ -402,11 +402,15 @@ public slots:
         const QVector<glm::vec3>& vertices, const QVector<glm::vec3>& normals);
 
 private:
+    using Mutex = std::mutex;
+    using Lock = std::unique_lock<Mutex>;
+
     ModelBlender();
     virtual ~ModelBlender();
 
     std::set<ModelWeakPointer, std::owner_less<ModelWeakPointer>> _modelsRequiringBlends;
```
jherico: should probably be `std::unordered_set` to get O(1) inserts, instead of O(LogN)

libraries/render-utils/src/Model.cpp
ZappoMan: we do not empty the set... we should be only pulling the first non-null (non-deleted/released) model that can be started... then we bail.

jherico: my mistake.

AndrewMeadows: AFAICT: it clears from the front until: the cleared element is a valid pointer AND it needs to blend (which we assume is true for most elements in the set) at which point it returns.  In short: it only ever pops one valid blend per frame.

libraries/render-utils/src/Model.h
ZappoMan: AFAICT - You can't have an `std::unordered_set<>` of `std::weak_ptr<>` because unordered set requires an `==` operator... `set<>` only requires a `less` operator which there is one for weak_ptr... namely `owner_less`... I would love to have it be an `unordered_set`... if you can point me to the solution I will gladly switch it... that being said, these sets are usually 0-2 in size, and the max I've seen is 8...

PR #7427, opened 2016-03-22
Add protection around ScriptEngine::runInThread being called twice

huffman:


huffman:
To test:

1. Try running and restarting a bunch of scripts. Verify that the scripts load as expected.

huffman:
I want to check one more thing, hold off on testing for now

hifi-gustavo:
Build console output for commit: 1a2f74f43a5013b023a3f7a47a1a28fa7be19672 available at: https://gist.github.com/e0fd95953f593fa801be<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7427.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7427.dmg

huffman:
Asana task: https://app.asana.com/0/26224634468389/102244079680957

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -158,6 +158,13 @@ void ScriptEngine::disconnectNonEssentialSignals() {
 }
 
 void ScriptEngine::runInThread() {
+    Q_ASSERT_X(!_isThreaded, "ScriptEngine::runInThread()", "runInThread should not be called more than once");
+
+    if (_isThreaded) {
```
ZappoMan: sounds like a crazy repro... but if it's what you suspect, we may need to make this member variable atomic to actually prevent a race of two thread starting at the same time

jherico: only if `runInThread` is being called from two different threads.  As far as I know only the main thread can launch a script engine.  if `runInThread` is being called more than once it's probably bad logic in `ScriptEngines.cpp`

ZappoMan: agreed ^^

huffman: re: crazy repro, agreed, although the dmp does seem to point here. I've tried actually reproing a few different ways without luck, so this is an intermediate fix.

OmegaHeron:
Tested Windows 7
General smoke test :+1: 
Loaded up a "ton" of scripts and repeatedly hammered restart all - no issues seen/no crashes. :+1: 
QA Pass

One thing (totally unrelated to this PR) I discovered.  If you just hammer reload scripts button without waiting for reload to finish you can end up with multiple running copies of some scrips.  Both handControllerHrab.js and squeezeHands.js seem especially prone to this.  Maybe something to think about in that.

PR #7428, opened 2016-03-22
Clear GL error before initializing qml to avoid mystery crash

zzmp:
Adds a check before calling oglplus operations in the OffscreenQmlRenderThread to see if there are any outstanding gl errors. If there is one, it will be cleared and logged as:
```
Clearing outstanding GL error to set up QML FBO: ERROR_TYPE
```

This should remove setupFbo crashes, but continue to log when the errors occur.

### DIFF
--- a/libraries/gl/src/gl/GLWindow.cpp
+++ b/libraries/gl/src/gl/GLWindow.cpp
@@ -39,6 +39,14 @@ GLWindow::~GLWindow() {
 bool GLWindow::makeCurrent() {
     bool makeCurrentResult = _context->makeCurrent(this);
     Q_ASSERT(makeCurrentResult);
+    
+    std::call_once(_reportOnce, []{
+        qDebug() << "GL Version: " << QString((const char*) glGetString(GL_VERSION));
+        qDebug() << "GL Shader Language Version: " << QString((const char*) glGetString(GL_SHADING_LANGUAGE_VERSION));
+        qDebug() << "GL Vendor: " << QString((const char*) glGetString(GL_VENDOR));
+        qDebug() << "GL Renderer: " << QString((const char*) glGetString(GL_RENDERER));
+    });
+    
     Q_ASSERT(_context == QOpenGLContext::currentContext());
     
     return makeCurrentResult;
--- a/libraries/gl/src/gl/GLWindow.cpp
+++ b/libraries/gl/src/gl/GLWindow.cpp
@@ -39,14 +39,6 @@ GLWindow::~GLWindow() {
 bool GLWindow::makeCurrent() {
     bool makeCurrentResult = _context->makeCurrent(this);
     Q_ASSERT(makeCurrentResult);
-    
-    std::call_once(_reportOnce, []{
-        qDebug() << "GL Version: " << QString((const char*) glGetString(GL_VERSION));
-        qDebug() << "GL Shader Language Version: " << QString((const char*) glGetString(GL_SHADING_LANGUAGE_VERSION));
-        qDebug() << "GL Vendor: " << QString((const char*) glGetString(GL_VENDOR));
-        qDebug() << "GL Renderer: " << QString((const char*) glGetString(GL_RENDERER));
-    });
-    
     Q_ASSERT(_context == QOpenGLContext::currentContext());
     
     return makeCurrentResult;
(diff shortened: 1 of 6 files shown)


zzmp:
https://app.asana.com/0/74051501495175/101905459350859

#### Testing

Play with qml surfaces. This bug manifested at startup, so restart Interface a lot.
This bug has not been reproducible, and was only observed on 2-3 machines. Treat this more like a smoke test.

zzmp:
@samcake or @jherico, is it dangerous to clear the context like this? We do it after glewInit already (presumably because of `glewExperimental = true` causing a GL_INVALID_ENUM).

hifi-gustavo:
Build console output for commit: 362e288be8ef18d466e13db1dbea4bb6a00a7b8b available at: https://gist.github.com/185a7cde2d7679bf0733<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7428.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7428.dmg

OmegaHeron:
Tested Windows 7 - all basics look good :+1: 

Ran/closed/repeated > 10 times with no issues.  Haven't searched all logs yet, but, so far, not seeing any hits for new error.  Smoke test = QA Pass

libraries/gl/src/gl/OffscreenGLCanvas.cpp
```diff
@@ -57,13 +53,6 @@ bool OffscreenGLCanvas::makeCurrent() {
     bool result = _context->makeCurrent(_offscreenSurface);
     Q_ASSERT(result);
     
-    std::call_once(_reportOnce, []{
```
ZappoMan: let's not remove this, let's keep this reporting in here

libraries/gl/src/gl/OffscreenGLCanvas.cpp
```diff
@@ -44,10 +44,6 @@ bool OffscreenGLCanvas::create(QOpenGLContext* sharedContext) {
         return true;
     }
 
-    qWarning() << "GL Version: " << QString((const char*) glGetString(GL_VERSION));
```
ZappoMan: let's not remove this, let's keep this reporting in here

jherico: Actually, this block should go, because a) it's using warning, not debug, and b) it's after the create, but not after a makeCurrent.  Create doesn't automatically make the created context current.

libraries/gl/src/gl/OffscreenGLCanvas.cpp
jherico: this reporting should stay.

libraries/gl/src/gl/OffscreenQmlSurface.cpp
```diff
@@ -187,6 +187,15 @@ bool OffscreenQmlRenderThread::event(QEvent *e) {
 void OffscreenQmlRenderThread::setupFbo() {
     using namespace oglplus;
     _textures.setSize(_size);
+
+    // Before making any ogl calls, clear any outstanding errors
+    // FIXME: Something upstream is polluting the context with a GL_INVALID_ENUM,
+    //        likely from glewExperimental = true
+    GLenum err = glGetError();
+    if (err != GLEW_OK) {
```
jherico: you should be testing against `GL_NO_ERROR`.

libraries/gl/src/gl/GLWindow.cpp
```diff
@@ -39,14 +39,6 @@ GLWindow::~GLWindow() {
 bool GLWindow::makeCurrent() {
     bool makeCurrentResult = _context->makeCurrent(this);
     Q_ASSERT(makeCurrentResult);
-    
-    std::call_once(_reportOnce, []{
```
jherico: this reporting should stay.

hifi-gustavo:
Build console output for commit: e28a876629636c74dbfcd8d27444847146d3ef3a available at: https://gist.github.com/185a7cde2d7679bf0733<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7428.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7428.dmg

PR #7419, opened 2016-03-21
Prevent deadlocks in AvatarUpdate

hyperlogic:
This has two parts:

1. Remove AvatarUpdate and all of it's internal locking.  This was left over from earlier avatar threading experiments, and is now unnecessary.
2. Prevent Application::idle from executing while still within Application::paintGL(), this can happen in rare cases where a Win32 dialog box is opened while inside of paintGL, apparently the nvidia driver can sometimes do this.

PR #7433, opened 2016-03-23
Avoid triggering deadlock on long local qml component instantion

zzmp:
- Expose DeadlockWatchdogThread::updateHeartbeat as a public slot on qApp.
- Make qml component instantiation `PreferSynchronous` explicitly.
- Update deadlock heartbeat before qml component instantiation to avoid "deadlock" trigger on long load times.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -239,11 +239,7 @@ class DeadlockWatchdogThread : public QThread {
 public:
     static const unsigned long HEARTBEAT_CHECK_INTERVAL_SECS = 1;
     static const unsigned long HEARTBEAT_UPDATE_INTERVAL_SECS = 1;
-#ifdef DEBUG
-    static const unsigned long MAX_HEARTBEAT_AGE_USECS = 600 * USECS_PER_SECOND;
-#else
     static const unsigned long MAX_HEARTBEAT_AGE_USECS = 10 * USECS_PER_SECOND;
-#endif
 
     // Set the heartbeat on launch
     DeadlockWatchdogThread() {
@@ -511,8 +507,8 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
     auto nodeList = DependencyManager::get<NodeList>();
 
     // Set up a watchdog thread to intentionally crash the application on deadlocks
-    auto deadlockWatchdog = new DeadlockWatchdogThread();
-    deadlockWatchdog->start();
+    _deadlockWatchdogThread = new DeadlockWatchdogThread();
+    _deadlockWatchdogThread->start();
 
     qCDebug(interfaceapp) << "[VERSION] Build sequence:" << qPrintable(applicationVersion());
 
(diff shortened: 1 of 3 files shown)


zzmp:
#### Intent

QML components sometimes triggered the deadlock timer, causing a crash logged as a deadlock.
This resets the timer just before component load/instantiation to give them about 10 seconds to instantiate.

#### Testing

General smoke test.

zzmp:
FYI @ZappoMan there is a crash here if someone tries to load a monstrously large qml component (as @jherico mentioned), but I don't think we care about it...yet.

interface/src/Application.cpp
```diff
@@ -239,11 +239,7 @@ class DeadlockWatchdogThread : public QThread {
 public:
     static const unsigned long HEARTBEAT_CHECK_INTERVAL_SECS = 1;
     static const unsigned long HEARTBEAT_UPDATE_INTERVAL_SECS = 1;
-#ifdef DEBUG
```
jherico: having a long debug heardbeat is probably still a good idea, since otherwise hitting a breakpoint can cause the watchdog to go off if you're looking at the code for more than a few seconds.

hifi-gustavo:
Build console output for commit: 745438148a760b10260d832dcfbeb99f5851c112 available at: https://gist.github.com/98aa8df8d7c987ee10ae<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7433.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7433.dmg

howard-stearns:
Honest question: Have we had any any actionable results from the deadlock watchdog?

I think the watchdog was a **great and noble** idea, but I want to make sure that we're not chasing our tails with "one more thing and then it will be useful." Of course, we don't want to give up just before it does become useful, or after it already has (and I've just missed it).

A searching the repo with
   is:pr deadlock in:title created:>2016-03-09 
suggests that there have been three prs to fix the watchdog, but only one pr that was an actual deadlock fix (and that pr #7419 doesn't cite a ticket or crash dump).

interface/src/Application.cpp
ZappoMan: I agree... let's not remove the extra long debug deadlock.

ZappoMan:
@howard-stearns the deadlock thread ABSOLUTELY has produced real actionable results. It's found at least half a dozen legit deadlocks.

interface/src/Application.cpp
zzmp: I removed this because someone completely disabled the deadlock trigger with a `#ifdef NDEBUG` on line 275.

ZappoMan: duh! Ok.

OmegaHeron:
Tested Windows 7 - General smoke test - QA Pass

PR #7445, opened 2016-03-23
fix memory leak in ShapeManager

AndrewMeadows:
- fix memory leak in ShapeManager

### DIFF
--- a/libraries/physics/src/ShapeFactory.cpp
+++ b/libraries/physics/src/ShapeFactory.cpp
@@ -120,3 +120,21 @@ btCollisionShape* ShapeFactory::createShapeFromInfo(const ShapeInfo& info) {
     }
     return shape;
 }
+
+void ShapeFactory::deleteShape(btCollisionShape* shape) {
+    assert(shape);
+    if (shape->getShapeType() == (int)COMPOUND_SHAPE_PROXYTYPE) {
+        btCompoundShape* compoundShape = static_cast<btCompoundShape*>(shape);
+        const int numChildShapes = compoundShape->getNumChildShapes();
+        for (int i = 0; i < numChildShapes; i ++) {
+            btCollisionShape* childShape = compoundShape->getChildShape(i);
+            if (childShape->getShapeType() == (int)COMPOUND_SHAPE_PROXYTYPE) {
+                // recurse
+                ShapeFactory::deleteShape(childShape);
+            } else {
+                delete childShape;
+            }
+        }
+    }
+    delete shape;
+}
--- a/libraries/physics/src/ShapeManager.cpp
+++ b/libraries/physics/src/ShapeManager.cpp
@@ -23,7 +23,7 @@ ShapeManager::~ShapeManager() {
     int numShapes = _shapeMap.size();
     for (int i = 0; i < numShapes; ++i) {
         ShapeReference* shapeRef = _shapeMap.getAtIndex(i);
-        delete shapeRef->shape;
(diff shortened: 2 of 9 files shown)


AndrewMeadows:
https://app.asana.com/0/26623005675174/103442220436762

@sethalves noticed that there was a memory leak in the ShapeManager.  The problem was that the child shapes of "compound shapes" are not yet in the ShapeManager, so we need to delete the children whenever we delete their parent.  Models that use collision hulls for their shape, and PolyVox entities use the "compound shape" feature.

AndrewMeadows:
Assigning to @sethalves to test and review: he knows the testplan which goes something like this:

(1) build on linux with this environment variable defined: export HIFI_MEMORY_DEBUGGING=1
(2) delete all controller plugins
(3) run the interface and capture all logs (including stderr)
(4) navigate to a domain with lots of compound objects
(5) logout and examine the tail of the logs for memory leak information

hifi-gustavo:
Build console output for commit: ecfe198e35fad5b07d09eba40ede017b7f1eb65c available at: https://gist.github.com/1b5e7d43cb5ecfbe7df5<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7445.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7445.dmg

libraries/physics/src/ShapeManager.cpp
```diff
@@ -58,14 +59,14 @@ btCollisionShape* ShapeManager::getShape(const ShapeInfo& info) {
...
-bool ShapeManager::releaseShape(const DoubleHashKey& key) {
+bool ShapeManager::releaseShapeByKey(const DoubleHashKey& key) {
     ShapeReference* shapeRef = _shapeMap.find(key);
     if (shapeRef) {
         if (shapeRef->refCount > 0) {
             shapeRef->refCount--;
             if (shapeRef->refCount == 0) {
                 _pendingGarbage.push_back(key);
-                const int MAX_GARBAGE_CAPACITY = 127;
+                const int MAX_GARBAGE_CAPACITY = 255;
```
AndrewMeadows: I'm bumping this up in a half-hearted attempt to reduce immediate garbage collection when deleting entities using compound shapes.  However, I noticed that many of our compound shapes are using 1000+ compound parts.  None of the child shapes are shared atm, unless their parent shape itself is shared.

libraries/physics/src/ShapeFactory.cpp
```diff
@@ -120,3 +120,21 @@ btCollisionShape* ShapeFactory::createShapeFromInfo(const ShapeInfo& info) {
...
+            if (childShape->getShapeType() == (int)COMPOUND_SHAPE_PROXYTYPE) {
+                // recurse
+                ShapeFactory::deleteShape(childShape);
+            } else {
+                delete childShape;
+            }
+        }
+    }
+    delete shape;
+}
```
AndrewMeadows: This is the real fix: recursive delete of child shapes.  Yes, I noticed at least one compound shape heirarchy that was two deep:  CompoundShape with one child: a CompoundShape.  This particular inefficiency should eventually be fixed but eventually we'll probably have CompoundShapes with many CompoundShape children (for articulated objects with complex parts).

sethalves:
I've tested this, and it looks good.

PR #7447, opened 2016-03-23
fix animated motor proteins

ZappoMan:
Fix animated entities that are being externally updated (e.g. from an AC script)

### DIFF
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.h
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.h
@@ -38,12 +38,6 @@ public:
                                                 EntityPropertyFlags& propertyFlags, bool overwriteLocalData,
                                                 bool& somethingChanged) override;
 
-    virtual void somethingChangedNotification() override {
-        // FIX ME: this is overly aggressive. We only really need to simulate() if something about
-        // the world space transform has changed and/or if some animation is occurring.
-        _needsInitialSimulation = true;
-    }
-
     virtual bool readyToAddToScene(RenderArgs* renderArgs = nullptr);
     virtual bool addToScene(EntityItemPointer self, std::shared_ptr<render::Scene> scene, render::PendingChanges& pendingChanges) override;
     virtual void removeFromScene(EntityItemPointer self, std::shared_ptr<render::Scene> scene, render::PendingChanges& pendingChanges) override;


ZappoMan:
This fixes this issue: https://app.asana.com/0/26320649018758/94449384699210

Testing notes:
- Basic Smoke Test.
- Make sure to go to cellscience and click on "inside the cell" -
- Observe that the motor proteins will properly animate even in desktop mode if not moving your camera (in production, their legs don't move as they move along their path).
- Test other cases of animated entities... (come to hell and verify that the dragon is flying)
- Multi-user tests: Make sure a second observer can see someone move and/or rotate an entity.

libraries/entities-renderer/src/RenderableModelEntityItem.h
```diff
@@ -38,12 +38,6 @@ class RenderableModelEntityItem : public ModelEntityItem {
                                                 EntityPropertyFlags& propertyFlags, bool overwriteLocalData,
                                                 bool& somethingChanged) override;
 
-    virtual void somethingChangedNotification() override {
```
ZappoMan: Here's the fix -- basically, this code was designed to detect any property change that might require the "Model.cpp" model to get it's bounds/rotation/position updated and simulated for the first time... but as the FIXME comment says, this is way too aggressive. The correct behavior is to only do simulation in cases where the bounds have changed or don't match the model... 

It turns out there is already code in the update and render methods to check the bounds and update accordingly, so this notification handler was superfluous.

But beyond being superfluous, it actually also caused the getAnimationFrame() to bail early, because if it really is a case of an initial simulation being needed the animation frames can't yet be calculated. So as a result, we'd bail early and never update the correct animation frame.

So basically, this old code is preventing the correct animation, and not providing extra support for changed entity properties that actually require model updates.

hifi-gustavo:
Build console output for commit: 3fa3b40f433251a1ae4fc758f2b58e492b45f364 available at: https://gist.github.com/1c4304bb3d8baeb83e29<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7447.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7447.dmg

ZappoMan:
test this please

hifi-gustavo:
Build console output for commit: 3fa3b40f433251a1ae4fc758f2b58e492b45f364 available at: https://gist.github.com/1c4304bb3d8baeb83e29<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7447.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7447.dmg

PR #7448, opened 2016-03-23
Added Refresh Buttons to entityProperties

Menithal:
This PR adds a simple "reload content" button at the end of each url entityProperty in the Edit windows.

It appends / updates to the url a parameter "HFTime" that is the current timestamp (in milliseconds) to request the latest version of an asset. If there is nothing on the field, pressing the button should do nothing.

-----

### Note this may be difficult to test via rawgit. I suggest downloading the zip and then using the example folder instead as a whole due to so many dependencies in the edit.js script

**Test Expectations:**
- if the url is empty, pressing the button should not do anything
- Pressing the button should always reload the property asset.
- If HFTime already exist, it should be removed, and a new one replacing it should appear.
- Pressing the button should not break the entity url to be invalid URI spec.

**Test cases:**
- Files/Atp? (this is a question mark)
- Plain url
- Url with multiple parameters
- Url parameter with existing HFTime parameter: HFTime Parameter is removed, and updated at the end of the url.

**Testing:**

1. Use edit.js
2. Select an entity with a url property. 
3. Update the url content.
4. Press the refresh button.
5. Updated entity should now load without having to touch the url. This url may now be used on other content using the same asset

### DIFF
--- a/examples/html/entityProperties.html
+++ b/examples/html/entityProperties.html
@@ -451,36 +451,36 @@
 
                 var elPreviewCameraButton = document.getElementById("preview-camera-button");
     
-		var urlUpdaters = document.getElementsByClassName("update-url-version");
-		var PARAM_REGEXP = /(?:\?)(\S+)/; // Check if this has any parameters.
-		var TIMESTAMP_REGEXP = /(&?HFTime=\d+)/;
-	
-		var refreshEvent = function(event){
-	  		var urlElement = event.target.parentElement.getElementsByClassName("url")[0];
-  			var content = urlElement.value;
-  			var date = new Date();
-  			var timeStamp = date.getTime();
+                var urlUpdaters = document.getElementsByClassName("update-url-version");
+                var PARAM_REGEXP = /(?:\?)(\S+)/; // Check if this has any parameters.
+                var TIMESTAMP_REGEXP = /(&?HFTime=\d+)/;
 
-  			if(content.length > 0){
-      				if(PARAM_REGEXP.test(content)){
-        			// Has params, so lets remove existing definition and append again.
-        				content = content.replace(TIMESTAMP_REGEXP,"") + "&";		
-      				}else{
-        				content += "?";
-      				}
-				content = content.replace("?&","?");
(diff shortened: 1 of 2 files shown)


hifi-gustavo:

cozza13:
ok to test

hifi-gustavo:
Build console output for commit: 12ecd592ab136506c7a1514570b87284d0aa9148 available at: https://gist.github.com/74ba7e6133b6b604a45e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7448.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7448.dmg

judasshuffle:
Works for me, I will miss typing ?v1234354, I will get over it, nice improvement

examples/html/entityProperties.html
```diff
@@ -451,6 +451,37 @@
...
+    			evt.initEvent("change", true, true ); 
+    			urlElement.dispatchEvent(evt);
+
+		};
+
+		for(var index = 0; index < urlUpdaters.length; index++){
+  			var urlUpdater = urlUpdaters[index];
+  			urlUpdater.addEventListener("click", refreshEvent);
+		}
+
```
AndrewMeadows: This appears to be indented using tabs instead of four spaces.  I assume we're enforcing space-only indentation on our HTML files.

Menithal: Will fix this to meet the requirement.

Menithal:
@AndrewMeadows  Updated the code styling

hifi-gustavo:
Build console output for commit: 24f2bdac7795ac5c3cb3fc3235f7821f031ba1be available at: https://gist.github.com/74ba7e6133b6b604a45e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7448.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7448.dmg

PR #7454, opened 2016-03-24
Fix async issues related to JS <-> App <-> QML <-> HTML communication

jherico:
Addresses a potential sources of crashing due to the asynchronous passing of requests between JS scripts and the QML objects.

Also re-implements the mechanism used to communicate between Interface JS and HTML JS

To test this you need to load scripts from `http://rawgit.com/jherico/hifi/qml_crashes/examples/defaultScripts.js`

One issue I saw in production was that repeatedly reloading all scripts would eventually crash.  With this build I'm unable to reproduce this.  Note that you should wait for the reload to finish before reloading again.  Simply spamming the reload button seems to have some issues.

PR #7422, opened 2016-03-22
Fix occasional crash on creating overlay tool windows

jherico:
There appears to be an issue with ownership and garbage collection when dynamically creating and destroying tabs in a QML `TabView`.

PR #7458, opened 2016-03-24
Guard destruction of half-initialized QmlOverlay.

howard-stearns:


howard-stearns:
Possible fix for https://app.asana.com/0/32622044445063/104506076099221

hifi-gustavo:
Build console output for commit: f7f2b14f43c1e6c2bf182be3da08500d7aaf84c5 available at: https://gist.github.com/01b7b1503e8eb6e5b196<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7458.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7458.dmg

howard-stearns:
Test notes: Do various quick shutdown things right after startup, with windows and (script) panels open or not open.

interface/src/ui/overlays/QmlOverlay.cpp
```diff
@@ -51,6 +51,9 @@ void QmlOverlay::buildQmlElement(const QUrl& url) {
 }
 
 QmlOverlay::~QmlOverlay() {
+    if (!_qmlElement) { // e.g., if shutdown while still building.
```
jherico: I don't understand how this will affect anything.  `_qmlElement.reset()` should have no effect if the element is null.

howard-stearns: Even with the deleter established when constructing the shared_ptr? (About 14 lines above. The body of the deleter is fine, but I'm worried about the existence/storage of the deleter lambda.)

There are crashes on the _qmlElement.reset() line. I'm hoping that's because of the above. Otherwise....wtf?

cozza13:
Been doing those tests. I have not specifically crashed, yet not really able to repro this on the production build. The system sometimes is not shutting down, however that is also happening on release and unrelated to this PR. I will give it a QA pass.

interface/src/ui/overlays/QmlOverlay.cpp
Atlante45: I just checked out that code, _qmlElement is owned by JavaScript and has a parent.
```
// If we already have a root, just set a couple of flags and the ancestry
    if (_rootItem) {
        // Allow child windows to be destroyed from JS
        QQmlEngine::setObjectOwnership(newObject, QQmlEngine::JavaScriptOwnership);
        newObject->setParent(_rootItem);
        if (newItem) {
            newItem->setParentItem(_rootItem);
        }
        return newObject;
    }
```
Wrapping it in an `std::shared_ptr` seems dangerous, it's probably getting deleted under our foot.

howard-stearns:
Given the continuing flux of related code (https://github.com/highfidelity/hifi/pull/7473, https://github.com/highfidelity/hifi/pull/7454, https://github.com/highfidelity/hifi/pull/7312, https://github.com/highfidelity/hifi/pull/7422), I think I'd like to withdraw this until the dust settles.

PR #7462, opened 2016-03-24
Lock cache's unusedResources access

zzmp:
When resources begin to fall out of scope and purge from the cache, the underlying `QMap` can have access violations from mutations across threads. This makes that member threadsafe.


zzmp:
https://app.asana.com/0/26225263936266/104719244308523

#### Testing

Make sure things still load. General smoke test.

Atlante45:
Oh it wasn't static then.
I guess the resources can be owned by any thread. and the shared ptr's door will run on whatever thread it gets destructed on.
That was the issue.

zzmp:
I have this being called multiple times on one thread - potential deadlock. Will update when I know more.

zzmp:
The issue was that I am nesting resources, so when one is purged from the cache, the AllReferencesCleared is called on the next one.

hifi-gustavo:
Build console output for commit: 7274a693c78782850293e7428ea3b75efaa108a4 available at: https://gist.github.com/8cfb9d3e6315537f3bc6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7462.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7462.dmg

zzmp:
I think all entity references are dropped from the main thread, so this PR is probably not necessary.

PR #7467, opened 2016-03-24
Drop resource data after load

zzmp:
Currently, everything using a subclass of `ResourceCache` holds onto the network data with the processed `Resource`.

This removes that `_data` member, which was never used, to reduce memory consumption.

PR #7471, opened 2016-03-25
Revert "Drop resource data after load"

ZappoMan:
Reverts highfidelity/hifi#7467

### DIFF
--- a/libraries/networking/src/ResourceCache.cpp
+++ b/libraries/networking/src/ResourceCache.cpp
@@ -423,12 +423,12 @@ void Resource::handleReplyFinished() {
     
     auto result = _request->getResult();
     if (result == ResourceRequest::Success) {
+        _data = _request->getData();
         auto extraInfo = _url == _activeUrl ? "" : QString(", %1").arg(_activeUrl.toDisplayString());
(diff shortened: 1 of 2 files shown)


Atlante45:
Did that break something?

ZappoMan:
Test says it's crashing on them... and @jherico - expressed concern that the signal was potentially sending bad data....

```
yeah, ok, that PR is unsafe....  the problem is specifically that the signal is using a const reference.  That means that if the signal handler is on a different thread than the emitter, then it will have a const reference to an object that might have been destroyed.
new messages

if you change the signal from `void loaded(const QByteArray& request);` to `void loaded(const QByteArray request);` it should be safe, because the QByteArray will be copied into the signal event and will travel with it, even if it gets handled later on another thread.  Because Qt's containers use shared backing memory, it shouldn't actually cause a duplicate of the memory to be a allocated
```

hifi-gustavo:
Build console output for commit: cd5beabf10cc8479637c0779f0587329b697b8cf available at: https://gist.github.com/a3cdf228118c88d31956<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7471.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7471.dmg

PR #7472, opened 2016-03-25
expire mortal entities correctly

AndrewMeadows:
- fix bug where sometimes entities with finite lifetimes did not appear to be deleted

### DIFF
--- a/libraries/entities/src/EntitySimulation.cpp
+++ b/libraries/entities/src/EntitySimulation.cpp
@@ -101,11 +101,14 @@ void EntitySimulation::expireMortalEntities(const quint64& now) {
                 prepareEntityForDelete(entity);
             } else {
                 if (expiry < _nextExpiry) {
-                    // remeber the smallest _nextExpiry so we know when to start the next search
+                    // remember the smallest _nextExpiry so we know when to start the next search
                     _nextExpiry = expiry;
                 }
                 ++itemItr;
             }
+        }
+        if (_mortalEntities.size() < 1) {
+            _nextExpiry = -1;
         }
     }
 }
--- a/libraries/physics/src/PhysicalEntitySimulation.cpp
+++ b/libraries/physics/src/PhysicalEntitySimulation.cpp
@@ -57,16 +57,18 @@ void PhysicalEntitySimulation::addEntityInternal(EntityItemPointer entity) {
 }
 
 void PhysicalEntitySimulation::removeEntityInternal(EntityItemPointer entity) {
-    EntitySimulation::removeEntityInternal(entity);
-    QMutexLocker lock(&_mutex);
-    _entitiesToAddToPhysics.remove(entity);
+    if (entity->isSimulated()) {
+        EntitySimulation::removeEntityInternal(entity);
+        QMutexLocker lock(&_mutex);
+        _entitiesToAddToPhysics.remove(entity);
 
-    EntityMotionState* motionState = static_cast<EntityMotionState*>(entity->getPhysicsInfo());
-    if (motionState) {
-        _outgoingChanges.remove(motionState);
-        _entitiesToRemoveFromPhysics.insert(entity);
-    } else {
-        _entitiesToDelete.insert(entity);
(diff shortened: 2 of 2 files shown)


libraries/entities/src/EntitySimulation.cpp
```diff
@@ -101,12 +101,15 @@ void EntitySimulation::expireMortalEntities(const quint64& now) {
...
-                    // remeber the smallest _nextExpiry so we know when to start the next search
+                    // remember the smallest _nextExpiry so we know when to start the next search
                     _nextExpiry = expiry;
                 }
                 ++itemItr;
             }
         }
+        if (_mortalEntities.size() < 1) {
+            _nextExpiry = -1;
+        }
```
AndrewMeadows: Minor optimization when _mortalEntities list is empty.

libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -57,16 +57,18 @@ void PhysicalEntitySimulation::addEntityInternal(EntityItemPointer entity) {
...
-    } else {
-        _entitiesToDelete.insert(entity);
+        EntityMotionState* motionState = static_cast<EntityMotionState*>(entity->getPhysicsInfo());
+        if (motionState) {
+            _outgoingChanges.remove(motionState);
+            _entitiesToRemoveFromPhysics.insert(entity);
+        } else {
+            _entitiesToDelete.insert(entity);
+        }
     }
```
AndrewMeadows: Wrapping code in `if (entity->isSimulated())` logic block.

libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -190,7 +192,7 @@ void PhysicalEntitySimulation::deleteObjectsRemovedFromPhysics() {
         entity->setPhysicsInfo(nullptr);
         delete motionState;
 
-        if (entity->isSimulated() && entity->isDead()) {
+        if (entity->isDead()) {
             _entitiesToDelete.insert(entity);
```
AndrewMeadows: These are the real fixes.  The order of operations is not always the same: sometimes the entity will be removed from the EntitySimulation base class (simulated bit goes false) **before** the RigidBody is removed from physics and sometimes it is the other way around -- depends mostly on who notices first that the entity needs to be deleted: the interface or the entity-server.  The check for isSimulated() (the base EntityServer class has it in its own lists) is not necessary here.  The isDead() bit is sufficient.

hifi-gustavo:
Build console output for commit: ac5c055a1ea1fe07ed2aa3664a8ce1ce366335f5 available at: https://gist.github.com/0a639066392945fbfbb3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7472.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7472.dmg

ZappoMan:
fixes: https://app.asana.com/0/26225263936266/104793925165111

AndrewMeadows:
https://app.asana.com/0/26623005675174/104793925165111

Reproducing the problem depends on a sort of race condition (who notices the entity should be deletes first: interface client or entity-server?) but here is the repro recipe that was most reliable for me:

(1) Create an entity
(2) Set its lifetime to 300 seconds
(3) Logout
(4) Login
(5) Set it lifetime to 5 seconds
--> should get deleted and disappear immediately
--> but the bug is that sometimes it doesn't disappear (it does get removed from the physics engine (can't collide against it) and it does get deleted on the server)

libraries/entities/src/EntitySimulation.cpp
Atlante45: Not blocking, but I think this would be better here, since that's what you're actually checking:
```cpp
if (_mortalEntities.isEmpty()) {
    ...
}
```

thoys:
Works for me aswell @PhilipRosedale , was just testing ;)

PR #7474, opened 2016-03-25
Drop network data after processing

zzmp:
See #7467. It was reverted, as it passed a const-ref in a signal which then went out of scope.
This passes a const in the signal instead, so it is copied into the Qt signal/slot framework.

### DIFF
--- a/libraries/animation/src/AnimNodeLoader.cpp
+++ b/libraries/animation/src/AnimNodeLoader.cpp
@@ -611,7 +611,7 @@ AnimNode::Pointer AnimNodeLoader::load(const QByteArray& contents, const QUrl& j
     return loadNode(rootVal.toObject(), jsonUrl);
 }
 
-void AnimNodeLoader::onRequestDone(const QByteArray& data) {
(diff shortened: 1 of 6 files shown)


hifi-gustavo:
Build console output for commit: aad40f5de0d1e152c20ba060f242610e0713cf33 available at: https://gist.github.com/088ec2bec6be99a9a45d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7474.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7474.dmg

zzmp:
General smoke test.

PhilipRosedale:
testing

PhilipRosedale:
looks fast on visiting/re-visiting multiple domains, all seems to be there.

samcake:
Trying To Revert in order to address a bug that started showing with it

samcake:
And i m going to think twice here, i don;t think this is the reason for the bug, looking at bugSplach

PR #7486, opened 2016-03-26
one more crack at fixing false deadlock

ZappoMan:
fixes the false deadlock that can occur if a context switch happened between samples to _heartbeat

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -270,10 +270,10 @@ public:
     void run() override {
         while (!_quit) {
             QThread::sleep(HEARTBEAT_UPDATE_INTERVAL_SECS);
-            auto now = usecTimestampNow();
 
-            // in the unlikely event that now is less than _heartbeat, don't rollover and confuse ourselves
-            auto lastHeartbeatAge = (now > _heartbeat) ? now - _heartbeat : 0; 
+            uint64_t lastHeartbeat = _heartbeat; // sample atomic _heartbeat, because we could context switch away and have it updated on us
+            uint64_t now = usecTimestampNow();
+            auto lastHeartbeatAge = (now > lastHeartbeat) ? now - lastHeartbeat : 0;
             auto sinceLastReport = (now > _lastReport) ? now - _lastReport : 0;
             auto elapsedMovingAverage = _movingAverage.getAverage();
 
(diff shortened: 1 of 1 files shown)


ZappoMan:
Fixes this crash - https://app.asana.com/0/26225263936266/105096043004287

Test plan - 
- smoke test
- philip seems to repro this easily, not sure how (maybe the scripts he's running)... but it has happened to him 4 or 5 times today.

interface/src/Application.cpp
```diff
@@ -270,10 +270,10 @@ class DeadlockWatchdogThread : public QThread {
     void run() override {
         while (!_quit) {
             QThread::sleep(HEARTBEAT_UPDATE_INTERVAL_SECS);
-            auto now = usecTimestampNow();
 
-            // in the unlikely event that now is less than _heartbeat, don't rollover and confuse ourselves
-            auto lastHeartbeatAge = (now > _heartbeat) ? now - _heartbeat : 0;
```
ZappoMan: this line is the problem... notice we're using the _heartbeat atomic twice... well - a context switch can happen between those to accesses and allow a thread to slip in and update it on us... which would cause an overflow.

interface/src/Application.cpp
```diff
@@ -270,10 +270,10 @@ class DeadlockWatchdogThread : public QThread {
     void run() override {
         while (!_quit) {
             QThread::sleep(HEARTBEAT_UPDATE_INTERVAL_SECS);
-            auto now = usecTimestampNow();
 
-            // in the unlikely event that now is less than _heartbeat, don't rollover and confuse ourselves
-            auto lastHeartbeatAge = (now > _heartbeat) ? now - _heartbeat : 0; 
+            uint64_t lastHeartbeat = _heartbeat; // sample atomic _heartbeat, because we could context switch away and have it updated on us
```
ZappoMan: this is the correct way to handle it... sample the atomic once... and use the local value.

hifi-gustavo:
Build console output for commit: f7e6f60e87193af1eeceaa511b967e2ef7a45931 available at: https://gist.github.com/bbc6d17a945a6afce79f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7486.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7486.dmg

PR #7488, opened 2016-03-27
Protect against deleted AssetRequest access

jherico:
A bugsplat dump showed a clear null pointer access in one of the AssetRequest callback methods.  it looks like progress message can sometimes be handled after a finished signal destroys the object.

This PR may be overzealous in terms of protection, as I'm not sure which callbacks can potentially be triggered after object deletion, but better to err on the side of caution, IMO.

### DIFF
--- a/libraries/networking/src/AssetClient.cpp
+++ b/libraries/networking/src/AssetClient.cpp
@@ -202,6 +202,7 @@ AssetUpload* AssetClient::createUpload(const QByteArray& data) {
 
 bool AssetClient::getAsset(const QString& hash, DataOffset start, DataOffset end,
                            ReceivedAssetCallback callback, ProgressCallback progressCallback) {
+    Q_ASSERT(thread() == QThread::currentThread());
     if (hash.length() != SHA256_HASH_HEX_LENGTH) {
         qCWarning(asset_client) << "Invalid hash size";
         return false;
@@ -240,6 +241,7 @@ bool AssetClient::getAsset(const QString& hash, DataOffset start, DataOffset end
 }
 
 bool AssetClient::getAssetInfo(const QString& hash, GetInfoCallback callback) {
+    Q_ASSERT(thread() == QThread::currentThread());
     auto nodeList = DependencyManager::get<NodeList>();
     SharedNodePointer assetServer = nodeList->soloNodeOfType(NodeType::AssetServer);
 
--- a/libraries/networking/src/AssetRequest.cpp
+++ b/libraries/networking/src/AssetRequest.cpp
@@ -59,80 +59,90 @@ void AssetRequest::start() {
     
     _state = WaitingForInfo;
     
+    QPointer<AssetRequest> that(this);
     auto assetClient = DependencyManager::get<AssetClient>();
-    assetClient->getAssetInfo(_hash, [this](bool responseReceived, AssetServerError serverError, AssetInfo info) {
-        _info = info;
+    assetClient->getAssetInfo(_hash, [that](bool responseReceived, AssetServerError serverError, AssetInfo info) {
+        if (that.isNull()) {
+            return;
+        }
+        that->_info = info;
 
         if (!responseReceived) {
-            _error = NetworkError;
+            that->_error = NetworkError;
         } else if (serverError != AssetServerError::NoError) {
             switch(serverError) {
                 case AssetServerError::AssetNotFound:
-                    _error = NotFound;
+                    that->_error = NotFound;
                     break;
                 default:
-                    _error = UnknownError;
+                    that->_error = UnknownError;
                     break;
             }
         }
 
-        if (_error != NoError) {
-            qCWarning(asset_client) << "Got error retrieving asset info for" << _hash;
-            _state = Finished;
-            emit finished(this);
+        if (that->_error != NoError) {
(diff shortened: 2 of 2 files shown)


jherico:
https://www.bugsplatsoftware.com/individualCrash/?id=2307&database=interface_alpha
possibly related to https://www.bugsplatsoftware.com/individualCrash/?id=2179&database=interface_alpha and https://app.asana.com/0/26225263936266/104506076099212

hifi-gustavo:
Build console output for commit: 279c275662974c114a8e43a9dad9cb12be578167 available at: https://gist.github.com/c93fd6e6cf35437efd9c<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7488.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7488.dmg

huffman:
Hmm, if the issue is with progress being emitted after finished then we probably have an issue in ReceivedMessage, PacketReceiver, or Connection. I can take a look, but I probably won't get to it until tomorrow morning. The bug you referenced is assigned to me and similar, probably the same issue.

libraries/networking/src/AssetClient.cpp
```diff
@@ -202,6 +202,7 @@ AssetUpload* AssetClient::createUpload(const QByteArray& data) {
 
 bool AssetClient::getAsset(const QString& hash, DataOffset start, DataOffset end,
                            ReceivedAssetCallback callback, ProgressCallback progressCallback) {
+    Q_ASSERT(thread() == QThread::currentThread());
```
birarda: this should already be guarded - the AssetRequest is on the AssetClient thread and `AssetRequest::start` ensures that the call is on the correct thread

libraries/networking/src/AssetRequest.cpp
```diff
@@ -59,80 +59,90 @@ void AssetRequest::start() {
     
     _state = WaitingForInfo;
     
+    QPointer<AssetRequest> that(this);
```
ZappoMan: I'm personally not really a fan of this `this` vs `that` naming pattern... But I won't block the merge for it... I'd rather the name be something like `pendingRequest` which I think is more clear...

Frankly - I think this is a case where lambdas make the code harder to read and understand...

The problem is the lambda as a callback can come at some very late unknown time... and so it can't trust the state of the objects referenced in it.

I don't have a better proposed solution, which is why I won't block the merge for this issue... but I think this underlying bug is a perfect example of how lamdas can get us in trouble.

libraries/networking/src/AssetClient.cpp
```diff
@@ -240,6 +241,7 @@ bool AssetClient::getAsset(const QString& hash, DataOffset start, DataOffset end
 }
 
 bool AssetClient::getAssetInfo(const QString& hash, GetInfoCallback callback) {
+    Q_ASSERT(thread() == QThread::currentThread());
```
birarda: same comment as above

birarda:
As per @huffman's comment I don't believe that this is the correct fix for the actual issue here. 

If @huffman already has something assigned to him that would cover what is causing this I would prefer this be closed in favour of a fix for the root issue.

ZappoMan:
@birarda - there appears to be nothing in the current code base that would protect that lambda from being called AFTER the asset request has been deleted... can you and/or @huffman identity what code prevents the `this` in this lamda from being previously deleted?

libraries/networking/src/AssetClient.cpp
ZappoMan: if already guarded then the assert is perfectly reasonable to leave in... as it's an assert() that if failing would be a problem.

birarda:
We do not expect an AssetRequest object to be deleted before it is completed or failed and as such do not protect it.

The bug is not the absence of guards in the callback in AssetRequest but the fact that the AssetRequest is being deleted before either one of progress or completion is being called. That is what we need to fix.

ZappoMan:
Fair enough - agreed that fixing the root cause is a better approach. Do you think you will find and fix the root cause in the next couple hours?

This PR, while admittedly not fixing the underlying root issue, will fix at least this one crash case which has been documented as happening on multiple cases. It should introduce any specific side effects not already present in the code. And so, if a fix to the root issue isn't apparent it seems this is a reasonable stop-gap to address the crashing. 

We should of course, keep another bug report open for the root issue. I agree that should ultimately be fixed.

birarda:
I don't think this is even an appropriate stop gap because of the issues I raised with the QPointer I raised above. 

We can prioritize finding this issue if it should take precedence over the other things we are currently chasing.

huffman:
I'm taking a look at this now

birarda:
Can we close this or at minimum mark it DNM while Huffman looks for the actual bug?

ZappoMan:
Good point about QPointer. Let's close this PR... please prioritize this crash as a P1 issue.

jherico:
@huffman I suggest you disable line 49 of AssetRequest.cpp in order to test.  The problem won't manifest if the file is loaded out of cache.

huffman:
@jherico Thanks, will do

PR #7489, opened 2016-03-28
Catch and report GL errors in QML render

zzmp:
This is in response to a crash in setupFbo caused by a presumably unrelated GL error that was not yet cleared. It seems to be the thread that does not catch an oglplus::Error, instead just crashing.

This adds oglplus::Error catching and reporting in instances where the library is used within the OffscreenQmlRenderThread.

### DIFF
--- a/libraries/gl/src/gl/OffscreenQmlSurface.cpp
+++ b/libraries/gl/src/gl/OffscreenQmlSurface.cpp
@@ -190,25 +190,21 @@ void OffscreenQmlRenderThread::setupFbo() {
     using namespace oglplus;
     _textures.setSize(_size);
 
-    // Before making any ogl calls, clear any outstanding errors
-    // FIXME: Something upstream is polluting the context with a GL_INVALID_ENUM,
-    //        likely from glewExperimental = true
-    GLenum err = glGetError();
-    if (err != GL_NO_ERROR) {
-        qDebug() << "Clearing outstanding GL error to set up QML FBO:" << glewGetErrorString(err);
+    try {
+        _depthStencil.reset(new Renderbuffer());
+        Context::Bound(Renderbuffer::Target::Renderbuffer, *_depthStencil)
+            .Storage(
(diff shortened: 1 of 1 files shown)


zzmp:
https://app.asana.com/0/inbox/74051501495174/105153848426914/105153848426916

hifi-gustavo:
Build console output for commit: 77bb8bafbb797c3878483a0986c17721440a3a55 available at: https://gist.github.com/54009123a3e23f88dbd4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7489.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7489.dmg

cozza13:
conducted a general smoke test resized windows (QML and application), ran edit and created boxes. teleported around. QA Pass

libraries/gl/src/gl/OffscreenQmlSurface.cpp
```diff
@@ -190,25 +190,21 @@ void OffscreenQmlRenderThread::setupFbo() {
     using namespace oglplus;
     _textures.setSize(_size);
 
-    // Before making any ogl calls, clear any outstanding errors
-    // FIXME: Something upstream is polluting the context with a GL_INVALID_ENUM,
-    //        likely from glewExperimental = true
-    GLenum err = glGetError();
```
ZappoMan: are you sure you want to remove this code? does it cause any downside?

ZappoMan:
apparently @jherico dev approved this 10 hours ago... if it's good enough for him, it's good enough for me... merging!

libraries/gl/src/gl/OffscreenQmlSurface.cpp
zzmp: That error should be caught in the try block, so it's ok to ignore it here.

PR #7493, opened 2016-03-28
Default to desktop when display plugins fail to load

zzmp:
- Add logging for oculus SDK errors.
- Check plugins when they are activated.
- Add fallback logic for display plugin changes.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -4892,7 +4892,17 @@ void Application::updateDisplayMode() {
 
         // FIXME probably excessive and useless context switching
         _offscreenContext->makeCurrent();
-        newDisplayPlugin->activate();
+
+        // If the new plugin fails to activate, fallback to first item on the list
+        if (!newDisplayPlugin->activate()) {
+            qWarning() << "Failed to activate plugin: " << newDisplayPlugin->getName();
+            newDisplayPlugin = displayPlugins.at(0);
+            qWarning() << "Activating fallback plugin: " << newDisplayPlugin->getName();
+            if (!newDisplayPlugin->activate()) {
+                qFatal("Failed to activate fallback plugin");
+            }
+        }
+
         _offscreenContext->makeCurrent();
         offscreenUi->resize(fromGlm(newDisplayPlugin->getRecommendedUiSize()));
         _offscreenContext->makeCurrent();
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -4886,6 +4886,7 @@ void Application::updateDisplayMode() {
     {
         std::unique_lock<std::mutex> lock(_displayPluginLock);
 
+        auto oldDisplayPlugin = _displayPlugin;
         if (_displayPlugin) {
             _displayPlugin->deactivate();
         }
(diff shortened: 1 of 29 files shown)


zzmp:
@cozza13 I'm posting this up for dev review, but I want to test the code myself before I put up a `QA Ready` label.

interface/src/Application.cpp
```diff
@@ -4892,7 +4892,17 @@ void Application::updateDisplayMode() {
 
         // FIXME probably excessive and useless context switching
         _offscreenContext->makeCurrent();
-        newDisplayPlugin->activate();
+
+        // If the new plugin fails to activate, fallback to first item on the list
+        if (!newDisplayPlugin->activate()) {
```
jherico: if `newDisplayPlugin == displayPlugins.at(0);` then you've entered an infinite loop here.

plugins/hifiSdl2/src/SDL2Manager.h
```diff
@@ -34,9 +34,7 @@ class SDL2Manager : public InputPlugin {
     virtual void init() override;
     virtual void deinit() override;
 
-    /// Called when a plugin is being activated for use.  May be called multiple times.
-    virtual void activate() override;
-    /// Called when a plugin is no longer being used.  May be called multiple times.
+    virtual bool activate() override;
```
jherico: please restore the comments.

zzmp: These are overrides, and the comments are on the base class. Still want them back?

hifi-gustavo:
Build console output for commit: 37a725b06dc8b9bc183609a2077ec54602c939bf available at: https://gist.github.com/8074627bd50839362db9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7493.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7493.dmg

plugins/hifiSdl2/src/SDL2Manager.h
ZappoMan: yes please

interface/src/Application.cpp
```diff
@@ -4892,7 +4892,17 @@ void Application::updateDisplayMode() {
...
+            qWarning() << "Failed to activate plugin: " << newDisplayPlugin->getName();
+            newDisplayPlugin = displayPlugins.at(0);
```
ZappoMan: wouldn't it be better to go back to the LAST active plugin?

zzmp: @ZappoMan My thinking was that desktop was the safest bet, and the most likely to always work.
It also does the least mouse capture, so would make it easiest to switch again.

hifi-gustavo:
Build console output for commit: b94e244844c99a8acf402634a31c0d8a62b115cc available at: https://gist.github.com/8074627bd50839362db9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7493.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7493.dmg

interface/src/Application.cpp
ZappoMan: ok - I am convinced.

hifi-gustavo:
Build console output for commit: 34a77b47a34a2f59d9fff4f0e4a6b78a0c3f4750 available at: https://gist.github.com/8074627bd50839362db9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7493.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7493.dmg

zzmp:
#### Intent

If a display plugin fails to load, the presentation thread will crash the system.
This fixes it so that instead of overall failure and log flooding, we fallback to the last display (if available), and then to the desktop display.

#### Testing

Do something to make Oculus (or OpenVR, I think that's Vive) fail *when switching to it*. This **does not** affect display failures when it is already active, only when it is first switched to. You should test two cases:
- Starting in HMD (with failure), and see it go to Desktop.
- Switching to HMD from something else (like 3D TV), and see it go back to whatever you had.

hifi-gustavo:
Build console output for commit: 7bcef1d3194cbfd7e7dc995b067df3d2db7a1df7 available at: https://gist.github.com/8074627bd50839362db9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7493.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7493.dmg

cozza13:
tested. Crashed the Oculus (powered off while interface was running). The display did not let me switch to Rift. In production is lets you do this and it will freeze the display. QA Pass

PR #7499, opened 2016-03-29
fix an overflow in portable high resolution clock on windows

birarda:
- fixes an overflow in portable high resolution clock on windows

### DIFF
--- a/libraries/shared/src/PortableHighResolutionClock.cpp
+++ b/libraries/shared/src/PortableHighResolutionClock.cpp
@@ -24,7 +24,7 @@ namespace {
 win_high_resolution_clock::time_point win_high_resolution_clock::now() {
     LARGE_INTEGER count;
     QueryPerformanceCounter(&count);
-    return time_point(duration(static_cast<rep>(double(count.QuadPart) * double(period::den) / double(g_Frequency))));
+    return time_point(duration(static_cast<rep>((double) count.QuadPart * (double) period::den / (double)g_Frequency)));
 }
 
 #endif
--- a/libraries/shared/src/PortableHighResolutionClock.cpp
+++ b/libraries/shared/src/PortableHighResolutionClock.cpp
@@ -24,7 +24,7 @@ namespace {
 win_high_resolution_clock::time_point win_high_resolution_clock::now() {
     LARGE_INTEGER count;
     QueryPerformanceCounter(&count);
-    return time_point(duration(count.QuadPart * static_cast<rep>(period::den) / g_Frequency));
+    return time_point(duration(static_cast<rep>(double(count.QuadPart) * static_cast<rep>(period::den) / g_Frequency)));
 }
 
 #endif
--- a/libraries/shared/src/PortableHighResolutionClock.cpp
+++ b/libraries/shared/src/PortableHighResolutionClock.cpp
@@ -24,7 +24,7 @@ namespace {
 win_high_resolution_clock::time_point win_high_resolution_clock::now() {
     LARGE_INTEGER count;
     QueryPerformanceCounter(&count);
-    return time_point(duration(static_cast<rep>(double(count.QuadPart) * static_cast<rep>(period::den) / g_Frequency)));
+    return time_point(duration(static_cast<rep>(double(count.QuadPart) * double(period::den) / double(g_Frequency))));
 }
 
 #endif


birarda:
To test, do not restart your machine over a 7 day stretch. Start Interface before the 7th day. Run Interface into the 7th day.

Comparisons between high resolution clock time points taken before and after the 7th day will be incorrect. In the `udt::SendQueue` this cause the send queue to attempt to sleep for long periods of time.

hifi-gustavo:
Build console output for commit: b134f22cfaf11f59ba77571b876f9d0f2513a9c9 available at: https://gist.github.com/28072b98343503f81614<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7499.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7499.dmg

AndrewMeadows:
test this please

hifi-gustavo:
Build console output for commit: b134f22cfaf11f59ba77571b876f9d0f2513a9c9 available at: https://gist.github.com/28072b98343503f81614<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7499.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7499.dmg

libraries/shared/src/PortableHighResolutionClock.cpp
```diff
@@ -24,7 +24,7 @@ namespace {
 win_high_resolution_clock::time_point win_high_resolution_clock::now() {
     LARGE_INTEGER count;
     QueryPerformanceCounter(&count);
-    return time_point(duration(count.QuadPart * static_cast<rep>(period::den) / g_Frequency));
+    return time_point(duration(static_cast<rep>(double(count.QuadPart) * static_cast<rep>(period::den) / g_Frequency)));
```
AndrewMeadows: There are some implicit casts going on here.  It isn't completely clear to me what is going on but maybe we can unravel it.  First some facts to help us:

rep = long long = 64 bit signed int = int64_t
LARGE_INTEGER::QuadPart is int64_t
g_Frequency is int64_t
period::den = std::nano::den has a type of ??? (I could not figure this out yet)

The operator precedence in C++ for multiplication and division is left-to-right so...

First we have a multiplication between count.QuadPart and period::den.  The explicit cast to double of the left forces an implicit cast of the right to double (after it was explicitly cast to int64_t).

Then we divide the product (a double) by g_Frequency (an int64_t) which forces an implicit cast of the right side to be a double.

Finally we explicitly cast the result (a double) to int64_t and feed that to a function called duration().

birarda: I think you've got it. Do you prefer that I make the implicit casts to double explicit? 

What I basically need is a result that is a double casted to a `rep` for the duration.

`period::den` is an `intmax_t` from `std::ratio` template

AndrewMeadows:
The test for this requires a computer that has been up for 6 days, but not yet 7.  So we either need to (a) find a tester with a fortunate uptime (b) take 7 days to test (assuming the tester can actually coax their Windows machine through 7 days of uptime (they'll probably have to disable automatic updates)).

Do we want to wait 7 days to merge this?

I note that there is no asana task mentioned in the comments of this PR so I can't tell if it is a P1 bug or not.

ZappoMan:
I am pretty sure my windows machine has not been restarted in the last 7
days... it's windows 8.1 -- will that work?

birarda:
It's likely that to run into the specific issue this is fixing you would need to have hit the edge (Interface running before and after the 7 day border)

Those testing notes are unlikely to actually be useful to test.

birarda:
Here's the Asana task for this -> https://app.asana.com/0/11984398816645/105027711866412.

birarda:
I need to correct myself - didn't do the cycle math correctly last night to end up with the 7 day border. Re-did it now to fully explain why this fix is required in lieu of being able to do full QA with a repro case. 

My hardware counter frequency is ~3.9MHz.
With the previous code, we would overflow at 2^63/1e9 or ~9.22e9 cycles. For my hardware counter that's ~0.656 hours.

It's entirely possible that comparing two time_points produced with the overflowing code does provide the correct interval. The issue in the `udt::SendQueue` for a client running before and after the overflow border. The `udt::SendQueue` has a `time_point` it keeps pushing forwards to maintain when it should send the next packet. If the original `time_point` is collected before the overflow border, it unsurprisingly will produce incorrect values when compared with a `time_point` collected before the border.

This can cause the `udt::SendQueue` to want to sleep for extremely long periods of time.

libraries/shared/src/PortableHighResolutionClock.cpp
AndrewMeadows: Coding standard section 3.1.2 says: "Never rely on implicit type conversion." which is why I point out the implicit casts.  Also, since there were explicit casts being made that appeared to be overridden by implicit casts I had to walk myself through it to convince myself that it was doing the right thing.

libraries/shared/src/PortableHighResolutionClock.cpp
```diff
@@ -24,7 +24,7 @@ namespace {
 win_high_resolution_clock::time_point win_high_resolution_clock::now() {
     LARGE_INTEGER count;
     QueryPerformanceCounter(&count);
-    return time_point(duration(count.QuadPart * static_cast<rep>(period::den) / g_Frequency));
+    return time_point(duration(static_cast<rep>(double(count.QuadPart) * double(period::den) / double(g_Frequency))));
```
ZappoMan: doesn't the coding standard say to do casts for PODTs as `(double)period::den / (double)g_Frequency)));`

hifi-gustavo:
Build console output for commit: 37950e99310af0d451023a3ff39f53b3979e6b76 available at: https://gist.github.com/28072b98343503f81614<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7499.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7499.dmg

hifi-gustavo:
Build console output for commit: b9ce427344afeda52326d4488908981b8cb060b6 available at: https://gist.github.com/28072b98343503f81614<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7499.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7499.dmg

ZappoMan:
Any reason why we shouldn't call this QA Ready, and Dev Approved?

although it will be hard to test the exact repro, we can at least smoke test this build to make sure nothing goes sideways.

ZappoMan:
@birarda - so based on your comment, above, do you feel that we can create a formula usable by the tester to determine the amount of time they need to keep their interface running to try to reproduce this test case? if they have a ~4ghz CPU can they assume that running for 1 hour will be sufficient to repro this potential overflow? Is there a way we can produce a table of CPU clock to wall clock for reproing this?

OmegaHeron:
So, @birarda, if I'm understanding this right - Windows needs to be near the rollover tick for time count - Interface needs to be running while it approaches max, hits max and rolls back to 0?  Other than that - basic smoke test shows no issue, yet.

cozza13:
@birarda and @ZappoMan based on @OmegaHeron comments is this good to go.

birarda:
I think a smoke test is sufficient here. I've done extensive testing on the overflow itself and confirmed this is a fix.

PR #7500, opened 2016-03-29
Update model URL on render thread

zzmp:
With the new ModelCache, changing a model's URL will no longer halt the render thread to load the geometry. This means that if the URL is changed during a render pass, the geometry will no longer be valid, despite the renderer already having checked that it can access it (very handwavy, but how it works).

This fixes these access violations, manifest as crashes on refresh of model URLs, by delaying the actual change of the URL on the model underlying an entity until the render thread is free. Because the render loop is event-based, this means the changes will go between render passes, and if a model is not fully loaded before the next pass it will no longer attempt to render it.

- Delays model URL changes to the end of the render pass.
- Cleans RenderableModelEntityItem to only set the URL through getModel.
- Cleans RenderableModelEntityItem::getModel to use the correct connection to call the entity tree.
- Cleans the entity tree model mutators to avoid extra allocations.

### DIFF
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -460,14 +460,17 @@ void EntityTreeRenderer::processEraseMessage(ReceivedMessage& message, const Sha
 
 ModelPointer EntityTreeRenderer::allocateModel(const QString& url, const QString& collisionUrl) {
     ModelPointer model = nullptr;
-    // Make sure we only create and delete models on the thread that owns the EntityTreeRenderer
+
+    // Only create and delete models on the thread that owns the EntityTreeRenderer
     if (QThread::currentThread() != thread()) {
         QMetaObject::invokeMethod(this, "allocateModel", Qt::BlockingQueuedConnection,
                 Q_RETURN_ARG(ModelPointer, model),
-                Q_ARG(const QString&, url));
+                Q_ARG(const QString&, url),
+                Q_ARG(const QString&, collisionUrl));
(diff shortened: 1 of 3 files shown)


zzmp:
https://app.asana.com/0/74051501495175/105559909158473

#### Testing

- Import a model.
- Refresh its URL.

This should crash master, but will work on this PR build.

hifi-gustavo:
Build console output for commit: 29dedd55241e08bc7f6458dbfbed5e7b15f856b6 available at: https://gist.github.com/b4974feee09c6b02966ec98d069fcce0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7500.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7500.dmg

Nex-Pro:
@ZappoMan   you want me to test this when it has built ?

ZappoMan:
@Nex-Pro - yes please! thanks!

Nex-Pro:
Seems to work just fine.  I uploaded several avatar models FBX and FST and an obj.
Tried multiple rapid clicking on the refresh ( up to about 40 times ).
Tried refresh button on a series of models as fast as possible.
No problems seen.  ( perhaps we could block multiple clicks as they seem to make the refresh take longer ) 
FST files refresh really fast of course.
This looks like a PASS to me.

Nex-Pro:
@zzmp   Yes I can confirm that the crashing on the master was instantaneous and consistent.

PR #7505, opened 2016-03-29
don't put ATP scripts in the ScriptCache

birarda:
- fixes an issue where the script cache could return an old version of a newly mapped ATP asset


ZappoMan:
why are we making this change? what specific issue is this fixing?

libraries/script-engine/src/ScriptCache.cpp
```diff
@@ -37,7 +39,9 @@ QString ScriptCache::getScript(const QUrl& unnormalizedURL, ScriptUser* scriptUs
     QString scriptContents;
 
     Lock lock(_containerLock);
-    if (_scriptCache.contains(url) && !reload) {
+
+    // don't cache ATP scripts here, allow the ResourceManager to handle
+    if (_scriptCache.contains(url) && !reload && url.scheme() != ATP_SCHEME) {
```
huffman: The scheme does not include the colon, so this won't compare correctly. IMO we shouldn't include it in `ATP_SCHEME`

birarda:
This is a fix for https://app.asana.com/0/26225263936266/106033407502859 which is one part of an issue that Eric is experiencing with ATP scripts.

To test, upload a entity script to ATP. Add an entity that uses that script. Make a change and overwrite the entity script in ATP. Add a new entity that uses the entity script. You will see `Found script in cache: atp:/entity-script.js` in your log. The version of the entity script running for the latest entity is the previous version, not your overwritten version.

libraries/script-engine/src/ScriptCache.cpp
birarda: Fixed.

ZappoMan:
We need to test the before and after performance of this against content like cell science and playa that may reuse many of the same scripts for multiple entitites.

It's very likely that there will be a significant performance hit.

If the intent is to simply make it easier for developers who are proactively testing changes to their content, then they can use the "Reload All Scripts" which should flush the script cache and allow them to get the new content.

birarda:
Just to be clear, because the ATP requests go through AssetResourceRequest, ATP scripts that have been previously downloaded will still be loaded from the disk cache.

This will incur an additional roundtrip to the Asset Server on load of a script to retrieve the ATP mapping, not the full re-download of the script every time.

PhilipRosedale:
does this change add extra round-trip requests to the entity server on initial connection to a domain?

birarda:
Yes @PhilipRosedale, same as with the asset requests that we tested for a migrated playa.

hifi-gustavo:
Build console output for commit: 18d8aaec96adf0097943c94bffa42ebe03a1dc84 available at: https://gist.github.com/3fd005ff074e5a43a37d422ef138c63e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7505.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7505.dmg

hifi-gustavo:
Build console output for commit: 217b6eb48b8d56ca67518352906c086082d8e787 available at: https://gist.github.com/3fd005ff074e5a43a37d422ef138c63e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7505.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7505.dmg

PhilipRosedale:
If one object has an entity script, and 1000 other objects nearby have the same referenced script, would this change cause all those objects to also do a network RT to check for the latest version?

PhilipRosedale:
In the case where, within a single session, an asset is cached locally, I think the right behavior is to never ask the network again unless 'reload all scripts' is pressed.  But maybe I am confused?  Come and explain if so.

birarda:
Depends on when you load the 1000 nearby objects. 

While initial script is loading the ScriptCache uses one single request for all pending requestors. If you request all 1001 at once you'll do the round trip once and they'll all hear back once you get the mapping and then you download the script or pull it from the ATP disk cache cache.

If the load of the subsequent entity scripts is staggered, you will incur the round-trip each time to be sure that you have the correct version of that asset.

ZappoMan:
1) if there's any argument for this with ATP the same argument exists for HTTP... and so if we sincerely believe we want to optimize against changes in resources DURING a sesion - then the correct fix is to eliminate the memory cache for scripts completely.

2) the reason we have a memory cache for script resources is because they are very often reused across multiple entities, and therefore there is a much greate price paid by doing this cache/network roundtrip.

We should not merge this PR... it's wrong to make this change for ATP only.

hifi-gustavo:
Build console output for commit: a11e16518fc13393384b1896bf858760b6bb5331 available at: https://gist.github.com/3fd005ff074e5a43a37d422ef138c63e<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7505.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7505.dmg

ZappoMan:
After chatting with @ericrius1 about why he asked for this feature...it seems he was using the "reload this script" icon in the running scripts dialog box...

I don't think that button flushes the cache for that individual script... but it certainly could be changed to do that... that would be a much more appropriate fix for this issue.

ZappoMan:
It appears as if the "reload this script" icon in the running scripts already removes the script from the memory cache... 

If it's not reloading from the network then there is some other root issue preventing it from working.

This is the wrong fix for the underlying issue. This PR should be closed.

ZappoMan:
I have confirmed that with the current master build, when you click the "reload this script" icon, for an atp url script, it will clear the contents from the memory cache for that script, and then call the regular resouce manager to load the script content....

```
    auto request = ResourceManager::createResourceRequest(nullptr, url);
    request->setCacheEnabled(!reload);
    connect(request, &ResourceRequest::finished, this, &ScriptCache::scriptDownloaded);
    request->send();
```

so... it seems that if this process is not actually reloading the content from the ATP server than the issue is below the `ResourceManager::createResourceRequest()` layer... the ScriptCache is not involved in this.

ZappoMan:
```
[03/29 16:23:58] [DEBUG] Reload script atp:/foo.js
[03/29 16:23:58] [DEBUG] Delete script from cache: "atp:/foo.js"
[03/29 16:23:58] [DEBUG] stopping script... "atp:/foo.js"
[03/29 16:23:58] [DEBUG] Script Engine shutting down (destructor) for script: "foo.js"
[03/29 16:23:58] [DEBUG] Got mapping for: "/foo.js" => "0e508afbdc47ea94a401548d39d1effc3ac8c11f5aea2ebb121b1618bcd8243c"
[03/29 16:23:58] [DEBUG] "atp:0e508afbdc47ea94a401548d39d1effc3ac8c11f5aea2ebb121b1618bcd8243c" loaded from disk cache.
[03/29 16:23:58] [DEBUG] Done downloading script at: "atp:/foo.js"
[03/29 16:23:58] [DEBUG] Logging activity "loaded_script"
[03/29 16:23:58] [DEBUG] script:print()<< 1459293838520
```

birarda:
After a discussion with @PhilipRosedale we will re-implement the in-memory cache for ATP mappings during the domain session.

PhilipRosedale:
To capture our discussion succinctly (hopefully):  

In the case of a gun firing bullets, or a number of identical trees sitting next to each other, we only want to do one network round trip per session.   

This is because these cases will be extremely common (guns and trees), and developers will also doubtless commonly mark assets NO CACHE and then expect their environments to still work fairly well.  In these cases we can periodically/lazily re-check assets who cache headers have expired.  But we still don't want to be firing off many re-requests for identical assets, or content will often perform poorly.

PR #7506, opened 2016-03-29
Fix ordering of GL backend state reset commands to Field key

zzmp:
Discovered while investigating crashes for GLBackend::resetPipelineState, although I do not think it is related.
This fixes the backends pipeline reset functors to be in sync with the fields that they reset.

### DIFF
--- a/libraries/gpu/src/gpu/GLBackend.h
+++ b/libraries/gpu/src/gpu/GLBackend.h
@@ -158,7 +158,6 @@ public:
         ~GLState();
 
         // The state commands to reset to default,
-        // WARNING depending on the order of the State::Field enum
         static const Commands _resetStateCommands;
 
         friend class GLBackend;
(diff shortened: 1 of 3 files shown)


zzmp:
#### Testing

General smoke test. Ensure things with alpha still work as expected.

hifi-gustavo:
Build console output for commit: 6ae03fe72e0fbb4783a6dc57f4ac2b3bc57fd39f available at: https://gist.github.com/3b5b0b9e261ad0532d7ce6680d9ef6e3<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7506.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7506.dmg

zzmp:
I wonder if having this incorrectly set to TRUE when there is no alpha available would result in the access violations we were seeing. I'm trying to test now...

samcake:
Great Catch, thanks for this

ZappoMan:
Smoked on Mac. I'm pretty sure, this is clean... let's merge.

PR #7507, opened 2016-03-30
fix "deadlock" when setting clock backwards

AndrewMeadows:
- fixed a bug whereby setting the system clock backwards would lock up the interface

### DIFF
--- a/libraries/entities/src/ParticleEffectEntityItem.cpp
+++ b/libraries/entities/src/ParticleEffectEntityItem.cpp
@@ -561,6 +561,12 @@ bool ParticleEffectEntityItem::needsToCallUpdate() const {
 
 void ParticleEffectEntityItem::update(const quint64& now) {
 
+    // we check for 'now' in the past in case users set their clock backward
+    if (now < _lastSimulated) {
+        _lastSimulated = now;
+        return;
+    }
+
     float deltaTime = (float)(now - _lastSimulated) / (float)USECS_PER_SECOND;
     _lastSimulated = now;
 
--- a/libraries/script-engine/src/ScriptEngine.cpp
+++ b/libraries/script-engine/src/ScriptEngine.cpp
@@ -673,11 +673,14 @@ void ScriptEngine::run() {
         }
 
         qint64 now = usecTimestampNow();
-        float deltaTime = (float) (now - lastUpdate) / (float) USECS_PER_SECOND;
 
-        if (!_isFinished) {
-            if (_wantSignals) {
-                emit update(deltaTime);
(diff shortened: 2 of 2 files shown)


libraries/entities/src/ParticleEffectEntityItem.cpp
```diff
@@ -561,6 +561,12 @@ bool ParticleEffectEntityItem::needsToCallUpdate() const {
 
 void ParticleEffectEntityItem::update(const quint64& now) {
 
+    // we check for 'now' in the past in case users set their clock backward
+    if (now < _lastSimulated) {
+        _lastSimulated = now;
+        return;
+    }
+
```
AndrewMeadows: Thisis the real fix.  When the clock was set back the particle system would compute a very large deltaTime and would proceed to "chew" on that in its simulation.  It wasn't a deadlock so much as a blockage of the main thread.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -673,11 +673,14 @@ void ScriptEngine::run() {
...
-        if (!_isFinished) {
-            if (_wantSignals) {
-                emit update(deltaTime);
+        // we check for 'now' in the past in case people set their clock back
+        if (lastUpdate < now) {
+            float deltaTime = (float) (now - lastUpdate) / (float) USECS_PER_SECOND;
+            if (!_isFinished) {
+                if (_wantSignals) {
+                    emit update(deltaTime);
+                }
```
AndrewMeadows: This prevents unreasonable deltaTimes from reaching the update() method on scripts when setting the clock back.

AndrewMeadows:
https://app.asana.com/0/26623005675174/105603974000695

This is the repro recipe:

1) run interface on windows
2) change system clock to go back 5 minutes
3) observe interface apparently not responding
4) wait 30 seconds for deadlock watchdog to force crash

Note, although the above problem is fixed, there is a new distinct problem when trying to logout after setting the clock back: deadlock on shutdown.  It is distinct and should be fixed in a separate PR.

hifi-gustavo:
Build console output for commit: 5ac0640cbec29b4a4979360ddfe6e7e3c416de96 available at: https://gist.github.com/1e0b49dc856c077cacdf770656cc9b83<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7507.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7507.dmg

PR #7508, opened 2016-03-30
Make export/import and copy/paste be parent/child-clean.

howard-stearns:
1. Make SpatiallyNestable::getParentPointer use the entity tree that we're known to be in (exportTree, clipboard, etc., if any) rather than the global. (This involves some type-safety vs library pixie-dust.)

2. New uuids generated when pasting from clipboard are now propogated to parentIDs.

3. Both export and import has to be able to deal with orphaned children (whose parent isn't in the data because it wasn't selected, or whatever). In those cases, we clear the result parentID and globalize the geometry using a new globalizeProperties method.

4. When we do offsets of position (including on orphaned children, whose properties.getPosition() is still local), do so only on global positions (skipping the offset for children).

5. remapIDs was the wrong thing at the wrong time, and the dead code removed. (remapIDs set id/parentID in the EntityItem, but without updating the tree, so later lookups of id or parent would fail. It also didn't globalize the geometry of oprphans. Finally, the remapping needs to be on the clipboard paste path.)

6. DRY up the code to avoid mistakes.

Notes:

a. One of the clipboard export paths was generating svos, with a deprecation warning in the log. Now it generates a json. (This wasn't done to make things "better", but as a consequence of bottlenecking the code-paths.)

b. The offsets on export/import (or copy/paste) are confusing and not what developers want. The Right Value(TM) is application specific, depends on how many are selected and whether parents were included, hard to explain in the various cases, and often require rotation and matching between cut/paste methodologies. However, I've left it in for bug compatability.

c. There may be thread safety issues, but that wasn't the task and I haven't attempted to change anything there.


howard-stearns:
Testing notes:
1. Select an object with edit.js, export it, and import the same file. Import again and make sure there are two copies plus the original. (You can delete the copies now to avoid confusion.)
2. Create a child object with a parent. Select BOTH and export them. Import twice, and make sure that in the original and each copy, moving a parent also moves its child. (You can delete the copies.)
3. Edit the file from #2 to remove the parent object (keeping the child and the reference to a missing parent). On import, observe that:
  - we don't crash
  - the child object comes in (although you may have to hunt around for the position)
  - the log should warn (on import) that the parent is being skipped.
  (You can delete the copy.)
4. Select just the original child, export, and import. Should work, and log should warn (in export only!) that the parent is being skipped. (You can delete the copies.)
5. Select just the original parent, export, and import. Should work (making a copy of just the parent, without the child).
6. Anything you can think to try with the clipboard.

hifi-gustavo:
Build console output for commit: 5381be6902b8fd774ba953e87610d9e098c6d7b9 available at: https://gist.github.com/5ab1d2a5e798f748fa5d246e6d94ed55<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7508.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7508.dmg

ericrius1:
This works for all test cases

ZappoMan:
I think @ericrius1 meant to QA Pass this... since his comment is about testing.

PR #7517, opened 2016-03-30
Check for valid qml gl context creation

zzmp:
This checks that a QML Renderer thread has successfully created a GL context.
If not, it will just quit the renderer thread, which may manifest as a (non-crashing) bug.

### DIFF
--- a/libraries/gl/src/gl/OffscreenGLCanvas.cpp
+++ b/libraries/gl/src/gl/OffscreenGLCanvas.cpp
@@ -38,11 +38,13 @@ bool OffscreenGLCanvas::create(QOpenGLContext* sharedContext) {
         _context->setShareContext(sharedContext);
     }
     _context->setFormat(getDefaultOpenGLSurfaceFormat());
+
     if (_context->create()) {
         _offscreenSurface->setFormat(_context->format());
         _offscreenSurface->create();
-        return true;
+        return _offscreenSurface->isValid();
     }
+    qWarning("Failed to create OffscreenGLCanvas context");
 
     return false;
 }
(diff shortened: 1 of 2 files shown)


zzmp:
@ZappoMan this will mask this crash from our reporting. Do we want to send events to metaverse when this happens, so that we can continue to track this bug, and follow repros?

zzmp:
https://app.asana.com/0/74051501495175/102829204238683

hifi-gustavo:
Build console output for commit: 530cb40528a6fac2fe449bdb3932e443944edf57 available at: https://gist.github.com/55dd66dd613d0dec6c6274b5be3dbdfa<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7517.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7517.dmg

ZappoMan:
I'm not sure if we want to suppress the crash like this... can this state be recovered from?

zzmp:
According to the [qt docs](http://doc.qt.io/qt-5/qopenglcontext.html#isValid),
>The typical way to handle context loss scenarios in applications is to check via this function whenever makeCurrent() fails and returns false. If this function then returns false, recreate the underlying native OpenGL context by calling create(), call makeCurrent() again and then reinitialize all OpenGL resources.

However, we are failing on create, as far as I can tell. So it seems like an unrecoverable state - for that QML Renderer thread.

We have not been able to reproduce this, however, and have no way of knowing if future GL context creation will continue to fail. Some other options I can think of:
- Web surfaces intermittently check their renderers to see if they are functional, and attempt to recreate them if they are not.
- Display some kind of alert, in debug mode, so that if it occurs inside of the office we can investigate it more.
- Report instance to metaverse.highfidelity.io, and monitor for them.
- Continue to attempt to handle this and crash, so that we continue to receive bug reports.

ZappoMan:
Test plan:
- Basic smoke test.
- test locations with web entities (like Forum)

PR #7519, opened 2016-03-30
Fix crash from updating/resetting avatar

zzmp:
In our application event loop:
- Scene updates with pending changes:
  - additions (aka resets)
  - updates
  - removals
- Engine runs (aka render pass)

The model URL, when changed, invalidates its geometry. Its geometry is accessed during updates and engine runs, but updates are queued during engine runs, so only two options are safe:
- Only change model URL during scene removal
- Add a check before performing updates, in the update functor, that the model geometry is still valid

Current code assumed the interlude between the engine run and the scene update was a good place, so it is crashing on update functors. This PR adds a guard in the update functor that was crashing.

### DIFF
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -859,7 +859,11 @@ void Avatar::scaleVectorRelativeToPosition(glm::vec3 &positionToScale) const {
 
 void Avatar::setSkeletonModelURL(const QUrl& skeletonModelURL) {
     AvatarData::setSkeletonModelURL(skeletonModelURL);
-    _skeletonModel->setURL(_skeletonModelURL);
+    if (QThread::currentThread() == thread()) {
+        _skeletonModel->setURL(_skeletonModelURL);
+    } else {
+        QMetaObject::invokeMethod(_skeletonModel.get(), "setURL", Qt::QueuedConnection, Q_ARG(QUrl, _skeletonModelURL));
+    }
 }
 
 // create new model, can return an instance of a SoftAttachmentModel rather then Model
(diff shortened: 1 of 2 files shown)


zzmp:
https://app.asana.com/0/74051501495175/106118303211919

#### Testing

In current build, enable `Developer > Advanced Menus`, and trigger `Edit > Reload Content (Clears all caches)`. Do this enough (probably <5 times) and you should crash.

In this PR build, you should no longer crash.

hifi-gustavo:
Build console output for commit: 00db0953067a4408344c8438fc608c5a732c2e32 available at: https://gist.github.com/140cac477b091495fd2861806749b71f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7519.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7519.dmg

hifi-gustavo:
Build console output for commit: 6e0d6ee45bdb2a609f7fb51cce7605e8a9537eff available at: https://gist.github.com/140cac477b091495fd2861806749b71f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7519.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7519.dmg

ZappoMan:
Smoke tested and did the two repro steps described above. rapidly reset my avatar almost 50+ times... same with clearing the cache... no crashes

PR #7522, opened 2016-03-31
optimize checkEnterLeaveEntities() to do distance check and stationary checks less frequently

ZappoMan:


hifi-gustavo:
Build console output for commit: 28978c8adbbba97b9b1064c889f19fc382144e47 available at: https://gist.github.com/1f3e786d482f51e070ca46801ebbc7bc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7522.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7522.dmg

ZappoMan:
This will fix this issue: https://app.asana.com/0/26320649018758/106608718702734
And hopefully address the root cause of this issue: https://app.asana.com/0/26320649018758/106541886040040

Test Plan:
- Smoke test
- Make sure zones still work
- Make sure enter/leave entity scripts still work
- Test creating a zone around your avatar while not moving, you should see the zone take effect even if you don't move

hifi-gustavo:
Build console output for commit: b76f8509aa0af0de2198fec7204cf9c1ca52e320 available at: https://gist.github.com/1f3e786d482f51e070ca46801ebbc7bc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7522.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7522.dmg

kevinmthomastr:
Successfully completed smoke test.  Everything is running as expected.

Successfully added a zone and modified its parameters to test for proper functionality.  Successfully tested on both the Music and Sandbox domain.  Zone functionality works well.
 
Successfully tested for enter/leave entity functionality to which I utilized my jsstreamplayerdomain-zone-entity.js.  Specifically when you enter the zone the script functionality works properly and toggles off upon exit.

Successfully created a zone in the Sandbox domain while not moving.  Verified that I was able to successfully see the zone take effect without moving my avatar.

kevinmthomas-carpool:
Successfully completed smoke test. Everything is running as expected.

Successfully added a zone and modified its parameters to test for proper functionality. Successfully tested on both the Music and Sandbox domain. Zone functionality works well.

Successfully tested for enter/leave entity functionality to which I utilized my jsstreamplayerdomain-zone-entity.js. Specifically when you enter the zone the script functionality works properly and toggles off upon exit.

Successfully created a zone in the Sandbox domain while not moving. Verified that I was able to successfully see the zone take effect without moving my avatar.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -218,12 +220,15 @@ bool EntityTreeRenderer::findBestZoneAndMaybeContainingEntities(const glm::vec3&
...
+    PerformanceTimer perfTimer("checkEnterLeaveEntities");
+    auto now = usecTimestampNow();
     bool didUpdate = false;
 
     if (_tree && !_shuttingDown) {
         glm::vec3 avatarPosition = _viewState->getAvatarPosition();
 
-        if (avatarPosition != _lastAvatarPosition) {
+        // If we've moved "enough" check to see our enter/leave state
+        if (glm::distance(avatarPosition, _lastAvatarPosition) > ZONE_CHECK_DISTANCE) {
```
AndrewMeadows: One probable way to reproduce the heavy lag is to go to a place that is "inside" the bounding box of many entities (such as the center of the Hell mansion) and then move around quickly such that this distance condition is constantly true.  The "moving" findBestZoneAndMaybeContainingEntities() case (which harvests  entitiesContainingAvatar) is more expensive than the "timeout" case.

ZappoMan: I've identified another potential optimization. Namely, we don't need to check the point inside for entities that don't have scripts... because only entities with scripts can get an enter/leave message... I'm going to add that optimization as well.

AndrewMeadows: I tested this a little bit in demo and cell science (Hell was unavailable) and could see "findBestZone" in the stats but couldn't make it get terrible, so we'll worry about it when we see problems.

AndrewMeadows:
Oops, I thought somehow I had accidentally added the DO_NOT_MERGE tag.  I'll put it back.

ZappoMan:
Ok - I've made some further optimizations... also I found another bug in the old code... namely the old code wouldn't handle the case of an entity with an "enter/leave" script moving around you if you were stationary...

New Test Plan:
- smoke test
- Make sure zones still work
- Make sure enter/leave entity scripts still work (NOTE: I've added a simple entity script that you can attach to a cube or sphere, and it will randomly change its color when you enter/leave it, make the cube not collide with myAvatar to test).
- Test creating a zone around your avatar while not moving, you should see the zone take effect even if you don't move
- Test "moving" an entity with an enter/leave script on to be around your avatar, you should still see the enter/leave events even if avatar is standing still

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -159,12 +160,14 @@ void EntityTreeRenderer::update() {
 
 bool EntityTreeRenderer::findBestZoneAndMaybeContainingEntities(const glm::vec3& avatarPosition, QVector<EntityItemID>* entitiesContainingAvatar) {
     bool didUpdate = false;
-    float radius = 1.0f; // for now, assume 1 meter radius
+    float radius = 0.01f; // for now, assume 0.01 meter radius, because we actually check the point inside later
```
ZappoMan: this is another small optimization, we really only care about entities that actually contain the point, but the EntityTree class doesn't have that method for searching. If we make the radius of the search sphere smaller, we reduce the potential for having to check entities that our sphere overlaps but don't actually contain our point.

hifi-gustavo:
Build console output for commit: 2e129a398f49b71b088476be0a40f435ed6269ca available at: https://gist.github.com/1f3e786d482f51e070ca46801ebbc7bc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7522.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7522.dmg

OmegaHeron:
smoke test - audio, placing/editing/removing objects, change domains all ok

Make sure zones still work - existing zones in Heron work as expected

Make sure enter/leave entity scripts still work - Tested with included script, works as test indicates it should

Test creating a zone around your avatar while not moving, you should see the zone take effect even if you don't move - Yes, it showed immediately without moving AV

Test "moving" an entity with an enter/leave script on to be around your avatar, you should still see the enter/leave events even if avatar is standing still - I think I tested this properly by moving cube with your test script to pass over my AV - it changed colors.

:+1: 

I ran into one issue that is likely unrelated to this PR, but, it happened during it for first time here.

I was in empty space, turned edit on, created a zone and played with its settings - deleted it once I was satisfied it was working for PR test.  Clicked button to turn off edit mode - save edit bar change but properties window remained open.  Couldn't close it with its closer icon.  No menu items would work, pressing enter wouldn't bring up location entry etc etc etc.  Have been unable to repro, but, have to mention it just in case.

ZappoMan:
@OmegaHeron - re edit window not hiding - that's a known bug unrelated to this PR

PR #7524, opened 2016-03-31
Allow display plugins to expose a preferred audio device

jherico:
This adds functionality to the display plugins to expose a preferred audio device, such as the Oculus headphones and mic.  This information is also exposed in the HMD scripting interface so that a script could automatically switch audio devices when the HMD is activated.

Note that apparently our QML layer doesn't use the same audio device as the AudioClient, so audio from QML surface and web entities will not respect this change (but neither will they respect a change to the audio device performed using the normal method in the menus.  It's a separate bug).

### DIFF
--- a/plugins/oculus/src/OculusDisplayPlugin.cpp
+++ b/plugins/oculus/src/OculusDisplayPlugin.cpp
@@ -6,7 +6,22 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 #include "OculusDisplayPlugin.h"
+
+#ifdef WIN32
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#include <Mmsystem.h>
+#include <mmdeviceapi.h>
+#include <devicetopology.h>
+#include <Functiondiscoverykeys_devpkey.h>
+#include <VersionHelpers.h>
+#endif
+
+#include <OVR_CAPI_Audio.h>
+#include <QtCore/QThread>
+
 #include <shared/NsightHelpers.h>
+#include <AudioClient.h>
 #include "OculusHelpers.h"
 
 const QString OculusDisplayPlugin::NAME("Oculus Rift");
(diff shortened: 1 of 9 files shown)


hifi-gustavo:
Build console output for commit: 3e55d6dd663c05ccb53003b4c9e816a46dca6112 available at: https://gist.github.com/72409f3ac9db929694d8c340c2c3d799<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7524.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7524.dmg

plugins/oculus/src/OculusDisplayPlugin.cpp
```diff
@@ -86,3 +130,64 @@ void OculusDisplayPlugin::hmdPresent() {
...
     }
 }
+
+bool OculusDisplayPlugin::isHmdMounted() const {
+    ovrSessionStatus status;
+    return (OVR_SUCCESS(ovr_GetSessionStatus(_session, &status)) && 
+        (ovrFalse != status.HmdMounted));
+}
+
+static QString audioDeviceFriendlyName(LPCWSTR device) {
```
ZappoMan: this should be in the Shared or Audio library -- please move it.

plugins/oculus/src/OculusDisplayPlugin.cpp
```diff
@@ -86,3 +130,64 @@ void OculusDisplayPlugin::hmdPresent() {
...
+    QString deviceName;
+
+    HRESULT hr = S_OK;
+    CoInitialize(NULL);
+    IMMDeviceEnumerator* pMMDeviceEnumerator = NULL;
+    CoCreateInstance(__uuidof(MMDeviceEnumerator), NULL, CLSCTX_ALL, __uuidof(IMMDeviceEnumerator), (void**)&pMMDeviceEnumerator);
+    IMMDevice* pEndpoint;
+    hr = pMMDeviceEnumerator->GetDevice(device, &pEndpoint);
+    if (hr == E_NOTFOUND) {
```
ZappoMan: this code is almost identical to the code in `defaultAudioDeviceForMode()` in `libraries/audio-client` I assume you either copied it from there, or you downloaded code from the internet... please change these two methods to use the same shard implementation

plugins/oculus/src/OculusDisplayPlugin.cpp
```diff
@@ -18,9 +33,38 @@ bool OculusDisplayPlugin::internalActivate() {
...
+    if (riftAudioIn != QString()) {
+        _savedAudioIn = audioClient->getDeviceName(QAudio::Mode::AudioInput);
+        QMetaObject::invokeMethod(audioClient.data(), "switchInputToAudioDevice", Q_ARG(const QString&, riftAudioIn));
+    } else {
+        _savedAudioIn.clear();
+    }
+
+    QString riftAudioOut = getPreferredAudioOutDevice();
+    if (riftAudioOut != QString()) {
+        _savedAudioOut = audioClient->getDeviceName(QAudio::Mode::AudioOutput);
```
ZappoMan: As we discussed in our stand up yesterday, I do not believe the plugin should be doing this switching automatically in the plugin. There are many reasons for this, but the main reason is that the state is now being implemented in multiple places. And so for example, if a script changes the audio device while in the HMD mode, this will switch away from the audio device the script chose and change it back to some other previous device when the user removes the HMD.

Instead I believe the plugin should ONLY be responsible for advertising it's desired audio devices. We should move all logic for what to do when the HMD is mounted or removed into the same place (which is currently `away.js`).

plugins/oculus/src/OculusDisplayPlugin.cpp
```diff
@@ -86,3 +93,26 @@ void OculusDisplayPlugin::hmdPresent() {
...
     }
 }
+
+bool OculusDisplayPlugin::isHmdMounted() const {
+    ovrSessionStatus status;
+    return (OVR_SUCCESS(ovr_GetSessionStatus(_session, &status)) && 
+        (ovrFalse != status.HmdMounted));
+}
+
+QString OculusDisplayPlugin::getPreferredAudioInDevice() const {
```
ZappoMan: but we need to expose it to scripts so `away.js` can access it... I assume we should just expose it on the `HMD` api... something like `HMD.preferredAudioInputDevice` `HMD.preferredAudioOutputDevice`

hifi-gustavo:
Build console output for commit: 6e3d03dc697bc41ba324bc4bce7dea44c3e008c8 available at: https://gist.github.com/72409f3ac9db929694d8c340c2c3d799<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7524.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7524.dmg

OmegaHeron:
I don't have hardware to properly test this, but, it doesn't break anything with my current complex setup and its plethora of audio devices + passes smoke test.  I'm sure an HMD equipped person will have more generally useful comments soon.

hifi-gustavo:
Build console output for commit: 97e96f7b6790531b51b1da100940e72e34fe2a34 available at: https://gist.github.com/72409f3ac9db929694d8c340c2c3d799<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7524.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7524.dmg

cozza13:
tested this with the Oculus and it is good to go.

PR #7528, opened 2016-03-31
Fbx reader safety

howard-stearns:
* Check whether qUncompress() failed, and bail from readFBX if so.
* ModelPackager and VHACDUtil utilities should check for failed fbx reads.


howard-stearns:
This is one tiny piece of https://app.asana.com/0/32622044445063/106103321129730, in which the hifi log file of the crash showed that qUcompress was failing, and we were building from the failed model.

However, it still doesn't explain or fix whatever caused the actual crash that followed.

howard-stearns:
Testing notes: Just make sure fbx read works. It would be good to exercise Edit->Package Model and the separate VHACD util.  The actual file that that had reported in the log as corrupt is actually fine, and a copy of it (as working) is at http://howard-stearns.github.io/models/junkok.fbx

hifi-gustavo:
Build console output for commit: 1df9a57e583ae3eecff728f2b72e6300e741c0f7 available at: https://gist.github.com/97f96d106dbc3318d7d915f566890283<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7528.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7528.dmg

libraries/fbx/src/FBXReader_Node.cpp
```diff
@@ -47,6 +47,9 @@ template<class T> QVariant readBinaryArray(QDataStream& in, int& position) {
         in.readRawData(compressed.data() + sizeof(quint32), compressedLength);
         position += compressedLength;
         QByteArray uncompressed = qUncompress(compressed);
+        if (uncompressed.isEmpty()) { // answers empty byte array if corrupt
+            throw QString("corrupt fbx file");
```
jherico: can we please throw `std::runtime_error` or something derived from it instead?

OmegaHeron:
Tested Windows 7 - No problems seen, but, I don't think this is synched with changes introduced in 7333 that leads to all my blender models having no textures.  If it is synched then it fixes...

Tested vhacd-util (linux compile) - Works as it always has.

libraries/fbx/src/FBXReader_Node.cpp
howard-stearns: Good idea, but changing the existing throws and catches classes should be a separate (non-P1) ticket.
https://app.asana.com/0/search/107509223713404/107214863719607

PR #7529, opened 2016-03-31
Add texture stress test

jherico:
Add a script that enables the user to stress test the result of overloading the GPU memory with texture data

This also fixes a small bug introduced in the mechanism used to pass messages from QML to JS


hifi-gustavo:
Build console output for commit: 27406f5b214b7af11a419239b5fa5472f9e3cf99 available at: https://gist.github.com/b79012556b87d239ec0cc99e404d8a09<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7529.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7529.dmg

examples/tests/textureStress.js
```diff
@@ -0,0 +1,67 @@
+Script.include("https://s3.amazonaws.com/DreamingContent/scripts/Austin.js");
```
ZappoMan: ??

libraries/ui/src/QmlWindowClass.cpp
```diff
@@ -122,13 +122,18 @@ void QmlWindowClass::initQml(QVariantMap properties) {
...
             // Forward messages received from QML on to the script
-            connect(_qmlWindow, SIGNAL(sendToScript(QVariant)), this, SIGNAL(fromQml(const QVariant&)), Qt::QueuedConnection);
+            connect(_qmlWindow, SIGNAL(sendToScript(QVariant)), this, SLOT(qmlToScript(const QVariant&)), Qt::QueuedConnection);
         });
     }
     Q_ASSERT(_qmlWindow);
     Q_ASSERT(dynamic_cast<const QQuickItem*>(_qmlWindow.data()));
 }
 
+void QmlWindowClass::qmlToScript(const QVariant& message) {
```
ZappoMan: what's this about? is it a bug fix?

libraries/ui/src/QmlWindowClass.cpp
```diff
@@ -122,13 +122,18 @@ void QmlWindowClass::initQml(QVariantMap properties) {
             object->setProperty(SOURCE_PROPERTY, _source);
 
             // Forward messages received from QML on to the script
-            connect(_qmlWindow, SIGNAL(sendToScript(QVariant)), this, SIGNAL(fromQml(const QVariant&)), Qt::QueuedConnection);
```
jherico: This is apparently not viable, because it sends the Hifi JS a QJSValue wrapped in a QVariant, which is not what we want.  Forcing it to go through a slot allows us to explicitly force the conversion using the `QJSValue.toVariant()` method, which *is* what we want.

ZappoMan: :ok_hand:

ZappoMan: please add a test plan to this PR that covers whatever this change my have an effect on....

cozza13:
Hi,

I tested this and it does do what it is supposed to. However the stats to not much up to the rendertest.js report, see image. 
capture

jherico:
@cozza13 the rendertest stats and my stats are using the same number.  1881886272 bytes is equal to 1794.7 gigabytes.

PR #7534, opened 2016-04-01
Application: don't call Model::simulate on network thread.

hyperlogic:
This can cause a crash.

On startup the Application::processOctreeStats method on the network thread will call into entities entity->isReadyToComputeShape() (without a tree lock) and trigger Model::simulate.  Model is NOT thread safe.  This was leading to a single model to be initialized on two threads simultaneously.

This might be somewhat rare, I only caught it because I was running a debug build.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -3215,6 +3215,25 @@ void Application::update(float deltaTime) {
 
     updateLOD();
 
+    if (!_physicsEnabled && _processOctreeStatsCounter > 0) {
+
+        // process octree stats packets are sent in between full sends of a scene.
+        // We keep physics disabled until we've recieved a full scene and everything near the avatar in that
+        // scene is ready to compute its collision shape.
+
+        if (nearbyEntitiesAreReadyForPhysics()) {
+            _physicsEnabled = true;
+            getMyAvatar()->updateMotionBehaviorFromMenu();
+        } else {
+            auto characterController = getMyAvatar()->getCharacterController();
+            if (characterController) {
+                // if we have a character controller, disable it here so the avatar doesn't get stuck due to
+                // a non-loading collision hull.
+                characterController->setEnabled(false);
+            }
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


hyperlogic:
https://app.asana.com/0/26225263936266/107126123683941

hifi-gustavo:
Build console output for commit: 3d43f73a882ddcad9a79424a8f0b35d016db90bf available at: https://gist.github.com/2244e011fe6812783b74f429a2d7b5d9<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7534.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7534.dmg

interface/src/Application.cpp
```diff
@@ -3215,6 +3215,25 @@ void Application::update(float deltaTime) {
 
     updateLOD();
 
+    if (!_physicsEnabled && _processOctreeStatsCounter > 0) {
+
+        // process octree stats packets are sent in between full sends of a scene.
+        // We keep physics disabled until we've recieved a full scene and everything near the avatar in that
+        // scene is ready to compute its collision shape.
+
+        if (nearbyEntitiesAreReadyForPhysics()) {
```
hyperlogic: I'm not 100% if we should lock the entity tree here, please advise.

sethalves: I don't think you need to lock here -- I see some entity-tree locking inside of nearbyEntitiesAreReadyForPhysics.

cozza13:
What kind of testing are you looking for?

hyperlogic:
Go into demo, walk very close to the blocks on the table.  Clear your cache.  quit.

The re start and make sure the boxes don't fall through the table and you don't crash. :)

cozza13:
I tested this, followed the instructions from above. QA Pss

PR #7544, opened 2016-04-01
faster kinematic motion for entities

AndrewMeadows:
- improve performance in domain with many moving kinematic objects (better, but completely solved)
- - faster kinematic integration (fewer mutex locks, fewer copies)
- - kinematic collision group no longer "touches" static or other kinematic objects
- - don't constantly update broadphase Aabb's of static objects that have not moved
- - faster isNaN() utils
- fix bug for kinematic objects: sudden jump when first set moving
- better Bullet profiler stats


AndrewMeadows:
This PR is the result of an effort to improve performance in the **cellscience** domain, which had started to show significant time spent in the "stepSimulation" stat.  The asana issues are:

https://app.asana.com/0/26623005675174/106408508227390
https://app.asana.com/0/26623005675174/106421247207454

hifi-gustavo:
Build console output for commit: 20914df3303e79cd61c79a1c91222171e4c15956 available at: https://gist.github.com/5e429efd4ff119b2d7f5a7c2f9e792ef<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7544.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7544.dmg

AndrewMeadows:
To test:

performance
(1) Login and navigate to cellscience domain, wait for overlay buttons to appear near window bottom
(2) Click on the "Inside the cell" button
(3) While you are waiting for all the models to load:
(3A) Enable stats overlay (press the / button, or enable Developer-->Stats menu option)
(3B) Enable detailed stats (Developer-->Timing-->Performance Timer-->Display Timing Details)
(3C) Look for the stat with path **/idle/update/physics/stepSimulation**
(4) Once the content is loaded look at the value of the **/idle/update/physics/stepSimulation** stat detail, which will vary over time.  Watch the stat for a minute or two and note its average/min/max values.
(5) The average/min/max values of the stat in (4) should be lower in this PR than in the previous release.

kinematic motion still works
(6) Make entityA (not a sphere!) that is NOT dynamic but nevertheless has a collision shape in the physics simulation (your avatar should be able to bump up against entityA)
(7) Make sure entityA has non-zero linear/angular damping values (so that it will eventually come to rest when it is set moving in step (8))
(8) Give entityA some reasonable velocity (1m/sec along one of the cardinal world axes)
--> entityA should move and slow down to a full stop (note there is a bug that prevents the velocity value in the entity properties from getting zeroed -- not new, we'll fix that in another PR)
(9) Verify that entityA collides correctly where it appears to be
(10) Give entityA some reasonable angular velocity (Note, the angular velocity is still in degrees/sec so a value of 60 should be noticeably fast enough)
--> entityA should turn and eventually rotate to a full stop
(11) Verify that entityA collides correctly for its apparent rotation

moved static object collides correctly
(12) Set the velocity of entityA to zero --> it is now "static" in the physics simulation
(13) Change the position of entityA a few meters to one side
(14) Verify that entityA collides correctly for its new apparent position.

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,76 @@ void EntityItem::simulate(const quint64& now) {
...
-            qCDebug(entities) << "    new position:" << position;
-            qCDebug(entities) << "    new velocity:" << velocity;
-            qCDebug(entities) << "    new AACube:" << getMaximumAACube();
-            qCDebug(entities) << "    old getAABox:" << getAABox();
-        #endif
+    }
+    setLocalEverything(transform, linearVelocity, angularVelocity);
+    if (!isMoving) {
+        // flag this entity to be removed from kinematic motion
+        _dirtyFlags |= Simulation::DIRTY_MOTION_TYPE;
```
AndrewMeadows: EntityItem::simulateKinematicMotion() was not performant.  These changes are an attempt to speed it up by reducing the number of locks and copies of large data structures (pass by reference instead).

libraries/physics/src/EntityMotionState.cpp
```diff
@@ -137,8 +137,6 @@ bool EntityMotionState::handleEasyChanges(uint32_t& flags) {
     if ((flags & Simulation::DIRTY_PHYSICS_ACTIVATION) && !_body->isActive()) {
         _body->activate();
     }
-
-    return true;
 }
```
AndrewMeadows: The return value of `EntityMotionState::handleEasyChanges()` wasn't actually being used.

libraries/physics/src/ObjectMotionState.cpp
```diff
@@ -39,7 +39,7 @@ const glm::vec3& ObjectMotionState::getWorldOffset() {
 }
 
 // static
-uint32_t worldSimulationStep = 0;
+uint32_t worldSimulationStep = 1;
```
AndrewMeadows: We start the global step at 1 now because when we initialize _lastKineticStep value when adding a kinematic object to the physics simulation we set it to `worldSimulationStep - 1` so that it starts moving on the first frame that it was set kinematic.

libraries/physics/src/ObjectMotionState.cpp
```diff
@@ -192,6 +196,10 @@ bool ObjectMotionState::handleEasyChanges(uint32_t& flags) {
         }
         worldTrans.setRotation(newRotation);
         _body->setWorldTransform(worldTrans);
+        if (!(flags & HARD_DIRTY_PHYSICS_FLAGS) && _body->isStaticObject()) {
+            // force activate static body so its Aabb is updated later
+            _body->activate(true);
+        }
```
AndrewMeadows: The normal Bullet code does not expect static objects to be active, but we're activating them here so that we can catch them later and update their Aabb's before the braodphase step.

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -50,6 +50,7 @@ void PhysicsEngine::init() {
         // default gravity of the world is zero, so each object must specify its own gravity
         // TODO: set up gravity zones
         _dynamicsWorld->setGravity(btVector3(0.0f, 0.0f, 0.0f));
+        _dynamicsWorld->setForceUpdateAllAabbs(false);
```
AndrewMeadows: We're changing the Bullet configuration to NOT update ALL RigidBody broadphase Aabb's every frame, only the active non-static ones.  This means we must track any static bodies that we've set active and manually update their Aabb's (and also manually deactivate them later).

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -80,6 +81,7 @@ void PhysicsEngine::addObjectToDynamicsWorld(ObjectMotionState* motionState) {
             body->setCollisionFlags(btCollisionObject::CF_KINEMATIC_OBJECT);
             body->updateInertiaTensor();
             motionState->updateBodyVelocities();
+            motionState->updateLastKinematicStep();
```
AndrewMeadows: Updating the lastKinematicStep when adding a KINEMATIC object to the world fixes a bug I noticed: kinematic objects would tend to "jump" forward when they started to move.

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -189,12 +191,18 @@ VectorOfMotionStates PhysicsEngine::changeObjects(const VectorOfMotionStates& ob
...
-                object->clearIncomingDirtyFlags();
-            } else {
-                stillNeedChange.push_back(object);
-            }
+            object->handleEasyChanges(flags);
+            object->clearIncomingDirtyFlags();
         }
+        if (object->getMotionType() == MOTION_TYPE_STATIC && object->isActive()) {
+            _activeStaticBodies.push_back(object->getRigidBody());
+        }
```
AndrewMeadows: This is where we build the list of active static bodies.

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -189,12 +191,18 @@ VectorOfMotionStates PhysicsEngine::changeObjects(const VectorOfMotionStates& ob
...
+    }
+    // active static bodies have changed (in an Easy way) and need their Aabbs updated
+    // but we've configured Bullet to NOT update them automatically (for improved performance)
+    // so we must do it outselves
+    for (size_t i = 0; i < _activeStaticBodies.size(); ++i) {
+        _dynamicsWorld->updateSingleAabb(_activeStaticBodies[i]);
     }
```
AndrewMeadows: And this is where we update the Aabb's of active static bodies.

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -388,6 +397,12 @@ const CollisionEvents& PhysicsEngine::getCollisionEvents() {
 
 const VectorOfMotionStates& PhysicsEngine::getOutgoingChanges() {
     BT_PROFILE("copyOutgoingChanges");
+    // Bullet will not deactivate static objects (it doesn't expect them to be active)
+    // so we must deactivate them ourselves
+    for (size_t i = 0; i < _activeStaticBodies.size(); ++i) {
+        _activeStaticBodies[i]->forceActivationState(ISLAND_SLEEPING);
+    }
+    _activeStaticBodies.clear();
```
AndrewMeadows: And this is where we clear the activation of static bodies, which happens at the end of the frame.

libraries/physics/src/ThreadSafeDynamicsWorld.cpp
```diff
@@ -143,3 +146,24 @@ void ThreadSafeDynamicsWorld::synchronizeMotionStates() {
...
+        btRigidBody* body = btRigidBody::upcast(colObj);
+        if (body && body->getActivationState() != ISLAND_SLEEPING)
+        {
+            if (body->isKinematicObject())
+            {
+                //to calculate velocities next frame
+                body->saveKinematicState(timeStep);
+            }
+        }
+    }
```
AndrewMeadows: This is basically a direct copy of the code out of the base class btDiscreteDynamicsWorld, which is why I've left the code formatting as it was found in the Bullet code.  The only difference here (for now) is the BT_PROFILE at the top.

libraries/shared/src/GLMHelpers.h
```diff
@@ -229,8 +229,8 @@ void generateBasisVectors(const glm::vec3& primaryAxis, const glm::vec3& seconda
 glm::vec2 getFacingDir2D(const glm::quat& rot);
 glm::vec2 getFacingDir2D(const glm::mat4& m);
 
-bool isNaN(glm::vec3 value);
-bool isNaN(glm::quat value);
+inline bool isNaN(const glm::vec3& value) { return isNaN(value.x) || isNaN(value.y) || isNaN(value.z); }
+inline bool isNaN(const glm::quat& value) { return isNaN(value.w) || isNaN(value.x) || isNaN(value.y) || isNaN(value.z); }
```
AndrewMeadows: Inlining these and the one for `float` in SharedUtil.h sped up a fast-loop release-build test of these methods 10X.

libraries/shared/src/SpatiallyNestable.cpp
```diff
@@ -849,3 +842,40 @@ AACube SpatiallyNestable::getQueryAACube() const {
...
+    // linear velocity
+    _velocityLock.withWriteLock([&] {
+        _velocity = localVelocity;
+    });
+    // angular velocity
+    _angularVelocityLock.withWriteLock([&] {
+        _angularVelocity = localAngularVelocity;
+    });
+    locationChanged();
+}
```
AndrewMeadows: These methods pass by reference and the `setLocalEverything()` uses only a single l`ocationChanged()` call.  Faster.

libraries/shared/src/Transform.h
```diff
@@ -145,7 +145,7 @@ class Transform {
     Vec4 transform(const Vec4& pos) const;
     Vec3 transform(const Vec3& pos) const;
 
-    bool containsNaN() const;
+    bool containsNaN() const { return isNaN(_rotation) || isNaN(glm::dot(_scale, _translation)); }
```
AndrewMeadows: This call is now 8X faster.  We're taking advantage of the fact that any NaN component in `_scale` or `_translation` will produce a NaN dot product, which is fast to compute and greatly reduces the number of `isNaN(float)` calls under the hood.

hifi-gustavo:
Build console output for commit: 0830c55bcfacef6c6f2fa49ad636acae83b4ed8d available at: https://gist.github.com/5e429efd4ff119b2d7f5a7c2f9e792ef<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7544.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7544.dmg

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,76 @@ void EntityItem::simulate(const quint64& now) {
...
-    qCDebug(entities) << "EntityItem::simulateKinematicMotion timeElapsed" << timeElapsed;
-#endif
-
-    const float MIN_TIME_SKIP = 0.0f;
-    const float MAX_TIME_SKIP = 1.0f; // in seconds
-
-    timeElapsed = glm::clamp(timeElapsed, MIN_TIME_SKIP, MAX_TIME_SKIP);
-
-    if (hasActions()) {
+    if (hasActions() || timeElapsed < 0.0f) {
```
zzmp: It seems strange that negative time may elapse.

libraries/physics/src/ObjectMotionState.cpp
zzmp: Add this as a comment.
Initialize with `{}`, not `=`.

libraries/physics/src/ObjectMotionState.cpp
zzmp: This is the last thing you do in both branches of this if-statement.
Can you just move it to after the if-statement, L204?

libraries/physics/src/ObjectMotionState.h
```diff
@@ -50,11 +50,12 @@ const uint32_t HARD_DIRTY_PHYSICS_FLAGS = (uint32_t)(Simulation::DIRTY_MOTION_TY
                                                      Simulation::DIRTY_COLLISION_GROUP);
 const uint32_t EASY_DIRTY_PHYSICS_FLAGS = (uint32_t)(Simulation::DIRTY_TRANSFORM | Simulation::DIRTY_VELOCITIES |
                                                      Simulation::DIRTY_MASS | Simulation::DIRTY_MATERIAL |
-                                                     Simulation::DIRTY_SIMULATOR_ID | Simulation::DIRTY_SIMULATION_OWNERSHIP_PRIORITY);
+                                                     Simulation::DIRTY_SIMULATOR_ID | Simulation::DIRTY_SIMULATION_OWNERSHIP_PRIORITY |
+                                                     Simulation::DIRTY_PHYSICS_ACTIVATION);
```
zzmp: What will this do?

libraries/physics/src/PhysicsEngine.cpp
zzmp: This seems important to know in the future. Add it as a comment.

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -189,12 +191,18 @@ VectorOfMotionStates PhysicsEngine::changeObjects(const VectorOfMotionStates& ob
...
+            object->handleEasyChanges(flags);
+            object->clearIncomingDirtyFlags();
         }
+        if (object->getMotionType() == MOTION_TYPE_STATIC && object->isActive()) {
+            _activeStaticBodies.push_back(object->getRigidBody());
+        }
+    }
+    // active static bodies have changed (in an Easy way) and need their Aabbs updated
+    // but we've configured Bullet to NOT update them automatically (for improved performance)
+    // so we must do it outselves
```
zzmp: typo

libraries/shared/src/GLMHelpers.h
zzmp: That's awesome. I'd love to know how you found this optimization.

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,76 @@ void EntityItem::simulate(const quint64& now) {
...
         return;
     }
 
-    if (hasLocalAngularVelocity()) {
```
zzmp: Are these `*local*` methods still used now that you made `*LocalTransformAndVelocities` methods?

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,76 @@ void EntityItem::simulate(const quint64& now) {
...
+    const float MAX_TIME_ELAPSED = 1.0f; // seconds
+    timeElapsed = glm::min(timeElapsed, MAX_TIME_ELAPSED);
+
+    Transform transform;
+    glm::vec3 linearVelocity;
+    glm::vec3 angularVelocity;
+    getLocalTransformAndVelocities(transform, linearVelocity, angularVelocity);
 
+    bool isMoving = false;
+    if (glm::length2(angularVelocity) > 0.0f) {
```
zzmp: Will this change existing behavior?

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,76 @@ void EntityItem::simulate(const quint64& now) {
...
-        float angularSpeed = glm::length(localAngularVelocity);
-
-        const float EPSILON_ANGULAR_VELOCITY_LENGTH = 0.0017453f; // 0.0017453 rad/sec = 0.1f degrees/sec
-        if (angularSpeed < EPSILON_ANGULAR_VELOCITY_LENGTH) {
-            if (setFlags && angularSpeed > 0.0f) {
-                _dirtyFlags |= Simulation::DIRTY_MOTION_TYPE;
-            }
-            localAngularVelocity = ENTITY_ITEM_ZERO_VEC3;
+        const float EPSILON_ANGULAR_VELOCITY_LENGTH_SQUARED = 0.0017453f * 0.0017453f; // 0.0017453 rad/sec = 0.1f degrees/sec
+        if (glm::length2(angularVelocity) < EPSILON_ANGULAR_VELOCITY_LENGTH_SQUARED) {
```
zzmp: Why are you squaring the right-hand side?

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,76 @@ void EntityItem::simulate(const quint64& now) {
...
+            angularVelocity = ENTITY_ITEM_ZERO_VEC3;
```
zzmp: Why are you no longer setting `Simulation::DIRTY_MOTION_TYPE`?

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,76 @@ void EntityItem::simulate(const quint64& now) {
...
         } else {
             // for improved agreement with the way Bullet integrates rotations we use an approximation
             // and break the integration into bullet-sized substeps
-            glm::quat rotation = getRotation();
+            glm::quat rotation = transform.getRotation();
             float dt = timeElapsed;
-            while (dt > PHYSICS_ENGINE_FIXED_SUBSTEP) {
-                glm::quat  dQ = computeBulletRotationStep(localAngularVelocity, PHYSICS_ENGINE_FIXED_SUBSTEP);
+            while (dt > 0.0f) {
+                glm::quat  dQ = computeBulletRotationStep(angularVelocity, glm::min(dt, PHYSICS_ENGINE_FIXED_SUBSTEP));
```
zzmp: There's a weird double-space here.

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,76 @@ void EntityItem::simulate(const quint64& now) {
...
+            angularVelocity = ENTITY_ITEM_ZERO_VEC3;
         } else {
             // for improved agreement with the way Bullet integrates rotations we use an approximation
             // and break the integration into bullet-sized substeps
-            glm::quat rotation = getRotation();
+            glm::quat rotation = transform.getRotation();
             float dt = timeElapsed;
-            while (dt > PHYSICS_ENGINE_FIXED_SUBSTEP) {
-                glm::quat  dQ = computeBulletRotationStep(localAngularVelocity, PHYSICS_ENGINE_FIXED_SUBSTEP);
+            while (dt > 0.0f) {
```
zzmp: Why not `glm::length2(dt)`?

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,76 @@ void EntityItem::simulate(const quint64& now) {
...
-        if (success) {
-            localAcceleration = glm::inverse(transform.getRotation()) * getAcceleration();
-        } else {
-            localAcceleration = getAcceleration();
-        }
-
+    if (glm::length2(linearVelocity) > 0.0f) {
         // linear damping
-        glm::vec3 localVelocity = getLocalVelocity();
         if (_damping > 0.0f) {
```
zzmp: Why not `glm::length2(_damping)`?

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,76 @@ void EntityItem::simulate(const quint64& now) {
...
-        // apply effective acceleration, which will be the same as gravity if the Entity isn't at rest.
-        localVelocity += localAcceleration * timeElapsed;
+        // integrate linearVelocity
+        linearVelocity += linearAcceleration * timeElapsed;
 
-        float speed = glm::length(localVelocity);
-        const float EPSILON_LINEAR_VELOCITY_LENGTH = 0.001f; // 1mm/sec
-        if (speed < EPSILON_LINEAR_VELOCITY_LENGTH) {
+        const float EPSILON_LINEAR_VELOCITY_LENGTH_SQUARED = 1.0e-6f; // 1mm/sec ^2
+        if (glm::length2(linearVelocity) < EPSILON_LINEAR_VELOCITY_LENGTH_SQUARED) {
```
zzmp: Why square the rhs?

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,76 @@ void EntityItem::simulate(const quint64& now) {
...
-            if (setFlags && speed > 0.0f) {
-                _dirtyFlags |= Simulation::DIRTY_MOTION_TYPE;
-            }
         } else {
-            setLocalPosition(localPosition);
-            setLocalVelocity(localVelocity);
+            // integrate position forward
+            // NOTE: we're using the NEW linear velocity, which is why we negate the acceleration term
+            glm::vec3 position = transform.getTranslation() + (linearVelocity * timeElapsed) - 0.5f * linearAcceleration * timeElapsed * timeElapsed;
+            transform.setTranslation(position);
```
zzmp: We are no longer setting the velocity.

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,76 @@ void EntityItem::simulate(const quint64& now) {
...
             setVelocity(ENTITY_ITEM_ZERO_VEC3);
-            if (setFlags && speed > 0.0f) {
-                _dirtyFlags |= Simulation::DIRTY_MOTION_TYPE;
-            }
         } else {
-            setLocalPosition(localPosition);
-            setLocalVelocity(localVelocity);
+            // integrate position forward
+            // NOTE: we're using the NEW linear velocity, which is why we negate the acceleration term
+            glm::vec3 position = transform.getTranslation() + (linearVelocity * timeElapsed) - 0.5f * linearAcceleration * timeElapsed * timeElapsed;
```
zzmp: Would it be more clear/faster to hold onto the old linear velocity?

libraries/entities/src/EntityItem.cpp
zzmp: Why not `glm::length2(timeElapsed)`?

AndrewMeadows: Regarding negative time: after I worked on the bug below I no longer take positive time for granted, especially if it is time that is computed by our own code:

https://app.asana.com/0/26225263936266/105603974000695/f

glm::length2() could have been called glm::lengthSquared().  It is faster than taking the length of a vector because it avoids a sqrt() call, which is why most Vector3 implementations include it in their API.  In this particular case timeElapsed is a scalar rather than a Vector3.

libraries/entities/src/EntityItem.cpp
AndrewMeadows: Definitely yes.

libraries/entities/src/EntityItem.cpp
AndrewMeadows: Using length2() here is faster than length() because it avoids a square root.  In this particular case we never actually need to know the true length, just whether it is more than zero.

libraries/entities/src/EntityItem.cpp
AndrewMeadows: glm::length2() computes the squared length of the vec3, so we must compare it to the squared minimum value.

libraries/entities/src/EntityItem.cpp
AndrewMeadows: Because dt is a scalar.

libraries/entities/src/EntityItem.cpp
AndrewMeadows: _damping is a scalar.

libraries/entities/src/EntityItem.cpp
zzmp: I did not read the documentation for length2 and assumed it truncated, not squared. I guess this answers most of my comments :).

libraries/entities/src/EntityItem.cpp
AndrewMeadows: This was a not-very-helpful optimization that introduces an (assumed negligible) approximation error when the object is coming to rest however...

... after looking at this code again I realize there is a bug: there is a particular case where we don't want to assume the object is actually stopping even though its velocity may be small: vertical ballistic motion in non-zero gravitational field near the peak of its trajectory.  I'll have to go back and fix this.

libraries/physics/src/ObjectMotionState.cpp
AndrewMeadows: No.  This is an `if + else if` block rather than an `if + else`.

libraries/physics/src/ObjectMotionState.h
AndrewMeadows: Moving the `DIRTY_PHYSICS_ACTIVATION` flag to `EASY_DIRTY_PHYSICS_FLAGS` guarantees that `ObjectMotionState::handleEasyChanges()` will be called later (we call that whenever any of the easy flags are set), which is where I put the `_body->activate(true);` call above.

libraries/shared/src/GLMHelpers.h
AndrewMeadows: Actually there were two things that sped these up: (1) passing by reference avoids a copy and (2) inlining.  I didn't test to see how much each change contributed to the final measure.  I tried summing the components and then making a single call to isNaN(float) but that was slower than checking each component once --> the cost of floating point math was larger than the function call and logic branching overhead.

Really I was trying to reduce the cost of Transform::containsNaN() which called these methods.  It looked innocuous at the surface but it was hiding at least 13 function calls under the hood --> low-hanging fruit for optimization.

hifi-gustavo:
Build console output for commit: 74058ac049336dbdf7a332b6458be7ee68b7e8f5 available at: https://gist.github.com/5e429efd4ff119b2d7f5a7c2f9e792ef<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7544.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7544.dmg

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,80 @@ void EntityItem::simulate(const quint64& now) {
...
-            qCDebug(entities) << "    timeElapsed:" << timeElapsed;
-            qCDebug(entities) << "    old AACube:" << getMaximumAACube();
-            qCDebug(entities) << "    old position:" << localPosition;
-            qCDebug(entities) << "    old velocity:" << localVelocity;
-            qCDebug(entities) << "    old getAABox:" << getAABox();
-            qCDebug(entities) << "    newPosition:" << newPosition;
-            qCDebug(entities) << "    glm::distance(newPosition, position):" << glm::distance(newLocalPosition, localPosition);
-        #endif
+        glm::vec3 linearAcceleration = _acceleration;
+        bool nonZeroAcceleration = (glm::length2(_acceleration) > 0.0f);
```
ZappoMan: is there a reason you didn't use `linearAcceleration` here?

libraries/entities/src/EntityItem.cpp
```diff
@@ -876,123 +876,80 @@ void EntityItem::simulate(const quint64& now) {
...
+        // integrate linearVelocity
+        linearVelocity += linearAcceleration * timeElapsed;
 
-        float speed = glm::length(localVelocity);
-        const float EPSILON_LINEAR_VELOCITY_LENGTH = 0.001f; // 1mm/sec
-        if (speed < EPSILON_LINEAR_VELOCITY_LENGTH) {
+        const float EPSILON_LINEAR_VELOCITY_LENGTH_SQUARED = 1.0e-6f; // 1mm/sec ^2
+        if (glm::length2(linearVelocity) < EPSILON_LINEAR_VELOCITY_LENGTH_SQUARED) {
             setVelocity(ENTITY_ITEM_ZERO_VEC3);
-            if (setFlags && speed > 0.0f) {
```
ZappoMan: this looks like a change in logic? 
1) have I misread it?
2) was it intentional to change the logic? was there a bug this change was fixing?
3) if you didn't intend to change the logic, can you confirm that the new logic is actually the same?

*Old logic:*
```
    float speed = glm::length(localVelocity);
    const float EPSILON_LINEAR_VELOCITY_LENGTH = 0.001f; // 1mm/sec
    if (speed < EPSILON_LINEAR_VELOCITY_LENGTH) {
        setVelocity(ENTITY_ITEM_ZERO_VEC3);
        if (setFlags && speed > 0.0f) {
            _dirtyFlags |= Simulation::DIRTY_MOTION_TYPE;
        }
    } else {
        setLocalPosition(localPosition);
        setLocalVelocity(localVelocity);
    }
```

* New Logic:*
```
    // this part is the same, just using faster length2() and an epsilon that is squared
    const float EPSILON_LINEAR_VELOCITY_LENGTH_SQUARED = 1.0e-6f; // 1mm/sec ^2
    if (glm::length2(linearVelocity) < EPSILON_LINEAR_VELOCITY_LENGTH_SQUARED) {
        setVelocity(ENTITY_ITEM_ZERO_VEC3); // that's the same...

        // this is where things start to get different... we used to be comparing "speed" or length of velocity
        // now we're comparing acceleration?
        if (nonZeroAcceleration) {
            isMoving = true;
        }
    } else {
         isMoving = true; // this is also different
    }

    // and this all looks very different
    setLocalTransformAndVelocities(transform, linearVelocity, angularVelocity);
    if (!isMoving) {
        // flag this entity to be removed from kinematic motion
        _dirtyFlags |= Simulation::DIRTY_MOTION_TYPE;
    }
```

libraries/shared/src/Transform.h
ZappoMan: How many multiplies does `dot()`, if you're saying that the multiplies are faster than the comparisons against self, then I wonder would we get a similar performance boost in the `isNaN(const glm::vec3& value)` and `isNaN(const glm::quat& value)` by multiplying the components together?

OmegaHeron:
Windows 7
Test 1 - Inside the cell.  Physics time about 5.2ms for me with production.  This PR about 3.5

Test 2 - Motion.  Linear - collision correct after decay to stop.  Yes.  Angular - collision correct after decay to stop.  Yes.  Extra credit linear and angular together - collision correct after decay to stop.  Yes

Test 3 - Entity to static - moved entity - collision correct.  Yes.

QA Pass

AndrewMeadows:
I just found a bug in the code.  I'll write a new testplan soon, after I fix it.

AndrewMeadows:
I fixed the bug and updated the test plan above: I added a section at the bottom for **very low gravity**.

hifi-gustavo:
Build console output for commit: a53cb2e532e568caed6f6e1b40d8cccecf2a4fa1 available at: https://gist.github.com/5e429efd4ff119b2d7f5a7c2f9e792ef<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7544.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7544.dmg

libraries/shared/src/Transform.h
AndrewMeadows: Yeah, I tried that for the **vec3** and **quat** methods.  Turns out the sequential check of all components was faster (for release build on linux).  For `containsNaN()` I also tried adding the two vectors and rotating them by the quaternion -- turns out the quaternion rotation operation is more expensive than you expect: approximately 3 * 4 multiplies and 3 additions... per component.

In theory, if we're using SIMD math instructions then the dot product is about as expensive as multiplying two floats, but I'm not positive it is actually hooked up.  In any case I measured empirically and tried all the combinations I could think of.

OmegaHeron:
Tested Windows 7

Test 1 - cellscience physics time.  Same as before - all good.
Test 2 - linear/angular velocity.  Same as before - all good.
Test 3 - Back to static.  Same as before - all good.
Test 4 - Very Low Gravity - Followed test instructions and it responded as test outlined

QA Pass

libraries/entities/src/EntityItem.cpp
AndrewMeadows: Good catch.  Yeah I lost the code that paid attention to setFlags.  Looking at how it was meant to be used I realize I should restore it.

OmegaHeron:
@AndrewMeadows - will I need to re-test after change in a08ab8e is applied in re-build?

AndrewMeadows:
Note, I reduced the speedup improvements I expect for this PR.  I still see cellscience go up to 9msec in the stepSimulation stat sometimes.

BTW, a dump of the time spent in various sub-parts of a single simulation step can be obtained from the interface logs right after pressing the 'f' key.  This debug feature has been around for a while so it works in this PR and older releases.

OmegaHeron:
@AndrewMeadows - In my case running on Windows I see it start highish then it steadily slows until it reaches a steady state running +/- .2 to .3 ms around the lowest value.

AndrewMeadows:
Yes, @OmegaHeron, wait for the final build.  There was a bug that zappoman found: time-delayed updates from the entity-server were incorrectly setting some dirty bits which would cause the PhysicsEngine to do a bit more work before the stepSimulation call (checking to see if RigidBody needed to be converted from KINEMATIC to STATIC).  I don't expect the bug to actually cause any weird behavior -- it would introduce a bit of extra useless work in the interface.

hifi-gustavo:
Build console output for commit: a08ab8e127d066aacf73ca12cac9679f1a8b19f1 available at: https://gist.github.com/5e429efd4ff119b2d7f5a7c2f9e792ef<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7544.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7544.dmg

OmegaHeron:
Ok - testing again now.

OmegaHeron:
Test 1 - Cellscience physics time - Pass
Test 2 - Linear/Angular velocity application - Pass
Test 3 - Return to static - Pass
Test 4 - VLG - Pass

All respond exactly as previous tests.  Cellscience step time might be a fraction of a ms faster, but that's likely within realm of random vs > good.

AndrewMeadows:
Seth noticed a showstopper bug that appears to be new to this PR: dynamic entities with gravity would "pop" from interpenetrations with the floor.  We rolled this back so that it wouldn't mess up hifi/master.

PR #7547, opened 2016-04-01
Lock trees on import export

howard-stearns:
Read-lock entity tree on both export methods.
Write-lock clipboard tree on import.
Read-lock clipboard when pasting from it to entity tree (during which the entity-tree was already write-locked, so no change there).

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -2815,60 +2815,68 @@ bool Application::exportEntities(const QString& filename, const QVector<EntityIt
     auto entityTree = getEntities()->getTree();
     auto exportTree = std::make_shared<EntityTree>();
     exportTree->createRootElement();
-
     glm::vec3 root(TREE_SCALE, TREE_SCALE, TREE_SCALE);
-    for (auto entityID : entityIDs) { // Gather entities and properties.
-        auto entityItem = entityTree->findEntityByEntityItemID(entityID);
-        if (!entityItem) {
-            qCWarning(interfaceapp) << "Skipping export of" << entityID << "that is not in scene.";
-            continue;
-        }
-
-        if (!givenOffset) {
-            EntityItemID parentID = entityItem->getParentID();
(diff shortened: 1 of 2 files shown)


howard-stearns:
Possible fix for https://app.asana.com/0/32622044445063/96807675573738

Although I have never been able to reproduce the "entity disappears on export", inspection of the code shows that these locks were missing.

hifi-gustavo:
Build console output for commit: cec4175e9ddfe4e2c7c4eddc4da4aa1e3771dfd1 available at: https://gist.github.com/d30eca19ae1d6659dfce5629168e40f0<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7547.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7547.dmg

kevinmthomas-carpool:
Successfully passed smoke test.  Testing on the Music domain with a Mac on OS X El Capitan.  First added an asset from the ATP to which it added, manipulated and deleted successfully.  Second test was to add an asset from Dropbox to which I also manipulated the size, collision properties and successfully deleted item.  Graphics experience had no change from main Mac release 4610 to which I ran my control comparison.  Final test was to visit the Sandbox domain.  At the time it was running 4 servers however the asset server was working properly.  I checked again with the main Mac release 4610 and it was also running 4 servers so that it was not an issue with the PR build.  I was successfully able to add an item from my Dropbox, manipulate it and delete it.  Successful test.

howard-stearns:
Cool. Thanks @kevinmthomas-carpool . Just want to make sure: the "adding" is by Edit->Import and/or there was exercising of Edit->Export (after first selecting one or more entities with edit.js)?

kevinmthomas-carpool:
Pleasure @howard-stearns and yes you are correct.

PR #7546, opened 2016-04-01
Release skybox textures when rendering stars

zzmp:
- Adds a virtual method `Skybox::clear` to release textures
- Calls clear when not rendering a skybox
- Amends procedural shader parsing to release unused resources (shader/textures)

---

#### Intent
Frees GPU texture memory when switching domains, or switching to a zone with no skybox.
(See: https://app.asana.com/0/74051501495175/107448718339742.)

##### Repro
1. Test twice; on one test, set your texture cache size to 0 (`Developer > Network > RAM Caches Size`).
2. Open `renderStats.js` and go to an area with no zone.
3. Load fully, and note the GPU gpu::Texture memory usage.
4. Go to `hifi://playa`, and wait for the skybox (celestial map) to load fully.
5. Go back to the area with no zone.

##### Expectations
In current build, the skybox textures will stick around. Your memory usage will be appreciably larger than it was before.
In this PR build, the skybox textures will go away. Your memory usage will be less than in current build.

This changes the procedural shaders implementation, so that should be given a smoke test as well.

PR #7549, opened 2016-04-02
Add a visualization of cacheable resource usage

zzmp:
- Move plotperf qml visualization to a shared folder
- Changes plotperf to poll and redraw at 10fps
- Clean ResourceCache to avoid exposing dangerous methods
- Add counter properties to ResourceCache for count and size, total and cached
- Add resources property to ResourceCache to list active resource URLs
- Expose Animation, Model, Texture, Sound caches to QML/JS
- Add cacheStats.js to show cache count/size

---

#### Intent
We are seeing slow rendering because of high resource usage. This is meant to show if resources are being freed or not, whether they are being held by something or cached appropriately, and whether the caches are releasing resources appropriately
(See: https://app.asana.com/0/74051501495175/105559909158465.)

#### Testing

- Ensure that renderStats.js still works.
- Ensure that cacheStats.js works as expected. It should show resource usage. Take it for a spin across a few domains :smile:.
- Check TextureCache.resources in the JS console (or ModelCache, AnimationCache, SoundCache) and see the loaded/loading resources in your domain.

*NOTE: You may notice a large amount of texture usage (400Mb) without it being cached after visiting the playa. This is fixed in #7546, and is exactly the kind of thing this tool is meant to help find.*


libraries/networking/src/ResourceCache.h
```diff
@@ -101,15 +111,21 @@ class ResourceCache : public QObject {
...
     void refresh(const QUrl& url);
 
+signals:
+    void dirty();
+
 public slots:
     void checkAsynchronousGets();
 
+protected slots:
+    void updateTotalSize(const qint64& oldSize, const qint64& newSize);
```
zzmp: This will be exposed to the JS/QML. It is only used for stats, so I considered it OK - not a vector for maliciousness.

libraries/networking/src/ResourceCache.h
```diff
@@ -101,15 +111,21 @@ class ResourceCache : public QObject {
     void refreshAll();
     void refresh(const QUrl& url);
 
+signals:
+    void dirty();
+
 public slots:
     void checkAsynchronousGets();
```
zzmp: This will be exposed to the JS/QML. It can only trigger gets on the networking thread, so I considered it OK - not a vector for maliciousness.

libraries/networking/src/ResourceCache.h
```diff
@@ -101,15 +111,21 @@ class ResourceCache : public QObject {
     void refreshAll();
     void refresh(const QUrl& url);
 
+signals:
+    void dirty();
```
zzmp: This will be exposed to the JS/QML. It is literally used to expose changes to QML, so I considered it OK - not a vector for maliciousness.

hifi-gustavo:
Build console output for commit: d78338330c70a8328f2c4a9ee604b5ffb78a3beb available at: https://gist.github.com/cf537a101ffbab4dcc7ebb072e66b9a4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7549.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7549.dmg

zzmp:
I should add a property exposing `ResourceCache::_resources.uniqueKeys` so that in the case of unreleased resources we can see which resource URLs are persisting. Even if it's not in the graphs, it would be valuable as a one-off call in JS.

hifi-gustavo:
Build console output for commit: 3c86191a59f7f68cdcb0526b071480151e598cb5 available at: https://gist.github.com/cf537a101ffbab4dcc7ebb072e66b9a4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7549.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7549.dmg

zzmp:
test this please

hifi-gustavo:
Build console output for commit: 3c86191a59f7f68cdcb0526b071480151e598cb5 available at: https://gist.github.com/cf537a101ffbab4dcc7ebb072e66b9a4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7549.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7549.dmg

libraries/networking/src/ResourceCache.cpp
```diff
@@ -56,6 +56,16 @@ void ResourceCache::refresh(const QUrl& url) {
     }
 }
 
+const QVariantList ResourceCache::getResourceList() const {
```
ZappoMan: It's not clear to me why you need this to be `const` -- since you're returning an object, the caller can assign it to their own local version which they can modify anyway...

libraries/networking/src/ResourceCache.h
```diff
@@ -101,15 +114,21 @@ class ResourceCache : public QObject {
...
 
+protected slots:
+    void updateTotalSize(const qint64& oldSize, const qint64& newSize);
+
 protected:
     /// Loads a resource from the specified URL.
     /// \param fallback a fallback URL to load if the desired one is unavailable
     /// \param delayLoad if true, don't load the resource immediately; wait until load is first requested
     /// \param extra extra data to pass to the creator, if appropriate
-    Q_INVOKABLE QSharedPointer<Resource> getResource(const QUrl& url, const QUrl& fallback = QUrl(),
```
ZappoMan: don't you need this to be invokable so you can marshal any callers back onto the correct thread? or his the cache thread safe from multiple threads?

libraries/networking/src/ResourceCache.cpp
jherico: Yes, it only really makes sense for a return value to be const if it's returning a reference.

libraries/networking/src/ResourceCache.h
```diff
@@ -79,8 +79,21 @@ class ResourceCacheSharedItems : public Dependency  {
 /// Base class for resource caches.
 class ResourceCache : public QObject {
     Q_OBJECT
+    Q_PROPERTY(int numTotal READ getNumTotalResources NOTIFY dirty)
```
jherico: I would prefer if these were named `getXXXCount` and `getXXXSize`.  Also, I would recomment using size_t for both of them.

libraries/networking/src/ResourceCache.h
```diff
@@ -79,8 +79,21 @@ class ResourceCacheSharedItems : public Dependency  {
...
+    Q_PROPERTY(int numCached READ getNumCachedResources NOTIFY dirty)
+    Q_PROPERTY(qint64 sizeTotal READ getSizeTotalResources NOTIFY dirty)
+    Q_PROPERTY(qint64 sizeCached READ getSizeCachedResources NOTIFY dirty)
+    Q_PROPERTY(QVariantList resources READ getResourceList NOTIFY dirty)
     
 public:
+    int getNumTotalResources() const { return _resources.size(); }
```
jherico: not thread safe

libraries/networking/src/ResourceCache.h
```diff
@@ -79,8 +79,21 @@ class ResourceCacheSharedItems : public Dependency  {
...
+    qint64 getSizeTotalResources() const { return _totalResourcesSize; }
```
jherico: not thread safe

libraries/networking/src/ResourceCache.h
```diff
@@ -79,8 +79,21 @@ class ResourceCacheSharedItems : public Dependency  {
...
+
+    int getNumCachedResources() const { return _unusedResources.size(); }
```
jherico: not thread safe

libraries/networking/src/ResourceCache.h
```diff
@@ -79,8 +79,21 @@ class ResourceCacheSharedItems : public Dependency  {
...
+    qint64 getSizeCachedResources() const { return _unusedResourcesSize; }
```
jherico: not thread safe

libraries/networking/src/ResourceCache.cpp
```diff
@@ -56,6 +56,16 @@ void ResourceCache::refresh(const QUrl& url) {
...
+    QVariantList list;
+    auto resources = _resources.uniqueKeys();
```
jherico: not thread safe

libraries/networking/src/ResourceCache.cpp
```diff
@@ -56,6 +56,16 @@ void ResourceCache::refresh(const QUrl& url) {
...
+    list.reserve(resources.size());
+    for (auto& resource : resources) {
```
jherico: not thread safe

libraries/networking/src/ResourceCache.cpp
```diff
@@ -377,6 +400,11 @@ void Resource::finishedLoading(bool success) {
     emit finished(success);
 }
 
+void Resource::setBytes(const qint64& bytes) {
```
jherico: this should really be 'setSize' as 'setBytes' implies you're providing data

kevinmthomas-carpool:
PR Build: 7549
Testing OS:  Windows 10
Testing Location:  hifi://music, hifi://sandbox, hifi://playa, hifi://cellscience, hifi://entry
Smoke Test:  Successful
Test Results:  Successful smoke test.  Initial test on Sandbox domain.  Verified renderstats.js working properly however the visual display was hard to read as the fonts were not working properly.  The respective data is as follows: Num Buffers 2406, CPU 2165, GPU 371; gpu::Buffer Memory 97.0 Mb, CPU 87.3 Mb, GPU 38.1 Mb; Num Textures 410, CPU 173, GPU 132; gpu::Texture Memory 808.0 Mb, CPU 210.6 Mb, GPU 727.2 Mb.  Successful load of cachestats.js which also had the same issue with a hard to read font.  To note, data math does not seem consistent in the following results: Animation Count 28.9, total 26.0, cached 9.0, Animation Size 15.0 Mb, total 13.5 Mb, cached 4.2 Mb, Model Count 24.4, total 22.0, cached 1.0; Model Size 251.5 Mb, total 226.3 Mb, cached 0.0 Mb, Texture Count 161.1, total 145.0, cached 6.0; Texture Size 853.3 Mb, total 767.9 Mb, cached 40.9 Mb.  Successful result of TextureCache.resources as it was a match to textures in domain.  Tested additionally on hifi://music, hifi://playa, hifi://cellscience, hifi://entry.  Verified  a large amount of texture usage without it being cached after visiting Playa.  Results were consistent throughout each domain, the renderstats.js data appeared accurate however the values in the cachestats.js did not equate.  As you can see in the results from the Sandbox domain, the count did not add up from the total and cached values on each data point.  Attached also were the results from the Music domain as a snapshot was taken of both cachestats.js and renderstats.js for further inspection.  Upon logging out of Interface a crash occurred.  Attached results from hifi-log.txt in addition the bugsplat was successfully sent.
[attachments.zip](https://github.com/highfidelity/hifi/files/203031/attachments.zip)

cozza13:
@zzmp here is the bugsplat for the crash https://www.bugsplatsoftware.com/individualCrash/?id=2633&database=interface_alpha

libraries/networking/src/ResourceCache.h
zzmp: I can change them to `size_t`.
Why do you want `getXXXCount`? I prefer them to be nouns, as they are properties and not methods, so they will not be invoked. This is the standard in JS, and these will be exposed there (I'll admit I don't know any standard QML practices).

libraries/networking/src/ResourceCache.h
zzmp: There is no marshalling done.
Instead, if the method is called from another thread, it is added to a `_resourcesToBeGotten` queue that needs to be processed through a call to the `checkAsynchronousGets` slot.

hifi-gustavo:
Build console output for commit: 801862cde1f0383fcf4153f1679e6425464a0547 available at: https://gist.github.com/cf537a101ffbab4dcc7ebb072e66b9a4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7549.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7549.dmg

zzmp:
@kevinmthomas-carpool the top line in the plots shows the scale of the plot, not the total. It should slowly fall to match the highest value as you leave it.

Cached is a subset of total, so you should only expect that the top line > total > cached.

Also, the numbers in cacheStats.js may be off from renderStats.js because they pull their data from different places. cacheStats looks at the assumed texture size, calculated before the texture is created on the GPU. renderStats looks at the actual texture size, pulled from the created texture. Eventually we need to make them accurate for better caching, but for now they will probably be slightly off.

hifi-gustavo:
Build console output for commit: 4ae8a4684dad07c365409e97ca6e5f4e44e6e985 available at: https://gist.github.com/cf537a101ffbab4dcc7ebb072e66b9a4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7549.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7549.dmg

kevinmthomas-carpool:
@zzmp Thanks Zach.  With that said the results are expected.

zzmp:
:) Thanks for testing @kevinmthomas-carpool

hifi-gustavo:
Build console output for commit: 107cfcf10b5acf675a9fb5102b25c737f29663d2 available at: https://gist.github.com/cf537a101ffbab4dcc7ebb072e66b9a4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7549.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7549.dmg

PR #7550, opened 2016-04-03
Model: fixed two crash issues when changing avatars

hyperlogic:
* When the GeometryReader has the last ref to the GeometryResource ptr
  It needs to hold on to the reference until invokeMethod is completed.
  Otherwise, invokeMethod will call a method on a deleted object, leading
  to memory corruption or crashes.

* When the Model URL is changed, the clusterMatrices are invalided and the
  RenderItemsSets are cleared.  However, there still might be renderItems in
  the scene pending changes list that might refer to those RenderItems and their
  clusterMatrices.  We need to guard against this access to prevent reading from
  memory that was previously freed.

Both of these issues were uncovered using the [avatar-thrasher](https://gist.github.com/hyperlogic/d82a61d141df43d576428501a82c5ee6) test script.

### DIFF
--- a/libraries/model-networking/src/model-networking/ModelCache.cpp
+++ b/libraries/model-networking/src/model-networking/ModelCache.cpp
@@ -144,10 +144,11 @@ void GeometryReader::run() {
                 throw QString("unsupported format");
             }
 
-            // Ensure the resource has not been deleted, and won't be while invoke method is in flight.
+            // Ensure the resource has not been deleted, and won't be while invokeMethod is in flight.
             auto resource = _resource.toStrongRef();
             if (!resource) {
                 qCWarning(modelnetworking) << "Abandoning load of" << _url << "; could not get strong ref";
+                delete fbxGeometry;
             } else {
                 QMetaObject::invokeMethod(resource.data(), "setGeometryDefinition", Qt::BlockingQueuedConnection, Q_ARG(void*, fbxGeometry));
             }
@@ -159,7 +160,7 @@ void GeometryReader::run() {
         qCDebug(modelnetworking) << "Error reading " << _url << ": " << error;
 
         auto resource = _resource.toStrongRef();
-        // Ensure the resoruce has not been deleted, and won't be while invoke method is in flight.
+        // Ensure the resoruce has not been deleted, and won't be while invokeMethod is in flight.
         if (!resource) {
             qCWarning(modelnetworking) << "Abandoning load of" << _url << "; could not get strong ref";
         } else {
--- a/libraries/model-networking/src/model-networking/ModelCache.cpp
+++ b/libraries/model-networking/src/model-networking/ModelCache.cpp
@@ -118,10 +118,8 @@ void GeometryReader::run() {
     }
     QThread::currentThread()->setPriority(QThread::LowPriority);
 
-    // Ensure the resource is still being requested
-    auto resource = _resource.toStrongRef();
-    if (!resource) {
-        qCWarning(modelnetworking) << "Abandoning load of" << _url << "; could not get strong ref";
+    if (!_resource.data()) {
+        qCWarning(modelnetworking) << "Abandoning load of" << _url << "; resource was deleted";
         return;
     }
 
@@ -146,14 +144,27 @@ void GeometryReader::run() {
                 throw QString("unsupported format");
             }
 
-            QMetaObject::invokeMethod(resource.data(), "setGeometryDefinition",
-                Q_ARG(void*, fbxGeometry));
+            // Ensure the resource has not been deleted, and won't be while invoke method is in flight.
+            auto resource = _resource.toStrongRef();
+            if (!resource) {
+                qCWarning(modelnetworking) << "Abandoning load of" << _url << "; could not get strong ref";
+            } else {
+                QMetaObject::invokeMethod(resource.data(), "setGeometryDefinition", Qt::BlockingQueuedConnection, Q_ARG(void*, fbxGeometry));
+            }
         } else {
             throw QString("url is invalid");
         }
     } catch (const QString& error) {
+
         qCDebug(modelnetworking) << "Error reading " << _url << ": " << error;
-        QMetaObject::invokeMethod(resource.data(), "finishedLoading", Q_ARG(bool, false));
+
+        auto resource = _resource.toStrongRef();
+        // Ensure the resoruce has not been deleted, and won't be while invoke method is in flight.
+        if (!resource) {
+            qCWarning(modelnetworking) << "Abandoning load of" << _url << "; could not get strong ref";
+        } else {
+            QMetaObject::invokeMethod(resource.data(), "finishedLoading", Qt::BlockingQueuedConnection, Q_ARG(bool, false));
+        }
     }
 
     QThread::currentThread()->setPriority(originalPriority);
--- a/libraries/render-utils/src/Model.cpp
+++ b/libraries/render-utils/src/Model.cpp
@@ -159,11 +159,13 @@ void Model::updateRenderItems() {
         Transform collisionMeshOffset;
         collisionMeshOffset.postTranslate(self->_offset);
 
+        uint32_t deleteGeometryCounter = self->_deleteGeometryCounter;
+
         render::PendingChanges pendingChanges;
         foreach (auto itemID, self->_modelMeshRenderItems.keys()) {
-            pendingChanges.updateItem<ModelMeshPartPayload>(itemID, [modelTransform, modelMeshOffset](ModelMeshPartPayload& data) {
+            pendingChanges.updateItem<ModelMeshPartPayload>(itemID, [modelTransform, modelMeshOffset, deleteGeometryCounter](ModelMeshPartPayload& data) {
                 // Ensure the model geometry was not reset between frames
-                if (data._model->isLoaded()) {
+                if (data._model && data._model->isLoaded() && deleteGeometryCounter == data._model->_deleteGeometryCounter) {
                     // lazy update of cluster matrices used for rendering.  We need to update them here, so we can correctly update the bounding box.
                     data._model->updateClusterMatrices(modelTransform.getTranslation(), modelTransform.getRotation());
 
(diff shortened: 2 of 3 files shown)


hyperlogic:
https://app.asana.com/0/45663370065758/107506113269077

hifi-gustavo:
Build console output for commit: 865a77ae2016c171287d0f65e5b3823f77beef49 available at: https://gist.github.com/449cc52a8c5636b1cd394ac8095b30a6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7550.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7550.dmg

libraries/render-utils/src/Model.cpp
```diff
@@ -159,11 +159,13 @@ void Model::updateRenderItems() {
...
 
+        uint32_t deleteGeometryCounter = self->_deleteGeometryCounter;
+
         render::PendingChanges pendingChanges;
         foreach (auto itemID, self->_modelMeshRenderItems.keys()) {
-            pendingChanges.updateItem<ModelMeshPartPayload>(itemID, [modelTransform, modelMeshOffset](ModelMeshPartPayload& data) {
+            pendingChanges.updateItem<ModelMeshPartPayload>(itemID, [modelTransform, modelMeshOffset, deleteGeometryCounter](ModelMeshPartPayload& data) {
                 // Ensure the model geometry was not reset between frames
-                if (data._model->isLoaded()) {
+                if (data._model && data._model->isLoaded() && deleteGeometryCounter == data._model->_deleteGeometryCounter) {
```
zzmp: I don't know if it is possible for `!data._model` in this codepath, as I don't think we ever clear the member once it is set, but I like the guard.

libraries/render-utils/src/Model.cpp
```diff
@@ -159,11 +159,13 @@ void Model::updateRenderItems() {
         Transform collisionMeshOffset;
         collisionMeshOffset.postTranslate(self->_offset);
 
+        uint32_t deleteGeometryCounter = self->_deleteGeometryCounter;
```
zzmp: Could this instead be a `bool` that is set to `true` when the functor is created and checked that it is still `true` from the reference to the `data._model` when the functor runs?
Is there an advantage to using `uint32_t`?

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -146,14 +144,27 @@ void GeometryReader::run() {
                 throw QString("unsupported format");
             }
 
-            QMetaObject::invokeMethod(resource.data(), "setGeometryDefinition",
-                Q_ARG(void*, fbxGeometry));
+            // Ensure the resource has not been deleted, and won't be while invoke method is in flight.
+            auto resource = _resource.toStrongRef();
```
zzmp: How is this different than generating the strong ref at line 122?

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -146,14 +144,27 @@ void GeometryReader::run() {
...
+            if (!resource) {
+                qCWarning(modelnetworking) << "Abandoning load of" << _url << "; could not get strong ref";
+            } else {
+                QMetaObject::invokeMethod(resource.data(), "setGeometryDefinition", Qt::BlockingQueuedConnection, Q_ARG(void*, fbxGeometry));
+            }
         } else {
             throw QString("url is invalid");
         }
     } catch (const QString& error) {
```
zzmp: Should `FBXGeometry* fbxGeometry` be declared before the `try` block so it can be `delete`d if we end up in the `catch` and it is non-null?

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -146,14 +144,27 @@ void GeometryReader::run() {
...
             }
 
-            QMetaObject::invokeMethod(resource.data(), "setGeometryDefinition",
-                Q_ARG(void*, fbxGeometry));
+            // Ensure the resource has not been deleted, and won't be while invoke method is in flight.
+            auto resource = _resource.toStrongRef();
+            if (!resource) {
+                qCWarning(modelnetworking) << "Abandoning load of" << _url << "; could not get strong ref";
+            } else {
+                QMetaObject::invokeMethod(resource.data(), "setGeometryDefinition", Qt::BlockingQueuedConnection, Q_ARG(void*, fbxGeometry));
```
zzmp: Should we be worried about performance from making this into a blocking operation?

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -146,14 +144,27 @@ void GeometryReader::run() {
                 throw QString("unsupported format");
             }
 
-            QMetaObject::invokeMethod(resource.data(), "setGeometryDefinition",
-                Q_ARG(void*, fbxGeometry));
+            // Ensure the resource has not been deleted, and won't be while invoke method is in flight.
+            auto resource = _resource.toStrongRef();
+            if (!resource) {
+                qCWarning(modelnetworking) << "Abandoning load of" << _url << "; could not get strong ref";
```
zzmp: If there is no resource we need to `delete` the `fbxGeometry`.

zzmp:
I think the same bug exists in the `ImageReader` `QThread` in `TextureCache.cpp`. It uses `Qt::AutoConnection` instead of `Qt::BlockingConnection`, but I think it should be fixed in a different PR so if any new issues arise we can more easily trace them.

kevinmthomas-carpool:
Successful smoke test on a Windows 10 machine.  Tested in the Sandbox domain to which I was also able to successfully manipulate an existing model and as a control did the exact operation at the Music domain to which both behaviors were expected and the same.  Final test was to add an item from the ATP and manipulate it.  Test was successful and behaved normally.

zzmp:
I just realized the resource readers (model & texture) using `Qt::AutoConnection` instead of `Qt::BlockingConnection` probably manifest as `Qt5Core` crashes, which are all over BugSplat (@ZappoMan)!

Example: I crashed using Tony's avatar-thrashing script and generated https://www.bugsplatsoftware.com/individualCrash/?id=2589&database=interface_alpha.

libraries/render-utils/src/Model.cpp
hyperlogic: Yeah, I'm just being paranoid at this point, but it's a pretty cheap check, so I added it.

libraries/render-utils/src/Model.cpp
hyperlogic: The counter has the advantage of never having to set it to false.  Performance wise it's the same as the bool.  Also, if we ever change the behavior and allow multiple updateItems to be queued up, the counter has the advantage of knowing exactly which lambda matches up with the current state of the model, with a boolean there's no way of knowing that.

libraries/model-networking/src/model-networking/ModelCache.cpp
hyperlogic: It's not all that different, however, I wanted to add another deletion check right after we do the heavy lifting of parsing the fbx file.  If the resource request is canceled, it would be likely done while the fbx file is parsing, in that case we won't call invoke method at all, where in the previous case we would.

The main fix here is to use the `Qt:BlockingQueuedConnection`, if the strong ref ends up being the last reference then the GeometryReader will delete the resource before invokeMethod has a chance to be called on the main thread.  Adding the blocking will prevent the main thread from calling a method on a deleted object.

libraries/model-networking/src/model-networking/ModelCache.cpp
hyperlogic: Yeah, i see your point, we could.  However, in practice I've never actually seen an exception in the FBXReader.  I guess, right before beta, I was shooting for a minimal diff, instead of changing the structure and inadvertently adding more bugs.

libraries/model-networking/src/model-networking/ModelCache.cpp
hyperlogic: Good catch!

libraries/model-networking/src/model-networking/ModelCache.cpp
hyperlogic: I don't think so, the GeometryReader thread will block on a semaphore. 

```
        QSemaphore semaphore;
        QCoreApplication::postEvent(object, new QMetaCallEvent(idx_offset, idx_relative, callFunction,
                                                        0, -1, 0, 0, param, &semaphore));
        semaphore.acquire();```

**Comment by hifi-gustavo** [comment]
Build console output for commit: 07cfa13a8249599f483c432eec06ec440eec2795 available at: https://gist.github.com/449cc52a8c5636b1cd394ac8095b30a6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7550.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7550.dmg

File: libraries/model-networking/src/model-networking/ModelCache.cpp [hunk 7]
```diff
@@ -146,14 +144,28 @@ void GeometryReader::run() {
...
+            }
         } else {
             throw QString("url is invalid");
         }
     } catch (const QString& error) {
+
         qCDebug(modelnetworking) << "Error reading " << _url << ": " << error;
-        QMetaObject::invokeMethod(resource.data(), "finishedLoading", Q_ARG(bool, false));
+
+        auto resource = _resource.toStrongRef();
```
zzmp: Nit - on line 147, the comment is before the `toStrongRef`.

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -118,10 +118,8 @@ void GeometryReader::run() {
     }
     QThread::currentThread()->setPriority(QThread::LowPriority);
 
-    // Ensure the resource is still being requested
-    auto resource = _resource.toStrongRef();
-    if (!resource) {
-        qCWarning(modelnetworking) << "Abandoning load of" << _url << "; could not get strong ref";
+    if (!_resource.data()) {
+        qCWarning(modelnetworking) << "Abandoning load of" << _url << "; resource was deleted";
```
zzmp: We need to add a reset to the thread priority if we return early, as in line 159. That, or move this `_resource.data` check to before we switch priorities.

zzmp:
@hyperlogic could you look at #7553? It takes this change and applies it to the `ImageReader` for textures.

PR #7551, opened 2016-04-03
Working on addressing stuttering in the frame while loading content - Part 1

jherico:
Testing has shown that large texture loading from images can cause the frame rate to stutter, which in turn can cause very bad effects in an HMD (though this is mitigated somewhat on the Oculus headset due to timewarp).

In order to combat this, the plan is to move the transfer of image data to the GPU from the primary rendering thread to a dedicated background thread.  This PR contains additional profiling used to identify the problem as well as some initial work to enable correcting it.

## Testing

This code only moves the GLTexelFormat class, and adds some additional profile ranges.  It should have no effect on the behavior of the app.


libraries/gl/src/gl/GLHelpers.cpp
```diff
@@ -12,7 +12,7 @@ const QSurfaceFormat& getDefaultOpenGLSurfaceFormat() {
         // Qt Quick may need a depth and stencil buffer. Always make sure these are available.
         format.setDepthBufferSize(DEFAULT_GL_DEPTH_BUFFER_BITS);
         format.setStencilBufferSize(DEFAULT_GL_STENCIL_BUFFER_BITS);
-        format.setVersion(4, 1);
+        format.setVersion(4, 5);
```
zzmp: This needs to be set [here](https://github.com/highfidelity/hifi/blob/1b52d700a0680671ad96aadb0610675185e7f576/libraries/gl/src/gl/Config.h#L22) as well so the proper version is checked for in OpenGLVersionChecker.

libraries/gpu/src/gpu/GLBackend.h
```diff
@@ -86,10 +86,21 @@ class GLBackend : public Backend {
...
         GLuint size() const { return _size; }
 
+        enum SyncState {
+            Idle,
+            Pending,
+            Transferred,
+        };
+
+        void setSyncState(SyncState syncState) { _syncState = syncState; }
+        SyncState getSyncState() const { return _syncState; }
```
zzmp: Should return a `const SyncState`.

hifi-gustavo:
Build console output for commit: f1d14d4d2e29f4e03a022349e6f68d8a5e626665 available at: https://gist.github.com/97d966f0ac5832a1aa1e531a4ece516f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7551.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7551.dmg

libraries/gpu/src/gpu/GLBackend.h
```diff
@@ -241,6 +252,12 @@ class GLBackend : public Backend {
     void renderPassTransfer(Batch& batch);
     void renderPassDraw(Batch& batch);
 
+    void initTextureTransferThread();
+    static void transferGPUObject(const TexturePointer& texture);
+
+    class TextureTransferThread;
+    static std::shared_ptr<TextureTransferThread> _textureTransferThread;
```
zzmp: Why does this need to be behind a `shared_ptr`?

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -44,20 +48,20 @@ class GLTexelFormat {
 
     static GLTexelFormat evalGLTexelFormat(const Element& dstFormat, const Element& srcFormat) {
         if (dstFormat != srcFormat) {
-            GLTexelFormat texel = {GL_RGBA, GL_RGBA, GL_UNSIGNED_BYTE};
+            GLTexelFormat texel = { GL_RGBA8, GL_RGBA, GL_UNSIGNED_BYTE };
```
zzmp: If any of this breaks build it will be conflated with the threaded changes. Can you make these changes in a separate PR so it is more isolated, in case something goes wrong?

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -399,194 +403,249 @@ class GLTexelFormat {
...
 
-    // need to have a gpu object?
-    if (!object) {
-        object = new GLTexture();
-        glGenTextures(1, &object->_texture);
-        (void) CHECK_GL_ERROR();
-        Backend::setGPUObject(texture, object);
+void transferTexture(const Texture& texture) {
+    PROFILE_RANGE(__FUNCTION__);
+    qDebug() << "Transferring texture";
```
zzmp: Is there any information (`QUrl`?) about the texture you can include in the debug statement?

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -399,194 +403,249 @@ class GLTexelFormat {
...
+        initialize(true, QThread::LowPriority);
+        _canvas = std::make_shared<OffscreenGLCanvas>();
+        _canvas->create(QOpenGLContext::currentContext());
+        _canvas->moveToThread(_thread);
+        _canvas->getContext()->moveToThread(_thread);
+    }
+
+protected:
+    bool processQueueItems(const Queue& messages) override {
+        _canvas->makeCurrent();
```
zzmp: This only needs to be done once on the thread, as the thread's context never changes and it has its own context. Can it be done in a `run` method?

zzmp: Check the return value to make sure this worked, and log something if not.

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -399,194 +403,249 @@ class GLTexelFormat {
...
+            }
+            transferTexture(*texturePointer);
         }
-        break;
+        return true;
     }
-    default:
-        qCDebug(gpulogging) << "GLBackend::syncGPUObject(const Texture&) case for Texture Type " << texture.getType() << " not supported";    
+
+    std::shared_ptr<OffscreenGLCanvas> _canvas;
```
zzmp: Why put this behind a `shared_ptr` if it is owned by the thread?

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -399,194 +403,249 @@ class GLTexelFormat {
...
-                    }
-                }
 
-                if (texture.isAutogenerateMips()) {
-                    glGenerateMipmap(GL_TEXTURE_CUBE_MAP);
-                    glTexParameteri(GL_TEXTURE_CUBE_MAP, GL_TEXTURE_MIN_FILTER, GL_LINEAR_MIPMAP_LINEAR);
-                }
+    // If the object hasn't been created, or the object definition is out of date, drop and re-create
+    GLTexture* object = Backend::getGPUObject<GLBackend::GLTexture>(texture);
+    if (object && object->_texture && GLTexture::Idle == object->getSyncState() &&
```
zzmp: Checks that object hasn't been created AND object is up-to-date. Will this ever be true?
Shouldn't it check that the object's state is `GLTexture::Transferred`?
Can you update the comment to reflect the conditional?

libraries/gpu/src/gpu/GLBackend.h
```diff
@@ -86,10 +86,21 @@ class GLBackend : public Backend {
         void setSize(GLuint size);
         GLuint size() const { return _size; }
 
+        enum SyncState {
```
zzmp: Can you add comments for these enums?
Specifically, can you note that `Idle` can mean either "Not yet transferred" or "Already transferred and updated", depending on the time stamps?

Not knowing this makes it confusing to read `syncGPUObject`.

libraries/gpu/src/gpu/GLBackend.h
jherico: so that the type `TextureTransferThread` can be forward declared

hifi-gustavo:
Build console output for commit: 56e58d5a9d8a1f57735457160836a0f4c8e99d1d available at: https://gist.github.com/97d966f0ac5832a1aa1e531a4ece516f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7551.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7551.dmg

libraries/gpu/src/gpu/GLBackend.h
jherico: returning a const only makes sense if the return type is a reference.

libraries/gpu/src/gpu/GLBackendTexture.cpp
jherico: I'll break this out into a distinct PR along with the performance measuring changes.

libraries/gpu/src/gpu/GLBackendTexture.cpp
jherico: I've added the texture ID which is all we really have at this point

libraries/gpu/src/gpu/GLBackendTexture.cpp
jherico: I've updated the code to only call makeCurrent once and fail if it doesn't succeed.

libraries/gpu/src/gpu/GLBackendTexture.cpp
jherico: in my latest commit I've moved this class into a header, and again, by making it a pointer type I can use only a forward declaration, to minimize the number of headers used in defining the class.

hifi-gustavo:
Build console output for commit: 38ac2aae10b26fe2e1e80465d893384601ac7462 available at: https://gist.github.com/97d966f0ac5832a1aa1e531a4ece516f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7551.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7551.dmg

hifi-gustavo:
Build console output for commit: d5d13204d29dad9c5793a182386fa8aee383b613 available at: https://gist.github.com/97d966f0ac5832a1aa1e531a4ece516f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7551.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7551.dmg

hifi-gustavo:
Build console output for commit: 515a9e3ebc695599ded6f2fc96196e89fdb34de5 available at: https://gist.github.com/97d966f0ac5832a1aa1e531a4ece516f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7551.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7551.dmg

hifi-gustavo:
Build console output for commit: 667c27866ed152019f2a5e0762aacc592d4fa747 available at: https://gist.github.com/97d966f0ac5832a1aa1e531a4ece516f<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7551.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7551.dmg

OmegaHeron:
Tested Windows 7 - I didn't see any difference between this and current production code and, that seems to be, about what I should expect according to description.

samcake:
I reviewed the code and i m good with it.

PR #7552, opened 2016-04-03
Avoid setting textures on empty models

zzmp:
- Check that textures exist to be set before setting them when taking models out of cache

---

#### Intent
Fixes a crash where the empty model (`QUrl()`) tried to set its textures when uncached (access violation).
(See: https://www.bugsplatsoftware.com/keycrash/?stackKeyId=394&database=interface_alpha.)

##### Repro
1. Upload a model with a collision hull.
2. Clear the collision hull through edit.js.
3. Add it back.
- This will not work if any other models are still using a null model (including set and unset collision hulls). Do this on a blank domain.

##### Expectations
In current build, you will crash. The callstack will show a crash in `GeometryResource::setTextures()`.
In this PR build, it should work fine.

### DIFF
--- a/libraries/model-networking/src/model-networking/ModelCache.cpp
+++ b/libraries/model-networking/src/model-networking/ModelCache.cpp
@@ -232,7 +232,7 @@ std::shared_ptr<NetworkGeometry> ModelCache::getGeometry(const QUrl& url, const 
     GeometryExtra geometryExtra = { mapping, textureBaseUrl };
     GeometryResource::Pointer resource = getResource(url, QUrl(), true, &geometryExtra).staticCast<GeometryResource>();
     if (resource) {
-        if (resource->isLoaded() && !resource->hasTextures()) {
+        if (resource->isLoaded() && resource->shouldSetTextures()) {
             resource->setTextures();
         }
         return std::make_shared<NetworkGeometry>(resource);
--- a/libraries/model-networking/src/model-networking/ModelCache.h
+++ b/libraries/model-networking/src/model-networking/ModelCache.h
@@ -107,7 +107,8 @@ protected:
     friend class GeometryMappingResource;
 
     // Geometries may not hold onto textures while cached - that is for the texture cache
-    bool hasTextures() const { return !_materials.empty(); }
+    // Instead, these methods clear and reset textures from the geometry when caching/loading
+    bool shouldSetTextures() const { return _geometry && _materials.empty(); }
     void setTextures();
     void resetTextures();
 


hifi-gustavo:
Build console output for commit: 2587cb3b848c4f0ddfd4d678ee077f5ee30182cb available at: https://gist.github.com/0134952d17972cec8cc9012cff08b2fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7552.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7552.dmg

kevinmthomas-carpool:
Tested on Windows 10 desktop machine, upon install of the PR, additionally installed the Sandbox to create the test environment for the blank domain to which is called for in the test plan.  Manually deleted all content to include the zone and ensured a completely blank domain by expanding the meter radius to 65535 to verify there was no content.  Successful smoke test as well.

Uploaded a model with a collision hull (kcommercialbuilding.fbx and respectively kcommercialbuilding10c.obj).  Manually cleared the collision hull through edit.js and upon remove a crash occurred.  This was NOT the expected result.

See attached files:
PR 7552 Interface55f16DQ7.dmp
PR 7552 hifi-log.txt 

An error report was also sent through the crash report system.

Tested utilizing the same procedures on a Windows 10 laptop machine.  Tested against build 4610 to which the idential steps and procedures above were taken and the result was a crash as expected.  

See attached files:
4610 Build Interface55f16DQ7.dmp
4610 Build hifi-log.txt 

An error report was also sent through the crash report system.

Recommendation is to have another user test and report to compare data as the results on the PR build were NOT expected.
[4610 Build Interface hifi-log.txt](https://github.com/highfidelity/hifi/files/201433/4610.Build.Interface.hifi-log.txt)


[PR 7552 hifi-log.txt](https://github.com/highfidelity/hifi/files/201434/PR.7552.hifi-log.txt)

zzmp:
@kevinmthomas-carpool I looked through my changes, and I was negating the boolean check that should have stopped the crash.

In other words - your test was accurate. I not only failed to fix it, but I ensured the crash would occur with my fix.

I pushed up an amended commit. It should actually fix the crash now. Could you test again?

hifi-gustavo:
Build console output for commit: fc19b60f8a247decb9f469717f7dd53046ae57e6 available at: https://gist.github.com/0134952d17972cec8cc9012cff08b2fc<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7552.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7552.dmg

kevinmthomas-carpool:
Uninstalled existing PR build and checked to ensure directory was removed.  Installed updated PR build based on amended commit.  Tested on Windows 10 desktop.  

Uploaded a model with a collision hull (kcommercialbuilding.fbx and respectively kcommercialbuilding10c.obj). Manually cleared the collision hull through edit.js and then manually re-added it.  No crash occurred and the behavior was expected.   Successful test.

PR #7553, opened 2016-04-03
Fix ImageReader threading issues

zzmp:
- Changes ImageReader to use a queued connection to avoid dropped reference
- Cleans up the exiting codepaths for ImageReader

---

#### Intent
@hyperlogic found that readers using `Qt::AutoConnection` could crash on access violations because the object on which they were invoking a method could be deleted, as the readers' strong ref expires before the invocation. #7550 fixes this for the `GeometryReader` - this is meant to mirror that change.

#### Testing

Test with the texture cache set to 0, and switch domains frequently, mid-loading.
Likely, you will crash in neither current nor PR build, as this fixes an edge-case threading race condition that may not even be exposed.


hifi-gustavo:
Build console output for commit: 022d08201de50faef4b1a605d33d66c74a503755 available at: https://gist.github.com/a767cd781d0b976bfecbd4eaa201b8fe<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7553.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7553.dmg

hifi-gustavo:
Build console output for commit: 5f1982125cf0d923614711480e6b67675097bc51 available at: https://gist.github.com/a767cd781d0b976bfecbd4eaa201b8fe<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7553.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7553.dmg

hifi-gustavo:
Build console output for commit: 79e528633584f111a2c9a9fcdec07e8b798838d0 available at: https://gist.github.com/a767cd781d0b976bfecbd4eaa201b8fe<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7553.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7553.dmg

kevinmthomas-carpool:
Testing Environment:  Windows 10 Desktop
Testing Location:  hifi://music
Smoke Test:  Successful
Test Results:  After 8 minutes of smoke testing, the domain content vanished to include all of the ATP assets and a new box entity that was created during the smoke test.  When I moved my avatar location the content re-appeared.  Upon leaving the avatar idle for another 3 minutes the content once again disappeared and again upon moving the avatar, the content reappeared.  This behavior was obviously unexpected.  I then left the avatar idle again for another 3-4 minutes and the aforementioned behavior did not occur in the same fashion.  I can deduce that the behavior was random.  There were no issues with the server count which was 5 and the render stats were normal compared to the release build of 4610.  The second test was to change avatars as the current avatar was the kavatar10.  I switched to the Albert avatar by selecting it in the Marketplace to which I and successfully completed a smoke test utilizing that avatar.  Switched back to the kavatar10 successfully as well.  There were no additional issues of any kind with any crashing or interruption of the experience while logged in.  Upon exiting the interface I did crash and send an error report. I have also attached the hifi-log.txt and InterfaceV6M90N76.dmp to this report for developer examination. 
[attachments.zip](https://github.com/highfidelity/hifi/files/201599/attachments.zip)

libraries/model-networking/src/model-networking/TextureCache.cpp
```diff
@@ -284,25 +284,25 @@ void ImageReader::run() {
     }
     QThread::currentThread()->setPriority(QThread::LowPriority);
 
-    auto texture = _texture.toStrongRef();
-    if (!texture) {
-        qCWarning(modelnetworking) << "Could not get strong ref";
+    if (!_resource.data()) {
+        qCWarning(modelnetworking) << "Abandoning load of" << _url << "; could not get strong ref";
+        QThread::currentThread()->setPriority(originalPriority);
```
jherico: FYI, there is a 'Finally' utility in the shared library that you can use to automatically execute a code block when it goes out of scope.

```
Finally restorePriority([&]{
    QThread::currentThread()->setPriority(originalPriority);
});
```

This enables you to do things without worrying about cleanup being missed because of an early return.

libraries/model-networking/src/model-networking/TextureCache.cpp
```diff
@@ -312,15 +312,31 @@ void ImageReader::run() {
...
+        auto resource = _resource.toStrongRef();
+        if (!resource) {
+            qCWarning(modelnetworking) << "Abandoning load of" << _url << "; could not get strong ref";
+            QThread::currentThread()->setPriority(originalPriority);
+            return;
+        }
+
+        auto url = _url.toString().toStdString();
+        texture = resource.dynamicCast<NetworkTexture>()->getTextureLoader()(image, url);
+    }
```
hyperlogic: I could be wrong, but there isn't much going on this thread between the first weakptr check and this check.  To make the code simpler, perhaps, we should just remove this toStrongRef() and just do the texture load without checking.  Leave the check right before we call invokeMethod tho.

zzmp: My assumption was that the two long operations were `QImage::fromData` and `NetworkTexture::getTextureLoader()`. I'm timing it right now to check.

hifi-gustavo:
Build console output for commit: 2bd53935fc77359fb292e42b267688873c2dec99 available at: https://gist.github.com/a767cd781d0b976bfecbd4eaa201b8fe<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7553.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7553.dmg

libraries/model-networking/src/model-networking/TextureCache.cpp
zzmp: I also must do the check before the texture load because I need the NetworkTexture to get the loader.

zzmp: I'm seeing times between 6 and 1434ms, so I'd like to keep all the checks in.

zzmp: On the largest texture in playa, I saw 1.2s for QImage::fromData and 0.7s for the textureLoader (and 0.014s for the setImage call).

zzmp:
@kevinmthomas-carpool were you having network issues while you tested this? That may have caused your scene to reset?

OmegaHeron:
What's a good test for this @zzmp - does avatarThrasher.js torture it or something else?

zzmp:
Ooo, I forgot to update that.

I could not find a good test :(. I believe the bug exists, because this code is very similar to the model loading code, and the bug could be induced there with avatarThrasher.js, but that should not work here as the texture is loaded.

You could *try* it with a texture cache size set to 0, but I think you may have to just do smoke testing.

---

The bug that this tries to fix is that a texture going out of scope mid-load should crash the system. However, that will only happen if it goes out of scope while it is loading, and the thread that it lives on is busy when the loading finishes, and the destructor (from the loading) runs before the setImage (from the loading) finishes. It is an edge-case threading race condition. I could not induce it.

OmegaHeron:
Ok - I'm going to set cache to 0, poke around a bit and run avatar thrasher while doing other things - if anything will kill it all that will.  :)

kevinmthomas-carpool:
@zzmp That is probably the case.  I have tracked issues with some dropouts from time to time.

OmegaHeron:
I've been running avatarThrasher.js with texture cache ram set to 0 and moving between domains for (lost track - maybe hypnotized by AV thrasher) 10-15 minutes.  No issues seen - recorded one crash at Playa but that's almost a certainty for me if I have texture cache at zero.  It pushes right up to my 2GB GPU RAM and add a AV or 2 it goes over rapidly.

PR #7554, opened 2016-04-04
Addressing stuttering in the frame while loading content - Part 2

jherico:
This moves the job of transferring of image data across the CPU/GPU boundary to a background thread / rendering context.

- MUST test on OSX to verify the legacy code path

### DIFF
--- a/libraries/gl/src/gl/GLHelpers.cpp
+++ b/libraries/gl/src/gl/GLHelpers.cpp
@@ -12,7 +12,7 @@ const QSurfaceFormat& getDefaultOpenGLSurfaceFormat() {
         // Qt Quick may need a depth and stencil buffer. Always make sure these are available.
         format.setDepthBufferSize(DEFAULT_GL_DEPTH_BUFFER_BITS);
         format.setStencilBufferSize(DEFAULT_GL_STENCIL_BUFFER_BITS);
-        format.setVersion(4, 1);
+        format.setVersion(4, 5);
 #ifdef DEBUG
         format.setOption(QSurfaceFormat::DebugContext);
 #endif
@@ -27,7 +27,7 @@ const QGLFormat& getDefaultGLFormat() {
     static QGLFormat glFormat;
     static std::once_flag once;
     std::call_once(once, [] {
-        glFormat.setVersion(4, 1);
+        glFormat.setVersion(4, 5);
         glFormat.setProfile(QGLFormat::CoreProfile); // Requires >=Qt-4.8.0
         glFormat.setSampleBuffers(false);
         glFormat.setDepth(false);
--- a/libraries/gpu/src/gpu/GLBackend.h
+++ b/libraries/gpu/src/gpu/GLBackend.h
@@ -24,6 +24,8 @@
 
 namespace gpu {
 
+class GLTextureTransferHelper;
+
 class GLBackend : public Backend {
 
     // Context Backend static interface required
@@ -35,7 +37,6 @@ class GLBackend : public Backend {
     explicit GLBackend(bool syncCache);
     GLBackend();
 public:
-
     virtual ~GLBackend();
 
     virtual void render(Batch& batch);
@@ -75,25 +76,63 @@ public:
 
     class GLTexture : public GPUObject {
     public:
-        Stamp _storageStamp;
-        Stamp _contentStamp;
-        GLuint _texture;
-        GLenum _target;
+        const Stamp _storageStamp;
+        Stamp _contentStamp { 0 };
+        const GLuint _texture;
+        const GLenum _target;
 
-        GLTexture();
+        GLTexture(const gpu::Texture& gpuTexture);
         ~GLTexture();
 
-        void setSize(GLuint size);
         GLuint size() const { return _size; }
 
+        enum SyncState {
+            // The texture is currently undergoing no processing, although it's content
+            // may be out of date, or it's storage may be invalid relative to the 
+            // owning GPU texture
+            Idle,
+            // The texture has been queued for transfer to the GPU
+            Pending,
+            // The texture has been transferred to the GPU, but is awaiting
+            // any post transfer operations that may need to occur on the 
+            // primary rendering thread
+            Transferred,
+        };
+
+        void setSyncState(SyncState syncState) { _syncState = syncState; }
+        SyncState getSyncState() const { return _syncState; }
+
+        // Is the storage out of date relative to the gpu texture?
+        bool invalid() const;
+
+        // Is the content out of date relative to the gpu texture?
+        bool outdated() const;
+
+        // Is the texture in a state where it can be rendered with no work?
+        bool ready() const;
+
+        // Move the image bits from the CPU to the GPU
+        void transfer() const;
+
+        // Execute any post-move operations that must occur only on the main thread
+        void postTransfer();
+
+        static const size_t CUBE_NUM_FACES = 6;
+        static const GLenum CUBE_FACE_LAYOUT[6];
+
     private:
-        GLuint _size;
+        void transferMip(GLenum target, const Texture::PixelsPointer& mip) const;
+
+        const GLuint _size;
+        // The owning texture
+        const Texture& _gpuTexture;
+        std::atomic<SyncState> _syncState { SyncState::Idle };
     };
-    static GLTexture* syncGPUObject(const Texture& texture);
+    static GLTexture* syncGPUObject(const TexturePointer& texture);
     static GLuint getTextureID(const TexturePointer& texture, bool sync = true);
 
     // very specific for now
-    static void syncSampler(const Sampler& sampler, Texture::Type type, GLTexture* object);
+    static void syncSampler(const Sampler& sampler, Texture::Type type, const GLTexture* object);
 
     class GLShader : public GPUObject {
     public:
@@ -240,6 +279,11 @@ public:
 protected:
     void renderPassTransfer(Batch& batch);
     void renderPassDraw(Batch& batch);
+
+    void initTextureTransferHelper();
+    static void transferGPUObject(const TexturePointer& texture);
+
+    static std::shared_ptr<GLTextureTransferHelper> _textureTransferHelper;
 
     // Draw Stage
     void do_draw(Batch& batch, size_t paramOffset);
@@ -484,6 +528,7 @@ protected:
 
     typedef void (GLBackend::*CommandCall)(Batch&, size_t);
     static CommandCall _commandCalls[Batch::NUM_COMMANDS];
+
 };
 
 };
--- a/libraries/gpu/src/gpu/GLBackend.h
+++ b/libraries/gpu/src/gpu/GLBackend.h
@@ -103,13 +103,13 @@ public:
         SyncState getSyncState() const { return _syncState; }
 
         // Is the storage out of date relative to the gpu texture?
-        bool invalid() const;
+        bool isInvalid() const;
 
         // Is the content out of date relative to the gpu texture?
-        bool outdated() const;
+        bool isOutdated() const;
 
         // Is the texture in a state where it can be rendered with no work?
-        bool ready() const;
+        bool isReady() const;
 
         // Move the image bits from the CPU to the GPU
         void transfer() const;
--- a/libraries/gpu/src/gpu/GLBackendTexture.cpp
+++ b/libraries/gpu/src/gpu/GLBackendTexture.cpp
@@ -19,8 +19,6 @@
 using namespace gpu;
 
 GLenum gpuToGLTextureType(const Texture& texture) {
-    // If we get here, we need to allocate and or update the content of the texture
-    // or it's already being transferred
     switch (texture.getType()) {
     case Texture::TEX_2D:
         return GL_TEXTURE_2D;
@@ -94,17 +92,17 @@ GLBackend::GLTexture::~GLTexture() {
     Backend::decrementTextureGPUCount();
 }
 
-bool GLBackend::GLTexture::invalid() const {
+bool GLBackend::GLTexture::isInvalid() const {
     return _storageStamp < _gpuTexture.getStamp();
 }
 
-bool GLBackend::GLTexture::outdated() const {
+bool GLBackend::GLTexture::isOutdated() const {
     return _contentStamp < _gpuTexture.getDataStamp();
 }
 
-bool GLBackend::GLTexture::ready() const {
+bool GLBackend::GLTexture::isReady() const {
     // If we have an invalid texture, we're never ready
-    if (invalid()) {
+    if (isInvalid()) {
         return false;
     }
 
@@ -112,7 +110,7 @@ bool GLBackend::GLTexture::ready() const {
     // as a special case
     auto syncState = _syncState.load();
 
-    if (outdated()) {
+    if (isOutdated()) {
         return Pending == syncState;
     }
 
@@ -167,7 +165,7 @@ void GLBackend::GLTexture::transfer() const {
 
         case Texture::TEX_CUBE:
             // transfer pixels from each faces
-            for (int f = 0; f < CUBE_NUM_FACES; f++) {
+            for (uint8_t f = 0; f < CUBE_NUM_FACES; f++) {
                 if (_gpuTexture.isStoredMipFaceAvailable(0, f)) {
                     transferMip(CUBE_FACE_LAYOUT[f], _gpuTexture.accessStoredMipFace(0, f));
                 }
@@ -188,15 +186,14 @@ void GLBackend::GLTexture::transfer() const {
 // Do any post-transfer operations that might be required on the main context / rendering thread
 void GLBackend::GLTexture::postTransfer() {
     setSyncState(GLTexture::Idle);
+    // At this point the mip pixels have been loaded, we can notify the gpu texture to abandon it's memory
     switch (_gpuTexture.getType()) {
         case Texture::TEX_2D:
-            // At this point the mip piels have been loaded, we can notify
             _gpuTexture.notifyMipFaceGPULoaded(0, 0);
             break;
 
         case Texture::TEX_CUBE:
             for (uint8_t f = 0; f < CUBE_NUM_FACES; ++f) {
-                // At this point the mip piels have been loaded, we can notify
                 _gpuTexture.notifyMipFaceGPULoaded(0, f);
             }
             break;
@@ -216,7 +213,7 @@ GLBackend::GLTexture* GLBackend::syncGPUObject(const TexturePointer& texturePoin
 
     // If the object hasn't been created, or the object definition is out of date, drop and re-create
     GLTexture* object = Backend::getGPUObject<GLBackend::GLTexture>(texture);
-    if (object && object->ready()) {
+    if (object && object->isReady()) {
         return object;
     }
 
@@ -224,7 +221,7 @@ GLBackend::GLTexture* GLBackend::syncGPUObject(const TexturePointer& texturePoin
     
     // Create the texture if need be (force re-creation if the storage stamp changes
     // for easier use of immutable storage)
-    if (!object || object->invalid()) {
+    if (!object || object->isInvalid()) {
         // This automatically destroys the old texture
         object = new GLTexture(texture);
     }
@@ -236,7 +233,7 @@ GLBackend::GLTexture* GLBackend::syncGPUObject(const TexturePointer& texturePoin
 
     // Object might be outdated, if so, start the transfer
     // (outdated objects that are already in transfer will have reported 'true' for ready()
-    if (object->outdated()) {
+    if (object->isOutdated()) {
         _textureTransferHelper->transferTexture(texturePointer);
     }
 
--- a/libraries/gpu/src/gpu/GLBackendTexture.cpp
+++ b/libraries/gpu/src/gpu/GLBackendTexture.cpp
@@ -9,19 +9,81 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 #include "GPULogging.h"
+
+#include <QtCore/QThread>
+
 #include "GLBackendShared.h"
 #include "GLTexelFormat.h"
+#include "GLBackendTextureTransfer.h"
 
 using namespace gpu;
 
-GLBackend::GLTexture::GLTexture() :
-    _storageStamp(0),
-    _contentStamp(0),
-    _texture(0),
-    _target(GL_TEXTURE_2D),
-    _size(0)
+GLenum gpuToGLTextureType(const Texture& texture) {
+    // If we get here, we need to allocate and or update the content of the texture
+    // or it's already being transferred
+    switch (texture.getType()) {
+    case Texture::TEX_2D:
+        return GL_TEXTURE_2D;
+        break;
+
+    case Texture::TEX_CUBE:
+        return GL_TEXTURE_CUBE_MAP;
+        break;
+
+    default:
+        qFatal("Unsupported texture type");
+    }
+    Q_UNREACHABLE();
+    return GL_TEXTURE_2D;
+}
+
+GLuint allocateSingleTexture() {
+    GLuint result;
+    glGenTextures(1, &result);
+    return result;
+}
+
+const GLenum GLBackend::GLTexture::CUBE_FACE_LAYOUT[6] = {
+    GL_TEXTURE_CUBE_MAP_POSITIVE_X, GL_TEXTURE_CUBE_MAP_NEGATIVE_X,
+    GL_TEXTURE_CUBE_MAP_POSITIVE_Y, GL_TEXTURE_CUBE_MAP_NEGATIVE_Y,
+    GL_TEXTURE_CUBE_MAP_POSITIVE_Z, GL_TEXTURE_CUBE_MAP_NEGATIVE_Z
+};
+
+// Create the texture and allocate storage
+GLBackend::GLTexture::GLTexture(const Texture& texture) : 
+    _storageStamp(texture.getStamp()), _texture(allocateSingleTexture()), 
+    _target(gpuToGLTextureType(texture)), _size((GLuint)texture.getSize()), _gpuTexture(texture) 
 {
     Backend::incrementTextureGPUCount();
+    Backend::updateTextureGPUMemoryUsage(0, _size);
+    Backend::setGPUObject(texture, this);
+
+    GLsizei width = texture.getWidth();
+    GLsizei height = texture.getHeight();
+    GLsizei levels = 1;
+    if (texture.maxMip() > 0) {
+        if (texture.isAutogenerateMips()) {
+            while ((width | height) >> levels) {
+                ++levels;
+            }
+        }
+        levels = std::max(1, std::min(texture.maxMip() + 1, levels));
+    }
+
+    GLTexelFormat texelFormat = GLTexelFormat::evalGLTexelFormat(texture.getTexelFormat());
+    withPreservedTexture(_target, [&] {
+        glBindTexture(_target, _texture);
+        (void)CHECK_GL_ERROR();
+        // GO through the process of allocating the correct storage 
+        if (GLEW_VERSION_4_2) {
+            glTexStorage2D(_target, levels, texelFormat.internalFormat, width, height);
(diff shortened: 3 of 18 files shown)


hifi-gustavo:
Build console output for commit: 61ea40f15636ddfacbfe1be6e7ad9b80c2108616 available at: https://gist.github.com/2539d6125f3a5ccb5cef351a0d835a2d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7554.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7554.dmg

kevinmthomas-carpool:
PR Build: 7554
Testing OS:  Windows 10
Testing Location:  hifi://music, hifi://sandbox, hifi://playa, hifi://free
Smoke Test:  Successful
Test Results:  Smoke test results successful and expected.  Successful import and manipulation of an fbx model to which the renderstats.js was closely monitored.  At location hifi://sandbox, values before upload were: [Num Buffers - CPU 2868, GPU 999; gpu::Buffer Memory 118.5 Mb, CPU 106.7 Mb, GPU 93.7 Mb; Num Textures 517, CPU 227, GPU 194; gpu::Texture Memory 1342.9 Mb, CPU 223.8 Mb, GPU 1208.6 Mb].  Values after fbx import: [Num Buffers - CPU 2890, GPU 1009; gpu::Buffer Memory 118.3 Mb, CPU 106.5 Mb, GPU 93.7 Mb; Num Textures 928, CPU 232, GPU 199; gpu::Texture Memory 1362.9 Mb, CPU 223.8 Mb, GPU 1226.7 Mb].  Successful deletion of fbx after test.  There were no effects on quality of experience of any kind.  While logged in, successful test and results were expected.  Upon logout of the interface a crash occurred.  Interface dmp and hifi-log.txt are attached for deeper inspection.
[attachments.zip](https://github.com/highfidelity/hifi/files/202304/attachments.zip)

libraries/gl/src/gl/GLHelpers.cpp
```diff
@@ -27,7 +27,7 @@ const QGLFormat& getDefaultGLFormat() {
     static QGLFormat glFormat;
     static std::once_flag once;
     std::call_once(once, [] {
-        glFormat.setVersion(4, 1);
+        glFormat.setVersion(4, 5);
```
samcake: I assume you want to make the version platform dependant?
we can stick to 4.1 right now

libraries/gpu/src/gpu/GLBackendTextureTransfer.cpp
```diff
@@ -0,0 +1,178 @@
...
+        // Texture no longer exists, move on to the next
+        if (!texturePointer) {
+            continue;
+        }
+        transferTextureSynchronous(*texturePointer);
+        GLBackend::GLTexture* object = Backend::getGPUObject<GLBackend::GLTexture>(*texturePointer);
+
+        // Now force a draw using the texture
+        try {
+            switch (texturePointer->getType()) {
```
samcake: I understand that you want to draw in order to make the texture resident in physical video memory.
I wouldn't recommend we do that, because by calling draws on this context we are polluting the driver in this thread/context with draw commands.
This context should only do the transfer calls and no draws.
Also, the texture at this point is ready for gpu Physical memory and validated. maybe it  is not truely resident in physical memory but it s good to go to be used in the main render thread.
As we move to 4.5 we will be actually able to nail this texture resident with the  glTexStorage functions and do memory management if we want to.
But ultimately, we do not really need this texture physically resident until it s really used for a drawcall in the main render thread.

libraries/gl/src/gl/GLHelpers.cpp
jherico: as far as I can tell, the Qt OpenGL system will pick the minimum of highest OpenGL level available and the level you request... so on OSX, you'll get a 4.1 context, but on windows and a recent nvidia card you'll get 4.5  Inside the code, if we want to use a > 4.1 function we need to put it inside a test like this

```
if (GLEW_VERSION_4_2) {
  ... 4.2 path ...
} else {
  ... 4.1 path ...
}
```

Some of the optimizations we're going to want to make for HMD's are basically going to require functionality  greater than OpenGL 4.1 (most AZDO stuff requires 4.2 and higher).

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -32,199 +94,164 @@ GLBackend::GLTexture::~GLTexture() {
     Backend::decrementTextureGPUCount();
 }
 
-void GLBackend::GLTexture::setSize(GLuint size) {
-    Backend::updateTextureGPUMemoryUsage(_size, size);
-    _size = size;
+bool GLBackend::GLTexture::invalid() const {
```
samcake: Can you had isInvalid instead of just invalid ?

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -32,199 +94,164 @@ GLBackend::GLTexture::~GLTexture() {
...
-            return object;
-        } else {
-            // Need to update the content of the GPU object from the source sysmem of the texture
-            needUpdate = true;
-        }
-    } else if (!texture.isDefined()) {
-        // NO texture definition yet so let's avoid thinking
-        return nullptr;
-    }
+bool GLBackend::GLTexture::outdated() const {
```
samcake: Same here, isOutdated instead of outdated

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -32,199 +94,164 @@ GLBackend::GLTexture::~GLTexture() {
...
+    return _contentStamp < _gpuTexture.getDataStamp();
+}
 
-    // need to have a gpu object?
-    if (!object) {
-        object = new GLTexture();
-        glGenTextures(1, &object->_texture);
-        (void) CHECK_GL_ERROR();
-        Backend::setGPUObject(texture, object);
+bool GLBackend::GLTexture::ready() const {
```
samcake: i guess isReady

hifi-gustavo:
Build console output for commit: d98abbc7df5d05b72a29a5015c81445df4f89499 available at: https://gist.github.com/2539d6125f3a5ccb5cef351a0d835a2d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7554.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7554.dmg

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -32,199 +94,164 @@ GLBackend::GLTexture::~GLTexture() {
...
-            if (needUpdate) {
-                if (texture.isStoredMipFaceAvailable(0)) {
-                    Texture::PixelsPointer mip = texture.accessStoredMipFace(0);
-                    const GLvoid* bytes = mip->readData();
-                    Element srcFormat = mip->getFormat();
-                
-                    GLTexelFormat texelFormat = GLTexelFormat::evalGLTexelFormat(texture.getTexelFormat(), srcFormat);
-
-                    glBindTexture(GL_TEXTURE_2D, object->_texture);
-                    glTexSubImage2D(GL_TEXTURE_2D, 0,
```
samcake: YEs now i understand that we were wrong with the arguments for glTexSubImage.
I guess we actually never really used it.
all the textures are created and the data is transfered in the gllTexImage call

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -32,199 +94,164 @@ GLBackend::GLTexture::~GLTexture() {
...
-                    if (texture.isStoredMipFaceAvailable(0, f)) {
-                        Texture::PixelsPointer mipFace = texture.accessStoredMipFace(0, f);
-                        Element srcFormat = mipFace->getFormat();
-                        GLTexelFormat texelFormat = GLTexelFormat::evalGLTexelFormat(texture.getTexelFormat(), srcFormat);
+// Do any post-transfer operations that might be required on the main context / rendering thread
+void GLBackend::GLTexture::postTransfer() {
+    setSyncState(GLTexture::Idle);
+    switch (_gpuTexture.getType()) {
+        case Texture::TEX_2D:
+            // At this point the mip piels have been loaded, we can notify
```
samcake: pixels :)

libraries/gpu/src/gpu/GLBackendTextureTransfer.cpp
```diff
@@ -0,0 +1,109 @@
...
+    _canvas->doneCurrent();
+    initialize(true, QThread::LowPriority);
+    _canvas->moveToThreadWithContext(_thread);
+#endif
+}
+
+void GLTextureTransferHelper::transferTexture(const gpu::TexturePointer& texturePointer) {
+    GLBackend::GLTexture* object = Backend::getGPUObject<GLBackend::GLTexture>(*texturePointer);
+#ifdef THREADED_TEXTURE_TRANSFER
+    TextureTransferPackage package { texturePointer, glFenceSync(GL_SYNC_GPU_COMMANDS_COMPLETE, 0) };
```
samcake: I don't understand why you need the sync fence and flush here 
this is going to be on the main thread/context and we don;t need to flush the command flow there to actually do what you need to do in the transfer thread isn't it ?
the gttexture object is already created once you get here.

See more comments below related (at the glFinish line)

hifi-gustavo:
Build console output for commit: b6ed1e4184ae1ab96f466a6c2836728299ac1049 available at: https://gist.github.com/2539d6125f3a5ccb5cef351a0d835a2d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7554.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7554.dmg

libraries/gpu/src/gpu/GLBackendTextureTransfer.cpp
```diff
@@ -0,0 +1,109 @@
...
+        // Texture no longer exists, move on to the next
+        if (!texturePointer) {
+            continue;
+        }
+
+        GLBackend::GLTexture* object = Backend::getGPUObject<GLBackend::GLTexture>(*texturePointer);
+        object->transfer();
+
+        // Now force a draw using the texture
+        try {
```
samcake: For some reason my previous comment at this line  disappeared ?
I wouldn't recommend we do any draw command here.
Once the texture globject is defined correctly (in the GLTexture::transfer() call) then these extra draws only force the texture resident in physical video memory for this draw call.
As we move to 4.5 we can control that residency explicitely if we wanted to (doing our own memory mangement getting ready for vulkan;)) but here this is extra work on this thread which is not needed.
Also the fact that we do draw commands on this transfer thread/context is not ideal for the driver as it will have to sync up with the main context for this. Ideally we shouldn't do any draw command on this transfer.

libraries/gpu/src/gpu/GLBackendTextureTransfer.cpp
jherico: I'm going to work on a performance test suite for this change, and I'll test a version that does the draw call and one that doesn't.  However, since this code only gets triggered from a `syncGpuObject` call, I kind of thought it was implicit that the texture was actually being used at this point.  We don't call `syncGpuObject` unless we're about to use it, right?

libraries/gpu/src/gpu/GLBackendTextureTransfer.cpp
```diff
@@ -0,0 +1,109 @@
...
+                default:
+                    qCWarning(gpulogging) << __FUNCTION__ << " case for Texture Type " << texturePointer->getType() << " not supported";
+                    break;
+            }
+        } catch (const std::runtime_error& error) {
+            qWarning() << "Failed to render texture on background thread: " << error.what();
+        }
+
+        glBindTexture(object->_target, 0);
+        glFinish();
```
samcake: Follow up to my comment from the transferTexture call:

So here i would actually create a fence on the texture after the GLtexture::transfer() call is done.
Then you can have a glClientWaitSync on it wich would do somewhat like the glFinish but more accurate.
Alternatively,  you could put that job aside and come back to it a few milliseconds and do the glClientWaitSync to guarantee the texture is ready to go and unlock it and set its state to transferred.

libraries/gpu/src/gpu/GLBackendTextureTransfer.cpp
jherico: we've issued a `glTexStorage` (or `glTexImage` if storage isn't available) on the main context, but because we do the `glTexSubImage` to transfer the mip data on another context, we need to know that the storage / image command has actually executed.  The flush is required to ensure the `glFenceSync` command has been pushed to the driver before we try to start using it in another tread.

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -32,199 +94,164 @@ GLBackend::GLTexture::~GLTexture() {
...
-
-                    object->_contentStamp = texture.getDataStamp();
-                }
+//#define USE_PBO
+
+// Move content bits from the CPU to the GPU for a given mip / face
+void GLBackend::GLTexture::transferMip(GLenum target, const Texture::PixelsPointer& mip) const {
+    GLTexelFormat texelFormat = GLTexelFormat::evalGLTexelFormat(_gpuTexture.getTexelFormat(), mip->getFormat());
+#ifdef USE_PBO
+    GLuint pixelBufferID;
```
samcake: PBO is really useful when downloading the texels of a gltexture from the video memory or copying within 2 gltextures.
When you upload the sysmem texels to the gltexture, this isn;t any faster than just doing glTexSubImage

libraries/gpu/src/gpu/GLBackendTextureTransfer.cpp
jherico: I'll do A/B testing with and without the draw to see how it impacts performance.  The issue is that I've read that unless you force the use of a texture by executing a draw, the driver may defer the actual transfer until it's actually in use, hence the common practice of executing a draw to force the transfer.

ZappoMan:
This PR definitely seems to improve the loading behavior. I am now not seeing big spikes in frame render time when new content is being loaded.

libraries/gpu/src/gpu/GLBackendTexture.cpp
jherico: Yeah, I was reading up on an article in `OpenGL Insights` last night and came to pretty much the same conclusion.  I'd like to do A/B testing on it to make sure though.

libraries/gpu/src/gpu/GLBackendTextureTransfer.cpp
jherico: Yeah, I was just having trouble finding the right params for client wait sync, and glFinish was a rough and dirty way to get the job done.  I'll do a proper fix.

libraries/gpu/src/gpu/GLBackendTexture.cpp
jherico: So, considering that we never saw bad textures as a result of this, I'm guessing the 'needUpdate' code path was never followed, implying that we've never had a texture where we updated the content after having synced it.

libraries/gpu/src/gpu/GLBackendTextureTransfer.cpp
samcake: I agree with the logic of putting a fence.
I don't think the Flush is needed.
what i m really questioning though is whether we wouldn't be better of doing the creation and the transfer on the transfer thread for standard textures.
There is the case of the render target textures which must be created in sync.
But aside from these, for regular textures, we could avoid this extra synchronization with the main thread.

libraries/gpu/src/gpu/GLBackend.h
```diff
@@ -75,25 +76,63 @@ class GLBackend : public Backend {
...
+            // any post transfer operations that may need to occur on the 
+            // primary rendering thread
+            Transferred,
+        };
+
+        void setSyncState(SyncState syncState) { _syncState = syncState; }
+        SyncState getSyncState() const { return _syncState; }
+
+        // Is the storage out of date relative to the gpu texture?
+        bool invalid() const;
```
zzmp: isInvalid (same with outdated, ready)

libraries/gpu/src/gpu/GLBackend.h
```diff
@@ -75,25 +76,63 @@ class GLBackend : public Backend {
...
-        GLuint _size;
+        void transferMip(GLenum target, const Texture::PixelsPointer& mip) const;
+
+        const GLuint _size;
+        // The owning texture
+        const Texture& _gpuTexture;
+        std::atomic<SyncState> _syncState { SyncState::Idle };
     };
-    static GLTexture* syncGPUObject(const Texture& texture);
+    static GLTexture* syncGPUObject(const TexturePointer& texture);
```
zzmp: Why did you change to a pointer?

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -9,19 +9,81 @@
...
 using namespace gpu;
 
-GLBackend::GLTexture::GLTexture() :
-    _storageStamp(0),
-    _contentStamp(0),
-    _texture(0),
-    _target(GL_TEXTURE_2D),
-    _size(0)
+GLenum gpuToGLTextureType(const Texture& texture) {
+    // If we get here, we need to allocate and or update the content of the texture
```
zzmp: and/or

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -9,19 +9,81 @@
...
+const GLenum GLBackend::GLTexture::CUBE_FACE_LAYOUT[6] = {
+    GL_TEXTURE_CUBE_MAP_POSITIVE_X, GL_TEXTURE_CUBE_MAP_NEGATIVE_X,
+    GL_TEXTURE_CUBE_MAP_POSITIVE_Y, GL_TEXTURE_CUBE_MAP_NEGATIVE_Y,
+    GL_TEXTURE_CUBE_MAP_POSITIVE_Z, GL_TEXTURE_CUBE_MAP_NEGATIVE_Z
+};
+
+// Create the texture and allocate storage
+GLBackend::GLTexture::GLTexture(const Texture& texture) : 
+    _storageStamp(texture.getStamp()), _texture(allocateSingleTexture()), 
+    _target(gpuToGLTextureType(texture)), _size((GLuint)texture.getSize()), _gpuTexture(texture)
```
zzmp: Why do you use `getSize` instead of `getStoredSize` or `evalTotalSize`?

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -9,19 +9,81 @@
...
+    GLTexelFormat texelFormat = GLTexelFormat::evalGLTexelFormat(texture.getTexelFormat());
+    withPreservedTexture(_target, [&] {
+        glBindTexture(_target, _texture);
+        (void)CHECK_GL_ERROR();
+        // GO through the process of allocating the correct storage 
+        if (GLEW_VERSION_4_2) {
+            glTexStorage2D(_target, levels, texelFormat.internalFormat, width, height);
+        } else {
+            glTexImage2D(_target, 0, texelFormat.internalFormat, width, height, 0, texelFormat.format, texelFormat.type, 0);
+        }
```
zzmp: At this point do we know the actual `_size`, and should we update it here?

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -9,19 +9,81 @@
...
+    GL_TEXTURE_CUBE_MAP_POSITIVE_Z, GL_TEXTURE_CUBE_MAP_NEGATIVE_Z
+};
+
+// Create the texture and allocate storage
+GLBackend::GLTexture::GLTexture(const Texture& texture) : 
+    _storageStamp(texture.getStamp()), _texture(allocateSingleTexture()), 
+    _target(gpuToGLTextureType(texture)), _size((GLuint)texture.getSize()), _gpuTexture(texture) 
 {
     Backend::incrementTextureGPUCount();
+    Backend::updateTextureGPUMemoryUsage(0, _size);
```
zzmp: Isn't this memory in CPU until the transfer?

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -32,199 +94,164 @@ GLBackend::GLTexture::~GLTexture() {
...
+    setSyncState(GLTexture::Idle);
+    switch (_gpuTexture.getType()) {
+        case Texture::TEX_2D:
+            // At this point the mip piels have been loaded, we can notify
+            _gpuTexture.notifyMipFaceGPULoaded(0, 0);
+            break;
+
+        case Texture::TEX_CUBE:
+            for (uint8_t f = 0; f < CUBE_NUM_FACES; ++f) {
+                // At this point the mip piels have been loaded, we can notify
```
zzmp: pixels here too

jherico:
hifi-snap-by-jherico-on-2016-04-04_15-56-42

These are the testing results from production

jherico:
hifi-snap-by-jherico-on-2016-04-04_16-06-32

Testing results from this branch

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -9,19 +9,81 @@
...
+        }
+        levels = std::max(1, std::min(texture.maxMip() + 1, levels));
+    }
+
+    GLTexelFormat texelFormat = GLTexelFormat::evalGLTexelFormat(texture.getTexelFormat());
+    withPreservedTexture(_target, [&] {
+        glBindTexture(_target, _texture);
+        (void)CHECK_GL_ERROR();
+        // GO through the process of allocating the correct storage 
+        if (GLEW_VERSION_4_2) {
```
samcake: one more for the road, instead of relying on the GLEW macro, can you make one in Config.h that reflect that level of features ?

hifi-gustavo:
Build console output for commit: 0cadebd3f46912b24148a0955186f98a4b997be8 available at: https://gist.github.com/2539d6125f3a5ccb5cef351a0d835a2d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7554.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7554.dmg

libraries/gpu/src/gpu/GLBackendTexture.cpp
samcake: This is something that i will need to adjust in my PR doing the compression and fixing some of the counters.
I d say don;t worry about this one there.

libraries/gpu/src/gpu/GLBackendTexture.cpp
jherico: not really, because I call `glTexStorage2D` or `glTexImage2D` to allocate GPU side storage for the texture, even if I haven't transferred any pixels.  If I only incremented this value when we transfer, we wouldn't ever account for texture memory usage in framebuffers, because we never transfer anything for them.

libraries/gpu/src/gpu/GLBackendTexture.cpp
jherico: Because `getSize()` is what the original code used.  The only real change here is to move a bunch of stuff that used to be in the `syncGPUObject` method into the constructor for the `GLTexture` objects.  The logic should be the same.

libraries/gpu/src/gpu/GLBackend.h
jherico: Because the object needs to go into a queue of textures waiting to be transferred.  You can't copy a `Texture`, so you can't put it in a queue, but you can copy a `TexturePointer`, or in this case, take a weak pointer to it.

libraries/gpu/src/gpu/GLBackendTextureTransfer.cpp
jherico: The flush is definitely needed if you're going to be testing the fence on a different context.  I feel like we've had this conversation before.  See https://www.opengl.org/wiki/Sync_Object#Flushing_and_contexts

libraries/gpu/src/gpu/GLBackendTextureTransfer.cpp
jherico: removing the draw has no impact on performance, so I've ifdefed it out.

libraries/gpu/src/gpu/GLBackendTexture.cpp
```diff
@@ -190,7 +190,7 @@ void GLBackend::GLTexture::postTransfer() {
     setSyncState(GLTexture::Idle);
     switch (_gpuTexture.getType()) {
         case Texture::TEX_2D:
-            // At this point the mip piels have been loaded, we can notify
+            // At this point the mip pixels have been loaded, we can notify
```
zzmp: You missed the comment under this one.

libraries/gpu/src/gpu/GLBackendTexture.cpp
jherico: _size is const and is based on whatever the `gpu::Texture` originally said it was.  I don't want to change this behavior.

libraries/gpu/src/gpu/GLBackendTexture.cpp
jherico: fixed

libraries/gpu/src/gpu/GLBackendTexture.cpp
jherico: fixed

libraries/gpu/src/gpu/GLBackendTexture.cpp
jherico: fixed

libraries/gpu/src/gpu/GLBackendTextureTransfer.cpp
samcake: OK as we discussed good enough for now

hifi-gustavo:
Build console output for commit: c168e2cc5894be2d612b50f15d9bfebcb8e8de40 available at: https://gist.github.com/2539d6125f3a5ccb5cef351a0d835a2d<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7554.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7554.dmg

samcake:
Tested again now on PC going to the standard places, it definitely improve the experience coming into a new domain.
Great job!

libraries/gpu/src/gpu/GLBackendTexture.cpp
jherico: the GLEW thing isn't a macro, it's actually a boolean that's populated when we call `glewInit()`

jherico:
Quick note on the test images above.  I created a script which would:

* teleport me to Dreaming
* clear my texture cache 
* then teleport me to Playa
* begin recording paintGL durations
* rotate the camera for 30 seconds 
* stop recording paintGL durations
* graph the results

The graphing is of the maximum frame time for subset of all the frames rendered (if the graph is 250 pixels wide and I rendered 1000 frames, then each line would be the max paintGL time for 4 frames)

The Y axis is time to render a frame, with each pixel being 10 ms (or roughly the time budget for one frame when running a 90 fps display).  The green line is 50 ms, the yellow is 100 ms and the red is 1000 ms.  

With the current synchronous loading, we can clearly see that the texture transfers sometimes force the render time into the > 1 second area, meaning you're staring at an obviously frozen screen for that long.  There's clearly still room for improvement, but this should mitigate the worst of the problem.

ZappoMan:
Smoke tested on mac - went to playa... things worked as expected. I could see stuff. Pass on Mac.

PR #7563, opened 2016-04-04
Application: fix for jittery hands when moving.

hyperlogic:
The issue was that the hands are sampled in sensor coordinates.
But when the avatar is moved via physics, the sensor to world matrix
must be updated to move the hands in the world by the correct amount.
These new hand positions are then computed before IK is performed.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1470,10 +1470,6 @@ void Application::paintGL() {
     // update the avatar with a fresh HMD pose
     getMyAvatar()->updateFromHMDSensorMatrix(getHMDSensorPose());
 
-    // update sensorToWorldMatrix for camera and hand controllers
-    getMyAvatar()->updateSensorToWorldMatrix();
-
-
     auto lodManager = DependencyManager::get<LODManager>();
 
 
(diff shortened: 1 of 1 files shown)


hyperlogic:
https://app.asana.com/0/45663370065758/107567199240596

hyperlogic:
To test:
  * Use hand controllers to move your avatar's hand in front of your face.  Then move forward or fly forward.  On master your hand will jitter and will jitter more the faster you are moving.  In this PR the jitter should be greatly reduced.

hifi-gustavo:
Build console output for commit: d7b13d226f5b469098c7556ebef75199165d0d7d available at: https://gist.github.com/d7ece1c760545c6c27742f05728be825<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7563.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7563.dmg

ZappoMan:
Whoa... Can we double check that this doesn't break time warp...

The line of code you're moving was recently moved to the prior location by @jherico late last week...

Can we make sure this isn't a regression.

ZappoMan:
after discussion with the group, this will not introduce problems with HMD head pose, so I retract my concern

cozza13:
I tested this and it works as described.

PR #7573, opened 2016-04-05
Update collision hulls of avatar children

sethalves:
- the collision hulls of entities which are children of avatars will now update when the avatar moves.


hifi-gustavo:
Build console output for commit: 294bdcb18eaaa24f7df5b5d3db077b28e033a274 available at: https://gist.github.com/1fba36b9f07b81a2db5010f2f09d12a6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7573.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7573.dmg

AndrewMeadows:
@sethalves, you need a testplan for this so that someone else has a hope of being able to test it.  How would someone witness and verify the effects of these changes?

sethalves:
test plan:
- run at least edit.js, grab.js, handControllerGrab.js
- make two boxes, use edit.js to make them dynamic (and grabbable)
- optionally, work over a table, give the boxes some gravity
- with one hand, use a hydra to bumper/parent grab one box
- gently bump the held box into the other box, see that they collide
- while still holding the first box, lift it out of the way
- with grab.js, slide the second box over where the first box was before you grabbed it
- make sure that the collision hull of the first box isn't hanging around where it was before the grab

TNS-TrueNorth:
No issues with collisions of hulls, tested with a few boxes while enabling dynamic & grabbable. Also played in demo, passes on Windows 10.

ZappoMan:
@AndrewMeadows @sethalves - any chance this is going to cause potential FPS issues?

sethalves:
I wouldn't expect this to change FPS.

I found and hopefully fixed a problem -- Another QA pass is in order.  I slightly updated the test plan.

hifi-gustavo:
Build console output for commit: 30d6734253c5f8e8475c2decbce5f5d9d400d695 available at: https://gist.github.com/1fba36b9f07b81a2db5010f2f09d12a6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7573.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7573.dmg

hifi-gustavo:
Build console output for commit: 8158d34b20df4d2ff625c6bcb340853bf203147d available at: https://gist.github.com/1fba36b9f07b81a2db5010f2f09d12a6<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7573.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7573.dmg

libraries/entities/src/EntityItem.cpp
```diff
@@ -1975,6 +1978,11 @@ QList<EntityActionPointer> EntityItem::getActionsOfType(EntityActionType typeToG
 
 void EntityItem::locationChanged() {
     requiresRecalcBoxes();
+    _dirtyFlags |= Simulation::DIRTY_TRANSFORM;
+    EntityTreePointer tree = getTree();
+    if (tree) {
+        tree->entityChanged(getThisPointer());
+    }
```
AndrewMeadows: I'm worried about this change.  `EntityItem::locationChanged()` is called in many places and it's meaning is something like: "**the entity's transform has changed and therefore the world-frame transforms of child entities need to be updated**" (right?).

The meaning of setting the Simulation::DIRTY_TRANSFORM flag is: "**the entity's transform has changed from outside Bullet therefore Bullet needs to update the corresponding RigidBody**".

If `EntityItem::locationChanged()` is ever called as a result of Bullet trying to update the Entity's transform then we'll end up in a constant update cycle: ...Bullet tells Entity tells Bullet tells Entity tells Bullet...

I'll have to audit and test the code to make sure we're not doing that.

AndrewMeadows: Also, I would want an assert here that the tree is locked.  If we ever call this when the tree is not locked then we'll have problems.

AndrewMeadows:
Besides the potential update loop, I don't see any performance problems.  The check for ancestor type only happens when we're computing the MotionType of the entity before inserting its RigidBody into the Bullet simulation, which does not happen very often.

hifi-gustavo:
Build console output for commit: 45733478da8f14d5bfbd5b41e92fe3a90adf9770 available at: https://gist.github.com/1fba36b9f07b81a2db5010f2f09d12a6<br/><br/>Installers available for download at: <br/>Windows: http://s3-us-west-1.amazonaws.com/hifi-pr-builds/HighFidelity-Alpha-PR7573.exe<br/>OS X: http://s3-us-west-1.amazonaws.com/hifi-pr-builds/HighFidelity-Alpha-PR7573.dmg

libraries/entities/src/EntityItem.cpp
AndrewMeadows: My two fears appear to be unfounded:

(1) I wasn't able to see or trigger the update cycle for kinematic objects that I was worrying about.

(2) I don't think it is possible to call EntityItem::getTree() without getting the lock.

cozza13:
I m getting this error when trying to grab 

[04/07 13:50:42] [WARNING]     update(0.016680000349879265) at http://rawgit.com/sethalves/hifi/update-collision-hulls-of-avatar-children/examples/controllers/handControllerGrab.js:1948
[04/07 13:50:42] [WARNING]     <global>() at -1

cozza13:
tested and works

PR #7574, opened 2016-04-05
Hold onto octree child after creation

zzmp:
- Hold onto octree element after creation so nothing can delete it before use.

---

#### Intent
Prevent a documented crash from trying to recurse into an empty element.
(See: https://app.asana.com/0/26225263936266/105096043004292.)

#### Testing

General smoke test. The original crash could not be reproduced, so just make sure things still work.

### DIFF
--- a/libraries/octree/src/Octree.cpp
+++ b/libraries/octree/src/Octree.cpp
@@ -384,9 +384,10 @@ int Octree::readElementData(OctreeElementPointer destinationElement, const unsig
         // check the exists mask to see if we have a child to traverse into
 
         if (oneAtBit(childInBufferMask, childIndex)) {
-            if (!destinationElement->getChildAtIndex(childIndex)) {
+            auto childAt = destinationElement->getChildAtIndex(childIndex);
+            if (!childAt) {
                 // add a child at that index, if it doesn't exist
-                destinationElement->addChildAtIndex(childIndex);
+                childAt = destinationElement->addChildAtIndex(childIndex);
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: ac30aed9482d9a4117b8d4c5a94b57a6ed8aa55f available at: https://gist.github.com/2dcda9f3b9b69dcd37bf27052493586a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7574.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7574.dmg

ZappoMan:
This change seems reasonable... but also I don't think it will fix the crash, because the entire tree should be looked for writing so no one should be changing the child elements on you while you're in these functions.

kevinmthomas-carpool:
PR Build: 7574
Testing OS:  Windows 10
Testing Location:  hifi://music, hifi://sandbox, hifi://playa
Test Results:  Successful smoke test on Music, Sandbox, Playa.  Successful atp import, manipulation and deletion on Music and successful fbx import, manipulation and deletion on Sandbox.  Graphics performed as expected and equivalent to the current release build.  Upon exit of Interface it caused a crash to which the hifi-log is attached for inspection.
[attachments.zip](https://github.com/highfidelity/hifi/files/203683/attachments.zip)

PR #7584, opened 2016-04-05
Avoid warning for expected unset PresentThread

zzmp:
- Add an `isSet` method to the `DependencyManager`
- Use `isSet` when initializing the `PresentThread` to avoid unnecessary warning

Adds a method to circumvent a warning where getting the present thread failed, even if you are only getting it to see if it exists. Does so by adding an `isSet` method to see if it exists (so the warning will still trigger if you use `get` and it DNE).

#### Testing

General smoke test on multiple devices (Oculus, Vive, Desktop).
In current PR build, you will *always* get a warning when you start a new device as its present thread is created.
In this PR build, you will only get a warning if something goes wrong.

### DIFF
--- a/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
+++ b/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
@@ -245,8 +245,10 @@ bool OpenGLDisplayPlugin::activate() {
 
 #if THREADED_PRESENT
     // Start the present thread if necessary
-    auto presentThread = DependencyManager::get<PresentThread>();
-    if (!presentThread) {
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 37efeb7f2952e8001bad3b7b4e04562e345ce56b available at: https://gist.github.com/421bfe628e13f54d748b5583bb5a7e09<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7584.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7584.dmg

cozza13:
tested - QA pass

PR #7595, opened 2016-04-06
Fix a random case of the depth reticle hiding when it shouldn't

ZappoMan:
- Fix a random case of the depth reticle hiding when it shouldn't
- Changes Entities.findRayIntersection() to block so we don't get non-answers
- fixes a potential thread safety issue in Reticle.allowMouseCapture


ZappoMan:
Test plan:
- Basic smoke test
- Make sure you're running the PR version of defaultScripts.js
- In the HMD verify that depthReticle still works and points at things
- Make sure the away.js still works as expected (ALT-TAB away, use ESC to enter/leave away, etc)
- Follow the following obscure corner case to test the autohide bug fix...

Testing the auto-hide case: This should fail on a production build, but succeed with this PR.
- In the HMD, with default scripts running.
- Open up the running scripts dialog, and set focus to the "search" field... notice you will see a blinking I-beam cursor in the edit control.
- slowly move your mouse around in a circle so that it moves over the running scripts dialog to over the scene
- **On production builds**: sometimes randomly (usually after about 10 seconds) when you move the reticle over the scene it will disappear completely and never show again even as you move it. << this is how production will fail.
- **On this PR**: this reticle will always remain visible as you move it.

examples/depthReticle.js
```diff
@@ -98,10 +98,21 @@ function seekToLookAt() {
 }
 
 function autoHideReticle() {
+    var now = Date.now();
+
+    // sometimes we don't actually get mouse move messages (for example, if the focus has been set
+    // to an overlay or web page 'overlay') in but the mouse can still be moving, and we don't want
+    // to autohide in these cases, so we will take this opportunity to also check if the reticle
+    // position has changed.
+    if (lastMouseX != Reticle.position.x || lastMouseY != Reticle.position.y) {
```
ZappoMan: this is the bug fix for the auto-hide problem. The issue is that when a QML window has focus, the mouseMove events don't always get forwarded to the JS Scripting layer... so we weren't getting mouse move messages, and our auto-hide was kicking in. This is a little belt and suspenders here, where the auto-hide logic doesn't depend on mouseMove events coming through to determine if it should auto-hide or not.

libraries/display-plugins/src/display-plugins/CompositorHelper.cpp
```diff
@@ -503,7 +501,7 @@ glm::mat4 CompositorHelper::getReticleTransform(const glm::mat4& eyePose, const
         mousePosition.y *= -1.0f;
 
         vec2 mouseSize = CURSOR_PIXEL_SIZE / canvasSize;
-        return glm::scale(glm::translate(glm::mat4(), vec3(mousePosition, 0.0f)), vec3(mouseSize, 1.0f));
+        result = glm::scale(glm::translate(glm::mat4(), vec3(mousePosition, 0.0f)), vec3(mouseSize, 1.0f));
```
ZappoMan: this was cleaning up the flow of this function, so that both paths exit at the same location - it should have no material change in the logic.

libraries/display-plugins/src/display-plugins/CompositorHelper.h
```diff
@@ -176,7 +176,7 @@ protected slots:
 
     bool _reticleOverQml { false };
 
-    bool _allowMouseCapture { true };
+    std::atomic<bool> _allowMouseCapture { true };
```
ZappoMan: this addresses a potential thread safety issue if two threads (like two scripts) are simultaneously accessing the `Reticle.allowMouseCapture` state.

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -475,13 +475,20 @@ QVector<QUuid> EntityScriptingInterface::findEntitiesInBox(const glm::vec3& corn
     return result;
 }
 
-RayToEntityIntersectionResult EntityScriptingInterface::findRayIntersection(const PickRay& ray, bool precisionPicking, const QScriptValue& entityIdsToInclude, const QScriptValue& entityIdsToDiscard) {
+RayToEntityIntersectionResult EntityScriptingInterface::findRayIntersection(const PickRay& ray, bool precisionPicking,
```
ZappoMan: formatting to match coding standard (line was too long)

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -475,13 +475,20 @@ QVector<QUuid> EntityScriptingInterface::findEntitiesInBox(const glm::vec3& corn
...
     QVector<EntityItemID> entitiesToInclude = qVectorEntityItemIDFromScriptValue(entityIdsToInclude);
     QVector<EntityItemID> entitiesToDiscard = qVectorEntityItemIDFromScriptValue(entityIdsToDiscard);
-    return findRayIntersectionWorker(ray, Octree::TryLock, precisionPicking, entitiesToInclude, entitiesToDiscard);
+    return findRayIntersectionWorker(ray, Octree::Lock, precisionPicking, entitiesToInclude, entitiesToDiscard);
 }
 
-RayToEntityIntersectionResult EntityScriptingInterface::findRayIntersectionBlocking(const PickRay& ray, bool precisionPicking, const QScriptValue& entityIdsToInclude, const QScriptValue& entityIdsToDiscard) {
+// FIXME - we should remove this API and encourage all users to use findRayIntersection() instead. We've changed
+//         findRayIntersection() to be blocking because it never makes sense for a script to get back a non-answer
+RayToEntityIntersectionResult EntityScriptingInterface::findRayIntersectionBlocking(const PickRay& ray, bool precisionPicking,
```
ZappoMan: formatting to match coding standard (line was too long)

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -475,13 +475,20 @@ QVector<QUuid> EntityScriptingInterface::findEntitiesInBox(const glm::vec3& corn
...
+    return findRayIntersectionWorker(ray, Octree::Lock, precisionPicking, entitiesToInclude, entitiesToDiscard);
 }
 
-RayToEntityIntersectionResult EntityScriptingInterface::findRayIntersectionBlocking(const PickRay& ray, bool precisionPicking, const QScriptValue& entityIdsToInclude, const QScriptValue& entityIdsToDiscard) {
+// FIXME - we should remove this API and encourage all users to use findRayIntersection() instead. We've changed
+//         findRayIntersection() to be blocking because it never makes sense for a script to get back a non-answer
+RayToEntityIntersectionResult EntityScriptingInterface::findRayIntersectionBlocking(const PickRay& ray, bool precisionPicking, 
+                const QScriptValue& entityIdsToInclude, const QScriptValue& entityIdsToDiscard) {
+
+    qDebug() << "Entities.findRayIntersectionBlocking() is obsolete, use Entities.findRayIntersection() instead.";
```
ZappoMan: warn users of obsolete function

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -475,13 +475,20 @@ QVector<QUuid> EntityScriptingInterface::findEntitiesInBox(const glm::vec3& corn
...
 }
 
-RayToEntityIntersectionResult EntityScriptingInterface::findRayIntersection(const PickRay& ray, bool precisionPicking, const QScriptValue& entityIdsToInclude, const QScriptValue& entityIdsToDiscard) {
+RayToEntityIntersectionResult EntityScriptingInterface::findRayIntersection(const PickRay& ray, bool precisionPicking, 
+                const QScriptValue& entityIdsToInclude, const QScriptValue& entityIdsToDiscard) {
+
     QVector<EntityItemID> entitiesToInclude = qVectorEntityItemIDFromScriptValue(entityIdsToInclude);
     QVector<EntityItemID> entitiesToDiscard = qVectorEntityItemIDFromScriptValue(entityIdsToDiscard);
-    return findRayIntersectionWorker(ray, Octree::TryLock, precisionPicking, entitiesToInclude, entitiesToDiscard);
+    return findRayIntersectionWorker(ray, Octree::Lock, precisionPicking, entitiesToInclude, entitiesToDiscard);
```
ZappoMan: there was some suspicion that these disappearing reticle bug could have been caused by a random failure to get a lock resulting in a bad `findRayIntersection()` call... I confirmed that that wasn't causing the hidden reticle, but in the process finally convinced myself these two different versions of the API are totally wrong.

There are many scripts that use the "non-blocking" version... but assume they will always get a correct answer, and aren't checking to see that the try-lock failed... those scripts are likely manifesting bugs we've not yet filed... this change will essentially fix those scripts.

hifi-gustavo:
Build console output for commit: 0f4e1e79fdf19f7b13940a9a27b974256ae4c916 available at: https://gist.github.com/a80381a545f4209c9fed58dc68e8fa3a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7595.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7595.dmg

libraries/entities/src/EntityScriptingInterface.cpp
huffman: I think it would be good to make this a `qWarning`

hifi-gustavo:
Build console output for commit: 7f164a79595e3eb899ede057e045b7c2b52a514b available at: https://gist.github.com/a80381a545f4209c9fed58dc68e8fa3a<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7595.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7595.dmg

cozza13:
In the HMD verify that depthReticle still works and points at things (CONFIRMED)
Make sure the away.js still works as expected (ALT-TAB away, use ESC to enter/leave away, etc) (CONFIRMED)

On this PR: this reticle will always remain visible as you move it. (Confirmed)

PR #7607, opened 2016-04-07
Less nauseating third person HMD camera

hyperlogic:
The idea is to never rotate the HMD camera while in third person, only translate it.
Also, it's important that any HMD translation/rotation from the head occurs in world space properly.

There are some drawbacks, though, it's very easy to move the avatar in a direction that is different than your current third person camera facing.  And it's easy to move the avatar out of your field of vision.

Feedback is welcome

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1560,10 +1560,59 @@ void Application::paintGL() {
             }
         } else if (_myCamera.getMode() == CAMERA_MODE_THIRD_PERSON) {
             if (isHMDMode()) {
-                auto hmdWorldMat = myAvatar->getSensorToWorldMatrix() * myAvatar->getHMDSensorMatrix();
-                _myCamera.setRotation(glm::normalize(glm::quat_cast(hmdWorldMat)));
-                _myCamera.setPosition(extractTranslation(hmdWorldMat) +
-                    myAvatar->getOrientation() * boomOffset);
+
+                if (!_thirdPersonHMDCameraEnabled) {
+
+                    // we've just entered third person mode.
+                    _thirdPersonHMDCameraEnabled = true;
+
+                    // save a snapshot of the current hmd (in sensor space)
+                    _thirdPersonCameraHMDInverse = glm::inverse(myAvatar->getHMDSensorMatrix());
+
+                    // also keep a reference of the world facing direction of the hmd.
+                    _thirdPersonCameraOrientationReference = glmExtractRotation(myAvatar->getSensorToWorldMatrix() * myAvatar->getHMDSensorMatrix());
+
+                    // calculate a starting camera position that is behind the current hmd, but is only offset horizontally.
+                    glm::vec3 offset = _thirdPersonCameraOrientationReference * Vectors::UNIT_Z;
+                    glm::vec3 horizontalOffset(offset.x, 0.0f, offset.z);
+                    if (glm::length(horizontalOffset) > EPSILON) {
+                        horizontalOffset = glm::normalize(horizontalOffset) * myAvatar->getBoomLength();
+                    } else {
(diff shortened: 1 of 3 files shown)


hyperlogic:
https://app.asana.com/0/45663370065758/103199619107231

ZappoMan:
"it's very easy to move the avatar in a direction that is different than your current third person camera facing" -- this was my concern when I heard the description of what we were going to attempt.

ZappoMan:
FWIW -- In the production builds... I've noticed that the 3rd person mode for the HMD feels ok and natural unless/until the avatar auto-rotates it's body... then the booming action is kind of distracting. Also I found there's an "independent" camera mode if you're running edit.js where you can move the camera around disembodied from your avatar... that's not at all nauseating...

hifi-gustavo:
Build console output for commit: 04d4e7d2e5cc37f623723c53cbceb34d49bd94de available at: https://gist.github.com/1d0898fc0dc0a69fb65c2149192fabd4<br/><br/>Installers available for download at: <br/>Windows: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7607.exe<br/>OS X: http://s3.amazonaws.com/hifi-public/pr-builds/HighFidelity-Alpha-PR7607.dmg

ZappoMan:
Here's my thoughts on this new mode:
1) the "boom swing" effect is gone, and that's better.
2) I was very surprised that when I used the arrow/wasd keys to drive my character - he walked away from me and I couldn't see him.
3) I really want to be able to follow my character... this new mode makes it very hard to do that if you do any sort of turning.

My recommendation:
- if the view "rotates" because of an action I am taking (say I'm holding down the arrow key) then that's not nauseating to me
- if the view "rotates" because of an unknown action (say the character's IK has decided to rotate) then that is nauseating (actually, I don't feel nauseated, but I am surprised by it)
- I think we should find a way to rotate the view when you're using drive keys or controller inputs, but not when the IK does the auto-correct.

hyperlogic:
Cool, thanks for the feedback.

> 2) I was very surprised that when I used the arrow/wasd keys to drive my character - he walked away from me and I couldn't see him.

One idea would be to rotate the drive key input based on the current third person camera orientation, this is typically what third person action games do.  i.e. when I move the stick "forward" my character should walk away from the camera facing.  This would make the movement of the character more predictable, but doesn't help the issue with causing the camera to walk too far away.

>  "I think we should find a way to rotate the view when you're using drive keys or controller inputs, but not when the IK does the auto-correct."

This turns out to be kinda hard.  Currently both these motions occur within the characterController and would be difficult to decouple in this way.  It's possible, but I'm not sure it's worth it.  The root issue would still remain, which would be that sometimes you are not directly behind your avatar....

@PhilipRosedale raised the idea of snapping position and rotation of the camera when it was determined to be outside of your fov, back to be directly behind your avatar.  Not sure if this would be nauseating, but snap turns are more comfortable than smooth transitions.

hifi-gustavo:
Build console output for commit: 0b799d0d5d75f9a06f2ef2e19fa225c3d5648ac3 available at: https://gist.github.com/1d0898fc0dc0a69fb65c2149192fabd4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7607.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7607.dmg

hyperlogic:
The latest commit, does a better job of making the stick control more deterministic.  It's harder to accidently move your avatar out of your view, but it is still possible to intentionally move it out of your view.

hyperlogic:
Closing, will re-open a new PR with a different approach.

PR #7623, opened 2016-04-08
fix export of animation properties

ZappoMan:
the firstFrame, lastFrame, loop, and hold properties were not being properly exported - this will fix that

### DIFF
--- a/libraries/entities/src/AnimationPropertyGroup.cpp
+++ b/libraries/entities/src/AnimationPropertyGroup.cpp
@@ -244,6 +244,10 @@ void AnimationPropertyGroup::markAllChanged() {
     _fpsChanged = true;
     _currentFrameChanged = true;
     _runningChanged = true;
+    _loopChanged = true;
+    _firstFrameChanged = true;
+    _lastFrameChanged = true;
+    _holdChanged = true;
 }
 
 EntityPropertyFlags AnimationPropertyGroup::getChangedProperties() const {


ZappoMan:
Testing Plan:
- run interface
- create an animate model - set it's firstFrame and lastFrame properties
- export it
- verify that the firstFrame and lastFrame items are included in the JSON

Fixes this bug: https://app.asana.com/0/26320649018758/104986966674731

Note: there is a separate bug, where sometimes when you export an entity, the entity that is selected gets deleted. this PR does not attempt to fix that bug (I'm still looking into that)... so, if that happens to you... don't be surprised, and that shouldn't count as a QA fail for this PR.

libraries/entities/src/AnimationPropertyGroup.cpp
```diff
@@ -244,6 +244,10 @@ void AnimationPropertyGroup::markAllChanged() {
     _fpsChanged = true;
     _currentFrameChanged = true;
     _runningChanged = true;
+    _loopChanged = true;
```
ZappoMan: dev notes: yes-- this is lame... the bug was caused by the original implementation of markAllChanged() not marking all the properties as changed. This behavior is depended on by the export logic. Yes... this is a perfect example of a part of entities that is brittle... I am well aware of that.

hifi-gustavo:
Build console output for commit: f1f8fd8a24de8d920e828aa12d53dda23deb878f available at: https://gist.github.com/09301910391070590a726e55039a3948<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7623.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7623.dmg

howard-stearns:
test notes: An example model URL / animation URL pair is http://hifi-content.s3.amazonaws.com/DomainContent/production/cow/newMooCow.fbx / http://hifi-content.s3.amazonaws.com/DomainContent/production/cow/Cow_Idle.fbx, with a first frame: 15 and lastFrame: 180.

thoys:
Does this also fix duplicating animated entities? I had a few awkward moments where I made multiple copies of certain objects and I was the only one seeing them :)

thoys:
I will just QA this and find it out :+1:

ZappoMan:
@thoys - I do not believe it fixes the duplicating issue... I think that's a different issue I am looking at as well. I'm doing this one step at a time.

ZappoMan:
NOTE: I figured out the repro case on "export" sometimes deleting the entity you're trying to export... So I can tell you how to NOT do that to yourself.

Namely, if you use the "delete" key or the "backspace" key in the "Save As" file selector dialog box... you will send that many delete key messages to the JS layer... and as a result, edit.js will delete the entity.

So long as you don't press the delete key while in that dialog you'll be ok

imgntn:
@ZappoMan OOF, that's a good find!!  backspace key in export during 'save as' while changing the file might have the trigger i kept pulling.

thoys:
## QA Test report

Animation first- and lastFrame are being exported correctly:
```
{
    "Entities": [
        {
            "animation": {
                "currentFrame": 15,
                "firstFrame": 15,
                "lastFrame": 180,
                "url": "http://hifi-content.s3.amazonaws.com/DomainContent/production/cow/Cow_Idle.fbx"
            },
            "created": "2016-04-11T10:04:56Z",
            "dimensions": {
                "x": 0.73958581686019897,
                "y": 1.6134631633758545,
                "z": 2.5299453735351562
            },
            "id": "{5daa1a23-e8ce-458b-b165-8d2acb2f202d}",
            "modelURL": "http://hifi-content.s3.amazonaws.com/DomainContent/production/cow/newMooCow.fbx",
            "queryAACube": {
                "scale": 3.0904488563537598,
                "x": -1.5452244281768799,
                "y": -1.5452244281768799,
                "z": -1.5452244281768799
            },
            "type": "Model"
        }
    ],
    "Version": 57
}

```
Everything works fine. Was stubborn and tested the copying of animated cows anyway, this doesn't fix it, the entities server crashed in fact while trying that. 

QA Successful :+1:

PR #7632, opened 2016-04-10
avoid loop where bullet moves an entity and then an entity tells bullet that it has moved.

sethalves:
- fix bug that causes physical entities to pop and gain energy when they shouldn't

### DIFF
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1976,14 +1976,16 @@ QList<EntityActionPointer> EntityItem::getActionsOfType(EntityActionType typeToG
     return result;
 }
 
-void EntityItem::locationChanged() {
+void EntityItem::locationChanged(bool tellPhysics) {
     requiresRecalcBoxes();
-    _dirtyFlags |= Simulation::DIRTY_TRANSFORM;
+    if (tellPhysics) {
+        _dirtyFlags |= Simulation::DIRTY_TRANSFORM;
+    }
     EntityTreePointer tree = getTree();
     if (tree) {
         tree->entityChanged(getThisPointer());
     }
-    SpatiallyNestable::locationChanged(); // tell all the children, also
+    SpatiallyNestable::locationChanged(tellPhysics); // tell all the children, also
 }
 
 void EntityItem::dimensionsChanged() {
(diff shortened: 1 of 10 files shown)


libraries/entities/src/EntityItem.cpp
```diff
@@ -1976,14 +1976,16 @@ QList<EntityActionPointer> EntityItem::getActionsOfType(EntityActionType typeToG
     return result;
 }
 
-void EntityItem::locationChanged() {
+void EntityItem::locationChanged(bool tellPhysics) {
     requiresRecalcBoxes();
-    _dirtyFlags |= Simulation::DIRTY_TRANSFORM;
+    if (tellPhysics) {
+        _dirtyFlags |= Simulation::DIRTY_TRANSFORM;
+    }
```
sethalves: This is the main change, the rest is plumbing.  If tellPhysics is false, don't set the physics related flags.

sethalves:
to test:
- reset demo room and observer that things don't explode
- equip (parent/bumper-grab) a golf club and make sure it can still hit balls
- make sure moving-platform.js still works

hifi-gustavo:
Build console output for commit: 43f735dd211e75a24673422b906638d5eab53670 available at: https://gist.github.com/7e64eb5e46091a4f978eefe80295ebcd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7632.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7632.dmg

AndrewMeadows:
I looked over this code and think it will fix the problem.  Once we get QA confirmation we should merge it.

thoys:
## QA report

- Tested with @kevinmthomas-carpool and @Nex-Pro  in hifi://lobby and pressed the tidy up button multiple times
    - Nothing explodes anymore :+1: 
- Philip's golfclub script still hits the golfball  :+1: 
- `moving-platform.js` does the same as in the release build. It is pretty jittery. When avatar collisions is disabled it does follow correctly without losing the platform, but I have no idea what is supposed to happen. :question: 

Based on that I can tell this QA is a success for fixing the lobby script. For the moving platform script it didn't change anything, so QA pass :+1:

kevinmthomas-carpool:
## QA Test
**PR Build:** 7632
**Testing OS:** Windows 10
**Testing Locations:** sandbox, forum
**Test Results Requested:** Successful test of pressing the clean-up button in forum to which the entities behaved normally.  In addition, tested with @thoys who was also running the current PR build and upon pressing the clean-up button the entities behaved normally.  As a final test of this procedure, @grayson2015 who was running the current release build was present at forum to which upon a click of the clean-up button the entities did not behave normally and became unstable.  Unsuccessful test of moving-platform.js as it's behavior was unstable and jittery to concur with @thoys and can also concur that when avatar collisions are disabled it does follow correctly without losing the platform.  Successful test of the golf club utilizing the hydras.
**Successful Logout/Crash (If  A Crash Occurs, Attach ‘hifi-log.txt’):** Successful logout of Interface.

Nex-Pro:
Windows 10 64bit 16gb ram gtx770
QA Test.  'TIDY UP' button fails to work reliably for me in this build. ( took a long time to become active) Once active and clicked it was fine.
Moving platform experience was very jittery. My avatar was randomly moving/flickering.

PR #7633, opened 2016-04-11
Fix raw pointers in shared resource cache items

Atlante45:


Atlante45:
https://app.asana.com/0/26551373415487/108673470539519

Atlante45:
Testing plan (from asana task):
1. Go to somewhere with lots to load.
2. Open up stats (Developer > Stats), and click it to open up the stats on pending requests.
3. Reload the cache repeatedly.

hifi-gustavo:
Build console output for commit: 5e76ab90c42678a14f914d65d4154d5672d547f3 available at: https://gist.github.com/2f80c44c3481d139f1c869c3f2aadb72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.dmg

hifi-gustavo:
Build console output for commit: 4313b2266c40f3f58286f495fae1b6cd77b22a35 available at: https://gist.github.com/2f80c44c3481d139f1c869c3f2aadb72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.dmg

kevinmthomas-carpool:
## QA Test
**PR Build:** 7633
**Testing OS:** Windows 10
**Testing Locations:** Sandbox, Playa, Cellscience, Forum
**Smoke Test Results:** Successful test.

                                  |Playa    |Cellscience |Forum   
----------------------------------|---------|------------|--------
Render Rate:                      |63       |62          |56           
Present Rate:                     |60       |60          |60          
Simrate:                          |63       |64          |57           
Render Stats Num Buffers:         |2737     |3073        |4029              
Render Stats gpu::Buffer Memory:  |51.8Mb   |65.4Mb      |53.3Mb              
Render Stats Num Textures:        |429      |182         |1082              
Render Stats gpu::Texture Memory: |942.8Mb  |344.4Mb     |893.8Mb              
Render Stats Triangles:           |480k     |564k        |1853k            
Render Stats Drawcalls:           |270      |17204       |1135             
Render Stats Items:               |238      |800         |544             

**sandbox ‘createtestblocks.js’ Result:** Successful test.
**sandbox Import FBX Result:** Successful test.
**sandbox ‘harmonicOscillator.js’ Result:** Successful test.
**sandbox Echo Server Audio Result:** Successful test.
**sandbox ‘reverbTest.js’ Result:** Successful test.
**Additional Test Results Requested:** Failure as successive refreshes of cache on Playa caused an Interface crash.  Upon a relog, visited Cellscience and repeated test, successful results.  Visited Forum and repeated test, successful results.  Log file included for deeper analysis.
**Successful Logout/Crash:** Failure.
[attachments.zip](https://github.com/highfidelity/hifi/files/213382/attachments.zip)

interface/src/scripting/GlobalServicesScriptingInterface.cpp
```diff
@@ -122,7 +122,7 @@ void DownloadInfoResultFromScriptValue(const QScriptValue& object, DownloadInfoR
 
 DownloadInfoResult GlobalServicesScriptingInterface::getDownloadInfo() {
     DownloadInfoResult result;
-    foreach(Resource* resource, ResourceCache::getLoadingRequests()) {
+    foreach(auto resource, ResourceCache::getLoadingRequests()) {
```
ZappoMan: maybe we should change this to a `for()` AFAICT `foreach()` results in a copy of the container being iterated across

interface/src/ui/Stats.cpp
```diff
@@ -210,7 +210,7 @@ void Stats::updateStats(bool force) {
         // If the urls have changed, update the list
         if (shouldUpdateUrls) {
             _downloadUrls.clear();
-            foreach (Resource* resource, loadingRequests) {
+            foreach (auto resource, loadingRequests) {
```
ZappoMan: maybe we should change this to a `for()` AFAICT `foreach()` results in a copy of the container being iterated across

libraries/animation/src/AnimNodeLoader.cpp
```diff
@@ -568,12 +568,21 @@ bool processStateMachineNode(AnimNode::Pointer node, const QJsonObject& jsonObj,
 }
 
 AnimNodeLoader::AnimNodeLoader(const QUrl& url) :
-    _url(url),
-    _resource(nullptr) {
+    _url(url)
+{
+    auto request = ResourceManager::createResourceRequest(this, url);
+    if (request) {
+        connect(request, &ResourceRequest::finished, this, [this, request]() {
```
ZappoMan: How do we ensure that the `AnimNodeLoader` lifetime out lasts the resource request? It looks like `AnimNodeLoader` is always owned by a `Rig` instance... `Rig`s appear to be managed through stl shared pointers. it seems like you'd want to make sure the `Rig` instance isn't freed before this callback completes.

Can you clarify?

libraries/networking/src/ResourceCache.cpp
```diff
@@ -41,9 +41,9 @@ void ResourceCache::refreshAll() {
     resetResourceCounters();
     
     // Refresh all remaining resources in use
-    foreach (auto resource, _resources) {
-        if (!resource.isNull()) {
-            resource.data()->refresh();
+    foreach (QSharedPointer<Resource> resource, _resources) {
```
ZappoMan: maybe we should change this to a `for()` AFAICT `foreach()` results in a copy of the container being iterated across

libraries/networking/src/ResourceCache.cpp
```diff
@@ -198,43 +198,64 @@ void ResourceCache::updateTotalSize(const qint64& oldSize, const qint64& newSize
...
 
-QList<QPointer<Resource>> ResourceCacheSharedItems::getPendingRequests() const {
-    Lock lock(_mutex);
-    return _pendingRequests;
+QList<QSharedPointer<Resource>> ResourceCacheSharedItems::getPendingRequests() {
+    QList<QSharedPointer<Resource>> result;
+
+    {
+        Lock lock(_mutex);
+        foreach(QSharedPointer<Resource> resource, _pendingRequests) {
```
ZappoMan: maybe we should change this to a `for()` AFAICT `foreach()` results in a copy of the container being iterated across -- admittedly, the Qt container classes do a copy on write, so presumably as long as these stay Qt we don't pay the price... but as we're seeing more and more comingly of the stl containers and qt containers, we might want to migrate to more use of `for()`

libraries/networking/src/ResourceCache.cpp
```diff
@@ -198,43 +198,64 @@ void ResourceCache::updateTotalSize(const qint64& oldSize, const qint64& newSize
...
 
-QList<Resource*> ResourceCacheSharedItems::getLoadingRequests() const {
-    Lock lock(_mutex);
-    return _loadingRequests;
+QList<QSharedPointer<Resource>> ResourceCacheSharedItems::getLoadingRequests() {
+    QList<QSharedPointer<Resource>> result;
+
+    {
+        Lock lock(_mutex);
+        foreach(QSharedPointer<Resource> resource, _loadingRequests) {
```
ZappoMan: `for()` vs `foreach()`

libraries/networking/src/ResourceCache.cpp
```diff
@@ -198,43 +198,64 @@ void ResourceCache::updateTotalSize(const qint64& oldSize, const qint64& newSize
...
+            if (resource) {
+                result.append(resource);
+            }
+        }
+        _loadingRequests.removeAll(QWeakPointer<Resource>());
```
ZappoMan: Clement and I are discussing this line of code off line... take note: this might be a source of problems.

interface/src/scripting/GlobalServicesScriptingInterface.cpp
zzmp: Also, `const auto&`. I'm not sure how foreach works, but if it allows `const auto&`, I'd assume that avoids another copy. Same for other `foreach` loops.

libraries/animation/src/AnimNodeLoader.cpp
zzmp: Would inheriting from Resource solve this issue? Is that a lot more work?

libraries/networking/src/ResourceCache.cpp
```diff
@@ -41,9 +41,9 @@ void ResourceCache::refreshAll() {
     resetResourceCounters();
     
     // Refresh all remaining resources in use
-    foreach (auto resource, _resources) {
-        if (!resource.isNull()) {
-            resource.data()->refresh();
+    foreach (QSharedPointer<Resource> resource, _resources) {
+        if (resource) {
```
zzmp: What is the case where this is true?

libraries/networking/src/ResourceCache.cpp
```diff
@@ -198,43 +198,64 @@ void ResourceCache::updateTotalSize(const qint64& oldSize, const qint64& newSize
...
+    QList<QSharedPointer<Resource>> result;
+
+    {
+        Lock lock(_mutex);
+        foreach(QSharedPointer<Resource> resource, _pendingRequests) {
+            if (resource) {
+                result.append(resource);
+            }
+        }
+        _pendingRequests.removeAll(QWeakPointer<Resource>());
```
zzmp: I don't think this will remove invalidated QWeakPointer.

libraries/networking/src/ResourceCache.cpp
zzmp: Same as GetPendingRequests.

libraries/networking/src/ResourceCache.h
```diff
@@ -62,21 +62,20 @@ class ResourceCacheSharedItems : public Dependency  {
...
     using Lock = std::unique_lock<Mutex>;
 public:
-    void appendPendingRequest(Resource* newRequest);
-    void appendActiveRequest(Resource* newRequest);
-    void removeRequest(Resource* doneRequest);
-    QList<QPointer<Resource>> getPendingRequests() const;
+    void appendPendingRequest(QWeakPointer<Resource> newRequest);
+    void appendActiveRequest(QWeakPointer<Resource> newRequest);
+    void removeRequest(QWeakPointer<Resource> doneRequest);
+    QList<QSharedPointer<Resource>> getPendingRequests();
```
zzmp: Why did you decide to make these QSharedPointer instead of QWeakPointer?
Can you add a comment that holding these shared pointers prevents the resources from being freed, even if they are no longer in the scene and/or cache?

libraries/networking/src/ResourceCache.h
```diff
@@ -105,13 +104,11 @@ class ResourceCache : public QObject {
...
-    static const QList<Resource*> getLoadingRequests() 
-        { return DependencyManager::get<ResourceCacheSharedItems>()->getLoadingRequests(); }
+    static QList<QSharedPointer<Resource>> getLoadingRequests();
 
-    static int getPendingRequestCount() 
-        { return DependencyManager::get<ResourceCacheSharedItems>()->getPendingRequestsCount(); }
+    static int getPendingRequestCount();
 
-    ResourceCache(QObject* parent = NULL);
+    ResourceCache(QObject* parent = nullptr);
```
zzmp: Why not just use a default ctor?

interface/src/scripting/GlobalServicesScriptingInterface.cpp
Atlante45: From what I understand, it is safer to use `foreach` for Qt container as it will only do a shallow copy, when `for(:)` might do a deep copy in some cases. (A guy on the internet said so -->)
Here's a nice table showing the different cases: https://youtu.be/Ov7s0GgBbOQ?t=670
You might argue though that it makes it dangerous, if we ever switch those containers to the STL, then it'll definitely do a deep copy.

zzmp: According to that video, making it `for(const auto& * : *)` is the worst thing you could do :).

zzmp: Here it is in the docs (basically the same thing, foreach for Qt, for(:) for stl): http://doc.qt.io/qt-5/containers.html#the-foreach-keyword.

libraries/networking/src/ResourceCache.h
Atlante45: Not sure what you mean, you can't use the default impl when the actor has arguments.

libraries/networking/src/ResourceCache.cpp
Atlante45: Yes, I just tested that with Brad, that was a hack to try and cleanup the list a little bit while we know it won't change. But that call won't do anything.

libraries/networking/src/ResourceCache.cpp
Atlante45: This is true for every single resource dropped from the unused list or that is not cacheable, since we never properly remove them.

libraries/networking/src/ResourceCache.h
Atlante45: Since this is an interface for Stats/JS, I didn't want users to have to deal with possibly invalidated weak pointers.
I think the benefits are worth the very occasional lifetime extension of a resource.

libraries/animation/src/AnimNodeLoader.cpp
Atlante45: The request is supposed to have `this` as a parent, but I just checked and that's not true anymore. 

Zach not sure what you mean, nothing should use `Resource` unless it is subclassed and has an associated `ResourceCache`. That's why this was an issue in the first place.

However, I think I'll go back to using a Resource here and use the same pattern I used in ResourceTest,
since it looks like this is doing more harm than good.
I don't really understand why we are not using the AnimationCache here, maybe @hyperlogic can enlighten us?

zzmp: I mean would it be easier to just switch over to using a cache here, instead of doing a one-off thing that is only kind of a resource.

hifi-gustavo:
Build console output for commit: faa3dca970de50e804d737944183034878d78093 available at: https://gist.github.com/2f80c44c3481d139f1c869c3f2aadb72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.dmg

hifi-gustavo:
Build console output for commit: c1e0c8b7497e29a77258fb130a28dd6afc2b43ce available at: https://gist.github.com/2f80c44c3481d139f1c869c3f2aadb72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.dmg

hifi-gustavo:
Build console output for commit: 364f935b893505afd6873816ecf2815ad7bb41dd available at: https://gist.github.com/2f80c44c3481d139f1c869c3f2aadb72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.dmg

Atlante45:
test this please

hifi-gustavo:
Build console output for commit: 364f935b893505afd6873816ecf2815ad7bb41dd available at: https://gist.github.com/ed82b4aedeb4906640bea7873241976e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.dmg

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -77,21 +78,26 @@ void GeometryMappingResource::downloadFinished(const QByteArray& data) {
...
         _shapes = _geometryResource->_shapes;
         _meshes = _geometryResource->_meshes;
         _materials = _geometryResource->_materials;
-    }
 
-    // Avoid holding onto extra references
-    _geometryResource.reset();
+        // Avoid holding onto extra references
+        _geometryResource.reset();
+    }
```
Atlante45: @zzmp Can you check this specific file. This is the actual fix to the crash.

zzmp: @Atlante45 you mean it was going into the `if (success)` branch with no `_geometryResource`?

Atlante45: @zzmp Yes, that happens if you refresh all resources.
_geometryResource will trigger the slot again, but the pointer has been reset since then.

zzmp: Good catch. Can you add a disconnect and a FIXME comment after the reset, just to be thorough?

hifi-gustavo:
Build console output for commit: 44e4dcd4233383601e16bac9ac1f1836d03309f8 available at: https://gist.github.com/2f80c44c3481d139f1c869c3f2aadb72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.dmg

hifi-gustavo:
Build console output for commit: 785eda44cd8df122f9ba37aae2058065b2aeb97c available at: https://gist.github.com/2f80c44c3481d139f1c869c3f2aadb72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.dmg

Atlante45:
test this please

hifi-gustavo:
Build console output for commit: 785eda44cd8df122f9ba37aae2058065b2aeb97c available at: https://gist.github.com/2f80c44c3481d139f1c869c3f2aadb72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.dmg

kevinmthomas-carpool:
## QA Test
**PR Build:** 7633
**Testing OS:** Windows 10
**Testing Locations:** localhost, playa, cellscience, forum

                                  |localhost |playa    |cellscience |forum  
----------------------------------|----------|---------|------------|-----
Render Rate:                      |59.82     |55.94    |51.95       |53.15    
Present Rate:                     |60.04     |60.04    |60.04       |60.04   
Simrate:                          |55.12     |57.09    |50.20       |54.13    
Render Stats Num Buffers:         |876       |4439     |2437        |4241    
Render Stats gpu::Buffer Memory:  |21.2Mb    |94.2Mb   |14.2Mb      |32.2Mb        
Render Stats Num Textures:        |279       |402      |213         |836    
Render Stats gpu::Texture Memory: |733.8Mb   |1094.3Mb |371.4Mb     |1636.8Mb         
Render Stats Triangles:           |562k      |490k     |557k        |1328k    
Render Stats Drawcalls:           |165       |9695     |1311        |663   
Render Stats Items:               |68        |259      |812         |350
GPU-Z Memory Size                 |4096Mb    |4096Mb   |4096Mb      |4096Mb
GPU-Z Initial Memory		  |256Mb     |256Mb    |256Mb       |256Mb
GPU-Z Memory Used		  |972Mb     |2028Mb   |1364Mb      |3155Mb
GPU-Z Memory Load		  |8%        |13%      |10%         |10%
**Additional Test Results Requested:** Successful test rapid reloading of cache on each of the aforementioned domains listed above with no issue of reload of cache.

interface/src/scripting/GlobalServicesScriptingInterface.cpp
zzmp: This should be a `foreach(const auto&...`.

interface/src/ui/Stats.cpp
zzmp: This should be a `foreach(const auto&...`

libraries/animation/src/AnimNodeLoader.cpp
```diff
@@ -568,12 +568,13 @@ bool processStateMachineNode(AnimNode::Pointer node, const QJsonObject& jsonObj,
...
-    _resource = new Resource(url);
-    connect(_resource, &Resource::loaded, this, &AnimNodeLoader::onRequestDone);
-    connect(_resource, &Resource::failed, this, &AnimNodeLoader::onRequestError);
+    _url(url)
+{
+    _resource = QSharedPointer<Resource>::create(url);
+    _resource->setSelf(_resource);
+    connect(_resource.data(), &Resource::loaded, this, &AnimNodeLoader::onRequestDone);
+    connect(_resource.data(), &Resource::failed, this, &AnimNodeLoader::onRequestError);
+    _resource->ensureLoading();
```
zzmp: This is new - can you explain why it was not necessary before?

libraries/networking/src/ResourceCache.cpp
```diff
@@ -198,43 +198,62 @@ void ResourceCache::updateTotalSize(const qint64& oldSize, const qint64& newSize
...
-QList<Resource*> ResourceCacheSharedItems::getLoadingRequests() const {
-    Lock lock(_mutex);
-    return _loadingRequests;
+QList<QSharedPointer<Resource>> ResourceCacheSharedItems::getLoadingRequests() {
+    QList<QSharedPointer<Resource>> result;
+
+    {
+        Lock lock(_mutex);
+        foreach(QSharedPointer<Resource> resource, _loadingRequests) {
+            if (resource) {
```
zzmp: As long as we have a lock, why don't we purge loading resources that are no longer valid?

libraries/networking/src/ResourceCache.cpp
```diff
@@ -198,43 +198,62 @@ void ResourceCache::updateTotalSize(const qint64& oldSize, const qint64& newSize
...
-QList<QPointer<Resource>> ResourceCacheSharedItems::getPendingRequests() const {
-    Lock lock(_mutex);
-    return _pendingRequests;
+QList<QSharedPointer<Resource>> ResourceCacheSharedItems::getPendingRequests() {
+    QList<QSharedPointer<Resource>> result;
+
+    {
+        Lock lock(_mutex);
+        foreach(QSharedPointer<Resource> resource, _pendingRequests) {
+            if (resource) {
```
zzmp: As long as we have a lock, why don't we purge pending resources that are no longer valid?

libraries/networking/src/ResourceManager.cpp
```diff
@@ -110,6 +110,9 @@ ResourceRequest* ResourceManager::createResourceRequest(QObject* parent, const Q
     }
     Q_ASSERT(request);
 
+    if (parent) {
+        QObject::connect(parent, &QObject::destroyed, request, &QObject::deleteLater);
```
zzmp: Is this fixing something that was broken earlier?
Is this equivalent to actually setting `parent` as `request`'s parent, via Qt?

tests/networking/src/ResourceTests.cpp
```diff
@@ -55,7 +55,8 @@ void ResourceTests::downloadFirst() {
 
     // download the Mery fst file
     QUrl meryUrl = QUrl("http://hifi-public.s3.amazonaws.com/marketplace/contents/e21c0b95-e502-4d15-8c41-ea2fc40f1125/3585ddf674869a67d31d5964f7b52de1.fst");
```
zzmp: Did you run this test? I could not download this file.

libraries/networking/src/ResourceCache.cpp
```diff
@@ -401,7 +424,7 @@ void Resource::init() {
 
 void Resource::attemptRequest() {
     _startedLoading = true;
-    ResourceCache::attemptRequest(this);
+    ResourceCache::attemptRequest(_self);
```
zzmp: attemptRequest takes a QSharedPtr. attemptRequest should have a check in case the pointer is not valid.

zzmp:
requestCompleted and attemptRequest take a QSharedPtr. They should have checks in case the pointer given them is not valid.

They are internal (private, I think) methods, so I think the check should be in the method, and not before a call to the method.

libraries/animation/src/AnimNodeLoader.cpp
Atlante45: Resources used to start themselves in the actor by invoking `ensureLoading` which is redundant since ResourceCache already does it, so I removed it from the ctor.
It was only needed for standalone resources and causing issue, so I moved to this design until we remove all those.

libraries/networking/src/ResourceCache.cpp
Atlante45: I don't think this is the correct approach, dead resource should clean themselves up in `allReferenceCleared`, that way, dead weak ptr would never grow out of control, but this is outside the scope of that PR.

libraries/networking/src/ResourceManager.cpp
Atlante45: Fixing a potential memory leak.
Those use to be parented but now they are in different thread, so we need to do it this way.
Normally, I'd expect users to clean up the memory correctly, but I know some devs rely on the parenting pattern for that.

Atlante45:
I disagree, `attemptRequest` and `requestCompleted` are not method that should handle possibly null resources, you should only call them if you have a request to pass.

An assert might be more appropriate in my opinion. Generally I would pass in a `Resource&` here, but since we need access to the ref count, I couldn't do that.

tests/networking/src/ResourceTests.cpp
Atlante45: It was already broken, I just updated to the new pattern.

libraries/networking/src/ResourceCache.cpp
```diff
@@ -291,19 +319,14 @@ Resource::Resource(const QUrl& url, bool delayLoad) :
...
-    }
 }
 
 Resource::~Resource() {
     if (_request) {
         _request->disconnect(this);
         _request->deleteLater();
         _request = nullptr;
-        ResourceCache::requestCompleted(this);
+        ResourceCache::requestCompleted(_self);
```
zzmp: If you are already in the dtor, _self will not be able to lock as a QSharedPointer.

zzmp:
OK, I agree with your point about `attemptRequest`/`requestCompleted`, but I pointed out a case where the `QWeakPointer` cannot lock.

Fix that, and change the `foreach (X` to `foreach (const X` and it's good to go.

libraries/networking/src/ResourceCache.cpp
Atlante45: Good point, that's not a problem since it still shares the same ref count and that's the only thing we need in there but still.
We don't need it to be shared, so I'll just leave it as a QWeakPointer for requestCompleted.

hifi-gustavo:
Build console output for commit: 4ed8573f74479fe0da71c18676417cc1266f0e83 available at: https://gist.github.com/2f80c44c3481d139f1c869c3f2aadb72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7633.dmg

PR #7634, opened 2016-04-11
cleanup unmapped asset server files

birarda:
- add removal of unmapped asset server files on server startup
- removed unmapped asset files when mappings are deleted (if no other mapping exists for that hash)

### DIFF
--- a/assignment-client/src/assets/AssetServer.cpp
+++ b/assignment-client/src/assets/AssetServer.cpp
@@ -136,6 +136,8 @@ void AssetServer::completeSetup() {
 
     performMappingMigration();
 
+    cleanupUnmappedFiles();
+
     nodeList->addNodeTypeToInterestSet(NodeType::Agent);
 }
 
@@ -183,6 +185,34 @@ void AssetServer::performMappingMigration() {
                 } else {
                     qDebug() << "\tCould not add migration mapping for" << hash << "since a mapping for" << fakeFileName
                         << "already exists.";
+                }
+            }
+        }
+    }
+}
+
+void AssetServer::cleanupUnmappedFiles() {
+    QRegExp hashFileRegex { "^[a-f0-9]{" + QString::number(SHA256_HASH_HEX_LENGTH) + "}" };
+
+    auto files = _filesDirectory.entryInfoList(QDir::Files);
+
+    // grab the currently mapped hashes
+    auto mappedHashes = _fileMappings.values();
+
+    qDebug() << "Performing unmapped asset cleanup.";
+
+    for (const auto& fileInfo : files) {
+        if (hashFileRegex.exactMatch(fileInfo.fileName())) {
+            if (mappedHashes.contains(fileInfo.fileName())) {
+                qDebug() << "\tLeaving" << fileInfo.fileName() << "in asset files directory since it is mapped";
+            } else {
+                // remove the unmapped file
+                QFile removeableFile { fileInfo.absoluteFilePath() };
+
+                if (removeableFile.remove()) {
+                    qDebug() << "\tDeleted" << fileInfo.fileName() << "from asset files directory since it is unmapped.";
+                } else {
+                    qDebug() << "\tAttempt to delete unmapped file" << fileInfo.fileName() << "failed";
                 }
             }
         }
--- a/assignment-client/src/assets/AssetServer.cpp
+++ b/assignment-client/src/assets/AssetServer.cpp
@@ -147,7 +147,7 @@ void AssetServer::cleanupUnmappedFiles() {
     // grab the currently mapped hashes
     auto mappedHashes = _fileMappings.values();
 
-    qDebug() << "Performing unmapped asset cleanup.";
+    qInfo() << "Performing unmapped asset cleanup.";
 
     for (const auto& fileInfo : files) {
         if (hashFileRegex.exactMatch(fileInfo.fileName())) {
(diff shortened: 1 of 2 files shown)


birarda:
To test:

1. Run this Asset Server against an existing directory of assets that has some unmapped assets.
  - it should report that it has removed unmapped assets on startup
  - it should leave the still mapped assets intact

2. Add two files, one that has a single mapping and another uploaded twice so it has two mappings.
  - from the Asset Browser, delete the single mapping - the Asset Server should report that it removed the file
  - from the Asset Browser, delete one of the mappings for the doubly mapped file - the Asset Server should not report that it removed the file

ZappoMan:
So this is going to delete all the files on my asset server that I added before mapping was a feature? That will cause content loss on the Hell domain for sure. Won't it likely cause content loss for other alpha users?

birarda:
@ZappoMan - those files were migrated with mappings. 
Assuming you haven't gone in and intentionally deleted those they will not be removed.

ZappoMan:
Ok. I will test. You're saying that my old files have mappings. How does a file end up without a mapping?

birarda:
Currently the only way you should expect a file to end up without a mapping is if you explicitly hit the delete button for a path in the Asset Browser.

hifi-gustavo:
Build console output for commit: 334dc3cb6c1a25cbf9394ef73fac98a2fa690991 available at: https://gist.github.com/fbd2a4a23c1c205928f209b32e4c9401<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7634.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7634.dmg

OmegaHeron:
I'll be testing this as well - it makes sense in that on the first update where mapping deployed - each atp asset received a mapping to sha256.extension - which you've either changed, left same or deleted - this would, in theory, only remove an asset that's orphaned via having been deleted.  Will put under microscope once it compiles for Linux.

ZappoMan:
I got the following result when running this build...
```
[04/11 09:27:25] [INFO] [14120] [asset-server] There are 27 asset files in the asset directory.
[04/11 09:27:25] [INFO] [14120] [asset-server] Performing unmapped asset cleanup.
[04/11 09:27:25] [DEBUG] [14120] [asset-server] Deleted "0e508afbdc47ea94a401548d39d1effc3ac8c11f5aea2ebb121b1618bcd8243c" from asset files directory since it is unmapped.
```

I am not sure what file that is... but I know that at no point have I ever "explicitly hit the delete button for a path in the Asset Browser."

birarda:
Have you overwritten any assets in the Asset Browser by uploading a newer version?

ZappoMan:
not that I recall - but possibly - that seems more likely

OmegaHeron:
Tested Linux - I did a base catalog of assets placed by extracting a list from models.json - then correlated to mappings - then correlated to physical files present.  In my case nothing was removed that had an existent ATP->Mapped entry.  Extra credit - I used a snapshot of a setup made before ATP was enhanced with mappings/browser and confirmed it handled migrating it to mapped files and didn't remove anything.  A change like this is fraught with peril, but, it behaved safely for me.  :+1:

assignment-client/src/assets/AssetServer.cpp
```diff
@@ -134,55 +134,31 @@ void AssetServer::completeSetup() {
 
     qInfo() << "There are" << hashedFiles.size() << "asset files in the asset directory.";
 
-    performMappingMigration();
+    cleanupUnmappedFiles();
```
huffman: I think we need to return early if `loadMappingsFromFile` fails, unless there is some way that it is returning early that I'm not seeing?

assignment-client/src/assets/AssetServer.cpp
```diff
@@ -605,6 +589,30 @@ bool AssetServer::deleteMappings(AssetPathList& paths) {
...
         // persistence succeeded we are good to go
+
+        // grab the current mapped hashes
+        auto mappedHashes = _fileMappings.values();
+
+        // enumerate the mapped hashes and clear the list of hashes to check for anything that's present
+        for (auto& hashVariant : mappedHashes) {
+            auto it = hashesToCheck.find(hashVariant.toString());
+            if (it != hashesToCheck.end()) {
+                hashesToCheck.erase(it);
```
huffman: Any reason to not do the delete here if `it == hashesToCheck.end()`? It would remove the need to erase from the set.

assignment-client/src/assets/AssetServer.cpp
```diff
@@ -566,6 +542,8 @@ bool AssetServer::deleteMappings(AssetPathList& paths) {
     // take a copy of the current mappings in case persistence of these deletes fails
     auto oldMappings = _fileMappings;
 
+    QSet<QString> hashesToCheck;
```
huffman: Maybe too verbose, but `hashesToCheckForDeletion` would probably be more clear to me if I were reading this out of context.

assignment-client/src/assets/AssetServer.cpp
birarda: Maybe I don't understand exactly what you mean here - There's nothing to delete if `it == hashesToCheck.end()` since it just means that the mapped hash wasn't one of the ones I needed to check. 

I can either enumerate the hashes to check and look for each in the current mappings (enumerating each time), or I can enumerate the current mappings once to cut down the list and then do the final clear.

hifi-gustavo:
Build console output for commit: 190d11e0bf68a6facac1b4cfe1086893eddd1b9a available at: https://gist.github.com/fbd2a4a23c1c205928f209b32e4c9401<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7634.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7634.dmg

huffman:
re: doing deletion earlier, the comment is gone, but you were right, I had it backwards. This looks good.

PR #7649, opened 2016-04-12
away.js: fix for permanent disabling avatar IK

hyperlogic:
The kneel animation is now triggered via MyAvatar.overrideAnimation() and MyAvatar.restoreAnimation() calls.  This is much simpler than using the animation state handler callback to animate IK via animation variables.

### DIFF
--- a/examples/away.js
+++ b/examples/away.js
@@ -34,49 +34,23 @@ var OVERLAY_DATA_HMD = {
     alpha: 1,
     scale: 2,
     isFacingAvatar: true,
-    drawInFront: true   
+    drawInFront: true
 };
 
-// ANIMATION
-// We currently don't have play/stopAnimation integrated with the animation graph, but we can get the same effect
-// using an animation graph with a state that we turn on and off through the animation var defined with that state.
-var awayAnimationHandlerId, activeAnimationHandlerId, stopper;
+var AWAY_INTRO = {
+    url: "http://hifi-content.s3.amazonaws.com/ozan/dev/anim/standard_anims_160127/kneel.fbx",
+    playbackRate: 30.0,
+    loopFlag: false,
+    startFrame: 0.0,
+    endFrame: 83.0
+};
+
 function playAwayAnimation() {
-    function animateAway() {
-        return {isAway: true, isNotAway: false, isNotMoving: false, ikOverlayAlpha: 0.0};
-    }
-    if (stopper) {
-        stopper = false;
-        MyAvatar.removeAnimationStateHandler(activeAnimationHandlerId); // do it now, before making new assignment
-    }
-    awayAnimationHandlerId = MyAvatar.addAnimationStateHandler(animateAway, null);
+    MyAvatar.overrideAnimation(AWAY_INTRO.url, AWAY_INTRO.playbackRate, AWAY_INTRO.loopFlag, AWAY_INTRO.startFrame, AWAY_INTRO.endFrame);
 }
+
 function stopAwayAnimation() {
-    MyAvatar.removeAnimationStateHandler(awayAnimationHandlerId);
-    if (stopper) {
-        print('WARNING: unexpected double stop');
-        return;
-    }
-    // How do we know when to turn ikOverlayAlpha back on?
-    // It cannot be as soon as we want to stop the away animation, because then things will look goofy as we come out of that animation.
-    // (Imagine an away animation that sits or kneels, and then stands back up when coming out of it. If head is at the HMD, then it won't
-    //  want to track the standing up animation.)
-    // The anim graph will trigger awayOutroOnDone when awayOutro is finished.
-    var backToNormal = false;
-    stopper = true;
-    function animateActive(state) {
-        if (state.awayOutroOnDone) {
-            backToNormal = true;
-            stopper = false;
-        } else if (state.ikOverlayAlpha) {
-            // Once the right state gets reflected back to us, we don't need the hander any more.
-            // But we are locked against handler changes during the execution of a handler, so remove asynchronously.
-            Script.setTimeout(function () { MyAvatar.removeAnimationStateHandler(activeAnimationHandlerId); }, 0);
-        }
-        // It might be cool to "come back to life" by fading the ik overlay back in over a short time. But let's see how this goes.
-        return {isAway: false, isNotAway: true, ikOverlayAlpha: backToNormal ? 1.0 : 0.0}; // IWBNI we had a way of deleting an anim var.
-    }
-    activeAnimationHandlerId = MyAvatar.addAnimationStateHandler(animateActive, ['ikOverlayAlpha', 'awayOutroOnDone']);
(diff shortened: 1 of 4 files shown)


hyperlogic:
https://app.asana.com/0/45663370065758/110729759800124

hifi-gustavo:
Build console output for commit: 87cba810a72530c7f79bb3413dc81d5200a06044 available at: https://gist.github.com/242c18638e3c7ba5ecc151820191b55d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7649.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7649.dmg

cozza13:
QA Pass.

libraries/animation/src/Rig.cpp
```diff
@@ -48,36 +48,47 @@ const glm::vec3 DEFAULT_NECK_POS(0.0f, 0.70f, 0.0f);
 
 void Rig::overrideAnimation(const QString& url, float fps, bool loop, float firstFrame, float lastFrame) {
```
howard-stearns: It seems like all of the changes here and in Rig.h are not fixing the P2 bug, but rather are changing the internal implementation for the purpose of cleaning it up. How do you know we're not breaking something? There doesn't seem to be any test instructions.

examples/away.js
```diff
@@ -34,49 +34,26 @@ var OVERLAY_DATA_HMD = {
...
-    MyAvatar.removeAnimationStateHandler(awayAnimationHandlerId);
-    if (stopper) {
-        print('WARNING: unexpected double stop');
-        return;
-    }
-    // How do we know when to turn ikOverlayAlpha back on?
-    // It cannot be as soon as we want to stop the away animation, because then things will look goofy as we come out of that animation.
-    // (Imagine an away animation that sits or kneels, and then stands back up when coming out of it. If head is at the HMD, then it won't
-    //  want to track the standing up animation.)
-    // The anim graph will trigger awayOutroOnDone when awayOutro is finished.
```
howard-stearns: It is not clear to me how the above is addressed. Specifically, how does IK avoid stretching the user's neck or pulling their hands up during the transition out of away. Maybe IK is never involved until the blend from overrideAnimation to normal is completed?

examples/away.js
```diff
@@ -34,49 +34,26 @@ var OVERLAY_DATA_HMD = {
...
-        } else if (state.ikOverlayAlpha) {
-            // Once the right state gets reflected back to us, we don't need the hander any more.
-            // But we are locked against handler changes during the execution of a handler, so remove asynchronously.
-            Script.setTimeout(function () { MyAvatar.removeAnimationStateHandler(activeAnimationHandlerId); }, 0);
-        }
-        // It might be cool to "come back to life" by fading the ik overlay back in over a short time. But let's see how this goes.
-        return {isAway: false, isNotAway: true, ikOverlayAlpha: backToNormal ? 1.0 : 0.0}; // IWBNI we had a way of deleting an anim var.
-    }
-    activeAnimationHandlerId = MyAvatar.addAnimationStateHandler(animateActive, ['ikOverlayAlpha', 'awayOutroOnDone']);
+    MyAvatar.restoreAnimation();
```
howard-stearns: The old behavior used artist-defined animation to stand up (frames 84-167). It looks like now we're just blending from the crouched position to normal. Is that intended? Desirable? How does it look? And if it does look ok, why even bother with the crouching animation at all -- just blend directly to crouched position, no?

examples/away.js
```diff
@@ -34,49 +34,26 @@ var OVERLAY_DATA_HMD = {
...
 function playAwayAnimation() {
-    function animateAway() {
-        return {isAway: true, isNotAway: false, isNotMoving: false, ikOverlayAlpha: 0.0};
-    }
-    if (stopper) {
-        stopper = false;
-        MyAvatar.removeAnimationStateHandler(activeAnimationHandlerId); // do it now, before making new assignment
-    }
-    awayAnimationHandlerId = MyAvatar.addAnimationStateHandler(animateAway, null);
+    MyAvatar.overrideAnimation(AWAY_INTRO.url, AWAY_INTRO.playbackRate, AWAY_INTRO.loopFlag, AWAY_INTRO.startFrame, AWAY_INTRO.endFrame);
```
howard-stearns: The old behavior looped over an idle animation specified in the anim graph son. This means that we or a user could have an away idle animation (e.g., gently rocking back and forth or something), just by changing the animation graph, without any programming change (which would require deactivating default scripts so that the user could run their own custom away script, which means they wouldn't get any new default scripts in future versions).  Is it ok to lose that?

examples/away.js
hyperlogic: I'm not sure what you're asking here, but when overrideAnimation is called the system will smoothly interpolate between the current pose, which can include IK, to the first frame of the specified animation without IK.

examples/away.js
hyperlogic: Yeah, it is intended.  I think it's desirable as well, for a few reasons.  
   1. Previously, if you quickly enter and exit away mode the avatar will wait for the entire 3+ second kneel and standing animation to complete.  With the new code the transition is quicker and smoother.
   2. The standing animation was too long, and kind of creepy when you were wearing an HMD. 
   3. Code simplicity and robustness for MVP.  I could add a separate call to overrideAnimation() for the standing transition, however I elected not to, because this requires an animationState handler to catch OnDone events, as well as a setTimeout() to handle the case where the OnDone does not trigger due to a avatar reset, or avatar URL change.

examples/away.js
hyperlogic: Yep, in practice though, we weren't using the idle, even after 6 months... so yeah, if we need it we can add it back.  You are right about losing the ability for users to customize their away animations via the avatar-animation.json, but they can just as easily modify the url in away.js.

hyperlogic: You know on second thought, I think we do need a way for avatar creators to specify an animation for the away kneel.  The most obvious place is in the animation.json.... Here's a compromise.  What if, I add it back to the animation.json in a future PR, but, without adding all of the complex state transitions from every state into and out of the idle animation.  In that case, away.js can just look up the animation url from the json, without using animVars to trigger it.  What do you think?

howard-stearns: sounds good.

examples/away.js
howard-stearns: Spoke with Tony. The answer to my question is that the old animation graph was applying IK to the result of the blended idle-vs-away, while the new is blending between the away clip vs the application of ik to idle. So there's no problem. Nice.

libraries/animation/src/Rig.cpp
howard-stearns: Spoke with Tony. The answer is that reset is now remembering the user state and re-applying it, rather than resetting everything back to "freshly loaded".  The changes below this point are for that purpose.

PR #7650, opened 2016-04-12
Clear atp mapping cache between domain

Atlante45:
make sure the the cached mappings are flushed on domain change.


Atlante45:
https://app.asana.com/0/26551373415487/106561590021925

Atlante45:
Testing plan:
 Put 2 different models in two different domains, but give them the same name on ATP.
switch back and forth between the domain and notice the now we display the correct model in each domain.

hifi-gustavo:
Build console output for commit: 0f62d5c9979d4718ae6637f503a3021ba150f3fa available at: https://gist.github.com/92257ea7a8d203ecd28829526deb6392<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7650.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7650.dmg

kevinmthomas-carpool:
## QA Test
**PR Build:** 7650
**Testing OS:** Windows 10
**Testing Locations:** Music, Sandbox
**Additional Test Results Requested:**  Successful test uploaded two different models with the name (kevin.fbx) on the ATP each on the Music domain (uploaded a drum set named kevin.fbx) and the Sandbox domain (uploaded a building named kevin.fbx) to which the correct model was displayed correctly in each domain.

libraries/networking/src/ResourceCache.cpp
```diff
@@ -20,39 +20,88 @@
...
 #define clamp(x, min, max) (((x) < (min)) ? (min) :\
                            (((x) > (max)) ? (max) :\
                                             (x)))
 
-ResourceCache::ResourceCache(QObject* parent) :
-    QObject(parent) {    
+ResourceCache::ResourceCache(QObject* parent) : QObject(parent) {
+    auto& domainHandler = DependencyManager::get<NodeList>()->getDomainHandler();
+    connect(&domainHandler, &DomainHandler::disconnectedFromDomain,
+            this, &ResourceCache::clearATPAssets, Qt::DirectConnection);
```
ZappoMan: Are we confident that this won't cause deadlock? I assume with direct connection, you expect that the node list/domain handler and the resource cache on on different threads (or else Auto would give you same behavior)... and since clearATPAssets requires a lock, are we sure that there won't be some other thread already locking the resource cache that also needs access to the node list?

Atlante45: Yes, AFAICT no locks of any kind are held when `disconnectedFromDomain` is emitted, so there's no possibility for a dead lock.
Also, these locks in ResourceCache protect simple, low-level data structures, they should never be held for anything more than quickly reading/modifying those data structure. I made sure this was the case here.

hifi-gustavo:
Build console output for commit: b3b2f98f6cf60226aa4a63730dc6192b7e041b1e available at: https://gist.github.com/92257ea7a8d203ecd28829526deb6392<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7650.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7650.dmg

PR #7654, opened 2016-04-13
new default LOD settings, and support for reset settings without crashing

ZappoMan:


ZappoMan:
Testing Notes:
- Smoke test
- Check the Developer > Ask To Reset Settings option, and verify that on next startup you are asked if you want to reset your settings
- Verify that after a crash you are still asked to reset your settings.
- Verify that wording about reset settings is slightly different between crash vs just ask.
- Verify that after resetting your settings your minimum FPS for LOD is now 20.

hifi-gustavo:
Build console output for commit: 93f8cfc80a88fdde27ad938ebb9682d240608b8a available at: https://gist.github.com/d892cf21b7dd6502b9f8c16768d7a9f9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7654.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7654.dmg

OmegaHeron:
Tested Windows 7
General smoke test seems fine - no difference seen to production in frame rates, function etc.

Check the Developer > Ask To Reset Settings option, and verify that on next startup you are asked if you want to reset your settings.  **Yes**
Verify that after a crash you are still asked to reset your settings.  **Yes**
Verify that wording about reset settings is slightly different between crash vs just ask.  **Yes**
Verify that after resetting your settings your minimum FPS for LOD is now 20.  **Yes**

I would consider this operator error, but, if you have reset settings option check and you don't actually reset them - then it offers that dialog each launch until you 1) reset settings or 2) un-check option.

ZappoMan:
@OmegaHeron - re: "if you have reset settings option check and you don't actually reset them - then it offers that dialog each launch until you 1) reset settings or 2) un-check option." -- that was the design I intended...

interface/src/LODManager.h
```diff
@@ -18,8 +18,8 @@
 #include <PIDController.h>
 #include <SimpleMovingAverage.h>
 
-const float DEFAULT_DESKTOP_LOD_DOWN_FPS = 30.0;
-const float DEFAULT_HMD_LOD_DOWN_FPS = 45.0;
+const float DEFAULT_DESKTOP_LOD_DOWN_FPS = 20.0;
```
zzmp: Does this fix belong here, or in a separate PR?

ZappoMan: This is the actual fix - the other is a feature request.

interface/src/CrashHandler.cpp
```diff
@@ -28,16 +28,23 @@
 static const QString RUNNING_MARKER_FILENAME = "Interface.running";
 
 bool CrashHandler::checkForAndHandleCrash() {
```
zzmp: This method now does more than its name implies.

zzmp:
It looks like it is only possible to reset settings on launch, and that's why you implemented it as a dialogue on next startup. Is this true?

interface/src/CrashHandler.cpp
ZappoMan: hmmm - correct, and in inspecting it, I noticed a logic error.

ZappoMan:
@zzmp - re: resetting on startup... yes - that was the easiest way to handle things... resetting all settings mid-flight would be difficult.

PR #7664, opened 2016-04-14
only export joint values if they've been explicitly set by scripts

ZappoMan:
This partially fixes the problem with exporting/importing and duplicating models that are being animated. Basically, if a model was animated, the client thought that the joint rotations and translations had been set and was including those in the properties, that was wasteful and caused the exported JSONs and property sets to be larger than they need to be. That would ultimately cause the "addEntity" to fail if it's properties couldn't fit in a single packet.

This fixes the erroneous inclusion of all the extra joint data.

### DIFF
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -748,6 +748,7 @@ glm::vec3 RenderableModelEntityItem::getAbsoluteJointTranslationInObjectFrame(in
 bool RenderableModelEntityItem::setAbsoluteJointRotationInObjectFrame(int index, const glm::quat& rotation) {
     bool result = false;
     _jointDataLock.withWriteLock([&] {
+        _jointRotationsExplicitlySet = true;
         resizeJointArrays();
         if (index >= 0 && index < _absoluteJointRotationsInObjectFrame.size() &&
             _absoluteJointRotationsInObjectFrame[index] != rotation) {
@@ -764,6 +765,7 @@ bool RenderableModelEntityItem::setAbsoluteJointRotationInObjectFrame(int index,
 bool RenderableModelEntityItem::setAbsoluteJointTranslationInObjectFrame(int index, const glm::vec3& translation) {
     bool result = false;
     _jointDataLock.withWriteLock([&] {
+        _jointTranslationsExplicitlySet = true;
         resizeJointArrays();
         if (index >= 0 && index < _absoluteJointTranslationsInObjectFrame.size() &&
             _absoluteJointTranslationsInObjectFrame[index] != translation) {
--- a/libraries/entities/src/EntityItemProperties.cpp
+++ b/libraries/entities/src/EntityItemProperties.cpp
@@ -541,13 +541,6 @@ QScriptValue EntityItemProperties::copyToScriptValue(QScriptEngine* engine, bool
     COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_LOCAL_POSITION, localPosition);
     COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_LOCAL_ROTATION, localRotation);
 
-    COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_JOINT_ROTATIONS_SET, jointRotationsSet);
-    COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_JOINT_ROTATIONS, jointRotations);
-    COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_JOINT_TRANSLATIONS_SET, jointTranslationsSet);
-    COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_JOINT_TRANSLATIONS, jointTranslations);
-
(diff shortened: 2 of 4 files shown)


libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -541,13 +541,6 @@ QScriptValue EntityItemProperties::copyToScriptValue(QScriptEngine* engine, bool
     COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_LOCAL_POSITION, localPosition);
     COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_LOCAL_ROTATION, localRotation);
 
-    COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_JOINT_ROTATIONS_SET, jointRotationsSet);
```
ZappoMan: these were all duplicates, they are already included above... this was just wasted code.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -748,6 +748,7 @@ glm::vec3 RenderableModelEntityItem::getAbsoluteJointTranslationInObjectFrame(in
 bool RenderableModelEntityItem::setAbsoluteJointRotationInObjectFrame(int index, const glm::quat& rotation) {
     bool result = false;
     _jointDataLock.withWriteLock([&] {
+        _jointRotationsExplicitlySet = true;
```
ZappoMan: we now track if you explicitly set the joint rotations/translations and only include those properties if you've actually set them.. this way, when animations are setting the values, we don't erroneously treat it as if the were set as properties.

libraries/entities/src/ModelEntityItem.cpp
```diff
@@ -447,31 +451,40 @@ void ModelEntityItem::setJointTranslationsSet(const QVector<bool>& translationsS
...
+        }
     });
     return result;
 }
 
 QVector<bool> ModelEntityItem::getJointRotationsSet() const {
     QVector<bool> result;
     _jointDataLock.withReadLock([&] {
-        result = _absoluteJointRotationsInObjectFrameSet;
+        if (_jointRotationsExplicitlySet) {
```
ZappoMan: we now track if you explicitly set the joint rotations/translations and only include those properties if you've actually set them

ZappoMan:
Test Plan:
- smoke test
- Create an entity that is being animated.
- Select it and "export" it... check the resulting json file, you should not see: `"jointRotations": [` properties
- Run the `setupDressingroomScipt.js` that will create a doppelganger that will mimic your avatar moves... 
- select the doppelganger model, and export it, check the resulting json file, you SHOULD see: `"jointRotations": [` properties

hifi-gustavo:
Build console output for commit: b48503a82d9af4a6842ec88be9fc538b86c87b71 available at: https://gist.github.com/a11655e9f605eb4079ac4bed709206d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7664.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7664.dmg

howard-stearns:
Just to check my own understanding: this kicks the can (a long way) down the road as to failures that legitimately don't fit in a packet?

ZappoMan:
This change is more that just about kicking the can down the road.

The joint properties were in fact never set, and so the getProperties in the client were violating rule #1 that the client and server should always match. 

The export of an animated entity was wrong. Plain and simple. This fixes that bug. 

It has a side effect of making most entities more exportable and duplicatable. But it doesn't fix that bug completely. 

I have another different bug assigned to me to fix that problem.

howard-stearns:
Excellent.

kevinmthomas-carpool:
## QA Test
**PR Build:** 7664
**Testing OS:** Windows 10
**Testing Locations:** localhost, sandbox
**Additional Test Results Requested:** Successful test.  Imported an animated fbx and successfully exported and examined JSON.  Confirmed there is no ```jointRotations``` properties (JSON included as an attachment in this PR if a more detailed look is necessary).  Partially successful test regarding the ```setupDressingroomScipt.js``` as the Albert avatar successfully mimiced my avatar's behavior however his joint rotation was incorrect, screen shot included in the attachment for further analysis).  Successful test of dragging and duplicating several entities.  Unsuccessful test of regarding the ```setupDressingroomScipt.js``` with a second avatar present @thoys we tested with the @thoys avatar successfully mimiced my avatar's behavior however his joint rotation was incorrect, screen shot included in the attachment for further analysis).  Attempting to change my avatar to the Albert avatar to attempt to properly test the true mimic behavior caused a system crash.  
**Successful Logout/Crash:** Failure/crash, log file enclosed for further analysis.
[attachments.zip](https://github.com/highfidelity/hifi/files/220013/attachments.zip)

ZappoMan:
@kevinmthomas-carpool - regarding the joint rotations with the doppelganger... I noticed they are basically reversed... that is the same behavior as production (I think it's a bug in that script) regardless... this PR just needs to verify that the API still works... which it seems to.

Regarding the crash on exit, I just checked that on Bugsplat and it does not appear to be at all related to the code changed in this PR.

PR #7667, opened 2016-04-14
fix entity-server crash when entity moves out of bounds

AndrewMeadows:
* fix entity-server crash when entity moves out of bounds

### DIFF
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -926,7 +926,8 @@ void EntityItem::simulateKinematicMotion(float timeElapsed, bool setFlags) {
             glm::quat  dQ = computeBulletRotationStep(localAngularVelocity, dt);
             rotation = glm::normalize(dQ * rotation);
 
-            setRotation(rotation);
+            bool success;
+            setOrientation(rotation, success, false);
         }
 
         setLocalAngularVelocity(localAngularVelocity);
@@ -1983,10 +1984,10 @@ void EntityItem::locationChanged(bool tellPhysics) {
     requiresRecalcBoxes();
     if (tellPhysics) {
         _dirtyFlags |= Simulation::DIRTY_TRANSFORM;
-    }
-    EntityTreePointer tree = getTree();
-    if (tree) {
-        tree->entityChanged(getThisPointer());
+        EntityTreePointer tree = getTree();
+        if (tree) {
+            tree->entityChanged(getThisPointer());
+        }
     }
     SpatiallyNestable::locationChanged(tellPhysics); // tell all the children, also
 }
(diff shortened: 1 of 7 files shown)


AndrewMeadows:
This is a theoretical fix.  I wasn't able to reproduce the crash, although I tried pretty hard.

The problem (in theory) was that the entity-server could integrate an object's motion forward until it was out of bounds and then promptly delete it from a container, thereby invalidating some iterators over that same container in a higher context.

To test I created a bunch of kinematic (moving, but non-dynamic) objects just above the bottom boundary of the region.  I'd give them velocity and gravity toward the domain boundary and either (a) log out immediately, letting the entity-server decide to delete them all or (b) stick around and let the interface also send delete messages for the entities when the left the boundary.  Unfortunately I couldn't trigger a crash on this PR branch or on hifi/master.

hifi-gustavo:
Build console output for commit: c36f2186584fbd3c3c83c75288cc435e4ea570c6 available at: https://gist.github.com/5f52282c286c204c3f7d3edce1d7bf8b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7667.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7667.dmg

hifi-gustavo:
Build console output for commit: a263091c94e9a02d4e252df318b514854bf4732e available at: https://gist.github.com/5f52282c286c204c3f7d3edce1d7bf8b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7667.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7667.dmg

OmegaHeron:
Tested stack on Linux host - much like Andrew I could not create a crash with this or any other version of stack/interface available.  I can give this a pass in that it doesn't seem to make anything worse.  If anyone comes up with a way to reproduce crash it's aimed a fixing would be glad to re-test.

bluetoe:
I tried this with cubes, spheres, text, web and fbx models. All except the models were deleted when they crossed the -16384 point. But the model kept going and appeared to still be there after logging off and going back. When I check positions or teleport to them it shows the models at y=-16384 but they look visually to be still moving many meters below.

Eventually all the models are deleted, or at least they don't show up in the entity list. I had no crashes from this PR so I guess it passes. I will try doing the same thing with a model on the production build to see if it is the same.

bluetoe:
I just crashed the PR interface on quitting but this probably has nothing to do with the PR. And I did get the crash reporter.

bluetoe:
RC 4707 does the same thing. cubes and spheres are deleted at the boundry but an .fbx model looks like it keeps going and is not deleted. But no crash.

bluetoe:
Hmm, I just tried to quit interface 4707 and to log back on to see if the model persisted and when I logged back onto the domain it was disconnected. looking at logs now

bluetoe:
My mistake .. I guess. The domain is still running with the model slowly sinking. Very strange though, there seems to be an issue with localhost verses placename on login. I will try to repro it elsewhere.

PR #7678, opened 2016-04-15
use a steady clock for mixer timings

birarda:
- use the assumed steady p_high_resolution_clock for mixer block timings
- improve audio stats sending by spreading out stats packets to clients

### DIFF
--- a/assignment-client/src/audio/AudioMixerClientData.cpp
+++ b/assignment-client/src/audio/AudioMixerClientData.cpp
@@ -189,6 +189,10 @@ void AudioMixerClientData::checkBuffersBeforeFrameSend() {
     }
 }
 
+bool AudioMixerClientData::shouldSendStats(int frameNumber) {
+    return (frameNumber % (int) ceil(1.0f / AudioConstants::NETWORK_FRAME_SECS)) == _frameToSendStats;
+}
+
 void AudioMixerClientData::sendAudioStreamStatsPackets(const SharedNodePointer& destinationNode) {
 
     auto nodeList = DependencyManager::get<NodeList>();
--- a/assignment-client/src/audio/AudioMixerClientData.cpp
+++ b/assignment-client/src/audio/AudioMixerClientData.cpp
@@ -190,7 +190,7 @@ void AudioMixerClientData::checkBuffersBeforeFrameSend() {
 }
 
 bool AudioMixerClientData::shouldSendStats(int frameNumber) {
-    return (frameNumber % (int) ceil(1.0f / AudioConstants::NETWORK_FRAME_SECS)) == _frameToSendStats;
+    return frameNumber == _frameToSendStats;
 }
 
 void AudioMixerClientData::sendAudioStreamStatsPackets(const SharedNodePointer& destinationNode) {
--- a/assignment-client/src/audio/AudioMixerClientData.cpp
+++ b/assignment-client/src/audio/AudioMixerClientData.cpp
@@ -9,6 +9,8 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 
+#include <random>
+
 #include <QtCore/QDebug>
 #include <QtCore/QJsonArray>
 
@@ -26,7 +28,14 @@ AudioMixerClientData::AudioMixerClientData(const QUuid& nodeID) :
     _outgoingMixedAudioSequenceNumber(0),
     _downstreamAudioStreamStats()
 {
+    // of the ~94 blocks in a second of audio sent from the AudioMixer, pick a random one to send out a stats packet on
+    // this ensures we send out stats to this client around every second
+    // but do not send all of the stats packets out at the same time
+    std::random_device randomDevice;
+    std::mt19937 numberGenerator { randomDevice() };
+    std::uniform_int_distribution<> distribution { 1, (int) ceil(1.0f / AudioConstants::NETWORK_FRAME_SECS) };
 
+    _frameToSendStats = distribution(numberGenerator);
 }
 
 AvatarAudioStream* AudioMixerClientData::getAvatarAudioStream() {
--- a/assignment-client/src/audio/AudioMixerClientData.h
+++ b/assignment-client/src/audio/AudioMixerClientData.h
@@ -59,7 +59,7 @@ public:
     quint16 getOutgoingSequenceNumber() const { return _outgoingMixedAudioSequenceNumber; }
 
     // uses randomization to have the AudioMixer send a stats packet to this node around every second
-    bool shouldSendStats(int frameNumber) { return frameNumber % _frameToSendStats == 0; }
+    bool shouldSendStats(int frameNumber);
 
 signals:
     void injectorStreamFinished(const QUuid& streamIdentifier);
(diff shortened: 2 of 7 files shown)


birarda:
This is a fix for https://app.asana.com/0/11984398816645/106094631217033.

`usecTimestampNow()` is not a steady clock and breaks timing when external clock adjustment occurs. `p_high_resolution_clock` (which is `std::high_resolution_clock` on platforms other than windows) is steady (though not guaranteed this has been the case on all platforms I have tested) and does not break timing when there are external clock changes.

To test:

- Run either an audio-mixer or avatar-mixer from current release. Change the system clock on your machine backwards and forwards. You will see messages from either the `audio-mixer` or the `avatar-mixer` about the mixer struggling or recovering, with many strange values for time/percentage/audibility. If connected to the audio-mixer during these clock changes, you should have broken audio.
- Run the same test with this PR. The audio-mixer and avatar-mixer are resilient to clock changes. You should continue to hear correct audio while changing your system clock around.
- Ensure you get audio stats (Developer->Audio->Audio Network Stats) that keep updating around every second.

hifi-gustavo:
Build console output for commit: 38481aac5352ac6f911e953ce3fb63420db298c9 available at: https://gist.github.com/e591e8e14efc0256efc9a508cff2867a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7678.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7678.dmg

OmegaHeron:
Windows 7 - Having some really odd issue with this build.  Server console launches, it downloads content - presents the you're all set message but doe not leave a tray icon for accessing it.  Process manager shows 8 instances of assignment-client, 1 domain-server and 2 server-console processes.

I manually terminated all - relaunched server-console and same thing.  Going to uninstall - and start over clean again.

birarda:
It would surprise me for this to have caused issues with the tray icon of the Server Console. I will test the same here to make sure it seems to install/setup correctly.

OmegaHeron:
It did me as well - uninstalling, cleaning, rebooting etc.  Will report back.

birarda:
Installed and got tray icon here with expected 6 assignment-clients shown on domain-server dashboard. @OmegaHeron let me know what you get after reboot.

OmegaHeron:
Ok - it was a previous PR build where I had told server-console to exit and it left AC/DS hung - that seemed to confuse this build's install and etc etc etc.  I cleaned all that up, rebooted and reinstalled this PR.  Now it's fine.  Will report back testing results in a while - I plan to test this with Windows localhost then on my linux remotes.  That plus dinner time will take a while.

OmegaHeron:
Phase 1 - Windows stack + Windows Interface running on localhost
Current build - adjusting clock leads to problems.
PR Build - adjust clock leads to temporary loss of audio-mixer (very short) then it returns with proper sound.

Phase 2 - Linux stack + Windows Interface over Internet
Same as above.

That's extent to which I can comfortably test this.  I can say audio recovers gracefully from a +/- 1 hour time change on Windows and Linux for server side.  Much testing remains for any other impacts especially with other avatars present.  Also, note to self - remember to turn off NTP if doing this on Linux or it doesn't do much good to change time.

assignment-client/src/audio/AudioMixerClientData.h
```diff
@@ -58,6 +58,9 @@ class AudioMixerClientData : public NodeData {
     void incrementOutgoingMixedAudioSequenceNumber() { _outgoingMixedAudioSequenceNumber++; }
     quint16 getOutgoingSequenceNumber() const { return _outgoingMixedAudioSequenceNumber; }
 
+    // uses randomization to have the AudioMixer send a stats packet to this node around every second
+    bool shouldSendStats(int frameNumber) { return frameNumber % _frameToSendStats == 0; }
```
huffman: If I understand the intent, I think what you actually want here is: `return (frameNumber % (int) ceil(1.0f / AudioConstants::NETWORK_FRAME_SECS)) == _frameToSendStats;`

assignment-client/src/avatars/AvatarMixerClientData.h
```diff
@@ -45,11 +48,11 @@ class AvatarMixerClientData : public NodeData {
 
     uint16_t getLastReceivedSequenceNumber() const { return _lastReceivedSequenceNumber; }
 
-    quint64 getBillboardChangeTimestamp() const { return _billboardChangeTimestamp; }
-    void setBillboardChangeTimestamp(quint64 billboardChangeTimestamp) { _billboardChangeTimestamp = billboardChangeTimestamp; }
+    HRCTime getBillboardChangeTimestamp() const { return _billboardChangeTimestamp; }
+    void flagBillboardChange() { _billboardChangeTimestamp = p_high_resolution_clock::now(); }
```
huffman: There doesn't seem to be any change in behavior here, but I'm curious where this gets set back to 0 / epoch ? I couldn't seem to find that anywhere.

birarda: It doesn't, but we don't need it to? It's initialized with `time_since_epoch` at 0 and then in the AvatarMixer code we're just making sure it's change since we last sent.

In any case - I hear we don't use the billboards so I'm going to drop the handling code from the mixer.

hifi-gustavo:
Build console output for commit: 1d256be7ee5e04825d24f6a80bbb5bbd8cedf8af available at: https://gist.github.com/e591e8e14efc0256efc9a508cff2867a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7678.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7678.dmg

assignment-client/src/audio/AudioMixerClientData.cpp
```diff
@@ -180,6 +189,10 @@ void AudioMixerClientData::checkBuffersBeforeFrameSend() {
     }
 }
 
+bool AudioMixerClientData::shouldSendStats(int frameNumber) {
+    return (frameNumber % (int) ceil(1.0f / AudioConstants::NETWORK_FRAME_SECS)) == _frameToSendStats;
```
huffman: One other issue: The sign of `x % y` is not well defined (afaik) if x or y is negative.

Couple solutions:

1. Make `frameNumber` unsigned.
2. Wrap `frameNumber` in broadcastMixes so that it is bounded between 0 and the number of frames per second. ie `frameNumber = ++frame % (int) ceil(1.0f / AudioConstants::NETWORK_FRAME_SECS)`. And then do `frameNumber == _frameToSendStats` here

assignment-client/src/avatars/AvatarMixerClientData.h
```diff
@@ -86,8 +86,7 @@ class AvatarMixerClientData : public NodeData {
     std::unordered_map<QUuid, uint16_t> _lastBroadcastSequenceNumbers;
     std::unordered_set<QUuid> _hasReceivedFirstPacketsFrom;
 
-    quint64 _billboardChangeTimestamp = 0;
-    quint64 _identityChangeTimestamp = 0;
+    p_high_resolution_clock::time_point _identityChangeTimestamp;
```
huffman: Seems like we should use `HRCTime` here?

assignment-client/src/avatars/AvatarMixerClientData.h
huffman: Yep, you are right!

hifi-gustavo:
Build console output for commit: 93cf5a103518531561c49d64608241672f1b7283 available at: https://gist.github.com/e591e8e14efc0256efc9a508cff2867a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7678.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7678.dmg

PR #7681, opened 2016-04-15
Memory leak on exit for CharacterController

hyperlogic:
Delete the btRigidBody and it's collision shape in the CharacterController destructor.

Note: there is only one instance of this class and it's lifetime is the same as the MyAvatar instance.

### DIFF
--- a/libraries/physics/src/CharacterController.cpp
+++ b/libraries/physics/src/CharacterController.cpp
@@ -72,6 +72,17 @@ CharacterController::CharacterController() {
     _pendingFlags = PENDING_FLAG_UPDATE_SHAPE;
 }
 
+CharacterController::~CharacterController() {
+    if (_rigidBody) {
+        btCollisionShape* shape = _rigidBody->getCollisionShape();
+        if (shape) {
+            delete shape;
+        }
+        delete _rigidBody;
+        _rigidBody = nullptr;
+    }
+}
+
 bool CharacterController::needsRemoval() const {
     return ((_pendingFlags & PENDING_FLAG_REMOVE_FROM_SIMULATION) == PENDING_FLAG_REMOVE_FROM_SIMULATION);
 }
--- a/libraries/physics/src/CharacterController.h
+++ b/libraries/physics/src/CharacterController.h
@@ -36,7 +36,7 @@ class btDynamicsWorld;
 class CharacterController : public btCharacterControllerInterface {
 public:
     CharacterController();
-    virtual ~CharacterController() {}
+    virtual ~CharacterController();
 
     bool needsRemoval() const;
     bool needsAddition() const;


hyperlogic:
https://app.asana.com/0/45663370065758/112973471736194

hifi-gustavo:
Build console output for commit: 0b7a5c4df83ff1425359f78042c87cfa7c7142bc available at: https://gist.github.com/fbdb6bda67cc935b8062fe7cc8fe513c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7681.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7681.dmg

libraries/physics/src/CharacterController.cpp
```diff
@@ -72,6 +72,17 @@ CharacterController::CharacterController() {
...
 }
 
+CharacterController::~CharacterController() {
+    if (_rigidBody) {
+        btCollisionShape* shape = _rigidBody->getCollisionShape();
+        if (shape) {
+            delete shape;
+        }
+        delete _rigidBody;
+        _rigidBody = nullptr;
```
zzmp: Why is it important to set it back to `nullptr` if it's already freed?

libraries/physics/src/CharacterController.cpp
```diff
@@ -72,6 +72,17 @@ CharacterController::CharacterController() {
     _pendingFlags = PENDING_FLAG_UPDATE_SHAPE;
 }
 
+CharacterController::~CharacterController() {
+    if (_rigidBody) {
+        btCollisionShape* shape = _rigidBody->getCollisionShape();
+        if (shape) {
+            delete shape;
```
zzmp: Why isn't this done in the dtor for the `_rigidBody`?

hyperlogic: _rigidBody is a class from the Bullet physics library.  They don't use smart ptrs, and they also avoid doing deletes, mostly because it gives people the option to statically allocate pools of objects instead of using the heap.

libraries/physics/src/CharacterController.cpp
hyperlogic: It's not strictly necessary, I'm just being extra careful...  Sometimes this can help detect use after free issues.  Because in that case the pointer will be null which will lead to crashes.

cozza13:
What is the best way to test this? Smoke test?

ZappoMan:
Test Plan:
- Basic Smoke Test

OmegaHeron:
General smoke test Windows 7
All basic functions match what I see in stable - couple of crashes at playa recorded, but, those are like same as seen in stable, bug splats sent just in case.  QA Pass

PR #7684, opened 2016-04-15
Fix ResourceCacheSharedItems::_loadingRequests removals

zzmp:
- Moves ResourceCacheSharedItems implementation to top of impl file to match the header file.
- Fixes removal from _loadingRequests for the case when it requests are freed (freed weak pointers cannot be used as a key for removal).

### DIFF
--- a/libraries/networking/src/ResourceCache.cpp
+++ b/libraries/networking/src/ResourceCache.cpp
@@ -73,7 +73,18 @@ QList<QSharedPointer<Resource>> ResourceCacheSharedItems::getLoadingRequests() {
 
 void ResourceCacheSharedItems::removeRequest(QWeakPointer<Resource> resource) {
     Lock lock(_mutex);
-    _loadingRequests.removeAll(resource);
+    // resource can only be removed if it still has a ref-count, as
+    // QWeakPointer has no operator== implementation for two weak ptrs, so
+    // manually loop in case resource has been freed.
+    for (int i = 0; i < _loadingRequests.size();) {
+        auto request = _loadingRequests.at(i);
+        // Clear our resource and any freed resources
+        if (!request || request.data() == resource.data()) {
+            _loadingRequests.removeAt(i);
+            continue;
+        }
+        i++;
+    }
 }
 
 QSharedPointer<Resource> ResourceCacheSharedItems::getHighestPendingRequest() {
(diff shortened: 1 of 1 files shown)


zzmp:
#### Testing

You will not be able to test this without breaking through the code.
If you want to do that, you can break at `removeRequest` and make sure that the `_loadingRequests` member is no longer growing without bound. You will need to make lots of requests to do this, with many being invalidated. I repro'ed it using a refresh script made by @sethalves ([reset.txt](https://github.com/highfidelity/hifi/files/221949/reset.txt)). Use this as an entity script with Tidy Guy in the new Home content, and push the tidy up button - it will repeatedly refresh. *There is an as yet undiscovered memory leak here, so looking at the memory usage is not an appropriate test!*

hifi-gustavo:
Build console output for commit: 6279d4b014c8b7a419e2a2d6cf20baa7ce622e27 available at: https://gist.github.com/33ca4d60610238703260496f5cbbbab5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7684.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7684.dmg

libraries/networking/src/ResourceCache.cpp
```diff
@@ -28,6 +28,99 @@
...
+            // Clear any freed resources
+            auto resource = _loadingRequests.at(i).lock();
+            if (!resource) {
+                _loadingRequests.removeAt(i);
+                continue;
+            }
+            i++;
+        }
+    }
+}
```
Atlante45: Since you don't actually need to increase the ref count, you can make this faster by checking:
```cpp
if (!_loadingRequests.at(i)) {
    _loadingRequests.removeAt(i);
    continue;
}
```

Also since you're looping once any way (`removeAll`), you might as well do both in one loop:

```cpp
void ResourceCacheSharedItems::removeRequest(QWeakPointer<Resource> resource) {
    Lock lock(_mutex);
    for (int i = 0; i < _loadingRequests.size();) {
        // Clear any freed resources
        auto req = _loadingRequests.at(i);
        if (!req || req.data() == resource.data()) {
            _loadingRequests.removeAt(i);
            continue;
        }
        ++i;
    }
}
```

hifi-gustavo:
Build console output for commit: 0d06d8ea936660b7e6464ffc5ede1ba0da4d0821 available at: https://gist.github.com/33ca4d60610238703260496f5cbbbab5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7684.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7684.dmg

sethalves:
I've run this on windows and with asan on Linux, it looks good.

PR #7690, opened 2016-04-16
Introduce "recommended overlay rectangle" to display plugins

ZappoMan:
- allow display plugins to recommend a preferred rectangle for placement of windows and overlays
- have toolbar.js honor that
- have QML windows honor that


ZappoMan:
Test Plan:
- smoke test
- when you resize the desktop window the edit and examples should still follow stay inside the window area
- when you switch to HMD mode, the mute and the edit toolbars should be "closer to the center of your view" 
- users.js is no longer on by default

hifi-gustavo:
Build console output for commit: d5a46d3390a3859fadfcb24420f9837978e2ad14 available at: https://gist.github.com/2ba42ec0c9b1985a9939b76ce128be08<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7690.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7690.dmg

interface/resources/qml/desktop/Desktop.qml
```diff
@@ -21,8 +21,29 @@ FocusScope {
     objectName: "desktop"
     anchors.fill: parent
 
-    onHeightChanged: d.repositionAll();
-    onWidthChanged: d.repositionAll();
+    property vector4d recommendedRect: vector4d(0,0,0,0);
+
+    onHeightChanged: {
+        var oldRecommendedRect = recommendedRect;
```
jherico: the code for `onHeightChanged` and `onWidthChanged` appears identical, or nearly identical.  it should be factored into a function, just as previously they both referenced a function inside `d`

interface/resources/qml/desktop/Desktop.qml
```diff
@@ -202,12 +223,23 @@ FocusScope {
 //            }
         }
 
+        function getAvatarInputsWindow() {
```
jherico: This functionality really belongs in the `hifi/Desktop.qml`.  The intent is to keep the `Desktop.qml` free of anything that's specific to the Interface application.

interface/resources/qml/desktop/Desktop.qml
```diff
@@ -238,32 +270,62 @@ FocusScope {
...
             // newPosition = Utils.clampVector(newPosition, minPosition, maxPosition);
             // newPosition = Utils.randomPosition(minPosition, maxPosition);
-            newPosition = Qt.vector2d(desktop.width / 2 - targetWindow.width / 2,
-                                      desktop.height / 2 - targetWindow.height / 2);
+
+            // center in new space
+            //newPosition = Qt.vector2d(desktop.width / 2 - targetWindow.width / 2,
+            //                          desktop.height / 2 - targetWindow.height / 2);
+
+            var oldRecommendedRect = recommendedRect;
```
jherico: you do this often enough that you may just wish to maintain properties for old / current rect / dimensions.

interface/resources/qml/desktop/Desktop.qml
```diff
@@ -238,32 +270,62 @@ FocusScope {
...
+            var oldRecommendedDimmensions = { x: oldRecommendedRect.z - oldRecommendedRect.x, y: oldRecommendedRect.w - oldRecommendedRect.y };
```
jherico: it might be worth storing this as a Qt.vector2d

interface/src/Application.cpp
```diff
@@ -4882,19 +4880,41 @@ QRect Application::getRenderingGeometry() const {
 }
 
 glm::uvec2 Application::getUiSize() const {
-    return getActiveDisplayPlugin()->getRecommendedUiSize();
+    glm::uvec2 result;
```
jherico: this should be initialized to some minimum value, if only 10x10.  This return value is going to be used to create/resize a framebuffer, an operation which may fail if the texture size is 0x0.

interface/src/Application.cpp
```diff
@@ -4882,19 +4880,41 @@ QRect Application::getRenderingGeometry() const {
...
+    if (_displayPlugin) {
+        result = getActiveDisplayPlugin()->getRecommendedUiSize();
+    }
+    return result;
+}
+
+glm::uvec4 Application::getRecommendedOverlayRect() const {
+    glm::uvec4 result;
```
jherico: ibid.

interface/src/Application.cpp
```diff
@@ -4882,19 +4880,41 @@ QRect Application::getRenderingGeometry() const {
...
+    if (_displayPlugin) {
+        result = getActiveDisplayPlugin()->getRecommendedOverlayRect();
+    }
+    return result;
 }
 
 QSize Application::getDeviceSize() const {
-    return fromGlm(getActiveDisplayPlugin()->getRecommendedRenderSize());
+    QSize result;
```
jherico: as before

interface/src/Application.cpp
```diff
@@ -4882,19 +4880,41 @@ QRect Application::getRenderingGeometry() const {
...
+    }
+    return false;
 }
 
 bool Application::hasFocus() const {
-    return getActiveDisplayPlugin()->hasFocus();
+    if (_displayPlugin) {
+        return getActiveDisplayPlugin()->hasFocus();
+    }
+    return false;
```
jherico: there's no reason not to fall back on checking the focus of the main window here.

libraries/shared/src/RegisteredMetaTypes.cpp
```diff
@@ -194,6 +206,46 @@ glm::vec3 vec3FromVariant(const QVariant &object) {
     return vec3FromVariant(object, valid);
 }
 
+glm::vec4 vec4FromVariant(const QVariant &object, bool& valid) {
+    glm::vec4 v;
+    valid = false;
+    if (!object.isValid() || object.isNull()) {
+        return v;
+    }
+    else if (object.canConvert<float>()) {
```
jherico: I would build a test case for this to ensure that you're not taking this codepath for undesirable types.

interface/resources/qml/desktop/Desktop.qml
ZappoMan: I have a solution. I will allow any child to have a "shouldReposition" property, if true, they will get re-positioned even if they are not a top level window. this way there's no knowledge of it being an avatar inputs, and we can support re-positioning other elements easily in the future.

interface/resources/qml/desktop/Desktop.qml
ZappoMan: will fix

libraries/shared/src/RegisteredMetaTypes.cpp
ZappoMan: I don't understand this comment, this code is identical to the `vec3FromVariant` code...

hifi-gustavo:
Build console output for commit: 6b5fd1e807d333287d3c904e0da2b00c62c3ce29 available at: https://gist.github.com/2ba42ec0c9b1985a9939b76ce128be08<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7690.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7690.dmg

hifi-gustavo:
Build console output for commit: 8f3d48190ad0f5313c53acd366e84b24df767805 available at: https://gist.github.com/2ba42ec0c9b1985a9939b76ce128be08<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7690.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7690.dmg

libraries/shared/src/RegisteredMetaTypes.h
```diff
@@ -71,9 +74,10 @@ glm::quat quatFromVariant(const QVariant &object);
 // Rect
 QScriptValue qRectToScriptValue(QScriptEngine* engine, const QRect& rect);
 void qRectFromScriptValue(const QScriptValue& object, QRect& rect);
-
-QVariant qRectToVariant(const QRect& rect);
 QRect qRectFromVariant(const QVariant& object, bool& isValid);
```
jherico: I'm not sure why I added these functions in the first place, since QVariant has a QRect constructor.

examples/defaultScripts.js
```diff
@@ -14,11 +14,9 @@ Script.load("edit.js");
 Script.load("examples.js");
 Script.load("selectAudioDevice.js");
 Script.load("notifications.js");
-Script.load("users.js");
```
jherico: is this intentional?

hifi-gustavo:
Build console output for commit: b62f1a135d1c1b67d77e048c54bd9173467129e4 available at: https://gist.github.com/2ba42ec0c9b1985a9939b76ce128be08<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7690.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7690.dmg

examples/defaultScripts.js
ZappoMan: Yes. See the asana tasks linked. Ryan wants this removed.

examples/libraries/toolBars.js
```diff
@@ -355,20 +355,36 @@ ToolBar = function(x, y, direction, optionalPersistenceKey, optionalInitialPosit
             });
         }
     };
-    that.windowDimensions = Controller.getViewportDimensions();
+
+    function clamp(value, min, max) {
+        return Math.min(Math.max(value, min), max);
+    }
+
+    var recommendedRect = Controller.getRecommendedOverlayRect();
```
howard-stearns: Doesn't the coding standard says no abbreviations?
(I couldn't care less for the variable, but the Controller property is very public.)

OmegaHeron:
Testing on Windows 7 - If I enable edit I get no edit properties window.  Tried reloading script and confirmed it's loading from PR installed source, tried resizing Interface window as well - no sign of edit properties window found.

And that would be because it set this...
[ToolWindow.Position]
x=1280
y=981

On a 1280x1024 monitor.

When I initially ran this Interface was set to begin as a maximized window on said 1280x1024 screen.

howard-stearns:
There's seems to be discrepancy in window resizing between the way QML tools vs scriptable toolbars.  1. Place one of each above the top edge in desktop, so that the tool/toolbar is off screen. 
2. Toggle to HMD and back to desktop. 
3. The toolbar overlay is now fully visible (as expected), but the QML tool has its title portion off screen. (It is as though the QML tool origin being clamped is the portion without any toolbar or border.)

(In order to avoid having tools become inaccessible, the spec calls for a recommended area margin on the right and bottom, which this PR doesn't have. I wouldn't call that a blocker as we can fix that later. However, it would be nice to resolve the discrepancy between QML tools and scriptable toolbars before merging this PR.)

OmegaHeron:
Manually reset ToolWindow.Position to something sane - laucnched this PR and it rewrote it as x,y 1280,981.  Launched again and tried various window sizes with edit mode on/off.

Placed interface on my 2nd monitor (1920x1080) and ini file changes to

[ToolWindow.Position]
x=1920
y=1037

howard-stearns:
Yes @OmegaHeron, if I understand correctly, you have the toolbar setting positioned horizontally at (or past) the right edge of the recommended area. This PR doesn't address that, but we have a separate spec that attempts to.

OmegaHeron:
Ok - fair enough, but, this PR removes my ability to use edit.js as its properties windows is pushed to a location where it can not possibly be accessed.  If that's intentional in that it will feed another PR that rectifies, fine.  If not then it's a fail.

OmegaHeron:
Users.js does not load with default scripts (as intended) - interesting change.
Toolbars follow window resizes
General smoke test finds no changes to operation other than issue with edit properties being located off screen.
Unable to perform HMD tests

If edit properties window behavior is intended and to be fixed in (rapidly pushed) followup PR then this is a pass for Desktop mode.

ZappoMan:
@OmegaHeron - the edit window behavior you are describing does not sound good... that sounds like a QA fail to me... but I haven't exactly studied what you're saying happened. Is it a two monitor thing??

OmegaHeron:
@ZappoMan - I have dual displays, but, what I'm seeing is that edit properties window is set to max value of whichever monitor it runs on at launch - it rewrites the saved posit each run to x max of monitor inferface is running on and near y max.  If I run current production build it detects that's an out of bounds and returns it to center of screen.  So - it's not necessarily a two monitor thing, I just tested on both to be sure it wasn't some reaction to having 2x displays.

howard-stearns:
@ZappoMan It can be repro'd by changing monitors between sessions, but more commonly, it will occur as follows:
1. In HMD, move an overlay to the right of the visible area. (E.g., it can be completely visible to the right and not clipped, but still outside the recommended area. )
2. Switch to Desktop. The left edge of the overly will be aligned with the right edge of the screen, effectively making the overlay invisible.

The spec (and my comment above) addresses this through the use of margins at right and bottom.

OmegaHeron:
@howard-stearns - I'd like to understand one thing.  Does this imply from now on edit props window will always be located at some point where there's just enough to grab it and pull back into usable space or will it stick where I place and expect it to return on toggles?  As it stands now even with margin allowing it to be grabbed it would not stick to a relocated posit across runs as Interface is rewriting its ini file posit on each run.

ZappoMan:
" as Interface is rewriting its ini file posit on each run." -- that sounds
like a bug...

howard-stearns:
@OmegaHeron, I understand there is some consideration of changing the layout for edit.js itself, but let's assume the current vertical toolbar used by edit.js. (After all anyone can write a script to do that.)

We only intended to make sure that things are "visible enough" after screen resize, including the resizing implied by a Desktop/HMD change. (The important part of the overlay to remain visible is the upper left of the overlay.) We do _not  _envision having different positions for different display toggles. The proposed behavior to implement these goals is exactly what is implemented in this PR -- _plus_ a margin along the right and bottom that would allow for the upper left portion of each overlay to be within the visible area even when the upper left corner of the overlay is along the right or bottom edge of the recommended area.

So, things would stay where you put them until a resize or toggle, which would act just like this PR plus the margin. The persisted value would be as resized.  I should already be the case that the value persisted is as a faction of the recommended size (rather than as pixels), so that a change in monitor size will get scaled.

ZappoMan:
@OmegaHeron - I found and fixed the bug you noticed with the Tool Window... so please retest when you have a chance, and I believe we will be good to go.

OmegaHeron:
@ZappoMan - will do so soon as it's ready and advise.

hifi-gustavo:
Build console output for commit: f5cddc996267717cf383b46b2c5558953d3427e1 available at: https://gist.github.com/2ba42ec0c9b1985a9939b76ce128be08<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7690.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7690.dmg

OmegaHeron:
Tested Windows 7
Edit Properties posit is preserved and unchanged :+1: 
Toolbar/icons for Go To, Examples and Edit move about with window resize
Users no longer loads with default scripts
General smoke test looks good

One thing I notice is with Examples.  It initially places with its title bar cut off due to being placed a bit high.  All its important content is within visible bounds and remains so as screen is resized.  Same with "Go To" its title area is placed outside visual space vertically.

PR #7692, opened 2016-04-16
Reset entities script engine on domain change

ZappoMan:
This fixes memory "holds" and security issues with entity scripts. Now, when you change domains, the script engine state for the entities scripts will be completely reset. This reduces the memory footprint for scripts as well as fixes potential security back doors.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -4177,14 +4177,12 @@ void Application::clearDomainOctreeDetails() {
 
     // if we're about to quit, we really don't need to do any of these things...
     if (_aboutToQuit) {
-        qCDebug(interfaceapp) << "No need to clear domain octree details we are shutting down...";
         return;
     }
 
-
     qCDebug(interfaceapp) << "Clearing domain octree details...";
-    // reset the environment so that we don't erroneously end up with multiple
 
+    // reset the environment so that we don't erroneously end up with multiple
     resetPhysicsReadyInformation();
 
     // reset our node to stats and node to jurisdiction maps... since these must be changing...
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1188,7 +1188,6 @@ void Application::cleanupBeforeQuit() {
     nodeList->getPacketReceiver().setShouldDropPackets(true);
 
     getEntities()->shutdown(); // tell the entities system we're shutting down, so it will stop running scripts
-    getEntities()->clear(); // this will allow entity scripts to properly shutdown
 
     DependencyManager::get<ScriptEngines>()->saveScripts();
     DependencyManager::get<ScriptEngines>()->shutdownScripting(); // stop all currently running global scripts
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1156,6 +1156,9 @@ void Application::aboutToQuit() {
 }
 
 void Application::cleanupBeforeQuit() {
+    qDebug() << __FUNCTION__ << "----- START -----";
+
+
     // add a logline indicating if QTWEBENGINE_REMOTE_DEBUGGING is set or not
     QString webengineRemoteDebugging = QProcessEnvironment::systemEnvironment().value("QTWEBENGINE_REMOTE_DEBUGGING", "false");
     qCDebug(interfaceapp) << "QTWEBENGINE_REMOTE_DEBUGGING =" << webengineRemoteDebugging;
@@ -1178,8 +1181,6 @@ void Application::cleanupBeforeQuit() {
     }
     _keyboardFocusHighlight = nullptr;
 
-    getEntities()->clear(); // this will allow entity scripts to properly shutdown
-
     auto nodeList = DependencyManager::get<NodeList>();
 
     // send the domain a disconnect packet, force stoppage of domain-server check-ins
@@ -1189,10 +1190,18 @@ void Application::cleanupBeforeQuit() {
     // tell the packet receiver we're shutting down, so it can drop packets
     nodeList->getPacketReceiver().setShouldDropPackets(true);
 
+    qDebug() << __FUNCTION__ << "about to call getEntities()->shutdown();";
     getEntities()->shutdown(); // tell the entities system we're shutting down, so it will stop running scripts
+
+    getEntities()->clear(); // this will allow entity scripts to properly shutdown
+    //DependencyManager::destroy<EntityTreeRenderer>();
+    qDebug() << __FUNCTION__ << "AFTER call to getEntities()->shutdown();";
+
+    qDebug() << __FUNCTION__ << "about to shutdown ScriptEngines...";
     DependencyManager::get<ScriptEngines>()->saveScripts();
     DependencyManager::get<ScriptEngines>()->shutdownScripting(); // stop all currently running global scripts
     DependencyManager::destroy<ScriptEngines>();
+    qDebug() << __FUNCTION__ << "AFTER shutdown ScriptEngines...";
 
     // first stop all timers directly or by invokeMethod
     // depending on what thread they run in
@@ -1202,10 +1211,14 @@ void Application::cleanupBeforeQuit() {
     pingTimer.stop();
     QMetaObject::invokeMethod(&_settingsTimer, "stop", Qt::BlockingQueuedConnection);
 
+    qDebug() << __FUNCTION__ << " line:" << __LINE__;
+
     // save state
     _settingsThread.quit();
     saveSettings();
     _window->saveGeometry();
+
+    qDebug() << __FUNCTION__ << " line:" << __LINE__;
 
     // stop the AudioClient
     QMetaObject::invokeMethod(DependencyManager::get<AudioClient>().data(),
@@ -1214,9 +1227,13 @@ void Application::cleanupBeforeQuit() {
     // destroy the AudioClient so it and its thread have a chance to go down safely
     DependencyManager::destroy<AudioClient>();
 
+    qDebug() << __FUNCTION__ << " line:" << __LINE__;
+
     // destroy the AudioInjectorManager so it and its thread have a chance to go down safely
     // this will also stop any ongoing network injectors
     DependencyManager::destroy<AudioInjectorManager>();
+
+    qDebug() << __FUNCTION__ << " line:" << __LINE__;
 
     // Destroy third party processes after scripts have finished using them.
 #ifdef HAVE_DDE
@@ -1226,7 +1243,11 @@ void Application::cleanupBeforeQuit() {
     DependencyManager::destroy<EyeTracker>();
 #endif
 
+    qDebug() << __FUNCTION__ << " line:" << __LINE__;
+
     DependencyManager::destroy<OffscreenUi>();
+
+    qDebug() << __FUNCTION__ << " ----- END -----";
 }
 
 Application::~Application() {
@@ -4174,6 +4195,14 @@ void Application::updateWindowTitle() const {
 }
 
 void Application::clearDomainOctreeDetails() {
+
+    // if we're about to quit, we really don't need to do any of these things...
+    if (_aboutToQuit) {
+        qCDebug(interfaceapp) << "No need to clear domain octree details we are shutting down...";
+        return;
+    }
+
+
     qCDebug(interfaceapp) << "Clearing domain octree details...";
     // reset the environment so that we don't erroneously end up with multiple
 
@@ -4189,7 +4218,9 @@ void Application::clearDomainOctreeDetails() {
     });
 
     // reset the model renderer
+    qCDebug(interfaceapp) << __FUNCTION__ << " --- BEFORE getEntities()->clear() --- _aboutToQuit:" << _aboutToQuit;
     getEntities()->clear();
+    qCDebug(interfaceapp) << __FUNCTION__ << " --- AFTER getEntities()->clear() ---";
 
     auto skyStage = DependencyManager::get<SceneScriptingInterface>()->getSkyStage();
     skyStage->setBackgroundMode(model::SunSkyStage::SKY_DOME);
@@ -4198,10 +4229,13 @@ void Application::clearDomainOctreeDetails() {
 }
 
 void Application::domainChanged(const QString& domainHostname) {
+    qCDebug(interfaceapp) << __FUNCTION__ << " --- BEGIN --- domainHostname:" << domainHostname;
+
     updateWindowTitle();
-    clearDomainOctreeDetails();
+    //clearDomainOctreeDetails();
     // disable physics until we have enough information about our new location to not cause craziness.
     resetPhysicsReadyInformation();
+    qCDebug(interfaceapp) << __FUNCTION__ << " --- END ---";
 }
 
 
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1156,9 +1156,6 @@ void Application::aboutToQuit() {
 }
 
 void Application::cleanupBeforeQuit() {
-    qDebug() << __FUNCTION__ << "----- START -----";
-
-
     // add a logline indicating if QTWEBENGINE_REMOTE_DEBUGGING is set or not
     QString webengineRemoteDebugging = QProcessEnvironment::systemEnvironment().value("QTWEBENGINE_REMOTE_DEBUGGING", "false");
     qCDebug(interfaceapp) << "QTWEBENGINE_REMOTE_DEBUGGING =" << webengineRemoteDebugging;
@@ -1190,18 +1187,12 @@ void Application::cleanupBeforeQuit() {
     // tell the packet receiver we're shutting down, so it can drop packets
     nodeList->getPacketReceiver().setShouldDropPackets(true);
 
-    qDebug() << __FUNCTION__ << "about to call getEntities()->shutdown();";
     getEntities()->shutdown(); // tell the entities system we're shutting down, so it will stop running scripts
-
     getEntities()->clear(); // this will allow entity scripts to properly shutdown
-    //DependencyManager::destroy<EntityTreeRenderer>();
-    qDebug() << __FUNCTION__ << "AFTER call to getEntities()->shutdown();";
 
-    qDebug() << __FUNCTION__ << "about to shutdown ScriptEngines...";
     DependencyManager::get<ScriptEngines>()->saveScripts();
     DependencyManager::get<ScriptEngines>()->shutdownScripting(); // stop all currently running global scripts
     DependencyManager::destroy<ScriptEngines>();
-    qDebug() << __FUNCTION__ << "AFTER shutdown ScriptEngines...";
 
     // first stop all timers directly or by invokeMethod
     // depending on what thread they run in
@@ -1211,14 +1202,10 @@ void Application::cleanupBeforeQuit() {
     pingTimer.stop();
     QMetaObject::invokeMethod(&_settingsTimer, "stop", Qt::BlockingQueuedConnection);
 
-    qDebug() << __FUNCTION__ << " line:" << __LINE__;
-
     // save state
     _settingsThread.quit();
     saveSettings();
     _window->saveGeometry();
-
-    qDebug() << __FUNCTION__ << " line:" << __LINE__;
 
     // stop the AudioClient
     QMetaObject::invokeMethod(DependencyManager::get<AudioClient>().data(),
@@ -1227,13 +1214,9 @@ void Application::cleanupBeforeQuit() {
     // destroy the AudioClient so it and its thread have a chance to go down safely
     DependencyManager::destroy<AudioClient>();
 
-    qDebug() << __FUNCTION__ << " line:" << __LINE__;
-
     // destroy the AudioInjectorManager so it and its thread have a chance to go down safely
     // this will also stop any ongoing network injectors
     DependencyManager::destroy<AudioInjectorManager>();
-
-    qDebug() << __FUNCTION__ << " line:" << __LINE__;
 
     // Destroy third party processes after scripts have finished using them.
 #ifdef HAVE_DDE
@@ -1243,11 +1226,7 @@ void Application::cleanupBeforeQuit() {
     DependencyManager::destroy<EyeTracker>();
 #endif
 
-    qDebug() << __FUNCTION__ << " line:" << __LINE__;
-
     DependencyManager::destroy<OffscreenUi>();
-
-    qDebug() << __FUNCTION__ << " ----- END -----";
 }
 
 Application::~Application() {
@@ -4218,9 +4197,7 @@ void Application::clearDomainOctreeDetails() {
     });
 
     // reset the model renderer
-    qCDebug(interfaceapp) << __FUNCTION__ << " --- BEFORE getEntities()->clear() --- _aboutToQuit:" << _aboutToQuit;
     getEntities()->clear();
-    qCDebug(interfaceapp) << __FUNCTION__ << " --- AFTER getEntities()->clear() ---";
 
     auto skyStage = DependencyManager::get<SceneScriptingInterface>()->getSkyStage();
     skyStage->setBackgroundMode(model::SunSkyStage::SKY_DOME);
@@ -4229,13 +4206,9 @@ void Application::clearDomainOctreeDetails() {
 }
 
 void Application::domainChanged(const QString& domainHostname) {
-    qCDebug(interfaceapp) << __FUNCTION__ << " --- BEGIN --- domainHostname:" << domainHostname;
-
     updateWindowTitle();
-    //clearDomainOctreeDetails();
     // disable physics until we have enough information about our new location to not cause craziness.
     resetPhysicsReadyInformation();
-    qCDebug(interfaceapp) << __FUNCTION__ << " --- END ---";
 }
 
 
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -4182,7 +4182,6 @@ void Application::clearDomainOctreeDetails() {
 
     qCDebug(interfaceapp) << "Clearing domain octree details...";
 
-    // reset the environment so that we don't erroneously end up with multiple
     resetPhysicsReadyInformation();
 
     // reset our node to stats and node to jurisdiction maps... since these must be changing...
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -140,6 +140,8 @@ void EntityTreeRenderer::init() {
 void EntityTreeRenderer::shutdown() {
     _entitiesScriptEngine->disconnectNonEssentialSignals(); // disconnect all slots/signals from the script engine, except essential
     _shuttingDown = true;
+
+    clear(); // always clear() on shutdown
 }
 
 void EntityTreeRenderer::setTree(OctreePointer newTree) {
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -78,6 +78,7 @@ EntityTreeRenderer::~EntityTreeRenderer() {
 int EntityTreeRenderer::_entititesScriptEngineCount = 0;
 
 void EntityTreeRenderer::setupEntitiesScriptEngine() {
+    QSharedPointer<ScriptEngine> oldEngine = _entitiesScriptEngine; // save the old engine through this function, so the EntityScriptingInterface doesn't have problems with it.
     _entitiesScriptEngine = QSharedPointer<ScriptEngine>(new ScriptEngine(NO_SCRIPT, QString("Entities %1").arg(++_entititesScriptEngineCount)), ScriptEngine::doDeleteLater);
     _scriptingServices->registerScriptEngineWithApplicationServices(_entitiesScriptEngine.data());
     _entitiesScriptEngine->runInThread();
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -92,7 +92,6 @@ void EntityTreeRenderer::clear() {
         _entitiesScriptEngine->stop();
     }
 
-    // this would be a good place to actuall delete and recreate the _entitiesScriptEngine
     if (_wantScripts && !_shuttingDown) {
         // NOTE: you can't actually need to delete it here because when we call setupEntitiesScriptEngine it will
         //       assign a new instance to our shared pointer, which will deref the old instance and ultimately call
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -88,12 +88,11 @@ void EntityTreeRenderer::clear() {
     leaveAllEntities();
     if (_entitiesScriptEngine) {
         _entitiesScriptEngine->unloadAllEntityScripts();
+        _entitiesScriptEngine->stop();
     }
 
     // this would be a good place to actuall delete and recreate the _entitiesScriptEngine
     if (_wantScripts && !_shuttingDown) {
-        _entitiesScriptEngine->stop();
-
         // NOTE: you can't actually need to delete it here because when we call setupEntitiesScriptEngine it will
         //       assign a new instance to our shared pointer, which will deref the old instance and ultimately call
         //       the custom deleter which calls deleteLater
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -71,13 +71,46 @@ EntityTreeRenderer::EntityTreeRenderer(bool wantScripts, AbstractViewStateInterf
 }
 
 EntityTreeRenderer::~EntityTreeRenderer() {
+    qDebug() << __FUNCTION__ << "---- BEGIN ----";
     // NOTE: We don't need to delete _entitiesScriptEngine because
     //       it is registered with ScriptEngines, which will call deleteLater for us.
+
+    /*
+    if (_entitiesScriptEngine) { 
+        qDebug() << __FUNCTION__ << "about to stop entity script engine.";
+        _entitiesScriptEngine->stop();
+        qDebug() << __FUNCTION__ << "about to delete entity script engine.";
+        delete _entitiesScriptEngine;
+        _entitiesScriptEngine = nullptr;
+    }
+    */
+    qDebug() << __FUNCTION__ << "---- END ----";
 }
 
+static int entititesScriptEngineCount = 0;
+
 void EntityTreeRenderer::clear() {
+    qDebug() << __FUNCTION__ << "---- BEGIN ----";
     leaveAllEntities();
-    _entitiesScriptEngine->unloadAllEntityScripts();
+    if (_entitiesScriptEngine) {
+        _entitiesScriptEngine->unloadAllEntityScripts();
+    }
+
+    // this would be a good place to actuall delete and recreate the _entitiesScriptEngine
+    qDebug() << __FUNCTION__ << "_shuttingDown:" << _shuttingDown;
+    if (_wantScripts && !_shuttingDown) {
+        qDebug() << __FUNCTION__ << " about to stop/delete current _entitiesScriptEngine";
+        _entitiesScriptEngine->stop();
+        _entitiesScriptEngine->deleteLater();
+        qDebug() << __FUNCTION__ << " AFTER stop/delete current _entitiesScriptEngine";
+
+        qDebug() << __FUNCTION__ << " about to create new _entitiesScriptEngine";
+        _entitiesScriptEngine = new ScriptEngine(NO_SCRIPT, QString("Entities %1").arg(++entititesScriptEngineCount));
+        _scriptingServices->registerScriptEngineWithApplicationServices(_entitiesScriptEngine);
+        _entitiesScriptEngine->runInThread();
+        DependencyManager::get<EntityScriptingInterface>()->setEntitiesScriptEngine(_entitiesScriptEngine);
+        qDebug() << __FUNCTION__ << " AFTER create new _entitiesScriptEngine";
+    }
 
     auto scene = _viewState->getMain3DScene();
     render::PendingChanges pendingChanges;
@@ -88,6 +121,7 @@ void EntityTreeRenderer::clear() {
     _entitiesInScene.clear();
 
     OctreeRenderer::clear();
+    qDebug() << __FUNCTION__ << "---- END ----";
 }
 
 void EntityTreeRenderer::reloadEntityScripts() {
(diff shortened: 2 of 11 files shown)


interface/src/Application.cpp
```diff
@@ -1178,8 +1181,6 @@ void Application::cleanupBeforeQuit() {
     }
     _keyboardFocusHighlight = nullptr;
 
-    getEntities()->clear(); // this will allow entity scripts to properly shutdown
```
ZappoMan: this should happen after the EntityTreeRenderer has been told we're shutting down, so that it can short circuit some work... honestly, we probably could delete this, since the entity scripts can't realistically count on the domain services being available to them... so there's not much cleanup they can actually do.

interface/src/Application.cpp
```diff
@@ -4174,6 +4195,14 @@ void Application::updateWindowTitle() const {
 }
 
 void Application::clearDomainOctreeDetails() {
+
+    // if we're about to quit, we really don't need to do any of these things...
+    if (_aboutToQuit) {
```
ZappoMan: there's no point in clearing the domain details if we're shutting down.

interface/src/Application.cpp
```diff
@@ -4198,10 +4229,13 @@ void Application::clearDomainOctreeDetails() {
 }
 
 void Application::domainChanged(const QString& domainHostname) {
+    qCDebug(interfaceapp) << __FUNCTION__ << " --- BEGIN --- domainHostname:" << domainHostname;
+
     updateWindowTitle();
-    clearDomainOctreeDetails();
+    //clearDomainOctreeDetails();
```
ZappoMan: this call was redundant, because we also call this when we get a `nodeKilled` signal.

interface/src/Application.cpp
```diff
@@ -4198,10 +4229,13 @@ void Application::clearDomainOctreeDetails() {
...
     // disable physics until we have enough information about our new location to not cause craziness.
     resetPhysicsReadyInformation();
```
ZappoMan: technically this looks redundant too... because we call this inside of `clearDomainOctreeDetails()` which will get called for the `nodeKilled` signal.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -71,13 +71,46 @@ EntityTreeRenderer::EntityTreeRenderer(bool wantScripts, AbstractViewStateInterf
 }
 
 EntityTreeRenderer::~EntityTreeRenderer() {
+    qDebug() << __FUNCTION__ << "---- BEGIN ----";
     // NOTE: We don't need to delete _entitiesScriptEngine because
     //       it is registered with ScriptEngines, which will call deleteLater for us.
+
+    /*
```
ZappoMan: I was experimenting with the idea of shutting down the script engine more proactively...

We apparently changed `EntityTreeRenderer` to a `Dependency` back in December, and as a result, the `EntityTreeRenderer` shutdown doesn't happen until an unspecified late point in the life of the application, so this change really didn't add much.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -71,13 +71,46 @@ EntityTreeRenderer::EntityTreeRenderer(bool wantScripts, AbstractViewStateInterf
...
+    qDebug() << __FUNCTION__ << "---- BEGIN ----";
     leaveAllEntities();
-    _entitiesScriptEngine->unloadAllEntityScripts();
+    if (_entitiesScriptEngine) {
+        _entitiesScriptEngine->unloadAllEntityScripts();
+    }
+
+    // this would be a good place to actuall delete and recreate the _entitiesScriptEngine
+    qDebug() << __FUNCTION__ << "_shuttingDown:" << _shuttingDown;
+    if (_wantScripts && !_shuttingDown) {
```
ZappoMan: this is really the money shot when you tell the EntityTreeRenderer to clear, we also clear and recreate the entities scripting engine... we don't need to do this if we're shutting down.

hifi-gustavo:
Build console output for commit: 87078b1ea29931e9de641a92876382f1bc710aad available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

ZappoMan:
In looking closer, I believe that this particular call is unsafe and will/could cause crashes when you rapidly switch domains....


```
void ScriptEngine::loadEntityScript(const EntityItemID& entityID, const QString& entityScript, bool forceRedownload) {
...
    DependencyManager::get<ScriptCache>()->getScriptContents(entityScript, [=](const QString& scriptOrURL, const QString& contents, bool isURL, bool success) {

```

Basically, the ScriptCache could call back into that lambda at some distant point in the future when the script engine has already been deleted from moving between domains... this wasn't a problem when we only had a single Entities script engine for the life of the application.

This mechanism needs to be repaired before this PR can be merged.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1073,13 +1077,21 @@ void ScriptEngine::loadEntityScript(const EntityItemID& entityID, const QString&
     qDebug() << "ScriptEngine::loadEntityScript() calling scriptCache->getScriptContents() on thread ["
         << QThread::currentThread() << "] expected thread [" << thread() << "]";
 #endif
-    DependencyManager::get<ScriptCache>()->getScriptContents(entityScript, [=](const QString& scriptOrURL, const QString& contents, bool isURL, bool success) {
+
+    QPointer<ScriptEngine> theEngine(this);
```
ZappoMan: This handles the case of the ScriptEngine being deleted while the `getScriptContents()` is still pending...

With this new code, and rapidly switching domains, I was able to get this to occur.... 

```
[04/16 19:35:55] [DEBUG] Done downloading script at: "https://rawgit.com/ericrius1/hifi/growingPlant/unpublishedScripts/DomainContent/Home/plant/growingPlantEntityScript.js?v10.54"
[04/16 19:35:55] [DEBUG] ScriptCache::getScriptContents() returned after our ScriptEngine was deleted...
```

hifi-gustavo:
Build console output for commit: 71d57295eb968cd6f66364253b6d75893d77cc34 available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: hmmm... this still isn't sufficient, because the ScriptEngine can still get deleted after we pass this point but before `theEngine->entityScriptContentAvailable()` completes

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -71,13 +71,46 @@ EntityTreeRenderer::EntityTreeRenderer(bool wantScripts, AbstractViewStateInterf
...
+        qDebug() << __FUNCTION__ << " about to stop/delete current _entitiesScriptEngine";
+        _entitiesScriptEngine->stop();
+        _entitiesScriptEngine->deleteLater();
+        qDebug() << __FUNCTION__ << " AFTER stop/delete current _entitiesScriptEngine";
+
+        qDebug() << __FUNCTION__ << " about to create new _entitiesScriptEngine";
+        _entitiesScriptEngine = new ScriptEngine(NO_SCRIPT, QString("Entities %1").arg(++entititesScriptEngineCount));
+        _scriptingServices->registerScriptEngineWithApplicationServices(_entitiesScriptEngine);
+        _entitiesScriptEngine->runInThread();
+        DependencyManager::get<EntityScriptingInterface>()->setEntitiesScriptEngine(_entitiesScriptEngine);
```
jherico: I believe this makes the code in `EntityScriptingInterface::callEntityMethod` unsafe.

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: in other words.... "these are two separate instructions with two separate calls to QPointer thus the thread can be interrupted somewhere in between the check and the call (so "if" is not atomic)."

libraries/entities-renderer/src/EntityTreeRenderer.cpp
ZappoMan: Can you elaborate?

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: So what probably needs to happen here, is that `ScriptEngine::loadEntityScript()` maybe should be made static, and should take a weak pointer to the `ScriptEngine` that we want to execute the load in, once the script content is available.

That way, once the `ScriptCache` has the contents, we can promote the weak pointer to a strong pointer and know that the `ScriptEngine` can't be deleted on us while we evaluate the script.

This would mean that the `EntityTreeRenderer` should keep a`QSharedPointer<>` to it's `_entitiesScriptEngine` instead of a raw pointer.

hifi-gustavo:
Build console output for commit: 9db0fe0d11594ac658e6a4d8afa08c19a8b6885c available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

hifi-gustavo:
Build console output for commit: f6c75d0530c3b41dfc9af287ac4318ec421dadcf available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

ZappoMan:
Testing Plan:
- Smoke test - particularly focus on entity scripts
- Change domains, and verify that you see the following messages in your logs...
```
[04/18 10:35:47] [DEBUG] Logging activity "changed_domain"
...
[04/18 10:35:47] [DEBUG] Clearing domain octree details...
[04/18 10:35:47] [DEBUG] Script Engine shutting down (destructor) for script: "Entities 1"
```
- Rapidly change domains over and over again, I find rotating between: Forum, Demo, Playa, and Hell are a good choice. Rotate through them all in random order and repeated times...
- verify that you see this in your log for every time you change domains... note: X will be a different increasing integer: 1, 2, 3, 4... also, these won't always appear immediately after changing domains, but they should appear eventually.
```
[04/18 10:35:47] [DEBUG] Script Engine shutting down (destructor) for script: "Entities X"

```

- Especially after the rapidly changing domains, you might see one or more of these messages in your log...
```
[04/18 10:35:51] [DEBUG] ScriptCache::getScriptContents() returned after our ScriptEngine was deleted...
```
If you see that, that's good... it means there would have been a crash, but the new code did the right thing and didn't crash.

Expected results: no crashes. :)

ZappoMan:
Not quite ready... I'm looking into an issue Austin noticed.

hifi-gustavo:
Build console output for commit: 47900df3e776bbb7b0dd05da3b4d8a0b2a890cd9 available at: https://gist.github.com/651ffa741516f321535bd693308ae37c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

hifi-gustavo:
Build console output for commit: 1c111e20bd6be7bdd201d5a4052111e076a2f920 available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

hifi-gustavo:
Build console output for commit: 1c111e20bd6be7bdd201d5a4052111e076a2f920 available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -75,9 +75,30 @@ EntityTreeRenderer::~EntityTreeRenderer() {
     //       it is registered with ScriptEngines, which will call deleteLater for us.
 }
 
+int EntityTreeRenderer::_entititesScriptEngineCount = 0;
+
+void EntityTreeRenderer::setupEntitiesScriptEngine() {
+    QSharedPointer<ScriptEngine> oldEngine = _entitiesScriptEngine; // save the old engine through this function, so the EntityScriptingInterface doesn't have problems with it.
```
ZappoMan: we want to keep one ref count on the old engine, until after we call `setEntitiesScriptEngine()` this is because possibly some other script (running on it's own thread) might call `Entities.callEntityMethod()` in which case it would attempt to call that method on the old naked pointer for the old engine.... it's safe to drop the ref count after we call `setEntitiesScriptEngine()` because that method and `callEntityMethod()` are guarded with a mutex.

interface/src/Application.cpp
```diff
@@ -1190,6 +1188,8 @@ void Application::cleanupBeforeQuit() {
     nodeList->getPacketReceiver().setShouldDropPackets(true);
 
     getEntities()->shutdown(); // tell the entities system we're shutting down, so it will stop running scripts
+    getEntities()->clear(); // this will allow entity scripts to properly shutdown
```
zzmp: As long as you are ordering it this way, can you put `clear()` inside of the `shutdown()` method? It is part of shutdown, I think it should be encapsulated in that call.

interface/src/Application.cpp
zzmp: If we want to quit as quickly as possible, this makes sense (RELEASE),
but if we want to quit cleanly with proper memory cleanup (DEBUG), I think we should keep this.

Can you put this behind `#ifndef DEBUG`?

interface/src/Application.cpp
```diff
@@ -4199,7 +4205,6 @@ void Application::clearDomainOctreeDetails() {
 
 void Application::domainChanged(const QString& domainHostname) {
     updateWindowTitle();
-    clearDomainOctreeDetails();
```
zzmp: Why did you remove this?

interface/src/Application.cpp
```diff
@@ -4174,9 +4174,15 @@ void Application::updateWindowTitle() const {
...
+
+    // if we're about to quit, we really don't need to do any of these things...
+    if (_aboutToQuit) {
+        return;
+    }
+
     qCDebug(interfaceapp) << "Clearing domain octree details...";
-    // reset the environment so that we don't erroneously end up with multiple
 
+    // reset the environment so that we don't erroneously end up with multiple
```
zzmp: This comment does not jive with the comment on L4208. Taken together, they are hard to understand. Can you reconcile them?

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -75,9 +75,30 @@ EntityTreeRenderer::~EntityTreeRenderer() {
...
+    _entitiesScriptEngine = QSharedPointer<ScriptEngine>(new ScriptEngine(NO_SCRIPT, QString("Entities %1").arg(++_entititesScriptEngineCount)), ScriptEngine::doDeleteLater);
```
zzmp: Why doesn't GitHub understand printf format?

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1213,6 +1200,16 @@ void ScriptEngine::unloadAllEntityScripts() {
         callEntityScriptMethod(entityID, "unload");
     }
     _entityScripts.clear();
+
+#ifdef DEBUG_ENGINE_STATE
```
zzmp: We have a lot of very specific debug flags. Are they documented somewhere? It seems messy to make a new flag for each new feature's debug, without some organization.

libraries/entities-renderer/src/EntityTreeRenderer.h
```diff
@@ -196,6 +198,8 @@ public slots:
     QHash<EntityItemID, EntityItemPointer> _entitiesInScene;
     // For Scene.shouldRenderEntities
     QList<EntityItemID> _entityIDsLastInScene;
+
+    static int _entititesScriptEngineCount;
```
zzmp: s/entitities/entities

zzmp: Why is this necessary?

interface/src/Application.cpp
ZappoMan: because it's redundant... we do this on `nodeKilled()` which always happens when you change domains... and so by having it in two places we create and shutdown the ScriptEngine twice per domain change.

interface/src/Application.cpp
ZappoMan: ok - I didn't add that comment, I don't know why github thinks I did...

hifi-gustavo:
Build console output for commit: c26115cf2ef72698921b5eeab8a49d88ced2f2f6 available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

libraries/entities-renderer/src/EntityTreeRenderer.h
```diff
@@ -126,6 +126,8 @@ public slots:
     }
 
 private:
+    void setupEntitiesScriptEngine();
```
zzmp: Will this always be called from the same thread?

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: I think I'm the only one that uses this technique -- I just do it instead of `#if 0` or commenting out the code, just to make it clear that if you want to debug that feature you can enable this code.

libraries/script-engine/src/ScriptEngine.h
```diff
@@ -70,6 +70,9 @@ class ScriptEngine : public QScriptEngine, public ScriptUser, public EntitiesScr
 
     ~ScriptEngine();
 
+    // Useful for QSharePointer<ScriptEngine>
+    static void doDeleteLater(ScriptEngine* engine) { engine->deleteLater(); }
```
zzmp: This is redundant. QSharedPointer can just use `QObject::deleteLater` as a deleter, it does not need an argument like `std::shared_ptr`.

libraries/script-engine/src/ScriptEngine.h
```diff
@@ -70,6 +70,9 @@ class ScriptEngine : public QScriptEngine, public ScriptUser, public EntitiesScr
 
     ~ScriptEngine();
 
+    // Useful for QSharePointer<ScriptEngine>
```
zzmp: typo

libraries/entities-renderer/src/EntityTreeRenderer.cpp
zzmp: Just use `&QObject::deleteLater`, no need for another method.

interface/src/Application.cpp
ZappoMan: ok - this is a totally old comment for code that no longer exists... I'm removing it

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -414,7 +414,13 @@ void EntityScriptingInterface::deleteEntity(QUuid id) {
     }
 }
 
+void EntityScriptingInterface::setEntitiesScriptEngine(EntitiesScriptEngineProvider* engine) {
+    std::unique_lock<std::mutex> lock(_entitiesScriptEngineLock);
```
zzmp: Why `unique_lock`s instead of `lock_guard`s?

libraries/entities/src/EntityScriptingInterface.h
```diff
@@ -71,7 +71,7 @@ class EntityScriptingInterface : public OctreeScriptingInterface, public Depende
 
     void setEntityTree(EntityTreePointer modelTree);
     EntityTreePointer getEntityTree() { return _entityTree; }
-    void setEntitiesScriptEngine(EntitiesScriptEngineProvider* engine) { _entitiesScriptEngine = engine; }
+    void setEntitiesScriptEngine(EntitiesScriptEngineProvider* engine);
```
zzmp: :+1:

libraries/entities/src/EntityScriptingInterface.h
```diff
@@ -214,6 +214,8 @@ public slots:
         bool precisionPicking, const QVector<EntityItemID>& entityIdsToInclude, const QVector<EntityItemID>& entityIdsToDiscard);
 
     EntityTreePointer _entityTree;
+
+    std::mutex _entitiesScriptEngineLock;
```
zzmp: Should be called `_entitiesScriptEngineMutex` per standard.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
ZappoMan: ok

libraries/entities-renderer/src/EntityTreeRenderer.h
ZappoMan: no and that's ok - it doesn't do anything that requires needing to be on the same thread.

libraries/script-engine/src/ScriptEngine.h
ZappoMan: fixing

libraries/entities-renderer/src/EntityTreeRenderer.cpp
zzmp: Can you leave this as a comment instead of the current comment?

libraries/entities/src/EntityScriptingInterface.h
ZappoMan: hmmm really? where does it say that in the standard? I see other cases where these are called `_fooLock`

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1047,39 +1048,25 @@ void ScriptEngine::forwardHandlerCall(const EntityItemID& entityID, const QStrin
...
-#endif
-
-        QMetaObject::invokeMethod(this, "loadEntityScript",
-                                  Q_ARG(const EntityItemID&, entityID),
-                                  Q_ARG(const QString&, entityScript),
-                                  Q_ARG(bool, forceRedownload));
-        return;
-    }
+void ScriptEngine::loadEntityScript(QWeakPointer<ScriptEngine> theEngine, const EntityItemID& entityID, const QString& entityScript, bool forceRedownload) {
+    // NOTE: If the script content is not currently in the caceh, The LAMBDA here, will be called on the Main Thread
```
zzmp: s/caceh/cache

zzmp: s/The/the

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1047,39 +1048,25 @@ void ScriptEngine::forwardHandlerCall(const EntityItemID& entityID, const QStrin
...
+    //       which means we're guarenteed that it's not the correct thread for the ScriptEngine. This means
```
zzmp: s/guarenteed/guaranteed

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1047,39 +1048,25 @@ void ScriptEngine::forwardHandlerCall(const EntityItemID& entityID, const QStrin
...
-            << QThread::currentThread() << "] expected thread [" << thread() << "]";
-#endif
-
-        this->entityScriptContentAvailable(entityID, scriptOrURL, contents, isURL, success);
+            strongEngine->entityScriptContentAvailable(entityID, scriptOrURL, contents, isURL, success);
+        } else {
+            // FIXME - I'm leaving this in for testing, so that QA can confirm that sometimes the script contents
+            //         returns after the ScriptEngine has been deleted, we can remove this after QA verifies the
+            //         repro case.
+            qDebug() << "ScriptCache::getScriptContents() returned after our ScriptEngine was deleted...";
```
zzmp: This is similar to other debugs that we log. I think it'd be fine to leave in, but would like it to include the entityScript.

interface/src/Application.cpp
ZappoMan: this won't effect memory cleanup... things would still get cleared on exit... this shouldn't prevent proper memory cleanup

libraries/script-engine/src/ScriptEngine.cpp
zzmp: I don't think this comment belongs here, as the actual invokeMethod is done in `entityScriptContentAvailable`. If anything, you can add a similar comment there, or a comment around the THREAD_DEBUGGING stating why that debugging is relevant. It is not relevant to the whole function, and the comment here confuses me because the action does not occur until later.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -75,9 +75,30 @@ EntityTreeRenderer::~EntityTreeRenderer() {
...
+
 void EntityTreeRenderer::clear() {
     leaveAllEntities();
-    _entitiesScriptEngine->unloadAllEntityScripts();
+    if (_entitiesScriptEngine) {
+        _entitiesScriptEngine->unloadAllEntityScripts();
+        _entitiesScriptEngine->stop();
+    }
+
+    // this would be a good place to actuall delete and recreate the _entitiesScriptEngine
```
zzmp: Fix this comment.

zzmp:
There is still an error with unloading a single script, in that it will not stop the timers. Not sure if you want to address that here or in another PR - I think either would be appropriate.

hifi-gustavo:
Build console output for commit: 9e13b1bbaef7fd3cbb06d62cfcc01a7396c9cf10 available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

libraries/entities-renderer/src/EntityTreeRenderer.h
zzmp: It takes a lock - cool :)

libraries/entities/src/EntityScriptingInterface.h
zzmp: https://readme.highfidelity.com/docs/coding-standard#section-1-1-10-generic-variables-should-have-the-same-name-as-their-type- but lock is fine here.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -75,9 +75,29 @@ EntityTreeRenderer::~EntityTreeRenderer() {
...
+    _entitiesScriptEngine->runInThread();
+    DependencyManager::get<EntityScriptingInterface>()->setEntitiesScriptEngine(_entitiesScriptEngine.data());
+}
+
 void EntityTreeRenderer::clear() {
     leaveAllEntities();
-    _entitiesScriptEngine->unloadAllEntityScripts();
+    if (_entitiesScriptEngine) {
+        _entitiesScriptEngine->unloadAllEntityScripts();
+        _entitiesScriptEngine->stop();
```
zzmp: Considering that we've just called unload, do we need to give entity scripts a moment to run their unload code before calling stop, which will abortEvaluation?

ZappoMan: let me think about that.

hifi-gustavo:
Build console output for commit: 55010cebb70a0898b5552dd1346aa558acd582c8 available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

hifi-gustavo:
Build console output for commit: bf8cdabd1bf965e80f35728ece99c58a006dacb0 available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

howard-stearns:
@ZappoMan can you summarize the expected behavior of running entity script code when changing domains?  For example, what already-running, timer-scheduled, or unload code will run? What behavior for Entity, AvatarList, and Script.setTimeout methods in progress?

hifi-gustavo:
Build console output for commit: 6ac173d19701fa5087db1e5e4988950c04d5e257 available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

hifi-gustavo:
Build console output for commit: 24003840df35ceb70fd9cc90062789c704a5dab3 available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

hifi-gustavo:
Build console output for commit: d6fa58c39c9feeb4b4a201f8dd9324472bb3e3d7 available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

hifi-gustavo:
Build console output for commit: ab7f3d573984ae41f1b772c4fa3f1b392a036e4f available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

libraries/entities-renderer/src/EntityTreeRenderer.cpp
ZappoMan: my testing notes show that the `unload` methods do get called, and in general things work as we would expect. So I think this is acceptable for now.

ZappoMan:
**Testing Plan:**
- Smoke test - particularly focus on entity scripts
- Make sure to do testing with hand grab script, and make sure things still work.
- Change domains, and verify that you see the following messages in your logs...
```
[04/18 10:35:47] [DEBUG] Logging activity "changed_domain"
...
[04/18 10:35:47] [DEBUG] Clearing domain octree details...
[04/18 10:35:47] [DEBUG] Script Engine shutting down (destructor) for script: "Entities 1"
```
- Rapidly change domains over and over again, I find rotating between: Forum, Demo, Playa, and Hell are a good choice. Rotate through them all in random order and repeated times...
- verify that you see this in your log for every time you change domains... note: X will be a different increasing integer: 1, 2, 3, 4... also, these won't always appear immediately after changing domains, but they should appear eventually.
```
[04/18 10:35:47] [DEBUG] Script Engine shutting down (destructor) for script: "Entities X"

```

- Especially after the rapidly changing domains, you might see one or more of these messages in your log...
```
[04/18 10:35:51] [DEBUG] ScriptCache::getScriptContents() returned after our ScriptEngine was deleted...
```
If you see that, that's good... it means there would have been a crash, but the new code did the right thing and didn't crash.

Expected results: no crashes. :)

- In some domain, create an entity and point it's entity script to this script: https://raw.githubusercontent.com/ZappoMan/hifi/resetScriptEngineOnDomainChange/examples/entityScripts/exampleUpdate.js

You should see the following messages flooding your logs...
```
[04/18 13:28:24] [DEBUG] Done downloading script at: "https://raw.githubusercontent.com/ZappoMan/hifi/resetScriptEngineOnDomainChange/examples/entityScripts/exampleUpdate.js"
[04/18 13:28:24] [DEBUG] script:print()<< preload - entityID:{cdd1b3f8-6844-4af3-8e8a-fdf994de7ceb}
[04/18 13:28:24] [DEBUG] script:print()<< update in entityID:{cdd1b3f8-6844-4af3-8e8a-fdf994de7ceb}
[04/18 13:28:24] [DEBUG] script:print()<< update in entityID:{cdd1b3f8-6844-4af3-8e8a-fdf994de7ceb}
```

When you switch domains, you should see:

```
[04/18 13:28:45] [DEBUG] Clearing domain octree details...
[04/18 13:28:45] [DEBUG] script:print()<< unload - entityID:{cdd1b3f8-6844-4af3-8e8a-fdf994de7ceb}
```

AND... the "update" messages should stop.

- In some domain, create an entity and point it's entity script to this script: https://raw.githubusercontent.com/ZappoMan/hifi/resetScriptEngineOnDomainChange/examples/entityScripts/exampleUpdateNoDisconnect.js

You should see the same behavior as the other script... Notice that with this PR... even though the script doesn't call the disconnect, the update signals stop running. In previous releases the update messages would continue to spew... this is an example of a "security" issue... since the entity script from an old domain continues to run... This PR fixes that.

- In some domain, create an entity and point it's entity script to this script: https://raw.githubusercontent.com/ZappoMan/hifi/resetScriptEngineOnDomainChange/examples/entityScripts/exampleTimeoutNoCleanup.js

You should see similar behavior, as the same behavior as the other script... This script will print out the following message every 3 seconds... 

```
[04/18 14:12:17] [DEBUG] script:print()<< timer interval in entityID:{26e7b595-44d0-44af-9be6-f68b4d4e1d4d}
```
Notice that with this PR... even though the script doesn't delete the timer the timer stops running when you change domains. In previous releases the timer messages would continue to spew... this is an example of a "security" issue... since the entity script from an old domain continues to run... This PR fixes that.



- In some domain, create an entity and point it's entity script to this script:  https://raw.githubusercontent.com/ZappoMan/hifi/resetScriptEngineOnDomainChange/examples/entityScripts/exampleSelfCallingTimeoutNoCleanup.js

You should see similar behavior, as the same behavior as the other script... This script will print out the following message every 3 seconds... 

```
[04/18 14:59:15] [DEBUG] script:print()<< timeOutFunction in entityID:{26e7b595-44d0-44af-9be6-f68b4d4e1d4d}
```
Notice that with this PR... even though the script doesn't delete the timer the timer stops running when you change domains. In previous releases the timer messages would continue to spew... this is an example of a "security" issue... since the entity script from an old domain continues to run... This PR fixes that.

ZappoMan:
@howard-stearns - I have moved the testing notes to the last comment so they are easier to find. I think it thoroughly discusses your question. I've provided several example scripts that fail with production builds (leave script leaking) and work as we desire in this build (upon switching domains, the script remnants are correctly cleaned up.)

hifi-gustavo:
Build console output for commit: c3a86f46d07f1bee738f19b8f8cbc5b1e2d362de available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

howard-stearns:
Nice. The additions are good. 

This comment in one of the examples is particularly useful: "unload() will be called when our entity is no longer available. It may be because we were deleted, or because we've left the domain or quit the application."  

Do we want to also explicitly define the behavior of calling any of the following within unload?  Entities.addEntity, Entity.editEntity, AvatarList.getAvatarIdentifiers, Script.setTimeout, etc.

It has always been the case that one might be disconnected and then reconnect, and of course this doesn't change that. Just wondering if its worth setting expectations for script writers now. E.g., undetermined (as to whether it is the old or new domain or neither) vs reliably fail vs possibly crash?

hifi-gustavo:
Build console output for commit: d5891e9073ec21abeae0448de448e312a2dc5f86 available at: https://gist.github.com/7bc9b6def96c2e2fb4c4b56c9cd3dd63<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7692.dmg

howard-stearns:
Brad answered my last question, where the short answer is that such behavior is not specified (either before or after this PR, i.e., no change).  The expectation is that such methods shouldn't apply to the new domain, and may or may not happen to the old (as if the wire might be cut), and if that's not true, it would be a separate ticket outside this PR.

OmegaHeron:
Testing Windows 7

Rapidly change domains over and over again  ...  verify that you see this in your log for every time you change domains... 

Yes - in all cases I (eventually) saw message in log.

Entity creating message flood test - saw the flood - changed domains and saw;
[04/19 11:56:36] [DEBUG] script:print()<< unload - entityID:{c012fbf7-52b5-4eaf-b967-89b2c00bac7b}
[04/19 11:56:36] [DEBUG] script:print()<< NOTE --- WE DID NOT CALL Script.update.disconnect()
[04/19 11:56:36] [DEBUG] Script Engine shutting down (destructor) for script: "Entities 29"

Returned to where ^^ was located and deleted it - the log flood continued until I left domain and returned.  That was unexpected - it seems deleting entity containing test script did not terminate it until I changed domain/returned or it had so much queued log spew it kept going.  Unsure there.

Created entity with timeOutFunction script - saw it logging its messages - changed domain - messages stopped - returned and they began again.

Deleted entity with timeOutFunction script and messages stopped.

General smoke test shows no change to stable version in render, audio, general function.

No tests with hand controllers made.

Crashed on closing - bugsplat sent.

ZappoMan:
@OmegaHeron - re: "Returned to where ^^ was located and deleted it - the log flood continued until I left domain and returned. That was unexpected - it seems deleting entity containing test script did not terminate it until I changed domain/returned or it had so much queued log spew it kept going. Unsure there." -- yes actually that would be expected behavior. Basically this is an intentionally poorly written entity script, it's not doing proper cleanup, and so it doesn't cleanup on unload. We can however clean it up when we do the full domain cleanup switching domains... so that's expected.

PR #7709, opened 2016-04-19
Introduce Shader Compilation versions

samcake:
This PR is preparation for stereo drawcalls.

This PR add a notion of version in the GLBackend::GLShader of the gpu::Shader allowing to create several version of the shaders from the same source but using different compilation defines.
THis will beused to create a new version of the shaders for the stereo case which will do the shader slightily differently.

The glsl version is now assigned at runtime to the glsl shader..

FOr the simple Uniform values (which we don't want to support down the road, this is not ideal for modern API) we need the location information to match for all the different versions so we introduce a UniformMapping for the versions after number 0

This is not used for now and will be used in a future PR.


samcake:
To Test:
Try everywhere and every thing should look the same as in MAster
I can imagine some procedural shaders or rare materials using a different shader path to show issues maybe. Tey should be reported naturally as non compiled shaders.

hifi-gustavo:
Build console output for commit: 7c7cd6e9e56cfc75640c20348837c54d1e9149ff available at: https://gist.github.com/aca58d54d2970ad1f46e1ee9513f19d4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7709.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7709.dmg

kevinmthomas-carpool:
## QA Test
**PR Build:** 7709
**Testing OS:** Windows 10
**Testing Locations:** localhost, playa, cellscience, forum, music, free, entry, capitola, polyworld, sandbox
**Additional Test Results Requested:** Successful test.  Visited multiple domains listed above and everything appears to look the same as the master.  Shaders looked normal and expected. 
**Successful Logout/Crash:** Failure, initial tidy bot crash however expected and not tied to this PR.  Successful logout on second attempt.  Log file included in the event of a need for deeper inspection.
[attachments.zip](https://github.com/highfidelity/hifi/files/225632/attachments.zip)

PR #7719, opened 2016-04-20
Add prefetching to exposed caches

zzmp:
- Add `prefetch` to JS `ResourceCache` interface
- Hold onto resources as long as JS holds a ref
- Remove `MyAvatar::prefetchAnimation` and `Rig::prefetchAnimation` (and update `away.js`)
- Return a `ScriptableResource` from the `prefetch` to query resource status (loaded, failed, progress)
- Add a `Script.requestGarbageCollection` method to JS


zzmp:
See: https://app.asana.com/0/74051501495175/109187623266499.
This is not a P1 nor a crash. It is a feature, so probably shouldn't be overly prioritized.

#### Intent

1) Allow content-makers to query their resources to know when they are loaded.
2) Allow content-makers to preload certain resources, to avoid delays from loading later on.

#### Testing

*I will make test scripts in a day or two, when I have more free time. QA may want to hold off until then.*

- Ensure `away.js` still works. This used the old API. Part of this PR is updating it.
- Open up cacheCounters.js. Try loading a resource from JS, and make sure it loads. When you release it and a garbage collection cycle runs, it should fall into the cache or off the counter completely. If this is built in DEBUG, you will have a method available to you to trigger garbage collection: `requestGarbageCollection`. Its effects may not be immediate.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -270,6 +270,20 @@ static void resultHandlerFromScriptValue(const QScriptValue& value, AnimVariantR
     assert(false);
 }
 
+// Templated qScriptRegisterMetaType fails to compile with raw pointers
+using ScriptableResourceRawPtr = ScriptableResource*;
+
+static QScriptValue scriptableResourceToScriptValue(QScriptEngine* engine, const ScriptableResourceRawPtr& resource) {
+    auto object = engine->newQObject(
+        const_cast<ScriptableResourceRawPtr>(resource),
+        QScriptEngine::ScriptOwnership);
```
zzmp: It may be good to `reportAdditionalMemoryCost` here, or through a connection on progress. However, multiple `ScriptableResources` may refer to the same memory, so I held off for now.

hifi-gustavo:
Build console output for commit: b19ad5be07b40671a3cd6422f00ebbcb4dfc32bc available at: https://gist.github.com/9a014704498441d15737189d5b403384<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.dmg

zzmp:
test this please

hifi-gustavo:
Build console output for commit: b19ad5be07b40671a3cd6422f00ebbcb4dfc32bc available at: https://gist.github.com/9a014704498441d15737189d5b403384<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.dmg

zzmp:
Marking DNM while I add texture specificity, report add'l memory usage, and write tests.

hifi-gustavo:
Build console output for commit: c59eac6630c4f9221d5087a96040a7f8fdacb3f5 available at: https://gist.github.com/9a014704498441d15737189d5b403384<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.dmg

hifi-gustavo:
Build console output for commit: 713fc44ec5738789dbca9973ddc539b44ce0de82 available at: https://gist.github.com/9a014704498441d15737189d5b403384<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.dmg

libraries/model-networking/src/model-networking/TextureCache.h
```diff
@@ -25,32 +26,83 @@
...
+    virtual bool isCacheable() const override { return _loaded; }
+
+    virtual void downloadFinished(const QByteArray& data) override;
+          
+    Q_INVOKABLE void loadContent(const QByteArray& content);
+    Q_INVOKABLE void setImage(gpu::TexturePointer texture, int originalWidth, int originalHeight);
+
+private:
+    Type _type;
+    TextureLoaderFunc _textureLoader;
```
AndrewMeadows: I was auditing the initialization of data members and noticed that `_textureLoader` is not initialized in one of the NetworkTexture ctors.  I know this bug existed before this PR but could you please fix it here?

libraries/networking/src/ResourceCache.cpp
```diff
@@ -119,6 +119,73 @@ QSharedPointer<Resource> ResourceCacheSharedItems::getHighestPendingRequest() {
...
+        result->_progressConnection = connect(
+            resource.data(), &Resource::onProgress,
+            result, &ScriptableResource::progressChanged);
+        result->_finishedConnection = connect(
+            resource.data(), &Resource::finished,
+            result, &ScriptableResource::finished);
+    }
+
+    return result;
+}
```
AndrewMeadows: I don't understand how the memory is managed here.  The `result`Could you explain how the raw pointer `result` is eventually deleted?

libraries/networking/src/ResourceCache.cpp
```diff
@@ -354,8 +422,8 @@ void ResourceCache::removeResource(const QUrl& url, qint64 size) {
     _totalResourcesSize -= size;
 }
 
-void ResourceCache::updateTotalSize(const qint64& oldSize, const qint64& newSize) {
-    _totalResourcesSize += (newSize - oldSize);
+void ResourceCache::updateTotalSize(const qint64& deltaSize) {
+    _totalResourcesSize += deltaSize;
```
AndrewMeadows: I would like to have some sanity checking here, such as an assert if _totalResourcesSize ever goes negative, and maybe also if it ever gets larger than some reasonable number... like a petabyte.

libraries/model-networking/src/model-networking/TextureCache.h
zzmp: I don't think this is a bug, unless you consider uninitialized members a bug in general.

The only time this member is used is from `getTextureLoader`, which only uses it if `_type == CUSTOME_TEXTURE`, i.e., if the appropriate ctor was used.

libraries/networking/src/ResourceCache.cpp
zzmp: I am moving the prefetch methods to be private, as they should only be called by JS/QML, not other cpp, and I want to make that self-evident.
I am adding comments on top of `prefetch` to (hopefully) make ownership more clear.

The ownership is handled by the `ScriptEngine`. It takes over ownership using `QScriptEngine::newQObject` in ScriptEngine.cpp, in a specific registration that I added for the `ScriptableResource*` metatype. I needed to specify `QScriptEngine::ScriptOwnership` when I transferred it, but that has the GC deleting the pointers memory from JS. See: https://github.com/highfidelity/hifi/pull/7719/files#diff-0e211914feb9a81572284cc68e8a39a3R280.

I probably need a QML test case as well.

libraries/model-networking/src/model-networking/TextureCache.h
AndrewMeadows: I would consider it a bug for two reasons:

(1)  it makes the NetworkTexture API buggy.  What is preventing an innocent developer in the future from using the wrong ctor and then calling getTextureLoader() on it?  It is a sharp edge that is waiting to cut someone -- we should smooth the edges of our software to reduce bleeding.

(2) The lack of initialization of a function pointer is a **wtf?** moment for future developers who read the code and worry about (1).  In other words, this:  http://www.osnews.com/images/comics/wtfm.jpg

libraries/networking/src/ResourceCache.cpp
zzmp: Why a petabyte? How about a terabyte?

hifi-gustavo:
Build console output for commit: febc2d018b281642077c83b7bab2359c60b3e3c8 available at: https://gist.github.com/9a014704498441d15737189d5b403384<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.dmg

hifi-gustavo:
Build console output for commit: 2fd88d9fe1414b0825413a34ff560bae0ed59d42 available at: https://gist.github.com/9a014704498441d15737189d5b403384<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.dmg

hifi-gustavo:
Build console output for commit: 6e12493ed73abdb56b05968afda0b3efeba9623e available at: https://gist.github.com/9a014704498441d15737189d5b403384<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.dmg

hifi-gustavo:
Build console output for commit: 4df417e7568ed7a744435df2fcb63f6680a365fe available at: https://gist.github.com/9a014704498441d15737189d5b403384<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.dmg

libraries/networking/src/ResourceCache.cpp
```diff
@@ -119,6 +119,73 @@ QSharedPointer<Resource> ResourceCacheSharedItems::getHighestPendingRequest() {
...
+    }
+}
+
+ScriptableResource* ResourceCache::prefetch(const QUrl& url, void* extra) {
+    auto result = new ScriptableResource(url);
+
+    if (QThread::currentThread() != thread()) {
+        // Must be called in thread to ensure getResource returns a valid pointer
+        QMetaObject::invokeMethod(this, "prefetch", Qt::BlockingQueuedConnection,
+            Q_RETURN_ARG(ScriptableResource*, result), Q_ARG(QUrl, url));
```
zzmp: This needs to pass `void*, extra` as another argument.

libraries/networking/src/ResourceCache.h
```diff
@@ -121,12 +166,24 @@ public slots:
...
+    // Pointers created through this method should be owned by the caller,
+    // which should be a QScriptEngine with ScriptableResource registered, so that
+    // the QScriptEngine will delete the pointer when it is garbage collected.
+    Q_INVOKABLE ScriptableResource* prefetch(const QUrl& url) { return prefetch(url, nullptr); }
+
+
+    // Prefetches a resource to be held by the QScriptEngine.
+    // Left as a protected member so subclasses can overload prefetch
+    // and delegate to it (see TextureCache::prefetch(const QUrl&, int).
+    ScriptableResource* prefetch(const QUrl& url, void* extra);
```
zzmp: This needs to be a slot or invokable in order to be usable with QMetaObject::invokeMethod.

hifi-gustavo:
Build console output for commit: 069065dc23a7e66f32aa896dc22b3a165e69c4b5 available at: https://gist.github.com/9a014704498441d15737189d5b403384<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.dmg

libraries/script-engine/src/ScriptEngine.h
```diff
@@ -131,6 +131,8 @@ class ScriptEngine : public QScriptEngine, public ScriptUser, public EntitiesScr
     Q_INVOKABLE void callEntityScriptMethod(const EntityItemID& entityID, const QString& methodName, const MouseEvent& event);
     Q_INVOKABLE void callEntityScriptMethod(const EntityItemID& entityID, const QString& methodName, const EntityItemID& otherID, const Collision& collision);
 
+    Q_INVOKABLE void requestGarbageCollection() { collectGarbage(); }
```
zzmp: With the new compiler flags, should this only be available for PR builds? I think it's a nice feature for production builds too, but don't want to be unilateral about it.

zzmp:
FYI: The black flicker that I mentioned in the testing comments is due to the texture being asynchronously copied from the CPU to the GPU on first use. It will last at least one frame for any new texture.

We can counter it (in the case of `prefetch`) by either starting this transfer earlier, and only firing a `loadedChanged` signal when it is transferred; or by marking the prefetched textures to be transferred synchronously. I think the former is better, because it helps to prevent stutter from long frames waiting for texture transfers. However, I think it belongs in a separate PR (but I've got rough drafts implemented).

libraries/networking/src/ResourceCache.cpp
```diff
@@ -119,6 +119,74 @@ QSharedPointer<Resource> ResourceCacheSharedItems::getHighestPendingRequest() {
...
+    if (_progressConnection) {
+        disconnect(_progressConnection);
+    }
+    if (_finishedConnection) {
+        disconnect(_finishedConnection);
+    }
+}
+
+ScriptableResource* ResourceCache::prefetch(const QUrl& url, void* extra) {
+    auto result = new ScriptableResource(url);
```
hyperlogic: Hey, so I think @AndrewMeadows is right, this is a leak.  When called on the script thread it allocates a new ScriptableResource, then dispatches to the main thread.  The main thread will then allocate another ScriptableResource() which is returned to the script thread.  The script thread will then overwrite it's resource ptr with the one from the main thread, without freeing the original ScriptableResource.

zzmp: You're right, the allocation should only occur if on the correct thread.

hyperlogic:
So, one issue I have about the ScriptableResource API.  It seems odd to me that there are two signals for loadedChanged and failedChanged and that loadedChanged becomes true even if there is an error.  A user might think that if loadedChanged is triggered the asset was loaded successfully, and when failedChanged occurs on failure.  But they'd be wrong...  How about using a finished signal with a boolean argument?  or how about following XHR and readyState property and a readyStateChanged signal that's a number not a boolean.

hyperlogic:
Hey, about the scriptableResourceTest script.  It's kind of broken looking, it just flashes black and white.  Is there a way you can either slow it down so that the source images are visible?  Or just remove the actual movie playback all together, and just maybe prefetch the resources and free them after you receive all of the finished notifications?

examples/tests/scriptableResourceTest.js
```diff
@@ -0,0 +1,76 @@
...
+    return s;
+}
+
+var pictureFrameProperties = {
+    name: 'scriptableResourceTest Picture Frame',
+    type: 'Model',
+    position: center,
+    modelURL: FRAME_URL,
+    dynamic: true,
+}
```
hyperlogic: semicolon.

examples/tests/scriptableResourceTest.js
```diff
@@ -0,0 +1,76 @@
...
+for (var i = 1; i <= NUM_FRAMES + 1; i++) {
+    var padded = pad(i, 3);
+    var filepath = MOVIE_URL + padded + '.jpg';
+    var texture = TextureCache.prefetch(filepath);
+    frames.push(texture);
+    if (!texture.loaded) {
+        numLoading++;
+        texture.loadedChanged.connect(function() {
+            numLoading--;
+            if (!numLoading) play();
```
hyperlogic: please add brackets to this if

examples/tests/scriptableResourceTest.js
```diff
@@ -0,0 +1,76 @@
...
+function play() {
+    var frame = 0;
+    var movieInterval = Script.setInterval(function() {
+        Entities.editEntity(pictureFrame, { textures: JSON.stringify({ Picture: frames[frame].url }) })
+        frame += 1;
+        if (frame > NUM_FRAMES) {
+            Script.clearInterval(movieInterval);
+            Entities.deleteEntity(pictureFrame);
+            // free the textures at the next garbage collection
+            while (frames.length) frames.pop();
```
hyperlogic: please add brackets to this while

script-archive/theBird.js
```diff
@@ -20,8 +20,6 @@ for (i = 0; i < l; i++) {
     print(roles[i]);
 }
 
-MyAvatar.prefetchAnimation(THE_BIRD_RIGHT_URL);
```
hyperlogic: please add an AnimationCache.prefetch here as well.

hifi-gustavo:
Build console output for commit: bc8a171fdeb815ea04a613875860680fb3bb0619 available at: https://gist.github.com/9a014704498441d15737189d5b403384<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7719.dmg

script-archive/theBird.js
zzmp: It is used immediately after, so the prefetch will not accomplish anything.

hifi-gustavo:
Build console output for commit: 5bf9112cf55e31866ddc576518f3bfc8606d8203 available at: https://gist.github.com/9a014704498441d15737189d5b403384<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7719.dmg

hifi-gustavo:
Build console output for commit: 854c0ea3e3ab36b0b527f900f5a1dc029a083835 available at: https://gist.github.com/9a014704498441d15737189d5b403384<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7719.dmg

kevinmthomas-carpool:
## QA Test
**PR Build:** 7719
**Testing OS:** Windows 10
**Additional Test Results:** Successful test.  Successful test of ```away.js```, screen shot enclosed.  Successful test of ```cacheCounters.js```.  Successful test of ```prefetchTest.js```, screen shots enclosed.  Upon loading ```createCow.js```, value of ```prefetchTest.js``` was 16.1.  After the load of ```createCow.js```, value of ```prefetchTest.js``` was 0, therefore it appears it It looks like it pulled it out of the cache. 
away
value before load of createcow
value after load of createcow

scripts/developer/tests/scriptableResource/lib.js
```diff
@@ -19,15 +19,15 @@ function getFrame(callback) {
     var FRAME_URL = "http://hifi-production.s3.amazonaws.com/tutorials/pictureFrame/finalFrame.fbx";
 
     var model = ModelCache.prefetch(FRAME_URL);
-    if (model.loaded) {
-        makeFrame(true);
+    if (model.state = Resource.State.FINISHED) {
```
hyperlogic: should be ==

scripts/developer/tests/scriptableResource/lib.js
```diff
@@ -19,15 +19,15 @@ function getFrame(callback) {
...
+        makeFrame(Resource.State.FINISHED);
     } else {
         model.stateChanged.connect(makeFrame);
     }
 
     function makeFrame(state) {
-        if (state == 4) { throw "Failed to load frame"; }
-        if (state != 3) { return; }
+        if (state == Resource.State.FAILED) { throw "Failed to load frame"; }
```
hyperlogic: It might be best to use === and !==, in general for these comparisons.  Because in JavaScript `"" == 0` is true.

hyperlogic:
Other than the = bug, this PR is ready to go.

hifi-gustavo:
Build console output for commit: 29e0744b89bb52652f084f243c92b3959428f7c0 available at: https://gist.github.com/9a014704498441d15737189d5b403384<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7719.dmg

PR #7726, opened 2016-04-20
Fix for path MTU discovery fail

Atlante45:
This patch fixes the case where pmtu_disc fails.

- Make sure the DontFragment flag is not set for the socket.
- Drop the max packet size from 1500 to 1492 with is a fairly common MTU.

https://app.asana.com/0/26551373415487/114763027727130

### DIFF
--- a/libraries/networking/src/udt/Constants.h
+++ b/libraries/networking/src/udt/Constants.h
@@ -18,7 +18,7 @@
 
 namespace udt {
     static const int UDP_IPV4_HEADER_SIZE = 28;
-    static const int MAX_PACKET_SIZE_WITH_UDP_HEADER = 1500;
+    static const int MAX_PACKET_SIZE_WITH_UDP_HEADER = 1492;
     static const int MAX_PACKET_SIZE = MAX_PACKET_SIZE_WITH_UDP_HEADER - UDP_IPV4_HEADER_SIZE;
     static const int MAX_PACKETS_IN_FLIGHT = 25600;
     static const int CONNECTION_RECEIVE_BUFFER_SIZE_PACKETS = 8192;
--- a/libraries/networking/src/udt/Socket.cpp
+++ b/libraries/networking/src/udt/Socket.cpp
@@ -38,6 +38,21 @@ Socket::Socket(QObject* parent) :
     _synTimer->start(_synInterval);
 }
 
+void Socket::bind(const QHostAddress& address, quint16 port) {
+    _udpSocket.bind(address, port);
+    setSystemBufferSizes();
+
+#if defined(Q_OS_LINUX)
+    auto sd = _udpSocket.socketDescriptor();
+    int val = IP_PMTUDISC_DONT;
+    setsockopt(sd, IPPROTO_IP, IP_MTU_DISCOVER, &val, sizeof(val));
+#elif defined(Q_OS_WINDOWS)
+    auto sd = _udpSocket.socketDescriptor();
+    int val = 0; // false
+    setsockopt(sd, IPPROTO_IP, IP_DONTFRAGMENT, &val, sizeof(val));
+#endif
+}
+
 void Socket::rebind() {
     quint16 oldPort = _udpSocket.localPort();
     
--- a/libraries/networking/src/udt/Socket.h
+++ b/libraries/networking/src/udt/Socket.h
@@ -60,7 +60,7 @@ public:
     qint64 writeDatagram(const char* data, qint64 size, const HifiSockAddr& sockAddr);
     qint64 writeDatagram(const QByteArray& datagram, const HifiSockAddr& sockAddr);
     
-    void bind(const QHostAddress& address, quint16 port = 0) { _udpSocket.bind(address, port); setSystemBufferSizes(); }
+    void bind(const QHostAddress& address, quint16 port = 0);
     void rebind();
     
     void setPacketFilterOperator(PacketFilterOperator filterOperator) { _packetFilterOperator = filterOperator; }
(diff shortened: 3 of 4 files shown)


hifi-gustavo:
Build console output for commit: da13d8cf3c89ed4697fd2682eb43763b38ac58f5 available at: https://gist.github.com/c9f34a920d9af916fabad3b595bdc3dc<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7726.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7726.dmg

libraries/networking/src/udt/Constants.h
```diff
@@ -18,7 +18,7 @@
 
 namespace udt {
     static const int UDP_IPV4_HEADER_SIZE = 28;
-    static const int MAX_PACKET_SIZE_WITH_UDP_HEADER = 1500;
+    static const int MAX_PACKET_SIZE_WITH_UDP_HEADER = 1492;
```
ZappoMan: Columbus would be proud... but in other news... are we sure we need to do this? I did a quick search, I can't find what packet size the audio mixer and avatar mixer use... this change will definitely change the entity server and message mixer.

Atlante45: We don't HAVE to, but it turns out that 1492 is a [fairly common MTU size](We don't HAVE to, but it turns out that 1492 is a fairly common MTU size).
So any network path with such MTU will suffer fragmentation on all maximum sized packets. This will approximately divide the packet rate by a factor of 2.

Atlante45: I'm looking at everything that would be affected by that change right now.

hifi-gustavo:
Build console output for commit: 778c19d7e404a3abf0ffdb421dcd6de5bf8f3202 available at: https://gist.github.com/c9f34a920d9af916fabad3b595bdc3dc<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7726.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7726.dmg

Atlante45:
Testing the actual issue this fixes is pretty complicated and requires several specific circumstances.
I already tested the fix and am confident it will work, so we will test for regressions instead.

Testing plan:
- General smoke test.
- Make sure Avatar/Entity animations go correctly over the wire.
- Audio Injectors are working.
- Drawing board works. (Polylines go correctly over the wire)

libraries/networking/src/udt/Constants.h
Atlante45: I trailed through everything using that constant and didn't find anything that would cause an issue.
The only thing I'm worried about is big entity properties. (i.e.. I know polylines limit themselves to fit on the wire)

ZappoMan: That doesn't concern me... basically it means there are 8 fewer bytes...
some props won't fit, but we've never made a claim about the max size of
property supported... in fact I've generally said ~1000 is the most.

OmegaHeron: Going back to the dim mists of modern internet.... an engineering group decided 1492 would be maximal unfragmented routed packet size - this was attributed to both limitations of hardware in that time, and, one of task board members was helping a child memorize - "In 1492 Columbus sailed the ocean blue".  That said - there are still routers out there, especially in lesser developed places, with a MTU of 1492.  :)

Atlante45: @ZappoMan I think @OmegaHeron just called your house a "lesser developed place" xD

seriously though, there's no guarantee about either:
IPv4: "At least 68, max of 64KB"
IPv6: "At least 1280, max of 64KB, but up to 4GB with optional jumbo gram"
[[source](https://en.wikipedia.org/wiki/Maximum_transmission_unit#Table_of_MTUs_of_common_media)]

Eventually, we'll want to use [Path MTU Discovery](https://en.wikipedia.org/wiki/Path_MTU_Discovery) either by utilizing the OS's one or implementing our own.
Also I'd like to point out this PR is only necessary because Ubuntu fails to find the correct PMTU for the path in question.

OmegaHeron: Agreed on discovery and.. naw... ;)  I was just exhibiting one of the few privileges to being 50 something years old and having sat through too many boring engineering meetings in dot com boom V1.0 circa mind 1990s - Also I'm compiling this to test on Ubuntu 16.04/QT5.6.0 as I write this.

Atlante45: Nice, the issue we were seeing is that the Music Ubuntu box was falling to find the MTU to Zappoman's machine (was set to 1500 when it was actually 1492).
And all packet add the DontFragment flag set by default (OS setting I believe), so when streaming a file (with all max sized packets) nothing would get through.

OmegaHeron: Yeah - or it does eventually get through as seen at Music - if you hang there 20+ minutes it will load - I see it with some of my content too so I'm hoping to see a change once I roll this to my testing server.

Atlante45: If this happens to you, you theoretically should see a ~2x decrease in ATP content load times.

OmegaHeron:
Initial testing on Linux shows an improvement loading standard home scene and my problematic scene that exhibits behavior similar to Kevin's Music domain.  I'm unable to test drawing board or viewing of other avatars at this time.  I think it would be wise to attempt a deploy of this on Kevin's Linux server and see how it works there.

Atlante45:
@OmegaHeron if @kevinmthomas-carpool or @thoys give me the green light, I can deploy that PR on Music to do a quick test.

kevinmthomas-carpool:
@Atlante45 green light for ```music```.

PR #7733, opened 2016-04-21
Allow prefetched textures to transfer synchronously to GPU

zzmp:
When switching to a new texture on a model, the texture is asynchronously loaded.
Because the texture is not loaded, the model renders with some default texture for a frame.

This allows prefetched textures to load synchronously. For large textures, the framerate may take a hit as they are transferred, but it allows use cases like playing a video made of textures.

It is based on the prefetching feature added in #7719. The only new commits are after 4df417e.


hifi-gustavo:
Build console output for commit: 1436424c355624ecfdbf36faba2249017ca48b2f available at: https://gist.github.com/d6b94e1714781d3d064125d15849c28e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7733.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7733.dmg

PR #7737, opened 2016-04-21
move debug for resource retry to when retry occurs

birarda:
- delay log for resource retry to show only during actual retry

### DIFF
--- a/libraries/networking/src/ResourceCache.cpp
+++ b/libraries/networking/src/ResourceCache.cpp
@@ -519,10 +519,9 @@ const int MAX_ATTEMPTS = 8;
 void Resource::attemptRequest() {
     _startedLoading = true;
 
-    auto timer = qobject_cast<QTimer*>(sender());
-    if (timer) {
-        qCDebug(networking).nospace() << "Server unavailable for" << _url
-            << "- retrying asset load after" << timer->interval() << "ms - attempt" << _attempts << " of " << MAX_ATTEMPTS;
+    if (_attempts > 0) {
+        qCDebug(networking) << "Server unavailable for" << _url
+            << "- retrying asset load - attempt" << _attempts << " of " << MAX_ATTEMPTS;
     }
 
     ResourceCache::attemptRequest(_self);
--- a/libraries/networking/src/ResourceCache.cpp
+++ b/libraries/networking/src/ResourceCache.cpp
@@ -514,8 +514,17 @@ void Resource::init() {
     }
 }
 
+const int MAX_ATTEMPTS = 8;
+
 void Resource::attemptRequest() {
     _startedLoading = true;
+
+    auto timer = qobject_cast<QTimer*>(sender());
+    if (timer) {
+        qCDebug(networking).nospace() << "Server unavailable for" << _url
+            << "- retrying asset load after" << timer->interval() << "ms - attempt" << _attempts << " of " << MAX_ATTEMPTS;
+    }
+
     ResourceCache::attemptRequest(_self);
 }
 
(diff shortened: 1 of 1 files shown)


libraries/networking/src/ResourceCache.cpp
```diff
@@ -514,8 +514,17 @@ void Resource::init() {
     }
 }
 
+const int MAX_ATTEMPTS = 8;
+
 void Resource::attemptRequest() {
     _startedLoading = true;
+
+    auto timer = qobject_cast<QTimer*>(sender());
+    if (timer) {
```
zzmp: Instead of checking timer, could you check the number of attempts?

libraries/networking/src/ResourceCache.cpp
```diff
@@ -602,13 +611,10 @@ void Resource::handleReplyFinished() {
...
             case ResourceRequest::Result::ServerUnavailable: {
                 // retry with increasing delays
-                const int MAX_ATTEMPTS = 8;
                 const int BASE_DELAY_MS = 1000;
                 if (_attempts++ < MAX_ATTEMPTS) {
                     auto waitTime = BASE_DELAY_MS * (int)pow(2.0, _attempts);
-                    qCDebug(networking) << "Server unavailable for" << _url <<
-                        "retrying in " << waitTime << "ms," <<
-                        "attempt " << _attempts + 1 << "of" << MAX_ATTEMPTS;
+
```
zzmp: Can we still log the failure here, and also log the retry there?

birarda:
This avoids an issue where the Interface log would show that it was retrying an ATP asset download if there was a pending download while switching domains.

To test:

- Go to an ATP heavy domain (forum or Sandbox content)
- Reload all content
- Immediately, while reloading, switch domains (ideally to a domain that is disconnected like nimdok since your log will pollute less)
- Reload all content and stop your Asset Server assignment

Test the above on both master and this PR. On master you will see a log entry when you switch domains for retry of ATP assets (assuming you catch at least one or two in progress when you switch domains). On this PR you will not see that debug.

On this PR you should still see the retry debug in the case where you are still connected to the same domain and you stop your Asset Server assignment.

libraries/networking/src/ResourceCache.cpp
birarda: Good call

libraries/networking/src/ResourceCache.cpp
birarda: I wanted the timer so I could say how long it had been waiting. Since that's in the fail debug now I'll make this simpler with attempts.

hifi-gustavo:
Build console output for commit: 3a8c2144bf6685d0b9ce29010271910a3460a3f8 available at: https://gist.github.com/f9af89a799256c57310b5633c7bee0b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7737.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7737.dmg

libraries/networking/src/ResourceCache.cpp
```diff
@@ -514,8 +514,17 @@ void Resource::init() {
...
 }
 
+const int MAX_ATTEMPTS = 8;
+
 void Resource::attemptRequest() {
     _startedLoading = true;
+
+    auto timer = qobject_cast<QTimer*>(sender());
+    if (timer) {
+        qCDebug(networking).nospace() << "Server unavailable for" << _url
```
zzmp: Why are you using nospace? Also, you need a space after the word "attempt"

hifi-gustavo:
Build console output for commit: 8cd330f4310ea62a5d9e69822690f1907b24a1db available at: https://gist.github.com/f9af89a799256c57310b5633c7bee0b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7737.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7737.dmg

birarda:
Killed Jenkins jobs for all but the last build.

hifi-gustavo:
Build console output for commit: 5ea8b83745aee65fafb4d54435dd2454c95aade8 available at: https://gist.github.com/f9af89a799256c57310b5633c7bee0b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7737.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7737.dmg

OmegaHeron:
Used hifi://o2tech for testing since I control it and it is a remote server with standard home scene.

Comparing logs from RC6 to this, I believe it is doing what PR intends.  General smoke test passes as well.  QA Pass

PR #7711, opened 2016-04-19
faster hifi kinematic motion calculations take-4

AndrewMeadows:
* added more Bullet profile info
* fixed bug where kinematic objects can "pop forward" when first set in motion
* slight performance improvements for kinematic motion calculations
* improved support for slow moving kinematic objects
* send fewer updates for fast-moving objects

PR #7739, opened 2016-04-22
fix entity-server crash when object leaves the domain bounds - take 2

AndrewMeadows:
* fix entity-server crash when object leaves the domain bounds (really this time)


libraries/entities/src/EntitySimulation.cpp
```diff
@@ -232,6 +233,7 @@ void EntitySimulation::changeEntity(EntityItemPointer entity) {
         }
         changeEntityInternal(entity);
     }
+    */
 }
```
AndrewMeadows: There are two places in EntitySimulation.cpp where we check for an object leaving the domain bounds:

(1) every frame we walk EntitySimulation::_entitiesToSort and delete any that are outside.  The only entities that are put on that list are those that were moved by the simulation itself (e.g. via kinematic integration -- this is the entity-server we're talking about).

(2) Whenever an entity's position is changed (via an update packet triggered by a script, for example) then that code was supposed to eventually call EntitySimulation::changeEntity() which would then do the out-of-bounds check, and would remove it if so.

Path (1) was for the EntitySimulation to use and path (2) was for anything outside of EntitySimulation.  Unfortunately a way into path (2) was added as a side effect of the work done in EntitySimulation::moveSimpleKinematics(), which can invalidate an iterator, which crashes.

I believe the bad side effects have been removed in #7711, but it hasn't merged yet, so this is the quick and easy fix -- just eliminate path (2).  A consequence of this fix is that if an entity is moved out of bounds by some external script the entity-server will not necessarily catch it.  If the entity is translating then the kinematic simulation stuff will eventually detect it and remove it, but if it is at rest then it will happily sit outside the bounds, which is better than crashing for the short term, hence this fix.

hifi-gustavo:
Build console output for commit: 691fa8033ebedc441d65fb2a1d49c9f188668de8 available at: https://gist.github.com/dab8b9a16e9564317baec8f30623bd55<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7739.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7739.dmg

ZappoMan:
we're going to close this and instead go with #7711 which has the fix

PR #7743, opened 2016-04-22
group several separate glUniform parameters into a single Uniform Buffer

samcake:
THis PR groups the separate uniforms used to communicate texcoordTransform and lightmap parameters into a uniform Buffer managed from the Material.
This avoids glUniform calls and setup work happening for every Shape items and replace it with a single SetUniformBuffer which calls a single glUniformBufferRange

Introducing the TexMapArray buffer in the material class to contain these data and more in the future.

This PR should show no differences in look or significant performance change


hifi-gustavo:
Build console output for commit: 43caca901fe020d0c4b30cde809f365eef066219 available at: https://gist.github.com/b78e438ea082fa26771a11089ab0f94e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7743.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7743.dmg

OmegaHeron:
Tested Windows 7 - Nvidia GTX660M
General smoke tests shows no difference between this and RC6 for performance.  All functions work as expected.  Depending upon what's in conflict this may need retesting.  I didn't notice conflict warning before test.  QA Pass unless conflict resolution impacts validity.

hifi-gustavo:
Build console output for commit: 5cb346330be15dacf657e65e94475eda7a960bad available at: https://gist.github.com/b78e438ea082fa26771a11089ab0f94e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7743.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7743.dmg

libraries/model/src/model/Material.h
```diff
@@ -298,9 +298,21 @@ class Material {
     // conversion from legacy material properties to PBR equivalent
     static float shininessToRoughness(float shininess) { return 1.0f - shininess / 100.0f; }
 
+    // Texture Map Array Schema
+    static const int NUM_TEXCOORD_TRANSFORMS{ 2 };
+    class TexMapArraySchema {
+    public:
+        glm::mat4 _texcoordTransforms[NUM_TEXCOORD_TRANSFORMS];
+        glm::vec4 _lightmapParams{ 0.0, 1.0, 0.0, 0.0 };
```
AndrewMeadows: Except for these numerical literals that should be explicit floats the code changes in this PR look good to me.

The "schema pattern" here is new to me: the sole purpose of `TexMapArraySchema` appears to be to provide an object instance of the correct size for memory allocations.

samcake: The main goal of the pr is to replace separate calls to glUniform* to set
the values of the corresponding variables required by the shaders...
... And instead gather this variables in a gpu::buffer bound as
Uniformbuffer expected by the shader
The gpu::buffer binding is done ine a single efficient gl call (triggered
by Batch::setUniformBuffer())
If needed, the video memory copy of the buffer is updated (memcpyed) by the
system memory version

The sysmem chunk is what is mapped to the Schema structure and guarantees
and simplify the authoring of the variables
If the gpuBuffer is accessed by a "edit<Schema>()" call then a version
stamp is bumped so we can keep track and compare with the last known stamp
of the gpu memory version

Since we can group these variables under the umbrella of a Material then
most if the time they are constant for a given material
The Material owns its instance of texMapArrayBuffer which is pretty much
constant for the whole life of the Material.
It s a very nice granularity to manage during rendering

Then when we need to use the Material, we can basicaly setup its values by
binding the 2 contained schemaBuffers and binding the textures it uses


Thanks

S

PR #7746, opened 2016-04-23
reduce warnings about truncated kinematic timesteps

AndrewMeadows:
* reduce warnings about kinematic timesteps

### DIFF
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -884,7 +884,24 @@ void EntityItem::simulate(const quint64& now) {
 }
 
 bool EntityItem::stepKinematicMotion(float timeElapsed) {
+    // get all the data
+    Transform transform;
+    glm::vec3 linearVelocity;
+    glm::vec3 angularVelocity;
+    getLocalTransformAndVelocities(transform, linearVelocity, angularVelocity);
+
+    // find out if it is moving
+    bool isSpinning = (glm::length2(angularVelocity) > 0.0f);
+    float linearSpeedSquared = glm::length2(linearVelocity);
+    bool isTranslating = linearSpeedSquared > 0.0f;
+    bool moving = isTranslating || isSpinning;
+    if (!moving) {
+        return false;
+    }
+
     if (timeElapsed <= 0.0f) {
+        // someone gave us a useless time value so bail early
+        // but return 'true' because it is moving
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


AndrewMeadows:
A recent change introduced a warning about "truncated kinematic timesteps" and we found out that we were seeing a lot of them.  The dire nature of the warnings did not properly reflect what was actually going on: most of those warnings were for non-moving objects that hadn't been changed for a long time --> kinematic extrapolation of their movement is a no-op.

This PR moves things around so that we don't get the warning so much and also properly initializes the `EntityServer::_lastSimulated` data member whenever an object is added to the list of non-physical kinematic objects (in EntitySimulation).

The warning still shows up at login sometimes for moving objects -- I think what is happening is either the packets really do sit for several seconds or there is an unknown clock skew between the client and server at that point which is eventually properly measured and goes away.  We'll leave it at "warning" level for now and consider dropping it to "debug" later.

hifi-gustavo:
Build console output for commit: d984ce47922594e5cc993e0d48869a0539973528 available at: https://gist.github.com/3565bc695edd0b6eb6d342605cc0b601<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7746.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7746.dmg

AndrewMeadows:
To test:

(1) Log into any domain
(2) Bring up the interface log window (I think it is ctrl+alt+shift+L)
(3) Check for a burst of logs that look like this:
```
[04/22 22:07:23] [WARNING] kinematic timestep = 80661.9 truncated to 1
[04/22 22:07:23] [WARNING] kinematic timestep = 80661.9 truncated to 1
[04/22 22:07:23] [WARNING] kinematic timestep = 80661.9 truncated to 1
[04/22 22:07:23] [WARNING] kinematic timestep = 32607.6 truncated to 1
[04/22 22:07:23] [WARNING] kinematic timestep = 80662 truncated to 1
[04/22 22:07:23] [WARNING] kinematic timestep = 80662 truncated to 1
[04/22 22:07:23] [WARNING] kinematic timestep = 80662.1 truncated to 1
```
(4) This PR reduces the big burst to just a few -- one for each kinematic object in view.  There appears to still be some long timesteps computed as the interface and entity-server figure out their clock skew.  We'll look into those later, but all of the innocuous warnings have been removed.

kevinmthomas-carpool:
## QA Test
**PR Build:** 7746
**Testing OS:** Windows 10
**Additional Test Results Requested:** Successful test.  Log shows no trace of any mention of ```kinematic timestep``` occurrences.  Log file included for deeper inspection.
**Successful Logout/Crash:** Failure.  Crash upon logout of interface.  Log file included for deeper inspection.  
[attachments.zip](https://github.com/highfidelity/hifi/files/233027/attachments.zip)

ZappoMan:
The crash that Kevin had was unrelated to this PR... so I'm going to call this QA Pass

PR #7747, opened 2016-04-23
Fixes for avatar eye tracking.

hyperlogic:
An incorrect Head::_eyePosition was causing all sorts for problems for eye tracking.  

  * The mouth position was an offset much higher vertically than normal, which caused the eye "saccades" to be incorrect.
  * All of the eye look at fix up code was not being activated, which caused the eyes to not properly look at the camera in desktop mode.
  * The eye lid offset was computed from the top of the head rather than the center of the eyes which caused the eyes to droop more than they should on all our avatars.

### DIFF
--- a/interface/src/avatar/Head.cpp
+++ b/interface/src/avatar/Head.cpp
@@ -230,7 +230,7 @@ void Head::simulate(float deltaTime, bool isMine, bool billboard) {
     }
     
     _leftEyePosition = _rightEyePosition = getPosition();
-    _eyePosition = calculateAverageEyePosition();
+    _eyePosition = getPosition();
 
     if (!billboard && _owningAvatar) {
         auto skeletonModel = static_cast<Avatar*>(_owningAvatar)->getSkeletonModel();
@@ -238,6 +238,8 @@ void Head::simulate(float deltaTime, bool isMine, bool billboard) {
             skeletonModel->getEyePositions(_leftEyePosition, _rightEyePosition);
         }
     }
+
+    _eyePosition = calculateAverageEyePosition();
 }
 
 void Head::calculateMouthShapes() {


hifi-gustavo:
Build console output for commit: c850880515425a20dbc404da9d3fe229757e7199 available at: https://gist.github.com/dd431647194c2536fe21931773c07340<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7747.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7747.dmg

interface/src/avatar/Head.cpp
```diff
@@ -230,14 +230,16 @@ void Head::simulate(float deltaTime, bool isMine, bool billboard) {
...
+    _eyePosition = getPosition();
 
     if (!billboard && _owningAvatar) {
         auto skeletonModel = static_cast<Avatar*>(_owningAvatar)->getSkeletonModel();
         if (skeletonModel) {
             skeletonModel->getEyePositions(_leftEyePosition, _rightEyePosition);
         }
     }
+
+    _eyePosition = calculateAverageEyePosition();
```
PhilipRosedale: why set _eyePosition to getPosition() and then calculateAverageEyePosition() without using it in between?

ZappoMan: I was wondering something similar, I assume that something in the middle of these lines of code depends on one version of the value, while other code apparently depends on the other version. 

I wonder why this difference exists? Maybe the cold that needs the head position should just use that directly.

PhilipRosedale:
Tested that we now correctly look down at matthews mouth, rather than up to the forehead, as is the case in production.

ZappoMan:
@PhilipRosedale did you do a full smoke test pass? Shouldn't we have QA still do a pass?

interface/src/avatar/Head.cpp
ZappoMan: In looking through the code more carefully (not on my iPhone) -- it seems to me that the call to `_eyePosition = getPosition();` at *line 233* is not needed and could cause unintended side effects.

I believe the root issue that Tony has found and fixed is that the call to `_eyePosition = calculateAverageEyePosition();` that he moved to *line 242* needs to happen AFTER the call to `skeletonModel->getEyePositions(_leftEyePosition, _rightEyePosition);` or else it basically just returns the average of the unmodified values for `_leftEyePosition`, `_rightEyePosition` which would be `getPosition()` e.g. the head's position.

So it's clear that the correct place for the `_eyePosition = calculateAverageEyePosition();` call is *line 242*.

However... it seems like resetting `_eyePosition = getPosition();` at *line 232* is redundant, and could cause a "blip" of bad eye position by any thread that accesses that value between the calls at line 232 and 242.

It's redundant, because it's only really guarding against the possibility that you can't get the correct individual eye positions from the model (as handled by `getEyePositions()`) but in that case, the result of `calculateAverageEyePosition()` is still the same.

I'd suggest we remove the call to `_eyePosition = getPosition();`

PhilipRosedale:
OK, I have removed the "QA Approved" label.  However does our smoke test yet involve inspecting the eye movements of mutiple different avatars in and out of HMD's?  @cozza13 is adding this, but probably not yet.  I think that I will head to the office and use HMD's there to test this with several avatars, since this is a change needed for the SVVR demos and for the video we are shooting monday.

ZappoMan:
Sounds reasonable... clearly testing the eye tracking is something we should be doing. I think this is a relatively safe change - it seems very unlikely it would introduce other issues. But I'm more just trying to make sure we continue to follow the process that Chris has put in place for QA Approved.

kevinmthomas-carpool:
## QA Test
**PR Build:** 7747
**Testing OS:** Windows 10
**Testing Locations:** localhost, playa, cellscience
**Smoke Test Results:** Successful test.

                                  |localhost |playa    |cellscience  
----------------------------------|----------|---------|-----------
Render Rate:                      |60.04     |60.01    |59.32
Present Rate:                     |60.00     |60.00    |60.01
Simrate:                          |60.03     |59.00    |59.99
Render Stats Num Buffers:         |3391      |2734     |2797
Render Stats gpu::Buffer Memory:  |65.4Mb    |70.6Mb   |49.7Mb
Render Stats Num Textures:        |331       |283      |124
Render Stats gpu::Texture Memory: |1001.0Mb  |895.5Mb  |468.0Mb
Render Stats Triangles:           |511k      |419k     |501k
Render Stats Drawcalls:           |250       |1379     |169
Render Stats State Changes:	  |233       |192      |74
Render Stats Items:               |292       |233      |782
Render Stats Timing:              |40.5ms    |6.1ms    |16.1ms
GPU-Z Memory Size                 |4096Mb    |4096Mb   |4096Mb
GPU-Z Initial Memory		  |262Mb     |262Mb    |262Mb
GPU-Z Memory Used		  |2365Mb    |1513Mb   |1628Mb
GPU-Z Memory Load		  |26%       |34%      |22%

**localhost ‘createtestblocks.js’ Result:** Successful test.
**localhost ATP Upload (kelectricguitar10.fbx):** Successful test.
**localhost ‘harmonicOscillator.js’ Result:** Successful test.
**localhost Echo Server Audio Result:** Successful test.
**localhost ‘reverbTest.js’ Result:** Successful test.
**localhost ‘playaPerformanceTest.js’ Result:** Successful test.  Test results significantly beneath the red line.  Screen shot enclosed in attachments.
**playa ‘playaPerformanceTest.js’ Result:** Successful test.  Test results significantly beneath the red line.  Screen shot enclosed in attachments.
**cellscience ‘playaPerformanceTest.js’ Result:** Successful test.  Test results significantly beneath the red line.  Screen shot enclosed in attachments.
**Additional Test Results Requested:** (none)
localhost
playa
cellscience

PR #7765, opened 2016-04-26
Improve eye tracking when face to face

hyperlogic:
Before this PR, an angle check between the center eye positions of two avatars was used to determine if one avatar was looking at another.  As two avatars become face to face, this test would unexpectedly fail as one character moved his head side to side, because the volume of the head was not taken into account.   With this PR, we use a different check, we use a cone sphere test.  The angle that is checked used is the angle between the cone and a point on the sphere that is "closest" to the cone.  This is less likely to fail as avatars come face to face.

### DIFF
--- a/libraries/shared/src/GeometryUtil.cpp
+++ b/libraries/shared/src/GeometryUtil.cpp
@@ -564,3 +564,17 @@ void swingTwistDecomposition(const glm::quat& rotation,
     // rotation = swing * twist  -->  swing = rotation * invTwist
     swing = rotation * glm::inverse(twist);
 }
+
+// calculate the minimum angle between a point and a sphere.
+float coneSphereAngle(const glm::vec3& coneCenter, const glm::vec3& coneDirection, const glm::vec3& sphereCenter, float sphereRadius) {
+    glm::vec3 d = sphereCenter - coneCenter;
+    float dLen = glm::length(d);
+
+    // theta is the angle between the coneDirection normal and the center of the sphere.
+    float theta = acosf(glm::dot(d, coneDirection) / dLen);
+
+    // phi is the deflection angle from the center of the sphere to a point tangent to the sphere.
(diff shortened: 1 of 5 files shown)


hyperlogic:
https://app.asana.com/0/45663370065758/116194190897617

hyperlogic:
To test:
  1. enter a domain with two avatars.
  2. enter HMD mode and move your face very close to the other avatar.
  3. In the release build the eye tracking between avatars would fail very easily as you moved your face side to side, as you were very close to the other avatar.  This PR fixes that, it should be less likely to loose eye contact when you are face to face with another avatar.

hifi-gustavo:
Build console output for commit: 71f67c99ccbd401810bcaa79ca8a736dfa28da90 available at: https://gist.github.com/313c137db2bcb2153da6ef5d5983f2fe<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7765.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7765.dmg

libraries/shared/src/GeometryUtil.cpp
```diff
@@ -564,3 +564,17 @@ void swingTwistDecomposition(const glm::quat& rotation,
     // rotation = swing * twist  -->  swing = rotation * invTwist
     swing = rotation * glm::inverse(twist);
 }
+
+// calculate the minimum angle between a point and a sphere.
```
ZappoMan: this comment seems wrong... isn't it between cone and sphere?

AndrewMeadows: Maybe it should be... "calculate the minimum angle between a ray and a sphere".

AndrewMeadows:
I tested.  It is better than before.

PR #7766, opened 2016-04-26
fix auto resize of models on import

ZappoMan:
The auto resize code wasn't properly setting the lastEdit time of the EntityItemProperties it was using to auto-resize the model. This would cause a race condition where the edit would be rejected if the servers scene data arrived before the resize was processed.

### DIFF
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -518,6 +518,7 @@ void RenderableModelEntityItem::update(const quint64& now) {
     if (!_dimensionsInitialized && _model && _model->isActive()) {
         if (_model->isLoaded()) {
             EntityItemProperties properties;
+            properties.setLastEdited(usecTimestampNow()); // we must set the edit time since we're editing it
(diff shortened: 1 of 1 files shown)


ZappoMan:
This fixes this bug: https://app.asana.com/0/26320649018758/107174761588361

Test plan:
- smoke test
- import a model using edit.js
- import a model from the Asset Server

hifi-gustavo:
Build console output for commit: b597f7001b06b81272876e71f3e1a8cbc2623d3d available at: https://gist.github.com/c549e20f963b76ea2f48f9b8fa3da1e1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7766.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Alpha-PR7766.dmg

Nex-Pro:
Used the Asset Browser to upload models to Sandbox and my Localhost.
Uploads succeeded without issue.

*Noticed that the edit icons are arranged horizontally rather than the normal vertical layout.
Edit icon was part way off the screen.*

#### On the issues requiring testing, Success

PR #7787, opened 2016-04-28
Prevent pure virtual calls from Resources after ResourceCache is freed

zzmp:
Orphan resources (`setCache(nullptr)`) when their cache is freed. If there are any remaining resources after the cache is freed, this prevents their dtor from invoking methods on the cache, which would result in pure virtual calls.

This does *not* prevent a crash if the ResourceCache is freed during normal operations - only when it is freed during shutdown, and Resources outlive it.


zzmp:
Hopefully, this prevents the pure virtual calls on shutdown (see: https://app.asana.com/0/74051501495175/117189807738560), but I could not repro those, so I do not know if this contrived repro is for the same issue.

#### Testing

*Windows only.*
- Run the build, and attach it to an instance of Visual Studio.Do this by going to the Task Manager, right-clicking on Interface, and selecting Debug.
- Set Visual Studio to break on all exception types. Do this by going to Debug > Exceptions... and checking all Thrown categories.
- Free the texture cache. Do this in the console by calling `TextureCache.deleteLater()`. *This should be protected, but is not (known issue). It will crash Interface.*
- In current build, you will crash on a pure virtual function call.<br>
In this PR build, you will crash on an access violation.<br>
These should both be visible through Visual Studio, in Output.

hifi-gustavo:
Build console output for commit: 13d68c507f1685f2b6924fe34fca7b1ed177ea12 available at: https://gist.github.com/6f9d66ef3ff856169996fec0dc4c330d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7787.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7787.dmg

Nex-Pro:
**PR Build: #7801**
**Testing Architecture: Desktop PC i7 4.4 (@ 4.0 ghz) 16gb nvidia gtx770 2gb**
**Testing OS: Win10 64** 

See attached Visual studio debug/output dumps.

[crash_output_window_contentsPR7787.txt](https://github.com/highfidelity/hifi/files/245799/crash_output_window_contentsPR7787.txt)

[crash_output_window_contents_build4821.txt](https://github.com/highfidelity/hifi/files/245800/crash_output_window_contents_build4821.txt)

Nex-Pro:
@zzmp  Let me know if this is adequate.

zzmp:
@Nex-Pro did VS say what kind of crashes they were? I noticed only the PR output said "Unhandled", which is probably right, but I can't be sure without knowing what kind of exceptions they were.

zzmp:
Tested with @Nex-Pro, he observed both crashes in the pure virtual call handler. QA Fail until I know why.

zzmp:
This does not prevent virtual calls, nor does it fix anything. Closing.

PR #7805, opened 2016-05-03
Improve the performance of idle event loops

zzmp:
Changes the mechanism by which `Application::idle` is called from a busy wait to a condition_variable.

With a busy wait, everything was always running hot. Throttled display plugins (e.g. Desktop) ran as hot (on the CPU) as anything else (e.g. DK2).
Using a condition_variable instead allows CPU use to drop back down when not using Interface, an observed problem on i5 chips.


hifi-gustavo:
Build console output for commit: ed9a669571f46c63fbecb81c7e06f47b8a9796b7 available at: https://gist.github.com/038f2db64ebeadbc1b2f2aa92062faae<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7805.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7805.dmg

interface/src/Application.cpp
```diff
@@ -1095,17 +1095,11 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
...
-    _idleTimer = new QTimer(this);
-    connect(_idleTimer, &QTimer::timeout, [=] {
-        idle(usecTimestampNow());
-    });
-    connect(this, &Application::beforeAboutToQuit, [=] {
-        disconnect(_idleTimer);
-    });
-    // Setting the interval to zero forces this to get called whenever there are no messages
-    // in the queue, which can be pretty damn frequent.  Hence the idle function has a bunch
-    // of logic to abort early if it's being called too often.
```
Atlante45: This comment is super important otherwise that code doesn't make much sense.

Atlante45:
I don't see how this improves performance, you removed the unused idle arg, made the timer a non member and moved the heartbeat, I don't think any of those change will affect idle perfs.
According to the comment you removed it's probably already running too often and has to be slowed down.

hifi-gustavo:
Build console output for commit: 9719a6a103afcd13c17aee2cbad1efce84476263 available at: https://gist.github.com/038f2db64ebeadbc1b2f2aa92062faae<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7805.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7805.dmg

Atlante45:
I might be missing something, but I think IdleThread's `notify` and `aboutToQuit` slots never actually get called.
IdleThread never processes events and all those calls come from different thread. So I think they just get put in the event list and never get processed.
But since the cv only waits for 1ms idle still gets called fairly often.

zzmp:
`notify` and `aboutToQuit` signals are sent through an [auto-connection](http://doc.qt.io/qt-5/qt.html#ConnectionType-enum), which are sent with a `Qt::DirectConnection` if the object lives in the same thread.

Because the `IdleThread` has the `Application` as a parent, [it lives in that thread](http://doc.qt.io/qt-5/qobject.html#thread-affinity), so the signals are sent. They are not processed on the `IdleThread` but in the main thread, which is fine, as they are sent over using the `condition_variable`.

As if threaded applications weren't confusing enough, now we have Qt ;).

You can verify that `notify` is called with breakpoints (I have verified).
Waiting for 1ms would be too long, as it would cost us 1/13th of our rendering time.

zzmp:
See: https://app.asana.com/0/74051501495175/119003179687438.

#### Testing

*This changes the basic logic preceding rendering.*
*Should be tested on Desktop and HMD.*
*Should be tested on Windows and Mac and UNIX.*

Make sure it runs correctly under light and heavy load. Open your Task Manager while you test to ensure it works.

##### Light load
Run on a disconnected domain, with no scripts running, on Desktop mode. When Interface is in focus, you should get 60fps. When you focus on a different window, you should throttle to 15fps. In the Task Manager, you should see that the CPU load also goes down - you are no longer running hot.

##### Heavy load
Run on playa with other people. Make sure you can maintain your frame rate as you did before.

Start in Desktop and switch to HMD. Make sure you can maintain your frame rate as you did before. Now go to playa with lots of other people. Make sure you can...

Run editEntityStressTest.js. Make sure it runs as well as it did before. It will degrade your performance - this is nothing new. Make sure you can still use your mouse while running this, in Desktop and HMD, as you could before. Try stopping the script while it is running in Desktop and HMD, as you could before.

Atlante45:
Gotcha, that makes sense for Application, but `incrementPresentCount` gets called from the PresentThread no?

zzmp:
It would still call the slot on the next loop of the main thread, because that's where the slot lives. It would be delayed by one loop, which is similar to current behavior.

jherico:
I can't really test this against a lower end card as I don't have one available.  Seems to pass a basic smoke test.

ZappoMan:
I'm kind of concerned about this PR... it seems like we're putting all of our eggs BACK INTO the Qt message system. We've previously found this to not be reliable.

I'd like to see a design review on this to better understand how we've verified that this works in all cases.

interface/src/Application.cpp
```diff
@@ -1474,21 +1468,9 @@ void Application::initializeUi() {
 
 
 void Application::paintGL() {
-    updateHeartbeat();
-    // Some plugins process message events, potentially leading to
-    // re-entering a paint event.  don't allow further processing if this
-    // happens
-    if (_inPaint) {
```
ZappoMan: we know that `paintGL()` is not reentrant, have we verified it can never be called reentrantly with this new code? Maybe this should be an `assert()`

interface/src/Application.cpp
```diff
@@ -1474,21 +1468,9 @@ void Application::initializeUi() {
...
-        return;
-    }
     _inPaint = true;
     Finally clearFlagLambda([this] { _inPaint = false; });
 
-    // paintGL uses a queued connection, so we can get messages from the queue even after we've quit
-    // and the plugins have shutdown
-    if (_aboutToQuit) {
```
ZappoMan: same issue here... maybe this should be an `assert()` or left in.

interface/src/Application.cpp
```diff
@@ -2618,59 +2595,30 @@ bool Application::acceptSnapshot(const QString& urlString) {
...
-    if (isThrottled && (timeSinceLastUpdateUs / USECS_PER_MSEC) < THROTTLED_SIM_FRAME_PERIOD_MS) {
-        // Throttling both rendering and idle
-        return; // bail early, we're throttled and not enough time has elapsed
+    // Throttle if requested
+    if (displayPlugin->isThrottled() && (msecondsSinceLastUpdate < THROTTLED_SIM_FRAME_PERIOD_MS)) {
+        return;
     }
 
     auto presentCount = displayPlugin->presentCount();
-    if (presentCount < _renderedFrameIndex) {
```
ZappoMan: please explain this change in the logic

interface/src/Application.cpp
```diff
@@ -2618,59 +2595,30 @@ bool Application::acceptSnapshot(const QString& urlString) {
...
-        _renderedFrameIndex = INVALID_FRAME;
-    }
-
-    // Don't saturate the main thread with rendering and simulation,
-    // unless display plugin has increased by at least one frame
-    if (_renderedFrameIndex == INVALID_FRAME || presentCount > _renderedFrameIndex) {
```
ZappoMan: please explain this change in the logic

interface/src/Application.cpp
```diff
@@ -2618,59 +2595,30 @@ bool Application::acceptSnapshot(const QString& urlString) {
...
-
-        // request a paint, get to it as soon as possible: high priority
-        postEvent(this, new QEvent(static_cast<QEvent::Type>(Paint)), Qt::HighEventPriority);
+    // Throttle to present rate
     } else {
-        // there's no use in simulating or rendering faster then the present rate.
         return;
     }
 
-    // NOTICE NOTICE NOTICE NOTICE
```
ZappoMan: please explain why the new design does not require this warning.

interface/src/Application.cpp
```diff
@@ -5094,8 +5050,9 @@ void Application::updateDisplayMode() {
         std::unique_lock<std::mutex> lock(_displayPluginLock);
 
         auto oldDisplayPlugin = _displayPlugin;
-        if (_displayPlugin) {
-            _displayPlugin->deactivate();
+        if (oldDisplayPlugin) {
+            disconnect(oldDisplayPlugin.get(), &DisplayPlugin::presentCountChanged, _idleThread, &IdleThread::notify);
```
ZappoMan: I'm concerned that we're using signal/slots to handle this communication between the display plugin and the idle thread. It strikes me that this depends too much on Qt's messaging and signal/slots... when a more traditional thread synchronization technique would be more reliable. Why not just use a basic wait condition or some other synchronization technique?

zzmp:
Marking DNM while I take time to address @ZappoMan's concerns.

interface/src/Application.cpp
jherico: The idle function is no longer wired to a timer with an interval of 0, so it no longer gets called every single time there are no other events in the processing queue.  However I'm not clear on what is triggering idle now.

ZappoMan: I figured out what causes idle to get called... do I win a prize?

interface/src/Application.cpp
jherico: neither the main thread or the present thread should *ever* block doing some sort of wait.  Ultimately, as long as the main thread is responsible for rendering a frame, it has to do so in response to an event injected into the event queue.  

However, I'm concerned that this addition of an idle thread is sort of overkill.  All it does is wait for something in the present thread to change and respond by issuing an event on the main thread.  What is the advantage of this over doing a postEvent directly from the present thread?

interface/src/IdleThread.cpp
```diff
@@ -0,0 +1,37 @@
+//
```
jherico: if we want to take this approach, this class should be in the DisplayPlugins library, and the connection / disconnection should be automatically handles as part of t a display plugins activation / deactivation.

interface/src/Application.cpp
```diff
@@ -2618,59 +2595,30 @@ bool Application::acceptSnapshot(const QString& urlString) {
...
 
     auto presentCount = displayPlugin->presentCount();
-    if (presentCount < _renderedFrameIndex) {
-        _renderedFrameIndex = INVALID_FRAME;
-    }
-
-    // Don't saturate the main thread with rendering and simulation,
-    // unless display plugin has increased by at least one frame
-    if (_renderedFrameIndex == INVALID_FRAME || presentCount > _renderedFrameIndex) {
+    if (_renderedFrameIndex != presentCount) {
```
zzmp: @ZappoMan This should be equivalent to the old logic.

interface/src/IdleThread.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "Application.h"
+
+IdleThread::IdleThread(QObject* parent) :
+    QThread(parent) {}
+
+void IdleThread::run() {
+    while (!_aboutToQuit) {
+        std::unique_lock<std::mutex> lock(mutex);
+        condition.wait_for(lock, std::chrono::milliseconds(1)); // spurious wakeups are fine
+        QMetaObject::invokeMethod(qApp, "idle", Qt::BlockingQueuedConnection); // prevent duplicate calls
```
zzmp: I tried using a postEvent here, but in practice it meant that heavy scripting would prevent mouse use.
Using a BlockingQueuedConnection instead allows GUI use during heavy load *and* guarantees that `idle` is never called reentrantly.

interface/src/IdleThread.cpp
zzmp: Even though it relies on an invocation to Application::idle?

interface/src/Application.cpp
zzmp: I tried doing a postEvent from the IdleThread and it resulted in poor GUI performance. I assume the same would be true in the present thread, but I can test it separately.

zzmp:
Per our conversation (with @ZappoMan and @jherico):
- Post idle events from the PresentThread, or
- Make DisplayPlugin a friend of IdleThread and avoid signal/slot mechanism
- Investigate invokeMethod and see how it is prioritized
- In priority order: don't ruin UI performance, don't busy wait, don't lose 1/2ms

zzmp:
@jherico is checking Win32 specific timers to see if they offer improved precision and CPU performance, might be viable instead of adding a thread.

zzmp:
Using the timer introduced in f96ce91, I got the following delays (from DisplayPlugin::incrementPresentCount the present count to an invocation of Application::idle):

| Architecture | Time (us) | Time (under load: entityEditStressTest.js) | Blocks GUI? |
| --- | --- | --- | --- |
| Emit to IdleThread | 90-150 | 8000-9000 | |
| postEvent from DisplayPlugin | 30-250 | 8000-9000 | X |
| postEvent from DisplayPlugin (with removePostedEvents) | 30-250 | 80-800 | |

##### Analysis:

- `postEvent` is the fastest, regardless of priority.
- `postEvent` can flood the event queue, freezing GUI. IdleThread does not have this problem.
- `postEvent` with `removePostedEvents` will still flood the queue, but will clear it before bad perf.

zzmp:
Closing in favor of #7816.

PR #7808, opened 2016-05-03
Increase maximum friction value from 0.99 to 10.0

ctrlaltdavid:


ctrlaltdavid:
Affects Interface and Stack Manager.

Testing:
- Run the Stack Manager, Interface.exe, and edit.js from the PR build and test.
- Value > 0.99 can be set and retrieved
- Effect of value > 0.99 is as expected

Atlante45:
Why is max friction not 1.0f?
cc @AndrewMeadows Is it a physical reason?

hifi-gustavo:
Build console output for commit: bc04d1b4779ce391381fd3c864b22644f9be749a available at: https://gist.github.com/70e73be325a7c6e4d8314d1b38c28b10<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7808.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7808.dmg

AndrewMeadows:
I don't remember why **friction** was limited to something less than 1.0.  While it is true that: **restitution** should always be a little bit less than 1.0 (for stability of the physics simulation), and **damping** coefficients must not exceed 1.0, Bullet is OK with friction coefficients that exceed 1.0.

Atlante45:
@ctrlaltdavid Consulted with @AndrewMeadows and we should just set the max friction to 1.0f.

ctrlaltdavid:
Thanks @Atlante45 - have increased max friction to 1.0

hifi-gustavo:
Build console output for commit: d1064bf5ea7a84ce9c086c80d2811964675adc8a available at: https://gist.github.com/70e73be325a7c6e4d8314d1b38c28b10<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7808.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7808.dmg

AndrewMeadows:
Sorry, I should have made it more clear: coefficients of friction greater than 1.0 are allowed.  Bullet would be fine with a friction coefficient of 10.0, however the actual behavior of a tumbling object with such a coefficient might not be very different from one with a coefficient of 2.0.  Let's cap it at some reasonable number greater than 1.0, such as:  2, 4, 8 (a power of 2), 5, or 10 (easy fractions of our base-10 counting system).

In Bullet, the coefficient of friction is a multiplicative factor to the "normal force" for computing the maximum possible magnitude of the tangential force of sliding contact.  If the friction possibility prevents slippage then it does, otherwise it impedes it and the maximum value is used, which is why, in theory, there isn't much visible difference between setting the friction coefficient to 2.0 or 10.0 -- both cases would tend to prevent slippage for most contact events.

AndrewMeadows:
BTW, internally Bullet caps the friction coefficient to 10.0.  It also clamps to -10, but I don't think "negative friction" makes sense, so we should clamp to 0.0 on the low side.

Menithal:
# Test 

Using a A 2x0.2x2 dynamic box on a large angled surface with a downward gravity of 9, restitution of 0, and with both surfaces having a friction of 1.

# Result

Box stickt to surface, until the surface slanted beyond 45 degrees. This is regardless of the surface area of the dynamic entity

With Avatars, avatars do not have a friction of 9, and started sliding around 27.5 degrees slant. Probably a good idea to increase the friction on avatars?

Nex-Pro:
What does friction of 1.0 represent ? If Menithal's cube slid off at 45deg. then surely that isn't high friction. If it was really high the cube would remain until it's C of G moved beyond the downhill edge at which time it would roll/tumble  off the slope.

Menithal:
I can confirm however, that the friction can be set to 1. but I do not see much difference yet.* and it has no effect on the avatars as of the moment.

AndrewMeadows:
That sounds about right... if the box starts sliding at >45 degree slopes for a friction coefficient value of 1.0 then Bullet is using it to represent the "static friction" of the surface where the two materials start to slip.  Note that the actual friction of a contact point is the product of the friction coefficients of the objects involved, so if they are both 1.0 then the effective friction will indeed be 1.0.

ctrlaltdavid:
I've increased friction to 10.0, Bullet's maximum. The minimum friction is unchanged at 0.0.

hifi-gustavo:
Build console output for commit: 4b05d8ed951373cdf0e49147b1a695dbaa853c7c available at: https://gist.github.com/70e73be325a7c6e4d8314d1b38c28b10<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7808.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7808.dmg

Menithal:
Can confirm, can set friction to 10 now. It has some... interesting consequences

http://www.norteclabs.com/HF/screenshots/safetosay.jpg

Unfortunately however, avatars still slide, but I think that is quite intended :P could use some tweaks to it, but it probably isnt part of this PR

AndrewMeadows:
Yes, the avatar does not settle down in the physics simulation ATM.  This is independent of the friction setting of the floor it is standing on.

PR #7810, opened 2016-05-03
Improve CPU performance of Win32 usleep

zzmp:
Win32 only provides a Sleep with ms precision, but we use usleep throughout the codebase, especially for ScriptEngine.

When running a lot of ScriptEngines, this means a poor usleep implementation will make the CPU run hot. This improves the performance of usleep to use a busy wait (as it was) with msleep and yield when it can, with minimal loss of precision.


zzmp:
See: https://app.asana.com/0/74051501495175/119348421923943.

#### Testing

Run Developer > Timing > Run Timing Tests and make sure they are sane. They will show up in the log.
Run a bunch of scripts (defaultScripts.js does this). CPU performance should be better.

#### Review

I left a busy wait in usleep when waitTime <= 2us. I could get rid of it, and always yield if we have not waited long enough, and it would further improve performance, but in testing the lower waitTime values ended up waiting longer (3 instead of 1 us). I don't know how worthwhile that is.

hifi-gustavo:
Build console output for commit: a858d4cddcc258990c6c82a1ec9b3d29707bb72f available at: https://gist.github.com/77e41fb1fbea9091a2dbed84fdc4c3a4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7810.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7810.dmg

kevinmthomas-carpool:
## QA Test
**PR Build:** 7810
**Testing OS:** Windows 10
**Additional Test Results:** Ran ```Developer > Timing > Run Timing Tests``` and log file included.
[hifi-log.txt](https://github.com/highfidelity/hifi/files/248047/hifi-log.txt)

libraries/shared/src/SharedUtil.cpp
```diff
@@ -455,19 +455,28 @@ void printVoxelCode(unsigned char* voxelCode) {
...
+    // Busy wait with sleep/yield where possible
+    while (true) {
+        quint64 now = usecTimestampNow();
+        if (now >= sleepUntil) {
+            break;
         }
+
+        // Sleep if we have at least 1.1 ms to spare
+        // msleep is allowed to overshoot, so give it a 0.1 ms berth
+        if (sleepUntil - now > 1100) {
```
ZappoMan: magic number

libraries/shared/src/SharedUtil.cpp
```diff
@@ -455,19 +455,28 @@ void printVoxelCode(unsigned char* voxelCode) {
...
+            unsigned long sleepFor = (sleepUntil - now - 100) / USECS_PER_MSEC;
```
ZappoMan: magic number

libraries/shared/src/SharedUtil.cpp
```diff
@@ -455,19 +455,28 @@ void printVoxelCode(unsigned char* voxelCode) {
...
+            QThread::msleep(sleepFor);
+        // Yield if we have at least 2 us to spare
+        } else if (sleepUntil - now > 2) {
```
ZappoMan: magic number

ZappoMan:
Additional Testing Notes:

This code will change the underlying code in the Entity Server as well as Entity Editing. So we need to do a smoke test that involves verifying the functionality of the server as well.

Test for:
- Do files persist correctly (persist file timing is dependent on this code)
- Can we send lots of edit packets, do we honor the edit sending rate?
- Can agents connect to servers (including ACs)
- full smoke otherwise

ZappoMan:
@zach - maybe we should write a unit test for usleep() also/is it possible that with c++11 and VS2013 they actually provide this function for us?

zzmp:
@ZappoMan:
- We have a unit test, under Developer > Timing > Run Timing Tests. It just checks the timing, and not the CPU usage. I do not know how to do that in an automated way - any ideas?
- c++11 has sleep_for/sleep_until. I'll recode this using those portable functions and see how they perform on Win32, but its possible they just delegate to Sleep (which is what QThread::usleep does). Marking DNM until that work is done.

zzmp:
@jherico tested sleep_for and it only had ms precision, so that's out.

hifi-gustavo:
Build console output for commit: 9bae9159e0dbb8a92cd814bfe40d98c68dcebe17 available at: https://gist.github.com/77e41fb1fbea9091a2dbed84fdc4c3a4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7810.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7810.dmg

hifi-gustavo:
Build console output for commit: 4e6430f1b54feb0c1be19eb94f86d3784e18573f available at: https://gist.github.com/77e41fb1fbea9091a2dbed84fdc4c3a4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7810.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7810.dmg

hifi-gustavo:
Build console output for commit: 526143b097039dad5bee78f86b3c6018cef899cd available at: https://gist.github.com/77e41fb1fbea9091a2dbed84fdc4c3a4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7810.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7810.dmg

libraries/shared/src/SharedUtil.cpp
```diff
@@ -455,19 +455,29 @@ void printVoxelCode(unsigned char* voxelCode) {
...
-        quint64 compTimeSleep = compTime - BUSY_LOOP_USECS;
-        while (true) {
-            if (usecTimestampNow() < compTimeSleep) {
-                QThread::msleep(1);
-            }
-            if (usecTimestampNow() >= compTime) {
-                break;
-            }
+void usleep(int waitTime) {
+    quint64 sleepUntil = waitTime + usecTimestampNow();
```
ZappoMan: `usecTimestampNow()` is intended to give you the usecs since epoch according to the system time. This is possibly a more restricted form of what you actually care about or want... for example, if there's an OS service that just gives you running CPU time in a unit that is measurable in usecs, then that service might be preferable.

Our `usecTimestampNow()` does a bunch of extra work to to give you usecs since epoch... and so it may be lower performance than other approaches. And since you don't care about having a common baseline time, you only care about time elapsed since the caller called usleep... there may be better options.

libraries/shared/src/SharedUtil.cpp
```diff
@@ -455,19 +455,29 @@ void printVoxelCode(unsigned char* voxelCode) {
...
+
+    // Busy wait with sleep/yield where possible
+    while (true) {
+        quint64 now = usecTimestampNow();
```
ZappoMan: ibid

libraries/shared/src/SharedUtil.cpp
ZappoMan: Since this is a windows specific implementation, I'm sure there are more efficient Windows specifc API calls you could make... especially since in the last 100usecs you might be calling this function frequently.

hifi-gustavo:
Build console output for commit: 1553f532ba11c34da25829166f38e23585c5ccbf available at: https://gist.github.com/77e41fb1fbea9091a2dbed84fdc4c3a4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7810.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7810.dmg

OmegaHeron:
Round three - Default home scene with default scripts on.

Build 4825 - Standard home scene (over internet) - Default scripts on

4825-homescene-defaultscripts

This PR - Standard home scene (over internet) - Default scripts on

7816-homescene-defaultscripts

Again - this PR shows somewhat better performance over 4825, settling to a CPU use of 70% (approx) vs 77% (approx) for 4825.  Not good in either case, but, every bit helps.  Where this PR shines a bit is GPU utilization - it shows far less tendency to have zero GPU use periods than 4825 and its CPU use tends to stabilize toward a more happy lower value faster.

Much much more testing is needed, but, preliminary result is my i5 system is nearly equivalent in experience to my i7.  Is it a great experience in either case?  No.  I think a lot of it is the difference between one with and without an HMD.  I think we try to do things in not HMD mode you'd never do with one on like trying to spin around and move rapidly while a scene loads.  My understanding is you'd lose your lunch if you tried that with an HMD on. ;)  But - that said - the jerky loading and frozen in place UI while loading is one of the first things people brought here to look at HF comment on - then the OMG my fans are screaming and there's not even anything here! Then the OMG it just crashed and said I'm out of GPU RAM and there's hardly anything here!! WHY?!?!  Even on a system that's somewhat above stated minimum requirements Interface is pretty dismal.  If the minimum requirement was an i7 and GTX980 I'd be less prone to complain that my i5 with GTX970 is not cutting it - just as I never complained that my GTX660M + i7 laptop wasn't great here.

OmegaHeron:
Crap - I posted in wrong PR - fixing

OmegaHeron:
I took a look at this same way as I did for #7816 and I certainly, again, see improvement.  CPU use once scene stabilizes is lower and overall it feels more responsive.  Looks like about 20% CPU reduction over 4825 with default scripts on.


7810-homescene-defaultscripts

libraries/shared/src/SharedUtil.cpp
```diff
@@ -455,19 +455,44 @@ void printVoxelCode(unsigned char* voxelCode) {
...
-            if (usecTimestampNow() < compTimeSleep) {
-                QThread::msleep(1);
-            }
-            if (usecTimestampNow() >= compTime) {
-                break;
-            }
+void usleep(int waitTime) {
+    // Use QueryPerformanceCounter for least overhead
+    LARGE_INTEGER now; // ticks
+    QueryPerformanceCounter(&now);
```
jherico: this query call is superfluous, `now` is never referenced before the other call to `QueryPerformanceCounter`

PR #7816, opened 2016-05-05
Drive render loop from DisplayPlugin instead of busy wait

zzmp:
Changes the mechanism by which `Application::idle` is called from a busy wait to a `QEvent` posted by the `DisplayPlugin` when it is digesting a frame.

With a busy wait, everything was always running hot. Throttled display plugins (e.g. Desktop) ran as hot (on the CPU) as anything else (e.g. DK2).
Using a `QEvent` instead allows CPU use to drop back down when not using Interface, an observed problem on i5 chips.

A `condition_variable` was also explored, but this was simpler and had better performance (see #7805).


zzmp:
See: https://app.asana.com/0/74051501495175/119003179687438.

#### Testing

*This changes the basic logic preceding rendering.*
*Should be tested on Desktop and HMD.*
*Should be tested on Windows and Mac and UNIX.*

Make sure it runs correctly under light and heavy load. Open your Task Manager while you test to ensure it works.

##### Light load
Run on a disconnected domain, with no scripts running, on Desktop mode. When Interface is in focus, you should get 60fps. When you focus on a different window, you should throttle to 15fps. In the Task Manager, you should see that the CPU load also goes down - you are no longer running hot.

##### Heavy load
Run on playa with other people. Make sure you can maintain your frame rate as you did before.

Start in Desktop and switch to HMD. Make sure you can maintain your frame rate as you did before. Now go to playa with lots of other people. Make sure you can...

Run editEntityStressTest.js. Make sure it runs as well as it did before. It will degrade your performance - this is nothing new. Make sure you can still use your mouse while running this, in Desktop and HMD, as you could before. Try stopping the script while it is running in Desktop and HMD, as you could before.

hifi-gustavo:
Build console output for commit: 050a65b172f9a99d67ffe63dbef119e4ffbabc0a available at: https://gist.github.com/3115b77cddea8e8cf8d1cbff3d1022b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.dmg

interface/src/Application.h
```diff
@@ -93,6 +94,11 @@ class Application : public QApplication, public AbstractViewStateInterface, publ
     friend class PluginContainerProxy;
 
 public:
+    enum Event {
```
zzmp: Alternatively, these events could be defined in a separate, shared file.

OmegaHeron:
I'm going to start with simple tests and work up to the formal full smoke test.

First impression is this improves things greatly - loading home scene has less UI freezes and doesn't lead to my other apps struggling to get CPU time (for example - VLC playing an MP3 stream will stutter while running current production build when Interface hogs all CPU for extended periods - it does not with this PR).  That's how it feels - then there's science.  ;)

Build 4825 - Disconnected Space - No Scripts

4825-disco-noscripts

Once CPU use stabilizes it's running 55% (approx)  - Looking at lower left graph for CPU idle states there's hardly any.

This PR - Disconnected Space - No Scripts

7816-disco-noscripts

CPU use runs 35% (approx) - Idle states shows yields to allow other things to happen and all graphs show an, overall, better behavior.

Now moving onto Disco space - Default scripts on with Build 4825

4825-disco-defaultscripts

CPU use is now up to 72% (approx) - more congestion is indicated by UI Delays graph (upper left)

Disco space - Default scripts - This PR

7816-disco-defaultscripts

CPU use is 64% (approx) - there's less congestion overall, but, one can hope there's some goodness in another PR to improve this situation.  Just rezzing into empty space and spinning around a few times with default scripts on shows far far far too much CPU use.

OmegaHeron:
Round Two - Standard Home Scene (accessed over Internet)

Build 4825 - No Scripts

4825-homescene-noscripts

This PR - No Scripts

7816-homescene-noscripts

Again, this PR is better, but, we also begin to see how fetching models and stuffing GPU gets messy.  At least this PR has a little more CPU time to do that and, overall again, it feels better from just trying to load home and have it usable.  Note CPU idle states show some yielding for other things to get done.  Unfortunately UI Delays are still heavy - but less severe types in this PR as those clear once the heavy scene loading is completed vs current production where additional UI Delays are seen.  This is explained by CPU use dropping around 30 seconds on this PR and not doing so until later on production (and then not as low as this PR)

Of greater concern is GPU utilization.  The next to no GPU use periods correspond to unresponsive Interface (it shows FPS around 2.6).  This is improved as well by this PR.  But, it seems GPU is being starved for data during scene loading and I'm still learning how to dig deeper into traces to validate my theory there.

OmegaHeron:
Round three - Default home scene with default scripts on.

Build 4825 - Standard home scene (over internet) - Default scripts on

4825-homescene-defaultscripts

This PR - Standard home scene (over internet) - Default scripts on

7816-homescene-defaultscripts

Again - this PR shows somewhat better performance over 4825, settling to a CPU use of 70% (approx) vs 77% (approx) for 4825.  Not good in either case, but, every bit helps.  Where this PR shines a bit is GPU utilization - it shows far less tendency to have zero GPU use periods than 4825 and its CPU use tends to stabilize toward a more happy lower value faster.

Much much more testing is needed, but, preliminary result is my i5 system is nearly equivalent in experience to my i7.  Is it a great experience in either case?  No.  I think a lot of it is the difference between one with and without an HMD.  I think we try to do things in not HMD mode you'd never do with one on like trying to spin around and move rapidly while a scene loads.  My understanding is you'd lose your lunch if you tried that with an HMD on. ;)  But - that said - the jerky loading and frozen in place UI while loading is one of the first things people brought here to look at HF comment on - then the OMG my fans are screaming and there's not even anything here! Then the OMG it just crashed and said I'm out of GPU RAM and there's hardly anything here!! WHY?!?!  Even on a system that's somewhat above stated minimum requirements Interface is pretty dismal.  If the minimum requirement was an i7 and GTX980 I'd be less prone to complain that my i5 with GTX970 is not cutting it - just as I never complained that my GTX660M + i7 laptop wasn't great here.

jherico:
I've marked dev ready but I think this needs testing across more platforms than just one in order to pass QA.

kevinmthomas-carpool:
testing

judasshuffle:
https://i.gyazo.com/ce3510bc76cea085fc067e399e1efcf5.png had a bit of a play with this in playa with the dk2 on i'm using 38% of cpu, with everything loaded it seems smooth enough.Grinds some on my domain 'shop' but thats carved from lag at the best of times

kevinmthomas-carpool:
**PR Build:** 7816
**Testing OS:** Windows 10
**Testing Locations:** localhost, playa, cellscience
**Smoke Test Results:** Partially successful test.

                                  |localhost |playa    |cellscience   
----------------------------------|----------|---------|-----------
Render Rate:                      |60.08     |58.97    |60.57       
Present Rate:                     |60.04     |60.00    |60.00       
Present New Rate:                 |60.03     |58.99    |59.99      
Render Stats Num Buffers:         |5048      |5138     |2531        
Render Stats gpu::Buffer Memory:  |39.7Mb    |79.4Mb   |25.9Mb      
Render Stats Num Textures:        |302       |339      |143          
Render Stats gpu::Texture Memory: |639.6Mb   |1377.3Mb |430.4Mb     
Render Stats Triangles:           |425k      |763k     |533k       
Render Stats Drawcalls:           |1175      |573      |400         
Render Stats State Changes:	  |240       |182      |55          
Render Stats Items:               |244       |288      |655        
Render Stats Timing:              |45.2ms    |31.2ms   |21.8ms      
GPU-Z Memory Size                 |4096Mb    |4096Mb   |4096Mb      
GPU-Z Initial Memory		  |314Mb     |314Mb    |314Mb       
GPU-Z Memory Used		  |1389Mb    |1384Mb   |2053Mb      
GPU-Z Memory Load		  |25%       |26%      |15%        

**localhost ‘createTestBlocks.js’ Result:** Successful test.
**localhost ATP Upload (kelectricguitar10.fbx):** Successful test.
**localhost Echo Server Audio Result:** Successful test.
**localhost ‘harmonicOscillator.js’ Result:** Successful test.
**localhost ‘reverbTest.js’ Result:** Successful test.
**localhost ‘playaPerformanceTest.js’ Result:** Successful test.  Test results significantly beneath the red line.  Screen shot enclosed in attachments.
**playa ‘playaPerformanceTest.js’ Result:** Successful test.  Test results significantly beneath the red line.  Screen shot enclosed in attachments.
**cellscience ‘playaPerformanceTest.js’ Result:** Successful test.  Test results significantly beneath the red line.  Screen shot enclosed in attachments.
**Additional Test Results/Observations:** Microphone UI moved to the bottom right of the screen and not in its usual place in the top left.  After running a ```playaPerformance.js``` test on ```localhost```, it hung on trying to reconnect and therefore had to restart server.  Dramatically high timing on ```playa``` which was ```587.1``` in addition to after performing the ```playaPerformance.js``` test, content never reloaded properly and simply hung in a loading bar status.  ```cellscience``` domain successfully reloaded content after ```playaPerformance.js``` test however upon an attempt to teleport back to ```playa``` a crash occured.  After relog, did another ```playaPerformance.js``` test on ```playa```, successful reload of content.  Log file included for deeper analysis.
**Successful Logout/Crash:** Crash.
localhost performance
playa performance
cellscience performance
[hifi-log.txt](https://github.com/highfidelity/hifi/files/252666/hifi-log.txt)

cozza13:
-My Mic ended up down the bottom right (similar to above)
- crash on shutdown (may not be related to this https://www.bugsplatsoftware.com/individualCrash/?id=3285&database=interface_alpha )

ZappoMan:
I don't see how this PR could have caused the Mic/Mute/Audio levels ui to move... that's a head scratcher

zzmp:
@cozza13 my mic has been there for at least a few dev builds, now. That was not caused in this PR.

zzmp:
This was well behind master, so I merged in all the latest bells and whistles. @kevinmthomas-carpool @cozza13 could you retest and see if you are still crashing?

hifi-gustavo:
Build console output for commit: 0551da2c6269eebadfc5582b091149a36ca543b7 available at: https://gist.github.com/3115b77cddea8e8cf8d1cbff3d1022b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.dmg

kevinmthomas-carpool:
re-testing

kevinmthomas-carpool:
**PR Build:** 7816
**Testing OS:** Windows 10
**Testing Locations:** localhost, playa, cellscience
**Additional Test Results/Observations:** Same behavior where the audio components are in the bottom right and after a ```playaPerformance.js``` test, the domain would not fully res.  No difference from last test.

interface/src/Application.cpp
```diff
@@ -2676,79 +2641,87 @@ void Application::idle(uint64_t now) {
...
+                _idleLoopStdev.reset();
+            }
         }
-    }
 
-    _overlayConductor.update(secondsSinceLastUpdate);
+        _overlayConductor.update(secondsSinceLastUpdate);
+
+        // check for any requested background downloads.
+        emit checkBackgroundDownloads();
```
zzmp: This may account for the slowdown/failure in loading content.

zzmp:
Hangs, too.

zzmp:
The Audio Level meter moving is really really weird...

zzmp:
@kevinmthomas-carpool so the three things I need to fix are:
- Audio Level UI in wrong place
- Bad playaPerformance.js
- Running playaPerformance.js and switching domains without hangs/reload issues

Is this right?

kevinmthomas-carpool:
@zzmp yep you got it!

hifi-gustavo:
Build console output for commit: 4711c23d9d5b612214689c9711dcd7a8709128fe available at: https://gist.github.com/3115b77cddea8e8cf8d1cbff3d1022b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.dmg

zzmp:
Compiling on Release with DEBUG_PAINT_DELAY, I measured a paint delay of 0.01-0.02us.

libraries/plugins/src/plugins/DisplayPlugin.cpp
```diff
@@ -23,4 +24,18 @@ void DisplayPlugin::deactivate() {
...
 }
 
+int64_t DisplayPlugin::getPaintDelayUsecs() const {
+    return _paintDelayTimer.isValid() ? _paintDelayTimer.nsecsElapsed() / NSECS_PER_USEC : 0;
+}
+
+void DisplayPlugin::incrementPresentCount() {
+#ifdef DEBUG_PAINT_DELAY
+    // Avoid overhead if we are not debugging
+    _paintDelayTimer.start();
```
jherico: `incrementPresentCount` is called from the presentation thread, while `getPaintDelayUsecs` is called from the main thread.  QElapsedTimer is not documented as thread safe, so this is not safe usage.

interface/src/Application.h
```diff
@@ -93,6 +94,12 @@ class Application : public QApplication, public AbstractViewStateInterface, publ
...
+        Idle = DisplayPlugin::Paint,
+        Paint = Idle + 1,
```
jherico: these two lines are confusing.  I suggest renaming DisplayPlugin::Paint to `DisplayPlugin::Presented` or something, and removing the use of the term 'Idle'.  instead just have the Presented event type trigger the `Idle` function.

jherico:
The DEBUG_PAINT_DELAY code has a couple issues.  First, it's not thread safe.  Second, I don't see any mechanism that resets the computed values when you change display plugins.  Different plugins could conceivably produce different values.  

However, since it's not enabled in production and only for debugging perhaps it's not a big deal, but at the very least I would suggest putting warnings on the places where you use the timer that warning that it might have thread safety issues.

hifi-gustavo:
Build console output for commit: dd6a4dd091b87262716db6ece18dc8a8dd77f665 available at: https://gist.github.com/3115b77cddea8e8cf8d1cbff3d1022b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.dmg

hifi-gustavo:
Build console output for commit: 2d4fd783bd7f808dc9c718092f45b8b25c0de327 available at: https://gist.github.com/3115b77cddea8e8cf8d1cbff3d1022b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.dmg

OmegaHeron:
I've been running this PR for a while today.  It's certainly better in so far as CPU use on i5 and general feel to using it.  playaPerformanceTest.js shows a consistent "less spikey" appearance compared to current RC and xperf+win performance analyzer shows cleaner CPU use compared to my earlier measurements listed (way) above.  One problem with playaPerformanceTest is it's not really a consistent metric.  Playa is dynamic... did you have another AV come in during run?  did you have UFO flying through view?  Did something new get dropped while someone there with edit rights does some test or demo?  I can run it time after time after time and get a plot like below (or a little better) 

playaplot

Then go back for another run and get one with spikes will above top line all over it.  Why?  AVs entering/exiting or a new object coming into scope.  Its data is only useful subjectively and with knowledge of why it might look like it's a really bad plot when, really, it's just the way interface works now.  Something new coming into scene tends to make it stop - process and then move on.

jherico:
This PR reverts to older behavior that we worked hard to correct, i.e. if we can't keep up with the required render rate, we immediately drop to 1/2 the render rate, then 1/3, 1/4, etc...

zzmp:
I think this may have happened because I reverted to different idle and paint events, and did not test as rigorously as I had before. I will see if consolidating them again fixes it.

jherico:
No, it's caused because of this:  `removePostedEvents(this, Present);`  By removing queued events you're guaranteeing that if the painting takes longer than the current frame budget, you're dropping some integer fraction of the present events.

hifi-gustavo:
Build console output for commit: 25deaf5bda523ef53c574e997dae4b515be42931 available at: https://gist.github.com/3115b77cddea8e8cf8d1cbff3d1022b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.dmg

jherico:
Running this build still seems to cause 1 core to go to 100%.  Sitting in a empty or disconnected domain, if I set the affinity for the process it approaches 100% as I reduce cores to 1.  This implies it's still busy-waiting somewhere.

zzmp:
@jherico I could not reproduce this. Can you reinstall and make sure it is still happening?

hifi-gustavo:
Build console output for commit: 65e81077cb5eb2d7913a887a9e66250bffd703dd available at: https://gist.github.com/3115b77cddea8e8cf8d1cbff3d1022b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7816.dmg

jherico:
OK, I removed all the scripts and went to an empty scene and I was able to set the processor affinity down to a single CPU without the usage going to 100%.  However, an geometry in the scene seems to cause the app to stutter severly even at 2 cores.  I'd suggest that this indicates that the main thread is being starved of the ability to process anything other than paint messages, or that there is some other blocking operation going on.  I'd suggest that we let this PR through as is and address other issues as they crop up.

zzmp:
@jherico did you file a bug or write down how to repro the stutter?

PR #7817, opened 2016-05-05
fix bug preventing sim ownership release

AndrewMeadows:


AndrewMeadows:
After I removed the artificial delay from my linux machine I noticed this bug where simulation ownership was not being release when objects settled down.  The problem was that the timestamp of the pending release would exactly match the corrected timestamp of the entity-server echo.

hifi-gustavo:
Build console output for commit: 967a67256e9cf7bcf2c8b4f0acebeaf0698201e8 available at: https://gist.github.com/e7e67f8aca711e8aa129c0aca0c73d9b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7817.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7817.dmg

PR #7819, opened 2016-05-05
Perf checkpoint for testing: DNM

zzmp:
This is a combination of #7810 and #7816 for the sake of perf testing.
It is not meant to be merged.
Do not merge.


hifi-gustavo:
Build console output for commit: cacedab0e6ad967f32c158a691a7b03c384be547 available at: https://gist.github.com/1fcb62fcc319c77ba2e213314e6970b0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7819.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7819.dmg

libraries/shared/src/SharedUtil.cpp
```diff
@@ -455,19 +455,44 @@ void printVoxelCode(unsigned char* voxelCode) {
...
+        int64_t sleepFor = ((sleepTicks - now.QuadPart) * USECS_PER_SECOND) / ticksPerSec - MIN_SLEEP_USECS_BERTH;
+        if (sleepFor > MIN_SLEEP_USECS) {
+            Sleep((DWORD)(sleepFor / USECS_PER_MSEC));
+        // Yield otherwise
+        } else {
+            // Use Qt to delegate, as SwitchToThread is only supported starting with XP
+            QThread::yieldCurrentThread();
         }
     }
+}
```
Atlante45: Did you try using either `std::thread::sleep_for()` or  `QThread:usleep()`?

OmegaHeron:
I'll be running comparison tests on this over next hour or two and report back.  TY @zzmp

libraries/shared/src/SharedUtil.cpp
ZappoMan: `std::thread::sleep_for()` has about 1ms accuracy on windows... so it's not a good implementation for `usleep()`

However... we have a plan... we're going to move forward with Zach's improved implementation of `usleep()` if it does indeed improve the CPU performance... (as it appears it does)...

But we're also going to go through the limited uses of `usleep()` and convert them over to `std::thread::sleep_for()` since the 4 cases we have in the code base today, probably can live with ms precision... Those tasks are already independently reported/assigned... and we'll do those as individual PRs which can be tested and audited thoroughly for the feature they change.

OmegaHeron:
As expected this gives CPU reduction of 7810 and 7816.  Compared to 4825 at 70%+ CPU with default scripts at standard home scene this one comes in around 45-48%.  And understanding why Qt5Script seems to be most costly thread in Interface - if that can be solved for Win then it will likely be a gigantic Win4Win. :+1:

OmegaHeron:
Now - a few problems with this - It doesn't always idle down to 15 FPS when not foreground and it crashes if you minimize window then restore.  Bugsplat sent for crash.

OmegaHeron:
The minimize/restore see white unresponsive screen seems to have been a fluke.  Perhaps some instability at system level from too many torture test rounds.  Will keep an eye out for any repeats and remember I also saw it in 4825 just in case it's not a fluke.

zzmp:
@OmegaHeron I suspect the not going to 15FPS when loading may be because it is not even hitting 15FPS. Did you check the dropped frames during that time?

OmegaHeron:
Yeah - that's what it is, it was fooling me into thinking it was.  > Instrumentation showed truth.

PR #7825, opened 2016-05-05
Linux qol

sethalves:
- small changes to make interface work better on Linux


sethalves:
to test:
- make sure a web entity works (on Windows)
- if possible, try this on Linux

interface/src/main.cpp
```diff
@@ -203,7 +203,7 @@ int main(int argc, const char* argv[]) {
     Application::shutdownPlugins();
 
     qCDebug(interfaceapp, "Normal exit.");
-#ifndef DEBUG
+#if defined(DEBUG) && !defined(Q_OS_LINUX)
```
Atlante45: Didn't you mean `!defined(DEBUG)`?

hifi-gustavo:
Build console output for commit: cb523a3a715ce41f55aaabc889bf8f786e8bec9a available at: https://gist.github.com/44af5afad289191509eab337ac08e8f5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7825.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7825.dmg

libraries/entities-renderer/src/RenderableWebEntityItem.cpp
```diff
@@ -174,9 +174,14 @@ void RenderableWebEntityItem::render(RenderArgs* args) {
...
 
     if (!_webSurface) {
+        #if defined(Q_OS_LINUX)
+        // these don't seem to work on Linux
+        return;
+        #else
         if (!buildWebSurface(static_cast<EntityTreeRenderer*>(args->_renderer))) {
             return;
         }
+        #endif
```
AndrewMeadows: **cry**...

This just admits the truth of the matter: linux is no longer supported as a platform for the interface.  It is a lobotomized debug option for those who are lucky enough to not crash at login.

OmegaHeron:
Tested Ubuntu 16.04 - QT5.5.1.  GTX660M + i7 cpu.  Got playa to load about 1/2 way and thought, no - it'll overload GPU RAM and attempted to run home to no content area at my domain.  Crashed.  All subsequent launches crashed.  Not going to spend a lot of time on this now as it seems a fool's errand.  It doesn't make Interface worse or better for Linux here.

Testing on Windows shows no impact from linux patch.

interface/src/main.cpp
sethalves: yep!

hifi-gustavo:
Build console output for commit: f8f86a0d451b991846dc3796398d36799df06e1d available at: https://gist.github.com/44af5afad289191509eab337ac08e8f5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7825.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7825.dmg

libraries/entities-renderer/src/RenderableWebEntityItem.cpp
ZappoMan: btw - I don't think these work on mac either... but I might be wrong.

zzmp: Does it crash? Or just not work?

interface/src/main.cpp
```diff
@@ -203,7 +203,7 @@ int main(int argc, const char* argv[]) {
     Application::shutdownPlugins();
 
     qCDebug(interfaceapp, "Normal exit.");
-#ifndef DEBUG
+#if !defined(DEBUG) && !defined(Q_OS_LINUX)
```
zzmp: Is this just because of ASan?

libraries/entities-renderer/src/RenderableWebEntityItem.cpp
Atlante45: @ZappoMan Just tried, works on RC8. Was able to play a youtube video with no issues.

sethalves: on Linux, crash.

interface/src/main.cpp
sethalves: yes.

zzmp: Shouldn't we reenable this for all builds at some point?

ZappoMan: yes

zzmp: Not now?

ZappoMan: we need to verify that we can cleanly shutdown without crashing -- then we
can remove this.

PR #7826, opened 2016-05-05
Defer mipmap gen to gpu

zzmp:
Removes code that attempted to generate mipmaps after setting them to autogenerate.

This was generating (expensive) QImage mipmaps and doing nothing with them, as setting autogeneration makes assignStoredMip a nop.

This code should speed up scene loading, as it reduces the CPU load for generating textures. In performance tests loading 1580 standard textures (scenes from Big Buck Bunny), current build takes 105.7% of the time to load a QImage to process it into a gpu::Texture (calls within ImageReader::run to TextureUsage::process2DTextureColorFromImage over QImage::fromData). In this build, it takes only 68.8% of the time.

### DIFF
--- a/libraries/model/src/model/TextureMap.cpp
+++ b/libraries/model/src/model/TextureMap.cpp
@@ -171,14 +171,6 @@ gpu::Texture* TextureUsage::process2DTextureColorFromImage(const QImage& srcImag
 
         if (generateMips) {
             theTexture->autoGenerateMips(-1);
-            auto levels = theTexture->maxMip();
-            uvec2 size(image.width(), image.height());
-            for (uint8_t i = 1; i <= levels; ++i) {
-                size >>= 1;
-                size = glm::max(size, uvec2(1));
-                image = image.scaled(size.x, size.y, Qt::IgnoreAspectRatio, Qt::SmoothTransformation);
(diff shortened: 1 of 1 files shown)


zzmp:
This means that mips will not be generated until the textures are transferred to the thread. It is gaining CPU at the expense of "stutter" in loading when you turn.

Alternatively, we could load things immediately (costly on the CPU) and not have this "stutter".

hifi-gustavo:
Build console output for commit: 6a6d6dd20fee85f77e2ed6ccdcfed268cbd44e3d available at: https://gist.github.com/652e5ba973ca403323baa07e6075c06a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7826.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7826.dmg

OmegaHeron:
I've gone over this multiple times - I can't find a measurable or visual difference between this, current RC and potential next RC.  I expected to get something discernible from it... either it would visually look more rapid loading or I'd see something with profile traces, but, using Playa - I simply can't tell there's any difference between this and any other build.  Win 10 64bit - i5 6500 - GTX970.

PR #7830, opened 2016-05-06
switch OctreeSendThread and OctreePersistThread to use std::this_thread::sleep_for() instead of usleep()

ZappoMan:
`usleep()` on windows has the side effect that it does a busy wait if the request sleep time is under 1ms, since the `OctreeSendThread` and `OctreePersistThread` don't actually care about that level of precision, we can use the less accurate by more CPU friendly `std::this_thread::sleep_for()` to sleep our threads.

### DIFF
--- a/assignment-client/src/octree/OctreeSendThread.cpp
+++ b/assignment-client/src/octree/OctreeSendThread.cpp
@@ -9,6 +9,9 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 
+#include <chrono>
+#include <thread>
+
 #include <NodeList.h>
 #include <NumericalConstants.h>
 #include <udt/PacketHeaders.h>
@@ -101,13 +104,16 @@ bool OctreeSendThread::process() {
         int elapsed = (usecTimestampNow() - start);
         int usecToSleep =  OCTREE_SEND_INTERVAL_USECS - elapsed;
 
-        if (usecToSleep > 0) {
-            PerformanceWarning warn(false,"OctreeSendThread... usleep()",false,&_usleepTime,&_usleepCalls);
-            usleep(usecToSleep);
-        } else {
+        if (usecToSleep <= 0) {
             const int MIN_USEC_TO_SLEEP = 1;
-            usleep(MIN_USEC_TO_SLEEP);
+            usecToSleep = MIN_USEC_TO_SLEEP;
         }
+
+        {
+            PerformanceWarning warn(false,"OctreeSendThread... usleep()",false,&_usleepTime,&_usleepCalls);
+            std::this_thread::sleep_for(std::chrono::microseconds(usecToSleep));
+        }
+
     }
 
     return isStillRunning();  // keep running till they terminate us
--- a/libraries/octree/src/OctreePersistThread.cpp
+++ b/libraries/octree/src/OctreePersistThread.cpp
@@ -9,6 +9,9 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 
+#include <chrono>
+#include <thread>
+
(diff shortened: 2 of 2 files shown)


ZappoMan:
Testing notes:
- Smoke test --- You must test the sandbox... this is a server side change.
- verify that content still serves from the entity server in the sandbox
- we should see a slight improvement on the entity server CPU usage

OmegaHeron:
My low end digital ocean droplet is going to send you much love for this.  Will test ASAP

hifi-gustavo:
Build console output for commit: 99d35c37db29163013c93c058798703d80150a7f available at: https://gist.github.com/a00bc030e1a3456608b5b89b84c003c8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7830.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7830.dmg

ZappoMan:
@OmegaHeron - I changed another part of the server code to use the other sleep... so wait till this next build finishes before doing any testing

OmegaHeron:
Will do, it'll be a bit anyhow - have to run an unexpected errand.

AndrewMeadows:
After the more recent changes I stand by my opinion of **Development Approved**.

hifi-gustavo:
Build console output for commit: fe14bc8e52cc0823bb4e475467fbaedc67758251 available at: https://gist.github.com/a00bc030e1a3456608b5b89b84c003c8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7830.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7830.dmg

OmegaHeron:
I'm going to have to pass this off for other(s) to test.  I'm about to be a great-uncle and will be distracted for a few hours more.  :)  If it's still around later will give it a whirl.

zzmp:
Testers, make sure you test this on multiple platforms! It shouldn't actually change anything unless you are on Windows (and if it does, we want to know)!

assignment-client/src/octree/OctreeSendThread.cpp
```diff
@@ -101,13 +104,16 @@ bool OctreeSendThread::process() {
...
         int usecToSleep =  OCTREE_SEND_INTERVAL_USECS - elapsed;
 
-        if (usecToSleep > 0) {
-            PerformanceWarning warn(false,"OctreeSendThread... usleep()",false,&_usleepTime,&_usleepCalls);
-            usleep(usecToSleep);
-        } else {
+        if (usecToSleep <= 0) {
             const int MIN_USEC_TO_SLEEP = 1;
-            usleep(MIN_USEC_TO_SLEEP);
+            usecToSleep = MIN_USEC_TO_SLEEP;
```
zzmp: Why do we sleep here at all? Why don't we yield instead?

ZappoMan: I guess yield would be better... let me try that.

zzmp: Is there a way to test to make sure that changing to yield doesn't kill the CPU in minimum supported windows machines?

ZappoMan: let's leave it as is for now... it's ok if it sleeps for a full millisecond/

OmegaHeron:
QA Pass

Testing Windows 10 64bit - i5 CPU - GTX970 -> Stack localhost

CPU use is lower for assignment-client task running persistent script vs current RC set.  Serving ATP content is good, placing new content is good, modify content is good.  All other services work as expected.

Testing Windows 10 Client -> Stack inet Linux.  No change to CPU use seen (as expected since it's Linux).  All functions work as expected.

PR #7838, opened 2016-05-07
Use sleep_until for ScriptEngine

zzmp:


zzmp:
@ZappoMan I tested in my own build that this throttles to 60fps, but I'm still fuzzy on what it is throttling. I assumed it was any interfacing with C++, so I tried
```
var start = Date.now();
var lastPrint = start;
var times = 60;
while(times--) {
    var now = Date.now();
    var elapsed = now - start;
    var elapsedSinceLastPrint = now - lastPrint;
    print("now:" + now + " elapsed:" + elapsed);
    lastPrint = now;
    print(TextureCache.prefetch('x').state);
}
```
but that was finished in 314ms.

My next guess is that it throttles callback-based interfaces.

Do you know if that is right? Do you know what I can do to test it from JS?

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -706,22 +709,45 @@ void ScriptEngine::run() {
...
 
+    clock::time_point startTime = clock::now();
     int thisFrame = 0;
 
     auto nodeList = DependencyManager::get<NodeList>();
     auto entityScriptingInterface = DependencyManager::get<EntityScriptingInterface>();
 
     qint64 lastUpdate = usecTimestampNow();
 
+    // TODO: Integrate this with signals/slots instead of this busy wait
```
zzmp: @ZappoMan I do think this should go in as is, but 'd like to start working on removing this busy wait loop entirely, if possible, per @Atlante45's suggestion. Hopefully there is some way to trigger this behavior from Qt instead.

hifi-gustavo:
Build console output for commit: b3158a92f8753359cd15033945de3467d7cb36ee available at: https://gist.github.com/213c66f4aa5bc7c24217a288f38cc9af<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7838.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7838.dmg

zzmp:
I changed a lot of `const`s to make the intent of this sleep more clear. It was also my motivation for using `sleep_until` instead of `sleep_for` here.

hifi-gustavo:
Build console output for commit: 173d2a590c4bf265b441824754848590ffdd7480 available at: https://gist.github.com/213c66f4aa5bc7c24217a288f38cc9af<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7838.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7838.dmg

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: I don't understand your suggestion here... what is the "busy wait" you're talking about?

ZappoMan:
@zzmp - the while loop in the script engine is basically controlling the calls to "update" that it sends out... This is admittedly a pretty old design. but many scripts depend on it.

zzmp:
#### Testing

Scripts should still update at 60FPS. That means that running the following script should print the same approximate values in this PR and current build: 0.1666... (repeating 6).

```
var total = 0;
var times = 0;

Script.update.connect(function(delta) {
  total += delta;
  times++;

  if (times % 60 === 0) {
    print(total / times);
  }
});
```

assignment-client/src/Agent.cpp
```diff
@@ -386,8 +386,7 @@ void Agent::sendAvatarBillboardPacket() {
 void Agent::processAgentAvatarAndAudio(float deltaTime) {
     if (!_scriptEngine->isFinished() && _isAvatar) {
         auto scriptedAvatar = DependencyManager::get<ScriptableAvatar>();
-        const int SCRIPT_AUDIO_BUFFER_SAMPLES = floor(((SCRIPT_DATA_CALLBACK_USECS * AudioConstants::SAMPLE_RATE)
-            / (1000 * 1000)) + 0.5);
+        const int SCRIPT_AUDIO_BUFFER_SAMPLES = AudioConstants::SAMPLE_RATE / SCRIPT_FPS + 0.5;
```
ZappoMan: what's this change about?

libraries/script-engine/src/ScriptEngine.cpp
zzmp: I'm suggesting that, instead of a while loop that throttles itself to 60FPS with sleeps (not a busy wait, you're right), we trigger this from a signal when we render or simulate physics.

Probably something to put off until later, but it would:
- consolidate the logic for throttling to 60FPS (or desired framerate) to one spot.
- Sync up the ScriptEngine with render or simulate loops, which may be a good thing.

assignment-client/src/Agent.cpp
zzmp: Readability. SCRIPT_DATA_CALLBACK_USECS means nothing to me - and seems to be named so because of legacy code (you have to dig through git to find it) - so I changed it to SCRIPT_FPS. This changed the unit from microseconds to hertz, so I changed this definition to match.

ZappoMan: ok! sounds good,

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: ok - repair the comment... since it's not a busy wait... and I will approve. ;)

OmegaHeron:
First I think you meant, 0.01666 !0.1666 (1/60).

Windows 10 64 bit - i5-6500 CPU - GTX970 GPU *Desktop mode only*
Standard functional tests pass - testing playa, cellscience, forum and my own testing domain space.
Render rates, memory use (GPU/CPU) and other stats match current RC levels.

Kind of boring, right?  No.  CPU use with no scripts running is down approx 10% from last RC.  More impressive is full default script set running at Playa while spinning AV at approx 50%.  Compare that to 80-90% in previous.  Even more impressive - loading default scripts in a "easy" location with little render/physics load shows **only** 2-4% increase over no scripts.

Timing test using script referenced above showed a near constant .01666

Loading playa from fresh arrival is CPU heavy and I used it to validate ticker could hold to .01666 (approx) while seeing Interface running at 95-100% use.  It did.  I only saw case where value was < .01666 and that's at script start where it's working its way to its moving average - exactly same as seen in current RC.  Timing was maintained at .01666 while Inteface was NOT active window and while minimized.  While writing this - tick value stabilized to .0166655

Would likely be good to see additional testing, but, @zzmp and #7838 are my i5's new best friend.

hifi-gustavo:
Build console output for commit: 7db082b91567c06602b068511f9f55e6912cae50 available at: https://gist.github.com/213c66f4aa5bc7c24217a288f38cc9af<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7838.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7838.dmg

OmegaHeron:
One minor amendment.  The initial tick count with RC builds begins closer to finial steady state - while this takes a few ticks to get there.  It's outside my scope to say if that's a problem or not.

OmegaHeron:
And I run it again - no scripts running at all and see a 51% CPU use from interface where it was 25...30% before.  Madness.  It make no sense whatsoever - I repeated test 3+ times, ran current PR before posting above and never saw it at these numbers for this PR.  Back to the hard way of xperf+WinPerf - sighs.

zzmp:
@OmegaHeron current build is more exacting in terms of wait times, using busy waits for those last few-thousand microseconds. This PR avoids that, just sleeping with the windows system_clock instead. Less CPU use, less precision.

I assume that is why you saw it take longer to get to steady state. Should be fine :).

PR #7845, opened 2016-05-08
Enable script debugging

jherico:
First pass at enabling the use of QScriptEngineDebugger in Interface.  

To enable a script for debugging, you may:
* Include a comment with string `#debug` somewhere in the script
* Hold down shift while the script is being loaded or reloaded

Doing either of these will cause the debugger to be attached to the script.  

You can pause force execution to pause by including the keyword `debugger` somewhere in your script.  When not running in debug mode, this keyword will be ignored, but when in debug mode it will act as a breakpoint and pop up the debug window.  Note that if you do this in order to set breakpoints, I recommended you make sure to put the statement after any `Script.include` statements, so that all the scripts will be loaded into the debugger.

### DIFF
--- a/libraries/script-engine/src/ScriptEngine.cpp
+++ b/libraries/script-engine/src/ScriptEngine.cpp
@@ -17,12 +17,18 @@
 #include <QtCore/QFileInfo>
 #include <QtCore/QTimer>
 #include <QtCore/QThread>
+#include <QtCore/QRegularExpression>
+
+#include <QtWidgets/QMainWindow>
+#include <QtWidgets/QApplication>
+
 #include <QtNetwork/QNetworkRequest>
 #include <QtNetwork/QNetworkReply>
-#include <QtScript/QScriptEngine>
+
 #include <QtScript/QScriptValue>
 #include <QtScript/QScriptValueIterator>
-#include <QtCore/QStringList>
+
+#include <QtScriptTools/QScriptEngineDebugger>
 
 #include <AudioConstants.h>
 #include <AudioEffectOptions.h>
@@ -34,6 +40,7 @@
 #include <NodeList.h>
 #include <udt/PacketHeaders.h>
 #include <UUID.h>
+#include <ui/Menu.h>
 
 #include <controllers/ScriptingInterface.h>
 #include <AnimationObject.h>
@@ -169,6 +176,61 @@ void ScriptEngine::disconnectNonEssentialSignals() {
     }
 }
 
+void ScriptEngine::runDebuggable() {
+    init();
+    _isRunning = true;
+    _debuggable = true;
+    _debugger = new QScriptEngineDebugger(this);
+    _debugger->attachTo(this);
+    static QMenu* scriptDebugMenu = qobject_cast<QMainWindow*>(qApp->activeWindow())->menuBar()->addMenu("Script Debug");
+    scriptDebugMenu->addMenu(_fileNameString)->addMenu(_debugger->createStandardMenu(qApp->activeWindow()));
+    _debugger->action(QScriptEngineDebugger::InterruptAction)->trigger();
+    QScriptValue result = evaluate(_scriptContents, _fileNameString);
+
+    _lastUpdate = usecTimestampNow();
+    QTimer* timer = new QTimer(this);
+    connect(this, &ScriptEngine::finished, [this, timer] { 
+        disconnect(timer); 
+    });
+
+    connect(timer, &QTimer::timeout, [this, timer] {
+        if (_isFinished) {
+            if (!_isRunning) {
+                return;
+            }
+            stopAllTimers(); // make sure all our timers are stopped if the script is ending
+            if (_wantSignals) {
+                emit scriptEnding();
+                emit finished(_fileNameString, this);
+            }
+            _isRunning = false;
+            if (_wantSignals) {
+                emit runningStateChanged();
+                emit doneRunning();
+            }
+            timer->deleteLater();
+            return;
+        }
+
+        qint64 now = usecTimestampNow();
+        // we check for 'now' in the past in case people set their clock back
+        if (_lastUpdate < now) {
+            float deltaTime = (float)(now - _lastUpdate) / (float)USECS_PER_SECOND;
+            if (!_isFinished) {
+                if (_wantSignals) {
+                    emit update(deltaTime);
+                }
+            }
+        }
+        _lastUpdate = now;
+        // Debug and clear exceptions
+        hadUncaughtExceptions(*this, _fileNameString);
+    });
+
+    timer->start(10);
+}
+
+
 void ScriptEngine::runInThread() {
     Q_ASSERT_X(!_isThreaded, "ScriptEngine::runInThread()", "runInThread should not be called more than once");
 
@@ -260,6 +322,10 @@ void ScriptEngine::loadURL(const QUrl& scriptURL, bool reload) {
 // FIXME - switch this to the new model of ScriptCache callbacks
 void ScriptEngine::scriptContentsAvailable(const QUrl& url, const QString& scriptContents) {
     _scriptContents = scriptContents;
+    static const QString DEBUG_FLAG("#debug");
+    if (QRegularExpression(DEBUG_FLAG).match(scriptContents).hasMatch()) {
+        _debuggable = true;
+    }
     if (_wantSignals) {
         emit scriptLoaded(url.toString());
     }
@@ -723,7 +789,7 @@ void ScriptEngine::run() {
     auto nodeList = DependencyManager::get<NodeList>();
     auto entityScriptingInterface = DependencyManager::get<EntityScriptingInterface>();
 
-    qint64 lastUpdate = usecTimestampNow();
+    _lastUpdate = usecTimestampNow();
 
     // TODO: Integrate this with signals/slots instead of reimplementing throttling for ScriptEngine
     while (!_isFinished) {
@@ -771,15 +837,15 @@ void ScriptEngine::run() {
         qint64 now = usecTimestampNow();
 
         // we check for 'now' in the past in case people set their clock back
-        if (lastUpdate < now) {
-            float deltaTime = (float) (now - lastUpdate) / (float) USECS_PER_SECOND;
+        if (_lastUpdate < now) {
+            float deltaTime = (float) (now - _lastUpdate) / (float) USECS_PER_SECOND;
             if (!_isFinished) {
                 if (_wantSignals) {
                     emit update(deltaTime);
                 }
             }
         }
-        lastUpdate = now;
+        _lastUpdate = now;
 
         // Debug and clear exceptions
         hadUncaughtExceptions(*this, _fileNameString);
--- a/libraries/script-engine/src/ScriptEngine.cpp
+++ b/libraries/script-engine/src/ScriptEngine.cpp
@@ -177,19 +177,51 @@ void ScriptEngine::disconnectNonEssentialSignals() {
 }
 
 void ScriptEngine::runDebuggable() {
+    static QMenuBar* menuBar { nullptr };
+    static QMenu* scriptDebugMenu { nullptr };
+    static size_t scriptMenuCount { 0 };
+    if (!scriptDebugMenu) {
+        for (auto window : qApp->topLevelWidgets()) {
+            auto mainWindow = qobject_cast<QMainWindow*>(window);
+            if (mainWindow) {
+                menuBar = mainWindow->menuBar();
+                break;
+            }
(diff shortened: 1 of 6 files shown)


hifi-gustavo:
Build console output for commit: 286d1103f8f1c38efa5afd108d5962d39f15745f available at: https://gist.github.com/44475813767fbf9b34df81dde76411fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.dmg

Triplelexx:
I had a brief look at this.
Really awesome! I wasn't aware this was a feature of Qt script.
Tried it on a few scripts, set some breakpoints and tried other functionality.

Did not encounter any issues.

hifi-gustavo:
Build console output for commit: 7ef9b3ba7574fa6832e7956255ab62bc76d54273 available at: https://gist.github.com/44475813767fbf9b34df81dde76411fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.dmg

hifi-gustavo:
Build console output for commit: 18f840941f03b22ac948aaa6ad484fe871ab54b5 available at: https://gist.github.com/44475813767fbf9b34df81dde76411fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.dmg

hifi-gustavo:
Build console output for commit: 454fb1293e142a1ee7a1b308d854a7436bb19abd available at: https://gist.github.com/44475813767fbf9b34df81dde76411fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.dmg

hifi-gustavo:
Build console output for commit: 471d04734802dfa3ac091708dceae520beec270b available at: https://gist.github.com/44475813767fbf9b34df81dde76411fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.dmg

hifi-gustavo:
Build console output for commit: 6e903b1195afb0ea83c68b5ec83d081e52f71ba6 available at: https://gist.github.com/44475813767fbf9b34df81dde76411fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.dmg

hifi-gustavo:
Build console output for commit: f52c8770820d3e3091b3048c4252c6cbeebebe8c available at: https://gist.github.com/44475813767fbf9b34df81dde76411fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.dmg

hifi-gustavo:
Build console output for commit: 5182e191c47ba6847178ff9009ad404a25c28bb1 available at: https://gist.github.com/44475813767fbf9b34df81dde76411fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.dmg

Nex-Pro:
I concur with Mr. @Triplelexx  Looks good.

Nex-Pro:
@jherico Nice !

imgntn:
a nice how to http://doc.qt.io/qt-5/qtscriptdebugger-manual.html

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -166,6 +173,61 @@ void ScriptEngine::disconnectNonEssentialSignals() {
...
+    _isRunning = true;
+    _debuggable = true;
+    _debugger = new QScriptEngineDebugger(this);
+    _debugger->attachTo(this);
+    static QMenu* scriptDebugMenu = qobject_cast<QMainWindow*>(qApp->activeWindow())->menuBar()->addMenu("Script Debug");
+    scriptDebugMenu->addMenu(_fileNameString)->addMenu(_debugger->createStandardMenu(qApp->activeWindow()));
+    _debugger->action(QScriptEngineDebugger::InterruptAction)->trigger();
+    QScriptValue result = evaluate(_scriptContents, _fileNameString);
+
+    qint64* lastUpdatePtr = new qint64();
```
Atlante45: Why use a pointer?
If it was because you couldn't assign it in the lambda, you can just declare the lambda `mutable` instead.

jherico: If it was a local variable, it would be out of scope by the time the timer lambda executed, and the timer lambda needs to have it retain state between calls.  I suppose I could have made it a member, but I wanted to minimize changes and make this code match the normal run loop as closely as possible.

Atlante45: Not if you take it by copy. Then it becomes a "member" of the lambda, and if the lambda is mutable, you can modify that member.

Basically, you could just have `qint64 lastUpdate = usecTimestampNow();`, take it by copy like you do for `lastUpdatePtr` and then just get rid of `auto& lastUpdate = *lastUpdatePtr;`

hifi-gustavo:
Build console output for commit: 40316498a4c404ad012f43261b222cadbcfe46ee available at: https://gist.github.com/44475813767fbf9b34df81dde76411fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.dmg

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -169,6 +176,61 @@ void ScriptEngine::disconnectNonEssentialSignals() {
...
 }
 
+void ScriptEngine::runDebuggable() {
+    init();
+    _isRunning = true;
+    _debuggable = true;
+    _debugger = new QScriptEngineDebugger(this);
+    _debugger->attachTo(this);
+    static QMenu* scriptDebugMenu = qobject_cast<QMainWindow*>(qApp->activeWindow())->menuBar()->addMenu("Script Debug");
+    scriptDebugMenu->addMenu(_fileNameString)->addMenu(_debugger->createStandardMenu(qApp->activeWindow()));
```
Atlante45: Is that menu getting cleaned up when the script engine goes away?
I don't see anything for that.

OmegaHeron:
Windows 10 64 bit - i5 6500 CPU - GTX970 GPU - **Desktop mode - keyboard/mouse**
Reproducible crash recorded and sent via bugsplat.

Open script editor - use this simple script (from another PR where we tested timing)

`#debug`
`var total = 0;`
`var times = 0;`
`Script.update.connect(function(delta) {`
`  total += delta;`
`  times++;`
`  if (times % 60 === 0) {`
`    print(total / times);`
`  }`
`});`

Save then run then crash.

Attempting to just run it with `#debug` as its first line from Run Script dialog results in fail to load;

`[05/10 11:40:35] [DEBUG] Load script file:///C:/Users/Joe/Desktop/7838timing.js`
`[05/10 11:40:35] [DEBUG] Done downloading script at: "file:///C:/Users/Joe/Desktop/7838timing.js"`
`[05/10 11:40:35] [WARNING] [SyntaxError] Illegal character in file:///C:/Users/Joe/Desktop/7838timing.js:1(1)`

jherico:
@OmegaHeron sorry, the `#debug` declaration needs to be in a comment.   `// #debug`

OmegaHeron:
DOH FACEPALMS.  You said that, but, I didn't see it and I must have read that line 10 times.  Sorry.

OmegaHeron:
Ok - it works with properly formatted debug statement, but, some idiot proofing to prevent crashing/bad things on `#debug` alone might be in order.

OmegaHeron:
Ok - so if I set first line to //#debug and run via run script command all is well - I'm presented with debug console and can do "stuff" \o/

But - if I open in script editor and run from there with properly formatted debug directive I still crash on invoking run.

hifi-gustavo:
Build console output for commit: d4c3fb1b3baa2db55753a5030fea10f2c5086c3d available at: https://gist.github.com/44475813767fbf9b34df81dde76411fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.dmg

hifi-gustavo:
Build console output for commit: a3c49e8ee87e6dd6ffc9b356383061bbb9b59052 available at: https://gist.github.com/44475813767fbf9b34df81dde76411fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.dmg

hifi-gustavo:
Build console output for commit: b117a4b8d7d9a7bc425d30d97c31e93758ad807b available at: https://gist.github.com/44475813767fbf9b34df81dde76411fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7845.dmg

jherico:
@OmegaHeron try giving it a shot now?

OmegaHeron:
Better.

Running script from run script menu option or adding via running scripts list is good.  Using `#debug` vs correct `\\#debug` leads to script not executing and no crash :+1: .  One change that is troublesome... previously a script launched that has debug directive you got script debugger window at launch - now you have to start - find menu option and interrupt to get debug window.  If it's not a necessary evil... previous behavior was better.

From script editor using `//#debug launches script and same as above needs manual interrupt to present debugger console.

Since it's an advanced tool - I won't make a huge deal of this, but, if a script is in interrupted state you can't terminate it from running scripts list - it does terminate once you resume operation.

I did notice with simple timing script that it seemed to be pretty laggy on tick response - likely due to having to pause - get debugger window - resume, but, this could be ugly for debugging anything that's timing critical.

General CPU/GPU/Mem stats tracing shows only minuscule impact while debugger is running over baseline numbers. :+1:  All other scripts run normally - no issues seen with any other functions - seems a useful + benign addition.

jherico:
@OmegaHeron  

> previous behavior was better.

I think it depends on the user and the situation.  If I'm running default scripts and I have shift held down when I launch the application, I don't want 10 debuggers popping up.  Also, this is only the first pass at the debugger, so the mechanism for getting a script into debug mode is pretty raw.  Ideally, the various UI mechanisms that let you launch scripts would have checkboxes that say whether you want a script in debug mode and possibly whether you want to pause at startup, similar to most other debugging systems.  Failing that, I'll look into doing something like allowing you to hold down `Ctrl` as well as `Shift` when launching a script so that you can have it optionally pause at start.  

Also bear in mind that you can insert the `debugger` javascript keyword anwhere in the script to force the it to pause and the debugger window to appear.  Personally, since most of my scripts end up using a lot of `Script.include` statements, I'd rather be able to use the `debugger` keyword to cause the script to pause after those have executed, so that I can put breakpoints in the included source as well as the main script.  



> I did notice with simple timing script that it seemed to be pretty laggy on tick response 

When not in debugging mode, the script thread runs in a pretty tight loop and the update callback will be called pretty reliably.  When in debugging mode, the update callback is connected to a Qt timer that will only trigger a maximum of 100 times a second, and the timer events are competing with everything else in the main thread, so yes, it may be less responsive.  Not really much I can do about that as I've found no way to attach the debugger to a script that isn't running on the main thread (trust me, I've tried).

OmegaHeron:
Fair enough and makes sense.  TY for taking time to implement and explain - use case > simple makes it a lot more logical now.

OmegaHeron:
Also - having ran standard test batter, CPU time traces and other bits... QA Pass for me.

PR #7854, opened 2016-05-10
Clean up resource cache's encapsulation of asynchronous loading

zzmp:
`delayLoad` did not do what it advertised, so this revises the logic of asynchronous loads to be self-contained within ResourceCache - instead of requiring polling from Application - and removes the `delayLoad` parameter from caches and resources.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -584,12 +584,6 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
     // put the NodeList and datagram processing on the node thread
     nodeList->moveToThread(nodeThread);
 
-    // Model background downloads need to happen on the Datagram Processor Thread.  The idle loop will
-    // emit checkBackgroundDownloads to cause the ModelCache to check it's queue for requested background
-    // downloads.
-    auto modelCache = DependencyManager::get<ModelCache>();
-    connect(this, &Application::checkBackgroundDownloads, modelCache.data(), &ModelCache::checkAsynchronousGets);
-
     // put the audio processing on a separate thread
     QThread* audioThread = new QThread();
     audioThread->setObjectName("Audio Thread");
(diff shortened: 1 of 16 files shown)


zzmp:
#### Testing

Smoke test, focusing on model loading (it should still work).

hifi-gustavo:
Build console output for commit: 09d879e19f4140ac8899f02b45e37c0ca02c03d8 available at: https://gist.github.com/277a5332dbdc72b91d169cef4f146f50<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7854.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7854.dmg

judasshuffle:
Had a quick none scientific smoke test. That is I was grumbling about slow loading times from cached items then noticed this pr, so tried it. Things were taking around 10 seconds to load in on this pr I counted around 8 seconds. Nb this is me going 1 Mississippi 2 Mississippi  not actually timing it but it does seem faster, so yay

OmegaHeron:
QA Pass
Windows 10 64 bit - i5 6500 CPU - GTX970 GPU - **Desktop Mode - Keyboard/mouse**
All standard tests pass.  CPU/GPU/Memory use shows improvement over current RC (CPU Use).

Loading Playa with this PR is **an order of magnitude** smoother while running playaPerformanceTest.js
Peak CPU use with this PR loading Playa flattened out at around 90% for a (comparatively) short time - while current RC pegs it at 100% for a long time and exhibits stuttering/frozen movement while test script attempts to rotate AV.  Similar results seen at Cellscience and Forum.

zzmp:
That's awesome, considering that perf was not my intention here at all.
I was just trying to clean things up with the code structure.

Atlante45:
Yeah, we're gonna have to solve the ref drop on async load eventually but that a good clean up for now.

zzmp:
@Atlante45 is the ref drop actually a problem anywhere?

Atlante45:
It is for sounds, which is why I had to do some trickery for collision sounds to be loaded from the main thread.
There's also the fact that we don't and can't pass int the `extra` pointer, because it usually points to the stack.

zzmp:
We could probably memcpy and then delete it. We could also set up a shell QSharedPointer<Resource> that resets itself on a signal when the true resource is loaded. We'd have to think about thread-safety, though.

PR #7855, opened 2016-05-10
Updated default idle and turn animations.

hyperlogic:
* idle is less masculine and works better on prescilla & art3mis avatars.
* idle to walk transition is updated to work with this new idle.
* turn in place animations are improved.
* fly animation is more symmetric.


hifi-gustavo:
Build console output for commit: 0f1ff87f783f938299541dfac150a2471d140a69 available at: https://gist.github.com/c4ed429e9e1587c7585ffad1765ce41b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7855.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7855.dmg

ZappoMan:
@hyperlogic - this is great... but isn't our plan to allow the FST file to specify animation URLs for the avatars that are custom for each avatar? It seems like finding "the average" animation for both male and female avatars is not ever going to make people happy...

hyperlogic:
Yep, we will indeed add support for per Avatar animation overrides. But in the short term, this just a small improvement to the default set we are already using

OmegaHeron:
QA Pass
Tested Windows 10 64 bit - i5 6500 CPU - GTX970GPU **Desktop mode - keyboard/mouse**
General function test sets all pass.  No negative change seen to performance compare to current RC.

Doing comparison between this and current RC it's clear it is loading slightly different animations for the changed items.  I don't see anything too radical about changes - better, worse?  Neither - it's subtle enough that it's probably a little better looking, but not so much as to be a **wow**.

PR #7862, opened 2016-05-10
Use move ctor for vec3 to avoid rare crash

zzmp:


zzmp:
See bugzid:481.

zzmp:
@ZappoMan how do I make the fogbugz integration work?

zzmp:
test this please

libraries/fbx/src/FBXReader.cpp
```diff
@@ -1587,7 +1587,8 @@ FBXGeometry* FBXReader::extractFBXGeometry(const QVariantHash& mapping, const QS
             ShapeVertices& points = shapeVertices[jointIndex];
             foreach (const glm::vec3& vertex, extracted.mesh.vertices) {
                 const glm::mat4 vertexTransform = meshToJoint * glm::translate(vertex);
-                points.push_back(extractTranslation(vertexTransform) * clusterScale);
+                // Using the copy ctor crashed here, see bugsplat 3364
+                points.push_back(std::move(extractTranslation(vertexTransform) * clusterScale));
```
Atlante45: Are you sure about this change? 
`extractTranslation(vertexTransform) * clusterScale` is already a prvalue, so using `std::move` will not change anything except preventing copy elision in some cases.

In that case since `push_back` already takes rvalue refs, I would even expect the assembly to be the same.

ZappoMan:
@zzmp - I haven't figured it out... I think it only works if you mention fogbugz in your git commit message... :(

hifi-gustavo:
Build console output for commit: 59ae870d3f939494fc8c7012535b9e73ba21ba5f available at: https://gist.github.com/b91afb1e2661bfb206d92aa9c728f91a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7862.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7862.dmg

libraries/fbx/src/FBXReader.cpp
zzmp: According to the stack trace, it was using the copy-ctor override for push_back.

Atlante45: According to the Bugsplat trace, it crashes in `interface!std::vector<glm::detail::tvec3<float,0>,std::allocator<glm::detail::tvec3<float,0> > >::_Reallocate`
First I think having a `reserve` call would help, but it might be that the array is just getting too big and hits the array size limit.

PR #7864, opened 2016-05-11
Fix for vive controllers sometimes not working

hyperlogic:
* Fixed bug with input devices that where added, removed then re-added.
  The default mappings were being ignored on the second add.
* Fixed potential crash when hardware inputPlugin device poses were polled from the JavaScript thread
  by taking the UserInputManager lock during pluginUpdate.
* Renamed Controller.Hardware.Vive.LB & RB to LeftGrip and RightGrip, to better match Oculus touch.
* Updated resource/controller/vive.json to reflect this new mapping.
* Exposed touch pad capacitive touch events to JavaScript as
  Controller.Hardware.Vive.LSTouch and RSTouch.
* Added viveTouchpadTest.js script to test LSTouch and RSTouch events.

### DIFF
--- a/interface/resources/controllers/vive.json
+++ b/interface/resources/controllers/vive.json
@@ -5,14 +5,14 @@
         { "from": "Vive.LX",  "when": "Vive.LS", "to": "Standard.LX" },
 
         { "from": "Vive.LT", "to": "Standard.LT" },
-        { "from": "Vive.LB", "to": "Standard.LB" },
+        { "from": "Vive.LeftGrip", "to": "Standard.LB" },
         { "from": "Vive.LS", "to": "Standard.LS" },
 
         { "from": "Vive.RY",  "when": "Vive.RS", "filters": "invert", "to": "Standard.RY" },
         { "from": "Vive.RX",  "when": "Vive.RS", "to": "Standard.RX" },
 
         { "from": "Vive.RT", "to": "Standard.RT" },
-        { "from": "Vive.RB", "to": "Standard.RB" },
+        { "from": "Vive.RightGrip", "to": "Standard.RB" },
         { "from": "Vive.RS", "to": "Standard.RS" },
 
         { "from": "Vive.LeftApplicationMenu", "to": "Standard.Back" },
--- a/interface/resources/controllers/vive.json
+++ b/interface/resources/controllers/vive.json
@@ -7,6 +7,7 @@
         { "from": "Vive.LT", "to": "Standard.LT" },
         { "from": "Vive.LeftGrip", "to": "Standard.LB" },
         { "from": "Vive.LS", "to": "Standard.LS" },
+        { "from": "Vive.LSTouch", "to": "Standard.LSTouch" },
 
         { "from": "Vive.RY",  "when": "Vive.RS", "filters": "invert", "to": "Standard.RY" },
         { "from": "Vive.RX",  "when": "Vive.RS", "to": "Standard.RX" },
@@ -14,6 +15,7 @@
         { "from": "Vive.RT", "to": "Standard.RT" },
         { "from": "Vive.RightGrip", "to": "Standard.RB" },
         { "from": "Vive.RS", "to": "Standard.RS" },
+        { "from": "Vive.RSTouch", "to": "Standard.RSTouch" },
 
         { "from": "Vive.LeftApplicationMenu", "to": "Standard.Back" },
         { "from": "Vive.RightApplicationMenu", "to": "Standard.Start" },
--- a/scripts/developer/tests/viveTouchpadTest.js
+++ b/scripts/developer/tests/viveTouchpadTest.js
@@ -46,25 +46,29 @@ function shutdown() {
 }
 Script.scriptEnding.connect(shutdown);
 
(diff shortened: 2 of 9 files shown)


hyperlogic:
https://highfidelity.fogbugz.com/f/cases/471/Vive-controllers-sometimes-do-not-work

hyperlogic:
https://highfidelity.fogbugz.com/f/cases/471/Vive-controllers-sometimes-do-not-work

hifi-gustavo:
Build console output for commit: 09a4e0eaa88b668d4fc76f4e71a1fb4b082c3162 available at: https://gist.github.com/d3b1c1cc10cf5866238ea6a89ff3862f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7864.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7864.dmg

jherico:
I'm not sure if this approach won't have undesirable side effects.  The mapping system is designed as a stack, so that if I create a JS file that does custom mappings that override the default, that mapping takes precedence.  However, from my initial reading of this PR, it looks like if I had such a script and then I disconnected and reconnected the Vive controllers, the device default mappings would be loaded and then supersede the script mapping.  I'm not positive if that's an issue though, since the same thing might happen if you disable and re-enable a given plugin.

interface/resources/controllers/vive.json
```diff
@@ -5,14 +5,14 @@
         { "from": "Vive.LX",  "when": "Vive.LS", "to": "Standard.LX" },
 
         { "from": "Vive.LT", "to": "Standard.LT" },
-        { "from": "Vive.LB", "to": "Standard.LB" },
+        { "from": "Vive.LeftGrip", "to": "Standard.LB" },
```
jherico: You should add the mappings for Vive -> Standard for the left and right stick touch....

howard-stearns:
Cool, @hyperlogic. Does this fix https://app.asana.com/0/search/119496169305857/118442457384587 bugzid:346

ZappoMan:
One of the things we need to do in the near future is eliminate the idea of "enabled" and "disabled" devices (at least from the end user's perspective)... Making users find the "Input Devices" and enable/disable them is simply wrong and confusing... it adds steps and makes the product hard to use. 

In the case of all of our "non-pose" devices: xbox, mouse, keyboard, joysticks/buttons on the other controllers... the end result of the filter graphs can always be "blended togther" if you hold down the left arrow key while simultaneously pushing the right joystick on the xbox to the right you get the sum of these two movements commands.

We need to get to this same place for pose data as well.

The point I'm making is what should happen from the end users perspective is that essentially ALL devices are always active and all filter graphs using them should be running at all times. If the device is not present then it's as if all the inputs are "zeros" or "quiet"... the moment the device is actually present, it begins presenting data.

I am not suggesting we shouldn't merge this fix.. but I want us to recognize that what I've described above is ultimately the "correct" implementation of the feature. We should work toward getting to that "correct" implementation.

hyperlogic:
@jherico The only simple way I see to work around this would be to never remove the device default mapping when the device/plugin is disabled.

hyperlogic:
@howard-stearns: yeah probably

hyperlogic:
@jherico Another option would be to split the device default mapping list from user mappings from script.  Then always process user mappings first.

hifi-gustavo:
Build console output for commit: c91455a291476ead2f85d436d2bfb68ac24eec47 available at: https://gist.github.com/d3b1c1cc10cf5866238ea6a89ff3862f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7864.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7864.dmg

scripts/developer/tests/viveTouchpadTest.js
```diff
@@ -0,0 +1,75 @@
...
+        lifetime: 60000
+    });
+}
+
+function shutdown() {
+    Entities.deleteEntity(boxEntity);
+}
+Script.scriptEnding.connect(shutdown);
+
+prevPose = Controller.getPoseValue(Controller.Hardware.Vive.LeftHand);
```
imgntn: this will throw an error on non vive systems because Controller.Hardware.Vive is undefined so the .LeftHand doesn't exist.  since this script is a vive test... probably doesn't matter.  but good to know for supporting multiple controllers to test for existence first

hifi-gustavo:
Build console output for commit: 95ef543f257d6f8ccb525b25b0621cad35418bd2 available at: https://gist.github.com/d3b1c1cc10cf5866238ea6a89ff3862f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7864.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7864.dmg

cozza13:
QA Pass. Used the vive controller and disconnected it in a way that would make it not work in stable. it connected when I turned it back on

PR #7867, opened 2016-05-11
Fix crash in JurisdictionMap

huffman:
This is an attempt to fix a crash in JurisdictionMap. The _rootOctalCode used a bare pointer, allowing access across multiple threads in an unsafe way. Changes:

 * Make _rootOctalCode and _endNodes use std::shared_ptr so that their lifetime can be properly extended when they are being accessed externally.
  * Add guards around access to _rootOctalCode and _endNodes
  * Add atomic `getRootAndEndNodeOctalCodes` to avoid race conditions when getting both values.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -3656,11 +3656,11 @@ void Application::queryOctree(NodeType_t serverType, PacketType packetType, Node
             } else {
                 const JurisdictionMap& map = (jurisdictions)[nodeUUID];
 
-                unsigned char* rootCode = map.getRootOctalCode();
+                auto rootCode = map.getRootOctalCode();
 
                 if (rootCode) {
                     VoxelPositionSize rootDetails;
-                    voxelDetailsForCode(rootCode, rootDetails);
+                    voxelDetailsForCode(rootCode.get(), rootDetails);
                     AACube serverBounds(glm::vec3(rootDetails.x * TREE_SCALE,
                                                   rootDetails.y * TREE_SCALE,
                                                   rootDetails.z * TREE_SCALE) - glm::vec3(HALF_TREE_SCALE),
@@ -3720,11 +3720,11 @@ void Application::queryOctree(NodeType_t serverType, PacketType packetType, Node
             } else {
                 const JurisdictionMap& map = (jurisdictions)[nodeUUID];
 
-                unsigned char* rootCode = map.getRootOctalCode();
+                auto rootCode = map.getRootOctalCode();
 
                 if (rootCode) {
                     VoxelPositionSize rootDetails;
-                    voxelDetailsForCode(rootCode, rootDetails);
+                    voxelDetailsForCode(rootCode.get(), rootDetails);
                     AACube serverBounds(glm::vec3(rootDetails.x * TREE_SCALE,
                                                   rootDetails.y * TREE_SCALE,
                                                   rootDetails.z * TREE_SCALE) - glm::vec3(HALF_TREE_SCALE),
@@ -4251,9 +4251,9 @@ void Application::nodeKilled(SharedNodePointer node) {
                 return;
             }
 
-            unsigned char* rootCode = _entityServerJurisdictions[nodeUUID].getRootOctalCode();
+            auto rootCode = _entityServerJurisdictions[nodeUUID].getRootOctalCode();
             VoxelPositionSize rootDetails;
-            voxelDetailsForCode(rootCode, rootDetails);
+            voxelDetailsForCode(rootCode.get(), rootDetails);
 
             qCDebug(interfaceapp, "model server going away...... v[%f, %f, %f, %f]",
                 (double)rootDetails.x, (double)rootDetails.y, (double)rootDetails.z, (double)rootDetails.s);
@@ -4378,7 +4378,7 @@ int Application::processOctreeStats(ReceivedMessage& message, SharedNodePointer 
             }
 
             VoxelPositionSize rootDetails;
-            voxelDetailsForCode(octreeStats.getJurisdictionRoot(), rootDetails);
+            voxelDetailsForCode(octreeStats.getJurisdictionRoot().get(), rootDetails);
 
             qCDebug(interfaceapp, "stats from new %s server... [%f, %f, %f, %f]",
                 qPrintable(serverType),
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -4372,8 +4372,11 @@ int Application::processOctreeStats(ReceivedMessage& message, SharedNodePointer 
             serverType = "Entity";
         }
 
+        bool found = false;
+
         jurisdiction->withReadLock([&] {
             if (jurisdiction->find(nodeUUID) != jurisdiction->end()) {
+                found = true;
                 return;
             }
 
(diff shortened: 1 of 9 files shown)


huffman:
No need to test yet, I want to test it a little more

hifi-gustavo:
Build console output for commit: 8b346bd9c646bcca951884b52231b8430654f9cf available at: https://gist.github.com/c67b01cdfbd61c8d5cff794edb9a35b8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.dmg

hifi-gustavo:
Build console output for commit: 331e22469086b3a1e03bfef9da9b17a4f16682ba available at: https://gist.github.com/c67b01cdfbd61c8d5cff794edb9a35b8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.dmg

hifi-gustavo:
Build console output for commit: 7d4dfd2d4188856ddad6c10468a3b86288e134fc available at: https://gist.github.com/c67b01cdfbd61c8d5cff794edb9a35b8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.dmg

hifi-gustavo:
Build console output for commit: b65d2b09926148a38f2c727471d93efba35ede8c available at: https://gist.github.com/c67b01cdfbd61c8d5cff794edb9a35b8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.dmg

huffman:
Rebased to include fps-fixes

hifi-gustavo:
Build console output for commit: 8c41e617ca5779270c3d80d19f5ff00c0f3654af available at: https://gist.github.com/c67b01cdfbd61c8d5cff794edb9a35b8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.dmg

libraries/octree/src/JurisdictionMap.cpp
```diff
@@ -17,11 +17,39 @@
 #include <DependencyManager.h>
 #include <NodeList.h>
 #include <udt/PacketHeaders.h>
-#include <OctalCode.h>
 
 #include "OctreeLogging.h"
 #include "JurisdictionMap.h"
 
+void myDebugOutputBits(unsigned char byte, bool withNewLine) {
```
ZappoMan: shouldn't these functions be in a shared library somewhere?

libraries/octree/src/JurisdictionMap.cpp
```diff
@@ -17,11 +17,39 @@
...
+        printf("%d", byte >> (7 - i) & 1);
+    }
+    printf(" ] ");
+
+    if (withNewLine) {
+        printf("\n");
+    }
+}
+
+void myDebugPrintOctalCode(const unsigned char* octalCode, bool withNewLine) {
```
ZappoMan: shouldn't these functions be in a shared library somewhere?

libraries/octree/src/JurisdictionMap.cpp
huffman: I can move them to a shared library, but I didn't create them, I had just moved them from another place in the file.

ZappoMan: I see that. I wonder who the jerk that created them and didn't put them in a shared library was! That guy sucks!

ZappoMan:
Suggested things for the test plan:

- Jurisdiction maps are used between any octree server (only the entity server at this point) and the client to determine which servers are handling which "shard" of the octree data... so these changes will impact the client/server conversation for entities.
- This changes should not change the protocol, so you should include cross version testing in your test plan. Do old clients talk to new servers, do new clients talk to old servers.
- This PR changes the octree scene stats code (since that touched jurisdictions) so make sure that those are still working as expected... 
- OctreeSceneStats can side effect the "start" of physics, make sure you test teleporting to a new domain and that physics starts as expected.
- Jurisdictions change when you change domains... make sure you test rapidly changing domains.

huffman:
Rebased, waiting for build

hifi-gustavo:
Build console output for commit: 013d902376b332df07a70b4aaee9642ee9d03498 available at: https://gist.github.com/c67b01cdfbd61c8d5cff794edb9a35b8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.dmg

huffman:
Test plan for jurisdiction changes

  * Test that communication between client and entity server still works:
    * Create and edit entities in different areas of your domain
    * Move between those areas and make sure you can see everything correctly
    * Restart your client and make sure you can still see everything correctly
    * Make sure that physics starts correctly when going to a new domain.
      * Go to a domain with a known floor (playa, home)
      * Make sure physics starts working after assets have loaded - with avatar collisions enabled you should fall to the ground when this happens.
      * Do this for at least 2 domains in the same session (ie don't close the application before going to a new domain)
    * Repeat this test for Old Client -> PR Server and PR Client -> Old Server. (You can use the current release for the old client/server)
  * Make sure octree scene stats is working correctly
    * Enable the developer menu, open Developer -> Entities -> Entity Statistics.
    * Ensure that the jurisdiction is shown under `Entity Server #`, it will be `active jurisdiction: 00 [0, 0, 0: 1]`
    * Ensure that other stats looks reasonable (Note: Entity Updates has a bug where it jumps to a high number when switching domains. That is OK)

hifi-gustavo:
Build console output for commit: 68665e6b2579fa63e9cf0af6ce293af78afdf1dd available at: https://gist.github.com/c67b01cdfbd61c8d5cff794edb9a35b8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7867.dmg

interface/src/Application.cpp
```diff
@@ -4372,27 +4372,33 @@ int Application::processOctreeStats(ReceivedMessage& message, SharedNodePointer
...
+            // store jurisdiction details for later use
+            // This is bit of fiddling is because JurisdictionMap assumes it is the owner of the values used to construct it
+            // but OctreeSceneStats thinks it's just returning a reference to its contents. So we need to make a copy of the
+            // details from the OctreeSceneStats to construct the JurisdictionMap
+            JurisdictionMap jurisdictionMap;
+            jurisdictionMap.copyContents(octreeStats.getJurisdictionRoot(), octreeStats.getJurisdictionEndNodes());
+            jurisdiction->withWriteLock([&] {
+                (*jurisdiction)[nodeUUID] = jurisdictionMap;
+            });
+        }
```
AndrewMeadows: `found` is initialized to `false` and the only context that sets it `true` returns immediately so... in this context at line 4391 it is guaranteed to be `false`.  In fact, the variable itself is unnecessary.

ZappoMan: pretty sure that return is from the lambda

OmegaHeron:
QA Pass
Starting at playa with a clean install (no cache of playa content
Test that communication between client and entity server still works - Yes, all sub-tests under work.
Standard tests pass.

interface/src/Application.cpp
huffman: Yes, that returns from the lambda. My understanding was that the intent was to store the JurisdictionMap if there wasn't an entry in the NodeToJurisdictionMap, otherwise return early. The existing code was setting the JurisdictionMap every time, even an entry already existed.

PR #7874, opened 2016-05-12
allow MyAvatar.addThrust() to raise avatar from standing on ground

AndrewMeadows:


ZappoMan:
How does this change interact with the existing `Controller.Actions` infrastructure? I see "double jump button" in the code... but there isn't really such a thing as a "jump button"... instead there is `Actions.Up` which can be driven by keyboard buttons, joysticks, scripts, or other input devices.

AndrewMeadows:
Someone noticed that MyAvatar.addThrust() would do nothing when the avatar was standing on something, but would work when flyting around.  This PR fixes the problem.

To test run the following script which creates some alternate movement keys at **i**, **j**, **k**, **l**, **m**, **o**.  Press the **o** key to move up and **m** to move down.

'''
// Example script for pushing the avatar around using MyAvatar.addThrust()

// each key corresponds to a direction (world frame!)
var directions = {
    "i" : { x: 0, y: 0, z: 1 },
    "j" : { x: 1, y: 0, z: 0 },
    "k" : { x: 0, y: 0, z: -1 },
    "l" : { x: -1, y: 0, z: 0 },
    "o" : { x: 0, y: 1, z: 0 },
    "m" : { x: 0, y: -1, z: 0 }
};
var states = {
    "i" : false,
    "j" : false,
    "k" : false,
    "l" : false,
    "o" : false,
    "m" : false
};

// the THRUST_AMPLITUDE is a "magic number".  You just have to make it bigger/smaller until it "works"
var THRUST_AMPLITUDE = 1000.0;
var thrust = { x: 0, y: 0, z: 0 };

function updateMotor() {
    var numKeysPressed = 0;
    thrust = { x: 0, y: 0, z: 0 };
    for (var key in directions) {
        if (states[key]) {
            var dir = directions[key];
            thrust = { x: thrust.x + dir.x, y: thrust.y + dir.y, z: thrust.z + dir.z };
            numKeysPressed++;
        }
    }
    if (numKeysPressed > 0) {
        var thrustLength = Math.sqrt(thrust.x * thrust.x + thrust.y * thrust.y + thrust.z * thrust.z);
        if (thrustLength > 0) {
            var scaleFactor = THRUST_AMPLITUDE / thrustLength;
            thrust = { x: scaleFactor * thrust.x, y: scaleFactor * thrust.y, z: scaleFactor * thrust.z};
        }
    }
}

function applyThrust(dt) {
    var scaledThrust = { x: dt * thrust.x, y: dt * thrust.y, z: dt * thrust.z };
    MyAvatar.addThrust(scaledThrust); // NOTE: thrust is always in the world-frame
}

function keyPressEvent(event) {
    if (!event.isAutoRepeat) {
        if (event.text in states) {
            states[event.text] = true;
            updateMotor();
        }
    }
}

function keyReleaseEvent(event) {
    if (!event.isAutoRepeat) {
        if (event.text in states) {
            states[event.text] = false;
            updateMotor();
        }
    }
}

// connect callbacks
Controller.keyPressEvent.connect(keyPressEvent);
Controller.keyReleaseEvent.connect(keyReleaseEvent);
Script.update.connect(applyThrust);
'''

hifi-gustavo:
Build console output for commit: 8d5620ad653a1bc4bd0fc0066c4b45d3492d66ab available at: https://gist.github.com/c86d6d8adb50d7c59eeab0dccef515f2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7874.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7874.dmg

libraries/physics/src/CharacterController.cpp
```diff
@@ -173,54 +173,50 @@ void CharacterController::preStep(btCollisionWorld* collisionWorld) {
...
+        btScalar tau = dt / timescale;
+        finalVelocity = tau * desiredHorizVelocity + (1.0f - tau) * actualHorizVelocity;
+
+        // only blend vertical part if the target velocity has non-zero vertical component
+        if (desiredVertVelocity.length2() > MIN_TARGET_SPEED_SQUARED) {
+            tau = dt / FLY_ACCELERATION_TIMESCALE;
+            finalVelocity += tau * desiredVertVelocity + (1.0f - tau) * actualVertVelocity;
+        } else {
+            finalVelocity += actualVertVelocity;
         }
```
AndrewMeadows: These changes apply vertical velocity preferences when standing/falling... when they are non-zero.  If the target velocities are zero then there is no slaving and the avatar is allowed to fall under gravity without interference.

libraries/physics/src/CharacterController.cpp
```diff
@@ -513,10 +509,17 @@ void CharacterController::preSimulation() {
...
+                btVector3 desiredVelocity = _targetVelocity;
+                if (desiredVelocity.length2() < MIN_TARGET_SPEED_SQUARED) {
+                    desiredVelocity = btVector3(0.0f, 0.0f, 0.0f);
+                }
+                bool vertTargetSpeedIsNonZero = desiredVelocity.dot(_currentUp) > MIN_TARGET_SPEED;
+                if ((jumpButtonHeld || vertTargetSpeedIsNonZero) && (_takeoffJumpButtonID != _jumpButtonDownCount)) {
+                    SET_STATE(State::Hover, "double jump button");
+                } else if ((jumpButtonHeld || vertTargetSpeedIsNonZero) && (now - _jumpButtonDownStartTime) > JUMP_TO_HOVER_PERIOD) {
+                    SET_STATE(State::Hover, "jump button held");
+                }
```
AndrewMeadows: These changes help the avatar transition into the "InAir" state when above a floor but not pushing up/down.

AndrewMeadows:
@ZappoMan.  There should be no new or adverse interaction with the `Controller.Actions` infrustructure.  My understanding is that those are masquerading as key presses which get routed to the normal `MyAvatar::applyKeyboardMotor()` logic, which I designed to "play well" with the "thrust" and "scriptedMotor" features.

However, I should note that the "scriptedMotor" and "thrust" behaviors are applied AFTER the keyboard motor (this has been the case for more than a year).  The order of operation is as follows:

(1) keyboard motor computes a "targetVelocity" for MyAvatar
(2) scripted motor (if it active) modifies the "targetVelocity"
(3) thrust (if any) is added to the "targetVelocity"
(4) the CharacterController class takes the "targetVelocity" and "makes it so" in the Bullet physics simulation
(5) at the end of the physics simulation we harvest avatar position changes

Note: how successful the scripted motor will be at defeating the keyboard motor at step (2) depends on the scripted motor's "timescale": short timescales will tend to "win", long timescales will "lose".  The "thrust" feature has no timescale... it is just added on top at step (3).

ZappoMan:
@AndrewMeadows - I haven't looked at the code to confirm... but if it's true that there's a method called `MyAvatar::applyKeyboardMotor()` we need to rename that function... because Application and MyAvatar shouldn't know anything about the keyboard... they should instead be responding to the actions. Keyboard goes through the controller input filter graph, finishes at actions, and actions can drive the avatar... but you can also disable the keyboard and drive through the xbox or others.

AndrewMeadows:
I am unfamiliar with the "controller input" and "actions" pipeline.  I'll will audit the code and try to understand how it all works.

AndrewMeadows:
I've decided to do some cleanup and add it to this PR.  Not finished yet.

AndrewMeadows:
@ZappoMan, I'm going to leave the "KeyboardMotor" name convention.  It translates the state of `MyAvatar::_driveKeys` into an intended motor velocity.  It appears the "Controller actions" get translated into `_driveKeys` state changes.  In any case, the "KeyboardMotor" stuff is not exposed to JS.

ZappoMan:
Would you consider calling it "ActionMotor" and "_driveActions"?

AndrewMeadows:
Sure, "keyboard" --> "action" is fine with me, it it helps.

hifi-gustavo:
Build console output for commit: e4cb6e2b87a639bcdae4662bf406002222aeab55 available at: https://gist.github.com/c86d6d8adb50d7c59eeab0dccef515f2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7874.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7874.dmg

AndrewMeadows:
Here is another simple test script that uses the "scripted motor" feature: you can set the avatar's velocity (its reference frame) and its timescale (time to achieve its motion) directly via script.  This motor is independent of the one that is driven by pressing keys or harvesting controller actions and translating them into key presses:

```
//
//  simpleAvatarMotor.js
//
//  Created by AndrewMeadows, 2016.04.11
//
//
// Example script for pushing the avatar around using the "scripted" motor.

// each key corresponds to a direction

var directions = {
    "i" : { x: 0, y: 0, z: 1 },
    "j" : { x: 1, y: 0, z: 0 },
    "k" : { x: 0, y: 0, z: -1 },
    "l" : { x: -1, y: 0, z: 0 },
    "o" : { x: 0, y: 1, z: 0 },
    "m" : { x: 0, y: -1, z: 0 }
};

var states = {
    "i" : false,
    "j" : false,
    "k" : false,
    "l" : false,
    "o" : false,
    "m" : false
};

var MOTOR_SPEED = 5.0;

function disableMotor() {
    // we disable the motor by giving it a very long timescale otherwise the motor will try
    // to effect 'zero velocity' (it will 'brake' and fight normal keyboard movement).
    // when turning off: be sure to set timescale BEFORE velocity.
    MyAvatar.motorTimescale = 1000000;
    MyAvatar.motorVelocity = {x: 0, y: 0, z: 0 };
}

function updateMotor() {
    var numKeys = 0;
    var totalDir = { x: 0, y: 0, z: 0 };
    for (var key in directions) {
        if (states[key]) {
            var dir = directions[key];
            totalDir = {
                x: totalDir.x + dir.x,
                y: totalDir.y + dir.y,
                z: totalDir.z + dir.z };
            numKeys++;
        }
    }
    if (numKeys > 0) {
        var dirLength = Math.sqrt(totalDir.x * totalDir.x + totalDir.y * totalDir.y + totalDir.z * totalDir.z);
        var velocity = { x: 0, y: 0, z: 0 };
        if (dirLength > 0) {
            velocity = {
                x: MOTOR_SPEED * totalDir.x / dirLength,
                y: MOTOR_SPEED * totalDir.y / dirLength,
                z: MOTOR_SPEED * totalDir.z / dirLength };
        }
        // set the referenceFrame in which the motorVelocity is defined
        MyAvatar.motorReferenceFrame = "avatar"; // possible values are: 'avatar', 'camera', and 'world'

        // when turning on: be sure to change velocity BEFORE the timescale
        MyAvatar.motorVelocity = velocity;
        MyAvatar.motorTimescale = 0.7;
    } else {
        // no keys pressed, so we disable the motor
        disableMotor();
    }
}

function keyPressEvent(event) {
    if (!event.isAutoRepeat) {
        var key = event.text;
        if (key in states) {
            states[key] = true;
            updateMotor();
        }
    }
}

function keyReleaseEvent(event) {
    if (!event.isAutoRepeat) {
        var key = event.text;
        if (key in states) {
            states[key] = false;
            updateMotor();
        }
    }
}

// connect callbacks
Controller.keyPressEvent.connect(keyPressEvent);
Controller.keyReleaseEvent.connect(keyReleaseEvent);
Script.scriptEnding.connect(disableMotor);
```

hifi-gustavo:
Build console output for commit: c64612c42b6e7ffb0affcf933024217663960562 available at: https://gist.github.com/c86d6d8adb50d7c59eeab0dccef515f2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7874.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7874.dmg

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1166,8 +1165,46 @@ controller::Pose MyAvatar::getRightHandControllerPoseInAvatarFrame() const {
...
+            motorRotation = glm::quat();
+        }
+        _characterController.addMotor(_scriptedMotorVelocity, motorRotation, _scriptedMotorTimescale);
+    }
+
+    // legacy support for 'MyAvatar::applyThrust()', which has always been implemented as a
+    // short-lived linearAcceleration
+    _characterController.setLinearAcceleration(_thrust);
+    _thrust = Vectors::ZERO;
+}
```
AndrewMeadows: Each motor has: velocity and two timescales (horizontal and vertical).  A 'timescale' is the time it takes for the motor to get some fraction closer to its target velocity (more or less) so a small timescale makes the motor get there very quickly while a long timescale makes it slow.

If only one timescale is specified then the vertical is set equal to the hoirizontal and it is a symmetric motor in 3D.  Otherwise the motor is asymmetric and can push/pull differently along horizontal and vertical components.

The motors are passed o the `_characterController` which will blend the two motors together in an order-independent way.  Typically this blending happens later (in the physics substep) however when the avatar is "removed from the physics simulation" the blending is computed by making an explicit call to `CharacterController::computeNewVelocity()`.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1499,163 +1541,97 @@ void MyAvatar::updateOrientation(float deltaTime) {
...
+                motorSpeed += MIN_AVATAR_SPEED * boostCoefficient;
+            } else if (motorSpeed > finalMaxMotorSpeed) {
+                motorSpeed = finalMaxMotorSpeed;
             }
         }
+        _actionMotorVelocity = motorSpeed * direction;
+    } else {
+        // we're interacting with a floor --> simple horizontal speed and exponential decay
+        _actionMotorVelocity = MAX_WALKING_SPEED * direction;
     }
```
AndrewMeadows: This code controls how fast the 'fly','walk', and 'brake' target velocity changes.  This is all hand-tuned to maintain the speed-up/slow-down behavior.

hifi-gustavo:
Build console output for commit: aa171578b7fb1d60463bc9532e3fcbc5b38f64f2 available at: https://gist.github.com/c86d6d8adb50d7c59eeab0dccef515f2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7874.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7874.dmg

hifi-gustavo:
Build console output for commit: aab26c5089935508c0256f180af52dec01c5abcd available at: https://gist.github.com/c86d6d8adb50d7c59eeab0dccef515f2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7874.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7874.dmg

hifi-gustavo:
Build console output for commit: 9786c0bfd390c3e4f43a71611fedddeee21735f3 available at: https://gist.github.com/c86d6d8adb50d7c59eeab0dccef515f2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7874.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7874.dmg

Menithal:
Some feedback:

> MyAvatar.addThrust({x:0,y:0,z:1000000}) 

- Works and so does

> MyAvatar.addThrust({x:0,y:1000,z:0})

- Good to see down thrust ignored while walking. I am fine with that. avoids the accidentally burying self.

> MyAvatar.addThrust({x:0, y:-1000,z:0})

- Works when avatar is not on ground
- Using my hookgun script I am working for a worklist ticket: works nicely now and doesnt drag you along the ground any longer.
- Default Avatar movement seems alot faster now on the ground. Feels Faster than when flying

Not in the scope of the PR: but feedback regardless: there should be probably some thrust and time limit to how much non-equipped entity scripts, such as entity scripts can push the avatar around. This is -ohso exploitable- atm for griefing by rezing an entity with a script that on update loop MyAvatar.addThrust(<huge number>) to everyone who runs it locally. Overall there should also be a hard cap for any case.

AndrewMeadows:
@Menithal: down thrust is NOT ignored when walking.  The collision with the ground is completely negating it.  Since the avatar's RigidBody has the default restitution (bounciness) of 0.5 it would normally bounce off the ground, however the continual adding of thrust is probably hiding any small bounce effect.

I'll look at the tuning of the default avatar motion.  I tried to tune it to be the same, except to make the avatar push forward a little harder (not faster) so that it could walk up steeper inclines and maybe get over slightly larger steps (it still easily snags on shallow steps -- will have to be fixed in a later PR).

Yes, `MyAvatar.addThrust()` and `MyAvatar.motorVelocity` magnitudes will remain uncapped for now but we'll soon have to figure out how best to protect these features from grief.

Menithal:
@AndrewMeadows Then its very good at keeping you on the ground even if one uses 

> MyAvatar.addThrust({x:0, y:-100000,z:0})

Which usually phases one through objects if flying :)

PR #7879, opened 2016-05-12
fix reticle position offset

howard-stearns:
Make uses of getReticlePosition consistent with respect to window offset.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -864,10 +864,7 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
 
         if (action == controller::toInt(controller::Action::RETICLE_CLICK)) {
             auto reticlePos = getApplicationCompositor().getReticlePosition();
-            QPoint globalPos(reticlePos.x, reticlePos.y);
-
-            // FIXME - it would be nice if this was self contained in the _compositor or Reticle class
-            auto localPos = isHMDMode() ? globalPos : _glWidget->mapFromGlobal(globalPos);
+            QPoint localPos(reticlePos.x, reticlePos.y); // both hmd and desktop already handle this in our coordinates.
             if (state) {
                 QMouseEvent mousePress(QEvent::MouseButtonPress, localPos, Qt::LeftButton, Qt::LeftButton, Qt::NoModifier);
                 sendEvent(_glWidget, &mousePress);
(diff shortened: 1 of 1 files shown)


howard-stearns:
Some uses of getReticlePosition adjusted for window offset, some did not, and some adjusted only in desktop mode.

When I changed getReticlePosition to adjust for window offset in desktop (which is the case that needed it), I didn't update these callers. This PR removes all the callee adjustments.

howard-stearns:
bugzid:504

howard-stearns:
To test:

Do as many different kinds of "mouse clicks" and "drag" as you can, in desktop, oculus, and vive.
In this case, a mouse click is left click hardware mouse, button 3 of the they hydra (either hand), and clicking the center of the trackpad with the vive controller (either hand).
Do this for system overlays (including moving, dragging and closing the windows, and the buttons and drag bars on those windows), 2d scripted overlays (such as buttons and transparent toolbar that the edit.js buttons are set into), 3d overlays like the edit.js drag and rotation overlays, and in world-objects like the home light switch.
Do this with the Interface window full size and also with it smaller and not located in the upper left corner of the monitor.

Note: in edit.js, entity selection must be done with either the mouse or the Entity List panel. Clicking an entity with the hand controllers does not work.

hifi-gustavo:
Build console output for commit: 40f7212b1e248898560dd1589a1613def415419f available at: https://gist.github.com/142ee4de2db30572e002ee5d06bacc81<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7879.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7879.dmg

OmegaHeron:
QA Pass for portion I'm able to test.
Testing Windows 10 64bit - i5 6500 CPU - GTX970 GPU - **Desktop mode - keyboard/mouse only**
No negative change to baseline CPU/GPU/Memory use from RC/Other PR.
Was able to interact with "clicky things" regardless of window size, placement or Fullscreen !fullscreen.  Ultimate torture test - stretched to cover 2 monitors of different sizes and was still able to due everything proper involving clicking.

ZappoMan:
This PR is a release blocker... because it fixes a regression that was introduced in the dev-builds and is not in the stable builds.

PR #7882, opened 2016-05-13
Wait and cleanly exit the scripting thread

zzmp:
The scripting thread can be in three states:
- "normal evaluation" (most scripts): calling `ScriptEngine::stop` will stop the thread.
- "hanging evaluation" (`while(true){}`): `ScriptEngine::abortEvaluation` will break out of the evaluation.
- "hanging signal" (`Script.scriptEnding.connect(function(){ while(true){} );`): calling `ScriptEngine::currentContext()->throwError()` will break out of the signal handler.

AFAICT, evaluation only occurs once per loop of the scripting thread, so calling `ScriptEngine::stop` guarantees that there will be (at most) one "hanging evaluation." A signal can be connected to multiple times, so there is no limit to "hanging signal" states, but [quitting a thread should prevent further asynchronous signals from being handled (as the event loop will be stopped)](http://doc.qt.io/qt-5/qthread.html#quit).

To guarantee that a scripting thread is properly destructed, the thread should be `stop`ped, and any hanging states broken out of until the thread is no longer running. [Deleting a running QThread (i.e. isFinished() returns false) will probably result in a program crash.](http://doc.qt.io/qt-5/qthread.html#dtor.QThread)

This is handled in `ScriptEngine::wait`, but is a blocking operation. It cannot be a part of `~ScriptEngine`: the dtor is triggered by `deleteLater` and the `ScriptEngine` lives on the scripting thread, so calling `wait` would cause the thread to wait on itself.

---

The `EntityTreeRenderer` holds a scripting thread, and needs to delete it correctly, so it must invoke the blocking `wait` at some point before `ScriptEngine::deleteLater`. However, it runs on the main thread, so doing so might cause visible lag.

To solve this, It does so in the QThreadPool::globalInstance. This necessitates wrapping `wait` and `deleteLater` in a QRunnable.


libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -77,26 +78,40 @@ EntityTreeRenderer::~EntityTreeRenderer() {
...
+    });
+
     _scriptingServices->registerScriptEngineWithApplicationServices(_entitiesScriptEngine.data());
     _entitiesScriptEngine->runInThread();
     DependencyManager::get<EntityScriptingInterface>()->setEntitiesScriptEngine(_entitiesScriptEngine.data());
 }
 
 void EntityTreeRenderer::clear() {
     leaveAllEntities();
-    if (_entitiesScriptEngine) {
```
zzmp: I've moved this into the deleter for better encapsulation.
This means that unloadAllEntityScripts needs to be called from shutdownScripting.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -155,24 +154,23 @@ ScriptEngine::ScriptEngine(const QString& scriptContents, const QString& fileNam
 }
 
 ScriptEngine::~ScriptEngine() {
-    qCDebug(scriptengine) << "Script Engine shutting down (destructor) for script:" << getFilename();
```
zzmp: I changed a lot of these comments so they shared the same form, and could easilly be read in a batch.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -155,24 +154,23 @@ ScriptEngine::ScriptEngine(const QString& scriptContents, const QString& fileNam
...
+    qCDebug(scriptengine) << "Script Engine shutting down:" << getFilename();
 
     auto scriptEngines = DependencyManager::get<ScriptEngines>();
     if (scriptEngines) {
         scriptEngines->removeScriptEngine(this);
     } else {
         qCWarning(scriptengine) << "Script destroyed after ScriptEngines!";
     }
-
-    waitTillDoneRunning();
```
zzmp: If we are in the dtor, then the `DeferredDeletion` event has been processed, so the run should be over.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -155,24 +154,23 @@ ScriptEngine::ScriptEngine(const QString& scriptContents, const QString& fileNam
...
 }
 
 void ScriptEngine::disconnectNonEssentialSignals() {
     disconnect();
-    QThread* receiver;
+    QThread* workerThread;
```
zzmp: Fixed a naming inconsistency.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -272,57 +267,71 @@ void ScriptEngine::runInThread() {
...
             //       The application thread will need to continue to process events, because
             //       the scripts will likely need to marshall messages across to the main thread, e.g.
             //       if they access Settings or Menu in any of their shutdown code. So:
             // Process events for the main application thread, allowing invokeMethod calls to pass between threads.
-            QCoreApplication::processEvents(); // thread-safe :)
+            QCoreApplication::processEvents();
 
-            // If we've been waiting a second or more, then tell the script engine to stop evaluating
-            static const auto MAX_SCRIPT_EVALUATION_TIME =  USECS_PER_SECOND;
+            // If the final evaluation takes too long, then tell the script engine to stop evaluating
```
zzmp: s/evaluating/running

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -272,57 +267,71 @@ void ScriptEngine::runInThread() {
...
+                    if (context) {
+                        context->throwError("Timed out during shutdown");
+                    }
+                }
+
+                // Wait for the scripting thread to stop running, as
+                // flooding it with aborts/exceptions will persist it longer
+                static const auto MAX_SCRIPT_QUITTING_TIME = 0.5 * MSECS_PER_SECOND;
+                if (workerThread->wait(MAX_SCRIPT_QUITTING_TIME)) {
+                    workerThread->terminate();
```
zzmp: `terminate` does not immediately stop the thread, and `wait`ing afterwards didn't make much of a difference in my testing, so I just continue through the loop.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1309,8 +1292,23 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
         setParentURL(scriptOrURL);
     }
 
+    const int SANDBOX_TIMEOUT = 0.25 * MSECS_PER_SECOND;
```
zzmp: Reviewer: Is this an appropriate duration?

libraries/script-engine/src/ScriptEngine.h
```diff
@@ -67,9 +67,9 @@ class EntityScriptDetails {
 class ScriptEngine : public QScriptEngine, public ScriptUser, public EntitiesScriptEngineProvider {
     Q_OBJECT
 public:
-    ScriptEngine(const QString& scriptContents = NO_SCRIPT,
-                 const QString& fileNameString = QString(""),
-                 bool wantSignals = true);
+    static const auto MAX_SCRIPT_EVALUATION_TIME = USECS_PER_SECOND;
```
zzmp: This does not need to be exposed. I will put it back inside of `wait`.

libraries/script-engine/src/ScriptEngine.h
```diff
@@ -84,6 +84,13 @@ class ScriptEngine : public QScriptEngine, public ScriptUser, public EntitiesScr
     void run();
 
     ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+    // NOTE - this is intended to be a public interface for Agent scripts, and local scripts, but not for EntityScripts
```
zzmp: I exposed these methods for use in the EntityTreeRenderer.

libraries/script-engine/src/ScriptEngine.h
```diff
@@ -166,6 +169,7 @@ class ScriptEngine : public QScriptEngine, public ScriptUser, public EntitiesScr
 public slots:
     void callAnimationStateHandler(QScriptValue callback, AnimVariantMap parameters, QStringList names, bool useNames, AnimVariantResultHandler resultHandler);
     void updateMemoryCost(const qint64&);
+    void abort() { abortEvaluation(); }
```
zzmp: This does not need to be exposed. I will remove it.

ZappoMan:
This looks like a pretty significant set of changes. We will need to do some very thorough testing of this.

hifi-gustavo:
Build console output for commit: 806d06b5529104c6f41f1748381204c1f4c6b9fc available at: https://gist.github.com/e0a53deb68745c748e0e4ba7826a3976<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7882.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7882.dmg

libraries/script-engine/src/ScriptEngines.cpp
```diff
@@ -157,20 +158,22 @@ void ScriptEngines::shutdownScripting() {
             // and stop. We can safely short circuit this because we know we're in the "quitting" process
             scriptEngine->disconnect(this);
 
-            // Calling stop on the script engine will set it's internal _isFinished state to true, and result
-            // in the ScriptEngine gracefully ending it's run() method.
+            // If this is an entity script, we need to unload any entities
+            scriptEngine->unloadAllEntityScripts();
```
ZappoMan: I'm not really a fan of this being here. Certainly I don't think we should be calling this for every script engine... since all but one of them don't have any entity scripts...

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -138,10 +138,9 @@ static bool hadUncaughtExceptions(QScriptEngine& engine, const QString& fileName
     return false;
 }
 
-ScriptEngine::ScriptEngine(const QString& scriptContents, const QString& fileNameString, bool wantSignals) :
+ScriptEngine::ScriptEngine(const QString& scriptContents, const QString& fileNameString) :
     _scriptContents(scriptContents),
     _timerFunctionMap(),
-    _wantSignals(wantSignals),
```
ZappoMan: why did you remove this feature?

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -272,57 +267,71 @@ void ScriptEngine::runInThread() {
...
 
-    moveToThread(workerThread);
-
-    // Starts an event loop, and emits workerThread->started()
     workerThread->start();
 }
 
-void ScriptEngine::waitTillDoneRunning() {
-    // If the script never started running or finished running before we got here, we don't need to wait for it
+void ScriptEngine::wait() {
```
ZappoMan: why rename this? `waitTillDoneRunning()` seems more clear that it will wait till it's done running... `wait()` seems ambiguous.

zzmp:
Excellent reminder for me to post my own tests!

I used [loops.zip](https://github.com/highfidelity/hifi/files/262433/loops.zip) for my testing:
- I ran `test.js` as a client script, which loaded a few different types of infinite loops.
- I ran `unloadloop.js` as an entity script.

Running these scripts, I made sure that I could cleanly and quickly:
- switch domains
- shut down

The loops that cannot be interrupted by interception (e.g. while and signal-based loops) log their lifetimes as they quit. As they quit, you can see how many attempts it takes to abort, throw, and terminate them. Sometimes it takes a few, but it shouldn't take more than 2-3 seconds.

I identified two griefing vectors here:
- Loading many while and signal-based looping scripts will take a long time to shut down, as they are shut down one-at-a-time, and each may take 2-3s (tested).
- Hooking up many signal-based loops in one script may require multiple throws to escape, and each one will wait 0.5s (untested).

I also identified and solved a griefing vector:
- Loading an entity script which inifinitely loops before returning ties up its sandbox ScriptEngine. Now, there is a hard 250ms limit on entity constructor loading in the sandbox.

libraries/script-engine/src/ScriptEngine.cpp
zzmp: It was never used, and made signals harder to reason about, adding complexity for no apparent benefit. Specifically, it filtered out the signals that could be used to delete the `workerThread`.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -945,14 +940,8 @@ void ScriptEngine::stopAllTimersForEntityScript(const EntityItemID& entityID) {
 
 void ScriptEngine::stop() {
     if (!_isFinished) {
-        if (QThread::currentThread() != thread()) {
```
ZappoMan: why did you remove this thread switching?

libraries/script-engine/src/ScriptEngine.cpp
zzmp: It's a threaded feature, and `wait` has a specific meaning for threads that I thought applied fine here, so I preferred the more familiar method name.

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: sure

libraries/script-engine/src/ScriptEngines.cpp
zzmp: That's fine with me, I can change this back.

libraries/script-engine/src/ScriptEngine.cpp
zzmp: It's guarding an atomic.

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: I assume this is because you had a test case of an entity script with an infinite loop in the constructor... if so, let's make sure we include this in the test case.

zzmp: Exactly, it's in the tests that I posted.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -77,26 +78,40 @@ EntityTreeRenderer::~EntityTreeRenderer() {
...
 
-void EntityTreeRenderer::setupEntitiesScriptEngine() {
-    QSharedPointer<ScriptEngine> oldEngine = _entitiesScriptEngine; // save the old engine through this function, so the EntityScriptingInterface doesn't have problems with it.
-    _entitiesScriptEngine = QSharedPointer<ScriptEngine>(new ScriptEngine(NO_SCRIPT, QString("Entities %1").arg(++_entitiesScriptEngineCount)), &QObject::deleteLater);
+void EntityTreeRenderer::resetEntitiesScriptEngine() {
+    // Keep a ref to oldEngine until newEngine is ready so EntityScriptingInterface has something to use
+    auto oldEngine = _entitiesScriptEngine;
+
+    auto newEngine = new ScriptEngine(NO_SCRIPT, QString("Entities %1").arg(++_entitiesScriptEngineCount));
+    _entitiesScriptEngine = QSharedPointer<ScriptEngine>(newEngine, [](ScriptEngine* engine){
```
ZappoMan: please explain this new code more clearly... what exactly are you doing here? what are the repro cases for the old technique failing.

zzmp: Explanation: see updated PR description.

Repro: I'll need to test in the morning, but one of these two (I'm guessing the second) may cause a crash at this point. I will repro in the morning, and include a stack trace and explanation of how it failed:
- `while(true){}`
- `Script.scriptEnding.connect(function(){ while(true){} });`

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: but it's doing more than just waiting... it's also stopping the execution/evaluation of scripts... pondering

zzmp:
**TODO: This has not been tested with debuggable scripts!**

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: right - seems reasonable

libraries/script-engine/src/ScriptEngine.h
```diff
@@ -138,10 +145,6 @@ class ScriptEngine : public QScriptEngine, public ScriptUser, public EntitiesScr
 
     Q_INVOKABLE void requestGarbageCollection() { collectGarbage(); }
 
-    ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-    // NOTE - this is intended to be a public interface for Agent scripts, and local scripts, but not for EntityScripts
-    Q_INVOKABLE void stop();
```
ZappoMan: This removes the `Script.stop()` feature that's used by agent scripts... we can't do that.

ZappoMan: I retract this false accusation!

libraries/entities-renderer/src/EntityTreeRenderer.cpp
zzmp: This needs to be moved back. It needs marshalling if called from a thread other than the main.

hifi-gustavo:
Build console output for commit: 2140dc77b3ccbb284c1c018e30d23589ae12378e available at: https://gist.github.com/e0a53deb68745c748e0e4ba7826a3976<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7882.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7882.dmg

ZappoMan:
@zzmp  this needs to be rebased... it has the bad FPS issue caused by my snafu you will likely get a QA Fail if you don't rebase before testing.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
ZappoMan: @zzmp didn't we agree you were going to move this out of a lambda and make it easier to read?

hifi-gustavo:
Build console output for commit: 96f0c9cbcf0f07e367b24e947145e225abf898b1 available at: https://gist.github.com/e0a53deb68745c748e0e4ba7826a3976<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7882.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7882.dmg

hifi-gustavo:
Build console output for commit: 7d4276d15fe229855f8de6e9afe2b912e02d6761 available at: https://gist.github.com/e0a53deb68745c748e0e4ba7826a3976<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7882.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7882.dmg

zzmp:
**QA**: Switching domains has reportedly been more brittle on macs. Please make sure this gets tested on a mac as well.

ZappoMan:
Although this bug is not a regression from stable, this is an issue that happens fairly often in release... and so we'd like to see this PR get tested and merged into stable asap.

OmegaHeron:
Testing Windows 10 64 bit - i5 6500 CPU - GTX970 GPU - **Desktop Mode - keyboard/mouse**

Running test.js is impossible with my system config.  It consume 100% CPU on all cores, if I can manage to get enough UI responsiveness I'm unable to stop scripts via Stop All or individual stop buttons.  Since it so overburdens CPU my entire UI (Windows UI) becomes nearly unusable - Nvidia driver goes into a loop of NV display driver has crashed/NV display driver has recovered.  In one case I was able to close interface via its closer action (X) and it does fully terminate.

But, there seems something still amiss here in that Interface can consume vast quantities of resources and ultimately destabilize entire system with some infinite loop scripts....

I'll try other tests shortly.

OmegaHeron:
Running loop.js in a cube entity... all was well for a few, I saw nothing in log other than its loading.  Moved to another domain and came back to domain with entity + loop.js - was looking at log file, CPU use suddenly went off the charts - Windows UI became unresponsive - after much struggle I got mouse cursor to closure X on Interface window and clicked resulting in a bugsplat report.

Moving on to next after I compare how current RC behaves with this test.

OmegaHeron:
While running loop.js here's what I see in log
[05/16 15:41:07] [DEBUG] Found script in cache: "atp:/loop.js"
[05/16 15:41:12] [DEBUG] Total allocation  "12.61 MB"
[05/16 15:41:56] [WARNING] udt::SendQueue wanted to sleep for 2808096 microseconds
[05/16 15:41:56] [WARNING] Capping sleep to 2000000
[05/16 15:41:56] [WARNING] PSP: 6740624 NPD: 6740624 NPT: 2556882936336 NOW: 2554074839590
[05/16 15:41:56] [DEBUG] Logging activity "sendqueue-sleep"

Then - once I make Interface INACTIVE to write this is when it goes bad.  Suddenly Interface has a huge spike in resource use and entire Windows UI becomes unresponsive - then NV driver crash/recover loop until I finally manage to terminate Interface.

With build 4884 running this same entity with loop.js active I do see high CPU use but no sign of death spiral while writing this as it runs.  Sorry - but something just isn't right with this one.  Whatever its aim to solve may be - it ultimately makes the perfect storm to grief Interface on my configuration.  Maybe it's just an i5 CPU thing...

ZappoMan:
@OmegaHeron - this PR doesn't change anything obvious about the "running state" of scripts... it only changes the shutdown behavior. I think you should retry your test on the latest "dev build"... because if there's a difference between stable and this build, it was probably introduced by some other PR.

OmegaHeron:
@zappoman - Will do.

OmegaHeron:
It's going to take a bit more time to sort this, but, @ZappoMan, you are correct.  It's not from this PR.  Still backtracking but need dinner break.

OmegaHeron:
I, reluctantly acknowledge I admit to when I screw up.  I had transposed shortcuts for this PR with another leading to failure for tests after first.

Test 1 - running test.js fails in that as soon as all the infinite loop creators spin up my system destabilizes to point of needing reboot.  I attempted to hit stop all scripts, but, only a few seemed to stop before deathspiral finish.

Test 2 - loop.js in an entity this PR **prevents** it from dragging system down.  It gives output in log as expected and disallows script's negative impact.

Test 3 - unloadloop.js in an entity - This PR allows one to leave domain where it's running while lastest RC/Dev crashes.

I'm not sure what to do about failure with Test 1 - it's not really fault or responsibility of this PR and, if you can manage to get close on Interface it **will** shutdown while on RC/Dev it does not.  That said there should very likely be some research into preventing script engine from being so... freely giving of its time.

zzmp:
test.js is less of a burdren on i7, because it only takes one core (AFAICT). It behaved well in my testing.

OmegaHeron:
I'm wondering if TBB behaves very differently on an i5 than i7 or if it even comes into play here.  Something is definitely different - I see 100% core utilization across all 4 physical cores while running test.js on my setup.  I'll look at it on my i7 laptop as soon as I can.  There's far too much delta in how this run on i5 vs i7 in too many ways.

cozza13:
tested with scripts above. QA pass

PR #7885, opened 2016-05-13
make users.js movable

ctrlaltdavid:
A draggable border is displayed when you mouse over the users list.


ctrlaltdavid:
QA Notes
- Hover over users.js and a border similar to that displayed for edit.js is displayed. You can click and drag on the border to reposition the users list.
- The users list scrolls when there's not enough room to display the users before hitting the top of the screen.
- The window's position (the position of the bottom of the window since the window expands vertically) is remembered over script and Interface restarts; it is remembered relative to the nearest window edge.
- The window's position is also maintained relative to the nearest edge when the Interface window is resized.

hifi-gustavo:
Build console output for commit: 1928f51b3e61f2c53ab6198de494b722b18d0485 available at: https://gist.github.com/1286fe272a8f84a1e738d9f49d18fb60<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7885.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7885.dmg

Nex-Pro:
#### Testing platform: Desktop PC 16gb ram
#### CPU: 4.4 ghz I7 @4.0ghz (devils canyon)
#### GPU: Nvidia gtx770 2gb.
#### OS: WIN 10 64.
#### HMD: None
#### Hand Controllers: None

@ctrlaltdavid Thank you for adding this feature.

**Initial test result: Was able to move the users panel around the screen.**

User window moves sluggishly and trails far behind the cursor and movement is not smooth.
Feels reluctant.

Position of user window persists after restarting interface.

First quit of this PR resulted in a zombie process in the taskmanager but no bugsplat.

Subsequent quits have not repeated that as yet.

Moving the users window towards the top of the screen results in slightly odd behaviour when nearing the top of the main window.  The users window appears to stop then as the mouse is moved further upwards the users window collapses from the bottom up to a point where the users are no longer visible. Then the window continues to move upwards in its newly collapsed state. I am not sure if this is intentional. I need to try this with a longer list of users and will do so later.

ZappoMan:
@ctrlaltdavid - does this use the "recommended" rectangle? So that in the HMD it's not all the way in the back?

scripts/system/users.js
```diff
@@ -792,13 +915,36 @@ var usersWindow = (function() {
         windowLineHeight = windowTextHeight + windowLineSpacing;
         Overlays.deleteOverlay(textSizeOverlay);
 
-        viewportHeight = Controller.getViewportDimensions().y;
+        viewport = Controller.getViewportDimensions();
```
ZappoMan: this is wrong. we need to get the recommended rectangle... not the viewport size.

ZappoMan:
@ctrlaltdavid - I'm going to call this QA Fail because it doesn't honor the recommended rectangle.

Nex-Pro:
@ctrlaltdavid Ok.

ctrlaltdavid:
@ZappoMan I gather that the recommended rectangle is a subrectangle of the rendered view in which UI elements can readily be seen when in HMD view? @Ryan asked me to make the users window draggable so that it could be dragged to be more visible in HMD view. If the users window is restricted to lie within the recommended rectangle then it will clutter up the desktop view. Perhaps what's really required is that when in desktop view it's positioned relative to the full viewport but when in HMD view its positioned relative to the recommended rectangle (i.e., changes position when you switch between desktop and HMD view)?

howard-stearns:
@ctrlaltdavid do you have access to https://highfidelity.atlassian.net/wiki/display/HOME/HUD ?
That spec's the positioning and resize behavior. (And excludes separate HMD/destkop settings for the stuff we provide, but doesn't prohibit it for third-party stuff.)

RyanDowne:
This worked as I would expect with the exception that the starting point of the window is still pegged to the far left corner of the (in my case Oculus) space and it's starting point should be by default somewhere within a comfortable viewing range.

ctrlaltdavid:
@howard-stearns No, sorry, I don't have axes to that doc.

@RyanDowne I'll change the default start-up position to be at the bottom left of the recommended rectangle.

@Nex-Pro Re sluggish movement: is it any more sluggish than moving the edit toolbar? (Both are equally sluggish for me, but then I get subpar FPS so am not surprised.)

howard-stearns:
OK, I've emailed a current copy, FYI.

Nex-Pro:
@ctrlaltdavid  I guess equally sluggish

Nex-Pro:
@ctrlaltdavid  BTW in pr#9896 the sluggishness of moving the edit panel has pretty much gone as I get 60fps.

ctrlaltdavid:
@Nex-Pro This PR's build has the FPS slow-down that was fixed in a new build today. I'll update this PR with the fix from the latest master code.

Nex-Pro:
@ctrlaltdavid  yep.

hifi-gustavo:
Build console output for commit: 33e2cde266c0cf814ce7b1d994f6a9b096be0a14 available at: https://gist.github.com/1286fe272a8f84a1e738d9f49d18fb60<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7885.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7885.dmg

Nex-Pro:
@ctrlaltdavid  check this video re this pr

https://dl.dropboxusercontent.com/u/96759331/hifi_scripts/2016-05-14-1622-55.mp4

ctrlaltdavid:
@Nex-Pro So ... the user.js window moves pretty responsively which is good (much more responsively than the less sluggishly-moving the edit.js toolbar which is bad) ... and with both you can click-through them and end up moving both the overlay and an entity at the same time which is bad. ... Correct?

Nex-Pro:
Yep. Video is worth 999 words.

PR #7897, opened 2016-05-13
reset hud on driving

howard-stearns:
Reset hud on driving (all platforms/modes), and don't show overlays while driving.

The purpose of this PR is to solicit input on:
 1) The behavior. We've talked about this, but want to see how people feel about using it.
        Too much reset?
        Different definition of movement? (Currently based on linear velocity, with hysteresis.)
 2) The implementation, which is a bit hacky.

When trying it out, please note that there is a separate bugzid:523 that causes the reset to return your point-of-view to whatever the last reset was. (This bug effects apostrophe, too.) So maybe the best way to try it is to keep your head straight and simply strafe your body a bit sideways. (Or we can fix the bug....)


hifi-gustavo:
Build console output for commit: 1f664671f91070e868f4a29b12835b588670d200 available at: https://gist.github.com/71b25ed752bb0007fa498a7e702b83ee<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.dmg

hifi-gustavo:
Build console output for commit: 654ed33dd4e870a6f1973ef6efe0b3b31b4b93dd available at: https://gist.github.com/71b25ed752bb0007fa498a7e702b83ee<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.dmg

hyperlogic:
So, calling displayPlugin->resetSensors is kind of heavy handed, IMHO, I don't think we should reset the HMD, which might interact badly with the rest of the avatar code.  Instead perhaps we should just reset the UI sphere instead, with an offset matrix based on the current HMD position orientation.  my two cents.

hyperlogic:
Two things about the PR based on testing:
* When I walk away or lean into the UI and it resets.  The world around me seems to 'snap' in orientation (and perhaps in position) as well.  This is pretty bad.
* The fading in and out of the overlay is not always consistent.  It seems to almost always fade out, but very rarely fade in, instead it will pop in with an alpha of 1.0.  And if it does fade in, it usually fades in after it's been visible for a few frames.

howard-stearns:
Re snapping: Yeah, thats why I'd like to get fogbugzid:523 fixed.

Re fading back in: I'm not sure what's going there. I've observed that if I start walking at the start of a session, while things are still loading, the overlays won't come back in while things still load. (Downloads in stats, progress bar, etc.) But once everything is loaded, it seems to reliably fade out and in. Weird.

Re reset sensors: No objection. Do you mean that something less than the three lines of OpenVrDisplayPlugin::resetSensors would be broken out and called, or do you mean that we would call getActiveDisplayPlugin()->resetSensors() conditionally (e.g., only on the Vive)?

Re implementation vs functionality generally: The risk that I'm worried about, going forward, is whether the concept is what we want, so that we don't go chasing our tail on functionality. If so, then I'm sure we can work through the bugs and the right JS exposure. How do you all feel about the concept (fade out on trigger, and reset on fade in on end-trigger), and on the feel of the trigger for it (velocity)?

hyperlogic:
> Re reset sensors: No objection. Do you mean that something less than the three lines of OpenVrDisplayPlugin::resetSensors would be broken out and called, or do you mean that we would call getActiveDisplayPlugin()->resetSensors() conditionally (e.g., only on the Vive)?

I don't think we should call resetSensors at all, but instead move the UI sphere.

hyperlogic:
> Re implementation vs functionality generally: The risk that I'm worried about, going forward, is whether the concept is what we want, so that we don't go chasing our tail on functionality. If so, then I'm sure we can work through the bugs and the right JS exposure. How do you all feel about the concept (fade out on trigger, and reset on fade in on end-trigger), and on the feel of the trigger for it (velocity)?

How do I feel? I think it's ok given the compromises.  If we could somehow magically guess the users intent, I'd want the UI to only fade-in or fade-out when they wanted it to.  For example: based on a 'button' click.  But yeah, we're all out of buttons...

howard-stearns:
Recording some functional feedback so far:
 1. Trigger for disappearing is too frequent. Maybe decouple from stepping, and base it on distance.
 2. Trigger for re-appearing is too frequent. Maybe introduce a delay in which you have to stay still for a moment before it comes back.
 3. Both of these will be less important when we implement an always-on hud on/off toggle button. E.g., you would often have hud off and so would not notice 1 and 2, and even if it was on and you were frustrated by the delay in 2, you could toggle it back on yourself.

hifi-gustavo:
Build console output for commit: 8381e74fb3478750362de176aa0710b369469c0f available at: https://gist.github.com/71b25ed752bb0007fa498a7e702b83ee<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.dmg

hifi-gustavo:
Build console output for commit: 7f131e2c4fe06eb2d2faf3901ebcece4ec3c14ec available at: https://gist.github.com/71b25ed752bb0007fa498a7e702b83ee<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.dmg

ZappoMan:
Some feedback:
- I can't seem to ever turn to see the mic mute
- I can't figure out how to use the hand controllers for the mouse.

hifi-gustavo:
Build console output for commit: 3088a840351707182b20f08fda069b5905835cfb available at: https://gist.github.com/71b25ed752bb0007fa498a7e702b83ee<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.dmg

howard-stearns:
Overall, this addresses fogbugzid:522 / 541
The OpenVrDisplayPlugin.cpp part here fixes fogbugzid:523.

hifi-gustavo:
Build console output for commit: 9a428947e4f9fa6c51e76d69d385dc2eec33b945 available at: https://gist.github.com/71b25ed752bb0007fa498a7e702b83ee<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.dmg

hifi-gustavo:
Build console output for commit: 097b16aadc5e0f135e2df0a93a6bf7a939abccbe available at: https://gist.github.com/71b25ed752bb0007fa498a7e702b83ee<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.dmg

hyperlogic:
build failed on linux due to a failed download of bullet.  test this please

hyperlogic:
So, after fixing the issue with sensor reset on oculus, there's still a pretty glaring problem.  When you do a sensor reset, either by hitting the `'` key or in this PR by moving your head, you end up seeing part of the black rectangle from time-warp for a single frame.  I'm not sure if this is a DK2 only issue, but it's pretty bad and I'm not sure how to fix it.

howard-stearns:
We've had a pretty bad black rectangle problem for a while now -- in playa and in the NYC apartment. Is this worse?

hifi-gustavo:
Build console output for commit: edfce0d5badcac66e0b857c24b47875c37d33b04 available at: https://gist.github.com/71b25ed752bb0007fa498a7e702b83ee<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.dmg

howard-stearns:
Test plan:

On all displays (Vive, Oculus, and desktop):
- Resetting position (apostrophe key) should not appreciably change your point of view. (In HMD, it should re-allign your feet/hips to be under your head, with your toes aligned with your head.)
- When you start to move with the hand controllers or keyboard, the HUD should fade out. (This should have the same effect as turning off View->Overlays. Specifically, system menus and scripted overlays should fade out, as should stats if they are on.)
- A moment after you finish moving, the HUD should fade back in.  If you start and stop repeatedly within the fade-in timeout, it should not flicker. Also, if View->Overlays is unchecked, the overlays should not go on when you stop moving.
- The same should happen after a step. So, for example, in room-scale Vive, you should be able to take several steps with the HUD fading out, and come back in only when you're done moving.

Dev plan:
- There are some internal changes to this planned for the next sprint, such as for example, chaning the reset code. However, this should cover the expected functionality.

OmegaHeron:
Is there an option to **disable** this action?  I'm fine with it being on by default, but, I can think of numerous circumstances where I do not want this.

OmegaHeron:
Ok - desktop mode, I'm looking downward at something... I take a few steps back, enough move to fade HUD then stop.  My head snaps back up.  **NO** - It should not do that.

thoys:
How would one have a speedometer hud? :)

OmegaHeron:
Testing Windows 10 64 bit - i5 6500 CPU - GTX970 GPU - **Desktop mode - keyboard/mouse**
Passes standard tests.

In so far as this seems to do exactly what author intends I can't fail it.  But.
Needs an option to **disable** this.
Needs an option not to snap my AV back to T-Pose when I stop moving.
Doesn't disable HUD on left/right movement - only forward/backward.
Is annoying as hell if you have HUD displays you actually **want to see** while moving.

This is too heavy handed.

howard-stearns:
Great feedback, folks.
* It does seem like a preference would be a good idea.
* It should not t-pose when stopping, and it should work on left/right, and it should not change your head pitch. I'll try to reproduce those three. (And I'll remove the QA ready until then.)
* The theory is that it's generally wrong to show any HUD stuff while moving - EXCEPT that we would soon introduce a small always-on button to toggle HUD, and that HUD elements would generally each have their own toggle to keep them 'always on' also. Anything 'always on' would not disappear on movement.  (I don't know whether that theory is right, but that's what we're trying to find out.)

OmegaHeron:
I'm all for a lot of optional things - there are times when what this does would be perfect.  Without it being optional, I dislike everything about it - with optional and not buried in devel/advanced menus it would be potentially awesome.

OmegaHeron:
Also @howard-stearns note that I amended my T-Pose snap comment.  It doesn't snap to T-Pose on stop, but it does reset AV like a ' command was issued without seeing it go into T-Pose mode like you see with ' action.  Head goes back to level view.

ZappoMan:
@howard-stearns -- "The theory is that it's generally wrong to show any HUD stuff while moving" -- is it possible this is a desktop vs HMD nuance? I mean... showing the HUD on desktop seems less problematic... Also I can imagine cases where as a content developer debugging things you will want to see the HUD while moving... like for example when attempting to optimize content... checking render stats, etc...

howard-stearns:
Right, there are clearly cases where the user may decide they don't want this. I've add a preference as requested. It may also turn out that we will want more complicated suppression -- we'll see. 

The thing prompting all this is to keep the hud positioned on you as you move, and secondarily to not have the hud visible while resetting. What's not clear (to me) about a preference is whether the preference should turn off everything, or just whether or not the hud is visible while resetting. Right now (when it finishes building), it turns off all the resetting.

It's also arguable whether turning should trigger any of this behavior. (I cannot reproduce a failure to do this when moving left or right, so I'm interpreting @OmegaHeron 's  report as not resetting on turn.) The theory is that that people do want to rotate to look at different parts of the hud, without resetting, and that there are no ill-effects from doing so. Thus it is only when there's an actual translation of your avatar that we reset (and clear).

I still owe a fix to the issue of resetting the camera when looking down....

hifi-gustavo:
Build console output for commit: 2accccef9ed2cffa33948b392183a1648752d876 available at: https://gist.github.com/71b25ed752bb0007fa498a7e702b83ee<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.dmg

hifi-gustavo:
Build console output for commit: dc7d6d470ddc6ba5866131dc04a32c1cb7de5bee available at: https://gist.github.com/71b25ed752bb0007fa498a7e702b83ee<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.dmg

OmegaHeron:
QA Fail

Unsetting Clear overlays when driving - exit interface - restart results in
Snap turn when in HMD disabled - Clear overlays when driving enabled.  Fix that and I'll be happy to give QA Pass for Desktop mode.

howard-stearns:
good catch! On it.

hifi-gustavo:
Build console output for commit: 991da9e248338898b7ac582a5acaa81fd9d9e83e available at: https://gist.github.com/71b25ed752bb0007fa498a7e702b83ee<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.dmg

howard-stearns:
Ok, that should do it.

OmegaHeron:
QA Pass **Desktop mode**
All standard tests pass.
Preferences apply, persist and restore correctly.
Look at head angle doesn't snap to default on hud unhide.
Suggest forum post **explaining this new behavior and how to enable/disable before rolling out as it will lead some to think Interface is broken**.
Needs HMD - hand controller QA.

howard-stearns:
I've merge from upstream because that contains a fix for a separate bug, which fix will make it easier to test this PR in HMD.

howard-stearns:
Additional test plan item for Room Scale Vive:
  When first going to HMD mode from desktop (into Vive), you should not be outside the hud sphere.(fogbugzid:652)

hifi-gustavo:
Build console output for commit: 9ff1a695f03745204f398417246e44f098e3c834 available at: https://gist.github.com/71b25ed752bb0007fa498a7e702b83ee<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7897.dmg

thoys:
## Room scale Vive test results

- I am not able to walk out of the chaperone bounds HUD sphere, also not on startup. :+1: 
- It is kinda strange to see the HUDs disappear at first, but they will stay if you only look around you. :+1: 

Test passed :white_check_mark:

ZappoMan:
Defaulting this to "false" --- once it builds - I will merge.

ZappoMan:
never mind - I'm going to merge now -- it can't fail the build for the code change....

PR #7906, opened 2016-05-16
improve hand-controller edit

howard-stearns:
Makes edit.js more usable with hand controllers:
  The box edit-handles are larger.
  The distinction between a click (selection) and a drag, leans slightly more towards click than had been the case. This makes is slightly easier to select (a quick click-release with a steady hand) than before. (This is not a general redesign.  Selecting through the the Entity List on the edit panel is still the primary  recommended means of selection when using hand controllers.)

### DIFF
--- a/scripts/system/libraries/entitySelectionTool.js
+++ b/scripts/system/libraries/entitySelectionTool.js
@@ -304,7 +304,7 @@ SelectionDisplay = (function() {
     var previousHandleColor;
     var previousHandleAlpha;
 
-    var grabberSizeCorner = 0.025;
+    var grabberSizeCorner = 0.025; // These get resized by updateHandleSizes().
     var grabberSizeEdge = 0.015;
     var grabberSizeFace = 0.025;
     var grabberAlpha = 1;
@@ -4320,7 +4320,7 @@ SelectionDisplay = (function() {
     that.updateHandleSizes = function() {
         if (selectionManager.hasSelection()) {
             var diff = Vec3.subtract(selectionManager.worldPosition, Camera.getPosition());
-            var grabberSize = Vec3.length(diff) * GRABBER_DISTANCE_TO_SIZE_RATIO;
+            var grabberSize = Vec3.length(diff) * GRABBER_DISTANCE_TO_SIZE_RATIO * 2;
             var dimensions = SelectionManager.worldDimensions;
             var avgDimension = (dimensions.x + dimensions.y + dimensions.z) / 3;
             grabberSize = Math.min(grabberSize, avgDimension / 10);
--- a/scripts/system/edit.js
+++ b/scripts/system/edit.js
@@ -720,7 +720,7 @@ var mouseCapturedByTool = false;
 var lastMousePosition = null;
 var idleMouseTimerId = null;
 var CLICK_TIME_THRESHOLD = 500 * 1000; // 500 ms
-var CLICK_MOVE_DISTANCE_THRESHOLD = 8;
+var CLICK_MOVE_DISTANCE_THRESHOLD = 20;
 var IDLE_MOUSE_TIMEOUT = 200;
 var DEFAULT_ENTITY_DRAG_DROP_DISTANCE = 2.0;
 


howard-stearns:
The purpose here is to make it all possible (not necessarily ideally practical) to use the existing edit.js with hand controllers, just so we can gain some collective experience with doing so.  I.e., just move the ball forward a bit. 

These changes (and more) will still be needed when go to a full grab-like-laser-and-simulated-mouse infrastructure. 

edit.js is quite dependent on distinguishing between a click and drag, which is quite hard to do with hand controllers (where there is always some movement). It is especially hard with a lighthouse-on-poles setup as we have in the office, where everything is constantly moving no matter how still your controllers are.

scripts/system/libraries/entitySelectionTool.js
```diff
@@ -4320,7 +4320,7 @@ SelectionDisplay = (function() {
     that.updateHandleSizes = function() {
         if (selectionManager.hasSelection()) {
             var diff = Vec3.subtract(selectionManager.worldPosition, Camera.getPosition());
-            var grabberSize = Vec3.length(diff) * GRABBER_DISTANCE_TO_SIZE_RATIO;
+            var grabberSize = Vec3.length(diff) * GRABBER_DISTANCE_TO_SIZE_RATIO * 2;
```
howard-stearns: I feel that this arguably violates the no-magic numbers coding standard. However, there is a lot of similar code nearby (e.g,  lines 4333 and 4347).   I don't want to change the usage throughout all the edit.js code, and I don't want to leave the code with internally inconsistent styles.

howard-stearns:
Test plan:

With a mouse, use edit.js: select objects, drag them, and drag the resize handles. It should feel like before, although the handles are bigger. Do this with several different sized objects, and several distances between user and object.

With hydra, do the same. This should work pretty well, although the click for selecting objects does have to be rather quick and steady.

With vive, I find this useble for drag and resize, but the lighthouse-on-a-pole setup at my desk office has too much movement to be practical for selection. After this builds, I'll try the caves with wall-mounted lighthouses.

hifi-gustavo:
Build console output for commit: 3f17852244b92dfdf449327b80c2e330639a925d available at: https://gist.github.com/ead2bb7df9dac920e524967470121b27<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7906.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7906.dmg

Menithal:
## QA Test:

**QA Tester:** Menithal
**Verdict:** Needs more work

**Results:**
Since only a js related test, only doing according to testplan without a smokescreen test.
I can test using Vive and The Mouse

### Mouse: 

- Worked just as in Main. However I managed to size my object to a tiny size of 0.001^3 and it became impossible to select again from first person mode without the edit window.
- At 0.0100^3 it is still drag around with edit, but its night impossible to resize or move it with the handles it even if right next to the object in first person mode (if the object is on the floor)
- At 0.1000^3 the object is fully able to be transformed and resized at a comfortable distances, but selection being more important.
- Similar occurs in Main, but it would be nice to have a safeguard to be able to resize objects that accidentally go too small.

### Vive: 

- Has issues when you go out of the center of the room, hard to see where the exactly mouse  is pointing, 
- UI rendering only in center of the room makes interacting with ui difficult.

Now when Centered in the room:
- UI interaction works when centered, but limited to 180 degree cone
- One handed only? You can even switch holding the active trackpad press down and it will still hold in the same hand that it was one first.
- Difficult to select the Vertical Widget if object is small enough
- Can easilly confuse between movement with trackpad 
- Selecting objects is very tricky.
- Switching between hand controller and HMD selection is really annoying
-  Behavior when dragging an object to the right seems to consistantly not behave the same way as the left, regardless of the rotation of the avatar. Feels very sticky and some relative rotations of the object are completely lost
- Unable to select objects behind the 180 degree cone of HUD direction
- Using edit field up / down arrows is very... annoying. have to be close to the window to do so. Would be better if I could select field and move hand up or down to change value higher or lower

Personally I think for the HMD and Handcontroller such as the Vive, the use of the QML simply doesnt feel to work properly.

howard-stearns:
That's good feedback, @Menithal, but I don't think it's on what this change is about. 
- edit of small objects, with mouse or otherwise. There's no change for that in this PR. There may be a rework or replacement to edit.js in the future, and we should note that issue for any such work.
- vive HUD not moving as you move about. There's no change for that in this PR. (There's an early straw man prototype for that in #7879, for which feedback is welcome!)

The changes here are only:
 1) Make the resize-boxes bigger.
 2) Make the existing ad-hoc distinction between "click" and "drag" lean slightly more towards "click".

For both, we want to make sure that we haven't somehow broken mouse behavior with these two changes, in a way that wasn't already broken.  It would also be nice if could confirm that the changes are directionally correct -- 
   for (1), given that an object is selected (through the Entity List in the edit panel), is at least possible to resize using the boxes (with hydra or vive). Without this PR, it was nearly impossible.
   for (2), I think it helps with selection on hydra, but isn't really enough of a help on vive.

Menithal:
I havent personally used the edit while in the Vive because its just been so funky and still is, so I do not have a reference point, which is probably *why the feedback is more of a shotgun blast than a pinpoint accuracy to the PR* it self, all of them are for the hand controllers :) But generally speaking for the two points:

1) Confirmed They are larger, but could be clamped after a certain size
2) Hand controller selections will snap when they do snap. But the selection of the vertical widge is still very flaky

PR #7934, opened 2016-05-19
reduce number of points in unreasonable convex hulls

AndrewMeadows:
* don't allow convex hulls with more than 42 points

### DIFF
--- a/libraries/physics/src/PhysicalEntitySimulation.cpp
+++ b/libraries/physics/src/PhysicalEntitySimulation.cpp
@@ -218,7 +218,6 @@ void PhysicalEntitySimulation::getObjectsToAddToPhysics(VectorOfMotionStates& re
             ShapeInfo shapeInfo;
             entity->computeShapeInfo(shapeInfo);
             int numPoints = shapeInfo.getMaxNumPoints();
-            const int MAX_HULL_POINTS = 42;
             if (numPoints > MAX_HULL_POINTS) {
                 glm::vec3 p = entity->getPosition();
                 qWarning().nospace() << "convex hull with " << numPoints
--- a/libraries/physics/src/PhysicalEntitySimulation.cpp
+++ b/libraries/physics/src/PhysicalEntitySimulation.cpp
@@ -219,11 +219,9 @@ void PhysicalEntitySimulation::getObjectsToAddToPhysics(VectorOfMotionStates& re
             entity->computeShapeInfo(shapeInfo);
             int numPoints = shapeInfo.getMaxNumPoints();
             if (numPoints > MAX_HULL_POINTS) {
-                glm::vec3 p = entity->getPosition();
-                qWarning().nospace() << "convex hull with " << numPoints
-                    << " points for entity " << entity->getName()
-                    << " at <" << p.x << ", " << p.y << ", " << p.z << ">"
-                    << " will be reduced";
+                qWarning() << "convex hull with" << numPoints
+                    << "points for entity" << entity->getName()
+                    << "at" << entity->getPosition() << " will be reduced";
             }
             btCollisionShape* shape = ObjectMotionState::getShapeManager()->getShape(shapeInfo);
             if (shape) {
--- a/libraries/physics/src/PhysicalEntitySimulation.cpp
+++ b/libraries/physics/src/PhysicalEntitySimulation.cpp
@@ -217,6 +217,13 @@ void PhysicalEntitySimulation::getObjectsToAddToPhysics(VectorOfMotionStates& re
         } else if (entity->isReadyToComputeShape()) {
             ShapeInfo shapeInfo;
             entity->computeShapeInfo(shapeInfo);
+            int numPoints = shapeInfo.getMaxNumPoints();
+            const int MAX_HULL_POINTS = 42;
+            if (numPoints > MAX_HULL_POINTS) {
+                glm::vec3 p = entity->getPosition();
+                qDebug() << "entity" << entity->getName() << "at <" << p.x << p.y << p.z << ">"
+                    << "has convex hull with" << numPoints << "points";
+            }
             btCollisionShape* shape = ObjectMotionState::getShapeManager()->getShape(shapeInfo);
             if (shape) {
                 EntityMotionState* motionState = new EntityMotionState(shape, entity);
--- a/libraries/physics/src/PhysicalEntitySimulation.cpp
+++ b/libraries/physics/src/PhysicalEntitySimulation.cpp
@@ -221,8 +221,10 @@ void PhysicalEntitySimulation::getObjectsToAddToPhysics(VectorOfMotionStates& re
             const int MAX_HULL_POINTS = 42;
             if (numPoints > MAX_HULL_POINTS) {
                 glm::vec3 p = entity->getPosition();
-                qDebug() << "entity" << entity->getName() << "at <" << p.x << p.y << p.z << ">"
-                    << "has convex hull with" << numPoints << "points";
+                qWarning().nospace() << "convex hull with " << numPoints
+                    << " points for entity " << entity->getName()
+                    << " at <" << p.x << ", " << p.y << ", " << p.z << ">"
+                    << " will be reduced";
             }
             btCollisionShape* shape = ObjectMotionState::getShapeManager()->getShape(shapeInfo);
             if (shape) {
--- a/libraries/physics/src/ShapeFactory.cpp
+++ b/libraries/physics/src/ShapeFactory.cpp
@@ -70,14 +70,15 @@ btConvexHullShape* ShapeFactory::createConvexHull(const QVector<glm::vec3>& poin
         // create hull approximation
         btShapeHull shapeHull(hull);
         shapeHull.buildHull(margin);
+        // we cannot copy Bullet shapes so we must create a new one...
         btConvexHullShape* newHull = new btConvexHullShape();
         const btVector3* newPoints = shapeHull.getVertexPointer();
         for (int i = 0; i < shapeHull.numVertices(); ++i) {
             newHull->addPoint(newPoints[i], false);
         }
+        // ...and delete the old one
         delete hull;
         hull = newHull;
-        qDebug() << "reduced hull with" << points.size() << "points down to" << hull->getNumPoints(); // TODO: remove after testing
     }
 
     hull->recalcLocalAabb();
--- a/libraries/physics/src/ShapeFactory.cpp
+++ b/libraries/physics/src/ShapeFactory.cpp
@@ -66,18 +66,16 @@ btConvexHullShape* ShapeFactory::createConvexHull(const QVector<glm::vec3>& poin
         hull->addPoint(btVector3(correctedPoint[0], correctedPoint[1], correctedPoint[2]), false);
     }
 
-    const int MAX_HULL_POINTS = 42;
     if (points.size() > MAX_HULL_POINTS) {
         // create hull approximation
-        btShapeHull* shapeHull = new btShapeHull(hull);
-        shapeHull->buildHull(margin);
+        btShapeHull shapeHull(hull);
+        shapeHull.buildHull(margin);
         btConvexHullShape* newHull = new btConvexHullShape();
-        const btVector3* newPoints = shapeHull->getVertexPointer();
-        for (int i = 0; i < shapeHull->numVertices(); ++i) {
+        const btVector3* newPoints = shapeHull.getVertexPointer();
+        for (int i = 0; i < shapeHull.numVertices(); ++i) {
             newHull->addPoint(newPoints[i], false);
         }
         delete hull;
-        delete shapeHull;
         hull = newHull;
         qDebug() << "reduced hull with" << points.size() << "points down to" << hull->getNumPoints(); // TODO: remove after testing
     }
--- a/libraries/physics/src/ShapeFactory.cpp
+++ b/libraries/physics/src/ShapeFactory.cpp
@@ -10,6 +10,7 @@
 //
 
 #include <glm/gtx/norm.hpp>
+#include <BulletCollision/CollisionShapes/btShapeHull.h>
 
 #include <SharedUtil.h> // for MILLIMETERS_PER_METER
 
@@ -64,6 +65,23 @@ btConvexHullShape* ShapeFactory::createConvexHull(const QVector<glm::vec3>& poin
         correctedPoint = (points[i] - center) * relativeScale + center;
         hull->addPoint(btVector3(correctedPoint[0], correctedPoint[1], correctedPoint[2]), false);
     }
+
+    const int MAX_HULL_POINTS = 42;
+    if (points.size() > MAX_HULL_POINTS) {
+        // create hull approximation
+        btShapeHull* shapeHull = new btShapeHull(hull);
+        shapeHull->buildHull(margin);
+        btConvexHullShape* newHull = new btConvexHullShape();
+        const btVector3* newPoints = shapeHull->getVertexPointer();
+        for (int i = 0; i < shapeHull->numVertices(); ++i) {
+            newHull->addPoint(newPoints[i], false);
+        }
+        delete hull;
+        delete shapeHull;
+        hull = newHull;
+        qDebug() << "reduced hull with" << points.size() << "points down to" << hull->getNumPoints(); // TODO: remove after testing
+    }
+
     hull->recalcLocalAabb();
     return hull;
 }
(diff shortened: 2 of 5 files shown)


libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -217,6 +217,13 @@ void PhysicalEntitySimulation::getObjectsToAddToPhysics(VectorOfMotionStates& re
         } else if (entity->isReadyToComputeShape()) {
             ShapeInfo shapeInfo;
             entity->computeShapeInfo(shapeInfo);
+            int numPoints = shapeInfo.getMaxNumPoints();
+            const int MAX_HULL_POINTS = 42;
```
zzmp: This should be exposed as a member of a class (maybe ShapeInfo?) so it does not need redefinition and risk different definitions.

libraries/physics/src/ShapeFactory.cpp
```diff
@@ -64,6 +65,23 @@ btConvexHullShape* ShapeFactory::createConvexHull(const QVector<glm::vec3>& poin
         correctedPoint = (points[i] - center) * relativeScale + center;
         hull->addPoint(btVector3(correctedPoint[0], correctedPoint[1], correctedPoint[2]), false);
     }
+
+    const int MAX_HULL_POINTS = 42;
```
zzmp: This should be exposed as a member of a class (maybe ShapeInfo?) so it does not need redefinition and risk different definitions.

libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -217,6 +217,13 @@ void PhysicalEntitySimulation::getObjectsToAddToPhysics(VectorOfMotionStates& re
...
+            if (numPoints > MAX_HULL_POINTS) {
+                glm::vec3 p = entity->getPosition();
+                qDebug() << "entity" << entity->getName() << "at <" << p.x << p.y << p.z << ">"
```
zzmp: If possible, this should be closer to where the reduction takes place.

libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -217,6 +217,13 @@ void PhysicalEntitySimulation::getObjectsToAddToPhysics(VectorOfMotionStates& re
...
+                    << "has convex hull with" << numPoints << "points";
+            }
```
AndrewMeadows: This debug log may help us identify assets that provide convex hull with unreasonably larger numbers of points.  Perhaps someday we'll track down just how such hulls are constructed and eliminate them at the source.  Downloading all of those useless points is a waste of CPU, memory, and bandwidth.

libraries/physics/src/PhysicalEntitySimulation.cpp
zzmp: This should be a `qWarning`, and should tell the user the consequence of loading something with so many points (i.e. "Entity X at <x, y, z> has more than <MAX> points in hull and will be reduced").

zzmp: The coordinates will be formatted strangely - instead, use `qDebug().nospace()` and add the explicit formatting.

libraries/shared/src/ShapeInfo.cpp
```diff
@@ -99,6 +99,18 @@ uint32_t ShapeInfo::getNumSubShapes() const {
     }
     return 1;
 }
+
+int ShapeInfo::getMaxNumPoints() const {
+    int numPoints = 0;
+    for (int i = 0; i < _points.size(); ++i) {
```
zzmp: Why not use a range-based loop?

libraries/physics/src/ShapeFactory.cpp
```diff
@@ -64,6 +65,23 @@ btConvexHullShape* ShapeFactory::createConvexHull(const QVector<glm::vec3>& poin
...
+        shapeHull->buildHull(margin);
+        btConvexHullShape* newHull = new btConvexHullShape();
+        const btVector3* newPoints = shapeHull->getVertexPointer();
+        for (int i = 0; i < shapeHull->numVertices(); ++i) {
+            newHull->addPoint(newPoints[i], false);
+        }
+        delete hull;
+        delete shapeHull;
+        hull = newHull;
+        qDebug() << "reduced hull with" << points.size() << "points down to" << hull->getNumPoints(); // TODO: remove after testing
```
zzmp: Should this be a warning? Do you want this to log multiple times for a single shape?

hifi-gustavo:
Build console output for commit: e0d73a5f8a8733e7971c1697514c3029ef2bc27e available at: https://gist.github.com/257e9d621e0f6e5d4083d5fd54bd7f05<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.dmg

AndrewMeadows:
To test:

(1) Login to playa, or any other domain with interesting collidable models
(2) Check the logs and notice lines that look something like this:
```
[05/18 23:20:03] [DEBUG] entity "playa_model_dumptruck" at < 0.875902 0.916185 0.119535 > has convex hull with 856 points
[05/18 23:20:03] [DEBUG] reduced hull with 856 points down to 42
```
(3) Find some of the objects that were trimmed and verify that they still collide correctly.

libraries/physics/src/ShapeFactory.cpp
```diff
@@ -64,6 +65,23 @@ btConvexHullShape* ShapeFactory::createConvexHull(const QVector<glm::vec3>& poin
         correctedPoint = (points[i] - center) * relativeScale + center;
         hull->addPoint(btVector3(correctedPoint[0], correctedPoint[1], correctedPoint[2]), false);
     }
+
+    const int MAX_HULL_POINTS = 42;
+    if (points.size() > MAX_HULL_POINTS) {
+        // create hull approximation
+        btShapeHull* shapeHull = new btShapeHull(hull);
```
zzmp: Make `shapeHull` and `newHull` on the stack to avoid raw pointers, and then move `newHull` into `*hull`.

libraries/physics/src/ShapeFactory.cpp
AndrewMeadows: This information is not particularly interesting, except for testing this PR to verify that the new code path really is being used.  I'll remove it after testing.

libraries/shared/src/ShapeInfo.cpp
AndrewMeadows: We're looping over a `std::vector< std::vector< glm::vec3 > >`, so a list of lists of vec3's.  I'm looking for the list that has the longest length.

libraries/physics/src/PhysicalEntitySimulation.cpp
AndrewMeadows: No, this is in the right spot.  I want to identify the specific entity that has the offending convex hulls.  I'm not actually interested in logging the details of the reduction.

I can change it to a warning and add some verbage about what it means.

Sigh... I didn't like the explicit formatting which is why I did it this way, but I see that someone appears to have intentionally added the type gobbledygook that was poking me in the eye.  I guess I'll settle for pretty code and ugly log entry.

hifi-gustavo:
Build console output for commit: 0c278708b27a5da44edbcce16b130bd75ab49330 available at: https://gist.github.com/257e9d621e0f6e5d4083d5fd54bd7f05<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.dmg

libraries/physics/src/PhysicalEntitySimulation.cpp
ZappoMan: How did we come up with this number? Other than being fans of HHGTTG

libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -217,6 +217,15 @@ void PhysicalEntitySimulation::getObjectsToAddToPhysics(VectorOfMotionStates& re
         } else if (entity->isReadyToComputeShape()) {
             ShapeInfo shapeInfo;
             entity->computeShapeInfo(shapeInfo);
+            int numPoints = shapeInfo.getMaxNumPoints();
+            const int MAX_HULL_POINTS = 42;
+            if (numPoints > MAX_HULL_POINTS) {
+                glm::vec3 p = entity->getPosition();
```
ZappoMan: coding standard: variable names should be descriptive

libraries/physics/src/PhysicalEntitySimulation.cpp
```diff
@@ -217,6 +217,15 @@ void PhysicalEntitySimulation::getObjectsToAddToPhysics(VectorOfMotionStates& re
...
+                qWarning().nospace() << "convex hull with " << numPoints
+                    << " points for entity " << entity->getName()
+                    << " at <" << p.x << ", " << p.y << ", " << p.z << ">"
```
ZappoMan: you don't need to do this... the vec3 already has a streaming operator to stream directly

libraries/physics/src/ShapeFactory.cpp
```diff
@@ -64,6 +65,21 @@ btConvexHullShape* ShapeFactory::createConvexHull(const QVector<glm::vec3>& poin
...
+        for (int i = 0; i < shapeHull.numVertices(); ++i) {
+            newHull->addPoint(newPoints[i], false);
+        }
+        delete hull;
+        hull = newHull;
+        qDebug() << "reduced hull with" << points.size() << "points down to" << hull->getNumPoints(); // TODO: remove after testing
+    }
+
     hull->recalcLocalAabb();
     return hull;
```
AndrewMeadows: This code is a recipe I stole from this Bullet wiki page:
http://www.bulletphysics.org/mediawiki-1.5.8/index.php?title=BtShapeHull_vertex_reduction_utility

Both `btConvexHullShape` instances **must** be created on the heap because Bullet does not support copying of shapes (*), however the `btShapeHull` has been made local variable.

(*) The class derivation hierarchy of btConvexHullShape can be found in the link below.  I started following the code back and confirmed my suspicion... at least one of the classes in the stack does special allocations and does not support copying: `btPolyhedralConvexShape`.

http://www.continuousphysics.com/Bullet/BulletFull/classbtConvexHullShape.html

libraries/physics/src/PhysicalEntitySimulation.cpp
AndrewMeadows: I didn't like the format of `QDebug& operator<<(QDebug& dbg, const glm::vec3& v)` for this log so I formatted things explicitly.

hifi-gustavo:
Build console output for commit: b76669511e058a61a66ebec0bb72eea65db6d577 available at: https://gist.github.com/257e9d621e0f6e5d4083d5fd54bd7f05<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.dmg

hifi-gustavo:
Build console output for commit: b447b9be34e83c9d1fbb7cfbf1bc12efba4effd5 available at: https://gist.github.com/257e9d621e0f6e5d4083d5fd54bd7f05<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.dmg

libraries/physics/src/PhysicalEntitySimulation.cpp
ZappoMan: why don't you fix the formatting then... as the person who wrote the original formatting... I really don't care what the format is.

libraries/shared/src/ShapeInfo.cpp
zzmp: Even so, why not loop to hide some of the complexity:
```
for (const auto& point : points) {
    numPoints = std::max(numPoints, points.size());
}
```

libraries/physics/src/PhysicalEntitySimulation.cpp
AndrewMeadows: I did fix the formatting... for this particular log.  I wouldn't presume to change the qDebug format of vec3 for all of the other existing logs out there -- I assume the verbose format has a purpose.  Specifically I'm talking about the following two formats:

{type='glm::vec3' x=1.0000, y=2.000, z=3.0000}
<1.0000, 2.0000, 3.0000>

I admit the complexity of this code snippet does provide a "wtf?!" moment for future readers.  I could probably make the code a little cleaner (but not any faster) by first streaming the vector to a string and then feeding that to the qDebug() object.

ZappoMan: I mean change the formatting in the shared code to be what you want... and use the shared code. No one will complain that the shared code's formatting changed. People will probably applaud you.

libraries/physics/src/ShapeFactory.cpp
zzmp: Might be worth noting this as a comment in case someone tries to "fix" it later.
Thanks for the explanation!

OmegaHeron:
@AndrewMeadows - does this have implications for stack side or Interface only?  If it impacts stack side I want to test Linux stack build as part of my QA.

OmegaHeron:
All standard tests pass.
Even see a minor reduction of CPU load at standard home scene (1-2%)
One crash on close recorded/sent via bugsplat.

AndrewMeadows:
This work only affects the Bullet simulation so only has client-side implications.

OmegaHeron:
QA Pass for me - I saw no change to physics interaction while comparing this to RC at Playa, Standard home scene or my own content where I'd definitely notice any change to compound shape interactions.

libraries/shared/src/StreamUtils.cpp
```diff
@@ -69,54 +69,37 @@ QDataStream& operator>>(QDataStream& in, glm::quat& quaternion) {
...
 }
 
 QDebug& operator<<(QDebug& dbg, const glm::quat& q) {
-    dbg.nospace() << "{type='glm::quat'"
-        ", x=" << q.x <<
-        ", y=" << q.y <<
-        ", z=" << q.z <<
-        ", w=" << q.w <<
-        "}";
+    dbg.nospace() << '(' << q.x << ", " << q.y << ", " << q.z << ", " << q.w << ')';
```
AndrewMeadows: A quick informal poll suggested that the preferred formating for vec3 and friends would be: (x, y, z).

Note some quaternion API's use (w, x, y, z) (such as glm::quat) and others use (x, y, z, w) (such as Bullet).  When it comes to printing things out we're going with the latter, mostly because that is how I'm familiar with it, but also because that was the old (explicitly labelled) order.

hifi-gustavo:
Build console output for commit: 5615e66d35c63e054b9f87a64303ce2369dcde3b available at: https://gist.github.com/257e9d621e0f6e5d4083d5fd54bd7f05<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.dmg

hifi-gustavo:
Build console output for commit: 83b8ef8131a6d44b40d8e35b73921c789f08c6e3 available at: https://gist.github.com/257e9d621e0f6e5d4083d5fd54bd7f05<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.dmg

zzmp:
@AndrewMeadows why not use the range-based loop in ShapeInfo.cpp?
I am OK with dev approving, but still think that could expose less complexity.

hifi-gustavo:
Build console output for commit: 2e30f0916cc9a365b5a93171527499e22f5385cb available at: https://gist.github.com/257e9d621e0f6e5d4083d5fd54bd7f05<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7934.dmg

PR #7961, opened 2016-05-23
vhacd-util improvements

AndrewMeadows:
* speedup when processing FBX meshes
* generate more correct convex hulls for some shapes
* add -v (verbose) option (defaults to running silent)
* enable progress callback

### DIFF
--- a/tools/vhacd-util/src/VHACDUtil.cpp
+++ b/tools/vhacd-util/src/VHACDUtil.cpp
@@ -87,19 +87,12 @@ void getTrianglesInMeshPart(const FBXMeshPart &meshPart, std::vector<int>& trian
     }
 }
 
-
-void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
-                                    uint32_t& meshPartCount,
-                                    uint32_t startMeshIndex, uint32_t endMeshIndex) const {
+void vhacd::VHACDUtil::fattenMesh(const FBXMesh& mesh, const glm::mat4& geometryOffset, FBXMesh& result) const {
     // this is used to make meshes generated from a highfield collidable.  each triangle
     // is converted into a tetrahedron and made into its own mesh-part.
 
     std::vector<int> triangleIndices;
     foreach (const FBXMeshPart &meshPart, mesh.parts) {
-        if (meshPartCount < startMeshIndex || meshPartCount >= endMeshIndex) {
-            meshPartCount++;
-            continue;
-        }
         getTrianglesInMeshPart(meshPart, triangleIndices);
     }
 
@@ -107,10 +100,13 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
         return;
     }
 
+    int indexStartOffset = result.vertices.size();
+
     // new mesh gets the transformed points from the original
+    glm::mat4 totalTransform = geometryOffset * mesh.modelTransform;
     for (int i = 0; i < mesh.vertices.size(); i++) {
         // apply the source mesh's transform to the points
-        glm::vec4 v = mesh.modelTransform * glm::vec4(mesh.vertices[i], 1.0f);
+        glm::vec4 v = totalTransform * glm::vec4(mesh.vertices[i], 1.0f);
         result.vertices += glm::vec3(v);
     }
 
@@ -118,7 +114,6 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
 
     const uint32_t TRIANGLE_STRIDE = 3;
     const float COLLISION_TETRAHEDRON_SCALE = 0.25f;
-    int indexStartOffset = result.vertices.size();
     for (uint32_t i = 0; i < triangleIndices.size(); i += TRIANGLE_STRIDE) {
         int index0 = triangleIndices[i] + indexStartOffset;
         int index1 = triangleIndices[i + 1] + indexStartOffset;
@@ -341,8 +336,9 @@ bool vhacd::VHACDUtil::computeVHACD(FBXGeometry& geometry,
 
         // each mesh has its own transform to move it to model-space
         std::vector<glm::vec3> vertices;
+        glm::mat4 totalTransform = geometry.offset * mesh.modelTransform;
         foreach (glm::vec3 vertex, mesh.vertices) {
-            vertices.push_back(glm::vec3(mesh.modelTransform * glm::vec4(vertex, 1.0f)));
+            vertices.push_back(glm::vec3(totalTransform * glm::vec4(vertex, 1.0f)));
         }
         uint32_t numVertices = (uint32_t)vertices.size();
 
--- a/tools/vhacd-util/src/VHACDUtil.cpp
+++ b/tools/vhacd-util/src/VHACDUtil.cpp
@@ -47,7 +47,7 @@ bool vhacd::VHACDUtil::loadFBX(const QString filename, FBXGeometry& result) {
         } else if (filename.toLower().endsWith(".fbx")) {
             geom = readFBX(fbxContents, QVariantHash(), filename);
         } else {
-            qWarning() << "unknown file extension";
+            qWarning() << "file has unknown extension" << filename;
             return false;
         }
         result = *geom;
@@ -63,18 +63,20 @@ bool vhacd::VHACDUtil::loadFBX(const QString filename, FBXGeometry& result) {
 
 
 void getTrianglesInMeshPart(const FBXMeshPart &meshPart, std::vector<int>& triangleIndices) {
-    // append all the triangles (and converted quads) from this mesh-part to triangles
-    std::vector<int> meshPartTriangles = meshPart.triangleIndices.toStdVector();
-    triangleIndices.insert(triangleIndices.end(), meshPartTriangles.begin(), meshPartTriangles.end());
+    // append triangle indices
+    triangleIndices.reserve(triangleIndices.size() + (size_t)meshPart.triangleIndices.size());
+    for (auto index : meshPart.triangleIndices) {
+        triangleIndices.push_back(index);
+    }
 
     // convert quads to triangles
     const uint32_t QUAD_STRIDE = 4;
-    uint32_t quadCount = meshPart.quadIndices.size() / QUAD_STRIDE;
-    for (uint32_t i = 0; i < quadCount; i++) {
-        uint32_t p0Index = meshPart.quadIndices[i * QUAD_STRIDE];
-        uint32_t p1Index = meshPart.quadIndices[i * QUAD_STRIDE + 1];
-        uint32_t p2Index = meshPart.quadIndices[i * QUAD_STRIDE + 2];
-        uint32_t p3Index = meshPart.quadIndices[i * QUAD_STRIDE + 3];
+    uint32_t numIndices = (uint32_t)meshPart.quadIndices.size();
+    for (uint32_t i = 0; i < numIndices; i += QUAD_STRIDE) {
+        uint32_t p0Index = meshPart.quadIndices[i];
+        uint32_t p1Index = meshPart.quadIndices[i + 1];
+        uint32_t p2Index = meshPart.quadIndices[i + 2];
+        uint32_t p3Index = meshPart.quadIndices[i + 3];
         // split each quad into two triangles
         triangleIndices.push_back(p0Index);
         triangleIndices.push_back(p1Index);
@@ -105,8 +107,6 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
         return;
     }
 
-    int indexStartOffset = result.vertices.size();
-
     // new mesh gets the transformed points from the original
     for (int i = 0; i < mesh.vertices.size(); i++) {
         // apply the source mesh's transform to the points
@@ -118,6 +118,7 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
 
     const uint32_t TRIANGLE_STRIDE = 3;
     const float COLLISION_TETRAHEDRON_SCALE = 0.25f;
+    int indexStartOffset = result.vertices.size();
     for (uint32_t i = 0; i < triangleIndices.size(); i += TRIANGLE_STRIDE) {
         int index0 = triangleIndices[i] + indexStartOffset;
         int index1 = triangleIndices[i + 1] + indexStartOffset;
@@ -178,26 +179,39 @@ AABox getAABoxForMeshPart(const FBXMesh& mesh, const FBXMeshPart &meshPart) {
     return aaBox;
 }
 
-struct TriangleEdge {
-    int indexA { -1 };
-    int indexB { -1 };
+class TriangleEdge {
+public:
     TriangleEdge() {}
-    TriangleEdge(int A, int B) : indexA(A), indexB(B) {}
-    bool operator==(const TriangleEdge& other) const {
-        return indexA == other.indexA && indexB == other.indexB;
+    TriangleEdge(uint32_t A, uint32_t B) {
+        setIndices(A, B);
     }
-    void sortIndices() {
-        if (indexB < indexA) {
-            std::swap(indexA, indexB);
+    void setIndices(uint32_t A, uint32_t B) {
+        if (A < B) {
+            _indexA = A;
+            _indexB = B;
+        } else {
+            _indexA = B;
+            _indexB = A;
         }
     }
+    bool operator==(const TriangleEdge& other) const {
+        return _indexA == other._indexA && _indexB == other._indexB;
+    }
+
+    uint32_t getIndexA() const { return _indexA; }
+    uint32_t getIndexB() const { return _indexB; }
+private:
+    uint32_t _indexA { (uint32_t)(-1) };
+    uint32_t _indexB { (uint32_t)(-1) };
 };
 
 namespace std {
     template <>
     struct hash<TriangleEdge> {
         std::size_t operator()(const TriangleEdge& edge) const {
-            return (hash<int>()(edge.indexA) ^ (hash<int>()(edge.indexB) << 1));
+            // use Cantor's pairing function to generate a hash of ZxZ --> Z
+            uint32_t ab = edge.getIndexA() + edge.getIndexB();
+            return hash<int>()((ab * (ab + 1)) / 2 + edge.getIndexB());
         }
     };
 }
@@ -212,14 +226,12 @@ bool isClosedManifold(const std::vector<int>& triangleIndices) {
     for (uint32_t i = 0; i < triangleIndices.size(); i += TRIANGLE_STRIDE) {
         TriangleEdge edge;
         // the triangles indices are stored in sequential order
-        for (int j = 0; j < 3; ++j) {
-            edge.indexA = triangleIndices[(int)i + j];
-            edge.indexB = triangleIndices[i + ((j + 1) % 3)];
-            edge.sortIndices();
+        for (uint32_t j = 0; j < 3; ++j) {
+            edge.setIndices(triangleIndices[i + j], triangleIndices[i + ((j + 1) % 3)]);
 
             EdgeList::iterator edgeEntry = edges.find(edge);
             if (edgeEntry == edges.end()) {
-                edges.insert(std::pair<TriangleEdge, int>(edge, 1));
+                edges.insert(std::pair<TriangleEdge, uint32_t>(edge, 1));
             } else {
                 edgeEntry->second += 1;
             }
@@ -243,6 +255,7 @@ void vhacd::VHACDUtil::getConvexResults(VHACD::IVHACD* convexifier, FBXMesh& res
 
     // create an output meshPart for each convex hull
     const uint32_t TRIANGLE_STRIDE = 3;
+    const uint32_t POINT_STRIDE = 3;
     for (uint32_t j = 0; j < numHulls; j++) {
         VHACD::IVHACD::ConvexHull hull;
         convexifier->GetConvexHull(j, hull);
@@ -251,20 +264,21 @@ void vhacd::VHACDUtil::getConvexResults(VHACD::IVHACD* convexifier, FBXMesh& res
         FBXMeshPart& resultMeshPart = resultMesh.parts.last();
 
         int hullIndexStart = resultMesh.vertices.size();
-        for (uint32_t i = 0; i < hull.m_nPoints; i++) {
-            float x = hull.m_points[i * TRIANGLE_STRIDE];
-            float y = hull.m_points[i * TRIANGLE_STRIDE + 1];
-            float z = hull.m_points[i * TRIANGLE_STRIDE + 2];
+        resultMesh.vertices.reserve(hullIndexStart + hull.m_nPoints);
+        uint32_t numIndices = hull.m_nPoints * POINT_STRIDE;
+        for (uint32_t i = 0; i < numIndices; i += POINT_STRIDE) {
+            float x = hull.m_points[i];
+            float y = hull.m_points[i + 1];
+            float z = hull.m_points[i + 2];
             resultMesh.vertices.append(glm::vec3(x, y, z));
         }
 
-        for (uint32_t i = 0; i < hull.m_nTriangles; i++) {
-            int index0 = hull.m_triangles[i * TRIANGLE_STRIDE] + hullIndexStart;
-            int index1 = hull.m_triangles[i * TRIANGLE_STRIDE + 1] + hullIndexStart;
-            int index2 = hull.m_triangles[i * TRIANGLE_STRIDE + 2] + hullIndexStart;
-            resultMeshPart.triangleIndices.append(index0);
-            resultMeshPart.triangleIndices.append(index1);
-            resultMeshPart.triangleIndices.append(index2);
+        numIndices = hull.m_nTriangles * TRIANGLE_STRIDE;
+        resultMeshPart.triangleIndices.reserve(resultMeshPart.triangleIndices.size() + numIndices);
+        for (uint32_t i = 0; i < numIndices; i += TRIANGLE_STRIDE) {
+            resultMeshPart.triangleIndices.append(hull.m_triangles[i] + hullIndexStart);
+            resultMeshPart.triangleIndices.append(hull.m_triangles[i + 1] + hullIndexStart);
+            resultMeshPart.triangleIndices.append(hull.m_triangles[i + 2] + hullIndexStart);
         }
         if (_verbose) {
             qDebug() << "    hull" << j << " vertices =" << hull.m_nPoints
@@ -419,7 +433,7 @@ bool vhacd::VHACDUtil::computeVHACD(FBXGeometry& geometry,
                 index = dupeIndexMap[index];
             }
 
-            // this time we don't care if the parts are close or not
+            // this time we don't care if the parts are closed or not
             uint32_t triangleCount = (uint32_t)(triangleIndices.size()) / TRIANGLE_STRIDE;
             if (_verbose) {
                 qDebug() << "  process remaining open parts =" << openParts.size() << ": "
--- a/tools/vhacd-util/src/VHACDUtil.cpp
+++ b/tools/vhacd-util/src/VHACDUtil.cpp
@@ -424,7 +424,7 @@ void vhacd::ProgressCallback::Update(const double overallProgress,
     int progress = (int)(overallProgress + 0.5);
 
     std::cout << "\b\b\b";
-    std::cout << progress << "%";
+    std::cout << progress << "%" << std::flush;
     if (progress >= 100) {
         std::cout << std::endl;
     }
--- a/tools/vhacd-util/src/VHACDUtil.cpp
+++ b/tools/vhacd-util/src/VHACDUtil.cpp
@@ -9,6 +9,7 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 
+#include <unordered_map>
 #include <QVector>
 #include "VHACDUtil.h"
 
@@ -49,7 +50,7 @@ bool vhacd::VHACDUtil::loadFBX(const QString filename, FBXGeometry& result) {
 
         reSortFBXGeometryMeshes(result);
     } catch (const QString& error) {
-        qDebug() << "Error reading " << filename << ": " << error;
+        qDebug() << "Error reading" << filename << ":" << error;
         return false;
     }
 
@@ -57,13 +58,12 @@ bool vhacd::VHACDUtil::loadFBX(const QString filename, FBXGeometry& result) {
 }
 
 
-unsigned int getTrianglesInMeshPart(const FBXMeshPart &meshPart, std::vector<int>& triangles) {
+void getTrianglesInMeshPart(const FBXMeshPart &meshPart, std::vector<int>& triangles) {
     // append all the triangles (and converted quads) from this mesh-part to triangles
     std::vector<int> meshPartTriangles = meshPart.triangleIndices.toStdVector();
     triangles.insert(triangles.end(), meshPartTriangles.begin(), meshPartTriangles.end());
 
     // convert quads to triangles
-    unsigned int triangleCount = meshPart.triangleIndices.size() / 3;
     unsigned int quadCount = meshPart.quadIndices.size() / 4;
     for (unsigned int i = 0; i < quadCount; i++) {
         unsigned int p0Index = meshPart.quadIndices[i * 4];
@@ -77,10 +77,7 @@ unsigned int getTrianglesInMeshPart(const FBXMeshPart &meshPart, std::vector<int
         triangles.push_back(p0Index);
         triangles.push_back(p2Index);
         triangles.push_back(p3Index);
-        triangleCount += 2;
     }
-
-    return triangleCount;
 }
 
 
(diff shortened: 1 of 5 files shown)


tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -177,134 +177,263 @@ AABox getAABoxForMeshPart(const FBXMesh& mesh, const FBXMeshPart &meshPart) {
...
+        }
+    }
+    // scan for outside edge
+    for (auto& edgeEntry : edges) {
+        if (edgeEntry.second == 1) {
+             return false;
+        }
+    }
+    return true;
+}
```
AndrewMeadows: We can now test meshes to see if they are "closed".  V-HACD is faster and more accurate for closed meshes.

OmegaHeron:
General smoke test - Linux (Ubuntu 16.04 64 Bit - QT5.6.0)
Compiles and runs fine.  Composed a few simple shapes and saw no issue with output obj files.  Would probably be worthwhile to test on some more complex shapes, but, will wait for any specific tests.

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -177,134 +177,263 @@ AABox getAABoxForMeshPart(const FBXMesh& mesh, const FBXMeshPart &meshPart) {
...
+            resultMeshPart.triangleIndices.append(index1);
+            resultMeshPart.triangleIndices.append(index2);
+        }
+        if (_verbose) {
+            qDebug() << "    hull" << j << " vertices =" << hull.m_nPoints
+                << " triangles =" << hull.m_nTriangles
+                << " FBXMeshVertices =" << resultMesh.vertices.size();
+        }
+    }
+}
```
AndrewMeadows: The code that converts the V-HACD convex hulls into FBXGeometry format has been pulled into this method.

hifi-gustavo:
Build console output for commit: 56f3f6ee026a1b130e6a1b886f2a22de4346f708 available at: https://gist.github.com/739b8dd05a69190ed744b0d1cb12e020<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.dmg

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -177,134 +177,263 @@ AABox getAABoxForMeshPart(const FBXMesh& mesh, const FBXMeshPart &meshPart) {
...
-                qDebug() << "V-HACD computation failed for Mesh : " << count;
-                count++;
-                continue;
+                // compute approximate convex decomposition
+                bool success = convexifier->Compute(&vertices[0].x, 3, (uint)numVertices, &triangles[0], 3, triangleCount, params);
+                if (success) {
+                    getConvexResults(convexifier, resultMesh);
+                } else if (_verbose) {
+                    qDebug() << "  failed to convexify";
+                }
```
AndrewMeadows: New flow: if we detect a closed mesh part, then we go ahead and convexify it (above).  Otherwise we save its index onto a list (below) and eventually merge all open parts together in the hopes that the welded result is closed -- often this works out.

tools/vhacd-util/src/VHACDUtilApp.cpp
```diff
@@ -329,10 +340,14 @@ VHACDUtilApp::VHACDUtilApp(int argc, char* argv[]) :
...
-        vhacd::ProgressCallback pCallBack;
+        vhacd::ProgressCallback progressCallback;
 
         //set parameters for V-HACD
-        params.m_callback = &pCallBack; //progress callback
+        if (verbose) {
+            params.m_callback = &progressCallback; //progress callback
+        } else {
+            params.m_callback = nullptr;
+        }
```
AndrewMeadows: Add the **-v** option at the command line to make the vhacd-util tool verbose.

tools/vhacd-util/src/main.cpp
```diff
@@ -23,5 +23,5 @@ using namespace VHACD;
 
 int main(int argc, char * argv[]) {
     VHACDUtilApp app(argc, argv);
-    return 0;
+    return app.getReturnCode();
```
AndrewMeadows: The vhacd-util will now return non-zero to the shell on failure.

hifi-gustavo:
Build console output for commit: 4564a252d4e74b1a0f7b5f6eb731255b0509d7ea available at: https://gist.github.com/739b8dd05a69190ed744b0d1cb12e020<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.dmg

AndrewMeadows:
To test:

(1) You must be able to build the codebase on linux or MacOS.  This work relates to the standalone **vhacd-util** command-line executable which is NOT distributed with the downloadable packages.  Note, it appears that the vhacd-util tool does not work on windows.

(2) Download some FBX files to test.  Here is a list of a few from the 'playa' domain with the approximate times it takes the old/new **vhacd-util** to convexify them with default parameters on my laptop.

http://hifi-content.s3.amazonaws.com/DomainContent/production/garden/gnome_blue.fbx
old/new times = 1.99251  /  2.15254
old/new hulls = 6 / 6

http://hifi-content.s3.amazonaws.com/caitlyn/dev/Blueprint%20Objects/GreatOutdoors/Potted_Tree/Potted_Tree.fbx
old/new times = 358.944  /  152.856
old/new hulls = 159 / 59

http://hifi-public.s3.amazonaws.com/models/ping_pong_gun/target.fbx
old/new times = 10.2444  /  0.275341
old/new hulls = 43 / 1

Note the discrepancy between the times:  the gnome actually takes longer in the new code.  This is because the gnome is a very clean single mesh that is completely closed, and the new checks and cleanup don't help.  Meanwhile the tree and target have multiple sub parts that are not closed.  The new code will close the target completely and some parts of the tree.

(3) Run the vhacd-util manually against the models like so (linux and MacOS):

old:  vhacd-util -g --maxvertices 42 -i input.fbx -o output.obj
new: vhacd-util -v -g --maxvertices 42 -i input.fbx -o output.obj

Examine the results using some OBJ viewer and verify that they are more or less correct using the new code.

OmegaHeron:
Issue on windows is you have to manually copy some DLL files to directory where vhacd-util is compiled.  I've almost exclusively used it on Linux as it easier to work with from there, but, briefly, jumped through hoops to make it work on Windows.  Too much effort for zero return.

hifi-gustavo:
Build console output for commit: 04c0261d92ee69225774adeb2578650f98e92dd9 available at: https://gist.github.com/739b8dd05a69190ed744b0d1cb12e020<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.dmg

OmegaHeron:
QA Pass
Tested (Ubuntu 16.04 QT5.6.0) against items above and a few of my own objects - my results for objects referenced in test match test notes.  My unique tests produce compound shapes matching my expectations of previous vhacd output.  Feeding it bad data leads to exit status reflecting that so now at long last I can use it reliably in automated cases.  :+1:

OmegaHeron:
On a side note - I made a mistake and exported an FBX that had a 512x512x32M sky island AND a previously generated vhacd compound - the compound was on another layer.  Feeding that "bad" export into old vhacd-util produced a crash while it, after a **very long processing time**, actually produced a fairly sane output with this PR!

OmegaHeron:
There's still an outstanding issue where units of measure aren't properly handled here.  This is complex, but, once again, here goes.
Take a model that's 512x512x32M in Blender.  Blender is set to units = Metric.  Unit Scale = 1.0. Export that as obj - feed it to vhacd-util and import generated obj back to Blender.  Notice it is 100X too large.

Now - set Blender to operate in CM.  This is done by setting units of measure to Metric and Unit Scale to 0.01
Export that as FBX - run through VHACD and re-import to Blender.  Note that OBJ now is same apparent size as original mesh.  Sounds good, right?  **Wrong**

Apply that compound to a model in HF and it will create a compound that is 100X size.

To make a compound from blender that actually works you have to set units to Metric - Unit scale = 1.0 then scale object to be processed to 0.01 (This is object scale NOT unit scale) - do not apply scale transform and export.  Run that through VHACD and import it back to Blender.  Set original object scale back to 1,1,1 and not now that original and compound have same sizes.  Using that compound shape in HF yields a shape that is proper.  I've only been bringing this up a few months after VHACD was added and a change was made that yielded this outcome.  Basically as I see it - it's reading the FBX and wanting CM but somehow isn't taking units of measure as CM, but, instead, object scale = 0.01 as CM - then exporting a model that uses same object scale as original.  Coming into HF it seems to do a reversal of this leading to the need to play tricks.  I've been using vhacd-util since it was added and it took many many many hours to sort all this out.

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -177,134 +177,263 @@ AABox getAABoxForMeshPart(const FBXMesh& mesh, const FBXMeshPart &meshPart) {
...
+    int indexA { -1 };
+    int indexB { -1 };
+    TriangleEdge() {}
+    TriangleEdge(int A, int B) : indexA(A), indexB(B) {}
+    bool operator==(const TriangleEdge& other) const {
+        return indexA == other.indexA && indexB == other.indexB;
+    }
+    void sortIndices() {
+        if (indexB < indexA) {
+            int t = indexA;
```
ZappoMan: wouldn't `std::swap()` be cleaner?

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -177,134 +177,263 @@ AABox getAABoxForMeshPart(const FBXMesh& mesh, const FBXMeshPart &meshPart) {
...
+    };
+}
+
+// returns false if any edge has only one adjacent triangle
+bool isClosedManifold(const std::vector<int>& triangles) {
+    using EdgeList = std::unordered_map<TriangleEdge, int>;
+    EdgeList edges;
+
+    // count the triangles for each edge
+    for (size_t i = 0; i < triangles.size(); i += 3) {
```
ZappoMan: coding standard: no magic.

On this one, I'm sure you're going to roll your eyes and say "damnit Brad! it's a triangle!" -- but here's the thing... there are multiple things about a triangle that comes in threes... vertices, edges, scalars in a vertex... I think it makes the code clearer if we use a named variable that tells us what we're iterating over.

I assume this is vertices per triangle... but that's also confusing because the variable you're iterating over is called "triangle"... but it's probably not an array of "triangle" objects... but instead an array of vertices??

ZappoMan: reading closer, I guess triangles is an array of indices of vertices?

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -177,134 +177,263 @@ AABox getAABoxForMeshPart(const FBXMesh& mesh, const FBXMeshPart &meshPart) {
...
+    for (unsigned int j = 0; j < numHulls; j++) {
+        VHACD::IVHACD::ConvexHull hull;
+        convexifier->GetConvexHull(j, hull);
+
+        resultMesh.parts.append(FBXMeshPart());
+        FBXMeshPart& resultMeshPart = resultMesh.parts.last();
+
+        int hullIndexStart = resultMesh.vertices.size();
+        for (unsigned int i = 0; i < hull.m_nPoints; i++) {
+            float x = hull.m_points[i * 3];
```
ZappoMan: here's another example... I think the 3 here is the number of scalars per point... right? let's use a variable name to make that more clear.

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -177,134 +177,263 @@ AABox getAABoxForMeshPart(const FBXMesh& mesh, const FBXMeshPart &meshPart) {
...
+        if (_verbose) {
+            qDebug() << "    hull" << j << " vertices =" << hull.m_nPoints
+                << " triangles =" << hull.m_nTriangles
+                << " FBXMeshVertices =" << resultMesh.vertices.size();
+        }
+    }
+}
 
+float computeDt(uint64_t start) {
+    return (float)(usecTimestampNow() - start) / 1.0e6f;
```
ZappoMan: magic... USECS_PER_SECOND?

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -177,134 +177,263 @@ AABox getAABoxForMeshPart(const FBXMesh& mesh, const FBXMeshPart &meshPart) {
...
+                    qDebug() << "  skip part" << partIndex << ":  dimension =" << largestDimension << "(too large)";
+                }
+                ++partIndex;
                 continue;
             }
 
+            // figure out if the mesh is a closed manifold or not
+            bool closed = isClosedManifold(triangles);
+            if (closed) {
+                unsigned int triangleCount = (unsigned int)(triangles.size()) / 3;
```
ZappoMan: VERTICES_PER_TRIANGLE? or POINTS_PER_VERTEX? or EDGES_PER_TRIANGLE?

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -177,134 +177,263 @@ AABox getAABoxForMeshPart(const FBXMesh& mesh, const FBXMeshPart &meshPart) {
...
+                }
 
-            // compute approximate convex decomposition
-            bool res = interfaceVHACD->Compute(&vertices[0].x, 3, (uint)nPoints, &triangles[0], 3, triangleCount, params);
-            if (!res){
-                qDebug() << "V-HACD computation failed for Mesh : " << count;
-                count++;
-                continue;
+                // compute approximate convex decomposition
+                bool success = convexifier->Compute(&vertices[0].x, 3, (uint)numVertices, &triangles[0], 3, triangleCount, params);
```
ZappoMan: VERTICES_PER_TRIANGLE? or POINTS_PER_VERTEX? or EDGES_PER_TRIANGLE?

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -177,134 +177,263 @@ AABox getAABoxForMeshPart(const FBXMesh& mesh, const FBXMeshPart &meshPart) {
...
-                for (unsigned int i = 0; i < hull.m_nTriangles; i++) {
-                    int index0 = hull.m_triangles[i * 3] + hullIndexStart;
-                    int index1 = hull.m_triangles[i * 3 + 1] + hullIndexStart;
-                    int index2 = hull.m_triangles[i * 3 + 2] + hullIndexStart;
-                    resultMeshPart.triangleIndices.append(index0);
-                    resultMeshPart.triangleIndices.append(index1);
-                    resultMeshPart.triangleIndices.append(index2);
-                }
+            // this time we don't care if the parts are close or not
+            unsigned int triangleCount = (unsigned int)(triangles.size()) / 3;
```
ZappoMan: VERTICES_PER_TRIANGLE? or POINTS_PER_VERTEX? or EDGES_PER_TRIANGLE?

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -177,134 +177,263 @@ AABox getAABoxForMeshPart(const FBXMesh& mesh, const FBXMeshPart &meshPart) {
...
+            if (_verbose) {
+                qDebug() << "  process remaining open parts =" << openParts.size() << ": "
+                    << " triangles =" << triangleCount;
             }
 
-            count++;
+            // compute approximate convex decomposition
+            bool success = convexifier->Compute(&vertices[0].x, 3, (uint)numVertices, &triangles[0], 3, triangleCount, params);
```
ZappoMan: VERTICES_PER_TRIANGLE? or POINTS_PER_VERTEX? or EDGES_PER_TRIANGLE?

tools/vhacd-util/src/VHACDUtilApp.cpp
```diff
@@ -301,17 +310,19 @@ VHACDUtilApp::VHACDUtilApp(int argc, char* argv[]) :
...
     if (!vUtil.loadFBX(inputFilename, fbx)){
-        cout << "Error in opening FBX file....";
+        _returnCode = VHACD_RETURN_CODE_FAILURE_TO_READ;
+        return;
     }
     auto end = std::chrono::high_resolution_clock::now();
     auto loadDuration = std::chrono::duration_cast<std::chrono::nanoseconds>(end - begin).count();
 
+    if (verbose) {
+        qDebug() << "load time =" << (double)loadDuration / 1000000000.00 << "seconds";
```
ZappoMan: USECS_PER_SECOND

tools/vhacd-util/src/VHACDUtil.cpp
AndrewMeadows: Yes, I'm definitely rolling my eyes for this one, but not necessarily some of the others.  Everything here is about triangles so a raw 3 should not confuse anyone.  Personally I would find an INDICES_PER_TRIANGLE to be slightly confusing since it obviously should be 3 but obfuscation of the true value (did the writer really define it as 3? I'd have to verify) makes it harder to read.  I'll make the change in the editor and consider whether  I can stomach it or appreciate the humor of some future reader coming across it and wondering why I bothered to spell it out, and if so I'll make the change.

ZappoMan: I guess then make the variable be `triangleVertices` at least then it's more clear you're operating on the verttices, not the edges or the scalars of the vertices... you have 3 used elsewhere in this file where it's one of those other things...

My point is not that people don't know that a triangle has 3 vertices... but that you can't tell at first glance what this thing is doing...

If I read the code out loud, it sounds like it's walking over an array of triangles and skipping 2 out of 3 of them for some unclear reason.

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -115,10 +116,12 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
...
-    for (unsigned int i = 0; i < triangleCount; i++) {
-        int index0 = triangles[i * 3] + indexStartOffset;
-        int index1 = triangles[i * 3 + 1] + indexStartOffset;
-        int index2 = triangles[i * 3 + 2] + indexStartOffset;
+    const uint32_t TRIANGLE_STRIDE = 3;
+    const float COLLISION_TETRAHEDRON_SCALE = 0.25f;
+    for (uint32_t i = 0; i < triangleIndices.size(); i += TRIANGLE_STRIDE) {
+        int index0 = triangleIndices[i] + indexStartOffset;
+        int index1 = triangleIndices[i + 1] + indexStartOffset;
+        int index2 = triangleIndices[i + 2] + indexStartOffset;
```
AndrewMeadows: I've changed the variable name from `triangles` to `triangleIndices` for more clarity.  I also use the variable `TRIANGLE_STRIDE` and eliminate some math inside the loop.  Note that the +0, +1, and +2 magic numbers in this loop exist because triangles have three vertices and their indices happen to be packed adjacent to each other -- I feel there is no need to explain this to future intrepid code readers.

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -155,156 +158,290 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
...
+
+    // count the triangles for each edge
+    const uint32_t TRIANGLE_STRIDE = 3;
+    for (uint32_t i = 0; i < triangleIndices.size(); i += TRIANGLE_STRIDE) {
+        TriangleEdge edge;
+        // the triangles indices are stored in sequential order
+        for (int j = 0; j < 3; ++j) {
+            edge.indexA = triangleIndices[(int)i + j];
+            edge.indexB = triangleIndices[i + ((j + 1) % 3)];
+            edge.sortIndices();
```
AndrewMeadows: The code is assuming (requiring) that the triangle indices start every three integers however external code could try to interleave the triangle indices with other data in which case a different stride would be required.  Even if this were done triangles would still have three vertices, so the two raw 3's above stay -- they are fundamental.

hifi-gustavo:
Build console output for commit: b0b683cc7a38fee708e38d412640ab74194651ff available at: https://gist.github.com/739b8dd05a69190ed744b0d1cb12e020<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.dmg

ZappoMan:
```
C:\jenkins\workspace\beta_pr-master\label\windows\tools\vhacd-util\src\VHACDUtil.cpp(333): warning C4267: 'initializing' : conversion from 'size_t' to 'uint32_t', possible loss of data [C:\jenkins\workspace\beta_pr-master\label\windows\build\tools\vhacd-util\vhacd-util.vcxproj]

```

hifi-gustavo:
Build console output for commit: 8e34a31da72f652b0ee07d323dacabc857a50ae0 available at: https://gist.github.com/739b8dd05a69190ed744b0d1cb12e020<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.dmg

AndrewMeadows:
@OmegaHeron I created a fogbugz bug entry for the "units of measure" problem you described here.  We'll fix it in a future PR.

OmegaHeron:
Thanks @AndrewMeadows - it's likely also related to the gymnastics one has to do to get collision capsule to match on Avatars as well.  It's a similar mechanism.

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -42,65 +47,61 @@ bool vhacd::VHACDUtil::loadFBX(const QString filename, FBXGeometry& result) {
         } else if (filename.toLower().endsWith(".fbx")) {
             geom = readFBX(fbxContents, QVariantHash(), filename);
         } else {
-            qDebug() << "unknown file extension";
+            qWarning() << "unknown file extension";
```
zzmp: It may be useful (especially if this tool will be integrated later) to give the filename with all warnings so the point of failure is more obvious.

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -42,65 +47,61 @@ bool vhacd::VHACDUtil::loadFBX(const QString filename, FBXGeometry& result) {
...
         return false;
     }
 
     return true;
 }
 
 
-unsigned int getTrianglesInMeshPart(const FBXMeshPart &meshPart, std::vector<int>& triangles) {
+void getTrianglesInMeshPart(const FBXMeshPart &meshPart, std::vector<int>& triangleIndices) {
     // append all the triangles (and converted quads) from this mesh-part to triangles
```
zzmp: /s/triangles/triangleIndices

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -42,65 +47,61 @@ bool vhacd::VHACDUtil::loadFBX(const QString filename, FBXGeometry& result) {
...
     std::vector<int> meshPartTriangles = meshPart.triangleIndices.toStdVector();
```
zzmp: This is an extra copy (not a blocker).

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -42,65 +47,61 @@ bool vhacd::VHACDUtil::loadFBX(const QString filename, FBXGeometry& result) {
...
         }
-        getTrianglesInMeshPart(meshPart, triangles);
+        getTrianglesInMeshPart(meshPart, triangleIndices);
     }
 
-    auto triangleCount = triangles.size() / 3;
-    if (triangleCount == 0) {
+    if (triangleIndices.size() == 0) {
         return;
     }
```
zzmp: indexStartOffset isn't used until you turn triangles into a tetrahedron, it can be moved down.

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -42,65 +47,61 @@ bool vhacd::VHACDUtil::loadFBX(const QString filename, FBXGeometry& result) {
...
-    unsigned int triangleCount = meshPart.triangleIndices.size() / 3;
-    unsigned int quadCount = meshPart.quadIndices.size() / 4;
-    for (unsigned int i = 0; i < quadCount; i++) {
-        unsigned int p0Index = meshPart.quadIndices[i * 4];
-        unsigned int p1Index = meshPart.quadIndices[i * 4 + 1];
-        unsigned int p2Index = meshPart.quadIndices[i * 4 + 2];
-        unsigned int p3Index = meshPart.quadIndices[i * 4 + 3];
+    const uint32_t QUAD_STRIDE = 4;
+    uint32_t quadCount = meshPart.quadIndices.size() / QUAD_STRIDE;
+    for (uint32_t i = 0; i < quadCount; i++) {
```
zzmp: You should use the same construct here as you do for the triangles in L121:
```
i += QUAD_STRIDE
```
and then not have to multiply by `QUAD_STRIDE` 4 times every iteration.

tools/vhacd-util/src/VHACDUtil.cpp
zzmp: You should duplicate this strategy in L74 (noted there too).

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -155,156 +158,290 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
...
-
-
 AABox getAABoxForMeshPart(const FBXMesh& mesh, const FBXMeshPart &meshPart) {
     AABox aaBox;
-    unsigned int triangleCount = meshPart.triangleIndices.size() / 3;
-    for (unsigned int i = 0; i < triangleCount; ++i) {
-        aaBox += mesh.vertices[meshPart.triangleIndices[i * 3]];
-        aaBox += mesh.vertices[meshPart.triangleIndices[i * 3 + 1]];
-        aaBox += mesh.vertices[meshPart.triangleIndices[i * 3 + 2]];
+    const int TRIANGLE_STRIDE = 3;
```
zzmp: `TRIANGLE_STRIDE` and `QUAD_STRIDE` are used enough, and general enough, that you can probably have them be global `const`s.

zzmp: Why does this function use `int` when other functions use `uint32_t`?

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -155,156 +158,290 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
...
+        aaBox += mesh.vertices[meshPart.quadIndices[i]];
+        aaBox += mesh.vertices[meshPart.quadIndices[i + 1]];
+        aaBox += mesh.vertices[meshPart.quadIndices[i + 2]];
+        aaBox += mesh.vertices[meshPart.quadIndices[i + 3]];
     }
 
     return aaBox;
 }
 
+struct TriangleEdge {
```
zzmp: Use a `class` instead of a `struct` (coding standards).

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -155,156 +158,290 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
...
+    int indexA { -1 };
+    int indexB { -1 };
+    TriangleEdge() {}
+    TriangleEdge(int A, int B) : indexA(A), indexB(B) {}
+    bool operator==(const TriangleEdge& other) const {
+        return indexA == other.indexA && indexB == other.indexB;
+    }
+    void sortIndices() {
```
zzmp: Instead of a `sortIndices` method, which might be forgotten, can you make an accessor for the indices as a `const std::array<int, 2>&` so that it actively guards against getting them out of order?

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -155,156 +158,290 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
...
+            std::swap(indexA, indexB);
+        }
+    }
+};
 
+namespace std {
+    template <>
+    struct hash<TriangleEdge> {
+        std::size_t operator()(const TriangleEdge& edge) const {
+            return (hash<int>()(edge.indexA) ^ (hash<int>()(edge.indexB) << 1));
```
zzmp: Is there a reason this works as a hash? This is not obvious to me.

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -155,156 +158,290 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
...
+bool isClosedManifold(const std::vector<int>& triangleIndices) {
+    using EdgeList = std::unordered_map<TriangleEdge, int>;
+    EdgeList edges;
+
+    // count the triangles for each edge
+    const uint32_t TRIANGLE_STRIDE = 3;
+    for (uint32_t i = 0; i < triangleIndices.size(); i += TRIANGLE_STRIDE) {
+        TriangleEdge edge;
+        // the triangles indices are stored in sequential order
+        for (int j = 0; j < 3; ++j) {
```
zzmp: If you make this a `uint32_t` you can avoid the explicit cast back to `int` below.

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -155,156 +158,290 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
...
+    const uint32_t TRIANGLE_STRIDE = 3;
+    for (uint32_t j = 0; j < numHulls; j++) {
+        VHACD::IVHACD::ConvexHull hull;
+        convexifier->GetConvexHull(j, hull);
+
+        resultMesh.parts.append(FBXMeshPart());
+        FBXMeshPart& resultMeshPart = resultMesh.parts.last();
+
+        int hullIndexStart = resultMesh.vertices.size();
+        for (uint32_t i = 0; i < hull.m_nPoints; i++) {
```
zzmp: `i += TRIANGLE_STRIDE`

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -155,156 +158,290 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
...
+            float x = hull.m_points[i * TRIANGLE_STRIDE];
+            float y = hull.m_points[i * TRIANGLE_STRIDE + 1];
+            float z = hull.m_points[i * TRIANGLE_STRIDE + 2];
+            resultMesh.vertices.append(glm::vec3(x, y, z));
+        }
+
+        for (uint32_t i = 0; i < hull.m_nTriangles; i++) {
```
zzmp: `i += TRIANGLE_STRIDE`

tools/vhacd-util/src/VHACDUtilApp.cpp
```diff
@@ -329,10 +341,14 @@ VHACDUtilApp::VHACDUtilApp(int argc, char* argv[]) :
 
     if (generateHulls) {
         VHACD::IVHACD::Parameters params;
-        vhacd::ProgressCallback pCallBack;
+        vhacd::ProgressCallback progressCallback;
```
zzmp: Are there plans to expose this to the caller, instead of just having it internal?

tools/vhacd-util/src/VHACDUtilApp.cpp
```diff
@@ -301,17 +310,20 @@ VHACDUtilApp::VHACDUtilApp(int argc, char* argv[]) :
...
-        cout << "Error in opening FBX file....";
+        _returnCode = VHACD_RETURN_CODE_FAILURE_TO_READ;
+        return;
     }
     auto end = std::chrono::high_resolution_clock::now();
-    auto loadDuration = std::chrono::duration_cast<std::chrono::nanoseconds>(end - begin).count();
 
+    if (verbose) {
+        auto loadDuration = std::chrono::duration_cast<std::chrono::nanoseconds>(end - begin).count();
+        const double NANOSECS_PER_SECOND = 1.0e9;
```
zzmp: This should come from NUMERIC_CONSTANTS.h

tools/vhacd-util/src/VHACDUtilApp.cpp
```diff
@@ -18,6 +18,9 @@
 using namespace std;
 using namespace VHACD;
 
+const int VHACD_RETURN_CODE_FAILURE_TO_READ = 1;
```
zzmp: Can you put these in the header so it is obvious to someone using the code (with the source) what their failure means?

tools/vhacd-util/src/VHACDUtil.cpp
```diff
@@ -155,156 +158,290 @@ void vhacd::VHACDUtil::fattenMeshes(const FBXMesh& mesh, FBXMesh& result,
...
 
-                for (unsigned int i = 0; i < hull.m_nTriangles; i++) {
-                    int index0 = hull.m_triangles[i * 3] + hullIndexStart;
-                    int index1 = hull.m_triangles[i * 3 + 1] + hullIndexStart;
-                    int index2 = hull.m_triangles[i * 3 + 2] + hullIndexStart;
-                    resultMeshPart.triangleIndices.append(index0);
-                    resultMeshPart.triangleIndices.append(index1);
-                    resultMeshPart.triangleIndices.append(index2);
-                }
+            // this time we don't care if the parts are close or not
```
zzmp: closed

tools/vhacd-util/src/VHACDUtil.cpp
AndrewMeadows: QVector<T>::size() returns an int (whereas std::vector<T>::size() returns size_t... which on linux64 is uint64_t).

tools/vhacd-util/src/VHACDUtilApp.cpp
AndrewMeadows: No.  This isn't being used anywhere else yet, and might never be.

tools/vhacd-util/src/VHACDUtilApp.cpp
AndrewMeadows: Yes, we'll be using this to more amazing effect in the Glorius Futcher.

AndrewMeadows:
@OmegaHeron: after so many changes this needs to be retested.

hifi-gustavo:
Build console output for commit: 402b7f22822b80d2c72f25e7bdb7d708d6ea28d5 available at: https://gist.github.com/739b8dd05a69190ed744b0d1cb12e020<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.dmg

OmegaHeron:
Will be first on my A.M. list.  US East coast here :)

tools/vhacd-util/src/VHACDUtilApp.cpp
zzmp: I'm pretty due nsecs per usec and usecs per sec are used elsewhere (together). Not a blocker though.

OmegaHeron:
QA Pass
Re-test vhacd-util on Ubuntu 16.04 + QT5.6.0
Same tests pass as before.

tools/vhacd-util/src/VHACDUtil.cpp
AndrewMeadows: No,  `indexStartOffset` needs to be grabbed before the loop below it where new vertices are pushed onto the result.

AndrewMeadows:
@OmegaHeron, I think I've fixed the bug where **vhacd-util** incorrectly scales the vertices of blender FBX models.  Also, if you could test the -f option that would be great -- it treats the mesh as a heightfield and computes a tetrahedron for each triangle, where the tetrahedron's new tip is pulled down away from the normal of the face.

hifi-gustavo:
Build console output for commit: 1eb0b6a2315da8b1954b9bffcf8a60ac049222d9 available at: https://gist.github.com/739b8dd05a69190ed744b0d1cb12e020<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7961.dmg

OmegaHeron:
@AndrewMeadows - I'll take a look this evening.  Added todo.

OmegaHeron:
@AndrewMeadows Fantastic :)

Using a blender model where world scale = metric = 1.0 (1 Unit = 1 Meter) and exporting to FBX -> process with VHACD-Util leads to a compound shape that works proper.  It should be noted that importing this obj file to blender leads one to believe they've generated an incorrect model as it's a fraction of source model size.  **But, this is correct.** Blender imports obj under incorrect assumptions leading to this.  Compound shape applied in HF is great.

Using same command line parameters as specified adding -f results in the best terrain compound shape I've ever generated for sky-island.  Previously I had to "massage" vhacd-util output to fix it up.

So much :+1:

OmegaHeron:
Also of note - my terrain compound shape is a fraction of complexity of previous vhacd-util generated + manual rework and is superior.  Tests without -f on various models result in good results as well.

PR #7962, opened 2016-05-24
fix throttling when Interface is not in focus

zzmp:

PR #7964, opened 2016-05-24
check for all protocols matching when connecting to a domain

ZappoMan:
- add "protocol version signature" to the domain connection request to verify that the client connecting has matching protocols
- show a warning message if the protocols don't match
- handle the case of connecting to an old domain to still get a warning message
- handle the case of old clients connecting to the new domain, so they will get a warning message
- include a "reason code" in the DomainConnectionDenied message, so we can more easily programatically determine the reason for the connection denied and respond accordingly
- don't show a login dialog box for cases like "protocol mismatch" or "too many users" which don't require login/authentication


hifi-gustavo:
Build console output for commit: 5668db9e4a6f54e4971fe34f9abce516f365b867 available at: https://gist.github.com/373f800b92916471e3c62e3a11a41573<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7964.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7964.dmg

hifi-gustavo:
Build console output for commit: 183f38e4f06c28fa69bcd4dfba5bcc4b669c3d65 available at: https://gist.github.com/373f800b92916471e3c62e3a11a41573<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7964.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7964.dmg

ZappoMan:
Note: This changes the client and the server protocols, so for smoke testing, you will need to test against a common server/client install.

Testing Plan:
- General Smoke Test - everything should work fine.
- Connect **new client** to **old domain** - you should get a "The location you are visiting is running an incompatible server version. Content may not display correctly."
- Connect **old client** to **new domain** - you should get a "The location you are visiting is running an incompatible server version. Content may not display correctly."
- From the developer menu - choose Developer > Networking > Send wrong protocol version - 
    - reconnect to the domain (quit/restart, or go to other domain/return)
    - you should get a "The location you are visiting is running an incompatible server version. Content may not display correctly."

domain-server/src/DomainGatekeeper.cpp
```diff
@@ -516,16 +530,19 @@ void DomainGatekeeper::publicKeyJSONCallback(QNetworkReply& requestReply) {
...
 
-void DomainGatekeeper::sendConnectionDeniedPacket(const QString& reason, const HifiSockAddr& senderSockAddr) {
+void DomainGatekeeper::sendConnectionDeniedPacket(const QString& reason, const HifiSockAddr& senderSockAddr, 
+            DomainHandler::ConnectionRefusedReason reasonCode) {
     // this is an agent and we've decided we won't let them connect - send them a packet to deny connection
     QByteArray utfString = reason.toUtf8();
     quint16 payloadSize = utfString.size();
     
     // setup the DomainConnectionDenied packet
-    auto connectionDeniedPacket = NLPacket::create(PacketType::DomainConnectionDenied, payloadSize + sizeof(payloadSize));
```
Atlante45: Can you keep and update the size here?
That way we only allocate the space we need instead of ~1500 bytes.

ZappoMan: sure... no problem.

interface/src/Application.cpp
```diff
@@ -1062,6 +1064,24 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
...
+            break;
+        case DomainHandler::ConnectionRefusedReason::TooManyUsers:
+        case DomainHandler::ConnectionRefusedReason::Unknown: {
+            QString message = "Unable to connect to the location you are visiting.\n";
+            message += reasonMessage;
+            OffscreenUi::warning("", message);
+            break;
+        }
+        default:
+            // nothing to do.
```
Atlante45: Couldn't we at least print the message if there's one?

interface/src/Application.cpp
```diff
@@ -4572,6 +4592,17 @@ void Application::setSessionUUID(const QUuid& sessionUUID) const {
     Physics::setSessionUUID(sessionUUID);
 }
 
+
+// If we're not getting anything back from the domain server checkin, it might be that the domain speaks an 
+// older version of the DomainConnectRequest protocal. We will attempt to send and older version of DomainConnectRequest.
```
Atlante45: Non-Blocking, typo: `protocal`

domain-server/src/DomainGatekeeper.cpp
```diff
@@ -370,7 +378,8 @@ bool DomainGatekeeper::verifyUserSignature(const QString& username,
             } else {
                 if (!senderSockAddr.isNull()) {
                     qDebug() << "Error decrypting username signature for " << username << "- denying connection.";
-                    sendConnectionDeniedPacket("Error decrypting username signature.", senderSockAddr);
+                    sendConnectionDeniedPacket("Error decrypting username signature.", senderSockAddr,
```
jherico: Why send the reason enum *and* the error string?  Just send the enum and let the client convert it to a string if and when it displays the reason.  This has the added bonus of being one less thing to fix when we eventually do localization.

interface/src/Application.cpp
ZappoMan: it actually already does print a message and handle the login box... that behavior is handled by the NodeList/DomainManager... this extra behavior is just for doing "extra" stuff at the application layer.

domain-server/src/DomainGatekeeper.cpp
ZappoMan: the error string has additional context that can be different for the same error code.

libraries/networking/src/udt/PacketHeaders.cpp
```diff
@@ -58,9 +60,13 @@ PacketVersion versionForPacketType(PacketType packetType) {
         case PacketType::AssetUpload:
```
Atlante45: I think you need to bump the PackeType::EntityData a few lines up.
Otherwise the trick where you send an EntityData packet to tell the version don't match won't work.

ZappoMan: ah yes - I was only testing against beta clients... but yes. if I bump the EntityData it will also work with dev builds. Will do.

hifi-gustavo:
Build console output for commit: fb9f6185db724c0fc3f30ec46ad70353e3a848de available at: https://gist.github.com/373f800b92916471e3c62e3a11a41573<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7964.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7964.dmg

libraries/networking/src/udt/PacketHeaders.cpp
ZappoMan: done

libraries/networking/src/udt/PacketHeaders.cpp
```diff
@@ -47,7 +49,7 @@ PacketVersion versionForPacketType(PacketType packetType) {
         case PacketType::EntityAdd:
         case PacketType::EntityEdit:
         case PacketType::EntityData:
-            return VERSION_ENTITIES_NO_FLY_ZONES;
+            return VERSION_ENTITIES_MORE_SHAPES;
```
ZappoMan: turns out this wasn't properly bumped in Austin's PR... so I just bumped it here. This will look like a new version to all beta or dev builds prior to this.

hifi-gustavo:
Build console output for commit: ff45633c21ca3f9a496fa620e5fb7490968af874 available at: https://gist.github.com/373f800b92916471e3c62e3a11a41573<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7964.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7964.dmg

Nex-Pro:
@ZappoMan keep me posted plz.

ZappoMan:
@nex-pro it's ready to be tested

Nex-Pro:
@ZappoMan  cool

Nex-Pro:
@ZappoMan same build as before ?

ZappoMan:
@Nex-Pro - there was a new build 15 hours ago... right above here... that's the one to use.

Nex-Pro:
ok

Nex-Pro:
@ZappoMan  Testing this with Michelle.  

Initially all points above  succeeded. ( ***apart from smoke test, not attempted*** ) 

But on further repeated attempts to connect to Michelle's #7964 server and dev-demo  I got no content. And occasionally, incorrectly,  the wrong version pop-up.

Also in my local #7964 server  collisions with cubes did not occur.( believe this is a known issue ) 

I was able to consistently connect to dev-demo  with PR #7962 and content would load.

I reset settings. Cleared cache. Deleted Cache, data8 and prepared.

I reset my router.

re-installed the PR     None of which improved the situation.

Noticed this in the log:

**[05/25 19:51:23] [DEBUG] OctreePacketProcessor - piggyback packet version mismatch on 41 (EntityData) - Sender QUuid({4b43ad8a-697f-47b6-8bc9-e3a6a4aff651}) sent 58 but 59 expected.**

Nex-Pro:
That log message was while connected to dev-demo

ZappoMan:
@Nex-Pro - dev-demo is running an old version (compared to this PR) so I would expect you to get the the "incompatible version" message.

Nex-Pro:
ah !

Nex-Pro:
@ZappoMan  Just to be clear I was able to connect to Michelle's 7964 server with this build and see content, everything loaded.   It was subsequent attempts that failed.

ZappoMan:
@Nex-Pro - I'm kind of lost in what you're saying your experience was. Can you please present your findings in a simple table.

Client PR7964 to Server PR7964 ---- should work normally
Client PR7964 to **any other server** --- should get "incompatible protocol message
**Any other client** to Server PR7964 --- should get "incompatible protocol message

Nex-Pro:
Client PR7964 to Server PR7964 ----  worked normally
Client PR7964 to any other server --- got "incompatible protocol message
Any other client to Server PR7964 --- got "incompatible protocol message

Michelle also installed this pr client and server. After some delay of several minutes the content showed on my client when connected to michelle.

That was fine for a while. 

But after multiple further attempts  no content loaded on connecting with michelle.

we tried everything including deleting everthing and reinstalling but didn't fix it.

Michelle was not able to connect to my local server of this pr.

Thoys has joined in this test.

Thoys was able to connect. to me before promptly crashing a second or two after I did. We sent you bug splats.

Thoys was subsequently unable to connect.

I restarted the server whilst I was logged in to it.  Then I had to go to sandbox and then go home in order to reconnect.

Also Thoys also mentioned that he got bumped out of my server the first time he connected in the same way I got bumped from Michelle's on the first occasion.
Neither of us had seen that before.

ZappoMan:
Does the domain in question have shape primitives like boxes/spheres --
it's possible there was a bug in master that contributed to your problems.

I have since updated with the latest.

Retest might be in order.

ZappoMan:
@Nex-Pro the two crashes in bugsplat are in totally different pieces of code that are not touched by this PR... I don't think they are related. This PR only changes the very initial connect behavior... I think you're running into independent issues.

ZappoMan:
@Nex-Pro - I do see the following in the log files of the crashes....
```
[05/25 22:37:38] [DEBUG] UNEXPECTED: parsing of the octal code would make UNREASONABLY_DEEP_RECURSION... numberOfThreeBitSectionsInStream: 128 This buffer is corrupt. Returning.
[05/25 22:37:38] [DEBUG] UNEXPECTED: readElementDataFromBuffer() only had  1  bytes before masks. Not enough for meaningful data.
[05/25 22:37:39] [DEBUG] UNEXPECTED: readElementDataFromBuffer() only had  1  bytes before masks. Not enough for meaningful data.
[05/25 22:37:39] [DEBUG] UNEXPECTED: readElementDataFromBuffer() only had  1  bytes before masks. Not enough for meaningful data.
```
This is very likely related to a bug that was introduced in master and is now fixed... once this PR rebuilds (it's pending now)... I suspect your problems will be addressed.

Nex-Pro:
cool

hifi-gustavo:
Build console output for commit: 84f541285fdecd6380c2a8b442ed8cbbcdb5c3cc available at: https://gist.github.com/373f800b92916471e3c62e3a11a41573<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7964.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7964.dmg

hifi-gustavo:
Build console output for commit: 78b6a6c6612d8ad8563e4fa07bc8cbca6faa4825 available at: https://gist.github.com/373f800b92916471e3c62e3a11a41573<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7964.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7964.dmg

kevinmthomas-carpool:
Log for @Nex-Pro report.
[hifi-log.txt](https://github.com/highfidelity/hifi/files/284077/hifi-log.txt)

ZappoMan:
@kevinmthomas-carpool - is there a QA report?

Nex-Pro:
@ZappoMan @Atlante45 

Grab a coffee....

Connect new client to old domain - you should get a "The location you are visiting is running an incompatible server version. Content may not display correctly.": **Partial success see notes.**
Connect old client to new domain - you should get a "The location you are visiting is running an incompatible server version. Content may not display correctly.": **SUCCESS**

From the developer menu - choose Developer > Networking > Send wrong protocol version -
reconnect to the domain (quit/restart, or go to other domain/return)
you should get a "The location you are visiting is running an incompatible server version. Content may not display correctly.": **FAIL**
make sure to uncheck that option after this test **done**

Test the following domain authentication features:

NOTE: these must be done from a client not running on the same machine as the server/domain. **CHECK.**

domain with unrestricted access - should work fine, should work for unlogged in or logged in case: **SUCCESS**

domain with restricted access - should pop up login when you connect without being logged in: **SUCCESS**

test logging in with an account that is not allowed: **SUCCESS** ***account not on list correctly blocked.***

Test the "Maximum User Capacity" feature:: **Partial success see notes.**

NOTE: these must be done from a client not running on the same machine as the server/domain. **CHECK.**

Set "Max User Capacity" to 1 : **DONE**
Log in from two machines : **DONE**

Verify you get a popup message that informs you that the server is over capacity: **SUCCESS**


**Notes:**

Would not automatically reconnect after server was restarted whilst in the server. Had to restart client to reconnect.

After getting the server full message, on Kevin's server, clicking go home fails first time it is clicked, second click works. ( happened several times. )

Thoys, Kevin and I all noticed an odd disconnect where we were bumped back out to the previous location after several seconds.

Connecting to dev-playa after sandbox does not issue the compatibility warning.

Returning to a location that previously issued a warning does not produce another.

Clicking on go home when at sandbox or dev-playa can trigger a behaviour where you are moved to localhost for approx 0.25 seconds then pinged back to the location you where previously at.
Similar to the events seen by Thoys, Kevin and I but much faster if it happens when connecting locally.

I'll leave the smoke test until you advise further.

Also the compatibility warning is about 7 seconds slower to appear then it is with the old client.

**Restricted client numbers:** 

	When set to 3 Kevin, Michelle and I could all be in Kevin's server.
	When set to 1 Michelle and I were unable to get in.

Nex-Pro:
I need a vacation.

domain-server/src/DomainGatekeeper.cpp
AndrewMeadows: When we do localization we'll have to expand the set of error codes to span the zoology of possible error conditions.

PR #7979, opened 2016-05-27
Fix crash on shutdown of SteamVR

jherico:


jherico:
## Testing

* Run interface
* Run Steam VR
* Ensure OpenVR input or display plugins are enabled 
* Exit Steam VR

In the, this will cause the app to crash.  In this build you should no longer crash.  If the OpenVR display plugin is enabled, you should automatically switch back to the 2D desktop display mode.

jherico:
Fixes https://highfidelity.fogbugz.com/f/cases/548/exit-from-steamvr-crashes-Interface

hifi-gustavo:
Build console output for commit: ac5912df01493075122b7d9680adedf5c0a5ed56 available at: https://gist.github.com/7faaf7c657c7a2fabd227caceb545f9d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7979.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7979.dmg

interface/src/Application.cpp
```diff
@@ -5098,9 +5104,17 @@ void Application::updateDisplayMode() {
...
             QObject::connect(displayPlugin.get(), &DisplayPlugin::recommendedFramebufferSizeChanged, [this](const QSize & size) {
                 resizeGL();
             });
+            QObject::connect(displayPlugin.get(), &DisplayPlugin::outputDeviceLost, [this, displayPluginName] {
+                PluginManager::getInstance()->disableDisplayPlugin(displayPluginName);
+                auto menu = Menu::getInstance();
+                if (menu->menuItemExists(MenuOption::OutputMenu, displayPluginName)) {
+                    menu->removeMenuItem(MenuOption::OutputMenu, displayPluginName);
+                }
+            });
```
AndrewMeadows: I was squinting at these `connect()` calls wondering what happens when the code creates a duplicate connection.  However when I examined the implementation of `Application::updateDisplayMode()` more carefully I realized that this code is inside the context of a `std::call_once()`.

thoys:
## QA Testing Report

This PR fixes the crash that occurs on shutting down SteamVR. It switches back to Desktop mode.

Non QA related: `Although when re-enabling SteamVR the OpenVR option is not available. Is this something fixable?`

This fixed a major annoyance, good job!.

:+1: Successfully Tested! :+1:  :chart:

PR #7980, opened 2016-05-27
fix domain max capacity to only count users (not scripted agents)

zzmp:


zzmp:
#### Testing

- Start a domain server.
- In settings for that server, set max capacity to 1 and run an ac script that creates a user agent (https://raw.githubusercontent.com/highfidelity/hifi/master/script-archive/acScripts/animatedAvatarAgent.js).
- Connect to that server nonlocally (local connections bypass this restriction).

In current build, the agent counts towards max capacity. In this build, it does not.

- Connect, and have another user connect. Ensure the real user is counted towards the capacity.

domain-server/src/DomainServer.cpp
```diff
@@ -715,9 +715,13 @@ void DomainServer::processListRequestPacket(QSharedPointer<ReceivedMessage> mess
 unsigned int DomainServer::countConnectedUsers() {
     unsigned int result = 0;
     auto nodeList = DependencyManager::get<LimitedNodeList>();
-    nodeList->eachNode([&](const SharedNodePointer& otherNode){
-        if (otherNode->getType() == NodeType::Agent) {
-            result++;
+    nodeList->eachNode([&](const SharedNodePointer& node){
+        // only count unassigned agents (i.e., users)
+        if (node->getType() == NodeType::Agent) {
```
Atlante45: Pretty sure scripted agents are of type "NodeType::Agent", no?

hifi-gustavo:
Build console output for commit: b280eebd03026c807f40a4d3c157879ec9993307 available at: https://gist.github.com/6b652e281aef9d73789fb97c20aaa81c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7980.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7980.dmg

domain-server/src/DomainServer.cpp
ZappoMan: right - you'd need to check the "Assigned"  property of the node... like @birarda did when adjusting the connected user counts.

Atlante45: Yes, I missed that, he actually does that underneath.
Nvm

Nex-Pro:
Start a domain server. **DONE**
In settings for that server, set max capacity to 1 and run an ac script that creates a user agent (https://raw.githubusercontent.com/highfidelity/hifi/master/script-archive/acScripts/animatedAvatarAgent.js). **DONE**
Connect to that server nonlocally (local connections bypass this restriction). **DONE**
In current build, the agent counts towards max capacity. In this build, it does not.

Connect, and have another user connect. Ensure the real user is counted towards the capacity. **DONE**

**With limit set to 1 and  with the AC script running, second user (on another PC) was able to log in to my sandbox along with my locally logged-in avatar.**

Nex-Pro:
In addition Interface exited cleanly.

Would not auto re-connect when server restarted. Had to jump to dev-demo and back.

ZappoMan:
Btw - I'm not saying we shouldn't merge this... but I am wondering if we really want to make this change.

The original intent of this feature was to provide some control over the input/output bandwidth burden placed on the domain assignment clients... it turns out that assigned agents still do have some network bandwidth implied... most of them today (from sandboxes do run on local host) and therefore don't impact the bandwidth... but they could be running remotely, and therefore they would still have a burden.

Anyway... I guess my question is, are we sure this is a bug? Or is the original behavior actually correct? Or is there a different fix we need to do?

OmegaHeron:
I'd be fine with this with addition of a max "NPC" or whatever you want to call it variable - I'm perfectly content not to see script created agents counted as "real", but, I'm not comfortable with there being no cap for them in case something "bad" happened.

zzmp:
Investigating this so I can give a more complete response.

zzmp:
Right now, it is impossible to differentiate an unknown agent (AC-generated agent connection, a script-runner) from a user (Interface-generated agent connection, a human, presumably).

Unless the agent's uuid is known beforehand, [they are both created as generic agent nodes](https://github.com/highfidelity/hifi/blob/4120ed9f0bcc17250ddbda8718bdc0819e4118cf/domain-server/src/DomainGatekeeper.cpp#L87) until [the agent's first assignment request](https://github.com/highfidelity/hifi/blob/4120ed9f0bcc17250ddbda8718bdc0819e4118cf/domain-server/src/DomainGatekeeper.cpp#L162). ([Both of these listeners are registered in the DomainServer ctor](https://github.com/highfidelity/hifi/blob/master/domain-server/src/DomainServer.cpp#L390).)

Knowing this, and considering:
- the Domain Operator has (or at least *should* have) complete control over number of agent scripts
- the remote agents should be limited to reduce network bandwidth

and assuming that we only care about network bandwidth, not UDP packets sent locally (i.e. domain-servers can have many assignment-client nodes colocated on the same machine without affecting this limit),

**I think that the settings page should be changed to say "Maximum Remote Agent Capacity", or at least that the explanation of the field should be updated,** instead of the change proposed in this PR.

What do you guys think?

birarda:
I think we should consider the differentiator to be unassigned vs assigned for remote agents. The cap should apply to all unassigned Agents (Interface or otherwise) while the assigned ones do not count.

There is no path in our current to connect an assignment-client Agent without an assignment, but I can imagine that being something supported in the future.

zzmp:
OK, then I'm still advocating for this change. I added a bit to the explanation in describe-settings.json.

hifi-gustavo:
Build console output for commit: a8228d42ee9efeccd0548809e14a9e3130c51eef available at: https://gist.github.com/6b652e281aef9d73789fb97c20aaa81c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7980.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7980.dmg

PR #7986, opened 2016-05-30
make navigating directories in file browser dialog possible without double-clicking

ctrlaltdavid:
This is for controllers in particular, where double-clicking on a folder to browser it is not so easy.

### DIFF
--- a/interface/resources/qml/dialogs/FileDialog.qml
+++ b/interface/resources/qml/dialogs/FileDialog.qml
@@ -26,10 +26,10 @@ import "fileDialog"
 ModalWindow {
     id: root
     resizable: true
-    implicitWidth: 640
-    implicitHeight: 480
+    implicitWidth: 480
+    implicitHeight: 360
 
-    minSize: Qt.vector2d(388, 240)
+    minSize: Qt.vector2d(360, 240)
     draggable: true
 
     HifiConstants { id: hifi }
--- a/interface/resources/qml/dialogs/FileDialog.qml
+++ b/interface/resources/qml/dialogs/FileDialog.qml
@@ -203,6 +203,7 @@ ModalWindow {
                 var row = fileTableView.currentRow;
 
                 if (row === -1) {
+                    openFolderButton.enabled = false;
                     return;
                 }
 
@@ -213,6 +214,7 @@ ModalWindow {
                 } else {
                     currentSelection.text = "";
                 }
+                openFolderButton.enabled = currentSelectionIsFolder
             }
 
             function navigateUp() {
@@ -388,6 +390,13 @@ ModalWindow {
                     });
 
                     rows++;
+                }
+
+                fileTableView.selection.clear();
+                if (model.count > 0 && fileTableView.activeFocus) {
+                    fileTableView.currentRow = 0;
+                    fileTableView.selection.select(0);
+                    d.update();
                 }
             }
         }
@@ -617,6 +626,16 @@ ModalWindow {
                 Keys.onReturnPressed: okAction.trigger()
                 KeyNavigation.up: selectionType
                 KeyNavigation.left: selectionType
+                KeyNavigation.right: openFolderButton
+            }
+
+            Button {
+                id: openFolderButton
+                text: "Open Folder"
+                enabled: false
+                action: openFolderAction
+                KeyNavigation.up: selectionType
+                KeyNavigation.left: openButton
                 KeyNavigation.right: cancelButton
             }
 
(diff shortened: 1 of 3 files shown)


ctrlaltdavid:
Testing

Browse to choose file (e.g., Running Scripts > From File):
- You can both double-click or use the Open Folder button to navigate down the directory tree.
- You can successfully choose a file.

Browse to choose folder (e.g., Settings > General > Snapshots directory):
- You can both double-click or use the Open Folder button to navigate down the directory tree.
- You can successfully choose a directory.

The "Open Folder" button is enabled only when a folder has the focus highlight.

hifi-gustavo:
Build console output for commit: 437607998f3e777cc9e645762bdf3b16b746a9c9 available at: https://gist.github.com/e81427479de48adec35d40f1d2d9a7dd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7986.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7986.dmg

ctrlaltdavid:
Test this please

hifi-gustavo:
Build console output for commit: 437607998f3e777cc9e645762bdf3b16b746a9c9 available at: https://gist.github.com/e81427479de48adec35d40f1d2d9a7dd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7986.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7986.dmg

ZappoMan:
btw - I noticed that on windows at least, for the system file dialog, the "Open" button will open a folder if the focus is currently on a folder... shouldn't we use that UI/UX since it's common?

AlanZimmerman:
@ZappoMan I brought up the same point and @ctrlaltdavid pointed out that we use the file browser to set folder locations as well. We then discussed several more elegant (and labor-intensive) solutions but punted on them for now in order to get this fix out. I can revisit this if you'd like, I definitely have ideas for how this could be made much better.

ZappoMan:
IMO -- "reusing the ideas of existing standard UIs" is always better...
because it allows previous "training" to allow users to discover the right
answer.

AlanZimmerman:
I completely agree. I don't think any of us are advocating against reusing existing standard UI's. The problem we've encountered is that standard UI's in this case rely on double-clicking or right clicking. 

If we are asking people to specify a save location, we need to let them open a series of folders in the file browser, and then select one. We can either make bigger click targets so that double clicking is easier or come up with a different mechanic. For what it's worth, I would LOVE it if we didn't have this third button thing, I just couldn't think of a better solution that didn't require a ton of additional work. The estimate I got from David on my alternative idea was that it was 4x the work.

ZappoMan:
The standard windows UI allow you to double click as a way of opening a
file or directory. BUT IT ALSO supports selecting the item in the list and
then clicking the "open" button. If the selected item is a file, it opens
it and closes the dialog, if the file is a directory, it opens the
directory... this is the standard UI... we should support that.

ZappoMan:
Also note: when it's a "save" dialog... the text of the button changes from
"Save" to "Open" if you select a folder in the list vs a file.

ctrlaltdavid:
@ZappoMan "The standard windows UI allow you to double click as a way of opening a
file or directory" ... This is still supported. The point of this PR is to cater for HMD users where double-clicking a directory is not easy: https://highfidelity.fogbugz.com/f/cases/682/new-open-file-dialog-doesn-t-enable-open-button-on-directories And in the case of the dialog being opened to select a directory, there needs to be a way, after clicking on a directory in the list, to distinguish between "navigate into this directory" versus "select this directory and close the dialog"; hence the third button.

hifi-gustavo:
Build console output for commit: b65938b6316802dc69e1a5a3dea3452e7f148601 available at: https://gist.github.com/e81427479de48adec35d40f1d2d9a7dd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7986.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7986.dmg

Nex-Pro:
### Nice work @ctrlaltdavid 

Browse to choose file (e.g., Running Scripts > From File):

You can both double-click or use the Open Folder button to navigate down the directory tree. **SUCCESS**

You can successfully choose a file. #### SUCCESS

Browse to choose folder (e.g., Settings > General > Snapshots directory): **PARTIAL SUCCESS**
***Button displays as 'OPEN' when selecting a folder as the snapshot directory.
Shouldn't it display as "CHOOSE" ?***

You can both double-click or use the Open Folder button to navigate down the directory tree. **SUCCESS**

You can successfully choose a directory. **SUCCESS**

The "Open Folder" button is enabled only when a folder has the focus highlight. **SUCCESS**

**Note:** "Open Folder" button displays as "OPEN"

ctrlaltdavid:
@Nex-Pro That's per spec but I've pinged the spec-ers about it, thanks.

AlanZimmerman:
@ctrlaltdavid This looks good to me, I think everything is as-spec'ed. @ZappoMan You wanna take a look to confirm these changes match what we were talking about yesterday?

PR #7987, opened 2016-05-31
handControllerGrab respects overlays

howard-stearns:
Draw grab balls in front of entities (not just the lines as before).
Don't intersect with grab balls (not just lasers as before).
Don't run grab lasers through (2d or 3d) overlays.
Don't try to manage reticle in handControllerGrab because other scripts do.

### DIFF
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -249,6 +249,18 @@ function entityIsGrabbedByOther(entityID) {
     return false;
 }
 
+// If another script is managing the reticle (as is done by HandControllerPointer), we should not be setting it here,
+// and we should not be showing lasers when someone else is using the Reticle to indicate a 2D minor mode.
+var EXTERNALLY_MANAGED_2D_MINOR_MODE = true;
+function isIn2DMode() {
+    return EXTERNALLY_MANAGED_2D_MINOR_MODE && Reticle.visible;
+}
+function restore2DMode() {
+    if (!EXTERNALLY_MANAGED_2D_MINOR_MODE) {
+        Reticle.setVisible(true);
+    }
+}
+
 function MyController(hand) {
     this.hand = hand;
     if (this.hand === RIGHT_HAND) {
@@ -302,7 +314,10 @@ function MyController(hand) {
     this.update = function() {
 
         this.updateSmoothedTrigger();
-
+        if (isIn2DMode()) {
+            _this.turnOffVisualizations();
+            return;
+        }
         switch (this.state) {
             case STATE_OFF:
                 this.off();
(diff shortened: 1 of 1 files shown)


howard-stearns:
Note that this is to be merged to vive-ui feature branch.

howard-stearns:
Test plan:

* Make sure search/grab lasers do not run through 2D HUD elements:
 - Hold the trigger partially down to get the blue search beam.
 - Move the beam to intersect with an open QML window. The beam should stop when you reach the border of the window, at the same time as the mouse cursor appears. (In master, the search beam will be displayed along with the mouse, and the mouse will flicker.)
 - The same thing should happen with scripted overlay windows such as the examples button or edit button/toolbar.
 - The same thing should happen with the "avatar inputs" (microphone mute and level), and with the stats display.

* Make sure search/grab lasers do not run through 3D overlay elements. (This is harder to test.)
 - Open edit, and select an object (e.g., a cube) with the mouse. (It will be hard to select it with the hand controller, with or without this PR.)
 - Manipulate your hand position and direction so as to have the search beam intersect the 3d overlays. Note that if you are using the "active" hand (the last one in which you clicked button 3 in the hydra, or center-clicked the trackpad in vive), you will see two balls along the blue search beam. A blue ball will be at the end of the search beam, where it intersects and object, and a red ball will be where the search beam intersects the HeadsUpDisplay.  The trick here is to see if the blue ball and the end of the laser terminates at the 3D overlays.  (What I did was position myself with the selected cube nearer to me, with a wall a good distance behind it.  As I point the search beam in the general direction of the cube, the beam and the blue ball terminate at the wall. When I manipulate the controller to have the beam pass though the up/down arrow of the selected object, the blue ball and the beam terminus "jumps" forward to the up/down arrow, instead of being back along the wall. (In master, the blue ball and beam terminus stays along the wall.)

* The mouse cursor and the seach/grab beam should not be shown at the same time.
 - Let go of the hand controller trigger. Move the mouse around and you should see the mouse cursor. With the mouse still being moved, use another hand to squeeze the hand controller trigger. You should not see the search/grab beam while the mouse is visible. If you keep the trigger squeezed, you should see the appropriate beam five seconds after you stop moving the mouse (when the mouse cursor disappears, and not before).
 - Let go of the hand controller trigger and the mouse, and wait five seconds until the mouse cursor disappears. Squeeze the trigger partially until the blue search beam appears. With the trigger still squeezed, move the mouse and observe that the mouse cursor does not show, while the beam stays on. (You may observe hints that the mouse is still being moved even though it is not visible. For example, you can move the invisible mouse over HUD buttons. This will be addressed in a separate PR.)

hifi-gustavo:
Build console output for commit: 9ad19a2eaf031d95e9be1a6059aafb4790bed4ff available at: https://gist.github.com/e65e17e8742aa094c36f548ebe2a99af<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7987.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7987.dmg

thoys:
## QA results

- Grab search lasers don't run through 2D overlays (they are not visible).
- Grab search lasers don't run through 3D overlays (the stop nicely at the nearest surface it finds).
- Mouse cursor and grab beam were never present at the same time.

QA Passed successfully :+1:

PR #7988, opened 2016-05-31
partial fix for glitchy remote simulation of object with action

AndrewMeadows:
* perform server-side kinematic integration of objects with actions

### DIFF
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -922,13 +922,11 @@ void EntityItem::simulate(const quint64& now) {
         qCDebug(entities) << "     ********** EntityItem::simulate() .... SETTING _lastSimulated=" << _lastSimulated;
     #endif
 
-    if (!hasActions()) {
-        if (!stepKinematicMotion(timeElapsed)) {
-            // this entity is no longer moving
-            // flag it to transition from KINEMATIC to STATIC
-            _dirtyFlags |= Simulation::DIRTY_MOTION_TYPE;
-            setAcceleration(Vectors::ZERO);
-        }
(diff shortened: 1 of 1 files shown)


AndrewMeadows:
There is a bug where when an object is under the influence of a "physics action" the local simulation looks good but the remote simulation is bad -- the object appears to pop forward in its simulation in big steps -- it is not smooth.

The bug has two parts:

(1) The entity-server's simple (non-colliding) simulation will not step forward for an object that has a physics action.  This is a simple logic error that is fixed in this PR, which helps but does not solve the whole problem...

(2) There is a bug where the "send everything in the camera view" logic will resend duplicate data for the last changed object several times in a row.  This causes a backup of stale updates that cause the object's motion to glitch.  Not fixed in this PR -- I tried but failed.  The internal issue for this is Bug-483.

kevinmthomas-carpool:
smoke test?

AndrewMeadows:
No, this one is not actually ready for test, but is posted just in case it helps some other debugging.

kevinmthomas-carpool:
Ok then.

hifi-gustavo:
Build console output for commit: 5735ee824c5f10e12842f62b434b9ee72faee348 available at: https://gist.github.com/cb817a90d5ba96e2e66deaeddb6f6765<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7988.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7988.dmg

kevinmthomas-carpool:
If this is now in QA Ready what are you looking for a smoke test or is there something else please advise?

ZappoMan:
I've included this fix in PR#8003

PR #7991, opened 2016-06-01
Update the build docs for a better OpenSSL security posture.

aoighost:
There are around 19 CVEs in the current recommended OpenSSL. This should help make it clearer to use the latest and keep up to date.


hifi-gustavo:

PR #7999, opened 2016-06-02
Do not send stale vive data as valid after deactivating.

howard-stearns:
Do not send stale vive data as valid.
It had been the case that, after switching into Vive HMD and back to desktop, the hand pose input data was sending unchanging invalid data and reporting it as valid. As a result, the hand positions were bad, and scripts that depended on valid data were bad (including handControllerPointer, which moved the mouse to where it was told).

### DIFF
--- a/plugins/openvr/src/OpenVrDisplayPlugin.cpp
+++ b/plugins/openvr/src/OpenVrDisplayPlugin.cpp
@@ -96,6 +96,11 @@ void OpenVrDisplayPlugin::internalDeactivate() {
     Parent::internalDeactivate();
     _container->setIsOptionChecked(StandingHMDSensorMode, false);
     if (_system) {
+        // Invalidate poses. It's fine if someone else sets these shared values, but we're about to stop updating them, and
+        // we don't want ViveControllerManager to consider old values to be valid.
+        for (int i = 0; i < vr::k_unMaxTrackedDeviceCount; i++) {
+            _trackedDevicePose[i].bPoseIsValid = false;
+        }
         releaseOpenVrSystem();
         _system = nullptr;
     }


howard-stearns:
Fixes fogbugzid:739
This is a vive-ui branch PR, but should be cherry-pickable to master if desired.

hifi-gustavo:
Build console output for commit: e6cc1fabe0d97006024e38de4a498b1f364fa860 available at: https://gist.github.com/f0684441a235e19bf15eefde92e24e8a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7999.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR7999.dmg

howard-stearns:
Test plan: requires vive

Make sure the default scripts that go with this build are loaded. (e.g., stop all scripts, then load defaultScripts.js, making sure the path is the one associated with this PR install.)

Go into vive hmd, and make sure hand controllers work. Then go back to desktop, and confirm the following (in desktop):

- Avatar should be "normal" (e.g., when viewed in third person), and not have its hand's pulled in strange ways.  Waving the hand controllers around should have no effect.

- The mouse should be useable and not pinned to the bottom of the screen (or pinned anywhere else). When the mouse times out after five seconds, and you bring the mouse back by moving, the cursor should be where you left it.

- In Edit->Console..., Controller.getPoseValue(Controller.Standard.RightHand).valid should be false, and similarly for LeftHand.

thoys:
## QA Testing Report

Loaded default scripts for this PR. Hand controllers work fine in Vive mode, switched back to Desktop mode with expected results: 
- Hand controllers have no effect;
- Avatar posture is normal, e.g. no hands up in the air; 
- The mouse is usable and not pinned;
- `Controller.getPoseValue(Controller.Standard.RightHand).valid` returns `false`
- Mouse cursor is usable.

Also the mouse cursor gets usable after 5 seconds not moving the head in OpenVR mode.

:+1: Successfully Tested :+1:

PR #8004, opened 2016-06-03
fix bad convex hull simplification

AndrewMeadows:
* fix bug where some high-point-count convex hulls would be incorrectly simplified
* use many colors for convex hull visualization
* cleanup and optimizations around collision shape generation


libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -599,11 +599,7 @@ bool RenderableModelEntityItem::isReadyToComputeShape() {
 
 void RenderableModelEntityItem::computeShapeInfo(ShapeInfo& info) {
     ShapeType type = getShapeType();
-    if (type != SHAPE_TYPE_COMPOUND) {
-        ModelEntityItem::computeShapeInfo(info);
-        info.setParams(type, 0.5f * getDimensions());
-        adjustShapeInfoByRegistration(info);
```
AndrewMeadows: This non-compound-shape case moved to the `else` block.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -612,25 +608,27 @@ void RenderableModelEntityItem::computeShapeInfo(ShapeInfo& info) {
...
         // the way OBJ files get read, each section under a "g" line is its own meshPart.  We only expect
         // to find one actual "mesh" (with one or more meshParts in it), but we loop over the meshes, just in case.
+        const uint32_t TRIANGLE_STRIDE = 3;
+        const uint32_t QUAD_STRIDE = 4;
         foreach (const FBXMesh& mesh, collisionGeometry.meshes) {
             // each meshPart is a convex hull
             foreach (const FBXMeshPart &meshPart, mesh.parts) {
-                QVector<glm::vec3> pointsInPart;
+                points.push_back(QVector<glm::vec3>());
+                QVector<glm::vec3>& pointsInPart = points[i];
```
AndrewMeadows: `points` is now a local variable here which allows me to remove `RenderableModelEntityItem::_points` data member (which was ever only used here when generating shape data for models that needed a compound shape).

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -612,25 +608,27 @@ void RenderableModelEntityItem::computeShapeInfo(ShapeInfo& info) {
         const FBXGeometry& renderGeometry = _model->getFBXGeometry();
         const FBXGeometry& collisionGeometry = _model->getCollisionFBXGeometry();
 
-        _points.clear();
-        unsigned int i = 0;
+        QVector<QVector<glm::vec3>>& points = info.getPoints();
+        points.clear();
+        uint32_t i = 0;
```
AndrewMeadows: `points` is now a reference to `ShapeInfo::_points` which allows us to avoid a copy and also remove the unecessary `RenderableModelEntityItem::_points` data member which was only ever used here to generate points to be copied over.

libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -101,7 +101,6 @@ void buildStringToShapeTypeLookup() {
     addShapeType(SHAPE_TYPE_NONE);
     addShapeType(SHAPE_TYPE_BOX);
     addShapeType(SHAPE_TYPE_SPHERE);
-    addShapeType(SHAPE_TYPE_ELLIPSOID);
```
AndrewMeadows: SHAPE_TYPE_ELLIPSOID was unused and unsupported.  Purging it now but will soon provide valid shapes for more objects.

libraries/entities/src/LineEntityItem.h
```diff
@@ -60,8 +60,6 @@ class LineEntityItem : public EntityItem {
     
     const QVector<glm::vec3>& getLinePoints() const{ return _points; }
     
-    virtual ShapeType getShapeType() const { return SHAPE_TYPE_LINE; }
-
```
AndrewMeadows: LineEntityItem's are not collidable.  Maybe they should be eventually, but we'll do it right when the time comes.

libraries/entities/src/ParticleEffectEntityItem.h
```diff
@@ -281,7 +278,6 @@ class ParticleEffectEntityItem : public EntityItem {
 
     QString _textures { DEFAULT_TEXTURES };
     bool _texturesChangedFlag { false };
-    ShapeType _shapeType { SHAPE_TYPE_NONE };
```
AndrewMeadows: ParticleEffectEntityItem not supported for collisions in the physics engine.  This looks like copy-n-paste error.

libraries/physics/src/ShapeFactory.cpp
```diff
@@ -66,15 +114,40 @@ btConvexHullShape* ShapeFactory::createConvexHull(const QVector<glm::vec3>& poin
...
+            for (uint32_t j = 0; j < finalIndices.size(); ++j) {
+                if (finalIndices[j] == bestIndex) {
+                    keep = false;
+                    break;
+                }
+            }
+            if (keep) {
+                finalIndices.push_back(bestIndex);
+            }
+        }
```
AndrewMeadows: This code is what fixes the collisions for AlphaVersionD's cruise ship.  It appears that Bullet has some bugs in their btShapeHull generation, but that is a wasteful way to prune the convex hull anyway so we'll use this simpler method.

libraries/render-utils/src/Model.cpp
```diff
@@ -35,7 +35,46 @@ int vec3VectorTypeId = qRegisterMetaType<QVector<glm::vec3> >();
...
+            float red = component[index];
+            float green = component[(index + greenPhase) % NUM_COLLISION_HULL_COLORS];
+            float blue = component[(index + bluePhase) % NUM_COLLISION_HULL_COLORS];
+            material->setAlbedo(glm::vec3(red, green, blue));
+            material->setMetallic(0.02f);
+            material->setRoughness(0.5f);
+            _collisionHullMaterials.push_back(material);
+        }
+    }
+}
```
AndrewMeadows: This code creates many colors for rendering convex hulls for easier visualization of how the parts join together.

libraries/shared/src/ShapeInfo.h
```diff
@@ -60,10 +58,10 @@ class ShapeInfo {
     const glm::vec3& getHalfExtents() const { return _halfExtents; }
     const glm::vec3& getOffset() const { return _offset; }
 
+    QVector<QVector<glm::vec3>>& getPoints() { return _points; }
```
AndrewMeadows: The ShapeInfo is really just a temporary data structure for passing info to the ShapeFactory which actually generates the collision shapes.  So there is no need to enthusiastically clear ShapeInfo::_points in the class methods.  The points can now be obtained as reference and can be cleared manually as necessary.

hifi-gustavo:
Build console output for commit: e8e87b5db63a928de7973d78729b8ba0df09df91 available at: https://gist.github.com/84a2ffa9bbe66ce2ff47cc44a601ab43<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8004.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8004.dmg

hifi-gustavo:
Build console output for commit: 4df45a9f62f5108bab90349524ebfa40ba455878 available at: https://gist.github.com/84a2ffa9bbe66ce2ff47cc44a601ab43<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8004.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8004.dmg

libraries/entities/src/LineEntityItem.h
ZappoMan: please explain this change... if you don't return type line - what's the resulting default? if that results in something collidable you're introduce unexpected behavior

libraries/entities/src/ParticleEffectEntityItem.cpp
```diff
@@ -406,7 +404,6 @@ int ParticleEffectEntityItem::readEntitySubclassDataFromBuffer(const unsigned ch
         READ_ENTITY_PROPERTY(PROP_EMITTING_PARTICLES, bool, setIsEmitting);
     }
 
-    READ_ENTITY_PROPERTY(PROP_SHAPE_TYPE, ShapeType, updateShapeType);
```
ZappoMan: this is a protocol version change... did you intend that...

ZappoMan:
This PR changes the wire format but does not properly bump the protocol. It's "Dev Rejected" because of that.

I also think this PR should be split into 2 PRs - one that does the actual bug fix, and the other refactoring/cleanup code should be independent.

AndrewMeadows:
I'll separate out the good parts and submit those in their own PR.

PR #8022, opened 2016-06-06
add resized, moved, and closed OverlayWebWindow events

ctrlaltdavid:


hifi-gustavo:
Build console output for commit: d600e6a88d4120248cb4c752b984c71cbc99f1ee available at: https://gist.github.com/3b78ed49b99404129867b25b3e068555<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8022.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8022.dmg

ctrlaltdavid:
Use the attached test script to verify that the events are emitted when you resize, move, or close the test window the script creates.
On a non-PR build, verify that the events are not emitted.
Verify that default scripts with UI still function as expected (in particular resizing, moving, and closing) in desktop and HMD modes.

[testOverlayWebWindow.zip](https://github.com/highfidelity/hifi/files/299840/testOverlayWebWindow.zip)

kevinmthomas-carpool:
testing

interface/resources/qml/windows-uit/DefaultFrame.qml
```diff
@@ -83,7 +83,10 @@ Frame {
                     id: closeClickArea
                     anchors.fill: parent
                     hoverEnabled: true
-                    onClicked: window.visible = false;
+                    onClicked: {
+                        window.visible = false;
+                        window.windowClosed();
```
jherico: A number of windows are designed to 'hide' rather than 'close' when the close button is clicked.  Even if they're not, they don't actually get destroyed until the fade out occurs.  I would suggest that this signal be emitted in the onVisibleChanged handler for window, rather than here in the frame.

jherico:
Unless this PR is focused on specific issue that needs resolving, you may wish to defer it until after the Vive-UI branch has been merged, which has significant changes to the QML windowing code and merges all the `qml/window-uit` contents into `qml/window`.  Alternatively, you could open it up against against the vive-ui branch to avoid any future merge issues.

kevinmthomas-carpool:
## QA Test
**QA Tester**: KevinMThomas
**Additional Test Results + Observations**:  Utilized ```testOverlayWebWindow.js``` to which a UI was created and verified in PR build that upon emit, resize, move and closing the test window events are output in the log.  Ran test on 4970 and there were no such results.  Checked in both desktop and HMD modes.

ctrlaltdavid:
@jherico  These three events – window moved, resized, closed - are missing in OverlayWebWindow c.f. WebWindow which is a nuisance for a (non-official) script I’m working on. However, I’ll work around their not being available by polling and emulating the events, and defer this PR until sometime after the Vive UI branch is merged.
Note on closed event: I’m reasonably sure that for WebWindow the closed event only fires if you close it with the “x” or Alt-F4 (i.e., not if one’s script closes it; hence why I implemented it on the “x” button for OverlayWebWindow. However, I’ll revisit when I resubmit this PR later on sometime.

jherico:
@ctrlaltdavid right, but for the webwindow, the X is a system level item that *always* actually closes the window, never hides it.  Also, I'm assuming the event actually comes from the closing of the window, not the click, so if you click on the x or use a keyboard shortcut or menu to close the window, you should still get the close event.  

For upating to the new UI, I agree that the moved, resized and closed events should all be supported, although we should probably add an additional 'shown/hidden' event.  Of course, ultimately I want to get the script interfaces we have in interface JS scripts all visible to both QML and HTML views, so that we don't have to have an interface script do all this communication with a QML window, but instead put the logic directly in the QML window or in the HTML JS.

PR #8019, opened 2016-06-05
Eliminate file IO contentions for settings

jherico:
This removes most of the use of the QSettings file from the main thread, as well as removing the cost of the frequently constructed / destroy QSettings derived Settings object.

The existing code was using short-lived Settings objects when Settings was still derived from QSettings.  This meant file IO was triggered on the main thread every time a Settings object was created or destroyed, which happens every 10 seconds in code called by the `Application::saveSettings`.  Every time this was run, both the `Menu` and `MyAvatar` classes would instantiate a Settings object, write to it and then destroy it, triggering multiple reads, parses and writes of the file.  This is especially painful on lower end machines or those with slower disks.  

Additionally, I've decoupled the creation and destruction of the Setting::Handle type from triggering nearly immediate disk writes, even if the handle was never set or the value was the same as the existing written value.

PR #8023, opened 2016-06-06
Eliminate extraneous writes for unchanged values

jherico:
This commit was supposed to go in my [earlier PR](https://github.com/highfidelity/hifi/pull/8019) for settings from this weekend, but apparently I forgot to `git push`.  This further reduces the number of file writes by causing any client which uses the old style `Settings` interface to set a value to the already extant value, or the remove a value that already doesn't exist, to become a no-op.

### DIFF
--- a/libraries/shared/src/SettingHandle.cpp
+++ b/libraries/shared/src/SettingHandle.cpp
@@ -28,7 +28,9 @@ Settings::~Settings() {
 }
 
 void Settings::remove(const QString& key) {
-    _manager->remove(key);
+    if (key == "" || _manager->contains(key)) {
+        _manager->remove(key);
+    }
 }
 
 QStringList Settings::childGroups() const {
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: bc81f00dc76e45a29f79800a5949a5fe7c0515ca available at: https://gist.github.com/69eeb16da2e6aca9af0fa8dac4f00ebc<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8023.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8023.dmg

jherico:
## Testing

This should have the same behavior as the existing dev build of Interface.  However, you *should* see fewer writes to the INI file listed in a monitoring application such as the windows `Resource Monitor` tool to the INI file.

image

Nex-Pro:
@jherico  You want this compared to 4987 ?

jherico:
@Nex-Pro Sure, that would be most appropriate.  It should have equivalent or better performance in all cases.  Unfortunately I think it's still not going to be optimal because I think the scripting system always clears and then re-saves the list of scripts, but it may reduce load from other things that are currently 'updating' the INI file with values that it already has.

Nex-Pro:
Ok

Nex-Pro:
### PASS  *** Way fewer entries showing in the Resource Monititor ***
4987activity
pr8023activity

** Good Job ** Now that we are BFFs can I ask you thousands of Android Java questions ? lol

PR #8025, opened 2016-06-06
Render equip hotspot when trigger is squeezed

hyperlogic:
For entities that have wearable attach points, render a green sphere around them when the trigger is depressed.  Eventually these equip hotspots will be more functional, but at the moment they are only visual, they don't actually change the grab behavior at all.

### DIFF
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -376,7 +376,7 @@ function MyController(hand) {
                 print("WARNING: could not find updateMethod for state " + stateToName(this.state));
             }
         } else {
-            print("WARNING: could not find state " + this.state + " " + CONTROLLER_STATE_MACHINE[this.state]);
+            print("WARNING: could not find state " + this.state + " in state machine");
         }
     };
 
@@ -390,7 +390,7 @@ function MyController(hand) {
         if (WANT_DEBUG || WANT_DEBUG_STATE) {
             var oldStateName = stateToName(this.state);
             var newStateName = stateToName(newState);
-            print("STATE (" + this.hand + "): " + oldStateName + " --> " + newStateName);
+            print("STATE (" + this.hand + "): " + newStateName + " <-- " + oldStateName);
         }
 
         // exit the old state
@@ -401,7 +401,7 @@ function MyController(hand) {
                 exitMethod.call(this);
             }
         } else {
-            print("WARNING: could not find this.state " + this.state);
+            print("WARNING: could not find state " + this.state + " in state machine");
         }
 
         // enter the new state
@@ -412,7 +412,7 @@ function MyController(hand) {
                 enterMethod.call(this);
             }
         } else {
-            print("WARNING: could not find newState " + newState);
+            print("WARNING: could not find newState " + newState + " in state machine");
         }
 
         this.state = newState;
@@ -943,6 +943,7 @@ function MyController(hand) {
             this.setState(STATE_RELEASE);
             return;
         }
+
         if (this.state == STATE_HOLD_SEARCHING && this.bumperReleased()) {
             this.setState(STATE_RELEASE);
             return;
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -897,28 +897,26 @@ function MyController(hand) {
 
                 var hotspotPos = grabProps.position;
 
-                // does this entity have an equip point?
+                // does this entity have an attach point?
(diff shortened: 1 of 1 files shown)


scripts/system/controllers/handControllerGrab.js
```diff
@@ -179,47 +180,85 @@ var COLLIDES_WITH_WHILE_MULTI_GRABBED = "dynamic";
...
-            return "continue_equip";
-        case STATE_WAITING_FOR_EQUIP_THUMB_RELEASE:
-            return "waiting_for_equip_thumb_release";
-        case STATE_WAITING_FOR_RELEASE_THUMB_RELEASE:
-            return "waiting_for_release_thumb_release";
-    }
+var CONTROLLER_STATE_MACHINE = {};
+
+CONTROLLER_STATE_MACHINE[STATE_OFF] = {
+    name: "off",
```
howard-stearns: As long as you're making this change, it would probably be good to change the STATE_OFF, STATE_SEARCHING, etc. from numeric values to strings, and the use name: STATE_OFF here (and similarly below).

Or even better, create a 
function defineState(state, updateMethod, optionalEnterMethod, optionalExitMethod) {
   CONTROLLER_STATE_MACHINE[state] = {
      name: state,
      updateMethod: updateMethod,
      enterMethod: optionalEnterMethod,
      exitMethod: optionalExitMethod
      };
}
The idea of the latter is that changing the state machine itself (as opposed to its parameters) shouldn't require changing so many different spots in the code.

Just a thought.

hifi-gustavo:
Build console output for commit: b836a580ffda8b2625e64ec404d3520cfcf13ed5 available at: https://gist.github.com/1a3891fc2e4775b6501498c1242eac26<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8025.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8025.dmg

scripts/system/controllers/handControllerGrab.js
```diff
@@ -249,6 +288,14 @@ function entityIsGrabbedByOther(entityID) {
...
 }
 
+function propsArePhysical(props) {
+    if (!props.dynamic) {
+        return false;
+    }
+    var isPhysical = (props.shapeType && props.shapeType != 'none');
+    return isPhysical;
+}
+
```
howard-stearns: why change from a method? because you can? (i.e., it doesn't reference 'this', and so it doesn't need to be a method?)

That's a valid reason. I'm just asking so that I understand what you're thinking.

scripts/system/controllers/handControllerGrab.js
```diff
@@ -852,6 +882,56 @@ function MyController(hand) {
...
+                if (wearableData) {
+                    var handJointName = this.hand === RIGHT_HAND ? "RightHand" : "LeftHand";
+                    if (wearableData[handJointName]) {
+                        // draw the hotspot around the attach point.
+                        hotspotPos = wearableData[handJointName][0];
+                    }
+                }
+
+                // draw a hotspot!
+                this.equipHotspotOverlays.push(Overlays.addOverlay("sphere", {
```
howard-stearns: I would assume that an addOverlay at startup, with an editOverlay to change position and visible (on/off) would be more efficient than repeated addOverlay/deleteOverlay, but I don't actually know that.

scripts/system/controllers/handControllerGrab.js
hyperlogic: Yeah, I can do this.  I just wanted to keep the changes minimal at first.

scripts/system/controllers/handControllerGrab.js
hyperlogic: IMHO, a method implies that it modifies or requires access to member variables.  Where a function  does not.  In this case it's actually a "pure" function with no side-effects either...

scripts/system/controllers/handControllerGrab.js
hyperlogic: yeah, I don't know either.  I guess I was just following the pattern of the spheres and lines that are added/destroyed dynamically.

howard-stearns:
After discussion with Tony, I'm going to go ahead and merge this without a QA round.
Future PRs with functional changes will have test plan.

RyanDowne:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;xDYmYCww&#x2F;11-add-visible-equip-hot-spot"> Add visible equip hot-spot.</a>

PR #8026, opened 2016-06-06
SteamVR keyboard functionality

jherico:
This update to the earlier SteamVR keyboard PR moves most of the code from the ViveControllerManager.cpp file to OpenVrHelpers.cpp.  Both the display plugin and the input plugin now handle OpenVR events due to an unrelated PR, but the event handling itself is done in the common code in OpenVrHelpers, so it made sense to put the bulk of the code there.  The only thing ViveControllerManager.cpp is still responsible for is enabling the keyboard functionality when it activates and disabling it when it deactivates.

The keyboard itself should now simulate a 'return' keypress to the top level offscreen QML window, so when you enter an address in the address bar and press done, you should actually get taken to your destination.  However I'm uncertain of the implications of sending a return key to every possible text input field, as the effect might be undesirable on some of them.  

Additionally, the current overlay show / hide functionality is annoying in combination with the keyboard.  If the target of the keyboard is hidden when the desktop goes into pinned mode, the steam keyboard goes away along with it, because the item no longer has text focus.  In the case of the address bar, when the bar becomes visible again it immediately grabs focus and so the steam vr keyboard overlay reappears, but both the address bar and the keyboard overlay have been reset.  

Unfortunately this sequence can be triggered just be leaning too far in a direction, because this is considered to be driving your avatar and thus makes the overlays hidden.

### DIFF
--- a/libraries/gl/src/gl/OffscreenQmlSurface.cpp
+++ b/libraries/gl/src/gl/OffscreenQmlSurface.cpp
@@ -751,30 +751,9 @@ void OffscreenQmlSurface::onFocusObjectChanged(QObject* object) {
         return;
     }
 
-    QVariant result;
-#if 1
-    auto invokeResult = QMetaObject::invokeMethod(object, "inputMethodQuery", Q_RETURN_ARG(QVariant, result),
-        Q_ARG(Qt::InputMethodQuery, Qt::ImEnabled),
-        Q_ARG(QVariant, QVariant()));
-#else
-
-    //static const char* INPUT_METHOD_QUERY_METHOD_NAME = "inputMethodQuery(Qt::InputMethodQuery, QVariant)";
-    static const char* INPUT_METHOD_QUERY_METHOD_NAME = "inputMethodQuery";
-    auto meta = object->metaObject();
-    qDebug() << "new focus " << object;
-    auto index = meta->indexOfMethod(INPUT_METHOD_QUERY_METHOD_NAME);
-    if (index < 0 || index >= meta->methodCount()) {
-        setFocusText(false);
-        return;
-    }
-
-    auto method = meta->method(index);
-    auto invokeResult = method.invoke(object,
-        Q_RETURN_ARG(QVariant, result),
-        Q_ARG(Qt::InputMethodQuery, Qt::ImEnabled),
-        Q_ARG(QVariant, QVariant()));
-#endif
-    setFocusText(invokeResult && result.toBool());
+    QInputMethodQueryEvent query(Qt::ImEnabled);
+    qApp->sendEvent(object, &query);
+    setFocusText(query.value(Qt::ImEnabled).toBool());
 }
 
 void OffscreenQmlSurface::setFocusText(bool newFocusText) {
--- a/libraries/gl/src/gl/OffscreenQmlSurface.cpp
+++ b/libraries/gl/src/gl/OffscreenQmlSurface.cpp
@@ -396,6 +396,8 @@ void OffscreenQmlSurface::create(QOpenGLContext* shareContext) {
 
     _renderer->_renderControl->_renderWindow = _proxyWindow;
 
+    connect(_renderer->_quickWindow, &QQuickWindow::focusObjectChanged, this, &OffscreenQmlSurface::onFocusObjectChanged);
+
     // Create a QML engine.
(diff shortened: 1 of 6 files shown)


RyanDowne:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;JZUdLA0i&#x2F;14-steam-keyboard-integration"> steam keyboard integration</a>

thoys:
What is the testplan on this @jherico ? Just see if it functions?

thoys:
## Test report

1. Pressed `Enter` key to get the location input. 
2. SteamVR keyboard input appears nicely :+1: .
3. Typed `playa` using the touchpads.
4. Pressed the return key using touchpad.
5. SteamVR crash (both Steam and openvr-plugin, could be the Steam crash initiating the plugin crash).
(Tried this for other placenames, same happens)

### QA Failed :-1:

jherico:
@thoys can you tell me what version of SteamVR you're running?

jherico:
I can't reproduce this crash either with my local builds or with the PR build.

howard-stearns:
I can't reproduce the crash, but I'm not getting the keyboard at all.
(PR build of client and server, SteamVR 1464744840 built 5/31/16.)

jherico:
@howard-stearns from my experience, SteamVR has to be running for the keyboard to display.  Even though OpenVR works fine for rendering without the SteamVR app running, without the app, the keyboard doesn't function.

howard-stearns:
SteamVR is running before starting Interface.
(e.g., Starting steam if not already running in system tray, and then system tray -> steam -> steamvr.)

howard-stearns:
Hmm, now I crashed. But then I ran steam (i.e., the store), and it's shutting everything down while it updates. Stand by...

howard-stearns:
There are three programs:
- "Vive" does not appear to effect the results. e.g., can be running or not.
- "Steam" needs to be running to start SteamVR. It does not seem to matter what state "Steam" needs to be in after that. (e.g., how many steam windows are open, vs just the tray app).
- "SteamVR" is started from Steam, before starting Interface.

I start Sandbox from this PR, and then "Go Home" from that. The following is then from HMD:

- If SteamVR was not running before starting Interface, bringing up Navigate->address bar results in just the address bar. No keyboard and no crash.

. If SteamVR was already running before starting Interface, I sometimes crash as soon as I go to HMD (using this PR). If I make it through that, bringing up the address bar crashes as follows:
   1. HMD display goes to the white steam room.
   2. Interface goes to desktop mode.
   3. After a moment, SteamVR AND Steam both exit, with no dump report or other diagnostic.

thoys:
I am running build 1464744840 as well @jherico

jherico:
Hmmm... I'm running 1464900072.  I'll take a look at the code.  Most likely there's some minor change in the keyboard API between the beta and release versions.  I probably need to be checking for an error code at least.

hifi-gustavo:
Build console output for commit: 9975e798b04ac5273ab1c3f0da73dcd59dab9bd8 available at: https://gist.github.com/3fa0c0e1c5285ba9a641e47f0b79d194<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8026.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8026.dmg

hifi-gustavo:
Build console output for commit: 79c68b2ecb203cb09453e53179eb22e7689d00df available at: https://gist.github.com/3fa0c0e1c5285ba9a641e47f0b79d194<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8026.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8026.dmg

hifi-gustavo:
Build console output for commit: 40778d7f299777d5c395e0e742532f214ee5c799 available at: https://gist.github.com/3fa0c0e1c5285ba9a641e47f0b79d194<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8026.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8026.dmg

Issue #7844, opened 2016-05-08
Controller API should the ability to call vibrate event.


Menithal:
Controller API should have some capacity to vibrate the controllers for a certain length of time and specific intensities.

SamGondelman:
@Menithal this functionality has been added (although it only works on vive right now).  Let me know if there's anything else you want for this, otherwise feel free to close the issue :D

Menithal:
Sorry, took a bit to notice this was commented.  This has been implemented. Cant wait to try it out on the hookguns :)

PR #8027, opened 2016-06-06
Expose haptics to JS

SamGondelman:
Addresses: https://github.com/highfidelity/hifi/issues/7844

Added some JavaScript functions that allow you to trigger controller vibrations.  Added support for Vive controllers, Oculus Touch, and SDL game controllers, but only the Vive API seems to be working right now.

There are two main methods:
- Controller.triggerHapticPulse(float strength, float duration, Hand hand = 2)
- Controller.triggerHapticPulseOnDevice(uint deviceID, float strength, float duration, Hand hand = 2)

And there are also "short" versions of both of those that don't take in a duration and automatically trigger pulses of 250 ms:
- Controller.triggerShortHapticPulse(float strength, Hand hand = 2)
- Controller.triggerShortHapticPulseOnDevice(uint deviceID, float strength, Hand hand = 2)

Strengths should range from 0-1.  Durations are in milliseconds.

The last arguments are optional and default to 2, which means both hands.  0 means just the left hand, and 1 means just the right hand.

If you trigger a pulse while another one is already in progress, the strength and duration of the one that would finish last are honored.

### DIFF
--- a/libraries/controllers/src/controllers/InputDevice.h
+++ b/libraries/controllers/src/controllers/InputDevice.h
@@ -31,6 +31,12 @@ namespace controller {
 class Endpoint;
 using EndpointPointer = std::shared_ptr<Endpoint>;
 
+enum Hand {
+    LEFT = 0,
+    RIGHT,
+    BOTH
+};
+
 // NOTE: If something inherits from both InputDevice and InputPlugin, InputPlugin must go first.
 // e.g. class Example : public InputPlugin, public InputDevice
 // instead of class Example : public InputDevice, public InputPlugin
@@ -56,7 +62,7 @@ public:
     const QString& getName() const { return _name; }
 
     // By default, Input Devices do not support haptics
-    virtual bool triggerHapticPulse(float strength, float duration, bool leftHand) { return false; }
+    virtual bool triggerHapticPulse(float strength, float duration, controller::Hand hand) { return false; }
 
     // Update call MUST be called once per simulation loop
     // It takes care of updating the action states and deltas
--- a/libraries/controllers/src/controllers/InputDevice.h
+++ b/libraries/controllers/src/controllers/InputDevice.h
@@ -51,12 +51,12 @@ public:
 
     float getValue(const Input& input) const;
     float getValue(ChannelType channelType, uint16_t channel) const;
-	Pose getPoseValue(uint16_t channel) const;
+    Pose getPoseValue(uint16_t channel) const;
 
     const QString& getName() const { return _name; }
 
-	// By default, Input Devices do not support haptics
-	virtual bool triggerHapticPulse(float strength, float duration, bool leftHand) { return false; }
+    // By default, Input Devices do not support haptics
+    virtual bool triggerHapticPulse(float strength, float duration, bool leftHand) { return false; }
 
     // Update call MUST be called once per simulation loop
     // It takes care of updating the action states and deltas
--- a/libraries/controllers/src/controllers/InputDevice.h
+++ b/libraries/controllers/src/controllers/InputDevice.h
@@ -51,9 +51,12 @@ public:
 
     float getValue(const Input& input) const;
     float getValue(ChannelType channelType, uint16_t channel) const;
-    Pose getPoseValue(uint16_t channel) const;
+	Pose getPoseValue(uint16_t channel) const;
 
     const QString& getName() const { return _name; }
+
+	// By default, Input Devices do not support haptics
+	virtual bool triggerHapticPulse(float strength, float duration, bool leftHand) { return false; }
 
     // Update call MUST be called once per simulation loop
     // It takes care of updating the action states and deltas
--- a/plugins/hifiSdl2/src/Joystick.cpp
+++ b/plugins/hifiSdl2/src/Joystick.cpp
@@ -25,7 +25,7 @@ Joystick::Joystick(SDL_JoystickID instanceId, SDL_GameController* sdlGameControl
     _instanceId(instanceId)
 {
     if (!_sdlHaptic) {
-        qDebug() << SDL_GetError();
+        qDebug() << QString(SDL_GetError());
     }
     SDL_HapticRumbleInit(_sdlHaptic);
 }
--- a/plugins/hifiSdl2/src/Joystick.cpp
+++ b/plugins/hifiSdl2/src/Joystick.cpp
@@ -21,9 +21,10 @@ Joystick::Joystick(SDL_JoystickID instanceId, SDL_GameController* sdlGameControl
         InputDevice("GamePad"),
     _sdlGameController(sdlGameController),
     _sdlJoystick(SDL_GameControllerGetJoystick(_sdlGameController)),
+    _sdlHaptic(SDL_HapticOpenFromJoystick(_sdlJoystick)),
     _instanceId(instanceId)
 {
-    
+    SDL_HapticRumbleInit(_sdlHaptic);
 }
 
 Joystick::~Joystick() {
@@ -31,6 +32,7 @@ Joystick::~Joystick() {
 }
 
 void Joystick::closeJoystick() {
+    SDL_HapticClose(_sdlHaptic);
     SDL_GameControllerClose(_sdlGameController);
 }
 
@@ -60,6 +62,13 @@ void Joystick::handleButtonEvent(const SDL_ControllerButtonEvent& event) {
     } else {
         _buttonPressedMap.erase(input.getChannel());
     }
+}
+
+bool Joystick::triggerHapticPulse(float strength, float duration, bool leftHand) {
+    if (SDL_HapticRumblePlay(_sdlHaptic, strength, duration) != 0) {
+        return false;
+    }
+    return true;
 }
 
 controller::Input::NamedVector Joystick::getAvailableInputs() const {
--- a/plugins/hifiSdl2/src/Joystick.cpp
+++ b/plugins/hifiSdl2/src/Joystick.cpp
@@ -25,9 +25,12 @@ Joystick::Joystick(SDL_JoystickID instanceId, SDL_GameController* sdlGameControl
     _instanceId(instanceId)
 {
     if (!_sdlHaptic) {
-        qDebug() << QString(SDL_GetError());
+        qDebug() << "SDL Haptic Open Failure: " << QString(SDL_GetError());
+    } else {
+        if (SDL_HapticRumbleInit(_sdlHaptic) != 0) {
+            qDebug() << "SDL Haptic Rumble Init Failure: " << QString(SDL_GetError());
+        }
     }
-    SDL_HapticRumbleInit(_sdlHaptic);
 }
 
 Joystick::~Joystick() {
@@ -35,7 +38,9 @@ Joystick::~Joystick() {
 }
 
 void Joystick::closeJoystick() {
-    SDL_HapticClose(_sdlHaptic);
+    if (_sdlHaptic) {
+        SDL_HapticClose(_sdlHaptic);
+    }
     SDL_GameControllerClose(_sdlGameController);
 }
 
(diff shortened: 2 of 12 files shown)


SamGondelman:
Testing:
- With hand controllers connected (only the Vive controllers for now), call Controller.triggerHapticPulse, triggerHapticPulseOnDevice, Controller.triggerShortHapticPulse, and Controller.triggerShortHapticPulseOnDevice with various strengths (0-1), durations (in ms), and hands (0 = left, 1 = right, 2 or default = both).

hifi-gustavo:
Build console output for commit: b59d597780fd4f3ea7789551e0634395a34aa6c1 available at: https://gist.github.com/76096480c3b7008dace275f39d3e5c5f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8027.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8027.dmg

thoys:
Oops,

Your Ubuntu build seems to have been failed with the following warning:
```
[ 82%] Building CXX object libraries/entities-renderer/CMakeFiles/entities-renderer.dir/src/RenderableShapeEntityItem.cpp.o
/data/jenkins/Jenkins/workspace/beta_pr-master/label/ubuntu/plugins/hifiSdl2/src/Joystick.cpp: In constructor ‘Joystick::Joystick(SDL_JoystickID, SDL_GameController*)’:
/data/jenkins/Jenkins/workspace/beta_pr-master/label/ubuntu/plugins/hifiSdl2/src/Joystick.cpp:28:30: warning: format not a string literal and no format arguments [-Wformat-security]
         qDebug(SDL_GetError());
                              ^
```

hifi-gustavo:
Build console output for commit: 7a3cba8580ab1fcfd1bcb04b18007a69b9785a4d available at: https://gist.github.com/76096480c3b7008dace275f39d3e5c5f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8027.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8027.dmg

thoys:
## QA test report

I have gone to ultimate's with this test, since it needs to work good. So I am a bit strict. I have tested with small and big amounts of time. Short times seems to all work fine. (e.g. 1 second) , but when I start to do a vibration for 10 seconds and have a force above 0.75 the controllers starts switching vibrations on and off back and forward in very random patterns.

The function calls work all great. My Vive had controller ID 3 and that worked well for both hands and separate.

I cannot give this QA Test a pass until we figure out if this is a bug within the Vive API itself or a implementation error.

By the way, I have also tried calling one of the functions for 60 seconds, and then I overwrote it with a 0 strength to cancel it out, but this didn't seem to stop the flashing on of the lights on the Vive Controllers.

:-1: Sorry QA Fail for now :disappointed:         ( Better Luck :four_leaf_clover:  next time :+1:  )

SamGondelman:
^ That ^ should fix the flashing lights and other issues related to setting the strength to 0.

As for the irregular pulses, the Vive API (https://github.com/ValveSoftware/openvr/wiki/IVRSystem::TriggerHapticPulse) is extraordinarily limited.  It only allows single pulses of up to 4 ms, and then it can't produce another pulse for 5 ms.  So I believe the irregularity is either a result of dropped function calls or temporarily loosing hand tracking.  The support for longer pulses is something I threw together on top of that.  So I'm afraid this is the best we can do (as far as I can tell).

However, on both the Touch and SDL gamepads, once the APIs or hardware issues or whatever issues exist are resolved, that won't be a problem because of how their APIs work.

plugins/hifiSdl2/src/Joystick.cpp
```diff
@@ -21,16 +21,21 @@ Joystick::Joystick(SDL_JoystickID instanceId, SDL_GameController* sdlGameControl
...
     _sdlGameController(sdlGameController),
     _sdlJoystick(SDL_GameControllerGetJoystick(_sdlGameController)),
+    _sdlHaptic(SDL_HapticOpenFromJoystick(_sdlJoystick)),
     _instanceId(instanceId)
 {
-    
+    if (!_sdlHaptic) {
+        qDebug() << QString(SDL_GetError());
+    }
+    SDL_HapticRumbleInit(_sdlHaptic);
```
jherico: this should be in an else clause, and technically we should be checking the return value for an error as well.

plugins/hifiSdl2/src/Joystick.cpp
```diff
@@ -21,16 +21,21 @@ Joystick::Joystick(SDL_JoystickID instanceId, SDL_GameController* sdlGameControl
...
 }
 
 Joystick::~Joystick() {
     closeJoystick();
 }
 
 void Joystick::closeJoystick() {
+    SDL_HapticClose(_sdlHaptic);
```
jherico: this should be in an `if (_sdlHaptic)` block.

plugins/hifiSdl2/src/Joystick.cpp
```diff
@@ -21,16 +21,21 @@ Joystick::Joystick(SDL_JoystickID instanceId, SDL_GameController* sdlGameControl
         InputDevice("GamePad"),
     _sdlGameController(sdlGameController),
     _sdlJoystick(SDL_GameControllerGetJoystick(_sdlGameController)),
+    _sdlHaptic(SDL_HapticOpenFromJoystick(_sdlJoystick)),
     _instanceId(instanceId)
 {
-    
+    if (!_sdlHaptic) {
+        qDebug() << QString(SDL_GetError());
```
jherico: This should be a more descriptive error

qDebug() << "SDL Haptic Init Failure: " << SDL_GetError();

hifi-gustavo:
Build console output for commit: ff2405437bbbda73d4caaf6df2efca9c94709be1 available at: https://gist.github.com/76096480c3b7008dace275f39d3e5c5f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8027.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8027.dmg

libraries/controllers/src/controllers/InputDevice.h
```diff
@@ -55,6 +61,9 @@ class InputDevice {
 
     const QString& getName() const { return _name; }
 
+    // By default, Input Devices do not support haptics
+    virtual bool triggerHapticPulse(float strength, float duration, controller::Hand hand) { return false; }
```
jherico: I would rather see the haptic functionality exposed similarly to the way we expose input values on controllers, except as outputs instead.  i.e. they should be controls on a device to which people can write in routes.

plugins/openvr/src/ViveControllerManager.cpp
```diff
@@ -442,6 +452,55 @@ void ViveControllerManager::InputDevice::handlePoseEvent(float deltaTime, const
...
+        _system->GetTrackedDeviceClass(deviceIndex) == vr::TrackedDeviceClass_Controller &&
+        _trackedDevicePose[deviceIndex].bPoseIsValid) {
+        float strength = leftHand ? _leftHapticStrength : _rightHapticStrength;
+        float duration = leftHand ? _leftHapticDuration : _rightHapticDuration;
+
+        // Vive Controllers only support duration up to 4 ms, which is short enough that any variation feels more like strength
+        const float MAX_HAPTIC_TIME = 3999.0f; // in microseconds
+        float hapticTime = strength * MAX_HAPTIC_TIME;
+        if (hapticTime < duration * 1000.0f) {
+            _system->TriggerHapticPulse(deviceIndex, 0, hapticTime);
```
jherico: bear in mind the wiki says `Note: After this call the application may not trigger another haptic pulse on this controller and axis combination for 5ms.`

hifi-gustavo:
Build console output for commit: ff2405437bbbda73d4caaf6df2efca9c94709be1 available at: https://gist.github.com/76096480c3b7008dace275f39d3e5c5f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8027.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8027.dmg

plugins/openvr/src/ViveControllerManager.cpp
```diff
@@ -442,6 +452,55 @@ void ViveControllerManager::InputDevice::handlePoseEvent(float deltaTime, const
...
+
+    if (_system->IsTrackedDeviceConnected(deviceIndex) &&
+        _system->GetTrackedDeviceClass(deviceIndex) == vr::TrackedDeviceClass_Controller &&
+        _trackedDevicePose[deviceIndex].bPoseIsValid) {
+        float strength = leftHand ? _leftHapticStrength : _rightHapticStrength;
+        float duration = leftHand ? _leftHapticDuration : _rightHapticDuration;
+
+        // Vive Controllers only support duration up to 4 ms, which is short enough that any variation feels more like strength
+        const float MAX_HAPTIC_TIME = 3999.0f; // in microseconds
+        float hapticTime = strength * MAX_HAPTIC_TIME;
```
jherico: this system seems confusing and error prone.  I suggest given the Steam API, you instead take the following approach...

when you've received a request for a haptic pulse of a given duration and strength, calculate the the end time of the pulse, and the length of the individual pulses and store them.  Then in the update, ever 5 or 6 milliseconds, if the end-time is not in the past, re-trigger the pulse for the current strength duration.

libraries/controllers/src/controllers/InputDevice.h
SamGondelman: I agree, but I feel like this is sufficient for now given that there are no devices with other output values and since so many people would like this functionality exposed as soon as possible

plugins/openvr/src/ViveControllerManager.cpp
SamGondelman: I can add a check for this but it seems like attempting to trigger a pulse would be better than deciding to not trigger a pulse without any verification of whether or not it would have worked.

plugins/openvr/src/ViveControllerManager.cpp
SamGondelman: I'm not sure I understand the difference, besides counting down vs checking an end-time.  Counting down makes it a little easier to compare against new triggered pulses to determine precedence.

I can store the strength as MAX_HAPTIC_TIME*strength instead of calculating hapticTime in here if that's what you mean but that seems like it would be inconsistent between device managers.

hifi-gustavo:
Build console output for commit: 2e71a635749350960c2c5d8b171b18a2391b77c0 available at: https://gist.github.com/76096480c3b7008dace275f39d3e5c5f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8027.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8027.dmg

thoys:
## QA Testing Report

I have re-tested the commands. Making it 0 kills the vibrations now which is good. For the long vibrations I guess we'd just have to make timers to keep sending vibrations which works for me.

I am giving this QA a pass :+1: , great work Sam!

libraries/controllers/src/controllers/InputDevice.h
ZappoMan: here's my quick take on this... although I like the idea of having the controllers support "outputs" like they do inputs... the challenge is that the haptic APis have 2 parameters (strength and duration) and I'm not sure how that works with the writeable output...

Do you have a suggestion for how to address that @jherico ?

jherico: I'd probably implement it as two output endpoints, one for strength and one for duration.  The strength value would be persistent, keeping the last value written, while the duration would decay over time.  
the plugin would be responsible for making the requisite calls to the device in the update handler.

plugins/openvr/src/ViveControllerManager.cpp
jherico: My only concern with that is that the API doesn't specify the behavior if you violate the constraint.  Maybe it is a no-op, but if the spec doesn't guarantee that, then it's undefined behavior, which could include anything up to the controller catching on fire.

However, for a first pass this is fine.

Issue #7900, opened 2016-05-14
Xbox one wireless controller stops working after it did fell asleep and you wake it.


Richardus:
When you use the xbox one wireless controller and it's going into sleep mode to save battery or you shutdown the controller manual, and after that turn the xbox controller on again it's not working anymore.

Steps to test it:

-     Have wireless .xbox one windows controller
-     Start high fidelity. 
-     Test controller.
-     Turn controller off (simulates standby sleep mode)
-     Turn controller on

No response after this test in interface

See topic for more info. https://forums.highfidelity.com/t/bug-xbox-one-controller-wireless-stops-responding/10515/3

Richardus:
Build 4970 , Interface is crashing now complete. after your turned the xbox controller off and on again.

Richardus:
Fixed with Build 5029. It works fine after turning it off and on again.

PR #8033, opened 2016-06-07
Game pads work after falling asleep and being reactivated

SamGondelman:
Addresses: https://github.com/highfidelity/hifi/issues/7900 and https://forums.highfidelity.com/t/bug-xbox-one-controller-wireless-stops-responding/10515

### DIFF
--- a/libraries/controllers/src/controllers/UserInputMapper.h
+++ b/libraries/controllers/src/controllers/UserInputMapper.h
@@ -141,9 +141,6 @@ namespace controller {
         std::vector<Pose> _poseStates = std::vector<Pose>(toInt(Action::NUM_ACTIONS));
         std::vector<float> _lastStandardStates = std::vector<float>();
 
-        int recordDeviceOfType(const QString& deviceName);
-        QHash<const QString, int> _deviceCounts;
-
         static float getValue(const EndpointPointer& endpoint, bool peek = false);
         static Pose getPose(const EndpointPointer& endpoint, bool peek = false);
 
--- a/libraries/controllers/src/controllers/UserInputMapper.h
+++ b/libraries/controllers/src/controllers/UserInputMapper.h
@@ -142,7 +142,7 @@ namespace controller {
         std::vector<float> _lastStandardStates = std::vector<float>();
 
         int recordDeviceOfType(const QString& deviceName);
-        QHash<const QString&, int> _deviceCounts;
+        QHash<const QString, int> _deviceCounts;
 
         static float getValue(const EndpointPointer& endpoint, bool peek = false);
         static Pose getPose(const EndpointPointer& endpoint, bool peek = false);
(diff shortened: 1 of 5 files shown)


SamGondelman:
Testing:

- Start Interface with a game pad connected and on, and the SDL input plugin activated-
- Move around with controller
- Turn controller off
- Turn controller on
- Controller should still work

hifi-gustavo:
Build console output for commit: 545dda0a9887652942e6368da1be64a7ace3b441 available at: https://gist.github.com/272d84bf7c17f1d3d586290d97f71210<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8033.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8033.dmg

libraries/controllers/src/controllers/UserInputMapper.h
```diff
@@ -142,7 +142,7 @@ namespace controller {
         std::vector<float> _lastStandardStates = std::vector<float>();
 
         int recordDeviceOfType(const QString& deviceName);
-        QHash<const QString&, int> _deviceCounts;
+        QHash<const QString, int> _deviceCounts;
```
ZappoMan: if you remove the reference you might as well remove the const too...

plugins/hifiSdl2/src/Joystick.cpp
```diff
@@ -64,7 +64,7 @@ void Joystick::handleButtonEvent(const SDL_ControllerButtonEvent& event) {
 
 controller::Input::NamedVector Joystick::getAvailableInputs() const {
     using namespace controller;
-    static const Input::NamedVector availableInputs{
```
ZappoMan: what's this change for?

SamGondelman: When a gamepad is disconnected and reconnected, it gets a new deviceID (also if you have multiple game pads connected at once).  Since makePair creates a new Input using _deviceID, they weren't properly creating their own new Inputs because availableInputs was statically initialized with only the original deviceID.

libraries/controllers/src/controllers/UserInputMapper.h
SamGondelman: Actually _deviceCounts doesn't seem to be used for anything at all.  Should I just get rid of it, or keep it in case it ends up being used for something later?

plugins/hifiSdl2/src/Joystick.cpp
ZappoMan: I see... but how often does `getAvailableInputs()` get called when the ID hasn't changed? Maybe we can cache the result unless/until the deviceID changes?

libraries/controllers/src/controllers/UserInputMapper.h
ZappoMan: if it's not used at all let's get rid of it...

ZappoMan: hmmmm - what was it used for before??? I wonder....

hifi-gustavo:
Build console output for commit: 2c1d20bd1af4abf59a30f86125f5b2c24210b769 available at: https://gist.github.com/272d84bf7c17f1d3d586290d97f71210<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8033.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8033.dmg

PR #8042, opened 2016-06-09
First pass prototype of laser rendering in OpenVR plugin

jherico:
This is an implementation of hand lasers rendered by the compositor, so that they are not obscured by the UI overlay.  

## Display plugin API changes

New functions in the DisplayPlugin class allow you to set set the laser mode for each hand using the function 

    DisplayPlugin::setHandLaser(
        uint32_t hands,
        HandLaserMode mode,
        vec4 color = vec4(1),
        vec3 direction = vec3(0, 0, -1))

The first parameter is a bitfield, where 1 represents the left hand and 2 represents the right hand.  So if you want to set the lasers for both hands, pass 3.

The second parameter is the laser mode, where the currently supported modes are `None` and `Overlay`.  None disables rendering of the laser, while `Overlay will show a laser from the hand to the intersection point with the overlay sphere.  

The third parameter is the color (including alpha) with which to render the laser.  

The fourth and last parameter is the direction of the laser, expressed in the hand pose reference frame.  A vector of 0,0,-1 will render a laser perfectly aligned with the Steam overlay lasers.  A vector of (1, 0, 0) will render pointing to the right from the controller.

## Scripting API changes

New functions in HMDScriptingInterface allow the lasers to be enabled and disabled by scripts.

    bool HMDScriptingInterface::setHandLasers(
        int hands, bool enabled, const glm::vec4& color, const glm::vec3& direction) const

This is an alias for calling `DisplayPlugin::setHandLaser` on the current display plugin.  In theory, the return value should indicate whether the display plugin can currently support hand lasers, but right now it will return true for all HMD derived plugins and false for all other plugins.  

    void HMDScriptingInterface::disableHandLasers(
        int hands) const

This is a convenience method for disabling the laser on one or both hands, so that you don't have to pass the color and direction parameter when turning the laser off.

The bulk of this functionality is expressed in the `HmdDisplayPlugin` class.  If a derived class supports hand controllers and these controllers are active, then it can enable rendering of the lasers by the `HmdDisplayPlugin` class by populating the `glm::mat4 _handPoses[2]` member variable with non-identity matrices in it's `beginFrameRender` method (see for example the changes to the OpenVrDisplayPlugin.cpp class).

### DIFF
--- a/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.h
+++ b/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.h
@@ -74,6 +74,7 @@ protected:
     virtual void compositeScene();
     virtual void compositeOverlay();
     virtual void compositePointer();
+    virtual void compositeExtra() {};
 
     virtual bool hasFocus() const override;
 
--- a/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.h
+++ b/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.h
@@ -152,7 +152,7 @@ protected:
     }
 
 private:
-    // Any resource shared by the main thread and the presntaion thread must 
+    // Any resource shared by the main thread and the presentation thread must
     // be serialized through this mutex
     mutable Mutex _presentMutex;
     ProgramPtr _activeProgram;
--- a/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.h
+++ b/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.h
@@ -109,7 +109,6 @@ protected:
     int32_t _alphaUniform { -1 };
     ShapeWrapperPtr _plane;
 
-    mutable Mutex _mutex;
     RateCounter<> _droppedFrameRate;
     RateCounter<> _newFrameRate;
     RateCounter<> _presentRate;
@@ -135,7 +134,27 @@ protected:
     BasicFramebufferWrapperPtr _compositeFramebuffer;
     bool _lockCurrentTexture { false };
 
+    void assertIsRenderThread() const;
+    void assertIsPresentThread() const;
+
+    template<typename F>
+    void withPresentThreadLock(F f) const {
+        assertIsPresentThread();
+        Lock lock(_presentMutex);
+        f();
+    }
+
+    template<typename F>
+    void withRenderThreadLock(F f) const {
+        assertIsRenderThread();
+        Lock lock(_presentMutex);
+        f();
+    }
+
 private:
+    // Any resource shared by the main thread and the presntaion thread must 
+    // be serialized through this mutex
+    mutable Mutex _presentMutex;
     ProgramPtr _activeProgram;
 };
 
--- a/libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
+++ b/libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
@@ -316,9 +316,8 @@ void HmdDisplayPlugin::compositePointer() {
         Uniform<glm::mat4>(*_program, _mvpUniform).Set(mvp);
         _plane->Draw();
     });
-
-    compositeLasers();
 }
+
 
 void HmdDisplayPlugin::internalPresent() {
 
@@ -412,24 +411,7 @@ bool HmdDisplayPlugin::setHandLaser(uint32_t hands, HandLaserMode mode, const ve
     return true;
 }
 
-// FIXME try to consolidate the duplication of logic between this function and a similar one in CompsitorHelper.
-static float calculateRayUiCollisionDistance(const glm::mat4& headPose, const glm::vec3& position, const glm::vec3& direction) {
-    auto relativePosition4 = glm::inverse(headPose) * vec4(position, 1);
-    auto relativePosition = vec3(relativePosition4) / relativePosition4.w;
-    auto relativeDirection = glm::inverse(glm::quat_cast(headPose)) * direction;
-    if (glm::abs(glm::length2(relativeDirection) - 1.0f) > EPSILON) {
-        relativeDirection = glm::normalize(relativeDirection);
-    }
-    // FIXME fetch the actual UI radius from... somewhere?
-    float uiRadius = 1.0f;
-    float instersectionDistance;
-    if (!glm::intersectRaySphere(relativePosition, relativeDirection, vec3(0), uiRadius * uiRadius, instersectionDistance)) {
-        return -1;
-    }
-    return instersectionDistance;
-}
-
-void HmdDisplayPlugin::compositeLasers() {
+void HmdDisplayPlugin::compositeExtra() {
     std::array<HandLaserInfo, 2> handLasers;
     std::array<mat4, 2> renderHandPoses;
     withPresentThreadLock([&] {
@@ -465,10 +447,16 @@ void HmdDisplayPlugin::compositeLasers() {
         const auto& laserDirection = handLaser.direction;
         auto model = renderHandPoses[i];
         auto castDirection = glm::quat_cast(model) * laserDirection;
+        if (glm::abs(glm::length2(castDirection) - 1.0f) > EPSILON) {
+            castDirection = glm::normalize(castDirection);
+        }
+
+        // FIXME fetch the actual UI radius from... somewhere?
+        float uiRadius = 1.0f;
 
         // Find the intersection of the laser with he UI and use it to scale the model matrix
-        float distance = calculateRayUiCollisionDistance(_currentPresentFrameInfo.presentPose, vec3(renderHandPoses[i][3]), castDirection);
-        if (distance < 0) {
+        float distance; 
+        if (!glm::intersectRaySphere(vec3(renderHandPoses[i][3]), castDirection, vec3(0), uiRadius * uiRadius, distance)) {
             continue;
         }
 
--- a/libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
+++ b/libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
@@ -408,9 +408,11 @@ bool HmdDisplayPlugin::setHandLaser(uint32_t hands, HandLaserMode mode, const ve
         }
     });
     // FIXME defer to a child class plugin to determine if hand lasers are actually 
+    // available based on the presence or absence of hand controllers
     return true;
 }
 
+// FIXME try to consolidate the duplication of logic between this function and a similar one in CompsitorHelper.
 static float calculateRayUiCollisionDistance(const glm::mat4& headPose, const glm::vec3& position, const glm::vec3& direction) {
     auto relativePosition4 = glm::inverse(headPose) * vec4(position, 1);
     auto relativePosition = vec3(relativePosition4) / relativePosition4.w;
@@ -418,6 +420,7 @@ static float calculateRayUiCollisionDistance(const glm::mat4& headPose, const gl
     if (glm::abs(glm::length2(relativeDirection) - 1.0f) > EPSILON) {
         relativeDirection = glm::normalize(relativeDirection);
     }
+    // FIXME fetch the actual UI radius from... somewhere?
     float uiRadius = 1.0f;
     float instersectionDistance;
     if (!glm::intersectRaySphere(relativePosition, relativeDirection, vec3(0), uiRadius * uiRadius, instersectionDistance)) {
--- a/libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
+++ b/libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
@@ -9,6 +9,7 @@
 
 #include <memory>
 #include <glm/gtc/matrix_transform.hpp>
+#include <glm/gtx/intersect.hpp>
 
 #include <QtCore/QLoggingCategory>
 #include <QtWidgets/QApplication>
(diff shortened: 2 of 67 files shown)


hifi-gustavo:
Build console output for commit: 84c394609b23cc35d6d5156413e1476ae38cc71e available at: https://gist.github.com/0d9c0180244f799b9ac9827d4ee3b28f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.dmg

hifi-gustavo:
Build console output for commit: ad58433957badb8628d5a54c5641e27ec748a7a4 available at: https://gist.github.com/0d9c0180244f799b9ac9827d4ee3b28f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.dmg

plugins/openvr/src/OpenVrDisplayPlugin.cpp
```diff
@@ -181,7 +214,7 @@ void OpenVrDisplayPlugin::hmdPresent() {
 void OpenVrDisplayPlugin::postPreview() {
     PROFILE_RANGE_EX(__FUNCTION__, 0xff00ff00, (uint64_t)_currentPresentFrameIndex)
 
-    vr::TrackedDevicePose_t currentTrackedDevicePose[vr::k_unMaxTrackedDeviceCount];
+        vr::TrackedDevicePose_t currentTrackedDevicePose[vr::k_unMaxTrackedDeviceCount];
```
howard-stearns: whitespace

howard-stearns:
In addition to having QA do performance testing in vive at some point, it looks like we should also do performance testing on a weaker 2d machine, yes?

RyanDowne:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;GkMpj61l&#x2F;21-make-laser-appear-on-overlays-and-be-able-to-press-buttons-etc"> Make laser appear on overlays and be able to press buttons etc</a>

hifi-gustavo:
Build console output for commit: da2a6570ca16d98aa71cc93c27c28006dfaec14d available at: https://gist.github.com/0d9c0180244f799b9ac9827d4ee3b28f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.dmg

hifi-gustavo:
Build console output for commit: 9917f7ffcd7394e362b5af6eb8939a429d427c41 available at: https://gist.github.com/0d9c0180244f799b9ac9827d4ee3b28f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.dmg

hifi-gustavo:
Build console output for commit: 0a8d9e0b890d93feb691c0284199a5b80a9a0e81 available at: https://gist.github.com/0d9c0180244f799b9ac9827d4ee3b28f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.dmg

libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.h
```diff
@@ -135,7 +134,27 @@ class OpenGLDisplayPlugin : public DisplayPlugin {
...
+
+    template<typename F>
+    void withRenderThreadLock(F f) const {
+        assertIsRenderThread();
+        Lock lock(_presentMutex);
+        f();
+    }
+
 private:
+    // Any resource shared by the main thread and the presntaion thread must
```
howard-stearns: typo "presentation". Not a blocker.

howard-stearns:
Should the locking be separated out and cherry-picked into master?
(I imagine they're just an opportunity for clean living, and not in response to a known problem, and therefore can wait until we merge back, but I'm not certain.)

howard-stearns:
If it makes sense, please create a test plan for QA that describes turning on lasers and what should be seen, and mentions appropriate performance checks for both use in vive (w/ and w/o lasers), and for non-hmd desktop.

libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
```diff
@@ -343,22 +376,122 @@ void HmdDisplayPlugin::setEyeRenderPose(uint32_t frameIndex, Eye eye, const glm:
...
+    info.direction = direction;
+    withRenderThreadLock([&] {
+        if (hands & Hand::LeftHand) {
+            _handLasers[0] = info;
+        }
+        if (hands & Hand::RightHand) {
+            _handLasers[1] = info;
+        }
+    });
+    // FIXME defer to a child class plugin to determine if hand lasers are actually
```
howard-stearns: "are actually..." what? (not a blocker)

libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
```diff
@@ -343,22 +376,122 @@ void HmdDisplayPlugin::setEyeRenderPose(uint32_t frameIndex, Eye eye, const glm:
...
+    if (glm::abs(glm::length2(relativeDirection) - 1.0f) > EPSILON) {
+        relativeDirection = glm::normalize(relativeDirection);
+    }
+    float uiRadius = 1.0f;
+    float instersectionDistance;
+    if (!glm::intersectRaySphere(relativePosition, relativeDirection, vec3(0), uiRadius * uiRadius, instersectionDistance)) {
+        return -1;
+    }
+    return instersectionDistance;
+}
```
howard-stearns: It would be cool if we could refactor this with _oculusUIRadius, calculateRayUICollisionPoint, sphereSurfaceFromOverlay, and the like. (Maybe then also expose that 1.0 radius to Javascript.)
If not now (which I'm ok with), then maybe note with a fixme?

hifi-gustavo:
Build console output for commit: d3337bf1d7497b8b6bc14bbc6f1b7ae463957c8f available at: https://gist.github.com/0d9c0180244f799b9ac9827d4ee3b28f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.dmg

libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
jherico: fixed.

libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
jherico: added notes

jherico:
## Testing

Testing should consist of ensuring that the application still has roughly the same performance characteristics as the current dev build, and ensuring that the new laser scripting functions work as advertised.

The two new functions `HMD.disableHandLasers` and `HMD.setHandLasers` both take bitfields as their first argument, so `HMD.disableHandLasers(3);` should disable lasers for both hands, while `HMD.disableHandLasers(1);` would do it for the left hand and `HMD.disableHandLasers(2);` would do it for the right hand.  

The following code should create a left hand red laser pointing 'up' from the hand, and a right hand blue laser pointing forward from the hand controller.

```
var LEFT_HAND = 1;
var RIGHT_HAND = 2;
HMD.setHandLasers(RIGHT_HAND, true, { x: 0, y: 0.5, z: 0.8, w: 1.0 }, { x: 0, y: 0, z: -1 });
HMD.setHandLasers(LEFT_HAND, true, { x: 1.0, y: 0, z: 0, w: 1.0 }, { x: 0, y: 1, z: -1 });
```

In the same script you could then use `HMD.disableHandLasers(LEFT_HAND | RIGHT_HAND);` to disable both lasers.

howard-stearns:
It looks as though the condition necessary to show the laser is currently two-fold:
  1. Turn the laser on, as documented with HMD.setHandLasers,
AND
  2. Ensure that Reticle.visible is true.

Is that second one a bug (possibly a bug in my interpretation of what I'm seeing)?
The driver for this is that we want the laser beam INSTEAD of the current laser pointer. So how would we do that if the cursor has to be showing to see the laser?

howard-stearns:
Also, it's hard for me to tell, but to my eyes (through thick glasses with astigmatism correction, etc.), it looks like the termination of the laser is not at the HUD. In some hand positions it appears to be stopping short of the QML dialog I'm pointing at, and in other hand positions it appears to be going through the QML dialog. Am I crazy?

howard-stearns:
I bet I know what the laser-termination problem is. I bet the laser and the sphere are being computed with different centers.  (In addition to the current issue, once this and something like #8043 are both merged, we'll have to re-coordinate with @hyperlogic about what center is being used.

howard-stearns:
I've made a DNM pr that illustrates the above issues. Run the handControllerPointer script from https://github.com/highfidelity/hifi/pull/8051 (instead of the normal one) to show how there is an offset between our handControllerGrab laser and this laser. Visible when pointing your hand at an overlay. You can partially squeeze the trigger when pointing slightly away from the overlay, and you will see this PR's laser as you swing your hand onto a HUD overlay.

To illustrate that this laser doesn't shown the reticle is off, you can turn it off manually in the console, or edit the
   Reticle.visible = true;
on line 468 to instead set false. You will get neither the Reticle nor the laser.

Also, we will need to have a way for a caller of HMD.setHandLasers to learn whether the lasers are actually supported by the active plugin. (Perhaps by answering true only when supported.) That way, the script can fall back to drawing the Reticle when the laser is not supported.

jherico:
>  Also, we will need to have a way for a caller of HMD.setHandLasers to learn whether the lasers are actually supported by the active plugin. (Perhaps by answering true only when supported.) 

That's the plan and why the new scripting methods return booleans.

> Ensure that Reticle.visible is true.

That's a bug....  the hand laser rendering code is called from inside the compositePointer method, so if no pointer is enabled it doesn't get called.  Will fix.

>  Also, it's hard for me to tell, but to my eyes (through thick glasses with astigmatism correction, etc.), it looks like the termination of the laser is not at the HUD

I'm going to add a small sphere with 0.3 alpha at the end of the laser.  This should make it easier to tell if it's terminating in the right point, since it should become a hemisphere if it's intersecting a 2D UI element.

thoys:
## QA Test report

Performance good getting 90 FPS at hifi://Playa while the lasers are visible. Both new script functions work fine. 

Noticing the same as @howard-stearns : The lasers are only visible when the cursor is displayed. Also are the lasers supposed to be visible through your avatars hands?

QA Fail for now

hifi-gustavo:
Build console output for commit: 130c0dda318335d70e1d26de6625fb1bc46112ee available at: https://gist.github.com/0d9c0180244f799b9ac9827d4ee3b28f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.dmg

howard-stearns:
It looks like the following issues are still outstanding, correct?:
 - Only works when reticle is visible.
 - Has offset from our hand position.
 - Might be wrong length (e.g., wrong origin), of which we expect the following item to help test.
 - Visualization of end point, to (eventually?) be controllable from script as to whether or not appears.

But it looks like one of the issues was resolved at some point, and should be correct now:
 - HMD.setHandLasers to answer true when it does something, and false when it doesn't (e.g., desktop plugin, I have not yet tested).

hifi-gustavo:
Build console output for commit: 62051ad297056a5fd0e07966c9ea0cc108f6f3f4 available at: https://gist.github.com/0d9c0180244f799b9ac9827d4ee3b28f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.dmg

jherico:
> * Only works when reticle is visible.

Fixed 

> * Has offset from our hand position.

Fixed 

> * Might be wrong length (e.g., wrong origin) ...

Fixed, same issue as above

> HMD.setHandLasers to answer true when it does something, and false when it doesn't

Sort of fixed.  The HMDDisplayPlugin always returns true for this, even if we're internally suppressing the lasers because of something like the keyboard being visible, or no controllers being found.  We can iterate on it later to determine whether the return value should indicate immediate visibility or just support for the functionality.

hifi-gustavo:
Build console output for commit: ccd0e1eb155fdff5e9b7b95ef77edce940e6b3f6 available at: https://gist.github.com/0d9c0180244f799b9ac9827d4ee3b28f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.dmg

hifi-gustavo:
Build console output for commit: 6707f889b82d2e8fdaa065ac32e4d536f24b8dd6 available at: https://gist.github.com/0d9c0180244f799b9ac9827d4ee3b28f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8042.dmg

howard-stearns:
Right now, the laser length is wrong -- presumably because of the new additional transform used by the HUD. We'll deal with that separately.

PR #8043, opened 2016-06-09
HMD UI re-centering while driving improvements.

hyperlogic:
Previously the HUD fading in/out would also recenter the hmd sensor and the avatar, which caused many problems including:
  * The user's view could shift vertically.
  * Your avatar would briefly go into t-pose
  * other users would see your avatar go into t-pose.

Now we now move the UI sphere instead, which results in a much smoother experience.

## Code Changes

MyAvatar: 
* added hasDriveInput method.

OverlayConductor:
* removed avatar and sensor reset, instead the overlay's modelTransform is changed.
* revived STANDING mode, which is active if myAvatar->getClearOverlayWhenDriving() is true and you are wearing an HMD.
* SITTING & FLAT mode should be unchanged.
* Instead of using avatar velocity to fade out/fade in the hud, We use the presence or absence of avatar drive input.
* Additionally, we check distance to the UI sphere, and quickly recenter the hud if the users head is too close to the actual hud sphere.

CompositorHelper:
* Bug fixes for ray picks not using the sphere modelTransform.

HmdDisplayPlugin:
* Bug fixes for rendering not using the overlay's modelTransform.


hifi-gustavo:
Build console output for commit: 5ef6847dc35b047ffbced088d07de9bcc79bda8e available at: https://gist.github.com/d3d1033578a522c13b65cac576eec063<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.dmg

interface/src/ui/OverlayConductor.h
```diff
@@ -19,6 +19,7 @@ class OverlayConductor {
     void update(float dt);
     void setEnabled(bool enable);
     bool getEnabled() const;
+    bool shouldCenterUI() const;
```
howard-stearns: Looks like the build is complaining that _wantsOverlays on line 37, below, is no longer used.
This had been used to capture whether the user was displaying overlays before we fade out, so that we can restore to the same state after. E.g., if we're not displaying overlays before driving, there's nothing to fade out, and more importantly, we don't want the fade-back-in to restore overlays that weren't being shown.  I haven't tried this out yet, but it's not clear to me from the code how you're achieving this now.

interface/src/ui/OverlayConductor.cpp
```diff
@@ -22,10 +22,31 @@ OverlayConductor::OverlayConductor() {
...
+        return true;
+    }
+    return false;
+}
+
 void OverlayConductor::update(float dt) {
 
     updateMode();
 
+    MyAvatar* myAvatar = DependencyManager::get<AvatarManager>()->getMyAvatar();
```
howard-stearns: What is this used for?

RyanDowne:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;pszjSO9c&#x2F;13-don-t-reset-avatar-when-hud-is-reset"> Don&#39;t reset avatar when HUD is reset.</a>

interface/src/ui/OverlayConductor.cpp
hyperlogic: bool nowDriving = myAvatar->hasDriveInput();

hyperlogic:
Test cases:
  * Ensure the "Clear Overlay When Driving" is checked in Avatar settings, found in the Settings > Avatar... menu.
  * Open some UI dialogs on the HMD overlay, for example: press control-b to bring up a web browser.
  * While connected to a domain, press the arrow keys or the track pad on the right vive controller to move.
  * when you press these keys the overlay should smoothly fade away.
  * when you stop pressing these keys it should smoothly fade in, but centered around your current facing.
  * additionally, when the overlay is visible, if you lean your head in toward the surface of the overlay, it should reposition itself such that you're had is never on the outside of the UI sphere.

hifi-gustavo:
Build console output for commit: 6e3057479ffacde155330039d249b242210adb92 available at: https://gist.github.com/d3d1033578a522c13b65cac576eec063<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.dmg

howard-stearns:
Doesn't work in desktop mode.
When walking in room scale, the hud kept popping back in with each step. (Supposed to stay off until a moment after I stop moving.)
Always brings up keyboard when I stop moving. Any of the following might be some consequence of the keyboard:
Sometimes gets into a state where no clicks work.
Turn off overlays (e.g., using the toggle), drive, and stop. Then turn toggle back on. The overlays don't come back.

ZappoMan:
Why are we adding back STANDING and SITTING modes? I thought we were trying to avoid these types of mode changes.

hyperlogic:
> Doesn't work in desktop mode.

Please clearify, what is not working?  My understanding is that we do not want to fade/hide the UI in desktop mode when the user presses the drive keys.

hyperlogic:
> When walking in room scale, the hud kept popping back in with each step. (Supposed to stay off until a moment after I stop moving.)

@howard-stearns   Define walking?  Physically walking?  or hitting the drive keys.  The current implementation should recenter the hud UI around your head when it gets too close to the HUD sphere geometry.   Is this bad?  what other options do we have?  Does this not meet our number one sprint goal:  "The HUD always stays centered in your view when “on”"

hyperlogic:
> Always brings up keyboard when I stop moving. 

What? maybe that's due to some keyboard focus issues, what UI elements are in the hud?

hyperlogic:
> Sometimes gets into a state where no clicks work.

@howard-stearns I get that all the time, even without this PR.

hyperlogic:
> Turn off overlays (e.g., using the toggle), drive, and stop. Then turn toggle back on. The overlays don't come back.

How did you turn off overlays?  Using the button on the UI, or the menu option?

hyperlogic:
> Why are we adding back STANDING and SITTING modes? I thought we were trying to avoid these types of mode changes.

@ZappoMan The current option for clearOverlayWhenDriving is essentially the same as STANDING mode, so I left it named STANDING, mode I can change the name if you'd like.

hifi-gustavo:
Build console output for commit: 685710d0ec088b60b13fe7aef0ba3054cb67155e available at: https://gist.github.com/d3d1033578a522c13b65cac576eec063<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.dmg

hyperlogic:
I don't know why the linux build failed, lets try this again.   test this please.

hyperlogic:
@howard-stearns This PR should address your feedback.

hifi-gustavo:
Build console output for commit: 685710d0ec088b60b13fe7aef0ba3054cb67155e available at: https://gist.github.com/d3d1033578a522c13b65cac576eec063<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.dmg

hyperlogic:
I noticed an issue where sometimes, say 1/20 times when the overlay fades out, not all windows on the overlay "desktop" fade out, similarly sometimes some of the windows don't fade back in.  Perhaps this is an issue with the QML alpha fading? @jherico

howard-stearns:
As discussed:
- We still want to be able to walk without the overlays coming back after each step. (It should wait until we've been stopped for long enough without restarting.)
- There's a problem when manually bringing back the hud (using the button) while the "hud back on" timeout is winding down.

hifi-gustavo:
Build console output for commit: f6ed5a1dae57820a320fa53162960b600bebce47 available at: https://gist.github.com/d3d1033578a522c13b65cac576eec063<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.dmg

RyanDowne:
test this please

howard-stearns:
- The "Clear overlays on driving" is now only the room-scale vive behavior. The preference had been requested as a way to control the overall overlay-clearing behavior, regardless of mode.

- The behavior in room-scale vive is getting triggered too late, such that I am able to get my head outside the hud-sphere and look back on it for a bit. So that needs some tuning.

hifi-gustavo:
Build console output for commit: ccf85c1f7c34fde1d42f4a5ce1a01fbbc8aee56e available at: https://gist.github.com/d3d1033578a522c13b65cac576eec063<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.dmg

hifi-gustavo:
Build console output for commit: ab4bef7d55cbef9063ad0a53e85d5353e5fb66eb available at: https://gist.github.com/d3d1033578a522c13b65cac576eec063<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.dmg

hifi-gustavo:
Build console output for commit: 21e67fc4d844c8811231029fbb6fe60e24ac5dfc available at: https://gist.github.com/d3d1033578a522c13b65cac576eec063<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.dmg

hifi-gustavo:
Build console output for commit: 1ca1f98034731127e00954405b8c8df5e8a36aff available at: https://gist.github.com/d3d1033578a522c13b65cac576eec063<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.dmg

howard-stearns:
test this please

hifi-gustavo:
Build console output for commit: 21e67fc4d844c8811231029fbb6fe60e24ac5dfc available at: https://gist.github.com/d3d1033578a522c13b65cac576eec063<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8043.dmg

howard-stearns:
I've tested this. Pass.

thoys:
## QA Report

The test cases work fine. But I went and do something else. I called `Avatar->Reset Sensors` while in HMD, and the Overlays disappeared and never came back until I called the same menu item one or more times and it came back. Please repro this @howard-stearns .  QA fail until we sort this out

howard-stearns:
Good catch @thoys. It looks like this happens the first time you do menu->avatar->reset sensors, and then not again after that. And it looks like this problem does not happen on master. We're trying now to figure out whether we should address this in this PR, or separately later.

howard-stearns:
We've decided to look at the disappearing overlay issue separately.

thoys:
QA successful it is then :+1:

PR #8030, opened 2016-06-07
fix paused interactions with mouse

howard-stearns:
- Paused desktop wasn't showing mouse (which one would want for doing things while paused).
- Hand controllers were driving mouse while paused (which they shouldn't).

PR #8051, opened 2016-06-10
use system laser

howard-stearns:
Use the new system laser pointer from #8042.

This is DNM because it is meant to be after #8042 and #8030 land, and because it illustrates some issues with #8042:
1.  This PR shows BOTH the mouse Reticle and the system laser, but we really only want the later. Right now, if you turn Reticle.visible = false, the system laser will disappear.  We'll also need a way to know whether HMD.setHandLasers is supported by the active display plugin, so that we can fall back properly to using Reticle.
2. The system pointer has an offset from the hand, so the laser is parallel to where we're pointing, but not going through the same point.


hifi-gustavo:
Build console output for commit: dfd03d5e6184ffb7a0f3b9a0b81e584dcf78e2a6 available at: https://gist.github.com/feba1909ba6532b6e098ebdfe2bf9ff2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8051.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8051.dmg

howard-stearns:
Closing in favor of #8064

PR #8053, opened 2016-06-12
Update BUILD_OSX.md

audreyt:
Qt5.6 removed QTWebKit which hifi depends on. Before we complete the migration listed in https://wiki.qt.io/Porting_from_QtWebKit_to_QtWebEngine the build instruction need to tap into qt55 in historical brew versions.


hifi-gustavo:

PR #8054, opened 2016-06-13
fix model export continued

ctrlaltdavid:


ctrlaltdavid:
Test Plan

Continuing on from https://github.com/highfidelity/hifi/pull/8044

Fixes

1) Default directory for browsing to import / export should be desktop directory.
If you change the directory and use it to import / export this should be remembered for next time you browse to import / export (or other script usages of Window.browse() that don't specify a directory), including between program runs.
To reset the browse directory, delete the "LastBrowseLocation" from interface.ini.
Test also that can successfully load and save recording files in \scripts\developer\utilities\record\recorder.js (if recorder.js is considered to be working - I seem to recall hearing that it isn't?).

2) Make the default directory for preferences such as Snapshots directory be the current preference value.
Note: The Snapshots directory is not saved between program runs. This broke in build 4987, unrelated to the file browser dialog. I've filed a FogBugz report on this: https://highfidelity.fogbugz.com/f/cases/905/Snapshots-directory-is-not-remembered However, for the purposes of testing this PR, Settings > General > "Load scripts from this directory" is saved.

3) The default selection when choosing a directory is the default directory.
This makes the starting point consistent with subsequent navigation to choose a directory.

4) Fix message box that occurs if you type in a path and filename to export a file to and the directory doesn't exist.
The message box used to say, "Unable to write to location <path>" with "Yes" and "No" buttons, which didn't make sense. Now it has just a single "OK" button.


Not Fixed

>>> I am accustomed to being able to click on an entry in the file list and that would subsequently appear in the input field for editing. This is not the case.

I had a bit of a look at implementing this but it's more involved than first appears. It's a good idea though!
https://highfidelity.fogbugz.com/f/cases/906/When-exporting-a-model-should-be-able-to-click-on-existing-JSON-file

>>> I typed C:\dodecahedron.json in the expectation that it would save dodecahedron.json in c:\ it did not. (daft thing to do but worth a check)

It works on drives other than C:. It fails silently on C:. This is a Windows permissions error: when I try to save a text file to the root of C: I get a permissions error. It shouldn't fail silently though!
https://highfidelity.fogbugz.com/f/cases/907/Export-entity-fails-silently-if-can-t-write-file

hifi-gustavo:
Build console output for commit: 098432152dd7d65590d0f00a4f08fd971a5def27 available at: https://gist.github.com/318fd9738ed0d112e274c826dbd9dc02<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8054.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8054.dmg

interface/src/scripting/WindowScriptingInterface.cpp
```diff
@@ -101,15 +107,32 @@ QString fixupPathForMac(const QString& directory) {
...
 /// Display an open file dialog.  If `directory` is an invalid file or directory the browser will start at the current
 /// working directory.
 /// \param const QString& title title of the window
 /// \param const QString& directory directory to start the file browser at
 /// \param const QString& nameFilter filter to filter filenames by - see `QFileDialog`
 /// \return QScriptValue file path as a string if one was selected, otherwise `QScriptValue::NullValue`
 QScriptValue WindowScriptingInterface::browse(const QString& title, const QString& directory, const QString& nameFilter) {
-    QString path = fixupPathForMac(directory);
+    QString path = directory;
+    if (path == "") {
```
ZappoMan: wouldn't `!path.isEmpty()` be better?

interface/src/scripting/WindowScriptingInterface.cpp
```diff
@@ -120,8 +143,17 @@ QScriptValue WindowScriptingInterface::browse(const QString& title, const QStrin
 /// \param const QString& nameFilter filter to filter filenames by - see `QFileDialog`
 /// \return QScriptValue file path as a string if one was selected, otherwise `QScriptValue::NullValue`
 QScriptValue WindowScriptingInterface::save(const QString& title, const QString& directory, const QString& nameFilter) {
-    QString path = fixupPathForMac(directory);
+    QString path = directory;
+    if (path == "") {
```
ZappoMan: maybe `!path.isEmpty()`

hifi-gustavo:
Build console output for commit: 5553c9f87fd0bcc6c8062aa8c7ddef80372aeaa6 available at: https://gist.github.com/318fd9738ed0d112e274c826dbd9dc02<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8054.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8054.dmg

Nex-Pro:
**PR Build:** #8054
**CPU:** 4.0 ghz i7
**RAM:** 16 GB
**GPU:** Nvidia gtx770 2gb
**Testing OS:** win 10 pro
**HMD:** None
**Hand Controllers:** None

### PASS

**1.**
Default directory for browsing to import / export should be desktop directory. **SUCCESS**

Changed to directory persists across restarts for import/export of entities and script loading. **SUCCESS**

**2.**
Changing directory for scripts persists. **SUCCESS**

Changing directory for snapshots behaves as per notes in PR.

**3.**
The default selection when choosing a directory is the default directory. **SUCCESS**

**4.**
Fix message box that occurs if you type in a path and filename to export a file to and the directory doesn't exist. **SUCCESS**

#### Remarks.  

***Does it seem a little odd to when selecting folder to default to by first clicking 'OPEN' then subsequently clicking 'CHOOSE' ? Would it make more sense to single click in the list and then click 'CHOOSE'***

Once again good work ctrlalt.

ctrlaltdavid:
@Nex-Pro Selecting a folder by first clicking OPEN then subsequently CHOOSE is explicitly intended so that Vive users can navigate the directory structure without (a) double-clicking the folders in the table, and (b) without a third button. I agree though that it seems odd

PR #8068, opened 2016-06-14
Vive UI Snapshot

howard-stearns:
This is a snapshot of our Vive UI sprint in-progress, and is meant to test some basic assumptions about mixing 3D and 2D interactions:
1. In-world objects can be manipulated with the hand controllers, with actions visualized by laser beams (like existing handControllerGrab).
2. Those same (apparent) laser beams can be used to interact with 2D displays, arrangeable on a HUD around you (like Vive Home).
3. The HUD can be correctly positioned an oriented after both driving (with hand controllers) and physically walking (in room scale).

Please see the following sections on:
 What's Included
 Open Questions
 Known Bugs
 Test Plan


howard-stearns:
# What’s Included

- Manipulate and interact with HUD elements using hand controller trigger. Text fields open the Vive virtual keyboard.
- You stay in the center of the HUD as you move around. **Please make sure “Settings->Avatar->Clear overlays when driving” is checked.**
 - The HUD is hidden while being adjusted (i.e., while you move). 
 - However, any HUD element with a “close” button also has a “pin” button. Pinned elements stay visible when the rest of the HUD is hidden.
- Drive with mouse pads - like release, the center portion of the trackpad is a deadzone. Also walk around in room-scale.
- Hand controller grab (and mouse grab) are as in release, except that any equip point is made visible when you squeeze the trigger a bit. (More will be done with this soon.) This is the only way to interact with 3D objects. (E.g., there are no in-world simulated mouse clicks, for light switch, edit.js, etc.)
- There are buttons to toggle the HUD, HMD-vs-desktop, and the address bar, in addition to the existing buttons for examples and edit. (We may gather these into a “tray”, but that is not in this build.)

This build does not have anything new for hold/equip, nor for teleporting.

howard-stearns:
# Open Questions
### keyboard
- The Vive virtual keyboard can be used with the trackpads or with the active hand laser and trigger. However, the angle of the Vive laser is 45 degrees higher than ours. How much of a problem is that?
- The Vive virtual keyboard floats at a position relative to your head. How is this working for you?

### HUD interactions
- Are you able to click in windows ok? Resize, scroll, etc?
- There is no indication of where the search beam intersects empty HUD “glass”. Are you able to find the intersection with HUD elements ok?
- There is no indication of which hand is driving the HUD elements. (It is the last one triggered.) Are you getting confused as to which hand drives?
- HUD elements always draw over nearer objects, such as hands that are in between your face and a HUD window. This worked well when we wanted to always keep HUD elements visible. Does this still feel right? Now we draw 3D lasers “inside” the HUD, even if we cover up the hands that hold those lasers. (There’s also a current bug involving lasers that pass wholly in front of a HUD element on their way to a 3D entity, see below. But I’ll fix that shortly.)

### In-world interactions
- We currently don’t allow any in-world activity “through” a HUD element (i.e., that you can’t see behind a window or button). Is this too restrictive?
 - One source of difficulty might be relieved by grouping the toggle buttons into a tray that can be moved out of the way en masse.
 - Even then however, it might still be an issue that we stop the laser when you pass a far-grabbed item behind a window.
 - Sometimes in-world near-grabs or far-grabs are intercepted by the 2D HUD, leading to surprising results.  Examples:
  - Placing your hand inside a grabbable block then pressing the trigger with the intent to grab it, can cause a mouse click event to be sent to the Exit HMD mode button causing you to inadvertently switch to desktop mode instead.
  - While far-grabbing an object if your laser passes over a HUD element it can cause the object you are far grabbing to stop moving and slingshot away when the laser rolls off the HUD element.

howard-stearns:
# Known Bugs
- The virtual keyboard only appears when you start SteamVR before starting Interface.
- The keyboard opens every time a window appears if that window has a text field selected -- including those that default to pre-selecting a text field. This behavior happens even when the window “appears” through toggling the HUD. 
- When using the mouse, it sometimes gets stuck at the wrong depth.
- HUD windows sometimes don’t come back after they disappear. 
- The laser is obscured by overlays that are behind them. (We are fine when actually pointing at an overlay, but not when the laser wholly passes in front of the overlay on its way to something in 3D.) I might fix this before we submit to review.

Repro cases welcome!

hifi-gustavo:
Build console output for commit: f5446cc4e44d87808a71925a24d13ed9cccf97ac available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

howard-stearns:
# Test Plan

### Static Hud
- Access the top level menu and navigate the nested structure using my controller
- The HUD can be turned on and off with the click of a hot button
- When overlays are toggled off, the button is still reactive.
- All overlays that have a "close" button also have a "pin" button: a pinned overlay is always be on no matter the overall ‘on/off’ state.
- Mute/Unmute my microphone using my controller
- The user can enter/exit HMD mode by clicking on a button

### Overlays
- Move UI elements on my overlay space using my controller
- Laser beams need to render on top of HUD overlays elements so the laser terminates the overlay.
- Make laser appear on overlays and be able to press buttons etc
- When the focus is on a text field, a virtual keyboard opens, which allows text entry using the hand controller.
- Bring up the address field and add a name to it using my controller. There is a button for this.

### HUD Positioning
- My HUD is always positioned consistently in my view no matter where I move or how I turn when it's 'on'
- When the HUD resets (after walking or driving), the avatar is not reset (e.g., no T pose)
- The HUD reset behavior (both walking and driving) is governed by "Settings->Avatar->Clear overlays when driving"

### Entity Manipulation
- Equippable entities show their equip spot when you partially squeeze the trigger.
- The user can pick up a grabbable object by touching the object and pulling/holding the trigger
- removed in-world ReticleClick behavior from handControllerPointer (e.g., track pad center not used for this any more, and no more editing with hand controller)

### Navigation
- Partitioned track pad. FIXME
- I can fly using my controller

hifi-gustavo:
Build console output for commit: 71bf9384ce60b64bc17b29a0522a113ec6a235be available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

hifi-gustavo:
Build console output for commit: faa2624ca7963bf0f67c1248494dc567fdeb9af9 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

howard-stearns:
This did NOT pick up the script changes from the last two merges!
Maybe it got C++ changes and then decided not to re-generate stuff for the later script-only changes?

howard-stearns:
test this please

howard-stearns:
test this please

hifi-gustavo:
Build console output for commit: faa2624ca7963bf0f67c1248494dc567fdeb9af9 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

hifi-gustavo:
Build console output for commit: 4f6d65a961caf53cb8ac4c2e8545383a5abd3a53 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

howard-stearns:
test this please

hifi-gustavo:
Build console output for commit: 4f6d65a961caf53cb8ac4c2e8545383a5abd3a53 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

hifi-gustavo:
Build console output for commit: 5bc8b53d1b4afb4f1543bc675a53d9ceeca00d58 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

hifi-gustavo:
Build console output for commit: e104429fa1d008662d2f16c1526e7c04bbe21007 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

hifi-gustavo:
Build console output for commit: aeebe8b3fbecc47f339fa6983da5e586ef16fe0f available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

hifi-gustavo:
Build console output for commit: 4aa5e87a2693f1f140f8b7aac7ea15f36eb562b1 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

thoys:
# Testing Results

### Static Hud

*   Access the top level menu and navigate the nested structure using my controller
    - **Menus are all accessable using the vive controllers :+1:**
*   The HUD can be turned on and off with the click of a hot button. This toggling should be consistent with the View->Overlays menu checkstate.
    - **hot button is consistent with View->Overlays button and works :+1:**
*   When overlays are toggled off, the button is still reactive.
    - **When overlays are toggled off all the hotbar buttons are still reactive :+1:**
*   All overlays that have a "close" button also have a "pin" button: a pinned overlay is always be on no matter the overall ‘on/off’ state.
    - **Tested with browser, works! :+1:**
*   Mute/Unmute my microphone using my controller
    - **Works :+1:**
*   The user can enter/exit HMD mode by clicking on a button
    *   Each transition will put you in pause mode (e.g., so that you can actually put on/off the headset)
    *   The Display menu items should reflect the actual state.
    - **Transitions put you in pause mode and display menu items reflect the actual selected display mode :+1:**
*   In desktop mode, wave mouse around while you press escape to go into pause mode. Mouse cursor should continue to be displayed (and continue to wave). Repeat, but first stop moving mouse until it disappears, then press escape (at which point the cursor should appear) and then move the mouse (at which point the already appeared cursor should move).
    - **Works :+1:**
*   Also in desktop mode, with hydra connected and no vive, wave hand controller over an overlay (such as running scripts dialog), so that the mouse cursor appears and follows your hand movement. Then press escape to pause, at which point the cursor should remain visible but frozen. Keeping the hand controller moving, start moving the hardware mouse as well (while still paused). The cursor should track the mouse only, not the hand controller. When you press any button to get out of pause, the cursor should go back to being controlled by the hand controller.
    - **Works :+1:**
*   In Vive HMD, with the hand controller aimed at an overlay, so that it tracks the movement with a laser on the overlay, press escape to pause. (You cannot use escape to pause when keyboard input is going to any overlay.) The laser should disappear from the HMD display, but should appear on the desktop preview. Pressing any button should bring you back to normal. Alternatively, while still paused, you should be able to move the hardware mouse and see it move only in the desktop preview display.
    - **Works :+1:**
*   When the mouse cursor is visible in HMD (e.g., after moving the mouse), it should appear at the correct depth, and of constant size (e.g., not get smaller as it moves away). This should be true even on overlays, and even when you have moved your body in room scale within the HUD. (However, it does not currently update the depth as you move your head, but rather will update after another mouse movement.)
    - **The cursor behaves well if you start HMD mode in first person, although the 3d mouse cursor looks rotated strangely after coming from 3rd person mode into 1st person mode. It looks angled. To fix it use the `Avatar -> Reset Sensors` menu item. :-1:**

### Overlays

*   Move UI elements on my overlay space using my controller
    - **Works :+1:**
*   Laser beams need to render on top of HUD overlays elements so the laser terminates the overlay.
    *   They should appear to terminate at the HUD depth, regardless of how your are leaning with your head.
    *   You should be able to switch the active hand (by partially squeezing the trigger of the hand that does not show the laser), and the laser should transfer without issues.
    - **Works :+1:**
*   Make laser appear on overlays and be able to press buttons etc
    - **Works :+1:**
*   In Vive, partially squeeze a trigger to get the search beam. Aim it a system overlay, stats display, microphone inputs display, or scripted overlay. The laser should show blue for partial trigger, and red for full trigger, and be terminated at the HUD overlay. Let go of the trigger, and the laser should not be shown any more.
    - **Works :+1:**
*   If possible, try the same in desktop with hydra hand controllers. It should be the same, but with a mouse cursor instead of a laser -- i.e., the cursor only shown when squeezing the trigger. Note: there are buggy interactions with the virtual keyboard, which will be dealt with separately.
    - **Works :+1:**
*   When the focus is on a text field, a virtual keyboard opens, which allows text entry using the hand controller.
    - **Partly works, when opening the web-browser it works. When opening the edit.js properties window and selecting a text-input-field it does not work :-1:**
*   Bring up the address field and add a name to it using my controller. There is a button for this.
    - **Works :+1:**
*   You should be unable to interact with either the world or the overlay UI while the OpenVR keyboard is visible
    - **Correct :+1:**
*   If in a script you call HMD.suppressKeyboard() and it returns true, the keyboard should no longer appear regardless of where the keyboard focus is. This will remain true until you call HMD.unsuppressKeyboard()
    - **`HMD.suppressKeyboard()` returns `true`, and prevents the keyboard being shown when I select the web-browser adressbar, `HMD.unsuppressKeyboard()` returns `undefined` and re-enables the functionality.**
*   Multiple calls to HMD.suppressKeyboard() that return true will require a balanced number of calls to HMD.unsuppressKeyboard()
    - **This works, although `HMD.unsuppressKeyboard()` could be called way more times, and it doesn't effect the future `HMD.suppressKeyboard()`, which is expected! :+1:**
*   If the keyboard was already visible when suppressKeyboard was called, it should return false. In all other cases it should return true.
    - **This is true :+1:**
*   Make sure search/grab lasers do not run through 2D HUD elements:
    *   Hold the trigger partially down to get the blue search beam.
    *   Move the beam to intersect with an open QML window. -- In HMD, the beam depth should stop at the HUD/windows when you reach the border of the window. -- In desktop, using Hydra or Oculus, the beam is replaced by mouse cursor on the HUD window when the beam reaches the border of the window. -- (In master, the search beam will be displayed along with the mouse, and the mouse will flicker.)
    *   The same thing should happen with scripted overlay windows such as the examples button or edit button/toolbar.
    *   The same thing should happen with the "avatar inputs" (microphone mute and level), and with the stats display.
    - **This works:+1:**
*   Make sure search/grab lasers do not run through 3D overlay elements. (This is harder to test.)
    *   Open edit, and select an object (e.g., a cube) with the mouse. (It cannot be selected with the hand controller.)
    *   Manipulate your hand position and direction so as to have the search beam intersect the 3d overlays. The trick here is to see if the blue ball at the end of the laser terminates at the 3D overlays. (What I did was position myself with the selected cube nearer to me, with a wall a good distance behind it. As I point the search beam in the general direction of the cube, the beam and the blue ball terminates at the wall. When I manipulate the controller to have the beam pass though the up/down arrow of the selected object, the blue ball and the beam terminus "jumps" forward to the up/down arrow, instead of being back along the wall. (In master, the blue ball and beam terminus stays along the wall.)
    - **This works:+1:**

### HUD Positioning

*   My HUD is always positioned consistently in my view no matter where I move or how I turn when it's 'on'
    - **This works (even in 3rd person) :+1:**
*   When the HUD resets (after walking or driving), the avatar is not reset (e.g., no T pose)
    - **Correct :+1:**
*   The HUD reset behavior (both walking and driving) is governed by "Settings->Avatar->Clear overlays when driving"
    - **Correct :+1:**
*   The HUD behavior (fade out on drive, and back when stopped) should work in desktop as well as vive (when enabled by preference)
    - **Correct :+1:**
*   Open some UI dialogs on the HMD overlay, for example: press control-b to bring up a web browser.
    *   While connected to a domain, press the arrow keys or the track pad on the right vive controller to move.
    *   when you press these keys the overlay should smoothly fade away.
    *   when you stop pressing these keys it should smoothly fade in, but centered around your current facing.
    *   additionally, when the overlay is visible, if you lean your head in toward the surface of the overlay, it will fade out and fade back in recentered about your head.
    *   As you walk around in room scale, the overlays should not come back on until a moment after you stop. In particular, they shouldn't go off and on repeatedly as you continuously walk 
    around.
    *   With the setting on, you should not be able to get your head outside the HUD sphere.
    *   Do Avatar->Reset Sensors, particularly while in Vive. The aim here is to not lose any overlays or buttons as a result of the reset.
    - **fading and unfading work perfectly, no windows are lost on Reset Sensors :+1:**

### Entity Manipulation

*   Equippable entities show their equip spot when you partially squeeze the trigger.
    *   Check the ping pong gun to ensure that a green ball appears around the guns handle.
    *   Check the blocks in demo to ensure that a green ball does not appear around each block.
    - **Both work fine, green ball only seems to appear on objects with defined grabbing position. :+1:**
*   The user can pick up a grabbable object by touching the object and pulling/holding the trigger
    - **Works, but not while grabbing the green search incidator, non-release-blocker see bottom of test report. :+1:**
*   removed in-world ReticleClick behavior from handControllerPointer (e.g., track pad center not used for this any more, and no more editing with hand controller)
    - **Correct :+1:**
*   Place your hand over a colored block on the table, press the trigger to near-grab the block. You should be able to pick it up and move it, while the trigger is pressed, it should drop when you release the trigger.
    - **Works :+1:**
*   Stand away from the table, press the trigger the far-grab laser should appear, when the far-grab laser lands on a colored block you should enter far-grab mode. Releasing the trigger should drop the object.
    - **Assuming the table is in `demo` (or `dev-demo`), this works :+1:**
*   Test the near-trigger behavior by placing your hand in the top marquee of the defender machine and pressing the trigger. This should reset all the blocks on the table.
    - **Works :+1:**
*   Test the near-hold behavior by moving your hand over the ping-pong gun and pressing the GRIP button. While the grip button is depressed, you should be able to shoot the gun with the trigger button. Releasing the GRIP button should drop the gun.
    - **Works :+1:**
*   Far grab an entity, and while still holding it, move the hand controller such that the beam passes through a HUD overlay. The far grab should continue as if the overlay wasn't there. However, you should not be able to initate a far grab through an overlay. (Triggering through an overlay when you are not already holding an entity should be treated as a click on the overlay.)
    - **Works :+1:**
*   Try particularly to pass the hold-laser through a button. The button should not fire when you're already holding something with that laser.
    - **Works :+1:**
*   Repeat for near grab and for equip. (Currently in the vive-ui branch, equip is involves a simultaneous trigger and press of the "grip" button on the underside of the vive controller.)
    - **Works :+1:**
*   Make sure you can far grab with one hand, and while still far grabbing, far grab with the other hand, too. The let go of the first. Repeat with near grab and with equip.
    - **Works :+1:**

### Navigation

*   Exercise trackpad navigation (left hand: strafe left/right, forward/back; right hand: turn left/right (with and without snap turns), and jump/fly (and ascend/descend while flying). These should work with thumb away from the center of the trackpad. The motion should not be mixed between "x" activities and "y" activities. (For example, jumping with snap turns on shouldn't turn while jumping.)
    - **Works :+1:**
*   Exercise presses using the center of the trackpad. These should make the avatar move.
    - **Works :+1:**
*   I can fly using my controller
    - **Works :+1:**

### Other

*   Go into vive hmd, and make sure hand controllers work. Then go back to desktop, and confirm the following (in desktop):
    *   Avatar should be "normal" (e.g., when viewed in third person), and not have its hand's pulled in strange ways. Waving the hand controllers around should have no effect.
        - **Works :+1:**
    *   The mouse should be useable and not pinned to the bottom of the screen (or pinned anywhere else). When the mouse times out after five seconds, and you bring the mouse back by moving, the cursor should be where you left it.
        - **Works :+1:**
    *   In Edit->Console..., Controller.getPoseValue(Controller.Standard.RightHand).valid should be false, and similarly for LeftHand.
        - **Works :+1:**
*   Please check overall performance vs master, in desktop and vive, and also compare pointing at an overlay vs not (in vive and if practical, desktop w/hydra).
    - ** Performance didn't change, framerate still at 90FPS in playa. The UI is some getting used to, support for the keyboard is great, and how it responds is intuitive :+1:**
*   If you happen to have a vive controller that generates trigger spam (@thoys!), the Vive controllers should no longer intermittently or permanently report non-zero values even when the trigger is not being pressed at all.
    - **Yes, It's me!, Works :+1:**
*   When using the debug flag to use OpenVR over Oculus (with Oculus hardware), the screen should no longer sit at an Oculus 'waiting' screen forever.
    - **Couldn't test, no free HDMI port for my Oculus CV1/ or DP-cable for the Vive.**


## Noticed bugs/expectations
* [Bug] The 3d mouse cursor looks rotated strangely after coming from 3rd person mode into 1st person mode. It looks angled. To fix it use the `Avatar -> Reset Sensors` menu item. 
* [Feature request] The Mute button on the hmd-hot-toolbar should display the current mute state of the audio
* [Bug] The hmd-toolbar is movable out of the viewable screen. This should either be restricted or a reset for the toolbar(s) should be present.
* [Paradox] The hand of the avatar is shown behind the hud, which makes it feel kinda unconfortable to use, is there a way to overcome this without having the hud interfere with surrounding 3d objects?
* [Minor-Bug] Green grab balls indicator should also trigger a grab event when the trigger is squeezed on them, or alternatively the grab/search beam should go straight through them.
* [Bug] Rendered hud cursor looks weird after switching the Interface window to another window sometimes while in HMD mode, probably related to the first bug that I reported. No solid repro found yet! (Also fixes when using `Avatar -> Reset Sensors` )
* [Bug] While toggling between Interface and my text editor to write this report, I lost the laserpointer in my left hand. I was unable to grip anything with it anymore. No solid repro yet. (Restarting `handControllerGrab.js` fixed this.)
* [Bug] Camera goes in front of the avatar in 3rd Person view. Expected is that the camera only goes behind the avatar. No solid repro.

QA Fail :-1: (Please see list of bugs above, may not all apply to this PR, good work sofar vive-ui team!)

howard-stearns:
Exellent, @thoys. Here's an echo of what I'm hearing, combined with where we are on them:
### Must fix:
- Toolbar tray positioning, both initially and when switching HMD/desktop.

### Under development:
- Equip green ball behavior. (See "Noticed bugs/expectations", above.)
- Other changes to handControllerGrab.
  => @thoys, when you lost your laser pointer that was fixed by reloading handControllerGrab.js, was there any "exception" or "backtrace" in the log?

### Should fix:
- Keyboard does not come up for:
  Entity Properties text fields.
  Browser text input boxes.

### Things that are on our "polishing" list after merge, as time permits:
- Cursor looks weird transitioning from 3rd person to 1st.
 => Not sure if I'm seeing the same thing. One thing we've discussed in making the cursor "flat". Is that the issue?
- Button color to reflect actual state. (Mute, HMD, Goto, Examples)

### Other:
- Camera may be in front of avatar in third person.
 => Need repro, but possibly when you are in room-scale at the back of the HUD sphere?
- Overlays rendered over nearer objects.
 => This is by design, so that the overlays are never obscured by hands, etc. Laser beams being the one exception.
    Whether this is too weird is one of our open questions. We want continued feedback on this. For now, the consensus is that this is ok.

jherico:
@thoys 

> Couldn't test, no free HDMI port for my Oculus CV1/ or DP-cable for the Vive.

This was a situation that occurred if you used the OpenVR plugin with only the Oculus connected.

thoys:
@howard-stearns I think the laser pointer got lost while Equipping the Bow script, I saw another PR with a fix for that.

hifi-gustavo:
Build console output for commit: cf57e71f1a4a02b0b9b3daeac703c149f8e4676f available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

hifi-gustavo:
Build console output for commit: 4f71f621f734a0905cad357e91702f6ac345d67b available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

hifi-gustavo:
Build console output for commit: cd09c61f31b9707dc151e6393162193ca23e3058 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

hifi-gustavo:
Build console output for commit: c928de6051e57ed2191b670aa0e9fcbcb8fe4b71 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

hifi-gustavo:
Build console output for commit: 7e88772e855596a0275fca50f1d2ed86f443010e available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

hifi-gustavo:
Build console output for commit: c600b65f457257da803bd94489d13eeee9dc550f available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

hifi-gustavo:
Build console output for commit: 223f9bda2e6dcf92487ab4dac4cd3b44a762eef7 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

hifi-gustavo:
Build console output for commit: a1844d8c19ab2d28ffc2870829bfc424c57aadb0 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

thoys:
@howard-stearns , should I fully re-test it?

howard-stearns:
@thoys I think the priorities are 1) Your top issues that were causing a fail (keeping in mind that some things will be dealt with later, as discussed), 2) The new things added since, which I enumerated at the end of the original test plan rather than editing it, and 3) everything.

We're also going to have @SamGondelman go over everything today, too.

howard-stearns:
By the way, there are new changes to master that are included here, which were new to us. So it's possible that there was some conflict that has broken things.

RyanDowne:
Some things that might come up but were not part of scope:

1) Anything to do with using edit.js. There are lots of issues here and we explicitly did not resolve them

2) Trying to activate the keyboard on a web-page. Trying to get the keyboard up in the examples/search won't work. If you open a web-page (ctrl-b) and try to enter text in something like google.com's search field, the keyboard doesn't come up. 

3) In-world web-entities

ZappoMan:
Note: I tested on the mac with DK2 this morning... and I kept getting the twisted mouse pointer. It seemed to get twisted sometimes on reset... so if I hit the `'` key or if I moved and then came to a rest... it would be twisted sometimes, and not twisted other times.

hifi-gustavo:
Build console output for commit: af0335496eb516d58f3eaf544cafdbf9c98d2095 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

interface/src/Application.cpp
```diff
@@ -1419,7 +1421,7 @@ void Application::initializeUi() {
 
     rootContext->setContextProperty("Overlays", &_overlays);
     rootContext->setContextProperty("Window", DependencyManager::get<WindowScriptingInterface>().data());
-    rootContext->setContextProperty("Menu", MenuScriptingInterface::getInstance());
+    rootContext->setContextProperty("MenuInterface", MenuScriptingInterface::getInstance());
```
ZappoMan: er -- what's up with this? why change this?

interface/src/scripting/HMDScriptingInterface.h
```diff
@@ -12,6 +12,8 @@
 #ifndef hifi_HMDScriptingInterface_h
 #define hifi_HMDScriptingInterface_h
 
+#include <atomic>
```
ZappoMan: ??

interface/src/scripting/ToolbarScriptingInterface.h
```diff
@@ -0,0 +1,26 @@
...
+#include <QtCore/QObject>
+
+#include <DependencyManager.h>
+
+class ToolbarProxy;
+
+class ToolbarScriptingInterface : public QObject, public Dependency {
+    Q_OBJECT
+public:
+	Q_INVOKABLE QObject* getToolbar(const QString& toolbarId);
```
ZappoMan: indent wrong

hifi-gustavo:
Build console output for commit: dc341810327c0fdcd92d3861de72833ab9295a20 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
```diff
@@ -357,22 +383,116 @@ void HmdDisplayPlugin::setEyeRenderPose(uint32_t frameIndex, Eye eye, const glm:
...
+            _handLasers[1] = info;
+        }
+    });
+    // FIXME defer to a child class plugin to determine if hand lasers are actually 
+    // available based on the presence or absence of hand controllers
+    return true;
+}
+
+void HmdDisplayPlugin::compositeExtra() {
+    std::array<HandLaserInfo, 2> handLasers;
```
ZappoMan: magic numbers: `NUM_LASERS`?

SamGondelman:
Testing results:

### Static Hud

- Access the top level menu and navigate the nested structure using my controller
    -  :heavy_check_mark:, although some menu items have undesirable behavior.  For example, if you switch to third person, the UI follows the camera (which is good), but you can't navigate the menu with your hands to switch back to first person (you can use the trackpad, which works alright)
- The HUD can be turned on and off with the click of a hot button. This toggling should be consistent with the View->Overlays menu checkstate.
    -  :heavy_check_mark: 
- When overlays are toggled off, the button is still reactive.
    -  :heavy_check_mark: 
- All overlays that have a "close" button also have a "pin" button: a pinned overlay is always be on no matter the overall ‘on/off’ state.
    -  :heavy_check_mark:.  Pinned state is not remembered between sessions, though, which might be desirable. 
- Mute/Unmute my microphone using my controller
    -  :heavy_check_mark: 
- The user can enter/exit HMD mode by clicking on a button
    -  :heavy_check_mark: 
 - Each transition will put you in pause mode (e.g., so that you can actually put on/off the headset)
     -  :heavy_check_mark: 
 - The Display menu items should reflect the actual state.
     -  :heavy_check_mark: 
- In desktop mode, wave mouse around while you press escape to go into pause mode. Mouse cursor should continue to be displayed (and continue to wave). Repeat, but first stop moving mouse until it disappears, then press escape (at which point the cursor should appear) and then move the mouse (at which point the already appeared cursor should move).
    -  :heavy_check_mark:, but if you press escape while in HMD mode and switch back to desktop mode using the menu, the OS mouse isn't hidden, so you can see both that pointer and our pointer  
- Also in desktop mode, with hydra connected and no vive, wave hand controller over an overlay (such as running scripts dialog), so that the mouse cursor appears and follows your hand movement. Then press escape to pause, at which point the cursor should remain visible but frozen. Keeping the hand controller moving, start moving the hardware mouse as well (while still paused). The cursor should track the mouse only, not the hand controller. When you press any button to get out of pause, the cursor should go back to being controlled by the hand controller.
    - ~ :heavy_check_mark:, this mostly works, except that when using the hydras, the mouse cursor was hidden unless I lightly held the trigger 
- In Vive HMD, with the hand controller aimed at an overlay, so that it tracks the movement with a laser on the overlay, press escape to pause. (You cannot use escape to pause when keyboard input is going to any overlay.) The laser should disappear from the HMD display, but should appear on the desktop preview. Pressing any button should bring you back to normal. Alternatively, while still paused, you should be able to move the hardware mouse and see it move only in the desktop preview display.
    -  :heavy_check_mark: 
- When the mouse cursor is visible in HMD (e.g., after moving the mouse), it should appear at the correct depth, and of constant size (e.g., not get smaller as it moves away). This should be true even on overlays, and even when you have moved your body in room scale within the HUD. (However, it does not currently update the depth as you move your head, but rather will update after another mouse movement.)
    -  :heavy_check_mark: 

### Overlays

- Move UI elements on my overlay space using my controller
    -  :heavy_check_mark: 
- Laser beams need to render on top of HUD overlays elements so the laser terminates the overlay.
    -  :heavy_check_mark: 
 - They should appear to terminate at the HUD depth, regardless of how your are leaning with your head.
    -  :heavy_check_mark: 
 - You should be able to switch the active hand (by partially squeezing the trigger of the hand that does not show the laser), and the laser should transfer without issues.
    -  :heavy_check_mark: 
- Make laser appear on overlays and be able to press buttons etc
    -  :heavy_check_mark: 
- In Vive, partially squeeze a trigger to get the search beam. Aim it a system overlay, stats display, microphone inputs display, or scripted overlay. The laser should show blue for partial trigger, and red for full trigger, and be terminated at the HUD overlay. Let go of the trigger, and the laser should not be shown any more.
    -  :heavy_check_mark: 
- If possible, try the same in desktop with hydra hand controllers. It should be the same, but with a mouse cursor instead of a laser -- i.e., the cursor only shown when squeezing the trigger. Note: there are buggy interactions with the virtual keyboard, which will be dealt with separately.
    -  :heavy_check_mark: 
- When the focus is on a text field, a virtual keyboard opens, which allows text entry using the hand controller.
    -  :x:.  This works when you first click on a text field, but when you press "Done" on the keyboard, the text field keeps focus, so you can't reclick on it to bring the keyboard up again.  Clicking "Done" should probably take focus off the text field. 
- Bring up the address field and add a name to it using my controller. There is a button for this.
    -  :x:, mostly works, but if you bring up the address bar and press the trigger while pointing at nothing, the keyboard disappears, but the address bar keeps focus, so you can't click on it again to bring the keyboard up unless you first click on a different overlay. 
- You should be unable to interact with either the world or the overlay UI while the OpenVR keyboard is visible
    -  :heavy_check_mark: 
- If in a script you call HMD.suppressKeyboard() and it returns true, the keyboard should no longer appear regardless of where the keyboard focus is. This will remain true until you call HMD.unsuppressKeyboard()
    -  :heavy_check_mark: 
- Multiple calls to HMD.suppressKeyboard() that return true will require a balanced number of calls to HMD.unsuppressKeyboard()
    -  :heavy_check_mark: 
- If the keyboard was already visible when suppressKeyboard was called, it should return false. In all other cases it should return true.
    -  :heavy_check_mark: 
- Make sure search/grab lasers do not run through 2D HUD elements:
 - Hold the trigger partially down to get the blue search beam.
 - Move the beam to intersect with an open QML window. 
  -- In HMD, the beam depth should stop at the HUD/windows when you reach the border of the window. 
  -- In desktop, using Hydra or Oculus, the beam is replaced by mouse cursor on the HUD window when the beam reaches the border of the window. 
  -- (In master, the search beam will be displayed along with the mouse, and the mouse will flicker.)
 - The same thing should happen with scripted overlay windows such as the examples button or edit button/toolbar.
 - The same thing should happen with the "avatar inputs" (microphone mute and level), and with the stats display.
    -  all :heavy_check_mark: 
- Make sure search/grab lasers do not run through 3D overlay elements. (This is harder to test.)
 - Open edit, and select an object (e.g., a cube) with the mouse. (It cannot be selected with the hand controller.)
 - Manipulate your hand position and direction so as to have the search beam intersect the 3d overlays. The trick here is to see if the blue ball at the end of the laser terminates at the 3D overlays. (What I did was position myself with the selected cube nearer to me, with a wall a good distance behind it. As I point the search beam in the general direction of the cube, the beam and the blue ball terminates at the wall. When I manipulate the controller to have the beam pass though the up/down arrow of the selected object, the blue ball and the beam terminus "jumps" forward to the up/down arrow, instead of being back along the wall. (In master, the blue ball and beam terminus stays along the wall.)
    -  :x:, looks like the ball still passes through the 3D overlays 


### HUD Positioning

- My HUD is always positioned consistently in my view no matter where I move or how I turn when it's 'on'
    -  :heavy_check_mark: 
- When the HUD resets (after walking or driving), the avatar is not reset (e.g., no T pose)
    -  :heavy_check_mark: 
- The HUD reset behavior (both walking and driving) is governed by "Settings->Avatar->Clear overlays when driving"
    -  mostly :heavy_check_mark:, but when off you can still move outside of the sphere and have the HUD disappear 
- The HUD behavior (fade out on drive, and back when stopped) should work in desktop as well as vive (when enabled by preference)
    -  :heavy_check_mark: 
- Open some UI dialogs on the HMD overlay, for example: press control-b to bring up a web browser.
 - While connected to a domain, press the arrow keys or the track pad on the right vive controller to move.
 - when you press these keys the overlay should smoothly fade away.
    -  :heavy_check_mark: 
 - when you stop pressing these keys it should smoothly fade in, but centered around your current facing.
    -  :heavy_check_mark: 
 - additionally, when the overlay is visible, if you lean your head in toward the surface of the overlay, it will fade out and fade back in recentered about your head.
    -  :heavy_check_mark: 
 - As you walk around in room scale, the overlays should not come back on until a moment after you stop. In particular, they shouldn't go off and on repeatedly as you continuously walk around.
    -  :heavy_check_mark: 
 - With the setting on, you should not be able to get your head outside the HUD sphere.
    -  :x: 
 - Do Avatar->Reset Sensors, particularly while in Vive. The aim here is to not lose any overlays or buttons as a result of the reset.
    -  :x:, one time I reset while the HUD was active and the HUD button stayed white but the buttons next to it disappeared.  I had to toggle the HUD to get them back.  I couldn't repro it. 

### Entity Manipulation

- Equippable entities show their equip spot when you partially squeeze the trigger.
 - Check the ping pong gun to ensure that a green ball appears around the guns handle.
 - Check the blocks in demo to ensure that a green ball does not appear around each block.
    -  :heavy_check_mark: 
- The user can pick up a grabbable object by touching the object and pulling/holding the trigger
    -  :heavy_check_mark: 
- removed in-world ReticleClick behavior from handControllerPointer (e.g., track pad center not used for this any more, and no more editing with hand controller)
    -  :heavy_check_mark: 
- Place your hand over a colored block on the table, press the trigger to near-grab the block. You should be able to pick it up and move it, while the trigger is pressed, it should drop when you release the trigger.
    -  :heavy_check_mark: 
- Stand away from the table, press the trigger the far-grab laser should appear, when the far-grab laser lands on a colored block you should enter far-grab mode. Releasing the trigger should drop the object.
    -  :heavy_check_mark: 
- Test the near-trigger behavior by placing your hand in the top marquee of the defender machine and pressing the trigger. This should reset all the blocks on the table.
    -  :heavy_check_mark: 
- Test the near-hold behavior by moving your hand over the ping-pong gun and pressing the GRIP button. While the grip button is depressed, you should be able to shoot the gun with the trigger button. Releasing the GRIP button should drop the gun.
    -  this doesn't use the grip button anymore, but the behavior is correct :heavy_check_mark: 
- Far grab an entity, and while still holding it, move the hand controller such that the beam passes through a HUD overlay. The far grab should continue as if the overlay wasn't there. However, you should not be able to initate a far grab through an overlay. (Triggering through an overlay when you are not already holding an entity should be treated as a click on the overlay.)
    -  :heavy_check_mark: for far grab passing over overlay, but :x: because it seems like far grab overrides non-QML overlays (can far grab while looking through them).  Behavior of things like right-click menu in a domain like demo seems generally broken 
- Try particularly to pass the hold-laser through a button. The button should not fire when you're already holding something with that laser.
    -  :heavy_check_mark: 
- Repeat for near grab and for equip. (Currently in the vive-ui branch, equip is involves a simultaneous trigger and press of the "grip" button on the underside of the vive controller.)
    -  this is no longer true 
- Make sure you can far grab with one hand, and while still far grabbing, far grab with the other hand, too. The let go of the first. Repeat with near grab and with equip.
    -  :heavy_check_mark: 

### Navigation

- Exercise trackpad navigation (left hand: strafe left/right, forward/back; right hand: turn left/right (with and without snap turns), and jump/fly (and ascend/descend while flying). These should work with thumb away from the center of the trackpad. The motion should not be mixed between "x" activities and "y" activities. (For example, jumping with snap turns on shouldn't turn while jumping.)
    -  :heavy_check_mark: 
- Exercise presses using the center of the trackpad. These should make the avatar move.
    -  :heavy_check_mark: 
- I can fly using my controller
    -  :heavy_check_mark: 

### Other
- Go into vive hmd, and make sure hand controllers work. Then go back to desktop, and confirm the following (in desktop):
 - Avatar should be "normal" (e.g., when viewed in third person), and not have its hand's pulled in strange ways. Waving the hand controllers around should have no effect.
    -  :heavy_check_mark: 
 - The mouse should be useable and not pinned to the bottom of the screen (or pinned anywhere else). When the mouse times out after five seconds, and you bring the mouse back by moving, the cursor should be where you left it.
    -  :heavy_check_mark: 
 - In Edit->Console..., Controller.getPoseValue(Controller.Standard.RightHand).valid should be false, and similarly for LeftHand.
    -  :heavy_check_mark: 
- Please check overall performance vs master, in desktop and vive, and also compare pointing at an overlay vs not (in vive and if practical, desktop w/hydra).
    -  no noticeable difference :heavy_check_mark: 
- If you happen to have a vive controller that generates trigger spam (@thoys!), the Vive controllers should no longer intermittently or permanently report non-zero values even when the trigger is not being pressed at all.
    -  N/A 
- When using the debug flag to use OpenVR over Oculus (with Oculus hardware), the screen should no longer sit at an Oculus 'waiting' screen forever.
    -  I don't know how to use the flag so I'll come back and test this later 

The following PRs are being added to this branch late, and each have their own test plans that should also be tested here. (They were not present when this PR was originally tested.):
https://github.com/highfidelity/hifi/pull/8099
    -  Mentioned earlier, this mostly works 
https://github.com/highfidelity/hifi/pull/8100
    -  mostly :heavy_check_mark:, if you pass through sideways you can still see the back with one eye :P 
https://github.com/highfidelity/hifi/pull/8102
    -  :x:, after playing around in demo, I got into a state where I could near grab equippable objects but they wouldn't ever be equipped.  Center of trackpad -> drop happened a lot to me accidentally while I was trying to move.  Can't far grab equipable objects through their green spheres.  I couldn't get the "shake" gesture to work. 
https://github.com/highfidelity/hifi/pull/8106
    -  :heavy_check_mark: 
https://github.com/highfidelity/hifi/pull/8108
    -  I still lose the HUD button all the time when restarting in HMD mode, but I mentioned that in the comments for this PR so :heavy_check_mark: 
https://github.com/highfidelity/hifi/pull/8112
    -  :heavy_check_mark: 
https://github.com/highfidelity/hifi/pull/8114
    -  :heavy_check_mark:, except for some issues mentioned earlier 
https://github.com/highfidelity/hifi/pull/8120
    -  need to redownload new version to test, will update soon 

### General comments:
- The ever-present, randomly floating HUD button is a little strange.  I feel like a button that is somehow connected to the Avatar would make a lot more sense (like a watch on your wrist or something)
- It's pretty easy to lose overlays outside of the HUD-able area while dragging them
- Pressing escape twice while in HMD mode causes the view to jump around
- I don't know the exact cause but afer doing a lot of stuff I went back to desktop mode and went to a different domain, and the progress bar was stuck in the top left corner.  I also lost the edit bar off to the right at some point.
- The keyboard pops up frequently when a new window is opened because a text field suddenly gets focus.
- I'm not sure I like the whole "HUD sphere" thing.  I'd prefer a flat surface in front of you with the buttons on it so you can see them all without wildly turning your head to search for them.

Overall, this is mostly functional, so I wouldn't necessarily say it's a fail right now.

libraries/gl/src/gl/OglplusHelpers.cpp
```diff
@@ -359,6 +361,94 @@ ShapeWrapperPtr loadSphereSection(ProgramPtr program, float fov, float aspect, i
...
+
+            /// Returns the winding direction of faces
+            FaceOrientation FaceWinding(void) const {
+                return FaceOrientation::CCW;
+            }
+
+            /// Queries the bounding sphere coordinates and dimensions
+            template <typename T>
+            void BoundingSphere(Sphere<T>& bounding_sphere) const {
+                bounding_sphere = Sphere<T>(0, 0, -0.5, 0.5);
```
ZappoMan: magic numbers -- what are these 0.5??

plugins/openvr/src/OpenVrHelpers.cpp
```diff
@@ -84,6 +92,145 @@ void releaseOpenVrSystem() {
...
+
+    
+    auto menu = container->getPrimaryMenu();
+    auto action = menu->getActionForOption(MenuOption::Overlays);
+
+    // When the overlays are revealed, suppress the keyboard from appearing on text focus for a tenth of a second. 
+    _overlayMenuConnection = QObject::connect(action, &QAction::triggered, [action] {
+        if (action->isChecked()) {
+            _overlayRevealed = true;
+            QTimer::singleShot(100, [&] { _overlayRevealed = false; });
```
ZappoMan: magic number

scripts/system/controllers/handControllerGrab.js
```diff
@@ -820,275 +858,400 @@ function MyController(hand) {
...
-        }
-        if (this.state == STATE_HOLD_SEARCHING && this.secondaryReleased()) {
-            this.setState(STATE_RELEASE);
-            return;
+    this.searchEnter = function() {
+        this.equipHotspotOverlays = [];
+
+        // find entities near the avatar that might be equipable.
+        var entities = Entities.findEntities(MyAvatar.position, HOTSPOT_DRAW_DISTANCE);
+        var i, l = entities.length;
```
ZappoMan: coding standard says variable names should be descriptive except for iterators, so in this case `l` should be `length` or similar.

howard-stearns:
Here's my summary of Sam's QA. I'll edit/add to this with more info from others:
### Known issues to try to hunt down (future work):
- Had instance of overlay not returning with the rest of HUD, that was recovered by toggling HUD.
- Had instance(s) of flinging overlays off the screen.

### Recommended improvements:
- After dismissing keyboard, must click on something else before clicking back on text field to get keyboard. Maybe dismissing keyboard should unfocus for you?
- Some individual QML windows should maybe not start with text field focus.
- Noted strangeness of HUD toggle button reset.
- Need some indication of lost focus when in HMD - clicking out of Interface or popping up an old-style QT system window.
- Prefers flatter HUD. (We have an action item/recommendation to explore having the user positioned closer to the front rather than the center, which would also have a flatter effect.)

### Regression, but arguable as to what is correct:
- New handControllerGrab act through 3D overlays.

### Test plan error:
- Wording wrong in test plan: can get your head outside of HUD sphere, but you can't see back of unpinned overlays.

### Other:
- Sam crashed. Will track down/examine report.
- Reset when coming out of escape can be disconcerting, e.g., if your headset calibration puts you on the floor. (This is present in master.)
- Report of far-grabbing through a non-QML overlay, but we could not reproduce. Might be related to varying update frequencies.
- Stuck progress bar on changing domain. Presumably not related to this PR.

thoys:
# Testing Results

### Static Hud

*   Access the top level menu and navigate the nested structure using my controller
    - :heavy_check_mark: 
*   The HUD can be turned on and off with the click of a hot button. This toggling should be consistent with the View->Overlays menu checkstate.
    - :heavy_check_mark: 
*   When overlays are toggled off, the button is still reactive.
    - :heavy_check_mark: 
*   All overlays that have a "close" button also have a "pin" button: a pinned overlay is always be on no matter the overall ‘on/off’ state.
    - :heavy_check_mark: 
*   Mute/Unmute my microphone using my controller
    - :heavy_check_mark: 
*   The user can enter/exit HMD mode by clicking on a button
    *   Each transition will put you in pause mode (e.g., so that you can actually put on/off the headset)
    *   The Display menu items should reflect the actual state.
    - :heavy_check_mark: 
*   In desktop mode, wave mouse around while you press escape to go into pause mode. Mouse cursor should continue to be displayed (and continue to wave). Repeat, but first stop moving mouse until it disappears, then press escape (at which point the cursor should appear) and then move the mouse (at which point the already appeared cursor should move).
    - **:heavy_check_mark:  Works with the rendered cursor, but the cursor from the operating system is still visible after switching back from HMD to Desktop mode**
*   Also in desktop mode, with hydra connected and no vive, wave hand controller over an overlay (such as running scripts dialog), so that the mouse cursor appears and follows your hand movement. Then press escape to pause, at which point the cursor should remain visible but frozen. Keeping the hand controller moving, start moving the hardware mouse as well (while still paused). The cursor should track the mouse only, not the hand controller. When you press any button to get out of pause, the cursor should go back to being controlled by the hand controller.
    - :heavy_check_mark:
*   In Vive HMD, with the hand controller aimed at an overlay, so that it tracks the movement with a laser on the overlay, press escape to pause. (You cannot use escape to pause when keyboard input is going to any overlay.) The laser should disappear from the HMD display, but should appear on the desktop preview. Pressing any button should bring you back to normal. Alternatively, while still paused, you should be able to move the hardware mouse and see it move only in the desktop preview display.
    - :heavy_check_mark:
*   When the mouse cursor is visible in HMD (e.g., after moving the mouse), it should appear at the correct depth, and of constant size (e.g., not get smaller as it moves away). This should be true even on overlays, and even when you have moved your body in room scale within the HUD. (However, it does not currently update the depth as you move your head, but rather will update after another mouse movement.)
    - **:x: Does not always work, for example, try bringing up the right-click-menu.**

### Overlays

*   Move UI elements on my overlay space using my controller
    - :heavy_check_mark:
*   Laser beams need to render on top of HUD overlays elements so the laser terminates the overlay.
    *   They should appear to terminate at the HUD depth, regardless of how your are leaning with your head.
    *   You should be able to switch the active hand (by partially squeezing the trigger of the hand that does not show the laser), and the laser should transfer without issues.
    - :heavy_check_mark:
*   Make laser appear on overlays and be able to press buttons etc
    - :heavy_check_mark:
*   In Vive, partially squeeze a trigger to get the search beam. Aim it a system overlay, stats display, microphone inputs display, or scripted overlay. The laser should show blue for partial trigger, and red for full trigger, and be terminated at the HUD overlay. Let go of the trigger, and the laser should not be shown any more.
    - :heavy_check_mark:
*   If possible, try the same in desktop with hydra hand controllers. It should be the same, but with a mouse cursor instead of a laser -- i.e., the cursor only shown when squeezing the trigger. Note: there are buggy interactions with the virtual keyboard, which will be dealt with separately.
    - :heavy_check_mark:
*   When the focus is on a text field, a virtual keyboard opens, which allows text entry using the hand controller.
    - **Still partly works, when opening the web-browser it works. When opening the edit.js properties window and selecting a text-input-field it does not work :x:**
*   Bring up the address field and add a name to it using my controller. There is a button for this.
    - :heavy_check_mark:
*   You should be unable to interact with either the world or the overlay UI while the OpenVR keyboard is visible
    - :heavy_check_mark:
*   If in a script you call HMD.suppressKeyboard() and it returns true, the keyboard should no longer appear regardless of where the keyboard focus is. This will remain true until you call HMD.unsuppressKeyboard()
    - :heavy_check_mark:
*   Multiple calls to HMD.suppressKeyboard() that return true will require a balanced number of calls to HMD.unsuppressKeyboard()
    - :heavy_check_mark:
*   If the keyboard was already visible when suppressKeyboard was called, it should return false. In all other cases it should return true.
    - :heavy_check_mark:
*   Make sure search/grab lasers do not run through 2D HUD elements:
    *   Hold the trigger partially down to get the blue search beam.
    *   Move the beam to intersect with an open QML window. -- In HMD, the beam depth should stop at the HUD/windows when you reach the border of the window. -- In desktop, using Hydra or Oculus, the beam is replaced by mouse cursor on the HUD window when the beam reaches the border of the window. -- (In master, the search beam will be displayed along with the mouse, and the mouse will flicker.)
    *   The same thing should happen with scripted overlay windows such as the examples button or edit button/toolbar.
    *   The same thing should happen with the "avatar inputs" (microphone mute and level), and with the stats display.
    - :heavy_check_mark:
*   Make sure search/grab lasers do not run through 3D overlay elements. (This is harder to test.)
    *   Open edit, and select an object (e.g., a cube) with the mouse. (It cannot be selected with the hand controller.)
    *   Manipulate your hand position and direction so as to have the search beam intersect the 3d overlays. The trick here is to see if the blue ball at the end of the laser terminates at the 3D overlays. (What I did was position myself with the selected cube nearer to me, with a wall a good distance behind it. As I point the search beam in the general direction of the cube, the beam and the blue ball terminates at the wall. When I manipulate the controller to have the beam pass though the up/down arrow of the selected object, the blue ball and the beam terminus "jumps" forward to the up/down arrow, instead of being back along the wall. (In master, the blue ball and beam terminus stays along the wall.)
    - :heavy_check_mark:

### HUD Positioning

*   My HUD is always positioned consistently in my view no matter where I move or how I turn when it's 'on'
    - :heavy_check_mark:
*   When the HUD resets (after walking or driving), the avatar is not reset (e.g., no T pose)
    - :heavy_check_mark:
*   The HUD reset behavior (both walking and driving) is governed by "Settings->Avatar->Clear overlays when driving"
    - :heavy_check_mark:
*   The HUD behavior (fade out on drive, and back when stopped) should work in desktop as well as vive (when enabled by preference)
    - :heavy_check_mark:
*   Open some UI dialogs on the HMD overlay, for example: press control-b to bring up a web browser.
    *   While connected to a domain, press the arrow keys or the track pad on the right vive controller to move.
    *   when you press these keys the overlay should smoothly fade away.
    *   when you stop pressing these keys it should smoothly fade in, but centered around your current facing.
    *   additionally, when the overlay is visible, if you lean your head in toward the surface of the overlay, it will fade out and fade back in recentered about your head.
    *   As you walk around in room scale, the overlays should not come back on until a moment after you stop. In particular, they shouldn't go off and on repeatedly as you continuously walk 
    around.
    *   With the setting on, you should not be able to get your head outside the HUD sphere.
    *   Do Avatar->Reset Sensors, particularly while in Vive. The aim here is to not lose any overlays or buttons as a result of the reset.
    - :heavy_check_mark:

### Entity Manipulation

*   Equippable entities show their equip spot when you partially squeeze the trigger.
    *   Check the ping pong gun to ensure that a green ball appears around the guns handle.
    *   Check the blocks in demo to ensure that a green ball does not appear around each block.
    - :heavy_check_mark:
*   The user can pick up a grabbable object by touching the object and pulling/holding the trigger
    - :heavy_check_mark:
*   removed in-world ReticleClick behavior from handControllerPointer (e.g., track pad center not used for this any more, and no more editing with hand controller)
    - :heavy_check_mark:
*   Place your hand over a colored block on the table, press the trigger to near-grab the block. You should be able to pick it up and move it, while the trigger is pressed, it should drop when you release the trigger.
    - :heavy_check_mark:
*   Stand away from the table, press the trigger the far-grab laser should appear, when the far-grab laser lands on a colored block you should enter far-grab mode. Releasing the trigger should drop the object.
    - :heavy_check_mark:
*   Test the near-trigger behavior by placing your hand in the top marquee of the defender machine and pressing the trigger. This should reset all the blocks on the table.
    - :heavy_check_mark:
*   Test the near-hold behavior by moving your hand over the ping-pong gun and pressing the GRIP button. While the grip button is depressed, you should be able to shoot the gun with the trigger button. Releasing the GRIP button should drop the gun.
    - :heavy_check_mark:
*   Far grab an entity, and while still holding it, move the hand controller such that the beam passes through a HUD overlay. The far grab should continue as if the overlay wasn't there. However, you should not be able to initate a far grab through an overlay. (Triggering through an overlay when you are not already holding an entity should be treated as a click on the overlay.)
    - :heavy_check_mark:
*   Try particularly to pass the hold-laser through a button. The button should not fire when you're already holding something with that laser.
    - :heavy_check_mark:
*   Repeat for near grab and for equip. (Currently in the vive-ui branch, equip is involves a simultaneous trigger and press of the "grip" button on the underside of the vive controller.)
    - :heavy_check_mark:
*   Make sure you can far grab with one hand, and while still far grabbing, far grab with the other hand, too. The let go of the first. Repeat with near grab and with equip.
    - :heavy_check_mark:

### Navigation

*   Exercise trackpad navigation (left hand: strafe left/right, forward/back; right hand: turn left/right (with and without snap turns), and jump/fly (and ascend/descend while flying). These should work with thumb away from the center of the trackpad. The motion should not be mixed between "x" activities and "y" activities. (For example, jumping with snap turns on shouldn't turn while jumping.)
    - :heavy_check_mark:
*   Exercise presses using the center of the trackpad. These should make the avatar move.
    - :heavy_check_mark:
*   I can fly using my controller
    - :heavy_check_mark:

### Other

*   Go into vive hmd, and make sure hand controllers work. Then go back to desktop, and confirm the following (in desktop):
    *   Avatar should be "normal" (e.g., when viewed in third person), and not have its hand's pulled in strange ways. Waving the hand controllers around should have no effect.
        - :heavy_check_mark:
    *   The mouse should be useable and not pinned to the bottom of the screen (or pinned anywhere else). When the mouse times out after five seconds, and you bring the mouse back by moving, the cursor should be where you left it.
        - :heavy_check_mark:
    *   In Edit->Console..., Controller.getPoseValue(Controller.Standard.RightHand).valid should be false, and similarly for LeftHand.
        - :heavy_check_mark:
*   Please check overall performance vs master, in desktop and vive, and also compare pointing at an overlay vs not (in vive and if practical, desktop w/hydra).
    - **90FPS and smooth movement, comparable to the production build :heavy_check_mark:**
*   If you happen to have a vive controller that generates trigger spam (@thoys!), the Vive controllers should no longer intermittently or permanently report non-zero values even when the trigger is not being pressed at all.
    - :heavy_check_mark:
*   When using the debug flag to use OpenVR over Oculus (with Oculus hardware), the screen should no longer sit at an Oculus 'waiting' screen forever.
    - **Skipped.**

**The following PRs are being added to this branch late, and each have their own test plans that should also be tested here. (They were not present when this PR was originally tested.):**
#8099 :heavy_check_mark:
#8100 :heavy_check_mark:
#8102 :heavy_check_mark:
#8106 :heavy_check_mark:
#8108 :heavy_check_mark:
#8112 :heavy_check_mark:
#8120 :heavy_check_mark:


## Noticed bugs/expectations
- [Bug] Native cursor overlaps the rendered cursor after switching back from HMD to Desktop mode.
- [HMD-Dizzyness-Bug] Double-click with your vive controller on an entity in the edit.js entitylist, The camera will stay in position even if you move your head arround. This makes people very motionsick :). Preferred the camera always moves while moving head.
- [Bug-in-HMD] In some cases the cursor doesn't switch back to 2D mode while moving it over overlays. I cannot reproduce this well, but messing with different overlays, switching display modes, and some randomness usually gets this. (To see if you have this, close one of your eyes, and switch. Not always clearly visible when looking at it with both eyes )


### Previous bugs/expections revisit
* [Fixed] ~~The 3d mouse cursor looks rotated strangely after coming from 3rd person mode into 1st person mode. It looks angled. To fix it use the `Avatar -> Reset Sensors` menu item~~. 
* [Feature request] The Mute button on the hmd-hot-toolbar should display the current mute state of the audio
* [Bug] The hmd-toolbar is movable out of the viewable screen. This should either be restricted or a reset for the toolbar(s) should be present. **<-- This seems partially fixed, I can still move my HMD toolbar out of the View while in VR mode.**
* [Paradox] The hand of the avatar is shown behind the hud, which makes it feel kinda unconfortable to use, is there a way to overcome this without having the hud interfere with surrounding 3d objects?
* [Minor-Bug] Green grab balls indicator should also trigger a grab event when the trigger is squeezed on them, or alternatively the grab/search beam should go straight through them.
* [Fixed] ~~Rendered hud cursor looks weird after switching the Interface window to another window sometimes while in HMD mode, probably related to the first bug that I reported. No solid repro found yet! (Also fixes when using `Avatar -> Reset Sensors` )~~
* [Fixed] ~~While toggling between Interface and my text editor to write this report, I lost the laserpointer in my left hand. I was unable to grip anything with it anymore. No solid repro yet. (Restarting `handControllerGrab.js` fixed this.)~~
* [Bug] Camera goes in front of the avatar in 3rd Person view. Expected is that the camera only goes behind the avatar. No solid repro. **(Found repro, avatar seems to be behind me, in third person, while using the HMD, this seems to do something weird with the mousewheel scrolling behavior)**

interface/src/Application.cpp
jherico: `Menu` is a QML type, so you can't have a global scripting interface named `Menu` as well.

libraries/gl/src/gl/OglplusHelpers.cpp
jherico: It's a unit line on the X axis, so it's bounding sphere is a half-unit sphere centered at 0.5 on the X axis.

howard-stearns:
test this please

hifi-gustavo:
Build console output for commit: 581d87d6536b45867ea751369d5a3bd534948014 available at: https://gist.github.com/6a4c1e454f6d7f5bad14c92721b51e72<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8068.dmg

howard-stearns:
Conclusion of the group was to treat any remaining issues separately, ASAP.
Pass, and merging.

PR #8069, opened 2016-06-14
Show sysem laser when we have the opportunity for in-world pointer

howard-stearns:
When a hand controller trigger is partially or full squeezed, draw the inside-the-hud system laser.
Without this, HUD elements always draw over the top of the search lasers.

### DIFF
--- a/scripts/system/controllers/handControllerPointer.js
+++ b/scripts/system/controllers/handControllerPointer.js
@@ -420,6 +420,10 @@ function updateVisualization(controllerPosition, controllerDirection, hudPositio
     var falseProjection = intersection3d(eye, Vec3.subtract(hudPosition3d, eye));
     Overlays.editOverlay(fakeProjectionBall, {visible: true, position: falseProjection});
     */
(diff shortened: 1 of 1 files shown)


howard-stearns:
I'm not sure whether this is desirable, but I'm completing this so that we can look at the issues:
- The hands and other inside-the-hud objects are still obscured by HUD elements? Does it look wrong to show the laser but not these?
- How do we make it match the grab laser? We could adjust color to match search-vs-far-trigger, but how would we match the far-grab particles?

hifi-gustavo:
Build console output for commit: 37df0c46e6394f49f754c27752521438f9cdcad7 available at: https://gist.github.com/30702fbabe5977c129d68ac5bbb97b14<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8069.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8069.dmg

howard-stearns:
I don't think we're going to need this. Closing for now.

PR #8079, opened 2016-06-15
No laser without trigger

howard-stearns:
"Partial trigger should be the way I search for a click/interaction. It should not be on by default when I am over an overlay."

Back when we used mouse cursors, the rule was "always show a mouse cursor when I'm pointing at an overlay". This PR changes the laser to only be displayed when triggering.

### DIFF
--- a/scripts/system/controllers/handControllerPointer.js
+++ b/scripts/system/controllers/handControllerPointer.js
@@ -377,7 +377,7 @@ function turnOffVisualization(optionalEnableClicks) { // because we're showing c
     if (!optionalEnableClicks) {
         expireMouseCursor();
         clearSystemLaser();
-    } else if (!systemLaserOn) {
+    } else if (!systemLaserOn && activeTrigger.state) {
         // If the active plugin doesn't implement hand lasers, show the mouse reticle instead.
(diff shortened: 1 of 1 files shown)


howard-stearns:
test plan:
- In Vive, partially squeeze a trigger to get the search beam. Aim it a system overlay, stats display, microphone inputs display, or scripted overlay. The laser should show blue for partial trigger, and red for full trigger, and be terminated at the HUD overlay. Let go of the trigger, and the laser should not be shown any more.
- If possible, try the same in desktop with hydra hand controllers. It should be the same, but with a mouse cursor instead of a laser -- i.e., the cursor only shown when squeezing the trigger.
Note: there are buggy interactions with the virtual keyboard, which will be dealt with separately.

hifi-gustavo:
Build console output for commit: 6366ca7508ce77de4230900141298669f3c49fcf available at: https://gist.github.com/5b3f44e98a5ecacc8f96a3d7a63b24d6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8079.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8079.dmg

PR #8081, opened 2016-06-15
Continue grabbing through overlays

howard-stearns:
"When I have an overlay or an entity selected using distance grab, the action should not be interrupted when I pass over an overlay."

### DIFF
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -363,15 +363,17 @@ function MyController(hand) {
 
     var suppressedIn2D = [STATE_OFF, STATE_SEARCHING, STATE_HOLD_SEARCHING];
     this.ignoreInput = function () {
-        return (-1 !== suppressedIn2D.indexOf(_this.state)) && isIn2DMode();
+        // We've made the decision to use 'this' for new code, even though it is fragile,
+        // in order to keep/ the code uniform without making any no-op line changes.
+        return (-1 !== suppressedIn2D.indexOf(this.state)) && isIn2DMode();
     };
 
     this.update = function() {
 
         this.updateSmoothedTrigger();
 
-        if (_this.ignoreInput()) {
-            _this.turnOffVisualizations();
+        if (this.ignoreInput()) {
+            this.turnOffVisualizations();
             return;
         }
 
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -361,11 +361,15 @@ function MyController(hand) {
 
     var _this = this;
 
+    this.ignoreInput = function () {
+        return (_this.state <= STATE_HOLD_SEARCHING) && isIn2DMode();
+    };
+
     this.update = function() {
 
         this.updateSmoothedTrigger();
 
-        if (isIn2DMode()) {
+        if (_this.ignoreInput()) {
             _this.turnOffVisualizations();
             return;
         }
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -361,8 +361,9 @@ function MyController(hand) {
 
     var _this = this;
 
+    var suppressedIn2D = [STATE_OFF, STATE_SEARCHING, STATE_HOLD_SEARCHING];
     this.ignoreInput = function () {
-        return (_this.state <= STATE_HOLD_SEARCHING) && isIn2DMode();
+        return (-1 !== suppressedIn2D.indexOf(_this.state)) && isIn2DMode();
     };
 
     this.update = function() {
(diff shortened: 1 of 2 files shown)


howard-stearns:
test plan:
- Far grab an entity, and while still holding it, move the hand controller such that the beam passes through a HUD overlay. The far grab should continue as if the overlay wasn't there. However, you should not be able to initate a far grab through an overlay. (Triggering through an overlay when you are not already holding an entity should be treated as a click on the overlay.)
- Try particularly to pass the hold-laser through a button. The button should not fire when you're already holding something with that laser.
- Repeat for near grab and for equip. (Currently in the vive-ui branch, equip is involves a simultaneous trigger and press of the "grip" button on the underside of the vive controller.)

hifi-gustavo:
Build console output for commit: 2c7642a3674aa100e46c2dd4a6ba2dc6f6ba5e80 available at: https://gist.github.com/d4a4145478d2ee2628c61f78228b70e5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8081.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8081.dmg

scripts/system/controllers/handControllerGrab.js
```diff
@@ -361,11 +361,15 @@ function MyController(hand) {
 
     var _this = this;
 
+    this.ignoreInput = function () {
+        return (_this.state <= STATE_HOLD_SEARCHING) && isIn2DMode();
```
hyperlogic: This is the first code introduced to handControllerGrab that relies on the fact that the states are numbers and have a specific ordering.  I would recommend instead that you use the === operator to pick the specific states that you wish to ignoreInput on.

scripts/system/controllers/handControllerGrab.js
```diff
@@ -361,11 +361,15 @@ function MyController(hand) {
...
+    };
+
     this.update = function() {
 
         this.updateSmoothedTrigger();
 
-        if (isIn2DMode()) {
+        if (_this.ignoreInput()) {
```
hyperlogic: 95% of the code in handcontrollergrab uses this, not _this.  We should probably stick to the pattern unless there is a specific reason not to.  In this case update is intended to be used as a method, not as a plain ole closure.

scripts/system/controllers/handControllerGrab.js
howard-stearns: Wilco. While we're at it, when you change handControllerGrab, it would be cool to have the states in an Array, and so they would then be explicitly ordered. (I would have written mine to use the index into the Array. But in any case I'll change it now.)  Also, it's more idiomatic in Javascript to just use the strings themselves as values rather than the values (numbers).

scripts/system/controllers/handControllerGrab.js
howard-stearns: 'this' should not be used within function values. This will fail depending on how it's used: e.g., x = foo.bar; ..... x(17) will not work if foo refers to 'this'. So if we standardize, we should stick to a variable ('_this' or 'that' or 'self'), not the special value 'this'.  In other words, standardize is good. But pick a standard that will actually work.

hyperlogic: I strongly disagree, for object methods, we should be using `this`, not `_this`. Regardless, the if you search for occurrences of `_this` there are less than 10, but do a similar search for `this` and there are over 400.  The precedence has already been set.

hyperlogic:
This fixes one class of problem, when you have already grabbed (near or far) an object and your hand sweeps over a UI overlay.  However it does not fix two other classes of problems.

   * When you want to begin near grabbing an object but your laser inadvertently is over a button when you pull the trigger.
   * When you have the ping pong gun equipped and you attempt to shoot it while the laser is over a button.

hifi-gustavo:
Build console output for commit: 1e3b20b66d9445f49368573902f29fd26cc5df8b available at: https://gist.github.com/d4a4145478d2ee2628c61f78228b70e5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8081.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8081.dmg

hifi-gustavo:
Build console output for commit: 6f671b7cde167005965a47539e048b09bf8f5171 available at: https://gist.github.com/d4a4145478d2ee2628c61f78228b70e5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8081.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8081.dmg

howard-stearns:
@hyperlogic quite correct. This PR makes no attempt to address those cases. It is strictly about taking a non-searching state that started outside of a hud overlay, and continuing it as the laser passes through a hud.

PR #8092, opened 2016-06-17
expand density range

AndrewMeadows:
* min density = 10kg/m^3 (aerogel)
* max density = 100000 kg/m^3 (5x density of platinum)


AndrewMeadows:
Note: when setting dynamic things very light you probably also want to reduce the linear and angular drag properties.

hifi-gustavo:
Build console output for commit: 7c948b08c24a11fbc86a37ad05d2e5badc391361 available at: https://gist.github.com/4367b36f5f30f1030609bcb658bc9849<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8092.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8092.dmg

Nex-Pro:
Aerogel is typically 1.5 kg m^3

Nex-Pro:
We should use aerographene at 160g m^3

Nex-Pro:
@ZappoMan can we get testing notes for this one plz.

AndrewMeadows:
@NexPro: The motivation for this was for Caitlyn to see if these settings allowed her to get a particular bit of content working.  Some other advice was given that I think helped her make some progress so I'll check with her to see where things stand.

I'd prefer to expand the density range only as much as is necessary.  So, if content creators can get by with just 1X density of platinum I'd prefer to do that.

AndrewMeadows:
Turns out expanding the density range was not required to make the content work after all.   Therefore I'm going to close this PR and NOT let it get merged.  We may expand the density later: I'm willing to hear arguments for why it is needed.

PR #8103, opened 2016-06-20
Reduce GLShader::compileShader log spam

SamGondelman:
Previously, if an entity had a shader attached to it with an error in it, the entire shader would constantly get printed in the log.  This should prevent the error from getting logged constantly (technically it's still printed twice but it was decided that this is ok).

Also, there were outdated GLSL calls in one of our skybox shaders, causing lots of errors to be logged on Macs.  This should now be fixed.

### DIFF
--- a/libraries/gpu/src/gpu/Shader.h
+++ b/libraries/gpu/src/gpu/Shader.h
@@ -120,6 +120,9 @@ public:
     bool isProgram() const { return getType() > NUM_DOMAINS; }
     bool isDomain() const { return getType() < NUM_DOMAINS; }
 
+    void setCompilationHasFailed(bool compilationHasFailed) { _compilationHasFailed = compilationHasFailed; }
+    bool compilationHasFailed() const { return _compilationHasFailed; }
+
     const Source& getSource() const { return _source; }
 
     const Shaders& getShaders() const { return _shaders; }
@@ -180,6 +183,9 @@ protected:
 
     // The type of the shader, the master key
     Type _type;
+
+    // Whether or not the shader compilation failed
+    mutable bool _compilationHasFailed { false };
 };
 
 typedef Shader::Pointer ShaderPointer;
--- a/libraries/gpu/src/gpu/Shader.h
+++ b/libraries/gpu/src/gpu/Shader.h
@@ -185,7 +185,7 @@ protected:
     Type _type;
 
     // Whether or not the shader compilation failed
-    mutable bool _compilationHasFailed { false };
+    bool _compilationHasFailed { false };
 };
 
 typedef Shader::Pointer ShaderPointer;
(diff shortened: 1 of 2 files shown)


SamGondelman:
Testing:

- On a Macbook Pro, go to Playa or load the Day/Night skybox from Examples (under "Shaders").  You should see the shader load properly with no errors logged.
- On any system, create an entity somewhere and attach the following user data: `{"ProceduralEntity":{"version":2,"shaderUrl":"https://hifi-content.s3.amazonaws.com/samuel/shaderError.fs","uniforms":{}}}`
- The shader error will be printed twice and will not be printed anymore unless you do something like reload your content or leave the domain and come back (if you add it to a zone, it should be printed any time you re-enter the zone)

hifi-gustavo:
Build console output for commit: 34d18da4e4ecd18f330c7ae5294003628e893dfa available at: https://gist.github.com/ae14afe3572a57fd3b691d69748a1c4f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8103.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8103.dmg

kevinmthomas-carpool:
testing

kevinmthomas-carpool:
## QA Test
**QA Tester**: KevinMThomas
**QA Pass**: **QA Pass** FAILURE
**Additional Test Results + Observations**: Unable to rez the sky shader on ```playa``` as I do not have permission.  Created a box entity an added ```{"ProceduralEntity":{"version":2,"shaderUrl":"https://hifi-content.s3.amazonaws.com/samuel/shaderError.fs","uniforms":{}}}``` to the user data and no error messages were printed to log.

SamGondelman:
@kevinmthomas-carpool sorry I meant either go to Playa and verify that the skybox is working OR go to your own domain and make the skybox from the examples.  They use the same shader so either would work.  The skybox in Playa was previously broken on Macbooks, so if you can see the skybox there or when you make your own, the shader was properly fixed.

And are you sure about shaderError.fs?  The shader definitely fails to compile for me.  I guess technically they're warnings but as soon as you click outside of the user data field the log should print the entire shader twice and then stop.  Let me know if that works!

kevinmthomas-carpool:
## QA Test
**QA Tester**: KevinMThomas
**Additional Test Results + Observations**: @SamGondelman the skybox did work in ```playa```.  When I click outside the user data field the log does not print the shader twice.

libraries/gpu/src/gpu/Shader.h
```diff
@@ -120,6 +120,9 @@ class Shader {
     bool isProgram() const { return getType() > NUM_DOMAINS; }
     bool isDomain() const { return getType() < NUM_DOMAINS; }
 
+    void setCompilationHasFailed(bool compilationHasFailed) { _compilationHasFailed = compilationHasFailed; }
```
samcake: Can you put that function non public so only the gpu::Context and Backend can assigned it ?

samcake:
I m good with the new feature,
Also,  we could think of attaching an error log information ?
this way the error message of the compilation could bubble up to the user of gpu and not only available in the inner guts of the Backend

PR #8112, opened 2016-06-21
do not tilt virtual keyboard

howard-stearns:
It used to track the vive head pose. Now it is just takes the rotation.

### DIFF
--- a/plugins/openvr/src/OpenVrHelpers.cpp
+++ b/plugins/openvr/src/OpenVrHelpers.cpp
@@ -138,7 +138,7 @@ void showOpenVrKeyboard(bool show = true) {
             if (vr::VROverlayError_None == showKeyboardResult) {
                 _keyboardShown = true;
                 // Try to position the keyboard slightly below where the user is looking.
-                mat4 headPose = toGlm(_trackedDevicePose[0].mDeviceToAbsoluteTracking);
+                mat4 headPose = cancelOutRollAndPitch(toGlm(_trackedDevicePose[0].mDeviceToAbsoluteTracking));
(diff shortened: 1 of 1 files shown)


howard-stearns:
Still angled away from you, and still dropped down a bit from horizontal. We might want to adjust those going forward, but this is an improvement.

Test plan:

In vive hmd, open "goto" address bar while looking in various directions, with various head tilts as well. (You can re-position the button tray to make it easier to click on while looking in various directions.)
The keyboard should be visible and sensible. 

For the goto case, the address bar and keyboard should both be visible. For other cases (such as a text field of a qml window such a Running Scripts), it will still be possible to position the text field such that the keyboard will obscure the text field.  We will deal with that separately, if at all.

hifi-gustavo:
Build console output for commit: 2ac6dc8798995490cf4a9f02cf83aac85c773ab8 available at: https://gist.github.com/4e9fbcf2c434b2b90393ab4f0786e9ca<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8112.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8112.dmg

RyanDowne:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;iRoSNzxv&#x2F;36-decide-what-to-do-about-the-keyboard-overlay-is-too-low-to-be-comfortable-and-should-be-angled-more-if-possible"> Decide what to do about The keyboard overlay is too low to be comfortable and should be angled more if possible.</a>

PR #8113, opened 2016-06-21
fix setting velocity of inactive kinematic RigidBody

AndrewMeadows:
* fix bug where a deactivated kinematic RigidBody does not wake up when new velocity is applied

### DIFF
--- a/libraries/physics/src/EntityMotionState.cpp
+++ b/libraries/physics/src/EntityMotionState.cpp
@@ -135,7 +135,14 @@ void EntityMotionState::handleEasyChanges(uint32_t& flags) {
         _nextOwnershipBid = 0;
     }
     if ((flags & Simulation::DIRTY_PHYSICS_ACTIVATION) && !_body->isActive()) {
-        _body->activate();
+        if (_body->isKinematicObject()) {
+            // only force activate kinematic bodies (dynamic shouldn't need force and
+            // active static bodies are special (see PhysicsEngine::_activeStaticBodies))
+            _body->activate(true);
+            _lastKinematicStep = ObjectMotionState::getWorldSimulationStep();
+        } else {
+            _body->activate();
+        }
     }
 }
 


AndrewMeadows:
This fixes a bug:

(1) Make a non-dynamic EntityA
(2) Set the linear velocity of EntityA to be 1.0 in X direction --> should drift to a stop (default linear damping = 0.4)
(3) Set the linear velocity of EntityA to be -1.0 in X direction --> EntityA does not move!

The problem was that in step (2) EntityA was transitioning from **static** to **kinematic** and the logic was in place for setting it active for the transition, however in step (3) we did not have logic for force activating kinematic objects.  This bug was introduced by PR-8039 when the logic that used to transition an at-rest kinematic object to be static was removed... to fix another bug where "held" Entities were not simulated correctly in the avatar's frame.

hifi-gustavo:
Build console output for commit: 31706a39093c88b87b5a58884d18a1d306b7ddfd available at: https://gist.github.com/ab96604f11ac411a46ccac982f96c9f1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8113.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8113.dmg

Nex-Pro:
Tested on Win 10 16gb  gtx770 2gb


Setting the linear V to be 1.0 
The entity immediately moves in the positive x direction with decaying speed.
Setting the linear V to be -1 X makes the entity return to the original location.

However IMO this is still buggy.

**Observations:**

The value for linear V is not updated. Clicking the entity results in a second push.

**To reproduce:**

Starting by setting linear V to -1.0 X   The entity moves with decaying speed.
When the entity appears to have come to rest the linear V value displayed remains at -1.0 X.
Clicking on the entity gives it a second push in the negative X direction. The entity moves further, with decaying speed, in negative X 
The displayed value reduces.
Same for positive X.

AndrewMeadows:
@Nex-Pro, the bug you reported about the velocity being re-applied when leaving edit selection on the entity is a bug in edit.js and is unrelated to the bug being fixed in this PR.  We'll have to fix that in another project.

PR #8126, opened 2016-06-23
fix reticle bugs

howard-stearns:
- Make context menu come up under mouse rather than to the side, so that in HMD you don't go cross-eyed looking past the context menu at the mouse in the distance.
- Stop seeking the mouse to the center of where you are looking when we get close enough. (Was never close enough.)
- Don't require a mouse move for either of the above.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -921,17 +921,10 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer) :
                 cycleCamera();
             } else if (action == controller::toInt(controller::Action::UI_NAV_SELECT)) {
                 if (!offscreenUi->navigationFocused()) {
-                    auto reticlePosition = getApplicationCompositor().getReticlePosition();
-                    offscreenUi->toggleMenu(QPoint(reticlePosition.x, reticlePosition.y));
+                    toggleMenuUnderReticle();
                 }
             } else if (action == controller::toInt(controller::Action::CONTEXT_MENU)) {
-                auto reticlePosition = getApplicationCompositor().getReticlePosition();
-                offscreenUi->toggleMenu(QPoint(reticlePosition.x, reticlePosition.y));
-            } else if (action == controller::toInt(controller::Action::UI_NAV_SELECT)) {
-                if (!offscreenUi->navigationFocused()) {
-                    auto reticlePosition = getApplicationCompositor().getReticlePosition();
-                    offscreenUi->toggleMenu(QPoint(reticlePosition.x, reticlePosition.y));
-                }
+                toggleMenuUnderReticle();
             } else if (action == controller::toInt(controller::Action::RETICLE_X)) {
                 auto oldPos = getApplicationCompositor().getReticlePosition();
                 getApplicationCompositor().setReticlePosition({ oldPos.x + state, oldPos.y });
@@ -1240,7 +1233,16 @@ QString Application::getUserAgent() {
     return userAgent;
 }
 
-
+void Application::toggleMenuUnderReticle() const {
+    // In HMD, if the menu is near the mouse but not under it, the reticle can be at a significantly
+    // different depth. When you focus on the menu, the cursor can appear to your crossed eyes as both
+    // on the menu and off.
+    // Even in 2D, it is arguable whether the user would want the menu to be to the side.
+    const float X_LEFT_SHIFT = 50.0;
+    auto offscreenUi = DependencyManager::get<OffscreenUi>();
+    auto reticlePosition = getApplicationCompositor().getReticlePosition();
+    offscreenUi->toggleMenu(QPoint(reticlePosition.x - X_LEFT_SHIFT, reticlePosition.y));
+}
 
 void Application::checkChangeCursor() {
     QMutexLocker locker(&_changeCursorLock);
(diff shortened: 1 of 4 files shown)


howard-stearns:
test plan: we will ask the QA team to test when merging to master. For this pr (to vive-ui branch), the dev's will test. In either case, the plan is:

Bring up context menu with a right mouse click in HMD. The mouse cursor should be visible at the same apparent depth as the menu.

If the cursor is not visible before using the vive hand controller to bring up the menu, it won't be visible after either.

A source of "reticle gets stuck to where you are lookiing" has been removed, so report if you see that.

hifi-gustavo:
Build console output for commit: f8391f00629535b9849d928b0075c0f032dc61f2 available at: https://gist.github.com/1975b5276ce1bd22f028f4b8cd17e0f1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8126.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8126.dmg

interface/src/Application.cpp
```diff
@@ -1240,7 +1233,16 @@ QString Application::getUserAgent() {
...
+void Application::toggleMenuUnderReticle() const {
+    // In HMD, if the menu is near the mouse but not under it, the reticle can be at a significantly
+    // different depth. When you focus on the menu, the cursor can appear to your crossed eyes as both
+    // on the menu and off.
+    // Even in 2D, it is arguable whether the user would want the menu to be to the side.
+    const float X_LEFT_SHIFT = 50.0;
+    auto offscreenUi = DependencyManager::get<OffscreenUi>();
+    auto reticlePosition = getApplicationCompositor().getReticlePosition();
+    offscreenUi->toggleMenu(QPoint(reticlePosition.x - X_LEFT_SHIFT, reticlePosition.y));
+}
```
howard-stearns: Most of the changes to this file are just refactoring the repeated code, to go though this new method.

The offset (explained in comments) might arguably be done in buildMenu in MenuMouseHandler.qml, which is already doing some frobbing of target it is given. (I think buildMenu ought to do what it is told and have the adjustment upstream, such as in this new toggleMenuUnderReticle method. But I don't know who might be depending on buildMenu's current behavior, so I've left it alone.)

libraries/ui/src/OffscreenUi.cpp
```diff
@@ -492,10 +492,9 @@ void OffscreenUi::unfocusWindows() {
     Q_ASSERT(invokeResult);
 }
 
-void OffscreenUi::toggleMenu(const QPoint& screenPosition) {
+void OffscreenUi::toggleMenu(const QPoint& screenPosition) { // caller should already have mapped using getReticlePosition
     emit showDesktop(); // we really only want to do this if you're showing the menu, but for now this works
-    auto virtualPos = mapToVirtualScreen(screenPosition, nullptr);
-    QMetaObject::invokeMethod(_desktop, "toggleMenu",  Q_ARG(QVariant, virtualPos));
+    QMetaObject::invokeMethod(_desktop, "toggleMenu", Q_ARG(QVariant, screenPosition));
 }
```
howard-stearns: mapToVirtual throws away its argument and re-reads the Reticle position directly, which isn't very sporting, and which defeats the purpose of supplying an offset value above.  As it turns out, all callers already have the Reticle value obtained correctly anyway, so there is no need to mapToVirtualScreen again.

howard-stearns: mapToVirtual throws away its argument and re-reads the Reticle position directly, which isn't very sporting, and which defeats the purpose of supplying an offset value above.  As it turns out, all callers already have the Reticle value obtained correctly anyway, so there is no need to mapToVirtualScreen again.

scripts/system/controllers/handControllerPointer.js
```diff
@@ -208,9 +208,9 @@ function isShakingMouse() { // True if the person is waving the mouse around try
     return isShaking;
 }
 var NON_LINEAR_DIVISOR = 2;
-var MINIMUM_SEEK_DISTANCE = 0.01;
-function updateSeeking() {
-    if (!Reticle.visible || isShakingMouse()) {
+var MINIMUM_SEEK_DISTANCE = 0.1;
```
howard-stearns: We never stopped seeking because normal head movements were more of a change than this.

scripts/system/controllers/handControllerPointer.js
```diff
@@ -208,9 +208,9 @@ function isShakingMouse() { // True if the person is waving the mouse around try
...
+function updateSeeking(doNotStartSeeking) {
+    if (!doNotStartSeeking && (!Reticle.visible || isShakingMouse())) {
```
howard-stearns: We were not updating during update, but only onMouseMove, which caused sticking.
New arg and use here allow us to safely updateSeeking on line 399.

scripts/system/controllers/handControllerPointer.js
```diff
@@ -353,6 +353,16 @@ clickMapping.from(rightTrigger.full).when(isPointingAtOverlayStartedNonFullTrigg
...
+clickMapping.from(Controller.Hardware.Keyboard.RightMouseClicked).peek().to(function () {
+    // Allow the reticle depth to be set correctly:
+    // Wait a tick for the context menu to be displayed, and then simulate a (non-hand-controller) mouse move
+    // so that the system updates qml state (Reticle.pointingAtSystemOverlay) before it gives us a mouseMove.
+    // We don't want the system code to always do this for us, because, e.g., we do not want to get a mouseMove
+    // after the Left/RightSecondaryThumb gives us a context menu. Only from the mouse.
+    Script.setTimeout(function () {
+        Reticle.setPosition(Reticle.position);
+    }, 0);
+});
```
howard-stearns: Note the code comment.

RyanDowne:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;TrvQylwb&#x2F;60-make-context-menu-come-up-under-mouse-rather-than-to-the-side"> Make context menu come up under mouse rather than to the side.</a>

RyanDowne:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;DRqx56Hf&#x2F;66-reticle-depth-sometimes-wrong-on-context-menu"> reticle depth sometimes wrong on context menu</a>

PR #8136, opened 2016-06-24
Don't reposition windows prior to selecting the initial display plugin

jherico:
The code that re-positions windows doesn't properly deal with the case where the user exited in HMD mode.  If the user exits in HMD mode all the window positions are relative to the HMD-sized overlay, and will have large X values because of the extremely wide overlay size.  

This change inhibits any desktop automatic re-positioning of windows at the construction time of the desktop until such time as we enable it.  In the code, I've added a line to enable automatic re-positioning after we've loaded the settings for the first time, since this will update the menu and thus the previous window positions should now all be valid against the current plugin (the assumed plugin at the time of shutting down previously).

### DIFF
--- a/interface/resources/qml/desktop/Desktop.qml
+++ b/interface/resources/qml/desktop/Desktop.qml
@@ -24,6 +24,13 @@ FocusScope {
     readonly property int invalid_position: -9999;
     property rect recommendedRect: Qt.rect(0,0,0,0);
     property var expectedChildren;
+    property bool repositionLocked: true
+
+    onRepositionLockedChanged: {
+        if (!repositionLocked) {
+            d.handleSizeChanged();
+        }
+    }
 
     onHeightChanged: d.handleSizeChanged();
     
@@ -52,11 +59,14 @@ FocusScope {
         readonly property real menu: 8000
     }
 
-
     QtObject {
         id: d
 
         function handleSizeChanged() {
+            if (desktop.repositionLocked) {
+                return;
+            }
+
             var oldRecommendedRect = recommendedRect;
             var newRecommendedRectJS = (typeof Controller === "undefined") ? Qt.rect(0,0,0,0) : Controller.getRecommendedOverlayRect();
             var newRecommendedRect = Qt.rect(newRecommendedRectJS.x, newRecommendedRectJS.y, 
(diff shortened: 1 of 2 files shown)


jherico:
## Testing

In the production build if you shut down and restart in HMD mode the tray will end up positioned incorrectly.  In this build the tray should appear in the same location as when you last shut down.

ZappoMan:
I wonder if it wouldn't be better to store the position values as floating point percentages of the desktop size? Instead of as the absolute values for the current desktop... this might solve several of these problems.

hifi-gustavo:
Build console output for commit: f90a351400c5de2a89a30c25dc72a00fe25ca8ab available at: https://gist.github.com/1ecd401dd6e0d559a9b839b7cd735b8b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8136.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8136.dmg

howard-stearns:
I've lost track of what problem we're solving. My bad.

My frame of reference is The Resize Algorithm described at https://highfidelity.atlassian.net/wiki/display/HOME/HUD#HUD-HUDelementvisibilityandpositioning

I know that there have been problems with the old hacky toolbars not using The Resize Algorithm in various situations, 
and I'm aware that The Resize Algorithm (was never intended) to keep a constant number of pixels from an edge,
but I'm not aware of any problem with The Resize Algorithm itself.

I had assumed that the problem we were trying to solve here was that either we weren't trying to use The Resize Algorithm, or had a bug in implementing it. But I'm not sure my assumption is good. Are we thinking that we have a better idea for an algorithm to be used only for some stuff in some situations?

RyanDowne:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;BaXtk0yp&#x2F;57-tray-sometimes-repositions-inaccessibly"> tray sometimes repositions inaccessibly</a>

jherico:
@ZappoMan  @howard-stearns  I don't think this PR is the time to expand the scope of the change into "Let's re-write the positioning logic _again_".

ZappoMan:
I agree @jherico

howard-stearns:
works well

PR #8137, opened 2016-06-25
Fix bugs with Entities.addEntity

AlexanderOtavka:
Bug Fixes:
1. Entities cannot be added locally if they don't have permission from the domain server.
2. If an entity fails to be added, Entites.addEntity will return null. (Fixes [FogBugz case 365][fogbugz]).

Testing:
1. Start up sandbox
2. Navigate to <http://localhost:40100/settings/#security> in a browser
3. Disable entity rez and temp rez
4. Go to localhost in interface
5. Run the below script (don't try to use edit.js, it does it's own permission checks)
6. No entity should be created. If a box shows up in front of the avatar, it fails the test.
6. Check the console (CTRL-SHIFT-L), make sure it printed "NEW ENTITY ID = null" somewhere

```js
// Create the box properties
var pos = Vec3.sum(MyAvatar.position, Quat.getFront(MyAvatar.orientation));  
var properties = {
  type: "Box",
  position: pos
};

// Add the box
var id = Entities.addEntity(properties);
print("NEW ENTITY ID = " + id);

// Kill script after entity is created (or not created)
Script.stop();
```

[fogbugz]: https://highfidelity.fogbugz.com/f/cases/356/Entities-addEntity-can-return-a-UUID-even-when-unsuccessful

### DIFF
--- a/libraries/entities/src/EntityScriptingInterface.cpp
+++ b/libraries/entities/src/EntityScriptingInterface.cpp
@@ -191,9 +191,11 @@ QUuid EntityScriptingInterface::addEntity(const EntityItemProperties& properties
     if (success) {
         emit debitEnergySource(cost);
         queueEntityMessage(PacketType::EntityAdd, id, propertiesWithSimID);
-    }
 
-    return id;
+        return id;
+    } else {
+        return QUuid();
+    }
 }
 
 QUuid EntityScriptingInterface::addModelEntity(const QString& name, const QString& modelUrl, const glm::vec3& position) {
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -320,6 +320,11 @@ EntityItemPointer EntityTree::addEntity(const EntityItemID& entityID, const Enti
         return nullptr;
     }
 
+    if (!properties.getClientOnly() && getIsClient() &&
+        !nodeList->getThisNodeCanRez() && !nodeList->getThisNodeCanRezTmp()) {
+        return nullptr;
+    }
+
     bool recordCreationTime = false;
     if (props.getCreated() == UNKNOWN_CREATED_TIME) {
         // the entity's creation time was not specified in properties, which means this is a NEW entity


hifi-gustavo:
Build console output for commit: a4da63c1ff33ad58f04ca35ed7d31647400b71bd available at: https://gist.github.com/566f6cd926b90319e4a00ccbb26e9023<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8137.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8137.dmg

kevinmthomasse:
testing

kevinmthomasse:
## QA Test
**QA Tester**: KevinMThomas
**Additional Test Results + Observations**: Ran PR ```sandbox``` and navigated to ```http://localhost:40100/settings/#security``` in a browser and disabled ```entity rez and temp rez```.  Restarted domain.  Went to ```localhost``` and ran above script.  Immediate crash and bugsplat sent and log included.  Re-visited ```localhost``` and re-ran script again the console and successfully it printed out ```[06/27 06:46:58] [DEBUG] script:print()<< NEW ENTITY ID = null```.  Second logout successful no crash.
[hifi-log.txt](https://github.com/highfidelity/hifi/files/334643/hifi-log.txt)

AndrewMeadows:
@kevinmthomasse, we're wondering why you passed this PR when the interface crashed immediately upon running the test script the first time.  Did you think the crash was unrelated when it did not reproduce the second time?  Have you seen the interface crash in a similar way on previous PRs/installers?

kevinmthomasse:
@AndrewMeadows I passed it as I was able to relog and successfully complete the script immediately after.  I did indicate the crash and included the log however I did not want to call it a fail as in the past I have said fail and the team would caution that it is not necessarily a fail.  I should have said QA Partial Fail.

AlexanderOtavka:
@kevinmthomasse I haven't been able to reproduce the crash.  Do you know how I might do so?  Could this just be a coincidence and the crash wasn't caused by my PR?

kevinmthomasse:
@AndrewMeadows and @AlexanderOtavka that is why I passed the PR originally.   I was not able to reproduce the crash.  I included the log so that the team could review to see if it was in fact something unrelated.  The fact that it worked immediately after indicated that it was not a complete fail.  I should have for clarity said partial success however I was very clear that it crashed and the log was included.

AlexanderOtavka:
@kevinmthomasse Ok, great, thanks for the clarification.

kevinmthomasse:
My pleasure @AlexanderOtavka thank you.

AndrewMeadows:
I looked at the log file.  It seems to indicate that our "deadlock watchdog" thread decided to throw an exception, which usually means that the main thread is deadlocked.  When I loaded the core dump from the crash report it appears that the main thread was NOT blocked, but the watchdog really had thrown an exception (wasn't a crash per-se).  All the other threads appeared to be in a wait condition as expected.

It is a mystery!  If it doesn't reproduce we can discard it as "probably unrelated", but we should at least try to reproduce a few times.

AndrewMeadows:
We did some more testing and could not repro the crash. We think it is unrelated.

PR #8141, opened 2016-06-26
fix - on Linux, don't mess with the menus from the hydra plugin

sethalves:
- on Linux, don't mess with the menus from the hydra plugin because it makes Qt stop

### DIFF
--- a/plugins/hifiSixense/src/SixenseManager.cpp
+++ b/plugins/hifiSixense/src/SixenseManager.cpp
@@ -77,6 +77,7 @@ bool SixenseManager::activate() {
     InputPlugin::activate();
 
 #ifdef HAVE_SIXENSE
+    #if !defined(Q_OS_LINUX)
     _container->addMenu(MENU_PATH);
     _container->addMenuItem(PluginType::INPUT_PLUGIN, MENU_PATH, TOGGLE_SMOOTH,
                            [this] (bool clicked) { setSixenseFilter(clicked); },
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: bbe9b931bee9364581feb1ab983de329f7d0c301 available at: https://gist.github.com/c66843b7ef6450a8c27c942920d95d23<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8141.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8141.dmg

sethalves:
to test:
- use hydra on linux

AndrewMeadows:
I will test this since the problem it fixes is definitely affecting me.

plugins/hifiSixense/src/SixenseManager.cpp
```diff
@@ -106,8 +108,10 @@ void SixenseManager::deactivate() {
     InputPlugin::deactivate();
 
 #ifdef HAVE_SIXENSE
+    #if !defined(Q_OS_LINUX)
     _container->removeMenuItem(MENU_NAME, TOGGLE_SMOOTH);
     _container->removeMenu(MENU_PATH);
+    #endif
```
AndrewMeadows: These changes solve the problem and I really appreciate it,  but the fix is a HACK or at least a WORKAROUND -- it needs a comment about why these menu operations are avoided on the Linux platform... for the intrepid future developers who look at the code but also to remind us to test undoing the changes when we upgrade to Qt-6 or higher.

AndrewMeadows:
Actually, the problem this fixes is blocking me on my current project.  I'll merge it now and add the comments myself.

plugins/hifiSixense/src/SixenseManager.cpp
sethalves: Yeah... the truth is I don't know why this fixes it.

PR #8149, opened 2016-06-28
limit hand controller actions to recommended area

howard-stearns:
This keeps things from being flung off screen.


hifi-gustavo:
Build console output for commit: c492d125e080a018d20b56b322a43a523ecb9e2e available at: https://gist.github.com/4d382a08874ab2436204b2ceb3164eb6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8149.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8149.dmg

RyanDowne:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;nTI2tKE9&#x2F;64-overlays-shouldn-t-get-flung-off-screen"> overlays shouldn&#39;t get flung off screen</a>

scripts/system/controllers/handControllerPointer.js
```diff
@@ -125,13 +125,12 @@ function ignoreMouseActivity() {
     weMovedReticle = false;
     return true;
 }
-var reticleMinX, reticleMaxX, reticleMinY, reticleMaxY;
+var MARGIN = 50;
```
jherico: This margin concerns me.  It might be useful for positioning windows, but it means that there might be buttons or fields you can't get to because they're positioned along the edge of the screen.

hifi-gustavo:
Build console output for commit: 58e230094310770eb1471e05b9af28b21e2e484d available at: https://gist.github.com/4d382a08874ab2436204b2ceb3164eb6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8149.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8149.dmg

scripts/system/controllers/handControllerPointer.js
howard-stearns: I agree with the concern, but this is true in general -- i.e., it is just as true for a margin of zero.

The particular number (50, as opposed to zero or any other number), was chosen to be no bigger than our buttons, and big enough to keep a recognizable/identifiable piece of the window on the screen.  So, no matter where you grab a window or button and slide it away, you'll always have a recognizable/identiable/grabbable piece that can be used to pull it back.  And yet you can still slide windows almost entirely out of view if you want, as you can now. (So it shouldn't be a big shock to folks.)

Now that I think of it, it probably ought to be perhaps half the size of a smallest solo button -- 25 px.

howard-stearns: Hmm, that's so weird: Why did github consider that a comment on an outdated diff? The line and the comments are still valid and shouldn't be hidden.

However, I think I will change the margin to 25, at which point it will be outdated. 
Strange. Github is predicting the future...

hifi-gustavo:
Build console output for commit: b649b30777a86733c98a596c3899a2c27506c850 available at: https://gist.github.com/4d382a08874ab2436204b2ceb3164eb6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8149.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8149.dmg

howard-stearns:
test plan:
try moving all sorts of hud elements off the various edges of the screen, using the hand controllers.
try fast and slow.
A pass indicates that you cannot flings things completely off the screen, and that you can get them mostly off screen if you want to. Mouse behavior is not changed.

scripts/system/controllers/handControllerPointer.js
```diff
@@ -125,8 +125,17 @@ function ignoreMouseActivity() {
     weMovedReticle = false;
     return true;
 }
+var MARGIN = 25;
+var reticleMinX = MARGIN, reticleMaxX, reticleMinY = MARGIN, reticleMaxY;
+function updateRecommendedArea() {
+    var dims = Controller.getViewportDimensions();
+    reticleMaxX = dims.x - MARGIN;
+    reticleMaxY = dims.y - MARGIN;
```
SamGondelman: should there be something like setReticlePosition(Reticle.getPosition()) here to prevent the reticle from being outside the recommended area after the viewport dimensions change?

thoys:
## QA Testing results

- Moved many different QML / Overlays to the edges of the screen, the constraints feel quite nice :+1: 
- When dragging the `users.js` list down completely until you only see the header. (while the user list is not collapsed). The handcontrollers seem to be unable to pick the user list back up. With the cursor this seems to work fine. :-1: 

 QA Fail ( User list should be drag-able by hand controllers, in any position)

scripts/system/controllers/handControllerPointer.js
howard-stearns: Right now (without this PR) the reticle position can be outside the area. It's not like things blow up if you go outside. We just want to reduce weird dragging.

Note that updateRecommendedArea isn't being called very often. Currently every 10 seconds. If we really cared about never ever letting the cursor go outside at any time -- say, just on principle -- we would have to recognize changes at the start of every update, before any chance of moving the mouse.

thoys:
I am ok with this,  as long as it will be dealt with :+1:

PR #8155, opened 2016-06-28
Cap static mesh vertex count

AndrewMeadows:
* boxify static meshes that use more that one million vertices

### DIFF
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -696,6 +696,7 @@ void RenderableModelEntityItem::computeShapeInfo(ShapeInfo& info) {
         QVector<glm::mat4> localTransforms;
         const FBXGeometry& geometry = _model->getFBXGeometry();
         int numberOfMeshes = geometry.meshes.size();
+        int totalNumVertices = 0;
         for (int i = 0; i < numberOfMeshes; i++) {
             const FBXMesh& mesh = geometry.meshes.at(i);
             if (mesh.clusters.size() > 0) {
@@ -706,6 +707,13 @@ void RenderableModelEntityItem::computeShapeInfo(ShapeInfo& info) {
                 glm::mat4 identity;
                 localTransforms.push_back(identity);
             }
+            totalNumVertices += mesh.vertices.size();
+        }
+        const int MAX_VERTICES_PER_STATIC_MESH = 1e6;
+        if (totalNumVertices > MAX_VERTICES_PER_STATIC_MESH) {
+            qWarning() << "model" << getModelURL() << "has too many vertices" << totalNumVertices << "and will collide as a box.";
+            info.setParams(SHAPE_TYPE_BOX, 0.5f * dimensions);
+            return;
         }
 
         updateModelBounds();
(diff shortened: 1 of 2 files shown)


AndrewMeadows:
To test:

(1) obtain a model with more that 1 milliion vertices
(2) create an entity item that uses the model
(3) set the entity's collision shape type to "Static Mesh"
--> the entity should collide like a box
--> the logs should provide a WARNING about how the mesh has too many vertices

Repeat the above steps for a model with several hundred thousand vertices but less than 1 milliion
--> the entity should collide like a static mesh

hifi-gustavo:
Build console output for commit: 74781c078638fbddd9a0fe5c5cbb6960a91526e9 available at: https://gist.github.com/a3c92ee566adc158f8b4441f074d6e49<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8155.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8155.dmg

OmegaHeron:
Working on this one - If anyone else wants to eval I'll post a few test models with < 1 Megavertex and > - in a bit.

OmegaHeron:
[06/29 18:51:41] [WARNING] model "atp:/1M45K779.fbx" has too many vertices 1045779 and will collide as a box.  And yes, that creates a box roughly shape of torturous sand-spur.

Adding another that's 522974 vertices results in a static mesh shape that follows circular outline.

1M+ on right, .5M on left.

hifi-snap-by-omegaheron-on-2016-06-29_18-58-21

OmegaHeron:
QA Pass for me.

Any chance we'll have ability to (eventually) visualize static mesh collision shapes as we do with compounds?

AndrewMeadows:
@OmegaHeron.  If the static-mesh objects are NOT rendering at all when trying to switching to collisions-debug mode to view the convex-hull meshes then... that would be a **bug**.  The static-mesh collision shape looks exactly like the normal mesh... except in the case where it has >1M vertices and gets boxified (as per this PR).

Yes, when it is boxified it should render that box in collisions-debug mode.  I'm currently doing some work on the render-pipeline that will hopefully make this easier to do.

OmegaHeron:
@AndrewMeadows - Enabling option to visualize physics shapes same as for compound shape shows nothing when using a static mesh.  Violent equipment frying thunderstorm in progress here now so I can't double-double check it, but, I've yet to see it display with statics.

OmegaHeron:
@AndrewMeadows - Confirmed up to DEV Build 5091 - Windows 10 - GTX970 GPU - No display for physics seen when using static mesh types.  In a domain with mixed static/compounds I *do* see for compounds.

AndrewMeadows:
@OmegaHeron I think the bug here is that if you set anything in the "compoundURL" property (even if it is an invalid URL, or when the shape type is not "Compound") then the render logic will try to use the "collision geometry" when drawing it.  That is, the reason your meshes are not rendering is that they have a non-empty "compoundURL" property.

I came up with a fix for it in hifi/mastuer but then checked and it appears the bug is fixed in PR-8169, so I'll probably just try to get that PR merged rather than fix it again and risk a merge conflict.

PR #8164, opened 2016-06-30
Added Mac preview image if VSync is enabled

SamGondelman:
On Mac, if you have VSync enabled, when you switch to HMD mode, an image will display telling you to put on your headset (as opposed to before, when it would just be stuck on the last frame).

Also includes:
- more robust HMD window picking when switching to dk2 (because of a bug in the legacy SDK)
- the ability to enable/disable vsync on mac (except it doesn't work in multi-context settings, like when you enable an HMD)


SamGondelman:
Testing:

Mac Preview:
- On Mac, make sure VSync is enabled.  Open Interface with an HMD connected and switch to HMD mode.  Instead of being stuck on the last frame, the desktop window should show a nice instructional image.
- Go to an empty domain or somewhere where your performance is high.  Your frame rate in the HMD should be limited to ~75, not 60.
- Resize the desktop window and drag it between monitors with different device pixel ratios.  The image should still display normally.
- Disable VSync.  Now when you switch to the HMD, the live preview should be shown.  The frame rate in the HMD should still not be limited to 60.

DK2 Window picking:
- This shouldn't have any noticeable effect.  If you've had problems with the DK2 display going to the wrong screen in the past, this might help, but it's still not perfect, because the SDK is buggy.  Just try switching to DK2 mode with your monitors configured in a few different orientations.  The DK2 window should usually find the right monitor.

hifi-gustavo:
Build console output for commit: 057fc8adcecfc059ebfc9898da3cc9b174e4c1aa available at: https://gist.github.com/85bf72e291bd4cb94d04d9c994c88903<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8164.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8164.dmg

ZappoMan:
Tested on mac - it works as advertised

libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
```diff
@@ -58,9 +64,42 @@ bool HmdDisplayPlugin::internalActivate() {
         _eyeInverseProjections[eye] = glm::inverse(_eyeProjections[eye]);
     });
 
+    if (_previewTextureID == 0) {
+        const QUrl previewURL("https://hifi-content.s3.amazonaws.com/samuel/preview.png");
```
ZappoMan: hmmm - this should actually be an embedded local resource like our other images... can you please fix that.

libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
```diff
@@ -373,6 +426,31 @@ void HmdDisplayPlugin::internalPresent() {
...
+            glBindTexture(GL_TEXTURE_2D, _previewTextureID);
+            glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, _previewTexture.width(), _previewTexture.height(), 0,
+                         GL_BGRA, GL_UNSIGNED_BYTE, _previewTexture.mirrored(false, true).bits());
+            using namespace oglplus;
+            Texture::MinFilter(TextureTarget::_2D, TextureMinFilter::Linear);
+            Texture::MagFilter(TextureTarget::_2D, TextureMagFilter::Linear);
+            glBindTexture(GL_TEXTURE_2D, 0);
+            _firstPreview = false;
+        }
+        useProgram(_previewProgram);
```
ZappoMan: we probably should be using our GPU layer instead of open gl directly... so that in the future we'll be able to swap out with a Metal or Vulkan backend.

SamGondelman: as it is now, we don't have access to the GPU layer at this point

hifi-gustavo:
Build console output for commit: d05a27e0e6ad9fe36dee768b9f9ed5e31718812a available at: https://gist.github.com/85bf72e291bd4cb94d04d9c994c88903<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8164.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8164.dmg

libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
ZappoMan: oh - duh -- I didn't notice this was in the HMD plugin... ok... carry on!

ZappoMan:
I will retest with the lastest changes!

PR #8167, opened 2016-06-30
equip-hotspots feature

hyperlogic:
* Allows the ability to override position and radius of equip-hotspot
* Allows multiple equip-hotspots per entity with different attach-points per hotspot.
* Allows the user to equip an object while near or far grabbed by the other hand.

### DIFF
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -324,6 +324,14 @@ EntityPropertiesCache.prototype.getWearableProps = function(entityID) {
         return undefined;
     }
 };
+EntityPropertiesCache.prototype.getEquipHotspotsProps = function(entityID) {
+    var props = this.cache[entityID];
+    if (props) {
+        return props.userData.equipHotspots ? props.userData.equipHotspots : {};
+    } else {
+        return undefined;
+    }
+};
 
 function MyController(hand) {
     this.hand = hand;
@@ -1083,8 +1091,8 @@ function MyController(hand) {
     };
 
     this.entityIsEquippableWithDistanceCheck = function (entityID, handPosition) {
+        var distance, handJointName;
         var props = this.entityPropertyCache.getProps(entityID);
-        var distance = Vec3.distance(props.position, handPosition);
         var grabProps = this.entityPropertyCache.getGrabProps(entityID);
         var debug = (WANT_DEBUG_SEARCH_NAME && props.name === WANT_DEBUG_SEARCH_NAME);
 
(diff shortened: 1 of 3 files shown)


hyperlogic:
QA test-plan:
   * go to dev-demo ensure that all the existing items on the table still function as expected.
   * Notice the object with the red, green and blue arrows, named axes.fbx.
   * The axes objectsshould display two equip-hotspots (green spheres) instead of one.  One centered about the white sphere connected to the red, green and blue arrows.  One centered near the tip of the red arrow.
   * The hotspot near the center white sphere should be larger than the one near the tip of the red arrow.
   * When equipping the object using the larger white-sphere equip-point the object should snap to your hand such that the whitesphere is aligned to your hand, and the blue arrow is forward.
   * When equipping the object using the smaller red-arrow equip-point the object should snap to your hand such that your wrist is near the tip of the red-arrow.

hifi-gustavo:
Build console output for commit: 6072487c9c388374e707a17fdb69b7d629e7f37b available at: https://gist.github.com/f2fdefa5d823dc63f14dacbd3de81ac6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8167.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8167.dmg

scripts/system/controllers/handControllerGrab.js
```diff
@@ -273,7 +275,7 @@ EntityPropertiesCache.prototype.clear = function() {
 EntityPropertiesCache.prototype.findEntities = function(position, radius) {
     var entities = Entities.findEntities(position, radius);
     var _this = this;
-    entities.forEach(function (x) {
+    entities.forEach(function(x) {
```
howard-stearns: See examples in https://readme.highfidelity.com/docs/coding-standard#section-4-2-8-multi-line-statements-must-have-all-n-1-lines-indented-at-least-one-level-four-spaces-

FWIW, it's also a violation of jslint. (I think eslint lets you specify which style you want, but jslint insists.)

howard-stearns: I guess you're going to tell me that you don't want to change all the other occurrences in the file.
That's fine, but here you did choose to change it.
Meanwhile, there's a heck of a lot that you did choose to change, in order to make things lintable. I think that's commendable. So why go backwards here.

(Just opinions. Not a blocker.)

scripts/system/controllers/handControllerGrab.js
```diff
@@ -1255,29 +1288,31 @@ function MyController(hand) {
...
         var entity;
-        if (equippableEntities.length > 0) {
+        if (equippableHotspots.length > 0) {
             // sort by distance
-            equippableEntities.sort(function (a, b) {
-                var aDistance = Vec3.distance(_this.entityPropertyCache.getProps(a).position, handPosition);
-                var bDistance = Vec3.distance(_this.entityPropertyCache.getProps(b).position, handPosition);
+            equippableHotspots.sort(function(a, b) {
+                var aDistance = Vec3.distance(a.worldPosition, handPosition);
+                var bDistance = Vec3.distance(b.worldPosition, handPosition);
```
howard-stearns: I realize this isn't a change in this PR, but still, a performance question:
  Should we cache the distances?
  Or maybe better yet, iterate over the hotspots to find the best one, without actually sorting all the others that we don't really care about.

scripts/system/controllers/handControllerGrab.js
hyperlogic: the most recent .eslintrc.json file checked into our mainline branch flags this as an error:

```"space-before-function-paren": ["error", "never"]```

RyanDowne:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;QptisZzY&#x2F;75-i-should-be-able-to-equip-when-near-or-distance-grabbing-an-object-with-one-hand-and-equipping-the-other"> I should be able to equip when near or distance grabbing an object with one hand and equipping the other.</a>

scripts/system/controllers/handControllerGrab.js
hyperlogic: Good feedback.  It is slower.

> Should we cache the distances?

We could, although the extra bookkeeping to accomplish this IMHO is not worth the extra complexity.  In general, the number of entities that will make it to this sort will be very small.

> Or maybe better yet, iterate over the hotspots to find the best one, without actually sorting all the others that we don't really care about.

We could do this, and it would be more efficient, however, it would complicate the logic for determining which is the best match. It would introduce mutable state into the iteration loop.  IMHO one of the reasons the old search logic was so complicated and difficult to debug was that it would mutate internal state as it was looping.  Currently after my recent changes for equip-hotspots the searching logic is written in a more pure-functional style, using map, filter and sort.  The loss of performance here I think is worth the trade off for more readable code.

hifi-gustavo:
Build console output for commit: 82b1cba81dcba949e1eaf3095c110a72ad90176e available at: https://gist.github.com/f2fdefa5d823dc63f14dacbd3de81ac6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8167.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8167.dmg

hifi-gustavo:
Build console output for commit: d8e5e3cbe4b1c7af668193d738e6ef0f941de7fa available at: https://gist.github.com/f2fdefa5d823dc63f14dacbd3de81ac6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8167.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8167.dmg

thoys:
## QA Testing report

- Tested using Hydras
- Spotted two attachment-points on the Axes object in `dev-demo` :+1: 
- The attachment-point on the red axis is smaller then the one in the white sphere :+1: 
- The big attachment-point makes the blue arrow point forwards consistently, tried multiple times with both hands. :+1:  
- The small attachment point makes the object attach with the red-arrow pointing towards my wrist. :+1: 
- I was able to grab the axes with one hand, and attach while held with the other hand :+1: 

- I was able to far and neargrab the pingpong gun with my one hand and attach it with my other hand :+1: 
- The position of the grabbed object is still adjustable with the other hand :+1: 
- Was not able to equip it with my other hand while it was already equipped :+1: 


QA Pass.  Great functionality!!! :+1:  :tada:

PR #8169, opened 2016-06-30
decouple Geometry from its GeometryResource in ModelCache

AndrewMeadows:
* Model class no longer has a NetworkGeometry::Pointer.  Instead it has a Geometry::Pointer and a GeometryResourceWatcher which connects the geometry to its resource after the resource is loaded.
* Several related classes have been renamed to make code easier to read
* Added new mesh collision shape options (simple-hull and simple-compound) but they are disabled in edit.js until we can visualize them.


AndrewMeadows:
The motivation for this PR was to generate collision hulls for mesh geometry on the fly, however the render pipeline API could not easily use mesh geometry that did not have an associated "resource" that had been loaded via a URL or file name.  The focus of the work switched to making this possible and now that a bunch of changes have been made it is going up for PR review sooner rather than later.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -808,12 +827,13 @@ void RenderableModelEntityItem::computeShapeInfo(ShapeInfo& info) {
...
+                points[i] = (points[i] * scaleToFit);
+            }
         }
 
-        pointCollection.push_back(points);
-        info.setParams(SHAPE_TYPE_STATIC_MESH, 0.5f * dimensions, _modelURL);
+        info.setParams(type, 0.5f * dimensions, _modelURL);
     } else {
         ModelEntityItem::computeShapeInfo(info);
         info.setParams(type, 0.5f * dimensions);
```
AndrewMeadows: Most of the new shape generation code is above this point.  The code is very similar for the three types of shapes that analyze the render geometry directly (STATIC_MESH, SIMPLE_HULL, and SIMPLE_COMPOUND) but differ in how many collections of vertices they make and whether they use the triangle mesh indices or not.

libraries/model-networking/src/model-networking/ModelCache.h
```diff
@@ -22,52 +22,30 @@
...
+    using GeometryMeshes = std::vector<std::shared_ptr<const model::Mesh>>;
+    using GeometryMeshParts = std::vector<std::shared_ptr<const MeshPart>>;
 
     // Mutable, but must retain structure of vector
     using NetworkMaterials = std::vector<std::shared_ptr<NetworkMaterial>>;
 
-    const FBXGeometry& getGeometry() const { return *_geometry; }
-    const NetworkMeshes& getMeshes() const { return *_meshes; }
+    const FBXGeometry& getFBXGeometry() const { return *_fbxGeometry; }
+    const GeometryMeshes& getMeshes() const { return *_meshes; }
```
AndrewMeadows: Variable names have been changed to make the code more readable.  There used to be three distinct implementations of getGeometry() for different (but related!) classes.  I removed one and renamed another (to getFBXGeometry()).

libraries/model-networking/src/model-networking/ModelCache.h
```diff
@@ -79,9 +57,9 @@ class Geometry {
     friend class GeometryMappingResource;
 
     // Shared across all geometries, constant throughout lifetime
-    std::shared_ptr<const FBXGeometry> _geometry;
-    std::shared_ptr<const NetworkMeshes> _meshes;
-    std::shared_ptr<const NetworkShapes> _shapes;
+    std::shared_ptr<const FBXGeometry> _fbxGeometry;
+    std::shared_ptr<const GeometryMeshes> _meshes;
+    std::shared_ptr<const GeometryMeshParts> _meshParts;
```
AndrewMeadows: The "Network" part of these class names was not helping so I've come up with more descriptive monikers.

libraries/model-networking/src/model-networking/ModelCache.h
```diff
@@ -118,30 +96,49 @@ class GeometryResource : public Resource, public Geometry {
...
+protected:
+    friend class GeometryMappingResource;
+
+    virtual QSharedPointer<Resource> createResource(const QUrl& url, const QSharedPointer<Resource>& fallback,
+        const void* extra);
+
+private:
+    ModelCache();
+    virtual ~ModelCache() = default;
 };
```
AndrewMeadows: ModelCache is not a new class.  It moved up here from below to handle name declaration dependencies that cropped up when trying to use a pre-existing typedef.

libraries/render-utils/src/Model.h
```diff
@@ -128,18 +128,18 @@ class Model : public QObject, public std::enable_shared_from_this<Model> {
...
+    const Geometry::Pointer& getCollisionGeometry() const { return _collisionGeometry; }
 
-    const QVariantMap getTextures() const { assert(isLoaded()); return _geometry->getGeometry()->getTextures(); }
+    const QVariantMap getTextures() const { assert(isLoaded()); return _renderGeometry->getTextures(); }
     void setTextures(const QVariantMap& textures);
 
     /// Provided as a convenience, will crash if !isLoaded()
     // And so that getGeometry() isn't chained everywhere
-    const FBXGeometry& getFBXGeometry() const { assert(isLoaded()); return getGeometry()->getGeometry()->getGeometry(); }
+    const FBXGeometry& getFBXGeometry() const { assert(isLoaded()); return _renderGeometry->getFBXGeometry(); }
```
AndrewMeadows: You're welcome.

libraries/model-networking/src/model-networking/ModelCache.h
```diff
@@ -118,30 +96,49 @@ class GeometryResource : public Resource, public Geometry {
...
 
 private slots:
     void resourceFinished(bool success);
     void resourceRefreshed();
 
 private:
     GeometryResource::Pointer _resource;
-    Geometry::Pointer _instance { nullptr };
+    Geometry::Pointer& _geometryRef;
+};
```
AndrewMeadows: The GeometryResourceWatcher class is key to the changes in this PR.  It replaces the NetworkGeometry class which "owned" the Geometry::Pointer for the Model class.  Now the Model class "owns" a Geometry::Pointer and the GeometryResourceWatcher get a reference to it so that the Model's geometry can be changed when a resource finishes loading.

In short: the Model has more direct access to its geometry and the watcher helps maintain it.

AndrewMeadows:
When it comes to testing don't bother testing the new collision shapes yet.  They are obfuscated from use for now and we'll test them properly in a future PR.  Instead these are the questions to answer:

(1) Do models still load and render?

(2) Do animated models that require mesh blending still render and animate properly?

(3) When  you change the modelURL of a ModelEntityItem does its visible shape still change accordingly?

(4) Does the "static-mesh" shape still collide correctly?

hifi-gustavo:
Build console output for commit: 342d9132541eeed8235707823f6a1b50f9dc1fea available at: https://gist.github.com/b43684f8a30ac4b0994b2115b47f957e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8169.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8169.dmg

hifi-gustavo:
Build console output for commit: 08b468f77d13ee8a9552a91fbfa0df5016ee6a42 available at: https://gist.github.com/b43684f8a30ac4b0994b2115b47f957e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8169.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8169.dmg

libraries/render-utils/src/GeometryCache.h
```diff
@@ -396,8 +396,6 @@ class GeometryCache : public Dependency {
     QHash<Vec2FloatPairPair, GridBuffer> _gridBuffers;
     QHash<int, GridBuffer> _registeredGridBuffers;
 
-    QHash<QUrl, QWeakPointer<NetworkGeometry> > _networkGeometry;
-
```
AndrewMeadows: This data member was declared but not used anywhere.

hifi-gustavo:
Build console output for commit: 73fcd55a69fab41c86bf81cc21ec09d97b1afb41 available at: https://gist.github.com/b43684f8a30ac4b0994b2115b47f957e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8169.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8169.dmg

OmegaHeron:
(1) Do models still load and render? - Yes
(2) Do animated models that require mesh blending still render and animate properly? - Untested as I don't have a good example to compare
(3) When you change the modelURL of a ModelEntityItem does its visible shape still change accordingly? - Yes
(4) Does the "static-mesh" shape still collide correctly? - Yes
(5) I noticed this bug and think maybe it doesn't reproduce on older versions - I don't always see behavior describe - sometimes it collides after being given its shape - sometimes it doesn't.  When it does fail to "solidify" changing domains or relog does correct.
(6) Another bug to test. - Again I find this to be intermittent.  There was a time many months ago that it was 100% 

I'm still testing this - just wanted to give some preliminary results.

libraries/render-utils/src/Model.h
thoys: LOL! The previous code looked pretty uh, descriptive.

AndrewMeadows:
Ryan tested in the "cell science" region which has some examples of animated meshes.  Also, I _think_ the animated avatar model goes through the same code path so that should test it too.

RyanDowne:
Per conversation: Modify this PR so that edit.js has access to the new hull types and we can test the new functionality.

AndrewMeadows:
After thinking this over I'm going to split this PR into two independent parts: new collision shapes (without visualization) and render-pipeline cleanup.

hifi-gustavo:
Build console output for commit: ddef2e98354a7720afa43608ae5ea6fd74c962ed available at: https://gist.github.com/b43684f8a30ac4b0994b2115b47f957e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8169.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8169.dmg

PR #8175, opened 2016-07-01
Fix eslint descrepencies

AlexanderOtavka:
Prohibit trailing commas, and fix function spacing rule to be more in
line with the coding standard.  Also prohibit leading decimal points,
also specified in coding standard.

Testing:
1. Make sure eslint is installed: `npm i -g eslint`, you need node for this: <https://nodejs.org/en/>
2. Create a file called `foo.js`
3. Copy-paste in the script below
4. Run `eslint -c path/to/hifi/.eslintrc.js --color foo.js` from the same directory as `foo.js`, replacing `path/to/hifi/` with the relevant path
5. It should throw errors for `space-before-function-paren`, `no-floating-decimal`, and `comma-dangle`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -52,13 +52,14 @@ module.exports = {
     },
     "rules": {
         "brace-style": ["error", "1tbs", { "allowSingleLine": false }],
-        "comma-dangle": ["error", "only-multiline"],
+        "comma-dangle": ["error", "never"],
         "camelcase": ["error"],
         "curly": ["error", "all"],
         "indent": ["error", 4, { "SwitchCase": 1 }],
         "keyword-spacing": ["error", { "before": true, "after": true }],
         "max-len": ["error", 128, 4],
         "new-cap": ["error"],
+        "no-floating-decimal": ["error"],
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 5b343c9b82e322f6452c4e3f48244fe55272457e available at: https://gist.github.com/fcfec34c27b0890d10e768e9c1989c3e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8175.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8175.dmg

imgntn:
is the trailing commas thing in the coding standard?  i sometimes find it helpful, for the reasons listed on the eslint page ```Trailing commas simplify adding and removing items to objects and arrays, since only the lines you are modifying must be touched. Another argument in favor of trailing commas is that it improves the clarity of diffs when an item is added or removed from an object or array:``` it's pretty minor overall so approving

imgntn:
aha -- i see it in 4.2.8...

AlexanderOtavka:
@imgntn Yeah, personally I like trailing commas, but I'm a slave to the coding standard.

PR #8179, opened 2016-07-01
Various snapshot improvements

SamGondelman:
- Snapshot sharing has been broken for a while for various reasons.  Unfortunately, it was decided to just remove it completely.
- Instead, added a signal that you can connect to.  Added that signal to notifications.js so that it shows you the path to your snapshot and a preview of the snapshot.
- the setting for where to save snapshots (under Settings -> Avatar) was not showing up, but it should now
- folder dialog selection pops up on first snapshot
- pressing CTRL-S to take a snapshot should no longer leave you stuck moving backwards
- pressing CTRL-S to take a snapshot after disabling overlays should not turn them back on


SamGondelman:
Testing:

- Take a snapshot.  No dialog should pop up.
- Take a snapshot with CTRL-S.  You shouldn't start moving backwards.

hifi-gustavo:
Build console output for commit: 218b26b5213cd1bc8af041064b4d5e82644e7334 available at: https://gist.github.com/1a39e9ddbc9346b4575b8249ae503d2b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8179.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8179.dmg

ZappoMan:
I think we should keep the preview/popup window... just remove the "share" button from it

hifi-gustavo:
Build console output for commit: 4eb94f8ccbd48d6b86f2ca23945c34fea2122d8e available at: https://gist.github.com/1a39e9ddbc9346b4575b8249ae503d2b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8179.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8179.dmg

hifi-gustavo:
Build console output for commit: b3513b543c50899edc63c9fda74bd386e4916e27 available at: https://gist.github.com/1a39e9ddbc9346b4575b8249ae503d2b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8179.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8179.dmg

hifi-gustavo:
Build console output for commit: 79864df08ea91abf042bbf20ffc02068d91c9693 available at: https://gist.github.com/1a39e9ddbc9346b4575b8249ae503d2b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8179.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8179.dmg

RyanDowne:
I don't see the preview. I see the notification for it but the actual snapshot is not showing up. I am also seeing and hearing things differently depending on if I am in Desktop or HMD mode

hifi-gustavo:
Build console output for commit: 890b22426c44b5773257c483d17e00e86809aab9 available at: https://gist.github.com/1a39e9ddbc9346b4575b8249ae503d2b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8179.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8179.dmg

nbq:
testing this now

scripts/system/notifications.js
```diff
@@ -336,12 +366,11 @@ function createNotification(text, notificationType) {
...
 
     if (Menu.isOptionChecked(PLAY_NOTIFICATION_SOUNDS_MENU_ITEM) &&
-        Menu.isOptionChecked(NotificationType.getMenuString(notificationType)))
-    {
+        Menu.isOptionChecked(NotificationType.getMenuString(notificationType))) {
         randomSounds.playRandom();
     }
 
-    return notify(noticeProperties, buttonProperties, height);
+    return notify(noticeProperties, buttonProperties, height, imageProperties);
```
AndrewMeadows: You're calling notify() with four arguments here but it takes five.  Is that ok?  Do you handle the case where the fifth argument is undefined?

SamGondelman: Yeah, the logic is a little circuitous because I was building it in on top of the existing system, but essentially:

- For existing notifications, imageProperties and image will both be undefined, so it won't get into the case at line 297.
- For the new snapshot notification:
    - notify is called here with imageProperties defined and image undefined
    - it will go in the case at line 297 because (imageProperties && !image) and call notify again with imageProperties defined and image set to true
    - it will NOT go into the if case at line 297 again, because image is true

Definitely let me know if you can devise an easier way to do this.

libraries/input-plugins/src/input-plugins/KeyboardMouseDevice.cpp
```diff
@@ -49,9 +49,11 @@ void KeyboardMouseDevice::InputDevice::focusOutEvent() {
 
 void KeyboardMouseDevice::keyPressEvent(QKeyEvent* event) {
     auto input = _inputDevice->makeInput((Qt::Key) event->key());
-    auto result = _inputDevice->_buttonPressedMap.insert(input.getChannel());
-    if (!result.second) {
-        // key pressed again ? without catching the release event ?
+    if (!(event->modifiers() & Qt::KeyboardModifier::ControlModifier)) {
```
ZappoMan: why is this change in this PR? Can you please elaborate? what does this have to do with snapshot support?

SamGondelman: Without this, pressing CTRL-S to take a snapshot would leave you moving backwards (and I believe also turn the overlays back on if you had toggled them off).  If this seems too hacky I can investigate a better solution.

ZappoMan: It looks like you've change the previous logic in a way that will impact all keys. IIRC - the original code is designed to detect and handle lost release events on all keys, which can happen if the focus changes mid press (I think). Your new code appears to remove that general logic. Can you verify please.

SamGondelman: That code is in Application::focusOutEvent and is untouched.

And actually, the issue does affect all keys.  In master, CTRL-A will leave you spinning to the left even after you release the keys.  You have to press A again (by itself, CTRL-A has no effect) to stop, so it isn't a focus issue.

It could be a QT bug, but here is the behavior I'm seeing in master (in Application):
- Press and release A by itself.  keyReleaseEvent is received.
- Press and release CTRL by itself.  keyReleaseEvent is received.
- Press CTRL-A, and let go of A, but keep pressing CTRL.  *NO keyReleaseEvent*.  Let go of CTRL and you'll get the CTRL keyReleaseEvent, but the event for A was lost.
- Press CTRL-A, and let go of CTRL, but keep pressing A.  Release A.  Both keyReleaseEvents received at the correct times.
- This is only an issue with the CTRL key AFAICT.  At least on mac (where Command == CTRL), other modifier keys work fine.

So my change just masks the issue by not forwarding CTRL modified to the KeyboardMouseDevice, but the state in Application is technically messed up.  I'm not sure what the best actual fix would be.

PR #8187, opened 2016-07-05
Test support for limiting texture quality

jherico:
Test limiting the texture quality to determine how much of an impact it has on lower end machines.  This PR limits textures to 512x512 at the file loading level.

### DIFF
--- a/libraries/model/src/model/TextureMap.cpp
+++ b/libraries/model/src/model/TextureMap.cpp
@@ -52,8 +52,18 @@ void TextureMap::setLightmapOffsetScale(float offset, float scale) {
     _lightmapOffsetScale.y = scale;
 }
 
+TextureQuality TextureUsage::_quality = TextureQuality::LOW;
+
 const QImage TextureUsage::process2DImageColor(const QImage& srcImage, bool& validAlpha, bool& alphaAsMask) {
     QImage image = srcImage;
+    if (_quality != TextureQuality::NATIVE) {
+        auto maxDimension = static_cast<int>(_quality);
+        auto size = image.size();
+        if (size.width() > maxDimension || size.height() > maxDimension) {
+            auto newSize = QSize(std::min(size.width(), maxDimension), std::min(size.height(), maxDimension));
+            image = image.scaled(newSize);
+        }
+    }
     validAlpha = false;
     alphaAsMask = true;
     const uint8 OPAQUE_ALPHA = 255;
@@ -202,6 +212,16 @@ gpu::Texture* TextureUsage::createNormalTextureFromNormalImage(const QImage& src
         image = image.convertToFormat(QImage::Format_RGB888);
     }
 
+    // FIXME we probably need special scaling functionality for normal maps.
+    if (_quality != TextureQuality::NATIVE) {
+        auto maxDimension = static_cast<int>(_quality);
+        auto size = image.size();
+        if (size.width() > maxDimension || size.height() > maxDimension) {
+            auto newSize = QSize(std::min(size.width(), maxDimension), std::min(size.height(), maxDimension));
+            image = image.scaled(newSize);
+        }
+    }
+
     gpu::Texture* theTexture = nullptr;
     if ((image.width() > 0) && (image.height() > 0)) {
         
@@ -233,6 +253,16 @@ gpu::Texture* TextureUsage::createNormalTextureFromBumpImage(const QImage& srcIm
     
     if (image.format() != QImage::Format_RGB888) {
         image = image.convertToFormat(QImage::Format_RGB888);
+    }
+
+    // FIXME we probably need special scaling functionality for normal maps.
+    if (_quality != TextureQuality::NATIVE) {
+        auto maxDimension = static_cast<int>(_quality);
+        auto size = image.size();
+        if (size.width() > maxDimension || size.height() > maxDimension) {
+            auto newSize = QSize(std::min(size.width(), maxDimension), std::min(size.height(), maxDimension));
+            image = image.scaled(newSize);
+        }
     }
 
     // PR 5540 by AlessandroSigna integrated here as a specialized TextureLoader for bumpmaps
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 8585d714f62963bce9a5ac728d519311cd6062ee available at: https://gist.github.com/569b25fbe07c291735502f943c7f98f7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8187.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8187.dmg

hifi-gustavo:
Build console output for commit: d7842604e6c83ab2f7ff0c834464ea837d89795c available at: https://gist.github.com/569b25fbe07c291735502f943c7f98f7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8187.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8187.dmg

Triplelexx:
Certainly seems to help me on my below spec machine. On production I will almost always crash when loading Playa, this build seems more stable. Once loaded my fps was between 30-60.
It seems to have an adverse effect on the Playa skybox though, resulting in very compressed/pixelated sky.

I'll see if I can find somewhere else to do some more tests.

jherico:
@Triplelexx The Playa skybox uses a 8k by 4k texture from [here](http://svs.gsfc.nasa.gov/vis/a000000/a003500/a003572/).  It's 23 MB _compressed_ and uncompressed it probably works out to be something 128 MB memory (8k *48k * 4 bytes per pixel).  Forcing it down to 512x512 from 16kx8k therefore has a drastic impact on it's visual quality, far more than any other part of the scene.  I need to refactor the skybox to a) use a pre-compressed version of the texture and store it on the GPU in compressed form, and b) use a cubemap version of the image rather than a single huge file.  

The former optimization will reduce the impact on lower end machines that might not have that much memory to spare for 'pretty sky'.  The latter optimization will ensure that even if we do limit the texture quality on load, the resulting image won't be quite as degraded, because we'll be using 6 images of 512x512 each, rather than just one.

Triplelexx:
That makes sense now. :) 
I tend to often only use 5 sides of a cubemap disregarding the bottom as well, using a star map that wouldn't be suitable if it was for a space scene.

hifi-gustavo:
Build console output for commit: cfe1ca0fc58020ce14c44c96daa62bc081d2dfca available at: https://gist.github.com/569b25fbe07c291735502f943c7f98f7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8187.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8187.dmg

hifi-gustavo:
Build console output for commit: cfe1ca0fc58020ce14c44c96daa62bc081d2dfca available at: https://gist.github.com/569b25fbe07c291735502f943c7f98f7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8187.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8187.dmg

hifi-gustavo:
Build console output for commit: 7cbec17f01882a1d4ff8fb4e76ce8740b09da9df available at: https://gist.github.com/569b25fbe07c291735502f943c7f98f7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8187-7cbec17.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8187-7cbec17.dmg

OmegaHeron:
Been playing with this as suggested in latest RC comments.  If you have a few moments @jherico, I have a some questions.  I want to understand what's happening under the hood with graphics so I can sort whether I'm seeing expected or unexpected behavior vs assuming it's all wrong.

Premise:  I'm connecting to standard home scene - insuring it's completely loaded and that I've moved through it and had every object come into fully visible scope.  I record a GPU VRAM use of X then I move to a known non-existent domain getting a star field and record GPU VRAM use of Y.  Y is < X by a minor amount.

I reconnect to standard home scene and, again, allow it to fully load and record a new GPU VRAM use of X2 - back to disconnected space and record GPU RAM Y2...  I keep doing this and, eventually, I see a maximal GPU VRAM use of 3500(ish) MB or on some runs 2050(ish) MB.

1) Are you basing max VRAM use on card type?  I ask because 3.5G is exactly what you'd cap on a GTX970 for max performance due to its upper 512MB being speed restricted.

2) Since I'm viewing identical content why does each return to standard home increase VRAM use?  Is it that you have no association to items fetched to compose a scene vs what's in GPU VRAM so you can't know you already have "something" in VRAM and use it?  Or something else?

3) I'm not sure I understand the trade-off between simply cleaning out GPU RAM of objects not even in scene vs moving it back to system RAM and doing a fairly expensive shuffle.  When I see log spewing that it's moving textures from GPU RAM to CPU RAM I see a corresponding choppy movement - lots of stutters/freezes as it manages those swaps.

4) Some runs I see peak GPU VRAM use never exceed 2050MB area - those runs I enter the realm of unresponsive much faster and more severely.  I'm **not** messing with setting GPU limit in devel menu - it's always on its default of auto (or whatever its precise wording is).

I'm just totally unclear on how so much of this works - it seems there's a lot of woe induced by not clearing out GPU VRAM on domain changes and it doesn't **seem** like there's a lot of gain by having GPU stuffed with stuff...  It's no faster loading standard home scene on 2nd visit after it's been fully rezzed than first.  In fact, each subsequent visit tends to be slower until GPU VRAM maxes out then it worsens.

Having some understanding of method to madness would help a lot in me recognizing something is just the way it is vs bug.  Also - I don't consider an i5 6500 CPU + GTX970 machine under spec.  It's, for CPU, a bit better than minimum spec and spot on target with GPU.  Compared to my Alienware laptop with i7m processor and GTX660M GPU the i5/GTX970 runs circles around it for everything but Interface.

jherico:
@OmegaHeron how are you recording GPU memory use?

OmegaHeron:
GPU-z, could it be it's fooling me?  I know interface shows + and - deltas via its reporting tools, but, GPU-z only shows minuscule decreases over run session.  I've watched other programs with it and see far more +/- delta over their runs.

OmegaHeron:
Here's a CSV dump from GPU-Z for a session where I only connected to home scene then no domain space repeatedly.

[GPU-Z Sensor Log.txt](https://github.com/highfidelity/hifi/files/369598/GPU-Z.Sensor.Log.txt)

jherico:
@OmegaHeron are you still enabling the OpenGL 4.5 engine?

OmegaHeron:
@jherico - No, I'm not, double checked that env var is not defined.

jherico:
@OmegaHeron found a serious issue with our texture management code that was causing our internal reporting stats to be completely off and lots of memory to be leaked.  Making a fix for it now.

jherico:
@OmegaHeron try https://github.com/highfidelity/hifi/pull/8266

PR #8191, opened 2016-07-06
basic hand-controller editing

howard-stearns:
Make the most basic edit.js operations work with hand controllers: selection, position, stretch, rotate.
Does not do multi-selection, copy-drag, or advanced editing features.
Also:
- fix minor regression in handControllerGrab: lasering through 3d overlays, again
- fix minor bug in handControllerGrab: far grabbing while editing.
- fix a couple of unguarded null property references in edit.js/entitySelectionTool.js
- makes the stretch boxes a bit bigger

Does NOT address separate existing issues:
- html text fields not opening the Vive keyboard.
- edit of a physical objects is jumpy
- rotation is jumpy and some rotations unobtainable (e.g., fine-grained 0)


hifi-gustavo:
Build console output for commit: b8e6572ebfe490cd587922b2a7617b2575915f4c available at: https://gist.github.com/896393c6da6629129948c1b7b2692529<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8191.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8191.dmg

howard-stearns:
**test plan:**

This isn't meant to fully support all entity editing through the hand controller -- see exclusions above. This just gets the basics working so that simple changes can be done in HMD, the team can explore what doesn't work well about 2d-style affordance-dragging, and we close the regression from recent builds.

With the **edit button activated in HMD**, one should be able to do the following with either left or right hand controller:
- use blue search laser, near grab, equip, etc., as before. However:
  -  the laser should not penetrate through the selected-entity affordances (up-down and rotate arrows, and the stretch boxes),
  - a full trigger should not far grab (while edit is open)
- select one entity at a time by full-triggering 
- drag the selected object in the x-z plane (by moving a full-triggered controller)
- stretch and rotate by full trigger dragging on the various affordances

howard-stearns:
test this please

hifi-gustavo:
Build console output for commit: b8e6572ebfe490cd587922b2a7617b2575915f4c available at: https://gist.github.com/896393c6da6629129948c1b7b2692529<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8191.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8191.dmg

scripts/system/controllers/handControllerGrab.js
```diff
@@ -221,6 +221,14 @@ function entityHasActions(entityID) {
     return Entities.getActionIDs(entityID).length > 0;
 }
 
+function findRayIntersection(pickRay, precise, include, exclude) {
+    var entities = Entities.findRayIntersection(pickRay, precise, include, exclude);
+    var overlays = Overlays.findRayIntersection(pickRay);
+    if (!overlays.intersects || (entities.distance <= overlays.distance)) {
```
huffman: Should `entities.intersects` also be checked here? I'm not sure what entities.distance defaults to.

howard-stearns: good catch. will do.

scripts/system/libraries/entitySelectionTool.js
```diff
@@ -1005,6 +1005,56 @@ SelectionDisplay = (function() {
...
+            var controllerPosition = Vec3.sum(Vec3.multiplyQbyV(MyAvatar.orientation, controllerPose.translation),
+                                              MyAvatar.position);
+            // This gets point direction right, but if you want general quaternion it would be more complicated:
+            var controllerDirection = Quat.getUp(Quat.multiply(MyAvatar.orientation, controllerPose.rotation));
+            return {origin: controllerPosition, direction: controllerDirection};
+        }
+    }
+    function generalComputePickRay(x, y) {
+        return controllerComputePickRay() || Camera.computePickRay(x, y);
+    }
```
huffman: This chunk of code seems better fit to live in edit.js where the rest of the input detection is, but it would require a fair amount of refactoring to do right. I think there is work in progress to create proper edit tools, but if we want to support mouse+hand edit.js long-term then we may want to consider refactoring edit.js a bit.

scripts/system/libraries/entitySelectionTool.js
```diff
@@ -4320,7 +4373,7 @@ SelectionDisplay = (function() {
     that.updateHandleSizes = function() {
         if (selectionManager.hasSelection()) {
             var diff = Vec3.subtract(selectionManager.worldPosition, Camera.getPosition());
-            var grabberSize = Vec3.length(diff) * GRABBER_DISTANCE_TO_SIZE_RATIO * 2;
+            var grabberSize = Vec3.length(diff) * GRABBER_DISTANCE_TO_SIZE_RATIO * 5;
```
huffman: @PhilipRosedale Might have input on this, I think he changed it semi recently (months). Probably hard to balance monitor vs hmd.

scripts/system/libraries/entitySelectionTool.js
howard-stearns: The idea is to have some basic functionality while the cooler stuff gets developed -- which could take a while before the dust settles on it.

A refactor could be pretty complicated, unless I'm looking at it wrong. The pick ray stuff is not used in edit.js, but only here in entitySelectionTool.js.  So that's why the activeHand and related stuff is here.

hifi-gustavo:
Build console output for commit: 110f5e9e2adaab0353cc8f3df3ecb01b5aacfe78 available at: https://gist.github.com/896393c6da6629129948c1b7b2692529<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8191.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8191.dmg

scripts/system/libraries/entitySelectionTool.js
huffman: re: pick ray stuff, the thought I had in mind was to add a pick ray object to what is now the mouse event object, so the callbacks/functions in this file don't have to worry about generating it, and thus don't need to worry about the source of the event. But, this would be more work than it's worth if this will eventually be replaced.

scripts/system/libraries/entitySelectionTool.js
howard-stearns: Yes. Philip suggested the increase.  It might be necessary to make it different HMD vs Desktop, but I'd like to see if we can get away with it being simpler.

If we really want to get into it, there's a bunch of grabberSizeCorner/Edge/Face stuff that doesn't seem to actually matter. (It gets used and the replaced.)  I opted to not attempt to clean that up. :-)

scripts/system/libraries/entitySelectionTool.js
howard-stearns: understood.

scripts/system/libraries/entitySelectionTool.js
huffman: Yeah, I think at this point it's safe to say that those grabbers won't be added back ;)

PR #8195, opened 2016-07-06
Exposed RSA key generation to scripting

aoighost:
https://worklist.net/20753

### DIFF
--- a/libraries/networking/src/CryptoScriptingInterface.cpp
+++ b/libraries/networking/src/CryptoScriptingInterface.cpp
@@ -1,0 +1,15 @@
+//
+//  Created by Peter Clemenko III on 2016/05/31
+// 
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include "CryptoScriptingInterface.h"
+
+#include "RSAKeypairGenerator.h"
+
+void CryptoScriptingInterface::generateKeypair() {
+	RSAKeypairGenerator rsaKeyPairGenerator;
+	rsaKeyPairGenerator.generateKeypair();
+}
--- a/libraries/networking/src/CryptoScriptingInterface.h
+++ b/libraries/networking/src/CryptoScriptingInterface.h
@@ -1,0 +1,27 @@
+//
+//  CryptoScriptingInterface.h
+//  libraries/networking/src
+//
+//  Created by Peter Clemenko III on 2016/05/31
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+
+#ifndef hifi_networking_CryptoScriptingInterface_h
+#define hifi_networking_CryptoScriptingInterface_h
+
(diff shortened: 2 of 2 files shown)


hifi-gustavo:

libraries/networking/src/CryptoScriptingInterface.h
```diff
@@ -0,0 +1,27 @@
...
+#define hifi_networking_CryptoScriptingInterface_h
+
+#include <QtCore/QObject>
+
+#include <DependencyManager.h>
+
+class CryptoScriptingInterface : public QObject, public Dependency {
+	Q_OBJECT
+public:
+	Q_INVOKABLE void generateKeypair();
```
huffman: Should this be returning the private/public keys?

libraries/networking/src/CryptoScriptingInterface.h
```diff
@@ -0,0 +1,27 @@
...
+
+
+#ifndef hifi_networking_CryptoScriptingInterface_h
+#define hifi_networking_CryptoScriptingInterface_h
+
+#include <QtCore/QObject>
+
+#include <DependencyManager.h>
+
+class CryptoScriptingInterface : public QObject, public Dependency {
```
huffman: It doesn't look like this is exposed to the script engine. There are examples of how to do this in ScriptEngine.cpp

huffman:
ok to test

hifi-gustavo:
Build console output for commit: 62becc6a000eb2cf0bff5e07d601f949d45fc7c2 available at: https://gist.github.com/87b7167e3f16b2f1c6ddfc54f4c56286<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8195.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8195.dmg

jherico:
This should include a script (or a pointer to a script on the net) that tests the functionality.

libraries/networking/src/CryptoScriptingInterface.cpp
```diff
@@ -0,0 +1,15 @@
...
+//
+
+#include "CryptoScriptingInterface.h"
+
+#include "RSAKeypairGenerator.h"
+
+void CryptoScriptingInterface::generateKeypair() {
+	RSAKeypairGenerator rsaKeyPairGenerator;
+	rsaKeyPairGenerator.generateKeypair();
+}
```
AndrewMeadows: @huffman is right: `RSAKeyPairGenerator::generateKeypair()` returns an `AsymmetricCipherKeyPair` instance, however that instance is not harvested here and is thus not passed to a higher scope for actual use.

It isn't clear to me: what would a JS script that uses crypto keypairs actually do?  Perhaps we need a hypothetical demo JS script that actually does something.  It could then be used as part of the test-plan for this PR to verify that everything works.

ZappoMan:
Please update this PR with fixes to the feedback. Thanks.

libraries/networking/src/CryptoScriptingInterface.cpp
CrackerHax: Use cases.....
https://en.wikipedia.org/wiki/Public-key_cryptography#Typical_use`

PR #8199, opened 2016-07-07
Start of OSVR HDK plug-in

ctrlaltdavid:


hifi-gustavo:
Build console output for commit: e78c721f1427f8aee3d033a85e11745bbea25ca3 available at: https://gist.github.com/ee28752a58bfa4e6223845cdd5f370b7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8199.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8199.dmg

interface/external/boost/readme.txt
```diff
@@ -0,0 +1,21 @@
+
+Instructions for adding the Boost library to Interface
+David Rowe, 6 Jul 2016
+
+1.  Download the Boost version 1.57 library from https://sourceforge.net/projects/boost/files/boost-binaries/ and unzip or
```
jherico: We don't really do this anymore.  Instead we should set up an external project for the boost files (if we really need them) and have it downloaded on demand, rather than requiring a user to download something.

plugins/osvrHMD/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+#
```
jherico: going forward, I think we should try to prefix the name hifi to all of our own plugins.  Additionally, I don't think we need the HMD suffix in the name.

plugins/osvrHMD/src/OsvrHmdDisplayPlugin.cpp
```diff
@@ -0,0 +1,47 @@
...
+//  Created by David Rowe on 5 Jul 2016.
+//  Copyright 2016 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include "OsvrHmdDisplayPlugin.h"
+
+#ifdef HAVE_OSVR
```
jherico: we should avoid the use of `HAVE_XXX` macros.  They make the code less readable.  Since we can detect whether we have the OSVR SDK from the CMake files, if it's not available, just don't build the plugin.  All the plugin code can therefore assume that the SDK is available, or it wouldn't be getting built.

interface/external/boost/readme.txt
ctrlaltdavid: Correct. The intent of this early PR was something concrete to discuss with Leo & Co. as to whether they want the auto-download of OSVR and Boost library files now or later.

plugins/osvrHMD/CMakeLists.txt
ctrlaltdavid: There are two sides to OSVR: the HMD side and the controllers side (not just hand controllers, but all sorts). This plugin is for the HMD side of things, hence the suffix.
You'd like the plugin called hifiOSVRHMD? (Or perhaps hifiOsvrHmd?)

plugins/osvrHMD/src/OsvrHmdDisplayPlugin.cpp
ctrlaltdavid: Where's the best place to disable building the plugin?

plugins/osvrHMD/CMakeLists.txt
ctrlaltdavid: Or maybe this is the appropriate place for OSVR controllers to be integrated at some stage, in which case it would be best to lose the "HMD" suffix?

plugins/osvrHMD/src/OsvrHmdDisplayPlugin.cpp
thoys: In the Cmake files probably :)

plugins/osvrHMD/CMakeLists.txt
jherico: That's what I was thinking.  Both the Oculus and OpenVR plugins support the hand controllers and the HMD in a single plugin.  It's also much easier to do this since each of them essentially gives you a handle or pointer to your SDK session.  If OSVR has such a pointer, it would be tricky to have it shared across multiple DLLs.

plugins/osvrHMD/src/OsvrHmdDisplayPlugin.cpp
jherico: In the CMakelists.txt for the plugin itself.  See the hifiSpacemouse plugin CMakelists.txt for an example.

plugins/osvrHMD/CMakeLists.txt
ctrlaltdavid: Yeah, I came to the same conclusion and have renamed to hifiOSVR, thanks.

plugins/osvrHMD/src/OsvrHmdDisplayPlugin.cpp
ctrlaltdavid: Thanks very much for the pointer. I almost had it but not quite; have it sorted now.

PR #8200, opened 2016-07-07
Batch optimizations

jherico:
Nearly half the time of rendering a frame consists of creating the batches.  Therefore, any optimizations to the core batch functions should have a direct impact on the rendering performance.  

This PR does two things.  

### Aggressive pre-allocation

The existing batch code tries to pre-allocate memory in _some_ of it's members based on the last batch size.  However, there are two problems.  It doesn't hit enough of the members (it doesn't touch the cache objects at all) and batch sizes can be highly variable, so a batch with few commands followed by one with lots of commands will end up having to do a lot of allocation.

In this PR, strategy is altered.  Batches now contain static max size members for each of their vectors.  When a batch is destroyed, the max size is changed to be the greater of the existing max size, or the current vector size.  When a batch is created, it enough space for that size vector.  A similar strategy is being used inside the `Batch::Cache` class to pre-allocate it's vector.  Additional memory is consumed for each batch, but the amount of memory is negligible on an application scale, and performance is measurably improved.

### Temporary variable usage reduction

All uses of `std::vector::push_back` in Batch and it's inner classes have been changed to `std::vector::emplace_back`.  _Push_ will invoke a default constructor and then a copy operator for the item being inserted into the list, while _emplace_ will simply invoke an argument constructor or a copy constructor, depending on the parameters.

For instance, many of the commands perform operations like `_params.push_back(srcViewport.y);` where the `srcViewport.y` is a float.  `_params` is a `std::vector<Batch::Param>` where `Batch::Param` is a struct that has a float constructor.  This results in a temporary `Batch::Param` being created, passed to `std::vector::push_back` which then either uses a default constructor and copy operator, or a just copy constructor.  Using `emplace_back` will pass the `srcViewport.y` parameter into function as is, where it will be used to do a placement new at the target location with it as the constructor parameter.  Only one instance of `Batch::Param` is created, as opposed to `push_back` which creates two.

## Results

Using my performance testing tool with the content from playa loaded, I got the following performance results:

**Production code**

    FPS  158.71
    FPS  168.68
    FPS  161.62
    FPS  165.86
    FPS  159.96
    FPS  161.23
    FPS  163.4
    FPS  157.03

**Aggressive Caching**

    FPS  166.12
    FPS  169.48
    FPS  169.06
    FPS  168.53
    FPS  168.57
    FPS  166.6
    FPS  168.46
    FPS  167.63

**Aggressive caching and emplace_back**

    FPS  179.69
    FPS  179.01
    FPS  177.9
    FPS  177.56
    FPS  179.92
    FPS  177.13
    FPS  176.89
    FPS  177.09

### DIFF
--- a/libraries/gpu/src/gpu/Batch.cpp
+++ b/libraries/gpu/src/gpu/Batch.cpp
@@ -26,7 +26,7 @@ ProfileRangeBatch::~ProfileRangeBatch() {
 }
 #endif
 
-#define ADD_COMMAND(call) _commands.push_back(COMMAND_##call); _commandOffsets.push_back(_params.size());
+#define ADD_COMMAND(call) _commands.emplace_back(COMMAND_##call); _commandOffsets.emplace_back(_params.size());
 
 using namespace gpu;
 
@@ -89,9 +89,9 @@ size_t Batch::cacheData(size_t size, const void* data) {
 void Batch::draw(Primitive primitiveType, uint32 numVertices, uint32 startVertex) {
     ADD_COMMAND(draw);
 
-    _params.push_back(startVertex);
-    _params.push_back(numVertices);
-    _params.push_back(primitiveType);
+    _params.emplace_back(startVertex);
+    _params.emplace_back(numVertices);
+    _params.emplace_back(primitiveType);
 
     captureDrawCallInfo();
 }
@@ -99,9 +99,9 @@ void Batch::draw(Primitive primitiveType, uint32 numVertices, uint32 startVertex
 void Batch::drawIndexed(Primitive primitiveType, uint32 numIndices, uint32 startIndex) {
     ADD_COMMAND(drawIndexed);
 
-    _params.push_back(startIndex);
-    _params.push_back(numIndices);
-    _params.push_back(primitiveType);
+    _params.emplace_back(startIndex);
+    _params.emplace_back(numIndices);
+    _params.emplace_back(primitiveType);
 
     captureDrawCallInfo();
 }
@@ -109,11 +109,11 @@ void Batch::drawIndexed(Primitive primitiveType, uint32 numIndices, uint32 start
 void Batch::drawInstanced(uint32 numInstances, Primitive primitiveType, uint32 numVertices, uint32 startVertex, uint32 startInstance) {
     ADD_COMMAND(drawInstanced);
 
-    _params.push_back(startInstance);
-    _params.push_back(startVertex);
-    _params.push_back(numVertices);
-    _params.push_back(primitiveType);
-    _params.push_back(numInstances);
+    _params.emplace_back(startInstance);
+    _params.emplace_back(startVertex);
+    _params.emplace_back(numVertices);
+    _params.emplace_back(primitiveType);
+    _params.emplace_back(numInstances);
 
     captureDrawCallInfo();
 }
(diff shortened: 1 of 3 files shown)


hifi-gustavo:
Build console output for commit: bad14f3beeb4297d45e7469282c82394b31779c7 available at: https://gist.github.com/90cabcd121405b8b00539bee370270f7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8200.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8200.dmg

libraries/gpu/src/gpu/Batch.cpp
```diff
@@ -26,28 +26,43 @@ ProfileRangeBatch::~ProfileRangeBatch() {
...
 
 using namespace gpu;
 
-Batch::Batch(const CacheState& cacheState) : Batch() {
-    _commands.reserve(cacheState.commandsSize);
-    _commandOffsets.reserve(cacheState.offsetsSize);
-    _params.reserve(cacheState.paramsSize);
-    _data.reserve(cacheState.dataSize);
-}
+size_t Batch::_commandsMax { 128 };
```
ZappoMan: Let's use a constant here... I assume 128 is arbitrary, let's give it a name and a comment for why you picked that as the default size.

libraries/gpu/src/gpu/Batch.cpp
```diff
@@ -458,22 +475,11 @@ void Batch::preExecute() {
     }
 }
 
-QDebug& operator<<(QDebug& debug, const Batch::CacheState& cacheState) {
```
ZappoMan: why are you removing this? Isn't there still value in having a debug stream operator?

libraries/gpu/src/gpu/Batch.h
```diff
@@ -491,6 +480,9 @@ class Batch {
     void captureDrawCallInfoImpl();
 };
 
+template <typename T>
+size_t Batch::Cache<T>::_max = 128;
```
ZappoMan: some constant... this is why we don't use magic numbers. please fix.

libraries/gpu/src/gpu/Batch.cpp
ZappoMan: never mind... I see this is for the CacheState which you removed

ZappoMan:
Dev Approved - tentative to you fixing the magic number.

hifi-gustavo:
Build console output for commit: 2ecb1f9bd471669d59b373dd1dceb2dd20529707 available at: https://gist.github.com/90cabcd121405b8b00539bee370270f7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8200.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8200.dmg

hifi-gustavo:
Build console output for commit: b73fe2484813d0177dffd9970abbff0617e248bc available at: https://gist.github.com/90cabcd121405b8b00539bee370270f7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8200.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8200.dmg

ZappoMan:
test this please

OmegaHeron:
Testing Windows 10 64 bit - i5 6500 CPU - GTX 970 GPU - Desktop Mode and DK2 HMD
As in #8202 I'm seeing FPS gains matching @jherico specs.  Playa load "feels" smoother and couldn't find any locations near playa landing point where FPS was less than 74.3 with DK2.  Standard home scene also shows better performance.  I couldn't find anything that looked/behaved differently with this PR vs current DEV/RC builds.  I suspect this plus other will be a nice booster.

ZappoMan:
once jenkins builds this. I will merge it

hifi-gustavo:
Build console output for commit: b73fe2484813d0177dffd9970abbff0617e248bc available at: https://gist.github.com/90cabcd121405b8b00539bee370270f7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8200.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8200.dmg

PR #8236, opened 2016-07-13
Update all asset server responses to be reliable

huffman:


hifi-gustavo:
Build console output for commit: b4b7b7ea34aeaf98a07bc8a98a0cab6f52484048 available at: https://gist.github.com/fcbe41bad6d292e72ba802ada69762fe<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8236-b4b7b7e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8236-b4b7b7e.dmg

huffman:
This solves _some_ issues with ATP requests never finishing, although it does seem like there is an additional bug somewhere else. It would be best to test this with a tool like clumsy (https://jagt.github.io/clumsy/) to test packet loss. Although you may still see failures on this PR, it should be markedly better than the current release.

To test:
 * Run clumsy with filtering set to `outbound` and Drop enabled with `5`% chance.
 * Connect to your local domain
 * Use "Edit -> Reload Content" to reload the content in home several times
 * The content will load slower than usual
 * Observe that all downloads eventually finish, although may sometimes have 1 or 2 assets not finish. This should be rare.

Please report back on the error rate. On release versions I reliably see ATP downloads not finish with 5% loss.

OmegaHeron:
QA Pass
Testing Windows 10 64 bit
Standard home scene via localhost, also accessed it via 2nd machine on LAN.
Drop enabled at 5, 10 and 15% all resulted in, eventual, full load of home scene after cache clear.  ATP upload and other operations work as expected.  Standard smoke tests pass (audio etc).
I did 5+ cache reloads, deleted cache from disk (to **really** clear it) twice and didn't have any transfers hang regardless of loss settings.  That's an improvement for sure from current RC/DEV.

PR #8252, opened 2016-07-15
suggestions

howard-stearns:
Show up to three suggested domains when you bring up the GoTo address bar. Clicking on picture takes you there. Can filter results by typing.


howard-stearns:
test plan:
- Bring up the address bar (e.g., GoTo button). It should display up to three suggestions. Each suggestion should have a picture, the principle name of the domain, and the number of users present. The number present should be in non-increasing order among the three. (I.e., prefers those with more people present, up to a max of 20.)
- You can click on a picture to go to that domain. (Note that in testing you will be running a newer protocol than most domains, and so you won't actually be able to display stuff once you are "there".) When you bring up the address bar again, the suggestions should NOT include the place you are now in.
- You can filter results by typing.  It is based on whether any of the domain's various place names or descriptions contain the given text, in any word order and case-insenstive. So, e.g., even though "Playa" is is displayed as "hq", you should be able to type "playa" and have it eventually display only "hq". You should be able to delete characters to bring more suggestions back. 
- If the picture is not available, a generic default should be shown. (During development, there was a domain named "Talk" that was up, but which had a picture that had something wrong. If that domain is up, you can try typing "talk" and, if it is up, you should still see a picture.
- Please repeat in HMD and desktop.

hifi-gustavo:
Build console output for commit: ce99146733deb50a425aa200ba213e056112a4b8 available at: https://gist.github.com/dc43c4deab1842e1c57efe5219f9a1d0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-ce99146.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-ce99146.dmg

hifi-gustavo:
Build console output for commit: 8faa961c8dfdb3f8586c73545708250bd6a4bca2 available at: https://gist.github.com/dc43c4deab1842e1c57efe5219f9a1d0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-8faa961.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-8faa961.dmg

interface/resources/qml/AddressBarDialog.qml
```diff
@@ -44,11 +46,52 @@ Window {
...
+                text: "Suggestions"
+            }
+            Row {
+                Card {
+                    id: s0;
+                    width: 200;
+                    height: 200;
+                    goFunction: goCard
+                }
+                HifiControls.HorizontalSpacer { }
```
jherico: This is unnecessary.  use the Row [`spacing`](http://doc.qt.io/qt-5/qml-qtquick-row.html#spacing-prop) property

interface/resources/qml/AddressBarDialog.qml
```diff
@@ -130,14 +173,148 @@ Window {
                 }
                 font.pixelSize: hifi.fonts.pixelSize * root.scale * 0.75
                 helperText: "Go to: place, @user, /path, network address"
+                onTextChanged: filterChoicesByText()
```
jherico: You may want to consider using a `Timer` with a small delay attached to it.  Instead of instantly modifying the suggestions on every keystroke, potentially making the UI laggy, restart the timer on each keystroke with a 200 ms timeout.  When the timer expires, then do updates to the UI.  this prevents you from wasting processing while the person is typing quickly with a destination in mind.  

We may also want to consider doing something similar to the chrome address bar, where if you start typing a destination that matches something in your history, you autofill the remainder of the address bar with the closest matching destination in a selected state, ensuring that the caret position is unaffected.

interface/resources/qml/controls-uit/HorizontalSpacer.qml
```diff
@@ -0,0 +1,19 @@
...
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+import QtQuick 2.5
+
+import "../styles-uit"
+
+Item {
```
jherico: as pointed out in the Row comment, this type is unnecessary.

interface/resources/qml/AddressBarDialog.qml
```diff
@@ -130,14 +173,148 @@ Window {
...
+        });
+        fill1(s0, filtered[0]);
+        fill1(s1, filtered[1]);
+        fill1(s2, filtered[2]);
+    }
+
+    function fillDestinations() {
+        suggestionChoices = null;
+        getDomains({minUsers: 0, maxUsers: 20}, function (error, domains) {
+            var here = addressBarDialog.getHost(); // don't show where we are now.
```
jherico: The address manager is exposed to QML, so you can directly call `AddressManager.hostname` here

interface/src/ui/AddressBarDialog.cpp
```diff
@@ -40,6 +40,10 @@ AddressBarDialog::AddressBarDialog(QQuickItem* parent) : OffscreenQmlDialog(pare
     _forwardEnabled = !(DependencyManager::get<AddressManager>()->getForwardStack().isEmpty());
 }
 
+QString AddressBarDialog::getHost() const {
```
jherico: unnecessary,  AddressManager is already exposed to QML.

interface/resources/qml/hifi/Card.qml
```diff
@@ -0,0 +1,64 @@
...
+        height: parent.height;
+        fillMode: Image.PreserveAspectCrop;
+        // source gets filled in later
+        anchors.verticalCenter: parent.verticalCenter;
+        anchors.left: parent.left;
+        onStatusChanged: {
+            if (status == Image.Error) {
+                console.log("source: " + source + ": failed to load " + JSON.stringify(userStory));
+                // FIXME: let's get our own
+                source = "http://www.davidluke.com/wp-content/themes/david-luke/media/ims/placeholder720.gif"
```
jherico: We should initialize the source value to this as well, so that if no source gets set it already has the placeholder image in it.

howard-stearns:
good feedback, @jherico . wilco.

interface/resources/qml/AddressBarDialog.qml
howard-stearns: I'm going to skip this one for now. I'm a very fast typist, and I can't get my (desktop machine) to fall behind.  This version is only designed for current scale, so there's only a couple hundred domains to filter.  In the long run, this will all be done differently.

hifi-gustavo:
Build console output for commit: cc367f7b49b635a10cc3a3cb7c6587364a52846f available at: https://gist.github.com/dc43c4deab1842e1c57efe5219f9a1d0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-cc367f7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-cc367f7.dmg

hifi-gustavo:
Build console output for commit: 99e9a76803d52b7cf07c84ad2e61b6e8ccc31186 available at: https://gist.github.com/dc43c4deab1842e1c57efe5219f9a1d0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-99e9a76.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-99e9a76.dmg

interface/resources/qml/AddressBarDialog.qml
```diff
@@ -44,11 +45,51 @@ Window {
...
+    property var suggestionChoices: [];
+    property int cardWidth: 200;
+    property int cardHeight: 152;
+
     AddressBarDialog {
         id: addressBarDialog
         implicitWidth: backgroundImage.width
         implicitHeight: backgroundImage.height
 
+        Column {
```
jherico: There's no reason to use a `Column` with only one child element.  You can use an `Item` or just child the `Row` directly to the `AddressBarDialog`

interface/resources/qml/AddressBarDialog.qml
```diff
@@ -130,14 +171,159 @@ Window {
...
+                return getDomains(options, function (error, others) {
+                    cb(error, domains.concat(others));
+                });
             }
+            cb(null, domains);
+        });
+    }
 
+    function filterChoicesByText() {
+        function fill1(target, data) {
```
jherico: This kind of filler function isn't needed.   You can define you `Card.qml` with a `property var data` field.  Then in the cards `Image` object, you can define `source` as `source: data ? data.lobby : defalutPicture`.  QML will recognize when `data` changes and all the fields that depend on it will be updated automatically.

interface/resources/qml/AddressBarDialog.qml
howard-stearns: Yup. I'll remove. (It was there from when there was an additional "Suggestions:" text above the row.)

interface/resources/qml/AddressBarDialog.qml
howard-stearns: I love dependency-directed-backtracking.
I'll change that next time around. (Right now, it's handy for logging stuff during debugging.)

hifi-gustavo:
Build console output for commit: b106dee3e9cb876cbead876246f79fcbb825d947 available at: https://gist.github.com/dc43c4deab1842e1c57efe5219f9a1d0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-b106dee.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-b106dee.dmg

hifi-gustavo:
Build console output for commit: f73c2b884dfa3ea4e3eb657d56ed19983f79bef0 available at: https://gist.github.com/dc43c4deab1842e1c57efe5219f9a1d0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-f73c2b8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-f73c2b8.dmg

interface/resources/qml/AddressBarDialog.qml
jherico: Bear in mind, even if you've got additional elements, nesting columns and rows is generally a pain in the ass in QML.  Better off making an item with the text at the top and anchoring the row beneath it.

interface/resources/qml/hifi/Card.qml
```diff
@@ -0,0 +1,96 @@
...
+        id: users;
+        size: textSize;
+        color: hifi.colors.white;
+        anchors {
+            bottom: parent.bottom;
+            right: parent.right;
+            margins: textPadding;
+        }
+    }
+    MouseArea {
```
jherico: By putting the mousearea last you're putting it at the top of the z-order.  Because it has `hoverEnabled: true` it will prevent any of the other items in this container from receiving hover events.  Moving it to the top will fix this.  In this case it probably doesn't matter because the card is essentially a giant read-only button, but if it did have fields that should respond to mouseover, and you still wanted to receive clicks over the entire surface, you'd need to break this up into two MouseAreas, one at the top of the z-order to get the clicks, and one at the bottom to get the hovers without masking other child object hover events.

interface/resources/qml/AddressBarDialog.qml
jherico: Yeah, whenever you declare or initialize a QML field with another field, or a piece of logic, or a function, you're actually creating a [property binding](http://doc.qt.io/qt-5/qtqml-syntax-propertybinding.html).  If you just assign to a field (i.e. `foo.bar = baz`) in a function, then you're just setting the value and breaking any binding, but it is possible to actually create a binding inside a function like that by doing `foo.bar = Qt.binding(function() { return baz; })`.  Then if `baz` changes, your `foo.bar` is updated automatically.  

It's pretty awesome.

interface/resources/qml/hifi/Card.qml
howard-stearns: interesting. Yeah, we'll need that in future versions.

hifi-gustavo:
Build console output for commit: d446e1bcf54eee13320b071910e689b5be910993 available at: https://gist.github.com/dc43c4deab1842e1c57efe5219f9a1d0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-d446e1b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-d446e1b.dmg

ZappoMan:
Too many domains can't be connected to... I can only connect to 1 out some 20 domains shown....

howard-stearns:
@ZappoMan, @RyanDowne and @zzmp: I'm going to make some assumptions here:
1. That the failure to connect reported by Zappoman is not from protocol mismatch, as that that gives a nice error message, and isn't going to be a general issue for people running beta.
2. That it's not generally from the three suggestions that come up without typing anything to filter on, because these are the ones with highest occupancy and that means they are running.
3. That the failure is when typing to filter, and thus getting one of the 240 or so domains that are in the database, but which might not be one of the 10-12 or so that are typically running.

As I understand it, we do not currently have a query option on https://metaverse.highfidelity.io/api/v1/domains/all to report only domains that are actually running, or a property in the returned values to tell us whether the domain is running. We will change that in due course, but that requires a change to a different repository and restarting the metaverse server, so it's tough to coordinate.

However, we do already have a query parameter for 'active', which doesn't mean running, but rather that it has one person in it, and of course that implies that it is running. This parameter works and I was using it before. I removed it because I thought that people would rather see a match for the thing they were typing, even if it was down, then to see nothing.  Zappoman's reaction tells me that I'm wrong.
 
So, I will make the change now to put that query parameter back. The suggestions will then only offer domains that have at least one person. We won't offer domains that are empty but running, because we don't yet have a way to check. When that is changed, we'll update this query.

Does that work for everyone?  I feel that it exactly this sort of issue that makes it important that we tease out these issues in actual use, so I think it's important to get this out there and see what else happens.

ZappoMan:
@howard-stearns - even when I just hit "enter" - I got suggestions I couldn't connect to... it's possible these were HiFi machines behind the hifi office firewall and they were misconfigured (not the right port)... they were often BlackLodge and IdiaLab

ZappoMan:
also @howard-stearns - I couldn't connect to domains that claimed to have a person in them... and indeed I was not getting protocol errors... I think it's possible these are either:
1) not actually running and there's a metaverse problem
2) are running, but networking prevents them from be useful.

Either way -- this feature as it stands right now... is too frustrating to release. We need to figure out why this is happening. Also note: @PhilipRosedale was testing with me... and he had the same experience.

howard-stearns:
Well fuck.

I've put the 'active' query flag in, and it's rebuilding. But that won't help with the "BlackLodge" problem. That's @hyperlogic 's dev machine. I assume what's happening there is that he's left it running with his avatar there, and concurrency is so low that this is enough to be listed. Meanwhile, it's probably not on a port that our office router lets traffic through. So, I _imagine_ that the domain data was designed to have domain servers initiate a "phone home to the metaverse" connection to report their data, without any check that domain is actually reachable.  Fuck.

hifi-gustavo:
Build console output for commit: cf025072e16f2fee538af06b7b10b6715ac463a8 available at: https://gist.github.com/dc43c4deab1842e1c57efe5219f9a1d0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-cf02507.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-cf02507.dmg

howard-stearns:
With the 'active' restriction, I don't now get "BlackLodge", so there must not be anyone there.  (You can check with https://metaverse.highfidelity.io/api/v1/domains/all?open&active) 

In my checks now, I find that the only active places I can't get into are protocol mismatches. So if I'm understanding correctly, this doesn't "fix" the problem because when Tony goes back into BlackLodge, it report as active but still not be accessible. However, does this mean that the _scale_ of the problem is brought down to our dev machines?  

On the other hand, there are a lot of machines with at least one person that I CAN get into, which means that a lot of people are running new protocol versions. I'm not sure what the implications are for that.

howard-stearns:
Current status is that only places that have a user present are suggested -- regardless of whether or not you type in a specific name.  Incompatible versions are not checked until you try to connect. (We are looking into what it would take to pre-filter that.)

birarda:
If you're unable to connect to machines running at the office, it's likely because those have been left with the default domain server port of 40102. The NATing at the office will only setup a firewall state for a single machine on that port, which makes any domain-server that spins up after the first unable to be reached. If Tony is running BlackLodge from the office have him confirm he's not using the default 40102 before it's deemed unreachable.

This issue is specific to our network setup at the office, and while it may be run into by other institutions with multiple domain-servers running behind their network I've specifically decided in the past not default to port randomization for the domain-server. I think this is debatable but tangential.

In the long run, it is unlikely to be sufficient to say that since somebody is in the domain we assume it will be reachable. I think we have a few of options here that we would run in tandem:

(1) The domain-servers should be able to do some tests to determine to the best of their ability if they will be reachable. This information should be clearly displayed by the Sandbox/domain-server so the domain owner knows what steps they can take if they are deemed unreachable. Interface clients behind difficult NATs should be provided similar information (why they cannot connect to some domains and how they would stick their port and open it on their router to assist).

(2) We have Interface clients determine the NAT type they are behind (easy to do with multiple STUN servers) and report connection success/failure to domains when they attempt to connect. We use this data to determine when domains appear to be unreachable from particular configurations or locations (geo-IP blocking, for example) and then hide those domains from users when we are pretty sure they will not be able to get there.

(3) If need be, we could also collect the same information in (2) by running probes at the central layer behind different NAT types.

I'll leave it up to those already discussing on what we should do in the short term. I think it is likely that this will still be valuable even with our low numbers of concurrency. It sounds like at least some of the connection failures for suggested places were due to domain-server default port on office machines.

howard-stearns:
Brilliant! I think that's exactly right, and particularly about surfacing the information to domain operators so that it is actionable by the right people.

cozza13:
testing this. 

"HQ" shows as 5 people when those 5 people are at the place name "Stacks" per for the concurrency page. (see images). If I click on "HQ" I end up going to a different X,Y,Z compared to going to "Stacks" X,Y,Z . 

capture2

capture

howard-stearns:
That's correct. Suggestions is by domain, not by place. The concurrency board is finer-grained wrt location, but courser grained wrt sampling frequency. So, the suggestions numbers should be right (for all places within the domain), and the aggregate of the sum of the concurrency board numbers for the same domain should be sort-of right, depending on timing.

The domain info api does not break down users by place.

birarda:
I would put it outside the scope of this PR, but we do have "referring hostname" data for domains now. We should leverage that to have the domain API return split counts by name, which would allow us to show where we really think people are.

We may also need to allow the domain owner to pick a "default place" where we bucket users that did not arrive via a place name. That would also be what we would use in the case where we are querying domains and not places.

It does seem odd to me that submissions would query for domains and not places - the images/names/descriptions are all tied to places, and there can be multiple distinct places in a single domain.

howard-stearns:
@birarda: Places api does not expose online_users data.  Suppose it did: we could sort by that, but then the client would have to correlate that info to the domain info -- unless the places API also did that for us and also provided the various filtering.

birarda:
Yes, I understand that it is the domains API that has been built out with some of the information you're beginning to leverage. I'm simply stating that I think it's possible we'll discover that what we're actually wanting to query for and display is places in online domains.

hifi-gustavo:
Build console output for commit: 17e64cf5e6e4cc46660590f2e50d5e09273fcf20 available at: https://gist.github.com/dc43c4deab1842e1c57efe5219f9a1d0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-17e64cf.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-17e64cf.dmg

howard-stearns:
Dev reviewer: This was Dev Approved, but the following changes have been made since:
1.  Activity reporting was added. These changes are the files UserActivityLogger.cpp, AddressManager.h/cpp, and AddressBarDialog.h/cpp.
2. Protocol versioning was added. This requires https://github.com/highfidelity/hifi/pull/8269 (which has not yet been merged, and so those changes to DomainServer and PacketHeaders show up here), and it is used in AddressManager.h/cpp.
3. The above is now used in AddressBarDialog.qml.

howard-stearns:
Note the DNM explantion that has been added to PR description, above.

hifi-gustavo:
Build console output for commit: 57955a2b5612d287ad33dac1b96cf33fd1b33fec available at: https://gist.github.com/dc43c4deab1842e1c57efe5219f9a1d0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-57955a2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-57955a2.dmg

howard-stearns:
test this please

hifi-gustavo:
Build console output for commit: 57955a2b5612d287ad33dac1b96cf33fd1b33fec available at: https://gist.github.com/dc43c4deab1842e1c57efe5219f9a1d0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-57955a2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-57955a2.dmg

hifi-gustavo:
Build console output for commit: 4a6132874f57b8ff463ee9117cd48ad9a6a15904 available at: https://gist.github.com/dc43c4deab1842e1c57efe5219f9a1d0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-4a61328.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8252-4a61328.dmg

howard-stearns:
Test note addendum:
This now shows only those domains which are:
- named domains (not temporary names)
- have a "place" defined
- Specified as open right now (which will be true by default).
- has at least one person already present
- are not restricted in who can enter
- MATCH YOUR PROTOCOL VERSION
- additionally, unless you are typing something that in the address bar, has a picture.

Thus, to test this, you will only see suggestions that meet the criteria above.
For now, I'm running two boxes with this PR build that do meet that criteria. (howard, and hfload).
Note: if you want multiple machines on the same network to be accessible, you must run the server on port 0, which is not the default. (Otherwise you'll collide over port 40102.)

interface/resources/qml/AddressBarDialog.qml
```diff
@@ -130,22 +170,178 @@ Window {
...
+        allDomains = suggestionChoices = [];
+        getDomains({minUsers: 0, maxUsers: 20}, function (error, domains) {
+            if (error) {
+                console.log('domain query failed:', error);
+                return filterChoicesByText();
+            }
+            var here = AddressManager.hostname; // don't show where we are now.
+            allDomains = domains.filter(function (domain) { return domain.name !== here; });
+            // Whittle down suggestions to those that have at least one user, and try to get pictures.
+            suggestionChoices = allDomains.filter(function (domain) { return domain.online_users; });
```
davidkelly: Is domain.online_users a boolean?  If not, maybe we should be explicit about > 0?  Super minor thing, if so.

cozza13:
I have gone through and tested this and it works.

interface/resources/qml/AddressBarDialog.qml
howard-stearns: It's a number. It's not supposed to be null or undefined, but in principle it could be.
In javascript, of course, 0, null, undefined and "" are all falsey.

howard-stearns:
With QA and dev approval, I'm merging.

PR #8256, opened 2016-07-15
Fix decoration inflation

jherico:
So, we can't use a mousearea for the whole window that uses hover events, or nothing _in_ the window can receive hover events.  However, we want the decorations to be inflated as long as the mouse is inside the outermost window.  

This PR changes the behavior on the exit detection to check if the global cursor is still inside the bounds of the decorations rectangle, and only deflates the window if it is not.  

In theory, we should always get an exit notification of the mouse leaving the window because it will have to cross the decorations again... however, I'm concerned that because windows can overlap and obfuscate one another, a window could get expanded and be left in that state if you moved onto another window that overlapped it before moving out of the first window's rectangle.

## Testing

When mousing over a window in the HMD using hand controllers, the decorations should stay large as long as you are anywhere within the bounds of the decorations, including over the pin and close controls.

### DIFF
--- a/interface/resources/qml/windows/Decoration.qml
+++ b/interface/resources/qml/windows/Decoration.qml
@@ -48,7 +48,23 @@ Rectangle {
         drag.target: window
         hoverEnabled: true
         onEntered: window.mouseEntered();
-        onExited: window.mouseExited();
+        onExited: {
+            if (!containsMouseGlobal()) {
+                window.mouseExited();
+            }
+        }
+        
+        function containsMouseGlobal() {
+            var reticlePos = Reticle.position;
+            var globalPosition = decorationMouseArea.mapToItem(desktop, 0, 0);
+            var localPosition = {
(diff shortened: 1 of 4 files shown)


jherico:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;mv5Pc0GV&#x2F;84-fix-window-decoration-expansion-without-breaking-mouse-hover-again"> Fix window decoration expansion without breaking mouse hover again</a>

hifi-gustavo:
Build console output for commit: 317dee1b235c12ddbdb316d3c36ef229e456d4a7 available at: https://gist.github.com/2aea61fa5ea5c9fdb01c82f1ee5df4bd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8256-317dee1.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8256-317dee1.dmg

PR #8257, opened 2016-07-16
Fix hover states for toolbar buttons

jherico:
The recent modifications to the edit toolbar required a change to the core toolbar API so that hovering logic is done by the caller rather than baked into the toolbar QML.

This PR updates all the existing system toolbar buttons to use the new mechanism to set hover and unhover states.

## Testing

With this build, if you mouse over one of the system toolbar buttons it should highlight in response to your mouse and return to it's original state when you move the mouse away.  In the current dev-master build, the buttons will not respond visually to mouse-over.

### DIFF
--- a/interface/resources/qml/hifi/toolbars/ToolbarButton.qml
+++ b/interface/resources/qml/hifi/toolbars/ToolbarButton.qml
@@ -50,12 +50,12 @@ Item {
         anchors.fill: parent
         onClicked: asyncClickSender.start();
         onEntered: {
-            if (hoverState > 0) {
+            if (hoverState >= 0) {
                 buttonState = hoverState;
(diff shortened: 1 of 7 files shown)


hifi-gustavo:
Build console output for commit: 9ebc0c28c846b309716b8313e4e9bff9f5cbddb8 available at: https://gist.github.com/0354ef80417a10cfe0b6481cac2e66f1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8257-9ebc0c2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8257-9ebc0c2.dmg

PR #8266, opened 2016-07-18
Actually delete textures we're not using

jherico:
Looks like we missed the actual GL deletion call in the destructor for textures... we only ever deleted textures that got used as a downsample source.  This should fix it.

### DIFF
--- a/libraries/gpu-gl/src/gpu/gl/GLTexture.cpp
+++ b/libraries/gpu-gl/src/gpu/gl/GLTexture.cpp
@@ -187,7 +187,11 @@ GLTexture::~GLTexture() {
         }
     }
 
-    Backend::decrementTextureGPUCount();
+    if (_id) {
+        glDeleteTextures(1, &_id);
+        const_cast<GLuint&>(_id) = 0;
+        Backend::decrementTextureGPUCount();
+    }
     Backend::updateTextureGPUMemoryUsage(_size, 0);
     Backend::updateTextureGPUVirtualMemoryUsage(_virtualSize, 0);
 }


jherico:
## Testing

With this build, the GPU memory consumed by the application as reported in external applications like GPU-Z and Process Explorer should bear a much closer relationship to the memory usage reported internally from such tools as renderStats.js.

Note the external tools will still typically report higher memory usage, but the relationship between the internal and external numbers should follow a mostly linear relationship.  

Additionally, when you exit a content filled domain for an empty domain, or simply disconnect from any domain (you can easily do this by typing the address 192.168.0.0 in the location bar), then you should see a sharp drop in GPU memory as reported by external tools, down to a relatively small value (on the order of a few hundred MB).  In production, the memory will not drop significantly on exiting a domain.

hifi-gustavo:
Build console output for commit: 0f240d39b61905313bd76127752d393cf65b6ff4 available at: https://gist.github.com/69e69381b8edab4f4c967efabc7b5c18<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8266-0f240d3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8266-0f240d3.dmg

jherico:
test this please

OmegaHeron:
I'll be running standard tests and data-log GPU mem stats as I did before.  Looks like Windows build is good, OSX and Linux failed.

hifi-gustavo:
Build console output for commit: 0f240d39b61905313bd76127752d393cf65b6ff4 available at: https://gist.github.com/69e69381b8edab4f4c967efabc7b5c18<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8266-0f240d3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8266-0f240d3.dmg

jherico:
@OmegaHeron note, my comment about the texture memory dropping back down to hundreds of MB doesn't seem to apply...  it looks like the consumption will typically stay high due to the ram caches.  However, repeatedly entering and leaving a domain should no longer cause the GPU memory to climb uncontrollably.  

Testing just now, I was going back and forth between dev-playa, dev-demo and 192.168.0.0 and the peak GPU memory usage for each was relatively stable.

OmegaHeron:
Yeah - I guess it would be good to compare if we see the log messages for swapping textures once it looks apparently high?  I'm still working through it as my first thought was it fails, but, I think it's more subtle when considering GPU may be caching textures at hardware/driver level keeping use figure high.  I'll play around a bit more with - I haven't yet pushed it into nosedive of stuttering like before, but the day is still young. :)

jherico:
@OmegaHeron you might be able to simplify testing by going into `Developer / Network / Ram Cache Sizes` and setting all the values to 0.  It doesn't actually seem to significantly affect my load speed, but it does get more stable values when switching locations.

OmegaHeron:
Windows 10 64 bit - i5 6500 CPU - GTX970 GPU

All RAM Caches set to zero as suggested.

I still think there's something amiss.  I'm including an annotated file from GPU-Z data logging.  A bugsplat event occurred from an (apparent) deadlock toward end.  I still, as far as interface is concerned, push this into aggressive texture swapping simply be going to home scene - then disconnected space and back to home scene repeatedly.  Not going to any other domains... just empty space/home scene.  No other AVs, no differing content.

I concatenated the rolled log files for this session into one file and added notes to GPU-Z data log as well.  I still wonder if the real culprit to a lot of this deadlocking points to the logging blocking actual useful things from happening.  I'd love to see a build with absolutely zero logging (and not just by directing it to dev/null) - totally not calling any write a string about whatever to wherever.

It's a lot to look at, but, if you run through the GPU-Z log and correlate to what's happening in Interface log it's an interesting read.

[GPU-Z-PR8266-run2.txt](https://github.com/highfidelity/hifi/files/370082/GPU-Z-PR8266-run2.txt)

[hifi-log.txt](https://github.com/highfidelity/hifi/files/370083/hifi-log.txt)

jherico:
I'm showing a very reliable transition between the max memory state and the min memory state as I move between Playa and a disconnected state:

image

It's possible there's something else inside the default home content that is causing an unrelated leak.  I'm trying to get it set up locally to test.

OmegaHeron:
I'd look at physics in respect to dynamic grabable objects especially.  It seems a very different beast as you begin removing things like that.

jherico:
I've verified that there is _something_ in the home content that is still leaking massive amounts of memory...  

https://files.slack.com/files-tmb/T025Q3X6R-F1SRW675E-b25f0daa7c/pasted_image_at_2016_07_18_02_40_pm_720.png

OmegaHeron:
And just to be clear - this isn't new.  Standard home scene has been this way for ages, I guess I never managed to raise the right flags in correct sequence to get proper attention.

jherico:
@OmegaHeron can you do me a favor.  In the home content there is a web entity.  Either hide or delete that entity and try your test again?

OmegaHeron:
Will do - be a few, collecting stats for another PR at the moment.

OmegaHeron:
@jherico - without web entity - preliminary test shows 2305MB used (via GPU-Z) - leave for disco space it drops to 1032MB - return 2315 - back to disco 1042.  Did about 5 more cycles and it's stayed very near to the 2315 at home scene, 1030 in disco space.  I'm satisfied web entity is culprit.

OmegaHeron:
It's so much better without web entity influence.  I can actually go places and return now without needing to restart.  Lucky none of other two places I have to go for QA rounds have web entities.  Now numbers GPU-Z gives match my general experience watching it with others games and SL/OS.

jherico:
It appears to be slightly more complicated than just the presence of web entities.  If I remove everything from a domain but the web entity, I don't leak memory.  If I remove only the web entity, I don't leak memory.  

I've narrowed it down to a small scene culled from the home content with the 6 avatar dolls and the web entity.  I suspect it's not specific to avatars, but rather to any entity with texture data, and that somehow the web entity is locking some set of textures loaded around the same time as it is loaded, and that avatars are showing the problem more clearly for some reason (possibly because they tend to have a lot of fairly high res textures).

OmegaHeron:
Needs the PSOD label.  Perfect Storm of Dysfunction.  I don't envy having to trace that one down, but, will be looking forward to testing any available fix when needed.

jherico:
@OmegaHeron [here](https://github.com/highfidelity/hifi/pull/8275).  This doesn't fix the actual leak, but it should stabilize the amount of leaked memory based on the number of web entities that you ever render concurrently.

PR #8276, opened 2016-07-19
Fix issues with mismatched codecs

ZappoMan:


ZappoMan:
**Note:** There might be some cases where the server has changed codecs and the client is not using the correct codec, which still result in the client being confused, but it won't play razor blades - instead it will be silent.

**Test Plan:**
- Note: this has a protocol change, so you have to test against this stack manager.
- Test audio... 
- Try to run a couple instances of this stack manager and switch between them.
- Try changing the codec preference list in the domain server and make sure audio still works

hifi-gustavo:
Build console output for commit: c484fec51d3e2137132427a484db41f107db491c available at: https://gist.github.com/10a304d5e40a932f6111a1b827213b4a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8276-c484fec.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8276-c484fec.dmg

cozza13:
ok to test

hifi-gustavo:
Build console output for commit: c484fec51d3e2137132427a484db41f107db491c available at: https://gist.github.com/10a304d5e40a932f6111a1b827213b4a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8276-c484fec.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8276-c484fec.dmg

hifi-gustavo:
Build console output for commit: 1ee5023f6df63d816e4b8ec29c8530a91b7318ad available at: https://gist.github.com/10a304d5e40a932f6111a1b827213b4a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8276-1ee5023.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8276-1ee5023.dmg

OmegaHeron:
QA Pass
Testing Windows 10 64 bit for Interface
Ubuntu 16.04 for Stack
DS Preference order = `hifiAC,zlib,pcm`

Here's what I did to torture it.  I had interface (on windows) connected to this PR stack (on linux) - I had mic open with echo server audio enabled - initially preference in DS was set to `hifiAC,zlib,pcm` - then I changed to `zlib,hifiAC,pcm` and restarted stack then changed to `pcm,zlib,hifiAC,` and Interface successfully detected change and negotiated stream properly.

I noticed this in log `Jul 19 16:11:32 hf-audio: [07/19 16:14:31] [DEBUG] [653] [audio-mixer] Codec mismatch: expected "pcm" got "zlib" writing silence` as the switchover took place and that seems exactly what it should be.

I've not managed to make any ear piercing buzzsaws yet, but, it's a fairly random/difficult thing to reproduce.

Bandwidth stats for each codec match expectation.  A funny - zlib beats 4:1 when there's no audio being injected, but, there's logic to that.  4:1 is superior for BW when actual audio is being sent.  It seems a good fix to me - I swapped between a domain on LAN and another remote both running this stack on Linux and it behaved well.

There's no way to be 100% sure this totally solves corrupted audio issue, but, it certainly solves one aspect of it (and hopefully that **is** the only aspect of it).  I didn't have good results using latest DEV Interface/stack and swapping codecs on the fly - so there's definite > good here.

OmegaHeron:
**crap** it just went into buzzsaw ear piercer with this PR Stack/interface!  Tracing now while it's doing it.

ZappoMan:
@OmegaHeron can you get me your interface and ac logs for when the buzzsaw happened??? That would help me.

OmegaHeron:
It's totally freaking out, there's no audio injector or anything sound producing in domain - mic is muted - no indication of anything incorrect in Interface log - I just see the ticking of
'[07/19 12:27:45] [DEBUG] Logging activity "stats"'

 every 5 seconds.  There's nothing being written to log from AC running audio.

AC log last entries are;
`Jul 19 16:11:32 hf-audio: [07/19 16:24:52] [DEBUG] [653] [audio-mixer] Selecting codec: "hifiAC"`
`Jul 19 16:11:32 hf-audio: [07/19 16:24:52] [DEBUG] [653] [audio-mixer] UNEXPECTED -- didn't have node linked data in  handleNegotiateAudioFormat`
`Jul 19 16:11:32 hf-audio: [07/19 16:24:52] [DEBUG] [653] [audio-mixer] selectedCodecName: "hifiAC"`
`Jul 19 16:11:32 hf-audio: [07/19 16:24:52] [DEBUG] [653] [audio-mixer] creating new AvatarAudioStream... codec: "hifiAC"`
`Jul 19 16:11:32 hf-audio: [07/19 16:24:53] [SUPPRESS] [653] [audio-mixer] 36 repeated log entries matching "Packet of type \d+ \([\sa-zA-Z:]+\) received from unknown node with UUID" - Last entry: "Packet of type 3 (Ping) received from unknown node with UUID 178c230d-eb43-46f5-a5e1-1f6d7d9c7976"`

None of that coincides with noise - It began about 15 minutes after last entry.  Note last AC log print was at 16:11:32 and noise began at 16:27:45 (adjusted for TZ change between Interface machine and server).

OmegaHeron:
I tried toggling mic while it was in that state and got a crash to desktop with windows dialog - no bugsplat.

ZappoMan:
@OmegaHeron is there a server I can test with? domain name?

OmegaHeron:
I found a possible cause - my server machine renegotiated its DHCP lease at approx time of failure - this leads to an approximate 500ms network down event - that may have been cause.  I've not been able to reproduce issue again despite my best efforts to.

OmegaHeron:
I've been trying for last (near) hour to reproduce issue - I've reproduced network issue on server leading to brief net outage - no problem.  I've introduced packet loss at ranges of 5-25% and no problem.  Others need to test this and stay on it for longer than a brief visit.  Maybe a cosmic ray hit something.  :)

libraries/networking/src/udt/BasePacket.cpp
```diff
@@ -154,8 +154,7 @@ qint64 BasePacket::writeString(const QString& string) {
     QByteArray data = string.toUtf8();
     uint32_t length = data.length();
     writePrimitive(length);
-    writeData(data.constData(), data.length());
-    seek(pos() + length);
```
jherico: can you explain this change?

ZappoMan: There's a couple things going on... 
1) writeData() - is actually supposed to be an internal function called back only by the Qt QIODevice when a write() call is made.
2) the seek() was code I added last week in an incorrect attempt to make the writeString() support writing multiple strings in a row, but that isn't needed if we correctly call write() instead of erroneously calling writeData()
3) this makes the writeString() match the code in writePrimitive() in it's use of the QIODevice

PR #8277, opened 2016-07-19
Ensure the pause overlay is unlit (emissive)

jherico:
## Testing

While in the HMD and running the `away.js` script, press escape or alt-tab from Interface in order to trigger the paused image.  Note that the image is fully lit (bright white text) and is unaffected by scene lighting.  In production builds it will be affected by scene lighting.


jherico:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ELaic2gY&#x2F;66-the-pause-screen-needs-to-be-unlit"> The Pause screen needs to be unlit</a>

PR #8287, opened 2016-07-20
fix bug breaking simple-compound shape type

AndrewMeadows:


libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -741,6 +741,9 @@ void RenderableModelEntityItem::computeShapeInfo(ShapeInfo& info) {
             pointCollection.resize(1);
         }
 
+        ShapeInfo::TriangleIndices& triangleIndices = info.getTriangleIndices();
+        triangleIndices.clear();
+
```
AndrewMeadows: This is the actual fix: grab and clear the `triangleIndices` up here rather than doing it for each mesh down below.

The problem was that for the second mesh the indices for the first would be completely overwritten.  Eventually the interface could crash when trying to build the shape inside `ShapeFactory` implementation.

AndrewMeadows:
To test:

(1) Create a ModelEntityItem using this URL: http://hifi-content.s3.amazonaws.com/ozan/dev/props/123D/star.fbx

(2) Set its shape type property to "Hull Per Submesh" using edit.js
--> this would crash on debug builds, might not crash for release but the report I heard was that the object would just not collide at all
--> in this PR the object should collide with the correct shape (set the object dynamic with gravity to see it tumble)

(3) Make sure objects that have "Static Mesh" shape still collide correctly.

hifi-gustavo:
Build console output for commit: 3fdfc0db141b6d465f6693088be841e632055a72 available at: https://gist.github.com/cbba86c914b1e056602681c2dca0ff88<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8287-3fdfc0d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8287-3fdfc0d.dmg

hifi-gustavo:
Build console output for commit: 20eebb90f88847718693c1bca9c242563893b4d5 available at: https://gist.github.com/cbba86c914b1e056602681c2dca0ff88<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8287-20eebb9.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8287-20eebb9.dmg

OmegaHeron:
QA Pass

Tested Windows 10 64 bit - i5 6500 CPU - GTX970 GPU

(2) Set its shape type property to "Hull Per Submesh" using edit.js - OK

--> in this PR the object should collide with the correct shape (set the object dynamic with gravity to see it tumble) - Yes giving some G on each axis then bumping shows actions I'd generally expect.

(3) Make sure objects that have "Static Mesh" shape still collide correctly. - Yes all I have set static still work as expected.

No negative changes to any typical performance stat seen.  :+1:

PR #8300, opened 2016-07-22
Actually audio was really broken

davidkelly:
Seems there were 2 issues:
* If you couldn't inject locally (say, no audio interface
  because you are an assignment client), we would not do the
  network injection at all.  That's bad.
* When we don't have a codec, like when using an injector, we
  were serializing the empty string in some problematic way.  I
  just use the built-in serialization, and it seems fine.

I tested this by first playing a sound in the console (or maybe you
want to tip some cows, that is ok too).  Then in _another_ interface
client on another machine in the same domain, I better hear the
sound.

Then, I added a script to play audio as a persistent script, which
just loops forever and ever.  You should hear that on both interface
clients also.

A detailed test plan to follow.

### DIFF
--- a/libraries/audio/src/AudioInjector.cpp
+++ b/libraries/audio/src/AudioInjector.cpp
@@ -203,8 +203,8 @@ bool AudioInjector::injectLocally() {
     }
 
     if (!success) {
-        // we never started so we are finished, call our stop method
-        stop();
+        // we never started so we are finished with local injection
+        finishLocalInjection();
     }
 
     return success;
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 615a838015e2a2fde3857d86bfee7863f98260bb available at: https://gist.github.com/2356af275b3e61f49172236f2311ba9a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8300-615a838.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8300-615a838.dmg

hifi-gustavo:
Build console output for commit: 4b48cc7c2e3e87c9d0918cfb53f32ffc18064110 available at: https://gist.github.com/2356af275b3e61f49172236f2311ba9a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8300-4b48cc7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8300-4b48cc7.dmg

PR #8211, opened 2016-07-08
Skin sub-surface lighting

ZappoMan:
Check performance

Use this Avatar: 
https://hifi-public.s3.amazonaws.com/sam/models/skinRenderingTest/matthew/matthew.fst

https://hifi-public.s3.amazonaws.com/sam/models/skinRenderingTest/priscilla/priscilla.fst

Make sure you do not see any graphics issues that are different to the curent beta build

Docs to create you own.  https://docs.google.com/document/d/1fYXVkCezYbMguRl1J1q244pELFqw_sf9fbuu9-C7pqg/edit?usp=sharing

PR #8301, opened 2016-07-23
Asset browser ui

AlexanderOtavka:
Add dynamic and collision shape selection to add to world flow from asset browser.

### Testing
1. Open asset browser
2. Select an fbx file and click add to world
3. A dialog identical to the add model dialog should appear
4. Try different configurations and different models
5. Shape type, dynamic state, and model URL should all be reflected in the created entities
6. When you add to world an object that has dynamic true, the object should have a gravity of -10.

### DIFF
--- a/libraries/entities/src/EntityScriptingInterface.h
+++ b/libraries/entities/src/EntityScriptingInterface.h
@@ -86,7 +86,8 @@ public slots:
     Q_INVOKABLE QUuid addEntity(const EntityItemProperties& properties, bool clientOnly = false);
 
     /// temporary method until addEntity can be used from QJSEngine
-    Q_INVOKABLE QUuid addModelEntity(const QString& name, const QString& modelUrl, const glm::vec3& position);
+    Q_INVOKABLE QUuid addModelEntity(const QString& name, const QString& modelUrl, const QString& shapeType, bool dynamic,
+                                     const glm::vec3& position, const glm::vec3& gravity);
 
     /// gets the current model properties for a specific model
     /// this function will not find return results in script engine contexts which don't have access to models
--- a/interface/resources/qml/AssetServer.qml
+++ b/interface/resources/qml/AssetServer.qml
@@ -213,7 +213,7 @@ ScrollingWindow {
                         // different scripting engine from QTScript.
                         gravity = Vec3.multiply(Vec3.fromPolar(Math.PI / 2, 0), 10);
                     } else {
-                        gravity = { x: 0, y: 0, z: 0 };
+                        gravity = Vec3.multiply(Vec3.fromPolar(Math.PI / 2, 0), 0);
                     }
 
                     print("Asset browser - adding asset " + url + " (" + name + ") to world.");
--- a/interface/resources/qml/AssetServer.qml
+++ b/interface/resources/qml/AssetServer.qml
@@ -207,7 +207,14 @@ ScrollingWindow {
                 } else if (url) {
                     var name = assetProxyModel.data(treeView.selection.currentIndex);
                     var addPosition = Vec3.sum(MyAvatar.position, Vec3.multiply(2, Quat.getFront(MyAvatar.orientation)));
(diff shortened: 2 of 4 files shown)


hifi-gustavo:
Build console output for commit: c25b87c33ff213a915b73eceba30812ed4c6ea17 available at: https://gist.github.com/925dd70fab4904193be1f510138f5828<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8301-c25b87c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8301-c25b87c.dmg

RyanDowne:
This is looking good but I am not seeing the -10 gravity when dynamic is selected.

hifi-gustavo:
Build console output for commit: 344ec25bcd3ede709e5edc0553e01769bad75f32 available at: https://gist.github.com/925dd70fab4904193be1f510138f5828<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8301-344ec25.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8301-344ec25.dmg

hifi-gustavo:
Build console output for commit: 90154158b6eb1a5099fae6ece1b0b3cd5dbbc2f6 available at: https://gist.github.com/925dd70fab4904193be1f510138f5828<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8301-9015415.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8301-9015415.dmg

Nex-Pro:
**PR Build:** #8211
**CPU:** 4.4 ghz Intel i7
**RAM:** 16 GB
**GPU:** Nvidia gtx770 2gb
**Testing OS:** win 10 pro
**HMD:** None
**Hand Controllers:** None

### QA PASS

Works fine.  Models with 'dynamic' set acquire a gravity of -10 Y.

Selecting different hull types results in the appropriate collisions. For example the firepit when simple has a smooth top. Otherwise avatars collide with the rocks in the fire.

***Only issue seen was some objects not acting dynamically unless collided with. But I don't think that relates to this PR.***

libraries/entities/src/EntityScriptingInterface.h
```diff
@@ -86,7 +86,8 @@ public slots:
     Q_INVOKABLE QUuid addEntity(const EntityItemProperties& properties, bool clientOnly = false);
 
     /// temporary method until addEntity can be used from QJSEngine
-    Q_INVOKABLE QUuid addModelEntity(const QString& name, const QString& modelUrl, const glm::vec3& position);
+    Q_INVOKABLE QUuid addModelEntity(const QString& name, const QString& modelUrl, const QString& shapeType, bool dynamic,
+                                     const glm::vec3& position, const glm::vec3& gravity);
```
AndrewMeadows: Normally I would object to the `EntityScriptingInterface::addModelEntity()` method: it is not general enough and is therefore less useful as general `EntityScriptingInterface` API -- what happens when I want to add a `ModelEntityItem` with a specific: rotation, velocity, dimensions, or any of a dozen other properties?

However, I note that the comment above says that this is a "temporary method until addEntity can be used from QJSEngine", and that it is only ever called in one spot so far.  If we want to continue to punt this technical debt down the road then it doesn't block this PR.  I'll ask around to see what others want regarding this issue.

AndrewMeadows:
The problem where dynamic objects are "inactive" until some other dynamic object moves near them (thereby waking them up in the physics simulation) is fundamental to how we load objects in the world and is not new to this PR.

libraries/entities/src/EntityScriptingInterface.h
AndrewMeadows: We're going to kick this forward, but hopefully for only a week or two.  Zander will try to use a QVariant.

PR #8307, opened 2016-07-23
Disable glow lines on Mac

jherico:
Our geometry shader use is apparently incompatible with OSX, so until we find a better solution for rendering glow lines, we should disable them.

### DIFF
--- a/libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
+++ b/libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
@@ -113,10 +113,12 @@ void HmdDisplayPlugin::customizeContext() {
 
     updateReprojectionProgram();
     updateOverlayProgram();
+#ifdef HMD_HAND_LASER_SUPPORT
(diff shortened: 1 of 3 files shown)


jherico:
## Testing

* Switch to the Oculus display plugin on a system running OSX....  
 * On production it will immediately crash.  In this build it should not.  

* Run this script on OSX:  http://s3.amazonaws.com/DreamingContent/scripts/tests/overlays.js
 * On production it will immediately crash.  In this build it should not (however, the glow line that is visible on windows should be rendered as a normal line instead.

hifi-gustavo:
Build console output for commit: 85b0440e32acbfc9d995678de6a51554c7399e5f available at: https://gist.github.com/6f80da8a9ef212c86bef188f91a397d9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8307-85b0440.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8307-85b0440.dmg

PR #8319, opened 2016-07-26
better allocation of CPU time while loading scenes and when slow scripts are running

ZappoMan:
This PR makes two changes to improve the CPU load and loading performance of scenes, as well as better handle scenes with complex scripts.

1) Scripts won't get "update" signals while the scene is still loading. Specifically during the time that physics is disabled during the scene load time. This allows the other CPU related tasks (like mesh parsing, or texture conversion) to get more access to the CPU. Scripts will still run, they just won't get updates sent to them.

2) Scripts that are taking a very long time in their update methods will get punished by being forced to sleep longer. This basically prevents poorly behaving scripts from hogging too much CPU time.

### DIFF
--- a/libraries/script-engine/src/ScriptEngine.cpp
+++ b/libraries/script-engine/src/ScriptEngine.cpp
@@ -894,7 +894,7 @@ void ScriptEngine::run() {
         qint64 now = usecTimestampNow();
 
         // we check for 'now' in the past in case people set their clock back
-        if (_isPhysicsEnabledFunc() && _lastUpdate < now) {
+        if (_emitScriptUpdates() && _lastUpdate < now) {
             float deltaTime = (float) (now - _lastUpdate) / (float) USECS_PER_SECOND;
             if (!_isFinished) {
                 auto preUpdate = clock::now();
--- a/libraries/script-engine/src/ScriptEngine.cpp
+++ b/libraries/script-engine/src/ScriptEngine.cpp
@@ -828,42 +828,44 @@ void ScriptEngine::run() {
 
     _lastUpdate = usecTimestampNow();
 
-    qint64 totalSleepFor = 0;
     std::chrono::microseconds totalUpdates;
-    auto lastLoopStart = clock::now();
 
     // TODO: Integrate this with signals/slots instead of reimplementing throttling for ScriptEngine
     while (!_isFinished) {
-        auto thisLoopStart = clock::now();
+        auto beforeSleep = clock::now();
 
         // Throttle to SCRIPT_FPS
+        // We'd like to try to keep the script at a solid SCRIPT_FPS update rate. And so we will 
+        // calculate a sleepUntil to be the time from our start time until the original target
+        // sleepUntil for this frame.
         const std::chrono::microseconds FRAME_DURATION(USECS_PER_SECOND / SCRIPT_FPS + 1);
-        const std::chrono::microseconds MINIMUM_SLEEP { FRAME_DURATION / 2 };
+        clock::time_point sleepUntil(startTime + thisFrame++ * FRAME_DURATION);
 
-        auto beforeSleep = clock::now();
-        clock::time_point sleepTime(startTime + thisFrame++ * FRAME_DURATION);
-        auto wouldSleep = (sleepTime - clock::now());
-        auto avgUpdates = totalUpdates / thisFrame;
+        // However, if our sleepUntil is not at least our average update time into the future
+        // it means our script is taking too long in it's updates, and we want to punish the
+        // script a little bit. So we will force the sleepUntil to be at least our averageUpdate
+        // time into the future.
+        auto wouldSleep = (sleepUntil - clock::now());
+        auto avgerageUpdate = totalUpdates / thisFrame;
 
-        if (wouldSleep < avgUpdates) {
-            sleepTime = beforeSleep + avgUpdates;
+        if (wouldSleep < avgerageUpdate) {
+            sleepUntil = beforeSleep + avgerageUpdate;
         }
 
-        std::this_thread::sleep_until(sleepTime);
+        std::this_thread::sleep_until(sleepUntil);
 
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: 83dc9ea6bba4391b2642d9dba405f1afe8e0697f available at: https://gist.github.com/092b96a060d2474cd8d9967c4c12a6fa<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8319-83dc9ea.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8319-83dc9ea.dmg

OmegaHeron:
Preliminary result = very promising.  I want to spend a few hours running this, trying different things and make sure I didn't get 1 or 2 good runs then not so good.

Early tests show I can run Interface and Stack on same i5 machine, load standard home and actually have it work - when Interface does enter CPU saturation it's recovering from it quicker and not losing connection to AC/DS processes.  Overall CPU use is lower by about 10-15%

I'll have more comments and numbers later.

OmegaHeron:
Using this with DK2 makes it even more apparent how much this improves "i5 experience" - for the CA people... it's a lot more like middle of nowhere I-5 vs Los Angeles gridlock I-5.

I've never been able to use DK2 and directly hop around domains, in past I had to be in Desktop mode - connect to domain of interest - wait for it to become stable - then, if it did, switch to HMD or, if not, restart Interface and try again.

I just successfully went from localhost in all its bloated glory to dev-play to dev-demo to my 2 domains - back to localhost - and repeated all in one session, 100% in HMD mode.  Until Interface needed to begin transferring textures from GPU->CPU RAM it was as smooth an experience as I've ever had in HF.

hifi-gustavo:
Build console output for commit: a12034cb45255d19fc1b880a2e8e4beaff8e7b9e available at: https://gist.github.com/092b96a060d2474cd8d9967c4c12a6fa<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8319-a12034c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8319-a12034c.dmg

hifi-gustavo:
Build console output for commit: 2d199fe3d007f3a6fb7f9c1099aac08c583e416d available at: https://gist.github.com/092b96a060d2474cd8d9967c4c12a6fa<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8319-2d199fe.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8319-2d199fe.dmg

OmegaHeron:
1) Do things "load" faster?

Yes and no.  Compared to RC/DEV current - when it works, individual pieces of scene don't load any slower/faster per piece.  The right question, I think, is does a scene like standard home load at all, and if so, is it faster.

For this PR that question is a resounding Yes - since it's no longer resetting connection multiple times forcing a restart of load loop - it finishes scene load consistently and is, as such, faster.

2) Is CPU usage better after loading?  Yes.

Now CPU use tends to track what one would think of as complexity of a scene.  I.E. Does it have a lot of content, lot of physics, lot of dynamics, lot of grab enabled items and etc.  Standard home shows around 35-40% CPU load with AV at rest once scene is fully loaded - moving upward to 40-45% as you walk around   Depending on where you walk and come to rest it can idle down to around 30% or a touch less.  Moving to a less complex space shows 15-16% CPu use AV at rest and 23-24% while walking.

3) Is FPS better while loading? After loading?

While, yes, most assuredly - unless there's a lot of active texture swapping then it can get a bit more like typical experience in RC/DEV with lots of FPS drops.  After loading.  No.  I've not had any FPS issues in standard home once I got it to load, ever, it yields around 120-150FPS on my GTX970 - and does same with this build.

4) Do you see a general improvement in loading: audio not breaking up as much, disconnections from domain less likely?

Yes.  Audio no longer breaks up at all, not one disco during repeated enter/exit cycles to standard home on localhost - that alone makes a huge difference as it doesn't enter death spiral infinite loop of trying to load, fail, load.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -828,24 +828,44 @@ void ScriptEngine::run() {
...
-        clock::time_point sleepTime(startTime + thisFrame++ * FRAME_DURATION);
-        std::this_thread::sleep_until(sleepTime);
+        clock::time_point sleepUntil(startTime + thisFrame++ * FRAME_DURATION);
+
+        // However, if our sleepUntil is not at least our average update time into the future
+        // it means our script is taking too long in it's updates, and we want to punish the
+        // script a little bit. So we will force the sleepUntil to be at least our averageUpdate
+        // time into the future.
+        auto wouldSleep = (sleepUntil - clock::now());
+        auto avgerageUpdate = totalUpdates / thisFrame;
```
huffman: If I understand correctly, we are keeping an average over the lifetime of the ScriptEngine. Thoughts on using a rolling average instead? My worry is that the last update duration will eventually become negligible as the ScriptEngine runs longer, and so will become nonreactive to changes in the update duration.

ZappoMan: in my testing it seems that poorly behaving scripts tend to behave consistently poorly... I guess it's possible that a script would be well behaving for a long time, and then start behaving poorly - but I haven't seen that in practice. Maybe we should get this out as is - and then do more extensive testing to see how scirpt behave.

In the scenario you describe - the method implemented here won't behave any worse than the current production code.

huffman: For entity scripts I think we can expect the characteristics of the ScriptEngine to change a fair amount over time because the engine itself is shared.

I agree, it won't be any worse, so getting it out sooner would be a priority.

PR #8323, opened 2016-07-27
Initial rough cut of selfies

davidkelly:
New camera mode, for the selfie, initial work to make both a selfie
and regular snapshot when you hit a button.

Lots to do still, but this basically works well enough to get us started


hifi-gustavo:
Build console output for commit: 83da75db52d6c74ac2205af6188f3c8b89660971 available at: https://gist.github.com/bd1f696413085016247d6b901fe1b3bf<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8323-83da75d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8323-83da75d.dmg

howard-stearns:
Please include a test plan, suitable for use by someone on #qatesters, or otherwise unfamiliar with the project.

hifi-gustavo:
Build console output for commit: f6aef1d079e12c51624c166dffbddad38c1d9d29 available at: https://gist.github.com/bd1f696413085016247d6b901fe1b3bf<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8323-f6aef1d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8323-f6aef1d.dmg

davidkelly:
# Test Plan

The intent here is to get some sort of ability to make selfies into interface.  For now, you just hit the snap button in the HUD.  There is a small delay and both a snapshot and a selfie are taken and saved to disk.  Speaking of delays, you will see the notification for both pics delayed until after the shots are taken (5 seconds).  That's temporary.

All we really want here is to be sure we are saving reasonable images for a first pass.  We will for sure make a much better UI/UX for this in the future.  Soon I'll send these off to the server to be stored, and then we will come back and add UI to selectively send them, etc...


### So, to test:
load `selfie.js`.  Note it will put a new button in the hud.  Hit that button, you will see the camera switch to selfie mode briefly.  The about 5 seconds later, you see the notifications for both the snapshot and selfie.  Should work the same (with different aspect ratio, etc...) in HMD and desktop.  They are saved in same folder as the snapshots.

hifi-gustavo:
Build console output for commit: c8667460960cad0fc3b204c9d896946559ce1118 available at: https://gist.github.com/bd1f696413085016247d6b901fe1b3bf<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8323-c866746.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8323-c866746.dmg

howard-stearns:
let's give it a more descriptive name?

howard-stearns:
I think we want this to be our best guess of what we want to deliver, so to that end:

- Include selfie.js in defaultScripts.js (probably in a way that puts the button ahead of edit?)

- I don't think we want this tied to the notification system:
   1) We will have a separate mechanism for the user to pick which snapshots to share, and we don't want to confuse the user with TWO indications of completeness.
  2) I don't think notifications work (or make sense?) in HMD, and I don't think we can come up with an explanation of how/why snapshots work differently between HMD and desktop that would make sense to users.

- I think we need a countdown display (both HMD and desktop), because I don't really "get" when the selfie is going to be taken. (If that's going to take a while and you want to get this merged in the meantime, please add an additional work-item card in trello so that we're aware of accumulating backlog.)

- I think we want the mirrored view in HMD (during selfie time), just like desktop:
  1.  As above, the difference feels gratuitous and wouldn't make sense to a user (in my opinion).
  2. I want to see what I'm going to get so I can pose right. That isn't any less true in HMD.
I don't know whether this is trivial to do or not, but I feel like it's important. (If we have to separate the work, make a card.)

howard-stearns:
The AvatarInputs tool and any other open windows are still visible in the snapshots. (Everything but the HUD button should go away if you Menu.setIsOptionChecked("Overlays", false).)

howard-stearns:
view -> selfie is confusing to me, and besides, it doesn't work right: utoggling the checkmark (using the menus) doesn't exit the mode, and it sticks that way in between sessions! 

Checkable menus are meant for "major modes" should be persisted in Settings. I don't feel that temporary modes (that are entered and exited by the system rather than the user) should be exposed to the user.

howard-stearns:
The plugin system generally uses inheritance to get the right behavior, and all HMD plugging inherit from HmdDisplayPlugin. Maybe we can make use of that rather than passing an hmdSelfie boolean around?

RyanDowne:
Part of getting this working fast was to determine if the front.back pictures being taken were capturing the environment sufficiently or were inline with the photo takers expectations. I've tested in four different domains, not giving myself too much load time and in all cases the pictures taken were all great!

I agree with @howard-stearns that we need to remove all UI elements or provide a toggle somewhere so people can decide. 

Selfie is not the right word for this I agree. I think snapshot.js or something would be a better name. 

Agree, much later, we can add things that will allow the user to know that the shot is about to be taken or a 'frame' that appears when the view changes. But, I think we could just remove the UI and start taking the next step to send these photos to the db. 

Nice work @davidkelly !!

ZappoMan:
Why are we adding a new camera mode?

davidkelly:
@ZappoMan - I added it so you could move around, get the shot you want before it takes the selfie.  But, if there is another way to go, I'm happy to do that.

ZappoMan:
@davidkelly - there is already support for independent camera mode, that allows you to move the camera completely "independent" from the avatar - it feels like the selfie feature would be best implemented using this independent camera mode with placement via  a script or some other UI for placement

davidkelly:
@ZappoMan - agreed - I see that now -- will do.

hifi-gustavo:

howard-stearns:
really gustavo?

howard-stearns:
add to whitelist

murillodigital:
test this please

hifi-gustavo:
Build console output for commit: c8667460960cad0fc3b204c9d896946559ce1118 available at: https://gist.github.com/bd1f696413085016247d6b901fe1b3bf<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8323-c866746.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8323-c866746.dmg

hifi-gustavo:
Build console output for commit: 8d64f72fa360f517af310b4ab912a4775b66115f available at: https://gist.github.com/bd1f696413085016247d6b901fe1b3bf<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8323-8d64f72.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8323-8d64f72.dmg

hifi-gustavo:
Build console output for commit: aaa65436f354e6f536a6b0a69d529e09ca9184a5 available at: https://gist.github.com/bd1f696413085016247d6b901fe1b3bf<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8323-aaa6543.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8323-aaa6543.dmg

PR #8328, opened 2016-07-28
Fix punished scripts stopping behavior and application shutdown while scripts being punished

ZappoMan:


ZappoMan:
Test Plan:
- Smoke test - make sure performance hasn't regressed
- Run a really bad script:
```
// an intentionally bad script to test the punish behavior

Script.update.connect(function(deltaTime){
    print("time since last update:" + deltaTime);
    var a = 1234567;
    var b = 89012345;
    for (var i = 0; i < 100000000; i++) {
        var c = a * b;
    }
});
```
- Make sure that while that really bad script is running, that you can still:
    - Close the app with Ctrl+Q and the red X button
    - Stop the really bad script from the running scripts dialog box

Notice with the current Dev-builds -- that exiting the program and closing the scripts will take a very long time.

hifi-gustavo:
Build console output for commit: 755989b8ffe1d3afe8683d552f5f89d47ca973d5 available at: https://gist.github.com/64b5b51b56039f2a1a4b7a20a8ca9cbe<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8328-755989b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8328-755989b.dmg

OmegaHeron:

Using DEV-5211 running bad script I can't say that I see either a huge delay in script termination or closure of Interface.  But - if I have default script set + bad script running and remove all scripts - bad script certainly takes longer to stop (1 second or so).  Same with closing with Ctrl Q or X.

Using this PR running bad script - behaves sames as ^^.  It looks like someone with hardware config that can repro issue below will need to test.  

"Notice with the current Dev-builds -- that exiting the program and closing the scripts will take a very long time."

I see no regression to performance original PR addressed - this PR continues to offer improved experience on my i5.

howard-stearns:
Running on my MacBook Pro (see ticket), this works as far as stopping an individual script from the running scripts dialog. However, the Remove All button doesn't work. 

repro: run this pr on an MBP, with default scripts. Wait for load. (Get about 10-11 fps.) Open Running Scripts and press Remove All button button. Nothing happens, even after several minutes. 

Clicking an individual script's 'X' button removes the script within a second. You can even run just one script -- say, dialTone.js -- and kill it individually. But running that same single script and using the Remove All button has no effect.

ZappoMan:
@howard-stearns - hmm - ok - I have another idea...

ZappoMan:
test this please

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -845,14 +845,29 @@ void ScriptEngine::run() {
         // it means our script is taking too long in it's updates, and we want to punish the
         // script a little bit. So we will force the sleepUntil to be at least our averageUpdate
         // time into the future.
-        auto wouldSleep = (sleepUntil - clock::now());
+        auto wouldSleep = (sleepUntil - beforeSleep);
         auto avgerageUpdate = totalUpdates / thisFrame;
```
zzmp: s/avgerage/average

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -845,14 +845,29 @@ void ScriptEngine::run() {
...
+        while (!_isFinished && clock::now() < sleepUntil) {
+            if (!firstTime) {
+                QCoreApplication::processEvents(); // before we sleep again, give events a chance to process
+            }
+            auto wouldSleepSlice = (sleepUntil - clock::now());
+            auto thisSleepUntil = sleepUntil;
+            if (wouldSleepSlice > FRAME_DURATION) {
+                thisSleepUntil = clock::now() + FRAME_DURATION;
+            }
+            std::this_thread::sleep_until(thisSleepUntil);
```
zzmp: You are sleeping for the `std::min(sleepUntil, clock::now() + FRAME_DURATION)` to avoid blocking for more than a frame at a time?

Can you replace this logic with `std::min(...)`? It seems easier to parse.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -845,14 +845,29 @@ void ScriptEngine::run() {
...
+            firstTime = false;
```
zzmp: This is related to the logic in L860-862 more than it is related to the sleep logic. Can you move it up to L863 to make it easier to follow?

zzmp:
Why do you not process events the first time? Why did you put process events before the sleep?

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: I will try - I'm not sure that chrono's support min()

ZappoMan:
@zzmp - later in the regular processing we do process events...

libraries/script-engine/src/ScriptEngine.cpp
zzmp: std::chrono::time_point implements operator< so it should just work.

libraries/script-engine/src/ScriptEngine.cpp
ZappoMan: fixing

ZappoMan:
@birarda @howard-stearns @zzmp - I believe I've addressed all your feedback:
- The logic should be much more clear (using `std::min()` and `std::max()`)
- The processEvents should allow shutdown even while in a "punish" script case
- The logic should now work even if there is no entity server and therefore physics will never get enabled

hifi-gustavo:
Build console output for commit: e879f22c952772658276153fb3bd8690affe31fe available at: https://gist.github.com/64b5b51b56039f2a1a4b7a20a8ca9cbe<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8328-e879f22.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8328-e879f22.dmg

sethalves:
I haven't done a lot of testing here, but I can say that the ambient-sound AC-script that eric wrote for playa (ACAudioSearchAndInject.js) fails in master and works with this PR.

OmegaHeron:

General smoke test = all pass.
Criteria for bad script test = all pass.
In ref to no entity server present.  As far as I can tell, yes - signals are passed.

PR #8329, opened 2016-07-28
Render collision geometry

AndrewMeadows:
* add CollisionRenderMeshCach class with unit tests
* cleanup: move collision GeometryResource knowledge out of Model class
* cleanup: lots of internal name changes (mostly in Model class) and some cruft removal
* render collision geometry for physics debugging purposes
* fallback to "simple-compound" when "compound" fails


AndrewMeadows:
To test:

(1) Create several model entities, each using a different FBX model:
EntityA = model using a valid "compoundShapeUrl" (need to find/creae an OBJ asset for this entity) and "shapeType" = "compound"
EntityB = model with "shapeType" = "box"
EntityC = model with "shapeType" = "sphere"
EntityD = model with "shapeType" = "simple-hull"
EntityE = model with "shapeType" = "simple-compound" (ideally this model should have multiple submeshes and/or multiple materials on different convex parts)
EntityF = model with "shapeType" = "static-mesh"
EntityG = model with "shapeType" = "none"

(2) Also create an EntityBox and and EntitySphere.

(3) Enable rendering of collision shapes:  Developer --> Physics --> Draw Collision Hulls
--> EntityA  through EntityE should change their rendered appearance showing exactly how they collide
--> EntityF through EntityG, and also the EntityBox and EntitySphere should NOT change their rendered appearance

(4) Disable rendering of collision shapes
--> EntityA through EntityE should recover their rendered appearance
-> EntityF through EntityG, and also the EntityBox and EntitySphere should NOT change their rendered appearance

(5) Re-enable rendering of collision shapes, logout with that debug option enabled, and then login again:
--> EntityA through EntityE should appear normal until they are given collision shapes at which point their collision shapes will be rendered

(6) While showing collision shapes change the shapeType of various entities
--> the collision shape changes should be visible as soon as their shapes actually change in the physics engine

(7) Collision geometry is only created when needed but it costs CPU cycles to generate it and this is done on the main thread.  What would happen if you went into a domain that had many collidable (but not using "static-mesh") models visible and enabled the debug option... would the main thread block long enough while it computes all of the geometry to cause the watchdog thread to throw an exception?  This should be tested.

Note: due to geometry sharing technology you'd need many _distinct_ models to make this stress-test valid -- many copies of the same mesh at the same size will create only one geometry instance, however distinct sizes of the same model will generate distinct geometry instances to render.  The more subshapes in the collision shape the higher the CPU cost of building the visible geometry.

libraries/entities-renderer/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 set(TARGET_NAME entities-renderer)
 AUTOSCRIBE_SHADER_LIB(gpu model procedural render render-utils)
 setup_hifi_library(Widgets Network Script)
-link_hifi_libraries(shared gpu procedural model model-networking script-engine render render-utils)
+link_hifi_libraries(shared gpu procedural model model-networking script-engine render render-utils physics)
```
AndrewMeadows: entities-renderer lib now depends on physics lib... I think maybe I needed this dependency in the middle of the project but that it is no longer needed.  I will test.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -509,43 +509,25 @@ ModelPointer EntityTreeRenderer::getModelForEntityItem(EntityItemPointer entityI
...
+ModelPointer EntityTreeRenderer::allocateModel(const QString& url) {
     ModelPointer model = nullptr;
 
     // Only create and delete models on the thread that owns the EntityTreeRenderer
     if (QThread::currentThread() != thread()) {
         QMetaObject::invokeMethod(this, "allocateModel", Qt::BlockingQueuedConnection,
                 Q_RETURN_ARG(ModelPointer, model),
-                Q_ARG(const QString&, url),
-                Q_ARG(const QString&, collisionUrl));
+                Q_ARG(const QString&, url));
```
AndrewMeadows: The collisionUrl is no longer needed by the Model class.  Instead the RenderableModelEntityItem class keeps the GeometryResource for the compoundShapeUrl (if any) and is responsible for giving the Model a collisionGeometry, or taking it away, as necessary.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -28,6 +29,9 @@
 #include "RenderableModelEntityItem.h"
 #include "RenderableEntityItem.h"
 
+static CollisionRenderMeshCache collisionMeshCache;
```
AndrewMeadows: The CollisionRenderMeshCache is new.  It isn't actually a singleton, per se, but you would only ever need one.  In our case we only need it in the RenderableModelEntityItem class context, so it is a static variable in this implementation.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -428,19 +432,35 @@ void RenderableModelEntityItem::render(RenderArgs* args) {
...
+                } else {
+                    // release mesh
+                    if (_collisionMeshKey) {
+                        collisionMeshCache.releaseMesh(_collisionMeshKey);
+                    }
+                    // clear model's collision geometry
+                    model::MeshPointer mesh = nullptr;
+                    _model->setCollisionMesh(mesh);
+                }
+            }
```
AndrewMeadows: The `RenderableModelEntityItem` now tracks whether it should `_showCollisionGeoemtry` and is responsible for giving its _`model` a `collisionGeometry`, or taking it away.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -908,9 +945,21 @@ void RenderableModelEntityItem::computeShapeInfo(ShapeInfo& info) {
...
+    const void* key = static_cast<const void*>(shape);
+    if (_collisionMeshKey != key) {
+        if (_collisionMeshKey) {
+            collisionMeshCache.releaseMesh(_collisionMeshKey);
+        }
+        _collisionMeshKey = key;
+        // toggle _showCollisionGeometry forces re-evaluation later
+        _showCollisionGeometry = !_showCollisionGeometry;
+    }
+}
```
AndrewMeadows: This method is called at the end of a new sequence of logic:

When the shape of a Bullet object is changed...
its `ObjectMotionState` calls its `setShape()`... 
which normally does nothinge except where it is overriden in `EntityMotionState`...
which calls `EntityItem::setCollisionShape()`...
which is normally does nothing except where it is overridden by `RenderableModelEntityItem`...
which stores the `_collisionMeshKey` (a void* copy of the raw `btCollisionShape` pointer!) for later
and toggles `_showCollisionGeometry` which triggers an update later, if necessary.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
AndrewMeadows: Note, this is where the _collisionMeshKey (which was stored previously) is actually used.  Some important things here:

_collisionMeshKey is a void* and actually points to a valid btCollisionShape instance in the ShapeManager.  It is not possible for this pointer to be invalid in this context: it is set by the EntityMotionState which also keeps a pointer to the shape and it ref-counted inside the ShapeManager class which guards all creation/deletion of the Bullet shapes.  The EntityMotionState will set/clear its entity's shape whenever it sets/clears its own pointer, and will be using the ShapeManager for its shapes.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -559,6 +578,18 @@ bool RenderableModelEntityItem::findDetailedRayIntersection(const glm::vec3& ori
...
+    ModelEntityItem::setShapeType(type);
+    if (_shapeType == SHAPE_TYPE_COMPOUND) {
+        if (!_compoundShapeResource && !_compoundShapeURL.isEmpty()) {
+            _compoundShapeResource = DependencyManager::get<ModelCache>()->getGeometryResource(getCompoundShapeURL());
+        }
+    } else if (_compoundShapeResource && !_compoundShapeURL.isEmpty()) {
+        // the compoundURL has been set but the shapeType does not agree
+        _compoundShapeResource.reset();
+    }
+}
```
AndrewMeadows: The `RenerableModelEntityItem` now tracks the `_compoundShapeResource` (a `GeometryResource::Pointer`) instead of that duty falling to its `_model`.

libraries/entities/src/ModelEntityItem.cpp
```diff
@@ -282,8 +282,8 @@ ShapeType ModelEntityItem::computeTrueShapeType() const {
         type = SHAPE_TYPE_COMPOUND;
     }
     if (type == SHAPE_TYPE_COMPOUND && !hasCompoundShapeURL()) {
-        // no compoundURL set --> fall back to NONE
-        type = SHAPE_TYPE_NONE;
+        // no compoundURL set --> fall back to SIMPLE_COMPOUND
+        type = SHAPE_TYPE_SIMPLE_COMPOUND;
```
AndrewMeadows: This is new behavior... we fall back to SIMPLE_COMPOUND if the entity is set to use COMPOUND but its compoundShapeUrl is empty.

Note: we still have problems if the compoundShapeUrl is non-empty but nevertheless invalid for other reasons.

libraries/model-networking/src/model-networking/ModelCache.h
```diff
@@ -111,7 +111,6 @@ class GeometryResourceWatcher : public QObject {
 
     QUrl getURL() const { return (bool)_resource ? _resource->getURL() : QUrl(); }
 
-private:
     void startWatching();
```
AndrewMeadows: Hrm... this change is not necessary.  I'll roll it back.

libraries/physics/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 set(TARGET_NAME physics)
 setup_hifi_library()
-link_hifi_libraries(shared fbx entities)
+link_hifi_libraries(shared fbx entities model)
```
AndrewMeadows: physics lib now depends on model lib.

libraries/render-utils/src/Model.cpp
```diff
@@ -1253,60 +1238,110 @@ void Model::segregateMeshGroups() {
...
+public:
+    CollisionRenderGeometry(model::MeshPointer mesh) {
+        _fbxGeometry = std::make_shared<FBXGeometry>();
+        std::shared_ptr<GeometryMeshes> meshes = std::make_shared<GeometryMeshes>();
+        meshes->push_back(mesh);
+        _meshes = meshes;
+        _meshParts = std::shared_ptr<const GeometryMeshParts>();
     }
-    return false;
+};
```
AndrewMeadows: `CollisionRenderGeometry`'s sole purpose is to provide a ctor of a `Geometry` that accepts a `MeshPointer` at construction.  Maybe I should just add a ctor to the base  class?  I wonder why I didn't think of this at the time.

AndrewMeadows:
test this please

murillodigital:
test this please

murillodigital:
test this please

murillodigital:
test this please

AndrewMeadows:
test this please

hifi-gustavo:
Build console output for commit: 3bd9a0c81dfcb8c42eb090e297993703ef149580 available at: https://gist.github.com/287f5b385ebff12ff127ec426ac554a4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-3bd9a0c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-3bd9a0c.dmg

hifi-gustavo:
Build console output for commit: c7a696ca7549fc0044f1530ac9df09fe970c9760 available at: https://gist.github.com/287f5b385ebff12ff127ec426ac554a4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-c7a696c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-c7a696c.dmg

Nex-Pro:
Testing.

Nex-Pro:
@AndrewMeadows 

#### Some alpha textures appear with solid white where there should be 100% transparency.
pr8329transparencyissue

I used the asylum bed model and fixed the collision model to be the same size.

When ( via a script I wrote ) I rez any number of the bed model, and set the correctly scaled collision model as the compound shape, the collision hull is not displayed at the same size and position as the model.

It takes large numbers of the models all rezzed at different scales to have any effect on frame rates. Literally thousands.

#### Picture of a few models with displaced/mis-sized collision hulls rezzed via script.

collisionhullerror

#### Here's the fbx and the obj of the collision hull.
fbx_and_obj_hull

#### Here's the hulls fitting correctly. Using edit.js
correcthulls

#### Here are images of 2000 beds note the frame rates  in each image.

2000beds

#### Misplaced and scaled hulls for above. Image taken from identical viewpoint.
2000_collisionmeshes

***Avatar collides correctly with the fbx models not the enlarged misplaced collision meshes.***

Nex-Pro:
### We're gonna need a lot of linen !

a_lot_of_sheets_needed

### Sixteen thousand beds prior to crash when choosing to show hulls.

*** Check the triangle count ! ***

16thousandbeds

Nex-Pro:
Multiple crash reports submitted.

AndrewMeadows:
Ok, I'm not going to worry about the theoretical crash mode.

The bad alpha textures is a blocker, but I don't have a good theory about how that broke.  I'll probably have to ask a fellow dev for help.

If the compoundShape is not scaling correctly then that would be a blocker, but at least I have an idea of how to fix it.

hifi-gustavo:
Build console output for commit: 48ddf1a6ff25184e7d8caf681f3a4a20858e1573 available at: https://gist.github.com/287f5b385ebff12ff127ec426ac554a4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-48ddf1a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-48ddf1a.dmg

hifi-gustavo:
Build console output for commit: f95c8fd9c34b01c5dd58021d9d80d36f4e9f4423 available at: https://gist.github.com/287f5b385ebff12ff127ec426ac554a4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-f95c8fd.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-f95c8fd.dmg

hifi-gustavo:
Build console output for commit: 0c4ff9964d133b14c6cdec06d82ade8ff15bae1c available at: https://gist.github.com/287f5b385ebff12ff127ec426ac554a4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-0c4ff99.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-0c4ff99.dmg

hifi-gustavo:
Build console output for commit: ab10b5a4edaa7598c888c41e6779800c3e736707 available at: https://gist.github.com/287f5b385ebff12ff127ec426ac554a4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-ab10b5a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-ab10b5a.dmg

hifi-gustavo:
Build console output for commit: b95cb566cd377a6ff968ede4852c3bd9b5eda5f6 available at: https://gist.github.com/287f5b385ebff12ff127ec426ac554a4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-b95cb56.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-b95cb56.dmg

AndrewMeadows:
This is ready for test again, however there is no need to test the performance.  What needs to be tested:

(1) Make sure transparent textures are not full opaque.  This was reported as a problem but I haven't been able to reproduce it, and can't imagine how the code changes in this PR could have introduced such a bug.  I'm skeptical, but please keep an eye out for where textures with alpha are being used and examine them.

(2) Make sure the visisible and collision geometries are rendered correctly for when...
(2a) ... the model is stretched to different dimensions
(2b) ... the model's registration point is not <0.5, 0.5, 0.5> (e.g. not at the center of the model's bounding box)
Also make sure the collision shape collides correctly under the above transformations.  Test shape types: **box**, **sphere**, **compound** (with OBJ collision model URL), **simple-hull** (aka "One hull"), and **simple-compound** (aka "Hull per submesh").  All of these shape types should produce a visible collision geometry when this option is enabled: **Developer --> Physics --> Draw collision shapes**

Note, the **sphere** collision shape does not yet stretch correctly for non-uniform dimensions -- it is always sphererical.  This is not a new bug and will have to be fixed in a future project.

(3) Make sure the non-static-mesh shape types can be set dynamic and appear to collide and tumble correctly.

hifi-gustavo:
Build console output for commit: d7bb5559877b21f7bfe6df3254a7e30012273a6b available at: https://gist.github.com/287f5b385ebff12ff127ec426ac554a4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-d7bb555.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8329-d7bb555.dmg

RyanDowne:
(1) Make sure transparent textures are not full opaque. This was reported as a problem but I haven't been able to reproduce it, and can't imagine how the code changes in this PR could have introduced such a bug. I'm skeptical, but please keep an eye out for where textures with alpha are being used and examine them.  :+1:


(2) Make sure the visible and collision geometries are rendered correctly for when...
(2a) ... the model is stretched to different dimensions :+1:
(2b) ... the model's registration point is not <0.5, 0.5, 0.5> (e.g. not at the center of the model's bounding box) :+1:
Also make sure the collision shape collides correctly under the above transformations. Test shape types: box, sphere, compound (with OBJ collision model URL), simple-hull (aka "One hull"), and simple-compound (aka "Hull per submesh"). All of these shape types should produce a visible collision geometry when this option is enabled: Developer --> Physics --> Draw collision shapes

:+1: :-1:

Note, the sphere collision shape does not yet stretch correctly for non-uniform dimensions -- it is always spherical. This is not a new bug and will have to be fixed in a future project.

(3) Make sure the non-static-mesh shape types can be set dynamic and appear to collide and tumble correctly. :-1:

Notes : If you import and model and give it a hull and dynamics on import, then change to non dynamic in edit.js, then try to give it a hull again and make dynamic, it doesn't seem to take. The object stays non-dynamic. 

If you are in hull debug mode and you visualize the per mesh hull them move to None, the hull doesn't go away. I think the same thing happens with Compound where there is no model specified in the field.

PR #8333, opened 2016-07-29
Miscellaneous cmake and render engine improvements

SamGondelman:
A few random fixes I made while working on the procedural particles.  That PR is no more (long live procedural particles), but I figured these fixes were still desirable.

This PR should have absolutely no effect, so there is no test plan.  If everything builds, it works.


libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -176,25 +176,6 @@ void RenderableModelEntityItem::doInitialModelSimulation() {
...
-    }
-    if (renderArgs && _model && _needsInitialSimulation && _model->isActive() && _model->isLoaded()) {
-        // make sure to simulate so everything gets set up correctly for rendering
-        doInitialModelSimulation();
-        _model->renderSetup(renderArgs);
-    }
-    bool ready = !_needsInitialSimulation && _model && _model->readyToAddToScene(renderArgs);
-    return ready;
-}
-
```
SamGondelman: This method seems important but it was never being called, I *believe* the work is now done in Model, and getModel is called elsewhere in this class.

libraries/gpu/src/gpu/Shader.h
```diff
@@ -53,19 +53,22 @@ class Shader {
         int32 _location{INVALID_LOCATION};
         Element _element;
         uint16 _resourceType{Resource::BUFFER};
+        uint32 _size { 0 };
```
SamGondelman: This is currently unused but is highly desirable for some shader improvements that will be made later, so I've included it now.

libraries/render-utils/src/AnimDebugDraw.cpp
```diff
@@ -15,6 +15,8 @@
 #include "GLMHelpers.h"
 #include "DebugDraw.h"
 
+#include <qmath.h>
```
SamGondelman: render-utils was previously getting M_PI from its dependency on procedural, I think.  That dependency is gone, so here we are.

hifi-gustavo:
Build console output for commit: 676f4bdfcfdbbe0e2b1d11559d458dc98d11e86a available at: https://gist.github.com/574fa4af0bd6c3eaf04b8c96d3d7e2ca<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8333-676f4bd.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8333-676f4bd.dmg

OmegaHeron:
QA Pass
Tested Windows 10 General smoke test = good.
Testing Linux - Ubuntu 16.04.1 All compiles, stack and interface function.

libraries/gpu/src/gpu/Shader.h
jherico: Can you elaborate on the use for it?  Also, what is this size representing?

libraries/render-utils/src/AnimDebugDraw.cpp
jherico: why `qmath.h` and not `math.h`?

Also, please put system headers before Qt headers and Qt headers before internal library headers.

libraries/gpu/src/gpu/Shader.h
```diff
@@ -53,19 +53,22 @@ class Shader {
...
  
-        Slot(const Slot& s) : _name(s._name), _location(s._location), _element(s._element), _resourceType(s._resourceType) {}
-        Slot(Slot&& s) : _name(s._name), _location(s._location), _element(s._element), _resourceType(s._resourceType) {}
-        Slot(const std::string& name, int32 location, const Element& element, uint16 resourceType = Resource::BUFFER) :
-             _name(name), _location(location), _element(element), _resourceType(resourceType) {}
+        Slot(const Slot& s) : _name(s._name), _location(s._location), _element(s._element), _resourceType(s._resourceType), _size(s._size) {}
+        Slot(Slot&& s) : _name(s._name), _location(s._location), _element(s._element), _resourceType(s._resourceType), _size(s._size) {}
+        Slot(const std::string& name, int32 location, const Element& element, uint16 resourceType = Resource::BUFFER, uint32 size = 0) :
+        _name(name), _location(location), _element(element), _resourceType(resourceType), _size(size) {}
```
jherico: missing tab

jherico:
A few comments but nothing blocking.

SamGondelman:
I might as well fix those few things tomorrow, please hold off on merging until then

libraries/gpu/src/gpu/Shader.h
SamGondelman: It represents the size in bytes of the uniform buffer.  It is useful in cases where the shader code is provided by a user or something like that.  Since we don't know what will be provided, we can't have a matching struct on the C++ size to take the sizeof().

libraries/render-utils/src/AnimDebugDraw.cpp
SamGondelman: math.h only defines M_PI and the like if _USE_MATH_DEFINES is defined.  We define that in the cmake file for interface, but by using qmath.h instead, we make sure that render-utils can still compile if you build it independently.

I'll fix the ordering tomorrow, though.  Thanks.

hifi-gustavo:
Build console output for commit: 9e20d922416e5278518fe3e097e1a77b0612a6e1 available at: https://gist.github.com/574fa4af0bd6c3eaf04b8c96d3d7e2ca<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8333-9e20d92.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8333-9e20d92.dmg

hifi-gustavo:
Build console output for commit: e35d453b37782f27e98a5b5e2d77f716b009632e available at: https://gist.github.com/574fa4af0bd6c3eaf04b8c96d3d7e2ca<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8333-e35d453.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8333-e35d453.dmg

PR #8353, opened 2016-08-02
fix for disabled physics with unloadable entity

birarda:
- fixes a bug where physics would be disabled upon entering a domain if a model entity failed to load

### DIFF
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -916,6 +916,16 @@ bool RenderableModelEntityItem::contains(const glm::vec3& point) const {
     return false;
 }
 
+bool RenderableModelEntityItem::shouldBePhysical() const {
+    // If we have a model, make sure it hasn't failed to download.
+    // If it has, we'll report back that we shouldn't be physical so that physics aren't held waiting for us to be ready.
+    if (_model && _model->didGeometryRequestFail()) {
+        return false;
+    } else {
+        return ModelEntityItem::shouldBePhysical();
+    }
+}
+
 glm::quat RenderableModelEntityItem::getAbsoluteJointRotationInObjectFrame(int index) const {
     if (_model) {
         glm::quat result;
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.h
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.h
@@ -65,6 +65,8 @@ public:
 
     virtual bool contains(const glm::vec3& point) const override;
 
+    virtual bool shouldBePhysical() const override;
+
     // these are in the frame of this object (model space)
     virtual glm::quat getAbsoluteJointRotationInObjectFrame(int index) const override;
     virtual glm::vec3 getAbsoluteJointTranslationInObjectFrame(int index) const override;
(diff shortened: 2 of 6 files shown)


birarda:
#### Test Plan

This is a fix for an easily reproducible bug. 

1. Run a Sandbox and connect to it.
2. Add an entity in front of your avatar with a URL that points to a model that does not exist.
3. Leave the domain or restart your Interface client and return to the domain. Be sure to reconnect close to the entity with the bad URL.

On master/stable, you should notice that physics will never be enabled in that domain. There will be a message in the log that starts with `Physics disabled until entity loads:`. If you are in a domain where there is a ground (like the home set) you will see that your avatar always stays flying.

With this PR, physics should be enabled even with the entity with a bad URL present. The `Physics disabled until entity loads` messages will not repeat in the log. Your avatar should correctly stand on the ground once physics are enabled.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -916,6 +916,16 @@ bool RenderableModelEntityItem::contains(const glm::vec3& point) const {
     return false;
 }
 
+bool RenderableModelEntityItem::shouldBePhysical() const {
+    // If we have a model, make sure it hasn't failed to download.
+    // If it has, we'll report back that we shouldn't be physical so that physics aren't held waiting for us to be ready.
+    if (_model && _model->didGeometryRequestFail()) {
```
Atlante45: Shouldn't that be `!_model || _model->didGeometryRequestFail()`

birarda: This checks if you have a `_model` and the request for said model failed. If you don't have a model yet, we don't know if the request will fail.

Atlante45: Yes, but shouldn't `shouldBePhysical` return `false` if we haven't yet made the request?

hifi-gustavo:
Build console output for commit: 5d3e125a0668d57a34bae8f35fb83303d6a2b21a available at: https://gist.github.com/41614e8f6c9754f4fc94c2cf0e576931<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8353-5d3e125.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8353-5d3e125.dmg

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
birarda: Nope - we hold the enabling of physics in the domain based on this - it's supposed to be physical as far as we know at that point.

kevinmthomasse:
testing

kevinmthomasse:
## QA Test
**QA Tester**: KevinMThomas
**QA Pass**: SUCCESS
**Additional Test Results + Observations**: Ran a Sandbox and connected to it and 
added an entity in front of my avatar with a URL that points to a model that does not exist as I called it ```hddff.fbx```.  Visited another domain and returned to the domain and reconnected close to the entity with the bad URL called ```hddff.fbx```.  Physics was enabled with the bad URL present. The Physics disabled until entity loads messages did not appear in the log and avatar correctly stood on the ground.

PR #5215, opened 2015-06-26
repair bad ordering of PROP_SCRIPT_TIMESTAMP

ZappoMan:

PR #8383, opened 2016-08-06
faster image processing on load

AndrewMeadows:
* faster analysis of transparent textures on load

### DIFF
--- a/libraries/model/src/model/TextureMap.cpp
+++ b/libraries/model/src/model/TextureMap.cpp
@@ -62,26 +62,27 @@ const QImage TextureUsage::process2DImageColor(const QImage& srcImage, bool& val
         if (image.format() != QImage::Format_ARGB32) {
             image = image.convertToFormat(QImage::Format_ARGB32);
         }
-        // Count the opaque and transparent pixels
+
+        // Figure out if we can use a mask for alpha or not
         int numOpaques = 0;
         int numTransparents = 0;
         int height = image.height();
         int width = image.width();
+        const int MAX_TRANSPARENT_PIXELS_FOR_ALPHAMASK = (int)(0.05f * (float)(width * height));
         for (int y = 0; y < height; ++y) {
             const QRgb* data = reinterpret_cast<const QRgb*>(image.constScanLine(y));
             for (int x = 0; x < width; ++x) {
                 auto alpha = qAlpha(data[x]);
                 numOpaques += (int)(alpha == OPAQUE_ALPHA);
-                numTransparents += (int)(alpha == TRANSPARENT_ALPHA);
+                if (alpha == TRANSPARENT_ALPHA) {
+                    if (++numTransparents > MAX_TRANSPARENT_PIXELS_FOR_ALPHAMASK) {
+                        alphaAsMask = false;
+                        break;
+                    }
+                }
             }
         }
-
-        // If alpha was meaningfull refine
-        auto totalNumPixels = height * width;
-        if (numOpaques != totalNumPixels) {
-            auto numTranslucents = totalNumPixels - numOpaques - numTransparents;
-            alphaAsMask = ((numTranslucents / (double)totalNumPixels) < 0.05);
-        }
+        validAlpha = (numOpaques != width * height);
     }
 
     if (!validAlpha && image.format() != QImage::Format_RGB888) {
@@ -658,13 +659,12 @@ const CubeLayout CubeLayout::CUBEMAP_LAYOUTS[] = {
 const int CubeLayout::NUM_CUBEMAP_LAYOUTS = sizeof(CubeLayout::CUBEMAP_LAYOUTS) / sizeof(CubeLayout);
 
 gpu::Texture* TextureUsage::processCubeTextureColorFromImage(const QImage& srcImage, const std::string& srcImageName, bool isLinear, bool doCompress, bool generateMips, bool generateIrradiance) {
-
-    bool validAlpha = false;
-    bool alphaAsMask = true;
-    QImage image = process2DImageColor(srcImage, validAlpha, alphaAsMask);
-
     gpu::Texture* theTexture = nullptr;
-    if ((image.width() > 0) && (image.height() > 0)) {
+    if ((srcImage.width() > 0) && (srcImage.height() > 0)) {
+        QImage image = srcImage;
+        if (image.format() != QImage::Format_RGB888) {
+            image = image.convertToFormat(QImage::Format_RGB888);
+        }
 
         gpu::Element formatGPU;
         gpu::Element formatMip;
@@ -672,7 +672,7 @@ gpu::Texture* TextureUsage::processCubeTextureColorFromImage(const QImage& srcIm
 
         // Find the layout of the cubemap in the 2D image
         int foundLayout = CubeLayout::findLayout(image.width(), image.height());
-        
+
         std::vector<QImage> faces;
         // If found, go extract the faces as separate images
         if (foundLayout >= 0) {
--- a/libraries/model/src/model/TextureMap.cpp
+++ b/libraries/model/src/model/TextureMap.cpp
@@ -59,29 +59,27 @@ const QImage TextureUsage::process2DImageColor(const QImage& srcImage, bool& val
     const uint8 OPAQUE_ALPHA = 255;
     const uint8 TRANSPARENT_ALPHA = 0;
     if (image.hasAlphaChannel()) {
-        std::map<uint8, uint32> alphaHistogram;
-
         if (image.format() != QImage::Format_ARGB32) {
             image = image.convertToFormat(QImage::Format_ARGB32);
         }
-
-        // Actual alpha channel? create the histogram
-        for (int y = 0; y < image.height(); ++y) {
+        // Count the opaque and transparent pixels
+        int numOpaques = 0;
+        int numTransparents = 0;
+        int height = image.height();
+        int width = image.width();
+        for (int y = 0; y < height; ++y) {
             const QRgb* data = reinterpret_cast<const QRgb*>(image.constScanLine(y));
-            for (int x = 0; x < image.width(); ++x) {
+            for (int x = 0; x < width; ++x) {
                 auto alpha = qAlpha(data[x]);
-                alphaHistogram[alpha] ++;
-                validAlpha = validAlpha || (alpha != OPAQUE_ALPHA);
+                numOpaques += (int)(alpha == OPAQUE_ALPHA);
+                numTransparents += (int)(alpha == TRANSPARENT_ALPHA);
             }
         }
 
         // If alpha was meaningfull refine
-        if (validAlpha && (alphaHistogram.size() > 1)) {
-            auto totalNumPixels = image.height() * image.width();
-            auto numOpaques = alphaHistogram[OPAQUE_ALPHA];
-            auto numTransparents = alphaHistogram[TRANSPARENT_ALPHA];
+        auto totalNumPixels = height * width;
+        if (numOpaques != totalNumPixels) {
             auto numTranslucents = totalNumPixels - numOpaques - numTransparents;
-
             alphaAsMask = ((numTranslucents / (double)totalNumPixels) < 0.05);
         }
     }
--- a/libraries/model/src/model/TextureMap.cpp
+++ b/libraries/model/src/model/TextureMap.cpp
@@ -65,24 +65,22 @@ const QImage TextureUsage::process2DImageColor(const QImage& srcImage, bool& val
 
         // Figure out if we can use a mask for alpha or not
         int numOpaques = 0;
-        int numTransparents = 0;
-        int height = image.height();
-        int width = image.width();
-        const int MAX_TRANSPARENT_PIXELS_FOR_ALPHAMASK = (int)(0.05f * (float)(width * height));
-        for (int y = 0; y < height; ++y) {
-            const QRgb* data = reinterpret_cast<const QRgb*>(image.constScanLine(y));
-            for (int x = 0; x < width; ++x) {
-                auto alpha = qAlpha(data[x]);
-                numOpaques += (int)(alpha == OPAQUE_ALPHA);
-                if (alpha == TRANSPARENT_ALPHA) {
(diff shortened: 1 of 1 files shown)


AndrewMeadows:
To test:

(1) Make sure the scene loads.

(2) Verify that transparent textures and partially transparent textures still work correctly.

(3) If possible find a texture that is a mixture of pull opaque and full transparent with no partially transparent pixels --> make sure it still looks right.

(4) Measure the time it takes for a scene with several textures to fully show up (Home should work fine, but make sure your cache is fresh: login once, wait for everything to show up, the logout and log back in and measure the time for the scene to load the second time)--> the time for this PR should not be slower than for the current RC or downloadable dev build under similar conditions.

libraries/model/src/model/TextureMap.cpp
```diff
@@ -59,29 +59,27 @@ const QImage TextureUsage::process2DImageColor(const QImage& srcImage, bool& val
...
-            auto totalNumPixels = image.height() * image.width();
-            auto numOpaques = alphaHistogram[OPAQUE_ALPHA];
-            auto numTransparents = alphaHistogram[TRANSPARENT_ALPHA];
+        auto totalNumPixels = height * width;
+        if (numOpaques != totalNumPixels) {
             auto numTranslucents = totalNumPixels - numOpaques - numTransparents;
-
             alphaAsMask = ((numTranslucents / (double)totalNumPixels) < 0.05);
         }
     }
```
AndrewMeadows: We were building a map of unique alpha values, but then not really using the histogram except for measuring the full opaque and full transparent pixels.  We can do the same thing much faster by not bothering to building the full histogram.

hifi-gustavo:
Build console output for commit: 4f2c00af46158387212c129b06951992835ecade available at: https://gist.github.com/cff6a09fd92754ccf1e6547b44ee011a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8383-4f2c00a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8383-4f2c00a.dmg

libraries/model/src/model/TextureMap.cpp
```diff
@@ -59,29 +59,27 @@ const QImage TextureUsage::process2DImageColor(const QImage& srcImage, bool& val
...
         // If alpha was meaningfull refine
-        if (validAlpha && (alphaHistogram.size() > 1)) {
-            auto totalNumPixels = image.height() * image.width();
-            auto numOpaques = alphaHistogram[OPAQUE_ALPHA];
-            auto numTransparents = alphaHistogram[TRANSPARENT_ALPHA];
+        auto totalNumPixels = height * width;
+        if (numOpaques != totalNumPixels) {
             auto numTranslucents = totalNumPixels - numOpaques - numTransparents;
-
             alphaAsMask = ((numTranslucents / (double)totalNumPixels) < 0.05);
```
ZappoMan: can't you optimize this further by breaking out of the loop early? Why touch every pixel? 

The outputs appear to be: 
- `validAlpha` - this is true if ANY pixel is not `OPAQUE_ALPHA`
- `alphaAsMask` - this is true if `numTranslucents` less than 5% of total pixels

So, can't you set `validAlpha` the first time you see a non `OPAQUE_ALPHA`...
And can't you break out of the loop as soon as you have seen more than 5% translucent pixels.

There may even be some other optimizations you could make to break earlier if you know it's not possible to reach future states.

ZappoMan:
@AndrewMeadows - also - in looking at this a little closer... there are actually two callers of `process2DImageColor` one of them uses the results of the alpha tests, the other one `processCubeTextureColorFromImage()` so all of the alpha processing is wasted time for those call... and worse than that - it tends to be that the cube textures are pretty large.  You could also get a performance boost by not doing the alpha processing when called from `processCubeTextureColorFromImage`

libraries/model/src/model/TextureMap.cpp
AndrewMeadows: Good point, an early exit could be done once we have > 5% alpha.  I'll implement and test it this evening.

ZappoMan: Btw - I hacked a version of this to see if it helps... and of course there is content that it does allow you to short circuit on early -- but interestingly enough... in playa there are about 40% of the textures that have alpha channels - but are 100% opaque -- in this case... we end up touching every pixel looking for that magical single transparent pixel...

We should document that content creators can improve the loading performance of their domains by not using RGBA images  for things that have no transparency... and we should probably fix playa to not have alpha channel images for content that doesn't need transparency.

ZappoMan: it's also worth noting that probably a bunch of the time in `ImageReader::run()` is spent in the Qt code to do image conversion... we might benefit from using a different image library that allows us to get this alpha data out while it's doing the conversion.

libraries/model/src/model/TextureMap.cpp
```diff
@@ -660,21 +659,20 @@ const CubeLayout CubeLayout::CUBEMAP_LAYOUTS[] = {
...
-    bool alphaAsMask = true;
-    QImage image = process2DImageColor(srcImage, validAlpha, alphaAsMask);
-
     gpu::Texture* theTexture = nullptr;
-    if ((image.width() > 0) && (image.height() > 0)) {
+    if ((srcImage.width() > 0) && (srcImage.height() > 0)) {
+        QImage image = srcImage;
+        if (image.format() != QImage::Format_RGB888) {
+            image = image.convertToFormat(QImage::Format_RGB888);
+        }
```
AndrewMeadows: We don't bother scanning for alpha for cube maps -- we weren't using the alpha information anyway.  Instead we just convert to RGB888 when necessary.

libraries/model/src/model/TextureMap.cpp
```diff
@@ -59,31 +59,30 @@ const QImage TextureUsage::process2DImageColor(const QImage& srcImage, bool& val
...
-        if (validAlpha && (alphaHistogram.size() > 1)) {
-            auto totalNumPixels = image.height() * image.width();
-            auto numOpaques = alphaHistogram[OPAQUE_ALPHA];
-            auto numTransparents = alphaHistogram[TRANSPARENT_ALPHA];
-            auto numTranslucents = totalNumPixels - numOpaques - numTransparents;
-
-            alphaAsMask = ((numTranslucents / (double)totalNumPixels) < 0.05);
-        }
+        validAlpha = (numOpaques != width * height);
     }
```
AndrewMeadows: There are three side effects of TextureUsage::process2DImageColor():
(1) update validAlpha to be true it the texture has at least 1 non-opaque pixel
(2) compute whether we can use alphaAsMask (true if 5% or fewer pixels are completely transparent)
(3) convert the image to RGB888 format if none of its pixels have any transparency

This work just speeds up the process:
(A) We don't bother building the full alphaHistogram since we only care about numbers of full opaque and full transparent
(B) We stop scanning pixels as soon as we know we can't use an alpha mask

AndrewMeadows:
I ran added some temporary debug logging and ran this PR in release build against a release build of master and can say: this PR will make some measurable differences in scene load for Playa and Home but yes @zappoman, you're right: there is also significant time being spent QImage::convertToFormat() when the image's actual format does not agree with how it will be used in the render pipeline.  In a most places we are converting to RGB888 (no alpha) but there are also a few places where we convert to ARGB32 and then to Grayscale8 (when converting a regular image to a roughness texture).  I'm with Austin on this one: ultimately we should pre-process the content once and store the textures in asset system using the format that they will be used for rendering.

In the meantime yes: our own Playa and Home content should be optimized by hand, and we should update the guidelines for content creators.

hifi-gustavo:
Build console output for commit: f759fd10330292982ae1ca342d4262d49bd2a176 available at: https://gist.github.com/cff6a09fd92754ccf1e6547b44ee011a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8383-f759fd1.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8383-f759fd1.dmg

libraries/model/src/model/TextureMap.cpp
ZappoMan: Your change appears to change the logic. Old code:
```
alphaAsMask = ((numTranslucents / (double)totalNumPixels) < 0.05);
```
Note: *numTranslucents *

New Code:
```
if (++numTransparents > MAX_TRANSPARENT_PIXELS_FOR_ALPHAMASK) {
```
Note: *numTransparents*

libraries/model/src/model/TextureMap.cpp
```diff
@@ -59,31 +59,30 @@ const QImage TextureUsage::process2DImageColor(const QImage& srcImage, bool& val
...
                 auto alpha = qAlpha(data[x]);
-                alphaHistogram[alpha] ++;
-                validAlpha = validAlpha || (alpha != OPAQUE_ALPHA);
+                numOpaques += (int)(alpha == OPAQUE_ALPHA);
+                if (alpha == TRANSPARENT_ALPHA) {
+                    if (++numTransparents > MAX_TRANSPARENT_PIXELS_FOR_ALPHAMASK) {
+                        alphaAsMask = false;
+                        break;
+                    }
+                }
```
ZappoMan: I think you want another test/break here... you're breaking out of the inner `for (int x` loop but not breaking out of the outer `for (int y` loop

hifi-gustavo:
Build console output for commit: 4e23ecfb2cd6d1ce70cb266d76e7c814f518f945 available at: https://gist.github.com/cff6a09fd92754ccf1e6547b44ee011a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8383-4e23ecf.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8383-4e23ecf.dmg

Nex-Pro:
pr: #8383

**CPU:** 4.4 ghz Intel i7
**RAM:** 16 GB
**GPU:** Nvidia gtx 1070 8gb
**Testing OS:** win 10 pro

IMO this is a **FAIL**

(1) Make sure the scene loads. **Check**

(2) Verify that transparent textures and partially transparent textures still work correctly. **Check**

(3) If possible find a texture that is a mixture of pull opaque and full transparent with no partially transparent pixels --> make sure it still looks right.
        **Examined the leaf texture on the home balcony, seems fine.**


Consistent theme with this is missing content in localhost.

I even restarted my PC.

I compared this to build #5215

I tried to time when content was loaded with a stopwatch but this proved tricky as the progress bar has a habit of vanishing for some time before re-appearing.

In this PR I cleared cache many times and items were missing even after several minutes on most occasions.
In buld 5215 All items loaded almost 100% of the time.

missingcontent1
missingcontent2
missingcontent3

AndrewMeadows:
@Nex-Pro, so the failure is that some objects did not show up?  BTW, was it only objects that were behind the camera at initial login?  or did some things not show up that were supposed to be in front at the moment of login?

It is hard to imagine how the changes made in this PR would cause that... unless there was some bug where processing textures faster caused objects to not appear.  I'd like to try to reproduce the bug you're seeing on the tip of master to verify that it is not happening there.

ZappoMan:
I am highly skeptical that this PR is causing these problems.
- Does this repro 100% of the time?
- Are we sure this doesn't happen on current Beta and/or current RC as well?

I'm not saying that things not appearing isn't a bug... but I don't think this change could possibly cause that to happen.

libraries/model/src/model/TextureMap.cpp
```diff
@@ -59,31 +59,28 @@ const QImage TextureUsage::process2DImageColor(const QImage& srcImage, bool& val
...
-            const QRgb* data = reinterpret_cast<const QRgb*>(image.constScanLine(y));
-            for (int x = 0; x < image.width(); ++x) {
-                auto alpha = qAlpha(data[x]);
-                alphaHistogram[alpha] ++;
-                validAlpha = validAlpha || (alpha != OPAQUE_ALPHA);
+        // Figure out if we can use a mask for alpha or not
+        int numOpaques = 0;
+        int numTranslucents = 0;
+        const int NUM_PIXELS = image.width() * image.height();
+        const int MAX_TRANSLUCENT_PIXELS_FOR_ALPHAMASK = (int)(0.05f * (float)(NUM_PIXELS));
```
Atlante45: I know you reused it from the old code, but do we have a name for that `0.05f`?
Is it an actual value from the format specification, or just something we set ourselves?

AndrewMeadows: I think this is a hand-tuned value that @samcake picked.  The code is simple enough and that magic number is now wrapped in a const with a descriptive name -- we'll tune it again later.

Nex-Pro:
@ZappoMan Items were not showing up that should have been in view.

ZappoMan:
@Nex-Pro - I don't believe that this PR is causing that issue. We should retest - I think something else is causing the problems you're seeing. I didn't QA Approve this, because I'm waiting for more info on this.... but @AndrewMeadows's change couldn't cause objects to not load.

ZappoMan:
@Nex-Pro - when I run this build - I am not seeing any loading problems.

Nex-Pro:
@ZappoMan I went back and forth between the two builds several times. I can try again if you want but it will have to be later today as it is now 3:10 am here.

Nex-Pro:
@ZappoMan  Cheking further I find there seem to be collidable but not viewable objects so it looks like the textures aren't showing up.

Nex-Pro:
@ZappoMan  this might be to do with the tidyguy script. If it fails to load all the models then no matter how many times the cache is cleared it won't bring back what was not there.

AndrewMeadows:
I've tried >10 times to reproduce this problem (some items do not appear on login) to no avail.  We need a few other people to test this out to see if we can get a reliable repro recipe.

OmegaHeron:

All standard tests + specific outlined here pass.  Standard home scene fully loaded for me, more or less, at same rate as current RC/DEV. QA Pass.

AndrewMeadows:
Nex-Pro said in the qa-tester's channel that he's now seeing this problem in dev-download build 5244 so probably not caused by this PR.  I'm going to set the QA-APPROVED label.

PR #8387, opened 2016-08-08
Fix Standard.A/B/X/Y primary thumb JSON mappings

SamGondelman:
We were mapping from Standard.A/B/X/Y to RightPrimaryThumb and Standard.DU/DL/DR/DD to LeftPrimaryThumb, but that mapping is only really true on the SDL controllers.  This should fix it on the Oculus Touches.

### DIFF
--- a/interface/resources/controllers/oculus_touch.json
+++ b/interface/resources/controllers/oculus_touch.json
@@ -1,8 +1,13 @@
 {
     "name": "Oculus Touch to Standard",
     "channels": [
-        { "from": "OculusTouch.A", "to": "Standard.RightPrimaryThumb" },
-        { "from": "OculusTouch.X", "to": "Standard.LeftPrimaryThumb" },
+        { "from": "OculusTouch.A", "to": "Standard.RightPrimaryThumb", "peek": true },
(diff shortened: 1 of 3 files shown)


hifi-gustavo:
Build console output for commit: a5f3089fdaf9f0244f95c3e61d53e33d211ac58a available at: https://gist.github.com/61a58dacbedd5aadfac9dbc9aeea2e3c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8387-a5f3089.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8387-a5f3089.dmg

SamGondelman:
I'm going to test this myself once I get my xbox controller working, and then I'll need someone else to test on the Touches.

Test plan:
- controllers should basically work as before, notably xbox and touch
- try with teleport.js to make sure that Left/RightPrimaryThumb still works (trigger + primary thumb to activate)

SamGondelman:
So *technically* this works on xbox controllers, although it revealed a separate, pre-existing bug with handControllerGrab and teleport.  Those will be worked out separately.

PR #8393, opened 2016-08-09
add JS Render.Performance API

zzmp:


zzmp:
Marking this DNM for now as @samcake thinks it will not work on Macs.

hifi-gustavo:
Build console output for commit: a4403d335f41619eba64065102a170e3afa05df3 available at: https://gist.github.com/572bf03fa06e7c7aaeb02c01e880d86e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8393-a4403d3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8393-a4403d3.dmg

zzmp:
Adds a new JS `Render.Performance` API.

`Render.Performance` maintains a "profiling list" of entities (which can be mutated through its API). 
Every other frame it will cycle through the list and not render the selected entity.
It will then record the that frame and an adjacent (control) frame as an "entry" for that entity.

It will expose aggregate entries, aggregating all entries for an entity. These entries will accumulate (and remain) until `Render.Performance.clear()` is called, whether or not the entity remains on the "profiling list".

An aggregate entry should be of the form:
```
{
  control: {
    // Number of control (full) frames recorded
    runs: Integer,
    // Total time (microseconds) across all entries
    time: Integer
    // Average control frame duration can be calculated as time / runs
  },
  profile: {
    // Number of profile (omitting the entity) frames recorded
    runs: Integer,
    // Total time (microseconds) across all entries
    time: Integer
    // Average profile frame duration can be calculated as time / runs
  },
  // Average time (microseconds) to render the entity
  // May be negative, if omitting the entity from the scene speeds up rendering
  time: Number,
  // Fraction of the frame render time used for the entity
  // Calculated as 1 - profile.time / control.time
  weight: Number
}
```

Methods on the API:
- `Render.Performance.entity(EntityItemID entity)`
  - Add *entity* to the "profiling list"
- `Render.Performance.clear()`
  - Remove all entities from the "profiling list"
-` Render.Performance.clearEntities()`
  - See `Render.Performance.clear()`
- `Render.Performance.clearEntities(EntityItemID entity)`
  - Remove *entity* from the "profiling list"
- `Render.Performance.getEntries()`
  - Returns an object mapping all profiled entities to their aggregated entries (see *Entry* definition below)
- `Render.Performance.getEntries(QVariantMap filter)`
  - Returns an object mapping the profiled entity matching the filter to its aggregated entry
  - *filter* should be an object with an `entity` property (`{ entity: ENTITY_ID }`)
- `Render.Performance.getEntriesByEntity(EntityItemID entity)`
  - Returns an object mapping *entity* to its aggregated entry
- `Render.Performance.clearEntries()`
  - Clears all aggregated entries

---

QA, please test all of these methods. You can find an `EntityItemID` through the edit.js entity explorer.
- Ensure it works with a single entity in the "profiling list"
- Ensure it works with multiple entities in the "profiling list"
- Ensure entries are not removed until `Render.Performance.clearEntries` is called, even if the entry is cleared from the "profiling list"

hifi-gustavo:
Build console output for commit: b1775a1ee8f768cd624a6f87ec10a5ad71233e41 available at: https://gist.github.com/572bf03fa06e7c7aaeb02c01e880d86e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8393-b1775a1.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8393-b1775a1.dmg

hifi-gustavo:
Build console output for commit: b8380b5546a657277ae8018df87d9e7c9a261256 available at: https://gist.github.com/572bf03fa06e7c7aaeb02c01e880d86e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8393-b8380b5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8393-b8380b5.dmg

zzmp:
test this please

hifi-gustavo:
Build console output for commit: b8380b5546a657277ae8018df87d9e7c9a261256 available at: https://gist.github.com/572bf03fa06e7c7aaeb02c01e880d86e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8393-b8380b5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8393-b8380b5.dmg

hifi-gustavo:
Build console output for commit: b07c7ca7f82f3850690876ef76c023ffb71e041c available at: https://gist.github.com/572bf03fa06e7c7aaeb02c01e880d86e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8393-b07c7ca.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8393-b07c7ca.dmg

libraries/render/src/render/CullTask.cpp
```diff
@@ -183,122 +183,98 @@ void CullSpatialSelection::run(const SceneContextPointer& sceneContext, const Re
...
-            for (auto id : inSelection.insideSubcellItems) {
-                auto& item = scene->getItem(id);
-                if (_filter.test(item.getKey())) {
-                    ItemBound itemBound(id, item.getBound());
-                    if (test.solidAngleTest(itemBound.bound)) {
-                        outItems.emplace_back(itemBound);
-                    }
+            std::for_each(inSelection.insideSubcellItems.begin(), inSelection.insideSubcellItems.end(), [&](unsigned int id) {
+                AABox bound;
+                if (notFiltered(id, bound) && inSolidAngle(bound)) {
```
samcake: For all these tests with culling enabled i would still like to keep the 2 if levels for readability and to make sure this is happening in the correct order.
1st filter against key
THen if it pass do the culling test

THis put my brain to peace when loooking at this code, not worrying that the bound is not defined when we enter the  cull test and unfortunately defined after.

libraries/render/src/render/CullTask.cpp
```diff
@@ -183,122 +183,98 @@ void CullSpatialSelection::run(const SceneContextPointer& sceneContext, const Re
...
-                if (_filter.test(item.getKey())) {
-                    ItemBound itemBound(id, item.getBound());
-                    if (test.frustumTest(itemBound.bound)) {
-                        if (test.solidAngleTest(itemBound.bound)) {
-                            outItems.emplace_back(itemBound);
-                        }
-                    }
+            std::for_each(inSelection.partialSubcellItems.begin(), inSelection.partialSubcellItems.end(), [&](unsigned int id) {
+                AABox bound;
+                if (notFiltered(id, bound) && inFrustum(bound) && inSolidAngle(bound)) {
```
samcake: Here too, definitely keep the 3 if blocks

libraries/render/src/render/CullTask.cpp
samcake: on second thought, i realize it s the same  and yes the && will happen in the correct order but still
just readabilty of the algorithm then, that s my only argument :(

samcake: on third thought i think i can live with it, i want to feel young again and read through if complex statement like i m 21
SO i d be ok to keep the one line if expression

libraries/render/src/render/Engine.h
```diff
@@ -45,9 +48,11 @@ namespace render {
     protected:
         SceneContextPointer _sceneContext;
         RenderContextPointer _renderContext;
+
+    private:
+        EngineStats* _stats;
```
samcake: If it becaomes a member why not a smart pointer ?

libraries/render/src/render/EngineStats.h
```diff
@@ -12,46 +12,76 @@
...
 #include <QElapsedTimer>
 
-#include "Engine.h"
+#include <gpu/Context.h>
 
 namespace render {
-
-    // A simple job collecting global stats on the Engine / Scene / GPU
-    class EngineStatsConfig : public Job::Config{
+    class EngineStats : public QObject {
```
samcake: I was about to ask you to update the scripts that use this but i realize the syntax is the same, cool

libraries/render/src/render/CullTask.cpp
```diff
@@ -183,122 +183,98 @@ void CullSpatialSelection::run(const SceneContextPointer& sceneContext, const Re
...
-            PerformanceTimer perfTimer("insideFitItems");
-            for (auto id : inSelection.insideItems) {
-                auto& item = scene->getItem(id);
-                if (_filter.test(item.getKey())) {
-                    ItemBound itemBound(id, item.getBound());
-                    outItems.emplace_back(itemBound);
-                }
+    // visibility cull if partially selected (octree cell containing it was partial)
+    // distance cull if was a subcell item (octree cell is way bigger than the item bound itself, so now need to test per item)
+    auto notFiltered = [&](unsigned int id, AABox& bound) {
```
samcake: Here i would definitely create 2 versions of the lambda, one checking about exception, the other not so in the best case scenario you avoid the test.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -438,11 +442,9 @@ void RenderableModelEntityItem::render(RenderArgs* args) {
...
                 _model->removeFromScene(scene, pendingChanges);
 
-                render::Item::Status::Getters statusGetters;
-                makeEntityItemStatusGetters(getThisPointer(), statusGetters);
-                _model->addToScene(scene, pendingChanges, statusGetters, _showCollisionHull);
-
-                scene->enqueuePendingChanges(pendingChanges);
+                assert(_myRenderer);
+                QMetaObject::invokeMethod(_myRenderer, "addingEntity", Qt::QueuedConnection,
+                    Q_ARG(EntityItemID, getEntityItemID()));
```
samcake: I don;t understand that?
why are we invoking Qt signal slot ?

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -811,7 +812,9 @@ void EntityTreeRenderer::addEntityToScene(EntityItemPointer entity) {
     render::PendingChanges pendingChanges;
     auto scene = _viewState->getMain3DScene();
     if (entity->addToScene(entity, scene, pendingChanges)) {
-        _entitiesInScene.insert(entity->getEntityItemID(), entity);
+        auto entityID = entity->getEntityItemID();
+        _entitiesInScene.insert(entityID, entity);
+        _entitiesRenderItems.insert(entityID, pendingChanges.getItems());
```
samcake: I don;t know, maybe it s just me
but keeping yet another list of all the renderItemID (riID) for all the enitities seems like a bad idea ?

As i udnerstand, you need this to know all the riID so you can perform the test.
COuldn;t you just go through the EntityTree, collect these riID on each individual entities when you start the performanceTest ?

It could be cool to have that list around though, just not convinced it s necessary since it's only used for test right now. building it in flight when calling EntityTree::getRenderItems could work too no ?

libraries/render/src/render/Engine.h
zzmp: It is a member of Engine so that the Engine instance can track it and update it in Engine::run without any lookup overhead, but its "owner" is actually its parent QObject, and it will be destroyed by Qt when the parent is destroyed. The parent, in this case, is the Engine's configuration.

libraries/render/src/render/EngineStats.h
zzmp: I needed to change this so that EngineStats updating wuldn't trigger every other configuration to update every frame.

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
zzmp: Every else around here is careful to make sure that _myRenderer is only called in its own thread, so I wanted to be defensive about it.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
zzmp: I saw the alternative as adding a virtual method to Entity to make entities give their render items, as they are already tracking them. However, this broke the separation between Entity and the rendering engine, which I didn't want to do.

This introduces the least new code, at the expense of one new structure.

@jherico may have had other reasons to prefer this way, we discussed it briefly before I implemented it.

libraries/render/src/render/CullTask.cpp
zzmp: I'll check it to see. Worst case scenario here is we just go away from these lambdas.

zzmp: I'm going to undo it. A good compiler should inline lambdas (according to stack overflow), but there's no way to guarantee it.

hifi-gustavo:
Build console output for commit: 46d8df7fb87f25ba1286cdc08bea52a394c6a380 available at: https://gist.github.com/572bf03fa06e7c7aaeb02c01e880d86e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8393-46d8df7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8393-46d8df7.dmg

PR #8399, opened 2016-08-09
Teleport works with xbox controllers

SamGondelman:
If a controller doesn't provide a valid hand pose (like with xbox controllers), use the avatar's head position and orientation instead.

This allows you to teleport with xbox controllers by controlling the teleporter with your head.

### DIFF
--- a/scripts/system/controllers/teleport.js
+++ b/scripts/system/controllers/teleport.js
@@ -234,17 +234,12 @@ function Teleporter() {
     this.rightRay = function() {
         var pose = Controller.getPoseValue(Controller.Standard.RightHand);
         var rightPosition = pose.valid ? Vec3.sum(Vec3.multiplyQbyV(MyAvatar.orientation, pose.translation), MyAvatar.position) : MyAvatar.getHeadPosition();
-        var rightRotation = pose.valid ? Quat.multiply(MyAvatar.orientation, pose.rotation) : MyAvatar.headOrientation;
-
-        var rightFinal = Quat.multiply(rightRotation, Quat.angleAxis(-90, {
-            x: 1,
-            y: 0,
-            z: 0
-        }));
+        var rightRotation = pose.valid ? Quat.multiply(MyAvatar.orientation, pose.rotation) :
+                                         Quat.multiply(MyAvatar.headOrientation, Quat.angleAxis(-90, {x: 1, y: 0, z: 0}));
 
         var rightPickRay = {
             origin: rightPosition,
-            direction: Quat.getUp(pose.valid ? rightRotation : rightFinal),
+            direction: Quat.getUp(rightRotation),
         };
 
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 956078bc91ef652ad1b38f66a3d767f87155de35 available at: https://gist.github.com/1db278c3d23bd2456a39bc7c53bb873d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8399-956078b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8399-956078b.dmg

scripts/system/controllers/teleport.js
```diff
@@ -232,22 +232,19 @@ function Teleporter() {
...
+        var rightFinal = Quat.multiply(rightRotation, Quat.angleAxis(-90, {
             x: 1,
             y: 0,
             z: 0
         }));
 
         var rightPickRay = {
             origin: rightPosition,
-            direction: Quat.getUp(rightRotation),
+            direction: Quat.getUp(pose.valid ? rightRotation : rightFinal),
```
seefo: this condition looks redundant since rightFinal's value already depends on pose.valid

SamGondelman:
Test plan:
- using an xbox controller, run teleport.js from this PR.  Press any of the 4 buttons or any of the buttons on the D-pad and you should see the teleportal destination appear and be controlled by your avatar's head
- should work as before with hand controllers

scripts/system/controllers/teleport.js
SamGondelman: I guess the naming is a little janky here but actually you only want to rotate rightRotation if the pose isn't valid ("up" for the hand is forward, but "up" for the avatar is actually up).

I'll refactor to eliminate the extra condition, though.

OmegaHeron:
Test Windows 10 - i5 6500 CPU - GTX 970 GPU

Using XBox controller and desktop - able to TP using controller and mouse to set target
Using XBox controller and DK2 am able to TP where I'm looking.

I tried some **very** long TP jumps in dev-playa, purposely looked at cracks between objects and teleported to etc etc.  In all cases I went where I would generally expect to have went and couldn't make myself end up in a place where I'd have a problem (like inside a crevice).

This worked rather well for me.  QA Pass.

hifi-gustavo:
Build console output for commit: 2360e90d395c5e9906188cfa3070b1913efc3909 available at: https://gist.github.com/1db278c3d23bd2456a39bc7c53bb873d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8399-2360e90.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8399-2360e90.dmg

PR #8406, opened 2016-08-10
Disable loading client scripts over ATP

AlexanderOtavka:
Client scripts don't make sense when loaded from ATP, and fail to load when interface is restarted or when changing domains.

Fixes [FogBugz 1409](https://highfidelity.fogbugz.com/f/cases/1409/Client-scripts-served-over-ATP-fail-to-load-on-interface-startup).

### Testing
1. Go to Edit>Open and Run Script from URL...
2. Type in a valid ATP URL pointing to a valid script in your ATP server
3. A warning message should show saying ATP client scripts are not allowed
4. Type in a valid ATP URL  to a script that doesn't exist
5. The same error should show
6. A valid HTTP URL to a script that exists should load
7. A valid HTTP URL to a script that doesn't exist should show an error message saying the script failed to load (different error from the ATP message)
8. An invalid URL should show the same error as step 7

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -5114,8 +5114,10 @@ void Application::setPreviousScriptLocation(const QString& location) {
 }
 
 void Application::loadScriptURLDialog() const {
-    auto newScript = OffscreenUi::getText(nullptr, "Open and Run Script", "Script URL");
-    if (!newScript.isEmpty()) {
+    auto newScript = OffscreenUi::getText(OffscreenUi::ICON_NONE, "Open and Run Script", "Script URL");
+    if (QUrl(newScript).scheme() == "atp") {
+        OffscreenUi::warning("Error Loading Script", "Cannot load client script over ATP");
+    } else if (!newScript.isEmpty()) {
         DependencyManager::get<ScriptEngines>()->loadScript(newScript);
     }
 }


hifi-gustavo:
Build console output for commit: 4e16998b450f72e13fc376fa19efb1b29e3c0a54 available at: https://gist.github.com/3d8d0e9b732902c9b69ddfe1b9ac3fc7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8406-4e16998.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8406-4e16998.dmg

humbletim:
> Client scripts don't make sense when loaded from ATP

ATP Client scripts are an awesome way to orchestrate Domain-wide UI functionality, such as managing a Navigation Sitemap Overlay.

> , and fail to load when interface is restarted or when changing domains.

Fair enough, but if possible please reconsider from Domain operator standpoint.

Maybe for now ATP Client scripts could just be omitted from the list getting saved to settings?

jherico:
@humbletim you could accomplish the same thing by adding an entity script to domain-wide zone that pushes the functionality onto the client.

AlexanderOtavka:
@humbletim yeah, this only prevents the user from loading it from the dialog, it doesn't disable anything under the hood.  The reasoning being that ATP client scripts are not a good experience for the user since they fail to load when changing domains, so they would need to manually reload the script from the URL every time they enter the domain.

humbletim:
@AlexanderOtavka OK, cool.

@jherico Sounds good too.

Just to mention, when working with UI windows and controller mappings Client scripts have been a significant time saver.  My top three reasons to prefer them in those cases are:

1. Visibility -- Client scripts "hide in plain sight" so when one misbehaves it's two keystrokes and one mouse click to find/terminate.
2. Deterministic -- During local development Client scripts only reload when told to.  Local entity scripts instead get reloaded automatically, though not predictably.
3. Autonomous -- Client scripts run in their own container, which offers several advantages such as being able to use balanced `Script.scriptEnding.connect`s for scheduling cleanup routines.

imgntn:
This works as described.

PR #8416, opened 2016-08-11
Don't always mark procedural shaders dirty

SamGondelman:
Previously, anytime you modified any part of a procedural shape entities user data, it would get marked as _shaderDirty.  As a result, repeatedly modifying the uniforms would also reset the shaders, and, in turn, would continually reset iGlobalTime to 0.  No more!

### DIFF
--- a/libraries/procedural/src/procedural/Procedural.cpp
+++ b/libraries/procedural/src/procedural/Procedural.cpp
@@ -115,6 +115,11 @@ bool Procedural::parseUrl(const QUrl& shaderUrl) {
         return false;
     }
 
+    // If the URL hasn't changed, don't mark the shader as dirty
+    if (_shaderUrl == shaderUrl) {
+        return true;
+    }
+
     _shaderUrl = shaderUrl;
     _shaderDirty = true;
 


SamGondelman:
Testing:
- run https://hifi-content.s3.amazonaws.com/samuel/createPokestop.js
- move farther than 5 meters from the pokestop so that it shrinks into a cube
- the cube should be animated and spinning

Previously:
- the cube would not spin, because iGlobalTime was continually set to 0

hifi-gustavo:
Build console output for commit: 0f988fb209cdc93a6a848495a6f66e7346e23fc9 available at: https://gist.github.com/9eb438fd6589300ee378a856351b50d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8416-0f988fb.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8416-0f988fb.dmg

imgntn:
QA pass -- that's pretty sweet.  have you tried your script while multiple avatars are around to see how the baton performs?  lets try that tomorrow!

PR #8428, opened 2016-08-12
Snapshot of Snapshot-Discovery to master

howard-stearns:
There is now a snapshot button in the toolbar:
- All HUD UI elements are removed for the picture, and it it works in HMD, including the notification. (The HMD snapshot is somewhat tall and has a wide field of view.)
- If you are logged in to a shareable place, you also get an option to share the snapshot to a public feed.

The "Go To" address bar has changed:
- The visual design is different, and the buttons highlight when you mouse-over or laser-over them.
- There are two new buttons on the right, which toggle (radio-button style) between the places suggestions and the public feed.
- The suggestions or feed activity is now scrollable and "flickable" (with velocity, and not limited to just the first three shown).
- The suggestions are now based on individual place names (although the concurrency number is for the whole domain).
- When you click on an item in the feed, a larger card comes up with more information.


howard-stearns:
Exclusions:
The following might be included in first release if there is time:
- A web page for each shared snapshot.
- Per-entry-place concurrency numbers, and including places that point to other places.
The following will not be included in the first release:
- Drilling down through expanded card and web page to show user and place, with their corresponding information and shared activities.
- Sharing to social media.
- Indivdiualized feeds.

howard-stearns:
Test Plan:
- Snapshot should work in both desktop and HMD, including the notification. No HUD UI should be visible (not even stats, open dialogs, or the HUD toggle button).
- The files should be saved in the place you have chosen for snapshots (in your settings). They are not deleted (e.g., when shared).
- If you are not in a shareable place, the notification should tell you that you cannot share from here. (Currently, the shareable places are any that you have entered through a named place, not localhost, ip address, domain id, nor temporary domain name.)
- Otherwise, if you are not logged in, the notification should tell you that you that you cannot share without being logged in.
- If you decline to share, the picture should not appear in the feed.
- The address bar should work in both desktop and HMD. The Vive keyboard should not obscure anything.
- The buttons should highlight with mouse-over or laser-over.
- The back and forward buttons should enable/disable appropriately as you visit more places and go back and forward through the session history.
- The suggestions/feed should be scrollable by dragging/flicking (mouse and HMD lasers). The card under the mouse/laser should highlight. (When highlighting, the card is made fully visible within the scroll. This may stop a "flick".)
- The places/feed buttons should operate as a toggling pair. The state should be retained whenever you bring up the address bar throughout the session (but not between sessions).
- The places should display the place name and the number of users present, with the most populated domains listed first. Only shareable places without entrance restrictions, and which point directly to a domain-id (not another place) are shown. If you haven't typed anything, it should only show such places that further have a preview picture and 1-20 current users. As you type (or delete) it should consider all places (regardless of picture or number of users), but only show those for which each word you type appears in the place name, place description, or domain-wide tags (case insensitive, but not yet canonicalizing/stripping accents). Clicking a place takes you to that place, and dismisses the address bar.
- The feed should show everything in reverse chronological order (currently through the first 100). Searching is as for place suggestions, but using the photographer's name, place name, and the word "snapshot". Clicking a feed item shows a larger version, with a description and a button to go to the specific location where the picture was taken (after which the address bar is no longer displayed, and when the address bar is next brought up, the enlarged card is not displayed). Clicking anywhere on the enlarged card other than the button will dismiss the card.

hifi-gustavo:
Build console output for commit: 2641c0fd76e015ba6de68e2a41f803cec4552dff available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-2641c0f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-2641c0f.dmg

Nex-Pro:
**PR** **#8428**

**CPU:** 4.4 ghz Intel i7
**RAM:** 16 GB
**GPU:** Nvidia gtx 1070 8gb
**Testing OS:** win 10 pro
**HMD:** HTC Vive.
**Hand Controllers:** Vive
**Leap Motion:** Not used

Test Plan:

Snapshot should work in both desktop and HMD, including the notification. No HUD UI should be visible (not even stats, open dialogs, or the HUD toggle button).: **SUCCESS**

The files should be saved in the place you have chosen for snapshots (in your settings). They are not deleted (e.g., when shared).: ***As this was a fresh install I was unaware at this point that I needed to drop to desktop to choose the folder so for some moments I was uninformed as to what was happening.***
**Files successfully saved to selected folder.**

If you are not in a shareable place, the notification should tell you that you cannot share from here. (Currently, the shareable places are any that you have entered through a named place, not localhost, ip address, domain id, nor temporary domain name.): **SUCCESS**

Otherwise, if you are not logged in, the notification should tell you that you that you cannot share without being logged in.: **SUCCESS**

If you decline to share, the picture should not appear in the feed. : **SUCCESS**

The address bar should work in both desktop and HMD. The Vive keyboard should not obscure anything.: **SUCCESS**

The buttons should highlight with mouse-over or laser-over.: **SUCCESS**

The back and forward buttons should enable/disable appropriately as you visit more places and go back and forward through the session history.: **SUCCESS**

The suggestions/feed should be scrollable by dragging/flicking (mouse and HMD lasers). The card under the mouse/laser should highlight. (When highlighting, the card is made fully visible within the scroll. This may stop a "flick".): **SUCCESS**

The places/feed buttons should operate as a toggling pair. The state should be retained whenever you bring up the address bar throughout the session (but not between sessions).: **SUCCESS**

The places should display the place name and the number of users present, with the most populated domains listed first. : **SUCCESS**

Only shareable places without entrance restrictions, and which point directly to a domain-id (not another place) are shown. **UNSURE HOW TO CONFIRM THIS** ( would need some location names to test against )

If you haven't typed anything, it should only show such places that further have a preview picture, aren't where you just entered, and which have 1-20 current users. : **SUCCESS**

As you type (or delete) it should consider all places (regardless of picture or number of users or where you entered), but only show those for which each word you type appears in the place name, place description, or domain-wide tags (case insensitive, but not yet canonicalizing/stripping accents). : **SUCCESS**

Clicking a place takes you to that place, and dismisses the address bar. : **SUCCESS**

The feed should show everything in reverse chronological order (currently through the first 100). Searching is as for place suggestions, but using the photographer's name, place name, and the word "snapshot". Clicking a feed item shows a larger version, with a description and a button to go to the specific location where the picture was taken (after which the address bar is no longer displayed, and when the address bar is next brought up, the enlarged card is not displayed). Clicking anywhere on the enlarged card other than the button will dismiss the card. : **SUCCESS but unable to check up to 100**

### NOTES: 

It seems odd to combine the location and snapshot feed on the address bar when the address bar is not also launched via the snapshot button. This seems counter-intuitive. IE opening or closing the snapshots feed is via the GOTO icon. This, surely, is an odd association to make. 

Also why not simply have a dismiss 'X' on the end of the address/feed bar where most people might expect it?

Thanks @Howard

howard-stearns:
noted needs:
1. Clearer close for address bar and/or user story card?
2. Clearer snapshot storage location or association with preference?
3. Clearer association between snapshot sharing and GoTo feed?

howard-stearns:
Re "Do Not Merge": Current plan is to not merge this until we do the following, this week:
1.  Core functionality: Produce Web page views of User Stories, Places, and Users that all link to each other, and replace the current User Story expanded QML card with a Web View.
2. UX improvements.

hifi-gustavo:
Build console output for commit: 3f42e644debd036f2a3203cb5721a925c5fe5d0d available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-3f42e64.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-3f42e64.dmg

hifi-gustavo:
Build console output for commit: 37dc324cec1472e68ed26d466093fe587b05036d available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-37dc324.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-37dc324.dmg

hifi-gustavo:
Build console output for commit: 01fcaa024d083055a59e80b5ce7d809c63f7aa1f available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-01fcaa0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-01fcaa0.dmg

interface/src/Application.h
```diff
@@ -263,6 +266,7 @@ class Application : public QApplication,
     void activeDisplayPluginChanged();
 
     void uploadRequest(QString path);
+    void receivedHifiSchemeURL(QString path);
```
howard-stearns: let's call this parameter url

interface/src/scripting/WindowScriptingInterface.cpp
```diff
@@ -96,7 +96,7 @@ void WindowScriptingInterface::alert(const QString& message) {
 /// \param const QString& message message to display
 /// \return QScriptValue `true` if 'Yes' was clicked, `false` otherwise
 QScriptValue WindowScriptingInterface::confirm(const QString& message) {
-    return QScriptValue((QMessageBox::Yes == OffscreenUi::question("", message)));
+    return QScriptValue((QMessageBox::Yes == OffscreenUi::question("", message, QMessageBox::Yes | QMessageBox::No)));
```
howard-stearns: Bug fix: without the additional arg to question(), it was just acting as a message box with no opportunity for the user to answer "yes", and always answering false to the caller.

scripts/system/snapshot.js
```diff
@@ -0,0 +1,156 @@
...
+    var persisted = Settings.getValue('openFeedAfterShare', true); // might answer true, false, "true", or "false"
+    return persisted && (persisted !== 'false');
+}
+function showFeedWindow() {
+    DialogsManager.showFeed();
+}
+
+var outstanding;
+function confirmShare(data) {
+    var dialog = new OverlayWebWindow('Snapshot Review', Script.resolvePath("html/ShareSnapshot.html"), 800, 470);
```
howard-stearns: let's change the file name to match the window title, which is the name we use in discussions.

hifi-gustavo:
Build console output for commit: 1eebf00d953665d6645dd9b681d69e78860169bb available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-1eebf00.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-1eebf00.dmg

interface/src/ui/Snapshot.cpp
```diff
@@ -141,3 +142,93 @@ QFile* Snapshot::savedFileForSnapshot(QImage & shot, bool isTemporary) {
...
+                                    QJsonDocument(rootObject).toJson());
+                                    
+    } else {
+        emit DependencyManager::get<WindowScriptingInterface>()->snapshotShared(false);
+    }
+}
+
+void SnapshotUploader::uploadFailure(QNetworkReply& reply) {
+    // TODO: parse response, potentially helpful for logging (?)
+    emit DependencyManager::get<WindowScriptingInterface>()->snapshotShared(false);
```
howard-stearns: let's have this pass errorDescriptionOrFalsey, localPath, so that clients can learn what went wrong.

interface/resources/qml/AddressBarDialog.qml
```diff
@@ -45,50 +47,92 @@ Window {
...
         toggleOrGo(true);
     }
-    property var allDomains: [];
-    property var suggestionChoices: [];
-    property var domainsBaseUrl: null;
+    property var allPlaces: [];
+    property var allStories: [];
     property int cardWidth: 200;
     property int cardHeight: 152;
+    property string metaverseBase: "https://metaverse.highfidelity.com/api/v1/";
```
howard-stearns: Can we make this come from NetworkingConstants?

hifi-gustavo:
Build console output for commit: e40b33e618a354f1162dd761b5d2ba12770d59bd available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-e40b33e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-e40b33e.dmg

howard-stearns:
Notes to code reviewers:

Almost all of the .cpp/.h changes are exposing things to .js/.qml:
- Window.takeSnapshot(notify = true, aspectRatio = 0.0f) and signal Window.snapshotTaken(localPath, notify)
 - Now uses the old notify only when asked and passes it to signal.
 - If aspectRatio is non-zero, carves out the largest chunk from the middle that has that width/height.
- Window.shareSnapshot(localPath) and signal Window.snapshotShared(localPath, notify).
- location.placename - where hostname can be ip address, domain id, etc., placename must be a shareable place name.
- Account.checkAndSignalForAccessToken() - used to safely open login when auth is needed
- Desktop.show(path, title) - brings up the specified settings dialog
- Dialogs.showFeed() - to show the feed within the address bar (switching to feed mode if necessary)
- app and AddressBarDialog signals receivedHifiScheme(url) when asked to process a hifi: url, which we use to close address bar when you click on a hifi link (e.g., from the address bar's feed display).

Bug fixes:
- OpenVrHelpers.cpp: QML text input fields get updated as each character is typed on the virtual keyboard, instead of just on enter. (We use this update the suggestions/feed as you type.)
- WindowScriptingInterface.cpp: confirm dialog now presents yes/no buttons and answers accordingly, instead of always answering false. (Used during development before the dedicated snapshot review dialog.)
- Button.qml -- make file self-contained so it can be used in more places
- TextInput.qml -- allow styling of helper/placeholder text

Otherwise, all the new behavior is in AddressBarDialog.qml (and the Card.qml it uses), or snapshot.js (and the SnapshotReview.html it uses).

howard-stearns:
Additional test plan (beyond that which was done/passed above):
- Snapshots:
 - Should not show cursor.
 - Should be the largest 1.91 width/height aspect ratio that fits within the viewable area, for desktop and hmd.
- When reviewing a snapshot you just took:
 - If not logged in, the login dialog should open when you press "share in feed", and the snapshot will not actually be shared.
 - The "open feed after" checkbox should control whether or not the feed is opened. This checkbox should persist across sessions. When opening the address bar to show the feed, it should switch to feed even if it had previously been showing "places". It should open the feed when asked even if it needs to show login (although login should be on top, and the snapshot will not actually be shared if not logged in).

hifi-gustavo:
Build console output for commit: e45f390fd206affb0ea9eb056a122887479562a0 available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-e45f390.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-e45f390.dmg

howard-stearns:
test this please

hifi-gustavo:
Build console output for commit: e45f390fd206affb0ea9eb056a122887479562a0 available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-e45f390.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-e45f390.dmg

hifi-gustavo:
Build console output for commit: c262b0a49728adc1d0e02238fcc118fbe866a59a available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-c262b0a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-c262b0a.dmg

hifi-gustavo:
Build console output for commit: 423083d24442da0b55d586797de453c163d1d734 available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-423083d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-423083d.dmg

hifi-gustavo:
Build console output for commit: 0a67fea6fe9dd0a2cceacd3774bc260fe10c7c45 available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-0a67fea.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-0a67fea.dmg

hifi-gustavo:
Build console output for commit: 35829f5f010206a9d2311e0fcd393c54ec261255 available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-35829f5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-35829f5.dmg

hifi-gustavo:
Build console output for commit: 1dd255f9d869c9ab0d0c54e5b7ab19c72a176f19 available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-1dd255f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-1dd255f.dmg

howard-stearns:
One known issue: when you click on a snapshot in the feed to bring up a snapshot card, the whole card isn't visible without some vertical scrolling. This might be addressed by styling on the server, or we might make the card bigger. In any case, the issue is known.

Nex-Pro:
Testing in progress.

Nex-Pro:
I ran through the above and checked all below.


Additional test plan (beyond that which was done/passed above):

**Snapshots:**

Should not show cursor. **SUCCESS**

Should be the largest 1.91 width/height aspect ratio that fits within the viewable area, for desktop and hmd.  **SUCCESS**

(Note that ctrl-s remains unchanged: it still shows HUD elements, and it still uses the full viewport, although it has now been changed to not show the double-view when in HMD.)  **SUCCESS**

**When reviewing a snapshot you just took:**

If not logged in, the login dialog should open when you press "share in feed", and the snapshot will not actually be shared.  **SUCCESS**

The "open feed after" checkbox should control whether or not the feed is opened. This checkbox should persist across sessions.  **SUCCESS**

When opening the address bar to show the feed, it should switch to feed even if it had previously been showing "places".  **SUCCESS**

It should open the feed when asked even if it needs to show login (although login should be on top, and the snapshot will not actually be shared if not logged in). **SUCCESS**

When you click on a snapshot in the feed, the card it brings up allows you to drill down to the user that took it (seeing their snapshot activity), and the place it was taken (seeing the other snapshot activity taken there). Clicking on a snapshot picture in this expanded view takes you there. **This failed with an error message....**  
***"We're sorry, but something went wrong.***
***If you are the application owner check the logs for more information."***

 **...when clicking on Howard's name. But this may be due to Howard and therefore correct behaviour rather than a problem with this PR as clicking on Ryan's name drilled down.**

**Feedback:**

**From not logged in... Take snapshot -> See login message -> log-in -> snapshot not sent to feed ?** Surely if the user goes to the trouble to log-in having taken a snapshot he/she might reasonably expect that the reward for logging in is that the snapshot makes it to the feed.

Also, and I think this is a biggie, If the user has to turn their head and look up to shoot the snapshot icon in HMD mode and they get the angle of view at the time of pulling the trigger then the user is not getting the snapshot they want. I found myself pre-positioning the HUD so the snapshot button was where I needed it for the POV I wanted. Hardly a recipe for spontaneous happy-snapping. I think this needs a re-design.

howard-stearns:
My conclusion, then, is pass, with the following notes:
- Server-side error for displaying the activity of some users. [HRS: Maybe it is usernames that contain dots?]
- UX flow for not-logged in is problematic: if user does log in soon enough at prompt, it would be nice to upload picture for them at that time. [HRS: if they push login window aside and then login an hour later, should we still upload pictures at that time? What about login in a subsequent session?]
- Potential problems with immediate-snapshot-rather-than-countdown: might be looking up or have hand in the way. (Mitigated by moving the HUD button tray.)

zzmp:
UX note:
- When using the address bar for place suggestions, you are taken to the place when you click on it.
- When using the address bar as a feed, you are taken to a picture, which you need to click again to go to.

You are already in an address bar, so you are *trying* to go somewhere. It feels broken that this behavior is incongruous. Instead of fixing the known issue of the card coming up too large for its container, can you just go to the location the picture was at? The user will already have seen the picture through the feed.

hifi-gustavo:
Build console output for commit: b3623da7725a140f0fbb3712bccdb48ea8ec6696 available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-b3623da.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-b3623da.dmg

interface/resources/qml/AddressBarDialog.qml
```diff
@@ -45,50 +47,80 @@ Window {
...
+    property string metaverseBase: addressBarDialog.metaverseServerUrl + "/api/v1/";
 
     AddressBarDialog {
         id: addressBarDialog
         implicitWidth: backgroundImage.width
         implicitHeight: backgroundImage.height
+        // The buttons have their button state changed on hover, so we have to manually fix them up here
+        onBackEnabledChanged: backArrow.buttonState = addressBarDialog.backEnabled ? 1 : 0;
+        onForwardEnabledChanged: forwardArrow.buttonState = addressBarDialog.forwardEnabled ? 1 : 0;
+        onUseFeedChanged: { updateFeedState(); }
```
zzmp: How come 75 uses brackets and 76 does not?

interface/resources/qml/AddressBarDialog.qml
```diff
@@ -45,50 +47,80 @@ Window {
...
             }
-            Card {
-                id: s2;
-                width: cardWidth;
-                height: cardHeight;
-                goFunction: goCard
+            highlightMoveDuration: -1;
+            highlightMoveVelocity: -1;
+            highlight: Rectangle { color: "transparent"; border.width: 4; border.color: "#1DB5ED"; z: 1; }
+            leftMargin: 50; // Start the first item over be about the same amount as the last item peeks through on the other side.
```
zzmp: s/be/by

howard-stearns:
Noted, and I agree that this is awkward, but it it the design for this round.
Other possibilities are (possibly in combinations);
- have the larger card come up on hover
- use larger cards to begin with
- make the larger cards swipable to the next, like a letterbox picture viewer
- have a specific button on the smaller card to go directly there, while clicking elsewhere on the small card brings up the larger one -- or vice versa.

The larger card will have more stuff in the next round, such as sharing to external social media, so we need the space.

Also, and this part isn't universally agreed yet, I feel that in the long term we won't have two different feeds that we switch between.  The places will be in amongst the other items in the feed, in score order. Thus there won't be two different mechanisms. (But of course, while that solves the different-behavior problem, it doesn't solve the don't-get-in-my-way-when-I-want-to-go-somewhere problem.

interface/resources/qml/AddressBarDialog.qml
howard-stearns: No good reason. (Historical.) will fix.

hifi-gustavo:
Build console output for commit: b3623da7725a140f0fbb3712bccdb48ea8ec6696 available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-b3623da.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-b3623da.dmg

interface/src/scripting/AccountScriptingInterface.cpp
```diff
@@ -26,6 +26,11 @@ bool AccountScriptingInterface::isLoggedIn() {
     return accountManager->isLoggedIn();
 }
 
+bool AccountScriptingInterface::checkAndSignalForAccessToken() {
```
zzmp: This (and `isLoggedIn`) could be inlined into the header file without breaking standard if you wanted to.

hifi-gustavo:
Build console output for commit: ed3b04f97ca6afbb118a437f72e7ac189dd1f255 available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-ed3b04f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-ed3b04f.dmg

interface/src/scripting/WindowScriptingInterface.cpp
```diff
@@ -203,3 +203,16 @@ void WindowScriptingInterface::copyToClipboard(const QString& text) {
     qDebug() << "Copying";
     QApplication::clipboard()->setText(text);
 }
+
+void WindowScriptingInterface::takeSnapshot(bool notify, float aspectRatio) {
+    // only evil-doers call takeSnapshot from a random thread
+    qApp->postLambdaEvent([notify, aspectRatio] {
```
zzmp: It would be better to expose a threadsafe method on Application, and only expose the thread-specific method as a private method, so that evil-doers are thwarted before they can begin.

interface/src/scripting/AccountScriptingInterface.cpp
howard-stearns: noted. I assume (but don't really know) that reason for existing pattern (isLoggedIn) is that someone including the .h shouldn't necessarily have to also include AccountManager.h.

Of course, in this particularly case, no one is ever going to include AccountScriptingInterface.h except for Application.cpp, and it already does include AccountManger.h. But I think that's the idea of not inlining it.

scripts/system/snapshot.js
```diff
@@ -0,0 +1,154 @@
...
+//
+// Created by David Kelly on 1 August 2016
+// Copyright 2016 High Fidelity, Inc
+//
+// Distributed under the Apache License, Version 2.0
+// See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+var SNAPSHOT_DELAY = 500; // 500ms
+var toolBar = Toolbars.getToolbar("com.highfidelity.interface.toolbar.system");
+var resetOverlays, recticleVisible;
```
zzmp: Everything should be declared on its own line, with its own `var`.

scripts/system/snapshot.js
```diff
@@ -0,0 +1,154 @@
...
+    imageURL: Script.resolvePath("assets/images/tools/snap.svg"),
+    visible: true,
+    buttonState: 1,
+    defaultState: 1,
+    hoverState: 2,
+    alpha: 0.9,
+});
+
+function shouldOpenFeedAfterShare() {
+    var persisted = Settings.getValue('openFeedAfterShare', true); // might answer true, false, "true", or "false"
```
zzmp: Why does a method that we wrote have such variance in return type? Can this be fixed?

scripts/system/snapshot.js
```diff
@@ -0,0 +1,154 @@
...
+    return persisted && (persisted !== 'false');
+}
+function showFeedWindow() {
+    DialogsManager.showFeed();
+}
+
+var outstanding;
```
zzmp: Move to top with other declarations.

scripts/system/snapshot.js
```diff
@@ -0,0 +1,154 @@
...
+function confirmShare(data) {
+    var dialog = new OverlayWebWindow('Snapshot Review', Script.resolvePath("html/SnapshotReview.html"), 800, 320);
+    function onMessage(message) {
+        // Receives message from the html dialog via the qwebchannel EventBridge. This is complicated by the following:
+        // 1. Although we can send POJOs, we cannot receive a toplevel object. (Arrays of POJOs are fine, though.)
+        // 2. Although we currently use a single image, we would like to take snapshot, a selfie, a 360 etc. all at the
+        //    same time, show the user all of them, and have the user deselect any that they do not want to share.
+        //    So we'll ultimately be receiving a set of objects, perhaps with different post processing for each.
+        var isLoggedIn, needsLogin = false;
```
zzmp: Two separate lines.

scripts/system/snapshot.js
```diff
@@ -0,0 +1,154 @@
...
+        switch (message) {
+        case 'ready':
```
zzmp: case should be indented.

scripts/system/snapshot.js
```diff
@@ -0,0 +1,154 @@
...
+            Desktop.show("hifi/dialogs/GeneralPreferencesDialog.qml", "GeneralPreferencesDialog");
+            break;
+        case 'setOpenFeedFalse':
+            Settings.setValue('openFeedAfterShare', false)
+            break;
+        case 'setOpenFeedTrue':
+            Settings.setValue('openFeedAfterShare', true)
+            break;
+        default:
+            dialog.webEventReceived.disconnect(onMessage); // I'm not certain that this is necessary. If it is, what do we do on normal close?
```
zzmp: Normal close doesn't need this because the script is destroyed, so any connections are disconnected.

scripts/system/snapshot.js
```diff
@@ -0,0 +1,154 @@
...
+            dialog.close();
+            dialog.deleteLater();
```
zzmp: This should be destroyed by the JS garbage collection without calling deleteLater. This could be dangerous code, and should be removed.

scripts/system/snapshot.js
howard-stearns: And spelling it correctly would be nice, too!
(I don't actually agree with the separate line generally -- and neither does jslint -- but for top level vars I don't object, so wilco.)

scripts/system/snapshot.js
howard-stearns: Currently assigned as a P2: https://highfidelity.fogbugz.com/f/cases/1514/Settings-getValue-of-boolean-gives-string-in-NEXT-session

scripts/system/snapshot.js
howard-stearns: I disagree on this one, feeling that when possible, things should be declared near where they are used.

scripts/system/snapshot.js
howard-stearns: I'm not following you. There's a new OverlayWebWindow dialog for each button push, which gets connected to webEventReceived at that time.  When the dialog is closed by the close button, the script does not end.

interface/src/scripting/WindowScriptingInterface.cpp
howard-stearns: I think maybe the (old) comment about evil-doers was trying to explain that the postLambdaEvent is perhaps not needed in practice, but there just for safety. Maybe the comment should be removed to avoid confusion.

But in any event, isn't the purpose of postLambdaEvent to make things threadsafe through the Application thread? Or am I confused?

hifi-gustavo:
Build console output for commit: ba72dc637510ffd624a6358d97e781e0f276b24d available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-ba72dc6.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-ba72dc6.dmg

interface/src/scripting/WindowScriptingInterface.cpp
```diff
@@ -205,7 +205,6 @@ void WindowScriptingInterface::copyToClipboard(const QString& text) {
 }
 
 void WindowScriptingInterface::takeSnapshot(bool notify, float aspectRatio) {
-    // only evil-doers call takeSnapshot from a random thread
     qApp->postLambdaEvent([notify, aspectRatio] {
```
zzmp: You are correct that `Application::postLambdaEvent` makes things threadsafe.

What I would like is for `Application::takeSnapshot` to be threadsafe, so that you can call it directly without an awkward call to `postLambdaEvent` and, more importantly, so that no one in the future unknowingly calls it from the wrong thread.
Could you reimplement it as a wrapper to `postLambdaEvent` that calls a `takeSnapshotInternal` to remove the possibility of bad threading?

I know this is a pattern that is already pervasive for consumers of `Application`, but that doesn't make it less dangerous, and I think it should be fixed as long as you are touching this part of the code.

scripts/system/snapshot.js
```diff
@@ -0,0 +1,155 @@
...
+                Desktop.show("hifi/dialogs/GeneralPreferencesDialog.qml", "GeneralPreferencesDialog");
+                break;
+            case 'setOpenFeedFalse':
+                Settings.setValue('openFeedAfterShare', false)
+                break;
+            case 'setOpenFeedTrue':
+                Settings.setValue('openFeedAfterShare', true)
+                break;
+            default:
+                dialog.webEventReceived.disconnect(onMessage); // I'm not certain that this is necessary. If it is, what do we do on normal close?
```
zzmp: It's necessary, as there can be multiple dialogues. You can remove the comment. I don't understand what the question is asking.

scripts/system/snapshot.js
```diff
@@ -0,0 +1,155 @@
...
+    button.writeProperty("defaultState", 1);
+    button.writeProperty("hoverState", 3);
+    Window.snapshotTaken.disconnect(resetButtons);
+
+    // last element in data array tells dialog whether we can share or not
+    confirmShare([ 
+        { localPath: path },
+        {
+            canShare: Boolean(location.placename),
+            isLoggedIn: true, // Just have the dialog act as though we are. To be removed at both ends later.
```
zzmp: Why is this necessary? It should just be removed, as it is checked at a later point.

scripts/system/snapshot.js
```diff
@@ -0,0 +1,155 @@
...
+    button.writeProperty("buttonState", 1);
+    button.writeProperty("defaultState", 1);
+    button.writeProperty("hoverState", 3);
+    Window.snapshotTaken.disconnect(resetButtons);
+
+    // last element in data array tells dialog whether we can share or not
+    confirmShare([ 
+        { localPath: path },
+        {
+            canShare: Boolean(location.placename),
```
zzmp: How does this work?

scripts/system/snapshot.js
howard-stearns: quite correct. that's dead code now.
(Originally, the .html displayed a message it was told that the user was not logged in.)

scripts/system/html/SnapshotReview.html
```diff
@@ -0,0 +1,242 @@
+<html>
+  <head>
+    <title>Share</title>
+    <link rel="stylesheet" type="text/css" href="css/edit-style.css">
+    <style type="text/css">
```
zzmp: This needs to be in a separate file. Put it in scripts/system/html/css/SnapshotReview.css.

scripts/system/snapshot.js
howard-stearns: I gather that Boolean() is there to make things clearer. :-)  (I was unfamiliar with it, myself.)

location.placename is a always a string, but may be the empty string (falsey) if the user is not in a shareable place -- e.g., if they teleported to an ip address, naked domain id, or currently, a temp place.   The SnapshotReview.html displays a message if canShare is not truthy.

scripts/system/html/SnapshotReview.html
```diff
@@ -0,0 +1,242 @@
...
...k6MSIKICAgICAgIGlkPSJjaXJjbGUxMyIKICAgICAgIHI9IjQuNDQwMDAwMSIKICAgICAgIGN5PSIxMjYuMTciCiAgICAgICBjeD0iMjAuNTQwMDAxIgogICAgICAgY2xhc3M9InN0MSIgLz48cGF0aAogICAgICAgc3R5bGU9ImZpbGw6I2ZmZmZmZjtmaWxsLW9wYWNpdHk6MSIKICAgICAgIGlkPSJwYXRoMTUiCiAgICAgICBkPSJtIDI4Ljg3LDEzOS4yNiBjIDAuMDEsLTAuMDEgMC4wMiwtMC4wMiAwLjAzLC0wLjAzIGwgMCwtMS44NiBjIDAsLTIuNjggLTIuMzMsLTQuNzcgLTUsLTQuNzcgbCAtNi40MiwwIGMgLTIuNjgsMCAtNC44NSwyLjA5IC00Ljg1LDQuNzcgbCAwLDEuODggMTYuMjQsMCB6IgogICAgICAgY2xhc3M9InN0MSIgLz48cGF0aAogICAgICAgc3R5bGU9ImZpbGw6I2ZmZmZmZjtmaWxsLW9wYWNpdHk6MSIKICAgICAgIGlkPSJwYXRoMTciCiAgICAgICBkPSJtIDM4LjE3LDEyMy40MiBjIDAsLTMuOTcgLTMuMjIsLTcuMTkgLTcuMTksLTcuMTkgbCAtMjAuMzEsMCBjIC0zLjk3LDAgLTcuMTksMy4yMiAtNy4xOSw3LjE5IGwgMCwxNC4xOCBjIDAsMy45NyAzLjIyLDcuMTkgNy4xOSw3LjE5IGwgMjAuMzEsMCBjIDMuOTcsMCA3LjE5LC0zLjIyIDcuMTksLTcuMTkgbCAwLC0xNC4xOCB6IG0gLTEuNzgsMTQuMjcgYyAwLDMuMDMgLTIuNDYsNS40OSAtNS40OSw1LjQ5IGwgLTIwLjMyLDAgYyAtMy4wMywwIC01LjQ5LC0yLjQ2IC01LjQ5LC01LjQ5IGwgMCwtMTQuMTkgYyAwLC0zLjAzIDIuNDYsLTUuNDkgNS40OSwtNS40OSBsIDIwLjMzLDAgYyAzLjAzLDAgNS40OSwyLjQ2IDUuNDksNS40OSBsIDAsMTQuMTkgeiIKICAgICAgIGNsYXNzPSJzdDEiIC8+PC9nPjxnCiAgICAgdHJhbnNmb3JtPSJ0cmFuc2xhdGUoMCwtMTEwKSIKICAgICBpZD0iTGF5ZXJfMiIgLz48L3N2Zz4=);
```
zzmp: This should not be inlined if you can help it. What is it?

scripts/system/html/SnapshotReview.html
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        input[type=button].naked:hover {
+            color: #00b4ef;
+            background: none;
+        }
+    </style>
+    <script type="text/javascript" src="qrc:///qtwebchannel/qwebchannel.js"></script>
+    <script type="text/javascript" src="js/eventBridgeLoader.js"></script>
+    <script>
```
zzmp: This needs to be in a separate file. Put it in scripts/system/html/js/SnapshotReview.js.

scripts/system/html/SnapshotReview.html
```diff
@@ -0,0 +1,242 @@
...
+    var paths = [], idCounter = 0, useCheckboxes;
+    function addImage(data) {
+        var div = document.createElement("DIV"),
```
zzmp: Please use jQuery. It is much easier to read, and already included in our application.

scripts/system/html/SnapshotReview.html
```diff
@@ -0,0 +1,242 @@
...
+            input = document.createElement("INPUT"),
+            label = document.createElement("LABEL"),
+            img = document.createElement("IMG"),
+            id = "p" + idCounter++;
+        function toggle() { data.share = input.checked; }
+        img.src = data.localPath;
+        div.appendChild(img);
+        data.share = true;
+        if (useCheckboxes) { // I'd rather use css, but the included stylesheet is quite particular.
+            // Our stylesheet(?) requires input.id to match label.for. Otherwise input doesn't display the check state.
```
zzmp: Why not do that?

interface/src/scripting/WindowScriptingInterface.cpp
davidkelly: After speaking with @zzmp, postLambdaEvent is the right thing to do.  But, it is his preference that it is unnecessary and that the function in `Application` instead do a repost itself to the main thread.  That means someone else using it doesn't have to know that it can only be called from the main thread.

That said, this pattern exists in a lot of places so maybe we can make that change after getting this in?  I don't want to hold things up for this, unless it is fairly important.

scripts/system/html/SnapshotReview.html
```diff
@@ -0,0 +1,242 @@
...
+                
+                // rest are image paths which we add
+                useCheckboxes = message.length > 1;
+                message.forEach(addImage);
+            });
+            EventBridge.emitWebEvent('ready');
+         });
+
+    };
+    // beware of bug: Cannot send objects at top level. (Nested in arrays is fine.)
```
zzmp: FYI the entityList gets around this by sending JSON instead of arrays, so that it can still use top-level objects.

plugins/openvr/src/OpenVrHelpers.cpp
```diff
@@ -187,6 +186,10 @@ void finishOpenVrKeyboardInput() {
     //QInputMethodEvent event(_existingText, QList<QInputMethodEvent::Attribute>());
     //event.setCommitString(newText, 0, _existingText.size());
     //qApp->sendEvent(_keyboardFocusObject, &event);
+}
+void finishOpenVrKeyboardInput() {
```
zzmp: Line between functions.

scripts/system/snapshot.js
zzmp: Is this just checking to see if it is there?
FYI, JS convention is to just use implicit casting: `!!location.placename` would be fine.

interface/src/ui/Snapshot.h
```diff
@@ -32,6 +32,16 @@ class SnapshotMetaData {
     QUrl _URL;
 };
 
+
+class SnapshotUploader: public QObject {
```
zzmp: Coding standard: This should have its own file.

interface/src/ui/Snapshot.cpp
```diff
@@ -141,3 +142,91 @@ QFile* Snapshot::savedFileForSnapshot(QImage & shot, bool isTemporary) {
...
+        emit DependencyManager::get<WindowScriptingInterface>()->snapshotShared(contents);
+    }
+}
+
+void SnapshotUploader::uploadFailure(QNetworkReply& reply) {
+    emit DependencyManager::get<WindowScriptingInterface>()->snapshotShared(reply.readAll());
+}
+
+void SnapshotUploader::createStorySuccess(QNetworkReply& reply) {    
+    emit DependencyManager::get<WindowScriptingInterface>()->snapshotShared("");
```
zzmp: Use `QString()` for an empty string.

interface/src/ui/Snapshot.cpp
```diff
@@ -141,3 +142,91 @@ QFile* Snapshot::savedFileForSnapshot(QImage & shot, bool isTemporary) {
...
+}
+
+void SnapshotUploader::uploadSuccess(QNetworkReply& reply) {
+    const QString STORY_UPLOAD_URL = "/api/v1/user_stories";
+    static SnapshotUploader uploader;
+
+    // parse the reply for the thumbnail_url
+    QByteArray contents = reply.readAll();
+    QJsonParseError jsonError;
+    auto doc = QJsonDocument::fromJson(contents, &jsonError);
```
zzmp: You can make this cleaner with https://github.com/highfidelity/hifi/pull/8456/commits/a9ae869ad4df58495fda450a2471aac0756b6948 if you want to. It will go in with #8456. Definitely not necessary.

interface/src/scripting/WindowScriptingInterface.cpp
```diff
@@ -203,3 +203,15 @@ void WindowScriptingInterface::copyToClipboard(const QString& text) {
     qDebug() << "Copying";
     QApplication::clipboard()->setText(text);
 }
+
+void WindowScriptingInterface::takeSnapshot(bool notify, float aspectRatio) {
```
zzmp: These can both be refactored to avoid threading issues.

interface/src/scripting/DesktopScriptingInterface.h
```diff
@@ -23,6 +23,7 @@ class DesktopScriptingInterface : public QObject, public Dependency {
 
 public:
     Q_INVOKABLE void setOverlayAlpha(float alpha);
+    Q_INVOKABLE void show(const QString& path, const QString&  title);
```
zzmp: If I had my way Q_INVOKABLE would all have to be threadsafe, same reason as the Application methods. But I don't think its a big deal here.

scripts/system/html/SnapshotReview.html
howard-stearns: Indeed, it was his design -- immediately before going on vacation. We can change that after he gets back.

scripts/system/html/SnapshotReview.html
howard-stearns: Because we don't have the in-house css expertise.  The existing (to my mind, weird) checkbox definitions are what we use within edit.js, and we don't have good turnaround on changing that across the board.

scripts/system/html/SnapshotReview.html
howard-stearns: That's a thought. Will consider for future.

scripts/system/snapshot.js
howard-stearns: I agree, but once got called out for that. I'll take this as an opportunity to do it this way with blessing.

scripts/system/html/SnapshotReview.html
howard-stearns: maybe next time. (I dislike jQuery.)

hifi-gustavo:
Build console output for commit: f9d8ccaeac0a260688619ac658803e55b0e3524c available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-f9d8cca.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-f9d8cca.dmg

interface/src/ui/Snapshot.cpp
howard-stearns: Want! (next time)

scripts/system/html/css/SnapshotReview.css
```diff
@@ -0,0 +1,131 @@
+/*
+//  edit-style.css
```
zzmp: Copypasta

howard-stearns: doh!

interface/resources/qml/hifi/Card.qml
```diff
@@ -17,28 +17,68 @@ import QtGraphicalEffects 1.0
...
+    property int onlineUsers: 0;
+    property bool isUserStory: userName && !onlineUsers;
+
     property int textPadding: 20;
     property int textSize: 24;
-    property string defaultPicture: "../../images/default-domain.gif";
+    property int textSizeSmall: 18;
+    property string defaultThumbnail: Qt.resolvedUrl("../../images/default-domain.gif");
     HifiConstants { id: hifi }
+    function pastTime(timestamp) { // Answer a descriptive string
```
zzmp: newline before new function

interface/resources/qml/hifi/Card.qml
```diff
@@ -17,28 +17,68 @@ import QtGraphicalEffects 1.0
...
+        source: thumbnail || defaultThumbnail;
         fillMode: Image.PreserveAspectCrop;
         // source gets filled in later
         anchors.verticalCenter: parent.verticalCenter;
         anchors.left: parent.left;
         onStatusChanged: {
             if (status == Image.Error) {
-                console.log("source: " + source + ": failed to load " + JSON.stringify(userStory));
-                source = defaultPicture;
+                console.log("source: " + source + ": failed to load " + hfiUrl);
```
zzmp: s/hfiUrl/hifiUrl

interface/resources/qml/hifi/Card.qml
zzmp: I think it'd be better to use a library for this, but I won't push it.

hifi-gustavo:
Build console output for commit: bfb7d728adf167179a0c2b96dbc44cac3eda61f6 available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-bfb7d72.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-bfb7d72.dmg

interface/resources/qml/hifi/Card.qml
```diff
@@ -79,18 +120,24 @@ Rectangle {
...
     }
+    property var hoverThunk: function () { };
+    property var unhoverThunk: function () { };
     MouseArea {
+        id: zmouseArea;
         anchors.fill: parent;
         acceptedButtons: Qt.LeftButton;
         onClicked: goFunction(parent);
         hoverEnabled: true;
+        onEntered: hoverThunk();
```
zzmp: What are these thunks for? You should add a comment - I'm assuming they coerce the entered/exited behavior into working.

interface/resources/qml/AddressBarDialog.qml
```diff
@@ -200,133 +270,217 @@ Window {
...
-        array.forEach(function (element) {
-            iterator(element, icb);
+        array.forEach(function (element, index) {
+            if (count < 0) { // don't keep iterating after we short-circuit
+                return;
+            }
+            iterator(element, function (error, mapped) {
+                results[index] = mapped;
+                if (error || !--count) {
+                    count = 1; // don't cb multiple times if error
```
zzmp: This should be `count = 0;`

interface/resources/qml/AddressBarDialog.qml
```diff
@@ -200,133 +270,217 @@ Window {
...
-    // short-circuits if error. Note that iterator MUST be an asynchronous function. (Use setTimeout if necessary.)
-    function asyncEach(array, iterator, cb) {
-        var count = array.length;
-        function icb(error) {
-            if (!--count || error) {
-                count = -1; // don't cb multiple times (e.g., if error)
-                cb(error);
-            }
-        }
+    function asyncMap(array, iterator, cb) {
```
zzmp: Why not use a library?

interface/resources/qml/AddressBarDialog.qml
```diff
@@ -200,133 +270,217 @@ Window {
...
+    }
+    function getDomainPage(pageNumber, cb) { // cb(error) after all pages of domain data have been added to model
+        // Each page of results is processed completely before we start on the next page.
+        // For each page of domains, we process each domain in parallel, and for each domain, process each place name in parallel.
+        // This gives us minimum latency within the page, but we do preserve the order within the page by using asyncMap and
+        // only appending the collected results.
+        var params = [
+            'open', // published hours handle now
+            // FIXME: should determine if place is actually running
+            'restriction=open', // Not by whitelist, etc.  FIXME: If logged in, add hifi to the restriction options, in order to include places that require login.
```
zzmp: Should these FIXMEs be in here?

interface/resources/qml/hifi/Card.qml
zzmp: I found where you used these, no need for comments.

interface/resources/qml/AddressBarDialog.qml
howard-stearns: Because that would quickly become a company-wide argument about promises vs callbacks.

zzmp: promises.

interface/resources/qml/AddressBarDialog.qml
howard-stearns: Well, until the domains and places api provides a mechanism to make such stuff available, yes...

interface/resources/qml/AddressBarDialog.qml
howard-stearns: good catch

hifi-gustavo:
Build console output for commit: f53ce94b200918999d8bb5beee74cd76f55931d7 available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-f53ce94.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-f53ce94.dmg

hifi-gustavo:
Build console output for commit: 7a9e8329597d8c66d2e6b26f0ace722461227dd2 available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-7a9e832.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-7a9e832.dmg

hifi-gustavo:
Build console output for commit: 97c706f4cbf0e45b4f18949ceff6bf8302858c5c available at: https://gist.github.com/11556053f9f14d74d0afadc0855c8646<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-97c706f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8428-97c706f.dmg

PR #8441, opened 2016-08-16
Better create locations for Entities

imgntn:
This PR makes it so that entities created using edit.js or imported from marketplace have better positions.

Previously, when you would import some objects from the marketplace they would end up in unpredictable locations or you would end up inside of them.  Now, you should never end up inside of an entity that you import.  Primitives should be created in predictable locations irrespective of the camera orientation.

To test:
1. Open edit.js
2. Import a cube.  Observe that it is about 1 meter in front of your avatar position about about a half meter above.
3.  Point your camera at the sky and repeat #2, cube should end up in the same place
4. Open up the marketplace and import "Dungeon Floor 6m x 6m"
5. Note that you are not inside of the floor.  In the release build, you end up inside of this and it jitters.
6. Test with some other marketplace items.

### DIFF
--- a/scripts/system/edit.js
+++ b/scripts/system/edit.js
@@ -1129,36 +1129,37 @@ function handeMenuEvent(menuItem) {
     tooltip.show(false);
 }
 
-// This function tries to find a reasonable position to place a new entity based on the camera
-// position. If a reasonable position within the world bounds can't be found, `null` will
-// be returned. The returned position will also take into account grid snapping settings.
 function getPositionToCreateEntity() {
-    var distance = cameraManager.enabled ? cameraManager.zoomDistance : DEFAULT_ENTITY_DRAG_DROP_DISTANCE;
-    var direction = Quat.getFront(Camera.orientation);
-    var offset = Vec3.multiply(distance, direction);
-    var placementPosition = Vec3.sum(Camera.position, offset);
-
-    var cameraPosition = Camera.position;
-
     var HALF_TREE_SCALE = 16384;
-
-    var cameraOutOfBounds = Math.abs(cameraPosition.x) > HALF_TREE_SCALE || Math.abs(cameraPosition.y) > HALF_TREE_SCALE ||
-                            Math.abs(cameraPosition.z) > HALF_TREE_SCALE;
-    var placementOutOfBounds = Math.abs(placementPosition.x) > HALF_TREE_SCALE ||
-                               Math.abs(placementPosition.y) > HALF_TREE_SCALE ||
-                               Math.abs(placementPosition.z) > HALF_TREE_SCALE;
-
-    if (cameraOutOfBounds && placementOutOfBounds) {
-        return null;
+    var direction = Quat.getFront(MyAvatar.orientation);
+    var distance = 1;
+    var position = Vec3.sum(MyAvatar.position, Vec3.multiply(direction, distance));
+    position.y +=0.5;
+    if (position.x > HALF_TREE_SCALE || position.y > HALF_TREE_SCALE || position.z > HALF_TREE_SCALE) {
+        return null
     }
-
-    placementPosition.x = Math.min(HALF_TREE_SCALE, Math.max(-HALF_TREE_SCALE, placementPosition.x));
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 1632f4cb3d448281bd6ca5134b286bbec67bf581 available at: https://gist.github.com/b6151b91bc12d8e3d91788ab06e88dff<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8441-1632f4c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8441-1632f4c.dmg

Nex-Pro:
Tested on Win10

In general this is working well but after trying a number of objects from the market place I got the Nevada moon rocks and it appeared approx 600 meters from my position. ( The zone containing it isn't that large.) If the rez distance is a function of the bounding box dimensions then this might need tweaking for larger objects.  It looks to me as if it is based on the view angle of the bounding box.

scripts/system/edit.js
```diff
@@ -1129,36 +1129,37 @@ function handeMenuEvent(menuItem) {
...
+    var HALF_TREE_SCALE = 16384;
+    var direction = Quat.getFront(MyAvatar.orientation);
+    var distance = 1;
+    if (dimensions.x > distance) {
+        distance = dimensions.x
+    }
+    if (dimensions.z > distance) {
+        distance = dimensions.z
     }
+    var position = Vec3.sum(MyAvatar.position, Vec3.multiply(direction, distance));
```
huffman: The previous implementation used the camera position because you might be in independent / orbit camera mode. This will change that behavior dramatically, thoughts?

Nex-Pro:
@huffman My gut feeling as a 3d modeller is that there should be some user options for this. This is good in as much as it is fairly predictable for most situations. But maybe the proximity should be scaled exponentially in some way, inverse square or some kind of log function.

With hand controllers I'd like to be able to shoot a spot on a surface and have the object rez there. Some kind of raycast and get back the UV of the ray intersection.

huffman:
@Nex-Pro Yeah, I like the idea of snapping to surfaces. This was actually something we talked about when originally working on edit.js, but back then we didn't have support for picking against polys. This would certainly be a good addition.

imgntn:
@huffman if you're in independent or entity mode, now it will use the camera position + orientation
@Nex-Pro thanks, i made some small adjustments, but the dimensions of that moon rocks group is 1200x1200x1200.   that thing is inside of a huge zone, so its going to end up a bit far away.  its still visible and you can move it closer if you want.  other objects seem to do well

hifi-gustavo:
Build console output for commit: 06787029b75abf97e5677dd5908a34de8ea66168 available at: https://gist.github.com/b6151b91bc12d8e3d91788ab06e88dff<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8441-0678702.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8441-0678702.dmg

hifi-gustavo:
Build console output for commit: e9c2a1d509671769247949e83f3d72d23ed25614 available at: https://gist.github.com/b6151b91bc12d8e3d91788ab06e88dff<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8441-e9c2a1d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8441-e9c2a1d.dmg

hifi-gustavo:
Build console output for commit: 0810ad0e7b0f66bb7cf53d9131cfb068fd5ff169 available at: https://gist.github.com/b6151b91bc12d8e3d91788ab06e88dff<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8441-0810ad0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8441-0810ad0.dmg

PR #8444, opened 2016-08-16
fix flashing in default skybox, and allow skybox lighting without textures

zzmp:
- Fixes flashing on a default skybox
- Fixes skybox behavior to allow keylights without a skybox texture
- Removes skybox functionality to revert to previous skybox, as it may no longer exist

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -4253,6 +4253,21 @@ namespace render {
         auto backgroundMode = skyStage->getBackgroundMode();
 
         switch (backgroundMode) {
+            case model::SunSkyStage::SKY_DEFAULT: {
+                static const glm::vec3 DEFAULT_SKYBOX_COLOR{ 255.0f / 255.0f, 220.0f / 255.0f, 194.0f / 255.0f };
+                static const float DEFAULT_SKYBOX_INTENSITY{ 0.2f };
+                static const float DEFAULT_SKYBOX_AMBIENT_INTENSITY{ 2.0f };
+                static const glm::vec3 DEFAULT_SKYBOX_DIRECTION{ 0.0f, 0.0f, -1.0f };
+
+                auto scene = DependencyManager::get<SceneScriptingInterface>()->getStage();
+                auto sceneKeyLight = scene->getKeyLight();
+                scene->setSunModelEnable(false);
+                sceneKeyLight->setColor(DEFAULT_SKYBOX_COLOR);
+                sceneKeyLight->setIntensity(DEFAULT_SKYBOX_INTENSITY);
+                sceneKeyLight->setAmbientIntensity(DEFAULT_SKYBOX_AMBIENT_INTENSITY);
+                sceneKeyLight->setDirection(DEFAULT_SKYBOX_DIRECTION);
+                // fall through: render a skybox, if available
+           }
             case model::SunSkyStage::SKY_BOX: {
                 auto skybox = skyStage->getSkybox();
                 if (skybox) {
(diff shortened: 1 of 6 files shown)


zzmp:
QA

- Create a default skybox, and turn in it.
  - In master, you will see flashes of full lighting - a bug
  - This PR should fix that
- Create a default skybox, and note the lighting
  - In master, a skybox without a texture will have default lighting
  - This PR should fix that, and a skybox can now define lighting while still using the default skybox texture
- Ensure other skyboxes continue to function
- Ensure that when passing through multiple skyboxes, you see the transitions that you would expect, and each skybox acts as coded
  - Test with a skybox with only lighting (default)
  - Test with a skybox with only ambient texture
  - Test with a skybox with a texture

hifi-gustavo:
Build console output for commit: 8c559bc65d7b1904d8de80153030d1a10f324695 available at: https://gist.github.com/a5fe2aaabf3e27c36bdefb92907f8143<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8444-8c559bc.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8444-8c559bc.dmg

hifi-gustavo:
Build console output for commit: d6d9871a9692870d99c7377752a7f080263878c6 available at: https://gist.github.com/a5fe2aaabf3e27c36bdefb92907f8143<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8444-d6d9871.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8444-d6d9871.dmg

Nex-Pro:
**PR** **#8444**

**CPU:** 4.4 ghz Intel i7
**RAM:** 16 GB
**GPU:** Nvidia gtx 1070 8gb
**Testing OS:** win 10 pro
**HMD:** HTC Vive.
**Hand Controllers:** Vive
**Leap Motion:** Not used

#### PASS

Create a default skybox, and turn in it
In master, you will see flashes of full lighting - a bug
This PR should fix that: **SUCCESS**

Create a default skybox, and note the lighting : **CHECK**

In master, a skybox without a texture will have default lighting : **CHECK**

This PR should fix that, and a skybox can now define lighting while still using the default skybox texture
Ensure other skyboxes continue to function : **SUCCESS**

Ensure that when passing through multiple skyboxes, you see the transitions that you would expect, and each skybox acts as coded: **SUCCESS**

Test with a skybox with only lighting (default) : **CHECK**
Test with a skybox with only ambient texture (you'll need a mirrored surface to test this) : **CHECK**
Test with a skybox with a texture : **CHECK**

Nex-Pro:
@zzmp ^

interface/src/Application.cpp
```diff
@@ -4253,40 +4253,44 @@ namespace render {
...
+                    sceneKeyLight->setAmbientMap(defaultSkyboxAmbientTexture);
+                }
+                // fall through: render defaults, if available
             }
-                break;
-
+            case model::SunSkyStage::SKY_DEFAULT_TEXTURE:
+                if (Menu::getInstance()->isOptionChecked(MenuOption::DefaultSkybox)) {
+                    qApp->getDefaultSkybox()->render(batch, args->getViewFrustum());
+                }
```
Atlante45: Aren't the last 2 cases missing a `break` inside their `if`?

This switch case is really confusing, everything falls through and ends up in the default case.
Isn't there a better way to write that?

I assume this is to reduce code duplication, but I think it'd be clearer to put those blocks in several functions and call those function in each block accordingly with no fall through.

zzmp: The last two cases are missing `break`s, but they don't do anything. I can add those.
I want everything to fall through, I thought that was the best use case for a switch statement - nearly every case does D, most cases, do C, some do B, few do A, so use `switch` to enter the codepath at the point where you start, A, B, C, D. I don't think its worth it to put it all in separate functions just to avoid a `switch`, when a switch does what I want it to do.

zzmp: Sorry, no `breaks` are missing. I want every case to fall through.
SKY_DEFAULT_TEXTURE includes NO_BACKGROUND.
SKY_AMBIENT_TEXTURE includes SKY_DEFAULT_TEXTURE.
SKY_BOX includes SKY_AMBIENT_TEXTURE if it is not set.
SKY_DEFAULT includes it all.

PR #8456, opened 2016-08-17
add an entity property for procedural shaders

zzmp:


libraries/entities/src/EntityItem.cpp
```diff
@@ -1204,6 +1211,19 @@ EntityItemProperties EntityItem::getProperties(EntityPropertyFlags desiredProper
     COPY_ENTITY_PROPERTY_TO_PROPERTIES(clientOnly, getClientOnly);
     COPY_ENTITY_PROPERTY_TO_PROPERTIES(owningAvatarID, getOwningAvatarID);
 
+    // migrate proceduralData to its own property
```
zzmp: On getting properties for an entity, migrate `userData::ProceduralEntity` to `proceduralData` if the latter is unset.

libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -726,6 +731,17 @@ void EntityItemProperties::copyFromScriptValue(const QScriptValue& object, bool
     COPY_PROPERTY_FROM_QSCRIPTVALUE(clientOnly, bool, setClientOnly);
     COPY_PROPERTY_FROM_QSCRIPTVALUE(owningAvatarID, QUuid, setOwningAvatarID);
 
+    // migrate proceduralData to its own property
```
zzmp: On setting properties for an entity, migrate `userData::ProceduralEntity` to `proceduralData` if the latter is unchanged.

hifi-gustavo:
Build console output for commit: 991b31f3b05bcc3772858898b16748b2a87f8403 available at: https://gist.github.com/5fcfd2fe7a0c6acc2c7b1aab26f0477e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8456-991b31f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8456-991b31f.dmg

libraries/entities/src/EntityItemProperties.cpp
ZappoMan: I don't think you need to do both these steps... the first part in EntityItem would work. But it doesn't hurt to do both...

zzmp:
@ZappoMan Should I worry about deprecating? I am worried that this might flood the log, so I didn't commit it.
```
diff --git a/libraries/entities/src/EntityItem.cpp b/libraries/entities/src/EntityItem.cpp
index 18399c8..ab6de2f 100644
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1219,6 +1219,10 @@ EntityItemProperties EntityItem::getProperties(EntityPropertyFlags desiredProper

         static const QString PROCEDURAL_USER_DATA_KEY = "ProceduralEntity";
         if (parsedUserData.contains(PROCEDURAL_USER_DATA_KEY)) {
+            static const QString proceduralEntityDeprecationNotice =
+                "userData's ProceduralEntity property is deprecated, please use proceduralData (entity %1)";
+            qWarning() << proceduralEntityDeprecationNotice.arg(_id.toString());
+
             auto parsedProceduralData = parsedUserData.take(PROCEDURAL_USER_DATA_KEY);
             properties.setProceduralData(JSON::stringify(parsedProceduralData));
         }
```

libraries/entities/src/EntityItemProperties.cpp
zzmp: I think that it will interact badly with the entityList.js unless this is here, as it won't report back if it changes the proceduralData.

ZappoMan: Let me think about that... but I think that since you handle it in the EntityItem::setProperties() you'll end up replacing all of the old userData with proceduralData on server load, and the server will only send to clients entities with the new properties.

ZappoMan:
I marked this as DO NOT MERGE until after we've done some demos on Friday.

hifi-gustavo:
Build console output for commit: ca5b23ca0c5e14fed94b8af18bafd0b9875a4976 available at: https://gist.github.com/5fcfd2fe7a0c6acc2c7b1aab26f0477e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8456-ca5b23c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8456-ca5b23c.dmg

Nex-Pro:
This has shown some odd behaviour.

I grabbed the Day/Night cycle skybox from the market and it didn't work
I then got the Ocean skybox from the market. The ocean procedural worked fine.
I deleted the Ocean skybox at which point the previously non-working day/night cycle skybox began to work. Issues discussed with Zach in slack.

zzmp:
stale, needs work, closing until that is done.

ZappoMan:
:(

PR #8462, opened 2016-08-17
Address issues with random camera locations 

jherico:
... caused by bad HMD sensor poses

I've been seeing issues where the avatar viewpoint is at some random location, with the X, Y or Z position being something really random like 3.015e30.  This completely breaks rendering, even of the default skybox and all you get is a black screen.

The exact cause of this is either uninitialized memory, or a race condition.  This PR addresses both possible causes by 
* initializing the pose data matrix to a zeroed out value
* doing more examination of the pose data validity during processing in the OpenVR plugin
* dropping obviously bad HMD sensor poses in the application code
* protecting access to the pose data with a mutex in the OpenVR `beginFrameRender` method, which was previously using a value that could potentially change underneath it.  

## Testing

Before each run, remove or rename the `address` value in the `AddressManager` section of the `interface.ini` file.  Start the application while already in OpenVR mode.  

In the current production build, you will occasionally end up with a black screen.  If you switch to desktop mode and have the stats display visible (using the `/` key to toggle) you will see the avatar position is at some random location, possibly light years away from the domain origin.

In this build you should end up in the default domain (playa) with a reasonable location 100% of the time.

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -516,13 +516,23 @@ glm::mat4 MyAvatar::getSensorToWorldMatrix() const {
     return _sensorToWorldMatrixCache.get();
 }
 
+// As far as I know no HMD system supports a play area of a kilometer in radius.  
+static const float MAX_HMD_ORIGIN_DISTANCE = 1000.0f;
 // Pass a recent sample of the HMD to the avatar.
 // This can also update the avatar's position to follow the HMD
 // as it moves through the world.
 void MyAvatar::updateFromHMDSensorMatrix(const glm::mat4& hmdSensorMatrix) {
     // update the sensorMatrices based on the new hmd pose
     _hmdSensorMatrix = hmdSensorMatrix;
-    _hmdSensorPosition = extractTranslation(hmdSensorMatrix);
+    auto newHmdSensorPosition = extractTranslation(hmdSensorMatrix);
+
+    if (newHmdSensorPosition != _hmdSensorPosition && 
+        glm::length(newHmdSensorPosition) > MAX_HMD_ORIGIN_DISTANCE) {
+        qWarning() << "Invalid HMD sensor position " << newHmdSensorPosition;
+        // Ignore unreasonable HMD sensor data
+        return;
+    }
+    _hmdSensorPosition = newHmdSensorPosition;
     _hmdSensorOrientation = glm::quat_cast(hmdSensorMatrix);
     _hmdSensorFacing = getFacingDir2D(_hmdSensorOrientation);
 }
--- a/plugins/openvr/src/OpenVrDisplayPlugin.cpp
+++ b/plugins/openvr/src/OpenVrDisplayPlugin.cpp
@@ -108,19 +108,20 @@ public:
     }
 
     void updateSource() {
-        Lock lock(_plugin._presentMutex);
-        while (!_queue.empty()) {
-            auto& front = _queue.front();
-            auto result = glClientWaitSync(front.fence, 0, 0);
-            if (GL_TIMEOUT_EXPIRED == result && GL_WAIT_FAILED == result) {
-                break;
-            }
+        _plugin.withNonPresentThreadLock([&] {
+            while (!_queue.empty()) {
+                auto& front = _queue.front();
+                auto result = glClientWaitSync(front.fence, 0, 0);
+                if (GL_TIMEOUT_EXPIRED == result && GL_WAIT_FAILED == result) {
+                    break;
+                }
 
-            glDeleteSync(front.fence);
-            front.fence = 0;
-            _current = front;
-            _queue.pop();
-        }
+                glDeleteSync(front.fence);
+                front.fence = 0;
+                _current = front;
+                _queue.pop();
+            }
+        });
     }
 
     void run() override {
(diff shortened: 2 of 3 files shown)


interface/src/avatar/MyAvatar.cpp
```diff
@@ -516,13 +516,23 @@ glm::mat4 MyAvatar::getSensorToWorldMatrix() const {
     return _sensorToWorldMatrixCache.get();
 }
 
+// As far as I know no HMD system supports a play area of a kilometer in radius.
```
Atlante45: This is gonna come back and bite us 10 years from now. I can feel it!

hifi-gustavo:
Build console output for commit: a32ab77b7dad3a9eaac11571584a7c3940835af2 available at: https://gist.github.com/eb5c9163077db6eee2865c8ba8a132ba<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8462-a32ab77.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8462-a32ab77.dmg

interface/src/avatar/MyAvatar.cpp
jherico: I _actually_ considered putting in a `FIXME re-evaluate in 2020` line in there.

plugins/openvr/src/OpenVrDisplayPlugin.cpp
```diff
@@ -170,15 +171,28 @@ class OpenVrSubmitThread : public QThread, public Dependency {
...
                 vr::VRCompositor()->WaitGetPoses(nextRender.vrPoses, vr::k_unMaxTrackedDeviceCount, nextSim.vrPoses, vr::k_unMaxTrackedDeviceCount);
-                {
-                    Lock lock(_plugin._presentMutex);
-                    _presentCount++;
-                    _presented.notify_one();
+
+                // Copy invalid poses in nextSim from nextRender
+                for (uint32_t i = 0; i < vr::k_unMaxTrackedDeviceCount; ++i) {
+                    if (!nextSim.vrPoses[i].bPoseIsValid) {
+                        nextSim.vrPoses[i] = nextRender.vrPoses[i];
```
Atlante45: I don't understand why you're copying the invalid values here?

plugins/openvr/src/OpenVrDisplayPlugin.cpp
```diff
@@ -366,19 +380,20 @@ bool OpenVrDisplayPlugin::beginFrameRender(uint32_t frameIndex) {
     }
     _currentRenderFrameInfo = FrameInfo();
 
+    PoseData nextSimPoseData;
     withNonPresentThreadLock([&] {
-        _currentRenderFrameInfo.renderPose = _nextSimPoseData.poses[vr::k_unTrackedDeviceIndex_Hmd];
```
Atlante45: Is this not needed anymore?
I see that there was a duplicate underneath.

plugins/openvr/src/OpenVrHelpers.h
```diff
@@ -66,8 +66,15 @@ struct PoseData {
     vec3 linearVelocities[vr::k_unMaxTrackedDeviceCount];
     vec3 angularVelocities[vr::k_unMaxTrackedDeviceCount];
 
+    PoseData() {
+        memset(vrPoses, 0, sizeof(vr::TrackedDevicePose_t) * vr::k_unMaxTrackedDeviceCount);
+    }
```
Atlante45: Since `vr::TrackedDevicePose_t`, it should get zero initialized.
Maybe this is required though:
`vr::TrackedDevicePose_t vrPoses[vr::k_unMaxTrackedDeviceCount]{ };`

Anyway, not blocking, at least this way we know for sure.

plugins/openvr/src/OpenVrDisplayPlugin.cpp
jherico: I'm copying from `nextRender` over the invalid poses in `nextSim`.

plugins/openvr/src/OpenVrHelpers.h
jherico: being a struct doesn't give you default initialization.  Only having initializers or a constructor gives you reliable values for a fresh object.  While `TrackedDevicePose_t foo {}` may give you proper values, I'm not sure that `vr::TrackedDevicePose_t vrPoses[vr::k_unMaxTrackedDeviceCount]{ };` would since the `{}` would be interpreted as an array initializer of the wrong length.

plugins/openvr/src/OpenVrDisplayPlugin.cpp
Atlante45: Duh! Thanks

Nex-Pro:
#### PASS

pr: #8462

**CPU:** 4.4 ghz Intel i7
**RAM:** 16 GB
**GPU:** Nvidia gtx 1070 8gb
**Testing OS:** win 10 pro
**HMD:** HTC Vive.
**Hand Controllers:** Vive
**Leap Motion:** None

TESTING: Before each run, remove or rename the address value in the AddressManager section of the interface.ini file. Start the application while already in OpenVR mode. **CHECK**

**Many runs on this over a period of hours with no issues. Each time the address data was observed and removed from interface.ini  and in every case where the address data was removed my avatar arrived at playa.**

PR #8485, opened 2016-08-20
Move steam.js to unpublishedScripts

Atlante45:


cozza13:
I requested this and have tested it.

PR #8497, opened 2016-08-22
Raised max iterations for IK to 16 from 4

hyperlogic:
* IK now returns early if solution is "good enough"
* IK now has nsight markers for profiling

As a result, time spent in the IK system, when the solution cannot be reached, has been increased from 200μs to 500μs

### DIFF
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -146,7 +146,6 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
     }
 
     float maxError = FLT_MAX;
-    int HACK_MAX_ERROR_JOINT = -1;
     int numLoops = 0;
     const int MAX_IK_LOOPS = 16;
     const float MAX_ERROR_TOLERANCE = 0.1f; // cm
@@ -180,13 +179,12 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
 
         // compute maxError
         maxError = 0.0f;
-        for (int i = 0; i < targets.size(); i++) {
+        for (size_t i = 0; i < targets.size(); i++) {
             if (targets[i].getType() == IKTarget::Type::RotationAndPosition || targets[i].getType() == IKTarget::Type::HmdHead ||
                 targets[i].getType() == IKTarget::Type::HipsRelativeRotationAndPosition) {
                 float error = glm::length(absolutePoses[targets[i].getIndex()].trans - targets[i].getTranslation());
                 if (error > maxError) {
                     maxError = error;
-                    HACK_MAX_ERROR_JOINT = targets[i].getIndex();
                 }
             }
         }
--- a/libraries/animation/src/AnimInverseKinematics.cpp
+++ b/libraries/animation/src/AnimInverseKinematics.cpp
@@ -13,6 +13,7 @@
 #include <GLMHelpers.h>
 #include <NumericalConstants.h>
 #include <SharedUtil.h>
+#include <shared/NsightHelpers.h>
 
 #include "ElbowConstraint.h"
 #include "SwingTwistConstraint.h"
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: e1f7dfc4e14836ce7d596b1ccf0fc617e4f5ab9e available at: https://gist.github.com/e2cc7a289e810ddc995697e9386cb116<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8497-e1f7dfc.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8497-e1f7dfc.dmg

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -173,6 +177,19 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
                 absolutePoses[i] = absolutePoses[parentIndex] * _relativePoses[i];
             }
         }
+
+        // compute maxError
+        maxError = 0.0f;
+        for (int i = 0; i < targets.size(); i++) {
```
sethalves: I get a signed-vs-unsigned comparison warning, here

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -173,6 +177,19 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
...
+            if (targets[i].getType() == IKTarget::Type::RotationAndPosition || targets[i].getType() == IKTarget::Type::HmdHead ||
+                targets[i].getType() == IKTarget::Type::HipsRelativeRotationAndPosition) {
+                float error = glm::length(absolutePoses[targets[i].getIndex()].trans - targets[i].getTranslation());
+                if (error > maxError) {
+                    maxError = error;
+                    HACK_MAX_ERROR_JOINT = targets[i].getIndex();
```
sethalves: what's this for?  it's unread?

hifi-gustavo:
Build console output for commit: 1e95e067c029cb26b8e6de76163788e1cbd81fc9 available at: https://gist.github.com/e2cc7a289e810ddc995697e9386cb116<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8497-1e95e06.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8497-1e95e06.dmg

libraries/animation/src/AnimInverseKinematics.cpp
```diff
@@ -144,9 +145,11 @@ void AnimInverseKinematics::solveWithCyclicCoordinateDescent(const std::vector<I
         accumulator.clearAndClean();
     }
 
+    float maxError = FLT_MAX;
     int numLoops = 0;
-    const int MAX_IK_LOOPS = 4;
-    while (numLoops < MAX_IK_LOOPS) {
+    const int MAX_IK_LOOPS = 16;
+    const float MAX_ERROR_TOLERANCE = 0.1f; // cm
+    while (maxError > MAX_ERROR_TOLERANCE && numLoops < MAX_IK_LOOPS) {
```
AndrewMeadows: Oh ok... I was looking at this rather large `MAX_ERROR_TOLERANCE` thinking "Surely that has to produce bad results" but now I see that the units are in **cm**, so it corresponds to **1mm** -- should be good enough.

Initializing `maxError` to `FLT_MAX` was also throwing me but I see what you're doing here... keep it a `while` loop rather than a `do while`.

AndrewMeadows:
Needs a test plan and a **QA_READY** label.

AndrewMeadows:
I tested the IK and can't find any problems with it.  This PR appears to slew a little faster so any quirky instabilities (when the hand target passes through the torso to the back, for example) snap just a little bit faster.  Is it real or just my imagination -- it is subtle.

PR #8556, opened 2016-08-31
add support for renderInfo properties to model entities

ZappoMan:


ZappoMan:
Test plan:
- smoke test - make sure we don't see any crazy drop in performance
- open up your Script console, and for some model entity that it in view, enter the following... (substitute the correct entity ID)
```
print(JSON.stringify(Entities.getEntityProperties("{003ba42b-e5d4-4fe8-b983-b5dc0659abea}").renderInfo));
```
you should see something like:
```
{"verticesCount":37170,"texturesSize":257040,"hasTransparent":true,"drawCalls":6,"texturesCount":4}
```

hifi-gustavo:
Build console output for commit: e1845e2c3f78b39fd8852d6b659cccc735766ec7 available at: https://gist.github.com/9e0cbdd460c6d587969e166423ab10e1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8556-e1845e2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8556-e1845e2.dmg

ctrlaltdavid:
Tested and it provides the expected data for model entities. No data is provided for other entity types, though I believe that this is expected.

ZappoMan:
@ctrlaltdavid - correct - I only implemented this for Model entity types. We can discuss how it might be implemented for other types in a future PR after you've merged yours... I would recommend you assume all types will eventually support this info, and display it if the entity provides it.

libraries/entities/src/EntityItemProperties.h
```diff
@@ -308,6 +321,12 @@ class EntityItemProperties {
     glm::vec3 _naturalDimensions;
     glm::vec3 _naturalPosition;
 
+    size_t _renderInfoVertexCount { 0 };
+    int _renderInfoTextureCount { 0 };
```
jherico: why the mix of int / size_t?  All of these are counts, except the bool.

libraries/render-utils/src/MeshPartPayload.cpp
```diff
@@ -71,8 +71,37 @@ void MeshPartPayload::updateTransform(const Transform& transform, const Transfor
 
 void MeshPartPayload::updateMaterial(model::MaterialPointer drawMaterial) {
     _drawMaterial = drawMaterial;
+    calculateMaterialSize();
 }
 
+bool MeshPartPayload::calculateMaterialSize() {
```
jherico: you go through a lot to calculate the return value, but you don't seem to use it.

libraries/entities/src/EntityItemProperties.h
ZappoMan: some of the gpu stuff returns size_t and some returns int... I don't know why I just kept the type so that the compiler wouldn't bitch at me.

libraries/render-utils/src/MeshPartPayload.cpp
ZappoMan: I do use it in the main caller in the Satanic Model.

libraries/render-utils/src/Model.cpp
```diff
@@ -161,6 +161,23 @@ void Model::setOffset(const glm::vec3& offset) {
     _snappedToRegistrationPoint = false;
 }
 
+size_t Model::getRenderInfoTextureSize() {
+    if (!_hasCalculatedTextureSize && isLoaded() && getGeometry()->areTexturesLoaded()) {
+        size_t textureSize = 0;
+        bool allTexturesLoaded = true;
+        foreach(auto renderItem, _modelMeshRenderItemsSet) {
+            auto meshPart = renderItem.get();
+            bool allTexturesForThisMesh = meshPart->calculateMaterialSize();
```
ZappoMan: @jherico - note -- I use the result here... apparently sometimes the mesh part isn't "ready" to be honest about the texture size... so this rigamarole is about trying again later.

jherico: I saw, and changed my comment earlier.

PR #8564, opened 2016-09-02
Updating stats display for threaded rendering

jherico:
Add a new counter... 'frame rate'... 

* Frame rate - the rate at which the main thread is creating new threads.
* Render rate - the rate at which the render thread is executing frames on the GPU
* Present rate - the rate at which new frames are being submitted to the display (on anything but OpenVR, this should be roughly the same as the render rate, less some noise from thread timing issues)
* Present new rate - the number of rendered frames that consisted of new frame data from the main thread
* Present drop rate - the excess frames dropped because more than one frame was in the submission queue when the render thread started the next frame

The drop rate and the new rate should add up to closely match the render rate.  On OpenVR for some reason, the drop rate is fairly high...  I don't yet understand this as the logic for when to generate a new frame should not allow for the application to 'get ahead' of the display.  

## Testing

When the stats display is on a new stat 'Frame rate' should appear and the new and drop rates should no longer stay at 0.

### DIFF
--- a/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
+++ b/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
@@ -475,32 +475,28 @@ bool OpenGLDisplayPlugin::eventFilter(QObject* receiver, QEvent* event) {
 }
 
 void OpenGLDisplayPlugin::submitFrame(const gpu::FramePointer& newFrame) {
-    if (_lockCurrentTexture) {
-        return;
-    }
-
     withNonPresentThreadLock([&] {
         _newFrameQueue.push(newFrame);
     });
 }
 
 void OpenGLDisplayPlugin::updateFrameData() {
+    if (_lockCurrentTexture) {
+        return;
+    }
     withPresentThreadLock([&] {
-        gpu::FramePointer oldFrame = _currentFrame;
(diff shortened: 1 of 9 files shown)


hifi-gustavo:
Build console output for commit: 91bfa7677c71b19366b8c88b7a1abe9b86ad4b76 available at: https://gist.github.com/c0bdf59c5de18134b412c61883b0be1a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8564-91bfa76.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8564-91bfa76.dmg

libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
```diff
@@ -475,32 +475,28 @@ bool OpenGLDisplayPlugin::eventFilter(QObject* receiver, QEvent* event) {
 }
 
 void OpenGLDisplayPlugin::submitFrame(const gpu::FramePointer& newFrame) {
-    if (_lockCurrentTexture) {
```
ZappoMan: how nervous should I be about these changes? they seem to change the logic of the `submitFrame` and `updateFrameData` methods

PR #8572, opened 2016-09-06
remove audio settings

zzmp:
- remove audio settings from Interface and Server
- cleanup some audio files


hifi-gustavo:
Build console output for commit: 645e6b4c590ae3a7f1e3f6d2ca3e6b12b321e10a available at: https://gist.github.com/70a78407a3f11056ca7763bee024b02f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-645e6b4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-645e6b4.dmg

hifi-gustavo:
Build console output for commit: f1a594800f11e369a8d656a5163c4cd49d57249a available at: https://gist.github.com/70a78407a3f11056ca7763bee024b02f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-f1a5948.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-f1a5948.dmg

zzmp:
Removing the Qt audio output buffer may have been too ambitious, but QA will tell.

QA:
- Use a performant and less-performant machine, to make sure we don't miss the edge cases.
- Smoke test audio.

---

If anybody used specific audio settings, and needs them for any reason, please LMK which ones (with a screenshot from master) and why you need them.

---

I talked to @OmegaHeron who uses the `Output buffer intial size (frames)` setting, so I am marking this DNM until I've added that back in.

ZappoMan:
@zzmp @OmegaHeron - we need to figure out why you need to set this... what is it about Omega's setup that requires this "intervention"... why doesn't the system automatically correct this?

OmegaHeron:
It's because I use a bluetooth stereo headset and, as I commented time and time again, closer to 2 years ago than now, the audio code in its rabid desire to be the least possible latent clobbers buffers on a device such as a BT connected device which has an additional level of latency over that attainable from a mobo connected device.  Some USB attached devices also suffer from this, though, the wireless connected devices suffer worst.

ZappoMan:
@OmegaHeron - what happens if you use the default setting? What *should* happen is that it relatively quickly determines that the buffers are too shallow and they grow to work.

OmegaHeron:
That said - Ken's new code seems to have cleaned that up and this may be a, finally, non-issue.  I did note I could run my BT device with 1 buffer vs 9 required before his changes.  My concern is that since only a small small sample of potential consumers have tried this - we go back to forum blowing up with my audio sounds like a motorboat or screams or whatever and then have no way to adjust for that.  Most of us ancients are running with settings > 3 (the old default) and have been so for almost our entire lifespan here (2+ years in my case).

hifi-gustavo:
Build console output for commit: ef88eecea3e322849a130036820f4d3ec543fece available at: https://gist.github.com/70a78407a3f11056ca7763bee024b02f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-ef88eec.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-ef88eec.dmg

ZappoMan:
My concern @OmegaHeron is that we don't actually document these values anywhere - and so basically it's black magic that someone would even *attempt* to change that value to fix some problem... I'm not saying we shouldn't have the option... I'm saying we need to understand when/why someone needs to change the value... and either document it - or better -- automatically detect that case and adjust accordingly.

OmegaHeron:
Fair enough and I'm not pushing that any/all of those remain.  I'm willing to QA it without any of that present and report findings.  My immediate question is this - with those settings removed, does Interface ignore their previously set values in ini file and attempt to start at, say, 1 and go from there?  Or.. does it read vars - accept and lock to user's formerly available pick?  Or... does it read those and, if high buffer count, attempt to lower to the new potentially better < value?  Knowing that I can make a proper eval of whether or not this breaks anything and report back.

ZappoMan:
expected behavior should be: ignore previously set values, use defaults instead.

OmegaHeron:
Audio device - bluetooth connected stereo headset for playback.  Mobo mic in for input.

Ok - initially, with echo server audio enable, I hear broken static then after 1 to 2 seconds of that I get no sound at all.  This behavior is consistent using sandbox on localhost, stack running on linux over LAN or stack running on linux from Inet.

Taking off the BT headset and going to mobo wired connection I hear nothing at all on server loopback - no noise - nothing.  If I enable loopback LOCAL audio I do hear myself.

zzmp:
whoa, hi. I just put the buffer back.

I suspect the buffer *is* necessary with certain devices, especially without Ken's backend, because of the latency to serve those devices' callbacks.

I've hidden it in the developer menu.

Expected behavior: use the setting, and grow the buffer as need be. Do not persist any growth.
The buffer defaults to 1 (the minimum).

There is no way to tell what the buffer is actually at through menus/scripting, but you should be able to see if it increases through your logs.

OmegaHeron:
Using current DEV 5354 I need 6 buffers to have stable playback connecting to heron with an avg ping time of 51ms, same with connection on LAN.  When I tested Ken's PR for audio, the PR number escapes me at the moment, I was able to use a value of 1 with my configuration - now I can not.  My new number is 5 versus 9 in the past - certainly not 1.

hifi-gustavo:
Build console output for commit: 1ca7f61469834b4d54c1290625df0b1f7f27f24e available at: https://gist.github.com/70a78407a3f11056ca7763bee024b02f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-1ca7f61.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-1ca7f61.dmg

hifi-gustavo:
Build console output for commit: 6228fd4d9c6b608b09b52a5be02ea2be8ff1da00 available at: https://gist.github.com/70a78407a3f11056ca7763bee024b02f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-6228fd4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-6228fd4.dmg

hifi-gustavo:
Build console output for commit: 52be0a03bc46f0ca511b372cd04443ad27d56075 available at: https://gist.github.com/70a78407a3f11056ca7763bee024b02f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-52be0a0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-52be0a0.dmg

hifi-gustavo:
Build console output for commit: fcbd9c539633eebafd9498d893e7af3d61414262 available at: https://gist.github.com/70a78407a3f11056ca7763bee024b02f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-fcbd9c5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8572-fcbd9c5.dmg

PR #8600, opened 2016-09-14
don't do hand-controller edit object-selection when pointed at a HUD element

howard-stearns:
fix https://highfidelity.fogbugz.com/f/cases/1249/Edit-js-with-Hand-Controllers-selecting-panel-de-selects-object

### DIFF
--- a/scripts/system/libraries/entitySelectionTool.js
+++ b/scripts/system/libraries/entitySelectionTool.js
@@ -1022,6 +1022,9 @@ SelectionDisplay = (function() {
                     // No switching while the other is already triggered, so no need to release.
                     activeHand = (activeHand === Controller.Standard.RightHand) ? Controller.Standard.LeftHand : Controller.Standard.RightHand;
                 }
+                if (Reticle.pointingAtSystemOverlay || Overlays.getOverlayAtPoint(Reticle.position)) {
+                    return;
+                }
                 var eventResult = that.mousePressEvent({});
                 if (!eventResult || (eventResult === 'selectionBox')) {
                     var pickRay = controllerComputePickRay();


howard-stearns:
test plan, using any hand controller (hydra in desktop or oculus, or Vive with hand controllers):
- Click the edit button and create two objects set slightly apart from each other (e.g., a box and a sphere).
- Make sure you can select either object using the mouse and the hand controller. (edit.js only supports single-object selection with hand controllers, regardless of this PR.)
- Position the entity panel over the unselected object, such that click targets on the entity panel (e.g., text fields) are positioned over the unselected object. Trigger on those targets using the hand controller and make sure that the unselected object does not become selected. (This was the bug.)
- Repeat with other overlays (such as system toolbar, microphone inputs, or user list) over the unselected objects. Again, the unselected object should not become selected by triggering "through" the overlay.

hifi-gustavo:
Build console output for commit: 03a8af76e3c3d41508cba17eb2b5eebb5ebf66c1 available at: https://gist.github.com/e00f9ea820f4a1151457255a52dd38c4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8600-03a8af7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8600-03a8af7.dmg

Nex-Pro:
**PR #8530**

**CPU:** 4.4 ghz Intel i7
**RAM:** 16 GB
**GPU:** Nvidia gtx 1070 8gb

**Testing OS:** win 10 pro

Tested at home location in the sandbox of this PR.
No other open GL applications running.


Click the edit button and create two objects set slightly apart from each other (e.g., a box and a sphere). **CHECK**

Make sure you can select either object using the mouse and the hand controller. (edit.js only supports single-object selection with hand controllers, regardless of this PR.) **CHECK**

Position the entity panel over the unselected object, such that click targets on the entity panel (e.g., text fields) are positioned over the unselected object. Trigger on those targets using the hand controller and make sure that the unselected object does not become selected. (This was the bug.) **SUCCESS**

Repeat with other overlays (such as system toolbar, microphone inputs, or user list) over the unselected objects. Again, the unselected object should not become selected by triggering "through" the overlay. **FAIL**

Users was impossible to reposition using the hand controllers. Any attempt to do so resulted in a background object becoming selected.

After moving users over the two test objects ( I used mouse to move it ) it became apparent that an object behind the users panel would become selected when using the dropdowns.

Also of note: When the two test objects were overlapped by about 50% it was very unreliable to select the un-selected object. Much more likely was that a background object would become selected or that the currently selected object would move while moderate pressure was applied to the trigger ( beam was still blue ). 

Outside the scope of this PR:

The above highlighted another issue with the hand controllers that has persisted for some time. The selected object would move rapidly and towards and away from the avatar with slight hand movements. To mitigate this I found that holding the hand controller above my head and pointing it down towards the selected object gave more control.

It is still not possible ( for me ) to have the up/down positioner arrows work on selected entities whilst using hand controllers.

@howard-stearns  thanks for fixing the shoot-thru edit panel issue.

Nex-Pro:
After a long run in HMD interface exited cleanly.

howard-stearns:
Thanks @Nex-Pro.  The users list definitely has some whacked out behavior that I don't even know how to characterize:
- As you say, you can't drag it with hand controllers.
- Also, you CAN click through it with the MOUSE to accidentally select objects for editing.
- Also, once selected, you can move objects around by dragging through it.
- Also, you can get the users list into some bizarre state in which it seems to suck all pointer input and even displays and interacts weirdly.
All of these (and probably more) are separate problems.
(https://highfidelity.fogbugz.com/f/cases/1687/user-list-behaves-oddly)

So, the question for this bug fix, is whether or not, after positioning the user list or objects in any way you can, do you accidentally hand-controller select through an otherwise functioning user list?  If so, can you give steps to reproduce?

Nex-Pro:
@howard-stearns  Re checking it now.

Nex-Pro:
@howard basically It will select an object behind the users list with hand controllers if the drop-downs lower-right are used. 

Also this may be diagnostic of a button-state issue. If the add/remove button is clicked on the friends list no entity gets selected. But if it is clicked a second time while the add friends dialog is on screen the click goes thru and selects the entity that is first behind the button.

howard-stearns:
Thanks @Nex-Pro, that's helpful. I've added that description to the other ticket.
I'll call this a limited pass, then.

PR #8604, opened 2016-09-15
Toggle Physics Option

rjobse:
Added an option in the settings tab labeled "toggle physics" which will
disable and enable physics.

I wrote this code a week ago and was attempting to debug it but was
unable to pull up the dev-demo because my branch was too far out of sync
with the master branch.  When I attempted to sync up with the master
branch I was getting compiler errors (possibly caused by me having
Qt5.5). Tom has encouraged me to submit my code despite not being able
to compile or test it for review so please let me know if it has a bug
or if there is anything you would like me to add.

The problem that I chose to solve with this submission is the creation
of a drop down tab that would disable physics until the user reenabled
it again. I accomplished this by adding in a flag inside of Application
(interface\src\Application.h) called _physicsManuallyDisabled which
would store if the user disabled or enabled physics from the drop menu.
I added in the flag inside of the update() when the _physicsEnabled flag
is checked so that if physics is manually disabled then _physicsEnabled
will never be allowed to be set high until physics is enabled again.
Lastly, I control the value of _physicsManuallyDisabled through an
addition to the settings dropdown menu with I labeled "toggle physics".
When the user selects this option it calls the togglePhysics() function
which toggles the _physicsManuallyDisabled flag (as well as sets
_physicsEnabled low if physics is disabled).

Please feel free to contact me if you are having any issues with my code
or would like for me to provide any more clarity.

Ryan J

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -3645,7 +3645,7 @@ void Application::update(float deltaTime) {
 
     updateLOD();
 
-    if (!_physicsEnabled) {
+    if (!_physicsEnabled && !_physicsManuallyDisabled) {
         // we haven't yet enabled physics.  we wait until we think we have all the collision information
         // for nearby entities before starting bullet up.
         quint64 now = usecTimestampNow();
@@ -3986,6 +3986,18 @@ int Application::sendNackPackets() {
 
 
     return packetsSent;
+}
+
+void Application::togglePhysics() {
+	if(_physicsManuallyDisabled) {
+		_physicsManuallyDisabled = false;
+	}
+	else {
+		//set enable physics low if physics is disabled
+		_physicsEnabled = false;
+		_physicsManuallyDisabled = true;
+	}
+	return;
 }
 
 void Application::queryOctree(NodeType_t serverType, PacketType packetType, NodeToJurisdictionMap& jurisdictions, bool forceResend) {
--- a/interface/src/Menu.h
+++ b/interface/src/Menu.h
@@ -184,6 +184,7 @@ namespace MenuOption {
     const QString ThreePointCalibration = "3 Point Calibration";
     const QString ThrottleFPSIfNotFocus = "Throttle FPS If Not Focus"; // FIXME - this value duplicated in Basic2DWindowOpenGLDisplayPlugin.cpp
     const QString ToolWindow = "Tool Window";
+	const QString TogglePhysics = "Toggle Physics";
     const QString TransmitterDrive = "Transmitter Drive";
     const QString TurnWithHead = "Turn using Head";
     const QString UseAudioForMouth = "Use Audio for Mouth";
--- a/interface/src/Application.h
+++ b/interface/src/Application.h
@@ -335,6 +335,8 @@ public slots:
 
     void rotationModeChanged() const;
 
+	void togglePhysics();
+	
     static void runTests();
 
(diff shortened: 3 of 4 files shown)


hifi-gustavo:

sethalves:
ok to test

hifi-gustavo:
Build console output for commit: 89ad422c7e7e21d9475b28a94213c0afd9abb5fb available at: https://gist.github.com/0422a6d42c3a0eaea8a3fba45d000bbd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8604-89ad422.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8604-89ad422.dmg

interface/src/Menu.h
```diff
@@ -184,6 +184,7 @@ namespace MenuOption {
     const QString ThreePointCalibration = "3 Point Calibration";
     const QString ThrottleFPSIfNotFocus = "Throttle FPS If Not Focus"; // FIXME - this value duplicated in Basic2DWindowOpenGLDisplayPlugin.cpp
     const QString ToolWindow = "Tool Window";
+	const QString TogglePhysics = "Toggle Physics";
```
AndrewMeadows: The menu text "Toggle Physics" is ambiguous when paired with the state of the checkbox -- "Do I check it to toggle physics ON or to toggle it OFF?".  The text should probably be something like "Disable physics simulation" such that the meaning of checked/unchecked state is more clear.

interface/src/Application.cpp
```diff
@@ -3988,6 +3988,18 @@ int Application::sendNackPackets() {
...
 }
 
+void Application::togglePhysics() {
+	if(_physicsManuallyDisabled) {
+		_physicsManuallyDisabled = false;
+	}
+	else {
+		//set enable physics low if physics is disabled
+		_physicsEnabled = false;
+		_physicsManuallyDisabled = true;
```
AndrewMeadows: This was confusing me... at toggle OFF you set both false but at toggle ON you set only one true.  But now I see... if one is set true then the update will happen and there was already code to set _physicsEnabled true on the next step.  It works but requires a [wtf-moment](http://www.osnews.com/images/comics/wtfm.jpg) for future readers.

Mysteriously, the "Toggle Physics" option is not working for me.  I installed it and can toggle the setting but don't notice a difference when I do.

hifi-gustavo:
Build console output for commit: f8e7926e7514f05f262b083fbe26608b625d39fa available at: https://gist.github.com/0422a6d42c3a0eaea8a3fba45d000bbd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8604-f8e7926.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8604-f8e7926.dmg

Nex-Pro:
I don't think this is QA Ready.

interface/src/Application.cpp
```diff
@@ -3645,7 +3645,7 @@ void Application::update(float deltaTime) {
 
     updateLOD();
 
-    if (!_physicsEnabled) {
+    if (!_physicsEnabled && !_physicsManuallyDisabled) {
```
AndrewMeadows: This is the wrong place to check `_physicsManuallyDisabled` in order to disable the physics simulation during runtime.

PR #8605, opened 2016-09-15
add kinematic character controller to enable physics sweep tests near avatar

AndrewMeadows:
Note: this PR is against an experimental branch instead of _master_.
- added CharacterGhostObject and CharacterController now has one
- the ghost follows the avatar's collision body around and caches overlapping bounding boxes of nearby objects in the physics simulation, which enables efficient sweep tests within its bounds
- there is some demo code for moving the avatar capsule around in a kinematic fashion (its body follows the ghost after it moves with sweeps) and debug options for testing it... but something broke it in this branch.  That's ok, because we still want the ghost for its efficient sweep capabilities in other work.

### DIFF
--- a/libraries/physics/src/CharacterController.cpp
+++ b/libraries/physics/src/CharacterController.cpp
@@ -124,9 +124,7 @@ void CharacterController::setDynamicsWorld(btDynamicsWorld* world) {
             _ghost.setCollisionShape(_rigidBody->getCollisionShape()); // KINEMATIC_CONTROLLER_HACK
         }
         // KINEMATIC_CONTROLLER_HACK
-        int16_t group = BULLET_COLLISION_GROUP_MY_AVATAR;
-        int16_t mask = BULLET_COLLISION_MASK_MY_AVATAR & (~ group);
-        _ghost.setCollisionGroupAndMask(group, mask);
+        _ghost.setCollisionGroupAndMask(_collisionGroup, BULLET_COLLISION_MASK_MY_AVATAR & (~ _collisionGroup));
         _ghost.setCollisionWorld(_dynamicsWorld);
         _ghost.setDistanceToFeet(_radius + _halfHeight);
         _ghost.setMaxStepHeight(0.75f * (_radius + _halfHeight)); // HACK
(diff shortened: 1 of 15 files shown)


libraries/physics/src/CharacterController.cpp
```diff
@@ -414,6 +444,10 @@ glm::vec3 CharacterController::getLinearVelocity() const {
     return velocity;
 }
 
+void CharacterController::setCapsuleRadius(float radius) {
+    _radius = radius;
+}
```
AndrewMeadows: I added this trivial method because I had some experimental/debug code that I wanted to execute whenever the radius was set.  This might still be useful so I left it in.

libraries/physics/src/CharacterController.cpp
```diff
@@ -522,6 +558,7 @@ void CharacterController::computeNewVelocity(btScalar dt, btVector3& velocity) {
         for (size_t i = 0; i < velocities.size(); ++i) {
             velocity += (weights[i] / totalWeight) * velocities[i];
         }
+        _simpleMotorVelocity /= totalWeight;
```
AndrewMeadows: The _simpleMotorVelocity is the intended velocity of the avatar capsule (rather than it's real velocity while bumping around) and is uses in the sweep tests during the kinematic motion demo.

libraries/physics/src/CharacterGhostShape.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include "CharacterGhostShape.h"
+
+void CharacterGhostShape::getAabb (const btTransform& t, btVector3& aabbMin, btVector3& aabbMax) const {
+    btCapsuleShape::getAabb(t, aabbMin, aabbMax);
+    // double the size of the Aabb by expanding both corners by half the extent
+    btVector3 expansion = 0.5f * (aabbMax - aabbMin);
+    aabbMin -= expansion;
+    aabbMax += expansion;
+}
```
AndrewMeadows: The purpose of the `CharacterGhostShape` class is to make a capsule that has an artificially large Aabb in the broadphase so it can capture more nearby overlapping pairs.  It is used for the shape of the `CharacterGhostObject`.

hifi-gustavo:
Build console output for commit: 821570fa76184ec61b157aa03a9074e4b3cfa35b available at: https://gist.github.com/f500171e79443b80ed0af9efbeb9d5a3<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8605-821570f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8605-821570f.dmg

PR #8614, opened 2016-09-16
Turn and drive the HMD with linear velocity

hyperlogic:
Turning action rotates the sensor to world matrix about the HMD, not the avatar.
Driving actions are converted to a single velocity relative to the HMD orientation.  This velocity is applied to the sensor to world matrix with NO acceleration, (other than the change acceleration introduced from to turning your HMD).

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -1300,12 +1300,18 @@ void MyAvatar::updateMotors() {
         }
         const float DEFAULT_MOTOR_TIMESCALE = 0.2f;
         const float INVALID_MOTOR_TIMESCALE = 1.0e6f;
-        if (_isPushing || _isBraking || !_isBeingPushed) {
-            _characterController.addMotor(_actionMotorVelocity, motorRotation, DEFAULT_MOTOR_TIMESCALE, INVALID_MOTOR_TIMESCALE);
+
+        if (!qApp->isHMDMode()) {
+            // OUTOFBODY_HACK: add default zero velocity motor to the characterController
+            _characterController.addMotor(glm::vec3(), motorRotation, DEFAULT_MOTOR_TIMESCALE, INVALID_MOTOR_TIMESCALE);
         } else {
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: c7caf9fe0667dbea92a5d017fc72ef14d3b4068c available at: https://gist.github.com/c70f3fea4e511b8aa3b74299c67d23c4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8614-c7caf9f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8614-c7caf9f.dmg

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1300,12 +1300,18 @@ void MyAvatar::updateMotors() {
...
+
+        if (!qApp->isHMDMode()) {
+            // OUTOFBODY_HACK: add default zero velocity motor to the characterController
+            _characterController.addMotor(glm::vec3(), motorRotation, DEFAULT_MOTOR_TIMESCALE, INVALID_MOTOR_TIMESCALE);
         } else {
-            // _isBeingPushed must be true --> disable action motor by giving it a long timescale,
-            // otherwise it's attempt to "stand in in place" could defeat scripted motor/thrusts
-            _characterController.addMotor(_actionMotorVelocity, motorRotation, INVALID_MOTOR_TIMESCALE);
+            if (_isPushing || _isBraking || !_isBeingPushed) {
+                _characterController.addMotor(_actionMotorVelocity, motorRotation, DEFAULT_MOTOR_TIMESCALE, INVALID_MOTOR_TIMESCALE);
```
AndrewMeadows: @hyperlogic, to get very close to immediate no-ramp-up constant velocity you can just set the first (horizontal) timescale argument here to a very low value, like 0.01 sec. Try it and see.

hyperlogic: True, but the point is that we don't want to apply any motor to the character controller.   Instead we want to move the sensor to world matrix.

PR #8635, opened 2016-09-22
fix missing local audition for audio injectors

zzmp:


libraries/script-engine/src/AudioScriptingInterface.cpp
```diff
@@ -45,7 +45,7 @@ ScriptAudioInjector* AudioScriptingInterface::playSound(SharedSoundPointer sound
         // stereo option isn't set from script, this comes from sound metadata or filename
         AudioInjectorOptions optionsCopy = injectorOptions;
         optionsCopy.stereo = sound->isStereo();
-        auto injector = AudioInjector::playSound(sound->getByteArray(), optionsCopy, _localAudioInterface);
+        auto injector = AudioInjector::playSound(sound->getByteArray(), optionsCopy);
```
davidkelly: Since the injector has the interface, does `_localAudioInterface` need to be part of `AudioScriptingInterface` at all?  I didn't really look hard, but at first glance it seems like it isn't used except in `setStereoInput` which now I'm not sure what it does.

hifi-gustavo:
Build console output for commit: 9e2ec27b29d8a7b186409a67ddad368067031132 available at: https://gist.github.com/94891273fc6ef2331e92f8bb6b7a0075<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8635-9e2ec27.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8635-9e2ec27.dmg

libraries/script-engine/src/AudioScriptingInterface.cpp
zzmp: I left it in because of `setStereoInput`. I don't know what the uses for that are, but I am afraid to break them.

davidkelly: K fair enough.  Worth eliminating if that is never used, and it looks like the _localAudioInterface now doesn't do anything, so that function is probably a no-op.  In fact IIRC we detect whether the sound is mono or stereo so (maybe) this function was never used?  Perhaps we can grep the source and verify?

ZappoMan:
Test plan:
- Smoke test
- Make sure collision sounds work
- Make sure normal scripted audio works

libraries/script-engine/src/AudioScriptingInterface.cpp
davidkelly: Especially since `_localAudioInterface` is private, and only now invoked in `setStereoInput`, and then (AFAICT) never used for anything, it seems like `setStereoInput` does nothing at all.  Seems worth removing unless there is something I'm missing here.

zzmp: `Audio.setStereoInput(true)` still works, and we still have codepaths in AudioClient.cpp specifically for dealing with stereo input. If you invoke that and enable server echo, you can hear that the audio is handled differently, and I expect it would work fine with stereo input (which I do not have to test with).

I don't think that we should remove a likely working feature to clean up AudioScriptingInterface. I'm OK cleaning the implementation if you can think of another way to do it, but removing all of the stereo input code seems out of scope for this PR.

davidkelly: ok cool, I've missed something in the code.  If it does something still, it should stay.

zzmp:
Note in your QA that restarted injectors may not work - this is a separate bug that will not be handled in this PR.

zzmp:
QA - please slack me after you QA this. I've got another release blocker PR to put up that depends on this PR, so I want to get it up ASAP (or fix any failures).

hifi-gustavo:
Build console output for commit: d0f4a29031205159448a0c88c2b089708b6baa17 available at: https://gist.github.com/94891273fc6ef2331e92f8bb6b7a0075<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8635-d0f4a29.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8635-d0f4a29.dmg

thoys:
# QA-testing report

Tested in a fresh sandbox with home on a Windows machine.

- Continuous sounds were running fine (Server side invoked injectors)
- Threw the books-plank over and heard collision sounds
- Used the pingpong gun to use the injector and it made the usual pong gun sound.

Everything else seems to be looking fine and smooth.

QA-pass

PR #8650, opened 2016-09-23
reload qml with scripts

zzmp:


libraries/script-engine/src/ScriptEngines.cpp
```diff
@@ -367,28 +369,21 @@ QStringList ScriptEngines::getRunningScripts() {
 
 void ScriptEngines::stopAllScripts(bool restart) {
     QReadLocker lock(&_scriptEnginesHashLock);
-    if (restart) {
```
ZappoMan: why are you removing this? I'm pretty sure we still want this.

zzmp:
QA:

- Save these to two files in the same folder, `test.js` and `test.qml`:
```
// test.js
var qml = Script.resolvePath('test.qml');
var window = new OverlayWindow({
  title: 'PR8650',
  source: qml,
  width: 400, height: 900
});
window.setPosition(25, 50);
```
```
// test.qml
import QtQuick 2.5
import QtQuick.Controls 1.4

Column {
  id: root

  Text {
    text: "Old"
    y: 30
  }
}
```
- Run test.js
- Update `test.qml`'s text to read "New", and save it
- Reload all scripts (enable `Settings > Advanced Menus` and select `Edit > Reload All Scripts`)
  - On master, the QML should not update
  - On this PR, the QML should update

libraries/script-engine/src/ScriptEngines.cpp
zzmp: It is redundant. `restart == true` only if called from a function that is already clearing the cache.

ZappoMan: ok

hifi-gustavo:
Build console output for commit: aeaae4ccc820a720efaf2f8c9706a58038419383 available at: https://gist.github.com/dfbd2773a79107624b0a8a3e3ff6a55a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8650-aeaae4c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8650-aeaae4c.dmg

OmegaHeron:
Testing Windows 10
Master acts as indicated, does not update QML.
With this PR reload all scripts does reload QML.

Not indicated in scope of QA Test, but, **only** reload all scripts forces a QML reload.  Reloading single script from running scripts dialog does **not** update QML nor does terminating script then manually loading again.  Just an FYI there.

zzmp:
@OmegaHeron unfortunately, the QML components are not sandboxed to the script engines that run them, so I could not empty only script-specific QML components from the cache. I felt that it would be unintuitive to totally empty the cache with single-script reloads, so I only did it on reload all, as you've found.

PR #8682, opened 2016-09-29
add developer menu support to enable/disable dynamic texture management

ZappoMan:


ZappoMan:
This adds a developer menu option that allows you to enable/disable incremental texture transfers and dynamic texture management. The default behavior is to enable these for windows machines with more than 4 cores... and disable it on windows machines with 4 or less cores.

Test plan:
- smoke test to make sure that basic rendering performance didn't degrade (not expected, since by default, this changes nothing)
- verify Developer > Render > Enable Incremental Texture Transfer and Enable Dynamic Texture Management menus exist
- verify on startup you should see something like the following message... 
```
[09/29 10:24:01] [DEBUG] [TEXTURE TRANSFER SUPPORT] 
[09/29 10:24:01] [DEBUG] 	idealThreadCount: 8 
[09/29 10:24:01] [DEBUG] 	RECOMMENDED enableSparseTextures: true 
[09/29 10:24:01] [DEBUG] 	RECOMMENDED enableIncrementalTextures: true
[09/29 10:24:01] [DEBUG] [TEXTURE TRANSFER SUPPORT] SETTING - Enable Incremental Texture Transfer: true
[09/29 10:24:01] [DEBUG] [TEXTURE TRANSFER SUPPORT] SETTING - Enable Sparse Textures and Dynamic Texture Management: true
```
or...
```
[09/29 10:24:01] [DEBUG] [TEXTURE TRANSFER SUPPORT] 
[09/29 10:24:01] [DEBUG] 	idealThreadCount: 4 
[09/29 10:24:01] [DEBUG] 	RECOMMENDED enableSparseTextures: false 
[09/29 10:24:01] [DEBUG] 	RECOMMENDED enableIncrementalTextures: false
[09/29 10:24:01] [DEBUG] [TEXTURE TRANSFER SUPPORT] SETTING - Enable Incremental Texture Transfer: false
[09/29 10:24:01] [DEBUG] [TEXTURE TRANSFER SUPPORT] SETTING - Enable Sparse Textures and Dynamic Texture Management: false
```

later in your log you will see something like this...
```
[09/29 10:24:01] [DEBUG] [TEXTURE TRANSFER SUPPORT] --- Enable Dynamic Texture Management menu option: true
[09/29 10:24:01] [DEBUG] [TEXTURE TRANSFER SUPPORT] SETTING - Enable Sparse Textures and Dynamic Texture Management: true
[09/29 10:24:01] [DEBUG] [TEXTURE TRANSFER SUPPORT] --- Enable Incremental Texture Transfer menu option: true
[09/29 10:24:01] [DEBUG] [TEXTURE TRANSFER SUPPORT] SETTING - Enable Incremental Texture Transfer: false
```

If you are on an i5 you should see recommended *false* and then the menu options should set it to *false*.
If you are on an i7 you should see recommended *true* and then the menu options should set it to *true*.

- Verify that you can change these options - and they will be remembered next time you start up.

libraries/gpu/src/gpu/Texture.cpp
```diff
@@ -28,6 +31,34 @@ std::atomic<uint32_t> Texture::_textureCPUCount{ 0 };
 std::atomic<Texture::Size> Texture::_textureCPUMemoryUsage{ 0 };
 std::atomic<Texture::Size> Texture::_allowedCPUMemoryUsage { 0 };
 
+std::atomic<bool> Texture::_enableSparseTextures { false };
```
jherico: why aren't you using the existing logic to set the initial values here?

libraries/gpu/src/gpu/Texture.h
```diff
@@ -144,6 +144,9 @@ class Texture : public Resource {
     static std::atomic<Size> _textureCPUMemoryUsage;
     static std::atomic<Size> _allowedCPUMemoryUsage;
     static void updateTextureCPUMemoryUsage(Size prevObjectSize, Size newObjectSize);
+
+    static std::atomic<bool> _enableSparseTextures;
```
jherico: I'm not sure this will behave the way we expect it to, since the code that may end up using this will be in a plugin (DLL), which will have it's own copy of the static.

libraries/gpu/src/gpu/Texture.cpp
jherico: OK, I see you've moved that logic into the Menu.cpp code, but I feel like it belongs here instead.

libraries/gpu/src/gpu/Texture.h
ZappoMan: in my testing - all access to the statics seemed to happen after the call to set the value from menu -- and it always seemed to get the same value as I had set it to...

But yeah - I was concerned about that too... 

- How else can we verify it's working?
- What other suggestions do you have?

libraries/gpu/src/gpu/Texture.cpp
ZappoMan: With the addition of the menu items, the environment variables don't make any sense anymore... and the logic seemed overly complicated.

hifi-gustavo:
Build console output for commit: 38eccd32a8d5f6ce25181e6bfc4be6e625ba9a16 available at: https://gist.github.com/14ba69b4637fb241398b7a0542884c0e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8682-38eccd3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8682-38eccd3.dmg

ZappoMan:
test this please

hifi-gustavo:
Build console output for commit: 38eccd32a8d5f6ce25181e6bfc4be6e625ba9a16 available at: https://gist.github.com/14ba69b4637fb241398b7a0542884c0e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8682-38eccd3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8682-38eccd3.dmg

libraries/gpu/src/gpu/Texture.h
jherico: If it's working then this is fine.  In general, the best alternative for storing a global value that is certain to be visible in all plugins is to use the `property` & `setProperty` functions on qApp.  You can typically use a `std::call_once` wrapper with a static `std::once_flag` and either initialize the property or fetch it if it's set.  That means that the value gets cached in the DLL local static and the cost of going into the qApp instance is amortized over the life of the application.

PR #8680, opened 2016-09-29
Backup old content on update

Atlante45:
- Backup old content instead of deleting it during a sandbox content update.

(UI still need some work, will cleanup tomorrow)

Test plan:
- Launch sandbox (content set should get installed).
- Close sandbox.
- Go to `C:\Users\Clement\AppData\Roaming\High Fidelity - PRXXXX\Server Console\config.json`.
- Change the homeContentLastModified value to an old date (ie: 2012 should do it).
- Start sandbox.
- You should get prompted for an update.
- Accept and notice the content set getting re-downloaded.
- There should be another window explaining how to restore the content
- Follow instructions and restore original content.

PR #8689, opened 2016-09-29
Backup old content on update - RC PR Mirror

Atlante45:
Mirror to the RC branch for #8680


server-console/src/content-update.css
```diff
@@ -0,0 +1,172 @@
...
+a:hover,
+a:active {
+    color: #B4B4B4;
+}
+
+a:hover {
+    color: #2D88A4;
+}
+
+.header{
```
huffman: Not a blocker, and I don't think it's really defined anywhere, but I would expect a space before {

server-console/src/main.js
```diff
@@ -488,27 +496,62 @@ function updateTrayMenu(serverState) {
...
+        console.log("Created directory " + folder);
+
+
+        var dsBackup = path.join(folder, '/domain-server');
+        fs.renameSync(getDomainServerClientResourcesDirectory(), dsBackup);
+        console.log("Moved directory " + getDomainServerClientResourcesDirectory());
+        console.log("to " + dsBackup);
+
+        var acBackup = path.join(folder, '/assignment-client');
+        fs.renameSync(getAssignmentClientResourcesDirectory(), acBackup);
```
huffman: One concern I would have about renaming is that if the ds rename is successful but the ac rename fails, you will be left in a bad state that you probably won't be able to easily recover from. An alternative would be to copy and then delete.

Also, this looks like it will fail if you already have a backup in the backup directory?

server-console/src/main.js
```diff
@@ -488,27 +496,62 @@ function updateTrayMenu(serverState) {
...
     } catch (e) {
         console.log(e);
+        return false;
     }
 }
 
-function deleteResourceDirectoriesAndRestart() {
+function openBackupInstructions(folder) {
+  // Explain user how to restore server
+  var window = new BrowserWindow({
```
huffman: indentation should be 4

server-console/src/main.js
```diff
@@ -488,27 +496,62 @@ function updateTrayMenu(serverState) {
...
+
+}
+function backupResourceDirectoriesAndRestart() {
     homeServer.stop();
-    deleteResourceDirectories();
-    maybeInstallDefaultContentSet(onContentLoaded);
+
+    var folder = getRootHifiDataDirectory() + "/Server Backup - " + Date.now();
+    if (backupResourceDirectories(folder)) {
+        maybeInstallDefaultContentSet(onContentLoaded);
```
huffman: If this fails the user would be left in a bad state, without the new content or their old content. Maybe we should move the backed up content back in this case.

hifi-gustavo:
Build console output for commit: b83243d65d5f85453eed1849d15878f08b24a79b available at: https://gist.github.com/5225f33f5e9fd3f7d2889d448d585353<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8689-b83243d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8689-b83243d.dmg

PR #8704, opened 2016-10-01
make calculateTextureInfo thread safe

ZappoMan:
Fixes crashes in calculateMaterialSize... which could be caused by scripts getting properties, while textures are being loaded or set from other threads.

### DIFF
--- a/libraries/model/src/model/Material.h
+++ b/libraries/model/src/model/Material.h
@@ -11,6 +11,8 @@
 #ifndef hifi_model_Material_h
 #define hifi_model_Material_h
 
+#include <QMutex>
+
 #include <bitset>
 #include <map>
 
@@ -324,7 +326,7 @@ public:
 
     // The texture map to channel association
     void setTextureMap(MapChannel channel, const TextureMapPointer& textureMap);
-    const TextureMaps& getTextureMaps() const { return _textureMaps; }
+    const TextureMaps& getTextureMaps() const { return _textureMaps; } // FIXME - not thread safe... 
     const TextureMapPointer getTextureMap(MapChannel channel) const;
 
     // Albedo maps cannot have opacity detected until they are loaded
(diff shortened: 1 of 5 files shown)


ZappoMan:
I have concrete evidence that the `Model::calculateTextureInfo()` will be called on scripted threads and the `Material::setTextureMap()` can be called on the Main Thread, Thread Pool threads, and potentially other scripts. This is likely the source of these crashes. This PR rearranges the code and adds locks accordingly.

Test Plan:
- Smoke test
- use the edit.js "expanded view" of entity list to verify that texture sizes are loading
- look around while content is loading

libraries/model/src/model/Material.h
```diff
@@ -324,7 +326,7 @@ class Material {
 
     // The texture map to channel association
     void setTextureMap(MapChannel channel, const TextureMapPointer& textureMap);
-    const TextureMaps& getTextureMaps() const { return _textureMaps; }
+    const TextureMaps& getTextureMaps() const { return _textureMaps; } // FIXME - not thread safe...
```
ZappoMan: note: I don't have any evidence that this is failing, but it looks dangerous... this method is called inside of the render items, and potentially other threads could be changing the textures of the material object.

hifi-gustavo:
Build console output for commit: 7fe16442faea5bac5dc4107ad20763c914ad91c1 available at: https://gist.github.com/d178167cde05546fe026506ca22bcb81<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8704-7fe1644.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8704-7fe1644.dmg

libraries/model/src/model/Material.h
```diff
@@ -344,12 +346,25 @@ class Material {
     };
 
     const UniformBufferView& getTexMapArrayBuffer() const { return _texMapArrayBuffer; }
+
+    int getTextureCount() const { calculateMaterialInfo(); return _textureCount; }
```
samcake: IF i understand correctly, it s because these set of get methods have a specific lock when accessing the values and the _textureMAps that  the compute code is now safe.
Before these feature, the getTextureMaps where probably never used with a potential multithreaded access
ok

PR #8707, opened 2016-10-03
fix canRez logic in tribble script

howard-stearns:
only need canRezTemp, and fail hard rather than retry

### DIFF
--- a/scripts/developer/tests/performance/tribbles.js
+++ b/scripts/developer/tests/performance/tribbles.js
@@ -54,42 +54,46 @@ function randomVector(range) {
     };
 }
 
-Script.setInterval(function () {
-    if (!Entities.serversExist() || !Entities.canRez()) {
-        return;
-    }
-    if (totalCreated >= NUMBER_TO_CREATE) {
-        print("Created " + totalCreated + " tribbles.");
-        Script.stop();
-    }
+if (!Entities.canRezTmp()) {
+    Window.alert("Cannot create temp objects here.");
+    Script.stop();
+} else {
+    Script.setInterval(function () {
+        if (!Entities.serversExist()) {
+            return;
+        }
+        if (totalCreated >= NUMBER_TO_CREATE) {
+            print("Created " + totalCreated + " tribbles.");
+            Script.stop();
+        }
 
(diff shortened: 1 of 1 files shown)


howard-stearns:
There's existing logic here that limits how many tribbles are created at a time. (E.g., can change the number of tribbles parameter to make a 1000, but it still won't overload the entity server by making too many at once.) That logic doesn't change here.

However, mixed in to that logic is a test for whether the entity server is ready (which is correct) and also a check for whether we canRez.  The test for canRez is wrong in two ways:
1) We really only care about rez'ing temp objects, which is a separate function.
2) If we don't have permission to test rez, we shouldn't keep trying, but should just fail hard with an obvious alert.

howard-stearns:
to test: make some tribbles in a place where you have temp rez permission, but not permanent rez. (Presumably dev-playa.)  It should make a bunch of tribbles that should hop around for a while and then stop.

HOWEVER: This may not work if there are other people present, because they won't have the referenced files for the entity script and library. That's a different issue outside the scope of this PR.

hifi-gustavo:
Build console output for commit: 76d9d4c635b2349bf48057cfcfd0c55bbe6f56b8 available at: https://gist.github.com/17cc52f87d6f654740e274a35f0584ab<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8707-76d9d4c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8707-76d9d4c.dmg

howard-stearns:
test his please

howard-stearns:
test this please

hifi-gustavo:
Build console output for commit: 76d9d4c635b2349bf48057cfcfd0c55bbe6f56b8 available at: https://gist.github.com/17cc52f87d6f654740e274a35f0584ab<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8707-76d9d4c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8707-76d9d4c.dmg

howard-stearns:
closing here and merging to domain-check branch instead

PR #8708, opened 2016-10-03
Fix the crash on start up if default ambient map fails to load

samcake:
THis is fogbuz #1776
https://highfidelity.fogbugz.com/f/cases/1776/Crash-in-interface-model-SunSkyStage-setSunAmbientSphere

In case the default ambient map is not loaded properly, then itdoesn;t exist.
THe crashing call was using the defaultAMbienMap pointer assuming it was existing.
I added a check for it.

Removed an assert in GL45BackendTexture that was triggering for the wrong reason in debug

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -4361,8 +4361,12 @@ namespace render {
                     auto scene = DependencyManager::get<SceneScriptingInterface>()->getStage();
                     auto sceneKeyLight = scene->getKeyLight();
                     auto defaultSkyboxAmbientTexture = qApp->getDefaultSkyboxAmbientTexture();
-                    sceneKeyLight->setAmbientSphere(defaultSkyboxAmbientTexture->getIrradiance());
-                    sceneKeyLight->setAmbientMap(defaultSkyboxAmbientTexture);
+                    if (defaultSkyboxAmbientTexture) {
+                        sceneKeyLight->setAmbientSphere(defaultSkyboxAmbientTexture->getIrradiance());
+                        sceneKeyLight->setAmbientMap(defaultSkyboxAmbientTexture);
+                    } else {
+                        qWarning() << "Failed to get a valid Default SKybox Ambient Texture ? probably because it couldn't be find during initialization step";
+                    }
                     // fall through: render defaults skybox
                 } else {
                     break;
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -4365,7 +4365,9 @@ namespace render {
                         sceneKeyLight->setAmbientSphere(defaultSkyboxAmbientTexture->getIrradiance());
                         sceneKeyLight->setAmbientMap(defaultSkyboxAmbientTexture);
                     } else {
-                        qWarning() << "Failed to get a valid Default SKybox Ambient Texture ? probably because it couldn't be find during initialization step";
+                        static QString repeatedMessage
+                            = LogHandler::getInstance().addRepeatedMessageRegex(
+                            "Failed to get a valid Default Skybox Ambient Texture ? probably because it couldn't be find during initialization step");
                     }
                     // fall through: render defaults skybox
                 } else {
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -2170,7 +2170,7 @@ bool Application::event(QEvent* event) {
     // handle custom URL
     if (event->type() == QEvent::FileOpen) {
 
-		QFileOpenEvent* fileEvent = static_cast<QFileOpenEvent*>(event);
+        QFileOpenEvent* fileEvent = static_cast<QFileOpenEvent*>(event);
 
         QUrl url = fileEvent->url();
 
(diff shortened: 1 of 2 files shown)


samcake:
Test Plan:
temporarly change the name of the default ambient skymap in the resource folder
C:\Program Files\High Fidelity\resources\images\Default-Sky-9-ambient.jpg
Change the file name to something else, or maybe remove it from the folder.
Now start interface.exe 
maybe move to an empty corner of a domain without a Zone, but the very first frames while loading the domain should trigger it anyway.
On current MAster, this should crash
With this PR, it should NOT crash and blast the log with the warning 
"Failed to get a valid Default SKybox Ambient Texture ? probably because it couldn't be find during initialization step"

THis PR has no other effects, so everything else should be equal.

hifi-gustavo:
Build console output for commit: afcb5b85576cd06c4511a128cbf153c75350abce available at: https://gist.github.com/5e2f6c1edf18ba370c4b07b03b6ac81f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8708-afcb5b8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8708-afcb5b8.dmg

interface/src/Application.cpp
```diff
@@ -4361,8 +4361,12 @@ namespace render {
                     auto scene = DependencyManager::get<SceneScriptingInterface>()->getStage();
                     auto sceneKeyLight = scene->getKeyLight();
                     auto defaultSkyboxAmbientTexture = qApp->getDefaultSkyboxAmbientTexture();
-                    sceneKeyLight->setAmbientSphere(defaultSkyboxAmbientTexture->getIrradiance());
-                    sceneKeyLight->setAmbientMap(defaultSkyboxAmbientTexture);
+                    if (defaultSkyboxAmbientTexture) {
+                        sceneKeyLight->setAmbientSphere(defaultSkyboxAmbientTexture->getIrradiance());
+                        sceneKeyLight->setAmbientMap(defaultSkyboxAmbientTexture);
+                    } else {
+                        qWarning() << "Failed to get a valid Default SKybox Ambient Texture ? probably because it couldn't be find during initialization step";
```
sethalves: Is this going to go crazy with logging?  Maybe there should be a addRepeatedMessageRegex here?

ZappoMan: yeah - let's not flood the log... let's add the suppression of repeated items

libraries/gpu-gl/src/gpu/gl45/GL45BackendTexture.cpp
```diff
@@ -511,9 +511,7 @@ void GL45Texture::stripToMip(uint16_t newMinMip) {
     _minMip = newMinMip;
     // Re-sync the sampler to force access to the new mip level
     syncSampler();
-    size_t oldSize = _size;
```
ZappoMan: do we need to remove this?

samcake: it s triggering on linux otherwise, it s not used

interface/src/Application.cpp
samcake: Just did

hifi-gustavo:
Build console output for commit: 0189e3aea7a08e66be47103472f8dfcf62f95012 available at: https://gist.github.com/5e2f6c1edf18ba370c4b07b03b6ac81f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8708-0189e3a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8708-0189e3a.dmg

hifi-gustavo:
Build console output for commit: 5adedbbb8bfaec1177b406bb0b477a0eb9759534 available at: https://gist.github.com/5e2f6c1edf18ba370c4b07b03b6ac81f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8708-5adedbb.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8708-5adedbb.dmg

libraries/gpu-gl/src/gpu/gl45/GL45BackendTexture.cpp
ZappoMan: ok

interface/src/Application.cpp
```diff
@@ -4361,8 +4361,13 @@ namespace render {
...
                     auto sceneKeyLight = scene->getKeyLight();
                     auto defaultSkyboxAmbientTexture = qApp->getDefaultSkyboxAmbientTexture();
-                    sceneKeyLight->setAmbientSphere(defaultSkyboxAmbientTexture->getIrradiance());
-                    sceneKeyLight->setAmbientMap(defaultSkyboxAmbientTexture);
+                    if (defaultSkyboxAmbientTexture) {
+                        sceneKeyLight->setAmbientSphere(defaultSkyboxAmbientTexture->getIrradiance());
+                        sceneKeyLight->setAmbientMap(defaultSkyboxAmbientTexture);
+                    } else {
+                        static QString repeatedMessage = LogHandler::getInstance().addRepeatedMessageRegex(
+                            "Failed to get a valid Default Skybox Ambient Texture ? probably because it couldn't be find during initialization step");
```
ZappoMan: AFAIK - this won't actually print the first log message... I believe you still want to do the `qWarning() << `

OmegaHeron:
Testing Windows 10
Confirmed current master with missing jpg crashes.
This PR no crash on missing file.  Also insured it work both missing file and from an totally non-existent cache.
Standard smoke tests pass.  Everything I looked at (my own content, dev-playa and dev-demo) looks same this PR as master.
I didn't see any log spam about Failed to get ambient sky texture - that's why I went back and double checked I have made source jpg unavailable and removed cache files.

samcake:
Thanks @OmegaHeron 

Like zappoman said, apparently i neededan extra call for the message to display.
I ll add something shortly
THanks for double checking!

PR #8692, opened 2016-09-30
Expose triggerOption to scripts, plus _physicsEnabled

davidkelly:
Now any menu item can be triggered, so you can do `Menu.triggerOption("Reload content (Clears all caches))"` or `Window.isPhysicsEnabled()`.  Allows our domain load test scripts to clear caches, detect when things have loaded.

PR #8709, opened 2016-10-03
initial domain-check script

howard-stearns:
First version of domain check script.

### DIFF
--- a/scripts/developer/tests/performance/domain-check.js
+++ b/scripts/developer/tests/performance/domain-check.js
@@ -1,0 +1,201 @@
+"use strict";
+/*jslint vars: true, plusplus: true*/
+/*globals Script, MyAvatar, Quat, Render, ScriptDiscoveryService, Window, LODManager, Entities, print*/
+//
+//  loadedMachine.js
+//  scripts/developer/tests/
+//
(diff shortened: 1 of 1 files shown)


howard-stearns:
Known issues, which will be addressed separately later, are:
- Refers to tribbles on my github pages. Will likely be addressed first by moving those three files to AWS and changing the reference here, and then later by making /~/ meaningful for entity script urls and within entity scripts.
- Requires manual cache clearing per prompts, and doesn't check for physics ready when checking load times. Will be addressed after https://github.com/highfidelity/hifi/pull/8692 is merged.
- Refers to non-dev domains, which is fine for now, but won't be fine when this script depends on ac fixes that are for a time only running on dev domains.
- Doesn't summon robot avatars.

hifi-gustavo:
Build console output for commit: ee801f4cc2ab13176392872dd75684896274726a available at: https://gist.github.com/8f85b54bcef176ae9d985edc0603d36e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8709-ee801f4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8709-ee801f4.dmg

PR #8714, opened 2016-10-03
Fix potential race condition on static memory

Atlante45:
Test plan:
- Clear your cache and go you several different domains with ATP assets.
- Make sure everything loads fine.
- Try uploading a couple assets to an ATP server to make sure this still works too.

### DIFF
--- a/libraries/networking/src/udt/Connection.cpp
+++ b/libraries/networking/src/udt/Connection.cpp
@@ -45,6 +45,20 @@ Connection::Connection(Socket* parentSocket, HifiSockAddr destination, std::uniq
     // set the initial RTT and flow window size on congestion control object
     _congestionControl->setRTT(_rtt);
     _congestionControl->setMaxCongestionWindowSize(_flowWindowSize);
+
+    // Setup packets
+    static const int ACK_PACKET_PAYLOAD_BYTES = sizeof(_lastSentACK) + sizeof(_currentACKSubSequenceNumber)
+                            + sizeof(_rtt) + sizeof(int32_t) + sizeof(int32_t) + sizeof(int32_t);
+    static const int LIGHT_ACK_PACKET_PAYLOAD_BYTES = sizeof(SequenceNumber);
+    static const int ACK2_PAYLOAD_BYTES = sizeof(SequenceNumber);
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: 573d5898ccc8a010ee6a1728d105ad520d261f46 available at: https://gist.github.com/0e97ed62552c62da16bbb365bb36ea93<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8714-573d589.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8714-573d589.dmg

ZappoMan:
Can you explain your fix here?

Atlante45:
It appears, we are getting a "Pure Virtual Call" on this line:
```cpp
ack2Packet->reset(); // ack2Packet appears to be null.
```

Since `ack2Packet` is a static variable local that's only local to that function and is never destroyed manually, it leave roughly 2 possibilities.
1) `new` is failing due to lack of RAM available or other reasons (Very unlikely).
2) Pointer is getting corrupted from multiple threads accessing it (Slightly less unlikely).

The interface is supposed to only have 1 udp socket that sequentially update it's connections.
So 2 connections should never touch that static memory at the same time.
After walking up the call tree, this appears to be true, but there might be a weird state we get in when we switch domain or reset/rebind the socket where 2 sockets live at the same time.

This would cause several connections to touch the same memory at the same time, so I put those static packets in the `Connection` object so that that doesn't happen.

sethalves:
any chance this is the cause of https://highfidelity.fogbugz.com/f/cases/1779/pure-virtual-function-call-in-latest-beta ?

Atlante45:
Seems unlikely. Did you manage to get the symbols for those couple Qt calls?

OmegaHeron:
Testing Windows 10 - i5 6500 CPU - GTX 970 GPU

All standard tests pass.  Specifically tested uploading content to ATP, all good.

Repeated cache clears in ATP heavy scene (standard home), my own content and dev-playa, dev-demo.  All good.  Completely removed cache (as in deleted from disk manually).  All good.

All content loaded as expected

QA Pass.

PR #8717, opened 2016-10-03
Remove frame lag and jitter from near grabbed physical objects.

hyperlogic:
In the game loop, physics occurs before avatar update.  Before this PR, when the avatar is moved during avatar update, near grabbed objects will not pick up this move until one frame later, when the physics is run on the next update.

After this PR, near grabbed objects are adjusted to reflect any position or rotation change that occurred during the avatar update.  This removes lag and jitter in the motion of held objects during driving and re-centering.

### DIFF
--- a/interface/src/avatar/AvatarActionHold.cpp
+++ b/interface/src/avatar/AvatarActionHold.cpp
@@ -25,14 +25,25 @@ AvatarActionHold::AvatarActionHold(const QUuid& id, EntityItemPointer ownerEntit
 {
     _type = ACTION_TYPE_HOLD;
     _measuredLinearVelocities.resize(AvatarActionHold::velocitySmoothFrames);
+
+    auto myAvatar = DependencyManager::get<AvatarManager>()->getMyAvatar();
+    if (myAvatar) {
+        myAvatar->addHoldAction(this);
+    }
+
 #if WANT_DEBUG
-    qDebug() << "AvatarActionHold::AvatarActionHold";
+    qDebug() << "AvatarActionHold::AvatarActionHold" << (void*)this;
 #endif
 }
 
 AvatarActionHold::~AvatarActionHold() {
+    auto myAvatar = DependencyManager::get<AvatarManager>()->getMyAvatar();
+    if (myAvatar) {
+        myAvatar->removeHoldAction(this);
+    }
+
 #if WANT_DEBUG
-    qDebug() << "AvatarActionHold::~AvatarActionHold";
+    qDebug() << "AvatarActionHold::~AvatarActionHold" << (void*)this;
 #endif
 }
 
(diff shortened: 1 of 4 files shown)


hyperlogic:
https://highfidelity.fogbugz.com/f/cases/448/Objects-held-in-hand-jitter-while-turning-walking-flying

hyperlogic:
Test plan:
   * Enter the dev-demo domain while wearing a HMD with hand controllers
   * Ensure that Settings > Advanced controller motion is enabled.
   * grab a block in one of your hands.
   * With the right controller rotate your avatar in place.  The object in your hand should be solid relative to the hand controller blue ball.  Even if Avatar > Settings > Snap turns are enabled or disabled.
   * With the left controller strafe sideways.  The object in your hand should be solid relative to the blue ball.

hifi-gustavo:
Build console output for commit: 327fcc970b85e7acecc88527fb6d395269a9ef16 available at: https://gist.github.com/8cf52c528fa2ff631fda379784938a93<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8717-327fcc9.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8717-327fcc9.dmg

interface/src/avatar/AvatarActionHold.cpp
```diff
@@ -460,3 +471,40 @@ void AvatarActionHold::deserialize(QByteArray serializedArguments) {
...
+    btTransform worldTrans = rigidBody->getWorldTransform();
+    AnimPose worldBodyPose(glm::vec3(1), bulletToGLM(worldTrans.getRotation()), bulletToGLM(worldTrans.getOrigin()));
+
+    // transform the body transform into sensor space with the prePhysics sensor-to-world matrix.
+    // then transform it back into world uisng the postAvatarUpdate sensor-to-world matrix.
+    AnimPose newWorldBodyPose = postAvatarUpdateRoomPose * prePhysicsRoomPose.inverse() * worldBodyPose;
+
+    worldTrans.setOrigin(glmToBullet(newWorldBodyPose.trans));
+    worldTrans.setRotation(glmToBullet(newWorldBodyPose.rot));
+    rigidBody->setWorldTransform(worldTrans);
```
sethalves: I doubt this setWorldTransform is needed.

hyperlogic: Turns out it is, I'm not 100% certain why.  But without this I do get occasional glitches.  I think it has to do with frames where there are no physics timesteps, which can happen somewhat rarely.

Nex-Pro:
tested on win 10

Tested with Thoys.

### PASS

Block was as steady as I could hold it.  No glitches observed. Tried strafing spinning waving and flying.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -2257,3 +2263,25 @@ glm::vec3 MyAvatar::getAbsoluteJointTranslationInObjectFrame(int index) const {
...
+    std::lock_guard<std::mutex> guard(_holdActionsMutex);
+    auto iter = std::find(std::begin(_holdActions), std::end(_holdActions), holdAction);
+    if (iter != std::end(_holdActions)) {
+        _holdActions.erase(iter);
+    }
+}
+
+void MyAvatar::updateHoldActions(const AnimPose& prePhysicsPose, const AnimPose& postUpdatePose) {
+    std::lock_guard<std::mutex> guard(_holdActionsMutex);
+    for (auto& holdAction : _holdActions) {
```
sethalves: without a lock on the entity tree, I'm not sure this is safe.

hyperlogic: Ok, I added an entity tree write lock around these calls.

hifi-gustavo:
Build console output for commit: dcd425a165802880321ea23a439c1854495c1159 available at: https://gist.github.com/8cf52c528fa2ff631fda379784938a93<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8717-dcd425a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8717-dcd425a.dmg

PR #8723, opened 2016-10-04
Dk/load time poll

davidkelly:
This is a merge of the commit that added `Window.isPhysicsEnabled` and `Menu.triggerOption(...)`, and then a small reworking of the `domain-check.js` script.  I also added a new message one could use to see when physics is enabled (`Window.physicsEnabled`).  However saw some annoying race conditions (which could be a bug in my script changes), so decided to just poll until we know that everything has been downloaded, and physics is enabled.  Will want to revisit this soon, as I'm sure I'm just stupid/sleepy right now.  But this does work.


hifi-gustavo:
Build console output for commit: f3c22ea03c66e48566b20de3358c89175e5f00c4 available at: https://gist.github.com/ac84ba9b8c500dbf3439f9c2bc0cab81<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8723-f3c22ea.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8723-f3c22ea.dmg

howard-stearns:
I don't understand what's gone wrong. The .cpp/.h should already be in the domain-check branch.

howard-stearns:
closing this, as I've instead pulled in just the polling parts.

This pr was based on an earlier version of domain-check, What I've done is roughly equivalent to resolving conflicts, and merging, but not quite:
- I didn't include the physics ready signalling. Not because we don't want it, but because we're not using it here, and I wanted to keep the diffs to a minimum.
- I didn't do the polling quite the same way -- partly to minimize diffs, and partly because I want to be able to switch to signal connections later with minimum changes.

PR #8725, opened 2016-10-04
possible fix for domain-server crash-on-exit

sethalves:


hifi-gustavo:
Build console output for commit: 7261fcae4a8860a6afe5c570a1851db6caa2dddb available at: https://gist.github.com/cd13e0cfa6e9d3f4e389531b04166291<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8725-7261fca.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8725-7261fca.dmg

sethalves:
to test:
- run servers from this PR
- change domain-server settings, restart domain-server
- make sure the settings stuck and everything is okay

sethalves:
https://highfidelity.fogbugz.com/f/cases/1783/Domain-Server-Crashes-check-the-crashes-channel-hosted-domains-are-crashing

libraries/shared/src/SettingInterface.cpp
```diff
@@ -34,7 +34,6 @@ namespace Setting {
         DependencyManager::destroy<Manager>();
 
         //
-        globalManager->deleteLater();
```
sethalves: before this PR, if I put QThread::sleep(2); right here, the domain-server crashes on exit every time.

ZappoMan:
How do we expect that this will impact the client? Isn't this code shared? Don't we need to include the client on this test?

sethalves:
Yes, you're right.  I've updated the test plan.

libraries/shared/src/SettingInterface.cpp
sethalves: ... and so does interface.

Atlante45: I'm not surprised, we were just getting lucky.
Something specific to gcc's codegen must be causing a context switch here to be more frequent.

sethalves:
#8742 contains everything here. closing this one

PR #8730, opened 2016-10-04
Add retrying of script requests

huffman:


hifi-gustavo:
Build console output for commit: 9dfa3e4a17b40fc2b2865c6470f895782f4172db available at: https://gist.github.com/8cd06e332a155fed37d3200d5ccd0226<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8730-9dfa3e4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8730-9dfa3e4.dmg

huffman:
This PR adds retrying of script requests. This means that if an http/atp/file request fails, it will be retried at some point in the future. The time it waits between each subsequent request increases after each fail (200 ms, 400 ms, 800 ms, 1600 ms, etc.), and ultimately fails after 5 failures.

**If you have scripting experience it would help in testing this PR**

Testing is a little tricky, here is what I would recommend doing:

Here is what you will want to test:

 1. HTTP script requests are retried when loading a client script, entity script, and a script that was `Script.include()`'d.
 1. ATP script requests are retried when loading an entity and a script that was `Script.include()`'d.

To test client scripts:
 1. Try loading a client script that has a bad http url (say, `http://some-bad-url.com/doesnotexist`) and paying attention to the logs to see if the script is rerequested and ultimately failed.
    1. If you can do this to a local http server where you add the file the request is looking for before it fails, that would be even better.

To test entity scripts:
 1. Try using bad ATP and HTTP urls on an entity script.
   1. With ATP, upload the file to the asset server before the rerequests completely fail, confirm that the next request is successful

To test Script.include:
 1. Create a script that has an `include()` to a url that doesn't exist, confirm that it retries.
   1. With ATP, do the same as with entity scripts and confirm that the include is successful once you upload the asset to the asset server.

libraries/networking/src/AssetClient.cpp
```diff
@@ -409,20 +411,24 @@ void AssetClient::handleCompleteCallback(const QWeakPointer<Node>& node, Message
...
+        qCDebug(asset_client) << "Got completed asset for a message that doesn't exist";
         return;
     }
 
 
     if (message->failed()) {
+        qCDebug(asset_client) << "Got failed asset";
         callbacks.completeCallback(false, AssetServerError::NoError, QByteArray());
     } else {
+        qCDebug(asset_client) << "Got asset";
```
Atlante45: I'm afraid this is gonna clog our log more than it already is.
I I would also set those debugs above as warnings, no?

libraries/networking/src/AssetRequest.cpp
```diff
@@ -148,6 +150,8 @@ void AssetRequest::start() {
             
             if (_error != NoError) {
                 qCWarning(asset_client) << "Got error retrieving asset" << _hash << "- error code" << _error;
+            } else {
+                qCDebug(asset_client) << "Received asset successfully: " << _hash;
```
Atlante45: Do we absolutely need that?
We already have 4-5 messages per downloaded assets.

libraries/networking/src/AssetResourceRequest.cpp
```diff
@@ -32,6 +37,25 @@ bool AssetResourceRequest::urlIsAssetHash() const {
     return hashRegex.exactMatch(_url.toString());
 }
 
+void AssetResourceRequest::setupTimer() {
+    Q_ASSERT(!_sendTimer);
+    static const int TIMEOUT_MS = 2000;
+
+    _sendTimer = new QTimer();
+    connect(this, &QObject::destroyed, _sendTimer, &QTimer::deleteLater);
```
Atlante45: You could just parent the timer here. (Not a blocker)

libraries/networking/src/AssetResourceRequest.cpp
```diff
@@ -102,11 +129,15 @@ void AssetResourceRequest::requestHash(const AssetHash& hash) {
...
-    connect(_assetRequest, &AssetRequest::progress, this, &AssetResourceRequest::progress);
+    connect(_assetRequest, &AssetRequest::progress, this, &AssetResourceRequest::onDownloadProgress);
     connect(_assetRequest, &AssetRequest::finished, this, [this](AssetRequest* req) {
         Q_ASSERT(_state == InProgress);
         Q_ASSERT(req == _assetRequest);
         Q_ASSERT(req->getState() == AssetRequest::Finished);
+
+        cleanupTimer();
+
+        qDebug() << "Asset request finished, " << req->getHash() << "error: " << req->getError();
```
Atlante45: Can we only print that as a warning in case of error?

libraries/script-engine/src/ScriptCache.h
```diff
@@ -51,7 +57,8 @@ private slots:
     ScriptCache(QObject* parent = NULL);
     
     Mutex _containerLock;
-    QMultiMap<QUrl, contentAvailableCallback> _contentCallbacks;
+    QMap<QUrl, ScriptRequest> _activeScriptRequests;
+    //QMultiMap<QUrl, contentAvailableCallback> _contentCallbacks;
```
Atlante45: Remove dead code

libraries/networking/src/AssetResourceRequest.cpp
huffman: Good call

huffman:
@Atlante45 I addressed your comments

hifi-gustavo:
Build console output for commit: 4901f386ca8a68e81a4f3b28def9667adc0e8d01 available at: https://gist.github.com/8cd06e332a155fed37d3200d5ccd0226<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8730-4901f38.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8730-4901f38.dmg

hifi-gustavo:
Build console output for commit: 4171fdd2e1372bb9465d3bae465769bd45b4467e available at: https://gist.github.com/8cd06e332a155fed37d3200d5ccd0226<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8730-4171fdd.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8730-4171fdd.dmg

PR #8734, opened 2016-10-05
Update entity scripting interface to allow modifications to unknown e…

huffman:
…ntities

### DIFF
--- a/libraries/entities/src/EntityScriptingInterface.cpp
+++ b/libraries/entities/src/EntityScriptingInterface.cpp
@@ -386,6 +386,15 @@ QUuid EntityScriptingInterface::editEntity(QUuid id, const EntityItemProperties&
         }
     });
 
+    // FIXME: We need to figure out a better way to handle this. Allowing these edits to go through potentially
+    // breaks avatar energy and entities that are parented.
+    //
+    // To handle cases where a script needs to edit an entity with a _known_ entity id but doesn't exist
+    // in the local entity tree, we need to allow those edits to go through to the server.
+    // if (!updatedEntity) {
+    //     return QUuid();
+    // }
+
     _entityTree->withReadLock([&] {
         EntityItemPointer entity = _entityTree->findEntityByEntityItemID(entityID);
         if (entity) {
--- a/libraries/entities/src/EntityScriptingInterface.cpp
+++ b/libraries/entities/src/EntityScriptingInterface.cpp
@@ -386,10 +386,6 @@ QUuid EntityScriptingInterface::editEntity(QUuid id, const EntityItemProperties&
         }
     });
 
-    if (!updatedEntity) {
-        return QUuid();
-    }
-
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 2f7e346f6ac26f0a97ca49ed338c15f694fcdd8c available at: https://gist.github.com/6c68a708e85751736ca8463e18fb28b4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8734-2f7e346.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8734-2f7e346.dmg

huffman:
This will require a small amount of scripting knowledge (and the console) to test. The change this PR makes is to allow you to make modifications to entities that are not in your local entity tree. An entity will generally not be in your local entity tree if it hasn't been in your view frustum or you haven't been near it.

 1. Create a red box entity far off in the distance (say, `10000,10000,10000`) and take note of the entity id
 1. Go back to `0,0,0`
 1. Restart interface
 1. Confirm that you do not have the entity in your local entity tree - you can do this by calling `Entities.getEntityProperties(THE_ENTITY_ID).color` and seeing that it is undefined
 1. Edit the entity to change its color
 1. Go to where your placed the entity and confirm that its color was updated despite it not having been in your entity tree when you made the edit

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -386,10 +386,6 @@ QUuid EntityScriptingInterface::editEntity(QUuid id, const EntityItemProperties&
         }
     });
 
-    if (!updatedEntity) {
```
ZappoMan: I think we should add a `FIXME` comment to come back and review this code... I'm looking at the code above this, and the early code definitely seems to assume that the server won't get an edit if the `updatedEntity` is false... The energy system stuff definitely seems to be broken by this change (although that's not actively used code)... the local only and parented entities may also be broken by this...

I won't block merge on this, but I think we should attempt to document this techical debt.

huffman: Good point, I will add a comment

hifi-gustavo:
Build console output for commit: c0c060ef40891113507bdd991fc8db5e101bdf8e available at: https://gist.github.com/6c68a708e85751736ca8463e18fb28b4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8734-c0c060e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8734-c0c060e.dmg

cozza13:
For some reason I can not see the comments from @captainoculus on here. So I am pasting in the omage he sent me 
pasted image at 2016_10_06 12_41 pm

PR #8735, opened 2016-10-05
Update teleport.js to ignore invisible or collisionless objects and disallow teleporting to  certain surfaces

huffman:


hifi-gustavo:
Build console output for commit: d50689b795b3f82b7d9c15db7ae0175ceded042d available at: https://gist.github.com/0e1a51b213684961a6a07fa50bc61368<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8735-d50689b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8735-d50689b.dmg

huffman:
To test, test the following changes:

Changes to teleport.js:
  1. Teleport ray no longer collides with invisible entities
  1. Teleport ray no longer collides with collisionless entities
  1. Teleport ray no longer collides with entities without a shape type set
  1. Smooth movement is removed
  1. You can now only teleport to surfaces that are pointing within some degree of UP
     1. This means you can't teleport into walls or a ceiling, or other surfaces that are extremely slanted
     1. This has some drawbacks, but we expect to address that in the future
     1. Test this with a variety of objects
        1. Models
        1. Cubes that are rotated in different ways (if you rotate a cube so that it's "upside down," can you still correctly teleport on to it?)

Minor change to handControllerGrab.js:
  1. The far grab ray will no longer intersect with invisible entities

Also test that there weren't any regressions with teleport.js and handControllerGrab.js, and also test that there weren't any regressions with selection and manipulating entities in edit.js, which makes heavy use of some of the APIs that were changed here.

scripts/system/away.js
```diff
@@ -274,6 +275,24 @@ function maybeGoAway() {
     }
 }
 
+function setEnabled(value) {
```
ZappoMan: What's this change all about? This doesn't appear to be teleport related.

Nex-Pro:
@huffman  How will you TP around in cellscience if this is implemented ?

huffman:
@Nex-Pro How do you tp in cell-science now? Are there places to tp to? It seems like an advanced-movement-only domain, which you can still do

huffman:
FWIW, I don't think this TP method is the perfect solution, but I think it's better in most cases.

RyanDowne:
Teleport ray no longer collides with invisible entities  - PASS
Teleport ray no longer collides with collisionless entities - PASS
Teleport ray no longer collides with entities without a shape type set - PASS
Smooth movement is removed- PASS
You can now only teleport to surfaces that are pointing within some degree of UP
This means you can't teleport into walls or a ceiling, or other surfaces that are extremely slanted
This has some drawbacks, but we expect to address that in the future - PASS
Test this with a variety of objects - PASS
Models -PASS
Cubes that are rotated in different ways (if you rotate a cube so that it's "upside down," can you still correctly teleport on to it?) - PASS

Minor change to handControllerGrab.js:

The far grab ray will no longer intersect with invisible entities -PASS
Also test: 

That there weren't any regressions with teleport.js and handControllerGrab.js
That there weren't any regressions with selection and manipulating entities in edit.js, which makes heavy use of some of the APIs that were changed here.
That away.js still works as expected - PASS - I went to demo and several other domains and was able to grab, equip etc without issue

PR #8736, opened 2016-10-05
Added user ability to disable rendering

druiz17:
The implementation of enabling the user to disable the rendering was done by first creating a clickable toggle rendering action for the menu item called rendering in the constructor of the menu class. In the beginning of the function paintGL in the application class, I checked if the toggle rendering action was clicked and if true the function would exit preventing the next frame to be rendered.


hifi-gustavo:

huffman:
ok to test

hifi-gustavo:
Build console output for commit: d56b690d48ad8b2cceb8363a7eb9c04d6f0dcfa4 available at: https://gist.github.com/e672387fca1b29bb6c5a2f945f43b7f4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8736-d56b690.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8736-d56b690.dmg

ZappoMan:
What is the "interface/src/Menu.h.gch" file?

sethalves:
what's interface/src/Menu.h.gch ?

interface/src/Application.cpp
```diff
@@ -1692,7 +1692,7 @@ void Application::initializeUi() {
 
 void Application::paintGL() {
     // Some plugins process message events, allowing paintGL to be called reentrantly.
-    if (_inPaint || _aboutToQuit) {
+    if (_inPaint || _aboutToQuit || Menu::getInstance()->isOptionChecked(MenuOption::ToggleRendering)) {
```
ZappoMan: instead of calling an expensive function like `Menu::getInstance()` and `isOptionChecked` you should handle the menu selection signal to set a member variable so that this check in `paintGL` can be as efficient as possible.

druiz17:
I think the "interface/src/Menu.h.gch" file was accidently generated on my size. I'll remove the file

cozza13:
I tested this and it seems to work. it pauses the sending of any new rendering. Within the HMD it creates the effect of the environment frozen at a point of time that you can move around.

druiz17:
Ok great. Do you guys still want me to submit the changes suggested above?

AndrewMeadows:
druiz17, no I think the importance of the optimization that @ZappoMan is pointing out would not be obvious to someone fresh out of university and unfamiliar with the challenges of maintaining a high frame rate for a VR HMD experience.

PR #8741, opened 2016-10-05
kill agents in avatar mixer when they stop being avatars

howard-stearns:
When an ac client becomes Agent.isAvatar true and then false, make the avatar actually go away on clients.

Additionally, don't spam identity packets for default avatars when identity hasn't actually changed..

### DIFF
--- a/assignment-client/src/Agent.cpp
+++ b/assignment-client/src/Agent.cpp
@@ -351,6 +351,21 @@ void Agent::setIsAvatar(bool isAvatar) {
             _avatarIdentityTimer->stop();
             delete _avatarIdentityTimer;
             _avatarIdentityTimer = nullptr;
+
+            // The avatar mixer never times out a connection (e.g., based on identity or data packets)
+            // but rather keeps avatars in its list as long as "connected". As a result, clients timeout
+            // when we stop sending identity, but then get woken up again by the mixer itself, which sends
+            // identity packets to everyone. Here we explicitly tell the mixer to kill the entry for us.
+            auto nodeList = DependencyManager::get<NodeList>();
+            auto packetList = NLPacketList::create(PacketType::KillAvatar, QByteArray(), true, true);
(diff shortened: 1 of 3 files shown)


howard-stearns:
The first is handled by the Agent.cpp change, and explained in a code comment.
The other files are for the second part.

howard-stearns:
Test plan:

1. Run scripts/developer/tests/performance/crowd-agent.js on a domain server. 
2. Run scripts/developer/tests/performance/summon.js as an interface script.  One or more avatars will appear near you, depending on how many instances of crowd-agent are running on that domain.  You many need to turn or back up to see them.  The Avatars count in stats should go up by that number.
3. Stop the summon.js interface script. The avatars should disappear and the Avatar count in stats should go back down.

Notes:
- The crowd-agent.js AC script does not work on Linux servers. That's being looked at separately.
- Because AC scripts cannot be local files, this script is at http://cdn.highfidelity.com/davidkelly/production/scripts/tests/performance/crowd-agent.js for convenience.
- It takes a long time to spin up a large number of AC scripts, so you probably want just 1-5 instances.
- The summon.js interface script will tell you how many are running if it is less than 25.

For the identity packet change:
- Run two clients,
- In one client, change the avatar between any avatar and the default, specifying the empty string for the default avatar appearance url.
- Observe that the changes are seen by the other client.
- You're probably not going to notice any bandwidth difference.

hifi-gustavo:
Build console output for commit: dc7a9f253cd46895d4b751d9f0940854f8960958 available at: https://gist.github.com/ee205f5e88f4727329c3d5e2b2aa14b7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8741-dc7a9f2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8741-dc7a9f2.dmg

zzmp:
should be two separate PRs

PR #8748, opened 2016-10-06
Add simple sound to avatars

davidkelly:
For now I just used the piano1.wav that howard suggested long ago
for another project.  We use Agent.playAvatarSound, and tell the
mixer to mix audio for each avatar by setting
Agent.setIsListeningToAudioStream to true.

### DIFF
--- a/scripts/developer/tests/performance/crowd-agent.js
+++ b/scripts/developer/tests/performance/crowd-agent.js
@@ -15,6 +15,9 @@
 //  a script like summon.js calls up to n avatars to be around you.
 
 var MESSAGE_CHANNEL = "io.highfidelity.summon-crowd";
+var soundIntervalId;
+var SOUND_POLL_INTERVAL = 500; // ms
+var SOUND_URL = "http://howard-stearns.github.io/models/sounds/piano1.wav";
 
 print('crowd-agent version 1');
 
@@ -41,9 +44,26 @@ function startAgent(parameters) { // Can also be used to update.
         Avatar.startAnimation(data.url, data.fps || 30, 1.0, (data.loopFlag === undefined) ? true : data.loopFlag, false, data.startFrame || 0, data.endFrame);
     }
     print('crowd-agent avatars started');
+    Agent.isListeningToAudioStream = true;
+    soundIntervalId = playSound();   
+    print('crowd-agent sound loop started');
+}
+function playSound() {
+    // Load a sound
+    var sound = SoundCache.getSound(SOUND_URL);
+    function loopSound(sound) {
+        // since we cannot loop, for now lets just see if we are making sounds.  If
+        // not, then play a sound.
+        if (!Agent.isPlayingAvatarSound) {
+            Agent.playAvatarSound(sound);
+        }
+    };
+    return Script.setInterval(function() {loopSound(sound);}, SOUND_POLL_INTERVAL);
 }
 function stopAgent(parameters) {
     Agent.isAvatar = false;
+    Agent.isListeningToAudioStream = false;
+    Script.clearInterval(soundIntervalId);
     print('crowd-agent stopped', JSON.stringify(parameters), JSON.stringify(Agent));
 }
 


hifi-gustavo:
Build console output for commit: 73aafaf518e0a9bb23677f159069359d4a63b128 available at: https://gist.github.com/3ec94d4f9c3f368c327359c05c9055ed<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8748-73aafaf.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8748-73aafaf.dmg

scripts/developer/tests/performance/crowd-agent.js
```diff
@@ -15,6 +15,9 @@
 //  a script like summon.js calls up to n avatars to be around you.
 
 var MESSAGE_CHANNEL = "io.highfidelity.summon-crowd";
+var soundIntervalId;
+var SOUND_POLL_INTERVAL = 500; // ms
+var SOUND_URL = "http://howard-stearns.github.io/models/sounds/piano1.wav";
```
howard-stearns: Can we make the sound data (sound, volume, loop) be a parameter passed to SUMMON? Note that SUMMON takes a number of parameters which can be used to start or change a value for the robot. (In the case of animation, there isn't a provision for stopping the animation, because one will always want to start 'idle' instead.)
The summon.js script would then be modified here to start the sound and control the parameters.

There are some subtle reasons for this:
- 25 robots making noise is interesting, but not realistic or likely to pass. 
  - If all of our tests fail, then the tool won't be used and no progress will be made.
  - If our system is good, then when working in a realistic setting no one thing will be the long pole. So pushing if we always fail on sound first in an unrealistic test, we'll never test anything else. (We still want to be able to push hard on sound, just that it shouldn't always do so.)
- It is much much easier to change a local interface script like summon.js, then to change an ac script like crowd-agent.js. So if we want to tune or explore anything, the parameters realistically need to be on the interface side.

scripts/developer/tests/performance/crowd-agent.js
```diff
@@ -41,9 +44,26 @@ function startAgent(parameters) { // Can also be used to update.
...
+    // Load a sound
+    var sound = SoundCache.getSound(SOUND_URL);
+    function loopSound(sound) {
+        // since we cannot loop, for now lets just see if we are making sounds.  If
+        // not, then play a sound.
+        if (!Agent.isPlayingAvatarSound) {
+            Agent.playAvatarSound(sound);
+        }
+    };
+    return Script.setInterval(function() {loopSound(sound);}, SOUND_POLL_INTERVAL);
```
howard-stearns: loopSound is cute, but given the above comments, I'm not sure the complexity (on this side) is worth it. I would say that it would be desirable but optional to support a loop parameter in an audioData parameter to SUMMON.

howard-stearns:
It buzzes for me (and then fades out) when I stop the summon.js script.

scripts/developer/tests/performance/crowd-agent.js
```diff
@@ -15,6 +15,9 @@
 //  a script like summon.js calls up to n avatars to be around you.
 
 var MESSAGE_CHANNEL = "io.highfidelity.summon-crowd";
+var soundIntervalId;
+var SOUND_POLL_INTERVAL = 500; // ms
+var SOUND_URL = "http://howard-stearns.github.io/models/sounds/piano1.wav";
 
 print('crowd-agent version 1');
```
howard-stearns: Recommend we spin version number when we update, so that we can tell from logs what was running. (Instead of relying on our understanding of caching.)

scripts/developer/tests/performance/crowd-agent.js
```diff
@@ -41,9 +44,26 @@ function startAgent(parameters) { // Can also be used to update.
...
+        // not, then play a sound.
+        if (!Agent.isPlayingAvatarSound) {
+            Agent.playAvatarSound(sound);
+        }
+    };
+    return Script.setInterval(function() {loopSound(sound);}, SOUND_POLL_INTERVAL);
 }
 function stopAgent(parameters) {
     Agent.isAvatar = false;
+    Agent.isListeningToAudioStream = false;
```
howard-stearns: It looks like the agent doesn't send PacketType::SilentAudioFrame on silence when isListeningToAudioStream, so if we're trying to exercise realistic behavior, we should:
1) have some gaps (which I claim should be controlled by summon.js) and
2) set Agent.isListeningToAudioStream to true.

howard-stearns:
Closing in favor of https://github.com/highfidelity/hifi/pull/8759 and https://github.com/highfidelity/hifi/pull/8751

PR #8749, opened 2016-10-06
Make AssetClient thread safe

Atlante45:
-- Make the Asset Client's callback maps thread safe.

BugzId: 470

Test plan:
- Load ATP assets with clear cache
- Upload an ATP asset
- Rename an ATP asset
- Delete an ATP asset
- Compare home content load time between current release and this PR.
    - Make sure there is no significant increase in load time


hifi-gustavo:
Build console output for commit: 11a3abb13d8d8c7601de74f03537cbea1d5b837e available at: https://gist.github.com/f056332a1e27d8db349ffab5ae88f7ab<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8749-11a3abb.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8749-11a3abb.dmg

huffman:
The original design of AssetClient was for it to be single threaded and for the AssetRequest and related objects to provide a thread-safe interface to making ATP requests. If AssetClient really is being called from different threads, and we don't want to fix that issue, then adding locks may not be enough. If AssetClient is getting called from different threads, then it means that the callbacks themselves might be called on a thread that is different from the thread that the call was originally made on.

OmegaHeron:
Testing Windows 10 - i5 6500 CPU - GTX 970 GPU

Standard home scene loads normally (Interface/stack localhost all) - Load time from a cached scene remains same as current master.  Load time with no cache present (physically deleted from local HD) remains same.  Load time from dev menu reload cache remains same.

ATP operations (upload, place, delete) pass.

CPU use shows no change from master.

Tested locations dev-demo, playa, standard home scene (localhost) and my own content.  All seem normal.

Skybox visual never loaded at playa - but that seems to be not related to this PR - it seems to do same in master (variably) as well.

PR #8755, opened 2016-10-07
openvr deadlock and crash on exit fixes

hyperlogic:
* fixes for deadlock when switching rapidly between desktop and openvr.
* workaround openvr crash on exit by calling submit before calling VR_Shutdown

### DIFF
--- a/libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
+++ b/libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
@@ -124,7 +124,6 @@ void HmdDisplayPlugin::uncustomizeContext() {
         batch.setFramebuffer(_compositeFramebuffer);
         batch.clearColorFramebuffer(gpu::Framebuffer::BUFFER_COLOR0, vec4(0));
     });
-    internalPresent();
     _overlayRenderer = OverlayRenderer();
     Parent::uncustomizeContext();
 }
--- a/plugins/openvr/src/OpenVrHelpers.cpp
+++ b/plugins/openvr/src/OpenVrHelpers.cpp
@@ -103,6 +103,15 @@ void releaseOpenVrSystem() {
             #if DEV_BUILD
                 qCDebug(displayplugins) << "OpenVR: zero refcount, deallocate VR system";
             #endif
+
+            // HACK: workaround openvr crash, call submit right before VR_Shutdown.
+            const GLuint NON_ZERO_GL_TEXTURE_HANDLE = 1;
+            vr::Texture_t vrTexture{ (void*)NON_ZERO_GL_TEXTURE_HANDLE, vr::API_OpenGL, vr::ColorSpace_Auto };
+            static vr::VRTextureBounds_t OPENVR_TEXTURE_BOUNDS_LEFT{ 0, 0, 0.5f, 1 };
+            static vr::VRTextureBounds_t OPENVR_TEXTURE_BOUNDS_RIGHT{ 0.5f, 0, 1, 1 };
+            vr::VRCompositor()->Submit(vr::Eye_Left, &vrTexture, &OPENVR_TEXTURE_BOUNDS_LEFT);
+            vr::VRCompositor()->Submit(vr::Eye_Right, &vrTexture, &OPENVR_TEXTURE_BOUNDS_RIGHT);
+
             vr::VR_Shutdown();
             _openVrQuitRequested = false;
             activeHmd = nullptr;
--- a/plugins/openvr/src/OpenVrHelpers.cpp
+++ b/plugins/openvr/src/OpenVrHelpers.cpp
@@ -104,9 +104,9 @@ void releaseOpenVrSystem() {
                 qCDebug(displayplugins) << "OpenVR: zero refcount, deallocate VR system";
             #endif
 
-            // HACK: workaround openvr crash, call submit right before VR_Shutdown.
-            const GLuint NON_ZERO_GL_TEXTURE_HANDLE = 1;
-            vr::Texture_t vrTexture{ (void*)NON_ZERO_GL_TEXTURE_HANDLE, vr::API_OpenGL, vr::ColorSpace_Auto };
+            // HACK: workaround openvr crash, call submit with an invalid texture, right before VR_Shutdown.
+            const GLuint INVALID_GL_TEXTURE_HANDLE = -1;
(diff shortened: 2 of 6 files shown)


hyperlogic:
https://highfidelity.fogbugz.com/f/cases/1813/Crash-in-d3d11-CLayeredObject-CContainedObject-Release

hifi-gustavo:
Build console output for commit: a4c7e832b5d5fc5d46ddc0d39bd6af31ccdaff5d available at: https://gist.github.com/c3b4de9602188e53a78977f33920b2d5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8755-a4c7e83.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8755-a4c7e83.dmg

libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
```diff
@@ -124,7 +124,6 @@ void HmdDisplayPlugin::uncustomizeContext() {
         batch.setFramebuffer(_compositeFramebuffer);
         batch.clearColorFramebuffer(gpu::Framebuffer::BUFFER_COLOR0, vec4(0));
     });
-    internalPresent();
```
hyperlogic: So, yeah...  There are three threads in play.
* The Main Thread: It calls PresentThread::setNewDisplayPlugin(),  It pushes the new displayPlugin onto the _newPluginQueue then blocks on PresentThread::_condition and until the _newPluginQueue is empty.  
* The OpenVR Submit Thread:  Increments the OpenVRSubmitThread::_presentCount after every present.
* The Present Thread: picks up the new display plugin off of the PresentThread::_newPluginQueue.  Then calls uncustomizeContext on the previous displayPlugin we are switching away from.

The deadlock would occur if the OpenVRSubmitThread object did not have a texture set up yet and we switch to desktop mode.  Because the texture was not present, OpenVRSubmitThread will never increment the presentCount.  However the Present Thread during the call to uncustomizeContext  will call internalPresent, which in turn calls OpenVRSubmitThread::waitForPresent(), which... wait for it... waits for the presentCount to be incremented.

The solution, is to stop calling internalPresent during uncustomizeContext().  This did not seem to have any negative implications... and avoids the deadlock.

plugins/openvr/src/OpenVrHelpers.cpp
```diff
@@ -103,6 +103,15 @@ void releaseOpenVrSystem() {
...
             #endif
+
+            // HACK: workaround openvr crash, call submit right before VR_Shutdown.
+            const GLuint NON_ZERO_GL_TEXTURE_HANDLE = 1;
+            vr::Texture_t vrTexture{ (void*)NON_ZERO_GL_TEXTURE_HANDLE, vr::API_OpenGL, vr::ColorSpace_Auto };
+            static vr::VRTextureBounds_t OPENVR_TEXTURE_BOUNDS_LEFT{ 0, 0, 0.5f, 1 };
+            static vr::VRTextureBounds_t OPENVR_TEXTURE_BOUNDS_RIGHT{ 0.5f, 0, 1, 1 };
+            vr::VRCompositor()->Submit(vr::Eye_Left, &vrTexture, &OPENVR_TEXTURE_BOUNDS_LEFT);
+            vr::VRCompositor()->Submit(vr::Eye_Right, &vrTexture, &OPENVR_TEXTURE_BOUNDS_RIGHT);
+
```
hyperlogic: I ran across this workaround in this [openvr issue](https://github.com/ValveSoftware/openvr/issues/242).  It fixed the crash issue for me.

libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
jherico: The point of the internal present at the shutdown is to not leave a final static frame frozen on the screen.  I believe this is mostly an issue with Oculus HMDs.  Perhaps there's a way we can resolve that without causing the deadlock though.

plugins/openvr/src/OpenVrHelpers.cpp
jherico: I'm wary of using an arbitrary texture here.  We don't know what, if anything, is actually being stored at texture ID 1.   For that matter, we don't know that it's a 2D textures that's compatible with OpenVR (maybe it's a framebuffer depth/stencil attachment).  If we want to do this, we need to create a specific texture specifically to serve this purpose, even if it's only a 1x1 black texture.

hyperlogic:
QA test plan:
* While using a Vive HMD: Use the Alt+1 and Alt-5 (sometimes Alt-4) buttons to rapidly switch between desktop and HMD modes.   It should never lock up interface.
* When using the Vive in HMD mode.  Exiting, by Pressing Ctrl-q, or any other method should no longer report a crash.

libraries/display-plugins/src/display-plugins/hmd/HmdDisplayPlugin.cpp
hyperlogic: So, I changed the code so that we still call internalPresent for oculus, but not for openvr.

plugins/openvr/src/OpenVrHelpers.cpp
hyperlogic: Here's what I learned.  If you call Submit with a valid texture openvr will still crash in VR_Shutdown (I was using the TextureCache::getWhiteTexture).  The fix, oddly, is to actually call Submit with an invalid opengl texture.  This is nuts, but it's the only thing I've found that prevents openvr from crashing.  So I renamed the constant to INVALID_GL_TEXTURE and set it to -1.

hifi-gustavo:
Build console output for commit: ec801cb9d8ba7b692ba00a30ade5e4051f4d2076 available at: https://gist.github.com/c3b4de9602188e53a78977f33920b2d5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8755-ec801cb.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8755-ec801cb.dmg

thoys:
## QA Testing report

- Repeatedly switched between desktop and HMD mode (Multiple times per second) and no crash 👍 
- Exited while in Vive HMD Mode without a crash 👍 



- Bug that was not handled in this PR. when exited in Vive HMD mode, and starting up again causes a crash if SteamVR is not running (or not set to run automatically), fallback to desktop mode would be better than a crash.

QA Passed

PR #8766, opened 2016-10-08
fix reticle depth when not moving mouse, but moving head relative to HUD

ZappoMan:


ZappoMan:
Test Plan:
- Basic smoke test to make sure this doesn't introduce some crazy performance problem
- Using an HMD & mouse
    - point at a 2D overlay (like the HUD window, or a dialog box)
    - notice the mouse should appear to rest at the depth of the surface of the 2D overlay 
    - *without moving the mouse* move your head in room scale back and forward as if closer to and further away from the HUD
    - notice that the reticle should adjust to stay on the surface of the 2D overlay

hifi-gustavo:
Build console output for commit: 4ec9299930c18495108da1125d66cdb89ec654ab available at: https://gist.github.com/b7be800cfa5d757b649c41e244f303e7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8766-4ec9299.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8766-4ec9299.dmg

howard-stearns:
Not a blocker, but it seems like we could now get rid of the other calls to hudReticleDistance().

Also, I guess we no longer have to shut down timers when a script ends. It feels like it would be confusing for anyone looking at the code, when we explicitly do that for some timers and not others within the same script, as will now be the case here.

ZappoMan:
I think it's still preferable to set the depth on a mouse move, since that can happen at faster than 20hz.

RyanDowne:
I did not do a smoke test but this functionally worked per test notes.

PR #8691, opened 2016-09-30
Room Scale HMD comfort features.

hyperlogic:
The goal of this PR is to increase comfort for HMD users, by eliminating instances where we move the HMD camera unexpectedly.

When walking in room scale you are no longer pushed by collision.  Your avatar will attempt to follow your HMD camera position in a physical way.  If your body cannot reach your HMD camera, because it is blocked by collision, you will separate from your body and the hand controller balls will no longer match your avatar's hands.  However, you can still grab and manipulate objects with the controller balls.  You can rejoin your body by walking back into it, teleporting, or moving out of collision in such a way that it can walk back to you.

Other features include:
- "Advanced Movement For Hand Controllers" in Settings menu, now drives the HMD with linear velocity and without collision.
- Snappier teleport behavior for HMD users.
- The avatar's hips will no longer embed into collision.
- The avatar collision can now drive over small bumps and obstructions.
- Avatar drive speed has been reduced to 2 m/s.
- Avatar no longer plays walk animation when motion is impeded by collision.
- Avatar walk animations for large or small avatars should have less foot sliding.

PR #8787, opened 2016-10-11
prevent avatar from recentering below ground

AndrewMeadows:
* fix bug where avatar would tunnel under floor when recentering on HMD position

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -251,6 +251,15 @@ void MyAvatar::centerBody() {
     auto worldBodyPos = extractTranslation(worldBodyMatrix);
     auto worldBodyRot = glm::normalize(glm::quat_cast(worldBodyMatrix));
 
+    if (_characterController.getState() == CharacterController::State::Ground) {
+        // the avatar's physical aspect thinks it is standing on something
+        // therefore need to be careful to not "center" the body below the floor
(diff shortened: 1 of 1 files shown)


AndrewMeadows:
Note: this is a PR into the out-of-body-experience branch which is currently sitting in PR #8691.  It fixes a bug that has the following repro recipe:

(1) Login with Vive while standing
(2) Navigate to Welcome domain
(3) Sit on floor
(4) Press ESC twice.  The first press goes into PAUSED mode, the second press unpauses and recenters the avatar so that its eyes line up with the known HMD position (which happens to be inside or below the floor).
--> on bugged versions the avatar's physical body will end up in deep penetration with the floor and will be resolved out, often out the bottom.
--> on fixed version the avatar's physical body will NOT be dropped... it the avatar was standing on the ground when the recenter happens.  If the avatar is hovering then the recenter will properly move the avatar's physical body down accordingly.

hifi-gustavo:
Build console output for commit: ef465d44438cbb05a48822db817f12f672443883 available at: https://gist.github.com/fe97280e2b367337895329ea5c4ae5ff<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8787-ef465d4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8787-ef465d4.dmg

PR #8792, opened 2016-10-12
allow hand controller operations when Interface doesn't have focus, but ONLY when in hmd

howard-stearns:
Fixes https://highfidelity.fogbugz.com/f/cases/1831/Laser-pointer-as-mouse-doesn-t-work-if-application-is-not-top-level-window

### DIFF
--- a/scripts/system/controllers/handControllerPointer.js
+++ b/scripts/system/controllers/handControllerPointer.js
@@ -471,7 +471,12 @@ function update() {
     if (!Menu.isOptionChecked("First Person")) {
         return off(); // What to do? menus can be behind hand!
     }
-    if (!Window.hasFocus() || !Reticle.allowMouseCapture) {
+    if ((!Window.hasFocus() && !HMD.active) || !Reticle.allowMouseCapture) {
+        // In desktop it's pretty clear when another app is on top. In that case we bail, because
+        // hand controllers might be sputtering "valid" data and that will keep someone from deliberately
+        // using the mouse on another app. (Fogbugz case 546.)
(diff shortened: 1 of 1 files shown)


howard-stearns:
test plan:
1. using vive in hmd, make sure you can use the hand lasers, including interactions with 2d buttons, menus, and overlays.  Try pausing and coming back, and make sure everything is still good.
2. Peeking under the HMD but leaving it on, bring another window on top of Interface. (It doesn't need to completely cover it, just so that Interface is not the active/focused window.) Use the hand controller trigger if necessary to get things rolling in Interface (e.g., if you accidentally paused), but make sure the other window is still on top.  Make sure you can still use the hand controller laser. (The mouse will probably NOT be functional until you click in the Interface window. But that's out of scope, as we're just trying to make sure the hand controller works.)
3. In the same session (and thus with the hand controllers still "hot"), switch to desktop mode. 
   - If interface is on top, the mouse should work as in master.
   - If interface is not on top, the mouse should work normally for other applications, without being "pinned" by the live hand controller.

hifi-gustavo:
Build console output for commit: 0099a4e549e985b29844eadcf993c521eeb11516 available at: https://gist.github.com/adc3589221c4e1222dfd85deb86aba71<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8792-0099a4e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8792-0099a4e.dmg

ZappoMan:
testing now

ZappoMan:
Tested - works as advertised.

PR #8793, opened 2016-10-12
prevent avatar from recentering below ground

AndrewMeadows:
* fix bug where avatar would tunnel under floor when recentering on HMD position

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -247,6 +247,15 @@ void MyAvatar::centerBody() {
     auto worldBodyPos = extractTranslation(worldBodyMatrix);
     auto worldBodyRot = glm::normalize(glm::quat_cast(worldBodyMatrix));
 
+    if (_characterController.getState() == CharacterController::State::Ground) {
+        // the avatar's physical aspect thinks it is standing on something
+        // therefore need to be careful to not "center" the body below the floor
+        float downStep = glm::dot(worldBodyPos - getPosition(), _worldUpDirection);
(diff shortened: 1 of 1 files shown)


AndrewMeadows:
Note: this is a cherry-pick of #8787 which was merged against the out-of-body-experience branch -- we'd like to fix it in master so that it does not depend on the other work.  It fixes a bug where the avatar's physical body could be pushed into the floor when in HMD mode.  The internal bug link is:

https://highfidelity.fogbugz.com/f/cases/1350

The repro recipe for the bug:

(1) Login with Vive while standing
(2) Navigate to Welcome domain
(3) Sit on floor
(4) Press ESC twice. The first press goes into PAUSED mode, the second press unpauses and recenters the avatar so that its eyes line up with the known HMD position (which happens to be inside or below the floor).
--> on bugged versions the avatar's physical body will end up in deep penetration with the floor and will be resolved out, often out the bottom.
--> on fixed version the avatar's physical body will NOT be dropped... it the avatar was standing on the ground when the recenter happens. If the avatar is hovering then the recenter will properly move the avatar's physical body down accordingly.

hifi-gustavo:
Build console output for commit: 867fc3ccbf5946f7c96e1941c0339a7cce252e31 available at: https://gist.github.com/6fdbd57bcdfaa652731cacbe1b1a4da5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8793-867fc3c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8793-867fc3c.dmg

ZappoMan:
Testing now.

ZappoMan:
Tested - worked as advertised.

PR #8807, opened 2016-10-13
fix entity-import with only tmp-rez rights

sethalves:
- it's now possible to import (with menu or market) entities in places where you have only tmp-rez rights

### DIFF
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -945,7 +945,9 @@ int EntityTree::processEditPacketData(ReceivedMessage& message, const unsigned c
                     properties.setLifetime(_maxTmpEntityLifetime);
                     // also bump up the lastEdited time of the properties so that the interface that created this edit
                     // will accept our adjustment to lifetime back into its own entity-tree.
-                    properties.setLastEdited(properties.getLastEdited() + LAST_EDITED_SERVERSIDE_BUMP);
+                    if (properties.getLastEdited() != UNKNOWN_CREATED_TIME) {
+                        properties.setLastEdited(properties.getLastEdited() + LAST_EDITED_SERVERSIDE_BUMP);
+                    }
                 }
             }
 
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -945,9 +945,10 @@ int EntityTree::processEditPacketData(ReceivedMessage& message, const unsigned c
                     properties.setLifetime(_maxTmpEntityLifetime);
                     // also bump up the lastEdited time of the properties so that the interface that created this edit
                     // will accept our adjustment to lifetime back into its own entity-tree.
-                    if (properties.getLastEdited() != UNKNOWN_CREATED_TIME) {
-                        properties.setLastEdited(properties.getLastEdited() + LAST_EDITED_SERVERSIDE_BUMP);
+                    if (properties.getLastEdited() == UNKNOWN_CREATED_TIME) {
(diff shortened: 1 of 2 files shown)


sethalves:
https://highfidelity.fogbugz.com/f/cases/1924/If-you-only-have-temp-rez-rights-on-a-server-the-lifetime-is-not-always-updated-correctly

sethalves:
to test:
- go to some place where you have only tmp-rez rights
- open the marketplace and rez something

RyanDowne:
Did huffman just check something like this in too?

sethalves:
these problems are cousins, but not the same.  both PRs are needed

hifi-gustavo:
Build console output for commit: 01c1cb328f7642fa7ed7703c87149e5aadd70c9f available at: https://gist.github.com/e86b24e2acbaf3501d56a26995d6624c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8807-01c1cb3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8807-01c1cb3.dmg

libraries/entities/src/EntityTree.cpp
```diff
@@ -945,7 +945,9 @@ int EntityTree::processEditPacketData(ReceivedMessage& message, const unsigned c
                     properties.setLifetime(_maxTmpEntityLifetime);
                     // also bump up the lastEdited time of the properties so that the interface that created this edit
                     // will accept our adjustment to lifetime back into its own entity-tree.
-                    properties.setLastEdited(properties.getLastEdited() + LAST_EDITED_SERVERSIDE_BUMP);
+                    if (properties.getLastEdited() != UNKNOWN_CREATED_TIME) {
+                        properties.setLastEdited(properties.getLastEdited() + LAST_EDITED_SERVERSIDE_BUMP);
```
ZappoMan: you know -- I'm not sure this is correct.

what cases are there for the properties having a bad "last edited" time? If that's what happens on create,  you'll still want to make sure that the client pays attention to the change here.

You might want an else clause here with an edit time of usecTimestampNow()

sethalves: imported entities have a last edited time of UNKNOWN_CREATED_TIME (which is zero).  added 1 to this doesn't make sense.  without this if, later code thinks the entity was created around 1970 and its lifetime is well over.

sethalves: you imagine something like
                    // also bump up the lastEdited time of the properties so that the interface that created this edit
                    // will accept our adjustment to lifetime back into its own entity-tree.
                    if (properties.getLastEdited() == UNKNOWN_CREATED_TIME) {
                        properties.setLastEdited(usecTimestampNow());
                    }
                    properties.setLastEdited(properties.getLastEdited() + LAST_EDITED_SERVERSIDE_BUMP);

ZappoMan: sure

hifi-gustavo:
Build console output for commit: fe09a1a57cd379acd88db6e5fb127e780afefb1c available at: https://gist.github.com/e86b24e2acbaf3501d56a26995d6624c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8807-fe09a1a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8807-fe09a1a.dmg

ZappoMan:
tested - passed

PR #8813, opened 2016-10-14
Avatar mixer data races

sethalves:


assignment-client/src/avatars/AvatarMixer.h
```diff
@@ -52,10 +52,12 @@ private slots:
 
     float _trailingSleepRatio { 1.0f };
     float _performanceThrottlingRatio { 0.0f };
+    mutable ReadWriteLockable _sleepRatioLock { };
 
     int _sumListeners { 0 };
     int _numStatFrames { 0 };
     int _sumIdentityPackets { 0 };
+    mutable ReadWriteLockable _statsLock { };
```
birarda: Could we use atomics for these instead of the ReadWriteLockable? Those would be less costly, right?

sethalves: yeah I think we could, but that wouldn't protect the integrity of one "bundle" of stats.  Maybe we don't care?

birarda: I don't think we do

hifi-gustavo:
Build console output for commit: 50d595794d9f43afbb7b52c2e8fb2f46f2918be1 available at: https://gist.github.com/99917a83cf8dcbdd1a62f3bca4131822<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8813-50d5957.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8813-50d5957.dmg

sethalves:
I'm going to split these up into a PR per data-race.

PR #8816, opened 2016-10-14
CellScience play button doesn't play anything

davidkelly:
Turns out an entity script is calling Audio.playSound(...), which is localOnly and stereo, but the position is far away from where you are.  Used to be that was fine -- localOnly just played into your ears.   Now we adjust the volume based on distance, and if mono we HRTF it too.  So, that sound just is too faint to hear.

Simple solution is use your avatar's position.  Also bumped the volume a bit since it was really low.

I've now uploaded this to the _correct_ spot in S3.

*Note that the old version of the script will be cached unless you wipe your cache - reload content should do the trick*

### DIFF
--- a/unpublishedScripts/DomainContent/CellScience/Scripts/showButtonToPlaySound.js
+++ b/unpublishedScripts/DomainContent/CellScience/Scripts/showButtonToPlaySound.js
@@ -6,7 +6,7 @@
 //
 
 (function() {
-    var baseURL = "https://hifi-production.s3.amazonaws.com/hifi-production/DomainContent/CellScience/";
+    var baseURL = "https://hifi-production.s3.amazonaws.com/DomainContent/CellScience/";
     var self = this;
     this.buttonImageURL = baseURL + "GUI/play_audio.svg?2";
 
--- a/unpublishedScripts/DomainContent/CellScience/Scripts/showIdentification.js
+++ b/unpublishedScripts/DomainContent/CellScience/Scripts/showIdentification.js
@@ -47,8 +47,7 @@
                 stereo: true,
                 loop: false,
                 localOnly: true,
-                volume: 0.035,
-                position: properties.position
+                volume: 0.45,
             };
             self.sound = SoundCache.getSound(self.soundURL);
             self.buttonImageURL = baseURL + "GUI/GUI_audio.png?" + version;
@@ -142,6 +141,7 @@
                 Overlays.editOverlay(self.button, {
                     visible: false
                 });
+                self.soundOptions.position = MyAvatar.position;
                 this.soundPlaying = Audio.playSound(self.sound, self.soundOptions);
             } else {
                 // print("not downloaded");


hifi-gustavo:
Build console output for commit: 18849d54166d83094780052a73bb29bcba34ff29 available at: https://gist.github.com/d56254c5264441c2902401880eadf47e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8816-18849d5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8816-18849d5.dmg

kevinmthomasse:
testing

kevinmthomasse:
## QA Test
**QA Tester**: KevinMThomas
**QA Pass**: **QA Pass** FAILURE
**Additional Test Results + Observations**: I read earlier comment from @davidkelly and I see that @ZappoMan has now added it Development Approved.  Tested in ```cellscience```.  There is no visible play button in addition when I click on the different UI's it does take me to each of the locations like ```The Ribosome```, etc however the additional issue is the audio is badly jittered.  It breaks up and jitters terribly.

RyanDowne:
I tried this in both cellscience and dev-cell and clicking the play button did not trigger the audio clip

ZappoMan:
@davidkelly -- did you update the content in cell science? opening a PR would not be sufficient to get cell science updated.

davidkelly:
I should have marked DNM until I'd updated and tested the content.  I thought the italicized part in the description above would be enough.  Cell cached the _old_ stuff, so it will not pick up the new things until I edit the models.json, etc...  Which I will do after verifying locally.  Marking DNM for now so I don't waste any qa cycles.

davidkelly:
I verified the script (`showIdentification.js`) works when using it in my sandbox in the same way the entities in cell science use it.  And I verified the bug was what I thought, by having a very large entity and being far from its origin.  However... there is a bug in master that is pegging the cpus on my `dev-david` version of cell science (and therefore in dev-cell`) so I cannot verify this works.  Also cannot try it in cell-science until monday as I don't have access to that machine to update the `models.json`

ZappoMan:
Did you deploy to cellscience or Dev-cellscience? The current non-Dev domain is not able to run Agent scripts because of a bug fixed by Howard last week that is not yet deployed.

davidkelly:
The script change is in S3.  I need to edit `models.json` on those machines, so they can pick that up, but I don't currently have ssh access to them.  Tomorrow morning I'll get marko to give me ssh to them, or coordinate editing the `models.json`.  Works locally, so _should_ be fine.

davidkelly:
# Test Plan
(1) Be sure cell science content has not been cached.  One way is to blow away _everything_ hifi related first.
(2) Go to a place that should have the play button (like `3272.76, 13703.3, 4378.75` for instance).  The play button should come up.
(3) hit play, learn you some biology.

PR #8819, opened 2016-10-15
never interact with hud elements while holding something

howard-stearns:
If you have near grabbed or far grabbed something, don't interact with hud elements.
Fixes at least the frist case of https://highfidelity.fogbugz.com/f/cases/1849/Double-lasers-when-over-HUD, and possibly the second.

Test plan:
- Far grab something and wave it behind or "in front" of a hud element. The hand-to-hud laser should not show up while you're holding it, and hud buttons should not highlight or interact.
- Repeat for near grabbing something.
Note that the hud elements are always drawn "on top", even when an object (grabbed or otherwise) is "closer". That does not change with this PR.

Also makes scripts use relative references, so you don't waste time debugging an included script that hasn't changed though you think you've edited it.

### DIFF
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -15,9 +15,9 @@
 
 (function() { // BEGIN LOCAL_SCOPE
 
-Script.include("/~/system/libraries/utils.js");
-Script.include("/~/system/libraries/Xform.js");
-Script.include("/~/system/libraries/controllers.js");
+Script.include("../libraries/utils.js");
+Script.include("../libraries/Xform.js");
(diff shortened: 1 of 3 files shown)


hifi-gustavo:
Build console output for commit: 140caf84d8140ab8369175c45e74107e5bde3422 available at: https://gist.github.com/39a07b7bb8f1c2f736df97af931ead86<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8819-140caf8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8819-140caf8.dmg

ZappoMan:
Tested - works as expected

PR #8821, opened 2016-10-16
fix cellscience ac script load

howard-stearns:
Get Cells and vesicles just once, 20 seconds after we start querying (and shut down queries).

Fixes https://highfidelity.fogbugz.com/f/cases/1960/cell-science-AC-scripts-cause-entity-server-to-spike-CPU

### DIFF
--- a/unpublishedScripts/DomainContent/CellScience/moveCellsAC.js
+++ b/unpublishedScripts/DomainContent/CellScience/moveCellsAC.js
@@ -39,11 +39,13 @@ var THROTTLE = true;
 var THROTTLE_RATE = 5000;
 
 var sinceLastUpdate = 0;
+var entitiesToMove = [];
 
 //print('cells script')
 
 function findCells() {
     var results = Entities.findEntities(basePosition, 60000);
+    Script.clearInterval(octreeQueryInterval); // we don't need it any more
 
     if (results.length === 0) {
         //    print('no entities found')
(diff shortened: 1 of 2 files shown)


howard-stearns:
@RyanDowne,  I have put links to this PR's two raw ac scripts on dev-cell, so you can look there to see if it is right.

hifi-gustavo:
Build console output for commit: e5c6fcaa70ef1754e5fee5b3927e21a14f553d90 available at: https://gist.github.com/d82e1b7588e23f9c83572302facff229<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8821-e5c6fca.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8821-e5c6fca.dmg

RyanDowne:
The translation is not smooth in dev-cellscience but they are moving now. I am also not sure if you worked only on the motorprotein script or the vesicle one too. I see a bunch of them in the edit list but the are really far away as if they got created too far away from the Inside the Cell scene. Anyway, it's an improvement but maybe something that's easier to look at together in the office.

ZappoMan:
My recommendation is that if @marko8904 says this drops the CPU load, then we merge it -- this is definitely a better design than what was in there before.

howard-stearns:
The changes are only to moveCellsAC and moveVesiclesAC.  (The cells script does only rotational velocity, and it's pretty darn subtle. The vesicles does the same subtle rotation, and some rather extreme linear velocity.)

I have not made any changes to the motorProteinControllerAC.

RyanDowne:
If you look at this pic {http://blog.highfidelity.com/blog/2015/07/high-fidelitys-stem-vr-challenge} you will see the large white vesicles that are around Inside the Cell. They are the ones that have Brownian motion. I don't see those at all in dev-cell.

PR #8844, opened 2016-10-18
Update default system toolbar position

huffman:


interface/resources/qml/hifi/Desktop.qml
```diff
@@ -53,9 +53,10 @@ OriginalDesktop.Desktop {
     Toolbar {
         id: sysToolbar;
         objectName: "com.highfidelity.interface.toolbar.system";
-        // Magic: sysToolbar.x and y come from settings, and are bound before the properties specified here are applied.
-        x: sysToolbar.x;
-        y: sysToolbar.y;
+        // These values will be overridden by sysToolbar.x/y if there is a saved position in Settings
+        // On exit, the sysToolbar position is saved to settings
+        x: 30
+        y: 50
```
huffman: The previous comment seemed incorrect in practice. If you know of a better way to do this, please let me know.

huffman:
To test:

 1. With your settings cleared, confirm that the HUD toolbar (the one that contains edit, snapshot, etc.) appears below the VU meter instead of on top of it.
 1. Confirm that moving the bar to a new position and restarting interface retains the new position.

interface/resources/qml/hifi/Desktop.qml
howard-stearns: Sounds like we're going to be constantly fighting timing on this. 
Alas, I don't know what The Right Answer(TM) is.

hifi-gustavo:
Build console output for commit: 3ad039cadc15e5a0a55f5581fda0663917cbc372 available at: https://gist.github.com/fcc24681e469797d6a8a58e1712597f9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8844-3ad039c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8844-3ad039c.dmg

howard-stearns:
For me, this does what it says in desktop. In HMD, it suffers from https://highfidelity.fogbugz.com/f/cases/1992/wacky-hud-element-placement-when-switching-between-desktop-and-hmd-loses-overlays, which cause a really unfortunate starting position in HMD.  (E.g., Vive start-in-VR users will never see their toolbar.)

I don't yet have an opinion as to whether we should move it left. I need to understand 1992 more....

jherico:
I'm still of the opinion that our window placement logic should be using the center of the screen as the origin, rather than the top left.

thoys:
@jherico what about anchor points?

huffman:
@jherico I agree, I think that's a better direction. This is only meant to fix an issue with the current approach.

PR #8845, opened 2016-10-18
openvr controller rendering + grab sphere adjustments

hyperlogic:
* hand controllers become visible when you enter away mode
* You can now display the hand controllers via the `MyAvatar.requestShowHandControllers()` and hide them with `MyAvatar.requestHideHandControllers()`
* You can now display the hand controllers via the "Developers > Hands > Show Hand Controllers" Menu
* grab sphere is always visible
* grab sphere is more aligned with the vive controller
* the vive controller is more aligned with the avatar's hand.


hyperlogic:
https://highfidelity.fogbugz.com/f/cases/1950/more-work-on-grab-indicators

hifi-gustavo:
Build console output for commit: d97a8c2e7eace2494d3142e8187022a33aedc060 available at: https://gist.github.com/974133c9cf554df96475374cb91011c2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8845-d97a8c2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8845-d97a8c2.dmg

hifi-gustavo:
Build console output for commit: d3db63de1b2f345e4ea86f0e801544a887480bad available at: https://gist.github.com/974133c9cf554df96475374cb91011c2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8845-d3db63d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8845-d3db63d.dmg

interface/src/avatar/MyAvatar.cpp
```diff
@@ -2305,3 +2305,15 @@ void MyAvatar::updateHoldActions(const AnimPose& prePhysicsPose, const AnimPose&
         });
     }
 }
+
+void MyAvatar::requestHideHandControllers() {
```
ZappoMan: hmmm - I'm not sure I'm a fan of this approach... what's the theory here?

interface/src/avatar/MyAvatar.h
```diff
@@ -282,6 +282,10 @@ class MyAvatar : public Avatar {
     void removeHoldAction(AvatarActionHold* holdAction);  // thread-safe
     void updateHoldActions(const AnimPose& prePhysicsPose, const AnimPose& postUpdatePose);
 
+    Q_INVOKABLE void requestHideHandControllers();
```
ZappoMan: why are these methods on MyAvatar? Why aren't these an HMD or Controllers interface? This seems like the wrong place to have these.

scripts/system/away.js
```diff
@@ -174,6 +174,7 @@ function goAway(fromStartup) {
     if (!wasMuted) {
         AudioDevice.toggleMute();
     }
+    MyAvatar.requestShowHandControllers();
```
ZappoMan: I think this should be on the HMD...

interface/src/avatar/MyAvatar.cpp
hyperlogic: HMD didn't fit, because this it relates to input plugins.  I could put it on Controller script interface, I just elected to add it to MyAvatar, as it already had several methods w.r.t. hand controller positions and the controller position is tied to the avatar anyhow.  Would you prefer me to add this to the Controller script interface?

interface/src/avatar/MyAvatar.h
hyperlogic: HMD didn't fit, because this it relates to input plugins.  I could put it on Controller script interface, I just elected to add it to MyAvatar, as it already had several methods w.r.t. hand controller positions and the controller position is tied to the avatar anyhow.  Would you prefer me to add this to the Controller script interface?

interface/src/avatar/MyAvatar.cpp
ZappoMan: We already have HMD.hasHandControllers... - so the idea that some hands come with hand controllers seems like a reasonable assumption at this point.

ZappoMan:
You should break this into to PRs -- one that makes the positioning and sizing changes to the grab spheres... and one that does the more controversial stuff.

plugins/openvr/src/ViveControllerManager.cpp
```diff
@@ -128,8 +61,15 @@ void ViveControllerManager::deactivate() {
...
-    _container->removeMenu(MENU_PATH);
+    render::PendingChanges pendingChanges;
+    for (auto& info : _payloadInfos) {
+        // remove from scene if necessary
+        if (info.addedToScene) {
+            pendingChanges.removeItem(info.itemID);
+            info.addedToScene = false;
+        }
+    }
+    _container->getMainScene()->enqueuePendingChanges(pendingChanges);
```
ZappoMan: Unless @jherico and @samcake convince me otherwise... this approach is a show stopper for me. We don't really want to expose the "scene" to the display plugins... they shouldn't know anything about the scene...

plugins/openvr/src/ViveControllerManager.h
```diff
@@ -17,20 +17,18 @@
...
-#include <RenderArgs.h>
-#include <render/Scene.h>
+#include <MeshPartPayload.h>
 
 namespace vr {
     class IVRSystem;
 }
 
-class ViveControllerManager : public InputPlugin {
+class ViveControllerManager : public InputPlugin, public std::enable_shared_from_this<ViveControllerManager> {
```
ZappoMan: That would be exactly why I'm opposed to messing around in the scene from the display plugin... basically this is all backwards.

hifi-gustavo:
Build console output for commit: ec2bc3495082dfb0b472948efa9d6881562ed3f9 available at: https://gist.github.com/974133c9cf554df96475374cb91011c2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8845-ec2bc34.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8845-ec2bc34.dmg

ZappoMan:
My recommendation:
All of this behavior can/should be done in scripts without any modifications to the C++ (with the exception of the weird grab point/laser line issue). You should just be using the model from the tutorial, and display that during away mode.

plugins/openvr/src/ViveControllerManager.h
hyperlogic: No, IMHO the current tutorial script approach is completely backwards.  This gets the correct mesh directly from openvr via their SDK, it is future proof.  It even supports users changing their hand controllers via the Steam Workshop.  I understand your concern about accessing the scene from input plugin. (even though this was how the original rendering was done).  I like the idea of exposing the mesh to the application.  And the application can then add a renderItem to the scene.  Would you like me to make that change?

hyperlogic:
I will split this into two PRs, even though they are both associated with this single fogbugs task.

https://highfidelity.fogbugz.com/f/cases/1950/more-work-on-grab-indicators

hyperlogic:
I'm going to leave this PR open so it can be tested by @RyanDowne and @PhilipRosedale

plugins/openvr/src/ViveControllerManager.h
huffman: I'm a fan of getting the mesh via the SDK, but there are a couple issues:

- We annotate the controller with text and images to point out specific buttons, so that work needs to stay as far as I can tell.
- Custom Steam Workshop controllers means the annotations and general tutorial experience may not work as expected. Maybe you can request the default controller?

huffman: I forgot to mention:
- We would also need a way to highlight the trigger

hifi-gustavo:
Build console output for commit: f329c29bac3212af76c5861ac8f7234152ea8ed6 available at: https://gist.github.com/974133c9cf554df96475374cb91011c2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8845-f329c29.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8845-f329c29.dmg

hyperlogic:
going to split this into separate pull requests.

PR #8851, opened 2016-10-19
Rate limit QML web surfaces

jherico:
Numerous QML web entities and overlays (such as the marketplace tablet and in-world web entities for locations in ova-welcome) may be contributing to the rendering slowdowns people are seeing.  This attempts to mitigate the issue by rate-limiting the surfaces to 10 frames per second.  This will be detrimental to video content rendered through a web entity, but c'est la vie.

## Testing

Application behavior should be unchanged.  Web pages may be less smooth in scrolling.  Video content on web pages will likely be 'jerky'

### DIFF
--- a/interface/src/ui/overlays/Web3DOverlay.cpp
+++ b/interface/src/ui/overlays/Web3DOverlay.cpp
@@ -79,6 +79,9 @@ void Web3DOverlay::render(RenderArgs* args) {
             });
         };
         _webSurface = QSharedPointer<OffscreenQmlSurface>(new OffscreenQmlSurface(), deleter);
+        // FIXME, the max FPS could be better managed by being dynamic (based on the number of current surfaces
+        // and the current rendering load)
+        _webSurface->setMaxFps(10);
         _webSurface->create(currentContext);
         _webSurface->setBaseUrl(QUrl::fromLocalFile(PathUtils::resourcesPath() + "/qml/controls/"));
         _webSurface->load("WebView.qml");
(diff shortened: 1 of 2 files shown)


kevinmthomasse:
## QA Test
**QA Tester**: KevinMThomas
**Additional Test Results + Observations**:  Application behavior is unchanged as expected.  Web page scrolling was not bad however image loading was a bit slower.  Solid framerate of 60 and tested on Oculus as the Marketplace tablet worked properly as expected.  To reiterate the Marketplace tablet scrolling I did not see any issue and the images loaded slightly slower.  Oculus framerate was 80 FPS.  Tested in ```ova-welcome```.

ZappoMan:
I'm not sure we want to do this

kevinmthomasse:
@ZappoMan standing by if you would like some specific additional testing on this.  Thanks!

ZappoMan:
Some suggestions:
- Only enable this if we detect some N number of web surfaces.
- Only enable this if the application simulation rate is ~75% of desired present rate??
- What if we set the value to 30fps? Does that help?

Also... can we construct an example where this change makes a positive difference? Or is this purely speculative? At what point (how many web pages) does it make a difference? What I noticed when testing yesterday is that even with ~10 web views visible, my frame rate was still 90fps.

howard-stearns:
Some other questions:
 - All Web surfaces are just some? (There's some parallel copied code, but maybe a subtle difference.)
 - Does it matter if they are created new each time, or just made visible/invisible? (Maybe something not going away like it should?)

ZappoMan:
As part of steam launch -- we don't have any launch content that shows video... so @RyanDowne has approved this change. Since it may be the source of some of the frame rate issues we've seen, we're going to merge this for now and retest/revisit.

PR #8856, opened 2016-10-19
Remove the use of implicit ids in geometry cache

jherico:
The use of implicit endpoints as keys can lead to leakage of GPU resources.  This changes the GeometryCache API to _require_ the use of a geometry ID for every simple draw, ensuring that a given item will re-use the same GPU buffers over and over, rather than potentially creating new ones.

## Testing

When using the hand lasers to hover over UI elements, the current production build will show a constantly increasing number of GPU buffers on the stats display (enabled with `/`).  With this build the stats should not keep climbing as you move the lasers around over the UI.

Application behavior should be otherwise unchanged.

### DIFF
--- a/interface/src/Util.cpp
+++ b/interface/src/Util.cpp
@@ -50,50 +50,69 @@ void renderWorldBox(gpu::Batch& batch) {
     static const float DASH_LENGTH = 1.0f;
     static const float GAP_LENGTH = 1.0f;
     auto transform = Transform{};
+    static std::array<int, 18> geometryIds;
+    static std::once_flag initGeometryIds;
+    std::call_once(initGeometryIds, [&] {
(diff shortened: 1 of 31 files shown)


hifi-gustavo:
Build console output for commit: 2705ee7030c96a0824db68283988ebb5dea57950 available at: https://gist.github.com/ab682746453ff29ce78e916d12d642be<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8856-2705ee7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8856-2705ee7.dmg

PhilipRosedale:
QA PASS:  Turned all UI elements on and off while inspecting buffers for growth

PR #8857, opened 2016-10-19
fix crowd animation again

howard-stearns:
fixes https://highfidelity.fogbugz.com/f/cases/1995/no-agent-animation-on-linux

On Linux, there are no /~/... resources. That doesn't matter if you just want to show the avatar, because the default url gets translated to each client and the client fetches and draws it. But to do the animation, the ac needs the skeleton.  So... we now have the interface script specify a valid .fst (which points to a valid .fbx).


hifi-gustavo:
Build console output for commit: 69a141587307c67fd2f5b6d2beb05b4dc56fc86b available at: https://gist.github.com/5fd9436f79f42375fc6ace1a3c0e5230<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8857-69a1415.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8857-69a1415.dmg

PR #8858, opened 2016-10-19
Update users.js to be minimizable when logged out, and by default

huffman:


huffman:
Test:

 1. Users window is minimized by default
 1. You can minimize and maximize the users window when logged out
 1. Everything looks correct on the users window when:
   1. Logged out
   1. Logged in
 1. The users window still works as expected (dropdowns work, clicking usernames works)

hifi-gustavo:
Build console output for commit: f42683cf8b3e2696f8d65f3d8565db16b91e4223 available at: https://gist.github.com/3a0051afa9d83dc656225cda1c5e9b2b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8858-f42683c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8858-f42683c.dmg

scripts/system/users.js
```diff
@@ -381,7 +381,7 @@ var usersWindow = (function () {
 
         isLoggedIn = false,
         isVisible = true,
-        isMinimized = true,
```
Atlante45: Why not keep the default to true?

RyanDowne:
This passed. We need to separately decide where the right location for it is when in HMD. It's current position is correct for desktop.

PR #8865, opened 2016-10-20
fix resize on plugin switch

howard-stearns:
fixes https://highfidelity.fogbugz.com/f/cases/1992/wacky-hud-element-placement-when-switching-between-desktop-and-hmd-loses-overlays

test plan:
- Start with an empty Interface.ini in desktop. Observe the location of overlays. Maybe open a few more, scattered around.
- Switch to HMD. All overlays should be within the "recommended area" which extends a bit to the sides, but NOT way behind you.
- Switch back to Desktop.  Overlays should be same as original.
- Repeat, but this time in HMD, move some of the overlays further around to both sides (and up and down) so that they start to get clipped (almost behind you).  Now when you go back to desktop, the ones that were clipped outside the recommended area (from the first HMD step) will be clipped, but still visible.
- Repeat with manual resizing of desktop Interface window.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -5681,6 +5681,9 @@ void Application::updateDisplayMode() {
     // Make the switch atomic from the perspective of other threads
     {
         std::unique_lock<std::mutex> lock(_displayPluginLock);
+        // Tell the desktop to no reposition (which requires plugin info), until we have set the new plugin, below.
+        bool wasRepositionLocked = offscreenUi->getDesktop()->property("repositionLocked").toBool();
+        offscreenUi->getDesktop()->setProperty("repositionLocked", true);
 
         auto oldDisplayPlugin = _displayPlugin;
         if (_displayPlugin) {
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 6f283a4893239a1df84ec30b41590557161a99f8 available at: https://gist.github.com/6f8c0af948b4e0acf8d7e61a0913c36e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8865-6f283a4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8865-6f283a4.dmg

RyanDowne:
This worked ok with the basic HUD but the edit.js HUD did not respect the place I moved it. When I opened it again it was off to the side not where I had placed it.

howard-stearns:
@RyanDowne the same device-specific memory behavior is present in the current beta release 5429, so it has nothing to do with this PR.

I can't say whether the edit toolbar behavior is a bug or not. My own opinion is that device-specific memory is wrong because it leads to exactly this sort of confusion, but maybe someone had some reason for making it work this way. I couldn't find where this behavior comes from, in a quick scan of the code.

Anyway, please update the pass fail with respect to 1992 as you feel appropriate.

PR #8868, opened 2016-10-20
Remove showing of overlays in away.js

huffman:
This change makes changes to away.js so that overlays are not automatically displayed when going into away mode.

To test:

 1. Confirm that you can see the away message in both HMD and Desktop while away
 1. Confirm that if you have overlays hidden (`View > Overlays` unchecked) while you are not in away mode that they are not shown when you go into away mode, but that **you can still see the away message**.
 1. Confirm that away.js works as expected

### DIFF
--- a/scripts/system/away.js
+++ b/scripts/system/away.js
@@ -54,7 +54,6 @@ var AWAY_INTRO = {
 var isEnabled = true;
 var wasMuted; // unknonwn?
 var isAway = false; // we start in the un-away state
-var wasOverlaysVisible = Menu.isOptionChecked("Overlays");
 var eventMappingName = "io.highfidelity.away"; // goActive on hand controller button events, too.
 var eventMapping = Controller.newMapping(eventMappingName);
 var avatarPosition = MyAvatar.position;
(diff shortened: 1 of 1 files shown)


ZappoMan:
hmmm - originally - we felt that it was important to show the away message to people even if they had overlays hidden - this sounds like a change.

hifi-gustavo:
Build console output for commit: fad0607bf6c66219f80de893c13fde5497fdd764 available at: https://gist.github.com/15c77f6a703b156a3b7a5b2379160059<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8868-fad0607.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8868-fad0607.dmg

RyanDowne:
Confirm that you can see the away message in both HMD and Desktop while away **PASS**
Confirm that if you have overlays hidden (View > Overlays unchecked) while you are not in away mode that they are not shown when you go into away mode **PASS**
Confirm that away.js works as expected PASS

One thing to note here, and this is not new in this PR... if this is meant to help with the tutorial, the single HUD button that flips the Overlays is always visible. We will need to hide this too.

ZappoMan:
@RyanDowne - I don't think we want this behavior "Confirm that if you have overlays hidden (View > Overlays unchecked) while you are not in away mode that they are not shown when you go into away mode" - that goes against the original spec for away mode

RyanDowne:
@ZappoMan I think you are confusing the Away overlay with the other ones. When you have overlays off and are in away mode, you still see the away indicator.

huffman:
Yes, @RyanDowne is correct, the overlays being off should not affect the away message in Desktop or HMD modes.

cozza13:
To test:

 1. Confirm that you can see the away message in both HMD and Desktop while away **YES**
 1. Confirm that if you have overlays hidden (`View > Overlays` unchecked) while you are not in away mode that they are not shown when you go into away mode, but that **you can still see the away message**. **FAIL THE HUD ALSO COMES BACK UP**
 1. Confirm that away.js works as expected **PASS EXCEPT FOR ABOVE**

PR #8879, opened 2016-10-21
add Bullet profile info to PerformanceTimer stats

AndrewMeadows:
* Bullet's performance profile data is now exposed in interface debug stats

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -3968,8 +3968,6 @@ void Application::update(float deltaTime) {
                 auto collisionEvents = _physicsEngine->getCollisionEvents();
                 avatarManager->handleCollisionEvents(collisionEvents);
 
-                _physicsEngine->dumpStatsIfNecessary();
-
                 if (!_aboutToQuit) {
                     PerformanceTimer perfTimer("entities");
                     // Collision events (and their scripts) must not be handled when we're locked, above. (That would risk
@@ -3982,6 +3980,13 @@ void Application::update(float deltaTime) {
                 }
 
                 myAvatar->harvestResultsFromPhysicsSimulation(deltaTime);
+
+                if (Menu::getInstance()->isOptionChecked(MenuOption::DisplayDebugTimingDetails) &&
+                        Menu::getInstance()->isOptionChecked(MenuOption::ExpandPhysicsSimulationTiming)) {
+                    _physicsEngine->harvestPerformanceStats();
+                }
+                // NOTE: the PhysicsEngine stats are written to stdout NOT to Qt log framework
+                _physicsEngine->dumpStatsIfNecessary();
             }
         }
     }
--- a/interface/src/Menu.h
+++ b/interface/src/Menu.h
@@ -105,6 +105,7 @@ namespace MenuOption {
     const QString ExpandMyAvatarTiming = "Expand /myAvatar";
     const QString ExpandOtherAvatarTiming = "Expand /otherAvatar";
     const QString ExpandPaintGLTiming = "Expand /paintGL";
+    const QString ExpandPhysicsSimulationTiming = "Expand /physics";
     const QString ExpandUpdateTiming = "Expand /update";
     const QString Faceshift = "Faceshift";
     const QString FirstPerson = "First Person";
--- a/interface/src/ui/Stats.cpp
+++ b/interface/src/ui/Stats.cpp
@@ -55,7 +55,9 @@ Stats::Stats(QQuickItem* parent) :  QQuickItem(parent) {
 bool Stats::includeTimingRecord(const QString& name) {
     if (Menu::getInstance()->isOptionChecked(MenuOption::DisplayDebugTimingDetails)) {
         if (name.startsWith("/idle/update/")) {
-            if (name.startsWith("/idle/update/myAvatar/")) {
+            if (name.startsWith("/idle/update/physics/")) {
+                return Menu::getInstance()->isOptionChecked(MenuOption::ExpandPhysicsSimulationTiming);
+            } else if (name.startsWith("/idle/update/myAvatar/")) {
                 if (name.startsWith("/idle/update/myAvatar/simulate/")) {
                     return Menu::getInstance()->isOptionChecked(MenuOption::ExpandMyAvatarSimulateTiming);
                 }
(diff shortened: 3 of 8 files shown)


interface/src/Menu.h
```diff
@@ -105,6 +105,7 @@ namespace MenuOption {
     const QString ExpandMyAvatarTiming = "Expand /myAvatar";
     const QString ExpandOtherAvatarTiming = "Expand /otherAvatar";
     const QString ExpandPaintGLTiming = "Expand /paintGL";
+    const QString ExpandPhysicsSimulationTiming = "Expand /physics";
```
AndrewMeadows: Added menu item for toggling visibility of Bullet stats inside the PhysicsEngine::stepSimulation() context.

interface/src/ui/Stats.cpp
```diff
@@ -363,7 +365,7 @@ void Stats::updateStats(bool force) {
             QString functionName = j.value();
             const PerformanceTimerRecord& record = allRecords.value(functionName);
             perfLines += QString("%1: %2 [%3]\n").
-                arg(QString(qPrintable(functionName)), 90, noBreakingSpace).
+                arg(QString(qPrintable(functionName)), -80, noBreakingSpace).
```
AndrewMeadows: Made the stats align on the left, and reduced the width of the space available.

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -286,6 +288,47 @@ void PhysicsEngine::stepSimulation() {
...
+    }
+    // recurse the children
+    for (int32_t i = 0; i < numChildren; ++i) {
+        profileIterator->Enter_Child(i);
+        recursivelyHarvestPerformanceStats(profileIterator, contextName);
+    }
+    // retreat back to parent
+    profileIterator->Enter_Parent();
+}
+
```
AndrewMeadows: Bullet already has its own performance profiling framework and some of its internal contexts are instrumented with it.  Up to this point the profiling data was exposed on a per-frame basis: if you press the `F` key then Bullet will print the next frame's worth of profiling data to **stdout** (yes, stdout rather than the Qt logs).  Occasionally I would use this feature when troubleshooting the physics simulation but I don't think any other hifi devs knew about it.

Here I'm just taking the exact same data and stuffing it into our own `PerformanceTimer` system so that they can be printed out with all the other stats.  The F key remains hooked up -- it prints one snapshop in a different format.

hifi-gustavo:
Build console output for commit: e44672180ae2e51ec59531f347e7502227a1d434 available at: https://gist.github.com/ae09d530a60d64915dac71d2dd0ab643<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8879-e446721.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8879-e446721.dmg

AndrewMeadows:
To test:

(1) Login and make the debug stats visible by pressing the `/` key.  Alternatively this can be done via the menu:  `Developer --> Stats`.
(2) Make sure stats details are enabled in the menu:  `Developer --> Timing --> Performance Timer --> Display Timing Details`
(3) Also expand the /physics details:  `Developer --> Timing --> Performance Timer --> Expand /physics`
(4) Note that the Bullet details start with ".../" because otherwise their context names would be too long for the UI.
(5) Make a pile of dynamic objects and set them tumbling against each other -->  when looking at the debug stats the time spent in the physics simulation should increase when the pile of dynamic objects is active.
(6) To make the physics simulation more expensive try creating many dynamic objects inside a concave model that uses a static mesh shape.
(7) Let the dynamic objects settle down --> time spent in the physics simulation according to the stats should drop accordingly.
(8) Delete all of the dynamic objects --> time spent in the physics simulation should drop.

Note: if you see the time spent in physics NOT go down when the dynamic objects settle down (7) or are deleted (8) then please note how that condition was achieved and supply a repro recipe.

AndrewMeadows:
The internal fogbugz issue related to this work is:  https://highfidelity.fogbugz.com/f/cases/2000

interface/src/Application.cpp
```diff
@@ -3982,6 +3980,13 @@ void Application::update(float deltaTime) {
                 }
 
                 myAvatar->harvestResultsFromPhysicsSimulation(deltaTime);
+
+                if (Menu::getInstance()->isOptionChecked(MenuOption::DisplayDebugTimingDetails) &&
+                        Menu::getInstance()->isOptionChecked(MenuOption::ExpandPhysicsSimulationTiming)) {
+                    _physicsEngine->harvestPerformanceStats();
+                }
+                // NOTE: the PhysicsEngine stats are written to stdout NOT to Qt log framework
+                _physicsEngine->dumpStatsIfNecessary();
```
ZappoMan: should this also be wrapped in an if?

AndrewMeadows: Not necessary.  _physicsEngine is properly created in the Application ctor is never changed after that.  Hrm... it makes me wonder why we're using a pointer for _physicsEngine but I won't be changing that right now.

cozza13:
(1) Login and make the debug stats visible by pressing the `/` key.  Alternatively this can be done via the menu:  `Developer --> Stats`. **Done**
(2) Make sure stats details are enabled in the menu:  `Developer --> Timing --> Performance Timer --> Display Timing Details` **Done**
(3) Also expand the /physics details:  `Developer --> Timing --> Performance Timer --> Expand /physics` **Done**
(4) Note that the names of the physics simulation stats start with ".../" because otherwise their context names would be too long for the UI. **True!**
(5) Make a pile of dynamic objects and set them tumbling against each other -->  when looking at the debug stats the time spent in the physics simulation should increase when the pile of dynamic objects is active. **YES**
(6) To make the physics simulation more expensive try creating many dynamic objects inside a concave model that uses a static mesh shape. **YES**
(7) Let the dynamic objects settle down --> time spent in the physics simulation according to the stats should drop accordingly. **YES**
(8) Delete all of the dynamic objects --> time spent in the physics simulation should drop. **YES**

Note: if you see the time spent in physics NOT go down when the dynamic objects settle down (7) or are deleted (8) then please note how that condition was achieved and supply a repro recipe.

libraries/physics/src/PhysicsEngine.cpp
AndrewMeadows: This work is not expensive as far as I can tell.  The extraction of Bullet stats into our own is inside the old "/idle/update/physics" stat context.  By enabling "Expand /update" timing details the /idle/update/physics stat is made visible and then when toggling "Expand /physics" on and off I can't measure a difference in the time spent in that stat.

PR #8880, opened 2016-10-21
Show hand controllers in away.js

hyperlogic:
* only works for openvr vive controllers
* unifies display scripts for tutorial and away.js.


hyperlogic:
QA Test:
  * while wearing a vive hmd.  Enter away mode by pressing ESC
  * the vive controllers should be visible, making it easier to grab them.
  * the vive controllers should not be visible when NOT in away mode.

hifi-gustavo:
Build console output for commit: c431c69d60848ed2372a17f274c532c0e76f00ef available at: https://gist.github.com/fecda5b1dcbb763bb9ac7a329fb862d6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8880-c431c69.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8880-c431c69.dmg

tools/atp-extract.py
```diff
@@ -0,0 +1,44 @@
+#
```
ZappoMan: this should be a different PR

RyanDowne:
This worked as promised. Two things that I am not sure are a result of this PR or if they are specific to the machine I was using..
1) When in the tutorial, the controllers yellow triggers are no longer animated. That's regression
2) When equipping the lighter, the alignment is off after being fixed. That's regression.

hifi-gustavo:
Build console output for commit: ae301602e71e6d74e5173013504a233f22af3285 available at: https://gist.github.com/fecda5b1dcbb763bb9ac7a329fb862d6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8880-ae30160.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8880-ae30160.dmg

hyperlogic:
@ryanDowne before you evaluate how this works with the tutorial, @huffman will need to update the tutorial assets with this version of tutorial.js.

hyperlogic:
@RyanDowne I should clarify, this change will work fine with the exiting tutorial.  Except, that the assets will potentially loaded twice from separate places if you go into away mode while inside the tutorial.  This PR includes a change to tutorial content that will share the same assets (bundled with the installer) and not bundled with the tutorial content.

hifi-gustavo:
Build console output for commit: 84d7eef4ba701ab9befb729787966ddd59208354 available at: https://gist.github.com/fecda5b1dcbb763bb9ac7a329fb862d6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8880-84d7eef.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8880-84d7eef.dmg

interface/src/scripting/HMDScriptingInterface.cpp
```diff
@@ -57,6 +57,20 @@ bool HMDScriptingInterface::isHandControllerAvailable() {
     return PluginUtils::isHandControllerAvailable();
 }
 
+void HMDScriptingInterface::requestShowHandControllers() {
```
ZappoMan: ok - I guess we accept it for now, but I think you're point about needing some better cross script syncing would be a better way to do this... since basically the C++ doesn't really do anything other than send a signal.

ZappoMan: I wonder if a `Script.sendGlobalSignal()` and `Script.connectGlobalSignal()` would be better? thoughts?

PR #8884, opened 2016-10-22
Add more GPU stats, make stats easier to read

jherico:
* Fixes the GPU texture memory calculation, which was not including non-sparse textures.
* Adds new stats for
 * sparse texture count
 * sparse texture memory as a subset of total GPU texture memory
 * whether sparse textures are enabled (the above two stats only display if they are)
* Make the stats easier to read by darkening the background, decreasing the background alpha, and giving the text an outline color.  

## Testing 

Application behavior should be unchanged.  This PR only modifies the code renders the debug stats.

If you enable stats (`/`) you should see the new values in the rightmost column.  If sparse textures is enabled, you should see the sparse texture count and the sparse texture memory consumed.

If you go to a dev-playa and use the developer / render / texture memory menu to set the texture memory to a very small value (like 64 MB) you should see the sparse texture memory fall to under that value.  The total texture memory may remain above that value due to other content, like framebuffers and the like.

### DIFF
--- a/libraries/gpu-gl/src/gpu/gl/GLTexture.cpp
+++ b/libraries/gpu-gl/src/gpu/gl/GLTexture.cpp
@@ -205,6 +205,8 @@ GLTexture::~GLTexture() {
                 qWarning() << "No recycler available for texture " << _id << " possible leak";
             }
         } else if (_id) {
+            // WARNING!  Sparse textures do not use this code path.  See GL45BackendTexture for 
+            // the GL45Texture destructor for doing any required work tracking GPU stats
             backend->releaseTexture(_id, _size);
         }
     }
@@ -275,7 +277,7 @@ void GLTexture::postTransfer() {
     setSyncState(GLSyncState::Idle);
     ++_transferCount;
 
-    // At this point the mip pixels have been loaded, we can notify the gpu texture to abandon it's memory
+    // At this point the mip pixels have been loaded, we can notify the g   pu texture to abandon it's memory
     switch (_gpuObject.getType()) {
         case Texture::TEX_2D:
             for (uint16_t i = 0; i < Sampler::MAX_MIP_LEVEL; ++i) {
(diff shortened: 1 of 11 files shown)


libraries/gpu-gl/src/gpu/gl/GLTexture.cpp
```diff
@@ -139,7 +139,7 @@ float GLTexture::getMemoryPressure() {
     }
 
     // Return the consumed texture memory divided by the available texture memory.
-    auto consumedGpuMemory = Context::getTextureGPUMemoryUsage();
+    auto consumedGpuMemory = Context::getTextureGPUSparseMemoryUsage();
```
jherico: Note, this change actually keeps the current behavior, since this PR also fixes a bug where only the sparse textures were being tracked in `Context::getTextureGPUMemoryUsage();`

hifi-gustavo:
Build console output for commit: b572fb2a91a0565c550363921123ad404d5f9493 available at: https://gist.github.com/b5619b955b5ca119da876a05a7891afa<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8884-b572fb2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8884-b572fb2.dmg

hifi-gustavo:
Build console output for commit: 7879bf37f248077e686d3e59d13f519c5007ad98 available at: https://gist.github.com/b5619b955b5ca119da876a05a7891afa<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8884-7879bf3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8884-7879bf3.dmg

samcake:
I m good with the fix yeah!
i would suggest replacing the word "sparse" which is very specific to the underneath technical implementation 
instead i would suggest use a word like "managed"
because these textures are read only and then the Backend  "manage" their upload to video mem and even quality.

PR #8906, opened 2016-10-25
more detailed stats in user heartbeat

ZappoMan:


ZappoMan:
Testing:
- simple smoke test - this should have no apparent impact on application behavior
- I have already checked that this info is sent to the metabase

ZappoMan:
we want to get this into the RC -- it's not a fix per se... but it's a feature we'd like in so we can better track poor rendering performance.

interface/src/Application.cpp
```diff
@@ -1222,6 +1231,10 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer, bo
         properties["active_display_plugin"] = getActiveDisplayPlugin()->getName();
         properties["using_hmd"] = isHMDMode();
 
+        auto glInfo = getGLContextData();
+        properties["gl_info"] = glInfo;
```
huffman: 1. We send some of this as part of the launch event, but maybe it should include everything
1. Does this ever change? Maybe this should only be in the launch event

hifi-gustavo:
Build console output for commit: e22a2e9e2fc424cb5584774d8f087b7c7c79e3ef available at: https://gist.github.com/c7d5a4240235781e7929691a3f36c1a8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8906-e22a2e9.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8906-e22a2e9.dmg

interface/src/Application.cpp
ZappoMan: in other words, I'm intentionally breaking normalization since we know that all of these values are json and not likely index optimized

huffman: Ok, sounds good

interface/src/Application.cpp
```diff
@@ -858,7 +858,7 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer, bo
         { "gl_version_int", glVersionToInteger(glContextData.value("version").toString()) },
         { "gl_version", glContextData["version"] },
         { "gl_vender", glContextData["vendor"] },
-        { "gl_sl_version", glContextData["slVersion"] },
+        { "gl_sl_version", glContextData["sl_version"] },
```
jherico: if we're having to fix a bug caused by a magic constant, we should fix the magic constant

ZappoMan: this isn't a bug fix, just a naming convention change... the only user of this key was this line here...

hifi-gustavo:
Build console output for commit: 822e088240d4f3353180785d813690f08a0bdea2 available at: https://gist.github.com/c7d5a4240235781e7929691a3f36c1a8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8906-822e088.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8906-822e088.dmg

PR #8909, opened 2016-10-25
Crash on exit fix for render items

hyperlogic:
Destroy the render scene & engine before the Application is destroyed.

Many render items/payloads contain smart pointers back to the objects that added them to the scene, including entity and avatar objects. Currently, those render items are destroyed when the scene is destroyed very late in the application life-cycle.

There are rare crashes that can occur when these render items are destroyed. Possibly, due to them referencing objects that have already been destroyed via raw pointers. In an effort to eliminate these crashes, we now destroy the scene earlier, within Application::aboutToQuit() which is connected to the QCoreApplication::aboutToQuit signal.  Also, we guard against null scene pointer dereferences.  Any location that accesses the scene off the main thread, now checks the validity of the scene pointer.

### DIFF
--- a/libraries/render/src/render/Scene.cpp
+++ b/libraries/render/src/render/Scene.cpp
@@ -48,6 +48,10 @@ Scene::Scene(glm::vec3 origin, float size) :
     _items.push_back(Item()); // add the itemID #0 to nothing
 }
 
+Scene::~Scene() {
+    qDebug() << "Scene::~Scene()";
+}
+
 ItemID Scene::allocateID() {
     // Just increment and return the proevious value initialized at 0
     return _IDAllocator.fetch_add(1);
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1576,6 +1576,12 @@ void Application::cleanupBeforeQuit() {
 #endif
 
     DependencyManager::destroy<OffscreenUi>();
+
+    // shutdown render engine
+    _main3DScene = nullptr;
+    _renderEngine = nullptr;
+
+    qCDebug(interfaceapp) << "Application::cleanupBeforeQuit() complete";
 }
 
 Application::~Application() {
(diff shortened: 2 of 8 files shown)


hyperlogic:
https://highfidelity.fogbugz.com/f/cases/308/BugSplat-2899-Unknown-pure-virtual-function-call
https://highfidelity.fogbugz.com/f/cases/2070/crash-on-shutdown-in-render-Payload-Payload

hyperlogic:
QA test:
Verify that quitting the application does not cause a crash.  Try a variety of scenarios including:
  * Quit while connected to a site containing other avatars.
  * Quit while connected to a site with entities that are being created and destroyed.  Probably the easiest way to test this is to have a user shooting the ping pong gun.
  * Quit while connected to a site with PolyVox entities, hifi://Eschatology being one.

hifi-gustavo:
Build console output for commit: 23aa6267551c7db26b1c18f2bacfcf1be9e36e41 available at: https://gist.github.com/1cf2a3f6d5d5876fa14eae3877ad2cf7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8909-23aa626.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8909-23aa626.dmg

libraries/render/src/render/Scene.cpp
```diff
@@ -48,6 +48,10 @@ Scene::Scene(glm::vec3 origin, float size) :
     _items.push_back(Item()); // add the itemID #0 to nothing
 }
 
+Scene::~Scene() {
+    qDebug() << "Scene::~Scene()";
```
sethalves: still want this qDebug ?

cozza13:
* Quit while connected to a site containing other avatars. **Tested**
  * Quit while connected to a site with entities that are being created and destroyed.  Probably the easiest way to test this is to have a user shooting the ping pong gun. **tested**
  * Quit while connected to a site with PolyVox entities, hifi://Eschatology being one. **Tested**

PR #8924, opened 2016-10-26
Fix race condition in BatchLoader

huffman:
Previously a thread could be inside BatchLoader's callback while it gets deleted. This attempts to fix that particular issue.

### DIFF
--- a/libraries/script-engine/src/BatchLoader.cpp
+++ b/libraries/script-engine/src/BatchLoader.cpp
@@ -50,18 +50,18 @@ void BatchLoader::start() {
 
         qCDebug(scriptengine) << "Loading script at " << url;
 
-		auto scriptCache = DependencyManager::get<ScriptCache>();
+        auto scriptCache = DependencyManager::get<ScriptCache>();
 
-		// Use a proxy callback to handle the call and emit the signal in a thread-safe way.
-		// If BatchLoader is deleted before the callback is called, the subsequent "emit" call will not do
-		// anything.
-		ScriptCacheSignalProxy* proxy = new ScriptCacheSignalProxy(scriptCache.data());
-		scriptCache->getScriptContents(url.toString(), [proxy](const QString& url, const QString& contents, bool isURL, bool success) {
-			proxy->receivedContent(url, contents, isURL, success);
-			proxy->deleteLater();
-		}, false);
+        // Use a proxy callback to handle the call and emit the signal in a thread-safe way.
+        // If BatchLoader is deleted before the callback is called, the subsequent "emit" call will not do
+        // anything.
+        ScriptCacheSignalProxy* proxy = new ScriptCacheSignalProxy(scriptCache.data());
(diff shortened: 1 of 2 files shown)


huffman:
This PR fixes a crash when loading scripts. We will want to test that script loading in general is still working, and hit a few domains multiple times to make sure a crash doesn't occur.

To test:

- Go to cellscience, both with a hot and cold cache
- Go to welcome, both with a hot and cold cache
- Make sure defaultScripts is working, in particular near grab, far grab, and teleport
- Try using a few different entities with entity scripts and ensure they are working

hifi-gustavo:
Build console output for commit: 10e6157ab9517ac1f58cfb8d263dac041b5026ba available at: https://gist.github.com/0b36c8b6c8e7897b3fcfba011c42c3db<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8924-10e6157.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8924-10e6157.dmg

libraries/script-engine/src/BatchLoader.cpp
```diff
@@ -44,33 +44,40 @@ void BatchLoader::start() {
...
-                qCDebug(scriptengine) << "Could not load" << url;
+                qCDebug(scriptengine) << "Could not load: " << url;
             }
 
             if (!_finished && _urls.size() == _data.size()) {
                 _finished = true;
                 emit finished(_data);
             }
-        }, false);
+		});
```
AndrewMeadows: Some of these changes have bad formatting -- they have a tab character instead of four spaces.

huffman: Thanks, I'll fix that.

hifi-gustavo:
Build console output for commit: 946f3782f36c44b751c89a566eda56e637f13546 available at: https://gist.github.com/0b36c8b6c8e7897b3fcfba011c42c3db<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8924-946f378.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8924-946f378.dmg

OmegaHeron:

Visited test locations (and others)
Scripts work, things load as expected.  Grab/TP and entity scripts seem to work as expected.

QA Pass

Not related to this PR 
Nav buttons in cellscience continue to be a random thing.  First visit 99.9999% all show, subsequent visits shows missing - finally after many leave/return cycles none show.

**I have a strong suspicion this has everything to do with me having Interface on monitor 1 and this on monitor 2 where I leave Interface to come here and type notes - it's all tied up in the FPS loss when Interface is not foreground app.  That also seems tied to doing it in cellscience (dev-cell) as it's only domain (so far) where I see that repro.  There's still an ugly fire breathing dragon hiding somewhere, but, it's nothing to do with this PR.**

Hexokinase continues to sometimes load as a black background void - eventually it loads zone visual.  Nothing new there.

PR #8930, opened 2016-10-27
fix the timing of the launchedFromSteam property

ZappoMan:


interface/src/Application.cpp
```diff
@@ -534,6 +534,8 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer, bo
     _maxOctreePPS(maxOctreePacketsPerSecond.get()),
     _lastFaceTrackerUpdate(0)
 {
+    setProperty("com.highfidelity.launchedFromSteam", launchedFromSteam);
```
Atlante45: Why not just use `SteamClient::isRunning()` instead of passing an argument?

ZappoMan: ok

hifi-gustavo:
Build console output for commit: c6d89f181a061e228c6c1d9c38d0b795c5916087 available at: https://gist.github.com/94855542a52ae01ffdb74b39131ce0a9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8930-c6d89f1.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8930-c6d89f1.dmg

PR #8947, opened 2016-10-28
Fix race condition in BatchLoader

huffman:


huffman:
This fixes a race condition when a script is in the script cache.

To test:

 1. Go to cellscience, both with a hot and cold cache
 1. Go to welcome, both with a hot and cold cache
 1. Make sure defaultScripts is working, in particular near grab, far grab, and teleport
 1. Try using a few different entities with entity scripts and ensure they are working

hifi-gustavo:
Build console output for commit: 52c74f776c27638f7ca9d2c36db4ba6c6638cbb4 available at: https://gist.github.com/3f3b87a0d6c50234dcb296768aa60da2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8947-52c74f7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8947-52c74f7.dmg

sethalves:
https://highfidelity.fogbugz.com/f/cases/2125/Mouse-pointer-not-moving-objects-in-HMD

howard-stearns:
I'm not sure this is working. I'm having problems with mouse pointer in hmd (with hand controllers turned off), and dev-cell doesn't have the four teleport buttons.  But these could be different problems....

huffman:
@howard-stearns Is there a build where mouse pointer in HMD is working for you?

howard-stearns:
I don't know when the last working build was...
But, with Seth's help, we found that IF I killed all scripts and then reloaded defaults, while not in a state of loading other content, then it works.  The theory is that maybe there's some load-order dependency among the default scripts. I have gotten it to fail again with lots of reloads of cache and getting new default scripts (e.g., while in dev-cell), but I have not gotten it to fail in clean-and-patient loading.

I don't know enough about the motivations for the loader changes to know whether to consider this a pass or fail. I defer to your judgement.

huffman:
Ah, got it. This fixes something different. This fixed a situation where defaultScripts would not finish loading at all, and you wouldn't even be able to stop or reload it.

ZappoMan:
Howard's bug sounds equally problematic... if the result is that scripts aren't working properly... do we have a theory on that? I doubt regular users will figure out the idea of reloading all scripts

huffman:
Maybe I'm misunderstanding. Is that a new bug that only exists on this PR? I agree it should be fixed, but if it's not related to this PR then we should file a bug.

PR #8955, opened 2016-10-28
fix obj textures

sethalves:
- fix a bug that could cause obj textures to be flipped around the wrong way

### DIFF
--- a/libraries/fbx/src/OBJReader.cpp
+++ b/libraries/fbx/src/OBJReader.cpp
@@ -123,7 +123,9 @@ glm::vec3 OBJTokenizer::getVec3() {
     return v;
 }
 glm::vec2 OBJTokenizer::getVec2() {
-    auto v = glm::vec2(getFloat(), 1.0f - getFloat());  // OBJ has an odd sense of u, v. Also N.B.: getFloat() has side-effect
+    float uCoord = getFloat();
+    float vCoord = 1.0f - getFloat();
+    auto v = glm::vec2(uCoord, vCoord);
     while (isNextTokenFloat()) {
         // there can be a w, but we don't handle that
         nextToken();


sethalves:
http://stackoverflow.com/questions/22616986/order-of-evaluation-of-arguments-in-function-calling

sethalves:
to test:
- load an obj model with a text
- make sure the texture lines up correctly

libraries/fbx/src/OBJReader.cpp
```diff
@@ -123,7 +123,9 @@ glm::vec3 OBJTokenizer::getVec3() {
     return v;
 }
 glm::vec2 OBJTokenizer::getVec2() {
-    auto v = glm::vec2(getFloat(), 1.0f - getFloat());  // OBJ has an odd sense of u, v. Also N.B.: getFloat() has side-effect
+    float uCoord = getFloat();
+    float vCoord = 1.0f - getFloat();
+    auto v = glm::vec2(uCoord, vCoord);
```
huffman: I must be missing something - doesn't this end up being the same?

hifi-gustavo:
Build console output for commit: 43536e0046963028be210749b32416a849f78aaa available at: https://gist.github.com/7c1df95ff6f00ddf70782afedbe302c8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8955-43536e0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8955-43536e0.dmg

libraries/fbx/src/OBJReader.cpp
huffman: Ah, got it, thanks!

sethalves: getFloat has the side-effect of moving the lexer forward.  If the compiler chooses to evaluate 1.0-f -getFloat() before getFloat(), it ends up swapping the u and v coordinates.

huffman: (I should have read the commit message)

ZappoMan:
do we have a test plan for this?

cozza13:
tested with an OBJ and multiple testures 

capture

cozza13:
Currenty Master

capture

PR #8956, opened 2016-10-28
Enable forced min spec failure for testing

jherico:
## Testing

On a system with SteamVR running, set the environment variable `HIFI_FAIL_MIN_SPEC` to any value and then run interface.  You should see the failure message in the HMD.

### DIFF
--- a/interface/src/main.cpp
+++ b/interface/src/main.cpp
@@ -41,7 +41,11 @@
 extern "C" {
     typedef int(__stdcall * CHECKMINSPECPROC) ();
 }
+
+static const QString DEBUG_FAIL_MIN_SPEC_FLAG("HIFI_FAIL_MIN_SPEC");
 #endif
+
+
 
 int main(int argc, const char* argv[]) {
 #if HAS_BUGSPLAT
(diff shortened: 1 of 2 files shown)


ZappoMan:
you don't need to do this... you just need to include a `steam_appid.txt` file in the path of the `interface.exe` (and that file needs to have the correct app id text in it, which @Atlante45 can tell you)

hifi-gustavo:
Build console output for commit: de8242e4eefffc9718225811f1b1ed565b293acc available at: https://gist.github.com/bc47435b79d624f6e09f2d4e30fc1017<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8956-de8242e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8956-de8242e.dmg

jherico:
@ZappoMan having the steam app ID spoofs the `SteamClient::isRunning()` but not the actual machine check.  Alan is doing testing of his graphic and doesn't have easy access to an i5.

ZappoMan:
@jherico -- oooook

cozza13:
tested and got the popup. confirmed that it did disappear after a while

ZappoMan:
I don't think we need to or want to merge this... if the purpose is for @AlanZimmerman to simply see/test/read the popup -- let him test this PR and not cloud the code base.

PR #8957, opened 2016-10-29
fix broken eyes in mirrored animations

AndrewMeadows:
* eyes now stay in their sockets during mirrored animations

### DIFF
--- a/libraries/animation/src/AnimSkeleton.cpp
+++ b/libraries/animation/src/AnimSkeleton.cpp
@@ -118,11 +118,26 @@ void AnimSkeleton::convertAbsoluteRotationsToRelative(std::vector<glm::quat>& ro
     }
 }
 
+void AnimSkeleton::saveNonMirroredPoses(const AnimPoseVec& poses) const {
+    _nonMirroredPoses.clear();
+    for (int i = 0; i < (int)_nonMirroredIndices.size(); ++i) {
+        _nonMirroredPoses.push_back(poses[_nonMirroredIndices[i]]);
+    }
+}
+
+void AnimSkeleton::restoreNonMirroredPoses(AnimPoseVec& poses) const {
+    for (int i = 0; i < (int)_nonMirroredIndices.size(); ++i) {
+        int index = _nonMirroredIndices[i];
+        poses[index] = _nonMirroredPoses[i];
+    }
+}
 
 void AnimSkeleton::mirrorRelativePoses(AnimPoseVec& poses) const {
+    saveNonMirroredPoses(poses);
     convertRelativePosesToAbsolute(poses);
     mirrorAbsolutePoses(poses);
     convertAbsolutePosesToRelative(poses);
+    restoreNonMirroredPoses(poses);
 }
 
 void AnimSkeleton::mirrorAbsolutePoses(AnimPoseVec& poses) const {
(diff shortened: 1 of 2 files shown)


AndrewMeadows:
This is a fix for a bug where, for some avatars and some animations, the eyes will wander outside of their head sockets.  The internal link is:

https://highfidelity.fogbugz.com/f/cases/1128/Avatar-eyeballs-incorrectly-animated

This is how to reproduce/test:

(1) Login in desktop mode
(2) Change avatar to use this URL: https://hifi-content.s3.amazonaws.com/ozan/dev/avatars/uploadvr/taylor/taylor.fst
(3) Press H to go into mirror mode where the camera is facing your avatar
(4) Using the arrow keys turn right and then turn left
--> for buggy interfaces the eyes will leave the eye sockets a little bit for one of the directions
--> for the interface in this PR the eyes should stay seated for both directions

libraries/animation/src/AnimSkeleton.cpp
```diff
@@ -118,11 +118,26 @@ void AnimSkeleton::convertAbsoluteRotationsToRelative(std::vector<glm::quat>& ro
...
+        poses[index] = _nonMirroredPoses[i];
+    }
+}
 
 void AnimSkeleton::mirrorRelativePoses(AnimPoseVec& poses) const {
+    saveNonMirroredPoses(poses);
     convertRelativePosesToAbsolute(poses);
     mirrorAbsolutePoses(poses);
     convertAbsolutePosesToRelative(poses);
+    restoreNonMirroredPoses(poses);
```
AndrewMeadows: I tried to selectively mirror parts of the animation but that just didn't work, so instead I just remember the non-mirrored pose values and restore them after the mirror operation.

hifi-gustavo:
Build console output for commit: c8d41463512803ffa3309796c91050585874ee42 available at: https://gist.github.com/527300686b9b5085e94677895f4a51c2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8957-c8d4146.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8957-c8d4146.dmg

hyperlogic:
I could be wrong, but I think you can do this without having to add a vector of ints and vector of poses to the AnimSkeleton.  Can't you just selectively ignore LeftEye and RightEye when building the mirrorMap?  Something like:

        if (_joints[i].name.startsWith("Left") && _joints[i].name != "LeftEye") {
            QString mirrorJointName = QString(_joints[i].name).replace(0, 4, "Right");
            mirrorJointIndex = nameToJointIndex(mirrorJointName);
        } else if (_joints[i].name.startsWith("Right") && _joints[i].name != "RightEye") {
            QString mirrorJointName = QString(_joints[i].name).replace(0, 5, "Left");
            mirrorJointIndex = nameToJointIndex(mirrorJointName);
        }

libraries/animation/src/AnimSkeleton.cpp
hyperlogic: See my comments below about using the _mirrorMap to handle this.

AndrewMeadows:
@hyperlogic, no we can't just not include the eyes in the mirror map because of the implementation of AnimSkeleton::mirrorAbsolutePoses() which mirrors _every_ pose rather than just _every mirrorable_ pose.                                                                                    
 ```
 void AnimSkeleton::mirrorAbsolutePoses(AnimPoseVec& poses) const {                                         
      AnimPoseVec temp = poses;                                                                              
      for (int i = 0; i < (int)poses.size(); i++) {                                                          
          poses[_mirrorMap[i]] = temp[i].mirror();                                                           
      }                                                                                                      
  }                                                                                                          
```
However, I now realize that if I were to modify that function to only loop over the poses that have an entry in _mirrorMap then maybe we could just omit the eyes from the map.  I'll try it.

cozza13:
tested on master and this PR. It works

hyperlogic:
Ok, I see, it's not that simple.  Thanks for taking a deeper look into this.

PR #8960, opened 2016-10-29
Introduce Instanced Stereo Drawcall

samcake:
This PR introduces the instanced stereo drawcall technique to acheive the  drawcall of a shape in stereo to the left and right veiwports of the stereo framebuffer in a single instanced drawcall.

The previous path for stereo drawcall was:
- set viewport to left side
- bind left camera
- draw
- set vieport to right side
- bind right camera 
- draw

THe new path does the following:
- draw instanced (x2)
-- in the shader the instanceID oddity gives us the side (left or right)
-- based on the side, the correct cameraTransform element is used (there is now an array of 2)
-- the vertex position (after clip) is transformed to the corresponding side of the framebuffer
-- the vertices are also clipped against the vertical side of their frustum to avoiding overlapping on the oother side of the framebuffer. this works with the viewport still covering the full region of the framebuffer touched by the stereo image beeing produced.

This new path require one more version of the shaders specialized for the purpose of stereo. this cost an extra compilation that could be seen as a  slow starting wait. we could try to mitigate that by beeing more accurate about what is really needed to be recompiled.

This path should show potentially slightly better performances for the drawing of shapes in stereo compared to current master.
THe gpu time should be equivalent since the amount of work is almost the same. the extra cost of the clipping shouldn't be an issue.
The cpu time for should be improved since we need less gl commands, no need for the change of viewport or rebinding of the camera ubo.

OVerall this PR could show improvments in scenes with a lot of drawcalls, meaning with a large number of shapes render item in view.
A good example is "polyword"

This PR paves the way for using NVidia specific stereo drawcall features on capable hardware...


hifi-gustavo:
Build console output for commit: 9adbb42a6ced55fd4a27369f02f370a1c5e6d40d available at: https://gist.github.com/ee44c2d436bfee3371d232423fa746fd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8960-9adbb42.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8960-9adbb42.dmg

hifi-gustavo:
Build console output for commit: d5f44befc68d904d5387442dc73fc67f7f02f6be available at: https://gist.github.com/ee44c2d436bfee3371d232423fa746fd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8960-d5f44be.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8960-d5f44be.dmg

hifi-gustavo:
Build console output for commit: 37b3c96131ea81e52a4f2a8957e34afb43e9ba36 available at: https://gist.github.com/ee44c2d436bfee3371d232423fa746fd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8960-37b3c96.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8960-37b3c96.dmg

samcake:
Test Plan:
Equivalent features  and look as master across platform.
Performances should be the same in mono.
This PR should NEVER be slower in Frame rate than current Master.
In domains with large number of opaque render items this should be faster in stereo.

For example, in Polyworld the counter of render item is very quickly around 2500.
THis should show better performances than master.

Nex-Pro:
### PR #8960

**CPU:** 4.4 ghz Intel i7
**RAM:** 16 GB
**GPU:** Nvidia gtx 1070 8gb

**Testing OS:** win 10 pro

**HMD:** HTC Vive.
**Hand Controllers:** Vive

### Good improvement.

Made numerous comparisons between this and current release.
In each case I waited for all content to load before proceeding.

Increased framerates at polyworld. There's a small screen right side of the door in the library.
When looking in that direction there's a large hit on the frame rate. This build was less badly affected.
Typically managing five to ten fps more than the current release build.

But 55 - 60 fps in HMD isn't great even in this build.

If you want a horrible experience
try flying and turning with your head
while doing a slalom in and out of the light cables hanging from the ceiling.

It isn't bad until you reach the far wall and turn back.
Then the frame rate plummets as you turn left and right making the return trip.

ZappoMan:
testing on mac - please don't merge yet

ZappoMan:
passes on mac with DK2

PR #8966, opened 2016-10-31
Send KillAvatar message to AudioMixer

davidkelly:
Similar to sending it to the `AvatarMixer`.  In this case, it is so it can remove the listening agent avatar audio streams.  Otherwise they stick around forever, and eventually swamp the audio mixer.

### DIFF
--- a/assignment-client/src/Agent.cpp
+++ b/assignment-client/src/Agent.cpp
@@ -424,7 +424,7 @@ void Agent::setIsListeningToAudioStream(bool isListeningToAudioStream) {
         },
             [&](const SharedNodePointer& node) {
             qDebug() << "sending KillAvatar message to Audio Mixers";
-            auto packet = NLPacket::create(PacketType::KillAvatar, NUM_BYTES_RFC4122_UUID);
+            auto packet = NLPacket::create(PacketType::KillAvatar, NUM_BYTES_RFC4122_UUID, true);
             packet->write(getSessionUUID().toRfc4122());
             nodeList->sendPacket(std::move(packet), *node);
         });
@@ -475,7 +475,7 @@ void Agent::setIsAvatar(bool isAvatar) {
             },
                 [&](const SharedNodePointer& node) {
                 qDebug() << "sending KillAvatar message to Avatar and Audio Mixers";
-                auto packet = NLPacket::create(PacketType::KillAvatar, NUM_BYTES_RFC4122_UUID);
+                auto packet = NLPacket::create(PacketType::KillAvatar, NUM_BYTES_RFC4122_UUID, true);
                 packet->write(getSessionUUID().toRfc4122());
                 nodeList->sendPacket(std::move(packet), *node);
             });
--- a/assignment-client/src/Agent.cpp
+++ b/assignment-client/src/Agent.cpp
@@ -88,6 +88,7 @@ void Agent::playAvatarSound(SharedSoundPointer sound) {
         QMetaObject::invokeMethod(this, "playAvatarSound", Q_ARG(SharedSoundPointer, sound));
         return;
     } else {
+        _numAvatarSoundSentBytes = 0;
         setAvatarSound(sound);
     }
 }
--- a/assignment-client/src/Agent.cpp
+++ b/assignment-client/src/Agent.cpp
@@ -424,9 +424,9 @@ void Agent::setIsListeningToAudioStream(bool isListeningToAudioStream) {
         },
             [&](const SharedNodePointer& node) {
             qDebug() << "sending KillAvatar message to Audio Mixers";
-            auto packetList = NLPacketList::create(PacketType::KillAvatar, QByteArray(), true, true);
-            packetList->write(getSessionUUID().toRfc4122());
-            nodeList->sendPacketList(std::move(packetList), *node);
+            auto packet = NLPacket::create(PacketType::KillAvatar, NUM_BYTES_RFC4122_UUID);
+            packet->write(getSessionUUID().toRfc4122());
+            nodeList->sendPacket(std::move(packet), *node);
         });
 
     }
@@ -475,9 +475,9 @@ void Agent::setIsAvatar(bool isAvatar) {
             },
                 [&](const SharedNodePointer& node) {
                 qDebug() << "sending KillAvatar message to Avatar and Audio Mixers";
-                auto packetList = NLPacketList::create(PacketType::KillAvatar, QByteArray(), true, true);
-                packetList->write(getSessionUUID().toRfc4122());
-                nodeList->sendPacketList(std::move(packetList), *node);
+                auto packet = NLPacket::create(PacketType::KillAvatar, NUM_BYTES_RFC4122_UUID);
+                packet->write(getSessionUUID().toRfc4122());
+                nodeList->sendPacket(std::move(packet), *node);
             });
         }
         emit stopAvatarAudioTimer();
--- a/assignment-client/src/Agent.cpp
+++ b/assignment-client/src/Agent.cpp
@@ -439,7 +439,8 @@ void Agent::setIsAvatar(bool isAvatar) {
             packetList->write(getSessionUUID().toRfc4122());
             nodeList->eachMatchingNode(
                 [&](const SharedNodePointer& node)->bool {
-                return node->getType() == NodeType::AvatarMixer && node->getActiveSocket();
+                return (node->getType() == NodeType::AvatarMixer || node->getType() == NodeType::AudioMixer)
+                        && node->getActiveSocket();
             },
                 [&](const SharedNodePointer& node) {
                 nodeList->sendPacketList(std::move(packetList), *node);
--- a/assignment-client/src/Agent.cpp
+++ b/assignment-client/src/Agent.cpp
@@ -404,8 +404,38 @@ QUuid Agent::getSessionUUID() const {
     return DependencyManager::get<NodeList>()->getSessionUUID();
 }
 
+void Agent::setIsListeningToAudioStream(bool isListeningToAudioStream) { 
+    // this must happen on Agent's main thread
+    if (QThread::currentThread() != thread()) {
+        QMetaObject::invokeMethod(this, "setIsListeningToAudioStream", Q_ARG(bool, isListeningToAudioStream));
+        return;
+    }
+    if (_isListeningToAudioStream) {
+        // have to tell just the audio mixer to KillAvatar. 
+
+        auto nodeList = DependencyManager::get<NodeList>();
+        nodeList->eachMatchingNode(
+            [&](const SharedNodePointer& node)->bool {
+            return (node->getType() == NodeType::AudioMixer)
+            && node->getActiveSocket();
+        },
+            [&](const SharedNodePointer& node) {
+            qDebug() << "sending KillAvatar message to Avatar and Audio Mixers";
+            auto packetList = NLPacketList::create(PacketType::KillAvatar, QByteArray(), true, true);
+            packetList->write(getSessionUUID().toRfc4122());
+            nodeList->sendPacketList(std::move(packetList), *node);
+        });
+
+    }
+    _isListeningToAudioStream = isListeningToAudioStream; 
+}
 
 void Agent::setIsAvatar(bool isAvatar) {
+    // this must happen on Agent's main thread
+    if (QThread::currentThread() != thread()) {
+        QMetaObject::invokeMethod(this, "setIsAvatar", Q_ARG(bool, isAvatar));
+        return;
+    }
     _isAvatar = isAvatar;
 
     if (_isAvatar && !_avatarIdentityTimer) {
@@ -435,14 +465,15 @@ void Agent::setIsAvatar(bool isAvatar) {
             // when we stop sending identity, but then get woken up again by the mixer itself, which sends
             // identity packets to everyone. Here we explicitly tell the mixer to kill the entry for us.
             auto nodeList = DependencyManager::get<NodeList>();
-            auto packetList = NLPacketList::create(PacketType::KillAvatar, QByteArray(), true, true);
-            packetList->write(getSessionUUID().toRfc4122());
             nodeList->eachMatchingNode(
                 [&](const SharedNodePointer& node)->bool {
                 return (node->getType() == NodeType::AvatarMixer || node->getType() == NodeType::AudioMixer)
                         && node->getActiveSocket();
             },
                 [&](const SharedNodePointer& node) {
+                qDebug() << "sending KillAvatar message to Avatar and Audio Mixers";
+                auto packetList = NLPacketList::create(PacketType::KillAvatar, QByteArray(), true, true);
+                packetList->write(getSessionUUID().toRfc4122());
                 nodeList->sendPacketList(std::move(packetList), *node);
             });
         }
--- a/assignment-client/src/Agent.cpp
+++ b/assignment-client/src/Agent.cpp
@@ -88,7 +88,10 @@ void Agent::playAvatarSound(SharedSoundPointer sound) {
         QMetaObject::invokeMethod(this, "playAvatarSound", Q_ARG(SharedSoundPointer, sound));
         return;
     } else {
-        _numAvatarSoundSentBytes = 0;
+        // TODO: seems to add occasional artifact in tests.  I believe it is 
+        // correct to do this, but need to figure out for sure, so commenting this
+        // out until I verify.  
+        // _numAvatarSoundSentBytes = 0;
         setAvatarSound(sound);
     }
 }
--- a/assignment-client/src/Agent.cpp
+++ b/assignment-client/src/Agent.cpp
@@ -420,11 +420,10 @@ void Agent::setIsListeningToAudioStream(bool isListeningToAudioStream) {
         auto nodeList = DependencyManager::get<NodeList>();
         nodeList->eachMatchingNode(
             [&](const SharedNodePointer& node)->bool {
-            return (node->getType() == NodeType::AudioMixer)
-            && node->getActiveSocket();
+            return (node->getType() == NodeType::AudioMixer) && node->getActiveSocket();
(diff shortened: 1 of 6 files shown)


hifi-gustavo:
Build console output for commit: 80e38c395ecf100d6948dd5f59beaa2b1eb951e3 available at: https://gist.github.com/c3541006e4c4b80bd5b26687d310f459<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-80e38c3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-80e38c3.dmg

ZappoMan:
I don't understand why the agent needs to do this? Why does the agent need special processing that's not needed by interfaces?

ZappoMan:
I want @birarda - to review and approve... but my expectations are that AC/Script Agents -- should look/act like interfaces... and so they shouldn't need this extra support.

birarda:
"Otherwise they stick around forever"

Nope. The AudioMixer cleans up the AudioMixerClientData when the node is removed from its list (which will happen after a period of inactivity). It also cleans up the HRTFs on other node AudioMixerClientData objects when the node is killing.

All we're doing here is removing the Agent approximately 5 seconds earlier than it would otherwise be removed. I'm not sure why we need this.

davidkelly:
They go away just fine when the script ends.  That isn't the issue.  We were seeing the listeners stuck at high numbers when we toggle the `Agent.isAvatar` between true and false.  The agent sticks around, but stops sending audio and tells the avatar mixer it is no longer an avatar.   Hence the node sticks around in `AudioMixer`, and the mixer continues to mix audio for it, even though is is no longer an avatar or sending audio.  So I'm letting the `AudioMixer` know similar to how we let the `AvatarMixer` know.   AFAICT interface doesn't have a 'not an avatar' mode where it doesn't want you to mix audio for it?

birarda:
I see - didn't expect a script actively toggling the isAvatar back and forth. Makes sense.

assignment-client/src/audio/AudioMixer.cpp
```diff
@@ -598,6 +599,14 @@ void AudioMixer::handleNodeKilled(SharedNodePointer killedNode) {
     });
 }
 
+void AudioMixer::handleKillAvatarPacket(QSharedPointer<ReceivedMessage> packet, SharedNodePointer sendingNode) {
+    auto clientData = dynamic_cast<AudioMixerClientData*>(sendingNode->getLinkedData());
+    if (clientData) {
+        clientData->removeAgentAvatarAudioStream();
+    }
```
birarda: If the node is going to stay connected, you probably need to do the same thing we're doing in `nodeKilled` to clean up HRTF objects for other nodes.

davidkelly: Yea for sure -- that is coming.  Was just experimenting with the lifetime stuff for the audio stream first, to be sure this did the trick.  I didn't realize people would be that interested in this while it was DNM.  Anyways that's coming shortly...

davidkelly: OH -- the reason we toggle is for doing load tests.   Summon a large number of avatars, have they scream and dance, etc...  Then go away.  And then maybe come back later, etc...  Probably not something we have done in the past?   Anyways that's how we ran into this issue...

hifi-gustavo:
Build console output for commit: e63e9bc0176cdfd5b12b0c974000b0fc0c9110e1 available at: https://gist.github.com/c3541006e4c4b80bd5b26687d310f459<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-e63e9bc.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-e63e9bc.dmg

hifi-gustavo:
Build console output for commit: 5714063eff61701955e4006cc1ba86fa3a967337 available at: https://gist.github.com/c3541006e4c4b80bd5b26687d310f459<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-5714063.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-5714063.dmg

hifi-gustavo:
Build console output for commit: fb2968a4191675610451b185d51fd0c1e362e4c8 available at: https://gist.github.com/c3541006e4c4b80bd5b26687d310f459<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-fb2968a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-fb2968a.dmg

davidkelly:
# Test Plan

(1) Serve up `scripts/developer/tests/performance/crowd-agent.js` locally (using maybe `python -m SimpleHTTPServer 8888` from that directory.
(2) Have sandbox serve up a bunch of instances of it (I use like 15), by putting something like `http://localhost:8888/crowd-agent.js?v=xxx` in the domain server's settings (`http://localhost:40100`), under Persistent Scripts. 
(3) save, let it restart, and just for fun quit and restart the sandbox
(4) wait until no more agents are queued (one starts every 15 seconds), by looking at the nodes page in the domain server.
(5) click on the audio mixer in the nodes section of the domain server's web interface.  You will see there is currently one listener
(6) in interface, load 'scripts/developer/tests/performance/summon.js`.  
(7) it should complain that you didn't get 24 total, but that's ok.
(8) listen, it should sound ok (but it will get old after awhile listening to it)
(9) there should be the correct number of listeners -- you plus however many agents you spun up.
(10) now stop the summon.js script.  
(11) listeners should go to 0 - it takes a bit as they go away over time.
(12) try this a few times.

hifi-gustavo:
Build console output for commit: 404c08b0c73f0edf974126f3d06e0dc807404f2e available at: https://gist.github.com/c3541006e4c4b80bd5b26687d310f459<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-404c08b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-404c08b.dmg

assignment-client/src/Agent.cpp
```diff
@@ -404,8 +408,38 @@ QUuid Agent::getSessionUUID() const {
...
+        // have to tell just the audio mixer to KillAvatar. 
+
+        auto nodeList = DependencyManager::get<NodeList>();
+        nodeList->eachMatchingNode(
+            [&](const SharedNodePointer& node)->bool {
+            return (node->getType() == NodeType::AudioMixer)
+            && node->getActiveSocket();
+        },
+            [&](const SharedNodePointer& node) {
+            qDebug() << "sending KillAvatar message to Avatar and Audio Mixers";
```
birarda: you're only sending to audio here, not avatar

assignment-client/src/Agent.cpp
```diff
@@ -404,8 +408,38 @@ QUuid Agent::getSessionUUID() const {
...
+        return;
+    }
+    if (_isListeningToAudioStream) {
+        // have to tell just the audio mixer to KillAvatar. 
+
+        auto nodeList = DependencyManager::get<NodeList>();
+        nodeList->eachMatchingNode(
+            [&](const SharedNodePointer& node)->bool {
+            return (node->getType() == NodeType::AudioMixer)
+            && node->getActiveSocket();
```
birarda: this should be on the same line as 423 or indented

assignment-client/src/audio/AudioMixer.cpp
```diff
@@ -598,6 +599,17 @@ void AudioMixer::handleNodeKilled(SharedNodePointer killedNode) {
     });
 }
 
+void AudioMixer::handleKillAvatarPacket(QSharedPointer<ReceivedMessage> packet, SharedNodePointer sendingNode) {
+    auto clientData = dynamic_cast<AudioMixerClientData*>(sendingNode->getLinkedData());
+    if (clientData) {
+        clientData->removeAgentAvatarAudioStream();
+        // when you don't specify a stream, this removes just the QUuid() 
+        // stream (which is the avatar's mic stream)
+        clientData->removeHRTFForStream(sendingNode->getUUID());
```
birarda: This would be removing the HRTF for the agent's own stream - that's not what we need here.  We need to remove the agent's HRTF from all other connected nodes.

zfox23:
I ran through @davidkelly's (very clear!) test plan a few times. The first time through, I ran into an issue where I only saw 15/16 listeners in the audio-mixer. I was unable to reproduce this issue throughout further testing, and saw/heard only intended behavior otherwise.

"QA Approved" label added.

assignment-client/src/audio/AudioMixer.cpp
davidkelly: yep - nice catch - will fix now

hifi-gustavo:
Build console output for commit: 0b80610e65c352f0cbc0b11d79b751e0aac2430e available at: https://gist.github.com/c3541006e4c4b80bd5b26687d310f459<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-0b80610.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-0b80610.dmg

hifi-gustavo:
Build console output for commit: c1feac971b009e0a0815b9dc7acdbe6de199e1d1 available at: https://gist.github.com/c3541006e4c4b80bd5b26687d310f459<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-c1feac9.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-c1feac9.dmg

assignment-client/src/Agent.cpp
```diff
@@ -404,8 +408,37 @@ QUuid Agent::getSessionUUID() const {
...
+        auto nodeList = DependencyManager::get<NodeList>();
+        nodeList->eachMatchingNode(
+            [&](const SharedNodePointer& node)->bool {
+            return (node->getType() == NodeType::AudioMixer) && node->getActiveSocket();
+        },
+            [&](const SharedNodePointer& node) {
+            qDebug() << "sending KillAvatar message to Audio Mixers";
+            auto packetList = NLPacketList::create(PacketType::KillAvatar, QByteArray(), true, true);
+            packetList->write(getSessionUUID().toRfc4122());
+            nodeList->sendPacketList(std::move(packetList), *node);
```
birarda: this doesn't need to be an NLPacketList - should just be an NLPacket. We should also be able to specify the exact size (NUM_BYTES_RFC4122_UUID) of the packet.

assignment-client/src/Agent.cpp
```diff
@@ -435,13 +468,15 @@ void Agent::setIsAvatar(bool isAvatar) {
...
                 [&](const SharedNodePointer& node)->bool {
-                return node->getType() == NodeType::AvatarMixer && node->getActiveSocket();
+                return (node->getType() == NodeType::AvatarMixer || node->getType() == NodeType::AudioMixer)
+                        && node->getActiveSocket();
             },
                 [&](const SharedNodePointer& node) {
+                qDebug() << "sending KillAvatar message to Avatar and Audio Mixers";
+                auto packetList = NLPacketList::create(PacketType::KillAvatar, QByteArray(), true, true);
+                packetList->write(getSessionUUID().toRfc4122());
                 nodeList->sendPacketList(std::move(packetList), *node);
```
birarda: This should also not be a packet list and we should specify the size.

assignment-client/src/audio/AudioMixer.cpp
```diff
@@ -598,6 +599,23 @@ void AudioMixer::handleNodeKilled(SharedNodePointer killedNode) {
     });
 }
 
+void AudioMixer::handleKillAvatarPacket(QSharedPointer<ReceivedMessage> packet, SharedNodePointer sendingNode) {
+    auto clientData = dynamic_cast<AudioMixerClientData*>(sendingNode->getLinkedData());
+    if (clientData) {
+        QUuid streamID = clientData->removeAgentAvatarAudioStream();
+        if (streamID != QUuid()) {
```
birarda: Won't this be the case for agent avatar audio? I think we want `removeHRTFForNode` here

assignment-client/src/Agent.cpp
davidkelly: I'd copied existing code -- will switch it

assignment-client/src/audio/AudioMixer.cpp
davidkelly: What I wanted to support was only removing that agent avatar stream's hrtf from each node -- in case perhaps we can have injectors in that agent.  I'll review my logic here though.

hifi-gustavo:
Build console output for commit: ec0382daf75e3d8be61926b95e42651bd4c45ed5 available at: https://gist.github.com/c3541006e4c4b80bd5b26687d310f459<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-ec0382d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-ec0382d.dmg

assignment-client/src/Agent.cpp
```diff
@@ -424,9 +424,9 @@ void Agent::setIsListeningToAudioStream(bool isListeningToAudioStream) {
         },
             [&](const SharedNodePointer& node) {
             qDebug() << "sending KillAvatar message to Audio Mixers";
-            auto packetList = NLPacketList::create(PacketType::KillAvatar, QByteArray(), true, true);
-            packetList->write(getSessionUUID().toRfc4122());
-            nodeList->sendPacketList(std::move(packetList), *node);
+            auto packet = NLPacket::create(PacketType::KillAvatar, NUM_BYTES_RFC4122_UUID);
+            packet->write(getSessionUUID().toRfc4122());
+            nodeList->sendPacket(std::move(packet), *node);
```
birarda: this was reliable before when it was a packet list - we might still want that to be the case

assignment-client/src/Agent.cpp
```diff
@@ -475,9 +475,9 @@ void Agent::setIsAvatar(bool isAvatar) {
             },
                 [&](const SharedNodePointer& node) {
                 qDebug() << "sending KillAvatar message to Avatar and Audio Mixers";
-                auto packetList = NLPacketList::create(PacketType::KillAvatar, QByteArray(), true, true);
-                packetList->write(getSessionUUID().toRfc4122());
-                nodeList->sendPacketList(std::move(packetList), *node);
+                auto packet = NLPacket::create(PacketType::KillAvatar, NUM_BYTES_RFC4122_UUID);
+                packet->write(getSessionUUID().toRfc4122());
+                nodeList->sendPacket(std::move(packet), *node);
```
birarda: this was reliable when it was a packet list - we may still want that to be the case?

assignment-client/src/audio/AudioMixer.cpp
```diff
@@ -602,16 +602,14 @@ void AudioMixer::handleNodeKilled(SharedNodePointer killedNode) {
...
-                    listenerClientData->removeHRTFForStream(sendingNode->getUUID(), streamID);
-                }
-            });
-        }
+        clientData->removeAgentAvatarAudioStream();
+        auto nodeList = DependencyManager::get<NodeList>();
+        nodeList->eachNode([sendingNode](const SharedNodePointer& node){
+            auto listenerClientData = dynamic_cast<AudioMixerClientData*>(node->getLinkedData());
+            if (listenerClientData) {
+                listenerClientData->removeHRTFForStream(sendingNode->getUUID(), QUuid());
```
birarda: `QUuid()` is the default, you can omit

assignment-client/src/Agent.cpp
davidkelly: Hmm yea I think that is worth doing...

hifi-gustavo:
Build console output for commit: 920cd90afae8e8826196e779d54cc681ca9fbe3f available at: https://gist.github.com/c3541006e4c4b80bd5b26687d310f459<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-920cd90.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8966-920cd90.dmg

assignment-client/src/audio/AudioMixerClientData.cpp
```diff
@@ -73,11 +73,19 @@ void AudioMixerClientData::removeHRTFForStream(const QUuid& nodeID, const QUuid&
     }
 }
 
+void AudioMixerClientData::removeAgentAvatarAudioStream() {
+    QWriteLocker writeLocker { &_streamsLock };
+    auto it = _audioStreams.find(QUuid());
+    if (it != _audioStreams.end()) {
+        _audioStreams.erase(it);
+    }
+    writeLocker.unlock();
```
birarda: Won't block for this but I think this is implicit based on the fact that writeLocker is about to go out of scope

howard-stearns: I agree. It's a no-op as is, but it would be cool to get rid of it. That's what I like about scoped QWriteLockers.

davidkelly: Yea -- true.  Some places we do this, so I copied assuming it was a style thing.  I sort of figured it was like being explicit, but I'm inclined to remove personally.

zfox23:
I'll run through the test plan on this PR again right now.

zfox23:
I ran through the test plan two times, both starting with a completely closed Sandbox. I fired up the server, waited for all agents to start, started summon.js, verified some numbers, stopped summon.js, then verified some more numbers. Everything happened as expected. :+1:

PR #8975, opened 2016-11-01
Add "Snapshot" text to snapshot cards in Address Bar Dialog

zfox23:
This text makes it easier for users to discern snapshot cards from concurrency cards.

### DIFF
--- a/interface/resources/qml/hifi/Card.qml
+++ b/interface/resources/qml/hifi/Card.qml
@@ -113,8 +113,8 @@ Rectangle {
     }
     FiraSansRegular {
         id: users;
-        visible: action === 'concurrency';
-        text: onlineUsers;
+        visible: true;
+        text: ((action === 'concurrency') ? (onlineUsers) : ('Snapshot'));
         size: textSize;
         color: hifi.colors.white;
         anchors {
--- a/interface/resources/qml/hifi/Card.qml
+++ b/interface/resources/qml/hifi/Card.qml
@@ -113,9 +113,8 @@ Rectangle {
     }
     FiraSansRegular {
         id: users;
-        visible: true;
-        text: ((action === 'concurrency') ? (onlineUsers) : ('Snapshot'));
-        size: textSize;
+        text: (action === 'concurrency') ? onlineUsers : 'snapshot';
+        size: (action === 'concurrency') ? textSize : textSizeSmall;
         color: hifi.colors.white;
         anchors {
             verticalCenter: usersImage.verticalCenter;


hifi-gustavo:

PhilipRosedale:
add to whitelist

interface/resources/qml/hifi/Card.qml
```diff
@@ -113,8 +113,8 @@ Rectangle {
     }
     FiraSansRegular {
         id: users;
-        visible: action === 'concurrency';
-        text: onlineUsers;
+        visible: true;
```
howard-stearns: not needed. that's the default.

interface/resources/qml/hifi/Card.qml
```diff
@@ -113,8 +113,8 @@ Rectangle {
...
+        text: ((action === 'concurrency') ? (onlineUsers) : ('Snapshot'));
```
howard-stearns: What, are you a Lisp programmer?
The extra parentheses are a little verbose. Should be sufficient to just have (action === 'concurrency') ? onlineUsers : 'snapshot'.

Also, my own opinion is that lowercase 'snapshot' is visually better, considering that some place names are lowercase. We don't want the word Snapshot to take more weight than the placename.  What do you think?

zfox23: 😆 I'm used to being verbose. I'll tone it down.

Your comment about visual weight makes sense. However, it feels to me like displaying the string "snapshot" is, in and of itself, visually commanding, especially when placed next to short place names. Thus, this doesn't feel like it should be a permanent change -- is the intention for it to be?

No matter what, "snapshot" feels better than "Snapshot" purely from a weight perspective, so I'll make the change. :+1:

hifi-gustavo:
Build console output for commit: c4b59eafbb3e8b28f987f0550e3461f2d759a2a8 available at: https://gist.github.com/ef0d002d9bb3913e85594f76d20e2972<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8975-c4b59ea.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8975-c4b59ea.dmg

hifi-gustavo:
Build console output for commit: b98c6cb50f3e9e074b5750667768217f8611e347 available at: https://gist.github.com/ef0d002d9bb3913e85594f76d20e2972<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8975-b98c6cb.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8975-b98c6cb.dmg

PR #8977, opened 2016-11-01
Use snapshot data from file

howard-stearns:
Allow movement (or domain change) between taking a snapshot and uploading it.
Fixes https://highfidelity.fogbugz.com/f/cases/2161/snapshots-recorded-for-wrong-place

### DIFF
--- a/interface/src/ui/SnapshotUploader.cpp
+++ b/interface/src/ui/SnapshotUploader.cpp
@@ -15,9 +15,13 @@
 #include "scripting/WindowScriptingInterface.h"
 #include "SnapshotUploader.h"
 
+SnapshotUploader::SnapshotUploader(QUrl inWorldLocation, QString pathname) :
+    _inWorldLocation(inWorldLocation),
+    _pathname(pathname) {
+}
+
 void SnapshotUploader::uploadSuccess(QNetworkReply& reply) {
     const QString STORY_UPLOAD_URL = "/api/v1/user_stories";
-    static SnapshotUploader uploader;
 
     // parse the reply for the thumbnail_url
     QByteArray contents = reply.readAll();
@@ -28,11 +32,8 @@ void SnapshotUploader::uploadSuccess(QNetworkReply& reply) {
         QString thumbnailUrl = dataObject.value("thumbnail_url").toString();
         QString imageUrl = dataObject.value("image_url").toString();
         auto addressManager = DependencyManager::get<AddressManager>();
-        QString placeName = addressManager->getPlaceName();
-        if (placeName.isEmpty()) {
-            placeName = addressManager->getHost();
-        }
-        QString currentPath = addressManager->currentPath(true);
+        QString placeName = _inWorldLocation.authority(); // We currently only upload shareable places, in which case this is just host.
+        QString currentPath = _inWorldLocation.path();
 
         // create json post data
         QJsonObject rootObject;
@@ -48,7 +49,7 @@ void SnapshotUploader::uploadSuccess(QNetworkReply& reply) {
         rootObject.insert("user_story", userStoryObject);
 
         auto accountManager = DependencyManager::get<AccountManager>();
-        JSONCallbackParameters callbackParams(&uploader, "createStorySuccess", &uploader, "createStoryFailure");
+        JSONCallbackParameters callbackParams(this, "createStorySuccess", this, "createStoryFailure");
 
         accountManager->sendRequest(STORY_UPLOAD_URL,
             AccountManagerAuth::Required,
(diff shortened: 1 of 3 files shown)


howard-stearns:
The data persisted in the file now uses shareable place info preferred to hostname.

howard-stearns:
test plan:
- Take a snapshot, and move to visibly different location before pressing the button to share it. Clicking on the address bar suggestion for that snapshot should take you to where it was taken, not where you were when you shared it.
- Repeat, but change domains entirely instead of just moving to a different location. Again, going to the suggestions should take you to the domain where it was taken, not the one you were in when you shared.

hifi-gustavo:
Build console output for commit: 4c34214e2d1f8b689fdbdb6110221ab276f9f8e3 available at: https://gist.github.com/a9dac014beba53431f86153559eb862c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8977-4c34214.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8977-4c34214.dmg

interface/src/ui/SnapshotUploader.cpp
```diff
@@ -63,13 +64,16 @@ void SnapshotUploader::uploadSuccess(QNetworkReply& reply) {
 }
```
Atlante45: Possible leak here, the else case should be deleting the uploader.
`else` clause is also badly formated.

interface/src/ui/SnapshotUploader.h
```diff
@@ -14,13 +14,19 @@
...
+public:
+    SnapshotUploader(QUrl inWorldLocation, QString pathname);
+public slots:
     void uploadSuccess(QNetworkReply& reply);
     void uploadFailure(QNetworkReply& reply);
     void createStorySuccess(QNetworkReply& reply);
     void createStoryFailure(QNetworkReply& reply);
+private:
+    QUrl _inWorldLocation;
+    QString _pathname;
```
Atlante45: Looks like this is never used.

zfox23:
Testing:
Test 1 passes. Taking a snapshot, then moving, then sharing the snapshot, then clicking the snapshot in the address bar takes me to the place where the snapshot was taken, not where I was when I clicked "share".

Test 2 passes. Same deal, except I moved to a different domain before sharing the snapshot.

I can confirm that the place name I am moved to when clicking the snapshot is, in fact, a place name instead of an ip:port.

**One thing I noticed:** When clicking on a snapshot in the address bar, the "Your location" field flashes to a string prepended with "hifi://" before the address bar window fades away. This doesn't feel like an optimal UX, but it's a small issue. This behavior also occurs in the latest dev-download.

interface/src/ui/SnapshotUploader.cpp
```diff
@@ -15,9 +15,13 @@
 #include "scripting/WindowScriptingInterface.h"
 #include "SnapshotUploader.h"
 
+SnapshotUploader::SnapshotUploader(QUrl inWorldLocation, QString pathname) :
+    _inWorldLocation(inWorldLocation),
```
howard-stearns: Good question, but I think it doesn't provide any additional user or debugging benefit.
First, it's already going to be a QUrl, although by static type checking alone, it could be an empty one, or a non-hifi-scheme one. Maybe those will be meaningful at some point in the future, in which case we'd have to go back and fix the validation.
Second, the path in is currently only though parseSnapshotData, which is parsing the URL in StrictMode.
Third, the result passes directly through to our server, and its up to it act on the data.  What would the user or the logs do differently if it were wrong, and for what definition of wrong?

(Admittedly, if you follow where uploadFailure and createStoryFailure are used, failures aren't presented to the user for action, which isn't great. That happens in snapshot.js.)

interface/src/ui/SnapshotUploader.cpp
howard-stearns: quite right

interface/src/ui/SnapshotUploader.h
howard-stearns: correct. I'd like for uploadFailure/createStoryFailure to pass on better error information (including the original image _pathname), but changing them (and their use in snapshot.js) was not strictly necessary for fixing the bug that this PR is for. So for now, I'm including the info only as far through the chain as I'm changing.

Atlante45:
Additional note:
I'm not particularly fan of objects that manage their lifetime themselves, as it's prone to leaks.
I think it'd be better if you could find an object capable of managing their lifetime and maybe store the pointers in a `std::unique_ptr` to be extra safe.

Won't block the PR for that though.

interface/src/ui/SnapshotUploader.cpp
zfox23: Got it, cool. Thanks for the explanation.

howard-stearns:
@Atlante45 - noted.
@zfox23 - yeah, I'm not sure what behavior was intended for updating the address bar like that. Arguable either way, but I agree that it's disconcerting to have it change as it is going away. But as you point out, that's outside this PR.

hifi-gustavo:
Build console output for commit: b6e440f8919b93bd63e6d5b13ed89cbf0d928d2c available at: https://gist.github.com/a9dac014beba53431f86153559eb862c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8977-b6e440f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR8977-b6e440f.dmg

PR #9009, opened 2016-11-05
fix numeric keyboard for hmd Window.prompt

howard-stearns:
fixes https://highfidelity.fogbugz.com/f/cases/2081/no-numeric-keys-in-virtual-keyboard-for-prompt

### DIFF
--- a/interface/resources/qml/dialogs/QueryDialog.qml
+++ b/interface/resources/qml/dialogs/QueryDialog.qml
@@ -118,6 +118,9 @@ ModalWindow {
             }
         }
 
+        property alias keyboardOverride: root.keyboardOverride
+        property alias keyboardRaised: root.keyboardRaised
+        property alias punctuationMode: root.punctuationMode
         Keyboard {
             id: keyboard
             raised: keyboardEnabled && keyboardRaised


howard-stearns:
test plan:
While in HMD, execute Window.prompt("how many?", 17 (e.g., in console). You should be able to enter and accept a number with the HMD virtual keyboard, after pressing the "123" key.

howard-stearns:
There's a lot of cargo-cult code here, in which Keyboard.qml handles toggling of the numeric keyboard by changing state in its parent. Ugh! In this case, the parent didn't define the property.

I'd love for all this code to be changed to OOP, but that's out of scope for this bug, which had been languishing.

hifi-gustavo:
Build console output for commit: 84eb1721e7272a230fbe3ececd55740be4cffe01 available at: https://gist.github.com/dcccd291ce8d494375884ffad27fee6f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9009-84eb172.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9009-84eb172.dmg

Nex-Pro:
### PASS

I ran Window.prompt("how many?", 17); in the console and got a prompt with keyboard. Clicking 123 switched to numeric. Result was returned to console correctly.
In desktop mode no keyboard was displayed.

Q: Shouldn't this actually show the keyboard already toggled to numeric ?

howard-stearns:
It is not expected for the test example to be initially in numeric mode. It would be nice if the Window.prompt API allowed specification of which keyboard should be initially shown, or to base the answer on the type of the default, but it doesn't.

PR #9017, opened 2016-11-06
Procedural vertex shader support

hyperlogic:
Added "vertexShaderUrl" field to the "ProceduralEntity" userdata block.
This expects a fragment of glsl vertex shader code that contains uniforms and the following function:

    void getProceduralVertex(inout vec4 position, inout vec4 color, inout vec2 texCoord0, inout vec3 normal)

This function can manipulate the inout parameters in anyway it sees fit.  It also has access to the glnoise functions in ProceduralCommon.slf.

For example: the following shader will morph each vertex position and color:

    uniform float amp;
    uniform float freq;
    uniform float timeScale;

    // position and normal are in model coordinates
    // color is SRGB
    void getProceduralVertex(inout vec4 position, inout vec4 color, inout vec2 texCoord0, inout vec3 normal) {
        vec4 morphPosition = vec4(position.x + amp * sin((freq * position.x) + (timeScale * iGlobalTime)),
                                  position.y + amp * sin((freq * position.y) + (timeScale * iGlobalTime)),
                                  position.z + amp * sin((freq * position.z) + (timeScale * iGlobalTime)), 1);

        vec4 morphColor = vec4(0.5 * sin((freq * position.x) + (timeScale * iGlobalTime)) + 0.5,
                               0.5 * sin((freq * position.y) + (timeScale * iGlobalTime)) + 0.5,
                               0.5 * sin((freq * position.z) + (timeScale * iGlobalTime)) + 0.5, 1);

        position = morphPosition;
        color = morphColor;
    }


hyperlogic:
QA Test Plan:
    * Create a sphere entity.
    * Enter edit mode
    * copy the following into the sphere's userData field.

    {
      "ProceduralEntity": {
        "version": 2,
        "shaderUrl": "https://s3.amazonaws.com/hifi-public/tony/shaders/matte.slf",
        "vertexShaderUrl": "https://s3.amazonaws.com/hifi-public/tony/shaders/wiggle.slv",
        "uniforms": {
          "amp": 0.05,
          "freq": 10,
          "timeScale": 3
        }
      }
    }

   * You should see the sphere begin to wiggle and have a morphing multi colored surface.
   * You should also be able to change the shape type to any other shape, and the object's vertices will continue to wiggle.

hifi-gustavo:
Build console output for commit: 68bb072c87362c548a22e329aeeab67e8e53f6f0 available at: https://gist.github.com/9345a0b66c9ec832a59c804463656d19<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9017-68bb072.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9017-68bb072.dmg

Triplelexx:
I tested the change with all shape types on my old AMD card, worked fine for me.
No performance hit was observed using the example, not that I was expecting any.

I did notice that on occasion after pasting the userData snippet it would have no effect.
Clearing the data and pasting again always gets it working on the second attempt.

humbletim:
I tested using image channels like `iChannel0` as displacement maps in the vertex shader -- and it works great!!!

Question: Should it be possible to do seamless UV calculations in the vertex shader now too?

Because when I try that it mostly works, except for this stubborn seam:
proceduraluvseam

(above in the vertex shader I was adding an extra `iGlobalTime` offset to the calculated `texCoord0.x` -- and expected the seam to rotate too if it's the UV calcs going wrong)

Same thing works fine if applied in the fragment shader:
```glsl
vec2 UV = vec2( clamp(vec2((((atan(_position.x, _position.z) / M_PI) + 1.0) / 2.0), (0.5-(asin(_position.y)/M_PI))),0.0,1.0));
```

libraries/procedural/src/procedural/ProceduralCommon.slf
```diff
@@ -19,9 +19,6 @@
 //               https://github.com/ashima/webgl-noise
 //
 
-<@include gpu/Transform.slh@>
```
samcake: THis actually must stay there in the procedural commons
so a procedural shader can query the transform information

libraries/procedural/src/procedural/ProceduralSkybox.cpp
```diff
@@ -18,9 +18,7 @@
 #include <model/skybox_vert.h>
 #include <model/skybox_frag.h>
 
-ProceduralSkybox::ProceduralSkybox() : model::Skybox() {
-    _procedural._vertexSource = skybox_vert;
-    _procedural._fragmentSource = skybox_frag;
+ProceduralSkybox::ProceduralSkybox() : model::Skybox(), _procedural(skybox_vert, skybox_frag) {
```
samcake: Can we avoid dropping the vertex shader for the Skybox ?
I don;t really see a good use case for it and this may break assumptions about ht einterface of the skybox as a background render item

libraries/procedural/src/procedural/ProceduralCommon.slf
samcake: After discussing with Tony...
We can avoid the repetition of the block if used in a vertex shader by protecting that scribe include  and declaration (next line) within a #ifdef GPU_VERTEX_SHADER / #endif

libraries/render-utils/src/simple.slf
```diff
@@ -14,6 +14,8 @@
 
 <@include DeferredBufferWrite.slh@>
 <@include model/Material.slh@>
+<@include gpu/Transform.slh@>
```
samcake: you can remove that and rely on it beeing in the "proceduralDefault.slh"

this applies to line 17 and line 18

libraries/procedural/src/procedural/ProceduralSkybox.cpp
hyperlogic: This is the same vertex shader that was previously used.  Because this shader does not have any off the magic //PROCEDURAL_FOO comments, it is used as is with no changes.

hyperlogic:
@samcake I made the changes you requested.  ProceduralCommon.slf & simple.slf are unchanged from master.  The new simple.slv now calls declareStandardObjectTransform() explicitly, there was no need to use the GPU_VERTEX_SHADER ifdefs

hifi-gustavo:
Build console output for commit: 6aba9e350938cd069adc72972295132c387fb1fc available at: https://gist.github.com/9345a0b66c9ec832a59c804463656d19<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9017-6aba9e3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9017-6aba9e3.dmg

hyperlogic:
@humbletim The gif you posted isn't a bug in the vertex shader, but it's due to this particular sphere geometry and the nature of UV coordinates and fragment interpolation.  The most appropriate way to do texture scrolling for this example is in the fragment shader.

OmegaHeron:
Would it be possible to get this merged to current code so I can connect to external domains?  Getting a proto mismatch on everything out there - with exception of this PR sandbox.

hifi-gustavo:
Build console output for commit: 7052978906648a42f22c49fe4d1b948a9a3d9dc6 available at: https://gist.github.com/9345a0b66c9ec832a59c804463656d19<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9017-7052978.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9017-7052978.dmg

OmegaHeron:
QA Fail

First thing I notice is, yes, it works and is cool!  But...  I placed a sphere - added user data fragment.  All subsequent primitives placed show shader effect.  Tried restarting interface - primitives placed *without* user data specified continue to show effect from first object that has data specified.  When I delete any entity containing shader code the others return to being "normal".

hyperlogic:
I'll re open this in the future when I have the time to diagnose what the problem is.

PR #9024, opened 2016-11-07
Fix userData in edit.js sometimes showing wrong value when locked

huffman:
To test:

 1. Create an entity 1 with no userData on it
 2. Create another entity 2 with JSON userData
 3. Select entity 2
 4. Select entity 1, look at the user properties, then lock it
 5. The userData should still be blank and should not show entity 2's userData
 6. Unlock entity 1
 7. Set the userData to a non-JSON value, like "asdf"
 8. Select entity 2
 9. Select entity 1, look at the user properties, then lock it
 10. The userData should still be the non-JSON value you set it to
 11. Lastly, please do general smoke testing to make sure userData is showing the correct value for a few different entities

### DIFF
--- a/scripts/system/html/js/entityProperties.js
+++ b/scripts/system/html/js/entityProperties.js
@@ -445,13 +445,11 @@ function hideUserDataTextArea() {
 };
 
 function showStaticUserData() {
-    $('#static-userdata').show();
-    $('#static-userdata').css('height', $('#userdata-editor').height())
     if (editor !== null) {
+        $('#static-userdata').show();
+        $('#static-userdata').css('height', $('#userdata-editor').height())
         $('#static-userdata').text(editor.getText());
     }
-
-
 };
 
 function removeStaticUserData() {


hifi-gustavo:
Build console output for commit: c4f35bf7bf8dbc62bdf3de45e88852a2a4e82272 available at: https://gist.github.com/4f58d85a8fd8b45102d2b788bed4f196<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9024-c4f35bf.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9024-c4f35bf.dmg

OmegaHeron:
Windows 10

Following this precisely - it works as advertised.

But, you can still end up with seeing wrong user data - if both entities are locked you do.  Perhaps a more elegant solution is to simply not show user data for a locked element.

The few entities I could find that use user data seemed to still work correctly.

PR #9029, opened 2016-11-08
Disable Sandbox auto-update popup when the application is running from Steam

Atlante45:
Disable Sandbox auto-update popup when the application is running from Steam


cozza13:
Ran through the test and it works

PR #9027, opened 2016-11-07
Fix crashes in OpenGLVersionChecker 

samcake:
Add more safe checks on the values returned from the opengl context creation to avoid crashing on non capable hardwares

THanks to @davidkelly for pointing out the issue

PR #9037, opened 2016-11-09
OpenGLVersionChecker passing when it shouldn't

davidkelly:
In bug 2084, it appears we are crashing when attempting to create a gl DC.  But, that is _after_ we checked the version and it should be ok.  

Looks like we shifted the minimum version too far, and ended up with a minimum major version of 0.  Also, where we try to create a gl DC we end up raising an exception, so now we log the error(s) first.  Chances
are that is a waste of time as it probably will not be flushed to disk but lets verify this is the issue, then worry about that.

No need to look to hard at this until we verify tomorrow.  Want to get this in here so @samcake can take a look, potentially find a way to reproduce 2084


hifi-gustavo:
Build console output for commit: e811224a71c5d93d016798ea96e84e00ea272cd0 available at: https://gist.github.com/2c137a74aa1ca6d1038acf19890f61ac<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9037-e811224.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9037-e811224.dmg

davidkelly:
Since we had another PR out there touching the same code, @samcake popped the corrected bit shifting from this PR in there (https://github.com/highfidelity/hifi/pull/9027).  Closing this.

PR #9043, opened 2016-11-10
add support for whitelist entity scripts

ZappoMan:
allows domain operator to limit entity scripts to only a specific white list of sources

### DIFF
--- a/assignment-client/src/entities/EntityServer.cpp
+++ b/assignment-client/src/entities/EntityServer.cpp
@@ -278,6 +278,13 @@ void EntityServer::readAdditionalConfiguration(const QJsonObject& settingsSectio
 
     tree->setWantEditLogging(wantEditLogging);
     tree->setWantTerseEditLogging(wantTerseEditLogging);
+
+    QString entityScriptSourceWhitelist;
+    if (readOptionString("entityScriptSourceWhitelist", settingsSectionObject, entityScriptSourceWhitelist)) {
+        tree->setEntityScriptSourceWhitelist(entityScriptSourceWhitelist);
+    } else {
+        tree->setEntityScriptSourceWhitelist("");
+    }
 }
 
 void EntityServer::nodeAdded(SharedNodePointer node) {
--- a/domain-server/resources/describe-settings.json
+++ b/domain-server/resources/describe-settings.json
@@ -1076,6 +1076,14 @@
           "advanced": true
         },
         {
+          "name": "entityScriptSourceWhitelist",
+          "label": "Entity Scripts Allowed from:",
+          "help": "The domains that entity scripts are allowed from. A comma separated list of domains that entity scripts are allowed from, if someone attempts to create and entity or edit an entity to have a different domain, it will be rejected. If left blank, any domain is allowed.",
+          "placeholder": "",
+          "default": "",
+          "advanced": true
+        },
+        {
           "name": "persistFilePath",
           "label": "Entities File Path",
           "help": "The path to the file entities are stored in.<br/>If this path is relative it will be relative to the application data directory.<br/>The filename must end in .json.gz.",
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -63,6 +63,11 @@ EntityTree::~EntityTree() {
     eraseAllOctreeElements(false);
 }
 
+void EntityTree::setEntityScriptSourceWhitelist(const QString& entityScriptSourceWhitelist) { 
+    _entityScriptSourceWhitelist = entityScriptSourceWhitelist.split(',');
+}
+
+
 void EntityTree::createRootElement() {
     _rootElement = createNewElement();
 }
@@ -925,6 +930,9 @@ int EntityTree::processEditPacketData(ReceivedMessage& message, const unsigned c
             quint64 startCreate = 0, endCreate = 0;
             quint64 startLogging = 0, endLogging = 0;
 
+            const quint64 LAST_EDITED_SERVERSIDE_BUMP = 1; // usec
+            bool suppressDisallowedScript = false;
+
             _totalEditMessages++;
 
             EntityItemID entityItemID;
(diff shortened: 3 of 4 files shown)


ZappoMan:
Test Plan:
- Smoke test
- Run this build, and go to Sandbox > Settings, Advanced, Entity Server Settings
- Add a domain name to the `Entity Scripts Allowed from:` setting.
- Try creating an entity with a script that's not in that list.
```
Entities.addEntity(
    {
        type: "Box", 
        position: MyAvatar.position, 
        visible: true,
        collisionless:true,
        name: "evil entity",
        dimensions: { x: 1, y: 1, z: 1},
        color: { red: 0, green: 255, blue: 0},
        script: "https://evilScripts.com/evilScript.js"
    });
```
- verify that the entity is not added (it will flash momentarily on the client but end up deleted)
- create a benign entity:
```
var id = Entities.addEntity(
    {
        type: "Box", 
        position: MyAvatar.position, 
        visible: true,
        collisionless:true,
        name: "benign entity",
        dimensions: { x: 1, y: 1, z: 1},
        color: { red: 0, green: 255, blue: 0}
    });
```
- verify that you can't change a script to be a script not on the white list:
```
Entities.editEntity(id,
    {
        type: "Box", 
        position: MyAvatar.position, 
        visible: true,
        collisionless:true,
        name: "benign entity",
        dimensions: { x: 1, y: 1, z: 1},
        color: { red: 0, green: 0, blue: 255},
        script: "https://evilScripts.com/evilScript.js"
    });
```
- You should see the box change blue, but the script will not be changed.

hifi-gustavo:
Build console output for commit: eec24eee504c4362c3596a83811b6501279f3529 available at: https://gist.github.com/fc49e4e6e8d9a6938732ab601b896042<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9043-eec24ee.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9043-eec24ee.dmg

hifi-gustavo:
Build console output for commit: 56a373acb1c16ed0b7b8b1b0521265466923aea8 available at: https://gist.github.com/fc49e4e6e8d9a6938732ab601b896042<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9043-56a373a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9043-56a373a.dmg

hifi-gustavo:
Build console output for commit: 8758e4a9ba0229c80285ab0416c1a84cd165dec7 available at: https://gist.github.com/fc49e4e6e8d9a6938732ab601b896042<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9043-8758e4a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9043-8758e4a.dmg

Nex-Pro:
### PASS

Please read notes:

I also ran the code in the console with a script from a permitted dropbox link. It set the script url appropriately.

With a single incorrect link as permitted in the server settings the previously working dropbox link was now rejected.

This is ok but domain admins will probably need a tool to trawl through every entity to check that only scripts permitted domains remain after settings are changed.  

Currently any scripts that exist from non-whitelisted domains persist.

huffman:
@ZappoMan I was late to the party, couple thoughts - 

 1. This seems like it will block marketplace items with scripts. We might want an easy way to include those prefixes
 1. I don't think this will retroactively disable or delete entities with "banned" scripts. It would be nice to improve this with that functionality at some point (if we let entity be created but don't let the script run on the client, that would be one way of approaching it).
 1. If you try bringing in something with multiple entities you can end up in a situation where some of the entities are created but some aren't. As a user this would be very confusing. I think we need to seriously think about how we are presenting this to the user. Or, more simply, I bring in an entity but it doesn't show up. This just makes it look like something is broken.

thoys:
@ZappoMan  What will this do to direct scripts (directly insert a script into the entity script field `(function() { this.preload = function(entityID) { print('This is the script for entity ' + entityID); })`, or local script files (for debugging purposes)?

ZappoMan:
@huffman - re your questions...
1) this gives the domain admin the control... it's off by default... 
2) correct - it's not intended to do that, in fact the way it's designed and admin could create content with scripts and then not allow anyone else to add or edit scripts
3) that's true - for now we have to live with that.

@thoys - you can add any string prefix to the allowed list, `file:` or `(function(` would both be allowed if you want to allow that in your domain.

huffman:
> 1) this gives the domain admin the control... it's off by default... 

What I mean is, I don't think it's obvious how to allow marketplace scripts (I don't know the prefix), even if you want to include those. Maybe we could add it to the wiki.

huffman:
> 2) correct - it's not intended to do that, in fact the way it's designed and admin could create content with scripts and then not allow anyone else to add or edit scripts

I would agree this is a good stop gap for now, but I think the better long term solution for that specific case is to have a permission to not allow or not allow users to set the script property. Maybe that is the intent.

PhilipRosedale:
Can we default to 'high fidelity marketplace scripts' and 'scripts i put on my own ATP server'?

Menithal:
- While off by default, and solving issues for possible "blackhat scripts" it may make accessibility to scripting more unclear for newcomers (which I am trying to solve with an [framework library](https://forums.highfidelity.com/t/making-scripting-easier-entities-simplified-wip/11748)). It also limits growth for alternative marketplaces as its extra configuring for a domain. It also may give the wrong impression to some, giving an illusion to a closed system. There is no UI event for end users putting in the script sayin the script is not whitelisted.
- This still leaves marketplace scripts open. Regardless of the whitelist/blacklist: anyone can put up a temporary script into the marketplace on an alt, even if it is just "hosted there" without going into-review or even "submitted" state. The hosted assets on the marketplace are still available to be placed on and run on entities and still be in the whitelist.
- It also means any admin scripts one may develop can still be put into the marketplace can be used by anyone who has temp rez rights.

At best the solution is a temporary work around.

However I'd be impartial to just setting it for objects with a lifetime (except I tend to use this quite often even If I have permissions for something). And not sure why is it necessary to have it done Server-side: aren't entities scripts after all loaded and run on the client side? why Not just ignore the script if your client doesn't like it?

I'd prefer to for entities to have ownership, and depending on your trust level on the user, the script can run if regardless of state. But I know this may add some overhead. Instead, this should always apply on temporary objects, but not on permanent object (where there should be a bigger level of trust)

And to fix the marketplace, I suggest non-submitted / in-review scripts to have a different path to default that could be turned on with "allow marketplace dev scripts"

thoys:
I would like to be able to set it in the server. Anyway the marketplace scripts could be solved by just showing unpublished items to the creator of the item, this also prevents from people using the marketplace for huge storage. I am sure you can hook a check up.

Issue #8667, opened 2016-09-27
Edit.js Undo parent entity with child entity does not restore the chilkd entity


Richardus:
If you have a set of entities, 1 parent and one child. and you delete the parent entity. both parenmt and child entity get deleted. If you do undo after that, Only the parent entity get restored.

PR #9057, opened 2016-11-11
fixed children of parent object not being added back when performing an undo

druiz17:
Fixed the issue of when the parent objects are deleted its children are deleted as well but not removed from the entity list. Also when removing a parent object with children, I made sure that the children are added back when performing an undo

link to issue - https://github.com/highfidelity/hifi/issues/8667

### DIFF
--- a/libraries/entities/src/EntityScriptingInterface.cpp
+++ b/libraries/entities/src/EntityScriptingInterface.cpp
@@ -1318,6 +1318,27 @@ QVector<QUuid> EntityScriptingInterface::getChildrenIDs(const QUuid& parentID) {
     return result;
 }
 
+bool EntityScriptingInterface::isChildOfParent(QUuid childID, QUuid parentID) {
+    bool isChild = false;
+
+    if(!_entityTree) {
+        return isChild;
+    }
+
+    _entityTree->withReadLock([&] {
+        EntityItemPointer parent = _entityTree->findEntityByEntityItemID(parentID);
+        parent->forEachDescendant([&](SpatiallyNestablePointer descendant) {
+            if(descendant->getID() == childID) {
+                isChild = true;
+                return; 
+            }
+        });
+    });
+
+    qDebug() << "Is Child of parent? : " << isChild;
+    return isChild;
+}
+
 QVector<QUuid> EntityScriptingInterface::getChildrenIDsOfJoint(const QUuid& parentID, int jointIndex) {
     QVector<QUuid> result;
     if (!_entityTree) {
--- a/libraries/entities/src/EntityScriptingInterface.cpp
+++ b/libraries/entities/src/EntityScriptingInterface.cpp
@@ -1321,7 +1321,7 @@ QVector<QUuid> EntityScriptingInterface::getChildrenIDs(const QUuid& parentID) {
 bool EntityScriptingInterface::isChildOfParent(QUuid childID, QUuid parentID) {
     bool isChild = false;
 
-    if(!_entityTree) {
+    if (!_entityTree) {
         return isChild;
     }
 
--- a/libraries/entities/src/EntityScriptingInterface.cpp
+++ b/libraries/entities/src/EntityScriptingInterface.cpp
@@ -1334,8 +1334,7 @@ bool EntityScriptingInterface::isChildOfParent(QUuid childID, QUuid parentID) {
             }
         });
     });
-
-    qDebug() << "Is Child of parent? : " << isChild;
+    
     return isChild;
 }
 
--- a/libraries/entities/src/EntityScriptingInterface.h
+++ b/libraries/entities/src/EntityScriptingInterface.h
@@ -255,7 +255,7 @@ public slots:
     Q_INVOKABLE QStringList getJointNames(const QUuid& entityID);
     Q_INVOKABLE QVector<QUuid> getChildrenIDs(const QUuid& parentID);
     Q_INVOKABLE QVector<QUuid> getChildrenIDsOfJoint(const QUuid& parentID, int jointIndex);
-    Q_INVOKABLE bool isChildOfParent(QUuid childID, QUuid parnetID);
+    Q_INVOKABLE bool isChildOfParent(QUuid childID, QUuid parentID);
 
     Q_INVOKABLE QUuid getKeyboardFocusEntity() const;
     Q_INVOKABLE void setKeyboardFocusEntity(QUuid id);
--- a/libraries/entities/src/EntityScriptingInterface.h
+++ b/libraries/entities/src/EntityScriptingInterface.h
@@ -255,6 +255,7 @@ public slots:
     Q_INVOKABLE QStringList getJointNames(const QUuid& entityID);
     Q_INVOKABLE QVector<QUuid> getChildrenIDs(const QUuid& parentID);
     Q_INVOKABLE QVector<QUuid> getChildrenIDsOfJoint(const QUuid& parentID, int jointIndex);
+    Q_INVOKABLE bool isChildOfParent(QUuid childID, QUuid parnetID);
 
     Q_INVOKABLE QUuid getKeyboardFocusEntity() const;
     Q_INVOKABLE void setKeyboardFocusEntity(QUuid id);
--- a/scripts/system/edit.js
+++ b/scripts/system/edit.js
@@ -1027,19 +1027,76 @@ function selectAllEtitiesInCurrentSelectionBox(keepIfTouching) {
     }
 }
 
+
+function sortDeleteSelected  (selected) {
+    var tempArray = selected;
+    var begin = 0;
+    while (begin < tempArray.length) {
+        var elementRemoved = false;
+        var next = begin + 1;
+        while (next < tempArray.length) {
+            var beginID = tempArray[begin];
+            var nextID = tempArray[next];
+            var beginProperties = Entities.getEntityProperties(beginID);
+            var nextProperties = Entities.getEntityProperties(nextID);
+            var nextParentID = nextProperties.parentID;
+            var beginParentID = beginProperties.parentID;
+
+            if (beginID == nextParentID) {
+                var temp = beginID;
+                tempArray[begin] = nextID
+                tempArray[next] = beginID
+                tempArray.splice(begin, 1);
+                elementRemoved = true;
+                break;
+            } else if (nextID == beginParentID) {
+                tempArray.splice(begin, 1);
+                elementRemoved = true;
+                break;
+            }
+            next++;
+        }
+        if (!elementRemoved) {
+            begin++;
+        }
+    }
+    return tempArray;
+}
+
+function recursiveDelete(entities, list) {
+    var entitiesLength = entities.length;
+    for (var i = 0; i < entitiesLength; i++) {
+        var entityID = entities[i];
+        var children = Entities.getChildrenIDs(entityID);
+        var childList = []
+        recursiveDelete(children, childList);
+        var initialProperties = Entities.getEntityProperties(entityID);
+        list.push({
+            entityID: entityID,
+            properties: initialProperties,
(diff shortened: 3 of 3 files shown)


huffman:
If you delete a parent via `Entities.deleteEntity` will it delete the children? I think maybe it doesn't and the children end up being parentless and possible moved to the origin (or, relative to it)

druiz17:
Yes, but the issue if you try to undo the deletion the children will not be added back

hifi-gustavo:
Build console output for commit: be08da273189b44f54a07c3b938a160ece6d15c6 available at: https://gist.github.com/321ac95b8ecbe89b0dcc6c19b2f280f8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9057-be08da2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9057-be08da2.dmg

druiz17:
To Test
1. In the interface enter edit mode and add multiple objects
2. On some of the objects set the parent field in properties to another objects id
3. delete the parent object and the children should be delete and also removed removed from entity list
 ( on master the child object are not removed from the list)
4. When performing an undo of the deleted parent its children should be added back as well

huffman:
Ok, I see now - I didn't look at the code previously. So now it's recursively deleting the children and then doing the proper thing on undo.

There are some gotchas here because you aren't guaranteed (afaik) to have all of the children of a particular entity in your local tree, which means there could still be children out there than you don't know about. This is difficult to design around (maybe we need special behavior for editing, with cooperation from the server), but this looks like an improvement over the current situation.

OmegaHeron:
Testing Windows 10
QA Pass

Master - created/linked 3 cubes.  Deleted and used undo to restore -- got one cube back.
This PR - same as above - undo returned full linked set.
Looked at models list to insure no dangling references left from link set deletes.  (Looked as in parsing models.json).

General smoke tests pass.  It would have been nice to check this against some other content where I know there's heavy use of linked sets, but, this PR is on old proto and everything with said content is on new...  As far as I can tell this should have no negative impact on existing linked content sets.

scripts/system/edit.js
```diff
@@ -1022,19 +1022,76 @@ function selectAllEtitiesInCurrentSelectionBox(keepIfTouching) {
     }
 }
 
+
+function sortDeleteSelected  (selected) {
+    var tempArray = selected;
```
huffman: If you want to get a copy of `selected` you will need to use `selected.slice()` here

scripts/system/edit.js
```diff
@@ -1248,6 +1305,16 @@ Controller.keyReleaseEvent.connect(function (event) {
     }
 });
 
+function recursiveAdd(newParnetID, parentData) {
```
huffman: Minor, but might be worth fixing this typo `newParnetID` -> `newParentID`

scripts/system/edit.js
```diff
@@ -1022,19 +1022,76 @@ function selectAllEtitiesInCurrentSelectionBox(keepIfTouching) {
...
+                elementRemoved = true;
+                break;
+            }
+            next++;
+        }
+        if (!elementRemoved) {
+            begin++;
+        }
+    }
+    return tempArray;
```
huffman: Will this handle a hierarchy of
```
A
|
B
|
C
```
Where B is a child of A and C is a child of B? In particular if A and C are selected, but not B. I think what we actually want to check is that Y is a descendant of X, whether it's direct or not. Thoughts?

druiz17: As it is, it would not handle the case above. I also agree that we should check if Y is a descendant of X but that might require a recursive call in for both checks, and I don't know if that a good idea or not since it would be in a double for loop. What's your thought on that?

huffman: IMO, I'd say worry about getting it right first, do it the simple way, then look at the performance. Deletes are rare, and scripts run on a separate thread, so it shouldn't have a huge effect on other parts of the system*.

I'd start by writing a function that checks if X is a descendant of Y. If performance is really bad, and we can identify that the `Entities.getEntityProperties` calls are the culprit, the solution might be to implement that on the c++ side. Or maybe there will be a completely different performance issue that we solve in a different way.

*It won't necessarily have 0 effect because the call might acquire locks that block other threads.

hifi-gustavo:
Build console output for commit: 17c9f749c405215ec47494cba84526fabe351275 available at: https://gist.github.com/321ac95b8ecbe89b0dcc6c19b2f280f8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9057-17c9f74.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9057-17c9f74.dmg

hifi-gustavo:
Build console output for commit: 713b5ac54534f721b58ccc398c049286d23f3e20 available at: https://gist.github.com/321ac95b8ecbe89b0dcc6c19b2f280f8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9057-713b5ac.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9057-713b5ac.dmg

druiz17:
After new changes the new test is 
1. In the interface enter edit mode and add multiple objects
2. On some of the objects set the parent field in properties to another objects id
3. delete the parent object and the children should be delete and also removed removed from entity list (also test if this works with selecting the grandchild first then the parent)
( on master the child object are not removed from the list)
4. When performing an undo of the deleted parent its children should be added back as well
5.

OmegaHeron:
Testing Windows 10

Test setup as follows;
5 cube entities 
Cube 1 Parent to cubes 2 and 3.
Cube 2 Parent to cube 4
Cube 4 Parent to cube 5

In edit mode;

Moving cube 1 results in 2, 3, 4 and 5 moving with.
Moving cube 2 results in 2, 4 and 5 moving with - 1 and 3 remain stationary.
Moving cube 3 results in 3 moving - 1, 2, 4 and 5 remain stationary.
Moving cube 4 results in 4 and 5 moving - 1, 2 and 3 remain stationary.
Moving cube 5 results in 5 moving - 1, 2, 3 and 4 remain stationary.

^^ Seems correct.

Deleting cube 1 results in all cubes deleted.
Undo delete results only in cube 1 being restored (Parent of all -- cube 1).

I think this is a fail at this stage.

OmegaHeron:
Actually - a simple 2 cube entity set with cube 2 child of cube 1 fails to restore second cube on undo delete... so - something has gone wrong somewhere.  *Unless there is something on server side that impacts this.* Does this PR Interface require this PR sandbox to work correctly or should it work as intended with any compatible sandbox version?

druiz17:
I believe that any PR sandbox should work

druiz17:
To Test after fix

1. In the interface enter edit mode and add multiple objects
2. On some of the objects set the parent field in properties to another objects id
3. delete the parent object and the children should be delete and also removed from entity list (also test if this works with selecting the grandchild first then the parent)
( on master the child object are not removed from the list)
4.When performing an undo of the deleted parent or child its children should be added back as well

hifi-gustavo:
Build console output for commit: 063aced6f5cc1a9498613ddab7a698c19f655132 available at: https://gist.github.com/321ac95b8ecbe89b0dcc6c19b2f280f8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9057-063aced.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9057-063aced.dmg

OmegaHeron:
Will re-test shortly.

OmegaHeron:
Testing Windows 10

QA Pass

Tested suggested linkset configuration + a much more complex one with children, grandchildren and great grandchildren.  Tried various combinations of deleting branches of family - entire family and undo operations.  In all cases undo worked as expected as did manipulating linkset from any branch of tree.

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -1318,6 +1318,26 @@ QVector<QUuid> EntityScriptingInterface::getChildrenIDs(const QUuid& parentID) {
     return result;
 }
 
+bool EntityScriptingInterface::isChildOfParent(QUuid childID, QUuid parentID) {
+    bool isChild = false;
+
+    if(!_entityTree) {
```
huffman: Coding standard: space after `if`

libraries/entities/src/EntityScriptingInterface.h
```diff
@@ -255,6 +255,7 @@ public slots:
     Q_INVOKABLE QStringList getJointNames(const QUuid& entityID);
     Q_INVOKABLE QVector<QUuid> getChildrenIDs(const QUuid& parentID);
     Q_INVOKABLE QVector<QUuid> getChildrenIDsOfJoint(const QUuid& parentID, int jointIndex);
+    Q_INVOKABLE bool isChildOfParent(QUuid childID, QUuid parnetID);
```
huffman: Minor, but typo: `parnet` -> `parent`

scripts/system/edit.js
```diff
@@ -1027,19 +1027,72 @@ function selectAllEtitiesInCurrentSelectionBox(keepIfTouching) {
     }
 }
 
+
+function sortDeleteSelected  (selected) {
```
huffman: Coding standard: no space after function name

scripts/system/edit.js
```diff
@@ -1027,19 +1027,72 @@ function selectAllEtitiesInCurrentSelectionBox(keepIfTouching) {
...
+    var begin = 0;
+    while (begin < array.length) {
+        var elementRemoved = false;
+        var next = begin + 1;
+        while (next < array.length) {
+            var beginID = array[begin];
+            var nextID = array[next];
+
+            if (Entities.isChildOfParent(beginID, nextID)) {
+                var temp = beginID;
```
huffman: Is `temp` needed? It looks like it isn't used

scripts/system/edit.js
```diff
@@ -1027,19 +1027,72 @@ function selectAllEtitiesInCurrentSelectionBox(keepIfTouching) {
...
+            next++;
+        }
+        if (!elementRemoved) {
+            begin++;
+        }
+    }
+    return array;
+}
+
+function recursiveDelete(entities, list) {
```
huffman: Not a blocker, but I would recommend using more descriptive argument names. `list` is really vague. Of course naming things is one of the hardest problems in Computer Science :P

druiz17:
retest to ensure that everything  still works

1. In the interface enter edit mode and add multiple objects
2. On some of the objects set the parent field in properties to another objects id
3. delete the parent object and the children should be delete and also removed from entity list (also test if this works with selecting the grandchild first then the parent)
( on master the child object are not removed from the list)
4.When performing an undo of the deleted parent or child its children should be added back as well

hifi-gustavo:
Build console output for commit: 348c0c95e5b5da5e73e8a54488582e692b8028e0 available at: https://gist.github.com/321ac95b8ecbe89b0dcc6c19b2f280f8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9057-348c0c9.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9057-348c0c9.dmg

OmegaHeron:
Will re-test shortly

OmegaHeron:
QA Pass

Tested with a complex "multi-generational" link set - behaves as expected.  Deleting all or parts of of linkset and using undo results in proper restore.  I guess exporting linkset still does not work (nothing to do with this PR) - exporting a linkset only exports the root object, but, I think that's the way it's always been..

PR #9061, opened 2016-11-12
tab-like filtering of suggestions by type

howard-stearns:
Prototype of Alan's tab-like goto filtering.
(Any ugliness is my fault. I'm just winging it.)

- Does not yet attempt to persist the tab choice between sessions, but it should be remembering when you bring window up and down within a session.
- Needs an image to replace the wagging finger on the right.

### DIFF
--- a/interface/resources/qml/AddressBarDialog.qml
+++ b/interface/resources/qml/AddressBarDialog.qml
@@ -24,7 +24,7 @@ Window {
     HifiStyles.HifiConstants { id: hifiStyleConstants }
 
     objectName: "AddressBarDialog"
-    title: "Go To"
+    title: "Go To:"
 
     shown: false
     destroyOnHidden: false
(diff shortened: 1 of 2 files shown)


howard-stearns:
BTW, this makes a new fetch when you change "tabs", and the fetch asks the server to do the type-filtering.  Thus on the 'places' tab, you get the 100 highest-scoring concurrency stories, not a filtering from the 100 highest stories of all types. 

This will mitigate the issue of people's places falling of the end of the 100, but the only way to really address that is to do the text filtering on the server, too. That will require some non-trivial server-side work.

hifi-gustavo:
Build console output for commit: 17428d7ecbf11a953d893b7703ad08b4c2602658 available at: https://gist.github.com/d557b57766e7828edb196a241153a28a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9061-17428d7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9061-17428d7.dmg

howard-stearns:
Superseed by https://github.com/highfidelity/hifi/pull/9071

PR #9062, opened 2016-11-13
group snapshots by place

howard-stearns:
Prototype of grouping snapshots by place in feed.

The difference from current behavior is subtle:
- When there are multiple snapshots among the (filtered) feed results that were taken in the same place, only the top-scoring card is shown in the feed, and the word "snapshot" is instead changed to "snapshots". 
- When you click on the camera icon on a card that shows plural snapshots, it goes to the place page (with all the top snapshots listed) rather than to the snapshot card. (There is a server PR https://github.com/highfidelity/data-web/pull/868 will allow you to page through ALL the snapshots for a given place, instead of just the top five.)

 Note that concurrency cards still appear separately, even for the same place. Otherwise, "grouping" would effectively just be showing concurrency cards only. The separate https://github.com/highfidelity/hifi/pull/9061 allows showing only concurrency, only snapshots, or both.

There are some other possibilities that I haven't yet explored:
- A more distinctive visual difference when multiple snapshots are grouped together.
- A fancy multi-snapshot action when you click on the card (instead of just the icon). For example, maybe it could animate some display of all the cards so that you could get the broader idea of what goes on in this place, and then pick the one you want to go to (which can be at different locations after the same user entry point). Instead, this PR currently just takes you to the in-world location of the top snapshot.

### DIFF
--- a/interface/resources/qml/AddressBarDialog.qml
+++ b/interface/resources/qml/AddressBarDialog.qml
@@ -114,6 +114,7 @@ Window {
                 timestamp: model.created_at;
                 onlineUsers: model.online_users;
                 storyId: model.metaverseId;
+                drillDownToPlace: model.drillDownToPlace;
                 hoverThunk: function () { ListView.view.currentIndex = index; }
                 unhoverThunk: function () { ListView.view.currentIndex = -1; }
             }
@@ -362,6 +363,7 @@ Window {
             tags: tags,
             description: description,
             online_users: data.details.concurrency || 0,
+            drillDownToPlace: false,
 
             searchText: [name].concat(tags, description || []).join(' ').toUpperCase()
         }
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: fac5d1e12eaa22283a1a11956fe450a4c2975ed8 available at: https://gist.github.com/8bbc0a7a140976e34ff059d32c7be1a3<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9062-fac5d1e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9062-fac5d1e.dmg

howard-stearns:
Superseed by https://github.com/highfidelity/hifi/pull/9071

PR #9065, opened 2016-11-14
Fix deleteLater() vulnerability in scripts

huffman:


hifi-gustavo:
Build console output for commit: 51586ae0a33cacc9496c69977f0ac3f859108634 available at: https://gist.github.com/24e2b9d24de857e1efa5226bad2dd1ba<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9065-51586ae.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9065-51586ae.dmg

huffman:
This PR removes the deleteLater function from objects that are exposed to JavaScript. For instance, in release builds you can call `Camera.deleteLater()` from the scripting console (and anywhere else where you can access Camera) which will end in a crash.

To test:
 
 * Confirm that you cannot call `deleteLater()` on global objects - it's sufficient to check only a few (`Camera`, `Entities`, `Overlays`)
 * Smoke test that client and entity scripts are still working as expected (defaultScripts.js and the ping pong gun from the marketplace)

hifi-gustavo:
Build console output for commit: 601aa51392f50448a6baac7ed721b78e908141d7 available at: https://gist.github.com/24e2b9d24de857e1efa5226bad2dd1ba<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9065-601aa51.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9065-601aa51.dmg

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -66,6 +66,8 @@
 #include "MIDIEvent.h"
 
 static const QString SCRIPT_EXCEPTION_FORMAT = "[UncaughtException] %1 in %2:%3";
+static const QScriptEngine::QObjectWrapOptions DEFAULT_QOBJECT_WRAP_OPTIONS =
+                QScriptEngine::ExcludeDeleteLater | QScriptEngine::ExcludeChildObjects;
```
ZappoMan: Don't we have a couple of case where we need the child object to be exposed? In particular the `MyAvatar` object I think many of it's methods are implemented by the subclasses `AvartarData`... will those methods still work?

huffman: Currently if a script QObject has a child QObject with a name, the child QObject will be exposed as a property on the script QObject. This should not affect inherited methods and properties of superclasses.

nbq:
Testing this now

PR #9071, opened 2016-11-14
suggestions filter/group

howard-stearns:
Provide tab-like control over whether suggestions shows snapshots, place concurrency, or all.
In the case of all, group all the non-concurrency suggestions (all the snapshots) that are from the same place.
When showing concurrency or multiple snapshots, drilling down displays the place card (which in-turn lists the snapshots), rather than the story card. (The concurrency story card isn't very useful, and which snapshots of a multiple would we show?)


howard-stearns:
This combines and superseeds https://github.com/highfidelity/hifi/pull/9062 and https://github.com/highfidelity/hifi/pull/9061, which will be withdrawn.

This is DNM pending styling, but I want to get ready in case someone else becomes available to work on the styling as a follow-on PR based on this one.

howard-stearns:
Test plan:
- There should be three options at the top of the "GoTo" suggestions dialog: 'All', 'Places', and 'Snaps'.
 -- Places should show only concurrency suggestions (showing the number of people present), one suggestion per place, ordered by most-concurrency first.
 -- Snaps should show only snapshot suggestions, in score order. This may produce interleavings between places, based on when the snapshots were taken or by whom.
 -- All should be an interleaving of both, except that all snapshots for the same place combined into the first snapshot entry for that place, in which case the word "snapshot" is replaced by "snapshots".
- When you click the icon in the lower right of a suggestion, it should show the snapshot larger, with more info ONLY for singular snapshot. In all other cases (concurrency or plural snapshots), it should show the place info (which shows concurrency along with a listing of snapshots).
- Typing to filter should work even as you change tabs. For example, if there are multiple snapshots available for some place and you type some of that name: 'All' should show one plural "snapshots" suggestion and one matching place; tabbing to 'Places' should show just the place concurrency; and tabbing to 'Snaps' should show several individual "snapshot" suggestions.
- The last-used tab should be remembered as the GoTo dialog is brought up and down during a session, but should start with 'All' each session.

hifi-gustavo:
Build console output for commit: 6078b833a161a98c6341dbcf188a5fb261864e6b available at: https://gist.github.com/06ba18bb0b1be2b3ec246a8f55207bd6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-6078b83.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-6078b83.dmg

interface/resources/qml/hifi/Card.qml
```diff
@@ -59,27 +66,58 @@ Rectangle {
 
     Image {
         id: lobby;
-        width: parent.width;
-        height: parent.height;
+        width: parent.width - (isConcurrency ? 0 : 8);
+        height: parent.height - messageHeight - (isConcurrency ? 0 : 4);
```
davidkelly: I am curious -- would it make sense to pull these magic numbers (4 and 8) out of here and into a property or something that would be easier to tweak later?  Really just curious - all I know about qml is how to spell it.

interface/resources/qml/hifi/Card.qml
```diff
@@ -59,27 +66,58 @@ Rectangle {
...
             if (status == Image.Error) {
                 console.log("source: " + source + ": failed to load " + hifiUrl);
                 source = defaultThumbnail;
             }
         }
     }
+    Rectangle {
+        id: shadow1;
+        visible: isStacked;
+        width: parent.width - 5;
```
davidkelly: same comment about this 5.  Again extremely minor

interface/resources/qml/hifi/Card.qml
```diff
@@ -59,27 +66,58 @@ Rectangle {
...
+        }
+        gradient: Gradient {
+            GradientStop { position: 0.0; color: "gray" }
+            GradientStop { position: 1.0; color: "white" }
+        }
+    }
+    Rectangle {
+        id: shadow2;
+        visible: isStacked;
+        width: shadow1.width - 5;
```
davidkelly: same comment

hifi-gustavo:
Build console output for commit: 9809515853d1c49ee494aa57c740c0f4ecebf915 available at: https://gist.github.com/06ba18bb0b1be2b3ec246a8f55207bd6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-9809515.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-9809515.dmg

hifi-gustavo:
Build console output for commit: 182050c4b94da7346c2e1a8c0b242573073a3979 available at: https://gist.github.com/06ba18bb0b1be2b3ec246a8f55207bd6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-182050c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-182050c.dmg

hifi-gustavo:
Build console output for commit: b0518feee42ae80440ae93b70fb80275067f9260 available at: https://gist.github.com/06ba18bb0b1be2b3ec246a8f55207bd6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-b0518fe.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-b0518fe.dmg

RyanDowne:
I think there needs to be more padding between the tabs and the cards or we need to make the header stand out more.

howard-stearns:
Latest has more padding per ^^, and on my machine at least, the "disappearing tab-lozenge text" problem is reduced to sometimes blinking-out-and-returning, which is way more acceptable then disappearing entirely.

hifi-gustavo:
Build console output for commit: fef68ac53a200d0e97d6159f0d5fb4e5b316360f available at: https://gist.github.com/06ba18bb0b1be2b3ec246a8f55207bd6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-fef68ac.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-fef68ac.dmg

interface/resources/qml/hifi/Card.qml
howard-stearns: And in the end, you were right. They did end up being related are now properties.

hifi-gustavo:
Build console output for commit: 050d3438493b6a5e6e37017a1ac6b23e2881d2c9 available at: https://gist.github.com/06ba18bb0b1be2b3ec246a8f55207bd6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-050d343.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-050d343.dmg

hifi-gustavo:
Build console output for commit: 5b57354624a9c25bea844f7fee1e772ba57d99e7 available at: https://gist.github.com/06ba18bb0b1be2b3ec246a8f55207bd6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-5b57354.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-5b57354.dmg

murillodigital:
test this please

hifi-gustavo:
Build console output for commit: 5b57354624a9c25bea844f7fee1e772ba57d99e7 available at: https://gist.github.com/06ba18bb0b1be2b3ec246a8f55207bd6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-5b57354.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9071-5b57354.dmg

RyanDowne:
One thing I noticed about this is the tab selection is not stateful and we might want it to be.

zfox23:
Testing on Win10/Rift/Hydra.

> There should be three options at the top of the "GoTo" suggestions dialog: 'All', 'Places', and 'Snaps'. Places should show only concurrency suggestions (showing the number of people present), one suggestion per place, ordered by most-concurrency first.

Yep! Some places (Mexico and Dreaming during testing) didn't have preview images, but I'm sure that's not due to this PR.

Regarding the number ("X" people), it looks like either the font size or the positioning of it is different from the "people" text. Is that true? Was it an active choice to not have those two strings line up? If we want the font size to be different, I think we should at least make the baseline consistent.



> Snaps should show only snapshot suggestions, in score order. This may produce interleavings between places, based on when the snapshots were taken or by whom.

Yes, this works. Problem: When usernames are too long, the "Snapshot" icon covers part of the end of the name string. Can we truncate the username strings before they go below the button?


> All should be an interleaving of both, except that all snapshots for the same place combined into the first snapshot entry for that place, in which case the word "snapshot" is replaced by "snapshots".

Yes, this is true. As a regular user, I'd be confused about _how_ the snapshots were grouped when they were grouped. I'm having trouble thinking of ways we could teach the user that they were grouped by place. I'm not sure that putting the place name on stacks is the right solution, but maybe...

Also, would it be possible to extend the click target for the snapshot groups to include the "stack?" I found myself clicking on the stack to learn more about it, but I had to click on the title bar of the top snapshot to actually see more info.


> When you click the icon in the lower right of a suggestion, it should show the snapshot larger, with more info ONLY for singular snapshot. In all other cases (concurrency or plural snapshots), it should show the place info (which shows concurrency along with a listing of snapshots).

Yep, this works! The fact that the place info is now more visible makes it much more obvious that we need a "back" button or breadcrumb navigation here. 😆 Also, the social media "Share" buttons on this page are slightly misaligned. It seems like the individual snapshot pages could use a little design love overall.


> Typing to filter should work even as you change tabs. For example, if there are multiple snapshots available for some place and you type some of that name: 'All' should show one plural "snapshots" suggestion and one matching place; tabbing to 'Places' should show just the place concurrency; and tabbing to 'Snaps' should show several individual "snapshot" suggestions.

Yep, this works!


> The last-used tab should be remembered as the GoTo dialog is brought up and down during a session, but should start with 'All' each session.

Yep, no problems here.


> Be sure to test in Desktop and HMD. E.g., are things still ok when you laser over the title bar? Can you still open/close the larger snapshot/place info and have it cover the virtual keyboard nicely, etc.

> Long place names and snapshot descriptions should be elided in the suggestions summaries (although I don't think the former will kick in with any current names)

Ah, I made a comment about this above. I saw a snapshot by "Richardus.Raymaker" where his name wasn't truncated.


> Clicking the big blue arrow/chevron/swiping-finger thingie should advance forward three pictures, until it reaches the end. (Re-clicking the tab button again reloads and puts you at the beginning.)

I don't see a big blue arrow/chevron/swiping-finger thingie. What am I missing? 😮

**A few final comments:**

- I don't think this bug is related to this PR, and I don't know if this is Hydra-only, but lasering on 2D menus is completely messed up for me. When hovering over menus or HUD elements, the laser is inaccurate, and I can't even see the laser beam (only semi-transparent circle where it lands).
- Scrolling through the list is a bit weird for a few reasons. 1. Hovering over a half-visible card makes that card fully visible, which causes the list to stop fling-scrolling and scroll erratically. 2. The list of cards doesn't support a mouse's vertical scroll wheel, but it does support the horizontal one. Is it possible to allow the user to scroll through that list using their vertical scroll wheel?
- This is GREAT and such a huge improvement over the original "Go To" menu! Very pumped about it. Awesome work.

howard-stearns:
Ryan - I also wonder about persisting the tab setting. I figured there was some value in resetting people to All, but maybe I was just being lazy. :-)

Zach:
- I agree that there are issues with the stack display. In general, I don't know what to do about these, but I'm inclined to first do the simplest thing that could possibly work, until a "clearly right answer" emerges:
 - It isn't obvious how they are grouped. (As you note.) Maybe it doesn't matter?
 - The chevron doesn't reach down to the stack. Should it? What about when the last card is not stacked? What about when the chevron is aligned with the gap between a stacked card and a non-stacked one?
 - Having something hang below makes it confusing to see what the click area is. (As you note.) Should it always include the area below, or only when stacked? Should it not include the corners? Should it highlight to show that it is part of the same hot area? If it highlighted, wouldn't that indicate that it is going to do something with the stack (such as spread them)?  Seems to me to open a can of worms.
- I might agree on back button. Not sure what to do, but in any case it's out of scope for this PR.
- I'd love to animate the movement of the suggestions - both when you click and as they fill. But I couldn't make it work (in previous development). Maybe you could take that on!
- We haven't discussed scroll wheel. Personally, I'm not keen on having a vertical scroll wheel do horizontal scrolling, and where not generally keen on people using mouse and keyboard.
- The bigger, vertically centered text for the number is on purpose. See Alan's spec.
- Fixed position is on purpose (and not different from master). The UX design context is that this is letterbox-like.
- The long names were fixed in a later version.
- Laser issues are in master. See https://github.com/highfidelity/hifi/pull/9086 and related.

zfox23:
Thank you for your thoughtful comments. I recognize the desire for "the simplest thing that works." I can certainly take a look at those animations :)

The one thing that I still feel personally uncertain about is the bigger, vertically-centered number text, but that's a design preference for sure.

Marking as QA Approved.

PR #9074, opened 2016-11-15
Bug fix for blank webEntities during long sessions

hyperlogic:
In prolonged sessions _currentWebCount would reach MAX_CONCURRENT_WEB_VIEWS and prevent new webEntities from being properly created and initialized.

The _currentWebCount inside of RenderableWebEntityItem would become inflated when buildWebSurface is called without a currentContext. The early return did not properly take this into account.  Instead we move the increment after the early return and right before the _webSurface is actually created.

### DIFF
--- a/libraries/entities-renderer/src/RenderableWebEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableWebEntityItem.cpp
@@ -31,7 +31,7 @@
 const float METERS_TO_INCHES = 39.3701f;
 static uint32_t _currentWebCount { 0 };
 // Don't allow more than 100 concurrent web views
-static const uint32_t MAX_CONCURRENT_WEB_VIEWS = 100;
+static const uint32_t MAX_CONCURRENT_WEB_VIEWS = 20;
 // If a web-view hasn't been rendered for 30 seconds, de-allocate the framebuffer
 static uint64_t MAX_NO_RENDER_INTERVAL = 30 * USECS_PER_SECOND;
 
--- a/libraries/entities-renderer/src/RenderableWebEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableWebEntityItem.cpp
@@ -69,8 +69,6 @@ bool RenderableWebEntityItem::buildWebSurface(QSharedPointer<EntityTreeRenderer>
         qWarning() << "Too many concurrent web views to create new view";
         return false;
     }
-    qDebug() << "Building web surface";
-
     QString javaScriptToInject;
     QFile webChannelFile(":qtwebchannel/qwebchannel.js");
     QFile createGlobalEventBridgeFile(PathUtils::resourcesPath() + "/html/createGlobalEventBridge.js");
@@ -85,12 +83,15 @@ bool RenderableWebEntityItem::buildWebSurface(QSharedPointer<EntityTreeRenderer>
         qCWarning(entitiesrenderer) << "unable to find qwebchannel.js or createGlobalEventBridge.js";
     }
 
-    ++_currentWebCount;
     // Save the original GL context, because creating a QML surface will create a new context
     QOpenGLContext * currentContext = QOpenGLContext::currentContext();
     if (!currentContext) {
         return false;
     }
+
+    ++_currentWebCount;
+    qDebug() << "Building web surface: " << getID() << ", #" << _currentWebCount << ", url = " << _sourceUrl;
+
     QSurface * currentSurface = currentContext->surface();
 
     auto deleter = [](OffscreenQmlSurface* webSurface) {
@@ -356,6 +357,8 @@ void RenderableWebEntityItem::destroyWebSurface() {
         QObject::disconnect(_hoverLeaveConnection);
         _hoverLeaveConnection = QMetaObject::Connection();
         _webSurface.reset();
+
+        qDebug() << "Delete web surface: " << getID() << ", #" << _currentWebCount << ", url = " << _sourceUrl;
     }
 }
 


hifi-gustavo:
Build console output for commit: eb401c125a2d92290cb397cca7fb1e9f2ca65515 available at: https://gist.github.com/0242c37ff5b0bda99e24aeca5153d353<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9074-eb401c1.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9074-eb401c1.dmg

hyperlogic:
QA Test:
   * jump between domains that have web entities within them.  (for example: earth & eschatology)
   * do this many ~20 times.
   * web entities should continue to work in each domain, and not be blank.

OmegaHeron:
Tested Windows 10

I setup my own test for this.  I created 50 web entities each in 2 domains.  Swapped between each.

Current RC (5695)  I couldn't even see all 50.

This PR - I can see all 50 in each domain.

First - yes, this solves issue of lost web entities when large numbers are present - be those in a single or multiple domains.  I think, though, it exposes a new problem in that with this PR code multiple visits to domains with large numbers of web entities eventually leads to a crash and, unlike current RC/master, huge amounts of resources can be consumed for all those web entities...

My test case was extreme, but, it highlights what could happen if interface is presented with large numbers of web entities.

One one attempt I was able to make 11 visits to each location then crashed.  As of 11th attempt I still saw content on all web entities.

Second attempt I crashed after 5th visit to each.
Third attempt I crashed after 8th visit to each.
Bugsplats sent for all.

PhilipRosedale:
What is our value for MAX_CONCURRENT_WEB_VIEWS?  

It should be a pretty small value, yes?  And then the largest ones only should be shown?  I would think otherwise this becomes a trivial content griefing vector.

OmegaHeron:
It's interesting that in RC it initially allows 17 QTWebEngine processes to launch, then clamps to 8 on domain change.  While this PR effectively seems to have no limits at all.  Whether that's new behavior or something that this bug fix exposes that was undetectable before is > my role to figure out.

cozza13:
Waiting on feedback fro, @hyperlogic to progress the PR

ZappoMan:
@PhilipRosedale - MAX_CONCURRENT_WEB_VIEWS = 100 right now.

ZappoMan:
All of @OmegaHeron's crashes were in `Qt5WebEngineCore` at the "same call stack"-- although it's not obvious from the bugsplat details if this is new or existing.

OmegaHeron:
I have my gut instinct it's existing but never seen until we started loading up lots more web entities.  Using 50 web entities in view - all with same default url also took my system to its knees.. pushing 100% memory use on an 8Gig system - and high CPU use loading all the web windows.  Granted - you wouldn't likely have such large numbers of web entities present in close locations, but, it's, as Philip pointed out, a powerful grief vector and - it has changed behavior from legacy code in that regardless of how many web entities were placed nearby one another - a very limited amount appeared/spawned qtWebProcess engines.

hyperlogic:
Now that the limit is functioning, it seems clear that allowing 100 web entities is too many.  I reduced it to 20.  @OmegaHeron give this a try.

OmegaHeron:
@hyperlogic - will do.  Thanks!

ZappoMan:
@hyperlogic - what happens when you hit the limit? do the web entities just not appear? do they eventually appear if others are removed -- seems like if there are 21 people in the domain, someone won't be able to open their tablet.

hifi-gustavo:
Build console output for commit: 066ddad25a245236e0ca9d7993f19568d964fcb2 available at: https://gist.github.com/0242c37ff5b0bda99e24aeca5153d353<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9074-066ddad.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9074-066ddad.dmg

OmegaHeron:
The behavior of master is this...

Not connected to any domain at all - i.e. pointed to an address with no stack -- you start with six qtWebEngineProcess.exe tasks spawned then if you visited a space with 50 web entities in a close cluster web process count rises to 10, displaying first 4 web entity objects tree presents -- then back to disconnected space web processes drop to 7 (why 7 now, not 6? And why 6 to begin with?) -- back to test zone it rises to eight showing one entity.

*This PR seems pointed to fixing why you show 4 then 1 visiting a space with lots of web entities and succeeds in it.*

Now this PR works similar in that it initially spools up 6 web process tasks - then visiting a domain with 50 in view you get first 20 tree presents (26 running qtWebEngine processes) - then back to disconnected space you, again, have 7 web processes running - back to test zone you have 27 web process tasks running - showing first 20 web entities tree presents.

So - this PR does faithfully show 20 web entities - it releases them on domain changes and does not, after many cycles between high web entity domains - present less than 20 (unlike master that does lose 3 after one visit).

Cans of worms are opening here though - not fault of this PR, but, web entities are costly - now being able to see larger numbers in a rezzing scene has definite impact and - this leaves questions as to how to deal with entering a space with say, 10 web entities in tree - plus agents possibly with attached web processes...

OmegaHeron:
Also - it seems Interface does **not** like changing domains while web entities are loading - I suspect this has always been case, but, now that you can have 20 all trying to populate in scene and leave it as that's in progress the crash is more apparent....  My bugsplats are all coming while a cluster of entities are attempting to load and I leave domain.  If I wait for them to all full render I've not been able to crash it on domain change.

hyperlogic:
@OmegaHeron the number of chrome helper processes is only loosely related to the number of web entities.  I could be wrong, but Qt will spin up processes on demand then leave some running in the background until they are needed again.

> @hyperlogic - what happens when you hit the limit? do the web entities just not appear? do they eventually appear if others are removed -- seems like if there are 21 people in the domain, someone won't be able to open their tablet.

@ZappoMan the web entities appear but with blank white content.  They will eventually appear when others are removed.  Yes, if there are 20 web entities in the domain and someone opens their tablet it will appear with a blank white surface.  We could add a priority system to address this, but at the moment, outside of test cases, there are no domains with 20 web entities.  So, I don't think the need is pressing.

hyperlogic:
@OmegaHeron Yes there are resource problems with web entities, including crashes.  I'll make sure to open a fogbugz ticket on this.  This PR doesn't fix the issue, but mitigates it somewhat.  Just FYI how often is this crashing for you when you switch domains, 100%, 20%?  Do you have a repro

OmegaHeron:
@hyperlogic - "number of chrome helper processes is only loosely related to the number of web entities." seems incorrect from empirical data.

On Windows the helper processes faithfully follow domain content as in - start with six - visit a domain with X web entities then six + X are active.  Leave domain for space with no web entities then 7 are active.  Return to another domain with Y web entities then seven + Y are active.  Leave to a domain with none and 7 remain active over at least 20 minutes I watched.  The initial 6 then 1 retain consistent PID numbers while the +X (X =< 20 currently) come and go with differing PIDS.

Repro for my crash is as follows;

Create an area with 20 web entities (or more if you like).  Create a second area with equal or greater count.  Enter space and once you see outline for cluster and they've began loading content you switch to alternate space - again - see the outline, content begins to load - move back.  My crash rate = 100% doing that.  The variable being how many times of cycling it takes.

hyperlogic:
@OmegaHeron yeah, I revised my comment.  It's 1:1, kind of like chrome tabs

hyperlogic:
@OmegaHeron About the repro:  when you say "Create an area" & "Create a second area"  Do you mean separate domains?  or two areas far away from each other in the same domain?

OmegaHeron:
@hyperlogic - I'm using two domains so I suggest same for repro as it may be the crash comes on domain change vs internal coordinate changes in a single domain....

ZappoMan:
I'm calling this a QA Pass -- it fixes the originally reported issue. The other problems identified in testing have been filed and need to be looked at independently.

PR #9079, opened 2016-11-15
out-of-body-experience: hover-drive works in HMD

AndrewMeadows:
* driving while in HMD works in hover-mode

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -465,7 +465,9 @@ void MyAvatar::simulate(float deltaTime) {
     // update sensorToWorldMatrix for camera and hand controllers
     // before we perform rig animations and IK.
 
-    updateSensorToWorldMatrix(_enableVerticalComfortMode ? SensorToWorldUpdateMode::VerticalComfort : SensorToWorldUpdateMode::Vertical);
+    if (_characterController.getState() != CharacterController::State::Hover) {
+        updateSensorToWorldMatrix(_enableVerticalComfortMode ? SensorToWorldUpdateMode::VerticalComfort : SensorToWorldUpdateMode::Vertical);
+    }
 
     {
         PerformanceTimer perfTimer("skeleton");
@@ -1356,7 +1358,8 @@ void MyAvatar::updateMotors() {
         }
 
         if (qApp->isHMDMode()) {
-            // OUTOFBODY_HACK: motors are applied differently in HMDMode
+            // OUTOFBODY_HACK: in HMDMode motors are applied differently: a "follow" motor is added
+            // during the CharacterController's substep
         } else {
             if (_isPushing || _isBraking || !_isBeingPushed) {
                 _characterController.addMotor(_actionMotorVelocity, motorRotation, DEFAULT_MOTOR_TIMESCALE, INVALID_MOTOR_TIMESCALE);
@@ -1887,12 +1890,17 @@ void MyAvatar::updateActionMotor(float deltaTime) {
 }
 
 void MyAvatar::applyVelocityToSensorToWorldMatrix(const glm::vec3& velocity, float deltaTime) {
-    glm::vec3 horizontalVelocity(velocity.x, 0.0f, velocity.z);
-    float speed2 = glm::length2(horizontalVelocity);
+    glm::vec3 newVelocity = velocity;
+    if (_characterController.getState() != CharacterController::State::Hover) {
+        newVelocity -= glm::dot(newVelocity, _worldUpDirection);
+    }
+    float speed2 = glm::length2(newVelocity);
     if (speed2 > MIN_AVATAR_SPEED_SQUARED) {
-        glm::vec3 position = extractTranslation(_sensorToWorldMatrix) + deltaTime * horizontalVelocity;
+        glm::vec3 position = extractTranslation(_sensorToWorldMatrix) + deltaTime * newVelocity;
         // update the position column of matrix
-        _sensorToWorldMatrix[3] = glm::vec4(position, 1);
+        glm::mat4 newSensorToWorldMatrix = _sensorToWorldMatrix;
+        newSensorToWorldMatrix[3] = glm::vec4(position, 1.0f);
+        setSensorToWorldMatrix(newSensorToWorldMatrix);
     }
 }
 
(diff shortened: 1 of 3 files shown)


interface/src/avatar/MyAvatar.cpp
```diff
@@ -465,7 +465,9 @@ void MyAvatar::simulate(float deltaTime) {
     // update sensorToWorldMatrix for camera and hand controllers
     // before we perform rig animations and IK.
 
-    updateSensorToWorldMatrix(_enableVerticalComfortMode ? SensorToWorldUpdateMode::VerticalComfort : SensorToWorldUpdateMode::Vertical);
+    if (_characterController.getState() != CharacterController::State::Hover) {
+        updateSensorToWorldMatrix(_enableVerticalComfortMode ? SensorToWorldUpdateMode::VerticalComfort : SensorToWorldUpdateMode::Vertical);
+    }
```
AndrewMeadows: _sensorToWorldMatrix doesn't slave vertically to avatar position when in State::Hover.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1887,12 +1890,17 @@ void MyAvatar::updateActionMotor(float deltaTime) {
...
+    float speed2 = glm::length2(newVelocity);
     if (speed2 > MIN_AVATAR_SPEED_SQUARED) {
-        glm::vec3 position = extractTranslation(_sensorToWorldMatrix) + deltaTime * horizontalVelocity;
+        glm::vec3 position = extractTranslation(_sensorToWorldMatrix) + deltaTime * newVelocity;
         // update the position column of matrix
-        _sensorToWorldMatrix[3] = glm::vec4(position, 1);
+        glm::mat4 newSensorToWorldMatrix = _sensorToWorldMatrix;
+        newSensorToWorldMatrix[3] = glm::vec4(position, 1.0f);
+        setSensorToWorldMatrix(newSensorToWorldMatrix);
     }
```
AndrewMeadows: After computing the new transform call setSensorToWorldMatrix() in this context (it has side effects that are required for things to work properly).

interface/src/avatar/MyCharacterController.cpp
```diff
@@ -160,20 +158,39 @@ bool MyCharacterController::testRayShotgun(const glm::vec3& position, const glm:
...
+                    if (rayResult.m_hitNormalWorld.dot(_currentUp) < _minFloorNormalDotUp) {
+                        result.walkable = false;
+                        // the bottom scan wasn't walkable
+                        // remove both forwardSlop and backSlop
+                        result.hitFraction = glm::min(1.0f, (closestRayResult.m_closestHitFraction * (backSlop + stepLength + forwardSlop) - backSlop) / stepLength);
+                        return result.hitFraction < 1.0f;
+                    }
+                }
+            }
+        }
```
AndrewMeadows: Adding these scan points makes it  a little less likely to tunnel through geometry when flying fast.

libraries/physics/src/CharacterController.cpp
```diff
@@ -277,10 +277,15 @@ void CharacterController::playerStep(btCollisionWorld* collisionWorld, btScalar
...
             vel += _followVelocity;
             const float HORIZONTAL_FOLLOW_TIMESCALE = 0.05f;
-            const float VERTICAL_FOLLOW_TIMESCALE = (_state == State::Hover) ? HORIZONTAL_FOLLOW_TIMESCALE : 20.0f;
+            float verticalFollowTimescale = 20.0f;
+            if (_state == State::Hover) {
+                verticalFollowTimescale = HORIZONTAL_FOLLOW_TIMESCALE;
+            } else {
+                // remove vertical component
+                vel -= vel.dot(_currentUp) * _currentUp;
+            }
```
AndrewMeadows: When hovering decrease the vertical timescale of the avatar's slave motor so that it can actually follow the HMD into the air.

libraries/physics/src/CharacterController.cpp
```diff
@@ -773,7 +778,8 @@ void CharacterController::updateState() {
             const btScalar MAX_WALKING_SPEED = 2.5f;
             bool flyingFast = _state == State::Hover && actualHorizVelocity.length() > (MAX_WALKING_SPEED * 0.75f);
 
-            if ((_floorDistance < MIN_HOVER_HEIGHT) && !jumpButtonHeld && !flyingFast) {
+            if ((_floorDistance < MIN_HOVER_HEIGHT) &&
+                    !(jumpButtonHeld || flyingFast || (now - _jumpButtonDownStartTime) > JUMP_TO_HOVER_PERIOD)) {
```
AndrewMeadows: Don't transition from State::Hover to State::InAir when we're mashing on the "jump" key.

AndrewMeadows:
To test:

(1) Login with HMD
(2) Press and hold the UP key (e) to jump and then fly into the air --> this used to not work.

hifi-gustavo:
Build console output for commit: d749697c46238ffd90f5de7800f52650a971b371 available at: https://gist.github.com/b84e4e440bb59a906564c6fc128174b3<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9079-d749697.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9079-d749697.dmg

interface/src/avatar/MyAvatar.cpp
hyperlogic: Hmm, why is this special case needed?

interface/src/avatar/MyAvatar.cpp
hyperlogic: Hmm, I believe you need to call this here, because you are no longer calling updateSensorToWorldMatrix in Avatar::simulate() (while hovering)

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1929,16 +1937,19 @@ void MyAvatar::updatePosition(float deltaTime) {
...
             if (_motionBehaviors & AVATAR_MOTION_ACTION_MOTOR_ENABLED) {
-                glm::quat liftRotation;
-                swingTwistDecomposition(glmExtractRotation(_sensorToWorldMatrix * getHMDSensorMatrix()), _worldUpDirection, liftRotation, motorRotation);
+                if (_characterController.getState() == CharacterController::State::Hover ||
+                        _characterController.getCollisionGroup() == BULLET_COLLISION_GROUP_COLLISIONLESS) {
+                    motorRotation = glmExtractRotation(_sensorToWorldMatrix * getHMDSensorMatrix());
+                } else {
+                    glm::quat liftRotation;
+                    swingTwistDecomposition(glmExtractRotation(_sensorToWorldMatrix * getHMDSensorMatrix()), _worldUpDirection, liftRotation, motorRotation);
+                }
```
hyperlogic: Oh, is this to fix vertical flying?

interface/src/avatar/MyAvatar.cpp
AndrewMeadows: In short: during `State::Hover` the avatar's physical capsule should just follow the HMD around, horizontally AND vertically.  Which means we don't want the HMD to be doing any following --> we can't call updateSensorToWorldMatrix() for this case.

PR #9080, opened 2016-11-15
Fix buffer leak in line entity

jherico:
Line entities need to release their GeometryCache IDs.

## Testing

Functionality should remain unchanged.  You can test this by creating a domain containing a line entity.  If you repeatedly disconnect (by connecting to some invalid location like "127.0.0.0") and reconnect, you will see the buffer count remain steady.  In the production build the buffer count will rise over time.

### DIFF
--- a/libraries/entities-renderer/src/RenderableLineEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableLineEntityItem.cpp
@@ -24,6 +24,13 @@ EntityItemPointer RenderableLineEntityItem::factory(const EntityItemID& entityID
     return entity;
 }
 
+RenderableLineEntityItem::~RenderableLineEntityItem() {
+    auto geometryCache = DependencyManager::get<GeometryCache>();
+    if (geometryCache) {
+        geometryCache->releaseID(_lineVerticesID);
+    }
+}
+
 void RenderableLineEntityItem::updateGeometry() {
     auto geometryCache = DependencyManager::get<GeometryCache>();
     if (_lineVerticesID == GeometryCache::UNKNOWN_ID) {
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: fda2cde0067cb6bd9e6e99c3fab8dfd5e0819f5f available at: https://gist.github.com/711a74d8ce3718879bac3365a38790ba<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9080-fda2cde.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9080-fda2cde.dmg

OmegaHeron:
What precisely is a "line entity"?  Is that an overall type for the cube primitive or something else I've never noticed?

Nex-Pro:
@OmegaHeron do you have an HMD and hand controllers ?

OmegaHeron:
@nex-pro - HMD yes, hand controllers no.  Guess that explains that.  :)  Thanks.  Turning this one back to QA pool.

Nex-Pro:
@OmegaHeron Ok I'll take it

Nex-Pro:
@OmegaHeron Basically they are polylines such as those created when using the whiteboard markers.

Nex-Pro:
# PASS on buffers not increasing due to set unchanging polylines.

Observed that buffers are not being released consistently over time.

I ran several long sessions in this and saw that the initial low buffer count at 127.0.0.0  would eventually climb after travelling to other domains and back. When moving between home where I had a whiteboard and back to 127.0.0.0 there was no increase. I tried with single polylines and large numbers. Each polyline stroke or segment seemed to add an additional buffer. 

To be clear: If the polyline count remained the same the buffer count where the lines were located did not creep up.

PR #9094, opened 2016-11-18
Turn Edit.js on when importing entities

Atlante45:
- Turn Edit.js on when importing entities

Test Plan:
- Import svo file with edit.js turned on/off
  - In each case it should be turned on after the import.

### DIFF
--- a/scripts/system/edit.js
+++ b/scripts/system/edit.js
@@ -240,11 +240,8 @@ var toolBar = (function () {
             hoverState: 3,
             defaultState: 1
         });
-        activeButton.clicked.connect(function () {
-            that.setActive(!isActive);
-            activeButton.writeProperty("buttonState", isActive ? 0 : 1);
-            activeButton.writeProperty("defaultState", isActive ? 0 : 1);
-            activeButton.writeProperty("hoverState", isActive ? 2 : 3);
+        activeButton.clicked.connect(function() {
+            that.toggle();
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: cfd4294743e4a8bd3624a53bfc4d4a7b46584739 available at: https://gist.github.com/07e3b13358de99fa5c418dac227aea81<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9094-cfd4294.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9094-cfd4294.dmg

Nex-Pro:
Tested on win10.

### PASS

works fine

scripts/system/edit.js
```diff
@@ -1185,8 +1192,6 @@ function importSVO(importURL) {
             if (isActive) {
                 selectionManager.setSelections(pastedEntityIDs);
             }
-
-            Window.raiseMainWindow();
```
huffman: Was there a specific reason to remove this? I admittedly don't remember why it needed to be here...

Atlante45: Maybe on mac if it's minimized and you import from the menu?
Not really useful anymore and it's causing the OpenGL window to lose focus.

PR #9105, opened 2016-11-21
Fix for unable to teleport to areas that you should

druiz17:
Fix the issue of unable to teleport in areas you should be able to. This usually occurs in areas where there are many objects in close proximity. 

To Test
1. In the domain eschatology, head over the balcony area.
2. Try to teleport to an area near one of the balcony archway/pillars (on master branch you would not be able to teleport into these areas)

 If you know of other places that this occurs in, please test this fix there.

### DIFF
--- a/libraries/entities/src/EntityTreeElement.cpp
+++ b/libraries/entities/src/EntityTreeElement.cpp
@@ -538,10 +538,12 @@ bool EntityTreeElement::findRayIntersection(const glm::vec3& origin, const glm::
     void** intersectedObject, bool precisionPicking) {
 
     keepSearching = true; // assume that we will continue searching after this.
+
     float distanceToElementCube = std::numeric_limits<float>::max();
     float distanceToElementDetails = distance;
     BoxFace localFace;
     glm::vec3 localSurfaceNormal;
+    
     // if the ray doesn't intersect with our cube, we can stop searching!
     if (!_cube.findRayIntersection(origin, direction, distanceToElementCube, localFace, localSurfaceNormal)) {
         keepSearching = false; // no point in continuing to search
@@ -552,6 +554,7 @@ bool EntityTreeElement::findRayIntersection(const glm::vec3& origin, const glm::
     if (!canRayIntersect()) {
         return false; // we don't intersect with non-leaves, and we keep searching
     }
+
     // if the distance to the element cube is not less than the current best distance, then it's not possible
     // for any details inside the cube to be closer so we don't need to consider them.
     if (_cube.contains(origin) || distanceToElementCube < distance) {
@@ -559,7 +562,6 @@ bool EntityTreeElement::findRayIntersection(const glm::vec3& origin, const glm::
         if (findDetailedRayIntersection(origin, direction, keepSearching, element, distanceToElementDetails,
                 face, localSurfaceNormal, entityIdsToInclude, entityIdsToDiscard, visibleOnly, collidableOnly,
                 intersectedObject, precisionPicking, distanceToElementCube)) {
-
 
             if (distanceToElementDetails < distance) {
                 distance = distanceToElementDetails;
@@ -580,9 +582,7 @@ bool EntityTreeElement::findDetailedRayIntersection(const glm::vec3& origin, con
     // only called if we do intersect our bounding cube, but find if we actually intersect with entities...
     int entityNumber = 0;
     bool somethingIntersected = false;
-
     forEachEntity([&](EntityItemPointer entity) {
-
         if ( (visibleOnly && !entity->isVisible()) || (collidableOnly && (entity->getCollisionless() || entity->getShapeType() == SHAPE_TYPE_NONE))
             || (entityIdsToInclude.size() > 0 && !entityIdsToInclude.contains(entity->getID()))
             || (entityIDsToDiscard.size() > 0 && entityIDsToDiscard.contains(entity->getID())) ) {
@@ -620,16 +620,14 @@ bool EntityTreeElement::findDetailedRayIntersection(const glm::vec3& origin, con
         glm::vec3 entityFrameDirection = glm::vec3(worldToEntityMatrix * glm::vec4(direction, 0.0f));
 
         // we can use the AABox's ray intersection by mapping our origin and direction into the entity frame
-        // and testing intersection there.localDistance = 0.0f;
+        // and testing intersection there.
         if (entityFrameBox.findRayIntersection(entityFrameOrigin, entityFrameDirection, localDistance,
                                                 localFace, localSurfaceNormal)) {
-
             if (entityFrameBox.contains(entityFrameOrigin) || localDistance < distance) {
                 // now ask the entity if we actually intersect
                 if (entity->supportsDetailedRayIntersection()) {
                     if (entity->findDetailedRayIntersection(origin, direction, keepSearching, element, localDistance,
                         localFace, localSurfaceNormal, intersectedObject, precisionPicking)) {
-
 
                         if (localDistance < distance) {
                             distance = localDistance;
--- a/libraries/entities/src/EntityTreeElement.cpp
+++ b/libraries/entities/src/EntityTreeElement.cpp
@@ -543,7 +543,7 @@ bool EntityTreeElement::findRayIntersection(const glm::vec3& origin, const glm::
     float distanceToElementDetails = distance;
     BoxFace localFace;
     glm::vec3 localSurfaceNormal;
-    
+
     // if the ray doesn't intersect with our cube, we can stop searching!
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: a176c624ab329bbbae0c36c982936122eee04387 available at: https://gist.github.com/5389be9a99de4362fa9719c2e6514da0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9105-a176c62.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9105-a176c62.dmg

Nex-Pro:
### PASS

tested on Win10 and Vive.

Huge improvement on chance of getting a Teleport Blue Ring. 

Did not notice any particular problem at Eschatology on master but this PR makes getting a TP Ring at Mexico hugely more reliable.

In master the ground surrounding the fountain at Mexico is difficult to TP to due to the laser not locking to the surfaces there.  In this PR there is no problem.

Note: Unrelated to this PR I still get orphaned TP Rings on occasion.

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -582,7 +580,11 @@ bool EntityTreeElement::findDetailedRayIntersection(const glm::vec3& origin, con
     // only called if we do intersect our bounding cube, but find if we actually intersect with entities...
     int entityNumber = 0;
     bool somethingIntersected = false;
+
+    float bestDistance = std::numeric_limits<float>::max();
+    float bestLocalDistance = std::numeric_limits<float>::max();
```
huffman: Is this needed? It looks like it's only used in one other place, and never assigned to elsewhere.

druiz17: bestDistance keeps track of the  distance that is returned from findDetialedRayIntersection and bestLocalDistacne should keep track of the best distance within that function scope. Thank you for pointing out that I never assign bestLocalDistance.

hifi-gustavo:
Build console output for commit: 2dfb745443466f620007cab26ce172c84e65742b available at: https://gist.github.com/5389be9a99de4362fa9719c2e6514da0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9105-2dfb745.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9105-2dfb745.dmg

hifi-gustavo:
Build console output for commit: daf12047cacbd5d8472a9e938c6f2103068a0f3f available at: https://gist.github.com/5389be9a99de4362fa9719c2e6514da0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9105-daf1204.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9105-daf1204.dmg

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -620,15 +620,18 @@ bool EntityTreeElement::findDetailedRayIntersection(const glm::vec3& origin, con
         glm::vec3 entityFrameDirection = glm::vec3(worldToEntityMatrix * glm::vec4(direction, 0.0f));
 
         // we can use the AABox's ray intersection by mapping our origin and direction into the entity frame
-        // and testing intersection there.
-        if (entityFrameBox.findRayIntersection(entityFrameOrigin, entityFrameDirection, localDistance,
+        // and testing intersection there.localDistance = 0.0f;
+        localDistance = 0.0f;
```
ZappoMan: why not initialize this back at line 598 where the variable is constructed/declared...

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -620,15 +620,18 @@ bool EntityTreeElement::findDetailedRayIntersection(const glm::vec3& origin, con
...
+        if (entityFrameBox.contains(entityFrameOrigin) || entityFrameBox.findRayIntersection(entityFrameOrigin, entityFrameDirection, localDistance,
                                                 localFace, localSurfaceNormal)) {
+
             if (localDistance < distance) {
```
ZappoMan: something seems wrong about this PR -- I must be missing something... we would expected that *most* distances are going to be larger than 0.0f "away" from the ray... and so the last "best" distance is likely to be larger than 0.

If the `entityFrameOrigin` *is not* within the `entityFrameBox` --- then that means the `entityFrameBox.findRayIntersection()` method will be called, and `localDistance` will get over-written (if it intersects)...

So likely if the `entityFrameOrigin` *is within* the `entityFrameBox` then you'll end up inside the if, when `localDistance` = 0.0... which means it will surely be less than the last good `distance` and into this code.... (see follow on comment below)...

ZappoMan:
I'm removing the QA Pass - because the code has been significantly changed since it was tested. I don't think the new version is going to work...

Test cases you need to try:
- ray picking from inside of a large model, and trying to pick against the model.
- ray picking inside a large model that has other smaller entities and models inside of it.
- ray picking inside a larger model that has non-detailed entities inside of it.

libraries/entities/src/EntityTreeElement.cpp
druiz17: The main problem this is trying to solve is when you are ray picking in a large model with small models inside of it. Since entityFrameBox.findRayIntersection will return the distance to bounding box... if a ray intersects a large model then a small model... the distance to the bounding box to the small model will be smaller then the large model since the small object bounding box is within the large. resulting in the If statement at line 628 to prevent the large model being selected as the closest. 

So the idea was to reset the localDistance and if the entityFrameOrigin is in entityFrameBox so we would end up in the if statement and get the correct distance but I realize as you stated  that if a model doesn't detail ray picking...there is a problem

hifi-gustavo:
Build console output for commit: 34d24b947e61ee08d8184854ec81f0eb7563ad47 available at: https://gist.github.com/5389be9a99de4362fa9719c2e6514da0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9105-34d24b9.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9105-34d24b9.dmg

sethalves:
further test -- import this raygun and use it to test raypicking in various situations:

{
    "Entities": [
        {
            "collidesWith": "static,dynamic,kinematic,otherAvatar,",
            "collisionMask": 23,
            "collisionsWillMove": 1,
            "dimensions": {
                "x": 0.118687704205513,
                "y": 0.18790049850940704,
                "z": 0.26457399129867554
            },
            "dynamic": 1,
            "gravity": {
                "x": 0,
                "y": -0.5,
                "z": 0
            },
            "modelURL": "http://23.253.109.180/~seth/hifi/raygun/raygun.obj",
            "name": "raygun",
            "rotation": {
                "w": 0.63811707496643066,
                "x": 0.30449378490447998,
                "y": -0.30458533763885498,
                "z": -0.63817811012268066
            },
            "script": "http://23.253.109.180/~seth/hifi/raygun/raygun.js",
            "shapeType": "box",
            "type": "Model",
            "userData": "{\"grabbableKey\":{\"grabbable\":true},\"wearable\":{\"joints\":{\"LeftHand\":[{\"x\":-0.06589794903993607,\"y\":0.14683279395103455,\"z\":0.030722394585609436},{\"x\":-0.5422520041465759,\"y\":0.47700217366218567,\"z\":0.5084763169288635,\"w\":0.46891868114471436}],\"RightHand\":[{\"x\":0.06325198709964752,\"y\":0.2050173580646515,\"z\":0.03773807734251022},{\"x\":0.41833728551864624,\"y\":0.4624304473400116,\"z\":0.5238471031188965,\"w\":-0.5799986720085144}]}}}"
        }
    ],
    "Version": 65
}

hifi-gustavo:
Build console output for commit: 104cbbff7e4ba3e09f38edb02df1ee7de228b4d4 available at: https://gist.github.com/5389be9a99de4362fa9719c2e6514da0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9105-104cbbf.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9105-104cbbf.dmg

hifi-gustavo:
Build console output for commit: 1b20cc8bd08a16896b65fc80bbf91fca3d5d5742 available at: https://gist.github.com/5389be9a99de4362fa9719c2e6514da0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9105-1b20cc8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9105-1b20cc8.dmg

cozza13:
tested with the ray gun. Approved

PR #9107, opened 2016-11-22
Audio fix for Windows 7

kencooke:
Refactored the audio device format detection.

Qt doesn't actually return the internal mix format on Windows 7, so fallback code is used.
Android and Mac OSX now use the native (hardware) format when possible.
Windows 8/10 using WASAPI is unchanged.
Linux using ALSA is unchanged.

### DIFF
--- a/libraries/audio-client/src/AudioClient.cpp
+++ b/libraries/audio-client/src/AudioClient.cpp
@@ -375,41 +375,53 @@ QAudioDeviceInfo defaultAudioDeviceForMode(QAudio::Mode mode) {
     return (mode == QAudio::AudioInput) ? QAudioDeviceInfo::defaultInputDevice() : QAudioDeviceInfo::defaultOutputDevice();
 }
 
+// attempt to use the native sample rate and channel count
+bool nativeFormatForAudioDevice(const QAudioDeviceInfo& audioDevice,
+                                QAudioFormat& audioFormat) {
+
+    audioFormat = audioDevice.preferredFormat();
+
+    audioFormat.setCodec("audio/pcm");
+    audioFormat.setSampleSize(16);
+    audioFormat.setSampleType(QAudioFormat::SignedInt);
+    audioFormat.setByteOrder(QAudioFormat::LittleEndian);
+
+    if (!audioDevice.isFormatSupported(audioFormat)) {
+        qCDebug(audioclient) << "WARNING: The native format is" << audioFormat << "but isFormatSupported() failed.";
+        return false;
+    }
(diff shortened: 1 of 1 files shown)


kencooke:
## Test Plan

Using this build of Interface..
1) Verify that audio works again on Windows 7
2) Verify that audio still works on Windows 8 or 10

If any problems, please include the portion of your log where the audio devices are opened.

hifi-gustavo:
Build console output for commit: d69d77d0eed12bc939e667bca0677a12de9d58ae available at: https://gist.github.com/b9696559a2844aea892cd4b7f6f25d9d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9107-d69d77d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9107-d69d77d.dmg

libraries/audio-client/src/AudioClient.cpp
```diff
@@ -375,41 +375,53 @@ QAudioDeviceInfo defaultAudioDeviceForMode(QAudio::Mode mode) {
...
+    // Mac OSX returns the preferred CoreAudio format
+    if (nativeFormatForAudioDevice(audioDevice, adjustedAudioFormat)) {
+        return true;
+    }
+#endif
+
+#if defined(Q_OS_WIN)
+    if (IsWindows8OrGreater()) {
+        // On Windows using WASAPI shared-mode, returns the internal mix format
+        if (nativeFormatForAudioDevice(audioDevice, adjustedAudioFormat)) {
```
huffman: Not sure if this is an issue, but: if nativeFormatForAudioDevice detects an unsupported sample rate, it returns false, but we will still continue below to possibly return true from `adjustedFormatForAudioDevice`. Previously we would always return false from `adjustedFormatForAudioDevice` if the sample rate was unsupported.

Nex-Pro:
Testing this for Win 10

libraries/audio-client/src/AudioClient.cpp
kencooke: I decided to let all cases (where native fails) also try the fallback code, since it should be harmless. There was one report of WASAPI audio drivers not returning the actual mix format.

Nex-Pro:
### Pass for me. 

Did not notice any issues either with avatars or audio injectors. Usual problems with some avatars having bandwidth issues but we confirmed it was the individual rather than this PR.

Nex-Pro:
@ZappoMan  You approved 3 seconds before I posted ... MINDREADER spookiness !

PR #9110, opened 2016-11-23
Capture original location in snapshots

howard-stearns:
Capture correct href and domain at start of snapshot, not the end, because it matters for time recordings such as animated .gif.
Also allow movement during recording.

(re-) fixes https://highfidelity.fogbugz.com/f/cases/2161/snapshots-recorded-for-wrong-place

### DIFF
--- a/interface/src/ui/Snapshot.cpp
+++ b/interface/src/ui/Snapshot.cpp
@@ -151,13 +151,17 @@ QFile* Snapshot::savedFileForSnapshot(QImage & shot, bool isTemporary) {
     return imageTempFile;
 }
 
-void Snapshot::uploadSnapshot(const QString& filename) {
+void Snapshot::uploadSnapshot(const QString& filename, const QUrl& href) {
 
     const QString SNAPSHOT_UPLOAD_URL = "/api/v1/snapshots";
-    // Alternatively to parseSnapshotData, we could pass the inWorldLocation through the call chain. This way is less disruptive to existing code.
-    SnapshotMetaData* snapshotData = Snapshot::parseSnapshotData(filename);
-    SnapshotUploader* uploader = new SnapshotUploader(snapshotData->getURL(), filename);
-    delete snapshotData;
+    SnapshotUploader* uploader;
+    if (href.isEmpty()) {
+        SnapshotMetaData* snapshotData = Snapshot::parseSnapshotData(filename);
+        uploader = new SnapshotUploader(snapshotData->getURL(), filename);
+        delete snapshotData;
+    } else {
+        uploader = new SnapshotUploader(href, filename);
+    }
     
     QFile* file = new QFile(filename);
     Q_ASSERT(file->exists());
--- a/interface/src/ui/Snapshot.cpp
+++ b/interface/src/ui/Snapshot.cpp
@@ -63,8 +63,6 @@ SnapshotMetaData* Snapshot::parseSnapshotData(QString snapshotPath) {
 
         // parsing URL
         url = QUrl(shot.text(URL), QUrl::ParsingMode::StrictMode);
-    } else if (snapshotPath.right(3) == "gif") {
-        url = QUrl(DependencyManager::get<AddressManager>()->currentShareableAddress());
     } else {
         return NULL;
     }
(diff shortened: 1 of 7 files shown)


howard-stearns:
The .cpp code changes are to allow the href to passed in on upload, because by the time that happens in .gif, we could be elsewhere. (.jpg embeds in file, and has zero time for movement, but both are treated uniformly now.)

The .js code changes are of three kinds:
1. Store the captured href in the SnapshotReview popup, in case there are several open for a long time before submitting, with location (and even domain changes) in between.
2. Work around an issue in which popups during clearOverlayWhenDriving can bork your state. (Turns that off during the recording, and restores it after the snapshot review popup.)
3. Capture the href and domainId at the start of recording, rather than the end.

hifi-gustavo:
Build console output for commit: 414371d1fc5b3b53e3742b14a5181f9c9807b55c available at: https://gist.github.com/4e1a4a055fef91c01e4c48c792d9a987<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9110-414371d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9110-414371d.dmg

howard-stearns:
Test plan:
- From a shareable place (e.g., Welcome or Playa), with "Settings->Avatar->Clear overlays when moving" and "Settings->General->Take Animated GIF Snapshot with HUD Button" both true, take a snapshot using the toolbar "SNAP" button.  Move to a recognizably different place during the recording. 
 - The Snapshot Review dialog should popup to tell you that recording is complete.  Do not share them yet.
 - Go to a different place (e.g. a different domain altogether).
 - Share both the animated picture and the still snapshot. 
 - Find your snapshots in the feed. (E.g., open the GoTo address bar, select "SNAPS", and start typing your username. They should appear.)  Click on both your snaps in turn. They should each take you back to the place in which the pictures were started.
- Repeat with different combinations, such as setting  "Settings->Avatar->Clear overlays when moving" to false, or taking another set of pictures from a different domain while the first set of pictures are not yet shared.

interface/src/ui/Snapshot.cpp
```diff
@@ -151,13 +151,17 @@ QFile* Snapshot::savedFileForSnapshot(QImage & shot, bool isTemporary) {
...
 
     const QString SNAPSHOT_UPLOAD_URL = "/api/v1/snapshots";
-    // Alternatively to parseSnapshotData, we could pass the inWorldLocation through the call chain. This way is less disruptive to existing code.
-    SnapshotMetaData* snapshotData = Snapshot::parseSnapshotData(filename);
-    SnapshotUploader* uploader = new SnapshotUploader(snapshotData->getURL(), filename);
-    delete snapshotData;
+    SnapshotUploader* uploader;
+    if (href.isEmpty()) {
+        SnapshotMetaData* snapshotData = Snapshot::parseSnapshotData(filename);
+        uploader = new SnapshotUploader(snapshotData->getURL(), filename);
```
davidkelly: should we protect against the `Snapshot::parseSnapshotData(...)` call from returning null here?  Seems worth an if, just in case.

howard-stearns: yup. wilco

hifi-gustavo:
Build console output for commit: 4e0c6e5d0f793c31cb242b4099e5da073bac5e20 available at: https://gist.github.com/4e1a4a055fef91c01e4c48c792d9a987<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9110-4e0c6e5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9110-4e0c6e5.dmg

interface/src/ui/Snapshot.cpp
```diff
@@ -151,13 +149,19 @@ QFile* Snapshot::savedFileForSnapshot(QImage & shot, bool isTemporary) {
...
 
     const QString SNAPSHOT_UPLOAD_URL = "/api/v1/snapshots";
-    // Alternatively to parseSnapshotData, we could pass the inWorldLocation through the call chain. This way is less disruptive to existing code.
-    SnapshotMetaData* snapshotData = Snapshot::parseSnapshotData(filename);
-    SnapshotUploader* uploader = new SnapshotUploader(snapshotData->getURL(), filename);
-    delete snapshotData;
+    QUrl url = href;
+    if (url.isEmpty()) {
+        SnapshotMetaData* snapshotData = Snapshot::parseSnapshotData(filename);
+        url = snapshotData->getURL();
```
davidkelly: this will be problematic if `Snapshot::parseSnapshotData` returns null, which I believe it can do under some circumstances.

howard-stearns: you're right of course. fixing.

hifi-gustavo:
Build console output for commit: bf0f56c138e2577c0e6cd3ba50197e9e14a4d64d available at: https://gist.github.com/4e1a4a055fef91c01e4c48c792d9a987<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9110-bf0f56c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9110-bf0f56c.dmg

ZappoMan:
code changed since last test, needs retest

PR #9099, opened 2016-11-19
first cut at personal space feature

ZappoMan:

PR #9114, opened 2016-11-23
bubble state

howard-stearns:
Changes to ignore-bubble's on/off state are observable to Javascript.
Default for new users is "on" (per Philip).

Fixes https://highfidelity.fogbugz.com/f/cases/2348/bubble-button-doesn-t-track-changes-to-state

### DIFF
--- a/libraries/networking/src/NodeList.cpp
+++ b/libraries/networking/src/NodeList.cpp
@@ -751,6 +751,7 @@ bool NodeList::sockAddrBelongsToDomainOrNode(const HifiSockAddr& sockAddr) {
 }
 
 void NodeList::ignoreNodesInRadius(float radiusToIgnore, bool enabled) {
+    bool isEnabledChange = _ignoreRadiusEnabled.get() != enabled;
     _ignoreRadiusEnabled.set(enabled);
     _ignoreRadius.set(radiusToIgnore);
 
@@ -759,6 +760,9 @@ void NodeList::ignoreNodesInRadius(float radiusToIgnore, bool enabled) {
     }, [this](const SharedNodePointer& destinationNode) {
         sendIgnoreRadiusStateToNode(destinationNode);
     });
+    if (isEnabledChange) {
+        emit ignoreRadiusEnabledChanged(enabled);
+    }
 }
 
 void NodeList::sendIgnoreRadiusStateToNode(const SharedNodePointer& destinationNode) {
@@ -885,10 +889,4 @@ void NodeList::muteNodeBySessionID(const QUuid& nodeID) {
         qWarning() << "NodeList::muteNodeBySessionID called with an invalid ID or an ID which matches the current session ID.";
 
     }
-}
-
-void NodeList::toggleIgnoreRadius() {
-    bool isIgnored = !getIgnoreRadiusEnabled();
-    ignoreNodesInRadius(getIgnoreRadius(), isIgnored);
-    emit ignoreRadiusEnabledChanged(isIgnored);
 }
--- a/libraries/networking/src/NodeList.cpp
+++ b/libraries/networking/src/NodeList.cpp
@@ -886,3 +886,9 @@ void NodeList::muteNodeBySessionID(const QUuid& nodeID) {
 
     }
 }
+
+void NodeList::toggleIgnoreRadius() {
+    bool isIgnored = !getIgnoreRadiusEnabled();
+    ignoreNodesInRadius(getIgnoreRadius(), isIgnored);
+    emit ignoreRadiusEnabledChanged(isIgnored);
+}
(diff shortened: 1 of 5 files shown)


libraries/networking/src/NodeList.cpp
```diff
@@ -886,3 +886,9 @@ void NodeList::muteNodeBySessionID(const QUuid& nodeID) {
 
     }
 }
+
+void NodeList::toggleIgnoreRadius() {
+    bool isIgnored = !getIgnoreRadiusEnabled();
+    ignoreNodesInRadius(getIgnoreRadius(), isIgnored);
+    emit ignoreRadiusEnabledChanged(isIgnored);
```
ZappoMan: This emit should be in the `ignoreNodesInRadius()` method instead... otherwise you only get the signal when the toggle method is called - and there are actually multiple paths to change the state.

hifi-gustavo:
Build console output for commit: 8b9efb461e5711b706bc7fb6772890462869a57d available at: https://gist.github.com/af15bcab61c9faf0de4014942d8ff09c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9114-8b9efb4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9114-8b9efb4.dmg

libraries/networking/src/NodeList.cpp
howard-stearns: good catch. changing...

libraries/networking/src/NodeList.h
```diff
@@ -154,7 +155,7 @@ private slots:
     tbb::concurrent_unordered_set<QUuid, UUIDHasher> _ignoredNodeIDs;
 
     void sendIgnoreRadiusStateToNode(const SharedNodePointer& destinationNode);
-    Setting::Handle<bool> _ignoreRadiusEnabled { "IgnoreRadiusEnabled", false };
+    Setting::Handle<bool> _ignoreRadiusEnabled { "IgnoreRadiusEnabled", true };
```
howard-stearns: I'll re-discuss with Philip next week. I hadn't questioned the change because it seemed obvious to me:

1) The reason the feature was added (as a critical fix, no less), was to keep new users from having a bad experience.  Finding a strange button in the strange HUD while they're being harassed is too much to ask of a new user.
2) We always make "protect me" choices for new users, even if that limits the possibilities: Snap turn when in HMD, not Advanced Movement For Hand Controllers. Constrain Toolbar Position to Horizontal Center, Clear overlays when moving, no default name, no need to pick an avatar before starting, ... (*)

What am I missing?
---
*: The only possible exception to this that I can think of is starting off unmuted, which I feel could be argued either way as to which way is "safer".

ZappoMan: Ok - I just asked again. turn it off.

hifi-gustavo:
Build console output for commit: b37feb20deb15c0c01ef8f281ee7d7742078bc9e available at: https://gist.github.com/af15bcab61c9faf0de4014942d8ff09c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9114-b37feb2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9114-b37feb2.dmg

OmegaHeron:
Testing Windows 10

Passes tests from #9099 

Clean start (no interface.ini present) privacy mode is **On** -- I lost track of whether it should be on or off on a clean start, but, I will say starting with it on seems to be catastrophically fraught with peril...

Users.toggleIgnoreRadius() does toggle bubble state, hud and world interaction matches toggle.  

General commentary, but, I'm not sure toggleIgnoreRadius really seems right name for function - togglePrivacy or togglePrivacyMode or toggleUserPrivacyMode seems more clear to me.

PhilipRosedale:
@OmegaHeron What do you mean by "fraught with peril"??  Why is it bad to have it on?  Email me, let's discuss, I don't want to make a mistake here.

RyanDowne:
Let's be mindful as we iterate on this feature. With the addition of a set radius we are already opening ourselves up to a situation where someone can use it to not just filter out people that approach them but those they just wish not to see. A natural next step would be to make "friends" ignore the bubble all together. This quite literally creates a chamber - or to be more specific an "echo-chamber" where you only need to see and hear the people that you like or that share your values. I worry that this is something we should be careful about doing too quickly. There have been several discussions as of late, especially after the election, that center around how controls such as these do us all a disservice even though they temporarily mitigate issues.

howard-stearns:
Summary: issues raised are now in design doc. Passes for current intentions.

PR #9119, opened 2016-11-28
Add the robustness flag to the GL context in 4.5

jherico:
Add the robustness flag to the OpenGL context when it's a 4.5 context.  This should prevent crashes caused by invalid vertex or index buffer access, instead leaving the results of such invalid draws as undefined.  

## Testing

Testing should focus on the recent crashes in the nVidia drivers, which neither Sam nor I have been able to reproduce.

### DIFF
--- a/libraries/gl/src/gl/Context.cpp
+++ b/libraries/gl/src/gl/Context.cpp
@@ -274,11 +274,14 @@ void Context::create() {
         contextAttribs.push_back(WGL_CONTEXT_PROFILE_MASK_ARB);
         contextAttribs.push_back(WGL_CONTEXT_CORE_PROFILE_BIT_ARB);
         contextAttribs.push_back(WGL_CONTEXT_FLAGS_ARB);
-        if (enableDebugLogger) {
-            contextAttribs.push_back(WGL_CONTEXT_DEBUG_BIT_ARB);
-        } else {
-            contextAttribs.push_back(0);
+        int flags = 0;
+        if (_version >= 0x0405) {
+            flags |= WGL_CONTEXT_ROBUST_ACCESS_BIT_ARB;
         }
+        if (enableDebugLogger) {
+            flags |= WGL_CONTEXT_DEBUG_BIT_ARB;
+        }
+        contextAttribs.push_back(flags);
         contextAttribs.push_back(0);
         auto shareHglrc = PRIMARY ? PRIMARY->_hglrc : 0;
         _hglrc = wglCreateContextAttribsARB(_hdc, shareHglrc, &contextAttribs[0]);


hifi-gustavo:
Build console output for commit: f30f9f49bd8fcb70010b7a3a369c027969941353 available at: https://gist.github.com/29b244238051cad7bafd2b884209c5a7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9119-f30f9f4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9119-f30f9f4.dmg

hifi-gustavo:
Build console output for commit: 5d7c7de49356dfb63aa9682fcbdb0fff45286386 available at: https://gist.github.com/29b244238051cad7bafd2b884209c5a7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9119-5d7c7de.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9119-5d7c7de.dmg

Nex-Pro:
@jherico I've had this running for some hours including updating to latest NVidia driver with no problems. Is there some specific content we should test against?

jherico:
@Nex-Pro We've been seeing lots of reports of crashes with polyline entities (created by writing on a whiteboard) with the current beta build.  If you can reproduce a crash in the beta build and then be reasonably certain that you're not seeing the same crash on the same content with this build, that would be considered success.

Nex-Pro:
@jherico Ok

Nex-Pro:
@jherico I'm dropping this due to inability to test. The only place I can try this is at localhost.

I draw many polylines on the whiteboard then exit hifi.

Restarting I appear near the whiteboard but no crash but also the polylines are not visible.  On one occasion I started drawing some more and after some time the lines that had gone missing re-appeared.

On other occasions the polylines never appeared.

As I am unable to get a reliable sequence of  draw - exit - start - observe I am dropping this.

cozza13:
from Balpien https://forums.highfidelity.com/t/beta-release-28-hotfix-update/11869/22?u=chris

jherico:
@Nex-Pro is that behavior only happening with this PR, and not with the current development build?

cozza13:
marked this as a QA fail based on comments from Balpien. Will wait for his video

jherico:
@cozza13 actually I think that this might indicate success, and prove that the problem is isolated to polyline.  The point of the robustness flag is to prevent things that would otherwise cause a crash because of screwups in the vertex or index buffers.  However, the result of drawing with the bad data is undefined.  So I would expect that if the polyline vertex and index buffer handling is bad, that the result would either be no rendering of the polyline or weird rendering, but no crash.  Technically it could do anything, but the fact that it's rendering 'weirdly' instead of crashing indicates that the flag is in fact doing it's job.  

However, I'm wary of enabling the flag globally as I've done in this PR, because it probably has a performance impact.  I'm going to continue investigating.

PR #9126, opened 2016-11-29
Working on OpenGL debuggability

jherico:
This is to support debugging issues where we are seeing crashes in the OpenGL drivers.  It does a number of different things.

* Create a shared location for global property names for items stored in the application (this is the easiest way to ensure that a given object is visible across DLL boundaries)
* Store the fact that we crashed on the last run as a global boolean
* Automatically enable OpenGL debug logging if we crashed on the previous run
* Fix a bug where external display plugins could not create shaders properly due to a missing callback for compiling the shaders
* Incomplete work to allow code to force the logging system to immediately flush any pending log messages to disk
** This is to allow code that performs logging that is likely leading up to a crash to ensure that the log file contains the critical information
** Unfortunately, so far it doesn't work, and leads to a deadlock I haven't been able to identify, so it's only use is currently commented out.  
* Update the render perf test to support a simple entity simulation mechanism (similar to that used in a server). 
** This is required in order to allow the render perf tool to render particle and polyline entities, which only produce output after an update
* Restore the proper pipeline setup code for rendering hand lasers

## Testing

This should not affect application behavior.  

Testers should explicitly verify that the laser pointers when aiming at the UI layer in an HMD still works.  

Additionally, if you trigger a crash intentionally (via the developer / crash menu) then on the next run you should see output in the log similar to this during startup

` GL Error:  GL_INVALID_ENUM error generated. Operation is not valid from the core profile.`

This particular error always occurs at least one or more times during startup as part of initializing the GLEW library and doesn't actually indicate a problem, but should serve to verify that the automated GL logging after crash mechanism is working .

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -45,6 +45,7 @@
 
 #include <gl/QOpenGLContextWrapper.h>
 
+#include <shared/GlobalAppProperties.h>
 #include <ResourceScriptingInterface.h>
 #include <AccountManager.h>
 #include <AddressManager.h>
@@ -536,7 +537,8 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer, bo
     _maxOctreePPS(maxOctreePacketsPerSecond.get()),
     _lastFaceTrackerUpdate(0)
 {
-    setProperty("com.highfidelity.launchedFromSteam", SteamClient::isRunning());
+    setProperty(hifi::properties::STEAM, SteamClient::isRunning());
+    setProperty(hifi::properties::CRASHED, _previousSessionCrashed);
 
     _runningMarker.startRunningMarker();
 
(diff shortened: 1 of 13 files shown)


hifi-gustavo:
Build console output for commit: 3b07e31eab766253a22a83e917dc2a911612e4d7 available at: https://gist.github.com/febcb0bbcd9604fa14362aa529e71fc9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9126-3b07e31.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9126-3b07e31.dmg

hifi-gustavo:
Build console output for commit: 63b3b7250d4c4484f177fba91fdfcd48a9f02a1f available at: https://gist.github.com/febcb0bbcd9604fa14362aa529e71fc9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9126-63b3b72.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9126-63b3b72.dmg

Nex-Pro:
### PASS

tested on win10 and vive:

Llasers work fine with UI elements.: **SUCCESS**

GL logging after crash: **SUCCESS**

got this error in the log by triggering a crash:

[11/30 17:43:45] [DEBUG] [hifi.glLogging] OpenGL:  GL_INVALID_ENUM error generated. Operation is not valid from the core profile.

PR #9128, opened 2016-11-29
Vec3 and shader helpers

Polyrhythm:
**Vec3.getAngle** (in radians) - handy for telling if an entity is facing within EPSILON radians of a given direction

**fbm** - classic noise algo for shaders that produces awesome cloud-like effects. Will probably be very popular for hifi modders for messing with skyboxes. 

**QA**:
Add a sphere entity with the following `userData`:

```
{
  "ProceduralEntity": {
    "version": 2,
    "shaderUrl": "https://raw.githubusercontent.com/Polyrhythm/hifi-scripts/master/clouds.fs"
  }
}
```

You should get a sphere that looks like this:

fbm-noise

This tests the shader helpers I added.

### DIFF
--- a/libraries/procedural/src/procedural/ProceduralCommon.slf
+++ b/libraries/procedural/src/procedural/ProceduralCommon.slf
@@ -264,6 +264,37 @@ float snoise(vec2 v) {
     return 130.0 * dot(m, g);
 }
 
+// https://www.shadertoy.com/view/lsfGRr
+float hash( float n )
+{
+    return fract(sin(n)*43758.5453);
+}
+
+float noise( in vec2 x )
+{
+    vec2 p = floor(x);
+    vec2 f = fract(x);
+
+    f = f*f*(3.0-2.0*f);
+
+    float n = p.x + p.y*57.0;
+
+    return mix(mix( hash(n+  0.0), hash(n+  1.0),f.x),
+               mix( hash(n+ 57.0), hash(n+ 58.0),f.x),f.y);
+}
+
+// https://www.shadertoy.com/view/MdX3Rr
+// https://en.wikipedia.org/wiki/Fractional_Brownian_motion
+float fbm(in vec2 p) {
+	const mat2 m2 = mat2(0.8, -0.6, 0.6, 0.8);
+	float f = 0.0;
+	f += 0.5000 * noise(p); p = m2 * p * 2.02;
+	f += 0.2500 * noise(p); p = m2 * p * 2.03;
+	f += 0.1250 * noise(p); p = m2 * p * 2.01;
+	f += 0.0625 * noise(p);
+
+	return f / 0.9375;
+}
 
 #define PROCEDURAL 1
 
--- a/libraries/procedural/src/procedural/ProceduralCommon.slf
+++ b/libraries/procedural/src/procedural/ProceduralCommon.slf
@@ -265,33 +265,31 @@ float snoise(vec2 v) {
 }
 
 // https://www.shadertoy.com/view/lsfGRr
-float hash( float n )
-{
-    return fract(sin(n)*43758.5453);
+float hifi_hash(float n) {
+    return fract(sin(n) * 43758.5453);
 }
 
-float noise( in vec2 x )
-{
+float hifi_noise(in vec2 x) {
     vec2 p = floor(x);
     vec2 f = fract(x);
 
-    f = f*f*(3.0-2.0*f);
+    f = f * f * (3.0 - 2.0 * f);
 
-    float n = p.x + p.y*57.0;
+    float n = p.x + p.y * 57.0;
 
-    return mix(mix( hash(n+  0.0), hash(n+  1.0),f.x),
-               mix( hash(n+ 57.0), hash(n+ 58.0),f.x),f.y);
+    return mix(mix(hifi_hash(n +  0.0), hifi_hash(n +  1.0), f.x),
+               mix(hifi_hash(n + 57.0), hifi_hash(n + 58.0), f.x), f.y);
 }
 
 // https://www.shadertoy.com/view/MdX3Rr
 // https://en.wikipedia.org/wiki/Fractional_Brownian_motion
-float fbm(in vec2 p) {
+float hifi_fbm(in vec2 p) {
 	const mat2 m2 = mat2(0.8, -0.6, 0.6, 0.8);
 	float f = 0.0;
-	f += 0.5000 * noise(p); p = m2 * p * 2.02;
-	f += 0.2500 * noise(p); p = m2 * p * 2.03;
-	f += 0.1250 * noise(p); p = m2 * p * 2.01;
-	f += 0.0625 * noise(p);
+	f += 0.5000 * hifi_noise(p); p = m2 * p * 2.02;
+	f += 0.2500 * hifi_noise(p); p = m2 * p * 2.03;
+	f += 0.1250 * hifi_noise(p); p = m2 * p * 2.01;
+	f += 0.0625 * hifi_noise(p);
 
 	return f / 0.9375;
 }
--- a/libraries/script-engine/src/Vec3.cpp
+++ b/libraries/script-engine/src/Vec3.cpp
@@ -83,15 +83,7 @@ glm::vec3 Vec3::fromPolar(float elevation, float azimuth) {
     return fromPolar(v);
 }
 
-float Vec3::angle(const glm::vec3& fromV, const glm::vec3& toV) {
-	glm::vec3 fromVNormalized = glm::normalize(fromV);
-	glm::vec3 toVNormalized = glm::normalize(toV);
-
-	float radians = glm::acos(glm::dot(fromVNormalized, toVNormalized));
-
-	return radians * 180.0f / PI;
+float Vec3::getAngle(const glm::vec3& v1, const glm::vec3& v2) {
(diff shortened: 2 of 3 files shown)


hifi-gustavo:

PhilipRosedale:
ok to test

hifi-gustavo:
Build console output for commit: fe774a853018d515917cd2e0d17376a55c469df5 available at: https://gist.github.com/28c3636aeaccbec82420f1cb3fbe4234<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9128-fe774a8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9128-fe774a8.dmg

ZappoMan:
add to whitelist

libraries/script-engine/src/Vec3.cpp
```diff
@@ -82,3 +82,16 @@ glm::vec3 Vec3::fromPolar(float elevation, float azimuth) {
     glm::vec3 v = glm::vec3(elevation, azimuth, 1.0f);
     return fromPolar(v);
 }
+
+float Vec3::angle(const glm::vec3& fromV, const glm::vec3& toV) {
+	glm::vec3 fromVNormalized = glm::normalize(fromV);
```
ZappoMan: formatting doesn't match coding standard. :)

Polyrhythm: Okay I'm just going to power through that dang doc

PhilipRosedale:
this is cool stuff, @Polyrhythm, thank you!

hifi-gustavo:
Build console output for commit: fe774a853018d515917cd2e0d17376a55c469df5 available at: https://gist.github.com/28c3636aeaccbec82420f1cb3fbe4234<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9128-fe774a8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9128-fe774a8.dmg

libraries/procedural/src/procedural/ProceduralCommon.slf
```diff
@@ -264,6 +264,37 @@ float snoise(vec2 v) {
     return 130.0 * dot(m, g);
 }
 
+// https://www.shadertoy.com/view/lsfGRr
+float hash( float n )
```
jherico: `hash`, `noise`, and `fbm` are common function names in shaders, and many of the existing procedural items probably already have them.  Adding these would cause them to no longer compile.  

Adding these would be reasonable if you rename them to insure no namespace conflicts, perhaps by naming them `hifi_noise` and similar.

libraries/script-engine/src/Vec3.cpp
```diff
@@ -82,3 +82,16 @@ glm::vec3 Vec3::fromPolar(float elevation, float azimuth) {
     glm::vec3 v = glm::vec3(elevation, azimuth, 1.0f);
     return fromPolar(v);
 }
+
+float Vec3::angle(const glm::vec3& fromV, const glm::vec3& toV) {
```
jherico: I would suggest exposing `dot`, rather than `angle` (or at least in addition to), allowing the client to do the acos in JS.  If you do want to keep `angle`, I'd suggest that you be explicit about the type of the returned value and name it `angleDegrees`, or have both `angleDegrees` and `angleRadians`.

cozza13:
Went through the QA steps and it looks good.

libraries/procedural/src/procedural/ProceduralCommon.slf
Polyrhythm: 👍 will fix up the namespacing

libraries/procedural/src/procedural/ProceduralCommon.slf
```diff
@@ -264,6 +264,37 @@ float snoise(vec2 v) {
...
+{
+    return fract(sin(n)*43758.5453);
+}
+
+float noise( in vec2 x )
+{
```
AndrewMeadows: I think the K&R bracing preference of the coding standard still applies to *.slh files, so it should be:
```
float foo(in arg) {
    return something;
}
```
Rather than:
```
float foo(in arg)
{
    return something;
}
```

https://github.com/highfidelity/hifi/wiki/Coding-Standard

libraries/script-engine/src/Vec3.cpp
AndrewMeadows: `Vec3.dot()` already exists.  I suppose `Vec3.angle()` is still useful.

libraries/script-engine/src/Vec3.cpp
```diff
@@ -82,3 +82,16 @@ glm::vec3 Vec3::fromPolar(float elevation, float azimuth) {
...
+	glm::vec3 toVNormalized = glm::normalize(toV);
+
+	float radians = glm::acos(glm::dot(fromVNormalized, toVNormalized));
+
+	return radians * 180.0f / PI;
+}
+
+glm::vec3 Vec3::lerp(const glm::vec3& v1, const glm::vec3& v2, float t) {
+	return v1 * (1.0f - t) + v2 * t;
+}
```
AndrewMeadows: `Vec3.lerp() `is a duplicate of `Vec3.mix()` however I prefer the `lerp` name.  I think we provided `mix` because that is what the glm library provides:  `glm::mix()`.  I guess we could provide both?

Normally I would make a comment about how `lerp` should probably just call `glm::mix()` however I believe we really want to implement the `Vec3` object completely in JS (it's faster!), so I will shut up.

libraries/script-engine/src/Vec3.cpp
```diff
@@ -82,3 +82,16 @@ glm::vec3 Vec3::fromPolar(float elevation, float azimuth) {
...
 }
+
+float Vec3::angle(const glm::vec3& fromV, const glm::vec3& toV) {
+	glm::vec3 fromVNormalized = glm::normalize(fromV);
+	glm::vec3 toVNormalized = glm::normalize(toV);
+
+	float radians = glm::acos(glm::dot(fromVNormalized, toVNormalized));
+
+	return radians * 180.0f / PI;
+}
```
AndrewMeadows: Vec3.angle() returns _degrees_.

*cry*

For those who don't know it: radians are the natural measure of angle.

AndrewMeadows: On a more serious note: our JS API is inconsistent.  Most methods are named like `foo()` and return radians but a few are named like `foo()` and return degrees (for example: `MyAvatar.getBodyYaw()`).  Then some are named like `fooRadians()` or `fooDegrees()` and return units as named.

This bothers me and I would have plowed through them a long time ago and changed everything to be either `foo()` or `fooDegrees()`, returning radians and degrees accordingly, but could not get consensus.  Something still needs to be done.

Polyrhythm: Sounds like we should settle on radians by default and label something as degrees otherwise? I can make the change with my PR.

libraries/script-engine/src/Vec3.cpp
Polyrhythm: I was thinking about how it would be nice to simply do all the helpers in JS and not have this middleman C++ stuff that's just kind of passing stuff around. Plus, if Vec3 objects can be constructed in JS, then they gain some inspectable properties instead of just being a vanilla object.

libraries/script-engine/src/Vec3.cpp
Polyrhythm: Also - worth nothing that the purpose of these helpers is merely convenience, and that convenience will be most readily appreciated by scripters who  aren't super knowledgeable on linear algebra or trig. Degrees are pretty intuitive for most people in most aspects, so probably worth at least having an option for people to make it more user friendly.

libraries/script-engine/src/Vec3.cpp
AndrewMeadows: I have written Vec3 and Quat implementations in JS that I use for my own JS scripts.  I gave those implementations to @huffman a few months ago because he was thinking about officially moving them to JS but it hasn't happened yet.

huffman: Yes, this is something that hasn't been a high priority but would be a welcome change IMO. I have the vec3/quat implementations plus a few polyfill things I'd like to get in there. We're inbetween projects right now so I'll try to make time to look into it this week.

jherico: I'll just leave this here:  http://s3.amazonaws.com/DreamingContent/scripts/Easings.js

Polyrhythm:
I've removed `Vec3.lerp` in deference to `Vec3.mix`, namespaced the shader helper functions by prepending `hifi_`, now return radians from `Vec3.getAngle`, and adjust naming conventions with everything to comply with the coding standards.

I also added two issues (visible above) which came out of the discussions in this PR.

hifi-gustavo:
Build console output for commit: 7b0756c74529274f33f157e4ac450ad83079ee93 available at: https://gist.github.com/28c3636aeaccbec82420f1cb3fbe4234<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9128-7b0756c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9128-7b0756c.dmg

PR #9130, opened 2016-11-30
Threaded Audio

zzmp:
Adds multithreading support for the audio-mixer assignment-client.


hifi-gustavo:
Build console output for commit: 2c42bb0aa8efc1dca6c307eaa44fd099680f5558 available at: https://gist.github.com/ed1a7d4bd28d1a3ade51a74ff1508aec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-2c42bb0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-2c42bb0.dmg

assignment-client/src/audio/AudioMixerSlavePool.cpp
```diff
@@ -0,0 +1,149 @@
...
+void AudioMixerSlavePool::setNumThreads(int numThreads) {
+    std::unique_lock<std::mutex> lock(_mutex);
+
+    // ensure slave are not running
+    assert(!_running);
+
+    // clamp to allowed size
+    {
+        // idealThreadCount returns -1 if cores cannot be detected - cast it to a large number
+        int maxThreads = (int)((unsigned int)QThread::idealThreadCount());
```
kencooke: This won't work (maxThreads will still be -1) because reinterpret cast doesn't change the value. You need something like:
```maxThreads = (maxThreads < 0) ? 0x7fffffff : maxThreads);```

**Comment by hifi-gustavo** [comment]
Build console output for commit: 800744a139af5839a014e37963eedf55392ddf1b available at: https://gist.github.com/ed1a7d4bd28d1a3ade51a74ff1508aec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-800744a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-800744a.dmg

**Comment by zzmp** [comment]
Fixed a few bugs, still have to give definite metrics, expect them tomorrow morning. This is ready for review, though.

**Comment by ZappoMan** [comment]
damn you jenkins!

**Comment by ZappoMan** [comment]
test this please

**Comment by hifi-gustavo** [comment]
Build console output for commit: 800744a139af5839a014e37963eedf55392ddf1b available at: https://gist.github.com/ed1a7d4bd28d1a3ade51a74ff1508aec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-800744a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-800744a.dmg

**Comment by hifi-gustavo** [comment]
Build console output for commit: 488a60e370c20c58f84c5dcbc1681df9536f81fa available at: https://gist.github.com/ed1a7d4bd28d1a3ade51a74ff1508aec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-488a60e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-488a60e.dmg

**Comment by hifi-gustavo** [comment]
Build console output for commit: bbfe58c56dedce40600e2802dc443fd64710a02c available at: https://gist.github.com/ed1a7d4bd28d1a3ade51a74ff1508aec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-bbfe58c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-bbfe58c.dmg

**Comment by hifi-gustavo** [comment]
Build console output for commit: d899391a1ad64bed394ea01a15468873f3f67d58 available at: https://gist.github.com/ed1a7d4bd28d1a3ade51a74ff1508aec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-d899391.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-d899391.dmg

**Comment by zzmp** [comment]
## Timing

### Setup
- HEAD (d899391) against BASE (master branch with timings added, 6bedad6)
- Server is running with no agents
- Agents are running on a separate machine as `assignment-client -t 2 -a SERVER -n N`
- Both machines are EC2 m4.4xlarge, availability zone us-east-2c, Ubuntu 16.04 Server
  - `assignment-client` reports `idealThreadCount: 16`, and audio threading is set to automatic
- Every agent runs [this bot script](https://github.com/highfidelity/hifi/files/627723/bot.txt)

### Results
- Frame mix timings are in milliseconds averaged over ten seconds at load
- `N/A` denotes that the audio-mixer build could not sustain the load, and began to throttle
  - This can be seen in the node stats, `performance_throttling_ratio`

*Frame mix timings (ms)*

| # Agents | HEAD | BASE |
| - | - | - |
| 1 | 0.2 | <0.1 |
| 2 | 0.3 | 0.1 |
| 10 | 0.4 | 0.9 |
| 20 | 0.7 | 3.1 |
| 30 | 1.2 | 6.8 |
| 40 | 1.8 | N/A |
| 50 | 3.0 | N/A |
| 80 | 7.0 | N/A |
| 90 | N/A | N/A |

## QA

- Run a server with this audio-mixer on a multi-core machine
- Ensure audio settings still work as expected
- Ensure the domain can host upwards of 30 people, without audio clipping issues
  - If you have issues, please note the machine that you used, especially the number of cores and bandwidth available

If anything goes wrong with the audio-mixer, please include a log. If it appears to freeze, and you can capture a callstack, please do; ping me on slack if this occurs and you need instructions.

**Comment by zzmp** [comment]
We noticed a problem where gradual addition of agents works, but adding a large number of agents at once ruins audio quality. Keeping this DNM until that is resolved.

**Comment by howard-stearns** [comment]
If it helps:
- This may be related to https://highfidelity.fogbugz.com/f/cases/2095/ ?
- Interface script scripts/developer/tests/performance/summon.js avoids that bug by phasing in avatars around you, of which a rotating group of 4 make noise. This phasing WORKS (last I checked) on a localhost domain running on an HMD-capable box. (The script will tell you that you need to run 24 instances of the AC script http://hifi-content.s3.amazonaws.com/howard/scripts/tests/performance/crowd-agent.js)
- If this fixes the problem, you should be able to change SPREAD_TIME_MS to zero in the interface script (no need to change or restart the ac script).

**Comment by zzmp** [comment]
Thanks @howard-stearns. That bug should be resolved.

In this case, when a large number of agents are started at once, the audio stutters across all streams, but no streams seem to be omitted.

**Comment by ZappoMan** [comment]
@zzmp - in the past (~2 years ago) - I did a lot of optimization work on the entity server to deal with burst connections, I suspect it has bit rotted a bunch since then... but the issue was likely related to lock contention and other performance issues in the lower level node list code. If you can confirm that the burst behavior is not a regression from master... then I think we can reasonably proceed with your PR.

**Comment by zzmp** [comment]
@ZappoMan I've tested multiple times on HEAD and BASE with 20 Agents. HEAD always comes out stuttery, BASE always comes out clean. It seems to be a regression.

**Comment by ZappoMan** [comment]
@zzmp - ok - seems like you need to figure that out and fix it.

**Comment by zzmp** [comment]
There is an obvious error in the `upstream` stats (under `z_listeners`) in HEAD, so I am treating the ring buffers incorrectly somewhere in the new codepaths. Progress :)!

File: assignment-client/src/audio/AudioMixerSlave.cpp [hunk 2]
```diff
@@ -0,0 +1,521 @@
...
+            for (auto& streamPair : streamsCopy) {
+                auto otherNodeStream = streamPair.second;
+                if (*otherNode != *node || otherNodeStream->shouldLoopbackForNode()) {
+                    addStreamToMix(*nodeData, otherNode->getUUID(), *nodeAudioStream, *otherNodeStream);
+                }
+            }
+        }
+    });
+
+    // use the per listner AudioLimiter to render the mixed data...
```
zzmp: typo

assignment-client/src/audio/AudioMixerSlave.cpp
```diff
@@ -0,0 +1,521 @@
...
+    nodeData->audioLimiter.render(_mixedSamples, _clampedSamples, AudioConstants::NETWORK_FRAME_SAMPLES_PER_CHANNEL);
+
+    // check for silent audio after the peak limitor has converted the samples
```
zzmp: s/limitor/limiter

assignment-client/src/audio/AudioMixerSlavePool.cpp
```diff
@@ -0,0 +1,191 @@
...
+#ifdef AUDIO_SINGLE_THREADED
+    functor(slave);
+#else
+    for (auto& slave : _slaves) {
+        functor(*slave.get());
+    }
+#endif
+}
+
+void AudioMixerSlavePool::start(Lock& lock, ConstIter begin, ConstIter end, unsigned int frame) {
```
hyperlogic: Lock is unused within this method...

assignment-client/src/audio/AudioMixerSlavePool.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+        // stop extra slaves...
+        auto slave = extraBegin;
+        while (slave != _slaves.end()) {
+            (*slave)->stop();
+            ++slave;
+        }
+
+        // ...cycle slaves with empty queue...
+        start(lock);
```
hyperlogic: calling start seems awkward, how about just waking up the slaves via _slaveCondition.notify_all()

assignment-client/src/audio/AudioMixerSlavePool.cpp
```diff
@@ -0,0 +1,191 @@
...
+    _running = false;
+}
+
+void AudioMixerSlavePool::setNumThreads(int numThreads) {
+    // clamp to allowed size
+    {
+        int maxThreads = QThread::idealThreadCount();
+        if (maxThreads == -1) {
+            // idealThreadCount returns -1 if cores cannot be detected
+            maxThreads = std::numeric_limits<int>::max();
```
hyperlogic: Hmm, maybe we should be conservative here instead, by limiting the max to something more reasonable like 4.

assignment-client/src/audio/AudioMixerSlavePool.h
```diff
@@ -0,0 +1,102 @@
...
+    friend void AudioMixerSlaveThread::wait();
+    friend void AudioMixerSlaveThread::notify();
+    friend bool AudioMixerSlaveThread::try_pop(SharedNodePointer& node);
+
+    // synchronization state
+    Mutex _mutex;
+    ConditionVariable _slaveCondition;
+    ConditionVariable _poolCondition;
+    int _numThreads { 0 };
+    int _numStarted { 0 };
```
hyperlogic: Because you increment this the slaveThread, you should make this an std::atomic

hyperlogic: Hmm, nevermind, I guess it's guarded by the pool._mutex.  but maybe you can get rid of the mutex if you make the _numStarted and _numFinished counters atomic.

libraries/networking/src/LimitedNodeList.h
```diff
@@ -174,7 +174,26 @@ class LimitedNodeList : public QObject, public Dependency {
...
+
+    // for use with algorithm
+    using value_type = SharedNodePointer;
+    using const_iterator = std::vector<value_type>::const_iterator;
+
+    // Cede control of iteration under a single read lock (e.g. for use by thread pools)
+    //   This allows multiple threads (i.e. a thread pool) to share a lock
+    //   without deadlocking when a dying node attempts to acquire a write lock
+    template<typename NodeAlgorithmLambda>
+    void algorithm(NodeAlgorithmLambda functor) {
```
hyperlogic: Where is this used? I'm maybe not super up to date on modern C++ but I don't see where this is used.

libraries/networking/src/LimitedNodeList.h
```diff
@@ -280,7 +299,7 @@ public slots:
 protected slots:
     void connectedForLocalSocketTest();
     void errorTestingLocalSocket();
-
```
hyperlogic: +1 for removing dangling whitespace!

assignment-client/src/audio/AudioMixerSlave.cpp
```diff
@@ -0,0 +1,521 @@
...
+
+void AudioMixerSlave::configure(ConstIter begin, ConstIter end, unsigned int frame) {
+    _begin = begin;
+    _end = end;
+    _frame = frame;
+}
+
+void AudioMixerSlave::mix(const SharedNodePointer& node) {
+    // check that the node is valid
+    AudioMixerClientData* data = (AudioMixerClientData*)node->getLinkedData();
```
hyperlogic: Is the AudioMixerClientData ever accessed by any other thread?  if so, you'll have to guard against that.

assignment-client/src/audio/AudioMixerSlavePool.cpp
zzmp: They need certain members set to guard against spurious wakeups, and I wrapped that in `start` so it could be reused here and in actual mixing.

libraries/networking/src/LimitedNodeList.h
zzmp: I can call this something else, but I couldn't really think of any clarifying names.

It is used by AudioMixerSlave to do nested iteration over nodes without requiring nested read locks. If a write lock is requested before a nested read lock (but after the unnested read lock on the same thread), there is UB and a potential deadlock if write lock is given precedence. This *does* occur if a node is killed in the middle of mixing, because mixing does a nested iteration across nodes.

Can think of a way to rephrase the comments to clarify this?

assignment-client/src/audio/AudioMixerSlave.cpp
zzmp: Yes, it is. Why do I have to guard against it?

The data is accessed in AudioMixer::prepareFrame, where it is modified. This all occurs serially, in one thread.
The data is accessed again in AudioMixerSlave::prepareMix, where it is not modified. This occurs in parallel, but should only take read locks.

hyperlogic: Hmm it was not clear to me that it is not modified, perhaps you should use a const* then.

zzmp:
The mixer seems to not recover once it starts throttling. I suspect it may be building up packets that are stalling the mixer in QCoreApplication::processEvents. I will add timings to diagnose it, but it may be something that can wait for another PR.

assignment-client/src/audio/AudioMixerSlavePool.h
zzmp: This shouldn't be the performance bottleneck, so I am open to changing the data types to make it more readable and maintainable.

hifi-gustavo:
Build console output for commit: b2f995d3df233ddd2143e21e66f6415d5af4d009 available at: https://gist.github.com/ed1a7d4bd28d1a3ade51a74ff1508aec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-b2f995d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-b2f995d.dmg

assignment-client/src/audio/AudioMixerSlavePool.h
zzmp: These two members need to be under a lock so threads don't increment them at the same time and go past numThreads. atomics are not a strong enough guarantee. (Tried to implement with atomics and this is where it failed.)

hifi-gustavo:
Build console output for commit: c4e435a1666d09c8191f71ef8dd66201dddc8ee1 available at: https://gist.github.com/ed1a7d4bd28d1a3ade51a74ff1508aec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-c4e435a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-c4e435a.dmg

assignment-client/src/audio/AudioMixer.cpp
```diff
@@ -762,202 +362,162 @@ AudioMixerClientData* AudioMixer::getOrCreateClientData(Node* node) {
...
 
 void AudioMixer::parseSettingsObject(const QJsonObject &settingsObject) {
+    if (settingsObject.contains(AUDIO_THREADING_GROUP_KEY)) {
+        QJsonObject audioThreadingGroupObject = settingsObject[AUDIO_THREADING_GROUP_KEY].toObject();
+        const QString AUTO_THREADS = "auto_threads";
+        bool autoThreads = audioThreadingGroupObject[AUTO_THREADS].toBool();
+        if (!autoThreads) {
+            const QString NUM_THREADS = "num_threads";
+            int numThreads = audioThreadingGroupObject[NUM_THREADS].toInt();
+            _slavePool.setNumThreads(numThreads);
```
hyperlogic: So for some reason on my machine numThreads is 8, where my actual settings are 4.

    [12/06 18:06:15] [DEBUG] [default] [16456] [assignment-client] AudioMixerSlavePool::resize: set 8 threads (was 0)
    [12/06 18:06:15] [DEBUG] [hifi.networking] [16456] [audio-mixer]  QJsonObject({"audio_buffer":{"dynamic_jitter_buffer":true,"max_frames_over_desired":"","repetition_with_fade":"","static_desired_jitter_buffer_frames":"1","use_stdev_for_desired_calc":"","window_seconds_for_desired_calc_on_too_many_starves":"","window_seconds_for_desired_reduction":"","window_starve_threshold":""},"audio_env":{"attenuation_coefficients":"","attenuation_per_doubling_in_distance":"0.5","codec_preference_order":"hifiAC,zlib,pcm","enable_filter":true,"noise_muting_threshold":"0.003","reverb":"","zones":""},"audio_threading":{"auto_threads":true,"num_threads":"4"}})

Maybe there's a race?  where reset gets called before the settings are acutally loaded?

assignment-client/src/audio/AudioMixerSlave.cpp
```diff
@@ -0,0 +1,519 @@
...
+            }
+
+            ++stats.manualEchoMixes;
+        }
+
+        return;
+    }
+
+    // get the existing listener-source HRTF object, or create a new one
+    auto& hrtf = listenerNodeData.hrtfForStream(sourceNodeID, streamToAdd.getStreamIdentifier());
```
hyperlogic: While you are mixing on a slave thread, a network packet can be received on the AudioMixer thread that will call AudioMixerClientData::removeHRTFForStream() or AudioMixerClientData::removeHRTFsForNode().  Access to the AudioMixerClientData::_nodeSourcesHRTFMap is not thread safe, you'll need to guard against this.

assignment-client/src/audio/AudioMixer.cpp
zzmp: You have `auto_threads` set to true, which uses all available threads on your machine, as detected by Qt.

Does that help text need to be updated on the settings page?

assignment-client/src/audio/AudioMixerSlave.cpp
zzmp: All calls to `AudioMixerClientData::removeHRTFForStream` are nested in a call to `LimitedNodeList::eachNode`, which requires a read lock taken in AudioMixer::start at `LimitedNodeList::nestedEach`...but the read lock is recursive, and it should need a write lock on the streams mutex.

This whole client data thing is messy. Good find.

assignment-client/src/audio/AudioMixer.cpp
hyperlogic: Yes you are right, however when I turn off auto_threads, with num_threads = 4. I get the following which doesn't seem right.
    [12/07 09:33:44] [DEBUG] [default] [15980] [assignment-client] AudioMixerSlavePool::resize: set 8 threads (was 0)
    [12/07 09:33:44] [DEBUG] [hifi.networking] [15980] [audio-mixer] Requesting settings from domain server
    [12/07 09:33:44] [DEBUG] [hifi.networking] [15980] [audio-mixer] Received domain settings: 
    [12/07 09:33:44] [DEBUG] [hifi.networking] [15980] [audio-mixer]  QJsonObject({"audio_buffer":{"dynamic_jitter_buffer":true,"max_frames_over_desired":"","repetition_with_fade":"","static_desired_jitter_buffer_frames":"1","use_stdev_for_desired_calc":"","window_seconds_for_desired_calc_on_too_many_starves":"","window_seconds_for_desired_reduction":"","window_starve_threshold":""},"audio_env":{"attenuation_coefficients":"","attenuation_per_doubling_in_distance":"0.5","codec_preference_order":"hifiAC,zlib,pcm","enable_filter":true,"noise_muting_threshold":"0.003","reverb":"","zones":""},"audio_threading":{"auto_threads":false,"num_threads":"4"}})
    [12/07 09:33:44] [WARNING] [default] [15980] [audio-mixer] AudioMixerSlavePool::setNumThreads: clamped to 0 (was 1)
    [12/07 09:33:44] [DEBUG] [default] [15980] [audio-mixer] AudioMixerSlavePool::resize: set 1 threads (was 8)

assignment-client/src/audio/AudioMixerSlave.cpp
zzmp: I don't think this is a problem.

`AudioMixerClientData` has some things (like _streamsLock) that are individually guarded, but they are not guarded through the whole mixing action, which might be more appropriate (don't add/delete streams while mixing, as state should be const during the mix).

However, all of the things that mutate `AudioMixerClientData` are triggered by signals/slots on `AudioMixer`. There are three cases:

- `packetReceiver.registerListener(PacketType::KillAvatar, this, "handleKillAvatarPacket");`

This becomes a case a a `Qt::AutoConnection` (see [PacketReceiver::registerListener](https://github.com/highfidelity/hifi/blob/d83600a22ee1579b15acf742e2fa83a84ce9951c/libraries/networking/src/PacketReceiver.cpp#L291) and [Qt::AutoConnection](http://doc.qt.io/qt-5/qt.html#ConnectionType-enum)), which uses a `Qt::QueuedConnection`. The packet is received on the node list, so the slot will not be invoked until `QtApplication::processEvents`, after mixing, when the `AudioMixer` thread is free. No lock is necessary.

- `connect(nodeList.data(), &NodeList::nodeKilled, this, &AudioMixer::handleNodeKilled);`

This is the same case - a `Qt::AutoConnection`, from the `NodeList` thread to the `AudioMixer` thread, so no lock is necessary.

- `connect(clientData, &AudioMixerClientData::injectorStreamFinished, this, &AudioMixer::removeHRTFsForFinishedInjector);`

`AudioMixerClientData` and `AudioMixer` live in the same thread, so `Qt::AutoConnection` becomes a `Qt::DirectConnection`. However, this is only signaled in [CheckBuffersBeforeFrameSend](https://github.com/highfidelity/hifi/blob/c4e435a1666d09c8191f71ef8dd66201dddc8ee1/assignment-client/src/audio/AudioMixerClientData.cpp#L211), which is called prior to mixing, and so should not cause any data contention.

---

I think that parts of `AudioMixerClientData` are spaghetti and need to be improved, but I think for now it will function without race conditions, and the improvements should wait for another PR.

assignment-client/src/audio/AudioMixer.cpp
zzmp: That definitely looks wrong. I'll look into it now.

OmegaHeron:
Tested Ubuntu 16.04.1 with the following configurations;

i7 8 core LAN connected machine
1 CPU/core digital ocean droplet
2 CPU/core Linode droplet

i5 4 core localhost (Windows 10)

Very informal smoke test only -- simply verified audio works as expected for 2 agents and that it did not "destabilize" anything on the droplet machines with 1/2 core(s).

Ideal thread count for each test machine matched what I'd expect it to.

The only concern I have is not a blocker, but... on someone trying to run sandbox/interface on their home machine with a low spec machine (i.e. i5 CPU) it may exacerbate not enough CPU to get the job done situation.  But -- that's not really a "your problem" situation.

zzmp:
@OmegaHeron I feel like using idealThreadCount for the thread count is a more sane option, but a home user *could* run this without tanking their CPUs by setting threadcount to `auto` off and `1` or `2`.

OmegaHeron:
@zzmp Agreed - It'll likely be good to document that for those wanting to try and do things on low spec machines that they, likely, shouldn't do.  ;)  In so far as this goes for running "serious" server installs it's a great step to glorious future lands.

cozza13:
I tested this on the machine that Zach provided on EC2 and it worked as expected.

zzmp:
@OmegaHeron @hyperlogic pointed out that all Steam users and normal users will have a local server running, so I am changing the default to not auto, with 1 thread.

hifi-gustavo:
Build console output for commit: ce9346f524e23e1fced13753365745344894990c available at: https://gist.github.com/ed1a7d4bd28d1a3ade51a74ff1508aec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-ce9346f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-ce9346f.dmg

hifi-gustavo:
Build console output for commit: f5b4fac46d8e14b14cb02433f9a9687be0c7b51d available at: https://gist.github.com/ed1a7d4bd28d1a3ade51a74ff1508aec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-f5b4fac.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-f5b4fac.dmg

hifi-gustavo:
Build console output for commit: 3508e717532e09a8237c8060903843df0e8658e7 available at: https://gist.github.com/ed1a7d4bd28d1a3ade51a74ff1508aec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-3508e71.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-3508e71.dmg

assignment-client/src/audio/AudioMixerSlavePool.cpp
```diff
@@ -154,12 +153,12 @@ void AudioMixerSlavePool::resize(int numThreads) {
...
         // ...cycle them until they do stop...
-        {
-            Lock lock(_mutex);
-            _numStopped = 0;
-            while (_numStopped != (_numThreads - numThreads)) {
-                _numStarted = _numFinished = _numStopped;
-                _slaveCondition.notify_all();
+        _numStopped = 0;
+        while (_numStopped != (_numThreads - numThreads)) {
+            _numStarted = _numFinished = _numStopped;
```
hyperlogic: Here you are changing the _numStarted and _numFinished a new value without the _mutex acquired, this will conflict with the access via the slave threads.

hifi-gustavo:
Build console output for commit: 5969c1eb06fabd1a282b8986507eb4e1fa00318f available at: https://gist.github.com/ed1a7d4bd28d1a3ade51a74ff1508aec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-5969c1e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9130-5969c1e.dmg

ZappoMan:
@OmegaHeron @cozza13 - looks like this needs another QA pass... since changes were made.

cozza13:
I have tested this on a large EC2 server and it works well

OmegaHeron:
Passes smoke test for the basics on small hardware as well - sorry it took me a bit to get to it.  Tested same as before on my Linux 1/2 cpu machines - i7 LAN machine (Linux) and i5 Win 10.

PR #9135, opened 2016-12-01
Fixed pressing the grip button causes a near grabbed object to drop

druiz17:
To Test
1. near or far grab an object...when pressing the grip button...the grabbed object should not be dropped
2. equip an object and pressing the grip button should drop the object.

fogbugz - https://highfidelity.fogbugz.com/f/cases/2368/Grip-button-causes-you-to-drop-a-near-grabbed-object

### DIFF
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -1012,7 +1012,7 @@ function MyController(hand) {
 
         // The value to check if we will allow the release function to be called
         var allowReleaseValue = 0.1;
-        if (value > 0 && _this.rawTriggerValue <= allowReleaseValue) {
+        if (value > 0 && _this.state == STATE_HOLD) {
             _this.release();
         }
     };
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -1009,7 +1009,10 @@ function MyController(hand) {
 
     this.secondaryPress = function(value) {
         _this.rawSecondaryValue = value;
-        if (value > 0) {
+
+        // The value to check if we will allow the release function to be called
+        var allowReleaseValue = 0.1;
+        if (value > 0 && _this.rawTriggerValue <= allowReleaseValue) {
             _this.release();
         }
     };
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -2346,6 +2346,7 @@ function MyController(hand) {
                 this.callEntityMethodOnGrabbed("releaseEquip");
             }
 
+
             //  Make a small release haptic pulse if we really were holding something
             Controller.triggerHapticPulse(HAPTIC_PULSE_STRENGTH, HAPTIC_PULSE_DURATION, this.hand);
 
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -2349,7 +2349,6 @@ function MyController(hand) {
                 this.callEntityMethodOnGrabbed("releaseEquip");
             }
 
-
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: b11ec38cca0e952834e835a31b12a921cff290af available at: https://gist.github.com/1a7e4af42c750f8c843aa7701e10855d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9135-b11ec38.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9135-b11ec38.dmg

scripts/system/controllers/handControllerGrab.js
```diff
@@ -1009,7 +1009,10 @@ function MyController(hand) {
 
     this.secondaryPress = function(value) {
         _this.rawSecondaryValue = value;
-        if (value > 0) {
+
+        // The value to check if we will allow the release function to be called
+        var allowReleaseValue = 0.1;
+        if (value > 0 && _this.rawTriggerValue <= allowReleaseValue) {
```
huffman: Is this better than just checking the current state and whether something is currently equipped?

druiz17: In my opinion, it is. They are many different states for grabbing an entity, like far grabbing , near grabbing and holding. Since all those states depend on the trigger being held its seemed better to check against the trigger value rather than all the different states.

huffman: Ok - I won't block this, but a couple thoughts:

 * The only state you should need to check is if `_this.state` is equal to `STATE_HOLD` (I think), which is better because this is what we actually care about - whether we are currently holding an entity or not.
 * Checking the trigger value here is brittle to changes in the future. It's using it as a proxy to determine whether an entity is equipped. If we ever change other methods of grabbing (or add new ones) that don't require the trigger to be held, this will break.

druiz17: - I do agree checking against the trigger value is a brittle to changes in the future, but when I tried to check if _this.state is equal to STATE_HOLD, it prevented you from dropping any equipped objects. I'll try to check against the state just to make sure

Nex-Pro:
tested on Win10 and Vive.

Smells like Victory !

## PASS

thanks for this one.

druiz17:
@huffman I re-tried the solution you proposed and it worked. I think made a mistake when I tried this approach last time.

huffman:
@druiz17 Awesome!

hifi-gustavo:
Build console output for commit: 09d5dc382f7a228238b55c17d23c84899eca2d55 available at: https://gist.github.com/1a7e4af42c750f8c843aa7701e10855d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9135-09d5dc3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9135-09d5dc3.dmg

druiz17:
Can I get a QA test for this implementation again

hifi-gustavo:
Build console output for commit: 08301d488d3b41fff52001e50d48c7baec8855d3 available at: https://gist.github.com/1a7e4af42c750f8c843aa7701e10855d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9135-08301d4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9135-08301d4.dmg

PhilipRosedale:
Testing

PhilipRosedale:
Verified on Vive controller that grip no longer drops near grabbed object, and also that grip still works for dropping the ping-pong gun.

PhilipRosedale:
QA PASS

PR #9136, opened 2016-12-01
Fix a crash if current shader is null

samcake:
This PR fixes a potential crash happening when a non compiled shader is used (and failed to compile) and a "_glUniform" is assigned on it.
This could be the case with procedural shaders failing to compile and beeing rendered. Their rendering sequence will potentially assign uniform values. THat s where we would get a crash.

Test Plan:
In current RC, some user experience a crash going to eschatology or looking at an entity with a failing procedural shader.
With this PR, this should be solved

THis should happen regardless of beeing desktop / HMD ( need to check both).

Smoke test otherwise, performances and quality should be the same

### DIFF
--- a/libraries/gpu-gl/src/gpu/gl/GLBackend.cpp
+++ b/libraries/gpu-gl/src/gpu/gl/GLBackend.cpp
@@ -406,7 +406,7 @@ void GLBackend::do_popProfileRange(const Batch& batch, size_t paramOffset) {
 
 // As long as we don;t use several versions of shaders we can avoid this more complex code path
 #ifdef GPU_STEREO_CAMERA_BUFFER
-#define GET_UNIFORM_LOCATION(shaderUniformLoc) _pipeline._programShader->getUniformLocation(shaderUniformLoc, (GLShader::Version) isStereo())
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: ff39ec675376d66b30ade6ee6a2da808a5f67abb available at: https://gist.github.com/01096c1fdcdb4b6a274e24bfc15f4846<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9136-ff39ec6.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9136-ff39ec6.dmg

hyperlogic:
works on my machine, i.e. I could no longer reproduce the glUniform crash

OmegaHeron:
Tested Windows 10 
General smoke test passes.

I can't say if this fixes intended problem as I can't reproduce crash with master (or this PR).  But -- I spent time in eschatology -- looked at several iterations of a skybox shader a friend is working on and visited usual test locations (dev-playa/dev-demo/old apartment scene).  Everything looks good, no change to rendering, FPS unchanged, CPU use unchanged.

PR #9137, opened 2016-12-01
Fixes for building on MSVC 2015 / Qt 5.8

jherico:
These are the required changes for enabling builds on Qt 5.8 and Visual Studio 2015.

The `#pragma comment(lib, "legacy_stdio_definitions.lib")` line ensures that functions which used to be exported from previous versions of the C stdio library, but are now inline still exist as linkable.  This is required because the OpenSSL binaries we rely on link against these functions.  

The remaining changes resolve compile errors or fix uninitialized variables.

## Testing

Smoke test only.  These changes should have no impact on application behavior or performance.

### DIFF
--- a/libraries/entities/src/EntityItemProperties.cpp
+++ b/libraries/entities/src/EntityItemProperties.cpp
@@ -147,15 +147,15 @@ QString getCollisionGroupAsString(uint8_t group) {
 }
 
 uint8_t getCollisionGroupAsBitMask(const QStringRef& name) {
-    if (0 == name.compare("dynamic")) {
+    if (0 == name.compare(QString("dynamic"))) {
         return USER_COLLISION_GROUP_DYNAMIC;
-    } else if (0 == name.compare("static")) {
+    } else if (0 == name.compare(QString("static"))) {
         return USER_COLLISION_GROUP_STATIC;
-    } else if (0 == name.compare("kinematic")) {
+    } else if (0 == name.compare(QString("kinematic"))) {
         return USER_COLLISION_GROUP_KINEMATIC;
-    } else if (0 == name.compare("myAvatar")) {
+    } else if (0 == name.compare(QString("myAvatar"))) {
         return USER_COLLISION_GROUP_MY_AVATAR;
-    } else if (0 == name.compare("otherAvatar")) {
+    } else if (0 == name.compare(QString("otherAvatar"))) {
         return USER_COLLISION_GROUP_OTHER_AVATAR;
     }
     return 0;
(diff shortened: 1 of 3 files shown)


hifi-gustavo:
Build console output for commit: 308662c027dd544ed15509171b29e26e4729110b available at: https://gist.github.com/a6a00af0716cd8bb60df1446721638c1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9137-308662c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9137-308662c.dmg

Nex-Pro:
Tested on Win10 and vive.

Seems generally ok but noticed some issues that may or may not be related to this PR,

Avatar can get into 3rd person mode without specifically selecting it or rolling the mouse wheel.

caitlyn observed an issue with rotating the avatar when in hover mode and grabbed some gifs of that.

Selecting Developer -> Avatar -> Move kinematically  triggers a crash each time. ( splats posted ) 

Frame rates are ok.

Avatar is able to mount stairs that regular avatars cannot.

Standing sideways on a slope triggers a rapid shuffling of the feet that seems to propogate all the way up to the hands.

Avatar does not tend to slide down slopes.

Show left foot trace works fine.

**Show right foot trace does not work.**

hifi-gustavo:
Build console output for commit: 5cbf7588acb94dbb68eaca1fea42e71e49e8587f available at: https://gist.github.com/a6a00af0716cd8bb60df1446721638c1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9137-5cbf758.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9137-5cbf758.dmg

OmegaHeron:
Running re-test on this per request.  @jherico - just to confirm, this does not actually bundle Qt5.8, it only makes it *possible* to use Qt5.8 specifically on Windows -- right?

OmegaHeron:
Testing windows 10.

Move kinematically results in crash to desktop in multiple versions of master releases - so it's not specific to this PR.

Avatar standing on slope - if sufficiently steep results in avatar sliding down slope which triggers walk animation - in desktop mode this looks like a normal slow walk down said slope.  In HMD it can look a little weird - but this does not come from this PR as all recent versions of master exhibit same (at least in so far as I can reproduce described behavior).

I couldn't find any stairs that this made any difference in.

Foot trace exhibits same in master - only one foot can display its move trace - left/right both work, just not at same time.

I was unable to reproduce 1st/3rd person view changing without my explicitly making it happen.

Weird avatar movement has been fixed in PR subsequent to this one - not this PR's problem.

FPS/CPU/General smoke test = all pass.

All issue found in first round QA exist in master currently (or have been fixed in pending ones).

PR #9155, opened 2016-12-05
Fix drop in text rendering quality when texture memory is low

jherico:
Texture downsampling is used in our application to ensure that we don't start dying when too much content is loaded, by reducing the memory consumed by textures.  However, there are some instances where we absolutely should not be downsampling a texture.  Two cases are the cursor textures used in UI compositing and the distance fields we use for text rendering.  In both these we can easily avoid downsampling by disabling mip generation for them. 

## Testing

On an AMD system launch the application and create a text entity in a scene with some textured content.  Then set the max texture memory to 4 MB in the Developer / Render / Max Texture Memory menu.  On production builds this will cause the text to become unreadable or disappear entirely.  On this build it should not affect text (entity) rendering at all.

### DIFF
--- a/interface/src/Menu.cpp
+++ b/interface/src/Menu.cpp
@@ -363,6 +363,7 @@ Menu::Menu() {
     QActionGroup* textureGroup = new QActionGroup(textureMenu);
     textureGroup->setExclusive(true);
     textureGroup->addAction(addCheckableActionToQMenuAndActionHash(textureMenu, MenuOption::RenderMaxTextureAutomatic, 0, true));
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: 2ca4d0ab29082de10ac023749c994a8a00b4ed51 available at: https://gist.github.com/e85f30083cf75aa5131af16e6db36168<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9155-2ca4d0a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9155-2ca4d0a.dmg

cozza13:
tested on AMD. QA Pass

capture

PR #9168, opened 2016-12-08
Optimized method for transforming axis aligned bounding boxes

jherico:
Testing shows that for scenes with many avatars a significant time is spent per frame in `Scene::processPendingChangesQueue` which can take up 20% of total runtime.  Nearly half of that time is spent inside `AABox::transform`.  The current implementation of `AABox::transform` is inefficient, calling far more matrix transformations than are strictly necessary.  This PR modifies the function to use a more efficient method found online.

This method was tested internally by allowing both implementations to exist in the function and produce results and finding the difference between the computed results, by subtracting one method result from the other and finding the length squared for the resulting value.  In all tested cases the result was below an epsilon value of 1 micrometer.

## Testing 

smoke test, application behavior should be unaffected.

### DIFF
--- a/libraries/shared/src/AABox.cpp
+++ b/libraries/shared/src/AABox.cpp
@@ -568,46 +568,20 @@ void AABox::transform(const Transform& transform) {
     translate(transform.getTranslation());
 }
 
+// Logic based on http://clb.demon.fi/MathGeoLib/nightly/docs/AABB.cpp_code.html#471
 void AABox::transform(const glm::mat4& matrix) {
-    auto minimum = _corner;
-    auto maximum = _corner + _scale;
+    auto halfSize = _scale * 0.5f;
+    auto center = _corner + halfSize;
+    halfSize = abs(halfSize);
+    auto newCenter = transformPoint(matrix, center);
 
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 3fcd45a23625ae8d6dabf27e1fc610c816cded6b available at: https://gist.github.com/a7e4805bbfa229973d7643e2b229f173<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9168-3fcd45a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9168-3fcd45a.dmg

ZappoMan:
Math FTW!

OmegaHeron:
Testing Windows 10
QA Pass
General smoke test passes - visited a few domains including welcome where there were more than a few AVs... all looked normal.

ZappoMan:
@OmegaHeron - I'm curious if you saw a CPU improvement... that's the other goal of this PR...

OmegaHeron:
@ZappoMan - minor less CPU - I suspect it would need a crowd to really show gains.

PR #9176, opened 2016-12-09
experimental kinect input plugin

ZappoMan:
Just opening this up for a little bit of daylight and transparency. Philip asked me to experiment with seeing how easy it would be to get a kinect camera working as an input plugin. The headline is, connecting to the kinect body tracking SDK is pretty straightforward, and I'm easily getting camera space joint positions (and rotations)...

This is mostly cargo culted from the neuron plugin, and in exploring that plugin, and experimenting with this it uncovers a couple problems with the input plugin API... here are some thoughts (some I've had for awhile, but I'm articulating them now in the context of this experiment)...

- The input plugin paradigm of exposing controller pose data in "sensor space" is kinda broken... because, as you can imagine, we really want to be able to support multiple plugins simultaneously providing pose data (think wearing an oculus while standing in front of a kinect, and/or using leap hands while wearing an oculus).

- We currently support "actions"/"pose sinks" for RightHand/LeftHand... that's great because it lets you either programatically write directly into the pose data for the joint from "sensor space" (or some other shared space) -- without dealing with the skeleton chain... the other approach (used by the neuronAvatar.js) is to translate pose data into a change of joint relative "mini-poses" -- that's a super pain in the ass, and error prone, especially if the input controller is something like kinect, oculus, or vive that has awareness of room scale.

**How can we allow a Kinect + Oculus to synchronize their sensor space?**
- One technique might be to allow the input plugin to ask the application for a world pose for standardized pose... for example, kinect has a head point in it's sensor space, similarly, the oculus has a head position. So they could both as (where is the head in the world) -- and then they could store that as the world to *their* sensor space.
- however this synchronization of sense spaces might require an explicit step by the users to calibrate. clearly you wouldn't want this step to happen until the oculus was on your head.

**Why doesn't the kinect.json automatically map the hands?**

- I've written a JS file that adds this mapping:
```
mapping.from(Controller.Hardware.Kinect.LeftHand).to(Controller.Standard.LeftHand);
```
- and the hand animates (albeit in a bad frame of reference)
- but why doesn't the json automatically do that mapping, it should

Anyway - this won't get merged until these problems can be better addressed, and I welcome input/feedback on how we can make input plugins that drive avatar joints easier to build.

### DIFF
--- a/interface/resources/qml/hifi/dialogs/GeneralPreferencesDialog.qml
+++ b/interface/resources/qml/hifi/dialogs/GeneralPreferencesDialog.qml
@@ -17,7 +17,7 @@ PreferencesDialog {
     id: root
     objectName: "GeneralPreferencesDialog"
     title: "General Settings"
-    showCategories: ["UI", "Snapshots", "Scripts", "Privacy", "Octree", "HMD", "Sixense Controllers", "Perception Neuron"]
+    showCategories: ["UI", "Snapshots", "Scripts", "Privacy", "Octree", "HMD", "Sixense Controllers", "Perception Neuron", "Kinect"]
     property var settings: Settings {
         category: root.objectName
         property alias x: root.x
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -3977,10 +3977,16 @@ void Application::update(float deltaTime) {
 
     controller::Pose leftHandPose = userInputMapper->getPoseState(controller::Action::LEFT_HAND);
     controller::Pose rightHandPose = userInputMapper->getPoseState(controller::Action::RIGHT_HAND);
+
+    qDebug() << __FUNCTION__ << "...............leftHandPose:" << leftHandPose.translation;
+
     auto myAvatarMatrix = createMatFromQuatAndPos(myAvatar->getOrientation(), myAvatar->getPosition());
     auto worldToSensorMatrix = glm::inverse(myAvatar->getSensorToWorldMatrix());
     auto avatarToSensorMatrix = worldToSensorMatrix * myAvatarMatrix;
     myAvatar->setHandControllerPosesInSensorFrame(leftHandPose.transform(avatarToSensorMatrix), rightHandPose.transform(avatarToSensorMatrix));
+
+    qDebug() << __FUNCTION__ << "POST transform leftHandPose:" << leftHandPose.translation;
+
 
     updateThreads(deltaTime); // If running non-threaded, then give the threads some time to process...
     updateDialogs(deltaTime); // update various stats dialogs if present
--- a/cmake/macros/TargetKinect.cmake
+++ b/cmake/macros/TargetKinect.cmake
@@ -1,0 +1,17 @@
+#
+#  Created by Brad Hefta-Gaub on 2016/12/7
+#  Copyright 2016 High Fidelity, Inc.
+#
+#  Distributed under the Apache License, Version 2.0.
+#  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+#
+macro(TARGET_KINECT)
+    # Kinect SDK data reader is only available on these platforms
+    if (WIN32)
+        #add_dependency_external_projects(kinect)
+        find_package(Kinect REQUIRED)
+        target_include_directories(${TARGET_NAME} PRIVATE ${KINECT_INCLUDE_DIRS})
+        target_link_libraries(${TARGET_NAME} ${KINECT_LIBRARIES})
+        add_definitions(-DHAVE_KINECT)
+    endif(WIN32)
+endmacro()
--- a/cmake/modules/FindKinect.cmake
+++ b/cmake/modules/FindKinect.cmake
@@ -1,0 +1,59 @@
+#
+#  FindKinect.cmake
+#
+#  Try to find the Perception Kinect SDK
+#
+#  You must provide a KINECT_ROOT_DIR which contains lib and include directories
+#
+#  Once done this will define
+#
+#  KINECT_FOUND - system found Kinect SDK
+#  KINECT_INCLUDE_DIRS - the Kinect SDK include directory
+#  KINECT_LIBRARIES - Link this to use Kinect
+#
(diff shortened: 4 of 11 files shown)


interface/resources/qml/hifi/dialogs/GeneralPreferencesDialog.qml
```diff
@@ -17,7 +17,7 @@ PreferencesDialog {
     id: root
     objectName: "GeneralPreferencesDialog"
     title: "General Settings"
-    showCategories: ["UI", "Snapshots", "Scripts", "Privacy", "Octree", "HMD", "Sixense Controllers", "Perception Neuron"]
+    showCategories: ["UI", "Snapshots", "Scripts", "Privacy", "Octree", "HMD", "Sixense Controllers", "Perception Neuron", "Kinect"]
```
ZappoMan: this is kinda lame... it sucks that we can't just drop in plugins and have them add preferences accordingly.

hifi-gustavo:
Build console output for commit: 4959a885815413b01058289d3e7e3614f160813e available at: https://gist.github.com/0bd3e8465a7e4f1ab2396b1a24a98586<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9176-4959a88.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9176-4959a88.dmg

hifi-gustavo:
Build console output for commit: b6838c6ec424fd190e04fe7d6a986a69fad14d31 available at: https://gist.github.com/0bd3e8465a7e4f1ab2396b1a24a98586<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9176-b6838c6.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9176-b6838c6.dmg

hifi-gustavo:
Build console output for commit: fdf9dc432de40aa4d3bc0187c1d64eab64d04d2c available at: https://gist.github.com/0bd3e8465a7e4f1ab2396b1a24a98586<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9176-fdf9dc4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9176-fdf9dc4.dmg

judasshuffle:
I use to use the kinnect with faceshift, then swapped over to the xtion pro which seemed to be better at capturing closer. Are the 2 devices like oil and water or can it support both? assumes you have a pile of xtion pros somwhere

ZappoMan:
@judasshuffle - I'm experimenting with the official xbox kinect 2 camera on this PR... this is doing full body tracking. I do have a prime sense camera (presumably the same as the Xtion PRO) - I haven't tried it with the kinect sdk.

judasshuffle:
Hi then I feel better, i dont have a kinect 2 and i gave my kinect 1 away about a week a go lol. excited to see what you come up with

ZappoMan:
@sethalves @jherico @hyperlogic - I was chatting with @PhilipRosedale a little about this... and described the following concept: 
- Wearing Oculus or Vive (in room scale)
- Kinect camera pointing at you to get body
- Leap motion on your HMD to get detailed hands

Imagine that setup... that's what I'd ultimately like to make possible for input plugins. This explains why having a solution for synchronizing pose data out of multiple different "sensor spaces" is important.

Of course one solution is to just use the HMD for head positioning, and then meat hooking the body joints after that... but that's less optimal in my mind. The problem is that the kinect can in fact do room scale too... so we don't want to force people to have an HMD just to get room scale...

One idea that Philip and I had was the idea of having "room scale" input plugins constantly report their sensor space for a "known pose" -- for example, the head pose. If all input plugins reported their head pose constantly (in their sensor space) then the system can do the correct synchronization for all the other pose data.

Thoughts?

jherico:
What you're talking about would probably be most easily accomplished not through our own input plugins but through OpenVR.  OpenVR already has the idea of a sensor space universe.  If you could create a synchronization mechanism where by multiple OpenVR device drivers could find a common frame of reference then it would basically be a matter of exposing all the various inputs as OpenVR controllers.  I suspect this would be an appropriate prelude to the Khronos API for VR, whenever that arrives.

interface/resources/qml/hifi/dialogs/GeneralPreferencesDialog.qml
jherico: this property could instead use something that is exposed to the QML context... i.e. `showCategories: globalPreferenceCategories`.  This could then be something that is modifiable at the C++ level so that a plugin that is active could add itself to the preferences list by modifying that global.

interface/src/Application.cpp
```diff
@@ -3977,11 +3977,17 @@ void Application::update(float deltaTime) {
 
     controller::Pose leftHandPose = userInputMapper->getPoseState(controller::Action::LEFT_HAND);
     controller::Pose rightHandPose = userInputMapper->getPoseState(controller::Action::RIGHT_HAND);
+
+    qDebug() << __FUNCTION__ << "...............leftHandPose:" << leftHandPose.translation;
```
ZappoMan: ok - yeah - this isn't ready for merging, I obviously haven't removed debug code.

ZappoMan:
@jherico - I don't understand how an OpenVR approach would work for someone who's using our system through the oculus store... I don't think that's a viable approach.

More importantly our current input plugin API is broken because of this. I have no way for kinect to correctly send pose data for the avatars head and hands... even when you don't consider synchronizing multiple plugins.

Polyrhythm:
One slightly more agnostic way of synchronizing device A with device B with Device C would be an IR marker. Pretty much any visual input device I can think of uses IR, so the procedure would be the same whether it's a hand tracker, face tracker, body tracker, etc.

This is, of course, incredibly consumer-unfriendly. Just me trying to find the a solution that would be flexible for any given device as that seems to be an implicit question in this PR - not specific to the Kinect itself.

ZappoMan:
I got the insight and feedback I wanted... closing for now.

PR #9187, opened 2016-12-12
Rescale yourself or entities using the grip buttons

PhilipRosedale:
This change allows you to rescale entities that you are holding by using the grip buttons.   While near or far-grabbing, hold both grip buttons down and move hands together or apart to rescale the entity.  

If no entities are grabbed, rescale will affect your avatar!


PhilipRosedale:
Test Plan:  

Check for normal operation for equip/dequip, and near and far grab of different objects.   Also test far trigger.   Check that the grip button will still drop a held object. 

Verify that you can rescale entities by squeezing both grip buttons and moving hands toward or apart from each other.

hifi-gustavo:
Build console output for commit: a142c830faefaee72b53c35dd18aff281b7c4e5c available at: https://gist.github.com/182ebd38aa5caf0cfa97eda706ef8975<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9187-a142c83.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9187-a142c83.dmg

Nex-Pro:
### PASS

Works well apart from two handed sword fighting !

And can't scale web entities.

Check for normal operation for equip/dequip, and near and far grab of different objects. **SUCCESS**
Also test far trigger. **SUCCESS**
Check that the grip button will still drop a held object.**SUCCESS**

Verify that you can rescale entities by squeezing both grip buttons and moving hands toward or apart from each other.**WORKS WELL**

Dynamic entities move as scaling takes place which might not always be desirable. 

Overall thumbs up. 

Two handed combateers might want to turn this off.

PhilipRosedale:
Nex-Pro were you serious about two handed swordfighting?  I don't want to break that - are there cases where we are needing to use both grip buttons at same time?

PhilipRosedale:
Could I get a re-test?   I added the ability to also rescale your own avatar by doing this when no object is being held.

hifi-gustavo:
Build console output for commit: 9370c8af6bf56fe6ed5a3bfa69bb174d8199d323 available at: https://gist.github.com/182ebd38aa5caf0cfa97eda706ef8975<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9187-9370c8a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9187-9370c8a.dmg

Nex-Pro:
Works as per description.

Notes:
tested this extensively with Michelle L.
Got an accidental attach, probably not due to this PR.

As I mentioned while talking about this at welcome yesterday:
This does not take into consideration users holding two objects and gripping controllers while moving them rapidly/agressively. Double sword fights etc.

I feel many users will grip the controllers when engaged in vigorous activities.
Having gripped items scale inadvertently may not be desirable.

Scaling avatar this way isn't very satisfactory as the user has no real idea what size/how tall they are after resize.
Some sort of visual clue other than hand size might be useful.

Once again I would suggest that this should be a toggled feature.

User has to drop gripped item/s if resize of avatar required.

PhilipRosedale:
Addressed QA Feedback: 

1.  Now you cannot rescale if you have objects in both hands, making it very difficult to accidentally rescale.  If you have two things held, you can't rescale at all. 

2.  Avatar rescaling now requires holding both triggers AND grips all the way down.  Not likely to do this accidentally!  When we add tablet UI, we can add a discoverable button for it. 

Test instructions:  

Same as before, but verify that you can *not* rescale if you have objects held in both hands.  And verify that to rescale avatar you must both hold down triggers AND grips.

Nex-Pro:
Tested on Win10 and Vive. ( Piper ) 

Thanks Philip.
Generally good. 


Check for normal operation for equip/dequip, and near and far grab of different objects. SUCCESS
Also test far trigger. SUCCESS
Check that the grip button will still drop a held object.SUCCESS

Verify that you can rescale entities by squeezing both grip buttons and moving hands toward or apart from each other. STILL WORKS WELL

Now you cannot rescale if you have objects in both hands, making it very difficult to accidentally rescale. If you have two things held, you can't rescale at all. : **SUCCESS**

Avatar rescaling now requires holding both triggers AND grips all the way down. Not likely to do this accidentally!  : **SUCCESS**

Plus points.

I found that using several simple line of sight experiments I was able to exactly scale my avatar to normal size. 

Avatar scaling is very precise,
seems to increment in millimeters.

Scaling of objects is very smooth.

**Only issues found are :**

Scaled objects in hand scale from object center rather than relative to hand, feels/looks odd.

When avatar is scaled to max it might take 3 attempts to initiate scaling down.

It is possible to scale an avatar up and become stuck in a collision cascade if in a restricted height location. See attached.

When avatar is scaled up it adopts a knee bent stance and needs some form of movement to trigger upright stance. This is not so using keyboard.

Might not be related to this PR but I found that large blocks scaled to a small size and then tossed to the ground would often fall through. See attached.

https://dl.dropboxusercontent.com/u/96759331/hifi-snap-by-Piper-Peppercorn-on-2016-12-15_17-14-36.gif

https://dl.dropboxusercontent.com/u/96759331/hifi-snap-by-Piper-Peppercorn-on-2016-12-15_17-18-46.gif

PR #9188, opened 2016-12-12
Fix and improve the WAV file loader

kencooke:
More robust WAV parser. Skips extraneous chunks, handles WAVE_FORMAT_EXTENSIBLE, supports arbitrary sample rates.

### DIFF
--- a/libraries/audio/src/Sound.cpp
+++ b/libraries/audio/src/Sound.cpp
@@ -195,8 +195,10 @@ struct RIFFHeader {
     char        type[4];        // "WAVE"
 };
 
-struct WAVEHeader {
-    chunk       descriptor;
+static const int WAVEFORMAT_PCM = 1;
+static const int WAVEFORMAT_EXTENSIBLE = 0xfffe;
+
+struct WAVEFormat {
     quint16     audioFormat;    // Format type: 1=PCM, 257=Mu-Law, 258=A-Law, 259=ADPCM
     quint16     numChannels;    // Number of channels: 1=mono, 2=stereo
     quint32     sampleRate;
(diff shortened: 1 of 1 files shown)


kencooke:
### Test Plan

1) Run Interface and verify that existing WAV assets load and play as before.
2) Modify a script, or use PlaySound via the JS console, to load a WAV file with sample rate other than 48khz, and verify it plays

OmegaHeron:
@kencooke -- any chance you might have some url(s) for test wav files we could use?

hifi-gustavo:
Build console output for commit: 0f10975a59371aef84ccc60bd99f935aac48b00d available at: https://gist.github.com/dcb51beaf5103b08a7e190cbfa6554d9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9188-0f10975.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9188-0f10975.dmg

OmegaHeron:
Testing Windows 10
QA Pass - I think - unless the one wav file that refused to play should have.  See last line of comments for source.

Sounds at playa work as expected along with other locations known to have sound assets.

Using test link provided - yes, rather loud 1k sine.  My ear drums thank you for cleaning.  Able to manipulate volume as expected.  I still have to trigger play sound *twice* the first time I play a sound before it will play.  I suspect that's a load/cache/time issue.

Tried a few samples from this page - all worked - static/white noise tests. http://riotorto.users.sourceforge.net/sound/sample/

This file fails.  http://riotorto.users.sourceforge.net/sound/sample/boo.wav

kencooke:
@OmegaHeron, that last one fails because it's an 8-bit WAV file, which we don't support. So that is the expected behavior.

OmegaHeron:
Then QA Pass - everything else I threw at it worked.  I had to dig deep to find that one.

PR #9200, opened 2016-12-13
Fix audio injector bug after space bubble changes

zfox23:


libraries/networking/src/udt/PacketHeaders.h
```diff
@@ -234,6 +234,7 @@ enum class AudioVersion : PacketVersion {
     CodecNameInAudioPackets,
     Exactly10msAudioPackets,
     TerminatingStreamStats,
+    SpaceBubbleChanges,
```
ZappoMan: how was this not already in there? was it because you only bumped a different packet format?

zfox23: Correct - I had only bumped the AvatarMixerPacketVersion.

hifi-gustavo:
Build console output for commit: 44c6eb303b3b5b1710ff05969dd01010cc0614cc available at: https://gist.github.com/0c6f5b4367acac2ece1efb23714291ee<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9200-44c6eb3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9200-44c6eb3.dmg

libraries/networking/src/udt/PacketHeaders.h
ZappoMan: ok - that's fine but for future reference bump both

davidkelly:
since I could hear the issue by just being in sandbox, I qa'd by just going to sandbox. Sounds perfect.  Just to be sure, played some audio from console which doesn't go through same pathway (plays locally of course) but that sounded right too.

PR #9206, opened 2016-12-14
Fix Oculus Touch flags check

Atlante45:
Test Plan:
- With Oculus Touch
- When controllers lose tracking or go into standby mode, hand should reset to idle position.

### DIFF
--- a/plugins/oculus/src/OculusControllerManager.cpp
+++ b/plugins/oculus/src/OculusControllerManager.cpp
@@ -206,7 +206,7 @@ void OculusControllerManager::TouchDevice::update(float deltaTime, const control
     }
 
     int numTrackedControllers = 0;
-    static const auto REQUIRED_HAND_STATUS = ovrStatus_OrientationTracked & ovrStatus_PositionTracked;
+    static const auto REQUIRED_HAND_STATUS = ovrStatus_OrientationTracked | ovrStatus_PositionTracked;
     auto tracking = ovr_GetTrackingState(_parent._session, 0, false);
     ovr_for_each_hand([&](ovrHandType hand) {
         ++numTrackedControllers;


plugins/oculus/src/OculusControllerManager.cpp
```diff
@@ -206,7 +206,7 @@ void OculusControllerManager::TouchDevice::update(float deltaTime, const control
     }
 
     int numTrackedControllers = 0;
-    static const auto REQUIRED_HAND_STATUS = ovrStatus_OrientationTracked & ovrStatus_PositionTracked;
+    static const auto REQUIRED_HAND_STATUS = ovrStatus_OrientationTracked | ovrStatus_PositionTracked;
```
ZappoMan: how did that EVER work? :)

Atlante45: Well, this causes the next `if` to believe that the controllers are always tracked so it works as expected when you use them.
But if you put them on your desk and they go into standby mode after a few secs, the hand would stay there instead of resetting to the idle position.

hifi-gustavo:
Build console output for commit: 07802eb284cdfd8542c74d1d78c14c353ee4edd5 available at: https://gist.github.com/7447a82aa82f8c57b5a37559ba7d1dfb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9206-07802eb.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9206-07802eb.dmg

birarda:
testing this

birarda:
Dropped controllers to desk and saw hands go idle. Pass.

PR #9209, opened 2016-12-14
Add framework for default frame handling, default frame for Oculus

jherico:
This PR adds behavior when no frame has been submitted to the display plugin, allowing it to render something custom.  This is typically for HMDs where it's preferable to have some simple loading screen instead of black space.  

## Testing

Start Interface in Oculus mode.  If the application doesn't start rendering right away you should see a color cube in the HMD for a second or two until the rest of the application initializes and starts rendering.


hifi-gustavo:
Build console output for commit: 5c0de385d006e8897867540b0d7699252c4928a3 available at: https://gist.github.com/b005eef1b80c672a3aaf604595d2e3ac<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9209-5c0de38.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9209-5c0de38.dmg

OmegaHeron:
Tested Windows 10 - DK2 HMD

General smoke test passes.  I was never able to get a default frame while starting Interface with Oculus on multiple times.  I'm not saying the PR doesn't work -- I don't think I ever hit a condition where it could work.

birarda:
@jherico A good suggestion from @Atlante45 - would it be possible now to load the display plugin as early as possible during application start so that the default frame is displayed as quickly as possible?

OmegaHeron:
That makes sense in a lot of ways @birarda -- before some time back the oculus plugin loaded first (or somewhere near it)... there was an update that sorted load order and it changed, obviously, when oculus loads - it went from, for instance, loading oculus runtime almost before you saw interface main window appear to sometime after it.  I suspect that's why I couldn't see the default as I was assured content streaming in before oculus plug ever loaded...

jherico:
@birarda @Atlante45 Re-ordering how initialization works would be tricky and probably fragile.  

If we can definitively know we're launching from the Oculus store we could possibly do some special case code to render the default frame, similar to the way we do the SteamVR min spec check.  But that would take a big chunk of additional work, maybe 2 engineer days.

ZappoMan:
soooo old... and it has merge conflicts... closing it out!

PR #9227, opened 2016-12-16
Session display names

howard-stearns:
Adds a sessionDisplayName to Avatar data. It is set by the server based on the user's chosen displayName, if any.

Part of https://highfidelity.atlassian.net/wiki/display/HOME/People+Action+List


howard-stearns:
Test Plan:
This is a protocol change, so you must use this PR's Interface and Sandbox together.
MyAvatar.sessionDisplayName should always have a non-empty string value when connected to a domain. (You can see it in the javascript console.)
The value should have the following characteristics:
- It should be based on MyAvatar.displayName
- Leading/trailing space is removed.
- A very limited number of American four-letter curse words are replaced with "*".
- If the base name ends up empty (see next), the base "anonymous" is used.
- If you change your displayName, you do NOT get a new MyAvatar.sessiondDisplayName (unless you leave and return). The sessionDisplayName should remain the same throughought the session.
- If there are multiple avatars in the domain with the same basic name, a "_1" is appended to the second, "_2" to the third, etc.
  - The numbering should result in each avatar AT THE TIME having a unique name.
  - As people come and go, lower numbers are NOT resused until the last person with the same base name has gone. At that point, the next person in with that same base name should not have a "_n", and the next should have "_1".  If you leave and return you are not likely to get the same number.
  - The few recognized curses are not considered unique from each other.
  - Any trailing whitespace..."_n..." in a name is removed, so as not to spoof an appended result. Thus "joe_27" and "joe" are both considered to be the same base name, with the first user getting the sessionDisplayName of "joe", and the second one "joe_1".
- This should work through rapid changes, spotty connections, etc. At least, as much as anything works under such conditions.
- If you bring up the "People Action List" (run Interface script system/pal.js from the Running Scripts dialog, and click the "People" toolbar button), the sessionDisplayName should be shown for yourself (at top), and for all other avatars present.

Limitations:
- Only a few naughty names are filtered.
- This does NOT assign special names to AC scripted agents.

zfox23:
[PAL display names](https://trello.com/c/TfjJaFUz/29-pal-display-names)

hifi-gustavo:
Build console output for commit: 70af96669c77493af72231dea637c6e6e831d061 available at: https://gist.github.com/704d6b792d989d7699842da0d0e7123f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9227-70af966.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9227-70af966.dmg

OmegaHeron:
Ok - just one comment before I fully get into testing this....

As an admin tool (that's why PAL is, right?), why would I want profanities filtered from display names?  I want to see what it is that some is/is not doing that might be/might !be "offensive"...  If someone has a nasty display name then that would presumably show unfiltered via the script that shows display name bubbles?  Just confused why sanitizing is a value added function for one watching what the visitors to his/her space are up to.

OmegaHeron:
I can't get this to work under following setup

Sandbox running on Windows 10 - my first AV connected to that sandbox via "localhost" - second av coming from second machine on LAN -  I bring second AV into my sandbox and I see this with PAL on

hifi-snap-by-omegaheron-on-2016-12-16_19-21-13

I'll pass this one back for others to test as they may have better results than I - I saw some crazy crazy stuff - my Windows machine could not see my second AV, my Linux machine could not see my windows machine AV but could see itself as my primary AV, not being of light as it should have been.  I reverted to stack/interface 5815 and all works as expected... AV 1 = my normal crappy mixamo fuse one while AV 2 = being of light.

howard-stearns:
good question. The PAL is for everybody, basically to replace the hard-to-click moving arrows on avatars that get put up by the "ignore"/"kick" toolbar button.  Just like that that button, it uses the same UI for all users -- it's just that admins get some extra stuff.

In this case, the extra stuff that the admins get are extra buttons, and also usernames when available. However, we want everyone to have the same names visible. For example, suppose someone in Welcome says, "Hey, greeter, there's a guy out behind the hanger attacking someone.  I've kicked him in my PAL, with the name "joe 3", but I can't stop the attack."  In this case, we want to make sure that the greeter has the same "joe 3" showing up on her PAL.  Meanwhile, we don't want people's obnoxious displayNames to be used as a source of griefing itself!  So, if we sanitize displayNames, we have to do it for everyone.

howard-stearns:
I'll see if I can repro the testing problems. Just to make sure -- all the connecting machines were using the same client from this PR, yes?  (For example, building from source for Linux could produce some skew based on exactly when you pulled from master. There's been some funky merges today...)

OmegaHeron:
All machines running same - my linux build box pulls the specific PR branch while the windows machine used binaries provided here.

Also - there's one other question I have and that's about what exactly is an admin.  Am I an admin while I'm on my own domain running on my own hardware and able to see all that extra data/buttons?

howard-stearns:
For the purpose of PAL projects, "admin" means someone for whom Users.canKick is true.

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -227,6 +241,27 @@ void AvatarMixer::broadcastAvatarData() {
             // setup a PacketList for the avatarPackets
             auto avatarPacketList = NLPacketList::create(PacketType::BulkAvatarData);
 
+            if (avatar.getSessionDisplayName().isEmpty() &&  // We haven't set it yet...
+                nodeData->getReceivedIdentity()) { // ... but we have processed identity (with possible displayName).
+                QString baseName = avatar.getDisplayName().trimmed();
```
davidkelly: This maybe could be a function, just for readability, like what you did for sendIdentityPacket.

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -416,6 +441,16 @@ void AvatarMixer::nodeKilled(SharedNodePointer killedNode) {
         && killedNode->getLinkedData()) {
         auto nodeList = DependencyManager::get<NodeList>();
 
+        {  // decrement sessionDisplayNames table and possibly remove
+           QMutexLocker nodeDataLocker(&killedNode->getLinkedData()->getMutex());
+           AvatarMixerClientData* nodeData = dynamic_cast<AvatarMixerClientData*>(killedNode->getLinkedData());
+           const QString& baseDisplayName = nodeData->getBaseDisplayName();
```
davidkelly: I think it is possible for getLinkedData() to return null, and this next line will dereference that null.  Sorry I see the guard comment, but don't follow.

assignment-client/src/avatars/AvatarMixer.cpp
howard-stearns: That's a good idea. I'll do that if I make other changes. We'll know more in testing. (sendIdentityPacket is actually called twice, so it's clearly justified there.)

assignment-client/src/avatars/AvatarMixer.cpp
howard-stearns: Two different, unrelated guards:
- The first two lines of the body of the method ensure that getLinkedData() isn't null. (New line 441.) That's not the guard I refer to in the comment on new line 448.
- In line 449, I use the [] operator on a QHash. That has the non-obvious effect of adding an entry to _sessionDisplayNames if the baseDisplayName isn't already there. The entry is put there on newline 255 (which also has a comment about the implicit behavior of []). But 255 only happens when an identity packet has been received. So, it is theoretically possible that we could somehow connect and then immediately get a nodeKilled packet before receiving an identity packet, and thus not already have an entry here for this baseDisplayName. (I don't think we can, actually, because I think the identity packet is the first one we get, and it's a so-called "reliable" send. But just in case...).  So, in that rare case, one might not like the idea of creating the entry in the hash, only to immediately remove it on line 450. When I first wrote it, I put a contains() guard around this code, but decided that it wasn't worth it: the adding+removing will still be end up with the correct behavior, and it's worth it in this exceptionally rare case to not have to do two hash lookups in the normal case.  That's what the 448 comment is about.  Not sure how to reword that to have it make sense. Suggestions?

By the way, in this bizarre state, baseDisplayName will be the empty string, which I don't think effects the analysis.

assignment-client/src/avatars/AvatarMixer.cpp
davidkelly: 👍

hifi-gustavo:
Build console output for commit: 9d9345cb65f7ddc974bc1bb041643cdf00567118 available at: https://gist.github.com/704d6b792d989d7699842da0d0e7123f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9227-9d9345c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9227-9d9345c.dmg

hifi-gustavo:
Build console output for commit: a45e4da75f84bef73d0bf2ea5c65ee9b7620b997 available at: https://gist.github.com/704d6b792d989d7699842da0d0e7123f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9227-a45e4da.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9227-a45e4da.dmg

hifi-gustavo:
Build console output for commit: 60cb7acb505ac1bc853f03cda8a1f8e8a6490d01 available at: https://gist.github.com/704d6b792d989d7699842da0d0e7123f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9227-60cb7ac.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9227-60cb7ac.dmg

davidkelly:
Verified that you see a session display name based on anonymous if you have no displayName set, and you are not logged in.  If you set the display name, the next time you pop into the domain, you see a session display name based on display name in PAL.  When 2 have same display name, you get an _1 for second one.  If that guy pops out and back in, you get _2, etc...  Changing display name during a session doesn't change the name shown in PAL, as expected.  Note if you login _during_ a session, you keep the display name you got when not logged in.  FWIW I think at some point, we should make this display your login name, but that is perhaps not something for this PR.

QA Pass for me.

howard-stearns:
Good question re sessiondDisplayName changes and logout:

The idea of the PAL list is that it is stable enough to find the person you want. The following could cause a PAL display and order change if we updated while it is open:
a. People exiting or entering.
b. People changing their displayName.
c. People changing their login status (we currently don't sort by login name, so no order change here).

I'm imagining that a future version might have a refresh button, that changes display to indicate when there has been some change. Theoretically:
- a change to the set of sessionIds.
- a login status change, and thus a change to username display for admins only. (Note that login/logout retains the same sessionUUID, and that the various actions are specified from Interface to the server visa sessionUUID. For ban, that resolves to username|fingerprint at the server.)
- a display name change (but I'm worried about coordinating everyone's display, because the idea of sessionDisplayName is that it's the same for everyone).

My instinct is to not go too far down that road just yet. Keep it simple until we get more experience with it.

PR #9245, opened 2016-12-20
measure avatar processing stats in Chrome trace framework

AndrewMeadows:
* measure `app.AvatarsPerSec` and `app.JointsPerSec` in chrome trace output

### DIFF
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -75,6 +75,19 @@ namespace render {
     }
 }
 
+static uint64_t timeProcessingJoints = 0;
+static int32_t numJointsProcessed = 0;
+
+float Avatar::getNumJointsProcessedPerSecond() {
+    float rate = 0.0f;
+    if (timeProcessingJoints > 0) {
+        rate = (float)(numJointsProcessed * USECS_PER_SECOND) / (float)timeProcessingJoints;
+    }
+    timeProcessingJoints = 0;
+    numJointsProcessed = 0;
+    return rate;
+}
+
 Avatar::Avatar(RigPointer rig) :
     AvatarData(),
     _skeletonOffset(0.0f),
@@ -281,7 +294,7 @@ void Avatar::updateAvatarEntities() {
 void Avatar::simulate(float deltaTime) {
     PerformanceTimer perfTimer("simulate");
 
-    if (!isDead() && !_motionState) {
+    if (!_motionState && !isDead()) {
         DependencyManager::get<AvatarManager>()->addAvatarToSimulation(this);
     }
     animateScaleChanges(deltaTime);
@@ -319,6 +332,7 @@ void Avatar::simulate(float deltaTime) {
         }
     }
 
+    uint64_t start = usecTimestampNow();
     if (_shouldAnimate && !_shouldSkipRender && (avatarPositionInView || avatarMeshInView)) {
         {
             PerformanceTimer perfTimer("skeleton");
@@ -345,6 +359,8 @@ void Avatar::simulate(float deltaTime) {
         PerformanceTimer perfTimer("skeleton");
         _skeletonModel->simulate(deltaTime, false);
     }
+    timeProcessingJoints += usecTimestampNow() - start;
+    numJointsProcessed += _jointData.size();
 
     // update animation for display name fade in/out
     if ( _displayNameTargetAlpha != _displayNameAlpha) {
@@ -627,6 +643,7 @@ glm::quat Avatar::computeRotationFromBodyToWorldUp(float proportion) const {
 }
 
 void Avatar::fixupModelsInScene() {
+#ifdef FOO // adebug
     _attachmentsToDelete.clear();
 
     // check to see if when we added our models to the scene they were ready, if they were not ready, then
@@ -650,6 +667,7 @@ void Avatar::fixupModelsInScene() {
     _attachmentsToDelete.insert(_attachmentsToDelete.end(), _attachmentsToRemove.begin(), _attachmentsToRemove.end());
     _attachmentsToRemove.clear();
     scene->enqueuePendingChanges(pendingChanges);
+#endif // adebug
 }
 
 bool Avatar::shouldRenderHead(const RenderArgs* renderArgs) const {
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -643,7 +643,6 @@ glm::quat Avatar::computeRotationFromBodyToWorldUp(float proportion) const {
 }
 
 void Avatar::fixupModelsInScene() {
-#ifdef FOO // adebug
     _attachmentsToDelete.clear();
 
     // check to see if when we added our models to the scene they were ready, if they were not ready, then
@@ -667,7 +666,6 @@ void Avatar::fixupModelsInScene() {
     _attachmentsToDelete.insert(_attachmentsToDelete.end(), _attachmentsToRemove.begin(), _attachmentsToRemove.end());
     _attachmentsToRemove.clear();
     scene->enqueuePendingChanges(pendingChanges);
-#endif // adebug
 }
 
 bool Avatar::shouldRenderHead(const RenderArgs* renderArgs) const {
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -294,7 +294,7 @@ void Avatar::updateAvatarEntities() {
 void Avatar::simulate(float deltaTime) {
     PerformanceTimer perfTimer("simulate");
 
-    if (!_motionState && !isDead()) {
+    if (!isDead() && !_motionState) {
         DependencyManager::get<AvatarManager>()->addAvatarToSimulation(this);
     }
     animateScaleChanges(deltaTime);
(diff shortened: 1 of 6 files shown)


interface/src/avatar/Avatar.cpp
```diff
@@ -75,6 +75,19 @@ namespace render {
...
+
+float Avatar::getNumJointsProcessedPerSecond() {
+    float rate = 0.0f;
+    if (timeProcessingJoints > 0) {
+        rate = (float)(numJointsProcessed * USECS_PER_SECOND) / (float)timeProcessingJoints;
+    }
+    timeProcessingJoints = 0;
+    numJointsProcessed = 0;
+    return rate;
+}
```
AndrewMeadows: The measurements of the joint processing times must live in the Avatar context, but they are harvested in the AvatarManager context, so we expose it with this static API.

interface/src/avatar/Avatar.cpp
```diff
@@ -281,7 +294,7 @@ void Avatar::updateAvatarEntities() {
 void Avatar::simulate(float deltaTime) {
     PerformanceTimer perfTimer("simulate");
 
-    if (!isDead() && !_motionState) {
+    if (!_motionState && !isDead()) {
```
AndrewMeadows: Ooops, I'll roll this change back.

interface/src/avatar/Avatar.cpp
```diff
@@ -345,6 +359,8 @@ void Avatar::simulate(float deltaTime) {
         PerformanceTimer perfTimer("skeleton");
         _skeletonModel->simulate(deltaTime, false);
     }
+    timeProcessingJoints += usecTimestampNow() - start;
+    numJointsProcessed += _jointData.size();
```
AndrewMeadows: We tally the times and number of joints here.  They totals are reset after they are harvested.

interface/src/avatar/AvatarManager.cpp
```diff
@@ -165,6 +166,10 @@ void AvatarManager::updateOtherAvatars(float deltaTime) {
 
     // simulate avatar fades
     simulateAvatarFades(deltaTime);
+
+    PROFILE_COUNTER(app, "NumAvatarsPerSec",
+            { { "NumAvatarsPerSec", (float)(size() * USECS_PER_SECOND) / (float)(usecTimestampNow() - start) } });
+    PROFILE_COUNTER(app, "NumJointsPerSec", { { "NumJointsPerSec", Avatar::getNumJointsProcessedPerSecond() } });
```
AndrewMeadows: These stats were supposed to have the form:  `simulation_avatar.NumFooPerSec` however I couldn't figure out how to make it work without using `app` as the parent context.  Maybe someone can help me fix this later.

libraries/shared/src/Profile.h
```diff
@@ -22,6 +22,7 @@ Q_DECLARE_LOGGING_CATEGORY(trace_resource_parse)
 Q_DECLARE_LOGGING_CATEGORY(trace_resource_network)
 Q_DECLARE_LOGGING_CATEGORY(trace_simulation)
 Q_DECLARE_LOGGING_CATEGORY(trace_simulation_animation)
+Q_DECLARE_LOGGING_CATEGORY(trace_simulation_avatar)
```
AndrewMeadows: These `trace_simulation_avatar` declarations were added with the hope of using them for the `NumFooPerSec` stats but I must be missing something since I couldn't get them to work.

hifi-gustavo:
Build console output for commit: 991a61b7222b9977314d68ec0a06e6b6cb0947e6 available at: https://gist.github.com/5f3628d465ea70337756b5d7fd975d7e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9245-991a61b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9245-991a61b.dmg

AndrewMeadows:
To test:

(1) Copy this script to a local file: test-pr-9245.js
```
var startTime = 0;
var endTime = startTime + 40;

Script.setTimeout(function() {
    var loggingRules = "" +
        "trace.*=false\n" +
        "trace.render.debug=true\n" +
        "trace.app.debug=true\n" +
        "";
    Test.startTracing(loggingRules);
}, startTime * 1000);


Script.setTimeout(function() {
    Test.stopTracing("test-pr-9245.json.gz");
    Test.quit();
}, endTime * 1000);
```

(2) Run the interface client via the command line by navigating to the directory with the interface.exe executable and invoking it like this:
    `./interface.exe --testScript path/to/script/int/part1/test-pr-9145.js`
**Be sure to login to a domain that already has at least one other avatar.**  The interface should exit after 40 seconds (as per the Test.quit() call in the final callback).    Note, the JS Test API will be unavailable unless the --testScript option is successfully passed to the interface executable on startup.

(3) Find the resulting test-pr-9245.json.gz file in the current directory.

(4) Open the Chrome browser and navigate to:  Chrome://tracing

(5) Press the Load button and select the test-pr-9145.json.gz file via the file-picker.
--> you should see two per-frame measurements of NumAvatarsPerSec and NumJointsPerSec in the data.  The values are not important yet.  Note: if there were never any other avatars in the domain then those measurements will not show up at all (see step (2)).

hifi-gustavo:
Build console output for commit: 4d711d869c35ec7c49d1154382cc7210a7bc1b14 available at: https://gist.github.com/5f3628d465ea70337756b5d7fd975d7e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9245-4d711d8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9245-4d711d8.dmg

interface/src/avatar/AvatarManager.cpp
```diff
@@ -165,6 +166,10 @@ void AvatarManager::updateOtherAvatars(float deltaTime) {
 
     // simulate avatar fades
     simulateAvatarFades(deltaTime);
+
+    PROFILE_COUNTER(simulation_avatar, "NumAvatarsPerSec",
+            { { "NumAvatarsPerSec", (float)(size() * USECS_PER_SECOND) / (float)(usecTimestampNow() - start) } });
+    PROFILE_COUNTER(simulation_avatar, "NumJointsPerSec", { { "NumJointsPerSec", Avatar::getNumJointsProcessedPerSecond() } });
```
jherico: unless it's critical that you get this every frame, you may want to make this a `SAMPLE_PROFILE_COUNTER`.  just add a float that indicates how often through the code path it should trigger... with 1.0 being always and 0.0 being never. 

So for one sample per 100 frames, you'd do...

`SAMPLE_PROFILE_COUNTER(0.01f, simulation_avatar, ...)`

libraries/shared/src/Profile.cpp
```diff
@@ -18,6 +18,7 @@ Q_LOGGING_CATEGORY(trace_resource_network, "trace.resource.network")
 Q_LOGGING_CATEGORY(trace_resource_parse, "trace.resource.parse")
 Q_LOGGING_CATEGORY(trace_simulation, "trace.simulation")
 Q_LOGGING_CATEGORY(trace_simulation_animation, "trace.simulation.animation")
+Q_LOGGING_CATEGORY(trace_simulation_avatar, "trace.simulation.avatar")
```
jherico: Just as an FYI, you don't have to put this into the global header.  In particular, if you're only going to use a category in one spot, you can also just say `Q_LOGGING_CATEGORY(trace_simulation_avatar, "trace.simulation.avatar")` in the cpp file you use it in, with no `Q_DECLARRE_LOGGING_CATEGORY` required.

hifi-gustavo:
Build console output for commit: fa18d548f7273cf0e94fd89dec587d08a6c155a2 available at: https://gist.github.com/5f3628d465ea70337756b5d7fd975d7e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9245-fa18d54.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9245-fa18d54.dmg

hifi-gustavo:
Build console output for commit: e43efa74de0708dffbbaf49e0548e87013029a82 available at: https://gist.github.com/5f3628d465ea70337756b5d7fd975d7e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9245-e43efa7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9245-e43efa7.dmg

OmegaHeron:
Running QA now

OmegaHeron:
Problem I run into is when starting interface up from a script it doesn't return to last connected domain - it just hangs in disconnected space.  Have no idea where Interface is going to - it's not localhost or anywhere "live".  I thought I might add a move AV to domain line, but, inexplicably, I have no idea how to do so nor can I find it in the JS docs in wiki........

OmegaHeron:
Testing Windows 10
QA Pass

Verify the trace data file was written in the directory you set in the script: test-pr-9245.json.gz. *Yes, Pass*

--> you should see two per-frame measurements of NumAvatarsPerSec and NumJointsPerSec in the data. *Yes, Pass*

9245trace

Smoke test pass.

Still not sure why interface running in test mode doesn't return to last known AV posit - I **thought** it did so when I did QA on original addition of auto test param to interface, but, that was - in QA space - a long time ago.

PR #9255, opened 2016-12-21
optimization of incoming avatar joint data in interface client

AndrewMeadows:
* faster math when copying joint data into animation rig
* remove unnecessary function call overhead when updating palm positions
* avoid unnecessary copies
* avoid scanning for animation pose overrides when there are none

### DIFF
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -1293,23 +1293,10 @@ void Avatar::setOrientation(const glm::quat& orientation) {
 void Avatar::updatePalms() {
     PerformanceTimer perfTimer("palms");
     // update thread-safe caches
-    glm::quat rotation;
-    int i = _skeletonModel->getLeftHandJointIndex();
-    if (_skeletonModel->getJointRotationInWorldFrame(i, rotation)) {
-        _leftPalmRotationCache.set(rotation);
-    }
-    glm::vec3 position;
-    if (_skeletonModel->getJointPositionInWorldFrame(i, position)) {
-        _leftPalmPositionCache.set(position);
-    }
-
-    i = _skeletonModel->getRightHandJointIndex();
-    if (_skeletonModel->getJointRotationInWorldFrame(i, rotation)) {
-        _rightPalmRotationCache.set(rotation);
-    }
-    if (_skeletonModel->getJointPositionInWorldFrame(i, position)) {
-        _rightPalmPositionCache.set(position);
-    }
+    _leftPalmRotationCache.set(getUncachedLeftPalmRotation());
+    _rightPalmRotationCache.set(getUncachedRightPalmRotation());
+    _leftPalmPositionCache.set(getUncachedLeftPalmPosition());
+    _rightPalmPositionCache.set(getUncachedRightPalmPosition());
 }
 
 void Avatar::setParentID(const QUuid& parentID) {
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -275,13 +275,16 @@ void Avatar::updateAvatarEntities() {
         }
 
         AvatarEntityIDs recentlyDettachedAvatarEntities = getAndClearRecentlyDetachedIDs();
-        _avatarEntitiesLock.withReadLock([&] {
-            foreach (auto entityID, recentlyDettachedAvatarEntities) {
-                if (!_avatarEntityData.contains(entityID)) {
-                    entityTree->deleteEntity(entityID, true, true);
+        if (!recentlyDettachedAvatarEntities.empty()) {
+            // only lock this thread when absolutely necessary
+            _avatarEntitiesLock.withReadLock([&] {
+                foreach (auto entityID, recentlyDettachedAvatarEntities) {
+                    if (!_avatarEntityData.contains(entityID)) {
+                        entityTree->deleteEntity(entityID, true, true);
+                    }
                 }
-            }
-        });
+            });
+        }
     });
 
     if (success) {
@@ -299,18 +302,25 @@ void Avatar::simulate(float deltaTime) {
     }
     animateScaleChanges(deltaTime);
 
-    bool avatarPositionInView = false;
-    bool avatarMeshInView = false;
+    bool avatarInView = false;
     { // update the shouldAnimate flag to match whether or not we will render the avatar.
         PerformanceTimer perfTimer("cull");
-        ViewFrustum viewFrustum;
         {
-            PerformanceTimer perfTimer("LOD");
+            // simple frustum check
+            PerformanceTimer perfTimer("inView");
+            ViewFrustum viewFrustum;
+            qApp->copyDisplayViewFrustum(viewFrustum);
+            avatarInView = viewFrustum.sphereIntersectsFrustum(getPosition(), getBoundingRadius())
+                || viewFrustum.boxIntersectsFrustum(_skeletonModel->getRenderableMeshBound());
+        }
+        PerformanceTimer lodPerfTimer("LOD");
+        if (avatarInView) {
             const float MINIMUM_VISIBILITY_FOR_ON = 0.4f;
             const float MAXIMUM_VISIBILITY_FOR_OFF = 0.6f;
+            ViewFrustum viewFrustum;
             qApp->copyViewFrustum(viewFrustum);
             float visibility = calculateRenderAccuracy(viewFrustum.getPosition(),
-                    getBounds(), DependencyManager::get<LODManager>()->getOctreeSizeScale());
+                getBounds(), DependencyManager::get<LODManager>()->getOctreeSizeScale());
             if (!_shouldAnimate) {
                 if (visibility > MINIMUM_VISIBILITY_FOR_ON) {
                     _shouldAnimate = true;
@@ -321,19 +331,11 @@ void Avatar::simulate(float deltaTime) {
                 qCDebug(interfaceapp) << "Optimizing" << (isMyAvatar() ? "myself" : getSessionUUID()) << "for visibility" << visibility;
             }
         }
-
-        {
-            PerformanceTimer perfTimer("inView");
-            // simple frustum check
-            float boundingRadius = getBoundingRadius();
-            qApp->copyDisplayViewFrustum(viewFrustum);
-            avatarPositionInView = viewFrustum.sphereIntersectsFrustum(getPosition(), boundingRadius);
-            avatarMeshInView = viewFrustum.boxIntersectsFrustum(_skeletonModel->getRenderableMeshBound());
-        }
     }
 
     uint64_t start = usecTimestampNow();
-    if (_shouldAnimate && !_shouldSkipRender && (avatarPositionInView || avatarMeshInView)) {
+    // CRUFT? _shouldSkipRender is never set 'true'
+    if (_shouldAnimate && avatarInView && !_shouldSkipRender) {
         {
             PerformanceTimer perfTimer("skeleton");
             _skeletonModel->getRig()->copyJointsFromJointData(_jointData);
@@ -643,6 +645,8 @@ glm::quat Avatar::computeRotationFromBodyToWorldUp(float proportion) const {
 }
 
 void Avatar::fixupModelsInScene() {
+#define ADEBUG
+#ifdef ADEBUG
     _attachmentsToDelete.clear();
 
     // check to see if when we added our models to the scene they were ready, if they were not ready, then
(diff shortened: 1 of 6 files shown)


interface/src/avatar/Avatar.cpp
```diff
@@ -275,13 +275,16 @@ void Avatar::updateAvatarEntities() {
...
+            _avatarEntitiesLock.withReadLock([&] {
+                foreach (auto entityID, recentlyDettachedAvatarEntities) {
+                    if (!_avatarEntityData.contains(entityID)) {
+                        entityTree->deleteEntity(entityID, true, true);
+                    }
                 }
-            }
-        });
+            });
+        }
```
AndrewMeadows: Do not unnecessarily lock Avatar::_avatarEntityData when there is nothing to do.

interface/src/avatar/Avatar.cpp
```diff
@@ -299,18 +302,25 @@ void Avatar::simulate(float deltaTime) {
...
         PerformanceTimer perfTimer("cull");
-        ViewFrustum viewFrustum;
         {
-            PerformanceTimer perfTimer("LOD");
+            // simple frustum check
+            PerformanceTimer perfTimer("inView");
+            ViewFrustum viewFrustum;
+            qApp->copyDisplayViewFrustum(viewFrustum);
+            avatarInView = viewFrustum.sphereIntersectsFrustum(getPosition(), getBoundingRadius())
+                || viewFrustum.boxIntersectsFrustum(_skeletonModel->getRenderableMeshBound());
```
AndrewMeadows: Avoid the second expensive intersection test if the first one succeeds.

interface/src/avatar/Avatar.cpp
```diff
@@ -725,7 +727,7 @@ glm::vec3 Avatar::getDisplayNamePosition() const {
     glm::vec3 bodyUpDirection = getBodyUpDirection();
     DEBUG_VALUE("bodyUpDirection =", bodyUpDirection);
 
-    if (getSkeletonModel()->getNeckPosition(namePosition)) {
+    if (_skeletonModel->getNeckPosition(namePosition)) {
```
AndrewMeadows: Avoid the overhead of calling Avatar::getSkeletonModel() inside the Avatar class implementation, which has direct access to _skeletonModel.

hifi-gustavo:
Build console output for commit: 0b0c3f0be4cee9aeea4a78f2239bd247082903fb available at: https://gist.github.com/8e6990fb0d3f343523572a35a8759edd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9255-0b0c3f0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9255-0b0c3f0.dmg

libraries/animation/src/Rig.cpp
```diff
@@ -1192,28 +1211,26 @@ void Rig::applyOverridePoses() {
...
+        absolutePosesOut.resize(relativePoses.size());
+        AnimPose geometryToRigTransform(_geometryToRigTransform);
+        for (int i = 0; i < (int)relativePoses.size(); i++) {
+            int parentIndex = _animSkeleton->getParentIndex(i);
+            if (parentIndex == -1) {
+                // transform all root absolute poses into rig space
+                absolutePosesOut[i] = geometryToRigTransform * relativePoses[i];
+            } else {
+                absolutePosesOut[i] = absolutePosesOut[parentIndex] * relativePoses[i];
+            }
```
AndrewMeadows: Hrm... I think I added a PerformanceTimer instance in this context, then removed it, and forgot to remove the context brackets.  I'll fix.

libraries/animation/src/Rig.cpp
```diff
@@ -1192,28 +1211,26 @@ void Rig::applyOverridePoses() {
...
         }
     }
-
-    // transform all absolute poses into rig space.
-    AnimPose geometryToRigTransform(_geometryToRigTransform);
-    for (int i = 0; i < (int)absolutePosesOut.size(); i++) {
-        absolutePosesOut[i] = geometryToRigTransform * absolutePosesOut[i];
-    }
```
AndrewMeadows: The important change to notice here is that by multiplying all of the root transforms by geometryToRigTransform in the previous loop further up we can eliminate the final individual transform of each joint.

libraries/animation/src/Rig.cpp
```diff
@@ -1251,62 +1268,36 @@ void Rig::copyJointsIntoJointData(QVector<JointData>& jointDataVec) const {
...
+            const JointData& data = jointDataVec.at(i);
+            _internalPoseSet._relativePoses[i].scale = Vectors::ONE;
+            _internalPoseSet._relativePoses[i].rot = rotations[i];
+            if (data.translationSet) {
+                // JointData translations are in scaled relative-frame so we scale back to regular relative-frame
+                _internalPoseSet._relativePoses[i].trans = _invGeometryOffset.scale * data.translation;
+            } else {
+                _internalPoseSet._relativePoses[i].trans = relativeDefaultPoses[i].trans;
             }
         }
```
AndrewMeadows: The optimizations in `Rig::copyJointsFromJointData()` are a little hard to follow because there is much transformations being done but the summary is:

(1) the transform between the Rig-frame and the Geometry-frame was just a rotation of π about the y-axis and much of that math can be avoided by transforming just the incoming joint data from Rig into Geometry.

(2) The `overrideFlag` is not used for `Avatar` instances... only for `MyAvatar` (not what we're dealing with here), so there is no need to check that bit or set it.

(3) Typically the vast majority of joints have changed rotations (that is, different from default pose) which means it is wasteful to check if the rotation has changed (it is almost always true anyway). Instead we just copy ALL rotation results into `_internalPoseSet`.

(4) We do check `data.translationSet` because... we don't even bother doing any linear translations until the very end, because the translation data was packed in the parent-relative frame.

AndrewMeadows:
Your mileage may vary but with all of these changes I was seeing a speedup of about 30% in the NumAvatarsPerSecond and NumJointsPerSecond trace metrics for the following setup:

OS = linux
build = Release
RAM = 16GB
CPU = i7
avatar count = 100

To test:

(0) run through steps (1) through (6) for this PR and for another already merged PR: PR-9245 or newer.

(1) Copy this code into a script called `test-PR-9255.js`:

```
var startTime = 0;
var endTime = startTime + 40;

Window.location = "hifi://dev-chris.highfidelity.io/-26.2322,-0.942479,51.8355";
Script.setTimeout(function() {
    var loggingRules = "" +
        "trace.*=false\n" +
        "trace.render.debug=true\n" +
        "trace.app.debug=true\n" +
        "trace.simulation.*=true\n"
        "";
    Test.startTracing(loggingRules);
}, startTime * 1000);


Script.setTimeout(function() {
    Test.stopTracing("path/trace-PR-9255.json.gz"); // Change path to something with write permissions
    Test.quit();
}, endTime * 1000);
```
(2) Start the interface from the command line with the following arguments:

`interface --testScript C:/path/to/your/script/test-PR-9255.js`

You should automatically connect to a domain with a big red box for a floor and about 100 NPC avatars running the walk animation.

(3) Wait for the avatars to load and interface to automatically quit (about 40 seconds after the script starts).  When it is done check to make sure the` trace-PR-9255.json.gz` file was successfully written to disk.  You probably should move this to the side with a unique name since you need to run through this sequence at least twice.

(4) In chrome browser navigate to `chrome://tracing`

(5) Press the Load button and use the file browser to select the trace file you want.

(6) Measure the peak values near the end of the trace of the `NumAvatarsPerSec` metric, after all the downloads are done.

(7) Compare the peak values in  step (6) for this PR vs PR-9245 (or newer) -- the values in this PR should be higher.

hifi-gustavo:
Build console output for commit: 890e35e96ec38e1aa24c892ac4b6a9efdf2bc064 available at: https://gist.github.com/8e6990fb0d3f343523572a35a8759edd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9255-890e35e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9255-890e35e.dmg

libraries/animation/src/Rig.cpp
hyperlogic: This code is actually smaller and easier to read now, good work.

interface/src/avatar/Avatar.cpp
AndrewMeadows: Yes, I would expect the bounding sphere to be more likely to succeed, which is why we test it first.  The old code would OR these two tests together, which is what happens here.  There is no reason to pick one to be first except for optimization.

interface/src/avatar/Avatar.cpp
```diff
@@ -1291,10 +1293,23 @@ void Avatar::setOrientation(const glm::quat& orientation) {
...
+        _leftPalmPositionCache.set(position);
+    }
+
+    i = _skeletonModel->getRightHandJointIndex();
+    if (_skeletonModel->getJointRotationInWorldFrame(i, rotation)) {
+        _rightPalmRotationCache.set(rotation);
+    }
+    if (_skeletonModel->getJointPositionInWorldFrame(i, position)) {
+        _rightPalmPositionCache.set(position);
+    }
```
AndrewMeadows: Yeah, it looks like I changed the behavior here for the translation part.  I'll have to study it to come up with a fix that is still optimized (if any).

AndrewMeadows: The `updatePalms()` call I was trying to optimize is inside `Avatar::simulate()`.  `MyAvatar` overrideds that method and never calls the base version, which means the context I was trying to optimize was for _other_ `Avatars` only.  Which raises the question:  "Why do we even bother to compute the palm positions for _other_ avatars?  Maybe we only need to know them for the `MyAvatar` instance?"

After an audit of the code I'm convinced they are not used for non-MyAvatar context in the C++.  Meanwhile, none of the existing new or old JS scripts in our official codebase use them, however it is remotely possible that some user is relying on the feature for some mysterious bit of content that monitors another Avatar's hand motions.

From my measurments caching the palm transforms using the inefficient code costs about 500 usec per 100 avatars.  Moving the Avatar.getRightHandPosition() API up to the MyAvatar context would save that 500 usec at the expense of a feature (that probably isn't being used yet).  Alternatively I could maybe shave 100 usec with some effort.

AndrewMeadows: It occurs to me... we recently had some changes where the AvatarMixer doesn't bother sending the full joint data for `Avatars` that are out of view, or who are far away.  This means that the states of _other_ `Avatar`'s palms may not even be valid were a script to query them --> the unused feature is kinda sometimes broken already.

AndrewMeadows:
I decided to roll back the changes to `Avatar::updatePalms()` which regains 500 usec of cost for processing 100 avatars (linux Release).  There are opportunities for cleanup there but to move forward would require design/feature decisions that need a consensus.

hifi-gustavo:
Build console output for commit: b3be7f0f3eedea240de5fd05401ce0dcc2c62977 available at: https://gist.github.com/8e6990fb0d3f343523572a35a8759edd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9255-b3be7f0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9255-b3be7f0.dmg

hifi-gustavo:
Build console output for commit: 4161775673c5709cbd76d7ee133d284b9f2c60e8 available at: https://gist.github.com/8e6990fb0d3f343523572a35a8759edd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9255-4161775.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9255-4161775.dmg

OmegaHeron:
Window 10 - i5 6500 CPU - GTX970GPU

Looked at NumAvatarsPerSec at ~35s

Master 5835 Run 1 17188.56 (Time 35051.3)
Master 5835 Run 2 16126.46 (Time 35131.46)

PR 9255 Run 1 23554.1 (Time 35120.35)
PR 9255 Run 2 43779.8 (Time 35041.07)

I don't see anything broken - seems to pass basic smoke tests.  Not assigning pass/fail as I'm unsure of the numbers real meanings.

hifi-gustavo:
Build console output for commit: a5efc08473ed05209a89a782617e3dde77c4eaf1 available at: https://gist.github.com/8e6990fb0d3f343523572a35a8759edd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9255-a5efc08.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9255-a5efc08.dmg

interface/src/avatar/Avatar.cpp
ZappoMan: I realize this code is better than the old code, but I'm wondering if the second clause `viewFrustum.boxIntersectsFrustum(_skeletonModel->getRenderableMeshBound());` is *ever* true with the first clause is false... My point is, maybe we can just completely eliminate the second clause, and make the code simpler to read and understand.

PR #9258, opened 2016-12-22
Immediate visual ignore

zfox23:
Currently, clicking "Ignore" for an avatar results in a ~5 second delay before the other avatar actually disappears. However, the audio for the ignored avatar cuts out immediately.

This PR causes the ignored avatar to disappear immediately, removing the delay and enabling behavior parity between the audio and visuals.

### DIFF
--- a/assignment-client/src/avatars/AvatarMixer.cpp
+++ b/assignment-client/src/avatars/AvatarMixer.cpp
@@ -558,6 +558,10 @@ void AvatarMixer::handleKillAvatarPacket(QSharedPointer<ReceivedMessage> message
 }
 
 void AvatarMixer::handleNodeIgnoreRequestPacket(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode) {
+    auto killPacket = NLPacket::create(PacketType::KillAvatar, NUM_BYTES_RFC4122_UUID + sizeof(KillAvatarReason));
+    killPacket->write(message->peek(NUM_BYTES_RFC4122_UUID));
+    killPacket->writePrimitive(KillAvatarReason::AvatarIgnored);
+    DependencyManager::get<NodeList>()->sendUnreliablePacket(*killPacket, *senderNode);
     senderNode->parseIgnoreRequestMessage(message);
 }
 


hifi-gustavo:
Build console output for commit: e7681e1e2738d9d7ef9f308df9313f9a0e1385fe available at: https://gist.github.com/a705cdac388829cf1360c359ce371ea2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9258-e7681e1.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9258-e7681e1.dmg

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -558,6 +558,10 @@ void AvatarMixer::handleKillAvatarPacket(QSharedPointer<ReceivedMessage> message
 }
 
 void AvatarMixer::handleNodeIgnoreRequestPacket(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode) {
+    auto killPacket = NLPacket::create(PacketType::KillAvatar, NUM_BYTES_RFC4122_UUID + sizeof(KillAvatarReason));
+    killPacket->write(message->peek(NUM_BYTES_RFC4122_UUID));
+    killPacket->writePrimitive(KillAvatarReason::AvatarIgnored);
+    DependencyManager::get<NodeList>()->sendUnreliablePacket(*killPacket, *senderNode);
```
ZappoMan: don't you also want to send the ignore packet to the other node as well? this basically makes the person disappear from you, but not you disappear from them... (unless that's handled elsewhere)

zfox23: I could do that for sure. The way this works after this change is such that the original behavior takes over for the other node. In other words: If Avatar A ignores Avatar B, Avatar B will disappear immediately for Avatar A, and it will take a couple seconds for Avatar A to disappear for Avatar B. 

If you'd prefer for the disappearing to happen simultaneously, I can certainly send the kill packet to the other node as well. I figured it was marginally faster (processing-wise) to not send two packets.

ZappoMan: with your PR it will give an unrealistic expectation that you've disappeared from them instantly... when really they still see you for 5 seconds... we should make it consistent all the way around. Unless you've identified some reason that this is not a good idea.

claudioshg:
Tested Windows 10 - CV1 HMD

The ~5 seconds delay is still there.
hifi-snap-by-claudioshg-on-201

zfox23:
Taking this PR down until further notice.

PR #9262, opened 2016-12-22
First attempt at PAL operations with ignore

howard-stearns:
This DNM PR is up so that we can look at the functionality of the PAL with respect to the ignore/unignore lifecycle, and following avatar-mixer optimizations.

Things to try, with two or more interfaces running this PR against the sandbox in this PR:
- Ignore the other avatar using the PAL:
 - The other avatar should see you disappear after a moment, as though you had left.
 - You will see the PAL ignore box checked for this user.
 - You should loose audio from that user.
 - You should see the other avatar disappear, but the clickable avatar node overlay should remain.
 - As the other avatar moves, the overlay should follow.
 - When you close the PAL, the overlay should disappear.
 - When you open the PAL again, the ignored user should be listed and marked as ignored, and the overlay should be visible at the right position. The overlay should indicate when the user is selected or unselected via the overlay or the list.
 - You should be able to click the activated ignore button to unignore. At that point, you and the other user should see and hear each other normally. (N.B.: 9253 addresses the PAL styling, including the checkboxes.)
- For an ordinary user, the above ignore/unignore behavior should ONLY work for those that you ignore. I.e., if A ignores B, and B (who does not have kick rights) opens the PAL, B must NOT be able to see A in the list nor have any overlay.
- For a user with kick permissions, the inclusion of an avatar in the list and overlay should include those that have ignored you. (The ignore button and state still refers to you ignoring them, not vice versa). I.e., if griefer ignores admin, both dissapear from each other. But if admin opens the PAL, he can see the griefer in the list, and their position with the overlay, and can take actions (such as ban).
- Generally, open PAL from either interface, in various combinations of one or both avatars:
 - newly arrived;
 - facing away from each other;
 - very far apart from each other;
 - various admin/non-admin and login combinations. N.B. that localhost typically has kick rights regardless of login state!
 - for avatars that are not in view, we don't _currently_ show the overlay, but we might later (e.g., around the edges of your display to indicate where to turn to find people, or some other "radar"). Check to make sure (e.g., in the console) that their avatar.position is changing when they move around.
 - Users.ignore(sessionIdString) and Users.unignore(sessionIdString) should work as you expect when the PAL is not open. (You can other sessionIDs in the console with JSON.stringify(AvatarList.getAvatarIdentifiers()).)
The PAL should populate, display sessionDisplayName, and for admins, the username|fingerprint.

This PR might not be how we want to implement this behavior. We're first checking out the behavior, and allowing further development. For example:
- I'm not sure what to test for interactions with physics, and whether or not this behavior is correct.
- This implementation sends identity/data packets in some PAL-opened conditions, that we might want to change this to be different packet types, a subset of the data, or at different frequencies, etc. (There's some special casing of avatar identity/data packets that might wouldn't be necessary if there's a separate packet type.)
- There's a proliferation of packet types, such as one for ignore and other for unignore, rather than one with a flag. I haven't made any attempt to simplify these. Maybe it's fine as is. (My point is that I haven't even thought about it, pending a sanity check on the actual behavior.)
- There's a proliferation of javascript slots and messages, such as Users.ignore and unignore. Same issues.
- I've deliberately left logging in for now, that should be removed before merge.
- There are likely cases where we don't get avatar data fast enough upon opening the PAL. Possibilities include a refresh button, or to have pal.js send new users whenever it notices that it is adding a new node in updateOverlays, or even when removing a node there. (There's some special handling of ignored nodes in Pal.qml that might not be necessary if we do that.)
- There's some data that should be cleared when changing domains. (See Pal.qml comments, but maybe more.)
- Instead of keeping track of ignore state in the Pal.js, we should really expose that to .js/.qml and ask it. That way things will stay in sync if some other code ignores (such as mod.js or equivalent).
- Generally, see FIXME (FIXME HRS) added by this PR.

### DIFF
--- a/assignment-client/src/avatars/AvatarMixer.cpp
+++ b/assignment-client/src/avatars/AvatarMixer.cpp
@@ -347,7 +347,6 @@ void AvatarMixer::broadcastAvatarData() {
                         && (forceSend
                             || otherNodeData->getIdentityChangeTimestamp() > _lastFrameTimestamp
                             || distribution(generator) < IDENTITY_SEND_PROBABILITY)) {
-                        qDebug() << "FIXME HRS sending identity to" << node->getUUID() << "from" << otherNode->getUUID() << "gets mine/all/view:" << getsIgnoredByMe << getsAnyIgnored << getsOutOfView ;
                         sendIdentityPacket(otherNodeData, node);
                     }
 
@@ -417,7 +416,6 @@ void AvatarMixer::broadcastAvatarData() {
                                         ? AvatarData::SendAllData : AvatarData::IncludeSmallData;
                         nodeData->incrementAvatarInView();
                     }
-                    //qDebug() << "FIXME HRS sending" << detail << "to" << node->getUUID() << "from" << otherNode->getUUID();
 
                     numAvatarDataBytes += avatarPacketList->write(otherNode->getUUID().toRfc4122());
                     numAvatarDataBytes += avatarPacketList->write(otherAvatar.toByteArray(detail));
@@ -552,14 +550,12 @@ void AvatarMixer::handleViewFrustumPacket(QSharedPointer<ReceivedMessage> messag
 void AvatarMixer::handleRequestsDomainListDataPacket(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode) {
     auto nodeList = DependencyManager::get<NodeList>();
     nodeList->getOrCreateLinkedData(senderNode);
-    qDebug() << "HRS FIXME received RequestsDomainListData packet from" << senderNode->getUUID();
 
     if (senderNode->getLinkedData()) {
         AvatarMixerClientData* nodeData = dynamic_cast<AvatarMixerClientData*>(senderNode->getLinkedData());
         if (nodeData != nullptr) {
             bool isRequesting;
             message->readPrimitive(&isRequesting);
-            qDebug() << "HRS FIXME handling RequestsDomainListData packet" << isRequesting << "from" << nodeData->getNodeID();
             nodeData->setRequestsDomainListData(isRequesting);
         }
     }
--- a/assignment-client/src/avatars/AvatarMixer.cpp
+++ b/assignment-client/src/avatars/AvatarMixer.cpp
@@ -110,6 +110,9 @@ void AvatarMixer::broadcastAvatarData() {
     const float CURRENT_FRAME_RATIO = 1.0f / TRAILING_AVERAGE_FRAMES;
     const float PREVIOUS_FRAMES_RATIO = 1.0f - CURRENT_FRAME_RATIO;
 
+    // only send extra avatar data (avatars out of view, ignored) every Nth AvatarData frame
+    const int EXTRA_AVATAR_DATA_FRAME_RATIO = 10; 
+    
     // NOTE: The following code calculates the _performanceThrottlingRatio based on how much the avatar-mixer was
     // able to sleep. This will eventually be used to ask for an additional avatar-mixer to help out. Currently the value
     // is unused as it is assumed this should not be hit before the avatar-mixer hits the desired bandwidth limit per client.
@@ -368,6 +371,11 @@ void AvatarMixer::broadcastAvatarData() {
                     AvatarDataSequenceNumber lastSeqToReceiver = nodeData->getLastBroadcastSequenceNumber(otherNode->getUUID());
                     AvatarDataSequenceNumber lastSeqFromSender = otherNodeData->getLastReceivedSequenceNumber();
 
+                    // this throttles the extra data to only be sent every Nth message
+                    if (getsOutOfView && lastSeqToReceiver % EXTRA_AVATAR_DATA_FRAME_RATIO > 0) {
+                        return;
+                    }
+                    
                     if (lastSeqToReceiver > lastSeqFromSender && lastSeqToReceiver != UINT16_MAX) {
                         // we got out out of order packets from the sender, track it
                         otherNodeData->incrementNumOutOfOrderSends();
--- a/assignment-client/src/avatars/AvatarMixer.cpp
+++ b/assignment-client/src/avatars/AvatarMixer.cpp
@@ -111,7 +111,7 @@ void AvatarMixer::broadcastAvatarData() {
     const float PREVIOUS_FRAMES_RATIO = 1.0f - CURRENT_FRAME_RATIO;
 
     // only send extra avatar data (avatars out of view, ignored) every Nth AvatarData frame
-    const int EXTRA_AVATAR_DATA_FRAME_RATIO = 10; 
+    const int EXTRA_AVATAR_DATA_FRAME_RATIO = 16; 
     
     // NOTE: The following code calculates the _performanceThrottlingRatio based on how much the avatar-mixer was
     // able to sleep. This will eventually be used to ask for an additional avatar-mixer to help out. Currently the value
@@ -370,11 +370,6 @@ void AvatarMixer::broadcastAvatarData() {
 
                     AvatarDataSequenceNumber lastSeqToReceiver = nodeData->getLastBroadcastSequenceNumber(otherNode->getUUID());
                     AvatarDataSequenceNumber lastSeqFromSender = otherNodeData->getLastReceivedSequenceNumber();
-
-                    // this throttles the extra data to only be sent every Nth message
-                    if (getsOutOfView && lastSeqToReceiver % EXTRA_AVATAR_DATA_FRAME_RATIO > 0) {
-                        return;
-                    }
                     
                     if (lastSeqToReceiver > lastSeqFromSender && lastSeqToReceiver != UINT16_MAX) {
                         // we got out out of order packets from the sender, track it
@@ -400,15 +395,21 @@ void AvatarMixer::broadcastAvatarData() {
                     nodeData->setLastBroadcastSequenceNumber(otherNode->getUUID(),
                                                              otherNodeData->getLastReceivedSequenceNumber());
 
-                    // start a new segment in the PacketList for this avatar
-                    avatarPacketList->startSegment();
-
                     // determine if avatar is in view, to determine how much data to include...
                     glm::vec3 otherNodeBoxScale = (otherNodeData->getPosition() - otherNodeData->getGlobalBoundingBoxCorner()) * 2.0f;
                     AABox otherNodeBox(otherNodeData->getGlobalBoundingBoxCorner(), otherNodeBoxScale);
+                    bool isInView = nodeData->otherAvatarInView(otherNodeBox);
+
+                    // this throttles the extra data to only be sent every Nth message
+                    if (!isInView && getsOutOfView && (lastSeqToReceiver % EXTRA_AVATAR_DATA_FRAME_RATIO > 0)) {
+                        return;
+                    }
+
+                    // start a new segment in the PacketList for this avatar
+                    avatarPacketList->startSegment();
 
                     AvatarData::AvatarDataDetail detail;
-                    if (!nodeData->otherAvatarInView(otherNodeBox) && !getsOutOfView) {
+                    if (!isInView && !getsOutOfView) {
                         detail = AvatarData::MinimumData;
                         nodeData->incrementAvatarOutOfView();
                     } else {
--- a/assignment-client/src/avatars/AvatarMixer.cpp
+++ b/assignment-client/src/avatars/AvatarMixer.cpp
@@ -51,7 +51,7 @@ AvatarMixer::AvatarMixer(ReceivedMessage& message) :
     packetReceiver.registerListener(PacketType::NodeIgnoreRequest, this, "handleNodeIgnoreRequestPacket");
     packetReceiver.registerListener(PacketType::NodeUnignoreRequest, this, "handleNodeUnignoreRequestPacket");
     packetReceiver.registerListener(PacketType::RadiusIgnoreRequest, this, "handleRadiusIgnoreRequestPacket");
-    packetReceiver.registerListener(PacketType::RequestDomainListData, this, "handleRequestDomainListDataPacket");
+    packetReceiver.registerListener(PacketType::RequestsDomainListData, this, "handleRequestsDomainListDataPacket");
 
     auto nodeList = DependencyManager::get<NodeList>();
     connect(nodeList.data(), &NodeList::packetVersionMismatch, this, &AvatarMixer::handlePacketVersionMismatch);
@@ -536,18 +536,18 @@ void AvatarMixer::handleViewFrustumPacket(QSharedPointer<ReceivedMessage> messag
     }
 }
 
-void AvatarMixer::handleRequestDomainListDataPacket(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode) {
+void AvatarMixer::handleRequestsDomainListDataPacket(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode) {
     auto nodeList = DependencyManager::get<NodeList>();
     nodeList->getOrCreateLinkedData(senderNode);
-    qDebug() << "HRS FIXME received requestDomainListData packet from" << senderNode->getUUID();
+    qDebug() << "HRS FIXME received RequestsDomainListData packet from" << senderNode->getUUID();
 
     if (senderNode->getLinkedData()) {
         AvatarMixerClientData* nodeData = dynamic_cast<AvatarMixerClientData*>(senderNode->getLinkedData());
         if (nodeData != nullptr) {
             bool isRequesting;
             message->readPrimitive(&isRequesting);
-            qDebug() << "HRS FIXME handling requestDomainListData packet" << isRequesting << "from" << nodeData->getNodeID();
-            nodeData->setRequestDomainListData(isRequesting);
+            qDebug() << "HRS FIXME handling RequestsDomainListData packet" << isRequesting << "from" << nodeData->getNodeID();
+            nodeData->setRequestsDomainListData(isRequesting);
         }
     }
 }
--- a/assignment-client/src/avatars/AvatarMixer.cpp
+++ b/assignment-client/src/avatars/AvatarMixer.cpp
@@ -585,6 +585,9 @@ void AvatarMixer::handleNodeIgnoreRequestPacket(QSharedPointer<ReceivedMessage> 
     senderNode->parseIgnoreRequestMessage(message);
 }
 
+void AvatarMixer::handleNodeUnignoreRequestPacket(QSharedPointer<ReceivedMessage> packet, SharedNodePointer sendingNode) {
+    sendingNode->parseUnignoreRequestMessage(packet);
+}
 void AvatarMixer::handleRadiusIgnoreRequestPacket(QSharedPointer<ReceivedMessage> packet, SharedNodePointer sendingNode) {
     sendingNode->parseIgnoreRadiusRequestMessage(packet);
 }
--- a/assignment-client/src/avatars/AvatarMixer.cpp
+++ b/assignment-client/src/avatars/AvatarMixer.cpp
@@ -49,6 +49,7 @@ AvatarMixer::AvatarMixer(ReceivedMessage& message) :
     packetReceiver.registerListener(PacketType::AvatarIdentity, this, "handleAvatarIdentityPacket");
     packetReceiver.registerListener(PacketType::KillAvatar, this, "handleKillAvatarPacket");
     packetReceiver.registerListener(PacketType::NodeIgnoreRequest, this, "handleNodeIgnoreRequestPacket");
+    packetReceiver.registerListener(PacketType::NodeUnignoreRequest, this, "handleNodeUnignoreRequestPacket");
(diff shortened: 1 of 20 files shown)


hifi-gustavo:
Build console output for commit: bb076990b569d5722947b6f775281ddb403eae8a available at: https://gist.github.com/e5a3cd7ea3f05a42dc6711ecdbe3473d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-bb07699.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-bb07699.dmg

hifi-gustavo:
Build console output for commit: 642fd02785d2c1f89a36a983fe749be02a3947e4 available at: https://gist.github.com/e5a3cd7ea3f05a42dc6711ecdbe3473d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-642fd02.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-642fd02.dmg

hifi-gustavo:
Build console output for commit: b77822131dd018109fed21de5ca2cbf1a1018a5a available at: https://gist.github.com/e5a3cd7ea3f05a42dc6711ecdbe3473d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-b778221.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-b778221.dmg

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -205,6 +207,11 @@ void AvatarMixer::broadcastAvatarData() {
             // use the data rate specifically for avatar data for FRD adjustment checks
             float avatarDataRateLastSecond = nodeData->getOutboundAvatarDataKbps();
 
+            // send extra data that is otherwise surpressed
+            bool getsOutOfView = nodeData->getRequestsDomainListData();
+            bool getsIgnoredByMe = nodeData->getRequestsDomainListData();
```
zfox23: Any reason not to make `getsIgnoredByMe = getsOutOfView` instead of reaching into nodeData for a function call and read op?

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -205,6 +207,11 @@ void AvatarMixer::broadcastAvatarData() {
             // use the data rate specifically for avatar data for FRD adjustment checks
             float avatarDataRateLastSecond = nodeData->getOutboundAvatarDataKbps();
 
+            // send extra data that is otherwise surpressed
```
zfox23: I propose adding to this comment to explain what "getRequestsDomainListData()" does, i.e.:
`// Right now, "getRequestsDomainListData()" returns whether or not the user currently has the PAL open.`
`// Other functions may request this extra data in the future `

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -205,6 +207,11 @@ void AvatarMixer::broadcastAvatarData() {
...
+            bool getsOutOfView = nodeData->getRequestsDomainListData();
```
zfox23: I feel like these variable names should be explained.
`// "getsOutOfView" will force the AvatarMixer to send full avatar data to clients for whom it is true.`

assignment-client/src/avatars/AvatarMixer.cpp
zfox23: `// When this is true, the AvatarMixer will send Avatar data to a client about avatars that they've ignored.`

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -205,6 +207,11 @@ void AvatarMixer::broadcastAvatarData() {
...
+            bool getsIgnoredByMe = nodeData->getRequestsDomainListData();
+            bool getsAnyIgnored = getsIgnoredByMe && node->getCanKick();
```
zfox23: `// When this is true, the AvatarMixer will send Avatar B's data to Avatar A, even if Avatar B has ignored Avatar A`
`// This is useful for domain administrators.`

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -333,7 +340,7 @@ void AvatarMixer::broadcastAvatarData() {
                         && (forceSend
                             || otherNodeData->getIdentityChangeTimestamp() > _lastFrameTimestamp
                             || distribution(generator) < IDENTITY_SEND_PROBABILITY)) {
-
+                        qDebug() << "FIXME HRS sending identity to" << node->getUUID() << "from" << otherNode->getUUID() << "gets mine/all/view:" << getsIgnoredByMe << getsAnyIgnored << getsOutOfView ;
```
zfox23: Any idea what "HRS" means?

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -388,14 +396,15 @@ void AvatarMixer::broadcastAvatarData() {
...
-                    if (!nodeData->otherAvatarInView(otherNodeBox)) {
+                    if (!nodeData->otherAvatarInView(otherNodeBox) && !getsOutOfView) {
                         detail = AvatarData::MinimumData;
                         nodeData->incrementAvatarOutOfView();
                     } else {
                         detail = distribution(generator) < AVATAR_SEND_FULL_UPDATE_RATIO
                                         ? AvatarData::SendAllData : AvatarData::IncludeSmallData;
                         nodeData->incrementAvatarInView();
                     }
+                    //qDebug() << "FIXME HRS sending" << detail << "to" << node->getUUID() << "from" << otherNode->getUUID();
```
zfox23: What's this for besides debugging? Is there anything to fix?

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -527,6 +536,22 @@ void AvatarMixer::handleViewFrustumPacket(QSharedPointer<ReceivedMessage> messag
     }
 }
 
+void AvatarMixer::handleRequestsDomainListDataPacket(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode) {
+    auto nodeList = DependencyManager::get<NodeList>();
+    nodeList->getOrCreateLinkedData(senderNode);
+    qDebug() << "HRS FIXME received RequestsDomainListData packet from" << senderNode->getUUID();
```
zfox23: HRS again?

interface/resources/qml/hifi/Pal.qml
```diff
@@ -188,8 +188,8 @@ Item {
             NameCard {
                 id: nameCard
                 // Properties
-                displayName: styleData.value
-                userName: model.userName
+                displayName: styleData.value;
+                userName: model && model.userName;
```
zfox23: Thanks for this bugfix. Please remove semicolons for consistency.

interface/resources/qml/hifi/Pal.qml
```diff
@@ -188,8 +188,8 @@ Item {
             NameCard {
                 id: nameCard
                 // Properties
-                displayName: styleData.value
-                userName: model.userName
+                displayName: styleData.value;
```
zfox23: Remove semicolon.

interface/resources/qml/hifi/Pal.qml
```diff
@@ -202,15 +202,26 @@ Item {
...
                     datum[styleData.role] = model[styleData.role] = newValue
-                    Users[styleData.role](model.sessionId)
-                    // Just for now, while we cannot undo things:
-                    userData.splice(model.userIndex, 1)
-                    sortModel()
+                    var key = styleData.role;
+                    if (!newValue) {
+                        key = 'un' + key;
+                    }
+                    if (styleData.role === 'ignore') {
```
zfox23: I'm going to be switching this section around a bit when I make the ignored{} list non-local. For now, I think it's fine.

interface/resources/qml/hifi/Pal.qml
```diff
@@ -294,6 +305,7 @@ Item {
     property var userData: []
     property var myData: ({displayName: "", userName: ""}) // valid dummy until set
     property bool iAmAdmin: false
+    property var ignored: ({}); // FIXME: reset when changing domains
```
zfox23: I'm confident this shouldn't be local. My "Personal Mute" changes should make this list grabble from the Mixers.

interface/resources/qml/hifi/Pal.qml
```diff
@@ -312,6 +324,16 @@ Item {
             myData = data[myIndex];
             data.splice(myIndex, 1);
             userData = data;
+            var ignoredID, index;
```
zfox23: Same deal, this part will be switched around after I make ignored{} non-local.

interface/src/avatar/AvatarManager.h
```diff
@@ -93,7 +93,6 @@ private slots:
     // virtual overrides
     virtual AvatarSharedPointer newSharedAvatar() override;
     virtual AvatarSharedPointer addAvatar(const QUuid& sessionUUID, const QWeakPointer<Node>& mixerWeakPointer) override;
-
```
zfox23: hehe :)

libraries/networking/src/Node.cpp
```diff
@@ -90,8 +90,14 @@ void Node::parseIgnoreRequestMessage(QSharedPointer<ReceivedMessage> message) {
...
 }
 
+void Node::parseUnignoreRequestMessage(QSharedPointer<ReceivedMessage> message) {
+    QUuid ignoredUUID = QUuid::fromRfc4122(message->readWithoutCopy(NUM_BYTES_RFC4122_UUID));
+    removeIgnoredNode(ignoredUUID);
+}
+
 void Node::addIgnoredNode(const QUuid& otherNodeID) {
     if (!otherNodeID.isNull() && otherNodeID != _uuid) {
+        QReadLocker lock { &_ignoredNodeIDSetLock };
```
zfox23: Does there have to be an associated "unlock" operation after this? Or does that happen automatically?

libraries/networking/src/Node.cpp
```diff
@@ -102,6 +108,20 @@ void Node::addIgnoredNode(const QUuid& otherNodeID) {
     }
 }
 
+void Node::removeIgnoredNode(const QUuid& otherNodeID) {
+    if (!otherNodeID.isNull() && otherNodeID != _uuid) {
+        // insert/find are read locked concurrently. unsafe_erase is not concurrent, and needs a write lock.
+        QWriteLocker lock { &_ignoredNodeIDSetLock };
```
zfox23: Does there have to be an associated "unlock" operation after this? Or does that happen automatically?

libraries/networking/src/NodeList.cpp
```diff
@@ -811,6 +811,38 @@ void NodeList::ignoreNodeBySessionID(const QUuid& nodeID) {
...
+        }, [&nodeID, this](const SharedNodePointer& destinationNode) {
+            // create a reliable NLPacket with space for the ignore UUID
+            auto ignorePacket = NLPacket::create(PacketType::NodeUnignoreRequest, NUM_BYTES_RFC4122_UUID, true);
+
+            // write the node ID to the packet
+            ignorePacket->write(nodeID.toRfc4122());
+
+            qCDebug(networking) << "Sending packet to unignore node" << uuidStringWithoutCurlyBraces(nodeID);
+
+            // send off this ignore packet reliably to the matching node
```
zfox23: *unignore* packet.

libraries/networking/src/NodeList.cpp
```diff
@@ -811,6 +811,38 @@ void NodeList::ignoreNodeBySessionID(const QUuid& nodeID) {
...
+    // enumerate the nodes to send a reliable ignore packet to each that can leverage it
+    if (!nodeID.isNull() && _sessionUUID != nodeID) {
+        eachMatchingNode([&nodeID](const SharedNodePointer& node)->bool {
+            if (node->getType() == NodeType::AudioMixer || node->getType() == NodeType::AvatarMixer) {
+                return true;
+            } else {
+                return false;
+            }
+        }, [&nodeID, this](const SharedNodePointer& destinationNode) {
+            // create a reliable NLPacket with space for the ignore UUID
```
zfox23: *unignore* UUID

libraries/networking/src/NodeList.cpp
```diff
@@ -811,6 +811,38 @@ void NodeList::ignoreNodeBySessionID(const QUuid& nodeID) {
     }
 }
 
+void NodeList::unignoreNodeBySessionID(const QUuid& nodeID) {
+    // enumerate the nodes to send a reliable ignore packet to each that can leverage it
```
zfox23: *unignore* packet

assignment-client/src/avatars/AvatarMixer.cpp
ZappoMan: If I'm understanding this correctly, you appear to be treating the DomainListData as a "send rate" consistent with the per frame send rate of other avatar data. I strongly oppose that choice. You shouldn't need this data at a full 45hz... you should be able to live with people who are out of view, but being requested for additional data at closer to 1-10hz

interface/resources/qml/hifi/Pal.qml
ZappoMan: I think we need to review our coding standard on this... IMO we should follow the same or similar coding standard as JS, since in this case QML is basically piggybacking JS. Our JS standard explicitly requires semicolons at the end of statements like this.

zfox23: A **vast** majority of our QML doesn't contain semicolons at the end of lines where we only define one property on that line. Out of all of our files I looked at for reference when learning QML, none of them contained EOL semicolons.

That doesn't mean I disagree with you - in fact, when I was revamping the look of the PAL, I started off by inserting EOL semicolons. However, when I found that the rest of our QML codebase doesn't have them, I removed them.

I'm not sure what to do here. It seems like we could either add EOL semicolons to all QML files, add them just to this file and change the standard moving forward, or remove them here and keep them removed moving forward. What do you think?

assignment-client/src/avatars/AvatarMixer.cpp
davidkelly: Howard's initials

interface/resources/qml/hifi/Pal.qml
davidkelly: I agree we keep it consistent with the rest of the codebase.  If we want to bring the codebase up to the standard, that should be its own task.

libraries/networking/src/Node.cpp
davidkelly: automatic when the lock object goes out of scope.  You can be explicit (say, to save time if there are lots of statements that don't need the lock before it goes out of scope), but usually used like this.

zfox23: Makes sense, thanks.

interface/resources/qml/hifi/Pal.qml
ZappoMan: I didn't request a change, because yeah...it's a random shit show... I think we should fix this (yes as a different task).

assignment-client/src/avatars/AvatarMixer.cpp
davidkelly: or maybe only one variable in the first place.  I think @howard-stearns was leaving room for these being decoupled at some point?  Perhaps I can leave them in and just make them equal to eachother as you suggested for now...

zfox23: I agree that these might be decoupled in the future, so I think leaving both in is a good idea.

assignment-client/src/avatars/AvatarMixer.cpp
davidkelly: Agreed - perhaps we can send it every Nth frame?

hifi-gustavo:
Build console output for commit: a84ad1a123d8d4dae66ae0b47e766bcb4e284a22 available at: https://gist.github.com/e5a3cd7ea3f05a42dc6711ecdbe3473d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-a84ad1a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-a84ad1a.dmg

hifi-gustavo:
Build console output for commit: 481d33218854ef19a8153d8688b515eb0bd0fe86 available at: https://gist.github.com/e5a3cd7ea3f05a42dc6711ecdbe3473d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-481d332.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-481d332.dmg

hifi-gustavo:
Build console output for commit: 45e595f43d7f907f049b30667770657dfa9fce34 available at: https://gist.github.com/e5a3cd7ea3f05a42dc6711ecdbe3473d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-45e595f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-45e595f.dmg

davidkelly:
The loop runs at 45 hz, so we would always be a factor of that.  Maybe every 8th or 16th one (say first 3 or 4 bits are all 0 in the sequence) would be fine too.  Popped that in as a strawman for us to start with.

zfox23:
Some recent change in here basically broke the functionality - other avatars now disappear in-world and from the PAL (after refresh) a few seconds after I open the PAL. Investigating...

hifi-gustavo:
Build console output for commit: 4fdc093cfab470ca78d155a174b3249eba310ffa available at: https://gist.github.com/e5a3cd7ea3f05a42dc6711ecdbe3473d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-4fdc093.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-4fdc093.dmg

hifi-gustavo:
Build console output for commit: d3c2ba5199b8c46224ce74f4eb277f1e14c6a3c0 available at: https://gist.github.com/e5a3cd7ea3f05a42dc6711ecdbe3473d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-d3c2ba5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-d3c2ba5.dmg

zfox23:
I'd like to merge this in. I've ensured it's in a working albeit imperfect state. Without merging in this change, I can't continue to work on my changes (like personal mute).

hifi-gustavo:
Build console output for commit: 800389cf20be3f0a002c50cd1ff879d5f80caa2d available at: https://gist.github.com/e5a3cd7ea3f05a42dc6711ecdbe3473d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-800389c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9262-800389c.dmg

OmegaHeron:
QA Pass

Tested as follows
Machine 1:  Windows 10 running this PR sandbox and Interface
Machine 2: Ubuntu 16.04.1 running this PR interface

Ignore the other avatar using the PAL:
The other avatar should see you disappear after a moment, as though you had left. **Pass**
You will see the PAL ignore box checked for this user. **Pass**
You should loose audio from that user. **Pass**
You should see the other avatar disappear, but the clickable avatar node overlay should remain. **Pass**
As the other avatar moves, the overlay should follow. **Pass**
When you close the PAL, the overlay should disappear. **Pass**
When you open the PAL again, the ignored user should be listed and marked as ignored, and the overlay should be visible at the right position. The overlay should indicate when the user is selected or unselected via the overlay or the list. **Pass**
You should be able to click the activated ignore button to unignore. At that point, you and the other user should see and hear each other normally. (N.B.: 9253 addresses the PAL styling, including the checkboxes.) **Pass**
For an ordinary user, the above ignore/unignore behavior should ONLY work for those that you ignore. I.e., if A ignores B, and B (who does not have kick rights) opens the PAL, B must NOT be able to see A in the list nor have any overlay. **Pass**
For a user with kick permissions, the inclusion of an avatar in the list and overlay should include those that have ignored you. (The ignore button and state still refers to you ignoring them, not vice versa). I.e., if griefer ignores admin, both dissapear from each other. But if admin opens the PAL, he can see the griefer in the list, and their position with the overlay, and can take actions (such as ban). **Pass**

As for the rest of the extensive test cases/conditions.  It seems to pass within the limits of only having 1 other agent to run against.

zfox23:
Thanks for the wonderful QA work, @OmegaHeron.

zfox23:
[requests domain list data](https://trello.com/c/mhqsJYHf/30-requests-domain-list-data)

zfox23:
This PR has been superseded by #9290.

PR #9279, opened 2016-12-28
Use supplied libquazip5.so for Debug builds

zzmp:
Current Debug builds search for libquazip5d.so, which is not distributed with the software.
This changes Debug to use libquazip5.so, as used for Release.

This fixes Debug builds on Ubuntu 16.04 (`-DCMAKE_BUILD_TYPE=Debug`).

### DIFF
--- a/cmake/externals/quazip/CMakeLists.txt
+++ b/cmake/externals/quazip/CMakeLists.txt
@@ -38,13 +38,13 @@ set(${EXTERNAL_NAME_UPPER}_DLL_PATH ${INSTALL_DIR}/lib CACHE FILEPATH "Location 
 
 if (APPLE)
   set(${EXTERNAL_NAME_UPPER}_LIBRARY_RELEASE ${INSTALL_DIR}/lib/libquazip5.1.0.0.dylib CACHE FILEPATH "Location of QuaZip release library")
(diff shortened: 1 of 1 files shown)


zzmp:
This should change no Release builds, so QA is not needed.

OmegaHeron:
Just to be sure - I built all with Ubuntu 16.04.1 using my methods (similar to all 3rd party builds documented) and pass.

zfox23:
Are there any issues with using the Release object of this library instead of the Debug one in Debug builds? I.e. is there a reason we didn't do this in the first place?

zzmp:
I don't think anybody actively builds Debug in Ubuntu.

I realized that all of my `assert`s were mostly for naught, because I am mostly in Ubuntu.
I needed to change this to get them to compile.

hifi-gustavo:
Build console output for commit: 0bb82d2b73614b31fa703ad0e5ec45fcfecd5f01 available at: https://gist.github.com/0aa348f6689d44b7d15fa0e5a5044e88<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9279-0bb82d2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9279-0bb82d2.dmg

zzmp:
Realized that may not have answered the question...

I know of no case where someone has needed the debug object for libquazip5. I don't even know what libquazip5 does.

As a similar example, we only build with a release version of Qt.

zfox23:
This seems fine, then. :D

PR #9280, opened 2016-12-28
Track dropped frames as reported by the Oculus SDK

Atlante45:
- Allow display plugins to return custom stats
- Track the dropped frames count reported by OVR.

### DIFF
--- a/plugins/oculus/src/OculusDisplayPlugin.cpp
+++ b/plugins/oculus/src/OculusDisplayPlugin.cpp
@@ -147,17 +147,25 @@ void OculusDisplayPlugin::hmdPresent() {
             logWarning("Failed to present");
         }
 
-        static int droppedFrames = 0;
+        static int compositorDroppedFrames = 0;
+        static int appDroppedFrames = 0;
         ovrPerfStats perfStats;
         ovr_GetPerfStats(_session, &perfStats);
         for (int i = 0; i < perfStats.FrameStatsCount; ++i) {
             const auto& frameStats = perfStats.FrameStats[i];
-            int delta = frameStats.CompositorDroppedFrameCount - droppedFrames;
+            int delta = frameStats.CompositorDroppedFrameCount - compositorDroppedFrames;
             _stutterRate.increment(delta);
-            droppedFrames = frameStats.CompositorDroppedFrameCount;
+            compositorDroppedFrames = frameStats.CompositorDroppedFrameCount;
+            appDroppedFrames = frameStats.AppDroppedFrameCount;
         }
+        _hardwareStats["app_dropped_frame_count"] = appDroppedFrames;
     }
     _presentRate.increment();
+}
+
+
+QJsonObject OculusDisplayPlugin::getHardwareStats() const {
+    return _hardwareStats;
 }
 
 bool OculusDisplayPlugin::isHmdMounted() const {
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: f79c6e08f5c40aa0d8c07cd428a6325760fe3e62 available at: https://gist.github.com/f63540a58b2e1e69f3560beb933b7ef0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9280-f79c6e0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9280-f79c6e0.dmg

RyanDowne:
What's the test plan for this?

plugins/oculus/src/OculusDisplayPlugin.cpp
```diff
@@ -147,19 +147,27 @@ void OculusDisplayPlugin::hmdPresent() {
...
         for (int i = 0; i < perfStats.FrameStatsCount; ++i) {
             const auto& frameStats = perfStats.FrameStats[i];
-            int delta = frameStats.CompositorDroppedFrameCount - droppedFrames;
+            int delta = frameStats.CompositorDroppedFrameCount - compositorDroppedFrames;
             _stutterRate.increment(delta);
-            droppedFrames = frameStats.CompositorDroppedFrameCount;
+            compositorDroppedFrames = frameStats.CompositorDroppedFrameCount;
+            appDroppedFrames = frameStats.AppDroppedFrameCount;
         }
+        _hardwareStats["app_dropped_frame_count"] = appDroppedFrames;
```
ZappoMan: we should probably add `compositor_dropped_frame_count` as well

Atlante45:
@RyanDowne Needs to be tested by someone that has access to the DB. Ideally @huffman could do it?

huffman:
@Atlante45 Yeah, I can test that part. Are there already entries in the db?

Atlante45:
There should be some from when I tested myself. Although I changed the names, so maybe not under that key.
I guess we could have a QA tester do a smoke test and you can check the DB then?

huffman:
@Atlante45 Yep, that works. Do they need to be using a DK2/CV1?

Atlante45:
@huffman Should be the same.

hifi-gustavo:
Build console output for commit: f46b8f3e6df4041426b4340b31647f67987462cc available at: https://gist.github.com/f63540a58b2e1e69f3560beb933b7ef0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9280-f46b8f3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9280-f46b8f3.dmg

plugins/oculus/src/OculusDisplayPlugin.cpp
```diff
@@ -147,19 +147,28 @@ void OculusDisplayPlugin::hmdPresent() {
...
+        _hardwareStats["compositor_dropped_frame_count"] = compositorDroppedFrames;
     }
     _presentRate.increment();
 }
 
+
+QJsonObject OculusDisplayPlugin::getHardwareStats() const {
```
jherico: This code is not thread safe and will cause crashes.  The `hmdPresent` function where the JSON object is being populated happens on the render thread, while the code calling `displayPlugin->getHardwareStats` is on the main thread.  You need to add some kind of lock to ensure that the object is not being written while it's being copied, or I guarantee this will eventually be the cause of crashes.

OmegaHeron:
Tested Windows 10 - i5 6500 CPU - GTX970 GPU - DK2

General smoke test = good, didn't see any difference between this and master.

@huffman you should have data from my tests if there's something you want to confirm on the other end.

plugins/oculus/src/OculusDisplayPlugin.cpp
Atlante45: Oh I see, the RateCounter have an atomic to keep track of the rate.
I made the the function thread safe.

OmegaHeron:
Tested again, same setup as above.
Smoke test = pass.

PR #9290, opened 2016-12-30
Personal mute; Quasi-ignore; Reversible PAL actions

zfox23:
This PR adds on to Howard's PR #9262. For a comparison between that PR and this one, [check out this compare page](https://github.com/howard-stearns/hifi/compare/monitored-ignore...zfox23:PersonalMute). I've made significant changes to Howard's PR, so I recommend that we merge _this_ PR into Master.

Standout features of this PR:
- **Adds the concept of "Personal Mute" to HiFi.** Users can now individually mute other users. This is separate from the previous "Mute" behavior, which I've renamed "Silence". Only domain administrators can Silence other users, which mutes them for all users across the domain.
- **Quasi-ignore allows users to ignore others, but still take action on them in the PAL.** Ignored users will still show up in the PAL, and their avatars' overlays will still appear in-world and remain selectable. This way, you can unignore them.
The Avatar Mixer will only send data for quasi-ignored users every 16 frames if the PAL is open; otherwise the Mixer will not send the data.
- **Reversible PAL actions means users can unignore and unmute users they've previously ignored or muted.** Note that administrator actions (Silence and Ban) are currently not reversible via the PAL.


zfox23:
[requests domain list data](https://trello.com/c/mhqsJYHf/30-requests-domain-list-data)

zfox23:
[Personal Mute](https://trello.com/c/CSFKcVR2/39-personal-mute)

hifi-gustavo:
Build console output for commit: 4a21eaa33fbf7a305a0275958e9d98053388faa5 available at: https://gist.github.com/e4018b106c72a08c5beb98ec5787e5b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-4a21eaa.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-4a21eaa.dmg

davidkelly:
QA results:

Both mute and ignore work as expected.  I can unignore or unmute, too.  When ignoring, the avatar disappears when you click now, which is nice.  So both are QA pass AFAICT.  One interesting caveat -- if you are an admin, and Silence someone, they disappear from the PAL.  I believe this didn't happen before, but I'll check in a bit.  If you reopen the PAL after they disappear, they are there again.  Also. when you silence someone, and dismiss the PAL, we don't remember we silenced them when you open PAL again (so, silence is unchecked, though they are still muted).  That is probably just a detail of silence we need to work out, and not an issue with this PR.  The disappearing from the PAL upon silence, that I'm unsure about....

davidkelly:
OH -- turns out that the disappearing when you silence (called mute in master awhile back) was an existing issue.  So, this is QA pass 👍

libraries/networking/src/NodeList.cpp
```diff
@@ -786,42 +786,118 @@ void NodeList::ignoreNodeBySessionID(const QUuid& nodeID) {
...
+            qCDebug(networking) << "Sending packet to" << (destinationNode->getType() == NodeType::AudioMixer ? "AudioMixer" : "AvatarMixer") << "to"
+                << (ignoreEnabled ? "ignore" : "unignore") << "node" << uuidStringWithoutCurlyBraces(nodeID);
 
             // send off this ignore packet reliably to the matching node
             sendPacket(std::move(ignorePacket), *destinationNode);
         });
 
-        QReadLocker setLocker { &_ignoredSetLock };
+        QReadLocker ignoredSetLocker { &_ignoredSetLock }; // write lock for insert and unsafe_erase
+        QReadLocker personalMutedSetLocker{ &_personalMutedSetLock }; // write lock for insert and unsafe_erase
```
davidkelly: These sets are designed for concurrent insertion, but the unsafe_erase is not thread-safe.  So perhaps the unsafe_erase calls should be protected by a write lock, instead of a read lock.  Initially we were only inserting, so the read lock is appropriate for sure, but here we are now erasing also.   We do get a write lock when clearing the set elsewhere.  

FWIW, these concurrent sets seem great - until you start erasing from them.  Then, since you need to insure no traversal or insertion is happening during an erase, you still need to acquire a read/write lock for all the operations.  Perhaps I'm misunderstanding something here, but to me they seem to be buying us very little.  Probably I'm missing something?

interface/resources/qml/controls-uit/CheckBox.qml
```diff
@@ -27,7 +27,9 @@ Original.CheckBox {
     readonly property int checkRadius: 2
 
     style: CheckBoxStyle {
-        indicator: Rectangle {
+        indicator:
+
+        Rectangle {
```
zfox23: You're right - I'm just gonna put this spacing back to how it was.

libraries/networking/src/NodeList.cpp
zfox23: Ah! this makes a lot of sense. I think I understand why I put it in this way, but it's definitely a mistake - great catch.

If, instead, I added a write lock for the "unsafe_erase()" operations and a read lock for the "insert()" operations, would that be sufficient?

zfox23: Like this: 
capture

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -108,6 +109,9 @@ void AvatarMixer::broadcastAvatarData() {
     const float CURRENT_FRAME_RATIO = 1.0f / TRAILING_AVERAGE_FRAMES;
     const float PREVIOUS_FRAMES_RATIO = 1.0f - CURRENT_FRAME_RATIO;
 
+    // only send extra avatar data (avatars out of view, ignored) every Nth AvatarData frame
+    const int EXTRA_AVATAR_DATA_FRAME_RATIO = 16;
```
howard-stearns: Can we get a comment that derives/explains the value 16? How does that relate to Hz, or however we want to think about this. The reason I care is that I eventually want to have out-of-view nodes visible on some sort of "radar" (such as around the edges of the viewing frustum as done in the peopleFinder prototype). In that case, we'll want them moving smoothly enough, and that may require changing this number, and at that time, I'll want to know what/why the number is what it is.

libraries/networking/src/NodeList.cpp
```diff
@@ -786,42 +786,118 @@ void NodeList::ignoreNodeBySessionID(const QUuid& nodeID) {
             } else {
                 return false;
             }
-        }, [&nodeID, this](const SharedNodePointer& destinationNode) {
+        }, [&nodeID, ignoreEnabled, this](const SharedNodePointer& destinationNode) {
             // create a reliable NLPacket with space for the ignore UUID
-            auto ignorePacket = NLPacket::create(PacketType::NodeIgnoreRequest, NUM_BYTES_RFC4122_UUID, true);
+            auto ignorePacket = NLPacket::create(PacketType::NodeIgnoreRequest, NUM_BYTES_RFC4122_UUID + sizeof(bool), true);
```
howard-stearns: I think the old code was broken, in that it was intended to send and receive a list of sessionIDs. I don't think it sent the right length before. I don't know if we still want to support that, but if we do, we should set the right length here, and I think the boolean will need to be first rather than last (unless you want to do more math to unpack it).

libraries/networking/src/NodeList.cpp
howard-stearns: I had kept the concurrent hash because that's what the packet team uses. In that case, the old code (which had did no erasing) did not need a lock for concurrency, but needs a "read lock" now, while the new unsafe_erase code needs a "write" lock.  This is a bit of a hack, in which "read" and "write" don't mean what they say, but rather just define different classes of locking. There's a discussion of this in #developers between myself and Birarda about this.

scripts/system/pal.js
```diff
@@ -132,10 +132,14 @@ function populateUserList() {
...
             Users.requestUsernameFromID(id);
         }
-        data.push(avatarPalDatum);
-        if (id) { // No overlay for ourself.
-            addAvatarNode(id);
+        // Request personal mute status and ignore status
+        // from NodeList (as long as we're not requesting it for our own ID)
+        if (id) {
+            avatarPalDatum['personalMute'] = Users.getPersonalMuteStatus(id);
+            avatarPalDatum['ignore'] = Users.getIgnoreStatus(id);
```
howard-stearns: It's not clear to me that we're picking up ignored items properly. In my version, I maintained a separate list of stuff that was ignored (in .qml), which is pretty hacky. This version is better in maintaining a list of ignored nodes in C++ and getting the status from there. That's great.

However, I don't see how the various lists are maintained properly. We populate by going the though the avatar hash map (AvatarList) in .js, when we open the list (or refresh), but at that time, the avatar hash map won't yet have populated with ignored nodes, because it will take a moment for our requestDomainListJData (sent at the same time) go to the mixer and get identity packets back to populate the avatar hash map.

interface/resources/qml/hifi/Pal.qml
```diff
@@ -169,7 +171,9 @@ Item {
             movable: false
             resizable: false
         }
-        model: userModel
+        model: ListModel {
```
zfox23: This makes lots of sense; I will re-add the separate list.

libraries/networking/src/NodeList.cpp
zfox23: I will talk to @birarda about his original intention when creating this.

libraries/networking/src/NodeList.cpp
```diff
@@ -786,42 +786,118 @@ void NodeList::ignoreNodeBySessionID(const QUuid& nodeID) {
...
+        if (ignoreEnabled) {
+            // add this nodeID to our set of ignored IDs
+            _ignoredNodeIDs.insert(nodeID);
+            // add this nodeID to our set of personal muted IDs
+            _personalMutedNodeIDs.insert(nodeID);
+            emit ignoredNode(nodeID);
+        } else {
+            _ignoredNodeIDs.unsafe_erase(nodeID);
+            _personalMutedNodeIDs.unsafe_erase(nodeID);
+            emit unignoredNode(nodeID);
```
zfox23: OK.

assignment-client/src/avatars/AvatarMixer.cpp
zfox23: Sure, I'll make this more clear.

scripts/system/pal.js
zfox23: This makes sense - I will do some thinking about this and fix it.

libraries/networking/src/NodeList.cpp
birarda: The current master code sends a packet with only one UUID (of size NUM_BYTES_RFC4122_UUID).

libraries/networking/src/NodeList.cpp
zfox23: After our discussion, I'm confident that what I the change I made in the screenshot above is the correct way of doing things - please correct me if I'm wrong.

libraries/networking/src/NodeList.cpp
zfox23: ^Discussed further with @birarda, maybeSendIgnoreSetToNode() can include multiple UUIDs. Thus, I gotta move some stuff around and fix the current implementation. Great catch.

libraries/networking/src/NodeList.cpp
zfox23: After looking into this, I'm not sure we should. Reason being this signal/slot connection: 
```
// when we hear that the user has ignored an avatar by session UUID
// immediately remove that avatar instead of waiting for the absence of packets from avatar mixer
connect(nodeList.data(), "ignoredNode", this, "removeAvatar");
```

Connecting these two means that the removeAvatar() call is receiving "true" for its second argument (the removal reason), which I don't like. How can I easily get around this?

howard-stearns:
Yeah, I don’t know what the right overall design is. (That’s why I wimped out and made a separate ignore/unignore message, consistently all the way through.)
I want our result to be consistent, but I don’t know what that would mean with a stamp-coupled ignore(isIgnored, id) message and signal. If the signal has a reason arg, maybe it’s ignore(isIgnored, id, reason)?

libraries/networking/src/NodeList.cpp
davidkelly: that's what I figured we should do -- LGTM!

libraries/networking/src/NodeList.cpp
zfox23: @howard, putting your reply here:
> Yeah, I don’t know what the right overall design is. (That’s why I wimped out and made a separate ignore/unignore message, consistently all the way through.)
> I want our result to be consistent, but I don’t know what that would mean with a stamp-coupled ignore(isIgnored, id) message and signal. If the signal has a reason arg, maybe it’s ignore(isIgnored, id, reason)?

Are you implying that I should also modify the declaration/definition of removeAvatar() ?

**Alternatively**, I could make a simple wrapper function for the slot associated with the ignoredNode() signal that passes the correct arguments to removeAvatar(). Do you think that is _more_ or _less_ complex than having two separate signals?

scripts/system/pal.js
zfox23: Fixed in upcoming commit.

assignment-client/src/avatars/AvatarMixer.cpp
zfox23: Fixed in upcoming commit.

libraries/networking/src/NodeList.cpp
zfox23: Fixed in upcoming commit.

interface/resources/qml/hifi/Pal.qml
zfox23: Fixed in upcoming commit.

libraries/networking/src/NodeList.cpp
zfox23: Fixed in upcoming commit.

interface/resources/qml/controls-uit/CheckBox.qml
zfox23: Fixed in upcoming commit.

hifi-gustavo:
Build console output for commit: 899dd21328cb790d4c150278e1aadd711474985d available at: https://gist.github.com/e4018b106c72a08c5beb98ec5787e5b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-899dd21.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-899dd21.dmg

interface/resources/qml/controls-uit/CheckBox.qml
```diff
@@ -52,6 +52,19 @@ Original.CheckBox {
             }
 
             Rectangle {
+                id: disabledOverlay
+                visible: !enabled
+                z: 100
```
howard-stearns: let's move this rectangle after the innerBox one (so that it gets drawn after the innerBox), and get rid of the explicit z value.
That will make things easier if we ever need to explicitly futz with z values in some compound higher-level assembly of these.

libraries/networking/src/NodeList.cpp
zfox23: As per our discussion, I'm going to make the wrapper function a lambda function and go for a single slot with a flag argument.

interface/resources/qml/controls-uit/CheckBox.qml
zfox23: Done - thanks!

interface/resources/qml/hifi/Pal.qml
```diff
@@ -183,36 +190,87 @@ Item {
...
             // This NameCard refers to the cell that contains an avatar's
             // DisplayName and UserName
             NameCard {
                 id: nameCard
                 // Properties
                 displayName: styleData.value
-                userName: model.userName
-                audioLevel: model.audioLevel
-                visible: !isCheckBox
+                userName: model ? model.userName : ""
```
howard-stearns: Is model ever falsey?  (here, line 202, 221, 224)

interface/resources/qml/hifi/Pal.qml
```diff
@@ -183,36 +190,87 @@ Item {
...
+            // This Button belongs in the columns that contain the stateless action buttons ("Silence" & "Ban" for now)
+            HifiControls.Button {
+                id: actionButton
+                color: 2 // Red
+                visible: isButton
+                anchors.centerIn: parent
+                width: 24
+                height: 24
+                onClicked: {
+                    if (styleData.role === "mute" || styleData.role === "kick") {
```
howard-stearns: Not critical, but what does this gain us? We're not visible if !isButton, which is is a repeat of this code.
(Seeing duplicated code always makes me question the logic and my understanding, so unless it actually gains us something....) Same issue on line 233.

interface/resources/qml/hifi/Pal.qml
```diff
@@ -183,36 +190,87 @@ Item {
...
                 boxSize: 24
                 onClicked: {
                     var newValue = !model[styleData.role]
-                    var datum = userData[model.userIndex]
-                    datum[styleData.role] = model[styleData.role] = newValue
-                    Users[styleData.role](model.sessionId)
-                    // Just for now, while we cannot undo things:
-                    userData.splice(model.userIndex, 1)
-                    sortModel()
+                    if (newValue === undefined) {
```
howard-stearns: how does this ever happen?

interface/resources/qml/hifi/Pal.qml
zfox23: I don't know why, but I've run into infrequent issues where I'll get some errors in my logs because model is undefined - perhaps when sorting. I'm not sure it'll still happen after my recent changes - I'll remove these lines and do some testing.

interface/resources/qml/hifi/Pal.qml
zfox23: Ah...I wanted to be explicit here in case, for some reason, we came through here and "styleData.role" was set to something we didn't expect. It's for protection's sake. I don't think it'll ever actually be an issue, and it's clearly not necessary, so I'll remove it.

interface/resources/qml/hifi/Pal.qml
zfox23: Same reasoning as before for this - I saw some rare warnings in my logs for this being undefined (usually when stress testing by doing things like sorting repeatedly), but I'm not sure it'll still happen. I'll remove and debug.

zfox23: I can't get the errors to occur when doing the same things as before. Conditional removed.

interface/resources/qml/hifi/Pal.qml
zfox23: Same deal, can't get the errors to occur anymore. It's been awhile since they've happened, anyway, and lots has changed. Reformatting this check and checks like it.

interface/resources/qml/hifi/Pal.qml
```diff
@@ -183,36 +190,87 @@ Item {
...
+            // I'm guessing this is a QT bug and not anything I can fix. I spent too long trying to work around it...
+            // I'm just going to leave the minor visual bug in.
             HifiControls.CheckBox {
+                id: actionCheckBox
                 visible: isCheckBox
                 anchors.centerIn: parent
+                checked: model ? model[styleData.role] : false
+                // If this is a "personal mute" checkbox, and the model is valid, Check the checkbox if the Ignore
+                // checkbox is checked.
+                enabled: styleData.role === "personalMute" ? ((model ? model["ignore"] : false) === true ? false : true) : true
```
howard-stearns: I think 221, 224, 236, and 250 (and related comments) could all be replaced with something like (untested):
  checked: model[styleData.role] && !actionCheckBox.enabled;
  enabled: (styleData.role === 'ignore') || !model.ignore;
The idea is to DRY up the code. "Use the force" of .qml's data-directed backtracking by specifying the property values we want in a declarative style. Then there's not reason to bash in new values, and no reason to re-establish any dependencies by bashing in the same computation that we undid with the first bashing.

interface/resources/qml/hifi/Pal.qml
```diff
@@ -309,18 +366,22 @@ Item {
...
+            width: Math.max(parent.width * 0.75, 400)
+            height: popupText.contentHeight*1.5
             anchors.centerIn: parent
             radius: hifi.dimensions.borderRadius
             color: "white"
             FiraSansSemiBold {
                 id: popupText
-                text: "This is temporary text. It will eventually be used to explain what 'Names' means."
+                text: "Bold names in the list are Avatar Display Names.\n" +
+                    "If a Display Name isn't set, a unique Session Display Name is assigned to them." +
```
howard-stearns: Is this working per @AlanZimmerman ?  (Pedantically, we still create a unique session display name even if a display name is set, because we sanitize it a bit and make the result unique. So the text isn't strictly true. It may be better for users than telling them the truth, though, which is why I ask what the UX is.)

interface/resources/qml/hifi/Pal.qml
zfox23: That's pretty cool...OK, I'll give statements like these a shot.

interface/resources/qml/hifi/Pal.qml
zfox23: I didn't consult Alan at all ^_^. I'll have him take a look at the text.

interface/resources/qml/hifi/Pal.qml
```diff
@@ -349,22 +407,37 @@ Item {
...
         case 'select':
             var sessionId = message.params[0];
             var selected = message.params[1];
             var userIndex = findSessionIndex(sessionId);
-            if (selected) {
-                table.selection.clear(); // for now, no multi-select
-                table.selection.select(userIndex);
-            } else {
-                table.selection.deselect(userIndex);
+            if (userIndex != -1) {
```
howard-stearns: Is this ever not true?
If it it can't happen, let's remove.
If it shouldn't happen and you want to guard against your assumptions being wrong, let's log the failure case.

zfox23: I'll remove it.

interface/resources/qml/hifi/Pal.qml
```diff
@@ -378,11 +451,13 @@ Item {
                 myData.userName = userName;
                 myCard.userName = userName; // Defensive programming
             } else {
-                // Get the index in userModel and userData associated with the passed UUID
+                // Get the index in userModel and userModelData associated with the passed UUID
                 var userIndex = findSessionIndex(userId);
-                // Set the userName appropriately
-                userModel.get(userIndex).userName = userName;
-                userData[userIndex].userName = userName; // Defensive programming
+                if (userIndex != -1) {
```
howard-stearns: same question as above (new line 434).

interface/resources/qml/hifi/Pal.qml
```diff
@@ -393,25 +468,23 @@ Item {
                     myData.audioLevel = audioLevel;
                     myCard.audioLevel = audioLevel; // Defensive programming
                 } else {
-                    console.log("userid:" + userId);
                     var userIndex = findSessionIndex(userId);
-                    userModel.get(userIndex).audioLevel = audioLevel;
-                    userData[userIndex].audioLevel = audioLevel; // Defensive programming
+                    if (userIndex != -1) {
```
howard-stearns: and here

interface/resources/qml/hifi/Pal.qml
zfox23: In this case, I'd like to log the failure case instead.

interface/resources/qml/hifi/Pal.qml
zfox23: In this case, I'd like to log the failure case instead.

interface/resources/qml/hifi/Pal.qml
zfox23: Your statement there didn't produce the same results. I'll think more about how to simplify these statements tomorrow.

hifi-gustavo:
Build console output for commit: 14def61eb8158147eb1087ed67d0e0044a604ab9 available at: https://gist.github.com/e4018b106c72a08c5beb98ec5787e5b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-14def61.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-14def61.dmg

interface/resources/qml/hifi/Pal.qml
```diff
@@ -393,25 +469,25 @@ Item {
                     myData.audioLevel = audioLevel;
                     myCard.audioLevel = audioLevel; // Defensive programming
                 } else {
-                    console.log("userid:" + userId);
```
howard-stearns: Thank you for removing this!

hifi-gustavo:
Build console output for commit: dde5f69ac9ff44505b50c397b77be4a8376b70b7 available at: https://gist.github.com/e4018b106c72a08c5beb98ec5787e5b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-dde5f69.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-dde5f69.dmg

hifi-gustavo:
Build console output for commit: c6958780255a5aec42b4385f4f9f4e30a804a7c1 available at: https://gist.github.com/e4018b106c72a08c5beb98ec5787e5b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-c695878.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-c695878.dmg

hifi-gustavo:
Build console output for commit: 03fbfc85f953274e9d480f11fd1b3714e48772bc available at: https://gist.github.com/e4018b106c72a08c5beb98ec5787e5b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-03fbfc8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-03fbfc8.dmg

hifi-gustavo:
Build console output for commit: ac41e8abc41cf453ed4f469686695324775cbe13 available at: https://gist.github.com/e4018b106c72a08c5beb98ec5787e5b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-ac41e8a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-ac41e8a.dmg

hifi-gustavo:
Build console output for commit: 675159be45bdf617dcaaeff7f7afd8ae1b170c64 available at: https://gist.github.com/e4018b106c72a08c5beb98ec5787e5b9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-675159b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9290-675159b.dmg

PR #9299, opened 2017-01-04
Fix settings hitch

Atlante45:
- Check setting value actually changed
- Set setting thread priority to lowest

Test Plan:
- Deactivate snap turns
- Spin in a circle
- Previous build should have a hitch every 10 sec, this one shouldn't

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1137,6 +1137,7 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer, bo
     _settingsTimer.moveToThread(&_settingsThread);
     _settingsTimer.setSingleShot(false);
     _settingsTimer.setInterval(SAVE_SETTINGS_INTERVAL); // 10s, Qt::CoarseTimer acceptable
+    _settingsThread.setPriority(QThread::LowestPriority);
     _settingsThread.start();
 
     if (Menu::getInstance()->isOptionChecked(MenuOption::FirstPerson)) {
--- a/libraries/shared/src/SettingHandle.cpp
+++ b/libraries/shared/src/SettingHandle.cpp
@@ -74,7 +74,9 @@ void Settings::endGroup() {
 }
 
 void Settings::setValue(const QString& name, const QVariant& value) {
-    _manager->setValue(name, value);
+    if (_manager->value(name) != value) {
+        _manager->setValue(name, value);
+    }
 }
 
 QVariant Settings::value(const QString& name, const QVariant& defaultValue) const {


interface/src/Application.cpp
```diff
@@ -1137,6 +1137,7 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer, bo
     _settingsTimer.moveToThread(&_settingsThread);
     _settingsTimer.setSingleShot(false);
     _settingsTimer.setInterval(SAVE_SETTINGS_INTERVAL); // 10s, Qt::CoarseTimer acceptable
```
jherico: Can we change this to something higher so that we don't have a precise alignment with the rate of stat updates?  If we set it to 30 and we see that one out of every 3 stat updates shows problems, it's going to be much easier to tie it to something to do with settings persistence.

hifi-gustavo:
Build console output for commit: 56a25dfdeeeeaabfac5c8c1dedeb84012904b51d available at: https://gist.github.com/3e96fb25dc9d3feee35e17cc0c69fb6f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9299-56a25df.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9299-56a25df.dmg

jherico:
If the settings persistence is actually triggering a long lock in the main thread, this will presumably reduce the frequency of the problem (by not saving every 10 seconds, even though nothing has changed) but may not solve the underlying problem itself.  When a setting does change, will it still trigger a pause in the main thread?  If so, that seems like it would be a problem.  We should create (or find if we have one) a script that can force a settings change at a given frequency to see if it causes the issue.  If it still recurs, I'd say we still have to solve the underlying problem.

Atlante45:
@austin Agreed, this is just a revert of a change I'd made in the JSON PR, plus I set a priority on the thread. So this should just take us back to the state we were at before.

We need to remove all those Settings call that just forward to QSettings, or make them not forward to QSettings. Because we have no control on when QSettings writes to file. And when we store 100+ settings in one go, QSettings ends up locking us to grab the file handle and writing to it 5+ times. And we have no control over that.
I'm looking at how much it would take to do just that.

OmegaHeron:
An interesting thing about this... I think it was specific to i7.  Never happens on my Windows box with i5 -- spun in circles for several minutes... then decided to check on Linux which runs on an older i7 CPU.  Sure enough... every 10 seconds freeze hits then recovers.

I'll come back with results of this PR on linux box once it compiles, but, it's interesting that we're now starting to see things on i7 similar to what was seen on i5 before some "constraints" on over utilizing a 4 core CPU were put in place.  Food for thought.

OmegaHeron:
Ubuntu 16.04.01 i7 CPU - no longer freezes every 10 seconds.
Windows 10 i5 6500 CPU - smoke test passes (does not exhibit 10 second freeze bug in master).

nbq:
Testing this one now on i7 Windows

PR #9314, opened 2017-01-05
Use a higher thread priority for Oculus present

jherico:
Allow OpenGL plugins to specify priority for the present thread.  Use time critical priority for Oculus.

## Testing

- Smoke test.
- There should be no visible change to the application behavior.
 - Frame dropping might be lower in this build compared to master.
- We should not see a significant increase in CPU usage.
- Make sure to test on i5.
- Make sure other CPU related tasks aren't negatively impacted:
    - audio shouldn't stutter
    - content shouldn't load significantly slower

### DIFF
--- a/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
+++ b/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
@@ -165,6 +165,7 @@ public:
 
                         if (newPlugin) {
                             bool hasVsync = true;
+                            QThread::setPriority(newPlugin->getPresentPriority());
                             bool wantVsync = newPlugin->wantVsync();
                             _context->makeCurrent();
 #if defined(Q_OS_WIN)
--- a/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.h
+++ b/libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.h
@@ -14,6 +14,7 @@
 #include <memory>
 #include <queue>
 
+#include <QtCore/QThread>
 #include <QtCore/QTimer>
 #include <QtGui/QImage>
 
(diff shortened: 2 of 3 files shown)


hifi-gustavo:
Build console output for commit: 9bf4804d3dd95f03e594efff6c3289f9d1f8842f available at: https://gist.github.com/3af673b8cfc5eb8f5191b89a84a555c9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9314-9bf4804.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9314-9bf4804.dmg

OmegaHeron:
Tested Windows 10 - i5 6500 CPU - GTX970 GPU

QA Pass.  

CPU use is not negatively impacted by this PR (in either mode - desktop checked just to be *sure*)
Scenes load, display as expected and at same rate of composing as in master.

Frame drops to Oculus is maybe... a tiny bit better, certainly not worse.

DK2 - old apartment scene - 5 runs, 1 minute each.
Blue = Current master 5866

5866dkvs9314

Desktop - same setup as above

5866dtvs9314

There's still (**nothing to do with this PR**) a weird case where some runs you see batch/gpu time run at approximately 3x normal values.  For example - old apartment runs batch around 2.1ms and GPU at 3.7ms once scene composes and AV is stationary.  Looking at the graphs below you'll see a few odd traces where CPU use is much higher for runtime...  those are accompanied by batch and GPU times of around 9ms and 14ms.  Same with HMD and, an interesting bit is when this happens the recently added missed oculus frame count skyrockets and *never* stops growing.

I only bring that up as it's not particularly rare - going back over my CPU graphs for several months I can see those runs where that's happening often and clearly.  Hopefully some of the new data collection points will shed some light on it someday...  it's not a just me thing, others complain of it -- they just haven't noticed the pattern and metrics showing something is odd.  Typical complaint is -- why does interface rez this scene awesome most of the time -- then another it just dies.  (My paraphrase).

PR #9385, opened 2017-01-14
Fix audio failures

kencooke:
This PR fixes audio failures that occur when a broken driver returns mixFormat that is invalid/unsupported. 

Qt internally uses mixFormat to populate the list of supportedSampleRates(), causing the fallback code to fail when mixFormat is invalid. The new fallback removes all dependence on mixFormat and attempts to find a supported format by iterating over all formats.

One remaining failure case (drivers which only support floating-point samples) will be addressed in a future PR.

### DIFF
--- a/libraries/audio-client/src/AudioClient.cpp
+++ b/libraries/audio-client/src/AudioClient.cpp
@@ -424,35 +424,25 @@ bool adjustedFormatForAudioDevice(const QAudioDeviceInfo& audioDevice,
     adjustedAudioFormat = desiredAudioFormat;
 
     //
-    // Attempt the device sample rate in decreasing order of preference.
+    // Attempt the device sample rate and channel count in decreasing order of preference.
     //
-    if (audioDevice.supportedSampleRates().contains(48000)) {
-        adjustedAudioFormat.setSampleRate(48000);
-    } else if (audioDevice.supportedSampleRates().contains(44100)) {
-        adjustedAudioFormat.setSampleRate(44100);
-    } else if (audioDevice.supportedSampleRates().contains(32000)) {
-        adjustedAudioFormat.setSampleRate(32000);
-    } else if (audioDevice.supportedSampleRates().contains(24000)) {
-        adjustedAudioFormat.setSampleRate(24000);
-    } else if (audioDevice.supportedSampleRates().contains(16000)) {
-        adjustedAudioFormat.setSampleRate(16000);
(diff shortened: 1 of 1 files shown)


kencooke:
### Test Plan

Verify mic input and audio output still work as before (on a normal system, the behavior is unchanged). On a system with broken audio drivers, audio should work where it failed before.

hifi-gustavo:
Build console output for commit: 42e13d1e9771107be1a9574cca0b3d8b8f761940 available at: https://gist.github.com/a63d65d39326316b9adefcb7779c6b75<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9385-42e13d1.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9385-42e13d1.dmg

libraries/audio-client/src/AudioClient.cpp
```diff
@@ -424,35 +424,25 @@ bool adjustedFormatForAudioDevice(const QAudioDeviceInfo& audioDevice,
...
-        // return the nearest in case it needs 2 channels
-        adjustedAudioFormat = audioDevice.nearestFormat(adjustedAudioFormat);
-        return true;
-    } else {
-        return false;
+    const int sampleRates[] = { 48000, 44100, 32000, 24000, 16000, 96000, 192000, 88200, 176400 };
+    const int inputChannels[] = { 1, 2, 4, 6, 8 };  // prefer mono
+    const int outputChannels[] = { 2, 4, 6, 8, 1 }; // prefer stereo, downmix as last resort
+
+    for (int channelCount : (desiredAudioFormat.channelCount() == 1 ? inputChannels : outputChannels)) {
```
ZappoMan: something about this `desiredAudioFormat.channelCount() == 1`  as a way to determine that this is an input seems janky... but I guess if that's the best we got... it seems like we could also be more explicit by calling this helper function with an `isInput` boolean. or maybe there's a way to ask the `audioDevice` if it's input or output

kencooke: It's actually best not to decide this by input/output, but by whether 1 or 2 channels is desired, since Interface (used to) allow stereo input for broadcasting music, etc.

ZappoMan: ok

ZappoMan:
Works on my Alienware!!

PR #9390, opened 2017-01-16
fix physics not starting up when a model is close behind

sethalves:
- fixed a bug that kept physics and scripts from starting when logging-in near (but not looking at) something that's physical

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -5294,15 +5294,17 @@ bool Application::nearbyEntitiesAreReadyForPhysics() {
     if (_nearbyEntitiesStabilityCount >= MINIMUM_NEARBY_ENTITIES_STABILITY_COUNT) {
         // We've seen the same number of nearby entities for several stats packets in a row.  assume we've got all
         // the local entities.
+        bool result = true;
         foreach (EntityItemPointer entity, entities) {
             if (entity->shouldBePhysical() && !entity->isReadyToComputeShape()) {
                 static QString repeatedMessage =
                     LogHandler::getInstance().addRepeatedMessageRegex("Physics disabled until entity loads: .*");
                 qCDebug(interfaceapp) << "Physics disabled until entity loads: " << entity->getID() << entity->getName();
-                return false;
+                // don't break here because we want all the relevant entities to start their downloads
+                result = false;
             }
         }
-        return true;
+        return result;
     }
     return false;
 }
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -646,6 +646,12 @@ bool RenderableModelEntityItem::isReadyToComputeShape() {
         // the model is still being downloaded.
         return false;
     } else if (type >= SHAPE_TYPE_SIMPLE_HULL && type <= SHAPE_TYPE_STATIC_MESH) {
+        if (!_model) {
+            EntityTreePointer tree = getTree();
+            if (tree) {
+                QMetaObject::invokeMethod(tree.get(), "callLoader", Qt::QueuedConnection, Q_ARG(EntityItemID, getID()));
+            }
+        }
         return (_model && _model->isLoaded());
     }
     return true;


sethalves:
to test:
- put interface in non-HMD mode
- position the avatar so that an entity with a "static-mesh" collision hull is close behind
- log out
- log in
- on master, unless you spin around to see the thing, physics will never enable.  with this PR, physics and scripts will start once the models are loaded.

Here is a place where this can be experienced:

    hifi://eschatology/90.0996,1.44495,80.1422/0,0.998945,0,-0.0459166

hifi-gustavo:
Build console output for commit: 050a7e87d24acd499601abee1b98b9220f9dd7b0 available at: https://gist.github.com/243e566575a90c83e2735d4853cae3ca<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9390-050a7e8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9390-050a7e8.dmg

nbq:
Testing this one now

nbq:
Tester: Coal
CPU: i7 6700k
GPU: GF 970

Testing interface in Desktop Mode
Went to `hifi://eschatology/90.0996,1.44495,80.1422/0,0.998945,0,-0.0459166`
logged out
logged back in
while facing forward I backed into the object behind me and it detected the collision hull because it would not let me keep backing up

Nex-Pro:
@sethalves   Is this fixing an intermittent problem? I cannot find any difference in behaviour.

I've set up a platform with that 'well' object located behind and out of view of the avatar. I created some dynamic spheres. In this and #5889 I can drag the spheres to bounce off the out-of-sight static .obj behind the avatar. Also backing up, as @nbq did, I collide with the static .obj. in both builds.  I tried restarting both clients and clearing cache.  No difference. Is this server specific ?

Nex-Pro:
@sethalves Do we also need to test an entity script in that out-of-sight object ?

sethalves:
I've just made a change.  Could I get a new review and QA pass?

Nex-Pro:
Will test when it builds.

hifi-gustavo:
Build console output for commit: bd67715cd2ff44c8f16c6d7966e37bb637559222 available at: https://gist.github.com/243e566575a90c83e2735d4853cae3ca<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9390-bd67715.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9390-bd67715.dmg

OmegaHeron:
*Wait for other testers to assign pass/fail please - I just wanted to look at this to see if it solves an issue that's annoyed me for... a long time.*

Windows 10

Master
Using test location provide, allowed scene to fully load via rotating AV 360 degrees (load indicator dismissed and AV showing physics via standing, not floating).  Exit Interface and start again from same location.  If avatar is not rotated 360 degrees it does not transition to standing. (Really need to be in 3rd person view to see how this changes things).

This PR
Scene loads physics (AV stands vs floating) without need to rotate 360 degrees on first run (cold cache) and on subsequent runs (hot cache).

Also tested with old apartment scene which exhibits problem this aims to fix.  From a cold cache it would never allow AV to transit to standing until rotated 360 degrees and *all* models loaded.  This PR allows AV to transit to standing once physics shell exists under its feet.

This also seems to solve an issue seen with loading old apartment from a cold cache where scene never fully rezzes/av left floating due to straggling models failing to ever download via ATP.  Now scene, eventually, fully loads without need to move to another domain and return or restart Interface.  👍

Basically the change this gives is; In current code your AV never transitions to standing until every model is fully loaded, if you're in desktop mode and enter a scene with cold (empty) cache you won't transit to standing until every model loadable is loaded (requiring you to rotate 360 degrees at least once) to actually load all models.  With this change out of view object no longer block load until you force them into view scope via moving AV.

Nex-Pro:
@sethalves  Just to confirm this does not fix the issue when starting interface and avatar position is inside a model such as the space cantina from the market.  The avatar still does not get physics for the cantina until they leave the  object and return after moving outside. This is a separate issue I believe.

PR #9361, opened 2017-01-11
Fix laser vs mouse again

howard-stearns:
The first commit undoes the previous fix, and should match https://github.com/highfidelity/hifi/commit/cb69e707811896852ae119b81c301c55b26995c6

The second commit is the fix.

Test plan:
- move your head around in welcome while the blue (or red laser is active). Should not get cursor. 
- stop triggering and shake the mouse to get it back. Should be able to find the mouse/cursor again.

PR #9399, opened 2017-01-16
handControllerGrab update loop change

howard-stearns:
make hand lasers smoother when heavily loaded by physics

### DIFF
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -3082,7 +3082,7 @@ var handleHandMessages = function(channel, message, sender) {
 
 Messages.messageReceived.connect(handleHandMessages);
 
-var TARGET_UPDATE_HZ = 50; // 50hz good enough (no change in logic)
+var TARGET_UPDATE_HZ = 60; // 50hz good enough, but we're using update
 var BASIC_TIMER_INTERVAL_MS = 1000 / TARGET_UPDATE_HZ; 
 var lastInterval = Date.now();
 
@@ -3095,7 +3095,7 @@ var updateTotalWork = 0;
 
 var UPDATE_PERFORMANCE_DEBUGGING = false;
     
-var updateIntervalTimer = Script.setInterval(function(){
+function updateWrapper(){
 
     intervalCount++;
     var thisInterval = Date.now();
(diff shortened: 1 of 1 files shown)


howard-stearns:
This is an "end the pain" fix, but does not address the three underlying issues:
1) While physics is settling, picking is expensive.
2) We have a lot of scripts that do picking in a polling loop. The picking is expensive, and we do the same picking multiple times in multiple default scripts.
3) We do a good job of penalizing script update signals (including those in the default scripts), but not evenly, but that effect isn't even on all kinds of polling. In particular, setTimeout/setInterval vs update signal.

This PR makes the same change that helped handControllerPointer.js recently by exploiting (3), in a way that I admit I don't understand. (cf https://github.com/highfidelity/hifi/pull/9361)

howard-stearns:
**test plan:**
- Make sure you:
 - are using the scripts from this PR;
 - have developer physics show ownership on, so that you can see when physical objects are "hot"
 - start in HMD (because switching between HMD and Desktop can leave you with odd frame rates).
- Go to welcome2 in HMD, and wiggle the base of the castle. The whole castle should light up with simulation data for a little while.
- While it's hot, pass a blue triggered laser over the objects in the castle. The laser should move smoothly.  (By contrast, in release, they will jump from one position to the next.)

- Test in other hot environments as well. 
- Try every hand-controller activity you can think of.
- Switch between developer debug defaultScripts (with reloading) so you can see if we haven't made something else be worse.

hifi-gustavo:
Build console output for commit: 0a3236dcc91d7ec0c0a392ed10e788094ed93e27 available at: https://gist.github.com/facdf7a037568bea824e426347b9fdea<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9399-0a3236d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9399-0a3236d.dmg

nbq:
testing this one now

nbq:
Tester Coal
CPU: i7 6700k
GPU: GF 970

Note: Could not use `Welcome2` as its running a different protocol version.

loaded up this build in hmd view
am using just defaultScripts.js from this build
turned on show physics ownership
went to `dev-welcome` and wiggled the base of castle to see the ownership and physics notifiers firing off
while everything was firing off I half triggered (blue lazer) over the castle and it moved smoothly
also tested by knocking over the dominos in `dev-welcome` and testing the blue lazer and making sure it moved smooth
switched between developer debug defaultScripts and reloaded default scripts to make sure nothing out of the ordinary happened
I also went over to the game area to test out some of the other functions of the hand controllers and it all seemed to work just fine without any problems.

P.S.: The unclickable dialog reared its head in this build. Upon clicking `Debug defaultScripts.js`, the box that came up notifying me I had to reload scripts for this to work was not able to be cleared using my hand controllers. I was only able to clear it by hitting enter.

PASS

PR #9430, opened 2017-01-18
Minimum edit entity filter

howard-stearns:
Highly experimental, optional entity-server filter so that arbitrary authoritative limits can be placed on entities.


hifi-gustavo:
Build console output for commit: 90c9c5b137dd1d3495534c26c68ff66252bad8c4 available at: https://gist.github.com/a75cd8902a34bc596edc88f0e4a71aa7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9430-90c9c5b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9430-90c9c5b.dmg

howard-stearns:
The following is included:
- Server settings -> Show Advanced -> Entity Server Settings -> Filter Entity Edits can be a javascript function definition:
  - It must be named 'filter'.
  - It is given an object that corresponds to the requested property edits.  It can return that object with or without modifications.
  - It can also return falsey, in which case the entire edit is rejected.
- All participants will eventually be told about the (possibly modified or rejected result.
   - If the edit was rejected or modified, it will be timestamped to just after the original edit. This is part of what helps the original editor learn about modifications and get back into sync.
   - If the edit was to a physical property, the editing Interface will have simulation ownership and will note that in the sent data. In this case, if the data is rejected or modified, the simulation owner will be set to no one.
- The entity server stats page (which can be turned on by specifying Server settings -> Show Advanced -> Entity Server Settings -> Status Hostname or Stat Port), will list the "Average Filter Time" under its "Entity Edit Statistics...".

The following are **not** included in this PR:
- The script has very limited global environment. In particular, it does not have the usual High Fidelity modules (such as Entities, Vec3, etc.).
- The script is entered as inline source rather than a reference to  a file (e.g., neither file:// nor http://).
- There is no means of combining separately written scripts. Just one big one for everything.
- There is finer-grained application of scripts (e.g., no zone or other entity). Just one big one for everything.

Currently, when a script is provided and rejects or modifies the edit, additional logging is done that is slows things down. This does not apply when there is no script, or when the script does not modify the edits, and the logging will eventually be removed.

howard-stearns:
Test plan:

Create one or more scripts for testing, perhaps along the lines of one of these:
function filter(p) { if (p.position) p.position.y = Math.min(1, p.position.y); return p; }
function filter(p) { if (p.name) {p.name += 'n'; print('fixme name', p. name);} if (p.position) {p.position.y += 1; print('fixme p.y', p.position.y);} return !p.userData && p; }

Make an entity and try edits (Entities.editEntity) that the filter will modify. Make sure it works as expected (e.g., with Entities.getEntityProperties). For frequent edits, such as through "grabbing" with lasers, mouse, or edit.js, beware of the time delay for logging.

howard-stearns:
still wrong for in some cases for physical properties....

hifi-gustavo:
Build console output for commit: 7b986dcac64f2b5aafb5a8cb1ea270ce48db64ee available at: https://gist.github.com/a75cd8902a34bc596edc88f0e4a71aa7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9430-7b986dc.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9430-7b986dc.dmg

libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -346,11 +346,15 @@ EntityPropertyFlags EntityItemProperties::getChangedProperties() const {
     return changedProperties;
 }
 
-QScriptValue EntityItemProperties::copyToScriptValue(QScriptEngine* engine, bool skipDefaults) const {
+QScriptValue EntityItemProperties::copyToScriptValue(QScriptEngine* engine, bool skipDefaults, bool allowUknownCreateTime, bool strictSemantics) const {
```
zfox23: Is "Uknown" a typo or on purpose?

libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -346,11 +346,15 @@ EntityPropertyFlags EntityItemProperties::getChangedProperties() const {
...
+    // If strictSemantics is true and skipDefaults is false, then all and only those properties are copied in the property flag
+    // is included in _desiredProperties, or is one of the specially enumerated ALWAYS properties below.
```
zfox23: This sentence reads strangely to me, and I don't understand it. I think there's some missing words in there?

Perhaps you meant "then all and only those properties *that* are copied in the property flag *are* included in _desiredProperties" ? But even that doesn't make total sense to me.

libraries/entities/src/EntityItemProperties.h
```diff
@@ -460,10 +462,6 @@ inline QDebug operator<<(QDebug debug, const EntityItemProperties& properties) {
 
     DEBUG_PROPERTY_IF_CHANGED(debug, properties, LastEditedBy, lastEditedBy, "");
 
-    properties.getAnimation().debugDump();
-    properties.getSkybox().debugDump();
-    properties.getStage().debugDump();
```
zfox23: Just curious - why did you remove these lines?

libraries/entities/src/EntityTree.cpp
```diff
@@ -918,6 +918,61 @@ void EntityTree::fixupTerseEditLogging(EntityItemProperties& properties, QList<Q
...
+    propertiesIn.setDesiredProperties(oldProperties);
+
+    auto in = QJsonValue::fromVariant(inputValues.toVariant()); // grab json copy now, because the inputValues might be side effected by the filter.
+    QScriptValueList args;
+    args << inputValues;
+
+    QScriptValue result = _entityEditFilterFunction.call(_nullObjectForFilter, args);
+
+    propertiesOut.copyFromScriptValue(result, false);
+    bool accepted = result.isObject(); // filters should return null or false to completely reject edit or add
```
zfox23: This information feels very important to add to whatever documentation we eventually write on entity filters.

libraries/entities/src/EntityItemProperties.cpp
howard-stearns: typo. will fix

libraries/entities/src/EntityItemProperties.cpp
howard-stearns: I'm trying to make no statement at all about what happens other than when strictSemantics is true and skipDefault is false. (Hence no "and only then".)  In all other cases, I couldn't figure out what the existing semantics were. skipDefault is only used by me, and it is only this case for which I know what behavior is intended.

libraries/entities/src/EntityItemProperties.h
howard-stearns: Good question. (I meant to put a CR comment, not in the code, but ran out of time and then forgot.)

The << operator sends output to a stream, accumulating results (which might then go someplace useful such as the debug stream).  Meanwhile, debugDump puts things directly into the debug stream, not whatever stream is handed to them.  So it is completely broken for operator<< to be calling debugDump, with the result that when you use the << operator, you get what you wanted PLUS the side effect of having animation, skybox, and stage sent to the debug stream.

This changes fixes the bug, but it does not attempt to re-implement << to include animation/skybox/stage. I need the basic property output to work properly (i.e., not have the bug above), and I don't need to fix animation/skybox/stage, as part of that.

libraries/entities/src/EntityTree.cpp
howard-stearns: I don't know how this is going to work as a domain-owner feature yet.

libraries/entities/src/EntityItemProperties.h
zfox23: Understood, thanks.

howard-stearns:
I'm closing this in favor of one against a branch, instead of against master.

PR #9457, opened 2017-01-20
Entity script server

Atlante45:
This PR adds a new Assignment type called the **Entity Script Server** as well as a new entity property called **serverScripts** (Only supports a single URL at the moment).

Those scripts will run on the entity server and have the same syntax as normal entity scripts.
They will have access to most of the APIs entity scripts have access minus the UI and Input related APIs.

Contrary to local entity scripts, entity server scripts will only ever have 1 instance  of themselves running at the time and it will run all the time (Even if no avatars are present in the domain).


hifi-gustavo:
Build console output for commit: 73248c356232e79feb338762e067393cd5bd6f65 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-73248c3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-73248c3.dmg

Atlante45:
test this please

hifi-gustavo:
Build console output for commit: 73248c356232e79feb338762e067393cd5bd6f65 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-73248c3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-73248c3.dmg

libraries/entities-renderer/src/EntityTreeRenderer.h
```diff
@@ -122,7 +122,7 @@ class EntityTreeRenderer : public OctreeRenderer, public EntityItemFBXService, p
 public slots:
     void addingEntity(const EntityItemID& entityID);
     void deletingEntity(const EntityItemID& entityID);
-    void entitySciptChanging(const EntityItemID& entityID, const bool reload);
+    void entityScriptChanging(const EntityItemID& entityID, const bool reload);
```
ZappoMan: :)

Atlante45: ;)

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -287,16 +288,48 @@ OctreeElement::AppendState EntityTreeElement::appendElementData(OctreePacketData
...
+                    }
+                }
 
                 if (hadElementExtraData) {
                     includeThisEntity = includeThisEntity &&
                         entityTreeElementExtraEncodeData->entities.contains(entity->getEntityItemID());
                 }
 
-                if (includeThisEntity || params.recurseEverything) {
+                if (params.usesFrustum && (includeThisEntity || params.recurseEverything)) {
```
ZappoMan: add comment about why this section only applies to `useFrustum` queries...

libraries/networking/src/AssetClient.h
```diff
@@ -22,10 +22,8 @@
 
 #include "AssetUtils.h"
 #include "LimitedNodeList.h"
-#include "NLPacket.h"
```
ZappoMan: looks like you've conflated some changes into this PR... why asset client changes? maybe make that a different PR to keep the changes clean.

libraries/networking/src/EntityScriptClient.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+GetScriptStatusRequest* EntityScriptClient::createScriptStatusRequest(QUuid entityID) {
+    auto request = new GetScriptStatusRequest(entityID);
+
+    request->moveToThread(thread());
+
+    return request;
+}
+
+bool EntityScriptClient::reloadServerScript(QUuid entityID) {
```
ZappoMan: maybe these messages should support multiple entity IDs at the same time, for future efficiency, since that's a likely desirable feature -- also maybe include a way to  "reload all"  since that's a very likely needed/requested feature. (note: could be punchlist if too hard)

libraries/networking/src/udt/PacketHeaders.cpp
```diff
@@ -48,7 +48,9 @@ PacketVersion versionForPacketType(PacketType packetType) {
         case PacketType::EntityAdd:
         case PacketType::EntityEdit:
         case PacketType::EntityData:
-            return VERSION_ENTITIES_LAST_EDITED_BY;
+            return VERSION_ENTITIES_SERVER_SCRIPTS;
+        case PacketType::EntityQuery:
+            return VERSION_ENTITIES_JSON_FILTER;
```
ZappoMan: I think you should follow the new coding standard on this message type, and it should be ~18 instead of 67 - since the query protocol hasn't changed.

libraries/octree/src/OctreeQueryNode.cpp
```diff
@@ -156,65 +155,71 @@ bool OctreeQueryNode::updateCurrentViewFrustum() {
...
-
-
-    { // if there has been a change, then recalculate
-        QMutexLocker viewLocker(&_viewMutex);
-        if (!newestViewFrustum.isVerySimilar(_currentViewFrustum)) {
-            _currentViewFrustum = newestViewFrustum;
-            _currentViewFrustum.calculate();
-            currentViewFrustumChanged = true;
+    
+    if (_usesFrustum) {
```
ZappoMan: I might be misreading this, but this looks backwards...

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1451,6 +1470,10 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
         testConstructor = sandbox.evaluate(program);
     }
     if (hadUncaughtExceptions(sandbox, program.fileName(), this)) {
+        newDetails.status = ERROR_RUNNING_SCRIPT;
+        newDetails.errorInfo = "Exception";
```
ZappoMan: are you able to include the actual exception text as well? I think that's needed for realistic debugging.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -106,6 +104,35 @@ void EntityTreeRenderer::resetEntitiesScriptEngine() {
...
+        };
+    };
+
+    auto makeSlotForSignalNoEvent = [&](QString name) -> std::function<void(const EntityItemID&)> {
+        return [newEngine, name](const EntityItemID& entityItemID) {
+            newEngine->callEntityScriptMethod(entityItemID, name);
+        };
+    };
+
+    connect(this, &EntityTreeRenderer::mousePressOnEntity, newEngine, makeSlotForSignal("mousePressOnEntity"));
```
ZappoMan: why are you making this change? I realize it sort of DRYs up the code... but is there some other specific reason?

More importantly, have you confirmed that this doesn't result in an extra delay in the calling of entity methods.

In particular, it used to be the case that if a `enterEntity` event was appropriate, we would emit the event and immediately call the entity method... presumably now you've changed it to call the entity method after the emit is actually signaled, will that not result in an extra message passing and delay in the actual calling of the method?

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -287,16 +288,48 @@ OctreeElement::AppendState EntityTreeElement::appendElementData(OctreePacketData
...
                 }
+                
+                auto entityNodeData = dynamic_cast<EntityNodeData*>(params.nodeData);
+                
+                if (entityNodeData) {
+                    // we have an EntityNodeData instance
+                    // so we should assume that means we might have JSON filters to check
+                    auto jsonFilters = entityNodeData->getJSONParameters();
+                    
+                    if (!jsonFilters.isEmpty()) {
```
birarda: 👍

libraries/entities-renderer/src/EntityTreeRenderer.h
Atlante45: Yeah but I heard that Qt is deprecating their QSciptEngine, so we swapped us to the new QScriptEngine, that I hear is much better.

libraries/entities/src/EntityTree.cpp
```diff
@@ -394,6 +394,10 @@ void EntityTree::emitEntityScriptChanging(const EntityItemID& entityItemID, cons
     emit entityScriptChanging(entityItemID, reload);
 }
 
+void EntityTree::emitEntityServerScriptChanging(const EntityItemID& entityItemID, const bool reload) {
```
Atlante45: Copied that code from `emitEntityScriptChanging` and didn't catch it.
I'll fix both.

libraries/octree/src/OctreeQueryNode.cpp
birarda: It was, fixing.

hifi-gustavo:
Build console output for commit: 2571be6edb6e531c4637ec8095042d1cb57c0a6c available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-2571be6.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-2571be6.dmg

libraries/networking/src/udt/PacketHeaders.cpp
birarda: 👍  Will fix.

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -287,16 +288,48 @@ OctreeElement::AppendState EntityTreeElement::appendElementData(OctreePacketData
                 EntityItemPointer entity = _entityItems[i];
                 bool includeThisEntity = true;
 
-                if (!params.forceSendScene && entity->getLastChangedOnServer() < params.lastViewFrustumSent) {
+                if (!params.forceSendScene && entity->getLastChangedOnServer() < params.lastQuerySent) {
                     includeThisEntity = false;
                 }
+                
+                auto entityNodeData = dynamic_cast<EntityNodeData*>(params.nodeData);
```
birarda: Will make this a static_cast

libraries/entities-renderer/src/EntityTreeRenderer.cpp
Atlante45: I did this at first because I wanted to move all the script engine related code in its own class to avoid duplication.
In the end, that proved to be too complicated to be achieved in the time frame we had, and I didn't want to just delete that code.

To answer your delay concerns, the entity script engine runs on its own thread and the signals are emitted from the main thread.
While we were calling `callEntityScriptMethod` right away, this caused `callEntityScriptMethod` to invoke itself back onto the script engine thread:
```cpp
void ScriptEngine::callEntityScriptMethod(const EntityItemID& entityID, const QString& methodName, const QStringList& params) {
        QMetaObject::invokeMethod(this, "callEntityScriptMethod",
                                  Q_ARG(const EntityItemID&, entityID),
                                  Q_ARG(const QString&, methodName),
                                  Q_ARG(const QStringList&, params));
        return;
    }
...
```

In this new code, the context object is the engine, so it will invoke `callEntityScriptMethod` on the right thread.
```cpp
connect(this, &EntityTreeRenderer::mousePressOnEntity, newEngine, makeSlotForSignal("mousePressOnEntity"));
```

Both are queued connections that switch thread. One uses `invokeMethod` and the other one a queued signal/slot connection, I don't have any data about whether one is faster than the other, but I'm pretty sure they use the same Qt event mechanism to do so.
I can gather some timing data if you want.

ZappoMan: I agree that presumably the emit process will obviate the need for the invoke... but to you're point we don't know for sure if Qt treats these the same. It sounds like this doesn't really add a new feature and is therefore just refactoring for the sake of changing. I'd suggest you revert this and add it in a different PR with a specific test plan and performance verification.

If you choose to leave it in, we need some performance numbers to show that it doesn't impact things. Note: I wouldn't be some adamant about this if it weren't for the fact that some small changes like this recently resulted in weeks of wasted effort related to problems with hand lasers. I seemingly innocuous change, which actually DID have a end user benefit (more accurate script timers) resulted in unexpected side-effects that introduced bugs.

If it ain't broke... don't fix it... unless you can prove it doesn't break other things...

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -287,16 +288,48 @@ OctreeElement::AppendState EntityTreeElement::appendElementData(OctreePacketData
...
+                    auto jsonFilters = entityNodeData->getJSONParameters();
+                    
+                    if (!jsonFilters.isEmpty()) {
+                        
+                        // if params include JSON filters, check if this entity matches
+                        bool entityMatchesFilters = entity->matchesJSONFilters(jsonFilters);
+                        
+                        if (entityMatchesFilters) {
+                            // we should include this entity unless it has already been excluded
+                            includeThisEntity = includeThisEntity && true;
```
birarda: I think you're right but I need to think about it a bit more.

hifi-gustavo:
Build console output for commit: b60ba0ff5cecd9f5d0379061a0c5a153fefd3783 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-b60ba0f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-b60ba0f.dmg

hifi-gustavo:
Build console output for commit: ab8a0a7d39c6d3d614b2624be483ba405d7bc93e available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-ab8a0a7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-ab8a0a7.dmg

libraries/entities-renderer/src/EntityTreeRenderer.cpp
Atlante45: I did a simple print test with both ways there and signals were always faster (I'm guessing they're earlier in the call queue, since they're wired from the start vs called on the spot.
But I don't have the time to do some more extensive testing today so I'll revert for now.

libraries/entities/src/EntityTreeElement.cpp
birarda: You're partially right - we might be able to avoid the filter check all together, but I need to make sure that doesn't affect the set of previously sent entities. However, in the case where nothing has changed (meaning `includeThisEntity` is false), we respect that and `includeThisEntity` would remain false.

I do need to make sure a full scene check is forced when the query has changed.

hifi-gustavo:
Build console output for commit: 95f07f3917fce73831b6ef3caa9d393220d5e483 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-95f07f3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-95f07f3.dmg

Atlante45:
test this please

hifi-gustavo:
Build console output for commit: a1ea57cf0b0a1b2e2a46714b58fb89a125a47425 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-a1ea57c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-a1ea57c.dmg

hifi-gustavo:
Build console output for commit: 95f07f3917fce73831b6ef3caa9d393220d5e483 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-95f07f3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-95f07f3.dmg

RyanDowne:
test this please

hifi-gustavo:
Build console output for commit: 95f07f3917fce73831b6ef3caa9d393220d5e483 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-95f07f3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-95f07f3.dmg

libraries/networking/src/AssetClient.h
huffman: The main change was to move `MessageID` out of AssetClient so it could be reused. There's another small change to that that will be incoming soon (currently in a PR on @Atlante45 fork)

hifi-gustavo:
Build console output for commit: 6da188b6ae248241ac48236542f522a8ffd6ca75 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-6da188b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-6da188b.dmg

birarda:
#### Test Plan

You will need to run the Sandbox and Interface from this PR because of protocol changes.

Here are two scripts that will help with testing:
https://gist.githubusercontent.com/birarda/641b466704dac409b53aa75d89935ac3/raw/78a4652caa63b8afad31e022695e35106100e50e/test-ess-1.js
https://gist.githubusercontent.com/birarda/58002c5d1e7e1cdd46bc5b4a36d93c84/raw/5ae90718edecdee40f984267575a8122b3688131/test-ess-2.js

1. Test the entity server with a normal smoke test with your interface client. You should not experience any changes to the behaviour of editing/seeing entities vs current master.
2. Ensure that your sandbox immediately has an "Entity Script Server" assigned. You should see this on your domain node listing (http://localhost:40100).
3. Check the Sandbox logs to ensure that there are no node connection issues.
4. Using edit.js, add an server script URL to an entity. Ensure that the entity script server output (in Sandbox logs) shows the output you expect. You should also see the status of the script in edit.js.
5. Reload the script using edit.js and confirm that you see that it unloads and preloads again.
6. Clear the server script URL in edit.js and confirm that you see that it unloads and stops.
7. Re-insert a server script URL in edit.js and confirm it starts running. Change the script URL to a different script and confirm that it unloads the first script and runs the second.
8. Add a server script URL in edit.js to a different entity and confirm via the entity script server logging that both scripts are running simultaneously.

NOTE: There is a "bug" in the current client entity handling that may cause you to refuse updates for an entity if you are continuously calling `editEntity` from your script update loop. This is something we will address soon but know that this may cause the entity script server to not stop a given server script when the server script URL is cleared nor changed.

nbq:
I will test this

nbq:
Tester Coal
CPU: i7 6700k
GPU: GF 970

1. tested the entity server with basic smoke test, showed no signs of any difference. Added entities to the sandbox and closed and reopened interface and they did persist.
2. Checked sandbox settings page to make sure I saw the new `entity-script-server` assigned.
3. Checked sandbox logs to verify there were no node connection issues, I did find these but I am not failing the PR because of it, I will leave it up to @birarda  ***Logs added below***
4. Added using edit.js, a server script url to an entity. Used provided one. Status changed to running.
5. Reloaded the script using edit.js to confirm it unloaded and was loaded again.
6. Cleared the server script url using edit.js and verified it did indeed unload and stop. It gave an obscure error at the bottom of the edit box once cleared that said `Could not find constructor`.
7. Using the first script provided I reinserted it into the field and then changed it to the second script provided and verified via logs that it did indeed unload the first and load the second. I did indeed see `unload` for 1 and `preload` for 2.
8. Using the first script provided I inserted it into a new entity and verified it started running and both are running together.

Overall I would say this is a **PASS** depending on what is viewed from the logs provided.

```
[01/21 16:06:51] [DEBUG] [hifi.networking] [3960] [entity-script-server] LimitedNodeList::sendPacket called without active socket for node "Agent" (I) {617b7df1-f404-4f14-bf8e-0fdf8e99af97} 71.76.205.19:36682 / 172.16.1.108:61588 - not sending
[01/21 16:06:52] [DEBUG] [hifi.networking] [3960] [entity-script-server] No response to UDP hole punch pings for node QUuid("{617b7df1-f404-4f14-bf8e-0fdf8e99af97}") in last second.
[01/21 16:06:52] [DEBUG] [hifi.networking] [3960] [entity-script-server] LimitedNodeList::sendPacket called without active socket for node "Agent" (I) {617b7df1-f404-4f14-bf8e-0fdf8e99af97} 71.76.205.19:36682 / 172.16.1.108:61588 - not sending
[01/21 16:06:53] [DEBUG] [hifi.networking] [3960] [entity-script-server] LimitedNodeList::sendPacket called without active socket for node "Agent" (I) {617b7df1-f404-4f14-bf8e-0fdf8e99af97} 71.76.205.19:36682 / 172.16.1.108:61588 - not sending
```

cozza13:
@omega to test on i5

OmegaHeron:
Still working on performance measurements with i5 CPU.  But.

PR needs some work and is a fail for me due to AC log being constantly filled with this repeating series of messages;

`[01/23 11:30:06] [DEBUG] [hifi.networking] [11676] [entity-script-server] Activating local socket for network peer with ID "6d2707cb-7cfc-4a10-b80b-8abdc23cbb8d"`
`[01/23 11:30:07] [DEBUG] [hifi.networking] [11676] [entity-script-server] LimitedNodeList::sendPacket called without active socket for node "Agent" (I) {c0b7f136-f8b8-40f6-827e-9a22ec5bf0a5} 70.127.20.117:55727 / 192.168.0.6:55727 - not sending`
`[01/23 11:30:07] [DEBUG] [hifi.networking] [11676] [entity-script-server] No response to UDP hole punch pings for node QUuid("{c0b7f136-f8b8-40f6-827e-9a22ec5bf0a5}") in last second.`
`[01/23 11:30:08] [DEBUG] [hifi.networking] [11676] [entity-script-server] LimitedNodeList::sendPacket called without active socket for node "Agent" (I) {c0b7f136-f8b8-40f6-827e-9a22ec5bf0a5} 70.127.20.117:55727 / 192.168.0.6:55727 - not sending`
`[01/23 11:30:08] [DEBUG] [hifi.networking] [11676] [entity-script-server] No response to UDP hole punch pings for node QUuid("{c0b7f136-f8b8-40f6-827e-9a22ec5bf0a5}") in last second.`
`[01/23 11:30:09] [DEBUG] [hifi.networking] [11676] [entity-script-server] LimitedNodeList::sendPacket called without active socket for node "Agent" (I) {c0b7f136-f8b8-40f6-827e-9a22ec5bf0a5} 70.127.20.117:55727 / 192.168.0.6:55727 - not sending`
`[01/23 11:30:09] [DEBUG] [hifi.networking] [11676] [entity-script-server] No response to UDP hole punch pings for node QUuid("{c0b7f136-f8b8-40f6-827e-9a22ec5bf0a5}") in last second.`
`[01/23 11:30:10] [DEBUG] [hifi.networking] [11676] [entity-script-server] LimitedNodeList::sendPacket called without active socket for node "Agent" (I) {c0b7f136-f8b8-40f6-827e-9a22ec5bf0a5} 70.127.20.117:55727 / 192.168.0.6:55727 - not sending`
`[01/23 11:30:11] [DEBUG] [hifi.networking] [11676] [entity-script-server] LimitedNodeList::sendPacket called without active socket for node "Agent" (I) {c0b7f136-f8b8-40f6-827e-9a22ec5bf0a5} 70.127.20.117:55727 / 192.168.0.6:55727 - not sending`
`[01/23 11:30:11] [DEBUG] [hifi.networking] [11676] [entity-script-server] No response to UDP hole punch pings for node QUuid("{c0b7f136-f8b8-40f6-827e-9a22ec5bf0a5}") in last second.`

Over multiple runs/restarts of stack those messages never stop.

OmegaHeron:
Another issue is this;

Add an entity
Assign a server script URL and note that it indicates running script in edit.js properties
Change server script URL to blank
Close edit tool
Open edit tool and select entity that had script assigned
Note that it now indicates an error each time object is selected

When removing a server script from an entity it seems it is not being "cleaned" of its former state of having an assigned script.  Put another way - an entity that has had a script assigned no longer behaves like one that never had a script once you remove its script value.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -1063,7 +1061,7 @@ void EntityTreeRenderer::entityCollisionWithEntity(const EntityItemID& idA, cons
         }
     }
 
-    if (isCollisionOwner(myNodeID, entityTree, idA, collision)) {
```
Atlante45: @AndrewMeadows @ZappoMan When I found this, it looked like a bug to me (Probably a copy paste error given the above block)
Can you confirm whether it is?

OmegaHeron:
In so far as performance is concerned for i5 running sandbox + Interface on same machine I'm not seeing any performance hit of note.  Obviously - if you load up a ton of heavy entity scripts it's going to eat CPU time and, possibly, lead to a condition where an i5 machine would not perform well.  But - that could be said of any CPU configuration... Someday the discussion of CPU limits for such things needs to come into scope (todo for The Glorious Future).

But - I'm concerned by 2 things;

1) Issue with an entity being assigned a script then having script removed (via setting its script value to none).  It seems server is still spending time evaluating an error condition for the empty script value as opposed to it not doing so for an entity that never had a script assigned.

Could this result in increased load over time if one assigned/removed scripts from a lot of entities (load in both CPU and memory)?

2) Indications of node networking failure messages.

hifi-gustavo:
Build console output for commit: c8e27ec31b593a278cbefe9775c1dba8d7122420 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-c8e27ec.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-c8e27ec.dmg

hifi-gustavo:
Build console output for commit: 9282f5555a3065aee5a949cd3409a607f0417f32 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-9282f55.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-9282f55.dmg

hifi-gustavo:
Build console output for commit: bd340f99cecd16be353b0030e96466b1b9827641 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-bd340f9.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-bd340f9.dmg

libraries/entities-renderer/src/EntityTreeRenderer.cpp
ZappoMan: I agree that looks like a typo

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -302,22 +302,23 @@ OctreeElement::AppendState EntityTreeElement::appendElementData(OctreePacketData
...
+                // if this entity has been updated since our last full send and there are json filters, check them
+                if (includeThisEntity && !jsonFilters.isEmpty()) {
 
                     // if params include JSON filters, check if this entity matches
                     bool entityMatchesFilters = entity->matchesJSONFilters(jsonFilters);
 
                     if (entityMatchesFilters) {
                         // we should include this entity unless it has already been excluded
-                        includeThisEntity = includeThisEntity && true;
+                        includeThisEntity = true;
```
ZappoMan: why set this to `true` isn't it already `true` or else it wouldn't be in this branch?

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -302,22 +302,23 @@ OctreeElement::AppendState EntityTreeElement::appendElementData(OctreePacketData
...
 
                         // make sure this entity is in the set of entities sent last frame
                         entityNodeData->insertEntitySentLastFrame(entity->getID());
 
                     } else {
                         // we might include this entity if it matched in the previous frame
                         if (entityNodeData->sentEntityLastFrame(entity->getID())) {
-                            includeThisEntity = includeThisEntity && true;
+                            includeThisEntity = true;
```
ZappoMan: why set this to `true` isn't it already `true` or else it wouldn't be in this branch?

libraries/entities/src/EntityTreeElement.cpp
birarda: good point

libraries/entities/src/EntityTreeElement.cpp
birarda: good point

hifi-gustavo:
Build console output for commit: 4ec5d70e7643c5ffaef34629c0ed1d1b4fe94ae0 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-4ec5d70.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-4ec5d70.dmg

birarda:
This is ready for another QA pass - we addressed the extremely helpful comments you added @OmegaHeron. Note that the test scripts have been updated in the test plan.

hifi-gustavo:
Build console output for commit: 5cd0af9136682e3816e9d18c81abb98f02fcc660 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-5cd0af9.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-5cd0af9.dmg

OmegaHeron:
@birarda - on the second test script,  what should `this.time` refer to?  It's currently giving undefined... if `this.time` is supposed to return a value as a property of script object then we have an issue.

`print("Update 2:", this.entityID, this.time);` -> `[01/25 12:01:51] [DEBUG] [hifi.scriptengine.script] [15572] [entity-script-server] script:print()<< Update 2: {fd521073-5edd-4b6d-96c2-0aeddbb77229} undefined`

OmegaHeron:
The following log spam needs to be addressed.  Seen anytime edit.js is running and an object is selected (regardless of it having, not having or have had/not had) a script assigned.  The debug output is constant (in Interface and is eating CPU time...).

`[01/25 12:20:43] [DEBUG] [hifi.scriptengine.script] script:print()<< Requesting status of script
[01/25 12:20:44] [DEBUG] [hifi.scriptengine.script] script:print()<< Got script status: true false error_loading_script`

Repro;  Enable edit, select an entity or make a new one, watch Interface log.  Anytime any object is selected = infinite logging of script status check.

huffman:
@OmegaHeron It looks like `this.time` is a remnant from a previous script, that shouldn't be a concern.

re: log spam, I will address that right now, thanks for pointing that out!

huffman:
@OmegaHeron The next build should no longer have the log spam.

OmegaHeron:
Thanks @huffman

hifi-gustavo:
Build console output for commit: f165e364e2185c444c22799c4cf7251b64dc651d available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-f165e36.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-f165e36.dmg

OmegaHeron:
QA Fail - reproduction as follows;

Create a text entity
Set its script (I used an ATP referenced script) to the following;

`var i = 0;`
`var testProps = {};`
`(function() {`
`    return {`
`        preload: function(entityID) {`
`            print("Preload 2: ", entityID);`
`            this.entityID = entityID;`
`            Script.update.connect(this, this.update);`
`        },`
`        unload: function() {`
`            print("Unload 2: ", this.entityID);`
`            Script.update.disconnect(this, this.update);`
`        },`
`        update: function(dt) {`
`	   testProps = { text: i}`
`	   Entities.editEntity(this.entityID, testProps);`
`	    i++;`
`        }`
`    }`
`});`

Assign that to text entity and close edit - you should see its display changing with ticks...

Now - go back to edit, select text entity and clear its script reference.  Note that text stops updating as expected.  Quit Interface and launch again, text entity will be showing ticks changing!  Looking at it in edit mode its script source is empty and shows not running.. yet the text entity is happily chugging away with an increase count display.....  restarting stack clears it, but, I was able to repeat this 100%.

Atlante45:
test this please

birarda:
@OmegaHeron I have a feeling you've run into this (my note from the test plan).

NOTE: There is a "bug" in the current client entity handling that may cause you to refuse updates for an entity if you are continuously calling editEntity from your script update loop. This is something we will address soon but know that this may cause the entity script server to not stop a given server script when the server script URL is cleared nor changed.

Not clear to us yet why the Interface would have stopped receiving the ticks until restarted. Additionally we MAY be able to add something to this PR that makes the server script URL change more likely to succeed.

OmegaHeron:
@birarda Yes!  That's likely it (facepalms) - was so deep into profiling CPU cycles and other stuff that I kind of forgot that caveat in test notes.

birarda:
@OmegaHeron can you re-perform the same test and see if you see the unload print from your script in the entity script server log? That would help a bunch.

OmegaHeron:
Doing it now

hifi-gustavo:
Build console output for commit: f165e364e2185c444c22799c4cf7251b64dc651d available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-f165e36.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-f165e36.dmg

OmegaHeron:
@birarda 

Created text entity and assigned it atp:/ess6.js getting;

`[01/25 13:54:02] [DEBUG] [default] [16112] [entity-script-server] Loading entity server script "atp:/ess6.js" for [entity-id: "{a6266463-4ac3-4869-9cbe-e29644d79d50}" ]`
`[01/25 13:54:02] [DEBUG] [hifi.networking] [16112] [entity-script-server] Got mapping for: "/ess6.js" => "590b60b9ec5c183694d996ab18a60a457f96279ed67771ff808ec8f285584522"`
`[01/25 13:54:02] [DEBUG] [hifi.networking.asset_client] [16112] [entity-script-server] "atp:590b60b9ec5c183694d996ab18a60a457f96279ed67771ff808ec8f285584522" loaded from disk cache.`
`[01/25 13:54:02] [DEBUG] [hifi.scriptengine] [16112] [entity-script-server] Done downloading script at: "atp:/ess6.js"`
`[01/25 13:54:02] [DEBUG] [hifi.scriptengine.script] [16112] [entity-script-server] script:print()<< Preload 2: {a6266463-4ac3-4869-9cbe-e29644d79d50}`

Noted text tick display is working.

Removed script reference via edit.js getting no indication of unloading (but visually text entity stopped updating).  So it looks like some weird in between states here... it doesn't fire script unload yet it stops counting but then is right back when interface is restarted.

Now for worse...

After logging back in to find entity still updating I deleted it.  That triggered 2 log notices

1) Script fired its unload event
2) Lots of repeating edit fail messages

`[01/25 13:59:43] [DEBUG] [hifi.entities] [12400] [entity-server] Edit failed. [ 45 (EntityEdit) ] entity id: [entity-id: {a6266463-4ac3-4869-9cbe-e29644d79d50} ] existingEntity pointer: 0x0`
`[01/25 13:59:43] [DEBUG] [hifi.scriptengine.script] [16112] [entity-script-server] script:print()<< Unload 2: {a6266463-4ac3-4869-9cbe-e29644d79d50}`

`[01/25 13:59:46] [SUPPRESS] [] [12400] [entity-server] 76 repeated log entries matching "^Edit failed.*" - Last entry: "Edit failed. [ 45 (EntityEdit) ] entity id: [entity-id: {a6266463-4ac3-4869-9cbe-e29644d79d50} ] existingEntity pointer: 0x0"`
`[01/25 13:59:46] [DEBUG] [hifi.entities] [12400] [entity-server] Edit failed. [ 45 (EntityEdit) ] entity id: [entity-id: {a6266463-4ac3-4869-9cbe-e29644d79d50} ] existingEntity pointer: 0x0`
`[01/25 13:59:51] [SUPPRESS] [] [12400] [entity-server] 148 repeated log entries matching "^Edit failed.*" - Last entry: "Edit failed. [ 45 (EntityEdit) ] entity id: [entity-id: {a6266463-4ac3-4869-9cbe-e29644d79d50} ] existingEntity pointer: 0x0"`
`[01/25 13:59:51] [DEBUG] [hifi.entities] [12400] [entity-server] Edit failed. [ 45 (EntityEdit) ] entity id: [entity-id: {a6266463-4ac3-4869-9cbe-e29644d79d50} ] existingEntity pointer: 0x0`

OmegaHeron:
It's pretty clear this is exactly the pre-existing bug you warned of @birarda 

It seems what happens is this;
I'm calling an edit model in loop
Due to bug - it's not actually fully registering that I removed script reference thus it locally stops updating the counter - but, really, hasn't stopped at all - in fact if you delete the script and see counter stop displaying then wait some random "long enough" time - it will begin showing updates again without restarting interface.  It shows a substantial jump from its count when stopped to value when it begins again showing that script never stopped (as expected since unload never fires).

Problem with all this is... until the bug is fix leading to this we're in dangerous territory rolling this new script server out.  Through, seemingly, no fault of its own, it's... unpredictable in its function.

birarda:
Agree with you 100% @OmegaHeron. We're discussing a potential server script URL specific interim fix for the general bug now.

Can you do one more test for us? As soon as you clear the server script URL can you move your view frustum and see if that kicks an update sooner that you receive it if your view is not moving?

OmegaHeron:
@birarda - It doesn't seem to but, oddly enough... taking snapshots seems to lead to an update faster.  While snapping my series to show it's ticking still - I noticed after a few that it would stop ticking.  Not sure why that might be and possibly just a gods of random at play.

huffman:
test this please

OmegaHeron:
While work is in progress on the bug we've been discussing - the good news.

Looking at CPU use shows no red flags (even on my i5 running interface/stack in parallel).  I created what amounts to several infinite loop entities running the second example script and watched entity-script-server CPU use via its process ID.  It seems its maximal execution time is limited compared to local (in interface) scripts.  This is reflected via both its lesser CPU use and by watching my counter display script running in local vs server mode.  Local mode loops radically faster and shows a matching much higher CPU burn.  Hopefully that's a feature, !bug of server entity script engine.  ;)

hifi-gustavo:
Build console output for commit: f165e364e2185c444c22799c4cf7251b64dc651d available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-f165e36.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-f165e36.dmg

marko8904:
test this please

OmegaHeron:
Next thing I'm noticing that's troublesome...

Running following script against a text entity I see a slow but steady climb in memory utilization by entity script server process.  It began reasonably low but over ~30 minutes or so run time has increased approximately 5x.  It began showing around 12,000K and was up to near 59,000K when I restarted stack.  It's a rather strange ongoing memory consumption as well... it varies between fairly rapidly allocating small increases to periods where it hardly changes at all.

`var i = 0;`
`var testProps = {};`
`(function() {`
`    return {`
`        preload: function(entityID) {`
`            print("Preload 2: ", entityID);`
`            this.entityID = entityID;`
`            Script.update.connect(this, this.update);`
`        },`
`        unload: function() {`
`            print("Unload 2: ", this.entityID);`
`            Script.update.disconnect(this, this.update);`
`        },`
`        update: function(dt) {`
`	   testProps = { text: i}`
`	   Entities.editEntity(this.entityID, testProps);`
`	    i++;`
`        }`
`    }`
`});`

hifi-gustavo:
Build console output for commit: f165e364e2185c444c22799c4cf7251b64dc651d available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-f165e36.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-f165e36.dmg

birarda:
@OmegaHeron fix on the way for the issue noted in test plan.

We're checking if the memory issues are present in Agent AC running the same script.

hifi-gustavo:
Build console output for commit: 9048db356faf0d584fb130c9e0e0ff2a8e14d022 available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-9048db3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-9048db3.dmg

birarda:
@OmegaHeron confirming that there is a bug that also exists on master (tested with an AC script) where you will stop receiving updates being done by an external script for an entity after moving it. We're filing that bug but that explains the issue you saw where the text entity stopped updating until you restarted interface.

birarda:
@OmegaHeron confirming that the leak you see in ESS happens in assignment-client Agent. We will file separately.

birarda:
@OmegaHeron merge incoming that will fix the thing we noted and you saw. If you could test that script changes made by an interface client are as expected handled by an entity script server that is continuously updating other properties of the entity, that should be the last thing we need to check.

OmegaHeron:
@birarda - will do and, yes, it all looks good otherwise.

birarda:
test this please

hifi-gustavo:
Build console output for commit: 3f34a9e83212c22bf64fe5dfef9e83af3b0f33ab available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-3f34a9e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-3f34a9e.dmg

hifi-gustavo:
Build console output for commit: f165e364e2185c444c22799c4cf7251b64dc651d available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-f165e36.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-f165e36.dmg

hifi-gustavo:
Build console output for commit: 3f34a9e83212c22bf64fe5dfef9e83af3b0f33ab available at: https://gist.github.com/953539252d867c32c4e6d0f3ea0edec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-3f34a9e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9457-3f34a9e.dmg

OmegaHeron:
Testing Windows 10

Latest fix is, so far, reliable *if* you set a script and then simply set it blank.  That case seems to be, with limited tests (10 or so tries) solid.  There's still issue with having set a script then replacing it - I mean, set a script value, see that it's running then change its value... things get a little less reliable in that one.  (This again applies to scripts triggering bugs not really owned by this PR).

I'm reasonably content saying QA Pass now.  Upcoming work should take care of outstanding issues.

Questions for The Glorious future;

Server scripts obey whitelist rules?  (Thinking ahead where I might want to control entity scripts of all kinds while allowing others edit rights).

How does all this play with having *both* a server script and a classic client script assigned to same object?  (I tried it and am still not sure how to be sure they didn't clobber one another other than programmer's job to see it doesn't...).

birarda:
@OmegaHeron you can have both a server script and client script and both will run independently. 

Will be opening up a separate PR so that entity server scripts respect the whitelist.

PR #9475, opened 2017-01-23
Fix invalid volume values in injector options silencing all injectors

Atlante45:
This fixes a bug where, if you pass an invalid value for the audio option volume, the latter would be set to `nan` and messup all the other injectors.

## Test plan:

- Go to a domain with injectors (Your sandbox should work)
- Create a cube entity
- Save this script to a local file and attach it to the cube entity:
```js
(function() {
  this.preload = function() {
    this.sound = SoundCache.getSound("https://s3-us-west-1.amazonaws.com/hifi-content/DomainContent/Welcome+Area/Sounds/low_pulse_1.L.wav");
};

  this.clickDownOnEntity = function(entityID, mouseEvent) {
    Audio.playSound(this.sound, {
      position: MyAvatar.position,
      volume: "bad value"
    });
  };
});
```
- Click the entity to play the sound
- This should quiet all the injectors on the current master, but play at the default volume in this PR.
- I also added some debug in the log so people know what's going on.

Fixes: https://highfidelity.fogbugz.com/f/cases/1746/

### DIFF
--- a/libraries/audio/src/AudioInjectorOptions.cpp
+++ b/libraries/audio/src/AudioInjectorOptions.cpp
@@ -9,9 +9,13 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 
+#include "AudioInjectorOptions.h"
+
+#include <QScriptValueIterator>
+
 #include <RegisteredMetaTypes.h>
 
-#include "AudioInjectorOptions.h"
+#include "AudioLogging.h"
 
 AudioInjectorOptions::AudioInjectorOptions() :
     position(0.0f, 0.0f, 0.0f),
@@ -22,7 +26,7 @@ AudioInjectorOptions::AudioInjectorOptions() :
     ambisonic(false),
     ignorePenumbra(false),
     localOnly(false),
-    secondOffset(0.0)
+    secondOffset(0.0f)
 {
 
 }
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 3f8b2f9cbb8726f3e269ed6af2abdbfdcb3d2ef7 available at: https://gist.github.com/7c007880c0e1ea0888fa4a6401fb1330<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9475-3f8b2f9.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9475-3f8b2f9.dmg

OmegaHeron:
Testing Windows 10
QA Pass

Tests against master (5924) exhibit blanking of sounds in Sandbox when entity clicked.

Tests against this PR - sound plays, other injected audio plays and voice on server echoback is fine as well.

Log message seen (as expected);
`[01/23 15:15:13] [WARNING] [hifi.audio] Audio injector options: volume is not a number`
`[01/23 15:15:13] [DEBUG] [hifi.audioclient] adding new injector`
`[01/23 15:15:15] [DEBUG] [hifi.audioclient] injector has no more data, marking finished for removal`
`[01/23 15:15:15] [DEBUG] [hifi.audioclient] removing injector`

You might consider changing the error message to something like Invalid volume value. Range of x ... y acceptable.  That would be clearer what the problem/solution is.  Just a thought.

libraries/audio/src/AudioInjectorOptions.cpp
```diff
@@ -40,31 +44,51 @@ QScriptValue injectorOptionsToScriptValue(QScriptEngine* engine, const AudioInje
...
-    }
-    
-    if (object.property("secondOffset").isValid()) {
-        injectorOptions.secondOffset = object.property("secondOffset").toNumber();
+
+    QScriptValueIterator it(object);
+    while (it.hasNext()) {
+        it.next();
+
+        if (it.name() == "position") {
```
zfox23: Any reason not to use switch statements here? It'd probably be faster, right?

Atlante45: I wish. But you can't switch on a string ;)

zfox23: Ooh, nice! I learned something today :)

Atlante45: Regarding whether it's faster, I would expect a good compiler to detect those cases and optimize it like would a switch statement.
But switch statements are still preferable whenever possible since they are much more readable.
Also Compilers will warn you when you forget cases (for enums only).

PR #9481, opened 2017-01-23
Update the name of the collision shapes used in edit.js properties

huffman:
Update the names to be consistent with what is shown on the add model
dialog.

### DIFF
--- a/scripts/system/html/css/edit-style.css
+++ b/scripts/system/html/css/edit-style.css
@@ -593,7 +593,7 @@ hr {
 .dropdown dl {
     font-family: FiraSans-SemiBold;
     font-size: 15px;
-    width: 172px;
+    width: 292px;
     height: 28px;
     padding: 0 28px 0 12px;
     color: #afafaf;
@@ -645,7 +645,7 @@ hr {
 .dropdown li {
     list-style-type: none;
     padding: 3px 0 1px 12px;
-    width: 200px;
+    width: 320px;
     height: auto;
     font-family: FiraSans-SemiBold;
     font-size: 15px;
--- a/scripts/system/html/entityProperties.html
+++ b/scripts/system/html/entityProperties.html
@@ -328,13 +328,13 @@
             <div class="model-group model-section zone-section property dropdown">
                 <label>Collision shape type</label>
                 <select name="SelectShapeType" id="property-shape-type">
-                    <option value="none">None</option>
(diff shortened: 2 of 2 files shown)


huffman:
The names of the collision shapes shown in edit.js have been updated to be consistent with the names that are used in the add model dialog.

The new names are:
```
None
No Collision
Box
Sphere
Compound
One Hull
Hull Per Submesh
Static Mesh
Basic - Whole model
Good - Sub-meshes
Exact - All polygons (non-dynamic only)
```

To test:

1. Ensure that the names have been updated, show up correctly, and can be selected.

nbq:
Will test this one

RyanDowne:
We should also check the if an entity had previously been set to dynamic and you change the hull to be exact, the dynamic checkbox gets unchecked.

nbq:
@RyanDowne I will test that also then in the build

hifi-gustavo:
Build console output for commit: f6d5a59c8ae2b4a82259a846d7b15f3c537be53b available at: https://gist.github.com/ef480560108a83cde60c5d43cebb5e8d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9481-f6d5a59.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9481-f6d5a59.dmg

nbq:
Tester: Coal
CPU: i7 6700k
GPU: GF 970
OS: Windows 10


* No Collision **(checked, worked)**
* Box **(checked, worked)**
* Sphere **(checked, worked)**
* Compound **(checked, worked, changed to `Good - Sub-meshes`)**
* Basic - Whole model **(checked, worked)**
* Good - Sub-meshes **(checked, worked)**
* Exact - All polygons (non-dynamic only) **(checked, worked)**


To test a collision hull change I brought in an EZ-Tube and set collision to be no collision and then I used edit menu to change it to exact and it worked just fine. I was able to walk through it with none and once set to exact it used the correct hull and I tested because I could not walk through it head on but I could fall through the middle so it acted as a tube.

When on `Exact` I could not check dynamic, also if I had it as dynamic on another setting, it unchecked when I set it to `Exact`, this test was for @RyanDowne 

I can say this is a QA PASS because everything was selectable and showed up appropriate on the draw collision shapes also

PR #9496, opened 2017-01-24
ENTITY FILTERS: "Can't change model" & "Can't exceed velocity X"

zfox23:
In this PR, I've included V1 of two entity filters.
1. A filter that prevents users from changing any entity properties under the "MODEL" header of the EDIT menu.
2. A filter that prevents any component of an entity's velocity from exceeding 5 units/second.

**To test these filters' efficacy:**
*Filter 1*
1. Enter your sandbox.
2. Spawn a beach ball from the Marketplace.
3. Use the domain settings page to apply Filter 1 to your sandbox, then apply the settings and restart the sandbox.
4. In your sandbox, open the Edit menu, then select the beach ball, then click the "Entity Properties" tab.
5. Attempt to change any field underneath the "MODEL" header.
For example, you can try changing the "Model URL" to `http://mpassets.highfidelity.com/6af97d4e-1338-4b4f-97bd-4d80449d5542-v1/CircusHammer.fbx`, which is the circus hammer model. Verify that changing any property under the MODEL header results in no change to the model appearance or behavior.
Note that some entity server rejections may apply but not be immediately visible in the "Entity Properties" edit window.

*Filter 2*
1. Enter your sandbox.
2. Spawn a beach ball from the Marketplace.
3. Use the domain settings page to apply Filter 2 to your sandbox, then apply the settings and restart the sandbox.
4. In your sandbox, grab the beach ball and attempt to throw it as hard as you can. Don't break your monitor.
5. Verify that the beach ball didn't go as fast as you wish it did.
You may notice the beach ball snap back a little as it flies through the air - this is normal, as it's the entity server rejecting your hard through. You're a regular Tom Brady.
Also note that the filter will apply even if you're far-grabbing an object - which means pulling the beach ball to your body quickly will result in the beach ball jumping around a little bit.

### DIFF
--- a/script-archive/entity-server-filter-example.js
+++ b/script-archive/entity-server-filter-example.js
@@ -16,16 +16,16 @@ function filter(p) {
 	/* Clamp velocity to maxVelocity units/second. Zeroing each component of acceleration keeps us from slamming.*/
 	var maxVelocity = 5;
     if (p.velocity) {
-		if (p.velocity.x > maxVelocity) {
-			p.velocity.x = maxVelocity;
+		if (Math.abs(p.velocity.x) > maxVelocity) {
+			p.velocity.x = Math.sign(p.velocity.x) * maxVelocity;
 			p.acceleration.x = 0;
 		}
-		if (p.velocity.y > maxVelocity) {
-			p.velocity.y = maxVelocity;
+		if (Math.abs(p.velocity.y) > maxVelocity) {
+			p.velocity.y = Math.sign(p.velocity.y) * maxVelocity;
 			p.acceleration.y = 0;
 		}
-		if (p.velocity.z > maxVelocity) {
-			p.velocity.z = maxVelocity;
+		if (Math.abs(p.velocity.z) > maxVelocity) {
+			p.velocity.z = Math.sign(p.velocity.z) * maxVelocity;
 			p.acceleration.z = 0;
 		}
     }
--- a/script-archive/entity-server-filter-example.js
+++ b/script-archive/entity-server-filter-example.js
@@ -9,6 +9,25 @@ function filter(p) {
 
     /* Can also reject altogether */
     if (p.userData) { return false; }
+	
+    /* Reject if modifications made to Model properties */
+    if (p.modelURL || p.compoundShapeURL || p.shape || p.shapeType || p.url || p.fps || p.currentFrame || p.running || p.loop || p.firstFrame || p.lastFrame || p.hold || p.textures || p.xTextureURL || p.yTextureURL || p.zTextureURL) { return false; }
+	
+	/* Clamp velocity to 10 units/second. Zeroing each component of acceleration keeps us from slamming.*/
+    if (p.velocity) {
+		if (p.velocity.x > 10) {
+			p.velocity.x = 10;
+			p.acceleration.x = 0;
+		}
+		if (p.velocity.y > 10) {
+			p.velocity.y = 10;
+			p.acceleration.y = 0;
+		}
+		if (p.velocity.z > 10) {
+			p.velocity.z = 10;
+			p.acceleration.z = 0;
+		}
+    }
 
     /* If we make it this far, return the (possibly modified) properties. */
     return p;
--- a/script-archive/entity-server-filter-example.js
+++ b/script-archive/entity-server-filter-example.js
@@ -13,18 +13,18 @@ function filter(p) {
     /* Reject if modifications made to Model properties */
     if (p.modelURL || p.compoundShapeURL || p.shape || p.shapeType || p.url || p.fps || p.currentFrame || p.running || p.loop || p.firstFrame || p.lastFrame || p.hold || p.textures || p.xTextureURL || p.yTextureURL || p.zTextureURL) { return false; }
 	
-	/* Clamp velocity to 10 units/second. Zeroing each component of acceleration keeps us from slamming.*/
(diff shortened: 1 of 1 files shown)


script-archive/entity-server-filter-example.js
```diff
@@ -9,6 +9,25 @@ function filter(p) {
 
     /* Can also reject altogether */
     if (p.userData) { return false; }
+	
+    /* Reject if modifications made to Model properties */
+    if (p.modelURL || p.compoundShapeURL || p.shape || p.shapeType || p.url || p.fps || p.currentFrame || p.running || p.loop || p.firstFrame || p.lastFrame || p.hold || p.textures || p.xTextureURL || p.yTextureURL || p.zTextureURL) { return false; }
+	
+	/* Clamp velocity to 5 units/second. Zeroing each component of acceleration keeps us from slamming.*/
+    if (p.velocity) {
+		if (p.velocity.x > 5) {
```
howard-stearns: We should probably put the 5 in a var, Maybe between lines 16 and 17?

hifi-gustavo:
Build console output for commit: 0e16686fb8ab0b5dab36e02c8b0749dacaa0236f available at: https://gist.github.com/f7923e51364ea7fdea800a26024f46b8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9496-0e16686.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9496-0e16686.dmg

script-archive/entity-server-filter-example.js
zfox23: Good point, makes sense.

hifi-gustavo:
Build console output for commit: 53173d8f037a71cf33898ede0b25869263b3045b available at: https://gist.github.com/f7923e51364ea7fdea800a26024f46b8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9496-53173d8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9496-53173d8.dmg

davidkelly:
(1) QA Pass works as expected
(2) Saw interesting things.  While throwing the ball around, it became a zombie ball.  Stayed in one place, couldn't far grab it, hit it with my hand, or anything.  That said, it definitely worked as advertised wrt the velocity and acceleration.  I don't know if this is an issue with the ball (like, that script had an issue), or with what we've done in this branch, or (doubtful) this filter.  A second one eventually decided to disappear underneath the floor of sandbox.   I didn't see these issues when I removed the filters, but then again they maybe don't reproduce easily?  And could be unrelated.  QA Pass?  Maybe another look?  I'll look at the logs and see if they tell a story.

davidkelly:
Also FWIW, I think we need to cap the velocity in both positive and negative directions, no?  This only checks for a velocity greater than 5 - rather than abs(p.velocity.x) > 5

script-archive/entity-server-filter-example.js
```diff
@@ -9,6 +9,26 @@ function filter(p) {
...
     /* Can also reject altogether */
     if (p.userData) { return false; }
+	
+    /* Reject if modifications made to Model properties */
+    if (p.modelURL || p.compoundShapeURL || p.shape || p.shapeType || p.url || p.fps || p.currentFrame || p.running || p.loop || p.firstFrame || p.lastFrame || p.hold || p.textures || p.xTextureURL || p.yTextureURL || p.zTextureURL) { return false; }
+	
+	/* Clamp velocity to maxVelocity units/second. Zeroing each component of acceleration keeps us from slamming.*/
+	var maxVelocity = 5;
+    if (p.velocity) {
+		if (p.velocity.x > maxVelocity) {
```
davidkelly: I think we want to test the abs of the velocity wrt each component.  I think this means you can have -100 for the velocity in the x (or y, or z) without problems.

davidkelly: BTW if you just programatically set the velocity to +10 in the x, and see how far it goes, and compare to -10, the difference is about what you'd expect if you were only really allowing it to be +5 vs -10.

zfox23: Oh, great idea! I didn't even consider negative velocities. Thanks.

zfox23:
@davidkelly Regarding your QA notes, did you happen to have a second observer client in your domain to see what the "zombie ball" looked like for others? It's possible that the ball appeared more "normal" to the observer, which would both make sense and be in-line with the spirit of these filters (i.e. it's OK for an entity to look a little strange to a client trying to break the rules; it's not OK for the entity to look strange to others).

davidkelly:
Oh that ball was permanently in that state. I didn't have a second observer in there, but I could try that when/if I reproduce it.

zfox23:
Interesting - I never encountered that in my own testing.

davidkelly:
K I think it could be just completely unrelated.  I'll re-test with new stuff shortly.

davidkelly:
well it is merged - so I'm sure it is fine.

hifi-gustavo:
Build console output for commit: 9796a476eacabd4775de99796acd327335811ef0 available at: https://gist.github.com/f7923e51364ea7fdea800a26024f46b8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9496-9796a47.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9496-9796a47.dmg

PR #9502, opened 2017-01-25
Throttle audio with PI controller, by node

zzmp:
- Updates timing statistics for audio-mixer
- Updates sleep mechanism for audio-mixer (verified by new statistics)
- Updates throttle controller to use time-to-work instead of time-to-sleep
- Updates throttle controller (`AudioMixer::manageLoad` is now `AudioMixer::throttle`) to a proportional-integral controller (was exponential) with annotated constants
- Updates throttle mechanism to throttle by a fraction of each listening node's streams (was done by global volume)

### DIFF
--- a/assignment-client/src/audio/AudioMixerSlave.cpp
+++ b/assignment-client/src/audio/AudioMixerSlave.cpp
@@ -199,121 +199,87 @@ void AudioMixerSlave::mix(const SharedNodePointer& node) {
     }
 }
 
-AudioStreamMap splitQuietestStreams(const AvatarAudioStream& listener, AudioStreamMap& streams, int numToRetain) {
-    using AudioStreamKey = AudioStreamMap::key_type;
-
-    std::vector<std::pair<float, AudioStreamKey>> streamVolumes;
-    // satisfy the push_heap precondition that streamVolumes be non-empty
-    streamVolumes.push_back(std::make_pair(0.0f, QUuid()));
-
-    // put streams into a heap by their volume
-    for (const auto& streamPair : streams) {
-        const auto& streamKey = streamPair.first;
-        const auto& stream = streamPair.second;
-
-        // calculate the volume
-        float distance = glm::length(stream->getPosition() - listener.getPosition());
-        float volume = stream->getLastPopOutputTrailingLoudness() / distance;
-
-        streamVolumes.push_back(std::make_pair(volume, streamKey));
-        std::push_heap(streamVolumes.begin(), streamVolumes.end());
-    }
-
-    // pop the loudest numToRetain streams off the heap
-    AudioStreamMap poppedStreams;
-    for (int i = 0; i < numToRetain; i++) {
-        std::pop_heap(streamVolumes.begin(), streamVolumes.end());
-
-        AudioStreamKey streamKey = streamVolumes.back().second;
-        poppedStreams[streamKey] = streams[streamKey];
-        streams.erase(streamKey);
-
-        streamVolumes.pop_back();
-    }
-    streams.swap(poppedStreams);
-    return poppedStreams;
-}
-
-bool AudioMixerSlave::prepareMix(const SharedNodePointer& node) {
-    AvatarAudioStream* nodeAudioStream = static_cast<AudioMixerClientData*>(node->getLinkedData())->getAvatarAudioStream();
-    AudioMixerClientData* nodeData = static_cast<AudioMixerClientData*>(node->getLinkedData());
+bool AudioMixerSlave::prepareMix(const SharedNodePointer& listener) {
+    AvatarAudioStream* listenerAudioStream = static_cast<AudioMixerClientData*>(listener->getLinkedData())->getAvatarAudioStream();
+    AudioMixerClientData* listenerData = static_cast<AudioMixerClientData*>(listener->getLinkedData());
 
     // zero out the client mix for this node
     memset(_mixSamples, 0, sizeof(_mixSamples));
 
-    // loop through all other nodes that have sufficient audio to mix
-    std::for_each(_begin, _end, [&](const SharedNodePointer& otherNode){
-        // make sure that we have audio data for this other node
-        // and that it isn't being ignored by our listening node
-        // and that it isn't ignoring our listening node
-        AudioMixerClientData* otherData = static_cast<AudioMixerClientData*>(otherNode->getLinkedData());
+    if (_throttlingRatio == 0.0f) {
+        std::for_each(_begin, _end, [&](const SharedNodePointer& node) {
+            AudioMixerClientData* nodeData = static_cast<AudioMixerClientData*>(node->getLinkedData());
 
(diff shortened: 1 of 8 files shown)


zzmp:
I tested this on a 16-CPU EC2 server.

I was able to use 80 avatars with no throttling (running [this script](https://gist.githubusercontent.com/cozza13/e8d997f2e2444885cda9774a2d888c11/raw/5441ed0cc808e0ce580d98e14a43cafd8f9eadc3/people.js)). ***The avatars must be run from a different machine as agents, or they will hog the CPU time that the `audio-mixer` needs to mix them.*** This is analogous to a real-life situation - 80 users will not be colocated on one machine.

When I ramped up to 100, there was stuttering for about 4 seconds until the throttling was enough, and then it was clear. When I turned my agents off and on again, the throttling remained, and I was able to clearly hear with 100 avatars present (note that I should not have been able to hear them all).

---

A lot of this was based on ideas thrown out by @ZappoMan in [*audio mixer appears to be too strict and sticky*](https://highfidelity.fogbugz.com/f/cases/2047/audio-mixer-fallback-appears-to-be-too-strict-and-sticky).

Code reviewers and QAers:
- Is there anything obvious that I missed (any big ideas I left out, that *should* be in there)?
- What do you think of throttling by node instead of by global volume? This means that when there is throttling, *everyone* will have the streams quietest to them filtered out. *Everyone* will have the same number of streams filtered out.
- If you try it, do you think the throttling is still sticky?
  - It still can take up to 3 minutes to reset, but the throttling is not as aggressive, so it will usually take half that or less.

hifi-gustavo:
Build console output for commit: 4ac0df8f8d7cfe9043a08c2b6aba858ddd277dea available at: https://gist.github.com/a12c5a97edf265fc0ac05318867f0bc3<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9502-4ac0df8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9502-4ac0df8.dmg

assignment-client/src/audio/AudioMixerSlave.cpp
```diff
@@ -196,75 +199,87 @@ void AudioMixerSlave::mix(const SharedNodePointer& node) {
...
-            } else if ((node->isIgnoreRadiusEnabled() || otherNode->isIgnoreRadiusEnabled())) {
-                // Define the minimum bubble size
-                static const glm::vec3 minBubbleSize = glm::vec3(0.3f, 1.3f, 0.3f);
-                AudioMixerClientData* nodeData = reinterpret_cast<AudioMixerClientData*>(node->getLinkedData());
-                // Set up the bounding box for the current node
-                AABox nodeBox(nodeData->getAvatarBoundingBoxCorner(), nodeData->getAvatarBoundingBoxScale());
-                // Clamp the size of the bounding box to a minimum scale
-                if (glm::any(glm::lessThan(nodeData->getAvatarBoundingBoxScale(), minBubbleSize))) {
-                    nodeBox.setScaleStayCentered(minBubbleSize);
+    if (_throttlingRatio == 0.0f) {
```
ZappoMan: maybe this should be on the inside of the `std::for_each(_begin, _end, [&](const SharedNodePointer& node) {` loop... it looks like the code would be DRYer in that case... thoughts?

ZappoMan:
```
AudioMixer.cpp:39, Clang (LLVM based), Priority: Normal
unused variable 'LOUDNESS_TO_DISTANCE_RATIO'
```

zzmp:
Thanks for the Jenkins copypasta.

zzmp:
275e337 - 46afa7e only DRY up the code. They should have no functional effects.

fb31254 adds a proportional term to the audio-mixer. This means if it is struggling really badly, it will throttle faster (it is no longer always linear). Conversely, if it is throttled badly and suddenly goes down to a small number of avatars to mix, it will recover much faster (30 seconds from full throttle to full recovery, instead of 3 minutes as it was).

hifi-gustavo:
Build console output for commit: fb31254f57366806ecbd453205b4ba4779519386 available at: https://gist.github.com/a12c5a97edf265fc0ac05318867f0bc3<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9502-fb31254.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9502-fb31254.dmg

hifi-gustavo:
Build console output for commit: 8ac9c6ba36fd215d3b95cafededc6cf8319d22cc available at: https://gist.github.com/a12c5a97edf265fc0ac05318867f0bc3<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9502-8ac9c6b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9502-8ac9c6b.dmg

zzmp:
### QA Plan

This changes low-level audio-mixer code, so it needs to be tested thoroughly.

@cozza13 should we coordinate who will do which test? I have done bot testing, but am not sure if that is sufficient / will help anyone else who wants to learn how.

You should do two types of load testing, and then a general test:
- Bot testing
  - Boot up a server
  - Bring it to load with agents running on a separate machine (if you use AWS EC2, you can use the table below to determine load)
  - Bring it past load and make sure it throttles appropriately
    - If it is just over load it should throttle slowly (you can see it in the logs)
    - If it is far over load it should throttle quickly
  - Bring it below load and make sure it recovers appropriately
    - If it is just below load it should recover slowly, &c.
- Live crowd testing
  - Same as bot testing, but with real people (@cozza13 is this feasible?)
- General test
  - Just stick around the server and make sure all audio sounds right

Appx load for AWS EC2 machines:

| Machine | # Avatars Approaching Load |
| - | - |
| m4.2xlarge | 45 |
| m4.4xlarge | 75 |
| m4.10xlarge | 120 |

cozza13:
cozza13 will test this

zzmp:
This does *not* use actual volume for throttling. It uses an approximation - `the volume of the stream / the distance from the stream`. It discards angle, domain and zone information, and gain adjustments per avatar. That means the `PAL: Per-Avatar Volume Slider` will not affect which streams are filtered when throttling.

This is in keeping with the old throttling behavior, so I think it is fine. These are things that can go in through a separate PR, which should consider:
- if the approximation is good (considering angle, domain, zone, adjustments)
- if the actual value is expensive (this can be profiled) to make sure throttling stays much cheaper than mixing

cozza13:
### QA Plan

- Bot testing **DONE**
- Bring it below load and make sure it recovers appropriately **DONE**
- If it is just below load it should recover slowly, &c. **STILL NEEDS TESTING**
- Live crowd testing **THIS MAY NEED TO HAPPEN ON FRIDAY**
- General test **DONE - WORKS**

PR #9518, opened 2017-01-26
Entity filters by URL

howard-stearns:
Entity server filters are now specified in settings as a url. When a value is specified, the contents must be successfully retrieved by the entity server before it is allowed to process edits.

Any syntax or runtime error is logged when the filter is read, which stops the server. Any runtime error is logged when the filter is run, which causes those edits to be rejected.

Note: The url contents are cached. If you want to force new contents, use a unique query parameter, such as http://localhost:800/my-filter.js?v=17


howard-stearns:
**test plan**
- Please note the "note" regarding content caching, above.
- Specify no filter url. The entity server should work normally.
- Specify a url, and the filter will be retrieved and used. File urls are rejected (with a log message), and the server will stop. (The monitor may cause it to repeatedly restart. BTW, the reason file urls are rejected is that the assignment client could be on a different machine. ATP urls are not supported at this time.)
- A bad url will be logged and the server not run. (In all these case, the monitor may cause it to repeatedly restart.)
- A filter with a syntax error or runtime error will be logged and the server not run.
- A filter that gets a runtime error when it turn will be logged and the edit rejected.

hifi-gustavo:
Build console output for commit: f36b968860c3ab89b7f49a85f8cca410b9157bf4 available at: https://gist.github.com/c9085c337c0b91c4bdad72daea5486df<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9518-f36b968.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9518-f36b968.dmg

davidkelly:
Test Results:
(1) no filter url, things work normally
(2) with a valid url and filter, things work as expected
(3)  file url (file://whatever.js) and entity server logs and stops, then restarts and does same.
(4) a valid url, but one that 404s, and we get the same as (3)
(5) syntax error in script, entity server logs it (a lot), but runs seemingly normally (of course, without a filter)
(6) a runtime error (calling non existent function on an object) in a filter does reject the edit, and log

A syntax error was expected to make the server stop (and then restart, in a loop), so lets look at that.

assignment-client/src/entities/EntityServer.cpp
```diff
@@ -286,11 +293,93 @@ void EntityServer::readAdditionalConfiguration(const QJsonObject& settingsSectio
...
+void EntityServer::scriptRequestFinished() {
+    auto scriptRequest = qobject_cast<ResourceRequest*>(sender());
+    const QString urlString = scriptRequest->getUrl().toString();
+    if (scriptRequest && scriptRequest->getResult() == ResourceRequest::Success) {
+        auto scriptContents = scriptRequest->getData();
+        qInfo() << "Downloaded script:" << scriptContents;
+        QScriptProgram program(scriptContents, urlString);
+        if (hasCorrectSyntax(program)) {
+            _entityEditFilterEngine.evaluate(scriptContents);
+            if (!hadUncaughtExceptions(_entityEditFilterEngine, urlString)) {
```
davidkelly: we need to call stop() if this is true.  But, we fall through and hit line 372 where we return without calling stop.

howard-stearns: quite right. A little bit more is needed, to. Fixed.

hifi-gustavo:
Build console output for commit: f9f0a4ea2d282023c3d25b25712be1d945b918fa available at: https://gist.github.com/c9085c337c0b91c4bdad72daea5486df<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9518-f9f0a4e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9518-f9f0a4e.dmg

PR #9530, opened 2017-01-27
compute Model's local bounds outside pending changes queue

AndrewMeadows:


AndrewMeadows:
This is PR is cleanup toward removing the `Model` backpointer from its render payload.  All we're doing here is moving some local bounds calculations out of a method `ModelMeshPartPayload::updateTransformForSkinnedMesh()` that is only invoked in a lambda function right before the render payload `ModelMeshPartPayload` is prepared for transaction to the render pipeline.  This makes for less work in the `render::Scene::processPendingChanges()` context and instead we only recompute the local bounds when the Model's joints have actually changed.

hifi-gustavo:
Build console output for commit: 489fbc629932d9f192158067411e57cd1591274b available at: https://gist.github.com/018ef8237137d990b5d155f922421d57<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9530-489fbc6.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9530-489fbc6.dmg

AndrewMeadows:
To test:

(1) Login to dev-chris.highfidelity.io and verify that the performance has not changed noticeably from whatever is the latest hifi/master.

(2) Login to a domain with plenty of models.  Move the camera view around and verify that object bounding boxes are correct -- the objects don't flicker out of view as the camera view is moved.

OmegaHeron:
Tested Windows 10 - i5 6500 CPU GTX970 GPU
QA Pass

Compared to master, no change to FPS/CPU load (well, minor improvement to both FPS and CPU but so minor to be within normal run to run variances).

General smoke test pass.

PR #9537, opened 2017-01-27
Add injector/avatar attenuation to audio-mixer throttling

zzmp:
Throttle sorting is done based on volume and distance.
It does not take into account injector gain or PAL adjustment gain (they are not misrepresented - they simply are *not* represented).

This adds injector gain and PAL adjustment gain into volume sorting (in throttling logic) before the quietest streams are filtered.


zzmp:
@kencooke this touches code that may affect audio-mixer performance - could you evaluate that?

hifi-gustavo:
Build console output for commit: e7e550d78c92b7324e046eb6d0fd79ed60f36c5c available at: https://gist.github.com/e0b129beec7effaa8beb3bbbaef47735<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9537-e7e550d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9537-e7e550d.dmg

assignment-client/src/audio/AudioMixerSlave.cpp
```diff
@@ -498,36 +498,30 @@ float computeGain(const AvatarAudioStream& listeningNodeStream, const Positional
...
         float angleOfDelivery = glm::angle(glm::vec3(0.0f, 0.0f, -1.0f),
                                            glm::normalize(rotatedListenerPosition));
 
         const float MAX_OFF_AXIS_ATTENUATION = 0.2f;
-        const float OFF_AXIS_ATTENUATION_FORMULA_STEP = (1 - MAX_OFF_AXIS_ATTENUATION) / 2.0f;
-
+        const float OFF_AXIS_ATTENUATION_STEP = (1 - MAX_OFF_AXIS_ATTENUATION) / 2.0f;
         float offAxisCoefficient = MAX_OFF_AXIS_ATTENUATION +
-        (OFF_AXIS_ATTENUATION_FORMULA_STEP * (angleOfDelivery / PI_OVER_TWO));
+            (OFF_AXIS_ATTENUATION_STEP * (angleOfDelivery / PI_OVER_TWO));
```
kencooke: Wouldn't hurt to code these divide-by-constants as multiplies:
```(OFF_AXIS_ATTENUATION_STEP * angleOfDelivery * (1.0f/PI_OVER_TWO)); ```
The compiler will only do this when inexact float optimizations are enabled.

File: assignment-client/src/audio/AudioMixerSlave.cpp [hunk 2]
```diff
@@ -162,6 +162,10 @@ bool AudioMixerSlave::prepareMix(const SharedNodePointer& listener) {
                     glm::vec3 relativePosition = nodeStream->getPosition() - listenerAudioStream->getPosition();
                     float gain = approximateGain(*listenerAudioStream, *nodeStream, relativePosition);
 
+                    // modify by hrtf gain adjustment
+                    auto& hrtf = listenerData->hrtfForStream(nodeID, nodeStream->getStreamIdentifier());
```
kencooke: The hash to do the map lookup here is kind of expensive. Since most people leave PAL volume at 1.0, it may not be worth it.

assignment-client/src/audio/AudioMixerSlave.cpp
```diff
@@ -501,10 +516,9 @@ float computeGain(const AvatarAudioStream& listeningNodeStream, const Positional
...
     if (streamToAdd.getType() == PositionalAudioStream::Injector) {
         gain *= reinterpret_cast<const InjectedAudioStream*>(&streamToAdd)->getAttenuationRatio();
-    }
 
-    // avatar: apply fixed off-axis attentuation to make them quieter as they turn away
-    if (!isEcho && (streamToAdd.getType() == PositionalAudioStream::Microphone)) {
+    // avatar: apply fixed off-axis attenuation to make them quieter as they turn away
+    } else if (!isEcho && (streamToAdd.getType() == PositionalAudioStream::Microphone)) {
         glm::vec3 rotatedListenerPosition = glm::inverse(streamToAdd.getOrientation()) * relativePosition;
         float angleOfDelivery = glm::angle(glm::vec3(0.0f, 0.0f, -1.0f),
```
kencooke: This could be optimized, not sure it matters

assignment-client/src/audio/AudioMixerSlave.cpp
zfox23: Agree with @kencooke on this one - I'd say that _most_ people leave _most_ of their volume sliders at default.

kencooke:
Profiling a throttling workload would help guide all this. Another opportunity for optimization is to compute some gain factors less often, since its overkill for some (like PAL volumes) to be computed 100 times/sec.

zzmp:
@kencooke the PAL volumes are pulled from the HRTF functors of that node's streams. How would you compute them less often? It would need to be cached in the stream somehow, but is there a way to do that that would be more obviously performant than just grabbing them from the HRTF functor?

zzmp:
As far as profiling, I tried throttling with the avatar turning attenuation, and the throttle amount went up by a lot. With this setup, the same throttling covers the same amount of avatars (so there is some experimentation to support that this is not a noticeably larger load for the throttling gain approximation). After I know if anything else will be changed (see above comments) I can run hotspots profiling as well, for finer profiles.

kencooke:
You're right, PAL is efficient since moving it into the HRTF. I was thinking of the other logic that mutes avatars (privacy bubbles?). Anyway my point was that the gain computation done every 10ms should be as lean as possible. Updating some computations less often might work, but would need to keep state in the HRTF node which is kind of ugly.

hifi-gustavo:
Build console output for commit: 2b50d9e8de2e0cc9d270f65a3e39d7df05508905 available at: https://gist.github.com/e0b129beec7effaa8beb3bbbaef47735<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9537-2b50d9e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9537-2b50d9e.dmg

hifi-gustavo:
Build console output for commit: eb00c7ba1115a39d65b48778c59240539f76ca02 available at: https://gist.github.com/e0b129beec7effaa8beb3bbbaef47735<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9537-eb00c7b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9537-eb00c7b.dmg

zzmp:
QA:
Run a bot test (see #9502).
Ensure that the audio-mixer performs well, and throttles well, on WIN32 and linux.
It should perform comparably to `master`, in that it should begin throttling at the same number of agents, and throttle about the same (i.e. if `master` throttles 80 avatars with a throttling ratio of 0.5, this PR should throttle to the same ratio for 80 avatars).

zzmp:
Pulling in master, as AVX2 was added to HRTF, so master will now run about 2x faster.

hifi-gustavo:
Build console output for commit: 1f56814714091715322404aaedf9b5a2d6507205 available at: https://gist.github.com/e0b129beec7effaa8beb3bbbaef47735<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9537-1f56814.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9537-1f56814.dmg

OmegaHeron:
General smoke test passes on Windows 10.

Unable to really do anything beyond a 2 or 3 AV test at this time.  Should likely get some more organized > load testing.

zzmp:
Tested on EC2.

#### master
110 users caused no throttling.
115 users caused throttling to 0.5.

#### #9537
105 users caused no throttling.
110 users caused throttling to 0.25.
115 users caused throttling to 0.425.

This causes a small degradation in performance, in exchange for incorporating PAL and injector gain into the throttle filter algorithm. I think it is worthwhile.

kencooke:
The PAL-aware gains appear to cause 5% hit (110->105) even when throttling is not occurring, is that expected? I hate to see it that high since anything in the mixing core is O(N^2) and becomes a problem again as we scale further.

ZappoMan:
The 5% hit doesn't really make much sense... I don't see any changes to the core logic that would account for a 5% hit in performance. I see only a couple small changes that might account for a performance it at all...

- The old code calculated the `distanceBetween` once and used it in two places. The new code recalculate the the `distance` using the `glm::length()` twice... that's clearly going to cost something... 5% seems high for that - but we could definitely reuse that calculated value instead of calculating twice.

- The old code directly calculated the gain in the `prepareMix()` function by just dividing loudness by distance, where the new code calls `approximateGain()` which basically does the same thing in the case of mic/avatar audio, but includes an if statement+multiply for injectors. This essentially is fixing a bug in how we considered injectors, but since the ratio of injectors to avatar audio should be low, this shouldn't impact performance in the general case... So the only question here is the function call introducing enough of a cache consistency hit to impact performance, and is there a way to shuffle this code around to prevent that impact.

- the old code manually did a clamp, the new code uses glm::clamp() - I guess this could be slower... 

- the new code introduces an extra multiply for all gain estimates to include the gain from the source/listener HRTF... since this is only impacting the gain estimate for sorting which streams to throttle, then it would seem like we could drop this multiply without a significant change in behavior.

kencooke:
The 5% makes more sense when you consider:
 - this is getting computed 100x100x100 = one million times every second.
 - the problem with injector and PAL gains is not the 2 multiplies, its fetching the 2 values. Each involves std::unordered_map lookup (key hash, cache misses?) that are protected by mutex (contention?).

ZappoMan:
@kencooke - I don't see any additional calls to `std::unordered_map` in this PR... can you highlight the line in this PR where that happens? I don't think that explains this slowdown.

Also - the fact that an extra multiply is happening 100x100x100 times does not explain a *percentage* change... if a single multiply was 5% of the per frame work load -- then sure... it would... but that doesn't seem like what's actually happening here.

kencooke:
@ZappoMan fetching the PAL gain here:
```
// modify by hrtf gain adjustment
auto& hrtf = listenerData->hrtfForStream(nodeID, nodeStream->getStreamIdentifier());
gain *= hrtf.getGainAdjustment();
```

does a lookup into _nodeSourcesHRTFMap here:
```
AudioHRTF& hrtfForStream(const QUuid& nodeID, const QUuid& streamID = QUuid()) { return _nodeSourcesHRTFMap[nodeID][streamID]; }
```

which looks to be a map-of-maps:
```
    using HRTFMap = std::unordered_map<QUuid, AudioHRTF>;
    using NodeSourcesHRTFMap = std::unordered_map<QUuid, HRTFMap>;
    NodeSourcesHRTFMap _nodeSourcesHRTFMap;
```

so each one results in hashes on QUuids and another std::_Hash to map them to buckets. That could be optimized by using std::pair<uuid, uuid> as key to a 1D map.

You're right, I don't see any new locks being acquired. I should have said "if any are protected by a mutex..."

assignment-client/src/audio/AudioMixerSlave.cpp
```diff
@@ -475,40 +488,54 @@ bool shouldIgnoreNode(const SharedNodePointer& listener, const SharedNodePointer
...
+        gain *= reinterpret_cast<const InjectedAudioStream*>(&streamToAdd)->getAttenuationRatio();
     }
 
+    // avatar: skip attenuation - it is too costly to approximate
+
+    // distance attenuation: approximate, ignore zone-specific attenuations
+    // this is a good approximation for streams further than ATTENUATION_START_DISTANCE
+    // those streams closer will be amplified; amplifying close streams is acceptable
+    // when throttling, as close streams are expected to be heard by a user
+    float distance = glm::length(relativePosition);
```
zzmp: This is done twice, and can be computed and then passed by ref instead to decrease computation.

zzmp:
Removed `Development Approved` pending:
- changes suggested by @ZappoMan 
- targeted tests (timers or VTune reports) showing exactly where the slowdown occurs

zzmp:
In response to the changes requested by @ZappoMan:
- `distanceBetween`: the old code calculated and used this in `gainForSource`. This has been renamed `approximateGain`, and is still only calculated once within the function. It is calculated *separately* for `approximateGain`, as this is done in a separate loop over nodes & streams. In order to only calculate once, it would need to be cached in the streams, which would add more complexity.
- I will run/report tests shortly to show the impact of `approximateGain`'s branching to get gain for specific injector types.
- `glm::clamp` is documented as `min(max(x, minVal), maxVal)`, which is what the code was. I will see if anything shows up through perf reports, though.
- The extra multiply for gain estimates, looked-up from source/listener HRTFs, is necessary as it implements PAL gain adjustments into throttle filtering.

Running perf tests now - will post shortly with results on why (which codepath) this PR slows audio mixing.

hifi-gustavo:
Build console output for commit: 9b0190f1cf829ad54f4216f8acd0b7bb24acbb89 available at: https://gist.github.com/e0b129beec7effaa8beb3bbbaef47735<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9537-9b0190f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9537-9b0190f.dmg

zzmp:
Rereading this, I see that the issue is adding 5% even when not throttling. This is likely just from rough testing - I didn't respond to it directly because I didn't understand the exact issue before, sorry about that.
Throttling occurs in both master and PR around 110-115. The test puts avatars in random positions, so there is some range.

---

I've added 9b0190f to this PR so that I can determine the effect on throttle time, in this and future PRs. To use it, `make assignment-client CXX_DEFINES='-DHIFI_AUDIO_THROTTLE_DEBUG'`.

I tested both master and PR on an m4.4xlarge with 115 avatars (which caused throttling), and noted the appx. amount of time spent per stream in gain approximation for throttling (*this only occurs during throttling*):
- master: ~1030 ns / stream
- PR: ~ 1250 ns / stream

This has increased throttling time per stream by 21.5%.
The point at which we are completely throttled should be at `N = sqrt(t_frame * no_threads / t_throttle_stream)` avatars (assuming only single-avatar nodes) (this is not the actual case). This reduces our max theoretical avatars from 394 to 354. It also gives us a basis from which to improve.

It should be allowed (despite the 21.5% increase in time) because it fixes bugs in throttling behavior, and only affects mixers with over 100 streams.

Upcoming PRs will further improve this time with memoization, cachine, lockless programming, &c.

zzmp:
@cozza13 I added QA Ready, but there was no significant code change. This was approved before, and can just be approved again once it builds.

@kencooke could you re Dev Approve this?

kencooke:
This seems completely reasonable.
As we continue to improve scaling, we can hopefully do both
- increase max avatars we can mix without throttling
- reduce overhead of throttling, to increase total avatars we can handle without server failure

PR #9548, opened 2017-01-30
fetch entity filter script asynchronously (but reject edits while waiting)

howard-stearns:
This is an attempt to workaround the problem with filters on Linux.

### DIFF
--- a/assignment-client/src/entities/EntityServer.cpp
+++ b/assignment-client/src/entities/EntityServer.cpp
@@ -294,7 +294,9 @@ void EntityServer::readAdditionalConfiguration(const QJsonObject& settingsSectio
     }
 
     if (readOptionString("entityEditFilter", settingsSectionObject, _entityEditFilter) && !_entityEditFilter.isEmpty()) {
-        // Fetch script from file synchronously. We don't want the server processing edits while a restarting entity server is fetching from a DOS'd source.
+        // Tell the tree that we have a filter, so that it doesn't accept edits until we have a filter function set up.
+        std::static_pointer_cast<EntityTree>(_tree)->setHasEntityFilter(true);
+        // Now fetch script from file asynchronously.
         QUrl scriptURL(_entityEditFilter);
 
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: 93414d802d740cfaa99d974f548924b5199ff381 available at: https://gist.github.com/2a6afdf7a7c43c161ac143ffc811661c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9548-93414d8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9548-93414d8.dmg

howard-stearns:
**test plan**

- Please note that filter file contents are cached, so use ?v=uniqueValue as needed.
- Specify no filter url. The entity server should work normally.
- Specify a url, and the filter will be retrieved and used. File urls are rejected (with a log message), and the server will stop. (The monitor may cause it to repeatedly restart. BTW, the reason file urls are rejected is that the assignment client could be on a different machine. ATP urls are not supported at this time.)
- A bad url will be logged and the server not run. (In all these case, the monitor may cause it to repeatedly restart.)
- A filter with a syntax error or runtime error will be logged and the server not run.
- A filter that gets a runtime error when it turn will be logged and the edit rejected.

davidkelly:
QA results on linux:
(1) with no filter url, everything is fine
(2) with url, filter is retrieved.  
(3) file url rejected with log, and entity server stops then is restarted, stops, etc...  
  (a) if the url is bad in the sense that we 404, we do the same
(4) syntax error causes the entity server die, but it gets restarted, the script is downloaded, and the entity server stops.  So it loops, but I think that is ok 
(5) couldn't test easily on linux.

One note -- for syntax errors, file:// and 404 urls, the restarts happen often.  I worry about this generating huge logs, which eventually kill servers and so on.  In a perfect world, we'd not restart the entity server at all, after logging.

howard-stearns:
We'll leave the (exceptional case) logging in for now, but we should cut it down later.
Calling it a qa pass as well.

PR #9551, opened 2017-01-30
Improved legibility of tablet in desktop mode.

hyperlogic:
In desktop mode we set the "isAA" property on the Web3DOverly to false, which causes it to be rendered after the FXAA pass.  It has the side effect of also making it render on top of everything else, because at that point the zbuffer has been cleared.


hifi-gustavo:
Build console output for commit: 98b53d1d29770087999cc0c1fa3d656adf5044c8 available at: https://gist.github.com/7ad0e0c1b942f799a8a5001eba4f3c93<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9551-98b53d1.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9551-98b53d1.dmg

PR #9556, opened 2017-01-30
Adding support for disabling texturing from the material

samcake:
a new toggle carried by the LightingModel object to eventually disable the usage of the material textures  and instead fall back to the default simple textures.

If disabled before any texture is used for rendering in a frame then this also a way to test the cost of the loading time of the textures by avoiding it and compare the time saved vs the normal case.

## Test Plan
By default there is zero change compared to the current master
Runnning the script debugDeferredLighting.js  (from this pr) you should see a new checkbox "Textures" in the first column at the bottom. When disabling it all the material textures should disappear.
When starting from an empty scene and disableing the Textures feature then going to a texture rich environment (such as playa for exemple) you should expect a much faster loading of the scene, no stuttering at all and controlling the gpu memory consumption in the stats window, no texture allocated.

### DIFF
--- a/libraries/render-utils/src/MeshPartPayload.cpp
+++ b/libraries/render-utils/src/MeshPartPayload.cpp
@@ -129,7 +129,7 @@ void MeshPartPayload::bindMesh(gpu::Batch& batch) const {
     }
 }
 
-void MeshPartPayload::bindMaterial(gpu::Batch& batch, const ShapePipeline::LocationsPointer locations) const {
+void MeshPartPayload::bindMaterial(gpu::Batch& batch, const ShapePipeline::LocationsPointer locations, bool enableTextures) const {
     if (!_drawMaterial) {
         return;
     }
@@ -148,7 +148,7 @@ void MeshPartPayload::bindMaterial(gpu::Batch& batch, const ShapePipeline::Locat
     }
 
     // Albedo
-    if (materialKey.isAlbedoMap()) {
+    if (enableTextures && materialKey.isAlbedoMap()) {
         auto itr = textureMaps.find(model::MaterialKey::ALBEDO_MAP);
         if (itr != textureMaps.end() && itr->second->isDefined()) {
             batch.setResourceTexture(ShapePipeline::Slot::ALBEDO, itr->second->getTextureView());
@@ -160,7 +160,7 @@ void MeshPartPayload::bindMaterial(gpu::Batch& batch, const ShapePipeline::Locat
     }
 
     // Roughness map
-    if (materialKey.isRoughnessMap()) {
+    if (enableTextures && materialKey.isRoughnessMap()) {
         auto itr = textureMaps.find(model::MaterialKey::ROUGHNESS_MAP);
         if (itr != textureMaps.end() && itr->second->isDefined()) {
             batch.setResourceTexture(ShapePipeline::Slot::MAP::ROUGHNESS, itr->second->getTextureView());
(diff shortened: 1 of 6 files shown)


libraries/render-utils/src/MeshPartPayload.cpp
```diff
@@ -230,15 +230,15 @@ void MeshPartPayload::bindMaterial(gpu::Batch& batch, const ShapePipeline::Locat
...
+    if (enableTextures && materialKey.isLightmapMap()) {
         auto itr = textureMaps.find(model::MaterialKey::LIGHTMAP_MAP);
 
         if (itr != textureMaps.end() && itr->second->isDefined()) {
             batch.setResourceTexture(ShapePipeline::Slot::MAP::EMISSIVE_LIGHTMAP, itr->second->getTextureView());
         } else {
             batch.setResourceTexture(ShapePipeline::Slot::MAP::EMISSIVE_LIGHTMAP, textureCache->getGrayTexture());
         }
-    } else if (materialKey.isEmissiveMap()) {
+    } else if (enableTextures && materialKey.isEmissiveMap()) {
```
AndrewMeadows: Not a blocker but: there is much checking of `enableTextures`, which is `true` the vast majority of the time.  Maybe provide an early exit?

```
if (!enableTextures) {
    batch.setResourceTexture(ShapePipeline::Slot::ALBEDO, textureCache->getWhiteTexture());
    batch.setResourceTexture(ShapePipeline::Slot::MAP::ROUGHNESS, textureCache->getWhiteTexture());
    batch.setResourceTexture(ShapePipeline::Slot::MAP::NORMAL, nullptr);
    batch.setResourceTexture(ShapePipeline::Slot::MAP::METALLIC, nullptr);
    batch.setResourceTexture(ShapePipeline::Slot::MAP::OCCLUSION, nullptr);
    batch.setResourceTexture(ShapePipeline::Slot::MAP::SCATTERING, nullptr);
    return;
}
```
Or maybe gpu::Batch has (or should have) a method that does all of that in one call?  `Batch::clearResourceTextures()` or something?

hifi-gustavo:
Build console output for commit: 3c56bd29600eb6066f367eb631dd986149d5d04b available at: https://gist.github.com/4c09a4616b8135432382d58a26e87e87<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9556-3c56bd2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9556-3c56bd2.dmg

libraries/render-utils/src/MeshPartPayload.cpp
samcake: good point

samcake: just updated that

hifi-gustavo:
Build console output for commit: 249ec80d74ab57c48f1bb063e95b1118f2fd5065 available at: https://gist.github.com/4c09a4616b8135432382d58a26e87e87<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9556-249ec80.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9556-249ec80.dmg

libraries/render-utils/src/MeshPartPayload.cpp
```diff
@@ -147,8 +147,19 @@ void MeshPartPayload::bindMaterial(gpu::Batch& batch, const ShapePipeline::Locat
         numUnlit++;
     }
 
+    if (!enableTextures) {
+        batch.setResourceTexture(ShapePipeline::Slot::ALBEDO, textureCache->getWhiteTexture());
+        batch.setResourceTexture(ShapePipeline::Slot::MAP::ROUGHNESS, textureCache->getWhiteTexture());
+        batch.setResourceTexture(ShapePipeline::Slot::MAP::NORMAL, nullptr);
```
jherico: don't we want the blue texture here?

ZappoMan:
can we make a script that just disables all textures and include that in this PR/test plan...

Nex-Pro:
@ZappoMan remove QA ready until then ?

ZappoMan:
@Nex-Pro - no - this is still QA ready -- my comment was more of a request for the devs

OmegaHeron:
Testing now

OmegaHeron:
Tested Windows 10
QA Pass

General smoke test = pass.

debugDeferredLighting.js does contain textures tick box and it toggles textures matching its checked/!checked state.

PR #9561, opened 2017-01-31
Fix simulation owned entities not to go to origin after cache reload (+ small priority fix)

thoys:
Fixes bug where the user is still in the same session while having simulation ownership over some entities but the entity tree is reset, this caused the entities to show up at origin of domain before this PR.

## QA Testing

- Have at least three interface controlled avatars in this test
- Physics should work as smooth as they work in the master branch
- Selecting an entity in edit mode followed by reloading the cache should not move the last selected entity to the origin anymore.
- Play around with the physics, reconnect a couple of times, reload cache a couple of times, make sure it all works like usual.


hifi-gustavo:
Build console output for commit: c3f9663ab0addb14ed9e3c6d64fe249d7f255b9f available at: https://gist.github.com/0705048a431b91e64dcecd9f505f289c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9561-c3f9663.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9561-c3f9663.dmg

libraries/entities/src/EntityItem.cpp
```diff
@@ -688,6 +688,14 @@ int EntityItem::readEntityDataFromBuffer(const unsigned char* data, int bytesLef
...
                 _simulationOwner.clearCurrentOwner();
             }
+        } else if (newSimOwner.matchesValidID(myNodeID) && !(_dirtyFlags & Simulation::DIRTY_SIMULATION_OWNERSHIP_FOR_POKE)
+            && !(_dirtyFlags & Simulation::DIRTY_SIMULATION_OWNERSHIP_FOR_GRAB)) {
+            // entity-server tells us that we have simulation ownership while we never requested this for this EntityItem,
+            // this could happen when the user reloads the cache and entity tree.
+            _dirtyFlags |= Simulation::DIRTY_SIMULATOR_ID;
+            somethingChanged = true;
+            _simulationOwner.clearCurrentOwner();
+            weOwnSimulation = false;
```
AndrewMeadows: I don't think this will work.  The DIRTY_SIMULATION_OWNERSHIP_FOR_POKE and _GRAB flags are intended to inform the "do I need to bid for this object" logic, and are promptly cleared when the bid goes out.

Also, bids at other priority levels could go out (triggered by collisions in the physics simulation for example) that never set the GRAB and POKE bits.

What you want to filter for is a truly unsolicited simulation ownership grant from the EntityServer.  There might be a clever way to figure it out (I can't think of what it would be yet) or maybe it will be necessary to add a bool that answers the question "Did I **ever** send a bid for simulation ownership of this object?".

thoys: I agree, especially since the dirty flags weren't very reliable and they didn't take account for the volunteer bids. I have changed the mechanism to use a boolean now which it sets to true the first time it tries to get simulation ownership over the entity:

hifi-gustavo:
Build console output for commit: 360899887775ae01049d57ab552f4cc83e6f479a available at: https://gist.github.com/0705048a431b91e64dcecd9f505f289c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9561-3608998.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9561-3608998.dmg

libraries/entities/src/EntityItem.cpp
AndrewMeadows: This part looks like it will work.

libraries/entities/src/EntityItem.cpp
```diff
@@ -1271,18 +1278,20 @@ void EntityItem::pokeSimulationOwnership() {
         // we don't own it yet
         _simulationOwner.setPendingPriority(SCRIPT_POKE_SIMULATION_PRIORITY, usecTimestampNow());
     }
+    checkForFirstSimulationBid(_simulationOwner);
```
AndrewMeadows: Let's not call `checkForFirstSimulationBid()` everywhere when updating the `EntityItem::_simulationOwner`.  Instead lets do the following:

(1) Remove all calls to `checkForFirstSimulationBid()`.

(2) Change the function to be something like `EntityItem::rememberHasBid() { _hasBidOnSimulation = true; }`

The reason I want the function simplified is for optimization: setting a bool false is faster than comparing two Uuid's for equality.

(3) I believe we only need to call `EntityItem::rememberHasBid()` in three places:

(a) EntityScriptingInterface.cpp near line 445 (in the `else` case)
(b) EntityScriptingInterface.cpp near line 232 (in the `if(_bidOnSimulationOwnership)` case)
(c) EntityMotionState.cpp near line 581 (in the first `else if` case)

That should do the trick.

libraries/entities/src/EntityItem.h
```diff
@@ -586,6 +588,9 @@ class EntityItem : public SpatiallyNestable, public ReadWriteLockable {
     static quint64 _rememberDeletedActionTime;
     mutable QHash<QUuid, quint64> _previouslyDeletedActions;
 
+    // per entity keep state if it ever bid on simulation, so that we can ignore false simulation ownership
+    mutable bool _hasBidOnSimulation = false;
```
AndrewMeadows: Hrm... we haven't updated the coding standard about this item but we prefer this format for initializing class data members at declaration:

`mutable bool _hasBidOnSimulation { false };`

thoys: Ah ok, should I update the others in the classes that I modify?

libraries/entities/src/EntityItem.cpp
thoys: Ok, but we need to make sure any new future bid functionality will also use the `remberHasBid` function, updating the code now.

hifi-gustavo:
Build console output for commit: 174a7ad5bdadc72f14875281889b26fb7706ac2b available at: https://gist.github.com/0705048a431b91e64dcecd9f505f289c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9561-174a7ad.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9561-174a7ad.dmg

Nex-Pro:
Testing this

Nex-Pro:
### PASS  

tested with three geographically separated Clients at dev-playa, dev-welcome and solo at localhost.  no problems seen. Good work, thanks.

PR #9565, opened 2017-01-31
Fix bug that causes GCC not to detect AVX2

kencooke:
There's a bug in the cpuid.h included with GCC that causes __get_cpuid(0x7) to return incorrect values. This results in failure to detect AVX2 support on Linux, and a severe loss in performance of the audio mixer.

Fixed by replacing __get_cpuid(0x7) with an alternate implementation.

### DIFF
--- a/libraries/shared/src/CPUDetect.h
+++ b/libraries/shared/src/CPUDetect.h
@@ -134,7 +134,7 @@ static inline bool cpuSupportsAVX() {
             result = true;
         }
     }
-    return result;    
+    return result;
 }
 
 static inline bool cpuSupportsAVX2() {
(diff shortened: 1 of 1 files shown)


kencooke:
### Test Plan

Verify that the Linux audio-mixer in this PR can mix substantially more avatars without audio throttling. There should be no functional changes, only performance.

hifi-gustavo:
Build console output for commit: 2ce357db3a1ea99eb93b28c702cba9df2d850ac2 available at: https://gist.github.com/c595e5fed964640ec1ced173a5ac8295<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9565-2ce357d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9565-2ce357d.dmg

OmegaHeron:
This compiles proper on Ubuntu 16.04.1.  Unless I'm missing it, there's no indication in log that AC audio server is or is not using AVX for its heavy lifting -- that might be a nice addition.

General smoke test passes
Audio from 2 agents shows a drop in CPU use over current dev master, so, *I think*, this is working.

PR #9581, opened 2017-02-01
Add log window for the EntityScriptServer in interface.

Atlante45:
- Added log streaming capability from the EntityScriptServer to allowed clients.
- Added log window for that purpose that subs/unsubs from the EntityScriptServer log automatically.
- Added HMD friendly log window.

## Test Plan:
- Go to a domain where you have rez rights
- `Developer > Entity Script Server Log` should be enabled
- Open it and run a simple entity server script
- Verify you can see the log
- Change domain settings so you don't have rez rights
- Log window should warn you that you have been unsubscribed from the log and the menu item should be grayed out.
- Come back to your domain.
- Restart the domain
- You should be warned the connection is lost. It should come back by itself.
- Run the essDebugWindow.js script
- Make sure the script window shows the same information and subs/unsub in the same way


OmegaHeron:
Will be testing this.  Windows & Linux

interface/src/ui/EntityScriptServerLogDialog.cpp
```diff
@@ -0,0 +1,58 @@
...
+//
+
+#include "EntityScriptServerLogDialog.h"
+
+EntityScriptServerLogDialog::EntityScriptServerLogDialog(QWidget* parent) : BaseLogDialog(parent) {
+    setWindowTitle("Entity Script Server Log");
+
+    auto nodeList = DependencyManager::get<NodeList>();
+    auto& packetReceiver = nodeList->getPacketReceiver();
+    packetReceiver.registerListener(PacketType::EntityServerScriptLog, this, "handleEntityServerScriptLogPacket");
```
ZappoMan: It would really be better if there was a way to access these logs from the HMD... or from a script.

If we could expose the following on the `EnityScriptingInterface` class we could implement a QML/HMD friendly log...
```
void setEnableEntityServerScriptLog(bool enabled);
bool getEnableEntityServerScriptLog();
bool getCanEnableEntityServerScriptLog();
// signal:
void entityServerScriptLog(QString newLogData);
```

marko8904:
test this please

interface/src/ui/EntityScriptServerLogDialog.cpp
Atlante45: Interesting, yes I guess we could expose something like that.
I haven't yet looked at you HMD friendly script log. Would it be easy to reuse that?

Ideally, I'd rather we port the current log windows to QML have a mechanism to unstick QML window from interface.
Any time I debug a script in interface I have my log window on my second screen, or sticking out behind my interface. Having them constrained to the opengl surface would be really annoying.
Although I understand sometimes it's nice to have it in the HMD.

ZappoMan: debugWindow.js is basically a QML window with a thin bridge to hook the log signal to adding a line of text to the QML... it's trivial... yes you could use it or borrow heavily for it... you could basically log into the same window or a second.

The current "window based" log window is useless for HMD development.

hifi-gustavo:
Build console output for commit: f7f87f3f23894c897baf260822b71c66617a509a available at: https://gist.github.com/cf61755cd4b7cb6333b58ec9d9df68b4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-f7f87f3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-f7f87f3.dmg

OmegaHeron:
Tested with Interface and Stack on Windows 10 and Ubuntu 16.04.  Tested with localhost and LAN transport sessions with Interface Win/Stack Linux, Interface win/Stack Linux, Interface Linux/Stack Win and Interface Linux/Stack Linux.

QA Pass, but consider my commentary on rez vs rez-temp before merging, please.

If agent has either rez or temp-rez rights it can view log.
If, during a session, those perms are removed log out shows new state and menu item is disabled.
If user never had rez/temp-rez right menu item is not enabled.

Now - if I were king of the world I would restrict this to those with rez rights.  Let me explain.  Temp rez implies a lesser level of trust than rez (at least in my mind).  I don't necessarily wish a user with *only* temp-rez perms to possibly see something in my ESS logs that I wouldn't have a problem with a user at the higher level of trust rez right conveys.

hifi-gustavo:
Build console output for commit: 94a4c17ca205d24da375b4dac14c754ded61fa22 available at: https://gist.github.com/cf61755cd4b7cb6333b58ec9d9df68b4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-94a4c17.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-94a4c17.dmg

OmegaHeron:
While waiting for Jenkins...

Tested Stack/Interface with Ubuntu 16.04.1

ESS log notifies and reconnects on a domain change.
ESS log now restricted to only those with rez rights entitlement.

Will test HMD magic in Windows once Jenkins catches up.

hifi-gustavo:
Build console output for commit: eb655b7c8f3996bdbb414eb92e09791b538e0286 available at: https://gist.github.com/cf61755cd4b7cb6333b58ec9d9df68b4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-eb655b7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-eb655b7.dmg

Atlante45:
@OmegaHeron Can you wait a bit? I'm still testing a couple things.
Will update the testing plan and tag you when it's ready.

OmegaHeron:
Be tomorrow then - on way out shortly.  Also check condition of having only temp-rez rights but still having menu to call up log (though you can't see any output).

Atlante45:
@OmegaHeron Sounds good. PR is ready to go. Updating test plan now.

scripts/developer/debugging/essDebugWindow.js
```diff
@@ -0,0 +1,28 @@
...
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+(function() { // BEGIN LOCAL_SCOPE
+
+// Set up the qml ui
+var qml = Script.resolvePath('debugWindow.qml');
+var window = new OverlayWindow({
+    title: 'Debug Window',
```
ZappoMan: you might want to make this title different from the other window... just to make it clear which is which.

interface/src/EntityScriptServerLogClient.cpp
```diff
@@ -0,0 +1,88 @@
...
+    QMetaObject::invokeMethod(this, "connectionsChanged", Qt::QueuedConnection);
+}
+
+void EntityScriptServerLogClient::disconnectNotify(const QMetaMethod& signal) {
+    // This needs to be delayed or "receivers()" will return completely inconsistent values
+    QMetaObject::invokeMethod(this, "connectionsChanged", Qt::QueuedConnection);
+}
+
+void EntityScriptServerLogClient::connectionsChanged() {
+    qDebug() << Q_FUNC_INFO << _subscribed << receivers(SIGNAL(receivedNewLogLines(QString)));
```
ZappoMan: log spam

interface/src/EntityScriptServerLogClient.cpp
```diff
@@ -0,0 +1,88 @@
...
+        if (_subscribed != enable) {
+            if (enable) {
+                emit receivedNewLogLines("====================== Subscribded to the Entity Script Server's log ======================");
+            } else {
+                emit receivedNewLogLines("==================== Unsubscribded from the Entity Script Server's log ====================");
+            }
+        }
+        _subscribed = enable;
+    } else {
+        qWarning() << "Entity Script Server not found";
```
ZappoMan: log spam -- use `qCWarning()`

assignment-client/src/scripts/EntityScriptServer.cpp
```diff
@@ -100,6 +133,67 @@ void EntityScriptServer::handleEntityScriptGetStatusPacket(QSharedPointer<Receiv
...
+        } else {
+            qWarning() << "Node not found";
+        }
+    }
+
+    auto node = nodeList->nodeWithUUID(*last);
+    if (node && node->getActiveSocket()) {
+        nodeList->sendPacket(std::move(packet), *node);
+    } else {
+        qWarning() << "Node not found";
```
ZappoMan: log spam -- use `qCWarning()`

assignment-client/src/scripts/EntityScriptServer.cpp
```diff
@@ -100,6 +133,67 @@ void EntityScriptServer::handleEntityScriptGetStatusPacket(QSharedPointer<Receiv
...
+    bool enable = false;
+    message->readPrimitive(&enable);
+
+    auto senderUUID = senderNode->getUUID();
+    auto it = _logListeners.find(senderUUID);
+
+    if (enable && senderNode->getCanRez()) {
+        if (it == std::end(_logListeners)) {
+            _logListeners.insert(senderUUID);
+            qInfo() << "Node" << senderUUID << "subscribed to log stream";
```
ZappoMan: log spam -- use `qCInfo()`

assignment-client/src/scripts/EntityScriptServer.cpp
```diff
@@ -100,6 +133,67 @@ void EntityScriptServer::handleEntityScriptGetStatusPacket(QSharedPointer<Receiv
...
+        }
+    } else {
+        if (it != std::end(_logListeners)) {
+            _logListeners.erase(it);
+            qInfo() << "Node" << senderUUID << "unsubscribed from log stream";
```
ZappoMan: log spam -- use `qCInfo()`

assignment-client/src/scripts/EntityScriptServer.cpp
```diff
@@ -100,6 +133,67 @@ void EntityScriptServer::handleEntityScriptGetStatusPacket(QSharedPointer<Receiv
...
+    auto nodeList = DependencyManager::get<NodeList>();
+
+    auto last = --std::end(_logListeners);
+    for (auto it = std::begin(_logListeners); it != last; ++it) {
+        auto node = nodeList->nodeWithUUID(*it);
+        if (node && node->getActiveSocket()) {
+            auto copy = NLPacket::createCopy(*packet);
+            nodeList->sendPacket(std::move(copy), *node);
+        } else {
+            qWarning() << "Node not found";
```
ZappoMan: log spam -- use `qCInfo/Warning()`

hifi-gustavo:
Build console output for commit: 2f295c1ffd1d23669e7db6ca220869883583941e available at: https://gist.github.com/cf61755cd4b7cb6333b58ec9d9df68b4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-2f295c1.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-2f295c1.dmg

hifi-gustavo:
Build console output for commit: 5e171a39c6a7017113d885659594933498e1afc7 available at: https://gist.github.com/cf61755cd4b7cb6333b58ec9d9df68b4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-5e171a3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-5e171a3.dmg

OmegaHeron:
I'm unsure why, but, with this PR goto (using enter key or hud button) has no content.

DEV-5994

hifi-snap-by-omegaheron-on-2017-02-03_11-22-42

This PR

hifi-snap-by-omegaheron-on-2017-02-03_11-24-58

If issue with locations/goto is sorted and it's either not fault of this PR or otherwise fixed then the rest is QA Pass.

Tested Windows 10 and Ubuntu 16.04.1

Verified that only agent permission to enable ESS log output is rez.

Log maintains function upon domain change.  If changing to another domain where agent has rez right then log output shows subscription to new ESS, otherwise it shows unsubscribed.

Log output stops upon leaving domain where ESS log was valid (as it should).

Log output in Interface matches log output from stack.


HMD ESS log via essDebugWindow.js works same as !HMD log display, but I did notice that I had to leave/return to domain where I could see ESS log the first time I ran it - I don't think that's a major issue, and trying it again a few times I didn't have to leave/return to get output.

davidkelly:
When there is a packet version change, this will happen.  We only show you domain that have matching protocols.

OmegaHeron:
:facepalms: :)  Ty @davidkelly - I had somehow, over all this time of doing these, never really noticed that.

QA Pass

assignment-client/src/AssignmentClientLogging.cpp
```diff
@@ -3,7 +3,7 @@
 //  assignment-client/src
 //
 //  Created by Clement on 12/14/15.
-//  Copyright 2015 High Fidelity, Inc.
+//  Copyright 2017 High Fidelity, Inc.
```
ZappoMan: technically you want to either keep the original copyright date when the work was created... and/or use a 2015-2017 for work that has been changed during the time. Basically all new work is copyrighted when it is created

assignment-client/src/scripts/EntityScriptServer.cpp
```diff
@@ -57,6 +78,18 @@ EntityScriptServer::EntityScriptServer(ReceivedMessage& message) : ThreadedAssig
 
     packetReceiver.registerListener(PacketType::ReloadEntityServerScript, this, "handleReloadEntityServerScriptPacket");
     packetReceiver.registerListener(PacketType::EntityScriptGetStatus, this, "handleEntityScriptGetStatusPacket");
+    packetReceiver.registerListener(PacketType::EntityServerScriptLog, this, "handleEntityServerScriptLogPacket");
+
+    static const int LOG_INTERVAL = MSECS_PER_SECOND / 50;
+    auto timer = new QTimer(this);
+    timer->setTimerType(Qt::PreciseTimer);
```
ZappoMan: does this really need to be a precise timer? I'm not sure if it does... ~50hz admittedly isn't very accurate on windows unless you mark it as precise... but... I'm not sure 50hz is needed... not a blocker, just something to consider. Precise timers are presumably more expensive (especially on windows).

assignment-client/src/scripts/EntityScriptServer.cpp
```diff
@@ -100,6 +133,63 @@ void EntityScriptServer::handleEntityScriptGetStatusPacket(QSharedPointer<Receiv
...
+        return;
+    }
+
+    auto packet = NLPacket::create(PacketType::EntityServerScriptLog, buffer.size(), true);
+    packet->write(buffer.data(), buffer.size());
+
+    auto nodeList = DependencyManager::get<NodeList>();
+
+    auto last = --std::end(_logListeners);
+    for (auto it = std::begin(_logListeners); it != last; ++it) {
```
ZappoMan: wouldn't a range based for loop be cleaner here? I mean, it certainly would make the code easier to read.

assignment-client/src/scripts/EntityScriptServer.cpp
```diff
@@ -154,9 +244,67 @@ void EntityScriptServer::run() {
...
+
 void EntityScriptServer::nodeActivated(SharedNodePointer activatedNode) {
-    if (activatedNode->getType() == NodeType::AudioMixer) {
-        negotiateAudioFormat();
+    switch (activatedNode->getType()) {
+        case NodeType::AudioMixer:
+            negotiateAudioFormat();
+            break;
+        case NodeType::Agent: {
+            auto activatedNodeUUID = activatedNode->getUUID();
```
ZappoMan: It looks the the intent of this code is to support agents coming & going (flapping) that have previously registered for logs to sort of "automatically re-register them for logs"... I can't think of any place else in the product where we do this... 

That's not a bad thing per se, but I think it might lead to a little bit of confusion about why this feature behaves differently from other features... (one example would be ignore... where this same basic concept could apply).

I would suggest at a minimum we should add some comments somewhere around this code or maybe the `_killedListeners` member variable to indicate that this feature behaves this way. And we probably want to add some documentation to the feature docs that explain this as well. Finally, we probably want to add some bugs to fogbugz about the other features that don't handle this flapping, and to note that really this should be generalized into NodeList or something...

davidkelly:
@OmegaHeron it isn't obvious what is/isn't shown in goto.  But we tried to make it stuff you definitely could go to.  In a perfect world we'd understand the permissions for all the domains, and only show you places you can go to based on the specific domain settings.  But for now, we show you domains with matching protocols.  Plus it does a check to see if you are logged in, and if not doesn't show you domains that require login for connect (IIRC).

interface/src/EntityScriptServerLogClient.h
```diff
@@ -0,0 +1,46 @@
+//
+//  EntityScriptServerLogClient.h
```
ZappoMan: Maybe this should be in a library like `Entities` or `Networking` this doesn't really need to be in interface, and I can imagine that someone might want to add logging support to an agent script or some other EXE in the platform...

ZappoMan: If it was just part of the `EntitiesScriptingInterface` it would also make sense, and would just automatically be available to agents.

assignment-client/src/AssignmentClientLogging.cpp
Atlante45: Never was quite sure how to handle those, that makes a lot of sense.

assignment-client/src/scripts/EntityScriptServer.cpp
Atlante45: Not absolutely needed. Admittedly 50Hz is not a value that's set in stone, 10Hz would probably be enough here. I just want the log to stay pretty responsive.

assignment-client/src/scripts/EntityScriptServer.cpp
Atlante45: I'm not sure if you mean what I think you mean when you say the agent is flapping. During a "real" flap (ie. Agents purposely disconnects/reconnects) then it will have a different UUID it's like it's an all new node so it's not an issue here.
If the agent loses connection to the ESS for a bit, it will get the `nodeKilled`/`nodeActivated` callbacks and should handle those accordingly.

This is specifically to handle the case where the Agent appears to flaps from the POV of the ESS when it didn't actually flap. So the connection was never broken from the POV of the Agent.
In this case, you would get unsubbed from the log without realizing.

If this is what you meant, then I don't think this statement is correct
> it might lead to a little bit of confusion about why this feature behaves differently from other features...
This is a bug in those other features that should be fixed.
If I connect to a node and establish some kind of contract with the node. I don't expect that contract to be broken without warning when I have never lost connection to that node.

Ignore would actually not be affected by this since you keep the same UUID and we don't clear those on node disconnects.
One feature that might be affected though is Messages. Admittedly this is a rare occurrence, but in the cases where it happens, then the Agent gets unsubbed from all his channels and has no way to realize it.

hifi-gustavo:
Build console output for commit: e488192227d0b809ec3dd81a531fb3972af6c477 available at: https://gist.github.com/cf61755cd4b7cb6333b58ec9d9df68b4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-e488192.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-e488192.dmg

assignment-client/src/scripts/EntityScriptServer.cpp
ZappoMan: I guess what I'm asking is... when/why does the ESS get a `nodeKilled` event when in fact the interface has not disconnected... that sounds like a bug in `NodeList` so it seems weird that we're writing code to handle this case here... but not other places.

Atlante45: This is expected behavior, if you don't hear from a node for over 5 sec, it gets killed from your list. Since the last heard timestamp cannot be perfectly symmetric, it is technically possible for 2 nodes to lose connection for a period of time that would trigger that case.

ZappoMan: does the interface client send something to the ESS? it's not sending pings to the ESS... what messages does the interface send to the ESS? isn't it only "I'm interested in logging"... wouldn't it constantly be dropping?

Atlante45: It works on the same model the messages do, it sends subscribe/unsubscribe packets.
Subscribed nodes will get the log forwarded to them.

Atlante45: Just got what you meant. The NodeList automatically sends keep-alive packets.
This would only happen if the network between the 2 nodes actually goes down or if it is heavily congested.

OmegaHeron:
Tested Windows 10 and Ubuntu 16.04.1

Same result as last test, QA Pass.

Atlante45:
@ZappoMan Do you need any more changes on this?

hifi-gustavo:
Build console output for commit: 574326d0d86bac211b7bb28eb5c8f71bc3385428 available at: https://gist.github.com/cf61755cd4b7cb6333b58ec9d9df68b4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-574326d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-574326d.dmg

hifi-gustavo:
Build console output for commit: 3a30de79aa052e41e2b845a241ca495d668df670 available at: https://gist.github.com/cf61755cd4b7cb6333b58ec9d9df68b4<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-3a30de7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9581-3a30de7.dmg

PR #9583, opened 2017-02-02
fix the edit.js toolbar buttons so they don't vanish when moused-over or clicked

sethalves:
- fix the edit.js toolbar buttons so they don't vanish when moused-over or clicked

### DIFF
--- a/scripts/system/edit.js
+++ b/scripts/system/edit.js
@@ -210,7 +210,10 @@ var toolBar = (function () {
         var button = toolBar.addButton({
             objectName: name,
             imageURL: imageUrl,
-            buttonState: 1,
+            imageOffOut: 1,
+            imageOffIn: 2,
+            imageOnOut: 0,
+            imageOnIn: 2,
             alpha: 0.9,
             visible: true
         });


sethalves:
to test:
- open edit.js
- mouse over the edit.js buttons (for example, the "sphere" button)
- make sure the button doesn't vanish
- also try this with hmd+hand-controllers and using the half-squeezed lasers on these same buttons

hifi-gustavo:
Build console output for commit: 6671324a33bff4c2426591a856e66b1561b0090b available at: https://gist.github.com/8c5125f6227db52ea4274391e208725b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9583-6671324.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9583-6671324.dmg

Nex-Pro:
Checking this.

Nex-Pro:
I think for this pr this is a pass but please read note.

In normal use the tool bar icons behave well.(HMD and desktop)

Note:
In HMD if the avatar is moving in any way via hand controllers and a toolbar icon is clicked eg: sphere or cube the toolbar icon becomes toggled to the 'over' condition ( 50% or so transparent ) subsequent pressing of the trigger clears it.

To repro:
 Drive backwards with left controller click cube or sphere icon on toolbar.
Observe Icon remains toggled.

Also happens with up/down etc.

sethalves:
okay, I'll make a note about the button-state problem.

PR #9584, opened 2017-02-02
don't auto-bid for sim ownership on entity creation in ESI

birarda:
- fixed an bug where non-dynamic entities created by an Interface client and updated by an agent or ESS would appear frozen to all clients

### DIFF
--- a/libraries/entities/src/EntityScriptingInterface.cpp
+++ b/libraries/entities/src/EntityScriptingInterface.cpp
@@ -223,17 +223,6 @@ QUuid EntityScriptingInterface::addEntity(const EntityItemProperties& properties
                     }
                 }
 
-                if (_bidOnSimulationOwnership) {
-                    // This Node is creating a new object.  If it's in motion, set this Node as the simulator.
-                    auto nodeList = DependencyManager::get<NodeList>();
-                    const QUuid myNodeID = nodeList->getSessionUUID();
-
-                    // and make note of it now, so we can act on it right away.
-                    propertiesWithSimID.setSimulationOwner(myNodeID, SCRIPT_POKE_SIMULATION_PRIORITY);
-                    entity->setSimulationOwner(myNodeID, SCRIPT_POKE_SIMULATION_PRIORITY);
-                    entity->rememberHasSimulationOwnershipBid();
-                }
-
                 entity->setLastBroadcast(usecTimestampNow());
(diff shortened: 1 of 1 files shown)


birarda:
#### Test Plan

The first part of the test plan is to confirm that this PR fixes the behaviour that helped us identify the bug. 

(1) Run https://raw.githubusercontent.com/huffman/hifi/8acb3f9b531dab813eff14ae99d8ecf7d9ab1d5b/unpublishedScripts/marketplace/stopwatch/spawnStopwatch.js from your Interface client. This will spawn a stopwatch in front of you.
(2) Click on the stopwatch to start it. The expected behaviour is that the hands move (seconds and minutes) and that you hear a tick sound every second.
(3) On master, you should not see the stopwatch hands moving until you quit Interface and return. On this PR the stopwatch hands should immediately start moving when you click the stopwatch after running the spawner script.

The second part of the test plan is to confirm that this PR does not effect the existing behaviour of dynamic entities. Interact with various objects from the marketplace (like the gun or the exploding gnomes) with at least one other client in Interface and confirm that the behaviour you see is what you expect. Effectively, perform a general smoke test of the dynamic behaviour of entities with another user.

hifi-gustavo:
Build console output for commit: 823d74d2d4516408e0529495bd63e11f24f98944 available at: https://gist.github.com/8844bd9715e3b23fc88d5203cb0663ef<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9584-823d74d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9584-823d74d.dmg

nbq:
I will test this

nbq:
Tester: Coal

* ran script on one interface and verified it worked ok, the hands started moving without having to relog so this is a PASS!
* I created an exploding gnome and threw it and both interfaces saw the gnome act exactly as it should
* Using the interface with Vive on, I created from marketplace the pingponggun, the grappling hook and the flare gun - Both Interfaces saw the items on the ground but the second interface could never see the item once it was picked up and equipped. I tried this multiple times restarting the Interfaces on both sides and no matter what, the other Interface could not see the item equipped in the other Interface's hands. This part is a FAIL for me.


**FAIL**

nbq:
Updated post above with a PASS

PR #9594, opened 2017-02-02
Cap fbx texture size

AndrewMeadows:
* absolute cap of 8192x8192 for all textures
* cap textures from FBX files to 2048x2048, with smaller caps for models that use many textures:
    1-8 unique textures = 2048x2048
    9-32 unique textures = 1024x1024
    33-128 unique textures = 512x512

### DIFF
--- a/libraries/fbx/src/FBXReader.h
+++ b/libraries/fbx/src/FBXReader.h
@@ -105,7 +105,7 @@ public:
     glm::mat4 inverseBindMatrix;
 };
 
-const int MAX_FBX_TEXTURE_SIZE = 1024 * 1024;
+const int MAX_FBX_TEXTURE_NUM_PIXELS = 1024 * 1024;
 
 /// A texture map in an FBX document.
 class FBXTexture {
@@ -115,7 +115,7 @@ public:
     QByteArray content;
 
     Transform transform;
-    int maxSize { MAX_FBX_TEXTURE_SIZE };
+    int maxNumPixels { MAX_FBX_TEXTURE_NUM_PIXELS };
     int texcoordSet;
     QString texcoordSetName;
 
--- a/libraries/fbx/src/FBXReader.h
+++ b/libraries/fbx/src/FBXReader.h
@@ -13,6 +13,7 @@
 #define hifi_FBXReader_h
 
 #include <QMetaType>
+#include <QSet>
 #include <QUrl>
 #include <QVarLengthArray>
 #include <QVariant>
@@ -105,7 +106,7 @@ public:
     glm::mat4 inverseBindMatrix;
 };
 
-const int MAX_FBX_TEXTURE_NUM_PIXELS = 1024 * 1024;
+const int MAX_NUM_PIXELS_FOR_FBX_TEXTURE = 2048 * 2048;
 
 /// A texture map in an FBX document.
 class FBXTexture {
@@ -115,7 +116,7 @@ public:
     QByteArray content;
 
     Transform transform;
-    int maxNumPixels { MAX_FBX_TEXTURE_NUM_PIXELS };
+    int maxNumPixels { MAX_NUM_PIXELS_FOR_FBX_TEXTURE };
     int texcoordSet;
     QString texcoordSetName;
 
@@ -145,6 +146,9 @@ public:
         emissiveColor(emissiveColor),
         shininess(shininess),
         opacity(opacity)  {}
+
+    void getTextureNames(QSet<QString>& textureList) const;
+    void setMaxNumPixelsPerTexture(int maxNumPixels);
 
     glm::vec3 diffuseColor{ 1.0f };
     float diffuseFactor{ 1.0f };
--- a/libraries/fbx/src/FBXReader.h
+++ b/libraries/fbx/src/FBXReader.h
@@ -100,10 +100,12 @@ public:
 /// A single binding to a joint in an FBX document.
 class FBXCluster {
 public:
-    
+
     int jointIndex;
     glm::mat4 inverseBindMatrix;
 };
+
+const int MAX_FBX_TEXTURE_SIZE = 1024 * 1024;
 
 /// A texture map in an FBX document.
 class FBXTexture {
@@ -111,11 +113,12 @@ public:
     QString name;
     QByteArray filename;
     QByteArray content;
-    
+
     Transform transform;
+    int maxSize { MAX_FBX_TEXTURE_SIZE };
     int texcoordSet;
     QString texcoordSetName;
-    
+
     bool isBumpmap{ false };
 
     bool isNull() const { return name.isEmpty() && filename.isEmpty() && content.isEmpty(); }
--- a/libraries/model-networking/src/model-networking/ModelCache.cpp
+++ b/libraries/model-networking/src/model-networking/ModelCache.cpp
@@ -472,7 +472,7 @@ QUrl NetworkMaterial::getTextureUrl(const QUrl& baseUrl, const FBXTexture& textu
 model::TextureMapPointer NetworkMaterial::fetchTextureMap(const QUrl& baseUrl, const FBXTexture& fbxTexture,
                                                         TextureType type, MapChannel channel) {
     const auto url = getTextureUrl(baseUrl, fbxTexture);
-    const auto texture = DependencyManager::get<TextureCache>()->getTexture(url, type, fbxTexture.content);
+    const auto texture = DependencyManager::get<TextureCache>()->getTexture(url, type, fbxTexture.content, fbxTexture.maxSize);
     _textures[channel] = Texture { fbxTexture.name, texture };
 
     auto map = std::make_shared<model::TextureMap>();
--- a/libraries/model-networking/src/model-networking/ModelCache.cpp
+++ b/libraries/model-networking/src/model-networking/ModelCache.cpp
@@ -472,7 +472,7 @@ QUrl NetworkMaterial::getTextureUrl(const QUrl& baseUrl, const FBXTexture& textu
 model::TextureMapPointer NetworkMaterial::fetchTextureMap(const QUrl& baseUrl, const FBXTexture& fbxTexture,
                                                         TextureType type, MapChannel channel) {
     const auto url = getTextureUrl(baseUrl, fbxTexture);
-    const auto texture = DependencyManager::get<TextureCache>()->getTexture(url, type, fbxTexture.content, fbxTexture.maxSize);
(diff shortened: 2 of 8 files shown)


libraries/fbx/src/FBXReader.h
```diff
@@ -100,22 +100,25 @@ class FBXJoint {
...
 class FBXTexture {
 public:
     QString name;
     QByteArray filename;
     QByteArray content;
-    
+
     Transform transform;
+    int maxSize { MAX_FBX_TEXTURE_SIZE };
     int texcoordSet;
```
AndrewMeadows: FBXTexture gets a new data member maxSize which is hard coded.

libraries/model-networking/src/model-networking/ModelCache.cpp
```diff
@@ -472,7 +472,7 @@ QUrl NetworkMaterial::getTextureUrl(const QUrl& baseUrl, const FBXTexture& textu
 model::TextureMapPointer NetworkMaterial::fetchTextureMap(const QUrl& baseUrl, const FBXTexture& fbxTexture,
                                                         TextureType type, MapChannel channel) {
     const auto url = getTextureUrl(baseUrl, fbxTexture);
-    const auto texture = DependencyManager::get<TextureCache>()->getTexture(url, type, fbxTexture.content);
+    const auto texture = DependencyManager::get<TextureCache>()->getTexture(url, type, fbxTexture.content, fbxTexture.maxSize);
```
AndrewMeadows: This is the one spot where `FBXTexture::maxSize` is used.

libraries/model-networking/src/model-networking/TextureCache.cpp
```diff
@@ -375,17 +381,32 @@ void ImageReader::run() {
...
+        int originalHeight = imageHeight;
+        imageWidth = (int)(scaleFactor * (float)imageWidth + 0.5f);
+        imageHeight = (int)(scaleFactor * (float)imageHeight + 0.5f);
+        //QSize newSize(imageWidth, imageHeight);
+        QImage newImage = image.scaled(QSize(imageWidth, imageHeight), Qt::IgnoreAspectRatio);
+        image.swap(newImage);
+        qCDebug(modelnetworking) << "Downscale image" << _url
+            << "from" << originalWidth << "x" << originalHeight
+            << "to" << imageWidth << "x" << imageHeight;
+    }
```
AndrewMeadows: This is where we apply the image size limits.  ATM we first parse the whole texture and then rescale it down if necessary.

libraries/model-networking/src/model-networking/TextureCache.h
```diff
@@ -23,6 +23,8 @@
 #include <ResourceCache.h>
 #include <model/TextureMap.h>
 
+const int ABSOLUTE_MAX_TEXTURE_SIZE = 8192 * 8192;
```
AndrewMeadows: ALL images have this absolute limit by default.

libraries/model-networking/src/model-networking/TextureCache.h
```diff
@@ -129,11 +132,11 @@ class TextureCache : public ResourceCache, public Dependency {
 
     /// Loads a texture from the specified URL.
     NetworkTexturePointer getTexture(const QUrl& url, Type type = Type::DEFAULT_TEXTURE,
-        const QByteArray& content = QByteArray());
+        const QByteArray& content = QByteArray(), int maxSize = ABSOLUTE_MAX_TEXTURE_SIZE);
```
AndrewMeadows: `TextureCache::getTexture()` has a default value for `maxSize`.  ATM the only call to this function where the `maxSize` is specified otherwise is where `FBXTextures` are used.

libraries/networking/src/ResourceCache.h
```diff
@@ -125,7 +125,7 @@ class ScriptableResource : public QObject {
      */
     Q_INVOKABLE void release();
 
-    const QUrl& getUrl() const { return _url; }
+    const QUrl& getURL() const { return _url; }
```
AndrewMeadows: These `getUrl()` method was contrary to our coding standards and was confusing me with other classes that had `getURL()` methods, so I made it conform.

libraries/fbx/src/FBXReader.h
```diff
@@ -100,22 +100,25 @@ class FBXJoint {
 /// A single binding to a joint in an FBX document.
 class FBXCluster {
 public:
-    
+
     int jointIndex;
     glm::mat4 inverseBindMatrix;
 };
 
+const int MAX_FBX_TEXTURE_SIZE = 1024 * 1024;
```
samcake: Maybe instead of SIZE we could find a different name for this since we are bounding the resolution aka the number of pixels more than the actual memory size in bytes.

AndrewMeadows: What do you think?  Is this a good hard-coded limit for model and avatar textures?

hifi-gustavo:
Build console output for commit: 64dbfd08194e9a61b253327657d0c0ea4e1a9fdb available at: https://gist.github.com/c3e726ecbbf1404a6f5f20a22fcc99fc<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9594-64dbfd0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9594-64dbfd0.dmg

AndrewMeadows:
test this please

hifi-gustavo:
Build console output for commit: 2f8823d10384af30a0cd8c452db557acbf583558 available at: https://gist.github.com/c3e726ecbbf1404a6f5f20a22fcc99fc<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9594-2f8823d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9594-2f8823d.dmg

hifi-gustavo:
Build console output for commit: 56911bdf1becc2a368e087f9c775252cade1187e available at: https://gist.github.com/c3e726ecbbf1404a6f5f20a22fcc99fc<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9594-56911bd.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9594-56911bd.dmg

AndrewMeadows:
What has changed:

We now set a maximum texture resolution per model based on how many unique textures it uses:

1 - 7 unique textures --> 2048x2048 max number of pixels
8 - 31  -->  1024x1024
32 - 127 --> 512x512
... etc

Note that there is a max resolution of 2048x2048 pixels for all FBX textures and an absolute maximum of 8192x8192 pixels for all other textures (such as skybox or UI).

The effects of these caps are:

(1) slightly faster load times in domains with several big textures
(2) lower quality textures on some models

To test:

(1) Copy this script to local file pr9594.js (yes, it is a one-line script):

`Window.location = "hifi://dev-chris.highfidelity.io/-26.2322,-0.942479,51.8355";`

(2) Find the `interface.exe` executable installed by this PR

(3) Launch the interface via the command line like this:

`interface.exe --testScript path/to/script/from/part/1/pr9594.js`

The domain should have about 100 avatars playing the walk animation.  The load will block and frame rate will be terrible until all of the avatars are loaded at which point the frame rate should be relatively stable and fast.  Note: the avatars are getting occasional updates from the server so their animation state will probably be much more clunky than the frame rate itself.

(4) Pay attention to how long it takes for the frame rate to settle down.  This time should not be longer than when logging into the same domain using the **hifi/dev-download** interface.

(5) Login to **dev-playa** domain and go to <x,y,z> = **<610, -8, 500>** which should be near a large model of a boy lying down on the playa looking at a lava lamp.  Note the quality of the texture on the boy's body (eyes, lips, and US flag texture on jacket pin).  This model uses a 4096x4096 texture which is clamped to 2048x2048 in this PR so there should be a slight reduction of texture quality.

hifi-gustavo:
Build console output for commit: 8b1f569132852cb61ffae0fe70ab1bea68460d78 available at: https://gist.github.com/c3e726ecbbf1404a6f5f20a22fcc99fc<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9594-8b1f569.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9594-8b1f569.dmg

AndrewMeadows:
test this please

hifi-gustavo:
Build console output for commit: 8b1f569132852cb61ffae0fe70ab1bea68460d78 available at: https://gist.github.com/c3e726ecbbf1404a6f5f20a22fcc99fc<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9594-8b1f569.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9594-8b1f569.dmg

Konshu:
I am curious what promoted the design to limit this? Was this a guesstimate to help performance? I ask only as this seems like it should be per domain variable that could be in an advanced menu on the domain to adjust per hostings desire. The only conflict I could see is if I designed my high texture quality avatar on my sim but went to say bobs, it may not work because our domains are configured differently.

PhilipRosedale:
@konshu:  Like you said, we think that a domain-specific restriction on sizes would not work because it would cause very confusing experiences when going to different places.    The logical place for a setting would be on the client, because you would want to match the setting to the download speeds and CPU capability of the client.  

Right now, large textures are both expensive to download and expensive to process into MIP levels on the client, resulting in big delays on first acquiring them.

AndrewMeadows:
@Konshu: Originally this PR used a simple 1024x1024 limit but after review and feedback the capping strategy has been modified and made more complicated.  I've updated the description to reflect that.

The overall goal is to speed up rendering on the client -- to allow the client to always stay around 90Hz with HMD on supported hardware regardless of what ridiculous content is thrown at it by the domain or other avatars.  This PR doesn't speed up the final render speed very much but does help maintain render speed while the textures and models are loaded.

I view this work as a mid-term HACK.  Long-term we need more comprehensive solutions, but they will also involve capping textures... just maybe the capping will be done before they are even loaded (client will request the MIP levels it wants).

samcake:
WE have been testing this PR intensively with no QA problem aside from potential concerns about the quality downgrade that could happen.
We think it s good to go in master and adjust the policy as needed in the future to keep the content creators and and the engineers happy.
Good to merge

PR #9608, opened 2017-02-03
provide a command-line argument to disable offscreenUi

sethalves:
- provide a command-line argument to disable offscreenUi


sethalves:
to test:
- no functional change

hifi-gustavo:
Build console output for commit: d2beb8d3cbe1962a88fc3127860158134ff5a76d available at: https://gist.github.com/26316a8cdab8be45ab5a247958303a28<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9608-d2beb8d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9608-d2beb8d.dmg

Nex-Pro:
@sethalves  Do you want us to test the command line argument ?

sethalves:
If you want, but I'm more concerned that nothing changes when the command-line argument isn't issued.

Nex-Pro:
Testing this

Nex-Pro:
had a crash on startup and on second run I had no scripts. But I was unable to repro.
On all other runs this was fine.

interface/src/ui/overlays/Overlays.cpp
```diff
@@ -639,11 +639,17 @@ void Overlays::setKeyboardFocusOverlay(unsigned int id) {
 }
 
 float Overlays::width() const {
+    if (!qApp->offscreenUiEnabled()) {
+        return 640.0f;
+    }
```
AndrewMeadows: I'm sometimes bummed to get my own magic numbers pointed out to me and yet I must point these out.  Maybe call this one NON_EXISTENT_OVERLAY_WIDTH?  I wonder why you use 640.0 here rather than 0.0?

interface/src/ui/OverlayConductor.cpp
```diff
@@ -100,6 +100,9 @@ void OverlayConductor::centerUI() {
 }
 
 void OverlayConductor::update(float dt) {
+    if (!qApp->offscreenUiEnabled()) {
+        return;
+    }
```
AndrewMeadows: In some places you query `qApp->offscreenUiEnabled()` and in others you check the validity of the `offscreenUi` instance below.  Both work.  Why use one and not the other?  I guess directly querying a bool data member of the class via pointer is actually faster than looking up a hash in a map so an efficiency argument could be made in favor of one.

jherico:
@sethalves you might want to look at the mechanism used to enable trace files based on the command line.  It uses a more flexible mechanism to populate global state variables that doesn't require changes in the Application API.  Look at `hifi::properties::TEST` in the Application.cpp code.  

If you used this mechanism, then in places in the code where you're currently doing something like this...

```
    auto offscreenUi = DependencyManager::get<OffscreenUi>();
    if (!offscreenUi) {
        ...
```

you could instead do this

```
    if (qApp->property(hifi::properties::DISABLE_UI).toBool()) { 
       ...
```

This is helpful because if you call `DependencyManager::get` for something that isn't there it creates a bunch of log spam.  It should also remove the need to add a new explicit property to `Application` and remove the need for an additional constructor argument.

hifi-gustavo:
Build console output for commit: a91f87a6c0e933d877c64ffa1b9b1071e27a87f5 available at: https://gist.github.com/26316a8cdab8be45ab5a247958303a28<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9608-a91f87a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9608-a91f87a.dmg

sethalves:
@AndrewMeadows @jherico I think I've addressed your comments.

hifi-gustavo:
Build console output for commit: a393d6db9271b2cd4a0815f34711260c17c04e93 available at: https://gist.github.com/26316a8cdab8be45ab5a247958303a28<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9608-a393d6d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9608-a393d6d.dmg

hifi-gustavo:
Build console output for commit: edb48baad5ba8ab02292bee5aa78eab0c2d73079 available at: https://gist.github.com/26316a8cdab8be45ab5a247958303a28<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9608-edb48ba.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9608-edb48ba.dmg

PR #9623, opened 2017-02-07
Optimize ignore zone checks for audio mixer

zzmp:


hifi-gustavo:
Build console output for commit: 5e9fb1794954a9853b8d9e933d522273aceed15a available at: https://gist.github.com/b5f5c5f98f583302d5894bdd7686ed1c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9623-5e9fb17.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9623-5e9fb17.dmg

zzmp:
Tested on a m4.4xlarge machines. `assignment-client` was compiled with `make assignment-client CXX_DEFINES='-DHIFI_AUDIO_THROTTLE_DEBUG' -j16`.

Started at an agent count (running [this script]()) that would not cause throttling, and let the agent number increase until throttling occured. `ns_per_throttle` was approximated after that, from watching it for a few seconds (it is *not exact*).

| commit | branch | max agents before throttling | ns_per_throttle @ 130 agents |
| :- | :- | -: | -: |
| ff56eb2 | master | 115 | 1475 |
| 5e9fb17 | #9623 | 123 | 1450 |

There doesn't appear to be a significant difference in `ns_per_throttle`. The time measured occurs after calls to `shouldIgnore*`, so this is expected.
There appears to be a slight increase in the number of avatars supportable. This is expected, as `shouldIgnore*` should have been optimized. I am surprised it is so small (~7%), and would like to run a more detailed perf before mergine.

*The above covers the load test portion of QA.*

### QA Plan

*This is not yet ready for QA.*

- Ensure the audio-mixer performs well, and sounds fine, with a few avatars in the domain.
- Ensure ignore bubbles continue to function as expected.
- Ensure server-related audio debugging features continue to function (echo server).

hifi-gustavo:
Build console output for commit: ee699d3fa60a51d88f1e2908eda51578cbf2e87c available at: https://gist.github.com/b5f5c5f98f583302d5894bdd7686ed1c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9623-ee699d3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9623-ee699d3.dmg

kencooke:
It looks like this gives about a 14% speedup (123/115)^2 of the mixing core. If the primary speedup is exploiting symmetry of the intersection-test (if A intersects B, B intersects A) then it looks great, about half of the original 30% that ignore() was taking. Nice.

Maybe we could get another 14% (hitting 130 avatars) by only computing every 100ms, with random initial counter values to distribute the load. Worth considering.

zzmp:
I also realized that the current `HIFI_AUDIO_THROTTLE_DEBUG` doesn't do what it advertises, so I am timing the entire mix instead.

| commit | branch | ns_per_mix @ 105 agents |
| :- | :- | -: |
| ff56eb2 + 553fffd | master | 7500 |
| 553fffd | #9623 | 5200 |

That's about a 30% speedup on the average case.

@kencooke, It may be that it only appeared as 14% because that was a worst-case test, and I looked at average-case here. It also should have been slightly more than a 2x speedup of `shouldIgnore*`, as it takes advantage of the symmetry mentioned as well as the fact that the intersection zone was recomputed for every node.

Either way, after looking at output with 553fffd, I am convinced that this is working as expected :). I'll talk to you offline about reducing the frequency of these computations.

hifi-gustavo:
Build console output for commit: 553fffd8aee5332b778793920071c1f4714c7373 available at: https://gist.github.com/b5f5c5f98f583302d5894bdd7686ed1c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9623-553fffd.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9623-553fffd.dmg

OmegaHeron:
I tested this with stack on both Windows and Linux along with interface on both.  No issue with privacy bubble, audio general or performance - caveat being that I had at most 2 avatars to test against.  Echo worked fine.  QA Pass for limited concurrent AV test I was able to perform.

PR #9640, opened 2017-02-09
New stats for avatar simulation

ZappoMan:
Test plan:
Part 1:
    - Run build
    - open onscreen stats "/"
    - click to expand stats
    - notice that you now have: "Fully Simulated Avatars", "Partially Simulated Avatars" in the first column, and "Avatars" under Frame Timing in the last column.

Part 2:
    - run the debugAvatarMixer.js script
    - notice you know have simulation stats that indicate how often that avatar has been simulated
    - also notice "hz" has been added to all the data rate columns, this indicates how often the data arrives as opposed to how much data is arriving

### DIFF
--- a/interface/src/avatar/Avatar.h
+++ b/interface/src/avatar/Avatar.h
@@ -182,6 +182,8 @@ public:
     void animateScaleChanges(float deltaTime);
     void setTargetScale(float targetScale) override;
 
+    Q_INVOKABLE float getSimulationRate(const QString& rateName = QString(""));
+
 public slots:
 
     // FIXME - these should be migrated to use Pose data instead
@@ -258,6 +260,12 @@ protected:
 
     void addToScene(AvatarSharedPointer self);
     void ensureInScene(AvatarSharedPointer self);
+
+    // Some rate tracking support
+    RateCounter<> _simulationRate;
+    RateCounter<> _skeletonModelSimulationRate;
+    RateCounter<> _jointDataSimulationRate;
+
 
 private:
     uint64_t _lastRenderUpdateTime { 0 };
--- a/interface/src/avatar/Avatar.h
+++ b/interface/src/avatar/Avatar.h
@@ -263,6 +263,7 @@ protected:
 
     // Some rate tracking support
     RateCounter<> _simulationRate;
+    RateCounter<> _simulationInViewRate;
     RateCounter<> _skeletonModelSimulationRate;
     RateCounter<> _jointDataSimulationRate;
 
--- a/interface/src/avatar/Avatar.h
+++ b/interface/src/avatar/Avatar.h
@@ -182,7 +182,7 @@ public:
     void animateScaleChanges(float deltaTime);
     void setTargetScale(float targetScale) override;
 
-    Q_INVOKABLE float getSimulationRate(const QString& rateName = QString(""));
+    Q_INVOKABLE float getSimulationRate(const QString& rateName = QString("")) const;
 
 public slots:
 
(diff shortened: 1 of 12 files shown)


hifi-gustavo:
Build console output for commit: dd2ddcc291831c01d7734551f48715c707c23ebc available at: https://gist.github.com/e6faed1c69c1f5be222f0e0bcfb38731<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9640-dd2ddcc.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9640-dd2ddcc.dmg

OmegaHeron:
Testing Windows 10

QA Pass

General smoke test = pass.

- notice that you now have: "Fully Simulated Avatars", "Partially Simulated Avatars" in the first column, and "Avatars" under Frame Timing in the last column.

Yes - zero based value :)  From a place with > 1 (myself) agents I see expected number as well.  I also see frame timing stat for avatar - it varies from 0.0 to 0.1 ms in a domain with other agents (2).


debugAvatarMixer.js loads and displays data as specified.

interface/src/avatar/AvatarManager.cpp
```diff
@@ -218,6 +230,9 @@ void AvatarManager::updateOtherAvatars(float deltaTime) {
     const uint64_t MAX_UPDATE_BUDGET = 2000; // usec
     uint64_t renderExpiry = startTime + RENDER_UPDATE_BUDGET;
     uint64_t maxExpiry = startTime + MAX_UPDATE_BUDGET;
```
AndrewMeadows: BTW, I'm thinking I'll probably move the `_startTime` measurement below the "sort" context.  Last time I checked the sort I was taking about 275 usec to handle 100 avatars (Windows Release).  In theory the sort scales linearly with number of avatars and so its cost could increase until we were unable to even process any avatar updates (somewhere around 700 or 800).  We could expand the overall time limits accordingly but the sort is non-negotiable so it should probably be moved out of the measure of negotiable processing.

interface/src/avatar/Avatar.h
```diff
@@ -182,6 +182,8 @@ class Avatar : public AvatarData {
     void animateScaleChanges(float deltaTime);
     void setTargetScale(float targetScale) override;
 
+    Q_INVOKABLE float getSimulationRate(const QString& rateName = QString(""));
```
AndrewMeadows: This method should be declared `const`.

interface/src/avatar/AvatarManager.cpp
ZappoMan: are you asking for a change here? or just making a comment...

interface/src/avatar/AvatarManager.cpp
```diff
@@ -218,6 +230,9 @@ void AvatarManager::updateOtherAvatars(float deltaTime) {
...
+
+    int fullySimulatedAvatars = 0;
+    int partiallySimulatedAvatars = 0;
```
ZappoMan: fixing `partiallySimulatedAvatars` - re the bail case, that's not tracked explicitly, but in that case the sum of `fullySimulatedAvatars` and `partiallySimulatedAvatars` is less than avatar count, so you can deduce it from the other stats.

interface/src/avatar/AvatarManager.cpp
AndrewMeadows: Just making a comment.

hifi-gustavo:
Build console output for commit: bb4654506a18c87838c8595477c59cc4d3c0be8d available at: https://gist.github.com/e6faed1c69c1f5be222f0e0bcfb38731<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9640-bb46545.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9640-bb46545.dmg

PR #9611, opened 2017-02-03
Add QMetaObject for AvatarEntityMap

kunalgosar:
Resolution to the issue described in: 
- https://github.com/highfidelity/hifi/issues/9447

Error Details:
[CRITICAL] [hifi.scriptengine] "[UncaughtException] TypeError: cannot call getAvatarEntityData(): unknown return type `AvatarEntityMap' (register the type with qScriptRegisterMetaType()) in Entities 8:32\n[Backtrace]\n (id = '{a275bceb-f49b-4554-bd9b-f51f30bae780}', ) at url:32\n () at -1"

Testing Instructions: 
- Call MyAvatar.getAvatarEntityData to get data into the scripting layer, ensure some entity is loaded otherwise data will be empty.
- Manipulate the data by changing some entity (color, position, etc)
- Call MyAvatar.setAvatarEntityData and check that changes follow what is expected

PR #9642, opened 2017-02-09
fix performance problem for invalid "avatarEntityData"

AndrewMeadows:
* fix a bug where invalid avatarEntityData could kill remote interface performance

### DIFF
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -231,13 +231,40 @@ void Avatar::updateAvatarEntities() {
     QScriptEngine scriptEngine;
     entityTree->withWriteLock([&] {
         AvatarEntityMap avatarEntities = getAvatarEntityData();
-        for (auto entityID : avatarEntities.keys()) {
+        AvatarEntityMap::const_iterator dataItr = avatarEntities.begin();
+        while (dataItr != avatarEntities.end()) {
+            // compute hash of data.  TODO? cache this?
+            QByteArray data = dataItr.value();
+            uint32_t newHash = qHash(data);
+
+            // check to see if we recognize this hash and whether it was already successfully processed
+            QUuid entityID = dataItr.key();
+            MapOfAvatarEntityDataHashes::iterator stateItr = _avatarEntityDataHashes.find(entityID);
+            if (stateItr != _avatarEntityDataHashes.end()) {
+                if (stateItr.value().success) {
+                    if (newHash == stateItr.value().hash) {
+                        // data hasn't changed --> nothing to do
+                        ++dataItr;
+                        continue;
+                    }
+                } else {
+                    // NOTE: if the data was unsuccessful in producing an entity in the past
+                    // we will try again just in case something changed (unlikely).
+                    // Unfortunately constantly trying to build the entity for this data costs
+                    // CPU cycles that we'd rather not spend.
+                    // TODO? put a maximum number of tries on this?
+                }
+            } else {
+                // remember this hash for the future
+                stateItr = _avatarEntityDataHashes.insert(entityID, AvatarEntityDataHash(newHash));
+            }
+            ++dataItr;
+
             // see EntityEditPacketSender::queueEditEntityMessage for the other end of this.  unpack properties
             // and either add or update the entity.
-            QByteArray jsonByteArray = avatarEntities.value(entityID);
-            QJsonDocument jsonProperties = QJsonDocument::fromBinaryData(jsonByteArray);
+            QJsonDocument jsonProperties = QJsonDocument::fromBinaryData(data);
             if (!jsonProperties.isObject()) {
-                qCDebug(interfaceapp) << "got bad avatarEntity json" << QString(jsonByteArray.toHex());
+                qCDebug(interfaceapp) << "got bad avatarEntity json" << QString(data.toHex());
                 continue;
             }
 
@@ -265,18 +292,22 @@ void Avatar::updateAvatarEntities() {
                 properties.setScript(noScript);
             }
 
+            // try to build the entity
             EntityItemPointer entity = entityTree->findEntityByEntityItemID(EntityItemID(entityID));
-
-            // NOTE: we don't bother checking whether updateEntity or addEntity succeed here:
-            // if not we'll try again next time the data is changed.  This is to close a performance DOS vector
-            // whereby invalid entity data is given to the AvatarMixer and constant retries kill performance.
+            bool success = true;
             if (entity) {
                 if (entityTree->updateEntity(entityID, properties)) {
                     entity->updateLastEditedFromRemote();
+                } else {
+                    success = false;
                 }
             } else {
                 entity = entityTree->addEntity(entityID, properties);
+                if (!entity) {
+                    success = false;
+                }
             }
+            stateItr.value().success = success;
         }
 
         AvatarEntityIDs recentlyDettachedAvatarEntities = getAndClearRecentlyDetachedIDs();
@@ -289,6 +320,14 @@ void Avatar::updateAvatarEntities() {
                     }
                 }
             });
+
+            // remove stale data hashes
+            foreach (auto entityID, recentlyDettachedAvatarEntities) {
+                MapOfAvatarEntityDataHashes::iterator stateItr = _avatarEntityDataHashes.find(entityID);
+                if (stateItr != _avatarEntityDataHashes.end()) {
+                    _avatarEntityDataHashes.erase(stateItr);
+                }
+            }
         }
     });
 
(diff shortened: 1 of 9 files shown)


interface/src/avatar/Avatar.cpp
```diff
@@ -295,9 +292,7 @@ void Avatar::updateAvatarEntities() {
         }
     });
 
-    if (success) {
-        setAvatarEntityDataChanged(false);
-    }
+    setAvatarEntityDataChanged(false);
```
AndrewMeadows: These changes in `Avatar::updateAvatarEntities()` are the actual **fix** for the bug.  The problem was that the `avatarEntityData` sent through the AvatarMixer was invalid such that at least one of the entities was never created, which caused these "avatar entities data" to be unpacked and applied every frame --> the interface would run out of time to update the render geometry of other avatars.

interface/src/avatar/Avatar.cpp
```diff
@@ -364,6 +359,9 @@ void Avatar::simulate(float deltaTime, bool inView) {
         measureMotionDerivatives(deltaTime);
         simulateAttachments(deltaTime);
         updatePalms();
+    }
+    {
+        PROFILE_RANGE(simulation, "entities");
         updateAvatarEntities();
```
AndrewMeadows: The offending contex has its own trace data point so we can troubleshoot this easier in the future.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -376,7 +376,9 @@ void MyAvatar::update(float deltaTime) {
         Q_ARG(glm::vec3, (getPosition() - halfBoundingBoxDimensions)),
         Q_ARG(glm::vec3, (halfBoundingBoxDimensions*2.0f)));
 
-    if (_avatarEntityDataLocallyEdited) {
+    uint64_t now = usecTimestampNow();
+    if (now > _identityPacketExpiry || _avatarEntityDataLocallyEdited) {
+        _identityPacketExpiry = now + AVATAR_IDENTITY_PACKET_SEND_INTERVAL_MSECS;
         sendIdentityPacket();
     }
```
AndrewMeadows: Not related to the bug per-se but something I noticed: we would sent the **AvatarIdentity** packet on a timer and then would also send it when things changed which would not reset the timer.  We now pretty much only ever send it in `MyAvatar::update()` and we compute the next expiry when we do.

libraries/avatars/src/AvatarData.cpp
```diff
@@ -2245,17 +2243,15 @@ void AvatarData::setAvatarEntityData(const AvatarEntityMap& avatarEntityData) {
...
-            foreach (auto entityID, previousAvatarEntityIDs) {
-                if (!_avatarEntityData.contains(entityID)) {
-                    _avatarEntityDetached.insert(entityID);
-                }
+        foreach (auto entityID, previousAvatarEntityIDs) {
+            if (!_avatarEntityData.contains(entityID)) {
+                _avatarEntityDetached.insert(entityID);
             }
         }
     });
```
AndrewMeadows: By making `AvatarData::setAvatarEntityData()` private we can dispense with most of the sanity checking logic at its beginning.  Also, this method should not be exposed to JS anyway.

libraries/entities/src/EntityEditPacketSender.cpp
```diff
@@ -93,7 +81,8 @@ void EntityEditPacketSender::queueEditEntityMessage(PacketType type,
         return; // bail early
     }
 
-    if (properties.getClientOnly()) {
+    if (properties.getClientOnly() && properties.getOwningAvatarID() == _myAvatar->getID()) {
+        // this is an avatar-based entity --> update our avatar-data rather than sending to the entity-server
         queueEditAvatarEntityMessage(type, entityTree, entityItemID, properties);
         return;
     }
```
AndrewMeadows: Similarly, by making `EntityEditPacketSender::queueEditAvatarEntityMessage()` private we can dispense with some sanity checking logic.

hifi-gustavo:
Build console output for commit: 4276d0a2769b88fc95c7ed6b83e3ac338cfe803d available at: https://gist.github.com/a1b6ab5a11a14357d214c96379d1024b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9642-4276d0a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9642-4276d0a.dmg

AndrewMeadows:
The problem here was: invalid "avatar entity data" was being passed through the AvatarMixer.  At least one of the entities could not be created on recipient interface clients and they would constantly retry every frame, which was CPU expensive.

The context where the CPU cycles were being spent was time-boxed to try to maintain high frame rates when trying to render many avatars --> the time would all get used up for processing the avatar entities and the render geometry of other avatars would never get updated.

To test:

(1) You need three avatars to test: UserA, UserB, UserC.

(2) UserA must hack their Interface.json settings file to send bogus avatarEntityData through the AvatarMixer like so:

(a) Find the local file. On windows it would be something like:
`C:\Users\YOUR_NAME\AppData\Roaming\High Fidelity - dev\Interface.json`
(b) Edit the Interface.json file with a text editor and remove any /Avatar/avatarEntityData lines from it.
(c) In their place paste in the text in the attached file which adds several entity items that are not valid
(d) Save the Interface.json file.

(3) Login all three users with UserA standing between UserB and UserC.

(4) With a bugged interface client UserB and UserC will not be able to see each other's avatars move or animate, but they can see changes to UserA.  UserA can see changes in UserB and UserC.

(5) With this PR all users should be able to see the other two change position and animate.

AndrewMeadows:
It has been pointed out that these changes conflict with the efforts in #9611 which was trying to expose the AvatarEntityData to JS as a valid feature.  Hrm... what to do?

AndrewMeadows:
Don't merge this yet.  After thinking it over I've decided that we should go ahead and expose the setAvatarEntityData() to JS but better protect other interfaces from bad data.  In particular the Interface should:

(1) flag data entries as bad after it fails to produce a viable entity so the Interface doesn't constantly try to parse it.
(2) recognize good data entries that have already been processed in earlier IdentityPackets.
(3) enforce a maximum number of AvatarEntityData entries (maybe a bytesize limit?): this feature cannot remain unbounded.

libraries/avatars/src/AvatarData.cpp
```diff
@@ -2202,14 +2200,24 @@ void AvatarData::setAttachmentsVariant(const QVariantList& variant) {
...
+        if (itr == _avatarEntityData.end()) {
+            if (_avatarEntityData.size() < MAX_NUM_AVATAR_ENTITIES) {
+                _avatarEntityData.insert(entityID, entityData);
+                _avatarEntityDataLocallyEdited = true;
+            }
+        } else {
+            itr.value() = entityData;
+            _avatarEntityDataLocallyEdited = true;
+        }
     });
```
AndrewMeadows: We now cap the total number of "avatar entities".

interface/src/avatar/Avatar.cpp
```diff
@@ -228,17 +228,43 @@ void Avatar::updateAvatarEntities() {
...
+                    // we will try again just in case something changed (unlikely).
+                    // Unfortunately constantly trying to build the entity for this data costs
+                    // CPU cycles that we'd rather not spend.
+                    // TODO? put a maximum number of tries on this?
+                }
+            } else {
+                // remember this hash for the future
+                stateItr = _avatarEntityDataHashes.insert(entityID, AvatarEntityDataHash(newHash));
+            }
+            ++dataItr;
```
AndrewMeadows: We now remember each "avatar entity" entry (by its hash) and don't bother parsing entries that have already been handled successfully.  This saves CPU cycles that were being consumed every AvatarIdentityPacket when someone was suing the "avatar entity" feature.

hifi-gustavo:
Build console output for commit: 66a315cb9cfb38efe906c0f014c98793887e863b available at: https://gist.github.com/a1b6ab5a11a14357d214c96379d1024b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9642-66a315c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9642-66a315c.dmg

nbq:
Testing with @Nex-Pro and @claudioshg

nbq:
Tester: Coal

* Using Piper and Claudioshg, we each tested this build.
* Coal being UserA hacked up the interface file (included below for reference) to included invalid entries from the pastebin provided.
* Saved Interface.json and logged back in and stood between the other two testers.
* We each tried logging in and out and closing our interface and coming back in to see if we triggered any freezes - it did not.
* We were all able to see each others fluid movements and no lock ups so this is a:

**PASS**

PR #9663, opened 2017-02-13
Don't allow a ban of any node on same machine as domain-server

davidkelly:
If you do, then all the assignment clients (if they run on same box as the domain-server) are also kicked, leaving the domain basically dead until someone fixes the permissions in the domain server setting page.

Ideally, we would have added a message (say, KickResponse) which gives an indication of success/failure and why, etc...  But that is really part of an additional feature, which allows for un-banning and so on.  

This only addresses this one issue where we ban something at 127.0.0.1 and the whole domain goes down.  Look for more fancy features building on this later!

### DIFF
--- a/domain-server/src/DomainServerSettingsManager.cpp
+++ b/domain-server/src/DomainServerSettingsManager.cpp
@@ -667,7 +667,17 @@ void DomainServerSettingsManager::processNodeKickRequestPacket(QSharedPointer<Re
                     auto& kickAddress = matchingNode->getActiveSocket()
                         ? matchingNode->getActiveSocket()->getAddress()
                         : matchingNode->getPublicSocket().getAddress();
-
+                    
+                    // probably isLoopback covers it, as whenever I try to ban an agent on same machine as the domain-server
+                    // it is always 127.0.0.1, but looking at the public and local addresses just to be sure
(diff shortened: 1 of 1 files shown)


davidkelly:
Test Plan

Startup in your sandbox.  Open the PAL (People), and ban the one agent you see in there (anonymous_1).  In master, you will be kicked from the domain in about 10 seconds.  In the domain server settings, you will see 127.0.0.1 has been banned (along with corresponding entries in the mac address and machine fingerprint sections.  

You do this in this PR, and nothing happens.  Later as part of other work, we will let you know we didn't really ban anyone, but this is enough to keep the above from happening.  You will see a warning in the domain server logs, however.

hifi-gustavo:
Build console output for commit: 45cb11f38a92c68b1c5eb1d7eb7e90c065bce95f available at: https://gist.github.com/e260b58df0d01df458bc23164725b884<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9663-45cb11f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9663-45cb11f.dmg

howard-stearns:
https://highfidelity.fogbugz.com/f/cases/2789/

howard-stearns:
test windows and mac

PR #9665, opened 2017-02-13
Fix entity scripts not loading in certain cases

huffman:
This fixes a bug where an entity script would be loaded but immediately
unloaded. This happens when the script property of the entity is changed shortly
after it is added.

Flow of events:

  Entity added
  => Entity added event emitted
  Entity edited
  => Script changing event emitted
  <= Entity added event received
  Script loaded
  <= Script changing event received
  Script stopped
  Tries to load script again, but because the script hasn't changed
    since it was last loaded, it is not loaded again.

The change here is to modify the behavior when receiving a script
changing event. Instead of always unloading and then conditionally
loading the script, it will either do both (unload + load) or neither.

This also fixes a bug where if you removed a script and then re-added the
original, the script would not load.


hifi-gustavo:
Build console output for commit: f1714e5715840cede1403ea3f299e8ffe2a28f88 available at: https://gist.github.com/82b966df697845507907c3ef55ac35cc<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9665-f1714e5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9665-f1714e5.dmg

PR #9669, opened 2017-02-14
inform gatekeeper immediately when kicking someone

davidkelly:
We seem to have not informed the gatekeeper that we'd banned someone, so it waited until it noticed (about 10s or so), and then kicked them.  Now, we tell gatekeeper right  away, and they are gone.  Seems like it takes about 2 seconds and they are gone.

### DIFF
--- a/domain-server/src/DomainServerSettingsManager.cpp
+++ b/domain-server/src/DomainServerSettingsManager.cpp
@@ -29,7 +29,7 @@
 #include <NLPacketList.h>
 #include <NumericalConstants.h>
 #include <SettingHandle.h>
-
+#include <AvatarData.h> //for KillAvatarReason
 #include "DomainServerNodeData.h"
 
 const QString SETTINGS_DESCRIPTION_RELATIVE_PATH = "/resources/describe-settings.json";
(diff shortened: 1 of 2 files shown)


davidkelly:
Test Plan:

Have a second client come into your sandbox.  Ban them multiple times.  Notice that it is much faster than it used to be.  For me, it was 10s or so in master, and is about 2s or so now.

davidkelly:
https://highfidelity.fogbugz.com/f/cases/2656

hifi-gustavo:
Build console output for commit: 3b344ca01a3e12e17282b2bc1787f742349314d0 available at: https://gist.github.com/2a753d9aab35596c61ee498963ad0597<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9669-3b344ca.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9669-3b344ca.dmg

domain-server/src/DomainServerSettingsManager.cpp
```diff
@@ -711,16 +711,26 @@ void DomainServerSettingsManager::processNodeKickRequestPacket(QSharedPointer<Re
                         }
                     }
                 }
+                // if we are here, then we kicked them, so send the KillAvatar message to everyone
+                auto packet = NLPacket::create(PacketType::KillAvatar, NUM_BYTES_RFC4122_UUID + sizeof(KillAvatarReason), true);
+                packet->write(nodeUUID.toRfc4122());
+                packet->writePrimitive(KillAvatarReason::NoReason);
```
howard-stearns: I don't have a strong opinion about what reason to use, and I don't know that it makes much of a difference.

It's plausible that KillAvatarReason::AvatarIgnored might someday get decorated in some appropriate way. (Avatar screaming "I'm melting!" as it disappears under its hat, or some such.)

On other other hand, I suppose there could be privacy concerns over the stigma of being banned inappropriately.

Nex-Pro: You might want to reconsider using 'KillAvatarReason' why not 'RemoveAvatarReason' seems less terminal even in the terminal.

davidkelly: Yes - I pondered a new KillAvatarReason, didn't think it was necessary, as the logic in AvatarMixer currently needed nothing special to know someone was being banned.  Later, if we do the wicked witch of the west animation, we'd need that reason to propagate back to the clients, and we can add it then.

howard-stearns:
works well:
- banning a logged in user quickly kicks by username
- banning a non-logged in user quickly kicks by fingerprint.
(The second actually kicks by fingerprint AND macid AND ip addreess, but that bug is in master as well.)

PR #9681, opened 2017-02-15
sort opaque render items by priority

AndrewMeadows:
* sort opaque render items by priority
* dynamically adjust max number of opaque render items to draw based on time spent trying to draw them in previous frames


libraries/render-utils/src/RenderDeferredTask.cpp
```diff
@@ -295,7 +295,28 @@ void DrawStateSortDeferred::run(const SceneContextPointer& sceneContext, const R
...
+            if (fabsf(_budget - _targetBudget) > CLOSE_ENOUGH * _targetBudget) {
+                // move halfway toward ideal
+                _budget += 0.5f * (_targetBudget - _budget);
+            }
+
+            // clamp the number we'll actually draw
+            _numDrawn = glm::min((int)(_budget), (int)inItems.size());
+            if (_maxDrawn != -1) {
+                _numDrawn = glm::min(_numDrawn, _maxDrawn);
+            }
```
AndrewMeadows: This is the feedback logic that computes the number of items to draw next frame.

libraries/render/src/render/SortTask.cpp
```diff
@@ -41,28 +45,25 @@ struct BackToFrontSort {
...
-        auto item = scene->getItem(itemDetails.id);
-        auto bound = itemDetails.bound; // item.getBound();
-        float distance = args->getViewFrustum().distanceToCamera(bound.calcCenter());
-
-        itemBoundSorts.emplace_back(ItemBoundSort(distance, distance, distance, itemDetails.id, bound));
+    glm::vec3 eye = args->getViewFrustum().getPosition();
+    for (const auto& itemDetails : inItems) {
+        // use "distance squared" for speed: sorts the same as 'distance' but avoids a sqrt
+        float distance2 = glm::distance2(eye, itemDetails.bound.calcCenter());
+        itemBoundSorts.emplace_back(ItemBoundSort(distance2, distance2, distance2, itemDetails.id, itemDetails.bound));
```
AndrewMeadows: Faster math for the naive "distance to center of item" logic, which we still use for transparent things (sorting by reverse order of course).

libraries/render/src/render/SortTask.cpp
```diff
@@ -118,3 +119,43 @@ void DepthSortShapes::run(const SceneContextPointer& sceneContext, const RenderC
...
+        sortedItems.emplace(ItemPriority(itemDetails.id, itemDetails.bound, priority));
+    }
+
+    // transfer queue to list
+    while (!sortedItems.empty()) {
+        const ItemPriority& item = sortedItems.top();
+        outItems.emplace_back(ItemBound(item.id, item.bound));
+        sortedItems.pop();
+    }
+}
```
AndrewMeadows: The priority sort uses "apparent size" multiplied by a exponential distance envelope which reduces the priority of far away objects at equal sizes.

hifi-gustavo:
Build console output for commit: eee540dac1884c052ec4d7167f2a1136868e4b08 available at: https://gist.github.com/addbe20c96fef554778a63cbce0c3ab7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9681-eee540d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9681-eee540d.dmg

hifi-gustavo:
Build console output for commit: 50c77a0fa520b775c858617642b80e175de337c7 available at: https://gist.github.com/addbe20c96fef554778a63cbce0c3ab7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9681-50c77a0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9681-50c77a0.dmg

AndrewMeadows:
This PR makes it possible to set a target time limit to spend in the `DrawOpaqueDeferred` context.  The number of items rendered will be reduced in an attempt to get close to the target time (has a slop of +/- 12%).  There is a default target currently hard coded to 8msec.  When items are dropped trying to hit the time budget the farther, smaller objects will tend to vanish first.

To test:

(1) Login to a domain with plenty of objects/avatar visible and spread out for larger distances.  dev-playa works well, but it helps if you fly up to a height where you can see stuff underneath you and stuff far away.

(2) Copy this script to disk and run it:
[code]var qml = Script.resolvePath('culling.qml');
var window = new OverlayWindow({
    title: 'Render Draws',
    source: qml,
    width: 300, 
    height: 200
});
window.setPosition(200, 50);
window.closed.connect(function() { Script.stop(); });[/code]
You should get some new UI that allows you to control some render debug tools.  In particular you can change the target max time spent in the `DrawOpaqueDeferred` context: the UI slider is titled "Draw Opaque [ms]".  Also you can change the max number of items to render using the "Opaque" slider.

(3) Look over the objects that are visible in the scene.  Some of the foreground stuff is big and small.  Some of the far stuff is big but looks small because it is so far away.  If you had to sacrifice rendering some of those objects in order to maintain frame rate... which would you give up first?

The logic in this PR makes that decision by assigning a priority to each item based on its size and distance from the camera.   As pressure is applied to reduce the number of items it will drop those of lower priority -- they will vanish from the rendered scene.

(4) Play with the "Opaque" slide.  It is initialized to `-1` which means "disabled".  Drag it up to the max and then slowly make the number smaller.  As you reduce the slider value you should see objects in the distance disappear first, but maybe also eventually smaller nearby objects.  Low "priority" objects are culled first.  Do you think they are sorted correctly?

When you're done set the "" slider to `-1` again.

(5) Now play with the "Draw Opaque [ms]" slider.  It starts at 8msec but dial it down to about 3 msec to see a reduction in the number of items rendered.  The order of elimination will be exactly the same as in (4) but where the count stops is dynamically determined.

(6) Keep the "Draw Opaque [ms]" slider at some low value so that some items are culled and wait several seconds.  There is "noise" in the time measurement and the system may adjust to different counts.  This is OK but what we don't want to see is objects flashing: on, off, on, off, etc.  The state of the scene should be relatively stable, but may adjust a little bit every few seconds.

(7) With the "Draw Opaque [ms]" slider set low: spin the view around so that very few objects are in your view, and then spin around and look at many objects.  Does the scene adjust quickly toward a final count?

Note: with the "Draw Opaque [ms]" time limit set low: some distant/small objects may flash on/off when other closer/larger objects that were out of view move into view, or as the camera view is slowly swung around to look at different densities of items.

cozza13:
A few vantage points that I have been going to:

hifi://dev-playa/-55.9979,34.2657,-87.9861/0,0.970123,0,-0.242612
hifi://dev-cell/13501,13501,13501/0,0.39221,0,0.919876
hifi://dev-welcome/147.452,31.3668,-263.574/0,0.935441,0,0.353483
hifi://dev-chris2.highfidelity.io/100.842,0.627618,25.6073/0,0.994947,0,0.100406
hifi://dev-chris.highfidelity.io/-16.6602,4.87944,-13.9103/0,0.909143,0,-0.416485

AndrewMeadows:
This project needs work.  I'm closing this PR and will open a new one later.

PR #9683, opened 2017-02-15
Clear caches on domain switch

jherico:
Currently, the ModelCache allows up to 100 MB of models that are not currently in use to remain in memory, consuming CPU and GPU resources.  However, the size calculation does not cover the textures attached to these models, which in fact will likely make up the vast majority of the memory on both the CPU and GPU.  This PR attempts to mitigate the situation by clearing out caches when switching between domains.

## Testing

Here is an example of dev-welcome after coming from [dev-chris2.highfidelity.io](hifi://dev-chris2.highfidelity.io/102.097,0.710212,26.7103/0,0.992718,0,0.120464).  Note the texture memory usage and the high count of textures (464).
image

Here is an example of dev-welcome after coming from a low texture domain (dreaming).  Note again the texture count and memory usage.  
image

This PR should produce something more like the latter result regardless of where the user is coming from.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -5152,6 +5152,7 @@ void Application::updateWindowTitle() const {
 #endif
     _window->setWindowTitle(title);
 }
+
 void Application::clearDomainOctreeDetails() {
 
     // if we're about to quit, we really don't need to do any of these things...
@@ -5181,6 +5182,12 @@ void Application::clearDomainOctreeDetails() {
     skyStage->setBackgroundMode(model::SunSkyStage::SKY_DEFAULT);
 
     _recentlyClearedDomain = true;
+
+    DependencyManager::get<AvatarManager>()->clearOtherAvatars();
+    DependencyManager::get<AnimationCache>()->clearUnusedResources();
+    DependencyManager::get<ModelCache>()->clearUnusedResources();
+    DependencyManager::get<SoundCache>()->clearUnusedResources();
+    DependencyManager::get<TextureCache>()->clearUnusedResources();
 }
 
 void Application::domainChanged(const QString& domainHostname) {
--- a/libraries/networking/src/ResourceCache.cpp
+++ b/libraries/networking/src/ResourceCache.cpp
@@ -221,7 +221,7 @@ ResourceCache::ResourceCache(QObject* parent) : QObject(parent) {
 }
 
 ResourceCache::~ResourceCache() {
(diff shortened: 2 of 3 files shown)


zzmp:
I thought caching between domains was an important feature.

hifi-gustavo:
Build console output for commit: 127a014b4694fdbd8dac2d16799249b82d49a203 available at: https://gist.github.com/6f6f7f12a3536437071154dc0f81383b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9683-127a014.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9683-127a014.dmg

Nex-Pro:
Is this a smart way to do this ?

Nex-Pro:
So if a whole bunch of friends  jump to another domain they have to reload all their friends avatars again ?

OmegaHeron:

I believe I understand the intent here - you have 100MB of model RAM cache, it could be holding a model that has a gigantic texture footprint that's now moved out of direct context, but remains in small RAM cache thus keeping its high texture RAM footprint locked.  So by clearing model cache you can (potentially) free up a large amount of texture space.  Is that, kind of right?  That said - in that context this actually makes a lot of sense to me.  Anyhow - end editorial mode.

General smoke test = pass.  Noticed no real change in how each domain loaded up on initial and subsequent same session visits - everything seems to work.  What I didn't see with master was the large VM memory stat use after visiting the 100 AV test domain then going to a light scene like dev-welcome.  That may be due to my GPU VRAM having hit zero in 100 AV test domain and remaining there as I moved to dev-welcome then anywhere else.

This PR;
Starting at heron with texture count 69 VM = 385 then to
dev-playa with texture count = 265 VM = 1186 then to
dev-chris with texture count = 389 VM = 4112 (my GPU free mem is now indicating 0 free) then to
dev-welcome with texture count = 157 VM = 497 then to (gpu free still showing 0) then to
heron with texture count = 71 VM = 385 (gpu free still 0).

Master (6062);

Starting at heron with texture count 69 VM = 385 then to
dev-playa with texture count = 264 VM = 1180 then to
dev-chris with texture count = 385 VM = 4172 (my GPU free mem is now indicating 0 free) then to
dev-welcome with texture count = 159 VM = 497 then to (gpu free still showing 0) then to
heron with texture count = 73 VM = 387 (gpu free still 0).

PR #9656, opened 2017-02-12
CR-1 21114 -- ScriptCache/ScriptEngine cleanup, loader states, accessible load errors

humbletim:
This PR contains several changes from the `require/module` work, separated out this way to split things up a little for code review and testing.

* Add new entity script statuses (PENDING, LOADING and UNLOADED).
* Provide specific network errors for failed scripts.
* Resolve FIXMEs related to script loading/caching; cleanup dead code.
* Refactor support for JavaScript syntax checking and exception reporting.
* Update the editor's "Server Script URL" panel to reflect the new statuses.

With @huffman's help I was able to get that last part working, which now doubles as a way to test the overall changes from the UI.

A cool thing for the user is that augmented errors show up in the tablet/editor panel for server scripts -- like when a syntax error, the problematic line number; or if a script fails to load, the underlying network reason.

### Testing:

* Launch the PR build of Interface + Sandbox.
* Perform a general smoke test to do with loading Client scripts
 *(app behavior should not have changed -- but Client scripts are now fetched using a different method so just want to be sure)*
* Rez a few Entities to work with.
* Try attaching a representative sample of script URLs (in the `Script URL` and `Server Script URL` fields, but not necessarily same time).
* For Server scripts -- detailed error messages should come back in the textarea:
image
* For regular entity scripts -- similar error messages should get reported to the debug log:
`02/11 20:41:08] [DEBUG] [hifi.scriptengine] "Bad syntax ([SyntaxError] Illegal character in https://httpbin.org/image/png:1(1))" "https://httpbin.org/image/png"`

test urls:
* https://httpbin.org/delay/10 => ~10 seconds of *Loading* => `Could not find constructor`
* http://httpbin.org/404 => `Failed to load script (Not Found)`
* [ftp://0.0.0.0:0000](ftp://0.0.0.0:0000) => `Failed to load script (AccessDenied)`
* [http:////:ptth](http:////:ptth) => `Failed to load script (ServerUnavailable)`
* https://httpbin.org/image/png => `...[SyntaxError] Illegal character in https://httpbin.org/image/png:1(1)`
* [/scripts/developer/tests/basicEntityTest/myEntityScript.js](https://raw.githubusercontent.com/highfidelity/hifi/master/scripts/developer/tests/basicEntityTest/myEntityScript.js) => *Running* (no error)
*Clear-out a populated Server Script URL* => *Unloaded* or *Not Running* (no error)

PR #9685, opened 2017-02-16
CR-2 21114 -- BaseScriptEngine.cpp, order-of-operations fixes, reworked JS Exceptions

humbletim:
Here is a rundown of the changes included with this PR -- feedback welcomed.

A key feature here is unlocking the standard JS exception mechanisms -- which are needed for consistent `require/module` support.

#### Extended support for JS syntax errors, exception tracking, stack traces, JS Error objects
* JS exceptions can now be used across nested C++/JS stack frames -- which was the main thing making it difficult to leverage them before.
Before a script had no (direct) way for example to know that an include had failed -- now exceptions could flow to the surface if wanted; making this sort of thing possible:

```javascript
try {
  include('preferred.js');
} catch(e) {
  include('fallback.js');
}
```

* A new helper class named `BaseScriptEngine::ExceptionEmitter` takes care of monitoring/escalating/clearing JS exceptions.  The idea was to make it as simple as using `QMutexLocker` is for locking -- declare an instance of the class inside of a C++ scope, and now that scope is protected from leaking/losing exceptions.
*  Here is the typical path an exception would take getting into the log files:
	* Code that involves JS errors now generates and ->throw's an actual JS `Error`.
	* `ExceptionEmitter` makes sure those get escalated where/when needed.
	*  Uncaught exceptions end up at `BaseScriptEngine::unhandledException(Error)`.  
	* ScriptEngine connects to that and responds by logging the errors as `[UncaughtException]` messages.

#### Entity bootstrapping process
* Errors while booting an Entity are now thrown as JS exceptions, which get logged by the same means as above.
* JS Errors can hold metadata -- for example, I was having trouble working backwards from Entity script errors to a specific point in the boot process, so went through and decorated the main errors to include better clues.
* Currently this extra detail gets printed with the exception.  This might make the log more verbose than people want by default so added a manual override Setting `com.highfidelity.experimental.enableExtendedJSExceptions`:

Off (false):
```
 [UncaughtExceptions] Error: Can't find variable: foobar in about:EmbeddedEntityScript:16\n[Backtrace]...
```

On (true):
```text
[UncaughtException (construct {1eb5d3fa-23b1-411c-af83-163af7220e3f})] Error: Can't find variable: foobar in about:EmbeddedEntityScript:16\n[Backtrace]...
```
* The loading process now takes into account multiple Entities trying to stampede-load from the same script URL.
  * Instead of letting them stall in ScriptCache, ScriptEngine now detects and queues the 2nd+ member of such sets locally.  When the first Entity reaches a RUNNING state -- the remaining are allowed through.  Or if it doesn't they are canceled as a set.  
  * note: stampede loads are very common right now -- for example, anytime someone visits Welcome (the 4+ portals share one script URL).  Or like when someone rezzes multiple copies of a scripted Entity; doesn't matter how slow they do it -- if they relog, the Entities will likely stampede-load; if anybody else shows up, same.

#### Misc

* Migrated overlapping logic for path verification from include() into PathUtils.
* Internal ScriptEngine names are now prefixed `about:`, which avoids `QUrl` from interpreting as relative, local file paths.
* `debugPrint` now invokes JS `Script.print` using the QScriptEngine API.
* Added `BaseScriptEngine::evaluateInClosure(locals, program)` (which `require` needs to created `module` scopes, but might also be useful for other things).
* Added `BaseScriptEngine::newLambdaFunction` which lets C++ lambdas become callable `QScriptValue` functions even when capturing things like `this`.  Only meant for use within the scripting engine itself.  This allows for example using `setInterval` and `setTimeout` to schedule lightweight C++ timers that automatically track/stop with the scripting engine.

### DIFF
--- a/libraries/script-engine/src/BaseScriptEngine.cpp
+++ b/libraries/script-engine/src/BaseScriptEngine.cpp
@@ -210,10 +210,9 @@ QScriptValue BaseScriptEngine::evaluateInClosure(const QScriptValue& closure, co
     qCDebug(scriptengine) << QString("[%1] evaluateInClosure %2").arg(isEvaluating()).arg(shortName);
 #endif
     {
-        ExceptionEmitter tryCatch(this, __FUNCTION__);
         result = BaseScriptEngine::evaluate(program);
-        if (tryCatch.hasPending()) {
-            auto err = tryCatch.pending();
+        if (hasUncaughtException()) {
+            auto err = cloneUncaughtException(__FUNCTION__);
 #ifdef DEBUG_JS_EXCEPTIONS
             qCWarning(scriptengine) << __FUNCTION__ << "---------- hasCaught:" << err.toString() << result.toString();
             err.setProperty("_result", result);
@@ -265,37 +264,6 @@ Lambda::Lambda(QScriptEngine *engine, std::function<QScriptValue(QScriptContext 
 QScriptValue Lambda::call() {
     return operation(engine->currentContext(), engine);
 }
-
-// BaseScriptEngine::ExceptionEmitter
-
-void BaseScriptEngine::_emitUnhandledException(const QScriptValue& exception) {
-    emit unhandledException(exception);
-}
-BaseScriptEngine::ExceptionEmitter::ExceptionEmitter(BaseScriptEngine* engine, const QString& debugName)
-    : _engine(engine), _debugName(debugName) {
-}
-bool BaseScriptEngine::ExceptionEmitter::hasPending() {
-    return _engine->hasUncaughtException();
-}
-QScriptValue BaseScriptEngine::ExceptionEmitter::pending() {
-    return _engine->cloneUncaughtException(_debugName);
-}
-QScriptValue BaseScriptEngine::ExceptionEmitter::consume() {
-    if (hasPending()) {
-        _consumedException = pending();
-        _engine->clearExceptions();
-    }
-    return _consumedException;
-}
-bool BaseScriptEngine::ExceptionEmitter::wouldEmit() {
-    return !_engine->isEvaluating() && _engine->hasUncaughtException();
-}
-BaseScriptEngine::ExceptionEmitter::~ExceptionEmitter() {
-    if (wouldEmit()) {
-        _engine->_emitUnhandledException(consume());
-    }
-}
-
 
 #ifdef DEBUG_JS
 void BaseScriptEngine::_debugDump(const QString& header, const QScriptValue& object, const QString& footer) {
--- a/libraries/script-engine/src/BaseScriptEngine.cpp
+++ b/libraries/script-engine/src/BaseScriptEngine.cpp
@@ -1,0 +1,321 @@
+//
+//  BaseScriptEngine.cpp
+//  libraries/script-engine/src
+//
+//  Created by Timothy Dedischew on 02/01/17.
+//  Copyright 2017 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include "BaseScriptEngine.h"
+
+#include <QtCore/QString>
+#include <QtCore/QThread>
+#include <QtCore/QUrl>
+#include <QtScript/QScriptValue>
+#include <QtScript/QScriptValueIterator>
+#include <QtScript/QScriptContextInfo>
+
+#include "ScriptEngineLogging.h"
+#include "Profile.h"
+
+const QString BaseScriptEngine::_SETTINGS_ENABLE_EXTENDED_EXCEPTIONS {
+    "com.highfidelity.experimental.enableExtendedJSExceptions"
+};
+
+const QString BaseScriptEngine::SCRIPT_EXCEPTION_FORMAT { "[%0] %1 in %2:%3" };
+const QString BaseScriptEngine::SCRIPT_BACKTRACE_SEP { "\n    " };
+
+// engine-aware JS Error copier and factory
+QScriptValue BaseScriptEngine::makeError(const QScriptValue& _other, const QString& type) {
+    auto other = _other;
+    if (other.isString()) {
+        other = newObject();
+        other.setProperty("message", _other.toString());
+    }
+    auto proto = globalObject().property(type);
+    if (!proto.isFunction()) {
+        proto = globalObject().property(other.prototype().property("constructor").property("name").toString());
+    }
+    if (!proto.isFunction()) {
+#ifdef DEBUG_JS_EXCEPTIONS
+        qCDebug(scriptengine) << "BaseScriptEngine::makeError -- couldn't find constructor for" << type << " -- using Error instead";
+#endif
+        proto = globalObject().property("Error");
+    }
+    if (other.engine() != this) {
+        // JS Objects are parented to a specific script engine instance
+        // -- this effectively ~clones it locally by routing through a QVariant and back
+        other = toScriptValue(other.toVariant());
+    }
+    // ~ var err = new Error(other.message)
+    auto err = proto.construct(QScriptValueList({other.property("message")}));
+
+    // transfer over any existing properties
+    QScriptValueIterator it(other);
+    while (it.hasNext()) {
+        it.next();
+        err.setProperty(it.name(), it.value());
+    }
+    return err;
+}
+
+// check syntax and when there are issues returns an actual "SyntaxError" with the details
+QScriptValue BaseScriptEngine::lintScript(const QString& sourceCode, const QString& fileName, const int lineNumber) {
+    const auto syntaxCheck = checkSyntax(sourceCode);
+    if (syntaxCheck.state() != QScriptSyntaxCheckResult::Valid) {
+        auto err = globalObject().property("SyntaxError")
+            .construct(QScriptValueList({syntaxCheck.errorMessage()}));
+        err.setProperty("fileName", fileName);
+        err.setProperty("lineNumber", syntaxCheck.errorLineNumber());
+        err.setProperty("expressionBeginOffset", syntaxCheck.errorColumnNumber());
+        err.setProperty("stack", currentContext()->backtrace().join(SCRIPT_BACKTRACE_SEP));
+        {
+            const auto error = syntaxCheck.errorMessage();
+            const auto line = QString::number(syntaxCheck.errorLineNumber());
+            const auto column = QString::number(syntaxCheck.errorColumnNumber());
+            // for compatibility with legacy reporting
+            const auto message = QString("[SyntaxError] %1 in %2:%3(%4)").arg(error, fileName, line, column);
+            err.setProperty("formatted", message);
+        }
+        return err;
+    }
+    return undefinedValue();
+}
+
+// this pulls from the best available information to create a detailed snapshot of the current exception
+QScriptValue BaseScriptEngine::cloneUncaughtException(const QString& extraDetail) {
+    if (!hasUncaughtException()) {
+        return QScriptValue();
+    }
+    auto exception = uncaughtException();
+    // ensure the error object is engine-local
+    auto err = makeError(exception);
+
+    // not sure why Qt does't offer uncaughtExceptionFileName -- but the line number
+    // on its own is often useless/wrong if arbitrarily married to a filename.
+    // when the error object already has this info, it seems to be the most reliable
+    auto fileName = exception.property("fileName").toString();
+    auto lineNumber = exception.property("lineNumber").toInt32();
+
+    // the backtrace, on the other hand, seems most reliable taken from uncaughtExceptionBacktrace
+    auto backtrace = uncaughtExceptionBacktrace();
+    if (backtrace.isEmpty()) {
+        // fallback to the error object
+        backtrace = exception.property("stack").toString().split(SCRIPT_BACKTRACE_SEP);
+    }
+    // the ad hoc "detail" property can be used now to embed additional clues
+    auto detail = exception.property("detail").toString();
+    if (detail.isEmpty()) {
+        detail = extraDetail;
+    } else if (!extraDetail.isEmpty()) {
+        detail += "(" + extraDetail + ")";
+    }
+    if (lineNumber <= 0) {
+        lineNumber = uncaughtExceptionLineNumber();
+    }
+    if (fileName.isEmpty()) {
+        // climb the stack frames looking for something useful to display
+        for(auto ctx = currentContext(); ctx && fileName.isEmpty(); ctx = ctx->parentContext()) {
+            QScriptContextInfo info { ctx };
+            if (!info.fileName().isEmpty()) {
+                // take fileName:lineNumber as a pair
+                fileName = info.fileName();
+                lineNumber = info.lineNumber();
+                if (backtrace.isEmpty()) {
+                    backtrace = ctx->backtrace();
+                }
+                break;
+            }
+        }
+    }
+    err.setProperty("fileName", fileName);
+    err.setProperty("lineNumber", lineNumber );
+    err.setProperty("detail", detail);
+    err.setProperty("stack", backtrace.join(SCRIPT_BACKTRACE_SEP));
+
+#ifdef DEBUG_JS_EXCEPTIONS
+    err.setProperty("_fileName", exception.property("fileName").toString());
+    err.setProperty("_stack", uncaughtExceptionBacktrace().join(SCRIPT_BACKTRACE_SEP));
+    err.setProperty("_lineNumber", uncaughtExceptionLineNumber());
+#endif
+    return err;
+}
+
+QString BaseScriptEngine::formatException(const QScriptValue& exception) {
+    QString note { "UncaughtException" };
+    QString result;
+
+    if (!exception.isObject()) {
+        return result;
+    }
+    const auto message = exception.toString();
+    const auto fileName = exception.property("fileName").toString();
+    const auto lineNumber = exception.property("lineNumber").toString();
+    const auto stacktrace =  exception.property("stack").toString();
+
+    if (_enableExtendedJSExceptions.get()) {
+        // This setting toggles display of the hints now being added during the loading process.
+        // Example difference:
+        //   [UncaughtExceptions] Error: Can't find variable: foobar in atp:/myentity.js\n...
+        //   [UncaughtException (construct {1eb5d3fa-23b1-411c-af83-163af7220e3f})] Error: Can't find variable: foobar in atp:/myentity.js\n...
+        if (exception.property("detail").isValid()) {
+            note += " " + exception.property("detail").toString();
+        }
+    }
+
+    result = QString(SCRIPT_EXCEPTION_FORMAT).arg(note, message, fileName, lineNumber);
+    if (!stacktrace.isEmpty()) {
+        result += QString("\n[Backtrace]%1%2").arg(SCRIPT_BACKTRACE_SEP).arg(stacktrace);
+    }
+    return result;
+}
+
+QScriptValue BaseScriptEngine::evaluateInClosure(const QScriptValue& closure, const QScriptProgram& program) {
+    PROFILE_RANGE(script, "evaluateInClosure");
+    if (QThread::currentThread() != thread()) {
+        qCCritical(scriptengine) << "*** CRITICAL *** ScriptEngine::evaluateInClosure() is meant to be called from engine thread only.";
+        // note: a recursive mutex might be needed around below code if this method ever becomes Q_INVOKABLE
+        return QScriptValue();
+    }
+
+    //_debugDump("evaluateInClosure.closure", closure);
+
+    const auto fileName = program.fileName();
+    const auto shortName = QUrl(fileName).fileName();
+
+    QScriptValue result;
+    QScriptValue oldGlobal;
+    auto global = closure.property("global");
+    if (global.isObject()) {
+#ifdef DEBUG_JS
+        qCDebug(scriptengine) << " setting global = closure.global" << shortName;
+#endif
+        oldGlobal = globalObject();
+        setGlobalObject(global);
+    }
+
+    auto ctx = pushContext();
+
+    auto thiz = closure.property("this");
+    if (thiz.isObject()) {
+#ifdef DEBUG_JS
+        qCDebug(scriptengine) << " setting this = closure.this" << shortName;
+#endif
+        ctx->setThisObject(thiz);
+    }
+
+    ctx->pushScope(closure);
+#ifdef DEBUG_JS
+    qCDebug(scriptengine) << QString("[%1] evaluateInClosure %2").arg(isEvaluating()).arg(shortName);
+#endif
+    {
+        ExceptionEmitter tryCatch(this, __FUNCTION__);
+        result = BaseScriptEngine::evaluate(program);
+        if (tryCatch.hasPending()) {
+            auto err = tryCatch.pending();
+#ifdef DEBUG_JS_EXCEPTIONS
+            qCWarning(scriptengine) << __FUNCTION__ << "---------- hasCaught:" << err.toString() << result.toString();
+            err.setProperty("_result", result);
+#endif
+            result = err;
+        }
+    }
+#ifdef DEBUG_JS
+    qCDebug(scriptengine) << QString("[%1] //evaluateInClosure %2").arg(isEvaluating()).arg(shortName);
+#endif
+    popContext();
+
+    if (oldGlobal.isValid()) {
+#ifdef DEBUG_JS
+        qCDebug(scriptengine) << " restoring global" << shortName;
+#endif
+        setGlobalObject(oldGlobal);
+    }
+
+    return result;
+}
+
+// Lambda
+
+QScriptValue BaseScriptEngine::newLambdaFunction(std::function<QScriptValue(QScriptContext *, QScriptEngine*)> operation, const QScriptValue& data, const QScriptEngine::ValueOwnership& ownership) {
+    auto lambda = new Lambda(this, operation, data);
+    auto object = newQObject(lambda, ownership);
+    auto call = object.property("call");
+    call.setPrototype(object); // ctx->callee().prototype() === Lambda QObject
+    call.setData(data);        // ctx->callee().data() will === data param
+    return call;
+}
+QString Lambda::toString() const {
+    return QString("[Lambda%1]").arg(data.isValid() ? " " + data.toString() : data.toString());
+}
+
+Lambda::~Lambda() {
+#ifdef DEBUG_JS_LAMBDA_FUNCS
+    qDebug() << "~Lambda" << "this" << this;
+#endif
+}
+
+Lambda::Lambda(QScriptEngine *engine, std::function<QScriptValue(QScriptContext *, QScriptEngine*)> operation, QScriptValue data)
+    : engine(engine), operation(operation), data(data) {
+#ifdef DEBUG_JS_LAMBDA_FUNCS
+    qDebug() << "Lambda" << data.toString();
+#endif
+}
+QScriptValue Lambda::call() {
+    return operation(engine->currentContext(), engine);
+}
+
+// BaseScriptEngine::ExceptionEmitter
+
+void BaseScriptEngine::_emitUnhandledException(const QScriptValue& exception) {
+    emit unhandledException(exception);
+}
+BaseScriptEngine::ExceptionEmitter::ExceptionEmitter(BaseScriptEngine* engine, const QString& debugName)
+    : _engine(engine), _debugName(debugName) {
+}
+bool BaseScriptEngine::ExceptionEmitter::hasPending() {
+    return _engine->hasUncaughtException();
+}
+QScriptValue BaseScriptEngine::ExceptionEmitter::pending() {
+    return _engine->cloneUncaughtException(_debugName);
+}
+QScriptValue BaseScriptEngine::ExceptionEmitter::consume() {
+    if (hasPending()) {
+        _consumedException = pending();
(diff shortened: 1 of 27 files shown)


hifi-gustavo:
Build console output for commit: b15956278c680bf94c38dbd3b8a259f0fc431d4b available at: https://gist.github.com/05e2159c339b6b39697ce3c33756b6eb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-b159562.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-b159562.dmg

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1244,35 +1245,46 @@ void ScriptEngine::stopTimer(QTimer *timer) {
...
+    } else {
+        qCDebug(scriptengine) << "stopTimer -- not in _timerFunctionMap" << timer;
     }
 }
 
 QUrl ScriptEngine::resolvePath(const QString& include) const {
     QUrl url(include);
     // first lets check to see if it's already a full URL
-    if (!url.scheme().isEmpty()) {
+    if (include.startsWith("/") || url.scheme().length() == 1) {
```
huffman: I don't understand this - what does it mean for the scheme to be length 1?

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1244,35 +1245,46 @@ void ScriptEngine::stopTimer(QTimer *timer) {
...
-        } else {
-            parentURL = QUrl(_parentURL);
+    // to the first absolute URL in the JS scope chain
+    QUrl parentURL;
+    auto ctx = currentContext();
+    do {
+        QScriptContextInfo contextInfo { ctx };
+        parentURL = QUrl(contextInfo.fileName());
+        ctx = ctx->parentContext();
+    } while (parentURL.isRelative() && ctx);
```
huffman: Do you have any examples of cases where we would have a relative path? Any relative paths should generally be expanded to an absolute path.

libraries/script-engine/src/BaseScriptEngine.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+    auto thiz = closure.property("this");
+    if (thiz.isObject()) {
+#ifdef DEBUG_JS
+        qCDebug(scriptengine) << " setting this = closure.this" << shortName;
+#endif
+        ctx->setThisObject(thiz);
+    }
+
+    ctx->pushScope(closure);
```
humbletim: Had the same thought.  One benefit of `pushClosure` is that it avoids baking JS code into the C++ code -- but if that API call ever goes away then a compatible fallback strategy is to just wrap the incoming source code in an actual `(function(...locals...) { /* source */ })(...locals...)` block (which coincidentally is how Node.js goes about creating [module closures](https://github.com/nodejs/node/blob/f8ee19761493dc2a444fc12d4753f9add637cb39/lib/internal/bootstrap_node.js#L506-L517)).

libraries/script-engine/src/BaseScriptEngine.cpp
```diff
@@ -0,0 +1,321 @@
...
+    emit unhandledException(exception);
+}
+BaseScriptEngine::ExceptionEmitter::ExceptionEmitter(BaseScriptEngine* engine, const QString& debugName)
+    : _engine(engine), _debugName(debugName) {
+}
+bool BaseScriptEngine::ExceptionEmitter::hasPending() {
+    return _engine->hasUncaughtException();
+}
+QScriptValue BaseScriptEngine::ExceptionEmitter::pending() {
+    return _engine->cloneUncaughtException(_debugName);
```
humbletim: Clones seem necessary for three main reasons:
* bound `QScriptValue`'s are not threadsafe (... and `Error` objects are bound values).
* bound `QScriptValue`'s cannot be used by another `QScriptEngine` instance.
* `QScriptEngine` tramples on its own exception states as recursive calls to `evaluate` are made.

Making an engine-local copy helps the data survive intact into the future and ensures it doesn't overlap with other threads or engine instances.  In part this is to allow the Entities preflight sandbox to throw/report exceptions consistently with the other bootstrapping errors.

On a side note -- any C++ functions returning a bound `QScriptValue` across a thread boundary are likely risking segfaults right now, unless the receiving side is specifically applying an exclusive lock over the originating engine instance.  From a thread safety standpoint, it seems especially dangerous to call bound `QScriptValue` methods like `value.toString()` or `value.property(name)` from another thread -- because internally those deep-dive into their bound engine instances.

libraries/script-engine/src/BaseScriptEngine.h
```diff
@@ -0,0 +1,86 @@
...
+#define hifi_BaseScriptEngine_h
+
+#include <functional>
+#include <QtCore/QDebug>
+#include <QtScript/QScriptEngine>
+
+#include "SettingHandle.h"
+
+// common base class for extending QScriptEngine itself
+class BaseScriptEngine : public QScriptEngine {
```
humbletim: Yes, I think so -- technically this class was already here, but only since my last PR.

It was originally imagined as a way to make the shared logic between the preflight engine and main script less [statically thissified](https://github.com/highfidelity/hifi/pull/9656/files#diff-0e211914feb9a81572284cc68e8a39a3L184) as part of sorting out exceptions.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -89,13 +93,9 @@ static QScriptValue debugPrint(QScriptContext* context, QScriptEngine* engine){
...
-    message = message.replace("\\", "\\\\")
-                     .replace("\n", "\\n")
-                     .replace("\r", "\\r")
-                     .replace("'", "\\'");
-
-    // FIXME - this approach neeeds revisiting. print() comes here, which ends up doing an evaluate?
-    engine->evaluate("Script.print('" + message + "')");
+    // FIXME - this approach neeeds revisiting. print() comes here, which ends up calling Script.print?
+    engine->globalObject().property("Script").property("print")
+        .call(engine->nullValue(), QScriptValueList({ message }));
```
humbletim: Ahh, that makes sense --  looks like global `print` is the only one of the two that logs to the main debug output; and `Script.print` at some point in the past became a one line [emit printedMessage](https://github.com/humbletim/hifi/blob/b15956278c680bf94c38dbd3b8a259f0fc431d4b/libraries/script-engine/src/ScriptEngine.cpp#L1297-L1299).

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -194,16 +157,31 @@ ScriptEngine::ScriptEngine(Context context, const QString& scriptContents, const
     DependencyManager::get<ScriptEngines>()->addScriptEngine(this);
 
     connect(this, &QScriptEngine::signalHandlerException, this, [this](const QScriptValue& exception) {
-        reportUncaughtException();
-        clearExceptions();
+        if (hasUncaughtException()) {
```
humbletim: That situation seems possibly in theory, but note that the previous handler code doesn't actually reference the incoming `exception` argument.

Do you think it's worth adding mismatch checking? (in an earlier version I had `exception.toString() != uncaughtException().toString()` detection, but it never caught anything so ended up backing it out).

Mismatched exceptions have happened in the past though -- like before the main event loop would very occasionally snarf exceptions from an evaluation that was still in progress; that case is now [specifically guarded against](https://github.com/humbletim/hifi/blob/b15956278c680bf94c38dbd3b8a259f0fc431d4b/libraries/script-engine/src/ScriptEngine.cpp#L311).

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -892,23 +881,28 @@ QScriptValue ScriptEngine::evaluate(const QString& sourceCode, const QString& fi
...
+    if (program.isNull()) {
+        // can this happen?
+        auto err = makeError("could not create QScriptProgram for " + fileName);
+        emit unhandledException(err);
+        return err;
     }
 
-    ++_evaluatesPending;
-    auto result = BaseScriptEngine::evaluate(program);
-    --_evaluatesPending;
```
humbletim: Nope (except for some transient code I introduced then backed-out along the way).  It was at one point useful to me for logging stack depth -- but in the end `QScriptEngine::isEvaluating()` proved sufficient for knowing about recursion.

libraries/script-engine/src/ScriptEngine.cpp
humbletim: ... just an extra FUD check based on this [apparent possibility](https://github.com/highfidelity/hifi/blob/d83600a22ee1579b15acf742e2fa83a84ce9951c/interface/src/ui/ScriptEditorWidget.cpp#L143).  Seems like "c:/" might be a thing?

libraries/script-engine/src/ScriptEngine.cpp
humbletim: Sure --  ScriptEngine::_fileNameString has historically been used to store "casual engine names" like `Entities 1`, `EmbeddedEntityScript`, `JSConsole`, etc. -- which code in some places assumes contains an absolute URL or real file path, so was effectively upgrading the engine name into a bogus `file:` URL..

This is where the "about:" prefixing comes from -- not an ideal solution, but at least `about:Entities 1` is already absolute and any relative paths resolved against it produce more benign `about:` URLs again.

Even if absolutizing everything we have influence over -- stack frames still may not contain usable `.fileName` fields at all levels -- for example, in strack traces can see as stuff like `forEach() at -1` and  `<global>() at -1`.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -892,23 +881,28 @@ QScriptValue ScriptEngine::evaluate(const QString& sourceCode, const QString& fi
...
-        result = uncaughtException();
-        reportUncaughtException(program.fileName());
-        emit evaluationFinished(result, true);
-        clearExceptions();
-    } else {
-        emit evaluationFinished(result, false);
+    QScriptValue result;
+    {
+        // catch any exceptions throw by nested operations
+        ExceptionEmitter tryCatch(this, __FUNCTION__);
```
humbletim: Yes -- I think one of the last things we want to happen inside nested stack frames is for the JS exceptions to get inadvertently printed and cleared   For standard JavaScript's `try/catch/finally` blocks, exceptions need to instead "flow upwards" -- until either caught or running out of chances to get caught.  Might be able to explain a little better using the whiteboard inworld or screen share.

libraries/script-engine/src/ScriptEngine.cpp
huffman: I think that if we can safely use a DirectConnection here then that would remove any worry about the ScriptEngine's last exception not being the exception that this signal is reporting. It probably ends up using a DirectConnection, but I think it would be worth making it explicit. Thoughts?

libraries/script-engine/src/BaseScriptEngine.cpp
huffman: Got it. I agree with the note on passing QScriptValues around. Are there places where this is happening now?

libraries/script-engine/src/BaseScriptEngine.cpp
```diff
@@ -0,0 +1,321 @@
...
+        detail = extraDetail;
+    } else if (!extraDetail.isEmpty()) {
+        detail += "(" + extraDetail + ")";
+    }
+    if (lineNumber <= 0) {
+        lineNumber = uncaughtExceptionLineNumber();
+    }
+    if (fileName.isEmpty()) {
+        // climb the stack frames looking for something useful to display
+        for(auto ctx = currentContext(); ctx && fileName.isEmpty(); ctx = ctx->parentContext()) {
```
ZappoMan: ctx as a variable name doesn't match coding standard... if it's an iterator, it should be a single letter variable... if it's not an iterator it needs to be more verbose "context"

libraries/script-engine/src/ScriptEngine.cpp
huffman: Ok, I believe I understand the general idea. I am confused about what the ExceptionEmitter does other than potentially emit an exception (if not evaluating). It has a consume function, but I'm not sure where that comes into play here.

One other concern: will this cause `Script.include`d scripts to bubble exceptions, causing existing scripts that would have continued evaluating to fail on a failed include?

libraries/script-engine/src/ScriptEngine.cpp
huffman: Ah, got it. Can a comment be added to make note of this?

libraries/script-engine/src/ScriptEngine.cpp
huffman: Got it! Thanks Tim.

libraries/script-engine/src/BaseScriptEngine.cpp
humbletim: @ZappoMan Ahh OK, didn't know a single letter would be acceptable here -- but can easily adopt.  

Could you remind me where the latest coding standard is at?  I thought I found it before through [CONTRIBUTING.md](https://github.com/highfidelity/hifi/blob/master/CONTRIBUTING.md), but that seems to currently lead to a [404 error page](https://wiki.highfidelity.com/v1.0/docs/coding-standard)...

libraries/script-engine/src/BaseScriptEngine.cpp
humbletim: Yup -- and apparently I've been patching-around these segfaults for over a year now without knowing what was causing them (or why they only ever seemed to target me!).

But I think I just connected the dots -- remember how I mentioned my Linux i5820k build (for whatever reason) only uses a single core?

I think that's why segfaults and timing conditions like "reload all" and "entity stampeding" have been singling me out -- Qt seems to saturate that single core, leading to a greater variety of thread-thread interleaving situations, and in fact serialized interleaving *all the time*.

If I am right about this, then others ought to be able to make "very crashy builds" too -- just by forcing Interface to run on a single core...

In any event, if interested I've a few ideas on addressing the problem more wholemeal, but here is [a specific example](https://github.com/humbletim/hifi/commit/3681ec99e741dad97b9c5a6b76d59168f7305b3b) of where I've seen mismatched engine / `QScriptValue` / threads emerge.

libraries/script-engine/src/ScriptEngine.cpp
humbletim: Yup, that makes perfect sense and will add in.

libraries/script-engine/src/ScriptEngine.cpp
humbletim: Maybe now that exactly where and when exception checks are needed it would make more sense inline things?

In most cases that means changing to use the below three lines.  These couldn't placed in the same exact spot though -- instead they'd need to go somewhere below the last possible moment something might emerge a JS exception:
```c++
if (!isEvaluating() && hasUncaughtException()) {
    emit unhandledException(cloneUncaughtException(__FUNCTION__));
}
```

Re: maintaining existing `include` behavior -- I am [covering the top level case]( https://github.com/humbletim/hifi/blob/b15956278c680bf94c38dbd3b8a259f0fc431d4b/libraries/script-engine/src/ScriptEngine.cpp#L1366-L1376), but maybe it should be expanded to consume exception from nested includes too?

hifi-gustavo:
Build console output for commit: 132a889f7474728dcfc5facd223f800854a378ff available at: https://gist.github.com/05e2159c339b6b39697ce3c33756b6eb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-132a889.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-132a889.dmg

libraries/script-engine/src/BaseScriptEngine.cpp
humbletim: Another example (`EntitiyItemProperties::copyToScriptValue`):
```log
[02/23 09:12:23] [DEBUG] [tim.EntityItemProperties.cpp] EntitiyItemProperties::copyToScriptValue -- not in same thread!?
[02/23 09:12:23] [DEBUG] [tim.EntityItemProperties.cpp] engine->thread()  QThread(0x33c4cd0, name = "Main Thread")
[02/23 09:12:23] [DEBUG] [tim.EntityItemProperties.cpp] QThread::currentThread()  QThread(0x6a6a580, name = "Script Thread:defaultScripts.js")
```

huffman: OK, I was focusing on the "bound QScriptValue's cannot be used by another QScriptEngine instance.", but these are mostly "bound QScriptValue's are not threadsafe (... and Error objects are bound values)." issues, right? So the copied values are not associated/bound to any QScriptEngine?

humbletim: ... all values discussed above are bound to a QScriptEngine.

Attempting to use a bound QScriptValue with a different QScriptEngine  essentially results in Qt saying "nope... can't do that" (the operation fails and produces an internal Qt log entry).

Attempting to use a bound QScriptVale from two different threads often results in Interface *immediately crashing*.

By specifically checking for mismatched threads above, both the requested `EntitiyItemProperties::copyToScriptValue` operation and a likely crash were aborted.

hifi-gustavo:
Build console output for commit: e794b64c31e748c5aaeff860b566a1d8ed2ee758 available at: https://gist.github.com/05e2159c339b6b39697ce3c33756b6eb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-e794b64.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-e794b64.dmg

hifi-gustavo:
Build console output for commit: 8ab7a8cad52e20ff48ccdb0c3765ab413a8ef5a0 available at: https://gist.github.com/05e2159c339b6b39697ce3c33756b6eb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-8ab7a8c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-8ab7a8c.dmg

libraries/script-engine/src/BaseScriptEngine.cpp
huffman: > Attempting to use a bound QScriptValue with a different QScriptEngine

Is that shown in one of your examples? I am trying to understand where that happens currently in the codebase.

To clarify, I understand that QScriptValue's, and it's bound QScriptEngine might be manipulated on threads that are different than that QScriptEngine's thread. I am wondering when we are trying to use a QScriptValue with a QScriptEngine that the QScriptValue isn't bound to.

libraries/script-engine/src/ScriptEngine.cpp
huffman: Yes, I think it should (it looks like you already changed it).

Regarding `ExceptionEmitter`, I think it makes more sense to replace it with a simple function call if it's not providing any benefit beyond it's dtor. I think it's just adding complexity unnecessarily if that is the case.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1485,43 +1494,138 @@ bool ScriptEngine::getEntityScriptDetails(const EntityItemID& entityID, EntitySc
...
-        EntityScriptDetails details = engine->_entityScripts[entityID];
-        if (details.status == EntityScriptStatus::PENDING || details.status == EntityScriptStatus::UNLOADED) {
-            engine->updateEntityScriptStatus(entityID, EntityScriptStatus::LOADING, QThread::currentThread()->objectName());
+    auto processDeferredEntities = [this](QScriptContext* context, QScriptEngine*) -> QScriptValue {
+        QList<DeferredLoadEntity> stillPending;
+
+        for (int i=0; i < RETRY_ATTEMPT_BATCH && !_pendingLoadEntities.isEmpty(); i++) {
+            auto retry = _pendingLoadEntities.takeFirst();
+            qCDebug(scriptengine) << "deferLoadEntityScript.retry[" << i << "]:" << retry.entityID << retry.entityScript;
+            if(!_entityScripts.contains(retry.entityID)) {
```
huffman: space after if

hifi-gustavo:
Build console output for commit: 0b73cef252f773e1e2a91a06ae301aa7d221cf83 available at: https://gist.github.com/05e2159c339b6b39697ce3c33756b6eb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-0b73cef.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-0b73cef.dmg

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1485,43 +1492,138 @@ bool ScriptEngine::getEntityScriptDetails(const EntityItemID& entityID, EntitySc
...
+    if (!_entityScripts.contains(entityID)) {
+        // make sure EntityScriptDetails has an entry for this UUID right away
+        // (which allows bailing from the loading/provisioning process early if the Entity gets deleted mid-flight)
+        updateEntityScriptStatus(entityID, EntityScriptStatus::PENDING, "...pending...");
+    }
+
+    // This lock/defer approach allows multiple Entities to boot from the same script URL while still taking
+    // full advantage of cacheable require modules.  This only affects Entities literally coming in back-to-back
+    // before the first one has time to finish loading.
+    auto &locker = _lockPerUniqueScriptURL[entityScript];
```
huffman: I don't know the answer to this - is this safe to do with `std::map`? You wouldn't want to do this with `std::vector`, for instance, because the object you are referring to may move (resizing the array, for instance).

huffman: (I am specifically talking about it being captured in the lambda below, I probably should have commented there.

hifi-gustavo:
Build console output for commit: 3ad3fd472e5720f1a67d202657ababab98b88d51 available at: https://gist.github.com/05e2159c339b6b39697ce3c33756b6eb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-3ad3fd4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-3ad3fd4.dmg

humbletim:
### Testing
* Perform general smoke test including a few familiar Entity/Client scripts.
* Run below three script tests from the Console... or Script Editor.
#### Script.resolvePath / Script.include regression tests:
* [/scripts/developer/tests/unit_tests/scriptUnitTests.js](
https://raw.githubusercontent.com/humbletim/hifi/21114-part2/scripts/developer/tests/unit_tests/scriptUnitTests.js)
  * There should be four "Uncaught Exceptions" visible in the console, but the Test Results themselves should all pass.
#### Entity stampede-loading test:
* [/scripts/developer/tests/entityStampedeTest.js](
https://raw.githubusercontent.com/humbletim/hifi/21114-part2/scripts/developer/tests/entityStampedeTest.js)
  * *NOTE: this test uses a simple preload so should be performed solo without any other Avatars around.*
  * Shortly after loading, a ring of **alternating black-white** Sphere-shaped Entities should appear.
  * Without stopping the script, try leaving the Domain and coming back.
    * ... the overall ring should now consist of **solid white** Cube Entities.
     *(it's OK if the black shapes are still slightly visible between the cubes)*
  * Without stopping the script, try leaving the Domain for a second time and coming back.
    * ... the ring should now have returned to previous **alternating black-white** state.
#### Entity bootstrap / Extended Exception tests:
* [testEntityStackTraces.js](
https://gist.githubusercontent.com/humbletim/6afab827fd5711b8282d09577ac8e147/raw/19966a18cc28fea92e8d7f9645a89f1d68ad30ce/testEntityStackTraces.js)
  * *NOTE: this test loads a variety of Entity script constructors with errors -- after the test completes it's the debug log that matters.*
  * Should see several detailed `[UncaughtException....] Error:`... lines reported.
  * Verify there are no exact duplicates (it might look at first like there are two sets of dupes, but their Error messages should be different...).  As a reference point here is [example log output](https://gist.githubusercontent.com/humbletim/6afab827fd5711b8282d09577ac8e147/raw/e56d712221ce3a8e7f03fbc2d43cff8db88b0835/example-output.txt)

libraries/script-engine/src/ScriptEngine.cpp
humbletim: ... dunno either, but now that I think about it more, probably not safe at that level in the case of the first entity.

The lock emerged early on -- as part of the battle to convince Entities to respect each other's load cycles (despite stampeding); `loadEntityScript` used to almost always be called on the Main Thread (but would quickly punt over to ScriptCache).

As of now though the `invokeMethod` above this snippet should mean the whole function always lives on the engine thread -- so a simple hash would probably now work just as well.

hifi-gustavo:
Build console output for commit: 02827552aebe347d2d80307da1439db97be77f66 available at: https://gist.github.com/05e2159c339b6b39697ce3c33756b6eb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-0282755.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-0282755.dmg

hifi-gustavo:
Build console output for commit: 3ac116545e20fcf136f45a386a662c91fe4d3fd6 available at: https://gist.github.com/05e2159c339b6b39697ce3c33756b6eb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-3ac1165.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-3ac1165.dmg

libraries/script-engine/src/BaseScriptEngine.cpp
humbletim: Ahh, OK -- the engine-engine mismatch here only happens between the Entities preflight sandbox and main ScriptEngine instance.

hifi-gustavo:
Build console output for commit: cdbb13ecff0e76a42249a6d30fd89256a0c9f7cd available at: https://gist.github.com/05e2159c339b6b39697ce3c33756b6eb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-cdbb13e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-cdbb13e.dmg

hifi-gustavo:
Build console output for commit: 731c01985bfafe6966eed5afd49c622d1a0c5087 available at: https://gist.github.com/05e2159c339b6b39697ce3c33756b6eb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-731c019.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-731c019.dmg

hifi-gustavo:
Build console output for commit: 5b11c0e00a92b55e12f4c433177d807b251b1a1c available at: https://gist.github.com/05e2159c339b6b39697ce3c33756b6eb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-5b11c0e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-5b11c0e.dmg

OmegaHeron:
Testing Windows 10

Script.resolvePath / Script.include regression tests:
scriptUnitTests.js *Pass*

EntityClientScript stampede-loading test:
entityStampedeTest.js *Pass*

EntityServerScript stampede-loading test:
entityServerStampedeTest.js

This test is unclear to me.  Do you want it ran as a script from domain-server scripts, or as a script in an entity (local interface entity script) or as a script in entity from server entity server script?  Please clarify.

Entity bootstrap / Extended Exception tests:
testEntityStackTraces.js *Pass*

Tested a selection of common well known (to me) scripts - no issues.  Should probably be double checked against some of the scripts those with hand controllers use (toys like bow, guns etc).

General smoke test pass.

@humbletim - Clarify how entityServerStampedeTest.js is to be ran and I'll finish this.

humbletim:
@OmegaHeron -- `entityServerStampedeTest.js` is another "regular" Client script so can be launched the same way as the others -- it just happens to create Entities with their `.serverScripts` property populated (instead of `.script` property like the "client" variation of this test).

OmegaHeron:
Thanks @humbletim - will finish QA shortly.

OmegaHeron:
EntityServerScript stampede-loading test:
entityServerStampedeTest.js *Pass*

Just to be thorough, I ran CPU use comparison (my standard set) to master.  No change seen 👍

QA Pass (caveat that I was not able to test a wide variety of items like bows, ping pong guns and others).  Famous last words, it should be fine with those from what I see of other tests.

hifi-gustavo:
Build console output for commit: 654c72fb7e19c6ee6f5d46249220b5257c5bea9e available at: https://gist.github.com/05e2159c339b6b39697ce3c33756b6eb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-654c72f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-654c72f.dmg

hifi-gustavo:
Build console output for commit: 2505a89b5e6c3aac60420918164a070ca9bdc5fb available at: https://gist.github.com/05e2159c339b6b39697ce3c33756b6eb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-2505a89.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9685-2505a89.dmg

PR #9686, opened 2017-02-16
CR 21114 -- Implement modules/require support into hifi

humbletim:
This PR integrates support for creating and using CommonJS/Node.js style JS Modules as part of High Fidelity scripting.  Worklist item: https://worklist.net/21114 

## Testing
Run the below test scripts from the Console or Script Editor.

#### Script.require / module.exports tests:
* `file:///~/developer/tests/unit_tests/moduleUnitTests.js`
    * All tests should pass.
    * *note: it's normal for Error messages to appear in the console during the tests*

#### Script.resolvePath / Script.include regression tests:
* `file:///~/developer/tests/unit_tests/scriptUnitTests.js`
    * All tests should pass.
    * During the test run, four Error messages should appear in the console:
        * Verify the Errors are now prefixed with `[Uncaught Exception evaluateInclude] ...`.
        * *note: ... in the previous PR these might have been displaying as `[Uncaught Exception operator()] ...`*
  
#### Entity bootstrap / Extended Exception regression tests:
* [testEntityStackTraces.js](
https://gist.githubusercontent.com/humbletim/6afab827fd5711b8282d09577ac8e147/raw/19966a18cc28fea92e8d7f9645a89f1d68ad30ce/testEntityStackTraces.js)
    * This test is from the earlier PR just to make sure extended Error prefixes are still working after related updates.
    * *reference point [example log output](https://gist.githubusercontent.com/humbletim/6afab827fd5711b8282d09577ac8e147/raw/e56d712221ce3a8e7f03fbc2d43cff8db88b0835/example-output.txt)*

-----------------------------------

## Summary / Notes

By adopting CommonJS/Node.js conventions a lot of the documentation online for standard JS modules now relates to creating modules within High Fidelity too.

Here are a few specific points to help bridge the gap:
### module.exports

- JS Modules export values using a placeholder object called `module.exports`.  The system automatically defines this placeholder as part of loading scripts with `Script.require`.
- To build-out an API as a module, properties can be assigned to `module.exports` one at a time.
- Or to export an entire API at once, the `module.exports` property can be replaced wholemeal.
- Here is a quick example of a module:
```javascript
// mymodule.js
module.exports = {
  foo: 'bar',
  add: function(a,b) {
    return a + b;
  }
};
```
- And using it:
```javascript
// myscript.js
var mymod = Script.require('./mymodule.js');
print(mymod.foo, mymod.add(2,2));
// prints "bar 4"
```

*(for compatibility with other environments, module scripts can also export values by assigning properties to the `exports` free variable: `exports.property = value;`.  However, for various reasons using `module.exports` is considered preferrable.)*

#### Module evaluation
- Module are loaded by `id`, which typically means relative filename or absolute URL to a script file.
- You can load Modules from the local filesystem, but access is limited in the same ways as `Script.include`.
- Each Module is evaluated in its own scope, so any `var's` and `const's` are always local to that module (ie: there's no need to manually wrap scripts within a global closure like `(function(){})()`).
- Modules can still access and create global variables if needed.
    - ... note that it's therefore possible to "upgrade" existing include files to work with both `Script.include` and `Script.require` at the same time.

#### The `module` Object
This system-defined Object is available to a module across its entire lifetime:  
* `module.id` contains the fully-resolved `id` that was used to load the script file.
* `module.parent` contains a reference to the parent Module (if any) that required the current one.
* `module.children` contains an Array of references to any child Modules that have been required.
* *note: for compatibility a few other properties are also defined; see source*

#### JSON Modules
- Module `id's` ending in `.json` will be fetched per usual, but the script contents will be interpreted as JSON -- with the parsed result becoming the module's export value.
- JSON modules can be loaded from the local filesystem -- which means if you need to store configuration alongside your script, you can now access that data as a relative module (and have things generally work without changes if later deploying to a web server or ATP folder).
- To explicitly load a Module as JSON (ie: regardless of file extension), the following workaround can be used:
   `var config = Script.require.call({ 'content-type': 'application/json' }, moduleURL);`

#### System Modules
*(... placeholder functionality / initial plumbing)*
* When a Module `id` has no file extension it gets interpreted to mean a "system module." 
* Other platforms sometimes leverage these abstract modules as a way to expose C++ or hybrid functionality.
* This extra layer of indirection also provides a mechanism for backwards-incompatible API changes to potentially remain *virtually compatible* with all existing scripts.
* This convention might also make sense for any "built-in" modules that ship with the App.
* Here is an example that demonstrates how system modules could be used to create a flex point between C++ and JS scripting layers: [scripts/modules/vec3.js](https://github.com/humbletim/hifi/blob/21114-part3/scripts/modules/vec3.js) (within this PR that module can be accessed as `var vec3 = Script.require('vec3');`).

#### Module Caching
- Modules are cached per script engine (by their fully-resolved `id`).  While cached, all subsequent `require's` for the same `id` return a reference to the previous `module.exports` value.
- The Module cache can be inspected from scripting through `Script.require.cache`.
- Given a valid Module `id`, a fully-resolved URL can be found using `Script.require.resolve(id)`
- Entities all share a single script engine -- and that means module cache too.  This makes cache invalidation a little more complicated, but to help simplify when you hit Control-R that now also invalidates the Entities module cache.
- To invalidate just the Module cache, call `Script.resetModuleCache(true);` from the corresponding script engine.
- To invalidate just a single module, delete that cache entry from the corresponding script engine. eg:
 `delete Script.require.cache[Script.require.resolve(moduleid)];`
- Hopefully one the above options can work better that applying cache busters (eg: '?v=NN'), but please let me know if you find situations where they still them necessary.

//


hifi-gustavo:
Build console output for commit: 60aa1ac4815261a718db3c87451b680e04fad090 available at: https://gist.github.com/4b08e00c9411f997475a02121220e899<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-60aa1ac.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-60aa1ac.dmg

hifi-gustavo:
Build console output for commit: b2d94f09f29f97eb9dde9ef3d226e1b9cc00d2c0 available at: https://gist.github.com/4b08e00c9411f997475a02121220e899<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-b2d94f0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-b2d94f0.dmg

hifi-gustavo:
Build console output for commit: e84c0d216092c71f2545c32ec7918b512a43225f available at: https://gist.github.com/4b08e00c9411f997475a02121220e899<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-e84c0d2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-e84c0d2.dmg

hifi-gustavo:
Build console output for commit: 6245bbcfc866c8fcc38a62e0315e00a70d036588 available at: https://gist.github.com/4b08e00c9411f997475a02121220e899<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-6245bbc.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-6245bbc.dmg

hifi-gustavo:
Build console output for commit: 8e723ffa9413a196fba0280a9048f38d622a6c8d available at: https://gist.github.com/4b08e00c9411f997475a02121220e899<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-8e723ff.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-8e723ff.dmg

hifi-gustavo:
Build console output for commit: a2866b480c1ced6f497b3d7c0b14e63cec0d572e available at: https://gist.github.com/4b08e00c9411f997475a02121220e899<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-a2866b4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-a2866b4.dmg

hifi-gustavo:
Build console output for commit: 1d7119cf279e21f33ccd0ff2445159e193902774 available at: https://gist.github.com/4b08e00c9411f997475a02121220e899<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-1d7119c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-1d7119c.dmg

hifi-gustavo:
Build console output for commit: c559838dbb50c1e33dbc68c10b7b19985b320876 available at: https://gist.github.com/4b08e00c9411f997475a02121220e899<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-c559838.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-c559838.dmg

OmegaHeron:
Will be testing this, per request, in a few hours.

OmegaHeron:
Tested Windows 10

General smoke test passes.  All default scripts run as normal and provided test scripts return correct responses.  @humbletim -- could you look at my log output for last test (testEntityStackTraces) for correctness?  I *think* it's right, but I had some excess output.  If it's correct then QA Pass.

moduleUnitTests.js

`[03/08 11:22:47] [DEBUG] [hifi.scriptengine.script] script:print()<< <span style="color:green">All enabled tests passed!</span>`
`[03/08 11:22:47] [DEBUG] [hifi.scriptengine.script] script:print()<< Tests completed in 21939ms.`

scriptUnitTests.js

`[03/08 11:25:34] [DEBUG] [hifi.scriptengine.script] script:print()<< <span style="color:green">All enabled tests passed!</span>`
`[03/08 11:25:34] [DEBUG] [hifi.scriptengine.script] script:print()<< Tests completed in 69ms.`

testEntityStackTraces.js

`[03/08 11:27:13] [CRITICAL] [hifi.scriptengine] "[UncaughtException {76e584d2-e96e-4240-972b-f059291b73cf}] Error: Can't find variable: Undefined_preload in about:EmbeddedEntityScript:17\n[Backtrace]\n    <anonymous>(uuid = '{76e584d2-e96e-4240-972b-f059291b73cf}', ) at about:EmbeddedEntityScript:17\n    <global>() at -1"`
`[03/08 11:27:13] [CRITICAL] [hifi.scriptengine] "[UncaughtException {76e584d2-e96e-4240-972b-f059291b73cf}] Error: Can't find variable: Undefined_setTimeout_setTimeout in about:EmbeddedEntityScript:15\n[Backtrace]\n    <anonymous>() at about:EmbeddedEntityScript:15\n    <global>() at -1"`
`[03/08 11:27:13] [CRITICAL] [hifi.scriptengine] "[UncaughtException signalHandlerException] Error: Can't find variable: Undefined_Script_update in about:EmbeddedEntityScript:8\n[Backtrace]\n    blah(0.017054999247193336) at about:EmbeddedEntityScript:8\n    <global>() at -1"`
`[03/08 11:27:13] [CRITICAL] [hifi.scriptengine] "[UncaughtException signalHandlerException] Error: Can't find variable: Undefined_Script_update_SCOPED in about:EmbeddedEntityScript:12\n[Backtrace]\n    blah(0.017054999247193336) at about:EmbeddedEntityScript:12\n    <global>() at -1"`
`[03/08 11:27:13] [CRITICAL] [hifi.scriptengine] "[UncaughtException {76e584d2-e96e-4240-972b-f059291b73cf}] Error: Can't find variable: Undefined_constructor_setTimeout in about:EmbeddedEntityScript:2\n[Backtrace]\n    <anonymous>() at about:EmbeddedEntityScript:2\n    <global>() at -1"`
`[03/08 11:27:15] [DEBUG] [hifi.scriptengine] unload called while !running [entity-id: "{968e5af1-e694-4056-80de-56e75d156538}" ] EntityScriptStatus_::EntityScriptStatus(ERROR_LOADING_SCRIPT)`
`[03/08 11:27:15] [DEBUG] [hifi.scriptengine] unload called while !running [entity-id: "{c0940e8b-2bb3-44ce-bbcf-a2a194e5a4dd}" ] EntityScriptStatus_::EntityScriptStatus(ERROR_LOADING_SCRIPT)`
`[03/08 11:27:15] [DEBUG] [hifi.scriptengine] unload called while !running [entity-id: "{27bb8826-8e7a-44f6-837d-13aca55578ef}" ] EntityScriptStatus_::EntityScriptStatus(ERROR_LOADING_SCRIPT)`
`[03/08 11:27:15] [DEBUG] [hifi.scriptengine] unload called while !running [entity-id: "{6ed77a46-810a-4cd9-a0d5-7bd103a94093}" ] EntityScriptStatus_::EntityScriptStatus(ERROR_LOADING_SCRIPT)`
`[03/08 11:27:15] [DEBUG] [hifi.scriptengine] unload called while !running [entity-id: "{69341f40-eba4-4dd6-9df0-02b81084728e}" ] EntityScriptStatus_::EntityScriptStatus(ERROR_LOADING_SCRIPT)`

humbletim:
That debug output looks OK in general.

I think maybe I should remove the extra [`unload called while !running`](https://github.com/highfidelity/hifi/blame/a744e0d11c2aae3b21f88886f8a3ec7bd628d9ed/libraries/script-engine/src/ScriptEngine.cpp#L1868) debug prints though.  @huffman ?

huffman:
@humbletim If they are harmless and would be spamming our logs then I agree they should be removed.

hifi-gustavo:
Build console output for commit: db73c80ba19f21416eac33e66db1cb929004d366 available at: https://gist.github.com/4b08e00c9411f997475a02121220e899<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-db73c80.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-db73c80.dmg

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1304,11 +1362,421 @@ void ScriptEngine::print(const QString& message) {
...
+        });
+
+        // this helps detect the case where stop() is invoked during the download
+        //  but not seen in time to abort processing in onload()...
+        connect(&monitor, &QTimer::timeout, this, [this, &loop, &loader]{
+            if (isStopping()) {
+                loop.exit(-1);
+            }
+        });
+        monitor.start(500);
```
huffman: This seems really short. It times out a request after 500 ms?

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1304,11 +1362,421 @@ void ScriptEngine::print(const QString& message) {
...
+
+    // module.require is a bound version of require that always resolves relative to that module's path
+    auto boundRequire = QScriptEngine::evaluate("(function(id) { return Script.require(Script.require.resolve(id, this.filename)); })", "(boundRequire)");
+    module.setProperty("require", boundRequire, READONLY_PROP_FLAGS);
+
+    return module;
+}
+
+// synchronously fetch a module's source code using BatchLoader
+QScriptValue ScriptEngine::fetchModuleSource(const QString& modulePath, const bool forceDownload) {
```
huffman: Does this need to return a QScriptValue?

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1304,11 +1362,421 @@ void ScriptEngine::print(const QString& message) {
...
+    }
+
+    {
+        // It seems that many JSON sources don't actually put .json in the URL...
+        // so for now as a workaround users wanting to indicate JSON parsing mode can
+        // do so by calling with a custom this context, eg:
+        //
+        //   var ITEMS_URL = 'https://highfidelity.io/api/v1/marketplace/items';
+        //   var thisObject = { 'content-type': 'application/json' };
+        //   var items = Script.require.call(thisObject, ITEMS_URL + '?category=everything&sort=recent');
```
huffman: Is there a precedent for doing it this way? If not, it seems like having a second optional `content-type` (or `options`) parameter would be more expected.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1304,11 +1362,421 @@ void ScriptEngine::print(const QString& message) {
...
+            _debugDump("this", thisObject);
+#endif
+            applyUserOptions(module, thisObject);
+        }
+    }
+
+    // evaluate the module
+    auto result = instantiateModule(module, sourceCode);
+
+    if (result.isError() && !result.strictlyEquals(module.property("exports"))) {
```
huffman: In what case is the latter true?

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1487,6 +1958,52 @@ void ScriptEngine::updateEntityScriptStatus(const EntityItemID& entityID, const
...
+            map["isError"] = true;
+            map["errorInfo"] = "Entity script details unavailable";
+            map["entityID"] = entityID.toString();
+        }
+    }
+    return map;
+}
+
+QFuture<QVariant> ScriptEngine::getLocalEntityScriptDetails(const EntityItemID& entityID) {
+    return QtConcurrent::run(this, &ScriptEngine::cloneEntityScriptDetails, entityID);
```
huffman: This has me confused - what is going on here? Looking at the docs, the first param should be a QThreadPool?

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -333,7 +345,11 @@ void ScriptEngine::runInThread() {
     // The thread interface cannot live on itself, and we want to move this into the thread, so
     // the thread cannot have this as a parent.
     QThread* workerThread = new QThread();
+#ifdef Q_OS_LINUX
+    workerThread->setObjectName(QString("js:") + getFilename());
```
humbletim: `pthread` names on Linux are capped to 15 chars (which means with full "Script Thread:" prefix only the first letter of thread names was available within `gdb` and segfault reports).

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -532,6 +548,40 @@ static QScriptValue createScriptableResourcePrototype(QScriptEngine* engine) {
     return prototype;
 }
 
+void ScriptEngine::resetModuleCache(bool deleteScriptCache) {
+    if (QThread::currentThread() != thread()) {
+        executeOnScriptThread([=](){ resetModuleCache(deleteScriptCache); });
```
humbletim: Except for the lambda aspect, roughly the same diff -- internally `executeOnScriptThread` just checks whether it's on the proper thread, and if not calls `invokeMethod` on itself.  Other benefits though include not having to hand-code Q_ARGs and it also avoids leaking arbitrary internal methods to JS (which marking something Q_INVOKABLE generally seems to do).

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1304,11 +1362,421 @@ void ScriptEngine::print(const QString& message) {
...
+    auto displayId = moduleId;
+    if (displayId.length() > MAX_DEBUG_VALUE_LENGTH) {
+        displayId = displayId.mid(0, MAX_DEBUG_VALUE_LENGTH) + "...";
+    }
+    auto message = QString("Cannot find module '%1' (%2)").arg(displayId);
+
+    auto throwResolveError = [&](const QScriptValue& error) -> QString {
+        raiseException(error);
+        maybeEmitUncaughtException("require.resolve");
+        return nullptr;
```
humbletim: Yup, good catch.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1304,11 +1362,421 @@ void ScriptEngine::print(const QString& message) {
...
+    auto idLength = url.toString().length();
+    if (idLength < 1 || idLength > MAX_MODULE_ID_LENTGH) {
+        auto details = QString("rejecting invalid module id size (%1 chars [1,%2])")
+            .arg(idLength).arg(MAX_MODULE_ID_LENTGH);
+        return throwResolveError(makeError(message.arg(details), "RangeError"));
+    }
+
+    // this regex matches: absolute, dotted or path-like URLs
+    // (ie: the kind of stuff ScriptEngine::resolvePath already handles)
+    QRegularExpression qualified ("^\\w+:|^/|^[.]{1,2}(/|$)");
```
humbletim: Since "system modules" are just a placeholder concept (per requirements), I thought it made more sense to have the regex capture what we know -- which is that supported module ids are going to initially consist of URLs, absolute paths, or relative paths.

We could probably "invert" the regex somehow -- but other variations seemed to sacrifice interesting options.  Like maybe the system wants to later have `Script.require('com.highfidelity.system.utils')`  work, which remains a possibility within the current scheme.

In general the regex isn't meant to restrict or validate -- just to partition; where there certain module names you were concerned about getting misinterpreted as system modules?

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1304,11 +1362,421 @@ void ScriptEngine::print(const QString& message) {
...
+    // this is for module.require (which is a bound version of require that's always relative to the module path)
+    if (!relativeTo.isEmpty()) {
+        url = QUrl(relativeTo).resolved(moduleId);
+        url = resolvePath(url.toString());
+    } else if (qualified.match(moduleId).hasMatch()) {
+        url = resolvePath(moduleId);
+    } else {
+        // check if the moduleId refers to a "system" module
+        QString defaultsPath = defaultScriptsLoc.path();
+        QString systemModulePath = QString("%1/modules/%2.js").arg(defaultsPath).arg(moduleId);
```
humbletim: As as scripter, no -- the parenting checks that immediately follow apply to all resulting local file paths -- system or otherwise.

Sure, will rename to `systemPath`.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1304,11 +1362,421 @@ void ScriptEngine::print(const QString& message) {
...
+
+    // if it looks like a local file, verify that it's an allowed path and really a file
+    if (url.isLocalFile()) {
+        QFileInfo file(url.toLocalFile());
+        QUrl canonical = url;
+        if (file.exists()) {
+            canonical.setPath(file.canonicalFilePath());
+        }
+
+        bool disallowOutsideFiles = !defaultScriptsLocation().isParentOf(canonical) && !currentSandboxURL.isLocalFile();
```
humbletim: Agreed... with the previous PRs I refactored the generic stuff like path stripping and filesystem case sensitivity out of `include` and into `PathUtils` (https://github.com/highfidelity/hifi/blob/master/libraries/shared/src/PathUtils.cpp#L87-L109), where it could then be shared with `require`.

There are actually several other places across the code base doing various (related) kinds of URL, path, or scheme verification/normalization that might benefit by rounding together into a few reusable `PathUtils` methods...

Maybe for now a local helper function that require/include both call into could work?

libraries/script-engine/src/ScriptEngine.cpp
humbletim: Not necessarily, but using a flexible hash map seems to work best as a way to accommodate error, content and metadata information like this when specific to scripting.

libraries/script-engine/src/ScriptEngine.cpp
huffman: > were there certain module names you were concerned about getting misinterpreted as system modules?

I was more concerned with not having full coverage over valid non-system-module-names. node.js will not resolve `something.js` relative to the current directory unless the current directory happens to be in the list of paths to search (NODE_PATH?), right?

libraries/script-engine/src/ScriptEngine.cpp
humbletim: It's a repeating timer that helps address the case where the script engine gets stopped during a long-running network request.

All I really wanted though was a way for incoming `stop()` request(s) to "cut in line" ahead of EventLoop processing -- without something like this, `isStopping()` seems to return `false` above when `onload` gets called (despite `stop()` -- which must be getting invoke-queued somewhere).

With this monitor, `isStopping()` checks are regularly interleaved into the event queue -- with benefit that at most ~500ms of waiting is ever spent holding for results that no longer matter.  Would be nicer if BatchLoader and ScriptCache supported ways of aborting requests..

libraries/script-engine/src/ScriptEngine.cpp
huffman: Couple opposing thoughts:

 * It feels wrong to use a QScriptValue for something that won't ever be used on the scripting side. If we want this kind of flexibility, a QVariantMap or QVariantHash would seem to be better suited.
 * We lose type safety, and information about what the returned value will contain.

Admittedly, I prefer using a struct here, but at a minimum I think a QVariant* value would be better suited.

libraries/script-engine/src/ScriptEngine.cpp
huffman: If we encapsulate the majority of the logic inside a single function then we should do that, otherwise we should leave it as-is.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1304,11 +1362,421 @@ void ScriptEngine::print(const QString& message) {
...
+// CommonJS/Node.js like require/module support
+QScriptValue ScriptEngine::require(const QString& moduleId) {
+    qCDebug(scriptengine_module) << "ScriptEngine::require(" << moduleId.left(MAX_DEBUG_VALUE_LENGTH) << ")";
+    if (!IS_THREADSAFE_INVOCATION(thread(), __FUNCTION__)) {
+        return unboundNullValue();
+    }
+
+    auto jsRequire = globalObject().property("Script").property("require");
+    auto cacheMeta = jsRequire.data();
+    auto cache = jsRequire.property("cache");
```
humbletim: Sure, storing this information on the JS side makes sense here I think because it's not actually C++ data (or meant to be) -- it's cached JS values that originate on the JS side and represent `require.cache` within the overall module pattern.

By storing the cache under the global object, it and its values automatically follow the life cycle of the engine -- by not storing them as reachable C++ properties (private or otherwise), the historical problem of having code assume ScriptEngine is threadsafe (or inadvertently access non-primitive engine values from other threads) is slightly-better protected against.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -1304,11 +1362,421 @@ void ScriptEngine::print(const QString& message) {
...
+        maybeEmitUncaughtException(__FUNCTION__);
+        return unboundNullValue();
+    }
+
+    // check the resolved path against the cache
+    auto module = cache.property(modulePath);
+
+    // modules get cached in `Script.require.cache` and (similar to Node.js) users can access it
+    // to inspect particular entries and invalidate them by deleting the key:
+    //   `delete Script.require.cache[Script.require.resolve(moduleId)];`
```
humbletim: Many, but nothing new or specific to module caching.

Local Entities already share the same engine and globals -- so at the moment Entity scripts pretty much always risk affecting each other.   For example, try setting `Script.foo="bar"` from one Entity script and reading it back from another (or setting `Entities.deleteEntity=null`).

Have some ideas to offer later that might help mitigate that sort of thing.

libraries/script-engine/src/ScriptEngine.cpp
humbletim: Was more to avoid changing the signature of `require`, but beyond that I don't really have a preference.

The primary value here I think is in having any way that works the same when bringing a relative JSON file in during local development and then once deployed remotely.

How about simply having an explicit API method for this?  `Script.requireJSON(id)`?

libraries/script-engine/src/ScriptEngine.cpp
huffman: Yeah, I agree about existing risks. I am less concerned about whether this is adding risk and more concerned with whether we are moving in a direction that improves it or locks us into a position where we can't improve it because we are adding features users will come to rely on that can't fundamentally be fixed.

libraries/script-engine/src/ScriptEngine.cpp
huffman: > The primary value here I think is in having any way that works the same when bringing a relative JSON file in during local development and then once deployed remotely.

I'm not sure I follow. My suggestion was: `require(id, options={})`. My understanding is that you can use `require` to load JSON in nodejs, so this seems preferable to `requireJSON`

libraries/script-engine/src/ScriptEngine.cpp
humbletim: JS modules can export any value -- including an `Error` instance.  So the latter case will be true for example with `module.exports = new Error("missing package dependency");`.

Normally `result` isn't relevant -- it reflects the implicit "final value" of the script evaluation (we only have to check it because that's how `QScriptEngine` happens to send back certain kinds of compile/runtime errors).

libraries/script-engine/src/ScriptEngine.cpp
humbletim: http://doc.qt.io/qt-5/qtconcurrentrun.html#using-member-functions

libraries/script-engine/src/ScriptEngine.cpp
humbletim: Right.  Could you elaborate on what you mean by full coverage?

Or would it address the same concern by adding a specific test and rejection message for when users attempt to use unanchored relative filenames? (eg: `require('relative.js')`)?

libraries/script-engine/src/ScriptEngine.cpp
humbletim: Switching to use a QVariantMap sounds good to me.  I think for now flexibility is important to keep intact -- BatchLoader and ScriptCache both hard coded assumptions about the type of information returned, which made them slightly harder to adapt later like when wanting to include the reason why a request failed.

libraries/script-engine/src/ScriptEngine.cpp
huffman: By full coverage I just meant covering all of the possible ways you might specify a relative or absolute path. My thought was that the system names are more constrained and easier to test, and anything that is valid in the eyes of QUrl or our local file checking is should be accepted.

My second question was coming off of this thinking, and realizing we probably shouldn't accept a relative path that doesn't begin with `./`, in which case this approach probably makes more concern.

My main concern is that we are only allowing what we want to accept. If we unintentionally certain strings and users start to rely on that behavior, we won't be able to change it later. But, I think we are ok here. Thoughts?

libraries/script-engine/src/ScriptEngine.cpp
humbletim: Well Node.js supports loading JSON modules built-in, but not *any* kind of module over HTTP -- so we are in uncharted territory here.

Part of the appeal of using `require/module.exports` is its simplicity at the "using modules" level, so rather than changing the standard signature of `require` would rather see this feature removed for now.  In most cases it could be brought back for anybody who finds need as a pure JS helper module.

What I meant before was that the primary purpose of JSON modules is in being able to code-once and run-anywhere in terms of a script.js doing `var config = Script.require(./config.json')` -- that same script works when loaded as a file:///, atp: or https:/// URL -- provided only that the two files and kept together as a pair.

hifi-gustavo:
Build console output for commit: 366c90ef6acfb46e5f85998cc433ba2fce1a7639 available at: https://gist.github.com/4b08e00c9411f997475a02121220e899<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-366c90e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-366c90e.dmg

libraries/script-engine/src/ScriptEngine.cpp
humbletim: I think we're probably OK because both anchored paths and absolute URLs unambiguously refer to a specific resource -- so are not subject to much interpretation anyhow (just concrete resolution in context).

To help avoid confusion though I've added a specific check for when the User attempts to load a relative file using an unanchored path -- letting the user know exactly what the system expects instead in that case.

hifi-gustavo:
Build console output for commit: 52a571558cc974eedae95cc3f2f4908433c40c0e available at: https://gist.github.com/4b08e00c9411f997475a02121220e899<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-52a5715.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-52a5715.dmg

libraries/script-engine/src/ScriptEngine.cpp
humbletim: This solution helps move things in a direction that helps improve the situation so I think we're on the same page.  The promise being made to scripts is only that modules loaded with `require` get reflected in the `require.cache` Object -- and if/when cleared, that those entries are no longer considered cached by the system.  Nothing here promises that the cache is shared with other scripts in any particular fashion.

libraries/shared/src/BaseScriptEngine.cpp
```diff
@@ -18,18 +19,26 @@
...
 #include "Profile.h"
 
-const QString BaseScriptEngine::_SETTINGS_ENABLE_EXTENDED_EXCEPTIONS {
-    "com.highfidelity.experimental.enableExtendedJSExceptions"
-};
-
 const QString BaseScriptEngine::SCRIPT_EXCEPTION_FORMAT { "[%0] %1 in %2:%3" };
 const QString BaseScriptEngine::SCRIPT_BACKTRACE_SEP { "\n    " };
 
+bool BaseScriptEngine::IS_THREADSAFE_INVOCATION(const QThread *thread, const QString& method) {
```
humbletim: added (for some reason github didn't auto-collapse this box)

libraries/script-engine/src/ScriptEngine.cpp
huffman: Great - one question on the regex: It looks like it will match `.` and `..` - when are those valid paths to match here?

libraries/script-engine/src/ScriptEngine.cpp
huffman: Ah, got it. Thanks for the explanation.

libraries/script-engine/src/ScriptEngine.cpp
huffman: I sent you a message on Slack about this.

libraries/script-engine/src/ScriptEngine.cpp
huffman: Nice! This seems like something we could be using more often.

libraries/script-engine/src/ScriptEngine.cpp
humbletim: right now those kinds of paths are allowed at this level because they are anchored and definitely not a system module; down below such ids will get rejected as not leading to an actual file.

other systems (such as QML, Node.js and Python) allow "directories" like "." and ".." to be imported.  this is forward-compatible (without risking users prematurely adopting the habit) with that sort of scheme later on.

basically, instead of rejecting if not "isFile," some day we could accept if "isDir" and a folder matching whatever conventions arrived at later (`qmldir`, `package.json`, `__init__.py` being examples of how other platforms earmark "module" directories).

libraries/script-engine/src/ScriptEngine.cpp
humbletim: OK, I'll suggest leaving as-is for now.  Later on, maybe some unit/characterization tests could be added to capture how these restrictions currently work -- so that a refactoring pass has a way to verify overall behavior isn't impacted when consolidating code.

hifi-gustavo:
Build console output for commit: 758c07c7d4d1ee354dbe9f2b2153d39c85defea4 available at: https://gist.github.com/4b08e00c9411f997475a02121220e899<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-758c07c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-758c07c.dmg

Nex-Pro:
Testing this.

hifi-gustavo:
Build console output for commit: 87d62ad1eb6162ff979b53052f055ffc22a4fcf1 available at: https://gist.github.com/4b08e00c9411f997475a02121220e899<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-87d62ad.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9686-87d62ad.dmg

Nex-Pro:
Fail  "PR build doesn't seem up to date yet; Tim working to resolve two merge conflicts resulting from recent PRs merged into master"

humbletim:
OK -- the latest PR build seems to have everything integrated now.

Please retest.  Note that the first two test scripts above should now be launched using those PR-local file:/// paths.

Nex-Pro:
Will do shortly 
@humbletim

Nex-Pro:
[2017-03-16 22:02:45.117] [about:Entities 1] [log] preload {f35acad2-7fa5-4a32-8495-b689e7491c0f}
[2017-03-16 22:02:45.117] [about:Entities 1] [error] [UncaughtException {f35acad2-7fa5-4a32-8495-b689e7491c0f}] Error: Can't find variable: Undefined_preload in about:EmbeddedEntityScript:17
[Backtrace]
    (uuid = '{f35acad2-7fa5-4a32-8495-b689e7491c0f}', ) at about:EmbeddedEntityScript:17
    () at -1
[2017-03-16 22:02:45.118] [about:Entities 1] [error] [UncaughtException (construct {33f087b4-597a-411c-bccf-d524be3afb23})] Error: Can't find variable: Undefined_constructor in about:EmbeddedEntityScript:2
[Backtrace]
    () at about:EmbeddedEntityScript:2
    () at -1
[2017-03-16 22:02:45.118] [about:Entities 1] [error] [UncaughtException (constructor {c464256c-0763-4751-ad3a-9895a940395a})] Error: failed to load entity script -- expected a function, got QString, not a constructor in javascript:"not a constructor":
[2017-03-16 22:02:45.124] [about:Entities 1] [error] [UncaughtException {f35acad2-7fa5-4a32-8495-b689e7491c0f}] Error: Can't find variable: Undefined_setTimeout_setTimeout in about:EmbeddedEntityScript:15
[Backtrace]
    () at about:EmbeddedEntityScript:15
    () at -1
[2017-03-16 22:02:45.124] [about:Entities 1] [error] [UncaughtException (preflight {d5f33528-7751-4266-8925-55da358259b9})] Error: throw in main script body in about:EmbeddedEntityScript:1
[Backtrace]
    () at about:EmbeddedEntityScript:1
[2017-03-16 22:02:45.125] [about:Entities 1] [error] [UncaughtException {2810f5ac-7df3-4832-8a60-81b4d26c08e1}] SyntaxError: Expected `;', `;' in about:EmbeddedEntityScript:1
[Backtrace]
    () at -1
[2017-03-16 22:02:45.126] [about:Entities 1] [error] [UncaughtException (construct {f44bce79-7ccc-4805-8dbd-3ec7200c7f79})] Error: throw in contructor body in about:EmbeddedEntityScript:1
[Backtrace]
    () at about:EmbeddedEntityScript:1
    () at -1
[2017-03-16 22:02:45.126] [about:Entities 1] [error] [UncaughtException signalHandlerException] Error: Can't find variable: Undefined_Script_update in about:EmbeddedEntityScript:8
[Backtrace]
    blah(0.017222000285983086) at about:EmbeddedEntityScript:8
    () at -1
[2017-03-16 22:02:45.127] [about:Entities 1] [error] [UncaughtException signalHandlerException] Error: Can't find variable: Undefined_Script_update_SCOPED in about:EmbeddedEntityScript:12
[Backtrace]
    blah(0.017222000285983086) at about:EmbeddedEntityScript:12
    () at -1
[2017-03-16 22:02:45.134] [about:Entities 1] [error] [UncaughtException {f35acad2-7fa5-4a32-8495-b689e7491c0f}] Error: Can't find variable: Undefined_constructor_setTimeout in about:EmbeddedEntityScript:2
[Backtrace]
    () at about:EmbeddedEntityScript:2
    () at -1

Nex-Pro:
@humbletim ^

humbletim:
@Nex-Pro those intentional exceptions / error output look good to me.

The first two test scripts concluded with "All enabled tests passed!" right?

Nex-Pro:
Looks like a **PASS**

kevinmthomasse:
smoke test request per @cozza13 testing

kevinmthomasse:
## QA Test
**QA Tester**: KevinMThomas
**localhost 'createTestBlocks.js' Result**: **localhost 'createTestBlocks.js'** SUCCESS
**localhost ATP Upload + Edit Result**: **localhost ATP** SUCCESS
**Echo Server Audio**: **Echo Server Audio** SUCCESS
**localhost 'harmonicOscillator.js' Result**: **localhost 'harmonicOscillator.js'** UNTESTED
**localhost 'reverbTest.js' Result**: **localhost 'reverbTest.js'** SUCCESS
**localhost 'playaPerformanceTest.js' Result**: **localhost 'playaPerformanceTest.js'** SUCCESS
**dev-playa 'playaPerformanceTest.js' Result**: **dev-playa 'playaPerformanceTest.js'** SUCCESS
**dev-demo 'playaPerformanceTest.js' Result**: **dev-demo 'playaPerformanceTest.js'** SUCCESS
**Additional Test Results + Observations**: Successful basic smoke test and successful Home Tutorial test with Oculus and Hand controllers.  There was a crash initially and sent a bugsplat however was unable to replicate. 


**Domain** |localhost |dev-playa |dev-demo
---|---|---|---
**Render Rate**: |313 |186 |217
**Present Rate**: |304 |188 |216
**Present New Rate**: |244 |186 |208
**CPU Buffer Count**: |2562 |3277 |998
**GPU Buffer Count**: |788 |700 |244
**CPU Memory Usage**: |31823120 |58583152 |12435568
**GPU Memory Usage**: |18228656 |19101104 |6165936
**CPU Texture Count**: |230 |258 |97
**GPU Texture Count**: |171 |149 |63
**GPU Transfer Count**: |0 |0 |0
**CPU Texture Memory Usage**: |242593680 |406035560 |88913660
**GPU Texture Memory Usage**: |553851531 |602587118 |225827042
**GPU Virtual Memory Usage**: |553851531|602587118|225827042
**Triangle Count**: |249195 |684518 |172743
**Triangle Rate**: |83065000 |136903600 |43185750
**API Drawcall Count**: |163 |285 |71
**Drawcall Count**: |2982 |526 |76
**Drawcall Rate**: |994000 |105200 |19000
**State Changes Texture Count**:|315 |273 |91
**Pipeline Count**: |72 |39 |31
**Item Opaques**: |246 |424 |109
**Item Translucents**: |98 |62 |15
**Item Lights**: |11 |49 |15
**Timing Opaques**: |0.984 |1.112 |0.289
**Timing Translucents**: |0.129 |0.073 |0.058
**Timing Lighting**: |0.02 |0.024 |0.042
**Render Frame**: |1.821 |2.325 |1.41

huffman:
@kevinmthomasse thanks for testing - I looked at the bugsplat and it appears unrelated

PR #9688, opened 2017-02-16
DNM: Introducing the KTX file format for texture serialization

samcake:
Work in progress

### DIFF
--- a/libraries/ktx/src/ktx/KTX.cpp
+++ b/libraries/ktx/src/ktx/KTX.cpp
@@ -15,6 +15,11 @@
 
 using namespace ktx;
 
+uint32_t evalPadding(size_t byteSize) {
+    auto padding = byteSize % PACKING_SIZE;
+    return (padding ? PACKING_SIZE - padding : 0);
+}
+
 const Header::Identifier ktx::Header::IDENTIFIER {{
     0xAB, 0x4B, 0x54, 0x58, 0x20, 0x31, 0x31, 0xBB, 0x0D, 0x0A, 0x1A, 0x0A
 }};
@@ -49,8 +54,8 @@ size_t Header::evalRowSize(uint32_t level) const {
     auto pixelWidth = evalPixelWidth(level);
     auto pixSize = evalPixelSize();
     auto netSize = pixelWidth * pixSize;
-    auto packing = netSize % PACKING_SIZE;
-    return netSize + (packing ? PACKING_SIZE - packing : 0);
+    auto padding = evalPadding(netSize);
+    return netSize + padding;
 }
 size_t Header::evalFaceSize(uint32_t level) const {
     auto pixelHeight = evalPixelHeight(level);
@@ -115,3 +120,4 @@ const Byte* KTX::getTexelsData() const {
         return nullptr;
     }
 }
+
--- a/libraries/ktx/src/ktx/KTX.cpp
+++ b/libraries/ktx/src/ktx/KTX.cpp
@@ -1,0 +1,92 @@
+//
+//  KTX.cpp
+//  ktx/src/ktx
+//
+//  Created by Zach Pomerantz on 2/08/2017.
+//  Copyright 2017 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include "KTX.h"
+
+#include <algorithm> //min max and more
+
+using namespace ktx;
+
+const Header::Identifier ktx::Header::IDENTIFIER {{
+    0xAB, 0x4B, 0x54, 0x58, 0x20, 0x31, 0x31, 0xBB, 0x0D, 0x0A, 0x1A, 0x0A
+}};
+
+Header::Header() {
+    memcpy(identifier, IDENTIFIER.data(), IDENTIFIER_LENGTH);
+}
+
+uint32_t Header::evalMaxDimension() const {
+    return std::max(pixelWidth, std::max(pixelHeight, pixelDepth));
+}
+
+uint32_t Header::evalMaxLevel() const {
+    return 1 + log2(evalMaxDimension());
+}
+
+uint32_t Header::evalPixelWidth(uint32_t level) const {
+    return std::max(pixelWidth >> level, 1U);
+}
+uint32_t Header::evalPixelHeight(uint32_t level) const {
+    return std::max(pixelHeight >> level, 1U);
+}
+uint32_t Header::evalPixelDepth(uint32_t level) const {
+    return std::max(pixelDepth >> level, 1U);
+}
+
+size_t Header::evalPixelSize() const {
+    return glTypeSize; // Really we should generate the size from the FOrmat etc
+}
+
+size_t Header::evalRowSize(uint32_t level) const {
+    auto pixelWidth = evalPixelWidth(level);
+    auto pixSize = evalPixelSize();
+    auto netSize = pixelWidth * pixSize;
+    auto packing = netSize % 4;
+    return netSize + (packing ? 4 - packing : 0);
+}
+size_t Header::evalFaceSize(uint32_t level) const {
+    auto pixelHeight = evalPixelHeight(level);
+    auto pixelDepth = evalPixelDepth(level);
+    auto rowSize = evalRowSize(level);
+    return pixelDepth * pixelHeight * rowSize;
+}
+size_t Header::evalImageSize(uint32_t level) const {
+    auto faceSize = evalFaceSize(level);
+    if (numberOfFaces == 6 && numberOfArrayElements == 0) {
+        return faceSize;
+    } else {
+        return (numberOfArrayElements * numberOfFaces * faceSize);
+    }
+}
+
+
+KTX::KTX() {
+}
+
+void KTX::resetStorage(Storage* storage) {
+    _storage.reset(storage);
+}
+
+const Header* KTX::getHeader() const {
+    if (_storage) {
+        return reinterpret_cast<const Header*> (_storage->_bytes);
+    } else {
+        return nullptr;
+    }
+}
+
+const Byte* KTX::getKeyValueData() const {
+    if (_storage) {
+        return (_storage->_bytes + sizeof(Header));
+    } else {
+        return nullptr;
+    }
+}
--- a/libraries/ktx/src/ktx/KTX.cpp
+++ b/libraries/ktx/src/ktx/KTX.cpp
@@ -49,8 +49,8 @@ size_t Header::evalRowSize(uint32_t level) const {
     auto pixelWidth = evalPixelWidth(level);
     auto pixSize = evalPixelSize();
     auto netSize = pixelWidth * pixSize;
-    auto packing = netSize % 4;
-    return netSize + (packing ? 4 - packing : 0);
+    auto packing = netSize % PACKING_SIZE;
+    return netSize + (packing ? PACKING_SIZE - packing : 0);
 }
 size_t Header::evalFaceSize(uint32_t level) const {
     auto pixelHeight = evalPixelHeight(level);
@@ -83,9 +83,34 @@ const Header* KTX::getHeader() const {
     }
 }
 
+
+size_t KTX::getKeyValueDataSize() const {
+    if (_storage) {
+        return getHeader()->bytesOfKeyValueData;
+    } else {
+        return 0;
+    }
+}
+
+size_t KTX::getTexelsDataSize() const {
+    if (_storage) {
+        return  _storage->size() - sizeof(Header) + getKeyValueDataSize();
+    } else {
+        return 0;
+    }
+}
+
 const Byte* KTX::getKeyValueData() const {
     if (_storage) {
         return (_storage->_bytes + sizeof(Header));
+    } else {
+        return nullptr;
+    }
+}
+
+const Byte* KTX::getTexelsData() const {
+    if (_storage) {
+        return (_storage->_bytes + sizeof(Header) + getKeyValueDataSize());
     } else {
         return nullptr;
     }
--- a/libraries/ktx/src/ktx/KTX.cpp
+++ b/libraries/ktx/src/ktx/KTX.cpp
@@ -15,10 +15,11 @@
 
 using namespace ktx;
 
-uint32_t evalPadding(size_t byteSize) {
+uint32_t Header::evalPadding(size_t byteSize) {
     auto padding = byteSize % PACKING_SIZE;
-    return (padding ? PACKING_SIZE - padding : 0);
+    return (uint32_t) (padding ? PACKING_SIZE - padding : 0);
 }
+
 
 const Header::Identifier ktx::Header::IDENTIFIER {{
     0xAB, 0x4B, 0x54, 0x58, 0x20, 0x31, 0x31, 0xBB, 0x0D, 0x0A, 0x1A, 0x0A
@@ -29,7 +30,7 @@ Header::Header() {
 }
 
 uint32_t Header::evalMaxDimension() const {
-    return std::max(pixelWidth, std::max(pixelHeight, pixelDepth));
+    return std::max(getPixelWidth(), std::max(getPixelHeight(), getPixelDepth()));
 }
 
 uint32_t Header::evalMaxLevel() const {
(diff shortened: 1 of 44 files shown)


hifi-gustavo:
Build console output for commit: 0d28d17e40f7ef1e6c72d432e0ce1f5b66280e1e available at: https://gist.github.com/1821b0af3a2dd4b3a2a1e8ad6f13a698<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9688-0d28d17.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9688-0d28d17.dmg

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+#include <list>
+#include <QtGlobal>
+
+namespace ktx {
+    class ReaderException: public std::exception {
+    public:
+        ReaderException(std::string explanation) : _explanation(explanation) {}
+        const char* what() const override {
+            return ("KTX deserialization error: " + _explanation).c_str();
```
jherico: This is invalid.  `c_str` is returning a pointer to a temporary.

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,186 @@
...
+//
+#include "KTX.h"
+
+#include <list>
+#include <QtGlobal>
+
+namespace ktx {
+    class ReaderException: public std::exception {
+    public:
+        ReaderException(std::string explanation) : _explanation(explanation) {}
```
jherico: should be `const std::string&`

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,186 @@
...
+        const char* what() const override {
+            return ("KTX deserialization error: " + _explanation).c_str();
+        }
+    private:
+        std::string _explanation;
```
jherico: should be `const std::string`

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,186 @@
...
+            }
+            break;
+        case Header::REVERSE_ENDIAN_TEST:
+            {
+                matching = false;
+                return true;
+            }
+            break;
+        default:
+            return false;
```
jherico: this should throw an exception.  If the endianness check doesn't match either value, then the file is invalid.

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,186 @@
...
+        }
+    }
+
+    bool checkIdentifier(const Byte* identifier) {
+        return memcmp(identifier, Header::IDENTIFIER.data(), Header::IDENTIFIER_LENGTH);
```
jherico: implicit cast from int to bool.  Also the logic seems to be reversed. should be `return (0 == memcmp(...)) ? true : false'

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    KeyValues getKeyValues(size_t length, const Byte* src) {
```
jherico: should probably be a static member of `KTX`.  Also I find you end up with more flexible functions if you declare things like this as `KeyValues getKeyValues(size_t length, const Byte* src, size_t offset = 0)`, or even `size_t getKeyValues(KeyValues& outValues, size_t length, const Byte* src, size_t offset = 0)` where the return value is the new offset after parsing the content.

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,186 @@
...
+            // advance offset/src
+            uint32_t keyValuePadding = 3 - ((keyValueByteSize + 3) % PACKING_SIZE);
+            offset += keyValueByteSize + keyValuePadding;
+            src += keyValueByteSize + keyValuePadding;
+        }
+
+        return keyValues;
+    }
+
+    Images getImagesTable(const Header& header, size_t mipsDataSize, const Byte* mipsData) {
```
jherico: same commentary...  consider `size_t getImagesTable(Images& outImages, const Header& header, size_t mipsDataSize, const Byte* mipsData, size_t offset = 0)`

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,186 @@
...
+            } else {
+                break;
+            }
+        }
+
+        return images;
+    }
+
+
+    bool KTX::checkHeaderFromStorage(const Storage& src) {
```
jherico: The Storage abstraction here is unnecessary, since the only thing we do with it is immediately convert it to a pointer and size.  consider `bool KTX::checkHeaderFromStorage(size_t length, void* data, size_t offset)`

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,186 @@
...
+            auto srcBytes = src.data();
+
+            // validation
+            if (srcSize < sizeof(Header)) {
+                throw ReaderException("length is too short for header");
+            }
+            const Header* header = reinterpret_cast<const Header*>(srcBytes);
+
+            if (!checkIdentifier(header->identifier)) {
+                throw ReaderException("identifier field invalid");
```
jherico: consider moving exception to function.

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,186 @@
...
+            }
+
+            bool endianMatch { true };
+            if (!checkEndianness(header->endianness, endianMatch)) {
+                throw ReaderException("endianness field has invalid value");
```
jherico: consider moving exception to function

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,186 @@
...
+            }
+
+            // TODO: endian conversion if !endianMatch - for now, this is for local use and is unnecessary
+
```
jherico: for now, if !endianMatch, we should throw an exception

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,186 @@
...
+            return true;
+        }
+        catch (ReaderException& e) {
+            qWarning(e.what());
+            return false;
+        }
+    }
+
+    std::unique_ptr<KTX> KTX::create(const Storage& src) {
+        auto srcCopy = std::make_unique<Storage>(src);
```
jherico: doesn't this create a deep copy of the `Storage` object?

libraries/ktx/src/ktx/Writer.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "KTX.h"
+
+
+namespace ktx {
+
+    class WriterException : public std::exception {
+    public:
+        WriterException(std::string explanation) : _explanation(explanation) {}
+        const char* what() const override {
+            return ("KTX serialization error: " + _explanation).c_str();
```
jherico: returning a pointer to a temporary, invalid.  See comment on ReaderException

libraries/ktx/src/ktx/Writer.cpp
```diff
@@ -0,0 +1,100 @@
...
+        }
+    private:
+        std::string _explanation;
+    };
+
+    std::unique_ptr<Storage> generateStorage(const Header& header, const KeyValues& keyValues, const Images& images) {
```
jherico: KeyValues is optional, so it should be last, and default to an empty object

libraries/ktx/src/ktx/Writer.cpp
```diff
@@ -0,0 +1,100 @@
...
+        size_t storageSize = sizeof(Header);
```
jherico: The calculation of the size of a KTX buffer should be abstracted out to it's own function.  This will enable coding like this

```
   size_t ktxSize = calculateSize(header, images);
   QFile f(...); // create or open file and set required size
   uint8_t* mapped = f.map(...);
   writeKtx(mapped, header, images);
```

This code is optimal because it doesn't require any intermediate memory... it writes directly from the source data to the file.

libraries/model/src/model/TextureMap.cpp
```diff
@@ -264,8 +266,23 @@ gpu::Texture* TextureUsage::process2DTextureColorFromImage(const QImage& srcImag
         if (generateMips) {
             ::generateMips(theTexture, image, formatMip, false);
         }
+        auto theKTX = Texture::serialize(*theTexture);
+        if (theKTX) {
+            // save that!
+            std::string filename("C://temp//ktxCache//texmex");
```
jherico: this should be using a path relative to something returned from `QStandardPaths::writableLocation`

libraries/ktx/src/ktx/Reader.cpp
samcake: yes
THat s why there is the second version right after where it s move

libraries/model/src/model/TextureMap.cpp
samcake: right this is just to try to spawn something

libraries/ktx/src/ktx/Reader.cpp
samcake: got it

libraries/ktx/src/ktx/Writer.cpp
samcake: got it

libraries/ktx/src/ktx/Writer.cpp
samcake: ok

libraries/ktx/src/ktx/KTX.h
```diff
@@ -0,0 +1,469 @@
...
+        Byte identifier[IDENTIFIER_LENGTH];
+        uint32_t endianness { ENDIAN_TEST };
+
+        uint32_t glType;
+        uint32_t glTypeSize { 0 };
+        uint32_t glFormat;
+        uint32_t glInternalFormat;
+        uint32_t glBaseInternalFormat;
+
+        uint32_t pixelWidth { 0 };
```
jherico: I'd suggest using a glm::uvec3 for the dimensions.  This will provide some benefits elsewhere.

libraries/ktx/src/ktx/KTX.cpp
```diff
@@ -0,0 +1,135 @@
...
+const Header::Identifier ktx::Header::IDENTIFIER {{
+    0xAB, 0x4B, 0x54, 0x58, 0x20, 0x31, 0x31, 0xBB, 0x0D, 0x0A, 0x1A, 0x0A
+}};
+
+Header::Header() {
+    memcpy(identifier, IDENTIFIER.data(), IDENTIFIER_LENGTH);
+}
+
+uint32_t Header::evalMaxDimension() const {
+    return std::max(getPixelWidth(), std::max(getPixelHeight(), getPixelDepth()));
```
jherico: if you store the dimensions in a uvec3 you can just return `glm::compMax(getPixelDimensions())`

libraries/ktx/src/ktx/KTX.cpp
```diff
@@ -0,0 +1,135 @@
...
+}
+
+uint32_t Header::evalMaxLevel() const {
+    return 1 + log2(evalMaxDimension());
+}
+
+uint32_t Header::evalPixelWidth(uint32_t level) const {
```
jherico: if you store the dimensions in a uvec3 you can replace this with 
```
glm::uvec3 Header::evalPixelDimensions(uint32_t level) const {
   glm::max(getPixelDimensions() >> level, glm::uvec3(1))
}
```

hifi-gustavo:
Build console output for commit: 2f7181fb322407d6b9fe0f7cb84c684a0036cf97 available at: https://gist.github.com/1821b0af3a2dd4b3a2a1e8ad6f13a698<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9688-2f7181f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9688-2f7181f.dmg

libraries/ktx/src/ktx/KTX.h
```diff
@@ -0,0 +1,469 @@
...
+        POS_Y = 2,
+        NEG_Y = 3,
+        POS_Z = 4,
+        NEG_Z = 5,
+        NUM_CUBEMAPFACES = 6,
+    };
+
+
+    // Chunk of data
+    struct Storage {
```
jherico: This storage class assumes ownership of the data.  I think we should drop it, but at the very least it shouldn't be a member of the KTX class.  The point of this is to allow us to read directly from memory mapped files.  Having a KTX have a bunch of heap storage to contain the data defeats that and makes the KTX class no less memory consuming than our current strategy.

libraries/ktx/src/ktx/KTX.h
```diff
@@ -0,0 +1,469 @@
...
+
+        Byte* data() {
+            return _bytes;
+        }
+        const Byte* data() const {
+            return _bytes;
+        }
+        size_t size() const { return _size; }
+
+        ~Storage() { if (_bytes) { delete _bytes; } }
```
jherico: if the Storage backing is from a memory mapped file, this will cause a problem.  You can't simply delete a memory mapped pointer.

samcake:
With the new calls you cam do with no ownership
I want that option

libraries/ktx/src/ktx/KTX.h
samcake: YEs but with the new static calls you can do it with pure functions.
I still want to keep the option though

libraries/ktx/src/ktx/KTX.h
samcake: yes but i m not going to use anything external (aside from standard) so that lib can become an external of it s own.

hifi-gustavo:
Build console output for commit: 8ee5defc6086a2cd2165b8e96672f489f965f40e available at: https://gist.github.com/1821b0af3a2dd4b3a2a1e8ad6f13a698<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9688-8ee5def.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9688-8ee5def.dmg

hifi-gustavo:
Build console output for commit: 2ee3f05713722d134b820d7cd69854c35df872b2 available at: https://gist.github.com/1821b0af3a2dd4b3a2a1e8ad6f13a698<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9688-2ee3f05.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9688-2ee3f05.dmg

hifi-gustavo:
Build console output for commit: 56d11206541c93fef18eecd9fd4ce4d9e1fc28d3 available at: https://gist.github.com/1821b0af3a2dd4b3a2a1e8ad6f13a698<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9688-56d1120.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9688-56d1120.dmg

libraries/ktx/src/ktx/Writer.cpp
```diff
@@ -0,0 +1,144 @@
...
+        result->resetImages(images);
+
+        return result;
+    }
+
+    size_t KTX::evalStorageSize(const Header& header, const Images& images, const KeyValues& keyValues) {
+        size_t storageSize = sizeof(Header);
+
+        if (header.bytesOfKeyValueData && !keyValues.empty()) {
+
```
jherico: missing code?

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,171 @@
...
+            break;
+        case Header::REVERSE_ENDIAN_TEST:
+            {
+                matching = false;
+                return true;
+            }
+            break;
+        default:
+            throw ReaderException("endianness field has invalid value");
+            return false;
```
jherico: unreachable code will generate warnings which will fail the build

libraries/ktx/src/ktx/Reader.cpp
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+
+    bool checkIdentifier(const Byte* identifier) {
+        if (!(0 == memcmp(identifier, Header::IDENTIFIER.data(), Header::IDENTIFIER_LENGTH))) {
+            throw ReaderException("identifier field invalid");
+            return false;
```
jherico: unreachable code will generate warnings which will fail the build

libraries/ktx/src/ktx/KTX.h
jherico: if you want to keep this model, then Storage needs to become pure virtual, with a one version for maintaining the contents in memory and another version for maintaining the contents as a memory mapped file.

libraries/ktx/src/ktx/KTX.cpp
samcake: I wont include glm in this code to keep it slim and easily to separate in its own repository

libraries/ktx/src/ktx/KTX.cpp
samcake: I wont include glm in this code to keep it slim and easily to separate in its own repository

samcake:
THis code is ogoing into our smarter_textures branch to upstream

PR #9718, opened 2017-02-21
fix sorting algorithm for avatar render updates

AndrewMeadows:


interface/src/avatar/AvatarManager.cpp
```diff
@@ -154,8 +154,7 @@ class AvatarPriority {
     AvatarPriority(AvatarSharedPointer a, float p) : avatar(a), priority(p) {}
     AvatarSharedPointer avatar;
     float priority;
-    // NOTE: we invert the less-than operator to sort high priorities to front
-    bool operator<(const AvatarPriority& other) const { return priority > other.priority; }
+    bool operator<(const AvatarPriority& other) const { return priority < other.priority; }
```
AndrewMeadows: The first problem: the sort order was wrong.

interface/src/avatar/AvatarManager.cpp
```diff
@@ -206,9 +205,8 @@ void AvatarManager::updateOtherAvatars(float deltaTime) {
             float radius = avatar->getBoundingRadius();
             const glm::vec3& forward = cameraView.getDirection();
             float apparentSize = radius / distance;
-            float cosineAngle = glm::length(offset - glm::dot(offset, forward) * forward) / distance;
-            const float TIME_PENALTY = 0.080f; // seconds
-            float age = (float)(startTime - avatar->getLastRenderUpdateTime()) / (float)(USECS_PER_SECOND) - TIME_PENALTY;
+            float cosineAngle = glm::length(glm::dot(offset, forward) * forward) / distance;
```
AndrewMeadows: The second problem: was computing sineAngle rather than cosineAngle... which is why I thought I needed to invert the sort order.

hifi-gustavo:
Build console output for commit: 27dd82b517c31bb3b65b6e2d07791688e4017a9e available at: https://gist.github.com/951142224c4fdebc6ebb9aaf59dd1d83<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9718-27dd82b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9718-27dd82b.dmg

hifi-gustavo:
Build console output for commit: dd447689e0cefb603814fc019e0dd87e213fcb88 available at: https://gist.github.com/951142224c4fdebc6ebb9aaf59dd1d83<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9718-dd44768.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9718-dd44768.dmg

interface/src/avatar/AvatarManager.cpp
```diff
@@ -205,15 +204,17 @@ void AvatarManager::updateOtherAvatars(float deltaTime) {
             float distance = glm::length(offset) + 0.001f; // add 1mm to avoid divide by zero
             float radius = avatar->getBoundingRadius();
             const glm::vec3& forward = cameraView.getDirection();
-            float apparentSize = radius / distance;
-            float cosineAngle = glm::length(offset - glm::dot(offset, forward) * forward) / distance;
-            const float TIME_PENALTY = 0.080f; // seconds
-            float age = (float)(startTime - avatar->getLastRenderUpdateTime()) / (float)(USECS_PER_SECOND) - TIME_PENALTY;
+            float apparentSize = 2.0f * radius / distance;
+            float cosineAngle = glm::length(glm::dot(offset, forward) * forward) / distance;
+            float age = (float)(startTime - avatar->getLastRenderUpdateTime()) / (float)(USECS_PER_SECOND);
```
AndrewMeadows: The second bug was that I was actually computing the `sineAngle` value rather than the `cosineAngle`.  This would tend to cancel out the first bug when all the avatars were far away (when everyone's `apparentSize` term was small).

I removed the TIME_PENALTY coefficient: it was too complicated and was tuned wrong for the new and improved AvatarMixer.

The `2.0f` coefficient that shows up in the `apparentSize` calculation was me making a hidden coefficient more obvious in an attempt to make tuning the new `_avatarSortCoefficient*` variables easier.

interface/src/avatar/AvatarManager.h
```diff
@@ -120,6 +124,11 @@ private slots:
     int _partiallySimulatedAvatars { 0 };
     float _avatarSimulationTime { 0.0f };
 
+    // TODO: remove this HACK once we settle on optimal sort coefficients
+    // These coefficients exposed for fine tuning the sort priority for transfering new _jointData to the render pipeline.
+    float _avatarSortCoefficientSize { 0.5f };
+    float _avatarSortCoefficientCenter { 0.25 };
+    float _avatarSortCoefficientAge { 1.0f };
```
AndrewMeadows: These coefficients are exposed to JS so that we can more easily tune the sort prioritization.  Eventually we'll hard code these coefficients when we have a better idea about what they should really be.

AndrewMeadows:
There were two bugs in the prioritization of "updates to the avatar render geometry" which would cause nearby avatars to not necessarily update in a timely manner when there were a lot of other avatars in the scene.

This is probably a good recipe for reproducing the problem:

(1) Login to dev-chris2.highfidelity.io -- there should be about 100 scripted test avatars inside the meeting space.

(2) Navigate such that your avatar is near one of the outer edge avatars, with all other avatars located further away behind.

(3) Notice how frequently the nearest avatar updates its joint positions:
--> will be updated at a very low rate for an older (bugged) interface, maybe even slower than avatars behind it.
--> will be updated at a faster rate for this PR, should be faster than other avatars behind and off to the side.

interface/src/avatar/AvatarManager.cpp
```diff
@@ -205,15 +204,17 @@ void AvatarManager::updateOtherAvatars(float deltaTime) {
             float distance = glm::length(offset) + 0.001f; // add 1mm to avoid divide by zero
             float radius = avatar->getBoundingRadius();
             const glm::vec3& forward = cameraView.getDirection();
-            float apparentSize = radius / distance;
-            float cosineAngle = glm::length(offset - glm::dot(offset, forward) * forward) / distance;
-            const float TIME_PENALTY = 0.080f; // seconds
-            float age = (float)(startTime - avatar->getLastRenderUpdateTime()) / (float)(USECS_PER_SECOND) - TIME_PENALTY;
+            float apparentSize = 2.0f * radius / distance;
+            float cosineAngle = glm::length(glm::dot(offset, forward) * forward) / distance;
```
PhilipRosedale: Why is this not cos(angleBetween)?  is this some sort of simplification?

PR #9720, opened 2017-02-22
Filter pal by view distance

howard-stearns:
There is now an "in view" checkbox: when checked (or when refreshed while checked), the PAL only shows avatars that are within 30m and in your view. Otherwise PAL attempts to show all in the domain.

Also relocates the refresh button to the upper right corner, and uses a hifi-standard button.

Also doesn't attempt to display avatars that have no identification (as you can't reliably do anything with them anyway).  This can happen, e.g., when completely overloaded and you don't get an identity packet in 5 seconds and so the avatar is removed, but then you do get a data packet.


howard-stearns:
test plan:
- Make sure the filter and refresh work as expected given the above description.  (The definition of in view may be a little broader than your actual frustum. But it should be reasonably close.)  So for example, go to a place with several avatars and set yourself up so that some of them are in view. Then toggle filter on or off.
- Be sure to check in HDM and desktop, and in places where you have kick rights vs where you don't.

howard-stearns:
One issue so far: the UI machinery now constantly recreates new .qml objects, so state is lost each time you open/close it. Have to think about what the right behavior is...

hifi-gustavo:
Build console output for commit: a20590a2f953e06601a6c6dbf6ca0e1fbefa9322 available at: https://gist.github.com/d7feceb9a8274df2f529c526a0d734ea<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9720-a20590a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9720-a20590a.dmg

howard-stearns:
The settings issues have been worked around, and styling changes made. Should be good to go.

hifi-gustavo:
Build console output for commit: de3a0d3d0c7c5504068d7237139459a84e0e440e available at: https://gist.github.com/d7feceb9a8274df2f529c526a0d734ea<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9720-de3a0d3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9720-de3a0d3.dmg

zfox23:
* The checkbox, checkbox caption, and refresh button are top-vertically-aligned. Is this intended, or just a side effect of not being explicit? It looks strange to me, and feels like it should be center-vertically-aligned. 
* Also, should the checkbox next to "in view" be the same size as the other PAL checkboxes?
* Also also, should "in view" be ALLCAPS? @AlanZimmerman?

image

howard-stearns:
(I had the same questions.) Alan reviewed the design, with top alignment and checkbox size particularly called out for attention. (We discussed the spacing, but not particularly the allcap-ness of the "in view". However, it is using our standard checkbox control, so if it's wrong, it's wrong in our standard.)

zfox23:
OK, cool. Still looks a little strange to me, but I'll hold my tongue and surely get used to it. 😄 

FYI, WRT the size: I added a property to the HiFi standard checkbox control that defines its size, and overrode that size for the other checkboxes in the PAL. Alan approved that decision.

hifi-gustavo:
Build console output for commit: cddba438372c28beee3f42dac0b417db1c98c178 available at: https://gist.github.com/d7feceb9a8274df2f529c526a0d734ea<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9720-cddba43.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9720-cddba43.dmg

howard-stearns:
test this please

hifi-gustavo:
Build console output for commit: cddba438372c28beee3f42dac0b417db1c98c178 available at: https://gist.github.com/d7feceb9a8274df2f529c526a0d734ea<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9720-cddba43.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9720-cddba43.dmg

Nex-Pro:
@howard-stearns If it is ok to ask, why a seemingly arbitrary 30m range ?  Why not 'Within earshot'.

Nex-Pro:
Testing this.

Nex-Pro:
Awaiting fresh server launch

Nex-Pro:
Issues:

Pal indicator sphere does not illuminate sufficiently for avatar selected in pal.( might be a previously seen graphical issue.)

Unable to edit my name on the PAL plus display name does not show.

Generally in desktop this seems to work.

Had some odd white outlining of PAL spheres.

After switching to HMD I could see zero avatars ( a couple of minutes ) When I opened the PAL they all rezzed.

This was repeatable.

Basically PAL is the on/off switch for displaying avatars.

Compared this to Master master generally displayed the avatars in HMD.

howard-stearns:
recording some discussion re questions above:
1. Not really sound based. This isn't so much a thing about matching up with volume as a quick "just show me people right in front of me in the list".  "right in front" is somewhat arbitrary, of course, (Separately, there's a PR in development to sort by average loudness.)
2. The 30m is a setting, so wizards have an out if this turns out to be a bad choice, but within the scope of #1, it seems like it's a good place to start without confusing the user with distances. (And confusing poor USians with the metric system.)

zfox23:
@Nex-Pro great catch re: "Basically PAL is the on/off switch for displaying avatars.". This is a known, filed bug that I'm currently working on (in coincidental conjunction with @ZappoMan) that isn't related to this PR.

howard-stearns:
@Nex-Pro  can you give me steps to repro the own-display-name problems. I was able to both edit and display my name, using both desktop and hmd. (Of course, the virtual keyboard wasn't much fun to use, particularly with 100 avatars present, but that's nothing to do with this PR.)

howard-stearns:
The expected/known behavior for the avatar nodes is this:
- If  there are less than 20 names displayed, each corresponding avatar should have both a cloud-and-centered-circle, billboarded to you, and a sphere.  None of these are visible in low light (which is rendering bug https://highfidelity.fogbugz.com/f/cases/3176/overlay-materials-and-effects).  The sphere part should be visible through walls and  other obstructions.  The sphere should be green when not selected, and (except for agents) should vary intensity with volume. Both the sphere and the circle-in-the-cloud should have hover and  an orangish/circled selected color.
- If there are more avatars shown, only the sphere is displayed.

Is that what you are seeing?

howard-stearns:
ok, I think we need to move forward, so I'm going to assume:
1. the in-world avatar node stuff is as per design with our current limitations ("expected/known" comment above), are not from this pr per se, and we can open a separate if needed.
2. display-name issues are not-repro  or from this PR and we should open separate ticket if there's a repro.
3. the on/off comment is as zfox says.

So, calling it a pass and let's open tickets as needed.

PR #9724, opened 2017-02-22
expose avatar update time budget to JS config

AndrewMeadows:


interface/src/avatar/AvatarManager.cpp
```diff
@@ -226,12 +226,14 @@ void AvatarManager::updateOtherAvatars(float deltaTime) {
...
     }
 
-    render::PendingChanges pendingChanges;
-    const uint64_t RENDER_UPDATE_BUDGET = 1500; // usec
-    const uint64_t MAX_UPDATE_BUDGET = 2000; // usec
-    uint64_t renderExpiry = startTime + RENDER_UPDATE_BUDGET;
+    // reset startTime and and compute expirys
+    startTime = usecTimestampNow();
+    uint64_t renderExpiry = startTime + _avatarUpdateTimeBudget;
+    const uint64_t MAX_UPDATE_BUDGET = _avatarUpdateTimeBudget + 500UL; // usec
```
AndrewMeadows: I reset `startTime` so that the sort operation is outside the time budget.  There was a risk the sort itself could consume all or most of the avatar update budget, given enough avatars on low-end platforms, however completing the sort is non-negotiable so it makes sense to pull it out of the time budget measurement.

I was seeing about 200 - 275 usec spent in the sort on my i7 in the 100-avatar case, so I've dropped the default time budget from 1500 usec down to 1300 (see where `_avatarUpdateTimeBudget` is initialized).  The absolute cuttoff is hard coded to 500 usec beyond that (in declaration of `MAX_UPDATE_BUDGET`).

interface/src/avatar/AvatarManager.h
```diff
@@ -126,6 +128,7 @@ private slots:
 
     // TODO: remove this HACK once we settle on optimal sort coefficients
     // These coefficients exposed for fine tuning the sort priority for transfering new _jointData to the render pipeline.
+    uint64_t _avatarUpdateTimeBudget { 1300 }; // usec
```
AndrewMeadows: `_avatarUpdateTimeBudget` is exposed to JS via the Q_INVOKABLE methods above in case anyone wants to play with it.  @ZappoMan, that is for you.

hifi-gustavo:
Build console output for commit: edc1fc0f1a8eeaf9ccdfd864fd3632eacce0f93c available at: https://gist.github.com/06578b2e2f8d4d95164bc2a0869163da<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9724-edc1fc0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9724-edc1fc0.dmg

AndrewMeadows:
Not much has changed in this PR except the time budget for handling avatar joint updates has been exposed to JS which may be helpful when devs are tuning things for large numbers of avatars.

**To test:**

**(1)** Login to a 100-avatar test domain: **dev-chris.highfidelity.io** or **dev-chris2.highfidelity.io**.

**(2)** With most of the scripted test avatars in view pay attention to how often the avatars appear to update their joint orientations.

**(3)** Run this on-line script to decrease the time available for processing avatar updates:

`AvatarManager.setAvatarUpdateTimeBudget(100);`

Pay attention to how often the avatars appear to update their joint orientations: we would expect them to be even more sluggish with their time budget reduced.

**(4)** You could try expanding the time budget above the default (1300 usec) with this script:

`AvatarManager.setAvatarUpdateTimeBudget(3000);`

The avatars should update at least as fast as case **(2)** but there will probably not be a noticeable improvement since there are other server-side factors that may limit avatar update rates.

hifi-gustavo:
Build console output for commit: 259cb6eaf197f8db829db15952e9cd7f6a407d0f available at: https://gist.github.com/06578b2e2f8d4d95164bc2a0869163da<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9724-259cb6e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9724-259cb6e.dmg

interface/src/avatar/AvatarManager.cpp
ZappoMan: I think this budget is way too small... what sort of actual "simulation" times are you seeing for 100 avatars in the scale testing?

interface/src/avatar/AvatarManager.cpp
```diff
@@ -596,7 +598,14 @@ RayToAvatarIntersectionResult AvatarManager::findRayIntersection(const PickRay&
 }
 
 // HACK
-float AvatarManager::getAvatarSortCoefficient(const QString& name) {
+void AvatarManager::setAvatarUpdateTimeBudget(uint32_t timeBudget) {
+    const uint32_t MIN_AVATAR_UPDATE_TIME_BUDGET = 0;
+    const uint32_t MAX_AVATAR_UPDATE_TIME_BUDGET = 11000U;
```
ZappoMan: what are the units here? usecs? msecs? comments please

interface/src/avatar/AvatarManager.cpp
AndrewMeadows: For the 100-avatar case on my i7 I'm seeing:

~1650 usec to sort and check 100 avatars that don't have any new _jointData.  The sort takes about 283 usec of that on average, so... about 1370 usec just too walk the list and do various per-frame checks that are not about copying the _jointData to the Rig **(1)**.  Note: that the true budget for walking this list is actually 1800 usec (MAX_UPDATE_BUDGET = 1300 + 500).

So the default budget as set may be too small, but I wouldn't characterize it as "way too small".  CPU cycles are going to be tight all around as we try to scale up avatar counts and render lots of entities and do physics.

The good news is that this PR would be opening up that budget for experimental adjustment and we could set a hard-coded default to a more correct value after experimentation.

**(1)** I think I could eliminate much of the overhead per-frame checks of the Avatars for various operations that might need to be done (e.g. does its model need to be loaded? does it have any attachments that need to be updated?  how about "avatar entities"?)... but my ideas would require us to maintain per-operation lists: I would describe the work as a "refactor".

hifi-gustavo:
Build console output for commit: 28970f442736af2705b1c09ff393c8c86ceed7a7 available at: https://gist.github.com/06578b2e2f8d4d95164bc2a0869163da<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9724-28970f4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9724-28970f4.dmg

AndrewMeadows:
Closing this PR.  Maye we'll make some of these changes later.

PR #9726, opened 2017-02-22
Avatar bandwidth budget improvements

ZappoMan:
This is the beginning of sharing the avatar priority sort used in simulation logic with the avatar mixer for bandwidth budgeting...

This PR includes:
- moving the actual `sortAvatars` method to be a static of `AvatarData` . I'm open to suggestions for a better place for this... this was someplace that is common to the mixer and the client. It may not actually need access to the AvatarData class... so maybe putting it in it's own namespace would be more OO friendly (data hiding from `AvatarData` members)

- adds a message between client and avatar-mixer when the client uses the priority tweak nobs (this is temporary only for debugging/fine-tuning)...

- uses the avatar sort order to handle bandwidth budget

- stops sending avatar data in the frame once bandwidth budget is reached, this results in never going over budget and as a result the "distance based random drops" don't kick in.

**Test Plan:**
- Smoke test with this sandbox/domain and a couple of different agents... you must test multi-user support.
- Make sure ignore still works
- Make sure when you turn away from avatars that your inbound avatar mixer bandwidth goes down
- Make sure that when you turn away from avatars they don't "disappear from your avatar list"
- Load test with more than 20 avatars... make sure that animation still seems reasonable.
    - when you get above ~15 avatars, the 5MB bandwidth budget is insufficient to send all joints at 45hz, so the system has to start prioritizing
    - the current master code uses a random distance based technique to drop avatars...
    - this build will prioritize avatars by apparent size, closeness to the center of your view, and time since last update
    - the overall expected result is that you should see everyone moving (they shouldn't disappear and/or freeze for long periods of time)
    - generally things closer and more in the center of your view should have a higher update rate than things further and off to the side of your view


hifi-gustavo:
Build console output for commit: 7cc32f555e57a2a35744411cf00e6b0552de96d3 available at: https://gist.github.com/dd7675eb92eb90ed231d3b406b08e1d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-7cc32f5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-7cc32f5.dmg

hifi-gustavo:
Build console output for commit: 10f465f8eaeef155405d2bae4d0e17f6166e87ff available at: https://gist.github.com/dd7675eb92eb90ed231d3b406b08e1d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-10f465f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-10f465f.dmg

hifi-gustavo:
Build console output for commit: 4c4506b1f7c12f23a3c3401c9b15cf1641806220 available at: https://gist.github.com/dd7675eb92eb90ed231d3b406b08e1d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-4c4506b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-4c4506b.dmg

hifi-gustavo:
Build console output for commit: edf7c016a12cd51ac5892d495a518142c44233e7 available at: https://gist.github.com/dd7675eb92eb90ed231d3b406b08e1d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-edf7c01.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-edf7c01.dmg

hifi-gustavo:
Build console output for commit: 06f0087459a56431f40cd6ec72f7501303b9c36d available at: https://gist.github.com/dd7675eb92eb90ed231d3b406b08e1d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-06f0087.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-06f0087.dmg

hifi-gustavo:
Build console output for commit: 40037bee55d063ae1f81c3811e9c7623370f37c5 available at: https://gist.github.com/dd7675eb92eb90ed231d3b406b08e1d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-40037be.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-40037be.dmg

hifi-gustavo:
Build console output for commit: 4f17498daf5c22d1e67e302d6b77ae248a5d8e49 available at: https://gist.github.com/dd7675eb92eb90ed231d3b406b08e1d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-4f17498.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-4f17498.dmg

hifi-gustavo:
Build console output for commit: dde9640c66a9be73e8c65c8d87a3b12d22f6a787 available at: https://gist.github.com/dd7675eb92eb90ed231d3b406b08e1d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-dde9640.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-dde9640.dmg

hifi-gustavo:
Build console output for commit: f61e16fccfe901116ffd75e56608c0be4ac7a6b5 available at: https://gist.github.com/dd7675eb92eb90ed231d3b406b08e1d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-f61e16f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-f61e16f.dmg

assignment-client/src/avatars/AvatarMixerSlave.cpp
```diff
@@ -147,49 +153,82 @@ void AvatarMixerSlave::broadcastAvatarData(const SharedNodePointer& node) {
...
+        AvatarSharedPointer thisAvatar = nodeData->getAvatarSharedPointer();
+        ViewFrustum cameraView = nodeData->getViewFrustom();
+        std::priority_queue<AvatarPriority> sortedAvatars = AvatarData::sortAvatars(
+                avatarList, cameraView,
 
-                // we're adjusting the full rate distance to target a bandwidth in the middle
-                // of the hysterisis gap
-                currentFullRateDistance *= (_maxKbpsPerNode * HYSTERISIS_MIDDLE_PERCENTAGE) / avatarDataRateLastSecond;
+                [&](AvatarSharedPointer avatar)->uint64_t{
+                    auto avatarNode = avatarDataToNodes[avatar];
```
AndrewMeadows: Accessing the map element by key will insert into the map if the key doesn't exist (to look without insert use `find()`)... which means the check for valid `avatarNode` below would always succeed... unless someone already inserted a shared pointer with a `nullptr` data... which maybe the access above just did?

assignment-client/src/avatars/AvatarMixerSlave.cpp
```diff
@@ -203,7 +242,6 @@ void AvatarMixerSlave::broadcastAvatarData(const SharedNodePointer& node) {
                 || (otherNode->isIgnoringNodeWithID(node->getUUID()) && !getsAnyIgnored)) {
 
                 shouldConsider = false;
-
             } else {
                 const AvatarMixerClientData* otherData = reinterpret_cast<AvatarMixerClientData*>(otherNode->getLinkedData());
```
AndrewMeadows: For the case where maybe the avatars should be ignoring each other (below this code)... `nodeBox` and `otherNodeBox` calculations are redone for every check.  This seems inefficient O(N^2).  It could be O(N) if each avatar's `box` were precomputed.

The overlap check is still O(N^2) of course.

assignment-client/src/avatars/AvatarMixerSlave.cpp
ZappoMan: in this case though, if it returns nullptr that's acceptable, since we check for that.

Basically, `find()` is about as expensive as `[]` and so instead of doing the `find()` and `[]` cost twice for the most likely case: that it's in there... this approach seems preferable.

ZappoMan: er... wait... rethinking... actually I don't think we can ever get here without having previously storing the key in the map....

hifi-gustavo:
Build console output for commit: 8cb8d686ecf33798645243e11b7c7e1c9cc68993 available at: https://gist.github.com/dd7675eb92eb90ed231d3b406b08e1d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-8cb8d68.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-8cb8d68.dmg

hifi-gustavo:
Build console output for commit: 1e91f74ce745929d6a502f7f7677640e64b617d6 available at: https://gist.github.com/dd7675eb92eb90ed231d3b406b08e1d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-1e91f74.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-1e91f74.dmg

hifi-gustavo:
Build console output for commit: 1b7ec7d61e445e20c9207effed8515527148e66c available at: https://gist.github.com/dd7675eb92eb90ed231d3b406b08e1d8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-1b7ec7d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9726-1b7ec7d.dmg

assignment-client/src/avatars/AvatarMixerSlave.cpp
```diff
@@ -147,266 +146,263 @@ void AvatarMixerSlave::broadcastAvatarData(const SharedNodePointer& node) {
...
+        AvatarSharedPointer thisAvatar = nodeData->getAvatarSharedPointer();
+        ViewFrustum cameraView = nodeData->getViewFrustom();
+        std::priority_queue<AvatarPriority> sortedAvatars = AvatarData::sortAvatars(
+                avatarList, cameraView,
+
+                [&](AvatarSharedPointer avatar)->uint64_t{
+                    auto avatarNode = avatarDataToNodes[avatar];
+                    assert(avatarNode); // we can't have gotten here without the avatarData being a valid key in the map
+                    return nodeData->getLastBroadcastTime(avatarNode->getUUID());
+                },
```
AndrewMeadows: This lambda is used to compute the "age" of the data that needs to go out, but it looks to me like it is using a global `LastBroadcastTime` which is a property of the node who owns the data.  The `LastBroadcastTime` is updated as soon as the first packet goes out carrying that data... which means subsequent recipients will be using a very recent (meaningless) time value in their sorting algorithm... e.g. based on the time when the update was sent to the previous recipient.

Unfortunately for accurate "time since foo was last sent to bar" measurement we would have to maintain more **state**: for each recipient we need a map of `<otherAvatarID, time>` pairs.

PR #9758, opened 2017-02-25
Fix Password Setting Bug

kunalgosar:
Resolution to: https://highfidelity.fogbugz.com/f/cases/3227/Unable-to-set-password-for-domain-server-web-control-interface

The form was comparing a hashed version of password and a non-hashed version of verified_password, which means that they would never match. Also it was saving a double hash of the password.

Testing instructions:
- Start domain server, and go to security
- Set a username and password
- Verify that it does not error saying mismatched passwords
- Verify that you can log in with username/password

### DIFF
--- a/domain-server/resources/web/settings/js/settings.js
+++ b/domain-server/resources/web/settings/js/settings.js
@@ -996,6 +996,10 @@ function saveSettings() {
       if (password && password.length > 0) {
         formJSON["security"]["http_password"] = sha256_digest(password);
       }
+      var verify_password = formJSON["security"]["verify_http_password"];
+      if (verify_password && verify_password.length > 0) {
+        formJSON["security"]["verify_http_password"] = sha256_digest(verify_password);
+      }
     }
 
     // verify that the password and confirmation match before saving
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: a4e0d970bf68c38c73968d5b777a7039f550c5d6 available at: https://gist.github.com/2ab38ab340849ead5ba1594264b11d7e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9758-a4e0d97.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9758-a4e0d97.dmg

nbq:
Testing this

nbq:
Tester: Coal

* Clean install
* Started domain-server and went to security
* Set a username and password, first tried to mismatch the passwords on purpose, then put in the correct one in both fields and verified it saved
* Was able to save just fine with the correct password
* Was prompted for the password on page reload and was able to log in just fine with the set password and username

**SUCCESS PASS**

PR #9789, opened 2017-02-28
Remember gain settings when pal is closed and reopened

davidkelly:
When you adjust an avatar's gain in the PAL and then close the PAL, that gain setting continues to be used.  However the PAL had no way of 'remembering' it, and if you reopen the PAL the slider would be reset to 0dB, which is completely wrong.  Now, when you do this, you will see the gain slider in the correct location when reopening PAL.

### DIFF
--- a/interface/resources/qml/hifi/NameCard.qml
+++ b/interface/resources/qml/hifi/NameCard.qml
@@ -345,7 +345,7 @@ Item {
             anchors.verticalCenter: nameCardVUMeter.verticalCenter
             // Properties
             visible: !isMyCard && selected
-            value: pal.gainSliderValueDB[uuid] ? pal.gainSliderValueDB[uuid] : Users.getAvatarGain(uuid)
+            value: Users.getAvatarGain(uuid)
             minimumValue: -60.0
             maximumValue: 20.0
             stepSize: 5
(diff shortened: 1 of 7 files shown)


davidkelly:
First off -- note that much of the diffs you see are just trailing whitespace cleanup.  My vim plugin gets rid of these turds automagically now.  Yay.

Test Plan

Open pal, adjust gain sliders for some people up/down.  Close and reopen PAL.  In master, all the sliders came back to 0dB.  In this PR, you will see them at the correct locations (where you left them when you last closed the PAL).

howard-stearns:
- What is gainSliderValueDB  (in Pal.qml) used for now. Can we get rid of it?
- Why are we sending messages to pal.js to set gain? Why not just set it directly from .qml?

hifi-gustavo:
Build console output for commit: 678020fed6d1ff95e8b8e06d0a331942865e6cfd available at: https://gist.github.com/72a9c36becb347d1bccc0092a34adec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9789-678020f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9789-678020f.dmg

davidkelly:
@howard-stearns - agree completely no need to cache the slider value any more, so I'm removing it.  I also think we never needed to send this message back to pal.js to set the gain and handle the analytics call - all could have been done on qml side in the first place.  Easy to fix.  Also, forgot to reset the gains I store when the node list is reset (doh!).  Will push that shortly.

hifi-gustavo:
Build console output for commit: efe44425a63cc54fd27c02742b19bd4e5b4c93a4 available at: https://gist.github.com/72a9c36becb347d1bccc0092a34adec1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9789-efe4442.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9789-efe4442.dmg

howard-stearns:
works for me.
One issue that I think is actually correct, is that when you go to another domain and come back, the gain is reset to 0. I think that's correct because the mixer gain will, in fact, be reset to zero. I think.

PR #9790, opened 2017-02-28
fix bug drawing avatar above ground

AndrewMeadows:

PR #9745, opened 2017-02-24
Faster avatar updates

AndrewMeadows:
* avoid unnecessary copy for sort of avatar updates
* avoid unnecessary animation blending for other avatars
* expand time budget for interface avatar render updates

PR #9798, opened 2017-03-02
avoid animation blending for other avatars

AndrewMeadows:


AndrewMeadows:
This PR reintroduces an optimization found in PR #9745, but it avoids the bug that had to e fixed in PR #9790.

libraries/animation/src/Rig.cpp
```diff
@@ -1346,8 +1346,13 @@ void Rig::copyJointsFromJointData(const QVector<JointData>& jointDataVec) {
...
+
+void Rig::computeExternalPoses(const glm::mat4& modelOffsetMat) {
+    _modelOffset = AnimPose(modelOffsetMat);
+    _geometryToRigTransform = _modelOffset * _geometryOffset;
+    _rigToGeometryTransform = glm::inverse(_geometryToRigTransform);
 
-    // build absolute poses and copy to externalPoseSet
     buildAbsoluteRigPoses(_internalPoseSet._relativePoses, _internalPoseSet._absolutePoses);
     QWriteLocker writeLock(&_externalPoseSetLock);
     _externalPoseSet = _internalPoseSet;
```
AndrewMeadows: This code which calculates the poses in the absolute frame had recently been added to `Rig::copyJointsFromJointData()`, but I'm moving it out into a new function: `Rig::computeExternalPoses()` so it can be called explicitly rather than being an implicit side effect of `Rig::copyJointsFromJointData()`.

interface/src/avatar/SkeletonModel.cpp
```diff
@@ -179,7 +179,9 @@ void SkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
 
         _rig->updateFromEyeParameters(eyeParams);
     } else {
-        Model::updateRig(deltaTime, parentTransform);
+        // no need to call Model::updateRig() because otherAvatars get their joint state
+        // copied directly from AvtarData::_jointData (there are no Rig animations to blend)
+        _needsUpdateClusterMatrices = true;
```
AndrewMeadows: This `else` case is for "other avatars" and we would like to avoid calling `Model::updateRig()` here because it performs a bunch of unnecessary calculations.  But we can only do that if we had already computed the "external poses" after copying the `AvatarData::_jointData` to the rig in the` Avatar::simulate()` context above.

AndrewMeadows:
**To test:**

Nothing visible has changed in this PR so the test would be to make sure avatar animations are still working correctly:

Login to a domain with other avatars and verify that their animations are updating as expected and their bodies are properly rendered standing on the ground in the same spot where they think they are located.

hifi-gustavo:
Build console output for commit: a55ab94cd9d96725c3b667d4448e57bdedb5badd available at: https://gist.github.com/c0747415d932a984b4ed5c31c46a17d0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9798-a55ab94.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9798-a55ab94.dmg

Nex-Pro:
### PASS

All looks good. I'm assuming this is a client fix.

PR #9804, opened 2017-03-03
Update Oculus SDK to 1.11

jherico:
Recent Oculus SDKs include an asynchronous spacewarp (ASW) feature that drops the submission rate for the application to half the present rate, while the compositor continues to run at full speed.  As a result, we see our present rate occasionally dropping, but since we don't know if the ASW feature is kicking in at a given moment, it's difficult to tell whether this is a failing in our application or explicit throttling from the Oculus runtime.

This PR exposes the ASW status from the SDK to the application.  Additionally, it modifies the 'target frame rate' property of the plugin to be ASW aware.  Finally, it adds a normalizedFrameRate property to display plugins, intended to be the fraction represented by the current frame rate divided by the (current) target frame rate.  

Note, the update of the Oculus SDK itself required a little bit of surgery on the external project.  The binaries provided by Oculus now use the dynamic c++ runtime library, rather than the static library previously used.  This means we can't use the lib file they provide any more.  We _can_ compile the shim ourselves, but they provide only a MSVC project, rather than any CMake support.  So this PR also includes a custom CMake file which is patched into their existing distribution and is used to build the target.


hifi-gustavo:
Build console output for commit: ce9d637b3f65194bb8b0c3140fe5ee84cde12bec available at: https://gist.github.com/f80024a58ac5a83d31778c1870aeb779<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9804-ce9d637.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9804-ce9d637.dmg

AndrewMeadows:
To test:

This just needs a smoke-test by someone with an Oculus CV-1 headset and hand controllers:

(1) Login with HMD mode using Oculus, make sure the view still works.
(2) Make sure the hand controllers still work.
(3) Switch to desktop mode, run around and change view.
(4) Switch back to Oculus HMD mode, make sure the view and hand controllers still work.

hifi-gustavo:
Build console output for commit: f9dcfa54aab2d5fa998dcd69df9dac87eb462640 available at: https://gist.github.com/f80024a58ac5a83d31778c1870aeb779<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9804-f9dcfa5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9804-f9dcfa5.dmg

libraries/display-plugins/src/display-plugins/OpenGLDisplayPlugin.cpp
```diff
@@ -654,6 +654,11 @@ float OpenGLDisplayPlugin::presentRate() const {
     return _presentRate.rate();
 }
 
+void OpenGLDisplayPlugin::resetPresentRate() {
+    // FIXME
+    // _presentRate = RateCounter<100>();
```
ZappoMan: does this need to be updated?

claudioshg:
Testing

PR #9761, opened 2017-02-26
WL#21166 Cloneable Equipment

Menithal:
https://worklist.net/21166

Adds four new grabbableKey fields that can be adjusted from the edit window

- cloneable: boolean value to allow for cloneable or non-cloneable.
- cloneLifetime: sets the life time for the clones
- cloneLimit: sets the max amount of clones you can have within a 50m radius
- cloneDynamic: Allows the cloned object to be dynamic

new edit window

To test you will need hand controllers, Vive or Oculus Touch:

# Test Prep

1. Rez A Bow, Pingpong gun, Flaregun, or hookshotgun from the the marketplace
2. Set Cloneable on the objects
3. Dynamic and Grabbable should go off (edit.js only)
4. Turn off Edit window

# General Clonable Test 

1. Attempt to far grab object. It should not allow you to drag the source.
2. Go near the source and grab it with the trigger (or grips, but hold grips),
 a new clone of object should equip like any other object would.
3. You should be able to fire / use the object  
4. You should be able to drop object.

#  Limit Test

Suggest setting Cloneable dynamic, and gravity on source, then use the grip buttons to clone and throw the clone off as many times until it spawns none.

You  should only be able to clone the amount of clones that is set in cloneLimit.  Default, if not defined in the userdata, should be 10 units

# Lifetime Test

All cloned objects should have a lifetime and should expire eventually. This is set according to cloneLifetime. Default, if not defined in the userdata, should be 300s.

# Lock Test

You should have lock permissions: locking the entity should still allow others to interact and grab a clone.

PR #9807, opened 2017-03-04
#9761 Cloneable Improvement

Menithal:
Changes:

- Clones now have a named based on the original entity id for easier identification
- Fixed range logic
- Limit is now calculated from the source instance, instead of just clone name to avoid issues with unnamed objects. This allows there to be multiple entities with the same name and be cloneable, but the limit is calculated per cloneable source instead of bound to all globally.
- Default limit is now 0, which disables limit by default, but still keeps the ability present
- Fixed Padding

To Test, see #9761 testplan.

### DIFF
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -274,24 +274,24 @@ CONTROLLER_STATE_MACHINE[STATE_OVERLAY_LASER_TOUCHING] = CONTROLLER_STATE_MACHIN
 
 // Object assign  polyfill
 if (typeof Object.assign != 'function') {
-  Object.assign = function(target, varArgs) {
-    'use strict';
-    if (target == null) {
-      throw new TypeError('Cannot convert undefined or null to object');
-    }
-    var to = Object(target);
-    for (var index = 1; index < arguments.length; index++) {
-      var nextSource = arguments[index];
-      if (nextSource != null) {
-        for (var nextKey in nextSource) {
-          if (Object.prototype.hasOwnProperty.call(nextSource, nextKey)) {
-            to[nextKey] = nextSource[nextKey];
-          }
+    Object.assign = function(target, varArgs) {
+        'use strict';
+        if (target == null) {
+            throw new TypeError('Cannot convert undefined or null to object');
         }
-      }
-    }
-    return to;
-  };
+        var to = Object(target);
+        for (var index = 1; index < arguments.length; index++) {
+            var nextSource = arguments[index];
+            if (nextSource != null) {
+                for (var nextKey in nextSource) {
+                    if (Object.prototype.hasOwnProperty.call(nextSource, nextKey)) {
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: d85cb645b0301338f5cedad9b547772d8e4b6890 available at: https://gist.github.com/7e96b9a5727a7d81ce4f3d52564e407f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9807-d85cb64.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9807-d85cb64.dmg

scripts/system/controllers/handControllerGrab.js
```diff
@@ -2601,31 +2601,29 @@ function MyController(hand) {
...
                                 var item = Entities.getEntityProperties(item, ["name"]);
-                                if (item.name === grabbedProperties.name) {
+                                if (item.name.indexOf('-clone-' + grabbedProperties.id) !== -1) {
                                     count++;
                                 }
                             })
+
+                            var limit = grabInfo.cloneLimit ? grabInfo.cloneLimit : 0;
+                            if (count >= limit && limit !== 0) {
+                                delete limit;
```
sethalves: what does "delete limit" do here?

Menithal: [delete](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete)s the variable limit from mem, leftover from the move but technically could just be removed.

Nex-Pro:
Testing this.

Nex-Pro:
Followed test plan for #9761 

Also:

Clones inherit UUID in their name of the source entity.

Limit is per source UUID

Limit set to zero is unlimited ( unable to check as there isn't enough time left in the current universe ).

***Only issue seen was that occasionally ( more than 5 times out of  a huge number of clone events ) the master object could be moved by the hand creating the clone.***

To repro: 
Set the green sphere at dev-playa to make clones with unlimited clone count and 10 second lifetime. ( This does not affect the outcome I just did not want to totally flood dev-playa ). 
Then as fast as  possible using both hand controllers create new clones.
Every now and then the master object will be moved while creating a clone.

Nex-Pro:
@Menithal ^^

Menithal:
I am not quite sure if that is actually the scope of the worklist as it seems a very big edge case. I have experienced something similar when working  with the hookgun script and the occasional not being able to grab userdata on some entities.

Meanwhile I can put a safeguard by caching the data on the cloneable, but since it is server related, I cant really do much on the front end side, since technically it could still happen regardless. This however is extra work and not in scope of the worklist item, would it be ok to add additional fee to work around it?

Otherwise, the best course of action is to lock the cloneable item in place to avoid this as the entity server checks the state.

Menithal:
^@cozza13

cozza13:
@Menithal I approved it

PR #9810, opened 2017-03-05
Fix unsafe SIMD optimizations

kencooke:
The SIMD glm_mat4_mul() takes __m128 arguments that are implicitly align(16) and the compiler will generate MOVDQA instructions. Casting from a (float*) is only guaranteed to be align(4) and may cause unaligned memory exceptions.

Fixed by adding SSE version using unaligned loads/stores. Performance is slightly improved, by reducing unnecessary mat4 copying.

### DIFF
--- a/libraries/shared/src/GLMHelpers.h
+++ b/libraries/shared/src/GLMHelpers.h
@@ -245,4 +245,53 @@ inline bool isNaN(const glm::quat& value) { return isNaN(value.w) || isNaN(value
 
 glm::mat4 orthoInverse(const glm::mat4& m);
 
+//
+// Safe replacement of glm_mat4_mul() for unaligned arguments instead of __m128
+//
+inline void glm_mat4u_mul(const glm::mat4& m1, const glm::mat4& m2, glm::mat4& r) {
+
+#if GLM_ARCH & GLM_ARCH_SSE2_BIT
+    __m128 u0 = _mm_loadu_ps((float*)&m1[0][0]);
+    __m128 u1 = _mm_loadu_ps((float*)&m1[1][0]);
+    __m128 u2 = _mm_loadu_ps((float*)&m1[2][0]);
+    __m128 u3 = _mm_loadu_ps((float*)&m1[3][0]);
+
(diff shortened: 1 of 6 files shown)


hifi-gustavo:
Build console output for commit: 818425707b182343ecbd0739fcc2d9df5b5cd689 available at: https://gist.github.com/a3c26daea4fe0f6ac41bbe332b697b24<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9810-8184257.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9810-8184257.dmg

kencooke:
### Test Plan
Using Interface from this build, smoke test that avatar models still load correctly.
The only observable change should be that Interface no longer crashes with unaligned memory faults.

libraries/shared/src/GLMHelpers.h
```diff
@@ -245,4 +245,53 @@ inline bool isNaN(const glm::quat& value) { return isNaN(value.w) || isNaN(value
 
 glm::mat4 orthoInverse(const glm::mat4& m);
 
+//
+// Safe replacement of glm_mat4_mul() for unaligned arguments instead of __m128
+//
+inline void glm_mat4u_mul(const glm::mat4& m1, const glm::mat4& m2, glm::mat4& r) {
+
+#if GLM_ARCH & GLM_ARCH_SSE2_BIT
```
Atlante45: I see you removed the OSX restriction, will that new code run on Macs?

kencooke: Yeah it should be fine on any x86 platform. I wasn't sure why the original code excluded OSX?

nbq:
testing

nbq:
Tester: Coal

* Tested build with my current avatar and it looks and moves fine
* Tested with photo-real male avatar
* Tested with Robot
* Tested with Skeleton
* Went to Welcome and checked out other avatars there and verified they looked like they should

**QA PASS**

PR #9811, opened 2017-03-06
Cleanup how Models do ray-picking and determination of point inside of convex hull

ZappoMan:
- Adds a `TriangleSet` class that does basic ray picking and convexHull point testing. This is essentially the same code as used to live directly in Model.cpp, but now it's been moved into a clean class where we can optimize the implementation without touching the Model class.
- Changes Model.cpp to use this new TriangleSet class.
- Changes Model.cpp to not constantly re-calculate the triangle list and mesh part boxes, and to instead simply use the details from the TriangleSets.

**NOTE:** because we were constantly recalculating the old triangle list every time the model moved or rotated, this new approach should see some slight performance improvement. This does not actually change the algorithm for doing ray-picking within a triangle set faster. That will be a subsequent PR.

------------
Test Plan:
- Smoke test
- Make sure to test Ray picking... e.g.:
    - using mouse to select things in edit.js
    - using hand controllers and lasers to click on things
- Make sure to test convex hull "zone shapes" to see if entering/leaving the zone still works correctly


hifi-gustavo:
Build console output for commit: 2a663cbcb16352c043e38f6f6744e1615045df5d available at: https://gist.github.com/52370e1a0acf3919d8758669f0a1414a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-2a663cb.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-2a663cb.dmg

Nex-Pro:
Looking at this

Nex-Pro:
### FAIL

Issues seen:

in HMD.  Was unable to teleport from the tutorial to outside. Could not get blue beam on anything outside of the Japanese house despite successfully completing tutorial. Went to welcome. Mostly unable to teleport. Could not trigger launch of EZ block ( was logged in ).  Could not move any EZ blocks. Could not go into edit mode as I was getting the permissions error.
Went to dev-playa

Unable to TP. In edit mode could only select primitives. Meshes ( locked or unlocked ) could not be selected. 

In desktop. Edit mode able to occasionally select fbx models.  Could always select primitives.

Notes:  Fresh install did not copy settings. After first issues were observed restarted this build several times. Checked that the correct script folder was being used. Reloaded scripts.

Compared this to 6115 and the 6095 no such issues.

Convex hull zone test not attempted.

libraries/shared/src/TriangleSet.cpp
```diff
@@ -0,0 +1,76 @@
...
+    for (const auto& triangle : _triangles) {
+        if (!isPointBehindTrianglesPlane(point, triangle.v0, triangle.v1, triangle.v2)) {
+            // it's not behind at least one so we bail
+            insideMesh = false;
+            break;
+        }
+
+    }
+    return insideMesh;
+}
```
AndrewMeadows: `TriangleSet::convexHullContains()` works as advertised... iff the mesh actually describes a convex hull.  However, looking further up the callstack there appear to be problems.  In short: `EntityItem::contains(point)` probably doesn't work correctly for most models people use these days.

The good news: that problem does not block this PR.

libraries/render-utils/src/Model.h
```diff
@@ -293,10 +287,10 @@ public slots:
 
     SpatiallyNestable* _spatiallyNestableOverride;
 
-    glm::vec3 _translation;
+    glm::vec3 _translation; // this is the translation in world coordinates to the model's registration point
     glm::quat _rotation;
     glm::vec3 _scale;
-    glm::vec3 _offset;
+    glm::vec3 _offset; // this is the translation for the minimum extent of the model (in original mesh coordinate space) to the model's registration point
```
hyperlogic: This comment is accurate for entity models but not for avatars.  For avatars this is the offset from the avatar's hips, as determined by the default pose, to the origin.  Not to be nit picky, just an observation

libraries/render-utils/src/Model.cpp
```diff
@@ -442,173 +429,98 @@ bool Model::convexHullContains(glm::vec3 point) {
...
+    _modelSpaceMeshTriangleSets.clear();
+    _modelSpaceMeshTriangleSets.resize(numberOfMeshes);
+
+    for (int i = 0; i < numberOfMeshes; i++) {
+        const FBXMesh& mesh = geometry.meshes.at(i);
+
+        for (int j = 0; j < mesh.parts.size(); j++) {
+            const FBXMeshPart& part = mesh.parts.at(j);
+
+            const int INDICES_PER_TRIANGLE = 3;
```
hyperlogic: To reduce the number of allocations within the TriangleSet vector, you can pre-compute the number of triangles here, and do a [QVector::reserve()](http://doc.qt.io/qt-5/qvector.html#reserve).

libraries/render-utils/src/Model.cpp
```diff
@@ -442,173 +429,98 @@ bool Model::convexHullContains(glm::vec3 point) {
...
+                    _modelSpaceMeshTriangleSets[i].insertTriangle(tri);
+                }
+            }
         }
     }
-
-    if (!_addedToScene && isLoaded()) {
-        createRenderItemSet();
-    }
 }
```
AndrewMeadows: Not a blocker:

This isn't new, but: for every triangle we allocate and copy all three points (rather than have one array of points and an array of triangle indices).  This trades CPU cycles for higher memory footprint but probably better memory coherence.  The win is not clear.

Nevertheless, let us count the distinct ways in which we store mesh data:

(1) FBXGeometry (vertices and indices)
(2) Physics shape (vertices and indices for static mesh)
(3) Render item (vertices and indices)
(4) Mesh triangles for ray-casts (vertices in triplicate, no indices)

Someday, as domains get more interesting and we continue our efforts to support mobile (and console?) platforms we may need to combine some of these formats for multiple purposes.

libraries/shared/src/TriangleSet.h
```diff
@@ -0,0 +1,33 @@
...
+public:
+    void insertTriangle(const Triangle& t);
+    void clear();
+
+    // Determine of the given ray (origin/direction) in model space intersects with any triangles
+    // in the set. If an intersection occurs, the distance and surface normal will be provided.
+    bool findRayIntersection(const glm::vec3& origin, const glm::vec3& direction, 
+        float& distance, BoxFace& face, glm::vec3& surfaceNormal, bool precision) const;
+
+    bool convexHullContains(const glm::vec3& point) const; // this point is "inside" all triangles
```
hyperlogic: Maybe this method should contain comment to the effect that this only works if the triangleSet is already a closed convex polytope.  It does not work properly on arbitrary meshes.

libraries/render-utils/src/Model.cpp
```diff
@@ -442,173 +429,98 @@ bool Model::convexHullContains(glm::vec3 point) {
...
-                    if (!isPointBehindTrianglesPlane(point, triangle.v0, triangle.v1, triangle.v2)) {
-                        // it's not behind at least one so we bail
-                        insideMesh = false;
-                        break;
-                    }
-
-                }
-                if (insideMesh) {
+        glm::mat4 meshToModelMatrix = glm::scale(_scale) * glm::translate(_offset);
+        glm::mat4 meshToWorldMatrix = meshToModelMatrix * glm::translate(_translation) * glm::mat4_cast(_rotation);
```
hyperlogic: It would probably be faster to call createMatFromQuatAndPos here, instead of doing the matrix multiply.  This line would become.

    glm::mat4 meshToWorldMatrix = meshToModelMatrix * createMatFromQuatAndPos(_rotation, _translation);

libraries/render-utils/src/Model.cpp
```diff
@@ -360,53 +357,43 @@ bool Model::findRayIntersectionAgainstSubMeshes(const glm::vec3& origin, const g
...
-                    } else {
-                        // this is the non-triangle picking case...
-                        bestDistance = distanceToSubMesh;
-                        intersectedSomething = true;
-                        face = subMeshFace;
-                        surfaceNormal = subMeshSurfaceNormal;
-                        extraInfo = geometry.getModelNameOfMesh(subMeshIndex);
-                    }
+        glm::mat4 meshToModelMatrix = glm::scale(_scale) * glm::translate(_offset);
+        glm::mat4 meshToWorldMatrix = meshToModelMatrix * glm::translate(_translation) * glm::mat4_cast(_rotation);
```
hyperlogic: Same here, It would be quicker to use createMatFromQuatAndPos

    glm::mat4 meshToWorldMatrix = meshToModelMatrix * createMatFromQuatAndPos(_rotation, _translation);

libraries/render-utils/src/Model.h
ZappoMan: hmmmmm - ok

libraries/shared/src/TriangleSet.h
ZappoMan: agreed - note that this is more of a case of me "moving existing code"... but I will add a comment

libraries/render-utils/src/Model.cpp
ZappoMan: cool - will fix

libraries/render-utils/src/Model.cpp
ZappoMan: cool - will fix

libraries/render-utils/src/Model.cpp
ZappoMan: As I see it, this is a implementation nuance of the TriangleSet datatype. Different data structures underneath may or may not be able to use indices...

libraries/shared/src/TriangleSet.h
```diff
@@ -0,0 +1,33 @@
...
+    const AABox& getBounds() const { return _bounds; }
+
+private:
+    QVector<Triangle> _triangles;
```
samcake: Can we use std vector instead

libraries/shared/src/TriangleSet.h
```diff
@@ -0,0 +1,33 @@
...
+//  libraries/entities/src
+//
+//  Created by Brad Hefta-Gaub on 3/2/2017.
+//  Copyright 2017 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include <QVector>
```
samcake: Can we use std vector instead

libraries/render-utils/src/Model.cpp
ZappoMan: fixed - but you left a comment on an unrelated line of code so it doesn't show as outdated.

libraries/render-utils/src/Model.cpp
samcake: I agree with Andrew's comment
The triangle set as of now is recreating a full copy of the mesh vertices
WE could ( and that s what i would advocate for) rely on the model::Mesh class to perform the algorithm on that storage
So that Triangle Set like you say would be an interface providing new features on the Mesh class
I understand that you replaced one copy by another which is more abstract, that s a step forward, i m saying this new class could be considered an algorithm class performing on the model::Mesh which is (in my mind at least) what we want to condsolidate against as our unique mesh class at runtime

libraries/shared/src/TriangleSet.h
ZappoMan: ok

libraries/render-utils/src/Model.cpp
ZappoMan: I don't disagree with this feedback... but the point of this code is to cleanup and provide an optimization. The current master code base has all the problems you've described and many more... mainly that it constantly recalculates the copied vertices every time the model changes the it's position in world space.

This PR is progress.

ZappoMan: btw -- why do we keep around the `FBXGeometry (vertices and indices)` what code uses that? I know the old ray pick code used it... and the new ray pick code uses it once... but never reuses it... do we have code that "reuses" it after the `model::Mesh` and the new `TriangleSet` code used it on init?

ZappoMan: We do call `getFBXGeometry()` throughout the code base... each of those should be inspected for use and determined if they can use `model::Mesh` instead.

hifi-gustavo:
Build console output for commit: af97e9bdd9f1ec8bece9166253f04044b828f8c7 available at: https://gist.github.com/52370e1a0acf3919d8758669f0a1414a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-af97e9b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-af97e9b.dmg

hifi-gustavo:
Build console output for commit: faba16033103ae971b91a01f8ddd739dfd7b9995 available at: https://gist.github.com/52370e1a0acf3919d8758669f0a1414a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-faba160.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-faba160.dmg

hifi-gustavo:
Build console output for commit: f27475613fe829e3f437cff48f7972871c38226f available at: https://gist.github.com/52370e1a0acf3919d8758669f0a1414a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-f274756.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-f274756.dmg

libraries/render-utils/src/Model.cpp
AndrewMeadows: I think we should merge this PR: it is complete and a step forward.  Now that we recognize the problem of multiple copies of the geometry we should put it on the big list of things to do, prioritize it, and deliver that in separate PRs.

BTW, I believe there is a way to use the `model::Mesh` for the physics shape data (when using `btBvhTriangleMeshShape`).  The static mesh shape project was an experiment that worked, but I didn't do it in the most optimal way the first time.

ZappoMan: @AndrewMeadows - it needs to pass QA first... I'm looking into the problems Piper saw... there's definitely something wrong.

ZappoMan:
@Nex-Pro - ok - so I found the bug that caused things to go so horribly wrong for you. This should now work.

@AndrewMeadows @hyperlogic - Some small changes since you reviewed it last.

hifi-gustavo:
Build console output for commit: 87934ee82d1c1a247311ec3886d3c278aa40d8a3 available at: https://gist.github.com/52370e1a0acf3919d8758669f0a1414a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-87934ee.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-87934ee.dmg

Nex-Pro:
#9811

Testing platform:

	OS: Win10 Pro 64
	CPU 4.0 ghz i7
	ram: 16gb
	gpu: Nvidia GTX1070 8gb Vram.
	HMD: HTC Vive.
	Hand Controllers: Vive.

Testing conditions:
	
	Removed previous installation and folders.

	Ensured no copies of interface.json existed prior to installation.

	Elected NOT to copy settings.

### Test Plan:
	Smoke test
Make sure to test Ray picking... e.g.:

**using mouse to select things in edit.js:** **SUCCESS**

**using hand controllers and lasers to click on things: **
	This could at best be described as Hit-Or-miss.
	Went to dev-playa and after needing to use the menu to re-enable the hud in HMD mode I entered edit mode.
	I tried selecting almost anything in sight and it is not reliable. Some items would not select until after another object was selected.
	Even with multiple tries or swapping hands. 

**Make sure to test convex hull "zone shapes" to see if entering/leaving the zone still works correctly:**

	Zone transitions very poor.
	Created a 10m cube zone set sky color to be deep red.
	Was able to move several meters into the zone and remain stationary and never get the red sky.
	If I got the red sky I could leave the zone and move to the default zone by a meter or two and either not get the default sky ( sky remains red ) or get a black sky.
	I could pass diagonally across the corner of a zone by more than a meter in from the bounding corner and never get the zone sky.
	Results seemed more dependent on how much/quickly I rotated my head rather than my position.

Additional:

	Teleporting seems to work fine now.
	Tablet is showing as private for some unknown reason.
	Noticed while eding entities with hand controllers or mouse that scaling them large would result in them lodding out completely including the handles.

Nex-Pro:
@ZappoMan ^

hifi-gustavo:
Build console output for commit: f7e824e47c004f2b9819163d42fe992568e64112 available at: https://gist.github.com/52370e1a0acf3919d8758669f0a1414a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-f7e824e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-f7e824e.dmg

ZappoMan:
test this please

hifi-gustavo:
Build console output for commit: f7e824e47c004f2b9819163d42fe992568e64112 available at: https://gist.github.com/52370e1a0acf3919d8758669f0a1414a<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-f7e824e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9811-f7e824e.dmg

Nex-Pro:
Retesting this shortly

Nex-Pro:
### Pass

PR #9815, opened 2017-03-06
Retain selection in PAL when sorting or reloading

davidkelly:
When you sort a column or reload the PAL, you probably expect the selected avatar to still be selected in the PAL afterwards.  But they were not.  Now they should be.


davidkelly:
TestPlan:

Open PAL in a place with several people
Select someone
Sort on volume, name, etc...
They should remain selected the whole time
Then reload PAL.  
They should still be selected (though in a different row)

davidkelly:
see https://highfidelity.fogbugz.com/f/cases/3377/pal-loses-selection-on-refresh-sort-filter-toggle#

hifi-gustavo:
Build console output for commit: f23f882595078612ed46d2bac9c15dc661db14f0 available at: https://gist.github.com/5492ae64cc24be8540e6d1db7a7f13c9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9815-f23f882.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9815-f23f882.dmg

hifi-gustavo:
Build console output for commit: 3d39aae35b0e91a594c9a3aedf8d6548938b8654 available at: https://gist.github.com/5492ae64cc24be8540e6d1db7a7f13c9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9815-3d39aae.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9815-3d39aae.dmg

OmegaHeron:
Tested Windows 10
QA Pass

Tried various sorting, selections of agents in list and states of ignore/mute.

Not from this PR (exists in master), but, would be a nice fix for glorious future PAL work.  Subtle thing...  In the Ignore column if you click column header to sort anywhere but at its left most side it sorts the column, but, makes it look as if you have ignored everyone as it selects all the elements.  This is in desktop mode - put another way... if mouse pointer is anywhere in column to right if I in ignore it does this.  Other columns do not do this.

davidkelly:
@OmegaHeron - great catch on that checkbox issue.  That is weird -- unsure why it happens and is totally worth fixing.  Thx!

howard-stearns:
Indeed, @OmegaHeron. https://highfidelity.fogbugz.com/f/cases/3314/sorting-by-table-column-in-qml-causes-buttons-to-be-highlighted-e-g-in-pal 
 
It's so frustrating and I have no idea why it's happening, but I know when: if the upper left corner of Interface is near the upper left corner of the Windows desktop. In that case, sort makes the hole column act as though it was being hovered by the mouse. Otherwise, if the corner is moved elsewhere, it doesn't happen. So bizarre.

OmegaHeron:
@howard-stearns - That really is strange!  And, yes, my pal window was somewhat toward the upper left side screen.  Most of my transient dialogs go in that general area, probably yet another penalty of being left handed since I tend to like my elements displayed to left when I have a choice.  :)

interface/resources/qml/hifi/Pal.qml
```diff
@@ -627,6 +642,7 @@ Rectangle {
 
         userModel.clear();
         var userIndex = 0;
+        var newSelectedIndexes=[];
```
howard-stearns: coding standard. space around '='

interface/resources/qml/hifi/Pal.qml
```diff
@@ -636,7 +652,14 @@ Rectangle {
             ['personalMute', 'ignore', 'mute', 'kick'].forEach(init);
             datum.userIndex = userIndex++;
             userModel.append(datum);
+            if (selectedIDs.indexOf(datum.sessionId) != -1) {
+                newSelectedIndexes.push(datum.userIndex);
+            }
         });
+        if (newSelectedIndexes.length > 0) {
+            table.selection.select(newSelectedIndexes);
+            table.positionViewAtRow(userIndex, ListView.Beginning);
```
howard-stearns: userIndex? Not newSelectedIndexes[0]? 
i.e., is it positioning the selected row in view, or the bottom row? (I'm just looking at the diff here, not running the code.)

howard-stearns:
As far as I can tell, it's not a function of where the PAL window is, but where the Interface.exe window is in relationship to the operating system desktop!

OmegaHeron:
If it's any help... my setup is (Windows 10).  Two displays.  Interface running on left display as a maximized window (*not* full screen).  Physically my left display is indicated as display number 2 via control panel and is set as my primary display.

howard-stearns:
right. So if it follows my results, you could move Interface down or to the right or both. (It would then fall off the bottom or right side unless you make it smaller.) At that point, I predict you won't see the problem. (I'm not advocating that you do this in general. It sucks. Just trying to prove the "theory".)

interface/resources/qml/hifi/Pal.qml
davidkelly: oh nice catch.

OmegaHeron:
Theory proved.  If Interface window is anchored into upper left most corner on a 2 display system where left monitor is second device it happens.  Place Interface anywhere else and it does not happen.

I disconnected my second monitor and confirmed windows was fully in single monitor mode.  Interface maximized on single display exhibits problem... out of maximized window it does not.  The key seems to be if the window is anchored to upper left corner or not.  I.E. You could have a window anchored to UL corner but not taking up full screen (vertically maximize only).

Weird weird stuff.

hifi-gustavo:
Build console output for commit: c6dbfc85f055bc95646b04b36b38c7b3526a00ad available at: https://gist.github.com/5492ae64cc24be8540e6d1db7a7f13c9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9815-c6dbfc8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9815-c6dbfc8.dmg

hifi-gustavo:
Build console output for commit: b760e8622606c91be1eac4995c30426ff56fb30a available at: https://gist.github.com/5492ae64cc24be8540e6d1db7a7f13c9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9815-b760e86.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9815-b760e86.dmg

PR #9816, opened 2017-03-06
ModelEntityItems that use 'sphere' for shapeType should collide as true spheres

AndrewMeadows:
* use implicit sphere for collision shape of ModelEntityItems with shapeType='sphere'


AndrewMeadows:
To test:

**(1)** Create a ModelEntityItem that is not spherical in appearance
**(2)** Using the Edit UI select the model from **(1)** and set its "Collision shape type" to "Sphere"
**(3)** Try to bump into, walk on, or throw objects against the entity... it should collide like a sphere.
**(4)** Enable rendering of collision shapes:  **Developer --> Physics --> Draw Collision Shapes**.  Verify that the entity's collision shape looks like a sphere (note: the rendered shape is probably less smooth as the object's collision shape -- it uses a finite number of triangles).

ZappoMan:
The title on this PR is weird, you say this is changing "model" entity types.. but you include a change to "shape" entity types. Which is it?

hifi-gustavo:
Build console output for commit: 10b3e7fc676736c46aafa354065bd6a879cfda8c available at: https://gist.github.com/073c679e5bb4caeb56da9cc08b02978c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9816-10b3e7f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9816-10b3e7f.dmg

libraries/entities/src/ShapeEntityItem.cpp
```diff
@@ -163,7 +163,7 @@ void ShapeEntityItem::appendSubclassData(OctreePacketData* packetData, EncodeBit
 // This value specifes how the shape should be treated by physics calculations.  
 // For now, all polys will act as spheres
 ShapeType ShapeEntityItem::getShapeType() const {
-    return (_shape == entity::Shape::Cube) ? SHAPE_TYPE_BOX : SHAPE_TYPE_SPHERE;
+    return (_shape == entity::Shape::Cube) ? SHAPE_TYPE_BOX : SHAPE_TYPE_ELLIPSOID;
```
AndrewMeadows: This is a hack to continue to support "ellipsoidal" shape for ShapeEntityItems with non-Box shape types (same as how it works in master).

Note: we never got around to supporting proper collision shapes for the other types: tetrahedron, octohedron, dodechahedron, and icosahedron, so they collide same as "sphere".

libraries/physics/src/ShapeFactory.cpp
```diff
@@ -257,8 +257,19 @@ const btCollisionShape* ShapeFactory::createShapeFromInfo(const ShapeInfo& info)
...
+            glm::vec3 halfExtents = info.getHalfExtents();
             float radius = halfExtents.x;
-            if (radius == halfExtents.y && radius == halfExtents.z) {
+            const float MIN_RADIUS = 0.001f;
+            const float MIN_RELATIVE_SPHERICAL_ERROR = 0.001f;
+            if (radius > MIN_RADIUS
+                    && fabs(radius - halfExtents.y) / radius < MIN_RELATIVE_SPHERICAL_ERROR
+                    && fabs(radius - halfExtents.z) / radius < MIN_RELATIVE_SPHERICAL_ERROR) {
+                // close enough to true sphere
                 shape = new btSphereShape(radius);
```
AndrewMeadows: Here where we actually build the Bullet collision shapes the new "ellipsoid" shape is basically the same as the old "sphere" shape... except we now test for things that are very close to spherical (rather than _exactly_spherical_)for which we use a true implicit sphere shape, otherwise we build an ellipsoidal convex hull.

libraries/entities/src/ShapeEntityItem.cpp
ZappoMan: er... "haven't yet" ;)

hifi-gustavo:
Build console output for commit: d749c0c8beb9936bfec689309a1c5c77d0dce341 available at: https://gist.github.com/073c679e5bb4caeb56da9cc08b02978c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9816-d749c0c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9816-d749c0c.dmg

OmegaHeron:
Windows 10

Create a ModelEntityItem that is not spherical in appearance
Using the Edit UI select the entity from (1) and set its "Collision shape type" to "Sphere"

**Pass** Try to bump into, walk on, or throw objects against the entity... it should collide like a sphere.

**Pass** Enable rendering of collision shapes: Developer --> Physics --> Draw Collision Shapes. Verify that the entity's collision shape looks like a sphere (note: the rendered shape is probably less smooth as the object's collision shape -- the rendered sphere uses a finite number of triangles).

Create a "sphere" entity using the Edit UI.
Scale the sphere to be ellipsoidal: change the 'dimension' of the entity to be different in X, Y, and Z.

**Pass** Set the sphere dynamic and give it some downward gravity --> verify that it bumps and rolls around like an ellipsoid rather than a true sphere. It should collide pretty close to what it looks like.

Sphere shape impacts then moves like one would expect a sphere to behave.  Ellipsoid contacts ground then rolls over and rocks until coming to rest as I would expect.

PR #9822, opened 2017-03-07
Basic proxy support. Will not work for UDP connections

vladest:
Need extra work for UDP sockets

### DIFF
--- a/assignment-client/src/main.cpp
+++ b/assignment-client/src/main.cpp
@@ -14,6 +14,7 @@
 
 #include "AssignmentClientApp.h"
 #include <BuildInfo.h>
+#include <QNetworkProxyFactory>
 
 int main(int argc, char* argv[]) {
     disableQtBearerPoll(); // Fixes wifi ping spikes
@@ -27,7 +28,8 @@ int main(int argc, char* argv[]) {
     qInfo() << "Starting.";
 
     AssignmentClientApp app(argc, argv);
-    
+    QNetworkProxyFactory::setUseSystemConfiguration(true);
+
     int acReturn = app.exec();
     qDebug() << "assignment-client process" <<  app.applicationPid() << "exiting with status code" << acReturn;
 
--- a/domain-server/src/main.cpp
+++ b/domain-server/src/main.cpp
@@ -20,6 +20,7 @@
 #include <BuildInfo.h>
 
 #include "DomainServer.h"
+#include <QNetworkProxyFactory>
 
 int main(int argc, char* argv[]) {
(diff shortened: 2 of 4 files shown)


hifi-gustavo:

ZappoMan:
ok to test

ZappoMan:
@vladest - can you write a test plan for this, namely how does a test verify that this is working as expected. Presumably from reading the Qt docs, one would need to verify that when changing the system wide proxy settings that the interface/domain server/etc use those configured values.

hifi-gustavo:
Build console output for commit: 2853666ba907465eadc45441c51d63964a9a7d94 available at: https://gist.github.com/32bdfd3b2a7d805d56132f0f9d0018fa<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9822-2853666.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9822-2853666.dmg

vladest:
@ZappoMan Hi. Thanks for your comment. Can you please point me to some kind of template and where to place the plan? thanks

sethalves:
Look at some of the other pull requests.  In some cases, the test plan can be quite simple.  In this case it might be difficult to know if it's working... is wireshark or something required?

vladest:
Test plan:
1. Run on system with proxy only  enabled
2. Setup system proxy 
3. Check that non-UDP based functionality works

Unfortunately, UDP specific parts of networking will not works thru HTTP proxy. Need to experiment with SOCKS5 proxies

OmegaHeron:
To be clear - what exactly does this proxy?  Just the http fetch of a model via URL and *everything* else bypasses proxy?  It's about as clear as mud and I was asked to QA it.

vladest:
yes, the proxy is working only for http requests

OmegaHeron:
If there's anyone who can QA this, please do so, maybe someone is around who actually uses proxy on a regular basis.  I started trying to configure my Win 10 box for proxy and totally screwed up my networking in the process.

sethalves:
@vladest could you describe the circumstances that caused to you write this code?  That might help with the testing.

ZappoMan:
If you're behind a firewall, that requires http traffic to go through a proxy (security, caching, etc) this will allow our http traffic to work. It might seem odd to have a firewall that blocks http but not udp, but I've seen it. 

To test this we need to set up an http Forward proxy. Squid or haproxy would do.

cozza13:
I am going to close this. Not sure if we would use it

PR #9829, opened 2017-03-08
Add double click message support to entities and overlays

ZappoMan:
Test plan:
- smoke test
- add [this script](https://raw.githubusercontent.com/ZappoMan/hifi/d7c3677b2211bd3b906bbf40c1d23b61f827b2d5/script-archive/entityScripts/doubleClickExample.js) to an entity, and double click on the entity, see that in the log you get the following output:
```
[03/07 17:58:00] [DEBUG] [hifi.scriptengine.script] script:print()<< clickDownOnEntity
[03/07 17:58:00] [DEBUG] [hifi.scriptengine.script] script:print()<< doubleclickOnEntity
```

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -3174,7 +3174,23 @@ void Application::mousePressEvent(QMouseEvent* event) {
     }
 }
 
-void Application::mouseDoublePressEvent(QMouseEvent* event) const {
+void Application::mouseDoublePressEvent(QMouseEvent* event) {
+    auto offscreenUi = DependencyManager::get<OffscreenUi>();
+    auto eventPosition = getApplicationCompositor().getMouseEventPosition(event);
+    QPointF transformedPos = offscreenUi->mapToVirtualScreen(eventPosition, _glWidget);
+    QMouseEvent mappedEvent(event->type(),
+        transformedPos,
+        event->screenPos(), event->button(),
+        event->buttons(), event->modifiers());
+
+    if (!_aboutToQuit) {
+        getOverlays().mouseDoublePressEvent(&mappedEvent);
+        if (!_controllerScriptingInterface->areEntityClicksCaptured()) {
+            getEntities()->mouseDoublePressEvent(&mappedEvent);
+        }
+    }
+
+
     // if one of our scripts have asked to capture this event, then stop processing it
     if (_controllerScriptingInterface->isMouseCaptured()) {
         return;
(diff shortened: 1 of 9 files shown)


hifi-gustavo:
Build console output for commit: d7c3677b2211bd3b906bbf40c1d23b61f827b2d5 available at: https://gist.github.com/b2d69091f2426f4343a54579814f96fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9829-d7c3677.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9829-d7c3677.dmg

OmegaHeron:
Tested Windows 10
QA Pass

Not a fail... just logic chopping impressions of this new functionality.  It works as advertised, but, I felt that the interval between clicks to be considered single vs double might be a touch long.  It seemed that I would get double click registered when a time longer than my experience feels should be a double vs two singles.

ZappoMan:
@OmegaHeron the interval between click and doubleclick is controlled by your OS, this is a setting in your Windows, Mac, or even unix windowing toolkit's settings.

OmegaHeron:
Fair enough @ZappoMan - it seems different in interface than elsewhere, but, that may have just been me paying too much attention to something looking for problems and fooling myself.

ZappoMan:
@OmegaHeron well- this code literally just listens to the OS message and sends it to scripts, there's no actual timing code in our code base. So it's got to be just how the OS decides to send those messages.

interface/src/Application.cpp
```diff
@@ -3174,7 +3174,23 @@ void Application::mousePressEvent(QMouseEvent* event) {
...
+        transformedPos,
+        event->screenPos(), event->button(),
+        event->buttons(), event->modifiers());
+
+    if (!_aboutToQuit) {
+        getOverlays().mouseDoublePressEvent(&mappedEvent);
+        if (!_controllerScriptingInterface->areEntityClicksCaptured()) {
+            getEntities()->mouseDoublePressEvent(&mappedEvent);
+        }
+    }
```
AndrewMeadows: Not a blocker since the code is functionally correct and the theoretical and minor performance issue (optimistic creation of something that might not be used later) only happens at shutdown but I mention it because it makes me say "WTF?... oh I see":

`mappedEvent` is only used inside the context of `if (!_aboutToQuit)` which makes me think: why bother creating it **outside** that context?  Why not only create it **inside**?

ZappoMan: good catch - this code by me was a copy/paste find/replace... of someone elses code.... so that optimization exists in a couple place... we can open a different PR to fix it...

PR #9842, opened 2017-03-09
improved noise gate

ZappoMan:
- fix a bug in mixers handling of silent packet transitions
- flush the encoder to silence when closing the noise gate on the client

**Test Plan**
- To test the noise gate behavior follow these steps:
    - Turn up your mic gain a lot so you get lots of background noise
    - Make sure Developer > Audio > Audio Noise Reduction is enabled
    - go to a quiet domain so you're the only one there.
    - Enable Server Loopback
    - Enable Audio Scope
    - Be as quiet as you can... watch that the audio is flat and you don't hear your loopback
    - make a relatively quiet noise (like a whisper or a jaw pop) -- listen for the server loop back. 
    - Verify you see a "clean" nearly identical echo in the scope... something like this:
image

NOTE: There is a fix to each side client and server which in all cases will "improve" this artifact. The ideal result is from using a new client and new server, but since clients can't be trusted the server has it's own guards against old or malformed clients to reduce the artifact.

- You can test this with OLD clients and this NEW server. You should see...
image

Notice that in this case it's not perfect, but it's pretty close.

- You can also test this with this NEW client and OLD servers. You should see...
image

Notice that in this case it's better and much closer to perfect.

For reference, here's an OLD client with an OLD server...
image

In this case you can clearly see that the first 5-6 frames of the echoed sample are completely wrong. That's because they are from the samples before the noise gate closed.

### DIFF
--- a/libraries/audio-client/src/AudioNoiseGate.cpp
+++ b/libraries/audio-client/src/AudioNoiseGate.cpp
@@ -58,8 +58,6 @@ void AudioNoiseGate::removeDCOffset(int16_t* samples, int numSamples) {
     }
 }
 
-#include <QDebug>
-
 void AudioNoiseGate::gateSamples(int16_t* samples, int numSamples) {
     //
     //  Impose Noise Gate
--- a/libraries/audio-client/src/AudioNoiseGate.cpp
+++ b/libraries/audio-client/src/AudioNoiseGate.cpp
@@ -77,8 +77,6 @@ void AudioNoiseGate::gateSamples(int16_t* samples, int numSamples) {
     //                      More means better rejection but also can reject continuous things like singing.
     // NUMBER_OF_NOISE_SAMPLE_FRAMES:  How often should we re-evaluate the noise floor?
 
-    _closedInLastFrame = false;
-    
     float loudness = 0;
     int thisSample = 0;
     int samplesOverNoiseGate = 0;
@@ -142,14 +140,13 @@ void AudioNoiseGate::gateSamples(int16_t* samples, int numSamples) {
         _sampleCounter = 0;
         
     }
+
     if (samplesOverNoiseGate > NOISE_GATE_WIDTH) {
         _isOpen = true;
         _framesToClose = NOISE_GATE_CLOSE_FRAME_DELAY;
     } else {
         if (--_framesToClose == 0) {
-            if (_isOpen) {
-                _closedInLastFrame = true;
-            }
+            _closedInLastFrame = !_isOpen;
             _isOpen = false;
         }
     }
--- a/libraries/audio-client/src/AudioNoiseGate.cpp
+++ b/libraries/audio-client/src/AudioNoiseGate.cpp
@@ -58,6 +58,7 @@ void AudioNoiseGate::removeDCOffset(int16_t* samples, int numSamples) {
     }
 }
 
+#include <QDebug>
 
 void AudioNoiseGate::gateSamples(int16_t* samples, int numSamples) {
     //
@@ -77,7 +78,8 @@ void AudioNoiseGate::gateSamples(int16_t* samples, int numSamples) {
     //  NOISE_GATE_FRAMES_TO_AVERAGE:  How many audio frames should we average together to compute noise floor.
     //                      More means better rejection but also can reject continuous things like singing.
     // NUMBER_OF_NOISE_SAMPLE_FRAMES:  How often should we re-evaluate the noise floor?
-    
+
+    _closedInLastFrame = false;
     
     float loudness = 0;
     int thisSample = 0;
@@ -147,6 +149,9 @@ void AudioNoiseGate::gateSamples(int16_t* samples, int numSamples) {
         _framesToClose = NOISE_GATE_CLOSE_FRAME_DELAY;
     } else {
         if (--_framesToClose == 0) {
+            if (_isOpen) {
+                _closedInLastFrame = true;
+            }
             _isOpen = false;
         }
     }
--- a/libraries/audio/src/InboundAudioStream.cpp
+++ b/libraries/audio/src/InboundAudioStream.cpp
@@ -163,7 +163,7 @@ int InboundAudioStream::parseData(ReceivedMessage& message) {
 
                     // Since the data in the stream is using a codec that we aren't prepared for,
                     // we need to let the codec know that we don't have data for it, this will
-                    // allowe the codec to interpolate missing data and produce a fade to silence.
+                    // allow the codec to interpolate missing data and produce a fade to silence.
                     lostAudioData(1);
 
                     // inform others of the mismatch
@@ -248,7 +248,7 @@ int InboundAudioStream::parseAudioData(PacketType type, const QByteArray& packet
 int InboundAudioStream::writeDroppableSilentFrames(int silentFrames) {
 
     // We can't guarentee that all clients have faded the stream down
-    // to silence and encoding that silence before sending us a 
+    // to silence and encoded that silence before sending us a 
     // SilentAudioFrame. If the encoder has truncated the stream it will
     // leave the decoder holding some unknown loud state. To handle this 
     // case we will call the decoder's lostFrame() method, which indicates
--- a/libraries/audio/src/InboundAudioStream.cpp
+++ b/libraries/audio/src/InboundAudioStream.cpp
@@ -137,7 +137,7 @@ int InboundAudioStream::parseData(ReceivedMessage& message) {
         }
         case SequenceNumberStats::Early: {
             // Packet is early. Treat the packets as if all the packets between the last
-            // OnTime packet and this packet was lost. If we're using a codec this will 
+            // OnTime packet and this packet were lost. If we're using a codec this will 
             // also result in allowing the codec to interpolate lost data. Then
             // fall through to the "on time" logic to actually handle this packet
             int packetsDropped = arrivalInfo._seqDiffFromExpected;
@@ -161,9 +161,9 @@ int InboundAudioStream::parseData(ReceivedMessage& message) {
                 } else {
                     qDebug(audio) << "Codec mismatch: expected" << _selectedCodecName << "got" << codecInPacket << "writing silence";
 
-                    // Since the data in the stream is using a codec that we're not prepared for,
+                    // Since the data in the stream is using a codec that we aren't prepared for,
                     // we need to let the codec know that we don't have data for it, this will
-                    // flush any internal codec state and produce fade to silence.
+                    // allowe the codec to interpolate missing data and produce a fade to silence.
                     lostAudioData(1);
 
                     // inform others of the mismatch
@@ -248,12 +248,12 @@ int InboundAudioStream::parseAudioData(PacketType type, const QByteArray& packet
 int InboundAudioStream::writeDroppableSilentFrames(int silentFrames) {
 
     // We can't guarentee that all clients have faded the stream down
-    // to silence and encode that silence before sending us a 
-    // SilentAudioFrame. The encoder may have truncated the stream and 
-    // left the decoder holding some loud state. To handle this case
-    // we will call the decoder's lostFrame() method, which indicates
-    // that it should interpolate from it's last known state (which may
-    // be some unknown loud state) down toward silence.
+    // to silence and encoding that silence before sending us a 
+    // SilentAudioFrame. If the encoder has truncated the stream it will
+    // leave the decoder holding some unknown loud state. To handle this 
+    // case we will call the decoder's lostFrame() method, which indicates
+    // that it should interpolate from its last known state down toward 
+    // silence.
     if (_decoder) {
         // FIXME - We could potentially use the output from the codec, in which 
         // case we might get a cleaner fade toward silence. NOTE: The below logic 
--- a/libraries/audio/src/InboundAudioStream.cpp
+++ b/libraries/audio/src/InboundAudioStream.cpp
@@ -136,9 +136,10 @@ int InboundAudioStream::parseData(ReceivedMessage& message) {
             break;
         }
         case SequenceNumberStats::Early: {
-            // Packet is early; write droppable silent samples for each of the skipped packets.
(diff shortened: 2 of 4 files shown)


hifi-gustavo:
Build console output for commit: 60daaba52511e79eac826685df77893c50c1eb2d available at: https://gist.github.com/6b4a937b05586623657c244ca91ff0ec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-60daaba.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-60daaba.dmg

libraries/audio/src/InboundAudioStream.cpp
```diff
@@ -136,9 +136,10 @@ int InboundAudioStream::parseData(ReceivedMessage& message) {
             break;
         }
         case SequenceNumberStats::Early: {
-            // Packet is early; write droppable silent samples for each of the skipped packets.
-            // NOTE: we assume that each dropped packet contains the same number of samples
-            // as the packet we just received.
+            // Packet is early treat the packets as if all the packets between the last
```
kencooke: grammar is rough here

libraries/audio/src/InboundAudioStream.cpp
```diff
@@ -136,9 +136,10 @@ int InboundAudioStream::parseData(ReceivedMessage& message) {
...
+            // OnTime packet and this packet was lost. If we're using a codec this will 
+            // also result in allowing the codec to flush its internal state. Then
```
kencooke: in this case, its clearer to say the codec is interpolating the lost audio

libraries/audio-client/src/AudioNoiseGate.cpp
```diff
@@ -77,7 +76,8 @@ void AudioNoiseGate::gateSamples(int16_t* samples, int numSamples) {
     //  NOISE_GATE_FRAMES_TO_AVERAGE:  How many audio frames should we average together to compute noise floor.
     //                      More means better rejection but also can reject continuous things like singing.
     // NUMBER_OF_NOISE_SAMPLE_FRAMES:  How often should we re-evaluate the noise floor?
-    
+
+    _closedInLastFrame = false;
```
zzmp: This state only changes with the opening/closing of the gate. Can you modify it there so that the logic is easier to follow?

libraries/audio/src/InboundAudioStream.cpp
zzmp: add back the semicolon

libraries/audio/src/InboundAudioStream.cpp
```diff
@@ -147,7 +148,10 @@ int InboundAudioStream::parseData(ReceivedMessage& message) {
         case SequenceNumberStats::OnTime: {
             // Packet is on time; parse its data to the ringbuffer
             if (message.getType() == PacketType::SilentAudioFrame) {
-                // FIXME - Some codecs need to know about these silent frames... and can produce better output
+                // If we recieved a SilentAudioFrame from our sender, we might want to drop
+                // some of the samples in order to catch up to our desired jitter buffer size.
+                // NOTE: If we're using a codec we will be calling the codec's lostFrame()
```
zzmp: This comment belongs in writeDroppableSilentFrames. It is not done here, so it is just confusing.

libraries/audio/src/InboundAudioStream.cpp
```diff
@@ -240,6 +249,21 @@ int InboundAudioStream::parseAudioData(PacketType type, const QByteArray& packet
 
 int InboundAudioStream::writeDroppableSilentFrames(int silentFrames) {
 
+    // if we have a decoder, we still want to tell the decoder about our
```
zzmp: this is really long - can you make it more concise?

libraries/audio/src/InboundAudioStream.cpp
```diff
@@ -240,6 +249,21 @@ int InboundAudioStream::parseAudioData(PacketType type, const QByteArray& packet
...
+    // lost frame. this will flush the internal state of the decoder
+    // we can safely ignore the output of the codec in this case, because
+    // we've enforced that on the sending side, the encoder ran at least
+    // one frame of truly silent audio before we sent the "droppable" silent
+    // frame. Technically we could leave this out, if we know for certain
+    // that the sender has really sent us an encoded packet of zeros, but
+    // since we can trust all encoders to always encode at least one silent
+    // frame (open source, someone code modify it), we will go ahead and
```
zzmp: s/code/could

libraries/audio/src/InboundAudioStream.cpp
```diff
@@ -240,6 +249,21 @@ int InboundAudioStream::parseAudioData(PacketType type, const QByteArray& packet
...
 int InboundAudioStream::writeDroppableSilentFrames(int silentFrames) {
 
+    // if we have a decoder, we still want to tell the decoder about our
+    // lost frame. this will flush the internal state of the decoder
+    // we can safely ignore the output of the codec in this case, because
+    // we've enforced that on the sending side, the encoder ran at least
+    // one frame of truly silent audio before we sent the "droppable" silent
+    // frame. Technically we could leave this out, if we know for certain
+    // that the sender has really sent us an encoded packet of zeros, but
+    // since we can trust all encoders to always encode at least one silent
```
kencooke: I think you mean "since we can't trust .."

libraries/audio/src/InboundAudioStream.cpp
```diff
@@ -240,6 +249,21 @@ int InboundAudioStream::parseAudioData(PacketType type, const QByteArray& packet
...
+    // frame (open source, someone code modify it), we will go ahead and
+    // tell our decoder about the lost frame.
```
zzmp: This isn't actually a lost frame, so the method name has become a misnomer. Can it be updated or aliased?

libraries/audio/src/InboundAudioStream.cpp
```diff
@@ -240,6 +249,21 @@ int InboundAudioStream::parseAudioData(PacketType type, const QByteArray& packet
 
 int InboundAudioStream::writeDroppableSilentFrames(int silentFrames) {
 
+    // if we have a decoder, we still want to tell the decoder about our
+    // lost frame. this will flush the internal state of the decoder
```
zzmp: This isn't really a lost frame.

libraries/audio/src/InboundAudioStream.cpp
```diff
@@ -240,6 +249,21 @@ int InboundAudioStream::parseAudioData(PacketType type, const QByteArray& packet
...
+    // we can safely ignore the output of the codec in this case, because
+    // we've enforced that on the sending side, the encoder ran at least
+    // one frame of truly silent audio before we sent the "droppable" silent
+    // frame. Technically we could leave this out, if we know for certain
+    // that the sender has really sent us an encoded packet of zeros, but
+    // since we can trust all encoders to always encode at least one silent
+    // frame (open source, someone code modify it), we will go ahead and
+    // tell our decoder about the lost frame.
+    if (_decoder) {
```
zzmp: So, because we don't trust the client to only label all 0s as a SilentFrame, we always decode it?
In that case, why do we use lostFrame instead of just a normal decode?

kencooke:
Some comments about the comments...

I wouldn't refer to calling LostData() as "flushing the internal state".
The intent (and normal usage) is to synthesize a buffer of interpolated audio, WITHOUT ANY INPUT DATA. It does slowly decay the internal state toward silence (after many frames) to avoid sounding annoying on huge packet losses. This is why it (kind of) flushed the artifacts before.

What is flushing the codec state is encoding at least one frame of silence, and then decoding the result.

libraries/audio/src/InboundAudioStream.cpp
ZappoMan: I have been assured by @kencooke  that `lostFrame` is an industry standard term and we should use it.

libraries/audio/src/InboundAudioStream.cpp
ZappoMan: No - this is because we can't trust that the client has made sure to send us a encode frame that actually reached 0... current clients, will start sending us "silentFrames" but right after sending a previously encoded frame... the result is that as far as the decoder knows, the very last encoded frame did have audio... basically this code branch restores the behavior from prior to Feb when a PR accidentally removed this code.

libraries/audio/src/InboundAudioStream.cpp
ZappoMan: I welcome your edits

libraries/audio/src/InboundAudioStream.cpp
ZappoMan: I have been assured by @kencooke that this is an industry standard term for any time you ask a codec to process audio without actual input.

libraries/audio/src/InboundAudioStream.cpp
ZappoMan: will fix

libraries/audio/src/InboundAudioStream.cpp
ZappoMan: will fix

libraries/audio/src/InboundAudioStream.cpp
ZappoMan: will fix

libraries/audio/src/InboundAudioStream.cpp
ZappoMan: will fix

libraries/audio/src/InboundAudioStream.cpp
ZappoMan: will fix

zzmp:
Can `lostFrame` be refactored to not take a `QByteArray`, if its intention is to synthesize a buffer without any input data?

libraries/audio/src/InboundAudioStream.cpp
kencooke: I named it because "Packet Loss Concealment" is the standard term for interpolating the audio for a lost frame.
But @zzmp is right about the usage here. We didn't lose a frame... we got a "silent" frame and don't trust that the encoding was done right.

kencooke:
the QByteArray is the output buffer (filled with synthesized audio). But in the usage here, the output is discarded. Its being done to fade the internal decoder state towards silence.

hifi-gustavo:
Build console output for commit: 79d8d90581103152926775f5cdd312bc1e2d0ecd available at: https://gist.github.com/6b4a937b05586623657c244ca91ff0ec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-79d8d90.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-79d8d90.dmg

hifi-gustavo:
Build console output for commit: 2d595c1f879ccc75709b164567b7e4c4c00c7010 available at: https://gist.github.com/6b4a937b05586623657c244ca91ff0ec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-2d595c1.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-2d595c1.dmg

sethalves:
test this please

hifi-gustavo:
Build console output for commit: 79d8d90581103152926775f5cdd312bc1e2d0ecd available at: https://gist.github.com/6b4a937b05586623657c244ca91ff0ec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-79d8d90.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-79d8d90.dmg

ZappoMan:
test this please

hifi-gustavo:
Build console output for commit: 79d8d90581103152926775f5cdd312bc1e2d0ecd available at: https://gist.github.com/6b4a937b05586623657c244ca91ff0ec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-79d8d90.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-79d8d90.dmg

hifi-gustavo:
Build console output for commit: bdab24d9bfc731b81db02ffdf409a7768881adb7 available at: https://gist.github.com/6b4a937b05586623657c244ca91ff0ec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-bdab24d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-bdab24d.dmg

hifi-gustavo:
Build console output for commit: bdab24d9bfc731b81db02ffdf409a7768881adb7 available at: https://gist.github.com/6b4a937b05586623657c244ca91ff0ec<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-bdab24d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9842-bdab24d.dmg

OmegaHeron:
Tested Windows 10 in my wind tunnel (i.e. work space with air conditioner and other cooling devices howling).  I use my own physical oscilloscope vs interface's set to capture transient pulses.

A direct comparison master vs this pr would be this;

Master - I enable noise reduction and engage mic - soon as I break squelch audio goes into oscillation (not feedback, oscillation from transient impulses creating a virtual oscillator - seen it a bazzillion times in RF DSP work).  Especially torturous for master is high white noise + a hard break in sound like emphasized T in something like tink. 

This PR - Same setup as above, works well.  Minimal harshness on squelch break, first time in however many years it's been since I began this that audio noise reduction is actually an acceptable option.

Likely be good to have some greater count of testers with differing audio environments, mics etc etc.

ZappoMan:
that sounds like a QA pass to me.

PR #9845, opened 2017-03-09
fix rare physics crash

AndrewMeadows:


libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -143,12 +143,35 @@ void PhysicsEngine::addObjectToDynamicsWorld(ObjectMotionState* motionState) {
...
+                        // swap with rbegin
+                        *itr = *(_activeStaticBodies.rbegin());
+                    }
+                    _activeStaticBodies.pop_back();
+                    break;
+                }
+                ++itr;
+            }
+        }
+    }
```
AndrewMeadows: Note the context here: we're in `PhysicsEngine::removeObjects()` which is only ever called right before the next `PhysicsEngine::stepSimulation()`:

**IF** `_activeStaticBodies` is not empty
  which means it was not cleared after the last `stepSimulation()`
  which means the last `stepSimulation()` didn't actually do anything (because our frame rate is faster than our physics step rate)
**THEN** scan `_activeStaticBodies` for bodies that we happen to be removing RIGHT NOW and clear them from the list, if any.

Since the list is a `std::vector` we scan it in reverse order and swap with the last element (rbegin) if we find something.

AndrewMeadows:
To test:

Reproducing this crash mode is theoretically hard, but this is how you might go about it:

(1) Configure or trick your interface into running faster than 90Hz.
(2) Write a script that moves static objects to new positions and then **promptly** deletes them.
(3) If you're "lucky" and you do (2) enough times the interface should eventually crash.

hifi-gustavo:
Build console output for commit: ffe69acb5cb0ec4acc53a3a2ab4d5c25c2b633dd available at: https://gist.github.com/12e95f96edbefacb8a722f9a8e7f73e3<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9845-ffe69ac.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9845-ffe69ac.dmg

Nex-Pro:
@AndrewMeadows   Are you saying test this until it crashes or do you want to compare how many crashes in master against none in this ?

Nex-Pro:
I think I'll give this a thrashing.

Nex-Pro:
This looks like a **PASS**

With VSYNC off and showing more than 400fps render rate.
Ran script I wrote that created a grid of 100 x 100 ( 10 thousand ) static FBX entities then randomly chose one at a time, moved and deleted.  This at ~60 x per second.  I could see the ents being moved and deleted. Tried this for Textured and non-textured.

I could not get a crash in this.  

Crash in master was fairly instantaneous.

AndrewMeadows:
Yay! @Nex-Pro, thank you for verifying the crash mode and testing this fix.

hifi-gustavo:
Build console output for commit: 2c141fd5552565a8c5e14fa6f2382b3080113c25 available at: https://gist.github.com/12e95f96edbefacb8a722f9a8e7f73e3<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9845-2c141fd.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9845-2c141fd.dmg

PR #9851, opened 2017-03-10
zero velocity of deactivated dynamic entities

AndrewMeadows:
* fix transform and velocities of dynamic entities whose RigidBody is deactivated in physics simulation 
* when an entity with remote simulation ownership comes to rest: try harder to flag it for deactivation

### DIFF
--- a/libraries/physics/src/EntityMotionState.cpp
+++ b/libraries/physics/src/EntityMotionState.cpp
@@ -10,7 +10,6 @@
 //
 
 #include <glm/gtx/norm.hpp>
-#include <iostream> // adebug
 
 #include <EntityItem.h>
 #include <EntityItemProperties.h>
@@ -27,19 +26,6 @@
 #ifdef WANT_DEBUG_ENTITY_TREE_LOCKS
 #include "EntityTree.h"
 #endif
-
-// adebug TODO BOOKMARK:
-// Consider an object near deactivation owned by userB and in view of userA...
-// The problem is that userB may deactivate and disown the object before userA deactivates
-// userA will sometimes insert non-zero velocities (and position errors) into the Entity before it is deactivated locally
-//
-// It would be nice to prevent data export from Bullet to Entity for unowned objects except in cases where it is really needed (?)
-// Maybe can recycle _serverPosition and friends to store the "before simulationStep" data to more efficiently figure out
-// if data should be used for non-owned objects.
-//
-// If we do that, we should convert _serverPosition and friends to use Bullet data types for efficiency.
-//
-// adebug
 
 const uint8_t LOOPS_FOR_SIMULATION_ORPHAN = 50;
 const quint64 USECS_BETWEEN_OWNERSHIP_BIDS = USECS_PER_SECOND / 5;
@@ -112,15 +98,6 @@ void EntityMotionState::updateServerPhysicsVariables() {
 }
 
 void EntityMotionState::handleDeactivation() {
-    // adebug
-    glm::vec3 pos = _entity->getPosition();
-    float dx = glm::distance(pos, _serverPosition);
-    glm::vec3 v = _entity->getVelocity();
-    float dv = glm::distance(v, _serverVelocity);
-    std::cout << "adebug deactivate '" << _entity->getName().toStdString()
-        << "' dx = " << dx << "  dv = " << dv << std::endl;  // adebug
-    // adebug
-
     // copy _server data to entity
     bool success;
     _entity->setPosition(_serverPosition, success, false);
@@ -145,10 +122,6 @@ void EntityMotionState::handleEasyChanges(uint32_t& flags) {
                 _outgoingPriority = 0;
                 const float ACTIVATION_EXPIRY = 3.0f; // something larger than the 2.0 hard coded in Bullet
                 _body->setDeactivationTime(ACTIVATION_EXPIRY);
-                bool verbose = _entity->getName() == "fubar"; // adebug
-                if (verbose) {
-                    std::cout << (void*)(this) << "  " << secTimestampNow() << "  adebug flag for deactivation" << std::endl;  // adebug
-                }
             } else {
                 // disowned object is still moving --> start timer for ownership bid
                 // TODO? put a delay in here proportional to distance from object?
@@ -156,13 +129,6 @@ void EntityMotionState::handleEasyChanges(uint32_t& flags) {
                 _nextOwnershipBid = usecTimestampNow() + USECS_BETWEEN_OWNERSHIP_BIDS;
             }
             _loopsWithoutOwner = 0;
-
-            // adebug BOOKMARK: the problem is that userB may deactivate and disown the Object
-            // but it may still be active for userA... who will store slightly non-zero velocities into EntityItem in the meantime
-            //
-            // It would be nice if we could ignore slight outgoing changes for unowned objects that WANT_DEACTIVATION until...
-            // (a) the changes exceed some threshold (--> bid for ownership) or...
-            // (b) they actually get deactivated (--> slam RigidBody positions to agree with EntityItem)
         } else if (_entity->getSimulatorID() == Physics::getSessionUUID()) {
             // we just inherited ownership, make sure our desired priority matches what we have
             upgradeOutgoingPriority(_entity->getSimulationPriority());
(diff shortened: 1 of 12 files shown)


libraries/physics/src/ThreadSafeDynamicsWorld.cpp
```diff
@@ -120,30 +120,41 @@ void ThreadSafeDynamicsWorld::synchronizeMotionState(btRigidBody* body) {
...
+                    _changedMotionStates.push_back(motionState);
+                    _activeStates.insert(motionState);
+                } else if (_lastActiveStates.find(motionState) != _lastActiveStates.end()) {
+                    // this object was active last frame but is no longer
+                    _deactivatedStates.push_back(motionState);
                 }
             }
         }
     }
+    _activeStates.swap(_lastActiveStates);
```
AndrewMeadows: We now track active objects so we can catch deactivation events and provide special treatment for entities that are being deactivated.

libraries/physics/src/EntityMotionState.cpp
```diff
@@ -97,6 +97,15 @@ void EntityMotionState::updateServerPhysicsVariables() {
...
 }
 
+void EntityMotionState::handleDeactivation() {
+    // copy _server data to entity
+    bool success;
+    _entity->setPosition(_serverPosition, success, false);
+    _entity->setOrientation(_serverRotation, success, false);
+    _entity->setVelocity(ENTITY_ITEM_ZERO_VEC3);
+    _entity->setAngularVelocity(ENTITY_ITEM_ZERO_VEC3);
+}
```
AndrewMeadows: The "special treatment" for deactivated entities is to make sure their velocities are zero and to restore their transform to the "last known position as known by the server".  If the entity's simulation was locally owned this last known transform is whatever was last sent to the entity-server.  If the entity's simulation was remotely owned it will be whatever was the last update from the server.

Hint: sometimes a remotely owned object would be deactivated and disowned, that data would arrive to a remote interface where the object would take a little longer to be deactivated... and in the meantime slight transform differences and small but non-zero velocities would be copied into the local EntityItem data.

libraries/physics/src/EntityMotionState.cpp
```diff
@@ -111,6 +120,8 @@ void EntityMotionState::handleEasyChanges(uint32_t& flags) {
                 flags &= ~Simulation::DIRTY_PHYSICS_ACTIVATION;
                 _body->setActivationState(WANTS_DEACTIVATION);
                 _outgoingPriority = 0;
+                const float ACTIVATION_EXPIRY = 3.0f; // something larger than the 2.0 hard coded in Bullet
+                _body->setDeactivationTime(ACTIVATION_EXPIRY);
```
AndrewMeadows: This is the second important change in this PR: when an entity's velocity goes to zero and its simulation ownership is removed by the remote interface, we force its local deactivation time to something beyond the timeout (2.0 seconds in Bullet) which helps (but does not guarantee!) that the object is actually deactivated in the local simulation.

The reason its deactivation is not guaranteed is because when the simulation is busy the EntityServer cannot not send the full state of the world at once (it is bandwidth limited), nor can it send the full state in one packet, so it is possible that a nearby entity is not properly updated in the local simulation which may cause this deactivating entity to collide with it (incorrectly, according to the remote simulation) and pick up a large enough velocity to reset its deactivation timer.

libraries/physics/src/EntityMotionState.cpp
```diff
@@ -221,12 +232,9 @@ void EntityMotionState::getWorldTransform(btTransform& worldTrans) const {
...
 
 // This callback is invoked by the physics simulation at the end of each simulation step...
-// iff the corresponding RigidBody is DYNAMIC and has moved.
+// iff the corresponding RigidBody is DYNAMIC and ACTIVE.
 void EntityMotionState::setWorldTransform(const btTransform& worldTrans) {
-    if (!_entity) {
-        return;
-    }
-
+    assert(_entity);
```
AndrewMeadows: The _entity is guaranteed to be valid in this context.

hifi-gustavo:
Build console output for commit: a16760278e5f84844604a8eda1e76dc379b19cfc available at: https://gist.github.com/dc1efae51acfb59e3a8696aae6fee6ab<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9851-a167602.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9851-a167602.dmg

AndrewMeadows:
This PR improves the compliance of a local simulation to the remote one (simulation owner).  Specifically, when the remote simulation owner decides that an entity has come to rest and should be deactivated the local simulation should now be more likely to actually follow suit: by deactivating the object and updating its final transform.

But... this PR doesn't solve ALL the problems --> it is still possible for a colliding pile of objects to settle remotely but not settle in the local simulation... which can cause the local simulation to take ownership of the pile and force it to activate on the remote simulation.  This has implications on how...

To test:

(1) Login UserA using this PR.  Enable visualization of physics simulation ownership:  **Developer --> Physics --> Highlight Simulation Ownership**

The sigils are cryptic (and upside down!) but here is a summary:  There are four sigils.  From left to right they are:

<1> Cyan downward arrow with circle around = data going out to EntityServer (only happens when local interface owns simulation)
<2> Red upward arrow with circle around it = data arriving from EntityServer
<3> Blue triangle with circle around it = object is **active** in local simulation
<4a> Blue head+torso icon (upside down) with circle around it = simulation owned by local interface
<4b> Red head+torso icon (upside down) with circle around it = simulation owned by remote interface

(2) Create a pile of dynamic objects that are all in close proximity to each other, such that when one object is bumped the rest of the pile activates.  Here is an optional script that creates a bunch of temporary boxes right in front of the avatar.  Only use if helpful:
```
"use strict";
/*jslint nomen: true, plusplus: true, vars: true*/
var Vec3, Quat, MyAvatar, Entities, Camera, Script, print;
var NUMBER_TO_CREATE = 100;
var LIFETIME = 300; // seconds
var SIZEX = 0.5;
var SIZEY = 1.5;
var SIZEZ = 0.5;
var TYPE = "Box";
var RATE_PER_SECOND = 800;
var SCRIPT_INTERVAL = 100;
var GRAVITY = { x: 0, y: -9.8, z: 0 };
var VELOCITY = { x: 0.0, y: 0, z: 0 };
var ANGULAR_VELOCITY = { x: 1, y: 1, z: 1 };
var DAMPING = 0.5;
var ANGULAR_DAMPING = 0.5;
var RANGE = 1;
var HOW_FAR_IN_FRONT_OF_ME = 9;
var HOW_FAR_UP = 3;
//
var totalCreated = 0;
var offset = Vec3.sum(Vec3.multiply(HOW_FAR_UP, Vec3.UNIT_Y),
                      Vec3.multiply(HOW_FAR_IN_FRONT_OF_ME, Quat.getFront(Camera.orientation)));
var center = Vec3.sum(MyAvatar.position, offset);
function randomVector(range) {
    return {
        x: (Math.random() -  0.5) * range.x,
        y: (Math.random() -  0.5) * range.y,
        z: (Math.random() -  0.5) * range.z
    };
}
//
Script.setInterval(function () {
    if (!Entities.serversExist() || !Entities.canRez()) {
        return;
    }
    if (totalCreated >= NUMBER_TO_CREATE) {
        print("Created " + totalCreated + " boxes.");
        Script.stop();
    }
    var i, numToCreate = RATE_PER_SECOND * (SCRIPT_INTERVAL / 1000.0);
    for (i = 0; (i < numToCreate) && (totalCreated < NUMBER_TO_CREATE); i++) {
        Entities.addEntity({
            type: TYPE,
            name: "box-" + totalCreated,
            position: Vec3.sum(center, randomVector({ x: RANGE, y: RANGE, z: RANGE })),
            dimensions: {x: SIZEX, y: SIZEY, z: SIZEZ},
            color: {red: Math.random() * 255, green: Math.random() * 255, blue: Math.random() * 255},
            velocity: VELOCITY,
            angularVelocity: Vec3.multiply(Math.random(), ANGULAR_VELOCITY),
            damping: DAMPING,
            angularDamping: ANGULAR_DAMPING,
            gravity: GRAVITY,
            dynamic: true,
            lifetime: LIFETIME
        });
        totalCreated++;
    }
}, SCRIPT_INTERVAL);
```
(3) Login UserB using some other build that does not have these changes in it:

(4) Create/bump the pile of objects and let it settle down.  Take turns bumping via UserA and then UserB.

When UserB owns the pile simulation and UserA is watching... UserA will be more likely to allow the pile to settle down immediately after UserB decides to release it.

When UserA owns the pile simulation and UserB is watching... UserB will be less likely than UserA to allow the pile to settle down... UserB may decide the pile is not finished and will volunteer to take ownership after UserA releases it.

AndrewMeadows:
I'm setting this to DO_NOT_MERGE.  I just realized that there is a change I should make.

hifi-gustavo:
Build console output for commit: 9c9eb879255d3136b321622d0b221196a27e96d9 available at: https://gist.github.com/dc1efae51acfb59e3a8696aae6fee6ab<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9851-9c9eb87.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9851-9c9eb87.dmg

AndrewMeadows:
Ok, ready to go.

hifi-gustavo:
Build console output for commit: bde2222dfd758292f8ac0ca6c3e26ec3dd6c492b available at: https://gist.github.com/dc1efae51acfb59e3a8696aae6fee6ab<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9851-bde2222.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9851-bde2222.dmg

OmegaHeron:
Tested Windows 10/Ubuntu 16.04

User A Windows 10, This PR
User B Ubuntu 16.04, DEV-6170
Stack - This PR running on Windows 10

First ran test using current master for both users, then user A this PR -- user B DEV-6170.

After multiple runs, would agree with eval given for test result desired outcome.

General smoke tests pass.  This, at least for me, isn't the easiest PR to evaluate pass/fail, but, after multiple runs of this and repeats with master, I'm convinced it's achieving goal.

PR #9906, opened 2017-03-15
Expose rendering in wireframe mode

samcake:
Simply expose to script a feature that we removed a while back... Wireframe rendering

The drawOpaque and draw Transparent passes now use a "global shapeKey" that is ORED with the individual item shapeKey so we can force global shapeKey flags such as "WIREFRAME".

THe wireframeENable value is taken from teh LightingModel object (which istsef can be configured from script).

TO test the Wireframe mode is working, from the script console call the line:
Render.getConfig("LightingModel").enableWireframe = true

And to go back to normal rendering
Render.getConfig("LightingModel").enableWireframe = false

### DIFF
--- a/libraries/render-utils/src/LightingModel.cpp
+++ b/libraries/render-utils/src/LightingModel.cpp
@@ -133,6 +133,7 @@ void LightingModel::setSpotLight(bool enable) {
 bool LightingModel::isSpotLightEnabled() const {
     return (bool)_parametersBuffer.get<Parameters>().enableSpotLight;
 }
+
 void LightingModel::setShowLightContour(bool enable) {
     if (enable != isShowLightContourEnabled()) {
         _parametersBuffer.edit<Parameters>().showLightContour = (float)enable;
@@ -142,6 +143,14 @@ bool LightingModel::isShowLightContourEnabled() const {
     return (bool)_parametersBuffer.get<Parameters>().showLightContour;
 }
 
+void LightingModel::setWireframe(bool enable) {
+    if (enable != isWireframeEnabled()) {
+        _parametersBuffer.edit<Parameters>().enableWireframe = (float)enable;
+    }
+}
+bool LightingModel::isWireframeEnabled() const {
+    return (bool)_parametersBuffer.get<Parameters>().enableWireframe;
+}
 MakeLightingModel::MakeLightingModel() {
     _lightingModel = std::make_shared<LightingModel>();
 }
@@ -167,6 +176,7 @@ void MakeLightingModel::configure(const Config& config) {
     _lightingModel->setSpotLight(config.enableSpotLight);
 
     _lightingModel->setShowLightContour(config.showLightContour);
+    _lightingModel->setWireframe(config.enableWireframe);
 }
 
 void MakeLightingModel::run(const render::SceneContextPointer& sceneContext, const render::RenderContextPointer& renderContext, LightingModelPointer& lightingModel) {
--- a/libraries/render-utils/src/LightingModel.h
+++ b/libraries/render-utils/src/LightingModel.h
@@ -64,6 +64,9 @@ public:
     void setShowLightContour(bool enable);
     bool isShowLightContourEnabled() const;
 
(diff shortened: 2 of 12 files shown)


hifi-gustavo:
Build console output for commit: ea2f1359bceb29a261a22487806aab0fdc0704a7 available at: https://gist.github.com/052df89f62a3a4a96a2f3f5b51547ca8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9906-ea2f135.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9906-ea2f135.dmg

ZappoMan:
This looks like a fail to me.. the primitives are not rendering as wireframe...

kevinmthomasse:
@ZappoMan sorry about that it didn't specify however I thought this was mesh-focused.  I should have asked for clarification.

samcake:
Good Point Mr Zappoman !
Looking why

samcake:
I actually see that as a feature....

samcake:
I added the wireframe mode to the primitive entities and and the polyvox entities.
I have a little issue with polyvox entities as they do not render the texture  color correctly.
I Think we can ship as is since this is  only a debugging tool

hifi-gustavo:
Build console output for commit: 0399249a03eb75cd1a27725ef41a0460e052bfff available at: https://gist.github.com/052df89f62a3a4a96a2f3f5b51547ca8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9906-0399249.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9906-0399249.dmg

libraries/render-utils/src/LightingModel.cpp
```diff
@@ -142,6 +143,14 @@ bool LightingModel::isShowLightContourEnabled() const {
     return (bool)_parametersBuffer.get<Parameters>().showLightContour;
 }
 
+void LightingModel::setWireframe(bool enable) {
+    if (enable != isWireframeEnabled()) {
+        _parametersBuffer.edit<Parameters>().enableWireframe = (float)enable;
```
sethalves: is casting to a (float) here really right?

libraries/render-utils/src/LightingModel.h
```diff
@@ -89,13 +92,12 @@ class LightingModel {
...
 
-        float showLightContour{ 0.0f }; // false by default
+        float showLightContour { 0.0f }; // false by default
 
         float enableObscurance{ 1.0f };
 
         float enableMaterialTexturing { 1.0f };
-
-        float spares{ 0.0f };
+        float enableWireframe { 0.0f }; // false by default
```
sethalves: why are bools floats, in this class?

libraries/render-utils/src/LightingModel.cpp
samcake: yes because the buffer Parameter member "enableWireframe" is a float scalar

libraries/render-utils/src/LightingModel.h
samcake: Because this struct layout is used in gpu  glsl and need to match exactly the same chunck of memory in the shader.
In the case of that structure which is a collection of flags, i wanted to go simple and have one scalar per boo, and i choose float to be able to have a simple way to convey "level" or "scale" kind of data

hifi-gustavo:
Build console output for commit: 4ef7fc61f05f684450523112eb0d800a703f571d available at: https://gist.github.com/052df89f62a3a4a96a2f3f5b51547ca8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9906-4ef7fc6.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9906-4ef7fc6.dmg

samcake:
I think we are good to go with this one

PR #9880, opened 2017-03-14
Fix avatars stuck in the sitting position after a domain disconnect

Atlante45:
## Test Plan:
- Test sitting down then standing up
- Test sitting down, jumping to other chairs, then standing up
- Test sitting on a chair already taken
- Test restarting the domain while seating
- Test jumping to a different domain while seating

[Bug](https://highfidelity.fogbugz.com/f/cases/3467/People-get-stuck-in-seated-animation-after-getting-up)

PR #9917, opened 2017-03-16
Cancel Avatar velocity when sitting down.

Atlante45:
- Added a mechanism to override an avatar's drive keys
- Canceled velocity of Avatar when sitting down

## Test plan:
- Import [this chair](https://gist.githubusercontent.com/Atlante45/81df7b998b768dcf3701ec5e5571e2ee/raw/30aeeb9a355716e3f8eb00df92fcfa8fbd2dcf94/green_chair.json) (Edit > Import Entities From URL)

- Try to sit while jumping/moving and make sure the avatar stays sitted.

This PR is based off [PR9880](https://github.com/highfidelity/hifi/pull/9880), so the diff won't be accurate until the later is merged.

[Bug](https://highfidelity.fogbugz.com/f/cases/3480/Avatar-pops-out-of-chair-when-trying-to-sit)

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -2091,14 +2091,78 @@ bool MyAvatar::getCharacterControllerEnabled() {
 }
 
 void MyAvatar::clearDriveKeys() {
-    for (int i = 0; i < MAX_DRIVE_KEYS; ++i) {
-        setDriveKey(i, 0.0f);
+    _driveKeys.fill(0.0f);
+}
+
+void MyAvatar::setDriveKey(int key, float val) {
+    try {
+        _driveKeys.at(key) = val;
+    } catch (const std::exception& exception) {
+        qCCritical(interfaceapp) << Q_FUNC_INFO << ": Index out of bounds";
+    }
+}
+
+float MyAvatar::getDriveKey(int key) const {
+    return isDriveKeyDisabled(key) ? 0.0f : getRawDriveKey(key);
+}
+
+float MyAvatar::getRawDriveKey(int key) const {
+    try {
+        return _driveKeys.at(key);
+    } catch (const std::exception& exception) {
+        qCCritical(interfaceapp) << Q_FUNC_INFO << ": Index out of bounds";
+        return 0.0f;
     }
 }
 
 void MyAvatar::relayDriveKeysToCharacterController() {
     if (getDriveKey(TRANSLATE_Y) > 0.0f) {
         _characterController.jump();
+    }
+}
+
+void MyAvatar::disableDriveKey(int key) {
+    try {
+        _disabledDriveKeys.set(key);
+    } catch (const std::exception& exception) {
+        qCCritical(interfaceapp) << Q_FUNC_INFO << ": Index out of bounds";
+    }
+}
+
+void MyAvatar::enableDriveKey(int key) {
+    try {
+        _disabledDriveKeys.reset(key);
+    } catch (const std::exception& exception) {
+        qCCritical(interfaceapp) << Q_FUNC_INFO << ": Index out of bounds";
+    }
+}
+
+void MyAvatar::disableDriveKeys(std::vector<int> key) {
+    try {
+        std::for_each(std::begin(key), std::end(key), [&](int val){
+            _disabledDriveKeys.set(val);
+        });
+    } catch (const std::exception& exception) {
+        qCCritical(interfaceapp) << Q_FUNC_INFO << ": Index out of bounds";
+    }
+}
+
+void MyAvatar::enableDriveKeys(std::vector<int> key) {
+    try {
+        std::for_each(std::begin(key), std::end(key), [&](int val) {
+            _disabledDriveKeys.reset(val);
+        });
+    } catch (const std::exception& exception) {
+        qCCritical(interfaceapp) << Q_FUNC_INFO << ": Index out of bounds";
+    }
+}
+
+bool MyAvatar::isDriveKeyDisabled(int key) const {
+    try {
+        return _disabledDriveKeys.test(key);
+    } catch (const std::exception& exception) {
+        qCCritical(interfaceapp) << Q_FUNC_INFO << ": Index out of bounds";
+        return false;
     }
 }
 
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -228,6 +228,21 @@ MyAvatar::~MyAvatar() {
     _lookAtTargetAvatar.reset();
 }
 
+void MyAvatar::registerMetaTypes(QScriptEngine* engine) {
+    QScriptValue value = engine->newQObject(this, QScriptEngine::QtOwnership, QScriptEngine::ExcludeDeleteLater | QScriptEngine::ExcludeChildObjects);
+    engine->globalObject().setProperty("MyAvatar", value);
+
+    QScriptValue driveKeys = engine->newObject();
+    auto metaEnum = QMetaEnum::fromType<DriveKeys>();
+    for (int i = 0; i < MAX_DRIVE_KEYS; ++i) {
+        driveKeys.setProperty(metaEnum.key(i), metaEnum.value(i));
+    }
+    engine->globalObject().setProperty("DriveKeys", driveKeys);
+
+    qScriptRegisterMetaType(engine, audioListenModeToScriptValue, audioListenModeFromScriptValue);
+    qScriptRegisterMetaType(engine, driveKeysToScriptValue, driveKeysFromScriptValue);
+}
+
 void MyAvatar::setOrientationVar(const QVariant& newOrientationVar) {
     Avatar::setOrientation(quatFromVariant(newOrientationVar));
 }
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: 48fa3b093b1d5a27fb9f792fea0e352604b9d01b available at: https://gist.github.com/d95489df691101789bd0da89f5829e76<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9917-48fa3b0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9917-48fa3b0.dmg

interface/src/avatar/MyAvatar.h
```diff
@@ -180,9 +184,15 @@ class MyAvatar : public Avatar {
...
     void clearDriveKeys();
-    void setDriveKeys(int key, float val) { _driveKeys[key] = val; };
+    void setDriveKey(int key, float val) { _driveKeys[key] = val; };
+    float getDriveKey(int key) const { return isDriveKeyCaptured(key) ? 0.0f : _driveKeys[key]; };
+    Q_INVOKABLE float getRawDriveKey(int key) const { return _driveKeys[key]; };
     void relayDriveKeysToCharacterController();
 
+    Q_INVOKABLE void captureDriveKey(int key) { _capturedDriveKeys.set(key); }
+    Q_INVOKABLE void releaseDriveKey(int key) { _capturedDriveKeys.reset(key); }
+    Q_INVOKABLE bool isDriveKeyCaptured(int key) const { return _capturedDriveKeys.test(key); }
```
AndrewMeadows: Is "captured" the right word here?  I know that people who design UI use the word "captured" to describe an event that is used in some UI context (usually by what has "focus") and is thereafter not revealed to other contexts, but in this case it seems to me that the drive key is being "disabled" rather than properly captured.  Maybe the language should be `disableDriveKey()`, `enableDriveKey()`, and `isDriveKeyEnabled()`?

Also, the keys are enumerated with integers, the mapping of which is non obvious.  Maybe the invokables should use string keys such as "TRANSLATE_X"?

In the current implementation it is not possible to disable all translation keys in a single script call.  Wouldn't it be cool if we could `MyAvatar.disableDriveKeys(["TRANSLATE_X", "TRANSLATE_Z"]);` ?

interface/src/avatar/MyAvatar.h
```diff
@@ -388,7 +397,9 @@ public slots:
     void clampScaleChangeToDomainLimits(float desiredScale);
     glm::mat4 computeCameraRelativeHandControllerMatrix(const glm::mat4& controllerSensorMatrix) const;
 
-    float _driveKeys[MAX_DRIVE_KEYS];
+    std::array<float, MAX_DRIVE_KEYS> _driveKeys;
```
AndrewMeadows: Switching to std::array<> is cool but... I don't see the advantage.  In particular this new method does no bounds checking and the std::array<> does not provide it either:

`Q_INVOKABLE float getRawDriveKey(int key) const { return _driveKeys[key]; };`

We have provided a way for JS to directly read memory outside of `_driveKeys`.

scripts/tutorials/entity_scripts/sit.js
```diff
@@ -183,47 +206,63 @@
...
+                this.lastTimeNoDriveKeys = Date.now();
+            }
+
+            // Allow some time for the IK to settle
+            if (ikError > MAX_IK_ERROR) {
+                var elapsed = now - this.sitDownTimestamp;
+                if (elapsed > IK_SETTLE_TIME) {
+                    shouldStandUp = true;
+                }
+            }
```
AndrewMeadows: Not a blocker, just an observation:

An alternative pattern to how `sitDownTimestamp` is used here is to use an expiry.  In this pattern: at the moment the timestamp would be taken we instead compute the time when something will expire:

`this.sitDownExpiry = now + IK_SETTLE_TIME`

Then when we want to know if time is up we can do this:

```
if (this.sitDownExpiry < now) {
    shouldStandUp = true;
}
```

The main reason I like this pattern is that it means we do the math at creation and not in the polling.  But I also find it simpler.  It could also work well for `lastTimeNoDriveKeys`: it could be replaced with a `standUpOnDriveExpiry` or something else with a better name.

claudioshg:
testing

claudioshg:
I'm having trouble with this PR. Out of the 5 chairs to download from the Marketplace, I can only sit in two. I followed the instructions for the other three chairs, but my avatar won't sit down on them. Out of the two chairs I can sit on, one chair does not let me move/jump while sitting, and the other chair does let me move/jump while sitting. How do I proceed?

Atlante45:
@claudioshg This is only for the new sit script I've been working on.
I should've added an example chair you can import, sorry about that.
I just added it to the test plan.

claudioshg:
### Desktop PC - Windows 10 Home - CPU 4.0GHz Intel i7 - GTX1070 - RAM 16GB - Tested with Oculus Rift

Tested: Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9917-48fa3b0.exe

I met with @Atlante45 in-world and resolved all issues. 
The goal of this PR was to see if the avatar stayed sitting after triggering the sit function on a chair, while in the middle of jumping. This PR is a PASS.

interface/src/avatar/MyAvatar.h
Atlante45: That makes sense, I think the reason I went with capture is because I wanted to stop the keys from propagating to the avatar while still being able to check it myself.
But I think that makes sense, will update.

I'm not really for using string. It's too easy to get it wrong and them you don't understand what's not working.
I want to expose the enum directly to JS but I'm having troubles doing so.

scripts/tutorials/entity_scripts/sit.js
Atlante45: I this makes sense for `sitDownExpiry` so I changed it, makes the code simpler, I agree.
I kept the drive key one though as it's more of a measure of time, and I think in that case, it makes more sense to actually measure it, rather than predicting it.

interface/src/avatar/MyAvatar.h
Atlante45: That's a good point, I should protect against bad access here.

hifi-gustavo:
Build console output for commit: 42316e713d9330e33d6cf3da4d08d3264cad442a available at: https://gist.github.com/d95489df691101789bd0da89f5829e76<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9917-42316e7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9917-42316e7.dmg

interface/src/avatar/MyAvatar.cpp
```diff
@@ -2102,6 +2121,51 @@ void MyAvatar::relayDriveKeysToCharacterController() {
...
+}
+
+bool MyAvatar::isDriveKeyDisabled(int key) const {
+    try {
+        return _disabledDriveKeys.test(key);
+    } catch (const std::exception& exception) {
+        qCCritical(interfaceapp) << Q_FUNC_INFO << ": Index out of bounds";
+        return false;
+    }
+}
```
AndrewMeadows: Maybe this should return 'true' in the failure case?  An invalid key is... 'disabled'?

If you disagree then consider changing the implementation of `MyAvatar::getDriveKey()`.  When it fails for bad key it fails twice.

hifi-gustavo:
Build console output for commit: 5a1ce4bb05292c80d0453486ff5ba17c563ca40b available at: https://gist.github.com/d95489df691101789bd0da89f5829e76<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9917-5a1ce4b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9917-5a1ce4b.dmg

hifi-gustavo:
Build console output for commit: 4a5e0b15ea6f5cb3db7a8001b8634196c5f61d29 available at: https://gist.github.com/d95489df691101789bd0da89f5829e76<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9917-4a5e0b1.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9917-4a5e0b1.dmg

Atlante45:
@AndrewMeadows I had to take out the functions that took arrays of driveKeys. I could not for the likes of me get Qt to convert it back and forth to ScriptValues.
And since we want that in master for the test tomorrow, I don't have the time to figure it out.
Since it's not essential I removed it for now, I can always revisit it later.

hifi-gustavo:
Build console output for commit: 698791295a6f2b60893619f1555a498084759e45 available at: https://gist.github.com/d95489df691101789bd0da89f5829e76<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9917-6987912.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9917-6987912.dmg

PR #9921, opened 2017-03-16
Fix possible crash in Menu::removeSeparator

jherico:
This PR fixes an assertion failure in `Menu::removeSeparator` in debug mode and a potential crash in release mode by preventing out of bounds array access.

## Dev Testing

Build and run the application in debug mode with default scripts.  A number of breakpoints will be hit inside the Qt WebEngine code... ignore them and hit continue on each one.  In master, eventually you will hit a breakpoint caused by an array index out of bounds error in the modified code because `textAt` will have a value of 0xffffffff which is not a valid index for the `menuActions` array.  In this branch, the array index out of bounds will not be triggered.

## Testing

Application behavior should remain unchanged.

### DIFF
--- a/libraries/ui/src/ui/Menu.cpp
+++ b/libraries/ui/src/ui/Menu.cpp
@@ -470,8 +470,8 @@ void Menu::removeSeparator(const QString& menuName, const QString& separatorName
     if (menu) {
         int textAt = findPositionOfMenuItem(menu, separatorName);
         QList<QAction*> menuActions = menu->actions();
-        QAction* separatorText = menuActions[textAt];
         if (textAt > 0 && textAt < menuActions.size()) {
+            QAction* separatorText = menuActions[textAt];
             QAction* separatorLine = menuActions[textAt - 1];
             if (separatorLine) {
                 if (separatorLine->isSeparator()) {


hifi-gustavo:
Build console output for commit: e265a0a1f88709bb201f0029bcb4c91e44edf422 available at: https://gist.github.com/7efef4cabb8f20a3648f493f355aa37b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9921-e265a0a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9921-e265a0a.dmg

ZappoMan:
@jherico - this test plan is not sufficient, you need to at a minimum write a test plan that exercises that features in the system that call this line of code.

jherico:
@ZappoMan are you asking for a test that can be done by a QA person or by a developer.  I'm fairly certain this will only show up in debug mode.  The issue is that `textAt` is being used as an index into an array before checking that it's a valid index.  This happens every time you start the application with default scripts running because `selectAudioDevice.js` calls `removeAudioMenus` and thus `Menu.removeSeparator("Audio", "Input Audio Device");` before the menu exists.  In release mode this most likely won't do anything at all because of the following conditional, but in debug mode the index is checked, so you get an assertion failure.

Nex-Pro:
Wields the mighty ax of MVS

libraries/ui/src/ui/Menu.cpp
```diff
@@ -470,8 +470,8 @@ void Menu::removeSeparator(const QString& menuName, const QString& separatorName
     if (menu) {
         int textAt = findPositionOfMenuItem(menu, separatorName);
         QList<QAction*> menuActions = menu->actions();
-        QAction* separatorText = menuActions[textAt];
         if (textAt > 0 && textAt < menuActions.size()) {
+            QAction* separatorText = menuActions[textAt];
             QAction* separatorLine = menuActions[textAt - 1];
             if (separatorLine) {
                 if (separatorLine->isSeparator()) {
```
AndrewMeadows: The initialization of `separatorText` could be moved down into this context where it is actually used.

AndrewMeadows:
I can confirm that this change fixes the crash mode.  I had independently discovered this crash on Sunday and figured out the same fix.  I can reliably crash on debug linux when running the defaultScripts.js and have been porting this fix to my builds as a workaround.

PR #9924, opened 2017-03-16
Use single frame buffer for local audio injectors

zzmp:


hifi-gustavo:
Build console output for commit: db956f4cc3a5e8a3e658de560e652d6ef941b358 available at: https://gist.github.com/9fc0c7daa78220b954eadcc2f83181a6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9924-db956f4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9924-db956f4.dmg

zzmp:
#### QA Plan

AUDIO-LATENCY-03 Test
Objective: Local injector latency should be <50msecs (i.e. unnoticeable)
Precondition: Enter any domain with limited audio.
// Load sounds
var dither = SoundCache.getSound("http://xiph.org/~xiphmont/demo/1kHz-0-dither.wav");
var A2 = SoundCache.getSound("http://mpassets.highfidelity.com/3fd92e5e-93cf-4bc1-b2f1-c6bae5629814-v1/A2.wav");
// Wait a moment for the sounds to load, then...

// Verify that a locally played sound has no noticeable latency, when it is the only sound played.
Audio.playSound(A2, { position: { x: 0, y: 0, z: 0 }, volume: 1 });

// Verify that a locally played sound has no noticeable latency, even if another local sound is already playing;
// this is the same as playing two sounds in a row:
for (var i = 0; i < 2; i++) Audio.playSound(A2, { position: { x: 0, y: 0, z: 0 }, volume: 1 });

// Verify that a local sound plays unbroken (the dither is much longer than A2).
Audio.playSound(dither, { position: { x: 0, y: 0, z: 0 }, volume: 0.1, localOnly: true });

OmegaHeron:
Tested Windows 10
General smoke tests pass.  Checked sounds in dev-playa/dev-welcome for any problems compared to master.
AUDIO-LATENCY-03 Test *Pass*
Compared to master this seems better, especially with multiple sounds.  In master there does at times seem to be lags to sounds starting when playing multiples while this PR does not.

Without setting up a way to measure time from keyboard press engaging sound to its output at speaker it's all reasonably subjective, but, having repeated test(s) multiple times between this PR and Master, subjectively, PR is better.

kencooke:
### Test Plan question:
Looks like the original code defaulted to 10 packets of ringbuffer, and now it starts at 1. Does this still work correctly when the buffers are manually set, and/or disabling the starvation-detect?

zzmp:
@kencooke yes, the local injectors have their own buffer, which should not be affected by manually setting the other buffers, nor by disabling starvation-detect. Those systems only affect the network audio.

PR #9938, opened 2017-03-17
Stop WebEntities Running Files and Commands

kunalgosar:
This addresses issues in: https://highfidelity.fogbugz.com/f/cases/3251/Raw-JS-files-are-not-displayed-in-web-entity

Previous behavior was that loading a script in WebEntity would ask all users in view to run that script. Similarly, loading a json (Entity Item) would rez that item once for each user in view (20 users cause item to rez 20x). Also, loading a url such as hifi://xyz would teleport all users in view to xyz.

This applies a profile and filter to requests originating from Browser/WebEntities to ensure that scripts are not unknowingly downloaded/run. Only plain text is allowed; this should be expanded to other file types as it is clear, which are causing bad behavior. 

It also stops WebEntity from running various files, such as scripts, entity items, avatar templates, teleports, etc.


kunalgosar:
Testing Instructions:

 - Load WebEntity and try various script/entity item URLs, they should not run for any one in the view. Javascript files should show up as plain text.
 - Running scripts, getting items through marketplace should follow default/old behavior.
 - Going to a .js file in browser or webEntity, if it is a general script (not necessary a hifi script) should not run for any one in view of that entity.
 - Old interceptor should follow default behavior: when opening a web entity or browser and going to highfidelity.io, you should be logged in if logged into the interface client.

hifi-gustavo:
Build console output for commit: 6a2f280c2e9e675263539d8ce12b1a13f2b477b0 available at: https://gist.github.com/f8b84255593183527870e358745ad13d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9938-6a2f280.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9938-6a2f280.dmg

OmegaHeron:
I started working on this, and am continuing to do so, but, this part of test is a bit hard to determine pass/fail on;

"Old interceptor should follow default behavior: when opening a web entity or browser and going to highfidelity.io, you should be logged in if logged into the interface client."

First - in master or this PR I can get no indication that going to highfidelity.io places me in a logged in state while I'm in a logged in state with interface.  The top menu bar doesn't display in a web entity (master or this pr) so there's no readily apparent way to see if you're logged in or not.  Going to a page that forces issue, like https://metaverse.highfidelity.io/user/domains , displays a login dialog, so -- at least for that it's failing (this PR and master) in that being logged in as my user name in interface is not logging me into HF web services from in world web entity....  unless the test quoted here is actually not valid then this is already a QA fail, but it is in master or PR.

kunalgosar:
I will look into this, I may have gotten the intended functionality wrong if that test is failing in master as well.

kunalgosar:
Hi @OmegaHeron, I looked into this and looks like there may be a bug breaking metaverse from logging you in automatically in the Release Build, which also caused this PR to fail. I will redefine the test to ensure that we can test just this PR:

Try opening a webEntity and the Browser, navigating to `metaverse.highfidelity.io/marketplace` and see that if you are logged into the interface, this logs you into marketplace. If this passes then the PR should pass for the test you mentioned.

OmegaHeron:
Thanks @kunalgosar - Will continue testing and try to finish QA on this.  Be a few hours until I can return to.

OmegaHeron:
Testing Windows 10

metaverse.highfidelity.io/marketplace - pass, it shows me logged in as my HF user name when I'm logged in via Interface.
Loading a url pointing to js - pass, shows source of JS, does not attempt to execute JS.
Loading a url pointing to FST file - pass, does nothing.
Loading a url pointing to FBX file - pass, does nothing.
Loading a url pointing to YouTube video - pass, plays video as expected.

Running scripts, getting items through marketplace should follow default/old behavior.  Pass.  Items obtained via marketplace continue to work and loading a script via URL works within Interface.

Using a second agent connected to sandbox via LAN - same results as above.

QA Pass for test(s) outlined here.

interface/src/networking/RequestFilters.cpp
```diff
@@ -0,0 +1,53 @@
...
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include "RequestFilters.h"
+
+#include <QtCore/QDebug>
+
+#include <AccountManager.h>
+
+bool static isAuthableHighFidelityURL(const QUrl& url) {
```
AndrewMeadows: Not a blocker but... the `static` declaration in this context makes me say "Wtf?".  For reference: http://www.osnews.com/images/comics/wtfm.jpg

After some research on the internet I've learned that declaring a local function as `static` is deprecated.  The new way to go forward is to use an "unnamed namespace":
```
namespace {
    bool static isAuthableHighFidelityURL(const QUrl& url) {
        // implementation details here...
    }
}

// the code down here can call the namespace method without special context
bool foo(const QUrl& url) {
    return isAuthableHighFidelityURL(const QUrl& url);
}
```

interface/src/Application.cpp
```diff
@@ -212,6 +213,7 @@ static const QString FST_EXTENSION  = ".fst";
 static const QString FBX_EXTENSION  = ".fbx";
 static const QString OBJ_EXTENSION  = ".obj";
 static const QString AVA_JSON_EXTENSION = ".ava.json";
+static const QString WEB_VIEW_TAG = "?noDowload=true";
```
kunalgosar: Fixed these typos, they were propagated through by copy-pastes.

interface/src/networking/RequestFilters.cpp
kunalgosar: Fixed this!

hifi-gustavo:
Build console output for commit: 5123fc5cb1f1ad74d59dbd59239a652f5b6b8f95 available at: https://gist.github.com/f8b84255593183527870e358745ad13d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9938-5123fc5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9938-5123fc5.dmg

interface/resources/qml/controls/WebView.qml
```diff
@@ -91,7 +96,7 @@ Item {
 
             // Required to support clicking on "hifi://" links
             if (WebEngineView.LoadStartedStatus == loadRequest.status) {
-                var url = loadRequest.url.toString();
+                var url = loadRequest.url.toString() + urlTag;
```
sethalves: Are we sure that the original url will never have query-args?  Can we use the api from http://doc.qt.io/qt-5/qurl.html on this `url` object and be more clever about appending to the query-args?

interface/src/Application.cpp
```diff
@@ -5605,7 +5608,9 @@ void Application::registerScriptEngineWithApplicationServices(ScriptEngine* scri
 
 bool Application::canAcceptURL(const QString& urlString) const {
     QUrl url(urlString);
-    if (urlString.startsWith(HIFI_URL_SCHEME)) {
+    if (urlString.endsWith(WEB_VIEW_TAG)) {
```
sethalves: Wont this fail if the url had its own query-args?

interface/src/networking/RequestFilters.cpp
```diff
@@ -0,0 +1,57 @@
...
+         static const QStringList HF_HOSTS = {
+             "highfidelity.com", "highfidelity.io",
+             "metaverse.highfidelity.com", "metaverse.highfidelity.io"
+         };
+
+         return url.scheme() == "https" && HF_HOSTS.contains(url.host());
+     }
+
+     bool isScript(const QString filename) {
+         return filename.contains(".js", Qt::CaseInsensitive);
```
sethalves: something.json ?

hifi-gustavo:
Build console output for commit: cc81bcab663dcc3698da8a0cd036afb98047b756 available at: https://gist.github.com/f8b84255593183527870e358745ad13d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9938-cc81bca.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9938-cc81bca.dmg

interface/resources/qml/controls/WebView.qml
kunalgosar: This is in the qml layer and so hard to use QUrl, but I added in a simple check here. There were other areas in the existing codebase that assumed there were no query-args, but I changed those.

interface/src/Application.cpp
kunalgosar: fixed this.

interface/src/networking/RequestFilters.cpp
kunalgosar: Added a check for json too, I was not sure what other files needed to be handled. But, I had discussed that we could trivially add them in if other files start causing problems.

hifi-gustavo:
Build console output for commit: 8749428ede5203c5f7e383a2d339b646a061dbf3 available at: https://gist.github.com/f8b84255593183527870e358745ad13d<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9938-8749428.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9938-8749428.dmg

PR #9942, opened 2017-03-18
Sit target experiment

Atlante45:
- Replace sit target with own avatar

hifi-snap-by-c-on-2017-03-17_19-25-20


hifi-gustavo:
Build console output for commit: 423ded70e0e0069801484dfeb75858f7b397a2e9 available at: https://gist.github.com/c58b99ffc9ace8cca5e7676f23190bc1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9942-423ded7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9942-423ded7.dmg

scripts/system/controllers/teleport.js
```diff
@@ -127,11 +127,20 @@ function Teleporter() {
...
     });
-    this.seatOverlay = Overlays.addOverlay("model", {
-        url: SEAT_MODEL_URL,
-        dimensions: TARGET_MODEL_DIMENSIONS,
-        visible: false
-    });
+
+    this.seatOverlay = null;
+    this.createSeatOverlay = function() {
+        this.seatOverlay = Entities.addEntity({
```
ZappoMan: so by making this an entity, you're making it visible to everyone? "local" entities are really "avatar attached entities"... they are basically new identity data, and result in everyone seeing them.

Is that what you wanted? Why?

Atlante45: No, I'm the only one that can see them. I tested with 2 clients and the second one wasn't seeing it.
It's not an avatar entity, it's a local entity. As I understand, it's just a normal entity that's not transmitted on the network, but maybe I'm wrong.

ZappoMan:
@Atlante45 -- I'm pretty sure (like 90% sure)... local entities are avatar entities (maybe poorly named)... they are visible to others... @sethalves correct me if I'm wrong.

scripts/system/controllers/teleport.js
ZappoMan: I'm pretty sure (like 90% sure) -- that local entities are avatar entities (it doesn't mean they are "attached" per se to your avatar, just that they are stored locally and transmitted through the avatar mixer... they are visible to other clients --- @sethalves correct me if I'm wrong.

If it wasn't visible to another client - it's some other bug... they should be visible.

sethalves: I agree with ZappoMan.  I don't know why you can't see them on another interface, but the extra "true" argument on addEntity makes it an avatar entity.  We don't have the concept of "local" entities, except for overlays.

ZappoMan: Was the reason you used entities was because overlay models don't support animations? If so, that's a bummer... but would be relatively easy to add that feature.

Atlante45: Yes, that's why. I kinda wish we'd implement that actually (local entities).

PR #9952, opened 2017-03-20
Return existing textures when matching hash for a previously found live texture

jherico:
This PR allows matching textures that are embedded in multiple model files or fetched from various URLs to be represented by the same live gpu::Texture object at runtime.  It also guards against some race and error conditions when loading or serializing KTX textures.

### DIFF
--- a/libraries/model-networking/src/model-networking/TextureCache.cpp
+++ b/libraries/model-networking/src/model-networking/TextureCache.cpp
@@ -188,6 +188,35 @@ NetworkTexturePointer TextureCache::getTexture(const QUrl& url, Type type, const
     return ResourceCache::getResource(url, QUrl(), &extra).staticCast<NetworkTexture>();
 }
 
+gpu::TexturePointer TextureCache::getTextureByHash(const std::string& hash) {
+    std::weak_ptr<gpu::Texture> weakPointer;
+    {
+        std::unique_lock<std::mutex> lock(_texturesByHashesMutex);
+        weakPointer = _texturesByHashes[hash];
+    }
+    auto result = weakPointer.lock();
+    if (result) {
+        qCWarning(modelnetworking) << "QQQ Returning live texture for hash " << hash.c_str();
+    }
+    return result;
+}
+
+gpu::TexturePointer TextureCache::cacheTextureByHash(const std::string& hash, const gpu::TexturePointer& texture) {
+    gpu::TexturePointer result;
+    {
+        std::unique_lock<std::mutex> lock(_texturesByHashesMutex);
+        result = _texturesByHashes[hash].lock();
+        if (!result) {
+            _texturesByHashes[hash] = texture;
+            result = texture;
+        } else {
+            qCWarning(modelnetworking) << "QQQ Swapping out texture with previous live texture in hash " << hash.c_str();
+        }
+    }
+    return result;
+}
+
+
 gpu::TexturePointer getFallbackTextureForType(NetworkTexture::Type type) {
     gpu::TexturePointer result;
     auto textureCache = DependencyManager::get<TextureCache>();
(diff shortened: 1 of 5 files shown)


hifi-gustavo:
Build console output for commit: 14c8c3f4db766ad3f7f09a3955bfea565cd8406c available at: https://gist.github.com/79b7c19c0c9a32ab98e95e19231c8f8c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9952-14c8c3f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9952-14c8c3f.dmg

OmegaHeron:
Windows 10

Did a general smoke test with my own content, dev-playa and dev-welcome.  Short of anything specific to look for I simply did a 1:1 compare between each on this PR and master for visual result.  Tested in desktop and DK2 HMD.  Looks same as master, nothing noticeable about FPS or CPU use or anything else raising red flags.

libraries/model-networking/src/model-networking/TextureCache.cpp
```diff
@@ -526,63 +570,43 @@ void ImageReader::read() {
...
-        auto ktx = gpu::Texture::serialize(*texture);
-        if (ktx) {
-            const char* data = reinterpret_cast<const char*>(ktx->_storage->data());
-            size_t length = ktx->_storage->size();
+        auto memKtx = gpu::Texture::serialize(*texture);
+        if (!memKtx) {
+            qCWarning(modelnetworking) << "Unable to serialize texture to KTX " << _url;
+        }
+
+        if (memKtx && textureCache) {
```
AndrewMeadows: This could be an `else if`.

libraries/model-networking/src/model-networking/TextureCache.cpp
```diff
@@ -188,6 +188,35 @@ NetworkTexturePointer TextureCache::getTexture(const QUrl& url, Type type, const
...
 
+gpu::TexturePointer TextureCache::getTextureByHash(const std::string& hash) {
+    std::weak_ptr<gpu::Texture> weakPointer;
+    {
+        std::unique_lock<std::mutex> lock(_texturesByHashesMutex);
+        weakPointer = _texturesByHashes[hash];
+    }
+    auto result = weakPointer.lock();
+    if (result) {
+        qCWarning(modelnetworking) << "QQQ Returning live texture for hash " << hash.c_str();
```
jherico: it's a prefix I use when debugging stuff so I can tail / grep my log file.  I usually remove it before pushing the code, but looks like I forgot.

libraries/model-networking/src/model-networking/TextureCache.cpp
jherico: technically yes, but the earlier if is just to box a warning.  I find it misleading to combine the two blocks into a single conditional.

libraries/model-networking/src/model-networking/TextureCache.cpp
```diff
@@ -188,6 +188,35 @@ NetworkTexturePointer TextureCache::getTexture(const QUrl& url, Type type, const
...
+    gpu::TexturePointer result;
+    {
+        std::unique_lock<std::mutex> lock(_texturesByHashesMutex);
+        result = _texturesByHashes[hash].lock();
+        if (!result) {
+            _texturesByHashes[hash] = texture;
+            result = texture;
+        } else {
+            qCWarning(modelnetworking) << "QQQ Swapping out texture with previous live texture in hash " << hash.c_str();
+        }
```
jherico: the expansion only injects a hash that will eventually be injected into the map anyway once the image is loaded.

PR #9960, opened 2017-03-21
Fix Ambient Lighting equation

samcake:
This pr fixes 2 issues in the ambient lighting equation we are using:
- The specular was multiplied TWICE by the ambient intensity... 
   this is just wrong, causing dimmer and dimmer reflections (compared to diffuse) as the ambient intensity go closer to 0.
   This is now fixed and should give a better balance of the metallic vs non metallic surfaces ambient lighting.

- THe spherical harmonics generation was not doing the correct rgb swizzling... So this is now fixed and the color of the ambient diffuse is matching the ambient map (or skymap).

This may cause turmoil in the cummonity as a domain might not look the same (to an accute observer) but this should be better and more accurate to what s expected. feel free to put complains below.

## TEST PLAN
Smoke test on performance (nothing should change)

The look of metallic and non metallic surfaces lit by ambient lighting (always true in a zone with a skymap) will be different in this PR compared to master and hopefully more accurate:

To be able to validate the new behavior, you need to run the tool script "debugDeferredLighting.js"
THis allows to disable the lighting equation piece by piece
FIrst, on the right column, disable everything but the first checkbox "ambient", you are now only lit with ambient light aka the ambient map from the Zone entity in which your avatar is. If there is no ambient map, then the same apply using the skybox map.

Second,  you can disable the albedo color to trouble your judgement, simply check off "albedo" from the bottom of the second column of the ui

Now you should see all the non metallic surfaces white-ish and the metallic be just what thery are, reflective. The surfaces are colored by the combination of the diffuse ambient (coming from the spherical harmonics genrated from the ambient map) and the specular response from that material (reflecting the ambient map more or less sharply depending on the roughness).

THis is working for both builds.
NOw you should be able to tell the difference between master and this pr:
- the diffuse ambient coloring in this PR is correct and matching the dominant colors of the ambient map (it s not in master)
- fulling around with the Zone ambient Intensity, the relative balance of reflections vs mostly diffuse should stay the same in terms of intensity. in master as you dim the ambient intensity closer to 0, the reflection loose way more intensity than diffuse, which is wrong.

Funny fact: in this PR, a perfect mirror surface (metallic, roughness to 0, albedo to white)in a zone with ambient intensity set to 0 should refelct the ambient map exactly the same as what you see in that ambient map used as a skybox map. (or you could comapre with an offline viewer and check this is the same image reflected with the same brightness).

TODO, provide snapshot of reference materials


hifi-gustavo:
Build console output for commit: e05abf2a2f36cb08b4fa9f371a5b1a4b359cbac9 available at: https://gist.github.com/02c70d5030c43055a9b7a4dee19635f8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9960-e05abf2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9960-e05abf2.dmg

hifi-gustavo:
Build console output for commit: e05abf2a2f36cb08b4fa9f371a5b1a4b359cbac9 available at: https://gist.github.com/02c70d5030c43055a9b7a4dee19635f8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9960-e05abf2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9960-e05abf2.dmg

AlanZimmerman:
Tested this using the method described above and compared it with master. Confirmed that metallic surfaces with very low roughness do appear to reflect more brightly the ambient map when the ambient intensity is near 1, and continue to have more visible reflections at lower ambient intensities. When using debugDeferredLighting.js, there is clearly an improvement in how the ambient map is used to light a model in this PR when compared with master. Also reviewed the lighting in this PR (both HMD and 2D) for popular domains (Welcome, Zaru, Archilogic) and saw small improvements in the lighting with no descernable difference in performance as compared with Master. PASS

Nex-Pro:
### Blue toned right hand side is this PR  left is 6183

I overlaid images and made a diagonal mask
dev_v_thisbuild

### This pr
hifi-snap-by-piper-peppercorn-on-2017-03-21_22-18-26

### 6183
hifi-snap-by-piper-peppercorn-on-2017-03-21_22-19-56

PR #9976, opened 2017-03-22
Worklist #21215 - Removing Duplicate Status Icons for Submeshes

mittens:
Based on some ideas/notes (and... well... most of the work) from @AndrewMeadows, I updated Model.h|cpp (And CauterizedModel.cpp) to prevent duplicate status icons showing per sub-mesh in an entity. 

Still oodles of icons, but they're now limited to one per entity.


hifi-gustavo:

mittens:
Modified the PR to exclude a file (though a benign one) from a different PR I was working on.

libraries/render-utils/src/Model.cpp
```diff
@@ -576,39 +576,39 @@ bool Model::addToScene(std::shared_ptr<render::Scene> scene,
...
             bool hasTransparent = false;
             size_t verticesCount = 0;
-            foreach(auto renderItem, _modelMeshRenderItemsSet) {
+            foreach(auto renderItem, _modelMeshRenderItems) {
                 auto item = scene->allocateID();
                 auto renderPayload = std::make_shared<ModelMeshPartPayload::Payload>(renderItem);
-                if (statusGetters.size()) {
+				if (_modelMeshRenderItemsMap.empty() && statusGetters.size()) {
                     renderPayload->addStatusGetters(statusGetters);
                 }
```
AndrewMeadows: The real fix is here: only add a "statusGetter" for the **first** renderItem of the model.

In order for this to work such that the same renderItem always gets the status icons no matter the model instance or the login session... the _`modelMeshRenderItemsSet` container must be changed a `QVector` rather than a `QSet` (the sort order of the `QSet` was effectively random: based on internal variables).  The good news is that we were already using that container like a _vector_ rather than a _set_.

The rest of the diff is name changes to make the code more correct.

AndrewMeadows:
ok to test

hifi-gustavo:
Build console output for commit: c71e57753be6ff9daf31e8404b84f53b7978206c available at: https://gist.github.com/f8cd0397cc9850d55bd0d224d355bb25<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9976-c71e577.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9976-c71e577.dmg

themelissabrown:
icons should now be limited to one per entity.

OmegaHeron:
This looks generally correct now - I'm no expert on the physics indicators, but...

QA Pass

Master looking at pingpong gun in dev-welcome at hifi://dev-welcome/109.468,3.81585,-164.53/0,-0.0432453,0,0.999065

hifi-snap-by-omegaheron-on-2017-03-23_14-27-22

PR looking at same

hifi-snap-by--on-2017-03-23_14-27-50

Seems reasonably compelling evidence that problem is solved.  I also compared master to PR while walking around dev-welcome trying to notice anything that seemed wrong and could find no case where indicators seemed to be missing in PR that should be present.

Grabbing and spawning a new gun showed single icons for each presented vs master that shows multiples of each status icon.

libraries/render-utils/src/Model.h
```diff
@@ -373,11 +373,11 @@ public slots:
 
     static AbstractViewStateInterface* _viewState;
 
-    QSet<std::shared_ptr<MeshPartPayload>> _collisionRenderItemsSet;
-    QMap<render::ItemID, render::PayloadPointer> _collisionRenderItems;
+    QVector<std::shared_ptr<MeshPartPayload>> _collisionRenderItems;
```
samcake: Could you move to std container too here ?
Not needed but cool,
we re trying to stop using the qt containers when possible

PR #9979, opened 2017-03-23
Fix crash in render-perf-test

jherico:
The render-perf tool does not provide a scripting engine to the entities it renders, so the code must be sensitive to the possibility that `_entitiesScriptEngine` may be null.  The code here does that imperfectly.  `shouldLoad` is only true if `_entitiesScriptEngine` is non null, however, because of the placement of the or clause, if `scriptUrl` is empty, then the conditional block will execute even if there is no script engine.  Line 1000 here will immediately crash.


hifi-gustavo:
Build console output for commit: 944255a37b4bf8e3c8f16500231f876279e64b16 available at: https://gist.github.com/61c97143bca6fb8a89b17a3010559782<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9979-944255a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR9979-944255a.dmg

PR #10034, opened 2017-03-28
bad serialization of Avatar Entity parentID

davidkelly:
Well, really just if you didn't give it one.  In that case, you would end up with `parentID: null` in the avatar entity json, which leads to the error in [this bug](https://highfidelity.fogbugz.com/f/cases/3536/Avatar-Entities-produce-lots-of-errors-in-logs), when deserializing it on the other clients.  The effect is: you create the avatar entity, and every other client in that domain sees this error in their logs.  And again every time you update it.  The fix is to not call jsonObject["parentID"] unless that key really is there, because it is there afterward (as a QJsonValue::Undefined).

### DIFF
--- a/libraries/entities/src/EntityEditPacketSender.cpp
+++ b/libraries/entities/src/EntityEditPacketSender.cpp
@@ -60,8 +60,10 @@ void EntityEditPacketSender::queueEditAvatarEntityMessage(PacketType type,
 
     // the ID of the parent/avatar changes from session to session.  use a special UUID to indicate the avatar
     QJsonObject jsonObject = jsonProperties.object();
-    if (QUuid(jsonObject["parentID"].toString()) == _myAvatar->getID()) {
-        jsonObject["parentID"] = AVATAR_SELF_ID.toString();
+    if(jsonObject.contains("parentID")) {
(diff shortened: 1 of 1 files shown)


davidkelly:
Test Plan:

Have 2 clients in a particular domain
On one of them, in the console, create an avatar entity:  `Entities.addEntity({type: "Box", color: {red: 0, green:255, blue:0}}, true);` will create a green box as an avatar entity (that's the true as the second parameter) at 0,0,0 for instance.  
If you are in master, the other client will have `[DEBUG] [hifi.shared] unhandled QScript type QVariant::Invalid` in the interface logs
If you do this where the client that adds (or edits for that matter) the entity is in this branch, no error.

hifi-gustavo:
Build console output for commit: 4b40633104ff76e09bf7d4c1abd81b4765da62e0 available at: https://gist.github.com/50e31416bf1ba68ab72bb14137eae26b<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10034-4b40633.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10034-4b40633.dmg

PR #10049, opened 2017-03-29
Avoid invisible avatars

AndrewMeadows:
* theoretical fix for bug that causes some avatars to be invisible for everybody

### DIFF
--- a/libraries/avatars/src/AvatarData.cpp
+++ b/libraries/avatars/src/AvatarData.cpp
@@ -1495,6 +1495,9 @@ void AvatarData::processAvatarIdentity(const Identity& identity, bool& identityC
         setAvatarEntityData(identity.avatarEntityData);
         identityChanged = true;
     }
+    // flag this avatar as non-stale by updating _averageBytesReceived
+    const int BOGUS_NUM_BYTES = 1;
+    _averageBytesReceived.updateAverage(BOGUS_NUM_BYTES);
 }
 
 QByteArray AvatarData::identityByteArray() const {
--- a/assignment-client/src/avatars/AvatarMixer.cpp
+++ b/assignment-client/src/avatars/AvatarMixer.cpp
@@ -186,8 +186,8 @@ void AvatarMixer::start() {
 }
 
 
-// NOTE: nodeData->getAvatar() might be side effected, must be called when access to node/nodeData
-// is guarenteed to not be accessed by other thread
+// NOTE: nodeData->getAvatar() might be side affected, must be called when access to node/nodeData
+// is guaranteed to not be accessed by other thread
 void AvatarMixer::manageDisplayName(const SharedNodePointer& node) {
     AvatarMixerClientData* nodeData = reinterpret_cast<AvatarMixerClientData*>(node->getLinkedData());
     if (nodeData && nodeData->getAvatarSessionDisplayNameMustChange()) {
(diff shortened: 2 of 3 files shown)


hifi-gustavo:
Build console output for commit: c5eb2e064200814fc94a8f106a7d415825161961 available at: https://gist.github.com/fdcd2e322d969afe309fff8cfabed163<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10049-c5eb2e0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10049-c5eb2e0.dmg

libraries/avatars/src/AvatarData.cpp
```diff
@@ -1495,6 +1495,9 @@ void AvatarData::processAvatarIdentity(const Identity& identity, bool& identityC
         setAvatarEntityData(identity.avatarEntityData);
         identityChanged = true;
     }
+    // flag this avatar as non-stale by updating _averageBytesReceived
+    const int BOGUS_NUM_BYTES = 1;
+    _averageBytesReceived.updateAverage(BOGUS_NUM_BYTES);
```
AndrewMeadows: The fix is to update AvatarData::_averageBytesRecieved because there is a side effect: it gets a new timestamp on the update.  This timestamp is tested for expiry by AvatarData::shouldDie() which is used to see if the avatar is still being updated by the AvatarMixer or if it should be deleted.

Note: this is the theoretical fix.  I wasn't able to reproduce the bug but I did verify that the "identity data" is sent first, followed by the "joint data" when the avatar first shows up to everyone else.  The theory goes like this:

When the "joint data" and the "identity data" were split apart (back in January?) the _averageBytesRecieved.updateAverage() side effect was kept in one of the update handlers (joints) and was not copied over to the other handler (identity).  Since the hidden timestamp is initialized to 0, if an identity packet is received but the joint packet arrives after too long of a delay then the interface main loop has a chance to check to see if the avatar is stale, and will delete it before the joint data arrives.  When the joint data does arrive the avatar is recreated and added to the list but it never gets another identity packet because those are only sent when identity info changes.

AndrewMeadows:
Now that I have a working theory I can suggest a possible semi-reliable method for reproducing the bug:

Using some previous PR where the bug is still happening...

(1) Fool your interface into running the main loop at a very high frame rate.  I think there is a way to do this by disabling vsync or something.

(2) UserA: login to a domain with lots of other avatars (it helps to have the AvatarMixer busy).

(3) Login with UserB: can UserA see UserB?  With the high frame rate it is more likely that the UserA's interface will decide to delete the first instance of UserB before the joint data arrives, which should reproduce the bug.  If not... try several times more.

(4) If UserB manages to arrive invisible according to UserA then Yay! you have reproduced the bug.  Test again with UserA using this PR to see if it never happens.

If you couldn't reproduce the bug then... Yay! the bug is really hard to reproduce!  Pass this PR anyway since it probably won't hurt.

AndrewMeadows:
build this please

hifi-gustavo:
Build console output for commit: c5eb2e064200814fc94a8f106a7d415825161961 available at: https://gist.github.com/fdcd2e322d969afe309fff8cfabed163<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10049-c5eb2e0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10049-c5eb2e0.dmg

PR #10052, opened 2017-03-29
Set z-order of comboboxes to make sure lowest will no overlap higher …

vladest:
…ones

Test plan:

- open Developer->Graphics menu
- make sure higher comboboxes not overlapped by lower ones

### DIFF
--- a/interface/resources/qml/hifi/tablet/tabletWindows/preferences/Section.qml
+++ b/interface/resources/qml/hifi/tablet/tabletWindows/preferences/Section.qml
@@ -89,14 +89,17 @@ Preference {
             if (categoryPreferences) {
                 console.log("Category " + root.name + " with " + categoryPreferences.length + " preferences");
                 for (var j = 0; j < categoryPreferences.length; ++j) {
-                    buildPreference(categoryPreferences[j]);
+                    //provide component position within column
+                    //lowest numbers on top
+                    buildPreference(categoryPreferences[j], j);
                 }
             }
         }
 
-        function buildPreference(preference) {
+        function buildPreference(preference, itemNum) {
             console.log("\tPreference type " + preference.type + " name " + preference.name)
             var builder;
+            var zpos;
             switch (preference.type) {
                 case Preference.Editable:
                     checkBoxCount = 0;
@@ -136,11 +139,14 @@ Preference {
                 case Preference.ComboBox:
                     checkBoxCount = 0;
                     builder = comboBoxBuilder;
+                    //make sure that combo boxes sitting higher will have higher z coordinate
+                    //to be not overlapped when drop down is active
+                    zpos = root.z + 1000 - itemNum
                     break;
             };
 
             if (builder) {
-                preferences.push(builder.createObject(contentContainer, { preference: preference, isFirstCheckBox: (checkBoxCount === 1) }));
+                preferences.push(builder.createObject(contentContainer, { preference: preference, isFirstCheckBox: (checkBoxCount === 1) , z: zpos}));
             }
         }
     }


hifi-gustavo:
Build console output for commit: 98221936448cdcf757576acbb257049eca826331 available at: https://gist.github.com/c70766e9b24f70a2f5ced5c4e10b7d70<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10052-9822193.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10052-9822193.dmg

interface/resources/qml/hifi/tablet/tabletWindows/preferences/Section.qml
```diff
@@ -136,11 +139,14 @@ Preference {
                 case Preference.ComboBox:
                     checkBoxCount = 0;
                     builder = comboBoxBuilder;
+                    //make sure that combo boxes sitting higher will have higher z coordinate
+                    //to be not overlapped when drop down is active
+                    zpos = root.z + 1000 - itemNum
```
zfox23: I believe this fixes this particular ComboBox, but does not fix the issue with ComboBox dropdowns as a whole. The fix for ComboBox Z-Ordering should go in ComboBox.qml.

vladest: This fixes issue with ambient occlusion settings when lower combobox have bigger z-order (since its created later)
Combobox itself cannot set its z-position since the combobox have no idea about other controls in particular window
What the issue with Combobox dropdown?

Nex-Pro: Fast response guys, I only posted about this a few days ago.

Nex-Pro:
### PASS

Specific test meets expectations.

I can now enable/disable ambient occlusion in desktop and HMD.

interface/resources/qml/hifi/tablet/tabletWindows/preferences/Section.qml
ctrlaltdavid: I think this is a good solution; can't see an obvious way of solving the z-order issue in ComboBox.qml itself.

zfox23: This screenshot shows what happens when using the new ComboBox.qml:
image

This screenshot shows what happens when using the old TabletComboBox.qml (without any other changes, including to positioning or z-order):
image

Something different is happening in TabletComboBox such that z-ordering is not an issue.

vladest: Zach, can you please provide details, how I can reproduce the issue (what is the screen, how to reach it), I'll take a look

zfox23: It's not totally trivial to reproduce this right now - sorry about that :)
1. Pull the latest code from the highfidelity/PAL_v2 branch.
2. Compile and run
3. Open the PAL, be happy about the evolving design, then use the drop-down menu in the top left of the window
4. Notice that the ComboBox appears and acts as you'd expect - this is using the old TabletComboBox.qml
5. Open PAL.qml in your build directory and replace `HifiControlsUit.TabletComboBox` with `HifiControlsUit.ComboBox`.
6. Reload Interface
7. Notice that the PAL's ComboBox now exhibits the issues shown above.

vladest: But I did not changed ComboBox, I've changed Section.qml
In Section.qml in tablet-ui branch uses exactly HifiControlsUit.ComboBox
so, its rather incorrect merging or something else
anyway, I'll check PAL_v2 branch to be sure

vladest: just checked. PAL_v2 branch does not contains my changes in Section.qml

zfox23: That's right - but Section.qml has no bearing on Pal.qml.

I shouldn't have to change the z-ordering of my ComboBox in PAL.qml if the previous TabletComboBox worked correctly. I'd like to avoid forcing the z parameter of any element in my QML, as that usually indicates some other problem.

vladest: ok, then I just cant understand the problem
if you build a preference page using several comboboxes the one created later will always have higher z-order in case of overlapping and  If create controls from top to bottom. my PR fixes exact this problem. it should not influence on other non-preference cases

zfox23: I'm not creating a preference page, though - Section.qml has no bearing on Pal.qml. This z-ordering problem will occur with all new ComboBoxes that aren't part of a preferences pane, will it not?

vladest: no, it will occur only in preferences pages, which built automatically, so it should not influence on Pal.qml

zfox23: I'm confused - the z-ordering problem is clearly present in Pal.qml AND present in the existing Section.qml.

This PR only fixes the z-ordering problem related to ComboBoxes in Section.qml, but does not fix the z-ordering problem related to ComboBoxes in Pal.qml.

vladest: exactly
if there is a problem with ComboBox drawing, its should be some other issue
There is also text missed, so, probably its not only z-order

vladest: Also, I've slightly investigate the issue:
both TabletComboBox and ComboBox are shown under Connections tab
the only reason, why there is empty list shown in ComboBox is line 219 in ComboBox.qml
which should be:
                        text: listView.model[index] ? listView.model[index] : (listView.model.get(index).text ? listView.model.get(index).text : "")

zfox23: Nice find! I've made that change on my PAL branch, but that seems like it deserves a fix beforehand.

OK, I concede - for now, I'm going to manually set the drawing order of my ComboBox within Pal.qml (since it will soon be replaced with a not-ComboBox). Dev-approved, with the caveat that we might run into ComboBox z-ordering issues in the future.

PR #10067, opened 2017-03-30
Fix Entities.setLocalJointTranslations

humbletim:
This PR fixes a typo that was preventing joint translations set with `Entities.setLocalJointTranslations` from being seen by others.

### Testing:

Two instances of Interface are needed -- a "transmitting" and a "receiving" side.  Only the receiving side actually has to be running this PR (in case that makes it easier to find someone to test with).

* Launch two instances somewhere with rez rights.
* On the transmitting side, run this Client script:
```javascript
var id = Entities.addEntity({
     type: 'Model',
     position: MyAvatar.position,
     modelURL: MyAvatar.skeletonModelURL,
     lifetime: 600,
});
var interval = Script.setInterval(function(){
  Entities.setLocalJointTranslations(id, [{ x: -100 * Math.random(), y: 0, z: 0 }]);
}, 1000);
```
* On the receiving side, while running this PR, you should see the T-pose Model Entity bouncing around randomly.

(Same test could be run using a recent beta on the receiving side to confirm the functionality was broken before.)

### DIFF
--- a/libraries/entities/src/ModelEntityItem.cpp
+++ b/libraries/entities/src/ModelEntityItem.cpp
@@ -429,7 +429,7 @@ void ModelEntityItem::setJointTranslations(const QVector<glm::vec3>& translation
         for (int index = 0; index < translations.size(); index++) {
             if (_localJointTranslationsSet[index]) {
                 _localJointTranslations[index] = translations[index];
-                _localJointTranslationsSet[index] = true;
+                _localJointTranslationsDirty[index] = true;
             }
         }
     });


hifi-gustavo:
Build console output for commit: d02fd06305b33242fbbfd918abfcd3dea23033a6 available at: https://gist.github.com/d2507b4893cb28bd00bbb1a9c0379429<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10067-d02fd06.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10067-d02fd06.dmg

sethalves:
D'OH

OmegaHeron:
Tested Windows 10 and Ubuntu 16.04
QA Pass

Test 1
Windows and Linux running Master 6325 on both as control
Observed: A copy of my AV was placed over my "live" AV in T-Pose doing nothing (Same observation from both A/B perspectives).

Test 2
Machine A - Win 10, This PR as transmitter
Machine B - Ubuntu 16.04, This PR as receiver
Observed:  A copy of my AV was placed over my "live" AV in T-Pose - after a few seconds it began to move randomly.

*Note:  This script and test could be added as a standard test like;
Interface A and B this build.  On Interface A run script (some URL reference to script above) while watching from Interface B.  Pass = a copy of avatar in use by user A should appear in T-Pose and within a second or two begin to move randomly from its original location.  Fail = Copy of user A avatar never appears or it appears but never moves.

PR #10079, opened 2017-03-31
Thread safety for property getter/setters on entities

jherico:
This is a first pass at making the property entity getters and setters thread safe.  Currently there is no guard against multiple threads accessing a property concurrently, and this is the cause of https://highfidelity.fogbugz.com/f/cases/3112 (reading a QString that is in the middle of being written can cause heap corruption).

This makes changes to the `EntityItem` base class and to the `TextEntityItem` class, but significantly more work is going to be required to secure all the different types.  However these changes do appear to specifically resolve the above bug

## Testing

* In a domain with edit rights create a text entity and record it's ID
* Write a small script similar to the following using the text entity ID

```
// REPLACE WITH YOUR TEXT ENTITY ID
var id = "{5b3d737d-dad1-4b5a-bd8a-cb72afe38aa1}"; 
for (var i = 0; i < 1000; ++i) {
  for (var j = 0; j < 1000; ++j) {
    Entities.editEntity(id, { text: "hello " + j });
  }
}
Script.stop();
```

* Execute the script

On the current production build, interface will likely crash after a short time with a bugsplat window.   This PR build should be able to successfully run the script to completion without crashing.

### DIFF
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -683,7 +683,7 @@ int EntityItem::readEntityDataFromBuffer(const unsigned char* data, int bytesLef
                 // However, for now, when the server uses a newer time than what we sent, listen to what we're told.
                 if (overwriteLocalData) weOwnSimulation = false;
             } else if (_simulationOwner.set(newSimOwner)) {
-                _dirtyFlags |= Simulation::DIRTY_SIMULATOR_ID;
+                markDirtyFlags(Simulation::DIRTY_SIMULATOR_ID);
                 somethingChanged = true;
                 // recompute weOwnSimulation for later
                 weOwnSimulation = _simulationOwner.matchesValidID(myNodeID);
@@ -695,19 +695,19 @@ int EntityItem::readEntityDataFromBuffer(const unsigned char* data, int bytesLef
             weOwnSimulation = true;
             if (!_simulationOwner.isNull()) {
                 // someone else really did own it
-                _dirtyFlags |= Simulation::DIRTY_SIMULATOR_ID;
+                markDirtyFlags(Simulation::DIRTY_SIMULATOR_ID);
                 somethingChanged = true;
                 _simulationOwner.clearCurrentOwner();
             }
         } else if (newSimOwner.matchesValidID(myNodeID) && !_hasBidOnSimulation) {
             // entity-server tells us that we have simulation ownership while we never requested this for this EntityItem,
             // this could happen when the user reloads the cache and entity tree.
-            _dirtyFlags |= Simulation::DIRTY_SIMULATOR_ID;
+            markDirtyFlags(Simulation::DIRTY_SIMULATOR_ID);
             somethingChanged = true;
             _simulationOwner.clearCurrentOwner();
             weOwnSimulation = false;
         } else if (_simulationOwner.set(newSimOwner)) {
-            _dirtyFlags |= Simulation::DIRTY_SIMULATOR_ID;
+            markDirtyFlags(Simulation::DIRTY_SIMULATOR_ID);
             somethingChanged = true;
             // recompute weOwnSimulation for later
             weOwnSimulation = _simulationOwner.matchesValidID(myNodeID);
@@ -994,7 +994,7 @@ SharedSoundPointer EntityItem::getCollisionSound() {
 }
 
 void EntityItem::simulate(const quint64& now) {
-    if (0 == getLastSimulated()) {
+    if (getLastSimulated() == 0) {
         setLastSimulated(now);
     }
 
@@ -1042,7 +1042,7 @@ void EntityItem::simulate(const quint64& now) {
     if (!stepKinematicMotion(timeElapsed)) {
         // this entity is no longer moving
         // flag it to transition from KINEMATIC to STATIC
-        _dirtyFlags |= Simulation::DIRTY_MOTION_TYPE;
+        markDirtyFlags(Simulation::DIRTY_MOTION_TYPE);
         setAcceleration(Vectors::ZERO);
     }
     setLastSimulated(now);
@@ -1294,7 +1294,7 @@ void EntityItem::getAllTerseUpdateProperties(EntityItemProperties& properties) c
 }
 
 void EntityItem::pokeSimulationOwnership() {
-    _dirtyFlags |= Simulation::DIRTY_SIMULATION_OWNERSHIP_FOR_POKE;
+    markDirtyFlags(Simulation::DIRTY_SIMULATION_OWNERSHIP_FOR_POKE);
     auto nodeList = DependencyManager::get<NodeList>();
     if (_simulationOwner.matchesValidID(nodeList->getSessionUUID())) {
         // we already own it
@@ -1306,7 +1306,7 @@ void EntityItem::pokeSimulationOwnership() {
 }
 
 void EntityItem::grabSimulationOwnership() {
-    _dirtyFlags |= Simulation::DIRTY_SIMULATION_OWNERSHIP_FOR_GRAB;
+    markDirtyFlags(Simulation::DIRTY_SIMULATION_OWNERSHIP_FOR_GRAB);
     auto nodeList = DependencyManager::get<NodeList>();
     if (_simulationOwner.matchesValidID(nodeList->getSessionUUID())) {
         // we already own it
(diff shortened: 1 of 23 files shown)


hifi-gustavo:
Build console output for commit: 350a6d49671bad5eb32db3338473a92ba8faa740 available at: https://gist.github.com/3c2c8aa4edd311798882e19fe7b77208<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10079-350a6d4.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10079-350a6d4.dmg

libraries/entities/src/EntityItem.cpp
```diff
@@ -941,43 +945,60 @@ void EntityItem::setMass(float mass) {
...
+        });
     }
-    return _collisionSound;
+    return result;
 }
 
 void EntityItem::simulate(const quint64& now) {
-    if (_lastSimulated == 0) {
-        _lastSimulated = now;
+    if (0 == getLastSimulated()) {
```
ZappoMan: coding standard says this is a NO NO

jherico: the use of the literal on the left you mean?

ZappoMan: correct - explicitly banned in the coding standard

libraries/entities/src/EntityItem.cpp
```diff
@@ -1189,10 +1213,10 @@ EntityItemProperties EntityItem::getProperties(EntityPropertyFlags desiredProper
     EntityItemProperties properties(propertyFlags);
     properties._id = getID();
     properties._idSet = true;
-    properties._created = _created;
-    properties._lastEdited = _lastEdited;
-    properties.setClientOnly(_clientOnly);
-    properties.setOwningAvatarID(_owningAvatarID);
+    properties._created = getCreated();
```
ZappoMan: if you were concerned about locking speed you could do a single read lock for this entire method.

jherico: received wisdom is to optimize for minimal locking duration, rather than minimal frequency of locks.

ZappoMan: Cool

hifi-gustavo:
Build console output for commit: 92e0a48d63ff7eb46503364fea2b918729a08c51 available at: https://gist.github.com/3c2c8aa4edd311798882e19fe7b77208<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10079-92e0a48.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10079-92e0a48.dmg

hifi-gustavo:
Build console output for commit: 5e8e32a9cb8802181c9e952ccec4733623f1bb65 available at: https://gist.github.com/3c2c8aa4edd311798882e19fe7b77208<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10079-5e8e32a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10079-5e8e32a.dmg

hifi-gustavo:
Build console output for commit: 656983e33d834b040498a62f9f3bfd20334e9f7f available at: https://gist.github.com/3c2c8aa4edd311798882e19fe7b77208<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10079-656983e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10079-656983e.dmg

hifi-gustavo:
Build console output for commit: cd7c1643a3492c8ed4c2824d98ed2cee8d67207e available at: https://gist.github.com/3c2c8aa4edd311798882e19fe7b77208<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10079-cd7c164.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10079-cd7c164.dmg

hifi-gustavo:
Build console output for commit: 15f7e895aafb0243cdb903eaa637e146985d704e available at: https://gist.github.com/3c2c8aa4edd311798882e19fe7b77208<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10079-15f7e89.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10079-15f7e89.dmg

Triplelexx:
I tested this a few times, I was able to run the script to completion. I also tried some other entity edits and did not encounter any problems.

libraries/entities/src/EntityItem.cpp
```diff
@@ -2327,3 +2357,443 @@ bool EntityItem::matchesJSONFilters(const QJsonObject& jsonFilters) const {
...
+        result = _collisionMask;
+    });
+    return result;
+}
+
+void EntityItem::setCollisionMask(uint8_t value) { 
+    withWriteLock([&] {
+        _collisionMask = value;
+    });
+}
```
AndrewMeadows: Why lock just to set or get a `uint8_t`?  I can understand locking when digging into a container or even when setting a vec3, but setting and getting fundamental types?

samcake:
I did a review pass  trying to understand the code in details.
I think it s a pass, i don;t see mistake just like that.

My only concern is about the performance cost of using the "withREad/WriteLock( lambda) pattern.
I would suggest we can do the same behavior without the cost of the lambda indirection with a macro.
THe macro execute the same code as the  withRead/WriteLock happening around the actual body of the lambda. 
Maybe the compiler recognize the pattern and avoiding all the behind the scene Lambda machinery but i m not really sure.

jherico:
@samcake my quick and dirty testing shows that the use of a lambda is statistically insignificant compared to the cost of locking the mutex, which itself is already pretty low (I can do it about 10 million times a second).  In other words, 

```

OmegaHeron:
This was merged and produced DEV-6292 where (Windows 10) I lost ability to load HTTP textures - for example, my avatar.  Models with embedded texture assets work as expected.

Looking at log I see Interface requesting texture via some (correct) URL like;
`[04/10 13:15:05] [DEBUG] [hifi.networking.resource] Starting request for: https://o2t.info/assets/av/tex/mega-wip-2/orig/mega-wip-4-clothed/mega-wip-4-clothed_Shoes_Normal.png`

Then;
`[04/10 13:15:06] [DEBUG] [hifi.networking] Error loading  QUrl("https://o2t.info/assets/av/tex/mega-wip-2/orig/mega-wip-4-clothed/mega-wip-4-clothed_Shoes_Normal.png")`

6292
hifi-snap-by-omegaheron-on-2017-04-10_13-16-51

6291

hifi-snap-by-omegaheron-on-2017-04-10_13-25-08

And up to latest (6296 as I write this).
hifi-snap-by-omegaheron-on-2017-04-10_13-28-07

Note result at dev-playa;

hifi-snap-by-omegaheron-on-2017-04-10_13-29-51

jherico:
@OmegaHeron can you verify that if you wipe your ktx_cache folder this still occurs?

OmegaHeron:
@jherico 

Removed user/appData/local/Interface/ktx_cache - no change.

Removed all folders under local/Interface - no change.

Used menu item to clear cache - no change.

jherico:
@OmegaHeron I feel like the issue here may be that the web content is gone, but in other builds you're seeing cached data.  I'm getting a 403 error accessing `https://o2t.info/assets/av/tex/mega-wip-2/orig/mega-wip-4-clothed/mega-wip-4-clothed_Shoes_Normal.png` from chrome, but if I switch my user-agent to mimic interface I get a 404 error.  

Also, I just loaded dev-playa in build 6920 and I'm seeing the missing textures there too, so while there seems to be a problem, I don't think it has to do with this PR.

OmegaHeron:
403 in chrome is because I disallow fetching from anything that doesn't ident as Interface so that's correct.

Checking 404 now.  I ssh'd to box and confirmed all is present, but - will check that something (i.e. updates) hasn't borked my nginx config.  Will advise.

jherico:
@OmegaHeron the problem with comparing against 6291 is that it's from the stable branch, so it doesn't have the actual problematic PR #10116 merged in yet

OmegaHeron:
@jherico - understood.  I'll drop my evil upon #10116 as I just ran against it per suggestion.  Unfortunately -- the typical trace back sequentially for dev builds to point of fail was flawed for this.  Objections to this PR withdrawn, apologies extended.

PR #10053, opened 2017-03-29
CR for Job #21245 - Created a Worklist job to show PR flow

cozza13:
Code Review for Job #21245 - Workitem available at https://www.worklist.net/21245

PR #9819, opened 2017-03-07
21204 - Turn the tetherball into a handheld toy

Triplelexx:
Added 2 scripts:

- createTetherballStick.js: Creates an equippable stick with a tethered ball
- tetherballStick.js: Entity script for an equippable stick with a tethered ball

**Test Plan**
Running createTetherballStick.js should create a stick and ball in front of the avatar.
The stick should be equippable in either hand and the ball should hang from the stick suspended by a line.
The ball should collide with things, yet ignore the user.

PR #10106, opened 2017-04-03
Added tetherball create and entity scripts

MrRoboman:
These scripts create a tetherball toy. It's stick with a ball hanging from the end of it!


hifi-gustavo:

themelissabrown:
add to whitelist

themelissabrown:
previous PR's https://github.com/highfidelity/hifi/pull/10053
Turn the tetherball into a handheld toy #9819
Work needed: 
@MrRobotman will:
-make string a bit longer and less rigid
-make toy equipable

themelissabrown:
@MrRoboman can you provide the script?

MrRoboman:
[createTetherballStick.zip](https://github.com/highfidelity/hifi/files/911702/createTetherballStick.zip)
@themelissabrown You mean like this?

MrRoboman:
Testing Notes:
Run createTetherballStick.js - Tetherball Stick and Ball should appear in front of you.
Move hand to Tetherball Stick - Blue 'Equip' sphere should appear.
Pull trigger - Tetherball Stick should equip to the correct hand with a natural looking postion and rotation. The ball should fall away, but be caught by the newly created blue 'string'. 
Move the Stick around, Fly fast, teleport - the ball should remain attached and act as if it's dangling from the stick.
Unequip by using the grip button - Stick should remain in current postion, Ball will return to the tip of the stick and the line will disappear.

OmegaHeron:
Tested Windows 10 - CV1 + Touch Controllers
QA Fail

If attached and you change to new domain then return to last location you see a "wrecked" tetherball toy.  Needs handling, gracefully, of agent having it attached and leaving domain while attached.

hifi-snap-by-omegaheron-on-2017-04-13_13-15-48

Works as described otherwise.

OmegaHeron:
Testing note - any time testing grabbed/equipped things - it should be required that test includes leaving domain where item was equipped for another domain then return to location in original domain where item was last equipped.  This will insure we don't have equipped items that leave bits around in undefined states (a problem with many equipped items).

MrRoboman:
Thanks, @OmegaHeron 
So the equipped item is left behind when leaving domain like other equippable entities like Ping Pong Gun. Is the desired fix that the stick and ball are left behind unbroken? Do we expect the toy to travel with you?

OmegaHeron:
That's a design question @MrRoboman so I can only give advice to take it up with your HiFi contact person for job.

My two cents worth as someone who's ran a domain here since nearly the beginning and seen his fair (unfair even) share of messes...  One huge problem with equipped and/or dynamic objects is that when an agent has hooked onto then leaves domain it sometimes leaves a mess...  my preference as a domain operator would be that an equipped object cleans up after an agent leaves domain, and for an object that does not equip a copy of its original versus taking the single item - it return to where it was first placed for equipping.

ImRichardJackson:
Tested Windows 10 - Touch Controllers
QA Pass

- Run createTetherballStick.js - Tetherball Stick and Ball should appear in front of you.
-Move hand to Tetherball Stick - Blue 'Equip' sphere should appear.
-Pull trigger - Tetherball Stick should equip to the correct hand with a natural looking position and rotation.
-Move the Stick around, Fly fast, teleport - the ball should remain attached and act as if it's dangling from the stick.
-Unequip by using the grip button - Stick should remain in current position, Ball will return to the tip of the stick and the line will disappear.

MrRoboman:
Fixed issue where the toy would be left behind in a broken state if the user equipping it leaves the domain by reducing its lifetime to 2 seconds.

OmegaHeron:
Re-testing now

OmegaHeron:
Tested Windows 10
QA Pass

Pass functional and cleanup tests.

themelissabrown:
I've never been so happy to change a PR to QA Approved 
Can I get a quick Dev approval?

PR #10124, opened 2017-04-04
send initial path in location update (e.g., on login)

howard-stearns:


howard-stearns:
I've tested this myself and confirmed that it sends the path and other location data when first logged in (regardless of availability), and then does not update after that if availability is set to none.

howard-stearns:
The reason for this PR is for the case where all of the following are true, in order:
1. A user sets their availability to no one, using the old released client. This deletes their location record.
2. The same user logs in with master. In master, this does not update any location info.
3. The same user then tries to shake hands with someone. This tries to update or create a location record (create, in this case) with no pathname info, which fails.  This creates an error on the server, and also doesn't let this user actually make a connection.

hifi-gustavo:
Build console output for commit: 6f79b2eaeb9c559abefe7b33560a36a0e6cf4bbd available at: https://gist.github.com/3c9cbc7f81f2681d439403c1c1e62903<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10124-6f79b2e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10124-6f79b2e.dmg

PR #10142, opened 2017-04-06
21258 - Don't render (or heavily use GPU) when window in minimized

Triplelexx:
**Test Plan:**

Run the PR in an area with content that should put a high load on your GPU. Observe the resources being used by interface. On this PR the resources should reduce when interface is minimized.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -6485,7 +6485,6 @@ void Application::windowMinimizedChanged(bool minimized) {
     }
 }
 
-
 void Application::postLambdaEvent(std::function<void()> f) {
     if (this->thread() == QThread::currentThread()) {
         f();
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1429,6 +1429,7 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer, bo
     entityPacketSender->setMyAvatar(myAvatar.get());
 
     connect(this, &Application::applicationStateChanged, this, &Application::activeChanged);
+    connect(_window, SIGNAL(windowMinimizedChanged(bool)), this, SLOT(windowMinimizedChanged(bool)));
     qCDebug(interfaceapp, "Startup time: %4.2f seconds.", (double)startupTimer.elapsed() / 1000.0);
 
     auto textureCache = DependencyManager::get<TextureCache>();
@@ -6475,6 +6476,15 @@ void Application::activeChanged(Qt::ApplicationState state) {
             break;
     }
 }
+
+void Application::windowMinimizedChanged(bool minimized) {
+    if (!minimized && !getActiveDisplayPlugin()->isActive()) {
+        getActiveDisplayPlugin()->activate();
+    } else if (minimized && getActiveDisplayPlugin()->isActive()) {
+        getActiveDisplayPlugin()->deactivate();
+    }
+}
+
 
 void Application::postLambdaEvent(std::function<void()> f) {
     if (this->thread() == QThread::currentThread()) {
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -2071,7 +2071,7 @@ void Application::initializeUi() {
 
 void Application::paintGL() {
     // Some plugins process message events, allowing paintGL to be called reentrantly.
-    if (_inPaint || _aboutToQuit) {
+    if (_inPaint || _aboutToQuit || _window->isMinimized()) {
         return;
     }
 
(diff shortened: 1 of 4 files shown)


hifi-gustavo:
Build console output for commit: 1b2887d5b1784a9ebd6af66ece5b3c4b5a315a15 available at: https://gist.github.com/138fd74e1a67a7a6ca6e74266fbaffbb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10142-1b2887d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10142-1b2887d.dmg

themelissabrown:
@Triplelexx this is for https://worklist.net/21258 correct?

PhilipRosedale:
cool ... testing

PhilipRosedale:
QA FAIL 

Strange - I use process explorer to look at both CPU and GPU and when I minimize the window, nothing happens!  Can you see if you see the GPU being not used with window minimized?

PhilipRosedale:
Wait, hang on, testing again.  May have had wrong build.

PhilipRosedale:
Nope - actually this doesn't work (yet).  @samcake says you also need to pause the displayPlugin, because there are two threads running.

samcake:
@Triplelexx  Sorry i was too fast reviewing the code
The check you did to avoid the PainGL is correct but it s not enough
When you are minimized you need to pause the true rendering happening in the "present THread" controlled by the display Plugin
I m going to bring @jherico aka Austin in here to help too

Triplelexx:
@samcake I've started having a look into the plugin's other thread now, I've not really looked at that. That explains the overhead, I put it down to other areas.

Triplelexx:
I think I've got a pretty good idea what needs to be done. I'm doing some more studying on Qt as well though to make things clearer.

hifi-gustavo:
Build console output for commit: 279997d9051420f3d2c577af9fe188cdf335d2cb available at: https://gist.github.com/138fd74e1a67a7a6ca6e74266fbaffbb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10142-279997d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10142-279997d.dmg

Triplelexx:
@samcake @jherico I have made a change to deactivate the display plugin. I am unsure if there might be any issue with the method though, I'm having some problems checking it effectively.

PhilipRosedale:
Testing at home and now this definitely drops the CPU user.  I may have seen a crash on attempting to logout with oculus active, from the minimized state.  Moving to work machine where I can test exact GPU% as well.

Triplelexx:
I think leaving the display plugin deactivated when you logout could have ramifications, if any crashing behavior is confirmed I'll look into it further.

PhilipRosedale:
QA Fail: 

CPU/GPU as now reduced as expected, tested in both desktop and HMD mode. 

However, minimize the app and then right-click and select "Close Window" from the toolbar icon, and nothing happens.   Do it a second time, and you get the dialog saying the application is not responding, and you have to force close it.  Need to fix this.

Triplelexx:
I'll have a look into it more later today

Triplelexx:
@PhilipRosedale I can't seem to repro that on Windows 10. I've got an idea to try though anyway. Where you using Windows 10?

Triplelexx:
I can repro in debug, but works fine in release build. I am also able to repro the same issue in master.

zfox23:
@Triplelexx If you can repro the issue in master, then it feels to me like this PR shouldn't be QA failed because of that issue. Thoughts, @themelissabrown?

themelissabrown:
@zfox23 I would agree that it's a QA pass and we need to figure out why what we do causes this: "However, minimize the app and then right-click and select "Close Window" from the toolbar icon, and nothing happens. Do it a second time, and you get the dialog saying the application is not responding, and you have to force close it. Need to fix this." in debug. I'll check fogbugz for a ticket.

themelissabrown:
Please QA in release build and check specifically for this: 
* Minimize the app and then right-click and select "Close Window" from the toolbar icon, and nothing happens. Do it a second time, and you get the dialog saying the application is not responding, and you have to force close it.

themelissabrown:
@zfox When all code was commented out, the same thing happened. This is potentially an issue with our debug mode. How do we get this PR cleared?

zfox23:
I'm sorry this took so long, @Triplelexx.

@themelissabrown: The crash that @PhilipRosedale is reporting should definitely be fixed, but is not caused by this PR. Could you please file a ticket against that bug?

PR #10146, opened 2017-04-06
avoid invisible avatars redux

AndrewMeadows:
* fixed bug where other avatars would not render when signing out of user account
* restored scale animation when other avatars disconnect
* removed delay of avatar scale animation when watching another avatar change size

### DIFF
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -192,6 +192,8 @@ void Avatar::animateScaleChanges(float deltaTime) {
             _isAnimatingScale = false;
         }
         setScale(glm::vec3(animatedScale)); // avatar scale is uniform
+
+        // flag the joints as having changed for force update to RenderItem
         _hasNewJointData = true;
 
         // TODO: rebuilding the shape constantly is somehwat expensive.
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -192,6 +192,7 @@ void Avatar::animateScaleChanges(float deltaTime) {
             _isAnimatingScale = false;
         }
         setScale(glm::vec3(animatedScale)); // avatar scale is uniform
+        _hasNewJointData = true;
 
         // TODO: rebuilding the shape constantly is somehwat expensive.
         // We should only rebuild after significant change.
@@ -200,8 +201,12 @@ void Avatar::animateScaleChanges(float deltaTime) {
 }
 
 void Avatar::setTargetScale(float targetScale) {
-    AvatarData::setTargetScale(targetScale);
-    _isAnimatingScale = true;
+    float newValue = glm::clamp(targetScale, MIN_AVATAR_SCALE, MAX_AVATAR_SCALE);
+    if (_targetScale != newValue) {
+        _targetScale = newValue;
+        _scaleChanged = usecTimestampNow();
+        _isAnimatingScale = true;
+    }
 }
 
 void Avatar::updateAvatarEntities() {
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -476,7 +476,7 @@ static TextRenderer3D* textRenderer(TextRendererType type) {
     return displayNameRenderer;
 }
 
-bool Avatar::addToScene(AvatarSharedPointer self, std::shared_ptr<render::Scene> scene, render::Transaction& transaction) {
+void Avatar::addToScene(AvatarSharedPointer self, std::shared_ptr<render::Scene> scene, render::Transaction& transaction) {
     auto avatarPayload = new render::Payload<AvatarData>(self);
     auto avatarPayloadPointer = Avatar::PayloadPointer(avatarPayload);
     _renderItemID = scene->allocateID();
@@ -486,9 +486,6 @@ bool Avatar::addToScene(AvatarSharedPointer self, std::shared_ptr<render::Scene>
     for (auto& attachmentModel : _attachmentModels) {
         attachmentModel->addToScene(scene, transaction);
     }
-
-    _inScene = true;
-    return true;
 }
 
 void Avatar::removeFromScene(AvatarSharedPointer self, std::shared_ptr<render::Scene> scene, render::Transaction& transaction) {
@@ -498,7 +495,6 @@ void Avatar::removeFromScene(AvatarSharedPointer self, std::shared_ptr<render::S
     for (auto& attachmentModel : _attachmentModels) {
         attachmentModel->removeFromScene(scene, transaction);
     }
-    _inScene = false;
 }
 
 void Avatar::updateRenderItem(render::Transaction& transaction) {
@@ -1435,7 +1431,7 @@ void Avatar::addToScene(AvatarSharedPointer myHandle) {
     }
 }
 void Avatar::ensureInScene(AvatarSharedPointer self) {
-    if (!_inScene) {
+    if (!render::Item::isValidID(_renderItemID)) {
         addToScene(self);
     }
 }
--- a/interface/src/avatar/AvatarManager.cpp
+++ b/interface/src/avatar/AvatarManager.cpp
@@ -396,9 +396,6 @@ void AvatarManager::clearOtherAvatars() {
     while (avatarIterator != _avatarHash.end()) {
         auto avatar = std::static_pointer_cast<Avatar>(avatarIterator.value());
         if (avatar != _myAvatar) {
-            //auto removedAvatar = avatarIterator.value();
-            //avatarIterator = _avatarHash.erase(avatarIterator);
-            //handleRemovedAvatar(removedAvatar);
             if (avatar->isInScene()) {
                 avatar->removeFromScene(avatar, scene, pendingChanges);
             }
--- a/interface/src/avatar/AvatarManager.cpp
+++ b/interface/src/avatar/AvatarManager.cpp
@@ -68,7 +68,7 @@ void AvatarManager::registerMetaTypes(QScriptEngine* engine) {
 }
 
 AvatarManager::AvatarManager(QObject* parent) :
-    _avatarFades(),
+    _avatarsToFade(),
     _myAvatar(std::make_shared<MyAvatar>(std::make_shared<Rig>()))
 {
     // register a meta type for the weak pointer we'll use for the owning avatar mixer for each avatar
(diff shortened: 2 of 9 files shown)


AndrewMeadows:
This purpose of this PR was to fix a bug with the following repro recipe:

(Bug-1) Other Avatars vanish on sign-out/sign-in:
(1) UserA: enter domain where there are other avatars
(2) UserA: toggle sign-in status: either sign-in with username+password or sign-out
(3) --> Bugged: On UserA's interface all other avatars will have vanished: they can be heard but not seen.
--> Expected: On UserA's interface all other avatars should vanish momentarily but re-appear as UserA reconnects to the domain.

That bug is fixed but there are a few other bugs that were fixed along the way:

(Bug-2) OtherAvatar's rescale animation is delayed:
(4) UserB: adjust avatar scale to either twice as big or twice as small
(5) UserA: watch UserB's avatar animate as its scale changes...
(6) --> Bugged: UserA will see UserB move to the correct location as if they had resized, and then a few seconds later their avatar will grow or shrink to the new size.
--> Expected: UserA will see UserB move and rescale simultaneously.


(Bug-3) OtherAvatars do not shrink small on disconnect.  (Note: this may be poor/confusing behavior, but this is how the code was designed to work and this behavior is broken)
(7) UserA: watch UserB as they logout
(8) --> Bugged: UserB vanishes immediately instead of shrinking small
--> Expected: UserB will first shrink to small size and then disappear


(Bug-4) `Scene.shouldRenderAvatars` = false should make all avatars disappear from the scene
(9) Open JS console: CTRL + ALT + J
(10) Type the following JS code:
Scene.shouldRenderAvatars = false
(11) --> Bugged: the avatars may blink off but they remain visible.  The MyAvatar instance will momentarily go into T-pose.
--> Expected: all avatars, including the MyAvatar instance should no longer be rendered.  However, they should still be in the physics engine so dynamic boxes will bounce off of the avatars' capsule shape.


To test:

First reproduce Bug-1 through Bug-4 on an older release to verify that the bugs really exist.  Then try them again on this PR to verify fixed.  Note: Bug-1, -3, and -4 should be fixed in this PR no matter the avatar-mixer vintage, as long as it's protocol is compatible, however Bug-2 will only prove fixed when using an avatar-mixer from this PR.

Also make sure:

(12) Avatar animation still works fine for UserA and UserB who watches.
(13) Changing avatar models still updates correctly, for UserA and UserB who watches.
(14) Avatar collision shape still works when avatar changes scale.
(15) Other avatar's collision shape is not left behind when other avatar disconnects.

interface/src/avatar/AvatarManager.cpp
```diff
@@ -353,26 +382,36 @@ void AvatarManager::handleRemovedAvatar(const AvatarSharedPointer& removedAvatar
...
+            AvatarMotionState* motionState = avatar->getMotionState();
+            if (motionState) {
+                _motionStatesThatMightUpdate.remove(motionState);
+                _motionStatesToAddToPhysics.remove(motionState);
+                _motionStatesToRemoveFromPhysics.push_back(motionState);
+            }
         }
+        ++avatarIterator;
     }
+    scene->enqueuePendingChanges(pendingChanges);
```
AndrewMeadows: This is the first half of the fix for Bug-1: just remove avatars from the world rather than delete them outright.

The second half of the fix was to properly reinsert them into the world if the interface receives an update for them from the AvatarMixer.  Look for other review comments where I point this code out.

Basically after calling `clearOtherAvatars()` the other Avatar instances will sit invisibly on the main list until they timeout.  This means that some scripts might be able to detect them before they timeout.  Also, they may persist when hopping from one domain to another... but will eventually timeout.  Hrm... I should add something about this to the test plan.

hifi-gustavo:
Build console output for commit: 8c5c64bc2e52e64250fd0b31132c9520cdbd64a2 available at: https://gist.github.com/b2fd6bce230196f2bfdc6a65d887c4c0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10146-8c5c64b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10146-8c5c64b.dmg

interface/src/avatar/AvatarManager.cpp
```diff
@@ -277,54 +290,70 @@ void AvatarManager::postUpdate(float deltaTime) {
...
+            } else if (_shouldRender) {
+                // very rare transition so we process pending changes immediately
+                render::ScenePointer scene = qApp->getMain3DScene();
+                render::PendingChanges pendingChanges;
+                avatar->addToScene(avatar, scene, pendingChanges);
+                scene->enqueuePendingChanges(pendingChanges);
+            }
+        }
     }
 }
```
AndrewMeadows: This is the second half of the fix for Bug-1: `processAvatarDataPacket()` has been made virtual in the `AvatarHashMap` base class and overridden by `AvatarManager` to properly insert or remove the avatar from the scene if its in-world state disagrees with its expected state.

`parseAvatarData()` has been added and returns a pointer to `AvatarData` instance that was added/updated, or a null pointer if the update data was bad.

libraries/avatars/src/AvatarData.cpp
```diff
@@ -123,6 +123,7 @@ void AvatarData::setTargetScale(float targetScale) {
     if (_targetScale != newValue) {
         _targetScale = newValue;
         _scaleChanged = usecTimestampNow();
+        _avatarScaleChanged = _scaleChanged;
```
AndrewMeadows: This is the fix for Bug-2: (the delay of an avatar rescale).  Since we were not setting this bit the new scale would only arrive when the AvatarMixer decided to send a "full" update rather than a partial one.

interface/src/avatar/AvatarManager.cpp
```diff
@@ -259,10 +258,24 @@ void AvatarManager::updateOtherAvatars(float deltaTime) {
...
+            while (itr != _avatarsToFade.end() && usecTimestampNow() > updateExpiry) {
+                auto avatar = std::static_pointer_cast<Avatar>(*itr);
+                avatar->animateScaleChanges(deltaTime);
+                avatar->simulate(deltaTime, true);
+                avatar->updateRenderItem(pendingChanges);
+                ++itr;
+            }
+        }
+        qApp->getMain3DScene()->enqueuePendingChanges(pendingChanges);
+    }
```
AndrewMeadows: Bug-2 was introduced when the rendering of other avatars was optimized by sort order and then time-boxed to just 2msec.  This change adds back the work required to update update the render items for the shrink animation.

interface/src/avatar/AvatarManager.cpp
```diff
@@ -475,26 +514,23 @@ void AvatarManager::handleCollisionEvents(const CollisionEvents& collisionEvents
 }
 
 void AvatarManager::updateAvatarRenderStatus(bool shouldRenderAvatars) {
-    if (DependencyManager::get<SceneScriptingInterface>()->shouldRenderAvatars()) {
+    _shouldRender = shouldRenderAvatars;
```
AndrewMeadows: To fix Bug-4 we cache the value of `shouldRenderAvatars` whenever it changes (note, we're actually using the argument here rather than redundantly grabbing the value from the `SceneScriptingInterface`)  and then use it to decide to add/remove/update Avatar render items.

interface/src/avatar/Avatar.cpp
```diff
@@ -193,15 +193,22 @@ void Avatar::animateScaleChanges(float deltaTime) {
...
 
 void Avatar::setTargetScale(float targetScale) {
-    AvatarData::setTargetScale(targetScale);
-    _isAnimatingScale = true;
+    float newValue = glm::clamp(targetScale, MIN_AVATAR_SCALE, MAX_AVATAR_SCALE);
+    if (_targetScale != newValue) {
+        _targetScale = newValue;
+        _scaleChanged = usecTimestampNow();
+        _isAnimatingScale = true;
+    }
```
AndrewMeadows: Oh yeah, there was also a hidden Bug-5 where each joint update packet would set the scale of the avatar and since we weren't checking for new values the `_isAnimatingScale` would be incorrectly set 'true' even when the scale hadn't actually changed.  This would cause the a bunch of unnecessary work in the "updateOtherAvatars" context on the main thread.  This fix will probably have a measurable (via trace dump) improvement to the time it takes to process the 100-avatar test case.

hifi-gustavo:
Build console output for commit: d73a0983ee3d0f71ded86b5364d668f3a0e6243d available at: https://gist.github.com/9c5759b8cb5022c2353ce32d2d5fbed1<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10146-d73a098.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10146-d73a098.dmg

AndrewMeadows:
After testing on the 100-avatar domain I think I introduced a crash bug on shutdown.  Will fix.

hifi-gustavo:
Build console output for commit: efb8270fa6b629e8ccc17cdf2f7750c75633374f available at: https://gist.github.com/b2fd6bce230196f2bfdc6a65d887c4c0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10146-efb8270.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10146-efb8270.dmg

AndrewMeadows:
Ok this is ready to test.

I ran some traces and found that this PR shaves some time out of the OtherAvatars context:

Context                  Master                     PR-10146
OtherAvatars            1.287                      1.013   msec
sort                            0.214                      0.125  msec
simulate                    0.014                      0.016  msec

hifi-gustavo:
Build console output for commit: 92495d9622e9a9f787a130f05d29a3b471978c98 available at: https://gist.github.com/b2fd6bce230196f2bfdc6a65d887c4c0<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10146-92495d9.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10146-92495d9.dmg

OmegaHeron:
Working on this - may take some time to complete...

OmegaHeron:
Windows 10;
QA Pass

Confirmed bugs with master.

Bug 1 - Used welcome as test bed.  14 Agents present and logging out (I was signed in when I got there) resulted in disconnect/connect and all agents reappeared.  Pass

Bug 2 - Used 2nd machine on my LAN.  Pass

Bug 3 - Used 2nd machine on my LAN (and saw one leave from Welcome).  Pass

Bug 4 - Used 2nd machine on my LAN and welcome.  Pass

Steps 12...15 Pass

PR #10157, opened 2017-04-07
Fixing the rendering bug triggered with Nvidia driver 381.65

samcake:
THis PR fixes the rendering bug that showed with a new nv driver (381.65).

THe bug is with the final evaluation of the occlusion value used to attenuate the ambient lighting.
This evaluation combine 3 potential sources of attenuation and keeps the min of it:
ambient provided by the material of a model,
ssao (if enabled)
and "curvature ambient" which is computed based on the curvature buffers

This last one seems to give bad results in the general case except for a very small portion of the screen.
I do not understand why this causes the bug that we see since when i debug the values individually (through the debug script) i don't see any problem.
This could be an actual bug from the driver when compiling the shader glsl code but i really don't know.
The very odd thing is that the exact same code works fine when shadow is enabled and it has nothing to do with that...

I fixed this problem by removing the "curvature ambient" contribution to the occlusion and only keep it for scattering pixels. Noit a big change visually but it fixes the problem. we can move on and i ll try to dig it when i have more time (ah ah)...

During that investigation i changed a few things in the render buffer debugger to better understand the issue and i kept it in this pr.

## TEST PLAN
This bug is specific to NVIDIA GPU using driver 381.65 on windows (i don't know about linux)
In master in welcome or anywhere with a ambient map /skymap there should be a big shadow accross the screen which is view dependent.

THe problem is gone with this pr hoppefully...

Optionally you could test the script "developer/utilities/render/debugDeferredLighting.js"
There is now the shadow checkbox to turn it on/off
Make sure that the combo box "Debug Framebuffer" shows the new item "Curvature Occlusion"

And voila

### DIFF
--- a/libraries/gl/src/gl/Config.h
+++ b/libraries/gl/src/gl/Config.h
@@ -29,7 +29,7 @@
 #include <GL/wglew.h>
 
 // Uncomment this define and recompile to be able to avoid code path preventing to be able to run nsight graphics debug
-//#define HIFI_ENABLE_NSIGHT_DEBUG 1
+#define HIFI_ENABLE_NSIGHT_DEBUG 1
 
 #endif
 
--- a/libraries/gl/src/gl/Config.h
+++ b/libraries/gl/src/gl/Config.h
@@ -29,7 +29,7 @@
 #include <GL/wglew.h>
 
 // Uncomment this define and recompile to be able to avoid code path preventing to be able to run nsight graphics debug
-#define HIFI_ENABLE_NSIGHT_DEBUG 1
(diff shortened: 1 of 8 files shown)


hifi-gustavo:
Build console output for commit: b1a4bc329be3ad754e4e079620e4f6d30ef0419e available at: https://gist.github.com/ae63a28e8a8f6eae9e5ee915b807de86<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10157-b1a4bc3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10157-b1a4bc3.dmg

PR #10159, opened 2017-04-07
Fixing the rendering bug triggered with Nvidia driver 381.65

samcake:
THis PR fixes the rendering bug that showed with a new nv driver (381.65).

THe bug is with the final evaluation of the occlusion value used to attenuate the ambient lighting.
This evaluation combine 3 potential sources of attenuation and keeps the min of it:
ambient provided by the material of a model,
ssao (if enabled)
and "curvature ambient" which is computed based on the curvature buffers

This last one seems to give bad results in the general case except for a very small portion of the screen.
I do not understand why this causes the bug that we see since when i debug the values individually (through the debug script) i don't see any problem.
This could be an actual bug from the driver when compiling the shader glsl code but i really don't know.
The very odd thing is that the exact same code works fine when shadow is enabled and it has nothing to do with that...

I fixed this problem by removing the "curvature ambient" contribution to the occlusion and only keep it for scattering pixels. Noit a big change visually but it fixes the problem. we can move on and i ll try to dig it when i have more time (ah ah)...

During that investigation i changed a few things in the render buffer debugger to better understand the issue and i kept it in this pr.

TEST PLAN

This bug is specific to NVIDIA GPU using driver 381.65 on windows (i don't know about linux)
In master in welcome or anywhere with a ambient map /skymap there should be a big shadow accross the screen which is view dependent.

THe problem is gone with this pr hoppefully...

Optionally you could test the script "developer/utilities/render/debugDeferredLighting.js"
There is now the shadow checkbox to turn it on/off
Make sure that the combo box "Debug Framebuffer" shows the new item "Curvature Occlusion"

And voila

### DIFF
--- a/libraries/render-utils/src/LightAmbient.slh
+++ b/libraries/render-utils/src/LightAmbient.slh
@@ -62,7 +62,7 @@ vec3 evalAmbientSpecularIrradiance(LightAmbient ambient, vec3 fragEyeDir, vec3 f
 
 <@if supportScattering@>
 float curvatureAO(in float k) {
-    return 1.0f - (0.0022f * k * k) + (0.0776f * k) + 0.7369;
+    return 1.0f - (0.0022f * k * k) + (0.0776f * k) + 0.7369f;
 }
 <@endif@>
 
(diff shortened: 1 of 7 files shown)


samcake:
test this please

hifi-gustavo:
Build console output for commit: f82310545cdafb14aba5519ff272c88b7d3f1138 available at: https://gist.github.com/c690e64ab8fdd5e6ff5022547448e8d6<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10159-f823105.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10159-f823105.dmg

StarmanF:
testing

StarmanF:
**Desktop PC - Windows 10 Education - CPU 4.0 GHz Intel i7 - GTX 1080 - RAM 16 GB - Vive**

Tested: Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10159-f823105.exe

**Pass**

Notes:

- Before testing this I had already updated my Nvidia drivers to 381.65.

- To begin I opened up the master welcome and confirmed that I was affected by the bug.

- Then I opened up this PR and the bug was gone.

Optional:

- Tried the shadow check box which put all the shadows on for the various in world objects, it was off by default, but didn't cause the bug to reappear. 

- Checked that the Curvature occlusion was present, turning it on made half the screen display white on any surfaces and grey on the skybox. Not sure what I was supposed to look out for there.

libraries/render-utils/src/LightAmbient.slh
```diff
@@ -62,7 +62,7 @@ vec3 evalAmbientSpecularIrradiance(LightAmbient ambient, vec3 fragEyeDir, vec3 f
 
 <@if supportScattering@>
 float curvatureAO(in float k) {
-    return 1.0f - (0.0022f * k * k) + (0.0776f * k) + 0.7369;
+    return 1.0f - (0.0022f * k * k) + (0.0776f * k) + 0.7369f;
```
AndrewMeadows: The shader code doesn't need the 'f' at the end of the literal, does it?  It must ignore that syntax because I'm seeing it used elsewhere and if it were a syntax error I'd expect lots of stuff to not work.

samcake: right glsl consider real as float by default (unlike c++)
but now i try to fix these when i see them

PR #10172, opened 2017-04-08
fix multi-line sessionDisplayName

howard-stearns:
fixes https://highfidelity.fogbugz.com/f/cases/4052/PAL-shouldn-t-accept-multi-line-display-names

### DIFF
--- a/assignment-client/src/avatars/AvatarMixer.cpp
+++ b/assignment-client/src/avatars/AvatarMixer.cpp
@@ -200,7 +200,7 @@ void AvatarMixer::manageDisplayName(const SharedNodePointer& node) {
         QString baseName = avatar.getDisplayName().trimmed();
         const QRegularExpression curses { "fuck|shit|damn|cock|cunt" }; // POC. We may eventually want something much more elaborate (subscription?).
         baseName = baseName.replace(curses, "*"); // Replace rather than remove, so that people have a clue that the person's a jerk.
-        const QRegularExpression trailingDigits { "\\s*_\\d+$" }; // whitespace "_123"
+        const QRegularExpression trailingDigits { "\\s*_\\d+(\\n[^$]*)?$" }; // trailing whitespace "_123" and any subsequent lines
         baseName = baseName.remove(trailingDigits);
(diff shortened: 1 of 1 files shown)


howard-stearns:
The fix is taken as part of stripping the user-supplied displayName, before the result is compared against other names for uniqueness. However, it does not alter the semantics or use of displayName itself.

Test plan: Try setting MyAvatar.displayName = "foo pr10172 bar_123\nred white   \nblue"
In open (or re-open) the PAL. The session display name should be one line, and not end in _n+ (unless there are other "foo pr10172 bar" present!)

hifi-gustavo:
Build console output for commit: c32879d807d573fa78a2ac23ccc7081b37fdf588 available at: https://gist.github.com/11e8d9c970b51bdb8286c0c6d4ba4626<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10172-c32879d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10172-c32879d.dmg

davidkelly:
This works perfectly for a display name with <whatever>_XXX\nYYY.  However, what about something like foo\nbar?  That will still show on multiple lines.  I think what we need is to strip new lines out, _then_ strip the _XXX off the end of that, to solve this issue in general.  Calling it QA Fail, but it is just another edge case.  It would be great to formulate one RegEx that can do this, but it can be easily done with 2...

howard-stearns:
doh! I'll fix.

hifi-gustavo:
Build console output for commit: 54888f85bec049e0ffb35439af39de7cae42fd83 available at: https://gist.github.com/11e8d9c970b51bdb8286c0c6d4ba4626<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10172-54888f8.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10172-54888f8.dmg

davidkelly:
Everything I tried seems to have worked 👍

PR #10178, opened 2017-04-10
Fix compound collision hulls

sethalves:
- fix bug that was causing compound collision shapes to be the convex union of their parts

### DIFF
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -14,6 +14,7 @@
 
 #include <QJsonDocument>
 #include <QtCore/QThread>
+#include <QUrlQuery>
 #include <glm/gtx/transform.hpp>
 
 #include <AbstractViewStateInterface.h>
@@ -621,8 +622,11 @@ void RenderableModelEntityItem::setShapeType(ShapeType type) {
 
 void RenderableModelEntityItem::setCompoundShapeURL(const QString& url) {
     auto currentCompoundShapeURL = getCompoundShapeURL();
-    QString hullURL = url + "?collision-hull";
-    ModelEntityItem::setCompoundShapeURL(hullURL);
+    QUrl hullURL(url);
+    QUrlQuery queryArgs(hullURL);
+    queryArgs.addQueryItem("collision-hull", "");
+    hullURL.setQuery(queryArgs);
+    ModelEntityItem::setCompoundShapeURL(hullURL.toString());
 
     if (getCompoundShapeURL() != currentCompoundShapeURL || !_model) {
         EntityTreePointer tree = getTree();
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -621,6 +621,10 @@ void RenderableModelEntityItem::setShapeType(ShapeType type) {
 }
 
 void RenderableModelEntityItem::setCompoundShapeURL(const QString& url) {
+    // because the caching system only allows one Geometry per url, and because this url might also be used
+    // as a visual model, we need to change this url in some way.  We add a "collision-hull" query-arg so it
+    // will end up in a different hash-key in ResourceCache.  TODO: It would be better to use the same URL and
+    // parse it twice.
     auto currentCompoundShapeURL = getCompoundShapeURL();
     QUrl hullURL(url);
     QUrlQuery queryArgs(hullURL);
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -611,7 +611,7 @@ void RenderableModelEntityItem::setShapeType(ShapeType type) {
     ModelEntityItem::setShapeType(type);
     if (getShapeType() == SHAPE_TYPE_COMPOUND) {
         if (!_compoundShapeResource && !getCompoundShapeURL().isEmpty()) {
-            _compoundShapeResource = DependencyManager::get<ModelCache>()->getGeometryResource(getCompoundShapeURL());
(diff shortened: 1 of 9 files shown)


sethalves:
https://highfidelity.fogbugz.com/f/cases/4031/compound-collisions-hulls-aren-t-working-right

sethalves:
to test:
- rez http://headache.hungry.com/~seth/hifi/well/brick-well.obj and stand it up
- press ctrl-J to open the running-scripts dialog
- type "dice" into the filter, and run createDice.js
- throw dice against the ring and through it

hifi-gustavo:
Build console output for commit: 554b23357c4c76b3c1402eb3afe04114cc60bbab available at: https://gist.github.com/0fca5d755c00228f9b86ea0478c9f7e2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10178-554b233.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10178-554b233.dmg

hifi-gustavo:
Build console output for commit: 717436e0fded47ab9e0172393e8aa5e5dc5a3c1e available at: https://gist.github.com/0fca5d755c00228f9b86ea0478c9f7e2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10178-717436e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10178-717436e.dmg

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -621,15 +622,19 @@ void RenderableModelEntityItem::setShapeType(ShapeType type) {
 
 void RenderableModelEntityItem::setCompoundShapeURL(const QString& url) {
     auto currentCompoundShapeURL = getCompoundShapeURL();
-    ModelEntityItem::setCompoundShapeURL(url);
+    QUrl hullURL(url);
+    QUrlQuery queryArgs(hullURL);
+    queryArgs.addQueryItem("collision-hull", "");
+    hullURL.setQuery(queryArgs);
+    ModelEntityItem::setCompoundShapeURL(hullURL.toString());
```
zzmp: Why is this being done?

sethalves: I added a comment:  ```    // because the caching system only allows one Geometry per url, and because this url might also be used
    // as a visual model, we need to change this url in some way.  We add a "collision-hull" query-arg so it
    // will end up in a different hash-key in ResourceCache.  TODO: It would be better to use the same URL and
    // parse it twice.
```

hifi-gustavo:
Build console output for commit: 6bd841857f6a00d6decc924493dc0939624d7eb7 available at: https://gist.github.com/0fca5d755c00228f9b86ea0478c9f7e2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10178-6bd8418.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10178-6bd8418.dmg

hifi-gustavo:
Build console output for commit: 8a3a52cbcf3b922a26944df6da87bb0dedb23b53 available at: https://gist.github.com/0fca5d755c00228f9b86ea0478c9f7e2<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10178-8a3a52c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10178-8a3a52c.dmg

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
zzmp: Could this be attacked? Do we care? What harm would it cause?

sethalves: A web server could deliver something different for the collision hull url.  Hm... I wonder if this works with atp...

OmegaHeron:
@sethalves - could you be a little more specific about how this should differ from master?  Want to make sure I'm not missing something here as it behaves precisely same in master vs PR.

Saw same behavior in master as PR - if dice hit ring they hit ring - if thrown through center they go through and land behind it.

I also set interface to draw collision shapes for a comparison.  In 6291 I see all the pieces while in this PR model disappears when draw collision shape = on.

I did this in master (6291)
Rezzed model, shape = compound shape model = same as asset model.
Got dice, threw them at ring, through ring against side of ring.
Repeated with this PR.

sethalves:
On master, the dice don't pass through the ring -- they bounce out as if it were a solid disk.

OmegaHeron:
Retesting master, that's not what I got.

sethalves:
make sure your collision type is compound, not exact

OmegaHeron:
With 6291 on Windows 10 I can walk through it, throw dice through it....  note params in screenshot, compund = on with compound model same as asset model.

hifi-snap-by-omegaheron-on-2017-04-10_18-16-27

hifi-snap-by-omegaheron-on-2017-04-10_18-17-13

sethalves:
Ah, I see.  The but was introduced after the last RC -- you'd have to get one from https://highfidelity.io/dev-download to see the bug.

OmegaHeron:
Sorry about that - @sethalves - then maybe QA Pass as your PR acts same intended with one oddity.  I can not see a collision shape in your PR if I ask to see one.  If that's expected then pass.

sethalves:
Yeah we broke that too.  This PR doesn't address the problem with showing the collision shape.  Thanks for the test

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
sethalves: It works in atp://.

PR #10181, opened 2017-04-10
re-enable the info button for user images on nearby tab

davidkelly:
So, hovering over a user's image in nearby tab in the PAL now will bring up an 'info' button, which you can click on.


davidkelly:
Test Plan

Bring up PAL
hover over a user image in the nearby tab
you should see the info button
click on it, you go to their user card 'card'
now go to connections
hover over user image
no info button should come up since there is a button for this below

davidkelly:
fixes https://highfidelity.fogbugz.com/f/cases/edit/4064/lost-info-on-nearby-people-in-pal

PR #10206, opened 2017-04-12
Cleanup dependencies in Avatar class in preparation for migration to new lib

AndrewMeadows:
* remove dependencies and cruft from Avatar implementation


AndrewMeadows:
We're getting ready to move the render parts of Avatar class out of interface and into a library.

There are zero expected behavior changes here.

To test:

(1) General smoke test.  In particular: all avatars should render correctly.
(2) The `Camera.cameraEntity` feature in JS should still work (I think it makes the camera follow a moving entity).  It is used in edit.js.

hifi-gustavo:
Build console output for commit: d7e4dd960993a43aadc2b523f5c9380c4ee877ce available at: https://gist.github.com/321ecddb78053c9a35b47e299e02d66f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10206-d7e4dd9.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10206-d7e4dd9.dmg

ZappoMan:
@AndrewMeadows - even though there are no expected behavior changes, we need to right a test plan, and include it in the test plan library for all the "expected" behaviors that you can think of that might "read on" this code. Please flesh out a test plan of the things that "have to still work" after making this change - and put it in the Test plan library.

hifi-gustavo:
Build console output for commit: f6727470092999a63340390a5cf467510148d965 available at: https://gist.github.com/321ecddb78053c9a35b47e299e02d66f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10206-f672747.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10206-f672747.dmg

interface/src/Camera.h
```diff
@@ -143,7 +133,6 @@ public slots:
     glm::quat _orientation;
     bool _isKeepLookingAt{ false };
     glm::vec3 _lookingAt;
-    EntityItemPointer _cameraEntity;
```
AndrewMeadows: This stuff moved to `FancyCamera`.

hifi-gustavo:
Build console output for commit: 64e8ef1d3a8937365c84e6a6d0fc40025e164a9d available at: https://gist.github.com/321ecddb78053c9a35b47e299e02d66f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10206-64e8ef1.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10206-64e8ef1.dmg

interface/src/avatar/Avatar.cpp
```diff
@@ -543,11 +532,11 @@ void Avatar::postUpdate(float deltaTime) {
     }
 }
 
-void Avatar::render(RenderArgs* renderArgs, const glm::vec3& cameraPosition) {
+void Avatar::render(RenderArgs* renderArgs) {
     auto& batch = *renderArgs->_batch;
     PROFILE_RANGE_BATCH(batch, __FUNCTION__);
 
-    if (glm::distance(DependencyManager::get<AvatarManager>()->getMyAvatar()->getPosition(), getPosition()) < 10.0f) {
+    if (glm::distance(DependencyManager::get<AvatarManager>()->getMyAvatarPosition(), getPosition()) < 10.0f) {
```
samcake: we could probably rely on the ViewFrustum pos  instead of  "myAvatarPosition"
And also, i don't know if the code triggered here make any sense in the genreal context of and Avatar ?
THis is drawing laser pointers if the hands are in the "FIngerPOintingFlag"
But that doesn't seem like a generic behavior ?

interface/src/avatar/Avatar.cpp
```diff
@@ -606,23 +595,16 @@ void Avatar::render(RenderArgs* renderArgs, const glm::vec3& cameraPosition) {
...
-            qApp->copyShadowViewFrustum(frustum);
-        } else {
-            qApp->copyDisplayViewFrustum(frustum);
-        }
-        if (!frustum.sphereIntersectsFrustum(getPosition(), getBoundingRadius())) {
-            return;
-        }
+    ViewFrustum frustum = renderArgs->getViewFrustum();
+    if (!frustum.sphereIntersectsFrustum(getPosition(), getBoundingRadius())) {
+        return;
```
samcake: Yep!

interface/src/avatar/AvatarManager.cpp
```diff
@@ -105,7 +105,7 @@ void AvatarManager::init() {
             this, &AvatarManager::updateAvatarRenderStatus, Qt::QueuedConnection);
 
     if (_shouldRender) {
-        render::ScenePointer scene = qApp->getMain3DScene();
+        const render::ScenePointer& scene = qApp->getMain3DScene();
```
samcake: I think here you actually want to keep the render::ScenePointer version 
not the const render::ScenePointer&
You want to actually create a new local shared ptr on the same Scene obtained from the getMain3DScene

interface/src/avatar/AvatarManager.cpp
```diff
@@ -306,12 +306,10 @@ void AvatarManager::simulateAvatarFades(float deltaTime) {
         if (avatar->getTargetScale() <= MIN_FADE_SCALE) {
             // fading to zero is such a rare event we push unique transaction for each one
             if (avatar->isInScene()) {
-                render::ScenePointer scene = qApp->getMain3DScene();
+                const render::ScenePointer& scene = qApp->getMain3DScene();
```
samcake: Same problem as above, this needs to be a local instance of the shared ptr, not a reference

interface/src/avatar/AvatarManager.cpp
```diff
@@ -343,21 +341,17 @@ void AvatarManager::processAvatarDataPacket(QSharedPointer<ReceivedMessage> mess
...
-                        render::Transaction transaction;
-                        avatar->removeFromScene(avatar, scene, transaction);
-                        scene->enqueueTransaction(transaction);
-                    }
-                }
-            } else if (_shouldRender) {
-                // very rare transition so we process the transaction immediately
-                render::ScenePointer scene = qApp->getMain3DScene();
-                if (scene) {
+                    const render::ScenePointer& scene = qApp->getMain3DScene();
```
samcake: Same as above, not a const ref but a local shared_ptr

interface/src/avatar/AvatarManager.cpp
```diff
@@ -343,21 +341,17 @@ void AvatarManager::processAvatarDataPacket(QSharedPointer<ReceivedMessage> mess
...
                     render::Transaction transaction;
-                    avatar->addToScene(avatar, scene, transaction);
+                    avatar->removeFromScene(avatar, scene, transaction);
                     scene->enqueueTransaction(transaction);
                 }
+            } else if (_shouldRender) {
+                // very rare transition so we process the transaction immediately
+                const render::ScenePointer& scene = qApp->getMain3DScene();
```
samcake: And again

interface/src/avatar/AvatarManager.cpp
```diff
@@ -394,7 +388,7 @@ void AvatarManager::handleRemovedAvatar(const AvatarSharedPointer& removedAvatar
 void AvatarManager::clearOtherAvatars() {
     // Remove other avatars from the world but don't actually remove them from _avatarHash
     // each will either be removed on timeout or will re-added to the world on receipt of update.
-    render::ScenePointer scene = qApp->getMain3DScene();
+    const render::ScenePointer& scene = qApp->getMain3DScene();
```
samcake: And again

interface/src/avatar/AvatarManager.cpp
```diff
@@ -522,7 +514,7 @@ void AvatarManager::handleCollisionEvents(const CollisionEvents& collisionEvents
 
 void AvatarManager::updateAvatarRenderStatus(bool shouldRenderAvatars) {
     _shouldRender = shouldRenderAvatars;
-    render::ScenePointer scene = qApp->getMain3DScene();
+    const render::ScenePointer& scene = qApp->getMain3DScene();
```
samcake: And again

interface/src/scripting/HMDScriptingInterface.cpp
```diff
@@ -105,9 +105,9 @@ QScriptValue HMDScriptingInterface::getHUDLookAtPosition3D(QScriptContext* conte
 }
 
 bool HMDScriptingInterface::getHUDLookAtPosition3D(glm::vec3& result) const {
-    Camera* camera = qApp->getCamera();
-    glm::vec3 position = camera->getPosition();
-    glm::quat orientation = camera->getOrientation();
+    const Camera& camera = qApp->getCamera();
```
samcake: You want a copy of the Shared ptr here, not a reference

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -437,7 +437,7 @@ void RenderableModelEntityItem::render(RenderArgs* args) {
         _model->renderDebugMeshBoxes(batch);
 #endif
 
-        render::ScenePointer scene = AbstractViewStateInterface::instance()->getMain3DScene();
+        const render::ScenePointer& scene = AbstractViewStateInterface::instance()->getMain3DScene();
```
samcake: Same as above, you need a copy, not a reference of the smart_ptr

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -526,7 +526,7 @@ ModelPointer RenderableModelEntityItem::getModel(QSharedPointer<EntityTreeRender
     // If we have no URL, then we can delete any model we do have...
     } else if (_model) {
         // remove from scene
-        render::ScenePointer scene = AbstractViewStateInterface::instance()->getMain3DScene();
+        const render::ScenePointer& scene = AbstractViewStateInterface::instance()->getMain3DScene();
```
samcake: same as above, copy, not ref

libraries/render-utils/src/Model.cpp
```diff
@@ -795,7 +793,7 @@ void Model::setURL(const QUrl& url) {
 
     {
         render::Transaction transaction;
-        render::ScenePointer scene = AbstractViewStateInterface::instance()->getMain3DScene();
+        const render::ScenePointer& scene = AbstractViewStateInterface::instance()->getMain3DScene();
```
samcake: nope, need to be a copy not a const ref

interface/src/avatar/AvatarManager.cpp
AndrewMeadows: @samcake, I was going to go through and make these changes however after looking at the code and doing some research on best-practices when dealing with `shared_ptr`... I must disagree.

The rule of thumb is to only copy a `shared_ptr` when you really want to have shared ownership of the pointer.  That is, the other context will hang onto it for a long time and might really end up being the one to own the final reference.

The cases in `AvatarManager` don't satisfy that condition.  The 3D scene of the `Application` is very stable and will outlive any of the temporary contexts in which the `AvatarManager` uses it.

An alternative path would be: give `AvatarManager` its own long-term `shared_ptr` copy as a regular data member.   We could then avoid most of the function calls for getting references or copies.

interface/src/avatar/Avatar.cpp
AndrewMeadows: The logic, as I understand it, is to draw any pointing lasers of any "nearby" avatars.  I agree that it should use origin of the frustum rather than the position of MyAvatar.

interface/src/avatar/AvatarManager.cpp
samcake: OK, you cinvinced me, i ll remove my comments

hifi-gustavo:
Build console output for commit: b8e1cad6fc79a19f05d97bdf10a6019248781e2f available at: https://gist.github.com/321ecddb78053c9a35b47e299e02d66f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10206-b8e1cad.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10206-b8e1cad.dmg

interface/src/avatar/Avatar.cpp
```diff
@@ -1437,8 +1419,7 @@ QList<QVariant> Avatar::getSkeleton() {
     return QList<QVariant>();
 }
 
-void Avatar::addToScene(AvatarSharedPointer myHandle) {
-    render::ScenePointer scene = qApp->getMain3DScene();
+void Avatar::addToScene(AvatarSharedPointer myHandle, const render::ScenePointer& scene) {
```
jherico: `const AvatarSharedPointer&`

interface/src/avatar/Avatar.cpp
```diff
@@ -1452,8 +1433,9 @@ void Avatar::addToScene(AvatarSharedPointer myHandle) {
         qCWarning(interfaceapp) << "AvatarManager::addAvatar() : Unexpected null scene, possibly during application shutdown";
     }
 }
-void Avatar::ensureInScene(AvatarSharedPointer self) {
+
+void Avatar::ensureInScene(AvatarSharedPointer self, const render::ScenePointer& scene) {
```
jherico: `const AvatarSharedPointer&`

interface/src/avatar/Avatar.cpp
```diff
@@ -85,7 +74,7 @@ namespace render {
     }
 }
 
-Avatar::Avatar(RigPointer rig) :
+Avatar::Avatar(QThread* thread, RigPointer rig) :
```
jherico: technically you don't need to pass the thread here.  Any code that has visibility on `QCoreApplication` can call `qApp->thread()`.  Not a big deal though.

AndrewMeadows: The purpose of this change is to remove dependency of `Avatar` on `qApp`.

interface/src/avatar/Avatar.cpp
AndrewMeadows: I'll fix this and the other case in a second PR.

Nex-Pro:
** This may be a FAIL please see comparison images.**

Using the latest dev-download...
(1) Login to "dev-playa" domain and walk around to get a feel for what the content should look like.
(2) Login to "dev-welcome" domain and look around to get a picture of what objects are where and how they should render

Using this PR...
(3) Login to "dev-playa" domain and verify stuff shows up correctly: FAIL see screenshots
(3a) Your avatar still renders in 1st and 3rd person, also in mirror mode (MyAvatar) SUCCESS
(3b) The skybox: See only default dark sly same as in master.
(3c) The signs and debris (models) See images.
(3d) The animated robot on top of the burger sign near <-37,-8,-16> (animated models): Success

(4) Login to "dev-welcome" domain and verify stuff still shows up: Missing primitives subject to angle of view see screenshots. This did not repro after restarting interface.

(4a) Other avatars (avartars) success : SUCCESS
(4b) Lights in hangars (lights): SUCCESS
(4c) Fire flames near <-13,0.2.0,-180> (particle systems) SUCCESS
(4d) UI and tablet: SUCCESS

(5) Create a polyvox object using this code in the script console (CTRL + ALT + j): SUCCESS

**POLYVOX**
polyvox

**Duct model in this PR**
this_duct

**Duct model in master**
master_duct

**Austin's sign in this PR**
this_austins

**Austins Sign in master**

master_austins

**Langton's sign in this PR**

this_langtons

**Langton's sign in master**

master_langtons

Nex-Pro:
Issue seen with change of view angle. Missing primitives. Unable to repro.

hifi-snap-by--on-2017-04-18_18-18-05
hifi-snap-by--on-2017-04-18_18-18-24

hifi-gustavo:
Build console output for commit: da55c96c4cc9a05f474b30e014e59d0e9d2956c0 available at: https://gist.github.com/321ecddb78053c9a35b47e299e02d66f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10206-da55c96.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10206-da55c96.dmg

AndrewMeadows:
@Nex-Pro I see the differences between your screenshots of the ductwork in dev-playa.

When I use the latest version of this PR I cannot reproduce: the ductwork as it does in your "**Duct model in this PR**" for both this PR and the latest dev-download.

Since I had rebased the changes here onto hifi/master right after (or shortly before) you started testing I also installed the previous build and tested that, with same results.

Could you re-test with the very latest dev-download?  Maybe we should track down the old version that was showing different results... perhaps a change was recently introduced in hifi/master that we didn't notice.  If so, the question becomes: what behavior is correct?  Did we fix something or did we introduce a bug?

samcake:
i have tested too on my i7 and haven't noticed any particular issue with culling so we can go ahead with that.

The rendering difference on the lighting over the Duct model is because of another bug on the normal mapping that we are currently addressing.
we are going to merge this pr
Thanks

PR #10212, opened 2017-04-13
avoid invisible avatars when source interface stops sending updates to avatar-mixer

AndrewMeadows:
* fix a mode by which other avatars can vanish if their interface is loaded
* avatar-mixer sends identity packets as 'reliable'

### DIFF
--- a/assignment-client/src/avatars/AvatarMixerSlave.cpp
+++ b/assignment-client/src/avatars/AvatarMixerSlave.cpp
@@ -69,7 +69,7 @@ void AvatarMixerSlave::processIncomingPackets(const SharedNodePointer& node) {
 int AvatarMixerSlave::sendIdentityPacket(const AvatarMixerClientData* nodeData, const SharedNodePointer& destinationNode) {
     int bytesSent = 0;
     QByteArray individualData = nodeData->getConstAvatarData()->identityByteArray();
-    auto identityPacket = NLPacket::create(PacketType::AvatarIdentity, individualData.size());
+    auto identityPacket = NLPacket::create(PacketType::AvatarIdentity, individualData.size(), true, true);
     individualData.replace(0, NUM_BYTES_RFC4122_UUID, nodeData->getNodeID().toRfc4122()); // FIXME, this looks suspicious
     bytesSent += individualData.size();
     identityPacket->write(individualData);
--- a/assignment-client/src/avatars/AvatarMixerSlave.cpp
+++ b/assignment-client/src/avatars/AvatarMixerSlave.cpp
@@ -67,15 +67,13 @@ void AvatarMixerSlave::processIncomingPackets(const SharedNodePointer& node) {
 
 
 int AvatarMixerSlave::sendIdentityPacket(const AvatarMixerClientData* nodeData, const SharedNodePointer& destinationNode) {
-    int bytesSent = 0;
     QByteArray individualData = nodeData->getConstAvatarData()->identityByteArray();
-    auto identityPacket = NLPacket::create(PacketType::AvatarIdentity, individualData.size(), true, true);
     individualData.replace(0, NUM_BYTES_RFC4122_UUID, nodeData->getNodeID().toRfc4122()); // FIXME, this looks suspicious
-    bytesSent += individualData.size();
-    identityPacket->write(individualData);
-    DependencyManager::get<NodeList>()->sendPacket(std::move(identityPacket), *destinationNode);
+    auto identityPackets = NLPacketList::create(PacketType::AvatarIdentity, QByteArray(), true, true);
+    identityPackets->write(individualData);
+    DependencyManager::get<NodeList>()->sendPacketList(std::move(identityPackets), *destinationNode);
     _stats.numIdentityPackets++;
-    return bytesSent;
+    return individualData.size();
 }
 
 static const int AVATAR_MIXER_BROADCAST_FRAMES_PER_SECOND = 45;
@@ -175,7 +173,6 @@ void AvatarMixerSlave::broadcastAvatarData(const SharedNodePointer& node) {
             }
         });
 
-        const uint64_t MIN_KEEP_ALIVE_PERIOD_FOR_OTHER_AVATAR = (AVATAR_UPDATE_TIMEOUT - 1) * USECS_PER_SECOND;
         AvatarSharedPointer thisAvatar = nodeData->getAvatarSharedPointer();
         ViewFrustum cameraView = nodeData->getViewFrustom();
         std::priority_queue<AvatarPriority> sortedAvatars;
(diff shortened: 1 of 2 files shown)


assignment-client/src/avatars/AvatarMixerSlave.cpp
```diff
@@ -69,7 +69,7 @@ void AvatarMixerSlave::processIncomingPackets(const SharedNodePointer& node) {
 int AvatarMixerSlave::sendIdentityPacket(const AvatarMixerClientData* nodeData, const SharedNodePointer& destinationNode) {
     int bytesSent = 0;
     QByteArray individualData = nodeData->getConstAvatarData()->identityByteArray();
-    auto identityPacket = NLPacket::create(PacketType::AvatarIdentity, individualData.size());
+    auto identityPacket = NLPacket::create(PacketType::AvatarIdentity, individualData.size(), true, true);
```
AndrewMeadows: Oops, avatar-mixer wasn't sending identity packets as 'reliable'.

ZappoMan: I am skeptical that this is the issue... the liklihood of packet loss for this seems very low and we've never sent this as reliable before.

assignment-client/src/avatars/AvatarMixerSlave.cpp
```diff
@@ -262,11 +263,17 @@ void AvatarMixerSlave::broadcastAvatarData(const SharedNodePointer& node) {
...
-                            ++numAvatarsHeldBack;
-                            shouldIgnore = true;
+                            // don't ignore this avatar if we haven't sent any update for a long while
+                            uint64_t lastBroadcastTime = nodeData->getLastBroadcastTime(avatarNode->getUUID());
+                            const AvatarMixerClientData* otherNodeData = reinterpret_cast<const AvatarMixerClientData*>(avatarNode->getLinkedData());
+                            if (lastBroadcastTime > otherNodeData->getIdentityChangeTimestamp() &&
+                                    lastBroadcastTime > startIgnoreCalculation - MIN_KEEP_ALIVE_PERIOD_FOR_OTHER_AVATAR) {
+                                ++numAvatarsHeldBack;
+                                shouldIgnore = true;
+                            }
```
AndrewMeadows: This is a workaround: if the source interface goes silent but not so silent that the node is actually removed, then the avatar-mixer will resend data at a long interval in an attempt to keep any remote avatars alive.

assignment-client/src/avatars/AvatarMixerSlave.cpp
```diff
@@ -302,8 +309,12 @@ void AvatarMixerSlave::broadcastAvatarData(const SharedNodePointer& node) {
...
-            // the time that Avatar B flagged an IDENTITY DATA change, send IDENTITY DATA about Avatar B to Avatar A.
-            if (nodeData->getLastBroadcastTime(otherNode->getUUID()) <= otherNodeData->getIdentityChangeTimestamp()) {
+            // the time that Avatar B flagged an IDENTITY DATA change
+            // or if no packet of any type has been sent for some time
+            // send IDENTITY DATA about Avatar B to Avatar A
+            uint64_t lastBroadcastTime = nodeData->getLastBroadcastTime(otherNode->getUUID());
+            if (lastBroadcastTime <= otherNodeData->getIdentityChangeTimestamp() ||
+                    startAvatarDataPacking > lastBroadcastTime + MIN_KEEP_ALIVE_PERIOD_FOR_OTHER_AVATAR) {
                 identityBytesSent += sendIdentityPacket(otherNodeData, node);
             }
```
AndrewMeadows: When sending keep-alive go ahead and resend identity data just in case the remote interface already deleted the corresponding AvatarData.

assignment-client/src/avatars/AvatarMixerSlave.cpp
howard-stearns: I'm also skeptical UNLESS there are consequences of the client not getting ones own identity packet. (Which happens now in order to get the mixer-assigned session display name.) Problems were observed for the avatars of struggling machines to not be visible. If the there was some consequence for the struggler missing their own identity.... but alas, I can't think of any (other than not knowing their own session display name, which would mess up the PAL.

ZappoMan:
This change doesn't make sense... it looks like it's going to cause identity packets to get sent much more frequently than they need to be sent.

assignment-client/src/avatars/AvatarMixerSlave.cpp
ZappoMan: Why is the source interface going silent? Is this a real thing? Source interfaces will typically be sending joint data for their idle animations.

assignment-client/src/avatars/AvatarMixerSlave.cpp
AndrewMeadows: This is not the fix you are looking for.

However, I felt it necessary to make this change so that I would stop mentioning this problem whenever yet another invisible avatar bug showed up.

AndrewMeadows:
Your understanding is incorrect.  At most the identity packet will be sent every 4 seconds... but only if no other data is being relayed.

Hint: the lastBroadcastTime is updated whenever **joint** or **identity** data are sent.

... which means there may well be a narrow race condition whereby joint data is broadcast **after** receipt of identity data, which would cause the identity data to never be sent (remote otherAvatar's identity data would not be updated).  However, it looks to me that the race condition doesn't exist iff joint data are sent and identity data are recieved on the same thread, which... might be the case

hifi-gustavo:
Build console output for commit: 63ea16f24b81a34e475e5b2e66f55b71348f1012 available at: https://gist.github.com/86bd6b39c0fa63ab07ae585f1b7f06ab<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10212-63ea16f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10212-63ea16f.dmg

assignment-client/src/avatars/AvatarMixerSlave.cpp
AndrewMeadows: Note: the `lastBroadcastTime` is update when sending either joint or identity data, which means there might be a small race condition in the first half of the `if` logic above.  That is, for absolute correctness we should be checking against `lastBroadcastIdentityTime`.

But... the race is probably narrow and might not exist at all if outgoing joint and incoming identity data are handled on the same thread.

assignment-client/src/avatars/AvatarMixerSlave.cpp
ZappoMan: as per slack conversation. I retract my concern, but we should make this an `NLPacketList` so it can be larger than an MTU

ZappoMan:
@AndrewMeadows - but if we're sending the identity packet reliably now... (so it can't be lost)... and we're going to still send duplicate "frame data" even if the sender hasn't sent us an update, then this extra sending of the identity packet is overkill... even every 4 seconds... if you have 1000 avatars resending this identity packet even every 4 seconds is going to be a lot of extra data... I don't think we need this extra belt & suspenders.... one pant elevation technique should be sufficient.

assignment-client/src/avatars/AvatarMixerSlave.cpp
```diff
@@ -302,8 +300,13 @@ void AvatarMixerSlave::broadcastAvatarData(const SharedNodePointer& node) {
...
 
             // If the time that the mixer sent AVATAR DATA about Avatar B to Avatar A is BEFORE OR EQUAL TO
-            // the time that Avatar B flagged an IDENTITY DATA change, send IDENTITY DATA about Avatar B to Avatar A.
-            if (nodeData->getLastBroadcastTime(otherNode->getUUID()) <= otherNodeData->getIdentityChangeTimestamp()) {
+            // the time that Avatar B flagged an IDENTITY DATA change
+            // or if no packet of any type has been sent for some time
+            // send IDENTITY DATA about Avatar B to Avatar A
+            const uint64_t AVATAR_UPDATE_STALE = AVATAR_UPDATE_TIMEOUT - USECS_PER_SECOND;
+            uint64_t lastBroadcastTime = nodeData->getLastBroadcastTime(otherNode->getUUID());
+            if (lastBroadcastTime <= otherNodeData->getIdentityChangeTimestamp() ||
```
ZappoMan: this is exactly the opposite change I asked for.

Remove resending identity, and only resend frame data every 4 seconds.

hifi-gustavo:
Build console output for commit: 14b60165efb5c5f6126a0c19eeee8a9c24461ae5 available at: https://gist.github.com/86bd6b39c0fa63ab07ae585f1b7f06ab<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10212-14b6016.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10212-14b6016.dmg

AndrewMeadows:
This PR doesn't actually fix the problem, but it does carve it down.  In some future PR we'll have to add a packet that allows the interface to request fresh identity data from the avatar-mixer for a particular avatar.

hifi-gustavo:
Build console output for commit: 5bdecfbfa0b59cf1406c73aa127d768d3799e6d9 available at: https://gist.github.com/86bd6b39c0fa63ab07ae585f1b7f06ab<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10212-5bdecfb.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10212-5bdecfb.dmg

cozza13:
QA test, went through the tutorial and went to another server.

PR #10223, opened 2017-04-14
Make setBlendshape changes persist

kunalgosar:
Fix for: https://highfidelity.fogbugz.com/f/cases/2782/MyAvatar-setBlendshape-Resets-after-a-single-frame

Testing Instructions:
- Try commands such as:
     > MyAvatar.setBlendshape("MouthSmile_R", 1.0);
     > MyAvatar.setBlendshape("MouthSmile_L", 1.0);
- Verify that the changes persist beyond just one frame

### DIFF
--- a/libraries/avatars/src/HeadData.cpp
+++ b/libraries/avatars/src/HeadData.cpp
@@ -39,7 +39,7 @@ HeadData::HeadData(AvatarData* owningAvatar) :
     _averageLoudness(0.0f),
     _browAudioLift(0.0f),
     _owningAvatar(owningAvatar),
-    _baseBlendshapeCoefficients(0, 0),
+    _baseBlendshapeCoefficients(0, 0)
 {
 
 }
@@ -96,6 +96,7 @@ const QVector<float>& HeadData::getSummedBlendshapeCoefficients() {
 
 void HeadData::setBlendshape(QString name, float val) {
     const auto& blendshapeLookupMap = getBlendshapesLookupMap();
+
     //Check to see if the named blendshape exists, and then set its value if it does
     auto it = blendshapeLookupMap.find(name);
     if (it != blendshapeLookupMap.end()) {
--- a/libraries/avatars/src/HeadData.cpp
+++ b/libraries/avatars/src/HeadData.cpp
@@ -38,7 +38,8 @@ HeadData::HeadData(AvatarData* owningAvatar) :
     _rightEyeBlink(0.0f),
     _averageLoudness(0.0f),
     _browAudioLift(0.0f),
-    _owningAvatar(owningAvatar)
+    _owningAvatar(owningAvatar),
+    _baseBlendshapeCoefficients(0, 0),
 {
 
 }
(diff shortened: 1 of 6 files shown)


hifi-gustavo:
Build console output for commit: d0b88c268392a7de45cb139ae10bb04e7951a211 available at: https://gist.github.com/421f5d59a658a5d737c759aded28dbb9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10223-d0b88c2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10223-d0b88c2.dmg

CaptainOculus:
image

libraries/avatars/src/HeadData.cpp
```diff
@@ -38,7 +39,8 @@ HeadData::HeadData(AvatarData* owningAvatar) :
     _rightEyeBlink(0.0f),
     _averageLoudness(0.0f),
     _browAudioLift(0.0f),
-    _owningAvatar(owningAvatar)
+    _owningAvatar(owningAvatar),
+    _baseBlendshapeCoefficients(QVector<float>(0, 0))
```
AndrewMeadows: @kunalgosar the build failed because there is a warning about order of instantiation in this constructor of the `HeadData` class.  The problem is that _baseBlendShapeCoefficients and _owningAvatar are declared in that order, but are initialized here in the reverse order.  The warning is saying that the compiler will reverse the order of the instantiation to agree with the order in which they were declared.

The code is technically correct, however we want to minimize all warnings so you'll have to fix it.  In this case, the best solution is to swap the two lines above rather than the order in the class declaration.

Also, use `0.0f` instead of `0` for floating point literal zero.

libraries/avatars/src/HeadData.cpp
```diff
@@ -86,6 +88,12 @@ static const QMap<QString, int>& getBlendshapesLookupMap() {
     return blendshapeLookupMap;
 }
 
+const QVector<float>& HeadData::getSummedBlendshapeCoefficients() {
+    for (int i = 0; i < _baseBlendshapeCoefficients.size(); i++) {
+        _blendshapeCoefficients[i] += _baseBlendshapeCoefficients[i];
+    }
+    return _blendshapeCoefficients;
+}
```
AndrewMeadows: I know we talked about this, and we were trying to avoid a big fat copy of QVector<float> but... this code looks dangerous to the naive reader, because it is summing _baseBlendshapeCoefficients into _blendshapeCoefficients and storing the results for later.  The only way this can be stable is if _blendshapeCoefficients is completely overwritten every frame... and it isn't clear here that is the case: the dev must go looking around elsewhere in the code to verify.

When I go looking for that proof I'm not satisfied.  It looks like each element of _blendshapeCoefficients could be overwritten by `HeadData::setBlendshape(QString name, float val)`.  But that only overwrites the coefficient one at a time and it isn't obvious that each coefficient is correctly singled out for overwrite every frame.

Alternatively they could be overwritten by `HeadData::setBlendshapes(const QVector<float>& blendshapeCoefficients)` but after looking at the code I'm pretty sure this is not necessarily called every frame.

So... I think we're going to have to bite the bullet and make that copy every frame.

BTW, there is a remote possibility that the two lists `_blendshapeCoefficients` and [_baseBlendshapeCoefficients](url) are not the same length --> there is a risk of crash for segfault!  I think the sum logic needs to check the sizes of the two vectors, only copy what is safe, and skip the rest.

libraries/avatars/src/HeadData.cpp
kunalgosar: Resolved this by switching the two statements

libraries/avatars/src/HeadData.cpp
kunalgosar: I resolved this by making a new vector, `_currBlendShapeCoefficients`, and each frame recopy the data from both `_blendshapeCoefficients` and `_baseBlendshapeCoefficients` into it.

hifi-gustavo:
Build console output for commit: ffdbc12968257b527d4f3cbb2376ca9bc30bc94e available at: https://gist.github.com/421f5d59a658a5d737c759aded28dbb9<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10223-ffdbc12.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10223-ffdbc12.dmg

libraries/avatars/src/HeadData.cpp
```diff
@@ -96,7 +117,10 @@ void HeadData::setBlendshape(QString name, float val) {
         if (_blendshapeCoefficients.size() <= it.value()) {
             _blendshapeCoefficients.resize(it.value() + 1);
```
AndrewMeadows: Right here, I think you need to: `_blendshapeCoefficients[it.value()] = 0.0f;`

AndrewMeadows: Ah nevermind.  I just read the documentation for QVector.

PR #10227, opened 2017-04-15
Introducing support for the ResourceBuffer binding point in the gpu lib, the Batch and the backends

samcake:
This PR introduces a new binding point for a gpu::buffer on the gpu::batch called ResourceBuffer.
A ResourceBuffer is implemented by binding the Buffer as a "GL Shader Storage Buffer Object" (SSBO) in the GL4.5 backend and by a "GL Texture Buffer Object" (TBO) in the GL 4.1 Backend.

ResourceBuffer are read only, are very comparable to the UniformBuffer  in terms of the usage, the main difference is that they can be much larger than UniformBuffer and their access latency is longer (because fetched in video memory like regular textures).

To test and prototype the feature, i modified the rendering of the DrawItemBounds job which is used to represent the bounding boxes of the render items through the debug tool script (debugDeferredLighting.js).
In that example, instead of drawing each item bounds individually with one drawcall per bound (passing the bound params as individual uniforms), we put all the bounds data in a gpu::Buffer and then render all the bounds in a single drawcall that fetch from the buffer binded as ResourceBuffer.

I had to change a few things to make it works, starting with grouping all the shader compilation related function as members from the GLBackend so we could specialize some part of the code depending on the Backend version, in particular how we handle that bind of the resource BUffer, in gl45 it s the simple ssbo, in gl41 we emulate the feature with tbo.

There are some constrains with the way things work right now that we may be able to improve on:
- The shader code is explicitly doing the 2 different code versions (one with ssbo, one with tbo)
   I would like to find a more integrated way of doing this where through a slh file we could provide an api to declare a ResourceBuffer and fetch from it without having to code the hugly details.
 For example see the code in drawItemBounds.slv, it s right there in you face #if gl41 else ....

- The elements stored in the resource buffer are expected to be multiple of vec4.
   THere is no detection of the element type that i would use in the the way we define the TBO in the gl41 case.
  We would want to have that supported at some point so we could have smaller than vec4 elements such as 2 scalars, 1 scalars, 2 Bytes, 1 Bytes and also support explicitely integer.

We can probably live with these constraints for a while since the main usage for it in the near future will work with it.
Feedback welcome...

While i was at it:
- the Transform Object buffer uses SSBO on gl45.
- I added a padding to the ITemBound class since that made it aligned with 2 Vec4 so for the DrawItemBound job, i can simply blit the input ItemBounds (std::vector) in a local Buffer which is feeding the shader directly. The performances are not affected and this gave me good ideas how we could use that padding with actual values and speed up other jobs of the pipeline... to be continued in future pr...
- Cleaned up the GL version dependent binding slots and glsl compilation headers

## TEST PLAN
Must test on WIndows, I5, I7 cpu, NVIDIA & AMD & intel gpu. toggling the env variable "HIFI_DISABLE_OPENGL_45" on and off to test with both GLBackends (41 and 45).
Must test on macos (just GL41).

Performances should be identical compared to dev-master (both backends) if not slightly faster in gl45 backend (because we now use SSBO for the TransformObjectBuffer).
THe difference in performance could be in the GPU timing.
Rendering shouldn't have changed.

Using the debug tool script debugDeferredLighting.js and checking the bottom check boxes that rely on the modified DrawItemBounds job to represent the bounding boxes of the different categories of items (opaques, transparents, ...) the rendering should just work and look the same in this pr as in dev master.
Eventually the batch timing and gpu timing should be slightly faster.
Using the debug tool script renderStats.js let you see the evolution of the number of drawcalls. as you toggle the check box of "opaques " for example (in the debugDeferredLIghting.js ui) you should see that in this pr, we just add 1 more drawcall to represent multiple boxes whereas with dev-master the number of drawcalls grow as much as the number of item bounds actually rendered.

That's it


libraries/gl/src/gl/GLShaders.h
```diff
@@ -17,12 +17,12 @@
...
 #ifdef SEPARATE_PROGRAM
-    bool compileShader(GLenum shaderDomain, const std::string& shaderSource, const std::string& defines, GLuint &shaderObject, GLuint &programObject);
+    bool compileShader(GLenum shaderDomain, const std::string& shaderSource, const std::string& defines, GLuint &shaderObject, GLuint &programObject, std::string& error = std::string());
 #else
-    bool compileShader(GLenum shaderDomain, const std::string& shaderSource, const std::string& defines, GLuint &shaderObject);
+    bool compileShader(GLenum shaderDomain, const std::string& shaderSource, const std::string& defines, GLuint &shaderObject, std::string& error = std::string());
 #endif
 
-    GLuint compileProgram(const std::vector<GLuint>& glshaders);
+    GLuint compileProgram(const std::vector<GLuint>& glshaders, std::string& error = std::string());
```
jherico: this will fail to compile on OSX.  can't pass a default argument to a non-const reference

libraries/gl/src/gl/GLShaders.h
```diff
@@ -17,12 +17,12 @@
 
 namespace gl {
 #ifdef SEPARATE_PROGRAM
-    bool compileShader(GLenum shaderDomain, const std::string& shaderSource, const std::string& defines, GLuint &shaderObject, GLuint &programObject);
+    bool compileShader(GLenum shaderDomain, const std::string& shaderSource, const std::string& defines, GLuint &shaderObject, GLuint &programObject, std::string& error = std::string());
 #else
-    bool compileShader(GLenum shaderDomain, const std::string& shaderSource, const std::string& defines, GLuint &shaderObject);
+    bool compileShader(GLenum shaderDomain, const std::string& shaderSource, const std::string& defines, GLuint &shaderObject, std::string& error = std::string());
```
jherico: this will fail to compile on OSX. can't pass a default argument to a non-const reference

hifi-gustavo:
Build console output for commit: 96960a505d8cbdabcacd7e031545f62135cab2c2 available at: https://gist.github.com/d26def7f7283cb9512c7597ea3a24709<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-96960a5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-96960a5.dmg

hifi-gustavo:
Build console output for commit: 038d01ce042f223060912fb3c4050b0c974cbcf8 available at: https://gist.github.com/d26def7f7283cb9512c7597ea3a24709<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-038d01c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-038d01c.dmg

hifi-gustavo:
Build console output for commit: 2b971710610e4a155cb8425e2599cf1d68331864 available at: https://gist.github.com/d26def7f7283cb9512c7597ea3a24709<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-2b97171.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-2b97171.dmg

hifi-gustavo:
Build console output for commit: 61ed5c3ce98252ee8f525380eb061cb58aaa7be8 available at: https://gist.github.com/d26def7f7283cb9512c7597ea3a24709<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-61ed5c3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-61ed5c3.dmg

OmegaHeron:
For anyone testing this (I'm doing Windows 10 with i5 and GTX970 plus Ubuntu 16 and GTX660M) the env var to set or not set for GL4.1 is HIFI_DISABLE_OPENGL_45 -- note extra _ before 45.  It will save you some time trying to figure out why you can't force GL4.1

@samcake - edit your test notes, please.  Env var is HIFI_DISABLE_OPENGL_45

OmegaHeron:

GL4.5
General smoke test = pass, everything looks same to me this PR to current Master.
Comparing same scene using renderStats and debugDeferredLighting I see draw calls with master go 72 to 157 when I toggle display opaques while with this PR it goes 72 to 73.

GL4.1
Same results as 4.5 - drawcalls down, things look same Master to PR.

Ubuntu 16.04 - i7 i7-3720QM - GTX660M
General smoke, GL4.5 and GL4.1 = same results as Windows testing.  Pass.

hifi-gustavo:
Build console output for commit: f8a59aa255e4f6d1352c4bf4b7805039a60e3ccd available at: https://gist.github.com/d26def7f7283cb9512c7597ea3a24709<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-f8a59aa.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-f8a59aa.dmg

libraries/gpu-gl/src/gpu/gl/GLBackendPipeline.cpp
```diff
@@ -191,11 +191,45 @@ void GLBackend::releaseResourceTexture(uint32_t slot) {
 }
 
 void GLBackend::resetResourceStage() {
+    for (uint32_t i = 0; i < _resource._buffers.size(); i++) {
+        releaseResourceBuffer(i);
```
samcake: Yes the release and bind calls are specialized in both backend's versions

samcake:
Currently testing on AMD  rx 470 (i5)
It s working fine with 4.1 backend, not with 4.5 backend, investigating

samcake:
Tested on my mac book and it is working as expected (it s using 4.1 backend)
It s a Pass on macos

hifi-gustavo:
Build console output for commit: 0a615f35a7bb758f934eb569c6f4d13ed7764581 available at: https://gist.github.com/d26def7f7283cb9512c7597ea3a24709<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-0a615f3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-0a615f3.dmg

samcake:
And this last change to the shaders is fixing it on AMD.
I ll review once more on nvidia.

hifi-gustavo:
Build console output for commit: 733e268391194780a131ec4c77a83053c648a1ed available at: https://gist.github.com/d26def7f7283cb9512c7597ea3a24709<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-733e268.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10227-733e268.dmg

samcake:
ANd finally checked again on nvidia after the last commit and it works fine in both 41 and 45
Pass

PR #10249, opened 2017-04-20
Add AvatarIdentityRequest protocol to avatar-mixer

AndrewMeadows:
* fix "invisible avatar" bug once and for all (I mean it this time)
* fix bug where old avatars remain visible for a few seconds after teleporting to new domain
* when signing out of (or into) a domain other avatars will vanish... but will be replaced by new copies

### DIFF
--- a/assignment-client/src/avatars/AvatarMixer.cpp
+++ b/assignment-client/src/avatars/AvatarMixer.cpp
@@ -394,6 +394,13 @@ void AvatarMixer::handleRequestsDomainListDataPacket(QSharedPointer<ReceivedMess
 }
 
 void AvatarMixer::handleAvatarIdentityRequest(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode) {
+    AvatarMixerClientData* nodeData = dynamic_cast<AvatarMixerClientData*>(senderNode->getLinkedData());
+    if (nodeData) {
+        while (message->getBytesLeftToRead() >= NUM_BYTES_RFC4122_UUID) {
+            QUuid otherNodeID = QUuid::fromRfc4122(message->readWithoutCopy(NUM_BYTES_RFC4122_UUID));
+            nodeData->clearLastBroadcastTime(otherNodeID);
+        }
+    }
 }
 
 void AvatarMixer::handleAvatarIdentityPacket(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode) {
--- a/assignment-client/src/avatars/AvatarMixer.cpp
+++ b/assignment-client/src/avatars/AvatarMixer.cpp
@@ -53,6 +53,7 @@ AvatarMixer::AvatarMixer(ReceivedMessage& message) :
     packetReceiver.registerListener(PacketType::NodeIgnoreRequest, this, "handleNodeIgnoreRequestPacket");
     packetReceiver.registerListener(PacketType::RadiusIgnoreRequest, this, "handleRadiusIgnoreRequestPacket");
     packetReceiver.registerListener(PacketType::RequestsDomainListData, this, "handleRequestsDomainListDataPacket");
+    packetReceiver.registerListener(PacketType::AvatarIdentityRequest, this, "handleAvatarIdentityRequest");
 
     auto nodeList = DependencyManager::get<NodeList>();
     connect(nodeList.data(), &NodeList::packetVersionMismatch, this, &AvatarMixer::handlePacketVersionMismatch);
@@ -390,6 +391,9 @@ void AvatarMixer::handleRequestsDomainListDataPacket(QSharedPointer<ReceivedMess
     }
     auto end = usecTimestampNow();
     _handleRequestsDomainListDataPacketElapsedTime += (end - start);
+}
+
+void AvatarMixer::handleAvatarIdentityRequest(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode) {
 }
 
 void AvatarMixer::handleAvatarIdentityPacket(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode) {
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -147,11 +147,6 @@ void Avatar::init() {
     _initialized = true;
 }
 
-void Avatar::processAvatarIdentity(const Identity& identity, bool& identityChanged, bool& displayNameChanged) {
-    AvatarData::processAvatarIdentity(identity, identityChanged, displayNameChanged);
-    _identityRequestExpiry = 0;
-}
-
 glm::vec3 Avatar::getChestPosition() const {
     // for now, let's just assume that the "chest" is halfway between the root and the neck
     glm::vec3 neckPosition;
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -147,6 +147,11 @@ void Avatar::init() {
     _initialized = true;
 }
 
+void Avatar::processAvatarIdentity(const Identity& identity, bool& identityChanged, bool& displayNameChanged) {
+    AvatarData::processAvatarIdentity(identity, identityChanged, displayNameChanged);
+    _identityRequestExpiry = 0;
+}
+
 glm::vec3 Avatar::getChestPosition() const {
     // for now, let's just assume that the "chest" is halfway between the root and the neck
     glm::vec3 neckPosition;
@@ -378,7 +383,12 @@ void Avatar::simulate(float deltaTime, bool inView) {
     {
         PROFILE_RANGE(simulation, "updateJoints");
         if (inView && _hasNewJointData) {
-            _skeletonModel->getRig()->copyJointsFromJointData(_jointData);
+            if (_skeletonModel->getRig()->getNumJoints() == _jointData.size()) {
+                _skeletonModel->getRig()->copyJointsFromJointData(_jointData);
+            } else {
+                // identity data is probably stale --> request new data
+                requestIdentityData();
+            }
             glm::mat4 rootTransform = glm::scale(_skeletonModel->getScale()) * glm::translate(_skeletonModel->getOffset());
             _skeletonModel->getRig()->computeExternalPoses(rootTransform);
             _jointDataSimulationRate.increment();
@@ -445,6 +455,21 @@ float Avatar::getSimulationRate(const QString& rateName) const {
         return _jointDataSimulationRate.rate();
     }
     return 0.0f;
+}
+
+void Avatar::requestIdentityData() {
+    uint64_t now = usecTimestampNow();
+	if (_identityRequestExpiry < now) {
+		// send the packet immediately
+    	auto packet = NLPacket::create(PacketType::AvatarIdentityRequest, NUM_BYTES_RFC4122_UUID);
+    	packet->write(getSessionUUID().toRfc4122());
+    	auto nodeList = DependencyManager::get<NodeList>();
+    	nodeList->broadcastToNodes(std::move(packet), NodeSet() << NodeType::AvatarMixer);
+
+		// and set expiry for next request
+		const uint64_t IDENTITY_REQUEST_PERIOD = 2 * USECS_PER_SECOND;
+		_identityRequestExpiry = now + IDENTITY_REQUEST_PERIOD;
+	}
 }
 
 bool Avatar::isLookingAtMe(AvatarSharedPointer avatar) const {
--- a/interface/src/avatar/AvatarManager.cpp
+++ b/interface/src/avatar/AvatarManager.cpp
@@ -362,8 +362,6 @@ void AvatarManager::processAvatarDataPacket(QSharedPointer<ReceivedMessage> mess
 }
 
 void AvatarManager::handleRemovedAvatar(const AvatarSharedPointer& removedAvatar, KillAvatarReason removalReason) {
-    AvatarHashMap::handleRemovedAvatar(removedAvatar, removalReason);
-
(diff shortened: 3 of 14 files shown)


AndrewMeadows:
There is a general problem where the avatar-mixer doesn't know that the interface has lost the identity information for some other Avatar in view.  When this happens the avatar-mixer will continue to send joint data to the interface, and the interface will create an Avatar instance to accept that data, but it will be unable to render it... until it knows what model to use.

This PR introduces a system whereby an interface can request new identity data for an Avatar that has mismatching joint data.  This will catch cases where the Avatar has no bones at all, or has a different number of them.

assignment-client/src/avatars/AvatarMixer.cpp
```diff
@@ -392,6 +393,16 @@ void AvatarMixer::handleRequestsDomainListDataPacket(QSharedPointer<ReceivedMess
...
 
+void AvatarMixer::handleAvatarIdentityRequest(QSharedPointer<ReceivedMessage> message, SharedNodePointer senderNode) {
+    AvatarMixerClientData* nodeData = dynamic_cast<AvatarMixerClientData*>(senderNode->getLinkedData());
+    if (nodeData) {
+        while (message->getBytesLeftToRead() >= NUM_BYTES_RFC4122_UUID) {
+            QUuid otherNodeID = QUuid::fromRfc4122(message->readWithoutCopy(NUM_BYTES_RFC4122_UUID));
+            nodeData->clearLastBroadcastTime(otherNodeID);
+        }
+    }
+}
```
AndrewMeadows: When the avatar-mixer gets an AvatarIdentityRequest packet it will reset the broadcast time for the target avatar in the sender's records.  This will trigger an fresh identity packet through the regular logic flow later.

interface/src/avatar/Avatar.cpp
```diff
@@ -378,7 +378,12 @@ void Avatar::simulate(float deltaTime, bool inView) {
     {
         PROFILE_RANGE(simulation, "updateJoints");
         if (inView && _hasNewJointData) {
-            _skeletonModel->getRig()->copyJointsFromJointData(_jointData);
+            if (_skeletonModel->getRig()->getNumJoints() == _jointData.size()) {
+                _skeletonModel->getRig()->copyJointsFromJointData(_jointData);
+            } else {
+                // identity data is probably stale --> request new data
+                requestIdentityData();
+            }
```
AndrewMeadows: This is where we check for bad identity data.  The only condition we use is that the incoming joint data count does not match the number of joints available in the `Rig`.

interface/src/avatar/Avatar.cpp
```diff
@@ -447,6 +452,21 @@ float Avatar::getSimulationRate(const QString& rateName) const {
...
+    	auto packet = NLPacket::create(PacketType::AvatarIdentityRequest, NUM_BYTES_RFC4122_UUID);
+    	packet->write(getSessionUUID().toRfc4122());
+    	auto nodeList = DependencyManager::get<NodeList>();
+    	nodeList->broadcastToNodes(std::move(packet), NodeSet() << NodeType::AvatarMixer);
+
+		// and set expiry for next request
+		const uint64_t IDENTITY_REQUEST_PERIOD = 2 * USECS_PER_SECOND;
+		_identityRequestExpiry = now + IDENTITY_REQUEST_PERIOD;
+	}
+}
```
AndrewMeadows: `Avatar::requestIdentityData()` will send immediately unless the expiry is not old enough.

interface/src/avatar/Avatar.cpp
ZappoMan: If the reason the joints don't match is because the model can't be downloaded... this is going to result in a DOS against the avatar mixer.

I am dubious of this approach.

interface/src/avatar/AvatarManager.cpp
```diff
@@ -399,18 +399,14 @@ void AvatarManager::clearOtherAvatars() {
...
-        }
-        ++avatarIterator;
+		if (avatar == _myAvatar || !avatar->isInitialized()) {
+            // don't remove myAvatar or uninitialized avatars from the list
+            ++avatarIterator;
+        } else {
+            auto removedAvatar = avatarIterator.value();
+            avatarIterator = _avatarHash.erase(avatarIterator);
+            handleRemovedAvatar(removedAvatar);
+		}
```
AndrewMeadows: This change restores old code that would remove all other avatars (rather than just try to remove them from the scene) whenever we reset connection to a domain (which happens when we "sign-out/in" and when we teleport to a new domain).

Note: I think what was happening in the "teleport to new domain" case is that, even though all other avatars were being removed from the scene (they would no longer render) there were joint data updates on the wire from the old avatar-mixer, which would resurrect the avatars into the scene before the connection to the old avatar-mixer had been closed.

Note: when signing out/in the interface will now lose all identity data for nearby avatars (they are being removed!)... and the avatar-mixer doesn't know this.  Unless the interface has a way to ask the avatar-mixer for new identity data the nearby avatars will go invisible.

interface/src/avatar/Avatar.cpp
ZappoMan: doesn't this result in clients always requesting identity data every 2 seconds?

libraries/networking/src/udt/PacketHeaders.h
```diff
@@ -206,6 +207,7 @@ const PacketVersion VERSION_ENTITIES_LAST_EDITED_BY = 65;
 const PacketVersion VERSION_ENTITIES_SERVER_SCRIPTS = 66;
 const PacketVersion VERSION_ENTITIES_PHYSICS_PACKET = 67;
 const PacketVersion VERSION_ENTITIES_ZONE_FILTERS = 68;
+const PacketVersion VERSION_AVATAR_IDENTITY_REQUEST = 70;
```
AndrewMeadows: I'm using `70` instead of `69` here because I know @sethalves has a pending PR that tries to introduce a `69`. This is just a courtesy to make the conflict resolution easier.

interface/src/avatar/Avatar.cpp
ZappoMan: nvrmnd

interface/src/avatar/AvatarManager.cpp
```diff
@@ -399,18 +399,14 @@ void AvatarManager::clearOtherAvatars() {
     AvatarHash::iterator avatarIterator =  _avatarHash.begin();
     while (avatarIterator != _avatarHash.end()) {
         auto avatar = std::static_pointer_cast<Avatar>(avatarIterator.value());
-        if (avatar != _myAvatar) {
-            if (avatar->isInScene()) {
```
ZappoMan: Did you mean to delete this code? Don't we need to remove the avatar from the scene?

hifi-gustavo:
Build console output for commit: 081cc154b005474f7b9556d1b92aa3890013b311 available at: https://gist.github.com/8c812709243f4f293cf8b0a52291fd8f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10249-081cc15.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10249-081cc15.dmg

hifi-gustavo:
Build console output for commit: cd165de4566be04f0d881f9825d91184d0c8d36e available at: https://gist.github.com/8c812709243f4f293cf8b0a52291fd8f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10249-cd165de.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10249-cd165de.dmg

interface/src/avatar/AvatarManager.cpp
AndrewMeadows: `AvatarManager::handleRemovedAvatar()` adds to `AvatarManager::_avatarsToFade` list whose elements will have their scale animated down to something small.  Eventually when the scale is small enough the `Avatar` will be removed from the scene and SharedPointer magic will delete the `Avatar` instance.

interface/src/avatar/AvatarManager.cpp
```diff
@@ -330,40 +330,7 @@ AvatarSharedPointer AvatarManager::newSharedAvatar() {
...
-                render::Transaction transaction;
-                avatar->addToScene(avatar, scene, transaction);
-                if (scene) {
-                    scene->enqueueTransaction(transaction);
-                }
-            }
-        }
-    }
-}
-
```
AndrewMeadows: This was an override of `AvatarHashMap::processAvatarDataPacket()` which would do extra work to reinsert Avatars into the scene whenever they received a data update but were not in the scene.  This was a workaround for the bug where avatars would go invisible when signing in/out of the domain.  I'm removing this workaround restoring the original version of `AvatarHashMap::processAvatarDataPacket()`.

hifi-gustavo:
Build console output for commit: 6547230681ebc12aed66e7cce3e8f661c3741828 available at: https://gist.github.com/8c812709243f4f293cf8b0a52291fd8f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10249-6547230.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10249-6547230.dmg

AndrewMeadows:
To test: 

PART-1
(1) Login with two accounts: AvatarA and AvatarB, such that they can see each other (the domain does not matter for PART-1 since it is an interface-only change).
(2) AvatarA: teleport to a different domain, but at a location such that you can still see the corresponding (X,Y,Z) position where AvatarB was
--> In this PR AvatarB should immediatly shrink to small size on teleport
--> In hifi/master AvatarB will probably be visible for about 5 seconds after teleport at which point it will timeout and shrink to small size.

PART-2
(3) Run a local domain (you need the avatar-mixer from this PR)
(4) Login AvatarA and AvatarB such that they can see each other
(5) AvatarA: sign out of the domain (don't log out, but just revoke your login credentials) or else sign in as the case may be.
--> In this PR the avatar is expected to shrink to small size, then reappear
--> In hifi/master the avarar is expected to disappear immediately (no shrinking), then reappear

PART-3
(6) Run a personal HTTP server
(7) Copy this asset to your personal HTTP server (old default legless robot):
http://mpassets.highfidelity.com/1b7e1e7c-6c0b-4f20-9cd0-1d5ccedae620-v1/bb64e937acf86447f6829767e958073c.fst
(8) Login AvatarA and use the new URL for AvatarA's appearance
(9) AFTER setting AvatarA's new appearance but BEFORE logging in AvatarB... remove the asset from your personal HTTP server so that AvatarB's asset fetch will fail
(10) Login AvatarB
--> In both cases (hifi/master and this PR) AvatarB should see AvatarA show up as the default avatar in T-
pose
(11) Verify that packets are passing through the avatar-mixer: that is, move AvatarA around
--> AvatarB should see AvatarA changing position and rotation but still in T-pose
(12) Restore the asset that was removed in step (9)
--> in this PR AvatarB should see AvatarA's correct model show up
--> in hifi/master AvatarB should see AvatarA remain as default in T-pose

ZappoMan:
I'm still not convinced this is the right approach. I don't see how we could possibly have gotten into an invisible avatar case... It seems to me that whatever the root cause of the avatar being culled from the client really needs to be addressed.

I also think you should de-conflate the invisible fix from the "disconnect from domain hides avatars" fix... these are really two different issues and should be two different PRs.

ZappoMan:
In digging through the log file that Salina posted to the greetsquad channel. It looks like the root cause of her invisible avatars was that *her client* froze for 5 and later 7 seconds. You can tell because you see this in here log:
```
[04/20 14:38:40] [WARNING] [hifi.entities] kinematic timestep =  5.60088  truncated to  1
[04/20 14:38:40] [WARNING] [hifi.entities] kinematic timestep =  5.60088  truncated to  1
[04/20 14:38:40] [WARNING] [hifi.entities] kinematic timestep =  5.60088  truncated to  1
[04/20 14:38:40] [WARNING] [hifi.entities] kinematic timestep =  5.60088  truncated to  1
[04/20 14:38:40] [WARNING] [hifi.entities] kinematic timestep =  5.60088  truncated to  1
[04/20 14:38:40] [WARNING] [hifi.entities] kinematic timestep =  5.60088  truncated to  1
[04/20 14:38:40] [WARNING] [hifi.entities] kinematic timestep =  5.60088  truncated to  1
```

Then immediately following this, you see that ALL of her local avatars are removed from the AvatarHashMap with "reason" 0...

```
[04/20 14:38:40] [DEBUG] [hifi.avatars] Removed avatar with UUID "d3b54f55-cb70-404b-9e8d-82e96971bc77" from AvatarHashMap 0
[04/20 14:38:40] [DEBUG] [hifi.avatars] Removed avatar with UUID "2cea632c-3924-4180-8633-d621e8d1df51" from AvatarHashMap 0
[04/20 14:38:40] [DEBUG] [hifi.avatars] Removed avatar with UUID "d79d0c17-e416-4efd-8478-081a0e368529" from AvatarHashMap 0
[04/20 14:38:40] [DEBUG] [hifi.avatars] Removed avatar with UUID "49c56731-574d-426b-bce3-c3644439a6e6" from AvatarHashMap 0
[04/20 14:38:40] [DEBUG] [hifi.avatars] Removed avatar with UUID "13a01546-3c08-4234-a54c-e21552d23759" from AvatarHashMap 0
```

I think we're approaching this problem from the wrong direction, and the actual problem is that the AvatarManager on the local client is too aggressive, and is removing avatars in cases when it's local client has been stalled for 5 seconds... Clearly we don't want to stall the main loop for 5 seconds, but if we do we shouldn't be removing the local avatars.

AndrewMeadows:
Closing this. We'll fix it another way.

PR #10255, opened 2017-04-21
Fix two cases for avatars stuck in T-Pose

ZappoMan:
Fixes to a couple of different bug vectors for avatars getting stuck in t-pose:

- **Issue 1**: Avatar Skeleton URL fails to download because of temporary poor network conditions -- we saw some document cases of this in the logs we reviewed. We now allow the resource cache to attempt to refetch, and we won't switch to the default avatar until multiple resource failures have occurred.

- **Issue 2**: Out of order identity packets from the avatar-mixer. After switching to reliable packets, we started seeing some out of order identity packets from the mixer, which can sometimes include empty skeleton URLs. Which would reset to the default skeleton after a previously correct url was already set.

This also fixes a related issue with Identity packets:
- **Issue 3**: The client was sending identity packets to the mixer every 1 ms... this was spamming the network with unneeded traffic.

This also fixes a related issue with invisible avatars:
- **Issue 4**: The client would remove its local copies of avatar identity information after any interruption of 5 seconds, this would result in invisible avatars.

**NOTE:** This PR has a protocol change, you will need to test with this build of sandbox and interface.

# **Test Plan:**

**T-Pose tests**
1. Smoke test the product paying particular attention to avatar loading, making sure you see people animating, wearing their correct avatar.
2. Make sure to test the PAL, handshake, and other things that related to identity packets.
3. Test with a network conditioner that will cause some packet loss and/or out of order packets.
4. If you can force a web server to timeout on some requests, use that to test case 1.
5. Be sure to test a domain server restart while client is connected.
6. Be sure to test kicking the client via the domain server page.
7. Test the "Sign Post Chat V1" app from the marketplace -- https://highfidelity.com/marketplace/items/dc8a64dd-658a-4031-97a9-d79df1322e4c .
8. Test the "handshake" - make sure it works the same as master
9. Test the "PAL" - make sure it works the same as master
10. Test a sandbox with an incorrectly set "system clock" and make sure that clients still behave normally for all these tests.
11. Verify that changing your avatar's skeleton or display name is properly sent out to all other clients connected to your domain.

**Invisible Avatar tests...**

12. Login to a domain that has at least one other avatar visible.
13. Sign into the domain (or sign out, as the case may be) via File --> Login / Sign Up
--> The other avatar (and entire world) should vanish but all come back in a few seconds

14. Login to a domain that has at least one other avatar visible.
15. Have the other avatar logout.
--> The other avatar should shrink to a small size and vanish.

[...]

PR #10277, opened 2017-04-25
fixes invisible avatars caused by client timeouts

AndrewMeadows:
* remove Avatar timeout and related code in interface
* remove unused cruft in Node

### DIFF
--- a/assignment-client/src/avatars/AvatarMixerSlave.cpp
+++ b/assignment-client/src/avatars/AvatarMixerSlave.cpp
@@ -263,16 +263,8 @@ void AvatarMixerSlave::broadcastAvatarData(const SharedNodePointer& node) {
                         // make sure we haven't already sent this data from this sender to this receiver
                         // or that somehow we haven't sent
                         if (lastSeqToReceiver == lastSeqFromSender && lastSeqToReceiver != 0) {
-                            // don't ignore this avatar if we haven't sent any update for a long while
-                            // in an effort to prevent other interfaces from deleting a stale avatar instance
-                            uint64_t lastBroadcastTime = nodeData->getLastBroadcastTime(avatarNode->getUUID());
-                            const AvatarMixerClientData* otherNodeData = reinterpret_cast<const AvatarMixerClientData*>(avatarNode->getLinkedData());
-                            const uint64_t AVATAR_UPDATE_STALE = AVATAR_UPDATE_TIMEOUT - USECS_PER_SECOND;
-                            if (lastBroadcastTime > otherNodeData->getIdentityChangeTimestamp() &&
-                                    lastBroadcastTime + AVATAR_UPDATE_STALE > startIgnoreCalculation) {
-                                ++numAvatarsHeldBack;
-                                shouldIgnore = true;
-                            }
+                            ++numAvatarsHeldBack;
+                            shouldIgnore = true;
                         } else if (lastSeqFromSender - lastSeqToReceiver > 1) {
                             // this is a skip - we still send the packet but capture the presence of the skip so we see it happening
                             ++numAvatarsWithSkippedFrames;
@@ -285,7 +277,7 @@ void AvatarMixerSlave::broadcastAvatarData(const SharedNodePointer& node) {
         int avatarRank = 0;
 
         // this is overly conservative, because it includes some avatars we might not consider
-        int remainingAvatars = (int)sortedAvatars.size(); 
+        int remainingAvatars = (int)sortedAvatars.size();
 
         while (!sortedAvatars.empty()) {
             AvatarPriority sortData = sortedAvatars.top();
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -115,8 +115,6 @@ Avatar::Avatar(QThread* thread, RigPointer rig) :
 }
 
 Avatar::~Avatar() {
-    assert(isDead()); // mark dead before calling the dtor
-
     auto treeRenderer = DependencyManager::get<EntityTreeRenderer>();
     EntityTreePointer entityTree = treeRenderer ? treeRenderer->getTree() : nullptr;
     if (entityTree) {
@@ -447,6 +445,24 @@ float Avatar::getSimulationRate(const QString& rateName) const {
     return 0.0f;
 }
 
+void Avatar::getIdentity(Identity& identity) const {
+    identity.uuid = getSessionUUID();
+    identity.skeletonModelURL = _skeletonModelURL;
+    identity.attachmentData = _attachmentData;
+    identity.displayName = _displayName;
+    identity.sessionDisplayName = _sessionDisplayName;
+    identity.avatarEntityData = _avatarEntityData;
+}
+
+void Avatar::setIdentity(const Identity& identity) {
+    assert(identity.uuid == getSessionUUID());
+    setSkeletonModelURL(identity.skeletonModelURL);
+    _attachmentData = identity.attachmentData;
+    _displayName = identity.displayName;
+    _sessionDisplayName = identity.sessionDisplayName;
+    _avatarEntityData = identity.avatarEntityData;
+}
+
 bool Avatar::isLookingAtMe(AvatarSharedPointer avatar) const {
     const float HEAD_SPHERE_RADIUS = 0.1f;
     glm::vec3 theirLookAt = dynamic_pointer_cast<Avatar>(avatar)->getHead()->getLookAtPosition();
@@ -510,12 +526,13 @@ static TextRenderer3D* textRenderer(TextRendererType type) {
 void Avatar::addToScene(AvatarSharedPointer self, const render::ScenePointer& scene, render::Transaction& transaction) {
     auto avatarPayload = new render::Payload<AvatarData>(self);
     auto avatarPayloadPointer = Avatar::PayloadPointer(avatarPayload);
-    _renderItemID = scene->allocateID();
-    transaction.resetItem(_renderItemID, avatarPayloadPointer);
-    _skeletonModel->addToScene(scene, transaction);
+    if (_skeletonModel->addToScene(scene, transaction)) {
+        _renderItemID = scene->allocateID();
+        transaction.resetItem(_renderItemID, avatarPayloadPointer);
 
-    for (auto& attachmentModel : _attachmentModels) {
-        attachmentModel->addToScene(scene, transaction);
+        for (auto& attachmentModel : _attachmentModels) {
+            attachmentModel->addToScene(scene, transaction);
+        }
     }
 }
 
--- a/interface/src/avatar/Avatar.cpp
+++ b/interface/src/avatar/Avatar.cpp
@@ -445,24 +445,6 @@ float Avatar::getSimulationRate(const QString& rateName) const {
     return 0.0f;
 }
 
-void Avatar::getIdentity(Identity& identity) const {
-    identity.uuid = getSessionUUID();
-    identity.skeletonModelURL = _skeletonModelURL;
-    identity.attachmentData = _attachmentData;
-    identity.displayName = _displayName;
-    identity.sessionDisplayName = _sessionDisplayName;
-    identity.avatarEntityData = _avatarEntityData;
-}
-
-void Avatar::setIdentity(const Identity& identity) {
-    assert(identity.uuid == getSessionUUID());
-    setSkeletonModelURL(identity.skeletonModelURL);
-    _attachmentData = identity.attachmentData;
-    _displayName = identity.displayName;
-    _sessionDisplayName = identity.sessionDisplayName;
-    _avatarEntityData = identity.avatarEntityData;
-}
-
 bool Avatar::isLookingAtMe(AvatarSharedPointer avatar) const {
     const float HEAD_SPHERE_RADIUS = 0.1f;
     glm::vec3 theirLookAt = dynamic_pointer_cast<Avatar>(avatar)->getHead()->getLookAtPosition();
--- a/interface/src/avatar/AvatarManager.cpp
+++ b/interface/src/avatar/AvatarManager.cpp
@@ -333,11 +333,6 @@ void AvatarManager::handleRemovedAvatar(const AvatarSharedPointer& removedAvatar
     // class in this context so we can call methods that don't exist at the base class.
     auto avatar = std::static_pointer_cast<Avatar>(removedAvatar);
 
-    // there is no way to request identity data from the avatar-mixer
-    // therefore whenever we remove an avatar we cache the identity in case we need it later
-    AvatarData::Identity identity;
-    avatar->getIdentity(identity);
-
     AvatarMotionState* motionState = avatar->getMotionState();
     if (motionState) {
         _motionStatesThatMightUpdate.remove(motionState);
--- a/interface/src/avatar/AvatarManager.cpp
+++ b/interface/src/avatar/AvatarManager.cpp
@@ -326,16 +326,6 @@ AvatarSharedPointer AvatarManager::newSharedAvatar() {
     return std::make_shared<Avatar>(qApp->thread(), std::make_shared<Rig>());
 }
 
-AvatarSharedPointer AvatarManager::addAvatar(const QUuid& sessionUUID, const QWeakPointer<Node>& mixerWeakPointer) {
-    AvatarSharedPointer avatarData = AvatarHashMap::addAvatar(sessionUUID, mixerWeakPointer);
-    QMap<QUuid, AvatarData::Identity>::iterator itr =  _identityCache.find(sessionUUID);
-	if (itr != _identityCache.end()) {
-        auto avatar = std::static_pointer_cast<Avatar>(avatarData);
-        avatar->setIdentity(*itr);
-    }
-    return avatarData;
-}
-
 void AvatarManager::handleRemovedAvatar(const AvatarSharedPointer& removedAvatar, KillAvatarReason removalReason) {
     AvatarHashMap::handleRemovedAvatar(removedAvatar, removalReason);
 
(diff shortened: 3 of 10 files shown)


assignment-client/src/avatars/AvatarMixerSlave.cpp
```diff
@@ -263,16 +263,8 @@ void AvatarMixerSlave::broadcastAvatarData(const SharedNodePointer& node) {
...
-                            uint64_t lastBroadcastTime = nodeData->getLastBroadcastTime(avatarNode->getUUID());
-                            const AvatarMixerClientData* otherNodeData = reinterpret_cast<const AvatarMixerClientData*>(avatarNode->getLinkedData());
-                            const uint64_t AVATAR_UPDATE_STALE = AVATAR_UPDATE_TIMEOUT - USECS_PER_SECOND;
-                            if (lastBroadcastTime > otherNodeData->getIdentityChangeTimestamp() &&
-                                    lastBroadcastTime + AVATAR_UPDATE_STALE > startIgnoreCalculation) {
-                                ++numAvatarsHeldBack;
-                                shouldIgnore = true;
-                            }
+                            ++numAvatarsHeldBack;
+                            shouldIgnore = true;
```
AndrewMeadows: I'm removing this hack in the avatar-mixer which would resend "frame" data in a flawed effort to keep the avatar alive on the receiving interface.

interface/src/avatar/Avatar.cpp
```diff
@@ -447,6 +445,24 @@ float Avatar::getSimulationRate(const QString& rateName) const {
...
+}
+
+void Avatar::setIdentity(const Identity& identity) {
+    assert(identity.uuid == getSessionUUID());
+    setSkeletonModelURL(identity.skeletonModelURL);
+    _attachmentData = identity.attachmentData;
+    _displayName = identity.displayName;
+    _sessionDisplayName = identity.sessionDisplayName;
+    _avatarEntityData = identity.avatarEntityData;
+}
```
AndrewMeadows: `Avatar` class knows how to get/set `AvatarData::Identity`.

interface/src/avatar/Avatar.cpp
```diff
@@ -510,12 +526,13 @@ static TextRenderer3D* textRenderer(TextRendererType type) {
...
-    _skeletonModel->addToScene(scene, transaction);
+    if (_skeletonModel->addToScene(scene, transaction)) {
+        _renderItemID = scene->allocateID();
+        transaction.resetItem(_renderItemID, avatarPayloadPointer);
 
-    for (auto& attachmentModel : _attachmentModels) {
-        attachmentModel->addToScene(scene, transaction);
+        for (auto& attachmentModel : _attachmentModels) {
+            attachmentModel->addToScene(scene, transaction);
+        }
```
AndrewMeadows: Only add `Avatar` to the scene if the `SkeletonModel` was successfully added.

interface/src/avatar/AvatarManager.cpp
```diff
@@ -213,10 +213,6 @@ void AvatarManager::updateOtherAvatars(float deltaTime) {
             }
         }
         avatar->animateScaleChanges(deltaTime);
-        if (avatar->shouldDie()) {
-            avatar->die();
-            removeAvatar(avatar->getID());
-        }
```
AndrewMeadows: `Avatars` will never timeout for lack of data.  They must be removed, which happens when (a) the **interface** resets its connection to the domain, or (b) the `avatar-mixer` sends a **KillAvatarPacket**.

assignment-client/src/avatars/AvatarMixerSlave.cpp
ZappoMan: Can you please elaborate on what the guarenteed flows from mixer to client is, and what the client expects?

interface/src/avatar/AvatarManager.cpp
```diff
@@ -330,44 +326,28 @@ AvatarSharedPointer AvatarManager::newSharedAvatar() {
...
-            } else if (_shouldRender) {
-                // very rare transition so we process the transaction immediately
-                const render::ScenePointer& scene = qApp->getMain3DScene();
-                render::Transaction transaction;
-                avatar->addToScene(avatar, scene, transaction);
-                if (scene) {
-                    scene->enqueueTransaction(transaction);
-                }
-            }
-        }
```
AndrewMeadows: We return to using the base class method: `AvatarHasMap::processAvatarDataPacket()`.

interface/src/avatar/Avatar.cpp
ZappoMan: shouldn't this be in `AvatarData`?

interface/src/avatar/AvatarManager.cpp
```diff
@@ -330,44 +326,28 @@ AvatarSharedPointer AvatarManager::newSharedAvatar() {
...
+AvatarSharedPointer AvatarManager::addAvatar(const QUuid& sessionUUID, const QWeakPointer<Node>& mixerWeakPointer) {
+    AvatarSharedPointer avatarData = AvatarHashMap::addAvatar(sessionUUID, mixerWeakPointer);
+    QMap<QUuid, AvatarData::Identity>::iterator itr =  _identityCache.find(sessionUUID);
+	if (itr != _identityCache.end()) {
+        auto avatar = std::static_pointer_cast<Avatar>(avatarData);
+        avatar->setIdentity(*itr);
     }
+    return avatarData;
 }
```
AndrewMeadows: We override `AvatarHashMap::addAvatar()` to restore old identity data when found.

interface/src/avatar/AvatarManager.cpp
```diff
@@ -330,44 +326,28 @@ AvatarSharedPointer AvatarManager::newSharedAvatar() {
...
     // class in this context so we can call methods that don't exist at the base class.
-    Avatar* avatar = static_cast<Avatar*>(removedAvatar.get());
-    avatar->die();
+    auto avatar = std::static_pointer_cast<Avatar>(removedAvatar);
+
+    // there is no way to request identity data from the avatar-mixer
+    // therefore whenever we remove an avatar we cache the identity in case we need it later
+    AvatarData::Identity identity;
+    avatar->getIdentity(identity);
+    _identityCache[avatar->getSessionUUID()] = identity;
```
AndrewMeadows: When removing an `Avatar` we cache its `Identity` data forever.

ZappoMan:
there are definitely going to be merge conflicts between this PR and my PR #10255 - we will need to coordinate efforts on this.

libraries/networking/src/Node.h
```diff
@@ -92,7 +89,6 @@ class Node : public NetworkPeer {
 
     QUuid _connectionSecret;
     std::unique_ptr<NodeData> _linkedData;
-    bool _isAlive;
```
AndrewMeadows: This` Node::_isAlive` stuff was unused cruft.

libraries/avatars/src/AvatarData.cpp
```diff
@@ -1497,9 +1497,6 @@ void AvatarData::processAvatarIdentity(const Identity& identity, bool& identityC
         setAvatarEntityData(identity.avatarEntityData);
         identityChanged = true;
     }
-    // flag this avatar as non-stale by updating _averageBytesReceived
-    const int BOGUS_NUM_BYTES = 1;
-    _averageBytesReceived.updateAverage(BOGUS_NUM_BYTES);
```
ZappoMan: yeah!!!

interface/src/avatar/AvatarManager.cpp
ZappoMan: when does this `_identityCache` get cleaned up? it looks like this will continue to grow over the life of the application run time... shouldn't it be cleared when you change domains?

interface/src/avatar/AvatarManager.cpp
```diff
@@ -330,44 +326,28 @@ AvatarSharedPointer AvatarManager::newSharedAvatar() {
...
-                render::Transaction transaction;
-                avatar->addToScene(avatar, scene, transaction);
-                if (scene) {
-                    scene->enqueueTransaction(transaction);
-                }
-            }
-        }
+AvatarSharedPointer AvatarManager::addAvatar(const QUuid& sessionUUID, const QWeakPointer<Node>& mixerWeakPointer) {
+    AvatarSharedPointer avatarData = AvatarHashMap::addAvatar(sessionUUID, mixerWeakPointer);
+    QMap<QUuid, AvatarData::Identity>::iterator itr =  _identityCache.find(sessionUUID);
```
ZappoMan: I don't get this... are you saying you can cause an avatar with a given session ID to be added/removed such that you will get an `addAvatar` call before the identity has arrived?

I can imagine the sessionID arriving in an avatar frame data before the identity has arrived... but I don't understand how you can have a previously sent identity...

What's the repro case that causes this?

hifi-gustavo:
Build console output for commit: 161222cf9d905af78c874f62b2d1852644e0d747 available at: https://gist.github.com/3f2a89fd200e43130cea364ec1c0d9a8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10277-161222c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10277-161222c.dmg

interface/src/avatar/AvatarManager.cpp
```diff
@@ -330,44 +326,28 @@ AvatarSharedPointer AvatarManager::newSharedAvatar() {
...
-void AvatarManager::processAvatarDataPacket(QSharedPointer<ReceivedMessage> message, SharedNodePointer sendingNode) {
-    PerformanceTimer perfTimer("receiveAvatar");
-    // enumerate over all of the avatars in this packet
-    // only add them if mixerWeakPointer points to something (meaning that mixer is still around)
-    while (message->getBytesLeftToRead()) {
-        AvatarSharedPointer avatarData = parseAvatarData(message, sendingNode);
-        if (avatarData) {
-            auto avatar = std::static_pointer_cast<Avatar>(avatarData);
-            if (avatar->isInScene()) {
-                if (!_shouldRender) {
```
ZappoMan: couple questions about removing this code...

1) this line with `_shouldRender` seems to be related to the dev/debug feature of enable/disable avatar rendering. Did you mean to remove this? 
2) Is this adding/removing from scene handled elsewhere? Can you point to how this is handled in this new code?

AndrewMeadows:
This PR is an attempt to finally fix the "invisible avatar" problem correctly.

An Avatar in the interface now never times out.  If the avatar-mixer creates an Avatar in the view it will stay there in that position forever... until it is updated or killed by the avatar-mixer, or removed **(1)** by the interface.

When an Avatar is killed or removed the interface will cache its identity data at that moment in case it needs it again later.

When an Avatar is added the interface will resurrect any identity data (if any) that it finds in cache until the avatar-mixer supplies new identity data.

**(1)** FYI: The interface will remove an Avatar when (a) moving to a new domain, or (b) signing out/in to the current domain (e.g. invoking or revoking its username+password information).

AndrewMeadows:
To test:

(1) Login to a domain that has at least one other avatar visible.
(2) Sign into the domain (or sign out, as the case may be) via **File --> Login / Sign Up**
--> The other avatar (and entire world) should vanish but all come back in a few seconds

(3) Login to a domain that has at least one other avatar visible.
(4) Have the other avatar logout.
--> The other avatar should shrink to a small size and vanish.

(5) Login to a domain that has at least one other avatar visible.
(6) Teleport to another domain at the same <X, Y, Z> coordinates and rotation.
--> The other avatar should vanish immediately on teleport (it shouldn't shrink and vanish).

There was another method that would sometimes create invisible avatars.  If this used to work for you in hifi/master then try it here:

(7) Login with low-end (i5) machine into a domain with a lot of other avatars but also with a lot of unique texture data and lots of models to load.
--> If the interface stalls for about 5 seconds or more while loading the scene it would sometimes allow other avatars to time-out and their identity data would be lost --> they would reappear but be invisible once the scene had loaded.

interface/src/avatar/AvatarManager.cpp
AndrewMeadows: Yes, it never gets cleaned up.  I was thinking: not a big memory problem for now.   We can consider doing some garbage collection when we are handling more avatars again.

We could clear the cache on changing domains yes, but the "sign out/in" path looks very much like the "go to new domain" path at the AvatarManager context and I am punting on distinguishing the difference between those two operations there.

If you're concerned about the size of the cache for long sessions of very busy users then I can do that extra work.

interface/src/avatar/AvatarManager.cpp
AndrewMeadows: The repro case here is:  sign out/in for current domain.  When you do this the interface will dismantle its connection to the domain (deleting all Avatars in the process) and then reassemble it... and the avatar-mixer will not notice this event at all, and will never send a new identity packet for all of the avatars that were in view.

interface/src/avatar/AvatarManager.cpp
AndrewMeadows: Yes, the `_shouldRender` feature is addressed whenever the setting is changed in `AvatarManager::updateAvatarRenderStatus()`.

The code above is hackery I added to handle the "sign out/in" problem back when we had Avatar timeouts, but introduced the "avatars are still visible when teleporting to new domain" bug.  Going forward we don't want this code.

assignment-client/src/avatars/AvatarMixerSlave.cpp
AndrewMeadows: The interface no longer has a timeout for an `Avatar`.  It creates one when necessary and then deletes it only when the avatar-mixer sends a KillAvatarPacket, or when moving to a new domain (or signing out/in for the current domain, which is more or less the same thing as far as the `AvatarManager` code knows).

Since there is no timeout there is no need make an exception so that a "keep alive" packet is sent from the avatar-mixer, which is what this code was doing.

interface/src/avatar/Avatar.cpp
AndrewMeadows: Yes, because it is data that the base `AvatarData` class knows about but no because the only context that needs it is where it is the derived `Avatar` class.

In other words: this new functionality is only used in the interface when it is dealing with `Avatar` instances.

interface/src/avatar/AvatarManager.cpp
AndrewMeadows: BTW, removing this code should also fix an occasional crash mode mentioned in internal issue: fogbugz-4245

interface/src/avatar/AvatarManager.cpp
AndrewMeadows: BTW, removing this code should also fix an occasional crash mode mentioned in internal issue: fogbugz-4245

interface/src/avatar/AvatarManager.cpp
birarda: There's a bug in a NodeList that's causing the session ID not to be refreshed when Interface logs in or out. I am fixing that now and will have a PR up shortly.

Then you will be able to operate on the assumption that a node changing their login status will return with a fresh session ID.

ZappoMan: this sounds like a bug in NodeList or the domain server... @birarda - can you comment?

interface/src/avatar/AvatarManager.cpp
ZappoMan: I'm all for removing unneeded code

interface/src/avatar/AvatarManager.cpp
ZappoMan: @AndrewMeadows - is this the only repro case you found? If so... then if @birarda fixes the bug in NodeList, then - is this cache needed any more?

AndrewMeadows: I believe the cache would be unnecessary if the domain sessionID were changed on sign out/in.

ZappoMan: excellent. let's get a fix from @birarda and then you can rework this PR assuming that fix.

interface/src/avatar/AvatarManager.cpp
birarda: https://github.com/highfidelity/hifi/pull/10281

interface/src/avatar/AvatarManager.cpp
birarda: https://github.com/highfidelity/hifi/pull/10281

AndrewMeadows:
Now that the sessionUUID is reset on sign out/in (#10281) the work here is mostly to remove previous work to fix the invisible avatars that would happen in that case.

hifi-gustavo:
Build console output for commit: 688bbb7f09442dfa6b369824956f4986b86c7fe5 available at: https://gist.github.com/3f2a89fd200e43130cea364ec1c0d9a8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10277-688bbb7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10277-688bbb7.dmg

hifi-gustavo:
Build console output for commit: efef25e4b1779461088260e5410aaf5e46cccf46 available at: https://gist.github.com/3f2a89fd200e43130cea364ec1c0d9a8<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10277-efef25e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10277-efef25e.dmg

ZappoMan:
@AndrewMeadows - I'm going to integrate these changes into my PR - because these two bugs too easily overlay which will cause people to fail each of our PRs without the other... I've moved your commits into this PR - https://github.com/highfidelity/hifi/pull/10255

PR #10303, opened 2017-04-27
Thread safety for sending location

davidkelly:
We store the last location, and only send an update if there's been a change.  That means we store the last location in a class variable, to use for comparison purposes.  But, in addition to the timer calling this, it happens right at startup plus whenever you change domains.  So, if calls happen to overlap, you can get into a state where you think you've not moved, but you have and never send  the update (until you move again).  Of course, this mostly happens for desktop users, and maybe more often after vsync is off (HMD->desktop).

This manifested when I was testing handshake - one person never saw the other since we look for it via the node id in the UserLocation on the backend.  There is also a PR for having the backend update the node ids a bit earlier, which helps eliminate this issue too.

### DIFF
--- a/interface/src/DiscoverabilityManager.cpp
+++ b/interface/src/DiscoverabilityManager.cpp
@@ -23,6 +23,8 @@
 #include "DiscoverabilityManager.h"
 #include "Menu.h"
 
+#include <QThread>
+
 const Discoverability::Mode DEFAULT_DISCOVERABILITY_MODE = Discoverability::Friends;
 
 DiscoverabilityManager::DiscoverabilityManager() :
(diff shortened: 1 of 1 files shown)


davidkelly:
Test Plan:

Run around in a few domains, and observe your location in metabase.  Yea, it is a bit of a PITA :(

hifi-gustavo:
Build console output for commit: 63468cc9e1483d2c92ea1547fb6b763d3d0819de available at: https://gist.github.com/4f80e43aa3946268422a4cbeb5293e14<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10303-63468cc.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10303-63468cc.dmg

davidkelly:
I did the QA (which, I know, is sort of suspect).  But you know, time constraints and all.

PR #10305, opened 2017-04-27
Load High Mips before Fbx after skybox

Atlante45:
Fix for RC-38, make the High Mips load before the following FBXs but after the skybox.


libraries/model-networking/src/model-networking/TextureCache.cpp
```diff
@@ -343,6 +344,8 @@ void NetworkTexture::makeRequest() {
     if (_ktxResourceState == PENDING_INITIAL_LOAD) {
         _ktxResourceState = LOADING_INITIAL_DATA;
 
+        setLoadPriority(this, HIGH_MIPS_LOAD_PRIORITY);
```
huffman: I don't think this will work, because at this point the request is already being attempted. We would need to do this around the time the resource is created, although I'm not sure we can do it in the ctor.

hifi-gustavo:
Build console output for commit: 2fecffea0bb4862f2ba5cd99c69a8c8937562d90 available at: https://gist.github.com/caf5d6d353bbdeefafecb93d2381c05e<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10305-2fecffe.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10305-2fecffe.dmg

PR #10312, opened 2017-04-27
add an entity file replacement via domain-server content page

birarda:
- added an entities file upload on content page in domain-server web interface
- added handling of replacement entities file to entity-server
- added clearing of domain octree information if disconnected from entity-server


hifi-gustavo:
Build console output for commit: a21a34a4a46b0c10ee87203f36f057962b846654 available at: https://gist.github.com/7a98d2b0e22be3982fb86f0fdb562fa5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10312-a21a34a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10312-a21a34a.dmg

birarda:
#### Test Plan

This feature allows a user with access to Domain Server settings to replace the `models.json.gz` in their Entity Server. The file is uploaded to the Domain Server and then sent to the Entity Server for processing.

(1) Run the Sandbox from this PR.
(2) Connect to the Sandbox from Interface and look at the content of the domain. Stay connected in Interface.
(3) Go to the domain server web interface (should be http://localhost:40100) and go to the 'Content' page.
(4) Upload a different models file using that page. Contact me if you need a models file to use.
(5) You should see your Interface client disconnect from the entity server, clear its local octree, and then load the new content from the replaced models file.
(6) You may also test the edge case of uploading a bad models file through the domain server page. The entity server should report in its log that it is refusing to perform replacement because the replacement file is invalid. It should not restart and the content in your domain should stay the same as it was before the attempted replacement.

hifi-gustavo:
Build console output for commit: 038ac745c8af5fe8ced7d1bd4e3bc093d71fff50 available at: https://gist.github.com/7a98d2b0e22be3982fb86f0fdb562fa5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10312-038ac74.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10312-038ac74.dmg

domain-server/resources/web/content/index.shtml
```diff
@@ -0,0 +1,41 @@
...
+          <h3 class="panel-title">Upload Entities File</h3>
+        </div>
+        <form id="upload-form" action="upload" enctype="multipart/form-data" method="post">
+          <div class="panel-body">
+            <p>
+              Upload an entities file (e.g.: models.json.gz) to replace the content of this domain.</br>
+              Note: <strong>Your domain's content will be replaced by the content you upload</strong>, but the backup files of your domain's content will not immediately be changed.
+            </p>
+            <p>
+              If your domain has any content that you would like to re-use at a later date, save a manual backup of your models.json.gz file, which is usually stored in <i>C:\Users\[user_name]\AppData\Roaming\High Fidelity\assignment-client</i>.
```
huffman: Should we add paths for other platforms?

domain-server/src/DomainServer.cpp
```diff
@@ -1633,13 +1633,23 @@ QString pathForAssignmentScript(const QUuid& assignmentUUID) {
     return directory.absoluteFilePath(uuidStringWithoutCurlyBraces(assignmentUUID));
 }
 
+QString DomainServer::pathForRedirect(QString path) const {
+    // make sure the passed path has a leading slash
+    if (!path.startsWith('/')) {
+        path.insert(0, '/');
+    }
+
+    return "http://" + _hostname + ":" + QString::number(_httpManager.serverPort()) + path;
```
huffman: Is there a reason to not use a relative path here?

domain-server/src/DomainServer.cpp
```diff
@@ -2560,3 +2588,20 @@ void DomainServer::setupGroupCacheRefresh() {
         _metaverseGroupCacheTimer->start(REFRESH_GROUPS_INTERVAL_MSECS);
     }
 }
+
+void DomainServer::handleOctreeFileReplacement(QByteArray octreeFile) {
+    // enumerate the nodes and find any octree type servers with active sockets
+
+    auto limitedNodeList = DependencyManager::get<LimitedNodeList>();
+    limitedNodeList->eachMatchingNode([](const SharedNodePointer& node){
```
huffman: space between `)` and `{`

domain-server/src/DomainServer.cpp
```diff
@@ -2560,3 +2588,20 @@ void DomainServer::setupGroupCacheRefresh() {
...
+        return node->getType() == NodeType::EntityServer && node->getActiveSocket();
+    }, [&octreeFile, limitedNodeList](const SharedNodePointer& octreeNode){
```
huffman: space between `)` and `{`

interface/src/Application.cpp
```diff
@@ -5273,33 +5276,8 @@ void Application::nodeKilled(SharedNodePointer node) {
...
-        });
-
-        // also clean up scene stats for that server
-        _octreeServerSceneStats.withWriteLock([&] {
-            if (_octreeServerSceneStats.find(nodeUUID) != _octreeServerSceneStats.end()) {
-                _octreeServerSceneStats.erase(nodeUUID);
-            }
-        });
+        // we lost an entity server, clear all of the domain octree details
+        clearDomainOctreeDetails();
```
huffman: `clearDomainOctreeDetails` seems to clear avatar stuff - do we want to be doing that when losing an entity node?

libraries/octree/src/OctreePersistThread.cpp
```diff
@@ -131,10 +132,46 @@ quint64 OctreePersistThread::getMostRecentBackupTimeInUsecs(const QString& forma
...
+        if (currentFile.exists()) {
+            static const QString FILENAME_TIMESTAMP_FORMAT = "yyyyMMdd-hhmmss";
+            auto backupFileName = _filename + ".backup." + QDateTime::currentDateTime().toString(FILENAME_TIMESTAMP_FORMAT);
+
+            if (currentFile.rename(backupFileName)) {
+                qDebug() << "Moved previous models file to" << backupFileName;
+            } else {
+                qWarning() << "Could not backup previous models file to" << backupFileName << "- removing replacement models file";
+
+                if (!QFile::remove(replacementFileName)) {
```
birarda: Huh ... looked for that in docs but only saw the static. Will fix.

domain-server/resources/web/content/index.shtml
birarda: Yup, will do.

domain-server/src/DomainServer.cpp
birarda: The original 302 RFC required absolute paths - when I originally added those redirects I don't think the new RFC had been accepted. I'd rather keep it absolute since we know it works and won't have to test across a bunch of different browsers.

huffman:
One other passing comment - I wish we could provide better feedback on whether the full upload was successful, but admittedly that would be a lot more work and is out of scope IMO for this particular punch list item.

birarda:
@huffman agreed - seeing the errors on the domain page would be a great improvement.

Nex-Pro:
@huffman @birarda  is this ready for QA ?

interface/src/Application.cpp
birarda: I moved the call to `clearOtherAvatars()` to a different slot.

I will check if `getMyAvatar()->setAvatarEntityDataChanged(true);` relates to the avatar mixer or the entity server.

domain-server/src/DomainServer.cpp
huffman: Huh, TIL. The issue I see with absolute paths is that you could have a setup that it won't be compatible with, for instance if you had a reverse proxy from port 80 to 40100 with port 40100 closed off to the public. (Unless I'm missing something?)

interface/src/Application.cpp
huffman: I was looking at:
```c++
    DependencyManager::get<AvatarManager>()->clearOtherAvatars();
    DependencyManager::get<AnimationCache>()->clearUnusedResources();
    DependencyManager::get<ModelCache>()->clearUnusedResources();
    DependencyManager::get<SoundCache>()->clearUnusedResources();
    DependencyManager::get<TextureCache>()->clearUnusedResources();
```

domain-server/src/DomainServer.cpp
birarda: You're right, it would not be compatible in that case. I'm going to add that to fogbugz as an issue separate from this PR.

birarda:
@Nex-Pro go for it

hifi-gustavo:
Build console output for commit: 9bc0609b1e2e6534d510354b234c330cbc72a1a5 available at: https://gist.github.com/7a98d2b0e22be3982fb86f0fdb562fa5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10312-9bc0609.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10312-9bc0609.dmg

RyanDowne:
Ran this per test plan and new content loaded up.

PR #10318, opened 2017-04-28
Pull request from LeonCrew

LeonCrew:
Create pull request from LeonCrew:

- Change to the orientation of the camera (top down)
- Teleport based on tap on the screen (or double tap?)
- Head turn based on swipe anywhere on the screen


hifi-gustavo:

PR #10328, opened 2017-04-29
fix announcement suggestions

howard-stearns:


howard-stearns:
Announcement were not display in the GOTO suggestions because our .qml doesn't use the Interface definition of XMLHttpRequest, and so doesn't frob the request into one that uses the login credentials.

Here we have the .qml use a subset of JSON-RPC to make requests that actually get sent to tablet-goto.js (which has a functioning request-via-XMLHttpRequest capability), that issues the given callbacks on response back. This allows the multiple feeds to make requests simultaneously while still processing the single channel that our event-bridge allows.

However, pal and goto both use this bridge. The intended design in the tablet is that scripts connect/disconnect the event bridge depending on which tablet app is actually in use, so this PR also fixes up pal.js as well as tablet-goto.js for this behavior

However, none of this wiring can happen until after the tablet is built, with its .qml Items, so the three Feed.qml items can no longer fillDestinations() when they are built. They have to queue a message to happen immediately after they are built. But .qml doesn't have either setTimeout or nextTick, so a Timer object is used.

howard-stearns:
test plan:

The 'Snapshot & GOTO Apps: Sharing Snaps (Including Blasting Announcements)' section of '"New Snapshot App/GoTo App/Discovery Backend" Test Plan' (https://highfidelity.atlassian.net/wiki/pages/viewpage.action?pageId=58143687)

hifi-gustavo:
Build console output for commit: b6ae0a5bde0705ab6e99ee8fbe696d7be4e5fc54 available at: https://gist.github.com/68421fd98ed62ced37fc405c56518bb7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10328-b6ae0a5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10328-b6ae0a5.dmg

howard-stearns:
tested in standup

PR #10339, opened 2017-05-01
Move "request" functionality into a Module

cfresquet:
Pulled out the "request" function from pal.js and makeUserConnection.js to a new request.js file in the modules directory.

Test:  Confirm handshake functionality still works as expected.
         Confirm connections/friends list & info still function as expected.


hifi-gustavo:
Build console output for commit: a3f637dc0ebdde001384d8df47e052b8aee1d47f available at: https://gist.github.com/440aedca788e9febb4741e156b367287<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10339-a3f637d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10339-a3f637d.dmg

scripts/system/makeUserConnection.js
```diff
@@ -11,7 +11,10 @@
 //  Distributed under the Apache License, Version 2.0.
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
-(function () { // BEGIN LOCAL_SCOPE
+
+(function() { // BEGIN LOCAL_SCOPE
+
+    var request = Script.require('../modules/request.js').request;
```
zfox23: IMO we should be using `Script.resolvePath()` for the path of `request.js`. That way, it'll be the same call to get `request.js` no matter the location of the script.

There's lots of examples of the use of `Script.resolvePath()` - check `teleport.js` for a few.

scripts/system/pal.js
```diff
@@ -14,6 +14,8 @@
 
 (function() { // BEGIN LOCAL_SCOPE
 
+    var request = Script.require('../modules/request.js').request;
```
zfox23: Same comment as above regarding using `Script.resolvePath()`

scripts/modules/request.js
```diff
@@ -0,0 +1,87 @@
...
+/* global module */
+// @module request
+//
+// This module contains the `request` module implementation
+
+// ===========================================================================================
+module.exports = {
+
+    // ------------------------------------------------------------------
+    test: function () {
```
howard-stearns: nice idea, but without any real benefit as-is, I'm thinking we should pull it out. But open to other ideas.

howard-stearns:
Also please fix up tablet-goto.js.  (Which did not use request when you started this, but does now. Things move fast around here!)

scripts/system/makeUserConnection.js
humbletim: On a related note -- the `system/modules/` directory was designed to support system-wide modules that can be unequivocally referenced by base name (instead of having to use hardcoded or lazy-resolved filepaths).

Long story short:
```javascript
var request = Script.require('request').request;
```

... could be used here just as well, because the module in question already lives inside of `system/modules/`.

cfresquet: attempting to use `Script.resolvePath()` I still needed to put in the relative path to get it to load (ie `"../modules/request.js"` was required to make it work.

`var request = Script.require('request').request;` worked perfectly and seems cleaner, so went with that.

howard-stearns:
superseeded by https://github.com/highfidelity/hifi/pull/10362

PR #10341, opened 2017-05-01
Prevent possible crash in texture buffering thread

jherico:
Transfer jobs are transient objects each holding lambdas to do two things... buffer data from a `gpu::Texture` object into local memory and then transfer that buffered data to the GPU.  Because the buffering may invoke disk IO, it's done asynchronously on a separate thread from the renderer.

However, the rendering thread may decide it needs to rebuild the queue of transfer jobs for a given texture because it's memory footprint has changed.  Because of this, in the current codebase, a TransferJob may be destroyed even while it's buffering lambda is being executed on the buffering thread.  When this happens, the lambda ceases to be valid and can potentially lead to the crash described [here](https://highfidelity.fogbugz.com/f/cases/4504/CRASH-in-gpu-Texture-accessStoredMipFace).

This PR fixes this.  It changes the queue of transfer jobs from `unique_ptr` to `shared_ptr` and creates a separate `_currentTransferJob` static member in the variable allocation support class.  While a given transfer job is active, it is stored in the `_currentTransferJob` member so that it cannot leave scope.  

## Testing

The crash related to this PR is problematic, but difficult to reproduce at will.  It is most likely to occur when first loading content that has already been cached on disk, and is more likely to occur when disk access is slow, and when textures are being frequently promoted and demoted.  The best chance for reproducing it is to start the application and loading an existing high-texture content domain like `dev-welcome` or `dev-playa` over and over.

### DIFF
--- a/libraries/gpu-gl/src/gpu/gl/GLTexture.cpp
+++ b/libraries/gpu-gl/src/gpu/gl/GLTexture.cpp
@@ -150,6 +150,7 @@ GLExternalTexture::~GLExternalTexture() {
 // Variable sized textures
 using MemoryPressureState = GLVariableAllocationSupport::MemoryPressureState;
 using WorkQueue = GLVariableAllocationSupport::WorkQueue;
+using TransferJobPointer = GLVariableAllocationSupport::TransferJobPointer;
 
 std::list<TextureWeakPointer> GLVariableAllocationSupport::_memoryManagedTextures;
 MemoryPressureState GLVariableAllocationSupport::_memoryPressureState { MemoryPressureState::Idle };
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: f81d01f179bd8eb1ff4760893e17f461e5b5c078 available at: https://gist.github.com/f7525ba2b55251891cee907088cdb8f3<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10341-f81d01f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10341-f81d01f.dmg

claudioshg:
testing

claudioshg:
### Desktop PC - Windows 10 Home - CPU 4.0GHz Intel i7 - GTX1070 - RAM 16GB - Tested with Oculus Rift

Tested: Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10341-f81d01f.exe

Not a Single Crash:
- The way I tested this PR:
- I loaded Sandbox first. Then loaded Interface from Sandbox.
- I was in Desktop Mode and I tested in dev-welcome.
- Step1: Load in to dev-welcome and wait for all content to load. 
Note: This step ^^^ was to cache all content to disk.
- Step2: Quit interface through the drop down menus.
- Step3: Restart Interface through Sandbox, and go to dev-welcome.
- Step4: I waited until the domain loaded. 
Note: This is were I would expect the crash to happen. This is where I'm loading for the first time, but with content that has already been cached on disk.
- Step5: Seeing no crash, I cleared all cache and start again from step 1.

I repeated steps 1 throug 5 over and over again for an hour. The interface never crashed for me.

PR #10358, opened 2017-05-03
Fix deadlock from minimizing

zzmp:


zzmp:
Minimizing had been changed to deactivate the display plugin (see #10142), which was the only thing sending the Application Present events. These are necessary because they call into `Application::idle`, which takes care of updating the deadlock watchdog and updating physics simulations.

This PR adds a new Idle event, which is sent by a timer when the application is minimized, that calls into idle without painting anything, so it continues updating the deadlock watchdog and updating physics simulations.

### QA Plan

- In `master`, minimizing the application will cause a crash within about thirty seconds from the deadlock watchdog.
- In this PR, minimizing should be fine. You should be able to look at the process and see that it takes less CPU/GPU when minimized, as it is both throttling and not painting anything. It will not crash.

hifi-gustavo:
Build console output for commit: ad8d3e549045846d10e0cf17048ca0eaf2fb82f7 available at: https://gist.github.com/d79675da68d77bb851e10a7ef473fe97<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10358-ad8d3e5.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10358-ad8d3e5.dmg

PR #10389, opened 2017-05-04
fix render of debug collision shapes

AndrewMeadows:
* fix debug option Developer --> Physics --> Show Collision Shapes

### DIFF
--- a/libraries/render-utils/src/Model.cpp
+++ b/libraries/render-utils/src/Model.cpp
@@ -573,7 +573,7 @@ bool Model::addToScene(const render::ScenePointer& scene,
 
     bool somethingAdded = false;
     if (_collisionGeometry) {
-        if (_collisionRenderItems.empty()) {
+        if (_collisionRenderItemsMap.empty()) {
             foreach (auto renderItem, _collisionRenderItems) {
                 auto item = scene->allocateID();
                 auto renderPayload = std::make_shared<MeshPartPayload::Payload>(renderItem);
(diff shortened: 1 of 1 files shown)


AndrewMeadows:
To test:

(1) Create a model entity but select the **Basic** or **Good** collision shape options (not **Exact**).
(2) Enable the debug menu option:  **Developer --> Physics --> Draw Collision Shapes**
--> in hifi/master the model disappears (bug!)
--> in this PR the model should render as a multi-color (or just blue, it it has only one part) object that shows its collision surface
(3) Disable the debug menu option: **Developer --> Physics --> Draw Collision Shapes**
--> the object should regain its normal appearance

hifi-gustavo:
Build console output for commit: f11d6eff92e0827b91a13532ae5dbdbf29db5526 available at: https://gist.github.com/d915eec252e3e1f13fb9aa26cbbf23c5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10389-f11d6ef.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10389-f11d6ef.dmg

OmegaHeron:
Tested Windows 10
QA Pass

Compared against latest master.

Placed object giving it basic shape *Pass*
Placed object giving it good shape *Pass*

In all tests model displayed physics outline when it should, model was never made invisible.
Also looked at it with all other shape variants just to be thorough.  Works as expected (i.e. shows a shape for all but exact and exact simply shows model vs turning it invisible or any other unexpected result).

PR #9707, opened 2017-02-18
MacOS: Fix for audio, toolbar and tablet.

hyperlogic:
There was a deadlock on the audio thread, This would prevent audio from playing, but it would also prevent defaultScripts from running properly.  Which affected pretty much everything, including the toolbar and tablet buttons.  Specifically selectAudioDevice.js would block when it called AudioDevice.getInputDevices().

PR #10394, opened 2017-05-05
Guard against local audio crashes

zzmp:
- Remove lock from audio device read
- Prevent overflow of local injectors buffer when switching devices
- Prevent data race when audio injectors are stopped during read

### DIFF
--- a/libraries/audio-client/src/AudioClient.cpp
+++ b/libraries/audio-client/src/AudioClient.cpp
@@ -1097,25 +1097,25 @@ void AudioClient::handleRecordedAudioInput(const QByteArray& audio) {
 }
 
 void AudioClient::prepareLocalAudioInjectors() {
-    if (_outputPeriod == 0) {
-        return;
-    }
-
-    int bufferCapacity = _localInjectorsStream.getSampleCapacity();
-    if (_localToOutputResampler) {
-        // avoid overwriting the buffer,
-        // instead of failing on writes because the buffer is used as a lock-free pipe
-        bufferCapacity -=
-            _localToOutputResampler->getMaxOutput(AudioConstants::NETWORK_FRAME_SAMPLES_PER_CHANNEL) *
-            AudioConstants::STEREO;
-        bufferCapacity += 1;
-    }
-
     int samplesNeeded = std::numeric_limits<int>::max();
     while (samplesNeeded > 0) {
-        // lock for every write to avoid locking out the device callback
-        // this lock is intentional - the buffer is only lock-free in its use in the device callback
+        // unlock between every write to allow device switching
         Lock lock(_localAudioMutex);
+
+        // in case of a device switch, consider bufferCapacity volatile across iterations
+        if (_outputPeriod == 0) {
+            return;
+        }
+
+        int bufferCapacity = _localInjectorsStream.getSampleCapacity();
+        if (_localToOutputResampler) {
+            // avoid overwriting the buffer,
+            // instead of failing on writes because the buffer is used as a lock-free pipe
+            bufferCapacity -=
+                _localToOutputResampler->getMaxOutput(AudioConstants::NETWORK_FRAME_SAMPLES_PER_CHANNEL) *
+                AudioConstants::STEREO;
+            bufferCapacity += 1;
+        }
 
         samplesNeeded = bufferCapacity - _localSamplesAvailable.load(std::memory_order_relaxed);
         if (samplesNeeded <= 0) {
--- a/libraries/audio-client/src/AudioClient.cpp
+++ b/libraries/audio-client/src/AudioClient.cpp
@@ -1115,7 +1115,7 @@ void AudioClient::prepareLocalAudioInjectors() {
     while (samplesNeeded > 0) {
         // lock for every write to avoid locking out the device callback
         // this lock is intentional - the buffer is only lock-free in its use in the device callback
-        RecursiveLock lock(_localAudioMutex);
+        Lock lock(_localAudioMutex);
 
         samplesNeeded = bufferCapacity - _localSamplesAvailable.load(std::memory_order_relaxed);
         if (samplesNeeded <= 0) {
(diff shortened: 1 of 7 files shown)


zzmp:
### Code Reviewer

Each commit fixes a specific issue. I could not reproduce, only speculate, so I will justify each commit.
Feel free to request them in separate PRs.

#### e882e7b648: Removes a lock in the audio device read callback

There should not be any unnecessary locks in the callback, as it delays audio.

The lock was also changed from recursive. This is no longer necessary, as the only recursion was caused when switching devices: calling start() on the new device could induce a callback before the lock was released (see the comments on #9707).

(@hyperlogic added this before, so I'm pinging him here if he wants to review as well.)

#### cdf6c332a6: Prevents overflow of the local injectors buffer when switching devices

In `AudioClient::prepareLocalAudioInjectors`, `bufferCapacity` was computed without a lock, even though it used guarded members. This puts that computation under a lock.

Because the members can be changed when switching devices, and the device switch can happen between iterations in `prepareLocalAudioInjectors` because the lock is released (to prevent blocking), this is recomputed every iteration.

(It is necessary to continue unlocking between each iteration so that, in the case of a device switch, a local injector is not buffered out to the old device before switching devices. I.e., unlocking ensures that no more than one frame of an injector will be lost to the old device before continuing to buffer it into the new device.)

#### 48af83a960: Prevents a datarace between AudioInjector and its local buffer

[Bug 4631](https://highfidelity.fogbugz.com/f/cases/4631/CRASH-when-changing-audio-settings) details a crash on [this line](https://github.com/highfidelity/hifi/blob/da1d93b86c712ff2d9f7c6c2814ee135d5cd0688/libraries/audio-client/src/AudioClient.cpp#L1180). AFAICT memset could not fail, but I think `getLocalBuffer` can fail in a datarace.
If the buffer is stopped and deleted, because it is on a separate thread, it is possible for the deletion to occur between checking for the buffer and reading it out.

To fix this, I move it onto the `AudioClient::_localAudioThread`. Other code changes followed.

libraries/audio-client/src/AudioClient.cpp
```diff
@@ -1097,26 +1097,26 @@ void AudioClient::handleRecordedAudioInput(const QByteArray& audio) {
...
+        }
+
+        int bufferCapacity = _localInjectorsStream.getSampleCapacity();
+        if (_localToOutputResampler) {
+            // avoid overwriting the buffer,
+            // instead of failing on writes because the buffer is used as a lock-free pipe
+            bufferCapacity -=
+                _localToOutputResampler->getMaxOutput(AudioConstants::NETWORK_FRAME_SAMPLES_PER_CHANNEL) *
+                AudioConstants::STEREO;
+            bufferCapacity += 1;
```
kencooke: The +1 here still looks a bit fishy to me. Is this to compute the min instead of max output from the resampler? Maybe a comment can clarify the intent.

hifi-gustavo:
Build console output for commit: 48af83a960bd231d2574a7de6c5a91dbc2f7c562 available at: https://gist.github.com/092d6f0b61b567d4d2bce3c630de4905<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10394-48af83a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10394-48af83a.dmg

libraries/audio-client/src/AudioClient.cpp
```diff
@@ -1168,16 +1168,18 @@ bool AudioClient::mixLocalAudioInjectors(float* mixBuffer) {
     memset(mixBuffer, 0, AudioConstants::NETWORK_FRAME_SAMPLES_STEREO * sizeof(float));
 
     for (AudioInjector* injector : _activeLocalAudioInjectors) {
-        if (injector->getLocalBuffer()) {
+        // the lock guarantees that injectorBuffer, if found, is invariant
+        AudioInjectorLocalBuffer* injectorBuffer;
+        if ((injectorBuffer = injector->getLocalBuffer())) {
```
kencooke: This assignment-as-test can look like an = instead of == bug.
It would be more clear as
```
AudioInjectorLocalBuffer* injectorBuffer = injector->getLocalBuffer()); 
if ((injectorBuffer) {

File: libraries/audio-client/src/AudioClient.cpp [hunk 3]
```diff
@@ -1317,23 +1319,25 @@ void AudioClient::setIsStereoInput(bool isStereoInput) {
...
 
 bool AudioClient::outputLocalInjector(AudioInjector* injector) {
-    Lock lock(_injectorsMutex);
-    if (injector->getLocalBuffer() && _audioInput ) {
-        // just add it to the vector of active local injectors, if 
-        // not already there.
-        // Since this is invoked with invokeMethod, there _should_ be
-        // no reason to lock access to the vector of injectors.
+    AudioInjectorLocalBuffer* injectorBuffer;
+    if ((injectorBuffer = injector->getLocalBuffer())) {
```
kencooke: same as above

zzmp:
### QA Plan

These fix issues that are either speculative or very hard to reproduce.

Do a general smoke test.

Play with local audio injectors. They should operate normally.
Switch audio devices. This should work normally.
Make sure to switch devices while local audio injectors are going.

Audio should not appear any different than with `master`.

hifi-gustavo:
Build console output for commit: 7db404eba01c7af86a400f4384523775e54b27cc available at: https://gist.github.com/092d6f0b61b567d4d2bce3c630de4905<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10394-7db404e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10394-7db404e.dmg

Nex-Pro:
Testing this

Nex-Pro:
### PASS
No apparent change from master.

Comments below apply to master also.

However worthy of note, I feel, is the fact that there is some crackle when files are played back. This crackle is not present in other playback apps and is not co-incident with the loop start/end of the track. 

I used a stopwatch to time the occurrence of crackle in this and master and the crackle I hear does not come at a regular interval. ( loop was 13 seconds )  

The crackle is not related to avatar movement. 

Nearby audio injectors, that are not in view, do not start playing unless the avatar rotates to get them in view. This is a particular problem with ambient sounds delivered via entities.

PR #10401, opened 2017-05-06
Fix sped up animations

jherico:
The recent refactor of avatar rendering into a library had one parent method being called multiple times per frame.  

## Testing

Walk around an environment with an avatar.  In master the walk animation will be sped up.  In this build it should be correctly paced.

### DIFF
--- a/interface/src/avatar/MySkeletonModel.cpp
+++ b/interface/src/avatar/MySkeletonModel.cpp
@@ -140,9 +140,6 @@ void MySkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
     auto orientation = myAvatar->getLocalOrientation();
     _rig->computeMotionAnimationState(deltaTime, position, velocity, orientation, ccState);
 
-    // evaluate AnimGraph animation and update jointStates.
-    Model::updateRig(deltaTime, parentTransform);
-
     Rig::EyeParameters eyeParams;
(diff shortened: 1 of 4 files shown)


interface/src/avatar/MySkeletonModel.cpp
```diff
@@ -140,9 +140,6 @@ void MySkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
     auto orientation = myAvatar->getLocalOrientation();
     _rig->computeMotionAnimationState(deltaTime, position, velocity, orientation, ccState);
 
-    // evaluate AnimGraph animation and update jointStates.
-    Model::updateRig(deltaTime, parentTransform);
```
jherico: This is the problematic extra call to the parent class.

hifi-gustavo:
Build console output for commit: b63015eaea681b8a97c594d486be8cd7e2e3f3db available at: https://gist.github.com/e54068deed0aefb2b9ded97eeef0583c<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10401-b63015e.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10401-b63015e.dmg

Nex-Pro:
## PASS
Walk cycle now working correctly

PR #10413, opened 2017-05-08
Fix lookAt eye snapping

humbletim:
This PR fixes a regression in master where lookAt eye snapping was no longer updating as of side-effect of some of the Avatar refactoring changes.

#### Testing:
* Launch two instances of this PR's Interface with avatar models where the eyeballs are visible (eg: "Photo-Real" Male or Female from the marketplace).
* Position your avatar in front of the other avatar and look/gaze around in some wide arcs.
* Verify that when you are approximately making eye contact (ie: +/- a heads-worth or two), both your and their eyes "snap" into making direct eye contact.

### DIFF
--- a/interface/src/avatar/MySkeletonModel.cpp
+++ b/interface/src/avatar/MySkeletonModel.cpp
@@ -38,6 +38,13 @@ void MySkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
 
     // make sure lookAt is not too close to face (avoid crosseyes)
     glm::vec3 lookAt = _owningAvatar->isMyAvatar() ?  head->getLookAtPosition() : head->getCorrectedLookAtPosition();
+    glm::vec3 focusOffset = lookAt - _owningAvatar->getHead()->getEyePosition();
+    float focusDistance = glm::length(focusOffset);
+    const float MIN_LOOK_AT_FOCUS_DISTANCE = 1.0f;
+    if (focusDistance < MIN_LOOK_AT_FOCUS_DISTANCE && focusDistance > EPSILON) {
+        lookAt = _owningAvatar->getHead()->getEyePosition() + (MIN_LOOK_AT_FOCUS_DISTANCE / focusDistance) * focusOffset;
+    }
+
     MyAvatar* myAvatar = static_cast<MyAvatar*>(_owningAvatar);
 
     Rig::HeadParameters headParams;
--- a/interface/src/avatar/MySkeletonModel.cpp
+++ b/interface/src/avatar/MySkeletonModel.cpp
@@ -147,6 +147,9 @@ void MySkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
     auto orientation = myAvatar->getLocalOrientation();
     _rig->computeMotionAnimationState(deltaTime, position, velocity, orientation, ccState);
 
+    // evaluate AnimGraph animation and update jointStates.
+    Model::updateRig(deltaTime, parentTransform);
+
     Rig::EyeParameters eyeParams;
     eyeParams.eyeLookAt = lookAt;
     eyeParams.eyeSaccade = head->getSaccade();
@@ -156,8 +159,5 @@ void MySkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
     eyeParams.rightEyeJointIndex = geometry.rightEyeJointIndex;
 
     _rig->updateFromEyeParameters(eyeParams);
-
-    // evaluate AnimGraph animation and update jointStates.
-    Parent::updateRig(deltaTime, parentTransform);
 }
 
--- a/interface/src/avatar/MySkeletonModel.cpp
+++ b/interface/src/avatar/MySkeletonModel.cpp
@@ -37,7 +37,7 @@ void MySkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
     Head* head = _owningAvatar->getHead();
 
     // make sure lookAt is not too close to face (avoid crosseyes)
-    glm::vec3 lookAt = _owningAvatar->isMyAvatar() ?  head->getLookAtPosition() : head->getCorrectedLookAtPosition();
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: ef0e32643d68ec4e76ce2b5fd85b260eaf53bbb1 available at: https://gist.github.com/6a93fd7a69e4970873a1be60503ba101<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10413-ef0e326.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10413-ef0e326.dmg

libraries/avatars-renderer/src/avatars-renderer/SkeletonModel.cpp
```diff
@@ -118,9 +118,6 @@ void SkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
 
         _rig->updateFromEyeParameters(eyeParams);
      }
-
-    // evaluate AnimGraph animation and update jointStates.
-    Parent::updateRig(deltaTime, parentTransform);
 }
```
AndrewMeadows: @humbletim, looking at this code I see a problems that were there before you started, but they need to be fixed:

(1) `MySkeletonModel::updateRig()` should NOT call `Parent::updateRig()` (which maps to `SkeletonModel::updateRig()`) because there is no code there that applies to the MyAvatar instance.  Instead `MySkeletonModel::updateRig()` should call `Model::updateRig()` because it needs the two side effects on that method.

(2) Meanwhile, `SkeletonModel::updateRig()` should NOT call `Parent::updateRig()` as you properly did (because OtherAvatar's don't actually animate, they just set the joints according to the data coming from the avatar-mixer), however there is one side effect of `Model::updateRig()` that it does need and that is the `_needsUpdateClusterMatrices = true;` line, so just add that line where you removed `Parent::updateRig()`.

(3) Optional: rather than wrap all of `SkeletonModel::updateRig()` implementation in an `if (!_owningAvatar->isMyAvagar())` logic block, just `assert(!_owningAvatar->isMyAvatar())` at the top and indent the rest of the code as normal without the `if` check.

AndrewMeadows:
I'm marking this as **CRITICAL_FIX** because Philip cares about the eye lookAt behavior and I'm guessing he would like to get this in sooner.

interface/src/avatar/MySkeletonModel.cpp
```diff
@@ -38,6 +38,13 @@ void MySkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
 
     // make sure lookAt is not too close to face (avoid crosseyes)
     glm::vec3 lookAt = _owningAvatar->isMyAvatar() ?  head->getLookAtPosition() : head->getCorrectedLookAtPosition();
```
jherico: you can eliminate this `? :` section... `MySkeletonModel` is only ever attached to `MyAvatar`

libraries/avatars-renderer/src/avatars-renderer/SkeletonModel.cpp
```diff
@@ -86,41 +87,36 @@ void SkeletonModel::updateRig(float deltaTime, glm::mat4 parentTransform) {
         lookAt = _owningAvatar->getHead()->getEyePosition() + (MIN_LOOK_AT_FOCUS_DISTANCE / focusDistance) * focusOffset;
     }
 
-    if (!_owningAvatar->isMyAvatar()) {
```
jherico: this method will get called by the derived `MySkeletonModel` so if there's work you can avoid for the `!isMyAvatar` case, you should do so.

hifi-gustavo:
Build console output for commit: 68ca0d921e9b4e7d16f5078e45684f0a74e1c059 available at: https://gist.github.com/6a93fd7a69e4970873a1be60503ba101<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10413-68ca0d9.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10413-68ca0d9.dmg

hifi-gustavo:
Build console output for commit: 29b1ac3572b5f7d1d4f13e856a2ffdc2e011de11 available at: https://gist.github.com/6a93fd7a69e4970873a1be60503ba101<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10413-29b1ac3.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10413-29b1ac3.dmg

libraries/avatars-renderer/src/avatars-renderer/SkeletonModel.cpp
humbletim: ... I've applied your suggested `? :` optimization to this code too, but were there any other operations you think can be avoided in this `!isMyAvatar` code path?

Note that with the changes @AndrewMeadows just requested, my understanding is that  `MySkeletonModel::updateRig` and `SkeletonModel::updateRig` are now "independent" code paths.  (Put another way -- `Model::simulate` virtually calls `MySkeletonModel::updateRig` _XOR_ `SkeletonModel::updateRig`).

AndrewMeadows: Correct.  `MySkeletonModel::updateRig()` now calls `Model::updateRig()` directly and not `SkeletonModel::updateRig()`, which is how it should be.

`MyAvatar` and `OtherAvatar` are doing mostly different things in `updateRig()` so we'll keep them separate.

OmegaHeron:
Tested Windows 10/Ubuntu 16.04
QA  Pass

Machine 1 - This PR Interface/stack Windows 10
Machine 2 - This PR Interface Ubuntu 16.04

Used show eye look at vector display.

Used my own Mixamo based AV and photo real male from Marketplace.

Lookat snapped to opposite avatar when in range, as expected.  Compared to current master (6472) where it never achieves lock.

PR #10415, opened 2017-05-09
Fix avatars' mouths not moving properly

ctrlaltdavid:


ctrlaltdavid:
Testing

Using two PCs, have two avatars with blend shapes enabled e.g., Being of Light) stand in front of each other. 
Make sure that Interface has focus on PC 2.
On PC 1, talk into the microphone such that the avatar 1's facial animation is activated.
Check that avatar 1's facial animation on PCs 1 and 2 look very much the same.

Repeat the test with Interface not having focus on PC 2. The result should be the same.

hifi-gustavo:
Build console output for commit: 4b4878cabd3b453a5eb1a7a6c3a5ef2a8c1f46d0 available at: https://gist.github.com/d3c3651d3375c3f69e8cafc602159ebb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10415-4b4878c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10415-4b4878c.dmg

libraries/avatars-renderer/src/avatars-renderer/Avatar.cpp
```diff
@@ -387,7 +388,12 @@ void Avatar::simulate(float deltaTime, bool inView) {
             Head* head = getHead();
             head->setPosition(headPosition);
             head->setScale(getUniformScale());
-            head->simulate(deltaTime);
+
+            qint64 now = usecTimestampNow();
+            float headDeltaTime = (float)(now - _lastHeadUpdate) / (float)USECS_PER_SECOND;
+            _lastHeadUpdate = now;
+
+            head->simulate(headDeltaTime);
```
AndrewMeadows: It does not make sense to me that we would need to track `headDeltaTime` independently of `deltaTime` passed in as an argument.  If the time periods differ significantly enough to break mouth blendshapes then something is messed up further down.

anshuman64:
QA Pass. However, let's not merge until Andrew's comment has been addressed as it seems important.

libraries/avatars-renderer/src/avatars-renderer/Avatar.cpp
ctrlaltdavid: `head->simulate(headDeltaTime)` uses the time delta to calculate average loudness and long term average loudness, assuming that the time delta is how long it's been since that method was last called.

In an extreme case - my dev PC calling the enclosing `Avatar::simulate()` at 500Hz and viewing an avatar running on my tests PC which is simulating at 20Hz - the if statement guarding `head->simulate(headDeltaTime)` means that it is called at only 20Hz.
And if the enclosing loops deltaTime of 1/500 is passed in the net effect is that, for example, jaw opening and closing runs at 20/500 = 1/25 the proper speed, i.e., it is severely slowed down. Mouth moves but very little, and closing jaw with now audio takes forever to decay to zero.

AndrewMeadows: So you're saying that `Avatar::simulate()` is being called with too short of a `deltaTime` and the code that drives the mouth movement is NOT invariant under different time steps.  I would conclude that there is a bug in the code that computes the mouth movement and now that you mention it... I think I know exactly where that is.

I'm sorry, but  I'm going to mark this PR for DO_NOT_MERGE but will keep it open for discussion until I have my own version of the fix posted.

ctrlaltdavid: Not exactly. Note how the call to `head->simulate()` is in one branch of the if statement but not in the other, so it is not necessarily called each time that `Avatar::simulate()` is called and so the deltaTime value passed into `Avatar::simulate()` won't be correct for `head->simulate()`.

AndrewMeadows: Ok you're right ctrlaltdavid and I was wrong. Your change does fix it, but I don't think adding another special tracked `deltaTime` is not the right way to do it.  I've got an alternate change that fixes it higher up.

AndrewMeadows:
Closing this in favor of #10424

PR #10422, opened 2017-05-09
Fix possible corruption of ktx cache files

huffman:
When creating a file like this:

  1. Create file
  2. Write file
  3. Close file

there is an opening before the file is written and closed where, if the
application were to crash (or not finish for any other reason), the file
would be left in an undefined state. Instead, this change updates it to
do this:

  1. Create temp file
  2. Write temp file
  3. Close temp file
  4. Move temp file to final path

This ensures that at step 3 we have a valid file, before we rename it.
We should never end up with a final file in an undefined state, but it is
possible to end up with a temp file that is an undefined state if we,
for instance, crash during step 2.

### DIFF
--- a/libraries/networking/src/FileCache.cpp
+++ b/libraries/networking/src/FileCache.cpp
@@ -112,9 +112,8 @@ FilePointer FileCache::writeFile(const char* data, File::Metadata&& metadata) {
     }
 
     QSaveFile saveFile(QString::fromStdString(filepath));
-    saveFile.open(QIODevice::WriteOnly);
-    saveFile.write(data, metadata.length);
-    if (saveFile.commit()) {
+    if (saveFile.open(QIODevice::WriteOnly) && saveFile.write(data, metadata.length) == metadata.length
+        && saveFile.commit()) {
         file = addFile(std::move(metadata), filepath);
     } else {
         qCWarning(file_cache, "[%s] Failed to write %s", _dirname.c_str(), metadata.key.c_str());
--- a/libraries/networking/src/FileCache.cpp
+++ b/libraries/networking/src/FileCache.cpp
@@ -17,6 +17,7 @@
 #include <unordered_set>
 
 #include <QDir>
+#include <QSaveFile>
 
 #include <PathUtils.h>
 
@@ -110,19 +111,13 @@ FilePointer FileCache::writeFile(const char* data, File::Metadata&& metadata) {
         return file;
     }
 
-    // write the data to a temporary file
-    std::string tmpFilepath = filepath + "_tmp";
-    FILE* saveFile = fopen(tmpFilepath.c_str(), "wb");
-    if (saveFile != nullptr && fwrite(data, metadata.length, 1, saveFile) && fclose(saveFile) == 0) {
-        int result = rename(tmpFilepath.c_str(), filepath.c_str());
-        if (result == 0) {
-            file = addFile(std::move(metadata), filepath);
-        } else {
-            qCWarning(file_cache, "[%s] Failed to rename %s to %s (%s)", tmpFilepath.c_str(), filepath.c_str(), strerror(errno));
-        }
+    QSaveFile saveFile(QString::fromStdString(filepath));
+    saveFile.open(QIODevice::WriteOnly);
+    saveFile.write(data, metadata.length);
+    if (saveFile.commit()) {
+        file = addFile(std::move(metadata), filepath);
     } else {
-        qCWarning(file_cache, "[%s] Failed to write %s (%s)", _dirname.c_str(), metadata.key.c_str(), strerror(errno));
-        errno = 0;
+        qCWarning(file_cache, "[%s] Failed to write %s", _dirname.c_str(), metadata.key.c_str());
     }
 
     return file;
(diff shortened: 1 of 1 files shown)


libraries/networking/src/FileCache.cpp
```diff
@@ -110,10 +110,16 @@ FilePointer FileCache::writeFile(const char* data, File::Metadata&& metadata) {
         return file;
     }
 
-    // write the new file
-    FILE* saveFile = fopen(filepath.c_str(), "wb");
+    // write the data to a temporary file
+    std::string tmpFilepath = filepath + "_tmp";
```
jherico: You may want to consider using [`QSaveFile`](http://doc.qt.io/qt-5/qsavefile.html) which does this work automatically.

hifi-gustavo:
Build console output for commit: d86071d783630393578046e389850953501df3ac available at: https://gist.github.com/6431a89fbe1ed5458b5b2d87c63833c5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10422-d86071d.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10422-d86071d.dmg

OmegaHeron:
This seems, I stress seems, to be ok.  Tested Windows 10.  Loaded dev-playa and dev-welcome from an empty cache (twice) and from populated cache (twice).  Everything looks ok - I didn't really see any difference between this and master for scenes unless it was something subtle I'd never notice.

Checked log and saw nothing suspicious, but, considering the mountain of log spew... hard to be sure something didn't fly by that's important.

libraries/networking/src/FileCache.cpp
huffman: Thanks for the suggestion. I checked it out and am using it here now.

huffman:
@OmegaHeron Thanks Omega. Yep, the hope is that there is no difference. I just made one more change that should also behave the same.

hifi-gustavo:
Build console output for commit: 542001b14f7e6b9ba3ab5dcac329a454c1308b0f available at: https://gist.github.com/6431a89fbe1ed5458b5b2d87c63833c5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10422-542001b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10422-542001b.dmg

libraries/networking/src/FileCache.cpp
```diff
@@ -110,13 +111,13 @@ FilePointer FileCache::writeFile(const char* data, File::Metadata&& metadata) {
         return file;
     }
 
-    // write the new file
-    FILE* saveFile = fopen(filepath.c_str(), "wb");
-    if (saveFile != nullptr && fwrite(data, metadata.length, 1, saveFile) && fclose(saveFile) == 0) {
+    QSaveFile saveFile(QString::fromStdString(filepath));
+    saveFile.open(QIODevice::WriteOnly);
+    saveFile.write(data, metadata.length);
```
jherico: technically, when writing you should check the return value to see how many bytes were written.  However, it looks like commit will only return true if all writes were successful.  While `fwrite` docs specifically say that it will only return a number different from the length requested if there's an error, I can't find any such statement for `QIODevice::write`.  

it might be safer to combine the write and commit in the conditional as  `if (metadata.length == saveFile.write(data, metadata.length) && saveFile.commit())`, but I'm not certain it would be critical.

huffman: I ended up omitting the check because commit will fail like you said, but I agree that checking couldn't hurt. I am adding it to the if.

hifi-gustavo:
Build console output for commit: f6c487c8ebb96d07e8c9416f7fa8b0009849425d available at: https://gist.github.com/6431a89fbe1ed5458b5b2d87c63833c5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10422-f6c487c.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10422-f6c487c.dmg

hifi-gustavo:
Build console output for commit: 6e4c5d1ab4bb336d9fbd386602a48b51e221796e available at: https://gist.github.com/6431a89fbe1ed5458b5b2d87c63833c5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10422-6e4c5d1.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10422-6e4c5d1.dmg

AndrewMeadows:
The ubuntu build failed for new warning: **comparison between signed and unsigned integer expressions [-Wsign-compare]** in `FileCache.cpp` line 115

huffman:
Thanks @AndrewMeadows, fixing

hifi-gustavo:
Build console output for commit: d92bdbf0ed5c5a33a9a75b40ad536465b1e6fba1 available at: https://gist.github.com/6431a89fbe1ed5458b5b2d87c63833c5<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10422-d92bdbf.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10422-d92bdbf.dmg

PR #10424, opened 2017-05-09
Fix avatar mouth movement when driven by audio loudness

AndrewMeadows:


AndrewMeadows:
This is an alternate fix to #10415.

To test:

Login UserA and UserB in a domain and stand facing each other.
UserA: go into mirror mode while UserB watches UserA's face.
UserA: speak into the microphone
--> the mouth movement seen for UserA should more or less match cadence on both screens.

Swap places such that UserB is in mirror mode and UserA watches UserB and repeat the test with UserB speaking into their microphone.

anshuman64:
Will QA once it builds

hifi-gustavo:
Build console output for commit: 2be6d19dfe5399438611780b1d6cd6995c849abf available at: https://gist.github.com/01908c89403d0befa75a4f4fdeeecedd<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10424-2be6d19.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10424-2be6d19.dmg

anshuman64:
So, I compared the three versions of this:
1. No fix on blend shapes = looks as if the person is taaaalkiiingg veeerryyy slooooowwly
2. ctrlaltdavid's fix = blend shapes move quickly and are expressive as one would expect
3. Andrew's fix = seems somewhere in between 1 & 2. I am not satisfied with what I am seeing.

Not sure if I'm just crazy, but it seems like ctrlaltdavid's fix performs better. Can someone confirm?

CainFoool:
anshuman64: Can confirm, tested both. David's works much better.

PR #10429, opened 2017-05-09
Close audio device when switching outputs

zzmp:


zzmp:
This is in response to [a crash when switching audio devices](https://highfidelity.fogbugz.com/f/cases/3983/).

I could not find nor reproduce the cause, but think it may be due to asynchronous closing of the audio output. The device is never closed. I don't know if that allows this or not, but to mitigate I am closing the device when switching between outputs.

### QA Plan

Audio smoke test. Make sure to switch outputs a few times.

kencooke:
Looks reasonable.

hifi-gustavo:
Build console output for commit: 5a11911e1d49bfa7e6b23ea3dd555b1f5c0c2aab available at: https://gist.github.com/5572e8e979aebcd34850553dde0f38fb<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10429-5a11911.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10429-5a11911.dmg

PhilipRosedale:
Testing;   Crashed few secs after switching audio output from rift to speakers.  Send bugsplat with "PR #10429 Crash on looking around few seconds after switching audio output."

Then, I caused a second crash on shutdown - this one without the crash reporter - by changing audio devices while banging a balloon against the xylphone keys while switching the audio device and shortly after, shutting down.  So basically there were a bunch of injectors playing, after I switched audio device, when I shut down.

zzmp:
[The bugsplat shows that the bug was physics related](https://www.bugsplat.com/individualCrash/?id=11679&database=interface_alpha).

@PhilipRosedale, when did the bugsplatless crash occur? While moving? When changing injectors? On shutdown? I'm afraid you may just be thoroughly testing and running into other undocumented bugs.

kencooke:
The bugsplat does appear to be unrelated physics crash in ObjectMotionState::handleEasyChanges()

I can't seem to repro the shutdown crash. My process was:
- start 32 simultaneous Ambisonic injectors
- marvel at how good our peak-limiter is
- switch audio devices 5 or 6 times
- quickly shutdown

For the no-bugsplat crash, the tail of your hifi log might help.

I believe one reason that some crashes fail to send a crash-report, is because we run the crash handler on the main thread (in the same process) as the crash occurs. The crash-handler does some heavy operations (create minidump, upload to Bugsplat) that can fail if the crash corrupts the heap/stack. Running the crash-handler as a server (in a separate process) would improve this, but is a big project.

Nex-Pro:
testing this shortly

Nex-Pro:
Tried hard to make this fall over by adding multiple sound sources and kept changing outputs (4)
and inputs. (3)

Couldn't get a fail. Zero crashes.

PR #10432, opened 2017-05-10
make nondynamic entities temporarily dynamic when far grabbing

MrRoboman:
Testing Notes
Ensure handControllerGrab.js is loaded properly.

In all cases entity's dynamic and grabbable properties should be unchanged upon release.
Far grab a grabbable, nondynamic entity. Entity becomes dynamic until release. 
Far grab with two hands and rotate.
Far grab, pull to near grab.
Rapidly far grab and release.

Far grab a dynamic, non grabbable entity.
Far grab a non dynamic, non grabbable entity.

Parent the different entity types to each other and try grabbing them.

### DIFF
--- a/scripts/system/controllers/handControllerGrab.js
+++ b/scripts/system/controllers/handControllerGrab.js
@@ -1036,8 +1036,8 @@ function getControllerJointIndex(hand) {
                                       "_CONTROLLER_RIGHTHAND" :
                                       "_CONTROLLER_LEFTHAND");
     }
-    
-    return MyAvatar.getJointIndex("Head");    
+
+    return MyAvatar.getJointIndex("Head");
 }
 
 // global EquipHotspotBuddy instance
@@ -1092,6 +1092,7 @@ function MyController(hand) {
     this.grabbedOverlay = null;
     this.state = STATE_OFF;
     this.pointer = null; // entity-id of line object
+    this.grabbedThingIsTempDynamic = false;
 
     this.triggerValue = 0; // rolling average of trigger value
     this.triggerClicked = false;
@@ -1331,7 +1332,7 @@ function MyController(hand) {
         if (this.stylus) {
             return;
         }
-        
+
         var stylusProperties = {
             name: "stylus",
             url: Script.resourcesPath() + "meshes/tablet-stylus-fat.fbx",
@@ -2132,7 +2133,7 @@ function MyController(hand) {
             return null;
         }
     };
-        
+
     this.chooseNearEquipHotspotsForFarToNearEquip = function(candidateEntities, distance) {
         var equippableHotspots = flatten(candidateEntities.map(function(entityID) {
                 return _this.collectEquipHotspots(entityID);
@@ -2289,7 +2290,7 @@ function MyController(hand) {
                 return;
             }
         }
-        
+
         if (isInEditMode()) {
             this.searchIndicatorOn(rayPickInfo.searchRay);
             if (this.triggerSmoothedGrab()) {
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: ac880f6d368e4fa52c7e0fc2e8d06c05eaded698 available at: https://gist.github.com/f2c1350512ea75f47aecedbf95973345<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10432-ac880f6.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10432-ac880f6.dmg

scripts/system/controllers/handControllerGrab.js
```diff
@@ -2314,7 +2315,20 @@ function MyController(hand) {
...
+            var isDynamic = props.dynamic;
+            var makeTempDynamic = isGrabbable && !isDynamic;
+
+            if(makeTempDynamic) {
+                this.grabbedThingIsTempDynamic = true;
+                Entities.editEntity(entity,{
+                    dynamic: true
+                });
+                entityPropertiesCache._updateCacheEntry(entity);
+            }
```
AndrewMeadows: I don't think this will succeed in general.  Many objects use the "Exact" collision shape (aka `shapeType : "static_mesh"`) which creates a `btBvhTriangleMeshShape` under the hood for which dynamic motion is unsupported -- such objects must be static.  This is a fundamental limitation of the Bullet physics engine.  Setting a static-mesh entity dynamic is prevented in the C++ (specifically in `EntityItem::updateDynamic()`).

I'm curious to know what is the motivation for the change?  Are static objects not grabbable, or it because we'd prefer they tumble when they bump other objects as they are pulled around?

Some may ask: "Why doesn't hifi just create correct dynamic shapes for all objects?".  A short-ish answer is: We have not yet invested in writing the code to make this possible, and since the computation of such shapes can be quite CPU-expensive the ideal solution would be one where we're not computing most them in real-time, but pre-computing and then downloading them as cacheable assets.

MrRoboman: Hey @AndrewMeadows thanks for the information. The motivation is this worklist item: https://worklist.net/21288 We want to be able to far grab nondynamic yet grabbable entities. Can you give me an example entity that this will fail on?
@PhilipRosedale can  you chime in on this?

AndrewMeadows: Ok, I read the worklist item and see that it suggested this very solution: set the static object dynamic so that it can actually move.  Unfortunately, my comment above still stands, AFAIK.

However, I guess I don't understand how far-grab works because I was under the impression that it could successfully move static objects.  I will have to test it myself.  If it turns out far-grab really can't move static objects then I would suggest a different way to fix it: change the C++ action to automatically accommodate static objects.  I won't go into the details here about what that would involve, but I'd be happy to help you with it in another channel.

I will do some testing and exploration to get a better idea of what might be breaking far-grab.

AndrewMeadows: Any model entity item that has **shapeType = "static-mesh"** cannot successfully be set **dynamic = true**.  Try it and then query the **dynamic** property: it should remain **"false"**.  If not, then that would be a bug.

PhilipRosedale:
@AndrewMeadows to your point regarding not being able to far-grab a static - yes should we not do it his way and instead modify the C++ action to accomodate statics?

AndrewMeadows:
@PhilipRosedale I described to **MrRoboman** via slack how the physics simulation interfaces with objects and how I thought this feature should work.  We definitely need to do it in the C++ and it turns out there are some tricky about moving static objects around in the middle of the `PhysicsEngine::stepSimulation()` that need to be worked around.

I'm closing this PR.  **MrRoboman** is still game to tackle the worklist project so he and I will continue our discussion on slack and he'll eventually open a new PR, or re-open this one.

PR #10433, opened 2017-05-10
Fix avatar recording playback mouth not moving

ctrlaltdavid:


hifi-gustavo:
Build console output for commit: 4212fc7d9d4f97296adf9972aa80ed192dd1d0a7 available at: https://gist.github.com/0c69e069af003a19af5f68ad8a441fd3<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10433-4212fc7.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10433-4212fc7.dmg

ctrlaltdavid:
JavaScript API's setBlendShape() calls need to set _baseBlendshapeCoefficients values used for skeleton animation whereas recording playback reads from JSON in HeadData::fromJson() needs to set _blendshapsCoefficients values used for head animation.

ctrlaltdavid:
Testing

Use the PR's Sandbox and Interface to make a recording with voice (check that your avatar's mouth moves when you make the recording) and play back the recording.

AC script: playRecordingAC.js
Interface script: record.js

Check that the playback avatar's mouth moves with the voice playback.

CainFoool:
Testing

CainFoool:
Testing Finished
 - Tried with three instances of playRecordingAC.js
- All worked successfully.

Test Complete, passed. Issue resolved.

AndrewMeadows:
I don't think this is quite right...

At one time we had a single set of `_blendshapeCoefficients` that were settable by script, but were also overwritten by FaceTracker or voice modulation.  This caused a problem where scripts would see their changes immediately erased by the next frame's dynamic coefficients.  This led to an internal bug being created (fogbugz case 2782) and a PR to fix it: #10223.

I think the 10223 solution was to create the `_baseBlendshapeCoefficients` which could be set via script and would persist at whatever they were set to until set otherwise.  The ephemeral `_blenshapeCoefficients` that were generated by voice modulation or by a FaceTracker would be added to the base as an offset to provide the final value that was actually used on the face.

Perhaps the recording should be using `HeadData::getSummedBlendshapeCoefficients()` for input data and then the playback script should just use the regular invokable `AvatarData::setBlendshape()` to write the the `_baseBlendshapeCoefficients`?

anshuman64:
I agree that this is a QA Pass. However, adding Do Not Merge until we can figure the larger issues out.

ctrlaltdavid:
@AndrewMeadows That makes sense. Which raises the point that recordings should record the summed blendshapes, not `_blendshapeCoefficients`.

ctrlaltdavid:
@AndrewMeadows Yes, that makes sense. Will submit a new PR.

PR #10444, opened 2017-05-11
Desktop users of different height can shake hands

davidkelly:
Well, _somewhat_ different heights.  We move your arm up/down to be close to the right height for the closest avatar to you, when in desktop.  However, once the short person's arm is over their head, and the tall person's arm is at their waist, that's it (for now).  So, a 2x difference in height is fine (depending on arm lengths, etc...), but that's about it.  

DNM while we play around and decide if this is sufficient.

### DIFF
--- a/scripts/system/makeUserConnection.js
+++ b/scripts/system/makeUserConnection.js
@@ -198,7 +198,7 @@
     }
 
     var animationData = {};
-    function updateAnimationData() {
+    function updateAnimationData(verticalOffset) {
         // all we are doing here is moving the right hand to a spot
         // that is in front of and a bit above the hips.  Basing how
         // far in front as scaling with the avatar's height (say hips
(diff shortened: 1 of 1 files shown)


davidkelly:
fixes https://highfidelity.fogbugz.com/f/cases/4263/desktop-xbox-hand-shake-height-cannot-be-dependent-on-avatar-height, more or less.

davidkelly:
This height difference seems to be about the limit

hifi-snap-by-yohighness-on-2017-05-11_14-21-50

davidkelly:
Test Plan:

You need 2 clients
(1) make the 2 avatars different heights
(2) attempt to shake hands
(3) for height differences up to roughly the image above, you should have success

hifi-gustavo:
Build console output for commit: 52c288bfed215e817081542302a3b4766a2f0da2 available at: https://gist.github.com/3e452cd82be9405c9de3546c382dd103<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10444-52c288b.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10444-52c288b.dmg

howard-stearns:
LGTM. let's try in standup

YuppyPlays:
Is this being marked QA ready?

I have 2 computers in the same room I can test this with in the same room if that is okay for the test.

howard-stearns:
thanks YuppyPlays, but I think we'll try this out here and decide if we need to do more with it. (If we do more, it won't have been ready now. If we decide it is ready now, then we will have tested it in deciding.)

davidkelly:
tested in standup, seems like everyone is happy with it.  Marking qa approved.

PR #10456, opened 2017-05-13
Market crash fix

vladest:
Test plan
- Goto Market
- Select Chat
- Enable Chat in MessageBox
- make sure no crash occurs


hifi-gustavo:
Build console output for commit: 49a45f0300a5a0d76096c9e6bd8ef9eff1cd2630 available at: https://gist.github.com/03efbe3fb12cae54c84f7387689c1ea7<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10456-49a45f0.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10456-49a45f0.dmg

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

Closest I could find is Text Chat in Market, downloaded and allowed script  (chat.js) to run.
Used the Chat app icon on the taskbar and went through some of the commands just to ensure no crashing occurs.

Assuming the above is the correct steps requested, then all is **Pass** as no crashes occurred. 

Let me know if there is anything else I can do/provide :)

Atlante45:
I just got around to looking at this and this isn't quite gonna work as expected.

The issue is that QSettings is a key/value store.
Even though it looks like there's support for a hierarchy, there's not.

What I mean is that if you have the following keys in the settings:
`MyGroup/KeyA` : 10
`MyGroup/KeyB` : 20

You cannot request `MyGroup` and get `{ KeyA: 10, KeyB: 20 }`. You need to request each key independently.

The issue with that patch is that the user will store:
`MyGroup` : { KeyA : 10, KeyB : 20 }
They'll then have to access the keys independently, they can't request `MyGroup` again.
(Actually, they will during the same session because of our caching layer, but they won't be able too after a restart)

It gets even weirder if they store something like this:
`MyGroup` : { KeyA : 10, KeyB : { KeyC: 20 } }

In this case the stored values are:
    "MyGroup/KeyA": 10,
    "MyGroup/KeyB": {
        "KeyC": 20
    },
Even weirder is that every time the interface is run, it'll inline one more level.
So on a subsequent run, this would become:
    "MyGroup/KeyA": 10,
    "MyGroup/KeyB/KeyC": 20

This is due QSettings' odd caching system.

I originally didn't add support for complex structures because it's impossible to satisfy every use case.
One might want to load/store a complex setting object all the time, and others might want to load the complex object once and then do modification key by key and vice versa.

To fix this appropriately, we need to decide whether we want to store the full object, or inline all the keys. Personally, I'd be in favor of storing the object so that we stay consistent with the other types. But this will still lead to confusions.

PR #10398, opened 2017-05-05
Fix the PAL not showing user info for user who opened PAL

zfox23:
There's a problem in Master where the PAL will frequently be very broken - the DisplayName field will sometimes be blank, the user's profile picture will not show, the user's admin status will not be reflected properly, etc. This PR fixes that.

Currently DNM because it's 5PM on Cinco de Mayo, it's a Friday, everyone else is drinking margaritas, and I want to make sure this is a valid solution.

**Test Plan:**
1. Log in.
2. Go into a domain in which you're an admin (your sandbox works).
3. Open the PAL. Verify that the "Admin" options appear for you. Verify that your profile picture, username, and display name show up in the top left of the PAL.
4. Go into a domain in which you're not an admin (like dev-welcome, probably).
5. Open the PAL. Verify that your profile picture, username, and display name show up in the top left of the PAL.

PR #10462, opened 2017-05-15
Force _identityUpdatedAt to stay above 0, fixing more identity packet bugs

zfox23:
PR #10398 addressed the fact that identity packet timestamps didn't take server clock skew into consideration. However, I forgot to include a check to prevent the timestamp from being recorded as negative, which has two problems:
1. Absolute negative timestamps don't make sense.
2. The timestamp data type is an unsigned integer; attempting to assign a negative timestamp to `_identityUpdatedAt` causes the value to underflow.

**Test Plan:**
Prerequisites:
- This plan requires two Interface clients running - Interface A and Interface B.
- The machine running Interface A should also be running Sandbox A.
- Use unique avatars for Interface A and Interface B.
- Set the system clock of the machine running Interface B to be five minutes **ahead of** Machine A's system clock.

Test Plan:
1. Using Interface B, connect to Sandbox A.
2. Wait 10 seconds.
3. Using Interface A, connect to Sandbox A.
4. Verify that both clients can see each other's avatars correctly, and that each appears in the other's PAL.
5. Using Interface A, disconnect from Sandbox A.
6. Wait 20 seconds.
7. Using Interface A, connect to Sandbox A.
8. Verify that both clients can see each other's avatars correctly, and that each appears in the other's PAL.
9. Using Interface B, disconnect from Sandbox A.
10. Wait 10 seconds.
11. Using Interface B, connect to Sandbox A.
12. Verify that both clients can see each other's avatars correctly, and that each appears in the other's PAL.
13. **Repeat steps 1-12** after setting the system clock of the machine running Interface B to be five minutes **behind** Machine A's system clock.

### DIFF
--- a/libraries/avatars/src/AvatarData.cpp
+++ b/libraries/avatars/src/AvatarData.cpp
@@ -1538,7 +1538,14 @@ void AvatarData::processAvatarIdentity(const Identity& identity, bool& identityC
 
     // use the timestamp from this identity, since we want to honor the updated times in "server clock"
     // this will overwrite any changes we made locally to this AvatarData's _identityUpdatedAt
-    _identityUpdatedAt = identity.updatedAt - clockSkew;
+    // Additionally, ensure that the timestamp that we try to record isn't negative, as
+    // "_identityUpdatedAt" is an *unsigned* 64-bit integer. Furthermore, negative timestamps
+    // wouldn't make sense.
+    if (identity.updatedAt - clockSkew >= 0) {
+        _identityUpdatedAt = identity.updatedAt - clockSkew;
+    } else {
+        _identityUpdatedAt = 0;
+    }
 }
 
 QByteArray AvatarData::identityByteArray() const {
--- a/libraries/avatars/src/AvatarData.cpp
+++ b/libraries/avatars/src/AvatarData.cpp
@@ -1541,7 +1541,7 @@ void AvatarData::processAvatarIdentity(const Identity& identity, bool& identityC
     // Additionally, ensure that the timestamp that we try to record isn't negative, as
     // "_identityUpdatedAt" is an *unsigned* 64-bit integer. Furthermore, negative timestamps
     // wouldn't make sense.
-    if (identity.updatedAt - clockSkew >= 0) {
+    if (identity.updatedAt > clockSkew) {
         _identityUpdatedAt = identity.updatedAt - clockSkew;
     } else {
         _identityUpdatedAt = 0;


libraries/avatars/src/AvatarData.cpp
```diff
@@ -1538,7 +1538,14 @@ void AvatarData::processAvatarIdentity(const Identity& identity, bool& identityC
 
     // use the timestamp from this identity, since we want to honor the updated times in "server clock"
     // this will overwrite any changes we made locally to this AvatarData's _identityUpdatedAt
-    _identityUpdatedAt = identity.updatedAt - clockSkew;
+    // Additionally, ensure that the timestamp that we try to record isn't negative, as
+    // "_identityUpdatedAt" is an *unsigned* 64-bit integer. Furthermore, negative timestamps
+    // wouldn't make sense.
+    if (identity.updatedAt - clockSkew >= 0) {
```
davidkelly: Maybe I'm misreading this -- but isn't identity.updatedAt a quint64?  If so this probably cannot be < 0.  If so, then the test needs to be identity.updatedAt > clockSkew perhaps?

zfox23: Discussed offline; I now understand that it's not 100% for sure that all compilers will put the result of this operation into a signed integer. I should change this test. Thanks!

hifi-gustavo:
Build console output for commit: a2a24b22a8ba82b9f2637e7408a83a81c7e8d72f available at: https://gist.github.com/e1d4abf4e7b56206d287e154af23df1f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10462-a2a24b2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10462-a2a24b2.dmg

hifi-gustavo:
Build console output for commit: e492ca9364c8b9f8aae61b09ed5ae4acaa97d4aa available at: https://gist.github.com/e1d4abf4e7b56206d287e154af23df1f<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10462-e492ca9.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10462-e492ca9.dmg

libraries/avatars/src/AvatarData.cpp
```diff
@@ -1538,7 +1538,14 @@ void AvatarData::processAvatarIdentity(const Identity& identity, bool& identityC
...
     // this will overwrite any changes we made locally to this AvatarData's _identityUpdatedAt
-    _identityUpdatedAt = identity.updatedAt - clockSkew;
+    // Additionally, ensure that the timestamp that we try to record isn't negative, as
+    // "_identityUpdatedAt" is an *unsigned* 64-bit integer. Furthermore, negative timestamps
+    // wouldn't make sense.
+    if (identity.updatedAt > clockSkew) {
+        _identityUpdatedAt = identity.updatedAt - clockSkew;
+    } else {
+        _identityUpdatedAt = 0;
+    }
```
AndrewMeadows: `clockSkew` can't be more than several seconds, right?  Meanwhile any valid timestamp would have to be greater than at least 46 years.  So... I don't think you need the `else` statement because anything smaller than a few seconds is close enough to zero for the one place where the value of `AvatarData::_identityUpdatedAt` actually matters.

zfox23: `clockSkew` can be anything - what makes you say that it can't be more than several seconds?

AndrewMeadows: Sorry, I deleted my comment, but:  any reasonable clockSkew is going to be small relative to the timestamp itself, which will be greater than 46 years (e.g. relative to beginning of epoch: Jan 1st 1970).

Now on to other things.  Consider this code:
```
    uint64_t foo = usecTimestampNow();
    int64_t bar = -1;
    bool fubar = (foo > bar);
    std::cout << "debug foo = " << foo << std::endl;  
    std::cout << "debug bar = " << bar << std::endl;  
    std::cout << "debug (foo > bar) = " << fubar << std::endl;
```
Which  produces this output:
```
debug foo = 1494889732013322
debug bar = -1
debug (foo > bar) = 0
```
Why does this happen?  It is explained on this stackoverflow page:

http://stackoverflow.com/questions/29735103/comparing-int64-t-and-uint64-t

What I think is happening is: these changes are constantly setting `_identityUpdatedAt` to zero, which is probably not what you wanted to do.

PR #10465, opened 2017-05-16
Fix domain settings not getting cleared on re-assignment

Atlante45:
Fix a bug where some audio settings would not get updated after clicking on "Save & Restart"

## Test Plan:

This is pretty tricky to test correctly. Ideally, this would be done by a dev or someone familiar with manual server setups.

- Point the audio mixer to the domain manually
- Make sure it gets the right assignment
- Create 2 audio zones
- Have someone else stand in the first zone, stand in the second one
- Change the attenuation values and makes sure it updates accordingly after server restarts
(You might have to kill the audio-mixer node manually so that the dedicated AC gets the assignment)

[Bug](https://highfidelity.fogbugz.com/f/cases/3198/Zone-attenuation-effects-don-t-get-cleared-out-when-removed-from-domain-settings)

### DIFF
--- a/assignment-client/src/audio/AudioMixer.cpp
+++ b/assignment-client/src/audio/AudioMixer.cpp
@@ -38,13 +38,14 @@
 #include "AudioMixer.h"
 
 static const float DEFAULT_ATTENUATION_PER_DOUBLING_IN_DISTANCE = 0.5f;    // attenuation = -6dB * log2(distance)
+static const int DEFAULT_NUM_STATIC_JITTER_FRAMES = -1;
 static const float DEFAULT_NOISE_MUTING_THRESHOLD = 1.0f;
 static const QString AUDIO_MIXER_LOGGING_TARGET_NAME = "audio-mixer";
 static const QString AUDIO_ENV_GROUP_KEY = "audio_env";
 static const QString AUDIO_BUFFER_GROUP_KEY = "audio_buffer";
 static const QString AUDIO_THREADING_GROUP_KEY = "audio_threading";
 
-int AudioMixer::_numStaticJitterFrames{ -1 };
+int AudioMixer::_numStaticJitterFrames{ DEFAULT_NUM_STATIC_JITTER_FRAMES };
 float AudioMixer::_noiseMutingThreshold{ DEFAULT_NOISE_MUTING_THRESHOLD };
 float AudioMixer::_attenuationPerDoublingInDistance{ DEFAULT_ATTENUATION_PER_DOUBLING_IN_DISTANCE };
 std::map<QString, std::shared_ptr<CodecPlugin>> AudioMixer::_availableCodecs{ };
@@ -56,7 +57,12 @@ QVector<AudioMixer::ReverbSettings> AudioMixer::_zoneReverbSettings;
 AudioMixer::AudioMixer(ReceivedMessage& message) :
     ThreadedAssignment(message) {
 
+    // Always clear settings first
+    // This prevents previous assignment settings from sticking around
+    clearDomainSettings();
+
     // hash the available codecs (on the mixer)
+    _availableCodecs.clear(); // Make sure struct is clean
     auto codecPlugins = PluginManager::getInstance()->getCodecPlugins();
     std::for_each(codecPlugins.cbegin(), codecPlugins.cend(),
         [&](const CodecPluginPointer& codec) {
@@ -232,7 +238,7 @@ void AudioMixer::sendStatsPacket() {
     }
 
     // general stats
-    statsObject["useDynamicJitterBuffers"] = _numStaticJitterFrames == -1;
+    statsObject["useDynamicJitterBuffers"] = _numStaticJitterFrames == DEFAULT_NUM_STATIC_JITTER_FRAMES;
 
     statsObject["threads"] = _slavePool.numThreads();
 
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Build console output for commit: cde7d6857018846c2b74b37faf12b3e48000cab7 available at: https://gist.github.com/6a5efc2061bb94ab66f4eaa4ce72d864<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10465-cde7d68.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10465-cde7d68.dmg

assignment-client/src/audio/AudioMixer.cpp
```diff
@@ -232,7 +238,7 @@ void AudioMixer::sendStatsPacket() {
     }
 
     // general stats
-    statsObject["useDynamicJitterBuffers"] = _numStaticJitterFrames == -1;
+    statsObject["useDynamicJitterBuffers"] = _numStaticJitterFrames == DEFAULT_NUM_STATIC_JITTER_FRAMES;
```
zzmp: This isn't checking against the default, but against a magic value that has us using jitter buffers. There should be another `static const` for this (which the default should be set equal to).

assignment-client/src/audio/AudioMixer.cpp
```diff
@@ -525,7 +541,7 @@ void AudioMixer::parseSettingsObject(const QJsonObject &settingsObject) {
             qDebug() << "Static desired jitter buffer frames:" << _numStaticJitterFrames;
         } else {
             qDebug() << "Disabling dynamic jitter buffers.";
-            _numStaticJitterFrames = -1;
+            _numStaticJitterFrames = DEFAULT_NUM_STATIC_JITTER_FRAMES;
```
zzmp: Same here - it shouldn't be "disabling" with the "default", but with a disabling magic number.

assignment-client/src/audio/AudioMixer.cpp
```diff
@@ -56,7 +57,12 @@ QVector<AudioMixer::ReverbSettings> AudioMixer::_zoneReverbSettings;
 AudioMixer::AudioMixer(ReceivedMessage& message) :
     ThreadedAssignment(message) {
 
+    // Always clear settings first
+    // This prevents previous assignment settings from sticking around
+    clearDomainSettings();
+
     // hash the available codecs (on the mixer)
+    _availableCodecs.clear(); // Make sure struct is clean
```
zzmp: Why can't this be cleared with `clearDomainSettings`?

Atlante45: It's not a domain setting per se, seemed semantically incorrect to group it with them.

assignment-client/src/audio/AudioMixer.cpp
Atlante45: Good point, I can call it `DISABLE_STATIC_JITER_FRAMES`. What do you think?

Atlante45: Or `USE_DYNAMIC_JITTER_FRAMES` would work too I guess.

Atlante45:
@zzmp When you restart the domain, the ACs lose their assignment (ie. they get back to an unassigned state), but the process doesn't end. So when they get assigned again, those static members are still set from the previous assignment if it happened to be an audio-mixer as well.

assignment-client/src/audio/AudioMixer.cpp
zzmp: Isn't it set in the audio settings in localhost:40102, as a comma-separated list?

assignment-client/src/audio/AudioMixer.cpp
zzmp: I like `DISABLE_STATIC_JITTER_FRAMES`, but I'm good with either.

zzmp:
I did not know that. I assumed the procs were all killed and restarted.

assignment-client/src/audio/AudioMixer.cpp
Atlante45: You're thinking of `_codecPreferenceOrder` which is in `clearDomainSettings`.
`_availableCodecs` is populated based on the available codecs on the machine and is then used to negotiate the codec to use with each client based on the `_codecPreferenceOrder` setting.

But since it's static, it's affected too (although it wouldn't do much in that case).
TBH I'd much rather make all that non static and give a reference to AudioMixer to anyone that needs those values (o maybe just pass those values in).
But I really didn't want to mess with the AM multithreaded code.

hifi-gustavo:
Build console output for commit: c799ddc9276b207d14e778d12920076c1ebf349f available at: https://gist.github.com/6a5efc2061bb94ab66f4eaa4ce72d864<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10465-c799ddc.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10465-c799ddc.dmg

YuppyPlays:
Ok to test?

assignment-client/src/audio/AudioMixer.cpp
zzmp: fair. so fair.

OmegaHeron:
Working on this now.  Be a bit for results.

Atlante45:
Thanks, @OmegaHeron!
Don't hesitate to ping me on Slack if you need anything from me.

OmegaHeron:
@Atlante45 - Slack seems to be having some issues, I can't connect to at this time.

Maybe it would be better to get one of the internal devs to check this - I can *not* successfully create a pair of zones in Master or this PR using model of stage/audience.  With slack being dead for me and my time for today almost over -- will have to pass to another tester or be later before I can attempt again.

Atlante45:
No worries, thanks for the heads up!

zzmp:
Set up zones with attenuation of 1. Could not hear sound between the zones (as expected).
Cleared the attenuation zones. Could hear sound again.
Ensured the audio-mixer was on the same node throughout.

QA Passed.

PR #10473, opened 2017-05-16
Remove unnecessary assert in Model.cpp

huffman:
This clause inside was not longer valid - it is a valid state for
part->_meshIndex < _modelMeshRenderItems.size(). A better check might be
to compare it to the size of _meshStates which we use below, but we use
_meshStates.at(...) which will do the bounds checking, which makes the
assert unnecessary.

### DIFF
--- a/libraries/render-utils/src/Model.cpp
+++ b/libraries/render-utils/src/Model.cpp
@@ -1052,7 +1052,6 @@ void Model::updateRig(float deltaTime, glm::mat4 parentTransform) {
 
 void Model::computeMeshPartLocalBounds() {
     for (auto& part : _modelMeshRenderItems) {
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Build console output for commit: 8eacf0ad4b8579a5f6f2be561002a1020924a45c available at: https://gist.github.com/2286c745550fae2bdb9878538dea9178<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10473-8eacf0a.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10473-8eacf0a.dmg

AndrewMeadows:
This is not testable with release builds.

PR #10483, opened 2017-05-17
cap scripted linear and angular velocity inputs

AndrewMeadows:


AndrewMeadows:
Users are starting to set unreasonable velocities on objects and the resulting crashes on are being put on my plate.  We're now going to filter unreasonable values from the obvious modes: velocities and acceleration.  Perhaps we'll add more caps on other aspects (e.g. size and complexity of dynamic objects) later.

AndrewMeadows:
Test Plan:

(1) Create a dynamic object and set its velocity as large as possible --> interface should not crash.
(2) Create a dynamic object and set its angularVelocity as large as possible --> interface should not crash.
(3) Create a dynamic object and set its gravity as large as possible --> interface should not crash.
(4) For bonus points set all three parameters as large as possible --> interface should not crash.

Note: if you make the object very large in step (2), spin it very fast, and allow it to collide with very small dynamic objects you might be able to get the interface to crash when not-a-number (NaN) gets introduced into the physics simulation.  This would not be considered a blocker: we may need to cap the dimensions of dynamic objects or cap angular velocity as a function of the object's size in a later PR.

hifi-gustavo:
Build console output for commit: 97ed7195069c864544ff56daa97960b954779455 available at: https://gist.github.com/070bc524b7ecf61b201be8e9964b9a64<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10483-97ed719.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10483-97ed719.dmg

libraries/entities/src/EntityItem.cpp
```diff
@@ -1735,14 +1749,20 @@ void EntityItem::updateAngularVelocity(const glm::vec3& value) {
         if (getShapeType() == SHAPE_TYPE_STATIC_MESH) {
             setLocalAngularVelocity(Vectors::ZERO);
         } else {
-            const float MIN_ANGULAR_SPEED = 0.0002f;
-            if (glm::length(value) < MIN_ANGULAR_SPEED) {
-                angularVelocity = ENTITY_ITEM_ZERO_VEC3;
-            } else {
-                angularVelocity = value;
+            float speed = glm::length(value);
+            if (!isnan(speed)) {
```
ZappoMan: shouldn't we implement the else and slam this to zero?

CainFoool:
Testing shortly!

CainFoool:
Test Complete

Was able to set the velocity, angular & gravity to obscene values without the game crashing however it quickly defaulted them to the highest 'possible', occasionally some values dropped to 0 when changing others.

Wasn't able to reproduce the spinning crash - seems to be okay.

**PASS**

libraries/entities/src/EntityItem.cpp
AndrewMeadows: The `updateAngularVelocity()` method takes data from `EntityItemProperties` that have been filled by JS script or by network packets.  I figured it would be better to ignore NaN values and keep the old value.  Suppose there were an object spinning fine and a script tried to set its angular velocity to NaN... we could either allow it to continue to spin fine (as implemented here) or we could stop it.  The second case seems more... unexpected and odd to me.

AndrewMeadows:
@CainFoool, thanks for testing despite the fact that I had forgotten to put up the `QA_READY` tag.

CainFoool:
No worries @AndreaMeadows - I know it looked like a fairly well laid out Test Plan so I assumed it was ready.

PR #10485, opened 2017-05-17
Add protections when accessing the mip bytes of a texture when provided from a KtxStorage

samcake:
This pr potentially address a problem and a crash if the KtxStorage file is failing to be mapped during the access sequence.
In theory this should not happen but we have seen crashes due to this:
https://www.bugsplat.com/individualCrash/?id=11792&database=interface_alpha

I have dsiabled the exception thrown from FileStorage thinking that since we don;t have a handler, this may be the reason fro another crash without a bugsplat...

This is attempting to fix this bug:
https://highfidelity.fogbugz.com/f/cases/4795/CRASH-in-storage-MemoryStorage-MemoryStorage

## TEST PLAN
No real repro for this so simply check that there is weird behavior different from current master

### DIFF
--- a/libraries/gpu/src/gpu/Texture_ktx.cpp
+++ b/libraries/gpu/src/gpu/Texture_ktx.cpp
@@ -210,7 +210,16 @@ PixelsPointer KtxStorage::getMipFace(uint16 level, uint8 face) const {
     auto faceSize = _ktxDescriptor->getMipFaceTexelsSize(level, face);
     if (faceSize != 0 && faceOffset != 0) {
         auto file = maybeOpenFile();
-        result = file->createView(faceSize, faceOffset)->toMemoryStorage();
+        if (file) {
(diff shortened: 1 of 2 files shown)


Nex-Pro:
Will test this shortly. When the build is available

hifi-gustavo:
Build console output for commit: ab3ea65d5417a661bb561de88e3f62b099b1f7af available at: https://gist.github.com/1c23c56a1111240624232369b56afb37<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10485-ab3ea65.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10485-ab3ea65.dmg

Nex-Pro:
Testing

Nex-Pro:
Gave this a soak test in locations with many textures. Nothing unusual/bad noticed.

### PASS

PR #10487, opened 2017-05-17
Remove extraneous calls to sendIdentityPacket()

zfox23:
Extraneous calls to sendIdentityPacket() could cause problems given our current system of timestamping identity packets. This PR removes those extraneous calls.

[Item number 2 of this FogBugz ticket.](https://highfidelity.fogbugz.com/f/cases/4801/T-posers-and-unable-to-use-PAL)

**Test Plan:**
Prerequisites:
- This plan requires two Interface clients running - Interface A and Interface B.
- The machine running Interface A should also be running Sandbox A.
- Use unique avatars and display names for Interface A and Interface B.

Test Plan:
1. Using Interface B, connect to Sandbox A.
3. Using Interface A, connect to Sandbox A.
4. Verify that both clients can see each other's avatars correctly, and that each appears in the other's PAL.
5. Using Interface A, disconnect from Sandbox A.
6. Wait 20 seconds.
7. Using Interface A, connect to Sandbox A.
8. Verify that both clients can see each other's avatars correctly, and that each appears in the other's PAL.
9. Using Interface B, disconnect from Sandbox A.
10. Wait 10 seconds.
11. Using Interface B, connect to Sandbox A.
12. Verify that both clients can see each other's avatars correctly, and that each appears in the other's PAL.
13. Set Interface B to use the same display name as Interface A, then refresh Interface B's PAL.
14. Verify that both Interface A and Interface B see the same Session Display Name - it should be "<Interface A's Display Name>_1"
15. Restart Interface B. Re-verify the results from Step 14.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -5258,9 +5258,8 @@ void Application::nodeActivated(SharedNodePointer node) {
     }
 
     if (node->getType() == NodeType::AvatarMixer) {
-        // new avatar mixer, send off our identity packet right away
+        // new avatar mixer, send off our identity packet on next update loop
         getMyAvatar()->markIdentityDataChanged();
-        getMyAvatar()->sendIdentityPacket();
         getMyAvatar()->resetLastSent();
     }
 }
--- a/libraries/avatars/src/AvatarData.cpp
+++ b/libraries/avatars/src/AvatarData.cpp
@@ -1588,8 +1588,6 @@ void AvatarData::setDisplayName(const QString& displayName) {
     _displayName = displayName;
     _sessionDisplayName = "";
 
-    sendIdentityPacket();
-
     qCDebug(avatars) << "Changing display name for avatar to" << displayName;
     markIdentityDataChanged();
 }


CainFoool:
Testing when avail

hifi-gustavo:
Build console output for commit: d4752926b8f35d6ff093d9702e6bda8082fba812 available at: https://gist.github.com/7de0135b4bd2dd8413e843a06651ac85<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10487-d475292.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10487-d475292.dmg

CainFoool:
Testing on two PC's now.

howard-stearns:
I think there's an error in the test plan at step 14, it says to restart B (but doesn't say to restart A), and then re-verify the results from "step 14":
1. It is step 14. Do you mean 13?
2. Restarting A but not B, I get "<Interface A's Display Name>_2", not ...1. This is actually what I expect, because the server doesn't know whether a new person is the same as before or not, and so it keeps adding to the numbers for a given name until the last person with that name has left, and THEN it starts at 1 again for the next person after that.

CainFoool:
Test Completed

Worked exactly as described. Could see each Interface Player just fine and at the last step when I named them to the same nick, when I re-joined I came back as TestBoxA_2 instead of TestBoxA_1; though I believe that's just how HiFi handles dupe entities.

**PASSED**

zfox23:
@howard-stearns you are correct about the numerical error. As for #2 - I had in mind to just restart B, but, in any case, @CainFoool got the directions correct and his noted behavior is a PASS.

Thanks to you both.

PR #10490, opened 2017-05-17
Enable Map/List storage in the settings

Atlante45:
This enables storing maps and lists in the Settings.

## Test Plan:
- Open your scripting console
- Enter the following
```js
Settings.setValue("Map", { a: 1, b: "2", c: [3, "4"] });
Settings.setValue("List", [ 1, "2", { a: 3, b: "4" }]);
```
- Restart Interface
- Open your scripting console
- Enter the following
```js
JSON.stringify(Settings.getValue("Map"));
```
- The console should print out:
```js
{"a":1,"b":"2","c":[3,"4"]}
```
- Enter the following
```js
JSON.stringify(Settings.getValue("List"));
```
- The console should print out:
```js
[1,"2",{"a":3,"b":"4"}]
```

PR #10493, opened 2017-05-18
include details in critical variantMap parsing error

zzmp:


zzmp:
No QA needed.

Atlante45:
Just FYI, I made this PR today to add support for those:
https://github.com/highfidelity/hifi/pull/10490
That was in response to my post in:
https://github.com/highfidelity/hifi/pull/10456

Technically could still happen for a Hash type, but that would require someone storing a `QHash` in the settings from C++, which I hope never happens!

PR #10497, opened 2017-05-18
fix crash for Zone with 'compoundURL' property

AndrewMeadows:
* fix crash on changing modelURL on zone

### DIFF
--- a/libraries/entities-renderer/src/RenderableZoneEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableZoneEntityItem.cpp
@@ -82,22 +82,18 @@ int RenderableZoneEntityItem::readEntitySubclassDataFromBuffer(const unsigned ch
     return bytesRead;
 }
 
-void RenderableZoneEntityItem::initialSimulation() {
-    _model->setScaleToFit(true, getDimensions());
-    _model->setSnapModelToRegistrationPoint(true, getRegistrationPoint());
-    _model->setRotation(getRotation());
-    _model->setTranslation(getPosition());
-    _model->simulate(0.0f);
-    _needsInitialSimulation = false;
-}
-
 void RenderableZoneEntityItem::updateGeometry() {
     if (_model && !_model->isActive() && hasCompoundShapeURL()) {
         // Since we have a delayload, we need to update the geometry if it has been downloaded
         _model->setURL(getCompoundShapeURL());
     }
     if (_model && _model->isActive() && _needsInitialSimulation) {
-        initialSimulation();
+        _model->setScaleToFit(true, getDimensions());
+        _model->setSnapModelToRegistrationPoint(true, getRegistrationPoint());
+        _model->setRotation(getRotation());
+        _model->setTranslation(getPosition());
+        _model->simulate(0.0f);
+        _needsInitialSimulation = false;
     }
 }
 
--- a/libraries/entities-renderer/src/RenderableZoneEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableZoneEntityItem.cpp
@@ -81,7 +81,7 @@ int RenderableZoneEntityItem::readEntitySubclassDataFromBuffer(const unsigned ch
 }
 
 Model* RenderableZoneEntityItem::getModel() {
-    Model* model = new Model(nullptr);
+    Model* model = new Model();
     model->setIsWireframe(true);
     model->init();
     return model;
--- a/libraries/entities-renderer/src/RenderableZoneEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableZoneEntityItem.cpp
@@ -36,15 +36,17 @@ void RenderableZoneEntityItem::changeProperties(Lambda setNewProperties) {
     QString oldShapeURL = getCompoundShapeURL();
     glm::vec3 oldPosition = getPosition(), oldDimensions = getDimensions();
     glm::quat oldRotation = getRotation();
-    
+
     setNewProperties();
-    
+
     if (oldShapeURL != getCompoundShapeURL()) {
         if (_model) {
-            delete _model;
+            _model.reset();
         }
-        
-        _model = getModel();
+
+        _model = std::make_shared<Model>();
+        _model->setIsWireframe(true);
+        _model->init();
         _needsInitialSimulation = true;
         _model->setURL(getCompoundShapeURL());
     }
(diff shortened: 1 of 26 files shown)


AndrewMeadows:
The `Model` ctor has one required argument: a pointer to a Rig.  A `RenderableZoneEntityItem` creates a model with a null `Rig` pointer... which means `Model::deleteGeometry()` must check the validity of `Model::_rig` before dereferencing it.

AndrewMeadows:
I will do some research and try to formulate a reliable repro recipe for this crash.

hifi-gustavo:
Build console output for commit: 8d3285f733461515e43f33abd3ef0b2d44893c4f available at: https://gist.github.com/43c68c11f3c84cee8d3640a15690ab58<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10497-8d3285f.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10497-8d3285f.dmg

AndrewMeadows:
Don't Merge.  I found a reliable repro for this problem.  The crashiness is bigger than I thought.  Needs more work.

AndrewMeadows:
To test:

(1) Run the script below that creates a zone with non-empty collisionShapeURL (used to define non-trivial bounds of the zone).
--> on hfii/master this will crash
--> does not crash in this PR
```
// create an entity with shape URL --> crash
var zoneId = Entities.addEntity({
    type: "Zone",
    position: { x: 5, y: 10, z: 5 },
    dimensions: { x: 10, y: 10, z: 10 },
    keyLightColor: { red: 0, green: 0, blue: 255 },
    keyLightIntensity: 0.75,
    keyLightDirection: { x: 0, y: 0, z: -1 },
    stageSunModelEnabled: false,
    lifetime: 60,
    shapeType: "compound",
    compoundShapeURL: "http://headache.hungry.com/~seth/hifi/cube.fbx"
});

// delete zone when this script is stopped
deleteEntity = function() {
    Entities.deleteEntity(zondId);
}
Script.scriptEnding.connect(deleteEntity);
```

hifi-gustavo:
Build console output for commit: 17b6cf29df50d53017906f17302532ab9886c525 available at: https://gist.github.com/43c68c11f3c84cee8d3640a15690ab58<br/><br/>Installers available for download at: <br/>Windows: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10497-17b6cf2.exe<br/>OS X: http://cachefly.highfidelity.com/HighFidelity-Beta-PR10497-17b6cf2.dmg

Nex-Pro:
Will test this.

Nex-Pro:
Testing.

Nex-Pro:
Crash confirmed on master.

I'm getting crashes running this in this build. Posted splats.
I ran script  from my desktop.
I'll hold awaiting advice.

Nex-Pro:
@AndrewMeadows  ^

AndrewMeadows:
Yeah, it looks like there are several crash modes here when we create a Model that has no Rig.  I'm going to have to step back and take a more critical look at the whole "Zone has a Model that it sometimes doesn't use" situation.

ZappoMan:
@AndrewMeadows - FYI - if a zone has a model - then it's intended to be a convex hull used for: rendering a boundary when visualize zones is enabled, AND more importantly - detect when the avatar is "inside" the convex hull/zone to enable enter/leave zone behavior.

libraries/entities-renderer/src/RenderableZoneEntityItem.h
```diff
@@ -48,14 +48,13 @@ class RenderableZoneEntityItem : public ZoneEntityItem  {
...
 
-    Model* getModel();
     void initialSimulation();
     void updateGeometry();
     
     template<typename Lambda>
     void changeProperties(Lambda functor);
     
-    Model* _model;
+    ModelPointer _model;
```
AndrewMeadows: The `RenderableZoneEntityItem` has a `_model` data member but only uses it when the Zone is given a funny shape via the `compoundURL` property.  The crashy problem, which has existed for at least a year but was so rarely tickled that we didn't notice, was that it would create a `Model` instance that had a null `Rig`... but the `Model` code was written with the assumption that a valid `Rig` pointer is always passed into the `Model` ctor.

After talking with @hyperlogic about the history of the Rig+Model relationship and how it is supposed to work I concluded the solution is to change `Model` to have a real `Rig` instance rather than just a `std::shared_ptr<Rig>`... which would force the truth of the assumptions that are implemented in the code.  The good news is that this cleans up some API and code logic.  There is no bad news except I had to change 20+ files to do it.

libraries/render-utils/src/Model.h
```diff
@@ -390,7 +391,7 @@ public slots:
     mutable bool _needsUpdateTextures { true };
 
     friend class ModelMeshPartPayload;
-    RigPointer _rig;
+    Rig _rig;
```
AndrewMeadows: Every `Model` has a `Rig`.

libraries/entities-renderer/src/RenderableZoneEntityItem.cpp
```diff
@@ -36,15 +36,17 @@ void RenderableZoneEntityItem::changeProperties(Lambda setNewProperties) {
...
+            _model.reset();
         }
-        
-        _model = getModel();
+
+        _model = std::make_shared<Model>();
+        _model->setIsWireframe(true);
+        _model->init();
         _needsInitialSimulation = true;
         _model->setURL(getCompoundShapeURL());
```
AndrewMeadows: The `RenderableModelEntityItem` blows away its `_model` and creates a new one whenever it its URL changes, rather than just updating the relevant bits.  A bit crude but it works.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/62/HighFidelity-Beta-PR10497-d3956b47d08ca164d6ce45ba972426a5a66142d0.exe

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/64/HighFidelity-Beta-PR10497-f80e7a374368b6b4d496df61db3267631e53e4d2.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/64/HighFidelity-Beta-PR10497-f80e7a374368b6b4d496df61db3267631e53e4d2.dmg

Nex-Pro:
will jump back on this shortly

CainFoool:
Testing.

CainFoool:
# Testing Failed

There is no crash now as the zone is spawned in, however:

- When you're inside the zone, the outer skybox flickers (in this case, the Garden skybox)
- Only once you step outside of the small zone will your skybox not flicker.
- Flicker only occurs when you're moving your camera it seems.

I have tested this on Master with the results:
- The skybox inside the zone is just black, and does not inherit it's parent.

AndrewMeadows:
The visual behavior changed so I'm marking this as QA_FAILED.

AndrewMeadows:
I spoke with @samcake.  He says that the zone skybox flicker is a known problem.  He's got a project in the works to revamp the logic flow on how skybox precidence happens for embedded zones.  So I'm going to mark this as **QA_PASSED**.

CainFoool:
Awesome.

PR #10516, opened 2017-05-19
Increment _identitySequenceId before sending ID packet for agents

zfox23:
Does what it says on the tin.

**Test Plan:**
1. On a clean installation (i.e. reset settings) of Interface and Sandbox, run Sandbox. Connect to your Sandbox from Interface.
2. Verify that your sandbox's logs are not spammed, once per second, with
`[avatar-mixer] Ignoring older identity packet for avatar QUuid("{00000000-0000-0000-0000-000000000000}") _identitySequenceId ( 4 ) is greater than 2`

Note that you may see the above message happen once with different values (instead of 4 and 2), but this should not be considered a fail. You might see those messages because we haven't fully fixed the way scripted agents send identity packets.

### DIFF
--- a/assignment-client/src/Agent.cpp
+++ b/assignment-client/src/Agent.cpp
@@ -584,6 +584,7 @@ void Agent::setIsAvatar(bool isAvatar) {
 void Agent::sendAvatarIdentityPacket() {
     if (_isAvatar) {
         auto scriptedAvatar = DependencyManager::get<ScriptableAvatar>();
+        scriptedAvatar->markIdentityDataChanged();
         scriptedAvatar->sendIdentityPacket();
     }
 }


assignment-client/src/Agent.cpp
```diff
@@ -584,6 +584,7 @@ void Agent::setIsAvatar(bool isAvatar) {
 void Agent::sendAvatarIdentityPacket() {
     if (_isAvatar) {
         auto scriptedAvatar = DependencyManager::get<ScriptableAvatar>();
+        scriptedAvatar->markIdentityDataChanged();
```
ZappoMan: this still results in agents sending many more packets than they need to... right? because it's quite likely that nothing has changed?

zfox23: Yep. This is absolutely a band-aid to fix the logspam. This explicitly does not fix the fact that we are still sending more identity packets than we need to. See the test plan comments above.

howard-stearns: correct. This does not attempt to fix that.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/70/HighFidelity-Beta-PR10516-487319f7efca5e05881e9df2a11730c9bcf64a30.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/70/HighFidelity-Beta-PR10516-487319f7efca5e05881e9df2a11730c9bcf64a30.dmg

PR #10530, opened 2017-05-23
Ensure user never loses its running scripts

Atlante45:
Change the way we save the running scripts to ensure users will never lose them.

## Test Plan:
- Launch interface, you should start with the default scripts.
- Modify your running scripts
- Restart interface
- Verify scripts were save correctly
- Modify your running scripts.
- Crash interface using a "Developer > Crash" menu option
- Verify scripts were save correctly
- Launch the current release build
- Modify your running scripts
- Override the PR's settings file with the release settings file
- Launch interface
- Verify scripts were save correctly

[Bug](https://highfidelity.fogbugz.com/f/cases/4690/Crash-leaves-me-with-no-running-scripts)

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1622,7 +1622,6 @@ void Application::cleanupBeforeQuit() {
     // Clear any queued processing (I/O, FBX/OBJ/Texture parsing)
     QThreadPool::globalInstance()->clear();
 
-    DependencyManager::get<ScriptEngines>()->saveScripts();
     DependencyManager::get<ScriptEngines>()->shutdownScripting(); // stop all currently running global scripts
     DependencyManager::destroy<ScriptEngines>();
 
--- a/libraries/script-engine/src/ScriptEngines.cpp
+++ b/libraries/script-engine/src/ScriptEngines.cpp
@@ -26,8 +26,12 @@
 #define __LOC__ __FILE__ "(" __STR1__(__LINE__) ") : Warning Msg: "
 
 static const QString DESKTOP_LOCATION = QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
-
 static const bool HIFI_SCRIPT_DEBUGGABLES { true };
+static const QString SETTINGS_KEY { "RunningScripts" };
+static const QUrl DEFAULT_SCRIPTS_LOCATION { "file:///~//defaultScripts.js" };
+// Using a QVariantList so this is human-readable in the settings file
+static Setting::Handle<QVariantList> runningScriptsHandle(SETTINGS_KEY, { QVariant(DEFAULT_SCRIPTS_LOCATION) });
+
 
 ScriptsModel& getScriptsModel() {
     static ScriptsModel scriptsModel;
@@ -61,19 +65,6 @@ ScriptEngines::ScriptEngines(ScriptEngine::Context context)
     _scriptsModelFilter.setSourceModel(&_scriptsModel);
     _scriptsModelFilter.sort(0, Qt::AscendingOrder);
     _scriptsModelFilter.setDynamicSortFilter(true);
-
-    static const int SCRIPT_SAVE_COUNTDOWN_INTERVAL_MS = 5000;
-    QTimer* scriptSaveTimer = new QTimer(this);
-    scriptSaveTimer->setSingleShot(true);
-    QMetaObject::Connection timerConnection = connect(scriptSaveTimer, &QTimer::timeout, [] {
-        DependencyManager::get<ScriptEngines>()->saveScripts();
-    });
-    connect(qApp, &QCoreApplication::aboutToQuit, [=] {
-        disconnect(timerConnection);
-    });
-    connect(this, &ScriptEngines::scriptCountChanged, this, [scriptSaveTimer] {
-        scriptSaveTimer->start(SCRIPT_SAVE_COUNTDOWN_INTERVAL_MS);
-    }, Qt::QueuedConnection);
 }
 
 QUrl normalizeScriptURL(const QUrl& rawScriptURL) {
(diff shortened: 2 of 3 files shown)


hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/105/HighFidelity-Beta-PR10530-d57cae81030a07ee15b8f1734a16b01fc00fcdd2.exe

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/108/HighFidelity-Beta-PR10530-10eb353126afa9239720adb59e36a75a1f29b459.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/108/HighFidelity-Beta-PR10530-10eb353126afa9239720adb59e36a75a1f29b459.dmg

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:
**PR Restart** Scripts saved and ran on restart. **Pass**
**PR Crash** Scripts saved and ran on restart **Pass**
**Overriding PR with release settings** Worked perfectly and carried over the scripts that were running in release. **Pass**

**Pass**

If there is anything else I can do/provide please let me know :)

interface/src/Application.cpp
```diff
@@ -1622,7 +1622,6 @@ void Application::cleanupBeforeQuit() {
     // Clear any queued processing (I/O, FBX/OBJ/Texture parsing)
     QThreadPool::globalInstance()->clear();
 
-    DependencyManager::get<ScriptEngines>()->saveScripts();
```
Atlante45: We don't save on shutdown anymore, we save as soon as a change happens.

libraries/script-engine/src/ScriptEngines.cpp
```diff
@@ -61,19 +65,6 @@ ScriptEngines::ScriptEngines(ScriptEngine::Context context)
...
-    scriptSaveTimer->setSingleShot(true);
-    QMetaObject::Connection timerConnection = connect(scriptSaveTimer, &QTimer::timeout, [] {
-        DependencyManager::get<ScriptEngines>()->saveScripts();
-    });
-    connect(qApp, &QCoreApplication::aboutToQuit, [=] {
-        disconnect(timerConnection);
-    });
-    connect(this, &ScriptEngines::scriptCountChanged, this, [scriptSaveTimer] {
-        scriptSaveTimer->start(SCRIPT_SAVE_COUNTDOWN_INTERVAL_MS);
-    }, Qt::QueuedConnection);
```
Atlante45: We moved from the old legacy settings that were really slow, so we don't need to delay saving by 5 sec to grab several at the same time when possible.
We can do it right as it happens.

libraries/script-engine/src/ScriptEngines.cpp
```diff
@@ -280,67 +271,72 @@ QVariantList ScriptEngines::getRunning() {
...
 }
 
 void ScriptEngines::loadScripts() {
-    // check first run...
-    Setting::Handle<bool> firstRun { Settings::firstRun, true };
-    if (firstRun.get()) {
-        qCDebug(scriptengine) << "This is a first run...";
-        // clear the scripts, and set out script to our default scripts
-        clearScripts();
-        loadDefaultScripts();
```
Atlante45: This is now down by the default value of the setting

libraries/script-engine/src/ScriptEngines.cpp
```diff
@@ -280,67 +271,72 @@ QVariantList ScriptEngines::getRunning() {
...
+        }
+    }
 }
 
 void ScriptEngines::saveScripts() {
+    // Do not save anything if we are in the process of shutting down
+    if (qApp->closingDown()) {
+        qWarning() << "Trying to save scripts during shutdown.";
+        return;
+    }
```
Atlante45: Don't save anything while the app is shutting down and unloading all the scripts.

libraries/script-engine/src/ScriptEngines.cpp
```diff
@@ -280,67 +271,72 @@ QVariantList ScriptEngines::getRunning() {
...
+
     // Saves all currently running user-loaded scripts
-    Settings settings;
-    settings.beginWriteArray(SETTINGS_KEY);
-    settings.remove("");
```
Atlante45: This is where I suspect the issue was.
We were clearing the array before refilling it, but if we happen to crash before we actually fill it, we'd lose everything.

libraries/script-engine/src/ScriptEngines.cpp
```diff
@@ -280,67 +271,72 @@ QVariantList ScriptEngines::getRunning() {
     return result;
 }
 
-
-static const QString SETTINGS_KEY = "RunningScripts";
-
 void ScriptEngines::loadDefaultScripts() {
-    QUrl defaultScriptsLoc = defaultScriptsLocation();
```
birarda: Is `defaultScriptsLocation()` still defined somewhere and should be removed?

libraries/script-engine/src/ScriptEngines.cpp
```diff
@@ -280,67 +271,72 @@ QVariantList ScriptEngines::getRunning() {
...
-    Settings settings;
-    settings.beginWriteArray(SETTINGS_KEY);
-    settings.remove("");
-    settings.endArray();
+    // loads all saved scripts
+    auto runningScripts = runningScriptsHandle.get();
+    for (auto script : runningScripts) {
+        auto string = script.toString();
+        if (!string.isEmpty()) {
+            loadScript(string);
```
birarda: Won't this double load scripts that were run by the user before? We load them with `loadScript` above, which will put them into the setting once loaded?

libraries/script-engine/src/ScriptEngines.cpp
Atlante45: It is, it's used by other things to find the script directory.

libraries/script-engine/src/ScriptEngines.cpp
Atlante45: We set `foundDeprecatedSetting` and return early in that case.

PR #10534, opened 2017-05-23
Fix collision signal emission for static objects

zfox23:
Fixes the bug noted in [this FogBugz ticket](https://highfidelity.fogbugz.com/f/cases/3146/Xylophone-not-working-with-hands).

**Test Plan:**

*Prerequisites:*
- Machine A must have an installation of Sandbox A and Interface A.
- Machine A must have an HMD attached.
- Machine B must have an installation of Interface B.
- Machine B must have an HMD attached.

*Test:*
1. Run Sandbox A.
2. On Interface A, GoTo Sandbox A. Enter your HMD. Do not go to Sandbox A on Interface B.
3. On Interface A, rez the Xylophone by [running this script](http://mpassets.highfidelity.com/13dda83b-a286-404e-99d3-8b8ee2a4b03a-v1/xylophoneRezzer.js).
4. On Interface A, play the xylophone using your hand controllers.
    - Hitting a key with a mallet results in the texture of that key turning gray.
    - Hitting a key with a mallet results in a single sound being audible.
    - There isn't any perceptible latency between hitting a key with a mallet and hearing a sound.
    - Both of your hand controllers vibrate when hitting a key with a mallet.
5. On Interface B, GoTo Sandbox A. Enter your HMD.
6. From Interface A, play the xylophone using your hand controllers. Verify the following:
    - Hitting a key with a mallet results in the texture of that key turning gray for both clients.
        - On Interface A, there is no perceptible latency between hitting the key and seeing the texture change.
        - It is acceptable for there to be a small amount of latency between Avatar A hitting the key and Interface B seeing the texture change.
    - Hitting a key with a mallet results in a single sound being audible on both clients.
        - On Interface A, there is no perceptible latency between hitting the key and hearing the single sound.
        - It is acceptable for there to be a small amount of latency between Avatar A hitting the key and Interface B hearing the sound.
    - Both of HMD A's hand controllers vibrate when Interface A hits a key with a mallet
    - Neither of HMD B's hand controllers vibrate when Interface A hits a key with a mallet.
7. From Interface B, play the xylophone using your hand controllers. Verify the following:
    - Hitting a key with a mallet results in the texture of that key turning gray for both clients.
        - On Interface B, there is no perceptible latency between hitting the key and seeing the texture change.
        - It is acceptable for there to be a small amount of latency between Avatar B hitting the key and Interface A seeing the texture change.
    - Hitting a key with a mallet results in a single sound being audible on both clients.
        - On Interface B, there is no perceptible latency between hitting the key and hearing the single sound.
        - It is acceptable for there to be a small amount of latency between Avatar B hitting the key and Interface A hearing the sound.
    - Both of HMD B's hand controllers vibrate when Interface B hits a key with a mallet
    - Neither of HMD A's hand controllers vibrate when Interface B hits a key with a mallet.
8. Using "Create" mode on either Interface client, make the leftmost xylophone key "Dynamic". To do this, you'll first have to change its collision hull type to "Good". Note that doing this may cause the xylophone key to flicker strangely, but that's OK.
9. Using both Interface A and Interface B, leave Sandbox A. Interface B is done.
10. Using Interface A, re-connect to Sandbox A.
11. Using the mallet, hit the leftmost xylophone key. Verify that the sound plays the very first time you hit the key.

### DIFF
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -1108,7 +1108,7 @@ void EntityTreeRenderer::entityCollisionWithEntity(const EntityItemID& idA, cons
         assert(!bothEntitiesStatic);
 #endif
 
-        if ((myNodeID == entityASimulatorID && entityAIsDynamic) || (myNodeID == entityBSimulatorID && !entityAIsDynamic)) {
+        if ((myNodeID == entityASimulatorID && entityAIsDynamic) || (myNodeID == entityBSimulatorID && (!entityAIsDynamic || entityASimulatorID.isNull()))) {
             playEntityCollisionSound(entityA, collision);
             emit collisionWithEntity(idA, idB, collision);
             if (_entitiesScriptEngine) {
@@ -1116,7 +1116,7 @@ void EntityTreeRenderer::entityCollisionWithEntity(const EntityItemID& idA, cons
             }
         }
 
-        if ((myNodeID == entityBSimulatorID && entityBIsDynamic) || (myNodeID == entityASimulatorID && !entityBIsDynamic)) {
+        if ((myNodeID == entityBSimulatorID && entityBIsDynamic) || (myNodeID == entityASimulatorID && (!entityBIsDynamic || entityBSimulatorID.isNull()))) {
             playEntityCollisionSound(entityB, collision);
             // since we're swapping A and B we need to send the inverted collision
             Collision invertedCollision(collision);
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -1092,24 +1092,29 @@ void EntityTreeRenderer::entityCollisionWithEntity(const EntityItemID& idA, cons
     // trigger scripted collision sounds and events for locally owned objects
     EntityItemPointer entityA = entityTree->findEntityByEntityItemID(idA);
     EntityItemPointer entityB = entityTree->findEntityByEntityItemID(idB);
-    QUuid entityASimulatorID = entityA->getSimulatorID();
-    QUuid entityBSimulatorID = entityB->getSimulatorID();
-    if ((bool)entityA && (myNodeID == entityASimulatorID || ((bool)entityB && myNodeID == entityBSimulatorID && entityASimulatorID.isNull()))) {
-        playEntityCollisionSound(entityA, collision);
-        emit collisionWithEntity(idA, idB, collision);
-        if (_entitiesScriptEngine) {
-            _entitiesScriptEngine->callEntityScriptMethod(idA, "collisionWithEntity", idB, collision);
-        }
-    }
+    if ((bool)entityA && (bool)entityB) {
+        QUuid entityASimulatorID = entityA->getSimulatorID();
+        QUuid entityBSimulatorID = entityB->getSimulatorID();
+        bool entityAIsStatic = !entityA->getDynamic() && !entityA->isMoving();
+        bool entityBIsStatic = !entityB->getDynamic() && !entityB->isMoving();
 
-    if ((bool)entityB && (myNodeID == entityBSimulatorID || ((bool)entityA && myNodeID == entityASimulatorID && entityBSimulatorID.isNull()))) {
-        playEntityCollisionSound(entityB, collision);
-        // since we're swapping A and B we need to send the inverted collision
-        Collision invertedCollision(collision);
-        invertedCollision.invert();
-        emit collisionWithEntity(idB, idA, invertedCollision);
-        if (_entitiesScriptEngine) {
-            _entitiesScriptEngine->callEntityScriptMethod(idB, "collisionWithEntity", idA, invertedCollision);
+        if (myNodeID == entityASimulatorID || (myNodeID == entityBSimulatorID && entityAIsStatic)) {
+            playEntityCollisionSound(entityA, collision);
+            emit collisionWithEntity(idA, idB, collision);
+            if (_entitiesScriptEngine) {
+                _entitiesScriptEngine->callEntityScriptMethod(idA, "collisionWithEntity", idB, collision);
+            }
+        }
+
+        if (myNodeID == entityBSimulatorID || (myNodeID == entityASimulatorID && entityBIsStatic)) {
+            playEntityCollisionSound(entityB, collision);
+            // since we're swapping A and B we need to send the inverted collision
+            Collision invertedCollision(collision);
+            invertedCollision.invert();
+            emit collisionWithEntity(idB, idA, invertedCollision);
+            if (_entitiesScriptEngine) {
+                _entitiesScriptEngine->callEntityScriptMethod(idB, "collisionWithEntity", idA, invertedCollision);
+            }
         }
     }
 }
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -1095,10 +1095,20 @@ void EntityTreeRenderer::entityCollisionWithEntity(const EntityItemID& idA, cons
     if ((bool)entityA && (bool)entityB) {
         QUuid entityASimulatorID = entityA->getSimulatorID();
         QUuid entityBSimulatorID = entityB->getSimulatorID();
-        bool entityAIsStatic = !entityA->getDynamic() && !entityA->isMoving();
-        bool entityBIsStatic = !entityB->getDynamic() && !entityB->isMoving();
+        bool entityAIsDynamic = entityA->getDynamic();
+        bool entityBIsDynamic = entityB->getDynamic();
 
-        if (myNodeID == entityASimulatorID || (myNodeID == entityBSimulatorID && entityAIsStatic)) {
+#ifdef WANT_DEBUG
+        bool bothEntitiesStatic = !entityAIsDynamic && !entityBIsDynamic;
+        if (bothEntitiesStatic) {
+            qCDebug(entities) << "A collision has occurred between two static entities!";
+            qCDebug(entities) << "Entity A ID:" << entityA->getID();
+            qCDebug(entities) << "Entity B ID:" << entityB->getID();
+        }
+        assert(!bothEntitiesStatic);
+#endif
+
+        if ((myNodeID == entityASimulatorID && entityAIsDynamic) || (myNodeID == entityBSimulatorID && !entityAIsDynamic)) {
             playEntityCollisionSound(entityA, collision);
             emit collisionWithEntity(idA, idB, collision);
             if (_entitiesScriptEngine) {
@@ -1106,7 +1116,7 @@ void EntityTreeRenderer::entityCollisionWithEntity(const EntityItemID& idA, cons
             }
         }
 
-        if (myNodeID == entityBSimulatorID || (myNodeID == entityASimulatorID && entityBIsStatic)) {
+        if ((myNodeID == entityBSimulatorID && entityBIsDynamic) || (myNodeID == entityASimulatorID && !entityBIsDynamic)) {
             playEntityCollisionSound(entityB, collision);
             // since we're swapping A and B we need to send the inverted collision
             Collision invertedCollision(collision);
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -1091,15 +1091,18 @@ void EntityTreeRenderer::entityCollisionWithEntity(const EntityItemID& idA, cons
 
     // trigger scripted collision sounds and events for locally owned objects
     EntityItemPointer entityA = entityTree->findEntityByEntityItemID(idA);
-    if ((bool)entityA && myNodeID == entityA->getSimulatorID()) {
+    EntityItemPointer entityB = entityTree->findEntityByEntityItemID(idB);
+    QUuid entityASimulatorID = entityA->getSimulatorID();
+    QUuid entityBSimulatorID = entityB->getSimulatorID();
+    if ((bool)entityA && (myNodeID == entityASimulatorID || ((bool)entityB && myNodeID == entityBSimulatorID && entityASimulatorID.isNull()))) {
         playEntityCollisionSound(entityA, collision);
         emit collisionWithEntity(idA, idB, collision);
         if (_entitiesScriptEngine) {
             _entitiesScriptEngine->callEntityScriptMethod(idA, "collisionWithEntity", idB, collision);
         }
     }
-    EntityItemPointer entityB = entityTree->findEntityByEntityItemID(idB);
-    if ((bool)entityB && myNodeID == entityB->getSimulatorID()) {
+
+    if ((bool)entityB && (myNodeID == entityBSimulatorID || ((bool)entityA && myNodeID == entityASimulatorID && entityBSimulatorID.isNull()))) {
         playEntityCollisionSound(entityB, collision);
         // since we're swapping A and B we need to send the inverted collision
         Collision invertedCollision(collision);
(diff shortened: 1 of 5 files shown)


libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -1091,15 +1091,18 @@ void EntityTreeRenderer::entityCollisionWithEntity(const EntityItemID& idA, cons
 
     // trigger scripted collision sounds and events for locally owned objects
     EntityItemPointer entityA = entityTree->findEntityByEntityItemID(idA);
-    if ((bool)entityA && myNodeID == entityA->getSimulatorID()) {
+    EntityItemPointer entityB = entityTree->findEntityByEntityItemID(idB);
+    QUuid entityASimulatorID = entityA->getSimulatorID();
+    QUuid entityBSimulatorID = entityB->getSimulatorID();
+    if ((bool)entityA && (myNodeID == entityASimulatorID || ((bool)entityB && myNodeID == entityBSimulatorID && entityASimulatorID.isNull()))) {
```
AndrewMeadows: I looked at the logic again and believe it is good.  I will paraphrase it for other reviewers:

We want to fire a collision event on ObjectA when:
(a) We own the simulation for ObjectA OR
(b) We own the simulation for ObjectB AND ObjectA is not owned by anyone  (this is NEW!)

... and symmetrically for firing collision events on ObjectB (below)

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -414,15 +414,17 @@ QUuid EntityScriptingInterface::editEntity(QUuid id, const EntityItemProperties&
...
             // make sure the properties has a type, so that the encode can know which properties to include
             properties.setType(entity->getType());
-            bool hasTerseUpdateChanges = properties.hasTerseUpdateChanges();
-            bool hasPhysicsChanges = properties.hasMiscPhysicsChanges() || hasTerseUpdateChanges;
-            if (_bidOnSimulationOwnership && hasPhysicsChanges) {
+            bool hasMiscPhysicsChanges = properties.hasMiscPhysicsChanges();
+            bool hasDynamicsChanges = properties.hasDynamicPhysicsChanges();
+            // _bidOnSimulationOwnership is set per-instance of the scripting interface.
+            // It essentially corresponds to "Am I an AC or an Interface client?" - ACs will never bid.
+            if ((_bidOnSimulationOwnership && ((hasMiscPhysicsChanges && entity->isMoving()) || hasDynamicsChanges))) {
```
AndrewMeadows: This is fixing the other bug where passers by would bid for ownership on the xylophone keys.  A bid for ownership would go out when changing any physical property, even when the object was static (e.g. setting the dimension of a static box).  This PR makes the bid only go out when changing (a) any physics property on a moving object OR (b) when changing a dynamics property.

zfox23:
Thanks for those clarifying comments, @AndrewMeadows.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/119/HighFidelity-Beta-PR10534-5101bbff63bb0d43302d5ac17ee2ba13cf644e66.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/119/HighFidelity-Beta-PR10534-5101bbff63bb0d43302d5ac17ee2ba13cf644e66.dmg

ZappoMan:
> Question: Should this be a packet version change? Clients without this change will still bid on physics ownership when they shouldn't, and that's very undesirable. So, I'm leaning towards "yes".

To answer this question, we employ the Rosedale Doctrine - any change to a client or server that would appear as a bug when the clients or servers are mismatched, even when that change does not change the wire format, should still require an upgrade of both client and server.

So the answer is yes. This is a "protocol change".

libraries/entities/src/EntityScriptingInterface.cpp
ZappoMan: The goal of this section of the code should be: Static objects should never have simulations owners. If this is achieved, then the code above in `EntityTreeRenderer.cpp` will correctly fire the collision events for the simulator of the dynamic object hitting the static object.

This code as written does not appear to guarantee that static objects don't have simulation owners.

ZappoMan: Additionally - in the case where an object was dynamic and is edited to become static, I believe we need to clear simulation ownership. I don't see that happening here.

zfox23:
Allow me to publicly echo some of @AndrewMeadows' ideas:
- The changes in this PR will severely reduce the cases under which a client bids for physics ownership of a static object - rare, but not impossible. @ZappoMan: When you say "This code as written does not appear to guarantee that static objects don't have simulation owners." - Andrew agrees that this is true, but does not think personally that fact should prevent a merge of this PR.
- Zappoman, you say: "Additionally - in the case where an object was dynamic and is edited to become static, I believe we need to clear simulation ownership. I don't see that happening here." Andrew thinks that this clearing happens elsewhere in the code. We can test for that.
- Andrew will look for the place in the code where we explicitly set ownership on entity. Then, we can make the change to ensure that ownership is not set when the object is static. However, again, he does not think that needs to be in this PR.

Next, here are some things I know I need to do for a more complete PR:
1. Clean up the xylophone script by removing the explicit set of the key dimensions during preload. Interface with Caitlyn to ensure that script gets updated on the Marketplace.
2. Write tests to verify that static objects do not have physics ownership.
3. Make this change a protocol change.

ZappoMan:
> Andrew will look for the place in the code where we explicitly set ownership on entity. Then, we can make the change to ensure that ownership is not set when the object is static. However, again, he does not think that needs to be in this PR.

Why not this PR? This "bug" has existed for a while (more than a year) -- if there are other cases where a static object can have a simulation owner this bug will continue to manifest... and we'll be back at it. Seems like we agree the collisions will again fail if a static object ever ends up with a simulation owner... so we should fix that as part of this fix. That's the root bug.

zfox23:
@AndrewMeadows has taken on the responsibility of finding/fixing the "bug" described above ("I don't know where the code is right now; I will look for it and fix this"). He will either clone this branch and make a new PR (rendering this one defunct), or give me a patch.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/123/HighFidelity-Beta-PR10534-71e0966848c9df0532926d17564199a90748a3c2.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/123/HighFidelity-Beta-PR10534-71e0966848c9df0532926d17564199a90748a3c2.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/130/HighFidelity-Beta-PR10534-2042aae667a43644f87b320ffeef2c028695a004.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/130/HighFidelity-Beta-PR10534-2042aae667a43644f87b320ffeef2c028695a004.dmg

ZappoMan:
I liked this PR, but I realized it was fragile with respect to simulation ownership of static objects. If the system allows someone to be the simulation owner of a static object, the new logic will fail... I asked Zach to look into making sure we never allow static objects to have simulation owners... and that resulted in a tangled web of other questions/issues...

So in stepping back, let me offer the following...

First some terms:

*static object* - this is the object like the key in the xyolophone... this is the type of object that has undefined behavior in the current client.

*colliding object* - I'm using this term, because dynamic object is overloaded and might not always be the case... but the colliding object is an object which is being held by the avatar and is striking the static object... the mallet in this case. We know that this object will always have a client that is the authoritative simulation owner. (edited)

*script executing client* - this is the client that will execute the collision script. We want there to only be one of these, and we want it to be logically chosen, there should be reason and consistency for why it's chosen.

the *script executing client* - in the case of a *colliding object* is working well, and fits our goal of being logically and consistently chosen, namely it is always the client that is the simulation owner. the root problem is that for *static objects* - we aren't consistently picking a *script executing client*... that's what we're trying to fix.

we really don't care about simulation ownership of static objects...


because other than this issue with *script executing clients* - there don't appear to be any negative side effects to static objects having flip-flopping simulation owners...

so *maybe* - the better fix here is something like....

in `EntityTreeRenderer::entityCollisionWithEntity()`
psuedo code...

```
bool myClientOwnsCollidingA = entityA->isDynamic() && (myNodeID == entityASimulatorID);
bool myClientOwnsCollidingB = entityB->isDynamic() && (myNodeID == entityBSimulatorID);
bool myClientOwnsStaticA = entityA->isStatic() && myClientOwnsCollidingB;
bool myClientOwnsStaticB = entityB->isStatic() && myClientOwnsCollidingA;
```

and then the emit signal logic is driven off of this...
I believe this will handle the cases of...
- my client owns 2 different colliding objects
- my client owns 1 colliding object that's hitting a static object
- my client owns 1 colliding object... that's hitting another colliding object owned by another client

zfox23:
I like the idea of driving the signal emission off of entity motion type (static vs not static) rather than basing it off of whether or not an entity has a simulation owner. I think this because we have realized that, as the code stands currently, static entities are not protected against having a simulation owner.

I will write some new code and update this PR.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -1091,22 +1091,30 @@ void EntityTreeRenderer::entityCollisionWithEntity(const EntityItemID& idA, cons
...
-        // since we're swapping A and B we need to send the inverted collision
-        Collision invertedCollision(collision);
-        invertedCollision.invert();
-        emit collisionWithEntity(idB, idA, invertedCollision);
-        if (_entitiesScriptEngine) {
-            _entitiesScriptEngine->callEntityScriptMethod(idB, "collisionWithEntity", idA, invertedCollision);
+    if ((bool)entityA && (bool)entityB) {
+        QUuid entityASimulatorID = entityA->getSimulatorID();
+        QUuid entityBSimulatorID = entityB->getSimulatorID();
+        bool entityAIsStatic = !entityA->getDynamic() && !entityA->isMoving();
```
ZappoMan: hmmm I'm not sure you want the `!isMoving`

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/143/HighFidelity-Beta-PR10534-f37010532a70d713b766bd1cdf466f63117fa43e.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/143/HighFidelity-Beta-PR10534-f37010532a70d713b766bd1cdf466f63117fa43e.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/147/HighFidelity-Beta-PR10534-22dcc5a49d5fff3ee7cadeb1c8164bfc178495dc.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/147/HighFidelity-Beta-PR10534-22dcc5a49d5fff3ee7cadeb1c8164bfc178495dc.dmg

zfox23:
More hiccups:

There is an important component that Brad and I were missing that renders my new code (committed this morning) incorrect. That component is: _nearGrabbed objects are not necessarily dynamic; they are kinematic_.

Because of that, there are cases in the latest code here under which collision events from both entities A and B _won't_ fire when they should.

I just finished a conversation with @AndrewMeadows, and he thinks that the code as of commit `1e55c852a8764105e3fcf626241045184ce2f5cb` is the correct direction, and we should go back to work on preventing static objects from getting simulation ownership.

To expand: The pseudocode for firing a collision event on Entity A should be:
```
if ((we own entity A) OR ((we own entity B) AND (we're reasonably sure nobody else will fire on Entity A, i.e. Entity A has no owner))) {
    (emit collision event signal on entity A);
}
```
which is what was there before this morning.

zfox23:
As per a discussion between @ZappoMan, @AndrewMeadows, @sethalves, and myself, I have a direction forward for this PR.

Of note is this documentable fact:
Two entities, both of which have their "Dynamic" checkbox unchecked, cannot collide with each other. Neither of the entities will trigger collision events.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/153/HighFidelity-Beta-PR10534-39bcbd56db96e7f36254a958a32185d6112ad459.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/153/HighFidelity-Beta-PR10534-39bcbd56db96e7f36254a958a32185d6112ad459.dmg

zfox23:
There is a case here that @AndrewMeadows pointed out where this code is broken:

1. There exists two dynamic entities (i.e. the "Dynamic" checkbox in the Entity Properties page is _checked_). They are at rest; nobody owns them.
2. Avatar A picks up Entity 1, thus taking ownership of it. Entity 2 is still un-owned.
3. Avatar A smacks Entity 2 with Entity 1.
4. Entity 1 emits the collision signal. Entity 2 should, but does not.

ZappoMan:
As per the chat between @zfox23 @AndrewMeadows and me... here's the proposed working solution.

```
if ((weOwnA && AisDynamic) || (weOwnB && (AisNotDynamic || (AIsDynamic && AisUnowned)) {
    fireEventOnA();
}
```

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/175/HighFidelity-Beta-PR10534-e5e0a07ff65a2ad636e95cc4dd9c92dc5967ee88.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/175/HighFidelity-Beta-PR10534-e5e0a07ff65a2ad636e95cc4dd9c92dc5967ee88.dmg

Nex-Pro:
@zfox23 Step 7 needs clarification. Is it that interface B doesn't vibrate when interface A strikes a key?

zfox23:
@Nex-Pro That's correct. Sec...

Nex-Pro:
Testing this to destruction.

Nex-Pro:
Awaiting fresh build as discussed.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/183/HighFidelity-Beta-PR10534-f2fab5718735e5ffce04dc8fca8a9cf6612514da.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/183/HighFidelity-Beta-PR10534-f2fab5718735e5ffce04dc8fca8a9cf6612514da.dmg

Nex-Pro:
@zfox23 

### FAIL
 Went thru all steps in exact sequence.
All good up until setting the leftmost key to be good and dynamic.
At that point the key sounded a note then shortly after disappeared. It seemed to return momentarily a couple of times but then it was gone.
Leaving and returning to sandbox  revealed the Xylophone with missing leftmost key.

feedback. We independently thought that it was strange/wrong that there should be a haptic pulse in each hammer even when only one is being used. This makes no sense and feels wrong.

Plays well !

zfox23:
@Nex-Pro:

Fortunately or unfortunately, that failure isn't part of this PR...this is a classic case of the test writer not writing a good enough test. 😢 I was kind of afraid of the test plan failing at that part...I'll chat with you on Slack about it.

Regarding your feedback: I agree that is wrong, but that's a limitation of the xylophone marketplace script, not this PR or our code. 😄

Nex-Pro:
Ok !   I do like to get goot test notes. It is important. @zfox23

PR #10552, opened 2017-05-25
Fix SendQueue not setting lastReceiverResponse when recv handshake ack

huffman:


huffman:
This is a small networking change that fixes an issue when it takes > 5 seconds to receive a handshake ack. This makes sure that once you receive that ack you don't immediately timeout and drop your SendQueue.

To test:
 1. Go to your local sandbox with a fresh cache. Everything should load. Try doing this 3 times.
 1. Go to Welcome. Make sure you can see other avatars.
 1. Go back to your local sandbox. Everything should load.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/184/HighFidelity-Beta-PR10552-f0c3a0ac2876c984900572d7123f89cdb985347e.exe

libraries/networking/src/udt/SendQueue.cpp
```diff
@@ -241,6 +241,9 @@ void SendQueue::handshakeACK(SequenceNumber initialSequenceNumber) {
             std::lock_guard<std::mutex> locker { _handshakeMutex };
             _hasReceivedHandshakeACK = true;
         }
+
+        _lastReceiverResponse = QDateTime::currentMSecsSinceEpoch();
```
ZappoMan: any chance we could use a `std::` function for this?

huffman: We could, but we use currentMSecsSinceEpoch in other areas of the code. I can make that change if we think it is important, but I would prefer doing it as a separate PR.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/184/HighFidelity-Beta-PR10552-f0c3a0ac2876c984900572d7123f89cdb985347e.dmg

Nex-Pro:
Checking this

libraries/networking/src/udt/SendQueue.cpp
ZappoMan: ok - but yes... where ever possible I'd like us to move toward stl instead of qt in cases like this

Nex-Pro:
@Huffman seems all ok.   
Apart from loading at welcome was very slow, took a couple of minutes to see the other avatars. Don't think it is down to this PR. 
Lemme know if you want logs.

huffman:
Thanks @Nex-Pro!

libraries/networking/src/udt/SendQueue.cpp
huffman: Agreed

PR #10553, opened 2017-05-25
Do not get stuck in floor

howard-stearns:
Avoids getting stuck in the floor in several ways, per https://highfidelity.fogbugz.com/f/cases/3534/Fix-people-getting-stuck-in-floor and subcases.

Test plan:
1. When you explicitly go to a troublesome location (e.g., locations from 5, below using the address bar), this PR should pretty reliably keep you safe: either the location is ok without getting stuck, or if you might get stuck it should put you at some safe place above that point.
2. This PR will also try to detect some cases of being stuck, and get you out of it as above. One way to test this by turning off Avatar => "Enable Avatar Collisions" and going to the same troublesome locations as in 1. You won't get pushed elsewhere while collisions are off. But if you then turn "Enable Avatar Collisions" on, this code will frequently recognize that you're now in trouble and move you away. This is not tied to the menu item, but rather to a detector that is running all the time. So you can also trigger the behavior in other ways, such as editing physical objects into existence that squish you into the floor.
3. Try 1 and 2 with a variety of avatars.
4. When you import a mesh and are given the option of defining the collider shape in various ways, wherever it tells you that exact mesh shells should not be used for dynamic objects, it should also now tell you to not use it for floors.
5. To find places where you might get stuck, there are some in the subcases of the ticket above, and some specific urls listed in a comment below with the text "Example screw cases". The file http://howard-stearns.github.io/models/torture.json can also be imported for testing. It defines two red boxes, between which there is not enough room for a normal-sized avatar to stand. (This is currently set up with this point at howard/100,100,100.) If you go there, you will reliably get stuck (shuddering).  There is then a green box that intersects the upper red box.   If you evaluate MyAvatar.safeLanding(a point between the two red boxes), you will end up on top of the green box.  There is also a purple box above that, but there is plenty of room to stand between the green and purple.)


hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/187/HighFidelity-Beta-PR10553-ac8cb6341928c1d5e728889b46ddb8af35f7ef97.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/187/HighFidelity-Beta-PR10553-ac8cb6341928c1d5e728889b46ddb8af35f7ef97.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/190/HighFidelity-Beta-PR10553-905b5261d1a57bc1051b76d93735db487cabf58e.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/190/HighFidelity-Beta-PR10553-905b5261d1a57bc1051b76d93735db487cabf58e.dmg

hyperlogic:
This is not a code review, just feedback.  I'm having trouble understanding the algorithm here.  Part of my problem is not lack of comments but the use of too many lambdas.  IMHO this code would be easier to follow if it instead used static functions (for things like isUp) or const methods for pure functions that close over "this".

Second, I'm not sure we should be using the entity tree to perform these line probes.  The visual mesh, is quite different from the static mesh and hulls that the character capsule actually collides with, so we should compare "apples to apples".

Third, highly doubt this code is thread safe, I would rather see the main body of this code run on the main thread via a blocking queued invokeMethod.

One minor nit, the capsule offset is in avatar frame not in world frame, so you'll need to account for that.

hyperlogic:
FWIW, you can directly cast rays against the physics simulation collision on the main thread.  See `CharacterGhostObject::rayTest()`

ZappoMan:
@howard-stearns @hyperlogic - to Tony's point about not using the visual mesh vs the collision mesh, I think he's correct about that, but let me rephrase. The Entity Tree may be the correct location to do this probing, it is after all the ground truth of what is around you, however - you would need to use the collision hulls for the entities not the visual meshes. And of course, you'd need to do filtering out of those entities that are "collision-less". This is currently not exposed but certainly could be. 

It may be that there's an existing more direct route to the collidable meshes/hulls in bullet. @AndrewMeadows  could give us more insight into whether or not there's a shortcut or improvement that could be had by directly querying bullet instead of the entity tree. 

Assuming that there isn't a "ray probe" available in bullet... or that you can't guarentee that bullet will have all of the collidable surfaces, then you could implement a version of EnttiyTree::findRayIntersection() that explicitly uses the collision meshes and accounts for the various collidable properties.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/223/HighFidelity-Beta-PR10553-92aeadd70f4477077d89066840c8b6b7ad5f876e.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/223/HighFidelity-Beta-PR10553-92aeadd70f4477077d89066840c8b6b7ad5f876e.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/248/HighFidelity-Beta-PR10553-07ea2e271bc1c0ca0e23b9768fe6b75a8b43fd78.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/248/HighFidelity-Beta-PR10553-07ea2e271bc1c0ca0e23b9768fe6b75a8b43fd78.dmg

murillodigital:
test this please

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/282/HighFidelity-Beta-PR10553-1e1dd3a633aca6aa4657832c8eafdbcea9d27865.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/282/HighFidelity-Beta-PR10553-1e1dd3a633aca6aa4657832c8eafdbcea9d27865.dmg

marko8904:
test this please

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/284/HighFidelity-Beta-PR10553-1e1dd3a633aca6aa4657832c8eafdbcea9d27865.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/284/HighFidelity-Beta-PR10553-1e1dd3a633aca6aa4657832c8eafdbcea9d27865.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/334/HighFidelity-Beta-PR10553-6b2e4c5abca650a627d53462aabe6c04f030a869.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/334/HighFidelity-Beta-PR10553-6b2e4c5abca650a627d53462aabe6c04f030a869.dmg

AndrewMeadows:
Bullet does have a `rayTest()` method that can be called against the `btCollisionWorld` (our `PhysicsEngine` class has a `btCollisionWorld`), however we should implement new API in PhysicsEngine to facilitate it use.  Some restrictions would be that the rayTest() shouldn't be called while the main thread is in the middle of Application::update() method which is where Bullet performs its stepSimulation() and where incoming/outgoing changes are handled.

Note: the `CharacterGhostObject` is an optimization used to help the "walk up steps" behavior.  The "ghost" feature in Bullet is basically a non-colliding object that maintains a list of overlapping objects that can provide a subset of the world objects for the purpose of doing `rayTests`.  Instead of testing a ray against ALL the objects in the world we can test against only those objects that overlap the "ghost".  We make the "ghost" follow the avatar around so we can reduce the cost of making large numbers of rayTests near the avatar.  One problem with using the ghost is that it only knows about objects that overlap it (and hence the avatar).  It does not necessarily know about all of the objects that might block the avatar's placement.

I've heard that Bullet supports "asynchronous collision detection", and by this I mean we can take a hypothetical object (or shape) then ask Bullet: "If I were to place this hypothetical thing here, what would it collide with?" and by "asynchronous" I mean we can ask that question outside of the main stepSimulation() call without screwing up any broadphase overlap state or physics dynamics.

Ultimately, if we want to do it with maxiumum success, we should use Bullet for evaluating rescue placement of the avatar.  The best way would probably be to expand the size of the "ghost" a bit to allow for rayTests() that are further out from the avatar (right now I think the "ghost's" shape is a box twice the size of the avatar's capsule) and to expose some API to allow the CharacterController to perform rayTests() against its current list of overlaps.

howard-stearns:
Example screw cases:
above the trouble in master: hifi://dev-demo/89.8525,-0.583215,42.1645/0,-0.223106,0,0.974794 
in model floor: hifi://dev-demo/89.8552,-1.91111,41.9404/0,-0.223106,0,0.974794
in lower floor hifi://dev-demo/89.9358,-3.98317,41.7584/0,-0.223103,0,0.974795
below floors: hifi://dev-demo/89.8525,-6.38928,42.1645/0,-0.223096,0,0.974796
should remain in tunnel, not on top: hifi://dev-demo/86.9668,-1.63248,52.1415/0,-0.0298017,0,0.999556
hifi://dev-playa/-17.835,-7.91472,-42.3409/0,0.64737,0,0.762176
into the small space between red blocks. should end up on top of green block: hifi://howard/100,100,100

howard-stearns:
good stuff, @AndrewMeadows. I've added your comments to https://highfidelity.fogbugz.com/f/cases/5003/findRayIntersection-has-option-to-use-collidableOnly-but-doesn-t-actually-use-colliders, as some poor slob (perhaps me) will need that info in looking at that ticket.

As I see this PR and it's ticket, we're trying to see if there are a set of appropriate heuristics that can be applied to avoid trouble and get us out of it (rather than the accuracy of the methods used by those heuristics, which is what 5003 is about).

interface/resources/qml/AssetServer.qml
```diff
@@ -177,7 +177,7 @@ ScrollingWindow {
                     SHAPE_TYPE_STATIC_MESH
                 ],
                 checkStateOnDisable: false,
-                warningOnDisable: "Models with 'Exact' automatic collisions cannot be dynamic"
+                warningOnDisable: "Models with 'Exact' automatic collisions cannot be dynamic, and should not be used as floors"
```
howard-stearns: These .qml wording changes are advising creative users not to use exact for floors. "Exact" shells don't work well with bullet, and shouldn't be used as floors. Zaru (dev-demo) has TWO exact shell floors that are close to each other, providing a terrific source of test cases!

interface/src/avatar/MyAvatar.cpp
```diff
@@ -294,6 +294,7 @@ void MyAvatar::simulateAttachments(float deltaTime) {
 QByteArray MyAvatar::toByteArrayStateful(AvatarDataDetail dataDetail) {
     CameraMode mode = qApp->getCamera().getMode();
     _globalPosition = getPosition();
+    // This might not be right! Isn't the capsule local offset in avatar space? -HR 5/26/17
```
howard-stearns: The notes here and on 459 don't have anything to do with this PR, except that Tony's comments might apply to this existing code as well. So, leaving a note for the next person to dive into that code.

libraries/render-utils/src/Model.cpp
```diff
@@ -332,7 +332,7 @@ void Model::initJointStates() {
```
howard-stearns: The algorithm in requiresSafeLanding requires that we be able to detect the exact thin shell that completely surrounds us. All the changes below are to allow findRayTriangleIntersection (in GeometryUtil.cpp) to find intersections with the backface of a mesh when used in precision picking by findRayIntersectionAgainstSubMeshes (in RenderableModelEntityItem.cpp).

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/406/HighFidelity-Beta-PR10553-05f41fb4f81c65db3687f4a518b720afc769e5c9.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/406/HighFidelity-Beta-PR10553-05f41fb4f81c65db3687f4a518b720afc769e5c9.dmg

howard-stearns:
test this please

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/409/HighFidelity-Beta-PR10553-05f41fb4f81c65db3687f4a518b720afc769e5c9.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/409/HighFidelity-Beta-PR10553-05f41fb4f81c65db3687f4a518b720afc769e5c9.exe

YuppyPlays:
Testing.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/414/HighFidelity-Beta-PR10553-ca526e07765cd316d91d7e50a87d1a9544a257de.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/414/HighFidelity-Beta-PR10553-ca526e07765cd316d91d7e50a87d1a9544a257de.dmg

YuppyPlays:
Testing.

YuppyPlays:
[Ran into some Permission problems, speaking to Dev via Slack]

YuppyPlays:
Testing complete:

Example Screw Cases, Character spawns in example places: **Fail** The location in Dev-playa caused issues with "Simple Robot" avatar.
Note: Could not connect to Howard place so recreated the conditions in my sandbox with the help of dev of the same name.

Test the torture.json between red boxes: **Pass** Spawns to top of green as expected. Location: hifi://plum-capacitor-1353/80,90,100
Test the torture.json inside the purple: **Conditional pass** There is a judder I had experienced before Howard joined me in my sandbox, however once he joined the shudder disappeared and worked flawlessly even after he had left. (Unsure why this is though) Location: hifi://plum-capacitor-1353/89,112,100

With multiple Avatars?: **Pass** Grey standard manequinn, Art3mis, simplerobot. All worked. 
Note: During Purple box test, very brief judder for new avatars but quickly fixed itself.



Please let me know if there is anything more I can do/provide to assist. :)

howard-stearns:
There's something very weird about the Simple Robot avatar. It as though it is much taller than twice it's capsule height. Investigating...

howard-stearns:
@YuppyPlays can you retest when this builds? I think the robot avatar should now work better.

YuppyPlays:
@howard-stearns No problem at all :) I shall test when built.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/435/HighFidelity-Beta-PR10553-3de243b093dd370fb2fdf00909bbb3f603590f32.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/435/HighFidelity-Beta-PR10553-3de243b093dd370fb2fdf00909bbb3f603590f32.dmg

YuppyPlays:
Testing complete:

I can confirm this has fixed the robot issue, it now spawns on top of the stair platform., **Pass**

interface/src/avatar/MyAvatar.cpp
```diff
@@ -2219,6 +2232,143 @@ void MyAvatar::goToLocation(const glm::vec3& newPosition,
...
+    if (QThread::currentThread() != thread()) {
+        bool result;
+        QMetaObject::invokeMethod(this, "safeLanding", Qt::BlockingQueuedConnection, Q_RETURN_ARG(bool, result), Q_ARG(const glm::vec3&, position));
+        return result;
+    }
+    glm::vec3 better;
+    if (!requiresSafeLanding(position, better)) {
+        return false;
+    }
+    qDebug() << "rechecking" << position << " => " << better << " collisions:" << getCollisionsEnabled() << " physics:" << qApp->isPhysicsEnabled();
```
ZappoMan: Did you intend to leave this in? If so, at least group it into a category. physics? avatar?

interface/src/avatar/MyAvatar.cpp
```diff
@@ -2219,6 +2232,143 @@ void MyAvatar::goToLocation(const glm::vec3& newPosition,
...
+    }
+
+    if (!findIntersection(capsuleCenter, down, lowerIntersection, lowerId, lowerNormal)) {
+        // Our head may be embedded, but our center is out and there's room below. See corresponding comment above.
+        return false; // nothing below
+    }
+ 
+    // See if we have room between entities above and below, but that we are not contained.
+    if (isDown(upperNormal) && isUp(lowerNormal)) {
+        // The surface above us is the bottom of something, and the surface below us it the top of something.
```
ZappoMan: personally - I think it reads better if this comment is above the if statement... since this comment describes what the control structure is testing.... but I won't consider this a blocker

interface/src/avatar/MyAvatar.cpp
```diff
@@ -2219,6 +2232,143 @@ void MyAvatar::goToLocation(const glm::vec3& newPosition,
...
+        // I.e., we are in a clearing between two objects.
+        auto delta = glm::distance(upperIntersection, lowerIntersection);
```
ZappoMan: personally I think `spaceBetween` would be more clear than `delta`... delta feels like it should be a "change" in something...

interface/src/avatar/MyAvatar.cpp
```diff
@@ -2219,6 +2232,143 @@ void MyAvatar::goToLocation(const glm::vec3& newPosition,
...
+        const float halfHeightFactor = 2.5f; // Until case 5003 is fixed (and maybe after?), we need a fudge factor. Also account for content modelers not being precise.
+        if (delta > (halfHeightFactor * halfHeight)) {
+            // There is room for us to fit in that clearing. If there wasn't, physics would oscilate us between the objects above and below.
+            // We're now going to iterate upwards through successive upperIntersections, testing to see if we're contained within the top surface of some entity.
+            // There will be one of two outcomes:
+            // a) We're not contained, so we have enough room and our position is good.
+            // b) We are contained, so we'll bail out of this but try again at a position above the containing entity.
+            for (;;) {
```
ZappoMan: how do we know this won't end up as an infinite loop? Should we have a fail-safe to always bail if we have searched 1000 times?

interface/src/avatar/MyAvatar.cpp
howard-stearns: good catch. removed

interface/src/avatar/MyAvatar.cpp
howard-stearns: done.

interface/src/avatar/MyAvatar.cpp
howard-stearns: done.

interface/src/avatar/MyAvatar.cpp
howard-stearns: done, with logging if failed

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/483/HighFidelity-Beta-PR10553-9cf027a68c3facf258f6fea15123f7675aae312e.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/483/HighFidelity-Beta-PR10553-9cf027a68c3facf258f6fea15123f7675aae312e.dmg

PR #10561, opened 2017-05-26
fix crash in EntityMotionState::handleDeactivation()

AndrewMeadows:


interface/src/Application.cpp
```diff
@@ -1682,6 +1682,10 @@ Application::~Application() {
 
     _physicsEngine->setCharacterController(nullptr);
 
+    // the _shapeManager should have zero references
+    _shapeManager.collectGarbage();
```
ZappoMan: will this slow down our shutdown times?

interface/src/Application.cpp
```diff
@@ -4481,12 +4485,13 @@ void Application::update(float deltaTime) {
...
                     PerformanceTimer perfTimer("handleOutgoingChanges");
-                    const VectorOfMotionStates& deactivations = _physicsEngine->getDeactivatedMotionStates();
-                    _entitySimulation->handleDeactivatedMotionStates(deactivations);
 
                     const VectorOfMotionStates& outgoingChanges = _physicsEngine->getChangedMotionStates();
                     _entitySimulation->handleChangedMotionStates(outgoingChanges);
                     avatarManager->handleChangedMotionStates(outgoingChanges);
+
+                    const VectorOfMotionStates& deactivations = _physicsEngine->getDeactivatedMotionStates();
+                    _entitySimulation->handleDeactivatedMotionStates(deactivations);
```
AndrewMeadows: This is the actual fix for the crash in `EntityMotionState::handleDeactivation()`.  The `PhysicsEngine::getChangedMotionStates()` call has a side-effect where it assembles the list of deactivated states which are now processed immediately (rather than waiting until next frame, at which point the list could have contained a pointer to a MotionState that had been deleted).

libraries/physics/src/PhysicsEngine.h
```diff
@@ -53,7 +53,7 @@ class PhysicsEngine {
     uint32_t getNumSubsteps();
 
     void removeObjects(const VectorOfMotionStates& objects);
-    void removeObjects(const SetOfMotionStates& objects); // only called during teardown
+    void removeSetOfObjects(const SetOfMotionStates& objects); // only called during teardown
```
AndrewMeadows: The rest of these changes were some sanity checking I was adding while trying to hunt down another crash.  I haven't been able to reproduce any violations in these checks but they should be added anyway.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/211/HighFidelity-Beta-PR10561-fa003b0c14bddfef711f175dcd629525ce3b2f29.exe

AndrewMeadows:
Now that I have a theory as to how the `EntityMotionState::handleDeactivation()` crash happens I can imagine a scenario that might be able to reproduce it on hifi/master.  I don't have time to implement the test script since I'm on vacation and appear to have forgotten my laptop battery but I'll describe it for the adventurous QA engineer:

Make a JS script that creates a static floor object (a big box that is about 30x1x30) and creates 100 small (0.2x0.2x0.2) dynamic boxes with gravity <0,-9,0>, and damping = 0.9, in a 10x10 grid 3m apart at rest just above the foor, such that none of the dynamic boxes overlap each ohter: each box has its own "simulation island" in the bullet physics simulation.

For each dynamic box:
(1) set its velocity to <0, 0.1, 0> to make it go "active"
(2) start a timer for 1.99 seconds
(3) delete the box at the end of the timer
(4) create a new dynamic box with gravity in its place
(5) Goto (1)

Set that script running and then make the Interface window run without focus so that its frame rate drops down to about 15 fps.  Wait for several minutes and hope that it crashes (in hifi/master... in theory this PR fixes the crash).

What we're trying to do is win a race condition where an object is deactivated in the physics simulation (e.g. come to rest and get set "inactive") right before it gets deleted, right before its deactivation is actually processed.  The low fps increases the odds that the race condition is close. Note: the physics simulation is configured to deactivate an object if its velocities are almost zero for at least 2.0 seconds, hence the 1.99 second timer.  Run 100 of these objects spaced out so that they don't influence each other, and replaced after they are deleted, and the odds of causing a crash will be increased.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/211/HighFidelity-Beta-PR10561-fa003b0c14bddfef711f175dcd629525ce3b2f29.dmg

interface/src/Application.cpp
AndrewMeadows: We still call `exit()` directly in `main()` (for Release build on non-linux platfirms) which bypasses the `Application` dtor which is where this code lives, so: No, it won't slow down anything we care about.

Even when `collectGarbage()` is called it should be fast.  The amount of memory used by the `ShapeManager` is much less than used by textures and mesh geometry.

OmegaHeron:
Testing

OmegaHeron:
Windows 10
QA Pass

Tested against master 6614 which exhibited crash on 2nd attempt.
This PR Windows 10 (i5 6500/GTX970)- 20 Runs no crash.
This PR Ubuntu 16.04 (i7-3720QM/GTX660M) - 20 Runs no crash.
_Note:  Did not smoke test against existing content in dev-welcome/dev-playa due to proto mismatch.  Wouldn't expect that to be a problem -- just in case FYI._

PR #10575, opened 2017-05-30
Fix showDialog to work without defaultScripts

zzmp:
- Updates Application::showDialog to work even if Tablet is not initialized via defaultScripts.js
- Changes behavior on showDialog to avoid closing an open dialog

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1522,12 +1522,12 @@ QString Application::getUserAgent() {
     return userAgent;
 }
 
-void Application::toggleTabletUI() const {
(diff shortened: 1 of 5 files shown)


zzmp:
### QA Plan

- Open the running scripts menu and make sure it works appropriately
  - Use the top-level menus to open it (on the top of the window, in Desktop mode)
  - Use the toolbar to open it
  - Use the tablet to open it
  - Use CTRL+J to open it
  - Open it with no defaultScripts running
  - Open it when it is already open (this used to close it, but now it will have no effect)

PR #10577, opened 2017-05-31
Changed the DEFAULT_DISCOVERABILITY_MODE to Connections

manapa01:
The default discoverability mode will be initialized to "Friends and Connections" instead of "Friends Only", fixing https://highfidelity.fogbugz.com/f/cases/4997

To Test:
-Reset System Settings
-Open up tablet and click on People
-Check that in right-hand corner that "Friends and Connections" is set

### DIFF
--- a/interface/src/DiscoverabilityManager.cpp
+++ b/interface/src/DiscoverabilityManager.cpp
@@ -25,7 +25,7 @@
 
 #include <QThread>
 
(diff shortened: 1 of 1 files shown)


hifi-gustavo:

manapa01:
will close and go again using a different branch

PR #10602, opened 2017-06-02
fixes for some simulation ownership problems

AndrewMeadows:
- faster simulation ownership bid for locally-created objects
- revoke ownership for non-moving static objects

### DIFF
--- a/libraries/physics/src/EntityMotionState.cpp
+++ b/libraries/physics/src/EntityMotionState.cpp
@@ -84,7 +84,7 @@ EntityMotionState::~EntityMotionState() {
 
 void EntityMotionState::updateServerPhysicsVariables() {
     assert(entityTreeIsLocked());
-    if (_entity->getSimulatorID() == Physics::getSessionUUID()) {
+    if (isLocallyOwned()) {
         // don't slam these values if we are the simulation owner
         return;
     }
@@ -135,7 +135,7 @@ void EntityMotionState::handleEasyChanges(uint32_t& flags) {
                 _nextOwnershipBid = usecTimestampNow() + USECS_BETWEEN_OWNERSHIP_BIDS;
             }
             _loopsWithoutOwner = 0;
-        } else if (_entity->getSimulatorID() == Physics::getSessionUUID()) {
+        } else if (isLocallyOwned()) {
             // we just inherited ownership, make sure our desired priority matches what we have
             upgradeOutgoingPriority(_entity->getSimulationPriority());
         } else {
@@ -315,7 +315,7 @@ bool EntityMotionState::isCandidateForOwnership() const {
     assert(_entity);
     assert(entityTreeIsLocked());
     return _outgoingPriority != 0
-        || Physics::getSessionUUID() == _entity->getSimulatorID()
+        || isLocallyOwned()
         || _entity->dynamicDataNeedsTransmit();
 }
 
@@ -489,7 +489,7 @@ bool EntityMotionState::shouldSendUpdate(uint32_t simulationStep) {
         return true;
     }
 
-    if (_entity->getSimulatorID() != Physics::getSessionUUID()) {
+    if (!isLocallyOwned()) {
         // we don't own the simulation
 
         // NOTE: we do not volunteer to own kinematic or static objects
@@ -597,7 +597,7 @@ void EntityMotionState::sendUpdate(OctreeEditPacketSender* packetSender, uint32_
         properties.clearSimulationOwner();
         _outgoingPriority = 0;
         _entity->setPendingOwnershipPriority(_outgoingPriority, now);
-    } else if (Physics::getSessionUUID() != _entity->getSimulatorID()) {
+    } else if (!isLocallyOwned()) {
         // we don't own the simulation for this entity yet, but we're sending a bid for it
         quint8 bidPriority = glm::max<uint8_t>(_outgoingPriority, VOLUNTEER_SIMULATION_PRIORITY);
         properties.setSimulationOwner(Physics::getSessionUUID(), bidPriority);
@@ -784,6 +784,10 @@ QString EntityMotionState::getName() const {
 void EntityMotionState::computeCollisionGroupAndMask(int16_t& group, int16_t& mask) const {
     assert(_entity);
     _entity->computeCollisionGroupAndFinalMask(group, mask);
+}
+
+bool EntityMotionState::isLocallyOwned() const {
+    return _entity->getSimulatorID() == Physics::getSessionUUID();
 }
 
 bool EntityMotionState::shouldBeLocallyOwned() const {
--- a/libraries/physics/src/EntityMotionState.cpp
+++ b/libraries/physics/src/EntityMotionState.cpp
@@ -65,8 +65,7 @@ EntityMotionState::EntityMotionState(btCollisionShape* shape, EntityItemPointer 
     _lastStep(0),
     _loopsWithoutOwner(0),
     _accelerationNearlyGravityCount(0),
-    _numInactiveUpdates(1),
-    _outgoingPriority(0)
+    _numInactiveUpdates(1)
 {
     _type = MOTIONSTATE_TYPE_ENTITY;
     assert(_entity);
@@ -75,6 +74,8 @@ EntityMotionState::EntityMotionState(btCollisionShape* shape, EntityItemPointer 
     // we need the side-effects of EntityMotionState::setShape() so we call it explicitly here
     // rather than pass the legit shape pointer to the ObjectMotionState ctor above.
     setShape(shape);
+
+    _outgoingPriority = _entity->getPendingOwnershipPriority();
 }
 
 EntityMotionState::~EntityMotionState() {
@@ -135,23 +136,23 @@ void EntityMotionState::handleEasyChanges(uint32_t& flags) {
                 _nextOwnershipBid = usecTimestampNow() + USECS_BETWEEN_OWNERSHIP_BIDS;
             }
             _loopsWithoutOwner = 0;
+            _numInactiveUpdates = 0;
         } else if (isLocallyOwned()) {
             // we just inherited ownership, make sure our desired priority matches what we have
             upgradeOutgoingPriority(_entity->getSimulationPriority());
         } else {
             _outgoingPriority = 0;
             _nextOwnershipBid = usecTimestampNow() + USECS_BETWEEN_OWNERSHIP_BIDS;
+            _numInactiveUpdates = 0;
         }
     }
     if (flags & Simulation::DIRTY_SIMULATION_OWNERSHIP_PRIORITY) {
-        // The DIRTY_SIMULATOR_OWNERSHIP_PRIORITY bits really mean "we should bid for ownership because
-        // a local script has been changing physics properties, or we should adjust our own ownership priority".
-        // The desired priority is determined by which bits were set.
-        if (flags & Simulation::DIRTY_SIMULATION_OWNERSHIP_FOR_GRAB) {
-            _outgoingPriority = SCRIPT_GRAB_SIMULATION_PRIORITY;
-        } else {
-            _outgoingPriority = SCRIPT_POKE_SIMULATION_PRIORITY;
-        }
+        // The DIRTY_SIMULATOR_OWNERSHIP_PRIORITY bit means one of the following:
+        // (1) we own it but may need to change the priority OR...
+        // (2) we don't own it but should bid (because a local script has been changing physics properties)
+        uint8_t newPriority = isLocallyOwned() ? _entity->getSimulationOwner().getPriority() : _entity->getSimulationOwner().getPendingPriority();
+        _outgoingPriority = glm::max(_outgoingPriority, newPriority);
+
         // reset bid expiry so that we bid ASAP
         _nextOwnershipBid = 0;
     }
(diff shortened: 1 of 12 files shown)


libraries/physics/src/EntityMotionState.cpp
```diff
@@ -135,23 +136,23 @@ void EntityMotionState::handleEasyChanges(uint32_t& flags) {
...
             _loopsWithoutOwner = 0;
-        } else if (_entity->getSimulatorID() == Physics::getSessionUUID()) {
+            _numInactiveUpdates = 0;
+        } else if (isLocallyOwned()) {
             // we just inherited ownership, make sure our desired priority matches what we have
             upgradeOutgoingPriority(_entity->getSimulationPriority());
         } else {
             _outgoingPriority = 0;
             _nextOwnershipBid = usecTimestampNow() + USECS_BETWEEN_OWNERSHIP_BIDS;
+            _numInactiveUpdates = 0;
```
AndrewMeadows: Clearing `_numInactiveUpdates` when releasing ownership helped fix a problem where the interface would sometimes take a while (one or two seconds) to actually release ownership of static objects for some cases.  Basically some state from the previous ownership event would leak into the subsequent ownership and would influence the logic flow based on old data.  This change clears that state.

libraries/physics/src/EntityMotionState.cpp
```diff
@@ -135,23 +136,23 @@ void EntityMotionState::handleEasyChanges(uint32_t& flags) {
...
-        if (flags & Simulation::DIRTY_SIMULATION_OWNERSHIP_FOR_GRAB) {
-            _outgoingPriority = SCRIPT_GRAB_SIMULATION_PRIORITY;
-        } else {
-            _outgoingPriority = SCRIPT_POKE_SIMULATION_PRIORITY;
-        }
+        // The DIRTY_SIMULATOR_OWNERSHIP_PRIORITY bit means one of the following:
+        // (1) we own it but may need to change the priority OR...
+        // (2) we don't own it but should bid (because a local script has been changing physics properties)
+        uint8_t newPriority = isLocallyOwned() ? _entity->getSimulationOwner().getPriority() : _entity->getSimulationOwner().getPendingPriority();
+        _outgoingPriority = glm::max(_outgoingPriority, newPriority);
```
AndrewMeadows: Here we are picking up the correct ownership priority hint (e.g. supplied by `EntityScriptingInterface`) when deciding to bid on something that we don't own yet.

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -174,19 +177,9 @@ void PhysicsEngine::removeObjects(const VectorOfMotionStates& objects) {
...
-                        // swap with rbegin
-                        *itr = *(_activeStaticBodies.rbegin());
-                    }
-                    _activeStaticBodies.pop_back();
-                    break;
-                }
-                ++itr;
+            std::set<btRigidBody*>::iterator itr = _activeStaticBodies.find(object->getRigidBody());
+            if (itr != _activeStaticBodies.end()) {
+                _activeStaticBodies.erase(itr);
```
AndrewMeadows: `PhysicsEngine::_activeStaticBodies` switched from a vector to a set so that we could add elements in new places and not end up with duplicate entries.

Elsewhere, after a successful `stepSimulation()` we now add the elements in `_activeStaticBodies` to the list of changed objects that might or might not need simulation ownership bids/revokes sent.  This helps us revoke ownership for objects that have suddenly been changed to static.

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -496,13 +491,23 @@ const CollisionEvents& PhysicsEngine::getCollisionEvents() {
...
     // so we must deactivate them ourselves
-    for (size_t i = 0; i < _activeStaticBodies.size(); ++i) {
-        _activeStaticBodies[i]->forceActivationState(ISLAND_SLEEPING);
+    std::set<btRigidBody*>::const_iterator itr = _activeStaticBodies.begin();
+    while (itr != _activeStaticBodies.end()) {
+        btRigidBody* body = *itr;
+        body->forceActivationState(ISLAND_SLEEPING);
+        ObjectMotionState* motionState = static_cast<ObjectMotionState*>(body->getUserPointer());
+        if (motionState) {
+            _dynamicsWorld->addChangedMotionState(motionState);
```
AndrewMeadows: This is where we add the contents of **_activeStaticBodies** to the outgoing changes.

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -129,6 +129,9 @@ void PhysicsEngine::addObjectToDynamicsWorld(ObjectMotionState* motionState) {
             }
             body->setCollisionFlags(btCollisionObject::CF_STATIC_OBJECT);
             body->updateInertiaTensor();
+            if (motionState->isLocallyOwned()) {
+                _activeStaticBodies.insert(body);
+            }
```
AndrewMeadows: Whenever adding or re-adding a static object to the world we make sure it is known to be inside `_activeStaticBodies` so that its ownership status will be addressed later.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/340/HighFidelity-Beta-PR10602-29a05d7eb4832102b8e6bf65f6a41514c288e63f.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/340/HighFidelity-Beta-PR10602-29a05d7eb4832102b8e6bf65f6a41514c288e63f.dmg

AndrewMeadows:
Test Plan:

(1) Copy the script below to local file.  When loaded it will create a temporary moving dynamic object and set it static after about 1 second of motion.

(2) Login and enable  **Developer --> Physics --> Highlight Simulation Ownership**

(3) Run the script from (1) and notice the simulation status icons.  They are from right to left:
(a) Red arrow pointing up = incoming data from entity-server
(b) Cyan arrow pointing down = outbound data to entity-server
(c) Blue circle with triangle = object is **active** in physics engine
(d) Blue (Red) circle with head/shoulder silhouette = you own (someone else owns) the object's simulation
 
--> In this PR the blue ownership icon (d) should show up immediately after creating the dynamic object.
--> Hifi/master may take about 1 second to eventually show the blue ownership icon.

(4) Note that the dynamic object is set static after about 1 second and has a lifetime of only 10 seconds.  You can reload the script to delete the current object and launch a new one.  If you reload the script and then promptly activate some other window such that Interface runs out of focus then the FPS will drop to about 15 and in hifi/master (not this PR) you will be more likely to achieve a situation where the ownership is grabbed AFTER the object has become static, in which case the ownership will remain to you indefinitely.

--> In this PR the ownership will quickly be revoked for objects that go static.

(5) Use the **Create** script/UI to select a static object that you did not create during this login session (e.g. it is unowned by you) and adjust its position.

--> in hifi/master the object will become owned by you and will stay that way indefinitely
--> in this PR the object will be momentarily owned by you but that ownership will quickly be revoked

```
// create a dynamic object, let it move for a few seconds, then set it static

// some general constants
var ORIGIN = { x: 0.0, y: 0.0, z: 0.0 };
var IDENTITY_ROTATION = { x: 0.0, y: 0.0, z: 0.0, w: 1.0 };

// compute a position to create the object relative to avatar
var forward = Quat.getFront(MyAvatar.orientation);
var forwardOffset = Vec3.multiply(2.0, forward);
var objectPosition = Vec3.sum(MyAvatar.position, forwardOffset);

// the properties of the object
var BOX_WIDTH = 1.0; // meters
var BOX_HEIGHT = 0.5; // meters
var BOX_DEPTH = 1.5; // meters
var LIFETIME = 60; // seconds
var properties = {
    type: "Box",
    name: "fubar",
    dimensions: {x: BOX_WIDTH, y: BOX_HEIGHT, z: BOX_DEPTH},
    position: objectPosition,
    lifetime: LIFETIME, // just in case this script crashes before it can cleanup the object
    dynamic: true,
    damping: 0.01,
    angularDamping: 0.01,
    velocity: forward,
    angularVelocity: forward
};

// create the object
var entityId = Entities.addEntity(properties);

// set object static
stopEntity = function() {
    // set object static
    var staticProperties = {
        dynamic: false,
        velocity: { x:0, y:0, z:0 },
        angularVelocity: { x:0, y:0, z:0 }
    };
    Entities.editEntity(entityId, staticProperties);
}

// start a timer
Script.setTimeout(stopEntity, 1000);

// delete the object when this script is stopped
deleteEntity = function() {
    Entities.deleteEntity(entityId);
}
Script.scriptEnding.connect(deleteEntity);
```

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

"Blue ownership icon should show up immediately after creating the dynamic object"
**Pass**

"object is static after 1 second"
**Fail** - I tested this once and it worked but then retried after the lifetime didn't seem to be 10 seconds, it then kept drifting off and didn't de-spawn.

"lifetime of only 10 seconds"
**Fail** - This goes on for longer, around 20 seconds.

"reload script to delete current object and launch a new one"
**Pass**

"activating another window should NOT make the fps drop as it would in hifi/master"
**Fail** - It seems to still act the same as hifi/master by dropping FPS to around 15.

"you will be more likely to achieve a situation where the ownership is grabbed AFTER the object has become static, in which case the ownership will remain to you indefinitely."
**Fail** - Inconclusive due to the previous fails.

"In this PR the ownership will quickly be revoked for objects that go static."
**Fail** - Inconclusive due to the previous fails.

"in this PR the object will be momentarily owned by you but that ownership will quickly be revoked"
**Fail** - I appear to be able to manipulate the object with no revoke of permission to stop me, waited a few minutes and tried again, still allows me.

Please let me know if there is anything more I can do/provide :)

AndrewMeadows:
@YuppyPlays 

(A) I made a mistake in the test-plan: the LIFETIME was actually set to 60 seconds rather than 10.  But the good news is that this gives you time to select the static object and manipulate it (for test) before it deletes itself.

(B) If the object kept drifting and didn't go static then the script failed for some reason.  It would e a mystery as to why this would happen.  The best thing to do in such a case is to re-testing the script many times to see if you can figure out a pattern as to why/when it fails.  If you can come up with a repro recipe for making this happen at least semi-reliable then that would be appreciated and we could file a new bug and then figure out how to fix it.

(C) I never claimed that moving the OS focus to a different window would allow the interface to run faster than 15 fps in this PR.  Nothing changed regarding that feature.  BTW, de-focusing the interface is a useful trick to force the physics engine to take multiple substeps per frame and to delay/chunk the incoming packets from the servers, which can help reveal race conditions and other bugs.

(D) In this case there is a race condition in hifi/master that will sometimes allow an object to become owned just as it goes static and inactive.  The logic that decides to revoke ownership depends on physics updates and so hifi/master is not equipped to properly disown such static objects.  When testing a race condition you must re-run the test many times to get a feel for how often the race is won/lost and what exactly it looks like for each case.  Reload the script, immediately select some other window to make the interface unfocused and watch for the final state of the ownership icon when the object comes to a stop.  No need to wait for the object's lifteime to expire: reload the script and repeat the test, several times.  Is the final state of the  ownership icon the same every time or does it vary?  Is there a behavior change between hfii/master and this PR when you repeat this test many times?

(E) You misunderstood: "the object will be momentarily owned by you but that ownership will quickly be revoked".  What I mean is: the blue ownership icon may blink visible when setting the position of a static object but it will not remain visible.  Try setting the position of a static object in hifi/master with the simulation ownership icons enabled and see what happens (the blue ownership icon will probably show up and remain visible until you log out) and then compare with what happens in this PR when you do the same thing.  Repeat the test many times for each version: is the behavior in both consistent or variable?

AndrewMeadows:
This is back up for test with a revised test plan.  If there are problems or confusion with the test plan ping me directly.

YuppyPlays:
Testing.

YuppyPlays:
Hi Andrew,

Here are my results with the revised tasks:

"In this PR the blue ownership icon (d) should show up immediately after creating the dynamic object."
**Pass**
"Hifi/master may take about 1 second to eventually show the blue ownership icon."
**Pass** (d) shows the icon just before it goes static then disappears again.
"the dynamic object is set static after about 1 second and has a lifetime of 60 seconds."
**Pass**
"If you reload the script and then promptly activate some other window such that Interface looses focus then the FPS will drop to about 15."
**Pass**
"reload the script to delete the current object and launch a new one: do this many times to get an idea of the final state of the ownership icon."
**?** Icons listed as (c) and (d) disappear as soon as the object goes static, C disappearing ever so slightly before  (d). (by I'd say 250ms?, barely noticeable). Did not encounter the issue I did previously with the object drifting off. Re-tried the script in all the same ways I did before but it's not reproducing the same issue.
However, I did receive a crash randomly during. Sent in crash log with what I was doing and other info.
"When this is done in hifi/master there is a possibility that ownership of the object will be gained just as it goes static... in which case the ownership will persist and the ownership icon will remain visible."
**Pass** I was able to replicate this b immediately clicking off the HF PR's window, it would then slow it down enough that the icon shows just as it's static and as said, doesn't disappear (until the object does)
"In this PR the ownership icon should always vanish shortly after the object goes static."
**Pass** Can confirm (d) icon disappears immediately as the object goes static.
"In hifi/master the object will become owned by you and will stay that way indefinitely"
**Pass**
"In this PR the object will be momentarily owned by you but that ownership will quickly be revoked"
**Pass** icon (d) shows whilst moving the object around, then once stopped the icon disappears again.


Notes: As said, I wasn't able to reproduce the issue I had previously which is very strange, but all in all works other than the crash. Not sure if this would be a 100% Pass but that's for you to decide :)

Please let me know if I happened to miss anything and by all means let me know if I can assist any further.

PR #10612, opened 2017-06-03
add blocking crash event sending to signal handlers

birarda:
This PR sends a crash event to our API when Interface catches a crash. This will only occur for builds that use Bugsplat and for users that are not opted-out of sending us usage data.

### DIFF
--- a/interface/src/CrashReporter.cpp
+++ b/interface/src/CrashReporter.cpp
@@ -21,6 +21,7 @@
 #include <csignal>
 #include <QDebug>
 
+#include "networking/ClosureEventSender.h"
 
 #pragma comment(lib, "Dbghelp.lib")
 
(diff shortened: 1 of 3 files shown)


birarda:
#### Test Plan

Forcibly crash your Interface client using each of the `Developer->Crash` menu items, in order. Finally, launch Interface and close it normally. Once complete, ping me to check the DB and confirm the events.

YuppyPlays:
Will test when built.

birarda:
I have a feeling that we will need to make this spawn a side process. As @kencooke explains to me, if the heap is corrupted and we do something that causes an allocation we might not get a bugsplat at all, which is absolutely not what we want.

birarda:
@YuppyPlays sorry, we're going to need a different approach here - no need to test

YuppyPlays:
@birarda No worries :)

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/393/HighFidelity-Beta-PR10612-4f726e42e83b6eb3489e642f054587cb9b7aa79d.dmg

PR #10617, opened 2017-06-04
Use HMD roll angle to turn while flying or walking in HMD

ctrlaltdavid:
Makes you turn while walking or flying in the HMD depending on the amount you roll your head; this instead of turning per where you're looking.


ctrlaltdavid:
**Testing**

This PR makes you turn while walking or flying in the HMD depending on the amount your roll your head; this instead of turning per where you're looking.
When Interface starts up, HMD roll control is enabled by default.

Menu item: <...>
- Enables/disables turning with head roll.
- When disabled, turn head turns after while walking and flying, as before.

Stationary, walking and flying:
- Rolling head has no effect.
- Turning head turns avatar.

Walking forwards/backwards in HMD:
- Rolling your head changes your direction.
- The more you roll your head the faster you turn.
- You can look left and right without changing your direction.

Flying forwards/backwards in HMD:
- Rolling your head changes your direction.
- The more you roll your head the faster you turn.
- You can look left and right without changing your direction.
- You fly up/down per your head pitch, as before.

Test both with and without Settings > Advanced Movement For Hand Controllers enabled.

Test with and without Settings > Avatar > Snap turn when in HMD enabled.

Test non-HMD motion works as before.

Use the developer script, hmdRollControl.js to enable/disable and change settings.

**Notes**

The following JavaScript API object properties are provided to control this:
- MyAvatar.hmdRollControlEnabled boolean
- MyAvatar.hmdRollControlDeadZone number - degrees
- MyAvatar.hmdRollControlSpeed number - degrees/second/degree ???

The following developer script is provided to enable/disable and tweak dead zone and turn speed: hmdRollControl.js

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/404/HighFidelity-Beta-PR10617-25a89c4b94d98309234a30fc89eb090eeab08938.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/404/HighFidelity-Beta-PR10617-25a89c4b94d98309234a30fc89eb090eeab08938.dmg

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

Stationary, walking and flying:
Rolling head has no effect. **Fail** I found my head when slightly tilted (roll) would randomly move in the direction of the roll when turning my head.
Turning head turns avatar. **Pass**

Walking forwards/backwards in HMD:
Rolling your head changes your direction. **Fail** I found that it got stuck in one roll and wouldn't change even when rolling head the opposite way.
The more you roll your head the faster you turn. **Pass**
You can look left and right without changing your direction. **Pass**

Flying forwards/backwards in HMD:
Rolling your head changes your direction. **Fail** Head roll works but now it seems the orientation is locked to the original front facing direction, for example if i turned around in real life then pressed forward while flying, it would send me in the direction i started in relative to my room as opposed to where I am facing. nap turn seems to change the direction it flies forward in.
The more you roll your head the faster you turn. **Pass**
You can look left and right without changing your direction. **Pass**
You fly up/down per your head pitch, as before. **Pass**
Test both with and without Settings > Advanced Movement For Hand Controllers enabled.

Test with and without Settings > Avatar > Snap turn when in HMD enabled. **Fail** Due to orientation issue, this I believe counts as a fail.

Test non-HMD motion works as before. **Pass**

Use the developer script, below, to enable/disable and change settings. **Fail** I'm not sure if I'm doing something wrong but I couldn't seem to load the script hmdRollControl.js, it just instantly turns off. Not sure if I'm misunderstanding?

When disabled, you turn with your head direction while walking and flying, as before.**Fail** See last step.

Please let me know if there is anything else I can do/provide to assist. :)

ctrlaltdavid:
@YuppyPlays Pushed a change which should fix turning problems, plus have split the script into two and updated the test plan with info. on them.

YuppyPlays:
Excellent, I am available and shall test when built :)

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/410/HighFidelity-Beta-PR10617-c788ff3c48abc10dd3dc811a39138c4c5ee31822.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/410/HighFidelity-Beta-PR10617-c788ff3c48abc10dd3dc811a39138c4c5ee31822.exe

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

Stationary, walking and flying:
Rolling head has no effect. **Pass**
Turning head turns avatar. **Pass**

Walking forwards/backwards in HMD:
Rolling your head changes your direction. **Pass**
The more you roll your head the faster you turn. **Pass**
You can look left and right without changing your direction. **Pass**

Flying forwards/backwards in HMD:
Rolling your head changes your direction. **Pass**
The more you roll your head the faster you turn.**Pass**
You can look left and right without changing your direction. **Pass**
You fly up/down per your head pitch, as before.**Pass**
Test both with and without Settings > Advanced Movement For Hand Controllers enabled. **Pass**

Test with and without Settings > Avatar > Snap turn when in HMD enabled. **Pass**

Test non-HMD motion works as before. **Pass**

Use the developer script, below, to enable/disable and change settings:
When disabled, you turn with your head direction while walking and flying, as before. **Fail** Tried loading in "running scripts": hmdRollControlDisable.js and hmdRollControlEnable.js but neither seems to make a difference.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/413/HighFidelity-Beta-PR10617-9968a239d07d79f0fdc9be4a3610a99400d69351.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/413/HighFidelity-Beta-PR10617-9968a239d07d79f0fdc9be4a3610a99400d69351.dmg

murillodigital:
test this please

YuppyPlays:
Testing.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/416/HighFidelity-Beta-PR10617-9968a239d07d79f0fdc9be4a3610a99400d69351.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/416/HighFidelity-Beta-PR10617-9968a239d07d79f0fdc9be4a3610a99400d69351.dmg

PhilipRosedale:
Looks good to me. Tested walking and flying, default turning rates look good.

YuppyPlays:
Testing complete:

Can confirm everything now works including the Disable.js. **Pass**

zfox23:
First of all, I'm so excited to see this PR!

I immediately noticed a few issues when using this:
1. The turn rate is very slow - I have to tilt my head at an uncomfortable angle to be able to "swoop" through things like the windows in Welcome. I recognize that I could change `hmdRollControlEnable.js` to increase this. I feel like the default value of `hmdRollControlSpeed` needs more tweaking via user testing.
2. For some reason, tilting my head left yields a faster turn rate than tilting my head right.
3. I was able to get into a state where holding my left control stick directly away from me caused me to turn. I don't think this is related to your PR (I think this happens in some cases after turning using the right stick), but it could be? I don't have repro steps.

tapalisa:
Yeah this seems good! I went through it without too much motion sickness (although now IRL tilting my head makes me sick. Go figure.)

1)It still works while walking. Is that intended? It's definitely more sickening to do while walking. 
2) I think I experienced the same thing that Zack experienced in (3) a couple times. I don't have repro steps either but it is very barfy. 

The slow turn rate is probably a good thing for delicate flowers like me, but it does highlight how uncontrollably fast you are forced to fly. We should brainstorm a fix for that.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1843,8 +1851,56 @@ void MyAvatar::updateOrientation(float deltaTime) {
...
+
+        // Turn with head roll.
+        if (speed >= MIN_CONTROL_SPEED) {
+            // Feather turn when stopping moving.
+            float speedFactor;
+            if (getDriveKey(TRANSLATE_Z) != 0.0f) {
+                _lastDrivenSpeed = speed;
+                speedFactor = 1.0;
+            } else {
+                speedFactor = glm::min(speed / _lastDrivenSpeed, 1.0f);
```
AndrewMeadows: There is danger of divide by zero here: `_lastDrivenSpeed` is initialized to `0.0` and there is no guarantee that it has been set otherwise in this context.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1843,8 +1851,56 @@ void MyAvatar::updateOrientation(float deltaTime) {
...
+            auto hmdSensorEulers = safeEulerAngles(_hmdSensorOrientation);
+            float rollAngle = fabsf(glm::degrees(hmdSensorEulers.z));
+            float dot = glm::dot(_hmdSensorOrientation * Vectors::UNIT_Y, 
+                glm::angleAxis(hmdSensorEulers.y, Vectors::UNIT_Y) * Vectors::UNIT_X);
+            float rollSign = dot < 0.0f ? 1.0f : -1.0f;
+            float yawChange = rollAngle > _hmdRollControlDeadZone ? rollSign * (rollAngle - _hmdRollControlDeadZone) : 0.0f;
+            totalBodyYaw += speedFactor * direction * yawChange * deltaTime * _hmdRollControlSpeed;
+        }
+ 
+        auto deltaYaw = glm::angleAxis(glm::radians(totalBodyYaw), avatarUp);
```
AndrewMeadows: Not a blocker.  I know people like to use degrees instead of radians but... **cry**.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1843,8 +1851,56 @@ void MyAvatar::updateOrientation(float deltaTime) {
...
+        _hmdRollControlOrientation = deltaYaw * _hmdRollControlOrientation;
```
hyperlogic: This seems more complex than it needs to be.  My thought was that you can determine the roll angle of the hmd.  using safeEulerAngles() or some other method.  And then just apply a y-180 angular velocity with a magnitude that is proportional to that angle.  Then the linear velocity is just = to the facing direction of the hmd.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1843,8 +1851,56 @@ void MyAvatar::updateOrientation(float deltaTime) {
...
+                _hmdRollControlOrientation);
+            _lastPitch = 0.0f;
+        }
+        _lastSpeed = speed;
+
+        // Climb/fall with head pitch.
+        auto cameraForward = getMyHead()->getCameraOrientation() * IDENTITY_FORWARD;
+        auto pitch = asin(glm::dot(avatarUp, cameraForward));
+        auto deltaPitch = glm::angleAxis(pitch - _lastPitch, getOrientation() * IDENTITY_RIGHT);
+        _hmdRollControlOrientation = deltaPitch * _hmdRollControlOrientation;
```
AndrewMeadows: There is danger that the resulting `_hmdRollControlOrientation` quaternion will drift to non-unity length **(1)**, which will cause interesting bugs.  Note: at high speeds `_hmdRollControlOrientation` is set to the result of swingTwistDecomposition() (above), which probably normalizes its output, which will hide this problem, but at low speeds for extended times there can be drift.

**(1)** Whenever we do `Q = dQ * Q` AND we are saving Q for use in the next frame we must normalize the result, else risk drifting into non-normalized quaternion space from accumulated floating point errro.  In other words do this:  `Q = glm::normalize(dQ * Q)`.

interface/src/avatar/MyAvatar.h
```diff
@@ -331,6 +335,13 @@ class MyAvatar : public Avatar {
     Q_INVOKABLE bool getClearOverlayWhenMoving() const { return _clearOverlayWhenMoving; }
     Q_INVOKABLE void setClearOverlayWhenMoving(bool on) { _clearOverlayWhenMoving = on; }
 
+    void setHMDRollControlEnabled(bool value) { _hmdRollControlEnabled = value; }
+    bool getHMDRollControlEnabled() const { return _hmdRollControlEnabled; }
+    void setHMDRollControlDeadZone(float value) { _hmdRollControlDeadZone = value; }
+    float getHMDRollControlDeadZone() const { return _hmdRollControlDeadZone; }
+    void setHMDRollControlSpeed(float value) { _hmdRollControlSpeed = value; }
+    float getHMDRollControlSpeed() const { return _hmdRollControlSpeed; }
+
```
hyperlogic: All of these assessors need to be thread safe.  I would suggest using std::atomic<>, or TheadSafeValueCache<> for the _hmdRoleControl* variables.

interface/src/avatar/MyAvatar.cpp
ctrlaltdavid: Thank you.

interface/src/avatar/MyAvatar.cpp
ctrlaltdavid: Well caught!

interface/src/avatar/MyAvatar.cpp
ctrlaltdavid: Just going with the flow of the rest of the method; will leave in degrees despite tears.

interface/src/avatar/MyAvatar.cpp
ctrlaltdavid: @hyperlogic I'm not sure I follow what you mean. If you apply an angular velocity to the avatar you would need to apply it according to change in head roll from previous loop, so that if you kept your hear rolled at one angle you don't keep accelerating the rate of turn, and so that stop turning when you make your head level. In practice need decouple direction of motion from both HMD orientation and avatar orientation (which can lag and cause unwanted side effects).

interface/src/avatar/MyAvatar.h
ctrlaltdavid: @hyperlogic If this is the case then there would appear to be many other properties / methods in this header that also need fixing? Perhaps would best be done as a separate PR?

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/430/HighFidelity-Beta-PR10617-91f2477a251d184a8fadc525352abda150e03603.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/430/HighFidelity-Beta-PR10617-91f2477a251d184a8fadc525352abda150e03603.dmg

ctrlaltdavid:
@zfox32, @tapalisa

1) Turn rate was purposefully lowered to current value from a value twice the current after others (Philip) found it was too high.

2) If you're strafing as well as turning it might appear that you're turning faster in the direction of strafe?

3) Hrmm?
	
Some possibilities:
- Separate turn rates for walking and flying.
- UI to configure turn rates.
- UI to configure turn-with-head-roll (new) versus move-in-head-direction (old), with separate settings for walking and flying.
- Default to turn-with-head-roll disabled until UI added.

zfox23:
@ctrlaltdavid, I'm not sure if your numbered list corresponds to my list or Alisa's, but:

1. Fair enough, but I still think the number should be more carefully considered before merging.
2. I was not strafing as well as turning. This is a repeatable problem for me.
3. The buggy state was: Hold the left control stick forward while looking straight ahead, witness turning to the left or right despite no other inputs.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1872,14 +1872,14 @@ void MyAvatar::updateOrientation(float deltaTime) {
...
-        _hmdRollControlOrientation = deltaPitch * _hmdRollControlOrientation;
+        _hmdRollControlOrientation = glm::normalize(deltaPitch * _hmdRollControlOrientation);
         _lastPitch = pitch;
 
         // Turn with head roll.
         if (speed >= MIN_CONTROL_SPEED) {
             // Feather turn when stopping moving.
             float speedFactor;
-            if (getDriveKey(TRANSLATE_Z) != 0.0f) {
+            if (getDriveKey(TRANSLATE_Z) != 0.0f && _lastDrivenSpeed != 0.0f) {
```
AndrewMeadows: This change doesn't look right: if `_lastDrivenSpeed == 0.0` then the logic falls into the `else` case and tries to divide by it (that would be bad).  Perhaps the `!=` operator is a typo?

zfox23:
I can confirm that I still turn way faster by tilting my head left versus right.

ctrlaltdavid:
@zfox23 Latest commits should fix getting into a state of turning when shouldn't (or turning much faster than should be). Should also fix occurrences of turn left/right rates being different.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/465/HighFidelity-Beta-PR10617-fb06f00ed5403962ea35bce05604c1defa3ad1f5.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/465/HighFidelity-Beta-PR10617-fb06f00ed5403962ea35bce05604c1defa3ad1f5.exe

ctrlaltdavid:
@PhilipRosedale 

Could you please try this out and see if it's what you want?
- When stationary, your avatar turns with your head.
- When moving, looking around does not change the direction of movement unless you roll your head to turn or pitch your head to fly up/down.
- When moving, arrow/joystick controls still turn you.
- There's a dead zone in head roll so that you don't turn with head almost level.
- Rate of turn increases with head roll.

Possible enhancements:
- Separate turn rates for walking and flying.
- UI to configure turn rates.
- UI to configure turn-with-head-roll (new) versus move-in-head-direction (old), with separate settings for walking and flying.
- Default to turn-with-head-roll disabled until UI added.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/473/HighFidelity-Beta-PR10617-8719400437eecbee4fc75f11e46346ecc7c98e9a.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/473/HighFidelity-Beta-PR10617-8719400437eecbee4fc75f11e46346ecc7c98e9a.dmg

PhilipRosedale:
@zfox23 does this resolve the bug you were seeing?

zfox23:
I was now able to get a the state where holding forward on the left stick made me move in a direction that didn't correspond with the way I was looking. But, I'm not sure that behavior is a problem with this PR. I was no longer able to get into the state where holding forward made me turn.

It also seems to me like looking up and down have different rates of turn between both each other and from roll left/right. However, I was not able to get into the state where tilting my head left yielded a faster turn rate than tilting it right.

After trying this some more, I am concerned about wanting to use this control scheme long-term. After using this PR, it feels like I should be flying where I'm looking, rather than using my head roll angle as a proxy for the right control stick. But perhaps I am too late in saying this.

PhilipRosedale:
@ctrlaltdavid This is working pretty well for me now.  

Like @zfox said, I think removing the 'turn toward the direction you are looking' that was also there was a mistaken idea from me in this design.  Can we try adding that back in - so that you travel/thrust the direction that you are looking, and also additionally turn with roll (as you've added)? 

When we get this right we are going to be happy!

ctrlaltdavid:
Replacement PR: https://github.com/highfidelity/hifi/pull/10685

PR #10639, opened 2017-06-07
[NEED FEEDBACK] Initial work on new third-person camera (outside HMD)

seefo:
Changed third-person outside of HMD to feel like an actual camera mode, looking for feedback from people.  Please leave feed back here in the comments with your thoughts and experience.

This also fixes the issue of a jagged camera when in first-person and you look up diagonally.


hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/481/HighFidelity-Beta-PR10639-c7bb96765510a74a5591384d77bc6da0a3c0dbdb.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/481/HighFidelity-Beta-PR10639-c7bb96765510a74a5591384d77bc6da0a3c0dbdb.dmg

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

First off, I'd just like to say I am really digging this for third person!

Turning left/right works flawlessly 
Dragging camera around:  some might say this feels off, but personally I like the way this handles when turning as even when avatar goes briefly off screen, it is smooth and fluent.
Walking forward/backwards: I found the backwards to make the avatar a tiny bit too close for my liking, but that's just me.
Jumping:  Works great
Flying: Flying works well, the camera pans out further than when walking which I'd imagine is expected (I quite like this though) However, if I fly backwards it will end up with the avatar behind the camera which i'm not sure if this is desired result or not. Also going too fast upwards/downwards when flying appears to take the avatar off camera too.

Overall I think this is fantastic. Keep up the awesome work!

zfox23:
I gave this a thorough test and I am not a fan of these changes. Overall, these changes take precise control away from me when I'm using the third person camera, even though they do add a smooth effect. More specifically:

# Standing Still
## Using Mouse Right-Click to Turn
I use the mouse to turn when I'm looking for quick, precise movements.
### First Person
- With these changes, precision is lost, especially when moving the camera in the Y direction.
- X and Y motion in first person has different properties
### Third Person
- This is almost dizzying. When I'm using the mouse to turn, I expect the camera to follow almost immediately behind my avatar. It might be OK if there were a tiny little bit of smoothing applied, but not nearly as much as is applied now.

## Using the Scroll Wheel to Zoom
- I'm OK with the loss of precision here, simply because it looks cool. I do wish the zoom responded a bit more quickly, though.
- Also, there is a bug here - when the camera is close to my head, scrolling has no effect for a few scrolls.

## Using A and D to Turn
I use A and D to turn when I'm driving my avatar somewhere.
I also use A and D in conjunction with "Shift" to strafe.
### First Person
- No perceptible changes.
### Third Person
- This is pretty cool. However, I do wish the camera responded a bit more quickly to my avatar's motion, i.e. `CAMERA_ROTATE_SPEED` were a bit higher (I _think_ that's the correct variable? :smile:)
- When strafing, the camera smoothly _starts_ following me, but abruptly _stops_ after I stop. Seems like a bug. Hard to explain, easy to reproduce.
- When the camera is zoomed way out, the effect is almost dizzying again.

## Using W and S to Drive
I use W and S when I'm driving my avatar somewhere.
I also use W and S in conjunction with "Shift" to move my head up and down, mostly to quickly "nod" to a coworker during a meeting.
### First Person
- No changes when driving forwards and backwards, which is good. BUT...
- When using W and S plus "Shift" to nod, this is terribly dizzying. There shouldn't be smooth camera motion applied in this case.
### Third Person
- Cool! This looks great. Same bug as above, though: the camera smoothly _starts_ following me, but abruptly _stops_ after I stop.

# Flying
Most of the above feedback applies when flying (regarding precision, dizzying effects, etc), with a few notable extra pieces of feedback:
- The effect of the camera panning back while flying is super cool - although I think it goes back too far when moving quickly.
  - When the camera is too far back, driving precision is greatly diminished. My common activity of trying to fly through the hangar openings in Welcome becomes a huge challenge!
  - When the camera is too far back, I can barely see my avatar.
- Bug: If I fly fast enough forwards or backwards, then turn, my avatar goes off-screen.
- Bug: If I ascend or descend fast enough, my avatar goes off-screen.

# In Closing...
Great work. I know that I was mostly negative above, but I do believe that the existing third-person camera needs some work, and it's excellent that you're starting to explore alternatives. I look forward to future iterations!

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1884,8 +1886,35 @@ void MyAvatar::updateOrientation(float deltaTime) {
...
+    // position
+    glm::vec3 boom_offset = _boomLength * -IDENTITY_FORWARD;
+    _cameraPositionTarget = getDefaultEyePosition() + (getHeadOrientation() * boom_offset);
+
+    float distance = glm::distance(_cameraPositionTarget, _cameraPosition);
+    glm::vec3 cameraSpeed = glm::normalize(_cameraPositionTarget - _cameraPosition) * (distance * CAMERA_MOVE_SPEED * deltaTime);
+
+    if (distance > CAMERA_MOVE_MIN_DELTA) {
+        _cameraPosition += cameraSpeed;
+    }
```
AndrewMeadows: `cameraSpeed` is a poorly named variable:

(1) It is a vector, so is not a speed.

(2) Its units are "meters" so it isn't a velocity.  It is a deltaPosition... based on how it is used.

A couple other points:

(3) The "deltaPosition" is only used when the `if` line evaluates to `true`, so I would say: "Move it inside the `if` block so that it is only computed when necessary".  However... you actually compute the distance between _cameraPositionTarget and _cameraPosition twice, once in the glm::distance() call and again in the glm::normalize() call.

(4) The glm::normalize() call will do bad stuff you feed it a zero-length vector.  I find it better to just avoid it when possible.

(5) The normalization is unnecessary because all it does is divide by the distance, but you multiply by distance which is effectively the inverse of the normalization.

Do this instead:
```
glm::vec3 nextStep = _cameraPositionTarget - _cameraPosition;
float offsetLength = glm::length(nextStep);
float constantStep = CAMERA_MOVE_SPEED * deltaTime;
if (length > constantStep) {
    nextStep *= constantStep / length;
}
_cameraPosition += nextStep;
```

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1884,8 +1886,35 @@ void MyAvatar::updateOrientation(float deltaTime) {
...
+
+    // orientation
+    _cameraPitchTarget += getDriveKey(PITCH);
+    
+    float pitch = getHead()->getBasePitch();
+    float pitch_delta = _cameraPitchTarget - pitch;
```
AndrewMeadows: Use `pitchDelta` rather than `pitch_delta` according to the coding standard [section 1.1.3](https://github.com/highfidelity/hifi/wiki/Coding-Standard#113-local-variables-must-be-in-mixed-case-and-not-prefixed-with-an-underscore).

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/765/HighFidelity-Beta-PR10639-a5830b78fbc0258c6f3f75c24fcfc9da0de78710.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/765/HighFidelity-Beta-PR10639-a5830b78fbc0258c6f3f75c24fcfc9da0de78710.exe

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1884,8 +1886,48 @@ void MyAvatar::updateOrientation(float deltaTime) {
...
+    const float CAMERA_MOVE_MIN_DELTA = 0.25f;
+    const float CAMERA_ROTATE_MIN_DELTA = 2.0f;
+
+    glm::vec3 boom_offset = _boomLength * -IDENTITY_FORWARD;
+    
+    // position
+    if (Menu::getInstance()->isOptionChecked(MenuOption::CenterPlayerInView)) {
+        float pitch = getHead()->getBasePitch() - 90;
+        float yaw = getBodyYaw();
+        float RAD_TO_DEG = 1 / 180.0f;
```
AndrewMeadows: Use `90.0f` instead of `90` and `1.0f` instead of `1` as per [coding standard section 3.5.4](https://github.com/highfidelity/hifi/wiki/Coding-Standard#354-when-using-a-single-precision-float-type-include-the-trailing-f).

BTW, `RAD_TO_DEG` is not needed.  You could use `glm::degrees()`.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1884,8 +1886,48 @@ void MyAvatar::updateOrientation(float deltaTime) {
...
+
+        float x = glm::asin(yaw * RAD_TO_DEG);
+        float y = glm::acos(pitch * RAD_TO_DEG);
+
+        _cameraPosition = getPosition() + (glm::vec3(0, y, 0) * getHeadOrientation());
```
AndrewMeadows: `0.0f`

interface/src/avatar/MyAvatar.h
```diff
@@ -774,6 +777,12 @@ public slots:
     void setAway(bool value);
 
     std::vector<int> _pinnedJoints;
+
+    // new third person camera
+    float _cameraPitchTarget{ 0 };
+    glm::vec3 _cameraPosition{ 0, 0, 0 };
+    glm::vec3 _cameraPositionTarget{ 0, 0, 0 };
```
AndrewMeadows: `float _cameraPitchTarget { 0.0f }`

And similarly for the lines below that.

ZappoMan:
Closing this PR. It's almost a month old, marked DNM, there's feedback that the feature is not desirable, and it's got merge conflicts. @seefo - can you please work with @birarda to develop a plan for either finishing this project or determine that we don't want to proceed. Also I suggest you get @RyanDowne and/or @cozza13 to find you a PM to guide this feature to completion. This PM should probably live in the client team.

seefo:
@ZappoMan Will do!

PR #10653, opened 2017-06-09
Spectator-camera : Provide a way to render the side camera always mono

samcake:
Quick fix to introduce a mechanism to force the context level stereo flag to be disabled

### DIFF
--- a/interface/src/PrototypeSelfie.cpp
+++ b/interface/src/PrototypeSelfie.cpp
@@ -1,6 +1,7 @@
 
 #include "PrototypeSelfie.h"
 
+#include <gpu/Context.h>
 
 void MainRenderTask::build(JobModel& task, const render::Varying& inputs, render::Varying& outputs, render::CullFunctor cullFunctor, bool isDeferred) {
 
@@ -51,9 +52,15 @@ public:
         // Caching/restoring the old values doesn't seem to be needed. Is it because we happen to be last in the pipeline (which would be a bug waiting to happen)?
         _cachedArgsPointer->_blitFramebuffer = args->_blitFramebuffer;
         _cachedArgsPointer->_viewport = args->_viewport;
+        _cachedArgsPointer->_displayMode = args->_displayMode;
         args->_blitFramebuffer = destFramebuffer;
         args->_viewport = glm::ivec4(0, 0, destFramebuffer->getWidth(), destFramebuffer->getHeight());
-        // FIXME: We're also going to need to clear/restore the stereo setup!
+        args->_viewport = glm::ivec4(0, 0, destFramebuffer->getWidth(), destFramebuffer->getHeight());
+        args->_displayMode = RenderArgs::MONO;
+
+        gpu::doInBatch(args->_context, [&](gpu::Batch& batch) {
+            batch.disableContextStereo();
+        });
 
         auto srcViewFrustum = args->getViewFrustum();
         srcViewFrustum.setPosition(_position);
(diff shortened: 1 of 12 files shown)


hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/528/HighFidelity-Beta-PR10653-b694256559216821b07af963c33e3cc0fdbfed20.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/528/HighFidelity-Beta-PR10653-b694256559216821b07af963c33e3cc0fdbfed20.dmg

zfox23:
A few things I noticed while testing this PR (that may or may not be due to this PR):

1. CTRL+S Snapshots now look like this:
hifi-snap-by-zfox-on-2017-06-09_10-47-04

I'm assuming this is because of the new stencil buffer being enabled? Is that right? If so, that doesn't warrant a "QA Failure" on this PR, but should be fixed.

2. Snapshots taken through the Snapshots app look like this:
hifi-snap-by-zfox-on-2017-06-09_10-48-05

Notice the small amount of black vignetting at the edges of the snapshot. Probably also stencil related? If that's right, that doesn't warrant a "QA Failure" on this PR, but should be fixed.

3. Most importantly, the camera preview texture has a strange line halfway through it - the scene seems to be lit one way to the left of the line, and the other way to the right of it? Please see the image below:

look

I've circled the "line" in the preview. Notice how, in the actual scene, the line is not present. This is a problem. Why is this?

howard-stearns:
Re 1 and 2: I don't know why snapshot app and ctl-s are different. Worth looking into. They should both look like the left eye, which should be similar to the mono preview. (The mono preview is actually doing some extra stuff that doesn't have anything to do with this pr or branch.)

zfox23:
Further investigation:

Both ctrl+s and the snapshot app correctly yield snapshots corresponding to the left eye's view.

The reason why the snapshot app produces a different image is because we specifically set the aspect ratio of the output snapshot - which is _very_ cropped in comparison to the full-res image. This seems bad, and also obvious, now that I'm thinking about it...

Here's a direct comparison:
Ctrl+S (aspect ratio not forced):
hifi-snap-by-zfox-on-2017-06-09_11-00-34

Snapshot App (aspect ratio forced, thus cropped):
hifi-snap-by-zfox-on-2017-06-09_11-00-39

I really don't like that the output of the snapshot app is cropped. I don't know why I haven't thought about this before. I can think of some solutions to this problem, but none of them are great...

zfox23:
Anyway, Issues (1) and (2) are definitively not related to this PR, so that leaves issue (3) to be fixed or addressed before this can be merged.

zfox23:
[turn off stereo in capture](https://trello.com/c/GJ8BuIct/169-turn-off-stereo-in-capture)

howard-stearns:
right. the snaps app takes all pictures in the aspect ratio that has been determined to be appropriate for external sharing as well as for use in our feed.

I'm not a fan of the stenciling in our snapshots (we could scissor out a smaller view instead of the largest view that fits the required aspect ratio), but I gather the stenciling is being considered separately (in master).

So, do I understand correctly then, that:
1. The snapshot behavior is the same as master.
2. The snapshot behavior is as designed (here and in master).
3. There is still a question about the lighting.

zfox23:
@howard-stearns. You understand all of that correctly.

howard-stearns:
There definitely is an issue, but let's merge and go from there. Meanwhile, we have to fix whatever is going on with the sizing.

PR #10673, opened 2017-06-13
Fix the background not rendering on the second frame

samcake:
This was simply due to a bug in the GLBackend which was actually expected to happen...
  // TODO: we will probably need to also check the write mask of stencil like we do
 // for depth buffer, but as would say a famous Fez owner "We'll cross that bridge when we come to it"

I guess we crossed that bridge today and fixed the stencilWriteMask not being set to xFF and preventing to write the Stencil buffer during the glClear call.


interface/src/Application.cpp
```diff
@@ -1873,8 +1873,8 @@ void Application::initializeGL() {
         isDeferred = false;
     }
 
-    _renderEngine->addJob<MainRenderTask>("MainFrame", cullFunctor, isDeferred);
     _renderEngine->addJob<SecondaryCameraRenderTask>("SecondaryCameraFrame", cullFunctor);
+    _renderEngine->addJob<MainRenderTask>("MainFrame", cullFunctor, isDeferred);
```
zfox23: Did you mean to flip the ordering in this PR?

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/636/HighFidelity-Beta-PR10673-f973f7b5e34c4e77f994ee225078e0f238e265aa.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/636/HighFidelity-Beta-PR10673-f973f7b5e34c4e77f994ee225078e0f238e265aa.exe

ZappoMan:
Bridges are overrated. Sometimes getting your feet wet is worth it.

interface/src/Application.cpp
samcake: yep that too

zfox23: Why is that necessary?

zfox23:
Yay, works! I'll resolve the conflicts and merge into `spectator-camera`.

PR #10678, opened 2017-06-13
Secondary camera and rendering improvements

zfox23:
This PR makes the following changes to spectator-camera:
1. ~Added a `SECONDARY_CAMERA_RENDER_MODE` to the `RenderArgs::RenderMode` enum, which doesn't do anything right now, but may be useful in the future~ Removed
2. Correctly set the projection of the secondary camera
    - ~None of the projection's parameters (vFoV, near clip distance, far clip distance) are available to JS _yet_; they are hardcoded right now.~ These parameters are now available to JS.
3. The secondary camera no longer uses the same culling frustum as the main camera.
    - This fixes most culling issues. I have seen some minor culling issues at the edges of the texture when in HMD mode, and I think it's because of the way we only use one eye's culling frustum? Not sure.
4. Fixed a bug in GLBackendTransform.cpp where the stereo enabled status wasn't checked properly
5. Removed dependencies on the `FramebufferCache` singleton for determining texture dimensions, instead correctly relying on viewport dimensions
6. Improved the scaling of the Spectator Camera Viewfinder Image3DOverlay when window size changes
    - "Improved", but not "perfected" - the new code will stretch the preview as the window aspect ratio changes, which should be fixed. I created a Trello card for this.
7. Removed the `flip()` helper function from `spectatorCamera.js`. Investigated and commented why a negative dimension for the viewfinder overlay is necessary.
    - This warrants more investigation when we have time.
8. Removed dead comments and code

At @samcake's request I will make another PR with changes from (4) and (5) to merge into Master.

Huge props to Sam for helping me out with bugfixes I never would have found on my own :)

**Test Plan:**
HMD + Hand Controllers Required
1. Don your HMD and navigate to dev-welcome. Verify that the "Display->Disable Preview" menu option is CHECKED. Also ensure that the Interface application window is maximized (fullscreen is unnecessary).
2. Using your Tablet, open the "Spectator" app
3. Tap the "Camera On" checkbox
4. Verify that a 3D representation of a video camera appears in front of you at a comfortable distance and height.
5. Verify that there exists a "viewfinder" overlay on top of the camera model that allows you to see things from the camera's point of view.
6. Using a hand controller, grab the camera and move it around. Verify that you are able to move it around.
7. Verify that, as you move the camera around, the viewfinder displays what you would expect the camera to capture i.e. if you grab rotate the camera 45 degrees to the left, the viewfinder shows content to your left.
8. Verify that, as you move the camera around, content does not "pop-in" on the viewfinder.
9. Point the camera's lens towards your avatar. Verify that all parts of your avatar's body and attachments are visible as you would expect.
    - KNOWN ISSUE: Head is cauterized and will not display. See https://trello.com/c/oktxxpnY/185-what-to-do-about-cauterized-head
10. Peek behind your HMD and manually resize the Interface window to be a square. Put your HMD back on. Verify that the camera's viewfinder displays a square image.
    - KNOWN ISSUE: Viewfinder preview will not automatically resize when window is resized. For now, use the tablet app to toggle the camera off and on, then verify the above, with the caveat that...
    - KNOWN ISSUE: Viewfinder preview currently stretches to fit the "glass" piece of the camera's model. For details regarding both this and the above KNOWN ISSUES, see: https://trello.com/c/dqtIcns3/170-respond-to-resizing-screen
11. Open the Spectator tablet app. Verify that the "MONITOR SHOWS" slider is currently set to its default, "HMD Preview".
12. While keeping the Interface window a square, use the "Display" menu to UNCHECK the "Disable Preview" option.
13. As you look around in your HMD, peek behind your HMD to verify that what shows up on your monitor (in the Interface application window) is a preview of what you're looking at.
14. In the Spectator tablet app, tap the "MONITOR SHOWS" slider to move the slider from "HMD Preview" to "Camera View".
15. Peek behind your HMD to verify that what's showing up on your monitor (in the Interface application window) is what the Spectator Camera is "seeing". Ensure the display is not corrupted in any way.
    - KNOWN ISSUE: Lighting problem - you might see "lines" or "tiles" of incorrect lighting on the display. See: https://trello.com/c/WUeqovYH/187-lighting-issue-vertical-line-about-2-3-across-to-the-right
16. Peek behind your HMD and manually resize the Interface window to be maximized (fullscreen is unnecessary). Put your HMD back on. Verify that the camera's viewfinder displays a wide image that fills the "glass" part of the camera model. Also, peek behind your HMD to verify that what displays on your monitor (in the Interface application window) is what the Spectator Camera is "seeing". Ensure the display is not corrupted in any way. 
    - KNOWN ISSUES: same as (10)'s KNOWN ISSUES
    - KNOWN ISSUE: What shows up on your monitor at this point will not be automatically resized, so you are likely to see some visual corruption in the Interface window. To fix this, use the Spectator tablet app to toggle the camera off and on.
17. Verify that, throughout the above testing, the camera viewfinder overlay did not move or rotate. You can do this by looking around the viewfinder and ensuring it still looks "centered" in the "glass" part of the camera model.
18. In the Spectator tablet app, tap the words "HMD Preview" under "MONITOR SHOWS" to switch the slider to "HMD Preview".
19. Verify that the camera's viewfinder preview still shows up correctly.
20. Verify that your main monitor now shows the "READY - Put on your Headset" image
    - KNOWN ISSUE: This doesn't currently work. Instead, the monitor displays a frozen view of what the Spectator Camera last "saw". A similar issue occurs if you have the camera on and switch this switch to "HMD Preview" with the "Disable Preview" option checked.
21. Quit Interface using Tablet->File->Exit. Verify that you don't crash.
    - KNOWN ISSUE: Sometimes, you crash. See: https://trello.com/c/TcsggLra/189-fix-crash-that-occurs-when-closing-interface-after-creating-a-spectator-camera


zfox23:
[Fix skybox not rendering in spectator camera](https://trello.com/c/yLxYEtJY/179-fix-skybox-not-rendering-in-spectator-camera)

zfox23:
[Investigate changing the NEAR CLIP plane of the secondaryCamera camera so that it's impossible to capture the front of the camera model when moving quickly](https://trello.com/c/8tUYPF8l/183-investigate-changing-the-near-clip-plane-of-the-secondarycamera-camera-so-that-its-impossible-to-capture-the-front-of-the-camera)

zfox23:
[wrong culling!](https://trello.com/c/xGVK8afO/188-wrong-culling)

zfox23:
[capture size in stereo (was turn off stereo in capture)](https://trello.com/c/GJ8BuIct/169-capture-size-in-stereo-was-turn-off-stereo-in-capture)

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/661/HighFidelity-Beta-PR10678-0d86d4f0cb3917acd52a447d0490ad28b4dddf64.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/661/HighFidelity-Beta-PR10678-0d86d4f0cb3917acd52a447d0490ad28b4dddf64.dmg

interface/src/SecondaryCamera.cpp
```diff
@@ -77,6 +78,10 @@ class BeginSecondaryCameraFrame {  // Changes renderContext for our framebuffer
             auto srcViewFrustum = args->getViewFrustum();
             srcViewFrustum.setPosition(_position);
             srcViewFrustum.setOrientation(_orientation);
+            srcViewFrustum.setProjection(glm::perspective(45.0f, ((float)args->_viewport.z / (float)args->_viewport.w), 0.1f, 100.0f));
```
howard-stearns: Magic numbers.

Shouldn't the 45 be wrapped in glm::radians()?

And should the fov and clipping planes come from the config rather than being hardcoded? (E.g., be instance variables of the BeginSecondaryCameraFrame, which are set by the config() method based on public properties of the BeginSecondaryCameraFrameConfig.)

scripts/system/spectatorCamera.js
```diff
@@ -132,14 +133,17 @@
...
-            scale: -0.16,
+            dimensions: { x: 0.16, y: -0.16 * windowAspectRatio / previewAspectRatio, z: 0 }
+            // Negative dimension for viewfinder is necessary for now due to the way Image3DOverlay
+            // draws textures.
+            // See Image3DOverlay.cpp:91. If you change the two lines there to:
+            //     glm::vec2 topLeft(-x, -y);
+            //     glm::vec2 bottomRight(x, y);
+            // the viewfinder will appear rightside up without this negative y-dimension.
+            // However, other Image3DOverlay textures (like the PAUSED one) will appear upside-down. *Why?*
+            // FIXME: This code will stretch the preview as the window aspect ratio changes. Fix that!
```
howard-stearns: How/when do we want to address this?

scripts/system/spectatorCamera.js
```diff
@@ -94,10 +90,15 @@
     //     spawn the camera entity.
     //
     var isUpdateRenderWired = false;
+    var windowAspectRatio;
+    var previewAspectRatio = 16/9;
```
howard-stearns: I'm not following what we're doing here.

The design spec was that we fill the Interface window, whatever size it is, without additional constraints or constants. The reason for that is to allow a lot of flexibility for surfacing issues while keeping everything visibly apparent. E.g., not a wall of numeric parameters in the app, and certainly not any hidden unchangeable ones if we can avoid it. (For now fov is fixed, but as I mention elsewhere, it should at least be config-able. But I'm not getting why 16x9 would be specified at all.)

Are we changing the MVP?

zfox23: This refers to the viewfinder preview.

scripts/system/spectatorCamera.js
zfox23: I made a Trello comment on the "react to resizing window" - I think this fix should be a part of that.

scripts/system/spectatorCamera.js
zfox23: The viewfinder model has a window that has an aspect ratio of approximately 16/9

interface/src/SecondaryCamera.cpp
```diff
@@ -62,13 +62,14 @@ class BeginSecondaryCameraFrame {  // Changes renderContext for our framebuffer
...
         if (destFramebuffer) {
-            // Caching/restoring the old values doesn't seem to be needed. Is it because we happen to be last in the pipeline (which would be a bug waiting to happen)?
             _cachedArgsPointer->_blitFramebuffer = args->_blitFramebuffer;
             _cachedArgsPointer->_viewport = args->_viewport;
             _cachedArgsPointer->_displayMode = args->_displayMode;
+            _cachedArgsPointer->_renderMode = args->_renderMode;
             args->_blitFramebuffer = destFramebuffer;
             args->_viewport = glm::ivec4(0, 0, destFramebuffer->getWidth(), destFramebuffer->getHeight());
             args->_displayMode = RenderArgs::MONO;
+            args->_renderMode = RenderArgs::SECONDARY_CAMERA_RENDER_MODE;
```
samcake: I would recommend to avoid that new mode
we re trying to get rid of it because that particular mode is almost exclusively application semantic that make it very hard to rely on it in a generic way within the jobs which are hopefully dumb and versatile enough to be reused in many different contexts.

howard-stearns: @samcake I think the idea might be related to looking ahead at how the head cauterization replacement by done. I don't know if this is the right way to go, but I could imagine the render commands (or even the shader) doing things conditionally based on the render mode. (But even if that's right, maybe such conditionalization should be more specific / lower level.)

scripts/system/spectatorCamera.js
howard-stearns: I see. But doesn't it defeat the purpose of a viewfinder to be displaying something different than is being captured?  Would it stretch? crop? letterbox? Please check with Alan, but this seams wrong to me.  

(Note that making it the largest that fits within the allotted space on the app, while keeping the aspect ratio of the capture, is essentially the same as letter boxing, without the ugly black strips. If the .js communicates the dimensions of the capture to the .qml, the .qml can change properties of the preview attributes, triggering a dynamic layout update. (This is what the various crop/stretch/etc. options do internally on the 'Image' item, but which we'll have to do explicitly for the "4. Composite in QML".))

howard-stearns:
test plan?

scripts/system/spectatorCamera.js
zfox23: The current code attempts to do what you say ("make it the largest that fits within the allotted space on the [viewfinder]"), but doesn't succeed in all cases. The purpose of this current code is not to attempt to display something different than is being captured.

The above is true for the old code, but the old code failed in more cases. I implemented this as a quick improvement (and a way for myself and others to learn why other implementations are wrong), not meaning for it to be the final solution.  I haven't been able to figure out the final solution as of now. 

It is a good idea to get signoff from Alan before any MVP implementation.

interface/src/SecondaryCamera.cpp
zfox23: True. 

vFoV is asked for in degree format. 

Yes. 😊 I can make that change for this PR.

zfox23:
@howard-stearns we haven't previously had a test plan on any of the PRs to be merged into `spectator-camera`. Would you like to change that now moving forward? I would prefer not to have to take the time to write them - I think detailed comments about each PR here or in Slack are enough to know what to look for and test.

howard-stearns:
we were supposed to all along. the first two were weird, but after that we should have been doing so.

interface/src/SecondaryCamera.cpp
zfox23: I'll get rid of this for now, since it doesn't do anything yet. For all of us to keep in mind: This code might be useful in the future.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/692/HighFidelity-Beta-PR10678-ec3b488897cd25bc3d251ee5038093d18ae1b34e.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/692/HighFidelity-Beta-PR10678-ec3b488897cd25bc3d251ee5038093d18ae1b34e.dmg

interface/src/SecondaryCamera.h
```diff
@@ -29,13 +29,19 @@ class MainRenderTask {
...
     Q_OBJECT
     Q_PROPERTY(glm::vec3 position MEMBER position NOTIFY dirty)  // of viewpoint to render from
     Q_PROPERTY(glm::quat orientation MEMBER orientation NOTIFY dirty)  // of viewpoint to render from
+    Q_PROPERTY(float vFoV MEMBER vFoV NOTIFY dirty)  // Secondary camera's vertical field of view. In degrees.
+    Q_PROPERTY(float nearClipPlaneDistance MEMBER nearClipPlaneDistance NOTIFY dirty)  // Secondary camera's near clip plane distance. In meters.
+    Q_PROPERTY(float farClipPlaneDistance MEMBER farClipPlaneDistance NOTIFY dirty)  // Secondary camera's far clip plane distance. In meters.
 public:
     glm::vec3 position{};
     glm::quat orientation{};
+    float vFoV{};
```
howard-stearns: defaults of zero don't seem right to me for fov and clipping.

zfox23: OK

scripts/system/spectatorCamera.js
```diff
@@ -100,6 +106,9 @@
         windowAspectRatio = sizeX/sizeY;
         spectatorFrameRenderConfig.resetSizeSpectatorCamera(sizeX, sizeY);
         spectatorFrameRenderConfig.enabled = beginSpectatorFrameRenderConfig.enabled = true;
+        beginSpectatorFrameRenderConfig.vFoV = vFoV;
+        beginSpectatorFrameRenderConfig.nearClipPlaneDistance = nearClipPlaneDistance;
+        beginSpectatorFrameRenderConfig.farClipPlaneDistance = farClipPlaneDistance;
```
howard-stearns: I think we want to have reasonable defaults in the config object itself. I.e., that applications don't _need_ to specify a bunch of startup stuff.  (See comment above.) Given that, I don't think we need to reset to the to the same values here.

zfox23: It feels to me like setting them in the script is a good idea for clarity, no? i.e. if we want to change these values in the future, we don't have to add extra code, just replace numeric values in the script.

Resetting the projection parameters to the same default values here doesn't seem like an irresponsible amount of overhead for the clarity benefit it provides, IMO.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/706/HighFidelity-Beta-PR10678-03c61e21914dcd6be8e75a90784a101af022cce7.exe

scripts/system/spectatorCamera.js
zfox23: If you passionately disagree, I'll remove these and relevant lines from the script

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/706/HighFidelity-Beta-PR10678-03c61e21914dcd6be8e75a90784a101af022cce7.dmg

scripts/system/spectatorCamera.js
howard-stearns: Cool. I've added some comments/clarifications to that card. Let me know if I'm understanding correctly.

zfox23: Yes, you are understanding correctly, with one missing caveat: Even this existing code doesn't stretch perfectly in all cases. For example, if you make the window sufficiently small, the entire viewfinder preview shrinks, too. It's those kinds of issues that should be fixed, too.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/710/HighFidelity-Beta-PR10678-8a217acd2bef505696dae04ba2a19076c3a364ee.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/710/HighFidelity-Beta-PR10678-8a217acd2bef505696dae04ba2a19076c3a364ee.dmg

scripts/system/spectatorCamera.js
howard-stearns: cool

PR #10685, opened 2017-06-14
Use HMD roll angle to turn while flying or walking in HMD

ctrlaltdavid:
Makes you turn while walking or flying in the HMD depending on the amount you roll your head; this in addition to turning per where you're looking.

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -1971,19 +1971,19 @@ void MyAvatar::updateOrientation(float deltaTime) {
             float speedFactor;
             if (getDriveKey(TRANSLATE_Z) != 0.0f || _lastDrivenSpeed == 0.0f) {
                 _lastDrivenSpeed = speed;
-                speedFactor = 1.0;
+                speedFactor = 1.0f;
             } else {
                 speedFactor = glm::min(speed / _lastDrivenSpeed, 1.0f);
             }
 
             float direction = glm::dot(getVelocity(), getRotation() * Vectors::UNIT_NEG_Z) > 0.0f ? 1.0f : -1.0f;
 
-            float rollAngle = glm::degrees(asin(glm::dot(IDENTITY_UP, _hmdSensorOrientation * IDENTITY_RIGHT)));
+            float rollAngle = glm::degrees(asinf(glm::dot(IDENTITY_UP, _hmdSensorOrientation * IDENTITY_RIGHT)));
             float rollSign = rollAngle < 0.0f ? -1.0f : 1.0f;
             rollAngle = fabsf(rollAngle);
+            rollAngle = rollAngle > _hmdRollControlDeadZone ? rollSign * (rollAngle - _hmdRollControlDeadZone) : 0.0f;
 
-            float yawChange = rollAngle > _hmdRollControlDeadZone ? rollSign * (rollAngle - _hmdRollControlDeadZone) : 0.0f;
-            totalBodyYaw += speedFactor * direction * yawChange * deltaTime * _hmdRollControlSpeed;
+            totalBodyYaw += speedFactor * direction * rollAngle * deltaTime * _hmdRollControlSpeed;
         }
     }
 
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -1961,6 +1961,32 @@ void MyAvatar::updateOrientation(float deltaTime) {
         totalBodyYaw += (speedFactor * deltaAngle * (180.0f / PI));
     }
 
+    // Use head/HMD roll to turn while walking or flying.
+    if (qApp->isHMDMode() && _hmdRollControlEnabled) {
+        // Turn with head roll.
+        const float MIN_CONTROL_SPEED = 0.01f;
+        float speed = glm::length(getVelocity());
+        if (speed >= MIN_CONTROL_SPEED) {
+            // Feather turn when stopping moving.
+            float speedFactor;
+            if (getDriveKey(TRANSLATE_Z) != 0.0f || _lastDrivenSpeed == 0.0f) {
+                _lastDrivenSpeed = speed;
+                speedFactor = 1.0;
+            } else {
+                speedFactor = glm::min(speed / _lastDrivenSpeed, 1.0f);
+            }
+
+            float direction = glm::dot(getVelocity(), getRotation() * Vectors::UNIT_NEG_Z) > 0.0f ? 1.0f : -1.0f;
+
+            float rollAngle = glm::degrees(asin(glm::dot(IDENTITY_UP, _hmdSensorOrientation * IDENTITY_RIGHT)));
+            float rollSign = rollAngle < 0.0f ? -1.0f : 1.0f;
+            rollAngle = fabsf(rollAngle);
+
+            float yawChange = rollAngle > _hmdRollControlDeadZone ? rollSign * (rollAngle - _hmdRollControlDeadZone) : 0.0f;
+            totalBodyYaw += speedFactor * direction * yawChange * deltaTime * _hmdRollControlSpeed;
+        }
+    }
+
     // update body orientation by movement inputs
     glm::quat initialOrientation = getOrientationOutbound();
     setOrientation(getOrientation() * glm::quat(glm::radians(glm::vec3(0.0f, totalBodyYaw, 0.0f))));
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -1983,7 +1983,7 @@ void MyAvatar::updateOrientation(float deltaTime) {
             rollAngle = fabsf(rollAngle);
             rollAngle = rollAngle > _hmdRollControlDeadZone ? rollSign * (rollAngle - _hmdRollControlDeadZone) : 0.0f;
 
-            totalBodyYaw += speedFactor * direction * rollAngle * deltaTime * _hmdRollControlSpeed;
+            totalBodyYaw += speedFactor * direction * rollAngle * deltaTime * _hmdRollControlRate;
         }
     }
 
--- a/interface/src/avatar/MyAvatar.h
+++ b/interface/src/avatar/MyAvatar.h
@@ -703,6 +703,7 @@ private:
     bool _hmdRollControlEnabled { true };
     float _hmdRollControlDeadZone { ROLL_CONTROL_DEAD_ZONE_DEFAULT };
     float _hmdRollControlSpeed { ROLL_CONTROL_SPEED_DEFAULT };
+    float _lastDrivenSpeed { 0.0f };
 
     // working copies -- see AvatarData for thread-safe _sensorToWorldMatrixCache, used for outward facing access
     glm::mat4 _sensorToWorldMatrix { glm::mat4() };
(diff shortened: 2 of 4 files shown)


ctrlaltdavid:
**Testing**

This PR makes you turn while walking or flying in the HMD depending on the amount your roll your head; this in addition to turning per where you're looking.
When Interface starts up, HMD roll control is enabled by default.

Stationary, walking and flying:
- Rolling head has no effect.
- Turning head turns avatar.

Walking forwards/backwards in HMD:
- Rolling your head changes your direction.
- The more you roll your head the faster you turn.
- Looking left/right also changes your direction (same as before.)

Flying forwards/backwards in HMD:
- Rolling your head changes your direction.
- The more you roll your head the faster you turn.
- Looking left/right also changes your direction (same as before.)
- You fly up/down per your head pitch (same as before).

Test both with and without Settings > Advanced Movement For Hand Controllers enabled.
Test with and without Settings > Avatar > Snap turn when in HMD enabled.
Test non-HMD motion works as before.

Use the developer script, below, to enable/disable and change settings.
- When disabled, you turn with your head direction while walking and flying, as before.


**Notes**

The following JavaScript API object properties are provided to control this:
- MyAvatar.hmdRollControlEnabled boolean
- MyAvatar.hmdRollControlDeadZone number - degrees
- MyAvatar.hmdRollControlSpeed number - degrees/second/degree

The following developer scripts are provided to enable/disable and tweak dead zone and turn speed:
- /scripts/developers/hmdRollControlEnable.js (Edit to tweak dead zone and turn speed.)
- /scripts/developers/hmdRollControlDisable.js

These scripts run once then stop after changing the settings (i.e., no UI). The settings are NOT remembered for next time Interface is run (i.e., roll control defaults to enabled).

YuppyPlays:
Will test once built :)

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/679/HighFidelity-Beta-PR10685-395d40f0f9026f4ef3935522b7a2a02f4d3405b0.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/679/HighFidelity-Beta-PR10685-395d40f0f9026f4ef3935522b7a2a02f4d3405b0.dmg

YuppyPlays:
Testing.

YuppyPlays:
Stationary, walking and flying:
Rolling head has no effect. **Pass**
Turning head turns avatar. **Pass**

Walking forwards/backwards in HMD:
Rolling your head changes your direction. **Pass**
The more you roll your head the faster you turn. **Pass**
Looking left/right also changes your direction (same as before.) **Pass**

Flying forwards/backwards in HMD:
Rolling your head changes your direction. **Pass**
The more you roll your head the faster you turn. **Pass**
Looking left/right also changes your direction (same as before.) **Pass**
You fly up/down per your head pitch (same as before). **Pass**

Test both with and without Settings > Advanced Movement For Hand Controllers enabled. **Pass**
Test with and without Settings > Avatar > Snap turn when in HMD enabled. **Pass**
Test non-HMD motion works as before. **Pass**

Also tried out the Enable/Disable scripts as previously this did not work for me, but now does **Pass**

Please let me know if there is anything more I can do/provide to assist :)

PhilipRosedale:
Works great for me now!  Nice.

scripts/developer/hmdRollControlDisable.js
```diff
@@ -0,0 +1,17 @@
...
+// Created by David Rowe on 4 Jun 2017.
+// Copyright 2017 High Fidelity, Inc.
+//
+// Distributed under the Apache License, Version 2.0
+// See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+var hmdRollControlEnabled = false;
+
+print("HMD roll control: " + hmdRollControlEnabled);
```
PhilipRosedale: Remove debug print

scripts/developer/hmdRollControlEnable.js
```diff
@@ -0,0 +1,21 @@
...
+//
+// Distributed under the Apache License, Version 2.0
+// See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+var hmdRollControlEnabled = true;
+var hmdRollControlDeadZone = 8.0;  // deg
+var hmdRollControlSpeed = 2.5;  // deg/sec/deg
+
+print("HMD roll control: " + hmdRollControlEnabled + ", " + hmdRollControlDeadZone + ", " + hmdRollControlSpeed);
```
PhilipRosedale: remove debug print

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/685/HighFidelity-Beta-PR10685-e3b4e4aa2045ac14ba7c31350164192a1b193798.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/685/HighFidelity-Beta-PR10685-e3b4e4aa2045ac14ba7c31350164192a1b193798.exe

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1961,6 +1961,32 @@ void MyAvatar::updateOrientation(float deltaTime) {
...
+    if (qApp->isHMDMode() && _hmdRollControlEnabled) {
+        // Turn with head roll.
+        const float MIN_CONTROL_SPEED = 0.01f;
+        float speed = glm::length(getVelocity());
+        if (speed >= MIN_CONTROL_SPEED) {
+            // Feather turn when stopping moving.
+            float speedFactor;
+            if (getDriveKey(TRANSLATE_Z) != 0.0f || _lastDrivenSpeed == 0.0f) {
+                _lastDrivenSpeed = speed;
+                speedFactor = 1.0;
```
AndrewMeadows: Use `1.0f` rather than `1.0`.

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1961,6 +1961,32 @@ void MyAvatar::updateOrientation(float deltaTime) {
...
+            } else {
+                speedFactor = glm::min(speed / _lastDrivenSpeed, 1.0f);
+            }
+
+            float direction = glm::dot(getVelocity(), getRotation() * Vectors::UNIT_NEG_Z) > 0.0f ? 1.0f : -1.0f;
+
+            float rollAngle = glm::degrees(asin(glm::dot(IDENTITY_UP, _hmdSensorOrientation * IDENTITY_RIGHT)));
```
AndrewMeadows: Use `asinf()` rather than `asin()`.

interface/src/avatar/MyAvatar.h
```diff
@@ -687,6 +698,13 @@ public slots:
     bool _useSnapTurn { true };
     bool _clearOverlayWhenMoving { true };
 
+    const float ROLL_CONTROL_DEAD_ZONE_DEFAULT = 8.0f; // deg
+    const float ROLL_CONTROL_SPEED_DEFAULT = 2.5f; // deg/sec/deg
+    bool _hmdRollControlEnabled { true };
+    float _hmdRollControlDeadZone { ROLL_CONTROL_DEAD_ZONE_DEFAULT };
+    float _hmdRollControlSpeed { ROLL_CONTROL_SPEED_DEFAULT };
+    float _lastDrivenSpeed { 0.0f };
```
AndrewMeadows: The _lastDrivenSpeed variable is interesting. I think I understand what it is doing but I'm curious to hear your explanation.

Meanwhile _hmdRollControlSpeed isn't really as "speed"... it is more of a "throttle coefficient" relating the head roll angle to a corresponding yaw speed, but it effectively has a "max yaw speed" mixed into it. That might be ok... but it makes the variable more of a "control knob with magical units" for anyone who uses it. Ultimately it makes the feature a bit wonky for any user who wants to tune it... but how many will bother? Since I view this feature as more of an experiment than the final say in motion control I say, "Let it slide."

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1961,6 +1961,32 @@ void MyAvatar::updateOrientation(float deltaTime) {
...
+            float rollSign = rollAngle < 0.0f ? -1.0f : 1.0f;
+            rollAngle = fabsf(rollAngle);
+
+            float yawChange = rollAngle > _hmdRollControlDeadZone ? rollSign * (rollAngle - _hmdRollControlDeadZone) : 0.0f;
+            totalBodyYaw += speedFactor * direction * yawChange * deltaTime * _hmdRollControlSpeed;
```
AndrewMeadows: `yawChange` is a misnomer.  It is actually a measure of `effectiveRollAngle` that is used as the throttle for yaw velocity after the dead-zone has been removed.  The product below that is added to `totalBodyYaw` is the "change in Yaw" from this frame.  The code would make more sense if `yawChange` were renamed: maybe just store the result in `rollAngle`?

zfox23:
Oh yes. This is SUPER fun. Great stuff, @ctrlaltdavid.

I'm still a little concerned about the difference in rate-of-change between the horizontal and vertical axes, but it's not a huge deal.

ctrlaltdavid:
@zfox23 I don't understand. Move-in-the-direction-you're looking (head yaw and pitch) hasn't changed. Only thing added is head roll turning you left/right.

interface/src/avatar/MyAvatar.h
ctrlaltdavid: `_lastDrivenSpeed` is the most recent speed of your avatar prior to you releasing your press on the drive control. It is used to decrease the amount of yaw each frame as your speed in each frame decreases towards 0.0. ... Otherwise, if you keep your head roll constant and take your finger off the accelerator, your turn rate ends abruptly.

ctrlaltdavid: I could perhaps change `_hmdRollControlSpeed` to `_hmdRollControlRate`?

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/715/HighFidelity-Beta-PR10685-16bebdd40945cf19878abf364f8dc4ad56f02e18.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/715/HighFidelity-Beta-PR10685-16bebdd40945cf19878abf364f8dc4ad56f02e18.exe

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

(Similar results as before but noticed something new and added as a note)

Stationary, walking and flying:
Rolling head has no effect. Pass
Turning head turns avatar. Pass

Walking forwards/backwards in HMD:
Rolling your head changes your direction. Pass
The more you roll your head the faster you turn. Pass
Looking left/right also changes your direction (same as before.) Pass

Flying forwards/backwards in HMD:
Rolling your head changes your direction. **Pass**
The more you roll your head the faster you turn. **Pass**
Looking left/right also changes your direction (same as before.) P**ass**
You fly up/down per your head pitch (same as before). **Pass**

Test both with and without Settings > Advanced Movement For Hand Controllers enabled. **Pass**
Test with and without Settings > Avatar > Snap turn when in HMD enabled. **Pass**
Test non-HMD motion works as before. **Pass**

Also tried out the Enable/Disable scripts as previously this did not work for me, but now does **Pass**

Please let me know if there is anything more I can do/provide to assist :)


Notes: I found if I strafed Left and rolled/tilted my head right, it would cause a bit of jittery movement and same with strafe right and roll/tilt head left.

Overall, I think that's a solid **Pass**

PR #10714, opened 2017-06-15
add broadcasting/replication to downstream avatar mixers and audio mixers

birarda:
- added re-broadcasting of avatar and audio data for selected logged in users in your domain.


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/776/HighFidelity-Beta-PR10714-cf4bc34572bd290ed205b67906e867487b1e87ba.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/776/HighFidelity-Beta-PR10714-cf4bc34572bd290ed205b67906e867487b1e87ba.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/782/HighFidelity-Beta-PR10714-f450e70f1e34b04d2fdea4884a6de478eddce57e.dmg

birarda:
test this please

assignment-client/src/audio/AudioMixer.cpp
```diff
@@ -103,6 +118,41 @@ void AudioMixer::queueAudioPacket(QSharedPointer<ReceivedMessage> message, Share
...
+                                                    message->getSenderSockAddr(), message->getSenderSockAddr(),
+                                                    true, true);
+    replicatedNode->setLastHeardMicrostamp(usecTimestampNow());
+
+    // construct a "fake" audio received message from the byte array and packet list information
+    auto audioData = message->getMessage().mid(NUM_BYTES_RFC4122_UUID);
+
+    PacketType rewrittenType;
+
+    if (message->getType() == PacketType::ReplicatedMicrophoneAudioNoEcho) {
```
ZappoMan: maybe this should be done with a table? or a switch? this repeated if approach just looks out of place

assignment-client/src/audio/AudioMixerClientData.cpp
```diff
@@ -97,6 +109,69 @@ void AudioMixerClientData::processPackets() {
...
+
+    // first, make sure that this is a packet from a node we are supposed to replicate
+    if (node.isReplicated()) {
+
+        auto nodeList = DependencyManager::get<NodeList>();
+
+        // now make sure it's a packet type that we want to replicate
+        PacketType mirroredType;
+
+        switch (message.getType()) {
```
ZappoMan: definitely feels like it would be easier to maintain if we had a common table of message types to their replicated equivalent... since we've got multiple places where people will have to remember to add new message types

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/793/HighFidelity-Beta-PR10714-f450e70f1e34b04d2fdea4884a6de478eddce57e.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/793/HighFidelity-Beta-PR10714-f450e70f1e34b04d2fdea4884a6de478eddce57e.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/871/HighFidelity-Beta-PR10714-b3824c6a25a7ac533e70c763af9b154e58c05ce8.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/871/HighFidelity-Beta-PR10714-b3824c6a25a7ac533e70c763af9b154e58c05ce8.dmg

Atlante45:
## Test Plan:

Requirements:
- You need access to several machines with publically (locally if on the same network) accessible IPs.
    - 3 where you can run the server.
    - At least 2 where you can run interface.

### Test 1: A -> B,C

1) Launch all 3 servers (You'll have to run them manually to pin the port of the Audio Mixer and the Avatar Mixer)
    - On Windows (You can replace `7000` and `7001` by whatever ports you want to use):
```
mintty --hold always --exec /c/Program\ Files/High\ Fidelity\ -\ PR10714/domain-server.exe &
mintty --hold always --exec /c/Program\ Files/High\ Fidelity\ -\ PR10714/assignment-client.exe -t0 -p 7000 &
mintty --hold always --exec /c/Program\ Files/High\ Fidelity\ -\ PR10714/assignment-client.exe -t1 -p 7001 &
mintty --hold always --exec /c/Program\ Files/High\ Fidelity\ -\ PR10714/assignment-client.exe -t2 &
mintty --hold always --exec /c/Program\ Files/High\ Fidelity\ -\ PR10714/assignment-client.exe -t3 &
mintty --hold always --exec /c/Program\ Files/High\ Fidelity\ -\ PR10714/assignment-client.exe -t4 &
mintty --hold always --exec /c/Program\ Files/High\ Fidelity\ -\ PR10714/assignment-client.exe -t5 &
mintty --hold always --exec /c/Program\ Files/High\ Fidelity\ -\ PR10714/assignment-client.exe -t6
```
- For the rest of this test plan, we'll assume that
    - Server A's Audio Mixer has IP:port combination: A0:a0
    - Server A's Avatar Mixer has IP:port combination: A1:a1
    - Server B's Audio Mixer have IP:port combination: B0:b0
    - Server B's Avatar Mixer have IP:port combination: B1:b1
    - Server C's Audio Mixer have IP:port combination: C0:c0
    - Server C's Avatar Mixer have IP:port combination: C1:c1
2) Verify that all 3 servers are running.
    - Connect to each of them with 2 clients
    - Verify each client can see/hear the other
3) Go to Server A with a client
4) Go to the settings page of Server A
    - Go to the `Broadcasting` section at the bottom
    - Add Server B and C as receiving servers:
receivers
    - Make sure Server A doesn't restart when you save the new settings
5) Leave a client connected to Server A at some location (ex: `/100,100,100`)
6) Go to the same location on Server B and C. You shouldn't see/hear anything.
7) Go to the settings page of Server A
    - Go to the `Broadcasting` section at the bottom
    - Make sure the client in Server A is logged in
    - Add the client's username to the list:
user
    - Make sure Server A doesn't restart when you save the new settings
8) The client on Server B/C should see the other avatar appear.
9) Confirm that you can see/hear the avatar in Server A from server B and C.
10) Go to the settings page of Server A
    - Go to the `Broadcasting` section at the bottom
    - Remove the client's username from the list
11) The client on Server B/C should see the other avatar disappear.

### Test 2: A -> B -> C: 

Requirements:
- Make sure to clear the broadcasting settings from the previous test

For this test you will execute the same steps as the previous plan with some modifications to step 4:
4) Go to the settings page of Server A
    - Add Server B as receiving servers:
receivers
    - Go to the settings page of Server B
    - Add Server C as receiving servers:
receivers

domain-server/src/DomainServer.cpp
```diff
@@ -2210,6 +2222,125 @@ void DomainServer::refreshStaticAssignmentAndAddToQueue(SharedAssignmentPointer&
...
+        }
+    );
+}
+
+bool DomainServer::shouldReplicateNode(const Node& node) {
+    QString verifiedUsername = node.getPermissions().getVerifiedUserName();
+    // Both the verified username and usernames in _replicatedUsernames are lowercase, so
+    // comparisons here are case-insensitive.
+    auto it = find(_replicatedUsernames.cbegin(), _replicatedUsernames.cend(), verifiedUsername);
+    return it != _replicatedUsernames.end() && node.getType() == NodeType::Agent;
```
AndrewMeadows: Why perform the relatively expensive `find()` when the type check is fast and would make it unnecessary?  How about something like:
```
if (node.getType() != NodeType::Agent) {
    return false;
}
 QString verifiedUsername = node.getPermissions().getVerifiedUserName();
// Both the verified username and usernames in _replicatedUsernames are lowercase, so
// comparisons here are case-insensitive.
auto it = find(_replicatedUsernames.cbegin(), _replicatedUsernames.cend(), verifiedUsername);
return it != _replicatedUsernames.end();
```

domain-server/src/DomainServer.cpp
```diff
@@ -2210,6 +2222,125 @@ void DomainServer::refreshStaticAssignmentAndAddToQueue(SharedAssignmentPointer&
...
+}
+
+void DomainServer::updateReplicatedNodes() {
+    // Make sure we have downstream nodes in our list
+    // TODO Move this to a different function
+    _replicatedUsernames.clear();
+    auto settings = _settingsManager.getSettingsMap();
+
+    static const QString REPLICATED_USERS_KEY = "users";
+    _replicatedUsernames.clear();
```
AndrewMeadows: There are two `_replicatedUsernames.clear();` above.  Are both necessary?

libraries/avatars/src/AvatarData.cpp
```diff
@@ -1504,62 +1483,108 @@ QUrl AvatarData::cannonicalSkeletonModelURL(const QUrl& emptyURL) const {
...
+
+    QUuid avatarSessionID;
+
+    // peek the sequence number, this will tell us if we should be processing this identity packet at all
+    udt::SequenceNumber::Type incomingSequenceNumberType;
+    packetStream >> avatarSessionID >> incomingSequenceNumberType;
+    udt::SequenceNumber incomingSequenceNumber(incomingSequenceNumberType);
+
+    if (!_hasProcessedFirstIdentity) {
+        _lastIncomingSequenceNumber = incomingSequenceNumber - 1;
```
AndrewMeadows: Is it safe to assume that `incomingSequenceNumber` will never be zero?  Because if not... maybe the line above would prevent future identity packets from being processed correctly?

libraries/avatars/src/AvatarData.cpp
```diff
@@ -1504,62 +1483,108 @@ QUrl AvatarData::cannonicalSkeletonModelURL(const QUrl& emptyURL) const {
...
+#ifdef WANT_DEBUG
+        qCDebug(avatars) << __FUNCTION__
+            << "identity.uuid:" << identity.uuid
+            << "identity.skeletonModelURL:" << identity.skeletonModelURL
+            << "identity.displayName:" << identity.displayName
+            << "identity.sessionDisplayName:" << identity.sessionDisplayName;
+#endif
 
+    } else {
+#ifdef WANT_DEBUG
```
AndrewMeadows: I don't think these middle `#endif` and `#ifdef` lines are necessary.  You could get by with one big `#ifdef` block.

libraries/avatars/src/AvatarData.h
```diff
@@ -785,7 +781,9 @@ public slots:
     float _audioAverageLoudness { 0.0f };
 
     bool _identityDataChanged { false };
-    quint64 _identitySequenceId { 0 };
+    udt::SequenceNumber _lastIncomingSequenceNumber { 0 };
+    udt::SequenceNumber _lastOutgoingSequenceNumber { 0 };
```
AndrewMeadows: The two new sequence numbers appear to be parallel and do not interact... which means we just need one.  My conclusion is that we could have got by with keeping `_identitySequenceId`.  I'm only 99% sure.

libraries/networking/src/LimitedNodeList.h
```diff
@@ -386,6 +400,9 @@ protected slots:
         }
     }
 
+
+    HifiSockAddr _mirrorSocket;
```
AndrewMeadows: The `_mirrorSocket` feature appears to be unused.

libraries/networking/src/NetworkPeer.h
```diff
@@ -76,6 +76,9 @@ class NetworkPeer : public QObject {
     float getOutboundBandwidth() const; // in kbps
     float getInboundBandwidth() const; // in kbps
 
+    bool isForcedNeverSilent() const { return _isForcedNeverSilent; }
+    void setIsForcedNeverSilent(bool isForcedNeverSilent) { _isForcedNeverSilent = isForcedNeverSilent; }
```
AndrewMeadows: `isForcedNeverSilent`.  I don't understand what that means, even when I read the code to see how it is used.  Maybe  we could come up with a better name?

libraries/networking/src/udt/PacketList.h
```diff
@@ -74,6 +76,8 @@ class PacketList : public QIODevice {
     
     PacketType _packetType;
     std::list<std::unique_ptr<Packet>> _packets;
+
+    bool _isOrdered = false;
```
AndrewMeadows: Why is `_isOrdered` not private in compliance with [coding standard section 3.2.3](https://github.com/highfidelity/hifi/wiki/Coding-Standard#323-class-variables-should-never-be-declared-public)?

domain-server/src/DomainServer.cpp
```diff
@@ -2210,6 +2222,125 @@ void DomainServer::refreshStaticAssignmentAndAddToQueue(SharedAssignmentPointer&
...
+                                           otherNode->getPublicSocket()) != downstreamNodesInSettings.cend();
+                if (!nodeInSettings) {
+                    qDebug() << "Removing downstream node:" << otherNode->getUUID() << otherNode->getPublicSocket();
+                    nodesToKill.push_back(otherNode);
+                }
+            }
+        });
+        for (auto& node : nodesToKill) {
+            handleKillNode(node);
+        }
```
birarda: I'll add a comment to the code, basically handleKill takes a write lock to remove the nodes from the unordered map. The public methods that enumerate the NodeList (like eachNode) only take a read lock and cannot recursively take a write lock.

libraries/avatars/src/AvatarData.cpp
birarda: The `SequenceNumber` class handles this. It's an unsigned 32 bit integer (minus some header bits) and the class handles the comparison. 

We assume that the minimum distance between A and B is the correct offset, meaning that a sequence number of 0 is correctly seen as the number following the maximum possible value (which is 134217728).

libraries/networking/src/udt/PacketList.h
birarda: It's protected, not public - so that it is usable by NLPacketList

libraries/avatars/src/AvatarData.h
birarda: You're right that we only need one though we could not have kept `_identitySequenceId` as it was. There was a bug where a receiving client would push the number every time it received an a change to one of the components of the identity packet (which can occur multiple times per packet). This caused the client to sometimes drop identity packets that it should have processed, because the sequence numbers were out of sync.

libraries/networking/src/NetworkPeer.h
birarda: Spent a good amount of time on this - though I agree the name isn't ideal I'm not sure what's better. I'll add a comment here.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/977/HighFidelity-Beta-PR10714-06a305d04b33127b5ebc6a08affdcd2acb628723.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/977/HighFidelity-Beta-PR10714-06a305d04b33127b5ebc6a08affdcd2acb628723.dmg

birarda:
Testing this internally with @Atlante45 given the complexity required to complete the test plan.

PR #10718, opened 2017-06-16
Cleanup and refactoring

jherico:
This is work that grew out of a project to move scene traversal to it's own thread, and off the main thread.  

## Fixes

Fixed a bug that will cause interface to hang if you startup with the shift key held.  I happened to discover this completely by accident while I was testing the other changes.

## CMake fixes

It turns out that the way we set up our `link_hifi_libraries` macro in CMake had a scoping / recursion bug in it.  The result was that many projects had semi-random (but deterministic) sets of dependency that would change depending on the order they were declared and encountered. This led to lots of situations where you'd change the order of some dependencies or add a new dependency and suddenly seemingly unrelated projects would break and stop compiling properly.

I've fixed the bug by making the  `link_hifi_libraries` call into a CMake function instead of a macro, as well as moving the recursion outside of the for loop where we apply the dependencies to the project.  

As a result, many of the implicit dependencies that were present are now gone, and I've had to do a pass to add them all back as explicit header include depedencies.  

## Cleanup changes
* Move `RenderArgs` out of `shared` and into `render`
  * `RenderArgs` required bizarre amounts of forward declarations because the concept of rendering was baked into the `OctreeRenderer` class (see below).
* Remove `RenderArgs::_whiteTexture`
  * This member was used in a few places to 'reset' a texture slot, but it was nowhere initialized, so it was always empty.  Where it was previously referenced, we're now explicitly setting the texture to nothing.
* Remove `QSharedPointer<OctreeRenderer> RenderArgs::_renderer;` in favor of `        `QSharedPointer<QObject> RenderArgs::_renderData;`
  * The `_renderer` type was always cast to a `QSharedPointer<EntityTreeRenderer>` in places where it was used, so there's no need to have this forward declaration anymore.
* Remove the concept of rendering from `octree` entirely
  * The class `OctreeRenderer` has been renamed `OctreeProcessor` (for lack of a better idea)
  * Move the `OctreeHeadlessViewer` and `EntityTreeHeadlessViewer` out of `octree` and `entities` and into `assignment-client`
  * Remove `OctreeRenderer::_viewFrustum`.  It was assigned in several places but never read, with clients using `RenderArgs::_viewFrustum` instead
  * Remove `OctreeRenderer::render()`.  It was used in only one place, the `LocalModelsOverlay` but ultimately it traversed the octree but did nothing, because it called `OctreeRenderer::renderElement` which had a default empty implementation as was never overridden.  
* Remove the concept of rendering from `entities` almost entirely
  * Ultimately I had to leave an opaque type `RenderableEntityInterface` in the `EntityItem` class, because I could not get dynamic casting from `EntityItem*` to `RenderableEntityInterface` to work in `entities-renderer`
* Remove `LocalModelsOverlay` and `Application::_entityClipboard`.  As mentioned above, these ultimately performed no rendering and weren't used for anything else.

## Performance changes
* Replace almost every instance of `OctreeElementPointer` when used as a parameter to `const OctreeElementPointer&`  
 * The implicit copying of these pointer whenever we're doing operations on the tree has a small but non-zero cost.  
* Change `RecurseOctreeOperation`
  * Old `typedef bool (*RecurseOctreeOperation)(OctreeElementPointer element, void* extraData);`
  * New `using RecurseOctreeOperation = std::function<bool(const OctreeElementPointer&, void*)>;`
  * Previously you could only pass an actual function pointer with a very specific signature.  With these changes will allow you to pass any of the following
    * `bool foo(OctreeElementPointer element, void* extraData)`
    * `bool bar(const OctreeElementPointer& element, void * extraData)`
    * `[](const OctreeElementPointer& element, void * extraData)->bool { return false; }`
* Replace almost every instance of `RecurseOctreeOperation` when used as a function parameter to `const RecurseOctreeOperation&`
* Replace many instances of `EntityItemPointer` when used as a parameter to `const EntityItemPointer&`  
  * This work wasn't nearly as exhaustive as the similar octree work.  There are still lots of raw `EntityItemPointer` function parameters

## Testing

Application behavior should be unchanged except for the small fix.  To test...

* Start Interface.  
* As soon as you've clicked on the icon to start, press and hold the shift key

In the current build, interface will become unresponsive and eventually crash from deadlock detection.  In this PR build the application should load and render normally.


interface/src/Application.cpp
```diff
@@ -5019,9 +5006,6 @@ QRect Application::getDesirableApplicationGeometry() const {
     return applicationGeometry;
 }
 
-// FIXME, preprocessor guard this check to occur only in DEBUG builds
-static QThread * activeRenderingThread = nullptr;
```
jherico: assigned but never read

interface/src/Application.h
```diff
@@ -535,7 +534,6 @@ private slots:
     PhysicalEntitySimulationPointer _entitySimulation;
     PhysicsEnginePointer _physicsEngine;
 
-    EntityTreeRenderer _entityClipboardRenderer;
```
jherico: Ultimately never used except by LocalModelsOverlay, which did nothing.

assignment-client/src/octree/OctreeHeadlessViewer.h
```diff
@@ -71,6 +60,7 @@ public slots:
     JurisdictionListener* _jurisdictionListener = nullptr;
     OctreeQuery _octreeQuery;
 
+    ViewFrustum _viewFrustum;
```
jherico: I'm not certain this is used anywhere, but it was maintained previously by the old `OctreeRenderer` class.

interface/src/scripting/MenuScriptingInterface.cpp
```diff
@@ -152,9 +152,13 @@ void MenuScriptingInterface::closeInfoView(const QString& path) {
 }
 
 bool MenuScriptingInterface::isInfoViewVisible(const QString& path) {
+    if (QThread::currentThread() == qApp->thread()) {
```
jherico: This is the bug fix.  If you hold down `shift` scripts will load in 'debug' mode meaning they're running on the main thread.  As such, calling `QMetaObject::invokeMethod` with `Qt::BlockingQueuedConnection` from such a script will hang the application.  Most of the `MenuScriptingInterface` methods already avoid this using the same boilerplate functionality that's shown here.

interface/src/ui/overlays/LocalModelsOverlay.cpp
```diff
@@ -1,50 +0,0 @@
...
-}
-
-void LocalModelsOverlay::render(RenderArgs* args) {
-    if (_visible) {
-        auto batch = args ->_batch;
-
-        Transform transform = Transform();
-        transform.setTranslation(args->getViewFrustum().getPosition() + getPosition());
-        batch->setViewTransform(transform);
-        _entityTreeRenderer->render(args);
```
jherico: The `OctreeRenderer::render();` method called here ultimately does nothing.  It traverses the tree calling `OctreeRenderer::renderElement()` for each item that is valid, but the default `renderElement` implementation did nothing and was never overridden.  As such `LocalModelsOverlay` is a giant expensive no-op.

libraries/gpu-gl/src/gpu/gl45/GL45BackendTexture.cpp
```diff
@@ -123,7 +123,7 @@ GLuint GL45Texture::allocate(const Texture& texture) {
     glCreateTextures(getGLTextureType(texture), 1, &result);
 #ifdef DEBUG
     auto source = texture.source();
-    glObjectLabel(GL_TEXTURE, result, source.length(), source.data());
+    glObjectLabel(GL_TEXTURE, result, (GLsizei)source.length(), source.data());
```
jherico: generated a warning in debug builds

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/785/HighFidelity-Beta-PR10718-fece4386023d12d4a73b89dfb634e6e96b8bcc93.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/785/HighFidelity-Beta-PR10718-fece4386023d12d4a73b89dfb634e6e96b8bcc93.exe

interface/src/ui/overlays/Image3DOverlay.cpp
```diff
@@ -112,7 +112,7 @@ void Image3DOverlay::render(RenderArgs* args) {
         _geometryId
     );
 
-    batch->setResourceTexture(0, args->_whiteTexture); // restore default white color after me
+    batch->setResourceTexture(0, nullptr); // restore default white color after me
```
samcake: i m afraid we need to specifically test this case because i think it was done on purpose for the case of the OVerlay

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

Tested an older PR just to confirm how the bug acts (Crashed / deadlocked as expected, had to click X)
Tested on this PR: Works perfectly, loads up as normal **Pass**.

libraries/render/CMakeLists.txt
```diff
@@ -3,6 +3,6 @@ AUTOSCRIBE_SHADER_LIB(gpu model)
 setup_hifi_library()
 
 # render needs octree only for getAccuracyAngle(float, int)
-link_hifi_libraries(shared ktx gpu model octree)
+link_hifi_libraries(shared networking ktx gpu model octree)
```
samcake: How can we avoid this dependency

Render lib  is not supposed to depend on networking

libraries/render/src/render/Forward.h
```diff
@@ -0,0 +1,26 @@
...
+#include <memory>
+
+namespace render {
+    class Scene;
+    using ScenePointer = std::shared_ptr<Scene>;
+    class ShapePipeline;
+    class Args;
+}
+
+class OctreeRenderer;
```
samcake: THis cannot be there

libraries/render-utils/src/MeshPartPayload.cpp
```diff
@@ -557,6 +557,7 @@ void ModelMeshPartPayload::render(RenderArgs* args) {
 
     if (_fadeState == FADE_WAITING_TO_START) {
         if (model->isLoaded()) {
+            // FIXME as far as I can tell this is the ONLY reason render-util depends on entities.
```
samcake: Great, one step closer !

interface/src/ui/overlays/Image3DOverlay.cpp
jherico: It's possible that it's specifically supposed to be a white texture and that there's some bug that it counters (like with websites with alpha values in them or something) but this change is at worst merely maintaining the current state of the code, because args->_whiteTexture was always an empty pointer.

libraries/render/src/render/Forward.h
jherico: fixed

libraries/render/CMakeLists.txt
jherico: fixed

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/789/HighFidelity-Beta-PR10718-a2547df1a203e4499e09f9d0ea77e03e6aab0875.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/789/HighFidelity-Beta-PR10718-a2547df1a203e4499e09f9d0ea77e03e6aab0875.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/818/HighFidelity-Beta-PR10718-a69cf31f4bb6aedd4878ce4bfd1b8d0d851eb778.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/818/HighFidelity-Beta-PR10718-a69cf31f4bb6aedd4878ce4bfd1b8d0d851eb778.dmg

PR #10727, opened 2017-06-16
Prevent deadlocks from different locking orders in FileCache

jherico:
Recent deadlock stack dumps have shown that depending on the usage pattern, the FileCache can deadlock because it has two mutexes, and one thread will be waiting on A while trying to get B, while another has B and is waiting on A.  

This PR changes to a single mutex (which is recursive) so that this pattern can't happen.  It also modifies the locations where the mutex is used.  All publicly visible functions now lock the mutex for the duration of public call, ensuring that private functions can work without fear of another thread modifying an object in the background.  The only exception is `addUnusedFile` which is a non-public function but still locks the mutex because it can be called the the destructor of a `File` object invoked by a shared pointer to a File leaving scope

This PR resolves issues with deadlocks and likely with missing avatars as well (loading avatars will trigger several texture loads concurrently, which is more likely to cause a deadlock.  Since the textures never finish loading, the avatar never becomes renderable).

## Testing

Unfortunately, there's no reliable way to trigger the deadlock.  Likely scenarios would involve going to high texture content domains or domains with many avatars (causing lots of concurrent texture loading).  It will be even more likely if the textures are already locally cached so that network transmission delays don't impact the testing. Reloading a test domain like dev-playa over and over, or going to a domain with lots of recorded avatars will likely eventually deadlock during the loading process, when running the master build.  

When running this build, there should be no deadlock (or if there is one, let us know and we'll try to inspect the bugsplat to verify that it's a different deadlock than this one)

### DIFF
--- a/libraries/networking/src/FileCache.cpp
+++ b/libraries/networking/src/FileCache.cpp
@@ -285,10 +285,12 @@ void FileCache::wipe() {
 }
 
 void FileCache::clear() {
+    Lock lock(_mutex);
+
     // Eliminate any overbudget files
     clean();
 
-    // Mark everything remaining as persisted
+    // Mark everything remaining as persisted while effectively ejecting from the cache
     for (auto& file : _unusedFiles) {
         file->_shouldPersist = true;
         file->_cache = nullptr;
--- a/libraries/networking/src/FileCache.cpp
+++ b/libraries/networking/src/FileCache.cpp
@@ -87,6 +87,12 @@ FileCache::~FileCache() {
 }
 
 void FileCache::initialize() {
+    Lock lock(_mutex);
+    if (_initialized) {
+        qCWarning(file_cache) << "File cache already initialized";
+        return;
+    }
+
     QDir dir(_dirpath.c_str());
 
     if (dir.exists()) {
@@ -120,21 +126,24 @@ FilePointer FileCache::addFile(Metadata&& metadata, const std::string& filepath)
         file->_cache = this;
         emit dirty();
 
-        Lock lock(_filesMutex);
         _files[file->getKey()] = file;
     }
     return file;
 }
 
 FilePointer FileCache::writeFile(const char* data, File::Metadata&& metadata, bool overwrite) {
-    assert(_initialized);
+    Lock lock(_mutex);
+
+    FilePointer file;
+    if (!_initialized) {
+        qCWarning(file_cache) << "File cache used before initialization";
+        return file;
+    }
 
     std::string filepath = getFilepath(metadata.key);
 
-    Lock lock(_filesMutex);
-
     // if file already exists, return it
-    FilePointer file = getFile(metadata.key);
+    file = getFile(metadata.key);
     if (file) {
         if (!overwrite) {
             qCWarning(file_cache, "[%s] Attempted to overwrite %s", _dirname.c_str(), metadata.key.c_str());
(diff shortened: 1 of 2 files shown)


libraries/networking/src/FileCache.cpp
```diff
@@ -188,31 +203,19 @@ std::string FileCache::getFilepath(const Key& key) {
...
+    _files[file->getKey()] = file;
+    _unusedFiles.insert(file);
+    _numUnusedFiles += 1;
+    _unusedFilesSize += file->getLength();
     clean();
 
     emit dirty();
 }
 
-void FileCache::removeUnusedFile(const FilePointer& file) {
```
jherico: this logic was only used in once place and the code has been made inline

libraries/networking/src/FileCache.h
```diff
@@ -125,16 +124,14 @@ class FileCache : public QObject {
     std::atomic<size_t> _totalFilesSize { 0 };
     std::atomic<size_t> _unusedFilesSize { 0 };
 
-    std::string _ext;
-    std::string _dirname;
-    std::string _dirpath;
+    const std::string _ext;
```
jherico: const objects are safer in MT code because you know they can't change on different threads

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/811/HighFidelity-Beta-PR10727-a675dfe7d2152fa2850f9dcd8cd06da26109128c.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/811/HighFidelity-Beta-PR10727-a675dfe7d2152fa2850f9dcd8cd06da26109128c.dmg

YuppyPlays:
Can test, any specific test plan?

jherico:
@YuppyPlays I've updated the comment.

YuppyPlays:
@jherico  Thank you :) Testing now.

YuppyPlays:
Testing complete:

**Not sure what to mark this as- I experienced a crash but unsure if related to this PR**

Loaded in and out of Dev-playa to sandbox 5 times, then wnt into Dev-Welcome and experienced a crash. Sent off a report and here is the log:
[hifi-log.txt](https://github.com/highfidelity/hifi/files/1082200/hifi-log.txt)

Please let me know if there is anything more I can do/provide to asssist.

libraries/networking/src/FileCache.cpp
```diff
@@ -294,7 +289,6 @@ void FileCache::clear() {
     clean();
 
     // Mark everything remaining as persisted
-    Lock unusedFilesLock(_unusedFilesMutex);
```
Atlante45: AFAICT this only runs in the dtor, but I'd leave a lock here in case it gets used somewhere else in the future.

jherico:
@cozza13 the crash was recorded [here](https://www.bugsplat.com/individualCrash/?id=12353&database=interface_alpha).  It appears to be related to [this bug](https://highfidelity.fogbugz.com/f/cases/5381/CRASH-interface-storage-Storage-createView-23-Crashes), which is unrelated to the [deadlock bug](https://highfidelity.fogbugz.com/f/cases/5410/Another-deadlock-issue-Interface-DeadlockWatchdogThread-run-345) that this PR is trying to fix.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/828/HighFidelity-Beta-PR10727-1632e3b35a8a4154daab42638df4f82e0ed5b109.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/828/HighFidelity-Beta-PR10727-1632e3b35a8a4154daab42638df4f82e0ed5b109.exe

PR #10738, opened 2017-06-19
Only change spectator cam config when parameters change

zfox23:
Un-prototype this specific part of the `spectator-camera` branch.

**Test Plan:**
- Do a quick smoke test of Spectator Camera functionality (start it up, move it around, verify that camera shows up on the main monitor when relevant settings are set, etc).

### DIFF
--- a/interface/src/SecondaryCamera.cpp
+++ b/interface/src/SecondaryCamera.cpp
@@ -54,11 +54,21 @@ public:
 
     void configure(const Config& config) {
         if (config.enabled || config.alwaysEnabled) {
-            _position = config.position;
-            _orientation = config.orientation;
-            _vFoV = config.vFoV;
-            _nearClipPlaneDistance = config.nearClipPlaneDistance;
-            _farClipPlaneDistance = config.farClipPlaneDistance;
+            if (config.position != _position) {
+                _position = config.position;
+            }
+            if (config.orientation != _orientation) {
+                _orientation = config.orientation;
+            }
+            if (config.vFoV != _vFoV) {
+                _vFoV = config.vFoV;
+            }
+            if (config.nearClipPlaneDistance != _nearClipPlaneDistance) {
+                _nearClipPlaneDistance = config.nearClipPlaneDistance;
+            }
+            if (config.farClipPlaneDistance != _farClipPlaneDistance) {
+                _farClipPlaneDistance = config.farClipPlaneDistance;
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


CainFoool:
Testing shortly.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/842/HighFidelity-Beta-PR10738-e69cde6c445b6594e992859153d442328b422656.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/842/HighFidelity-Beta-PR10738-e69cde6c445b6594e992859153d442328b422656.exe

CainFoool:
# Test Passed
Everything seemed to work okay as expected. From previous PR (The text not changing properly) that issue was occurring again, though expected as the code wasn't ported over until the stable merge.

Switching from HMD View to Camera View works as intended.
Switching the Camera Off and On works as expected.
Switching over to Desktop Mode with the Camera On removes the Camera, works as expected.

interface/src/SecondaryCamera.cpp
```diff
@@ -54,11 +54,21 @@ class BeginSecondaryCameraFrame {  // Changes renderContext for our framebuffer
...
+            if (config.vFoV != _vFoV) {
+                _vFoV = config.vFoV;
+            }
+            if (config.nearClipPlaneDistance != _nearClipPlaneDistance) {
+                _nearClipPlaneDistance = config.nearClipPlaneDistance;
+            }
+            if (config.farClipPlaneDistance != _farClipPlaneDistance) {
+                _farClipPlaneDistance = config.farClipPlaneDistance;
+            }
         }
```
howard-stearns: What is being gained with the inequality tests? Did you do any tests to verify that theory?

Alternatively, it is conceivable that there are performance issues in running configure at all. For example, what if config takes a lock that stops rendering, which you could find out by putting a breakpoint in this method and seeing if any code on the stack holds a lock.

My use of dirty in BeginSecondaryCameraFrameConfig was cargo cult: it could just be for clean living - a signal for anyone who wants to watch. Or, it might actually be used within the render::Task system to decide wither call Config::configure. I don't know. If calling configure isn't ever expensive (e.g., no locks, etc.), then there's no reason to avoid it.  

If there _is_ a reason to avoid it, and if dirty isn't used internally to do so, then one could create a specific BeginSecondaryCameraFrameConfig.isDirty flag. (E.g., define specific setters that set this flag, and test/clear that flag in the conditional on line 56 here. But again, only if it's needed.

zfox23: Point taken. I didn't have enough of an understanding about what the "right thing" was when opening this PR, so I'll close it. Someone (or I) can revisit this optimization when the time is right.

PR #10755, opened 2017-06-20
Fixed an issue where toolbar buttons duplicated themselves when reloading scripts

seefo:
Should fix https://highfidelity.fogbugz.com/f/cases/5278/In-Desktop-mode-when-reloading-all-scripts-sometimes-get-duplicates-of-all-HUD-buttons

Test Plan:
1. Start interface in Desktop mode
2. Open the running scripts panel
3. Remove all scripts and then press the default scripts button
4. Spam the 'Reload Scripts' button several times, verify that the buttons do not duplicate themselves and only appear once

### DIFF
--- a/libraries/script-engine/src/ScriptEngines.cpp
+++ b/libraries/script-engine/src/ScriptEngines.cpp
@@ -352,6 +352,13 @@ QStringList ScriptEngines::getRunningScripts() {
 void ScriptEngines::stopAllScripts(bool restart) {
     QVector<QString> toReload;
     QReadLocker lock(&_scriptEnginesHashLock);
+
+    if (_isReloading) {
+        return;
+    } else {
+        _isReloading = true;
+    }
+
     for (QHash<QUrl, ScriptEngine*>::const_iterator it = _scriptEnginesHash.constBegin();
         it != _scriptEnginesHash.constEnd(); it++) {
         ScriptEngine *scriptEngine = it.value();
@@ -371,7 +378,7 @@ void ScriptEngines::stopAllScripts(bool restart) {
     }
     // wait for engines to stop (ie: providing time for .scriptEnding cleanup handlers to run) before
     // triggering reload of any Client scripts / Entity scripts
-    QTimer::singleShot(500, this, [=]() {
+    QTimer::singleShot(1000, this, [=]() {
         for(const auto &scriptName : toReload) {
             auto scriptEngine = getScriptEngine(scriptName);
             if (scriptEngine && !scriptEngine->isFinished()) {
@@ -386,6 +393,7 @@ void ScriptEngines::stopAllScripts(bool restart) {
             qCDebug(scriptengine) << "stopAllScripts -- emitting scriptsReloading";
             emit scriptsReloading();
         }
+        _isReloading = false;
     });
 }
 
--- a/scripts/system/snapshot.js
+++ b/scripts/system/snapshot.js
@@ -41,7 +41,16 @@ var numStillSnapshotUploadsPending = 0;
 
 // It's totally unnecessary to return to C++ to perform many of these requests, such as DELETEing an old story,
 // POSTING a new one, PUTTING a new audience, or GETTING story data. It's far more efficient to do all of that within JS
-var request = Script.require('request').request;
+var request;
+
+try {
+    // Due to an issue where if the user spams 'script reload', this call could cause an exception
+    // preventing our scriptEnding to not properly be initialized resulting in the tablet button
+    // duplicating itself where you end up with a bunch of SNAP buttons on your toolbar
+    request = Script.require('request').request;
+} catch(err) {
+    print('Failed to resolve request api, error: ' + err);
+}
 
 function openLoginWindow() {
     if ((HMD.active && Settings.getValue("hmdTabletBecomesToolbar", false))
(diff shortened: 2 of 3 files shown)


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/895/HighFidelity-Beta-PR10755-00a3fc77a113dfa05296a6f248b86b7815c1e7ef.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/895/HighFidelity-Beta-PR10755-00a3fc77a113dfa05296a6f248b86b7815c1e7ef.exe

CainFoool:
# Test Passed

Spam-clicked reload scripts after loading defaults, there were no duplicates in the toolbar.

libraries/script-engine/src/ScriptEngines.cpp
```diff
@@ -371,7 +378,7 @@ void ScriptEngines::stopAllScripts(bool restart) {
     }
     // wait for engines to stop (ie: providing time for .scriptEnding cleanup handlers to run) before
     // triggering reload of any Client scripts / Entity scripts
-    QTimer::singleShot(500, this, [=]() {
+    QTimer::singleShot(1000, this, [=]() {
```
zfox23: How did you determine this `1000`?

scripts/system/snapshot.js
```diff
@@ -41,7 +41,16 @@ var numStillSnapshotUploadsPending = 0;
 
 // It's totally unnecessary to return to C++ to perform many of these requests, such as DELETEing an old story,
 // POSTING a new one, PUTTING a new audience, or GETTING story data. It's far more efficient to do all of that within JS
-var request = Script.require('request').request;
+var request;
```
zfox23: What happened to the ` = Script.require('request').request;` ?

scripts/system/snapshot.js
```diff
@@ -41,7 +41,16 @@ var numStillSnapshotUploadsPending = 0;
...
+
+try {
```
zfox23: Does this block have to exist in all scripts using `Script.require()`?

scripts/system/snapshot.js
```diff
@@ -750,11 +759,10 @@ Script.scriptEnding.connect(function () {
     }
     if (tablet) {
         tablet.removeButton(button);
+        tablet.screenChanged.disconnect(onTabletScreenChanged);
```
zfox23: Is this code missing from any other `defaultScripts`?

seefo: I moved this up from below the conditional in-case `tablet `was undefined.

zfox23: ah okay, cool.

scripts/system/snapshot.js
seefo: For now, yes.  I wrote a comment [here ](https://github.com/highfidelity/hifi/pull/10755/files#diff-a296bc8f887a56903753737d51373764R47) explaining why, there is a FogBugz report for that [here](https://highfidelity.fogbugz.com/f/cases/5479/Script-require-request-should-fail-silently-if-Script-Engine-is-going-down).

zfox23: It seems like you should update the other defaultscripts then too, no?

libraries/script-engine/src/ScriptEngines.cpp
seefo: At 500ms, we see a lot of duplication of buttons.  I believe this is largely in part due to scripts being loaded and then before they finish initializing, they're told to shutdown -- leaving some things undefined, in some cases this can prevent a script from completing its scriptEnding (which is usually responsible for removing a scripts toolbar button).

I also tried 2500ms but felt too long, 1000ms seemed to be a stable and sufficient for scripts.

zfox23: Do you think there's a way to determine this value programmatically, or is it required for us to determine it empirically?

scripts/system/snapshot.js
seefo: Snapshot.js was the only script I saw that experienced the problem described below -- all the others don't seem to use `Script.require` or if they do, they use it in a way that is not affected by the script engine restarting.

libraries/script-engine/src/ScriptEngines.cpp
seefo: I definitely think that there are potential ways to calculate this time, but I believe waiting for all scripts to finish reloading is outside the scope of this fix and that empirically, 1000ms seemed to be a stable and sufficient time frame.  Additionally, any script that takes longer than 1000ms to shutdown is prematurely terminated, as can be seen [here](https://github.com/highfidelity/hifi/blob/master/libraries/script-engine/src/ScriptEngine.cpp#L378).

scripts/system/snapshot.js
zfox23: Are you positive that's true for `makeUserConnection.js`, `pal.js`, and `tablet-goto.js` ?

zfox23: @seefo, if you are not positive, please change those scripts.

If you are 100 positive, I will approve this PR.

seefo: I've looked into those three scripts and I think we're okay.  The only reason we are performing that try-catch here with `snapshot.js`, is because otherwise our tablet button would be left on our toolbar.  It looks like in `tablet-goto.js`, that the first thing that they try to do is setup `request`, preventing the tablet button from getting added if it fails.

I wonder if we should follow how `tablet-goto.js` does this, instead of this try-catch?  

Additionally, I realize I incorrectly answered your initial question about whether or not all scripts using `Script.require` will need the try-catch -- scripts will only need to perform that if full initialization is necessary to not leave the interface, avatar, or HUD in an incomplete state.  For example, `makeUserConnection.js` doesn't need it as, for example, it doesn't add a tablet button or initiate any processes in the beginning of its initialization that could be interrupted.

zfox23: I'm OK with the way you're doing things now. Thanks for the explanation.

PR #10759, opened 2017-06-20
Fix crash when cache entry lifetime exceeds cache lifetime

jherico:
Local testing of an unrelated bug revealed that when we shut down the application a crash can be caused by the `cache::File` being destroyed after the `cache::FileCache`.  This can happen because the `gpu::Texture` holds on to the cache entry and is only destroyed at the final cleanup in the `Application` destructor (technically any such objects should have been destroyed earlier, but even if they haven't been, the app should not crash on exit).  

## Testing

Repeatedly start and close interface while content is loading.  In the master build you will eventually crash during the exit (this only takes a few attempts for me).

In this PR build you should not crash on exit.

### DIFF
--- a/libraries/model-networking/src/model-networking/KTXCache.cpp
+++ b/libraries/model-networking/src/model-networking/KTXCache.cpp
@@ -24,9 +24,10 @@ const int KTXCache::INVALID_VERSION = 0x00;
 const char* KTXCache::SETTING_VERSION_NAME = "hifi.ktx.cache_version";
 
 KTXCache::KTXCache(const std::string& dir, const std::string& ext) :
-    FileCache(dir, ext) {
-    initialize();
+    FileCache(dir, ext) { }
 
+void KTXCache::initialize() {
+    FileCache::initialize();
     Setting::Handle<int> cacheVersionHandle(SETTING_VERSION_NAME, INVALID_VERSION);
     auto cacheVersion = cacheVersionHandle.get();
     if (cacheVersion != CURRENT_VERSION) {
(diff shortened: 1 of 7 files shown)


libraries/model-networking/src/model-networking/KTXCache.cpp
```diff
@@ -24,9 +24,10 @@ const int KTXCache::INVALID_VERSION = 0x00;
 const char* KTXCache::SETTING_VERSION_NAME = "hifi.ktx.cache_version";
 
 KTXCache::KTXCache(const std::string& dir, const std::string& ext) :
-    FileCache(dir, ext) {
-    initialize();
```
jherico: calling a virtual member from the ctor is a bad idea.  Clients will need to call initialize explicitly when using a FileCache object.

libraries/shared/src/shared/FileCache.cpp
```diff
@@ -300,18 +312,32 @@ void FileCache::clear() {
...
+    Lock lock(_mutex);
+    if (file->_locked) {
+        addUnusedFile(FilePointer(file, std::bind(&File::deleter, file)));
     } else {
-        deleteLater();
+        delete file;
+    }
+}
+
+void File::deleter(File* file) {
```
jherico: We have to use a static function here to avoid `delete this` in the case where the weak pointer to the parent cache is no longer valid.

tests/shared/src/FileCacheTests.cpp
```diff
@@ -83,26 +50,38 @@ void FileCacheTests::testUnusedFiles() {
     {
         for (int i = 0; i < 100; ++i) {
             std::string key = getFileKey(i);
-            auto file = cache->writeFile(TEST_DATA.data(), TestFileCache::Metadata(key, TEST_DATA.size()));
+            auto file = cache->writeFile(TEST_DATA.data(), FileCache::Metadata(key, TEST_DATA.size()));
+            QVERIFY(file->_locked);
             inUseFiles.push_back(file);
-            forceDeletes();
```
jherico: the deletes of files are no longer event driven, so this is no longer required.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/918/HighFidelity-Beta-PR10759-e3fb42eefd951dc9a920420c2b2cb3dd016450b2.dmg

CainFoool:
Testing when built.

jherico:
test this please

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/921/HighFidelity-Beta-PR10759-e3fb42eefd951dc9a920420c2b2cb3dd016450b2.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/921/HighFidelity-Beta-PR10759-e3fb42eefd951dc9a920420c2b2cb3dd016450b2.dmg

CainFoool:
# Test Passed
I started and exited out of Interface 15x times on this PR and was not able to replicate the crashing issue which was happening on Master following the same steps (after 5x times).

PR #10799, opened 2017-06-23
Fix warnings about creating QObjects with parents in different thread

jherico:
Qt will emit a warning if you create a QObject derived instance and pass a parent QObject that was created in another thread.  There are at least two places in the code we do this.  The first is in the application where we create a timer to checkin to the domain, passing the `nodeList` as a parent.  The second is in the GL backend where we create a new threadpool for texture transfers, passing the application as a parent.  

## Testing

Application behavior should be unchanged, but starting the application should no longer produce any log entries like this
`[06/23 10:08:46] [WARNING] [default] QObject: Cannot create children for a parent that is in a different thread.`

Test that application startup and shutdown proceed normally and that the application does not hang or crash on exit (any more than it currently does, i.e. any crashes reported as related to this should be reliably reproducible, not the sporadic crashes deep inside Qt that we're currently seeing).

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -713,9 +713,13 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer, bo
     updateHeartbeat();
 
     // setup a timer for domain-server check ins
-    QTimer* domainCheckInTimer = new QTimer(nodeList.data());
+    QTimer* domainCheckInTimer = new QTimer(this);
     connect(domainCheckInTimer, &QTimer::timeout, nodeList.data(), &NodeList::sendDomainServerCheckIn);
     domainCheckInTimer->start(DOMAIN_SERVER_CHECK_IN_MSECS);
+    connect(this, &QCoreApplication::aboutToQuit, [domainCheckInTimer] {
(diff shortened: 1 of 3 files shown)


CainFoool:
Testing when built

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1059/HighFidelity-Beta-PR10799-416852c0cd216a772558f4079efaba9dfd8b2142.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1059/HighFidelity-Beta-PR10799-416852c0cd216a772558f4079efaba9dfd8b2142.dmg

CainFoool:
# Test Passed

Test | Result
-------- | --------
Test that application startup and shutdown proceed normally and that the application does not hang or crash on exit (any more than it currently does, i.e. any crashes reported as related to this should be reliably reproducible, not the sporadic crashes deep inside Qt that we're currently seeing). | Pass

Appears to work as normal, not able to find the log specified in OP.

Let me know if you need anything else tested!

PR #10792, opened 2017-06-22
Trying to fix QtQml/QtScript crashes

jherico:
This PR attempts to resolve the increased frequency of bugsplat crashes in QtQml and QtScript.  I believe that the primary issue is race conditions between the scripting thread which might be inadvertently calling QML functions on the script thread, corrupting memory and resulting in these crashes deep inside the Qt DLLs.  

* Move `ToolbarScriptingInterface` from `interface` to `ui`
* Move `TabletScriptingInterface` from `script-engine` to `ui`
* Remove circular dependency between `ui` and `script-engine` (in fact both directions have been removed) 
* Remove dependency from `networking` to `QtWebEngine` and all `QtWebEngine` dependant types from `networking` to `ui` (fixes bug in building server only components on systems with no QtWebEngine library installed)
* Migrate `ToolbarScriptingInterface` and `TabletScriptingInterface` to to rely on the main thread
  * Forced all code that might be called from a script in these interfaces should use Qt method invocation to execute on the main thread.  All other code should verify it's running on the main thread (mostly by Q_ASSERT calls right now)
  * All proxy objects returned by these interfaces should be owned by the main thread.  Will emit warnings if not
  * Removed all mutexes from `TabletScriptingInterface` and related proxy classes in favor of forcing execution on the main thread
  * Similary, destructors for the proxy classes verify that they are running on the main thread, or log warnings
* Fix a bug in `RunningScripts.qml` that triggers a heap corruption warning Visual Studio

## Likely related fogbugs

[5535 CRASH in QtScript](https://highfidelity.fogbugz.com/f/cases/edit/5535/CRASH-in-QtScript)
[5518 CRASH in QtQml](https://highfidelity.fogbugz.com/f/cases/5518/CRASH-in-QtQml-maybe-in-Steam)

## Testing

* Verify functionality of scripts that manipulate toolbars / tablets.  
* Verify that the behavior is correct (well, unchanged anyway) for both tablet and toolbar modes when using the tablet interface.  
* Verify that the button behavior is correct when _switching_ between tablet and toolbar mode for a script that uses the tablet interface

[...]

PR #10827, opened 2017-06-27
Fix toolbar button activation states

jherico:
Fix for [FB5640](https://highfidelity.fogbugz.com/f/cases/5640).  This bug popped up during the refactor of the Tablet and Toolbar scripting interfaces in #10792. It turns out the scripts that were using the Tablet interface we relying on a very specific ordering of operations.  Specifically, when a button was clicked and triggered the `onClicked` method in a script, the script would then call something like `tablet.gotoMenuScreen();`.  The tablet then emits a `screenChanged` event in response to the tablet location being changed, and all these scripts had behavior triggered on this event.  

However, they all relied on the fact that the `onScreenChanged` handler would be called synchronously when you used a `tablet.gotoXXX` method, i.e. while you were still in the `onClicked` handler.  This producing a complicated order of operations where a button would get activated, flags would get cleared, and then the `onClicked` would re-set some of those flags to true.  

However, the refactor meant that the `onScreenChanged` handler is now called asynchronously, well after `onClicked` has returned.  As a result the logic in the scripts needs to change, so that rather than clearing flags and then resetting them further up the stack, we just use one flag and we explicitly set it in the `onScreenChanged` handler based on the inputs `type` and `url`.  

## Testing

Open Interface and click on any one of the buttons "Menu", "Audio", "People", "Help" or "Market".  

In the current dev build, clicking on a button will show the appropriate dialog, but the button state isn't set to active.  Additionally, clicking the button again may or may not close it.

In this PR, the corresponding button should light up, as well as properly un-light if you close the window manually.  Additionally, if you click the button while it's lit, it should also close the corresponding window.

### DIFF
--- a/scripts/system/audio.js
+++ b/scripts/system/audio.js
@@ -15,6 +15,7 @@
 
 var TABLET_BUTTON_NAME = "AUDIO";
 var HOME_BUTTON_TEXTURE = "http://hifi-content.s3.amazonaws.com/alan/dev/tablet-with-home-button.fbx/tablet-with-home-button.fbm/button-root.png";
+var AUDIO_QML_SOURCE = "../audio/Audio.qml";
 
 var MUTE_ICONS = {
     icon: "icons/tablet-icons/mic-mute-i.svg",
@@ -34,7 +35,6 @@ function onMuteToggled() {
     }
 }
 
-var shouldActivateButton = false;
 var onAudioScreen = false;
 
 function onClicked() {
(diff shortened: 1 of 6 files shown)


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1121/HighFidelity-Beta-PR10827-de88a34e977b1cd33913216a911ab813129e195d.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1121/HighFidelity-Beta-PR10827-de88a34e977b1cd33913216a911ab813129e195d.exe

CainFoool:
Testing shortly.

Nex-Pro:
Testing this

Nex-Pro:
Pass. Works in HMD and Desktop mode.

Noted an issue with the toolbar being absent after switching to HMD. Had to reload scripts. This is not related to this PR.

CainFoool:
# Test Passed

Working as expected..

PR #10837, opened 2017-06-28
Fix projection & lighting issues identified with 2nd rendering pipeline

samcake:
This pr fixes the projection&  lighting pass applied to the spectator Camera frame (secondary) when the main frame is in stereo mode (or HMD).

In HMD/stereo mode some algorithms would check on the args->_context->isStereo() function to select a stereo path, when it should select the mono path instead because rendering the secondary frame.
I replaced most of these tests to instead rely on a isStereo() function on the args which is representing the correct state at the time when the function is asked.

On top of this, the View Correction matrix is applied on all view transform tagged as "camera"
but here again we don;t  want it to be applied to the case of the secondary frame.
I  introduced a way to disable / restore the capability from the Batch allowing to control if the correction matrix is used or not and specifically disabled during the secondary camera.
This is mostly visible when using the HMD Simulator which uses a spectacular ViewCorrection for the sake of debug.

## TEST PLAN
Verify that the lighting bug & projection bug are solved in this pr compared to the spectator camera main branch.
Testing the HMD simulator (available in the choice of displays when the env variable HIFI_DEBUG_HMD is defined), verify that the rendered camera frame is correct.

### DIFF
--- a/libraries/gpu-gl/src/gpu/gl/GLBackendPipeline.cpp
+++ b/libraries/gpu-gl/src/gpu/gl/GLBackendPipeline.cpp
@@ -77,8 +77,14 @@ void GLBackend::do_setPipeline(const Batch& batch, size_t paramOffset) {
     if (_pipeline._invalidProgram) {
         glUseProgram(_pipeline._program);
         if (_pipeline._cameraCorrectionLocation != -1) {
-            auto cameraCorrectionBuffer = syncGPUObject(*_pipeline._cameraCorrectionBuffer._buffer);
+            gl::GLBuffer* cameraCorrectionBuffer = nullptr;
+            if (_transform._viewCorrectionEnabled) {
+                cameraCorrectionBuffer = syncGPUObject(*_pipeline._cameraCorrectionBuffer._buffer);
+            } else {
+                cameraCorrectionBuffer = syncGPUObject(*_pipeline._cameraCorrectionBufferIdentity._buffer);
+            }
             glBindBufferRange(GL_UNIFORM_BUFFER, _pipeline._cameraCorrectionLocation, cameraCorrectionBuffer->_id, 0, sizeof(CameraCorrection));
+
         }
         (void) CHECK_GL_ERROR();
         _pipeline._invalidProgram = false;
--- a/interface/src/SecondaryCamera.cpp
+++ b/interface/src/SecondaryCamera.cpp
@@ -79,6 +79,7 @@ public:
 
             gpu::doInBatch(args->_context, [&](gpu::Batch& batch) {
                 batch.disableContextStereo();
+                batch.disableContextViewCorrection();
             });
 
             auto srcViewFrustum = args->getViewFrustum();
@@ -112,6 +113,7 @@ public:
 
         gpu::doInBatch(args->_context, [&](gpu::Batch& batch) {
             batch.restoreContextStereo();
+            batch.restoreContextViewCorrection();
         });
     }
 };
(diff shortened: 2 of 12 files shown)


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1151/HighFidelity-Beta-PR10837-37e5454421c4f924fb035540a0ff9b949af49b16.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1151/HighFidelity-Beta-PR10837-37e5454421c4f924fb035540a0ff9b949af49b16.exe

libraries/gpu-gl/src/gpu/gl/GLBackendPipeline.cpp
```diff
@@ -77,8 +77,14 @@ void GLBackend::do_setPipeline(const Batch& batch, size_t paramOffset) {
...
         glUseProgram(_pipeline._program);
         if (_pipeline._cameraCorrectionLocation != -1) {
-            auto cameraCorrectionBuffer = syncGPUObject(*_pipeline._cameraCorrectionBuffer._buffer);
+            gl::GLBuffer* cameraCorrectionBuffer = nullptr;
+            if (_transform._viewCorrectionEnabled) {
+                cameraCorrectionBuffer = syncGPUObject(*_pipeline._cameraCorrectionBuffer._buffer);
+            } else {
+                cameraCorrectionBuffer = syncGPUObject(*_pipeline._cameraCorrectionBufferIdentity._buffer);
+            }
             glBindBufferRange(GL_UNIFORM_BUFFER, _pipeline._cameraCorrectionLocation, cameraCorrectionBuffer->_id, 0, sizeof(CameraCorrection));
```
jherico: wouldn't it be more efficient to have a single camera correction buffer with two CameraCorrection objects in it and just change the offset here?

howard-stearns:
worked great with spectator camera, but in hmd simulator I got this. Notice that their appears to be a brighter region with a corner in the lower left? I'm not familiar with the hmd simulator and do not know what to expect, but naively, this looks like a light-tiling artifact.
image

howard-stearns:
Question: is this a pr against spectator camera just to facilitate testing?  It turns out that everything from spectator camera _except_ the spectatorCamera.js script is in master, so I would think that if this PR passes, there's no reason not to migrate these changes to master as well, correct?

zfox23:
[lighting issue - vertical line about 2/3 across to the right](https://trello.com/c/WUeqovYH/187-lighting-issue-vertical-line-about-2-3-across-to-the-right)

libraries/gpu-gl/src/gpu/gl/GLBackendPipeline.cpp
samcake: just changing the offset is pretty much equivalent to that since you still need to call the same function, eventually the driver can be smart and detect the bo is the same just with a different offset. but i doubt it make any difference in that case.

samcake:
@howard-stearns the lighting issue you pointing out is due to the HMD simulator and expected.
IN that case, the "last minute" view correction is very extreme and look in regions not covered by the lighting data structure, hence the difference of appearance.

The lighting data structure is updated in the main rendering at the render rate and cannot be modified or updated at the same rate the last minute view correction can (from the hmd head pose sensor)
That correction is injected in the DIsplayPlugin at the display frame rate of 90Hz just before we execute the gpu::Batch of the frame in opengl.

samcake:
@howard-stearns Yes the pr changes are here so we can easily validate with spectator camera
i cherry picked the exact commit in a different PR against main and will put it up.
If you green the QA here, then we can merge the other PR confidently

PR #6859, opened 2016-01-19
Add shadows behind a render-menu flag

zzmp:
Adds shadows behind a menu-item (off by default). Also fixes some shadow-related issues:
- Shadows no longer hold references to old scene items (caused crash on location change)
- Keylight frustum now culls correctly
- Head is rendered, not cauterized, as a shadow
- A light along (0, -1, 0) will retain its orientation (orientation calculation edge-case)
- Extra render buffers are no longer used for debugging

PR #10847, opened 2017-06-28
faster EntityItem lookup by EntityItemID

AndrewMeadows:
* faster and cleaner lookup EntityItem by EntityItemID
* fixed "Developer --> Physics --> Highlight Simulation Ownership" menu option

### DIFF
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -188,7 +188,7 @@ bool EntityTree::updateEntity(EntityItemPointer entity, const EntityItemProperti
                 bool success;
                 AACube queryCube = entity->getQueryAACube(success);
                 if (!success) {
-                    qCDebug(entities) << "Warning -- failed to get query-cube for" << entity->getID();
+                    qCWarning(entities) << "failed to get query-cube for" << entity->getID();
                 }
                 UpdateEntityOperator theOperator(getThisPointer(), containingElement, entity, queryCube);
                 recurseTreeWithOperator(&theOperator);
@@ -1512,7 +1512,7 @@ void EntityTree::addEntityMapEntry(EntityItemPointer entity) {
     QWriteLocker locker(&_entityMapLock);
     EntityItemPointer otherEntity = _entityMap.value(id);
     if (otherEntity) {
-        qCDebug(entities) << "EntityTree::addEntityMapEntry() found pre-existing id " << id;
+        qCWarning(entities) << "EntityTree::addEntityMapEntry() found pre-existing id " << id;
         assert(false);
         return;
     }
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -91,11 +91,14 @@ void EntityTree::eraseAllOctreeElements(bool createNewRoot) {
         _simulation->clearEntities();
     }
     {
-        QWriteLocker locker(&_entityToElementLock);
-        foreach(EntityTreeElementPointer element, _entityToElementMap) {
-            element->cleanupEntities();
+        QWriteLocker locker(&_entityMapLock);
+        foreach(EntityItemPointer entity, _entityMap) {
+            EntityTreeElementPointer element = entity->getElement();
+            if (element) {
+                element->cleanupEntities();
+            }
         }
-        _entityToElementMap.clear();
+        _entityMap.clear();
     }
     Octree::eraseAllOctreeElements(createNewRoot);
 
@@ -136,29 +139,24 @@ void EntityTree::postAddEntity(EntityItemPointer entity) {
 }
 
 bool EntityTree::updateEntity(const EntityItemID& entityID, const EntityItemProperties& properties, const SharedNodePointer& senderNode) {
-    EntityTreeElementPointer containingElement = getContainingElement(entityID);
+    EntityItemPointer entity;
+    {
+        QReadLocker locker(&_entityMapLock);
+        entity = _entityMap.value(entityID);
+    }
+    if (!entity) {
+        return false;
+    }
+    return updateEntity(entity, properties, senderNode);
+}
+
+bool EntityTree::updateEntity(EntityItemPointer entity, const EntityItemProperties& origProperties,
+        const SharedNodePointer& senderNode) {
+    EntityTreeElementPointer containingElement = entity->getElement();
     if (!containingElement) {
         return false;
     }
 
-    EntityItemPointer existingEntity = containingElement->getEntityWithEntityItemID(entityID);
-    if (!existingEntity) {
-        return false;
-    }
-
-    return updateEntityWithElement(existingEntity, properties, containingElement, senderNode);
-}
-
-bool EntityTree::updateEntity(EntityItemPointer entity, const EntityItemProperties& properties, const SharedNodePointer& senderNode) {
-    EntityTreeElementPointer containingElement = getContainingElement(entity->getEntityItemID());
-    if (!containingElement) {
-        return false;
-    }
-    return updateEntityWithElement(entity, properties, containingElement, senderNode);
-}
-
-bool EntityTree::updateEntityWithElement(EntityItemPointer entity, const EntityItemProperties& origProperties,
-                                         EntityTreeElementPointer containingElement, const SharedNodePointer& senderNode) {
     EntityItemProperties properties = origProperties;
 
     bool allowLockChange;
@@ -331,8 +329,7 @@ bool EntityTree::updateEntityWithElement(EntityItemPointer entity, const EntityI
      }
 
     // TODO: this final containingElement check should eventually be removed (or wrapped in an #ifdef DEBUG).
-    containingElement = getContainingElement(entity->getEntityItemID());
-    if (!containingElement) {
+    if (!entity->getElement()) {
         qCDebug(entities) << "UNEXPECTED!!!! after updateEntity() we no longer have a containing element??? entityID="
                 << entity->getEntityItemID();
         return false;
@@ -1502,27 +1499,38 @@ int EntityTree::processEraseMessageDetails(const QByteArray& dataByteArray, cons
 }
 
 EntityTreeElementPointer EntityTree::getContainingElement(const EntityItemID& entityItemID)  /*const*/ {
-    QReadLocker locker(&_entityToElementLock);
-    EntityTreeElementPointer element = _entityToElementMap.value(entityItemID);
-    return element;
+    QReadLocker locker(&_entityMapLock);
+    EntityItemPointer entity = _entityMap.value(entityItemID);
+    if (entity) {
+        return entity->getElement();
+    }
+    return EntityTreeElementPointer(nullptr);
 }
 
-void EntityTree::setContainingElement(const EntityItemID& entityItemID, EntityTreeElementPointer element) {
-    QWriteLocker locker(&_entityToElementLock);
-    if (element) {
-        _entityToElementMap[entityItemID] = element;
-    } else {
-        _entityToElementMap.remove(entityItemID);
+void EntityTree::addEntityMapEntry(EntityItemPointer entity) {
+    EntityItemID id = entity->getEntityItemID();
+    QWriteLocker locker(&_entityMapLock);
+    EntityItemPointer otherEntity = _entityMap.value(id);
+    if (otherEntity) {
+        qCDebug(entities) << "EntityTree::addEntityMapEntry() found pre-existing id " << id;
+        assert(false);
+        return;
     }
+    _entityMap.insert(id, entity);
+}
+
+void EntityTree::clearEntityMapEntry(const EntityItemID& id) {
+    QWriteLocker locker(&_entityMapLock);
+    _entityMap.remove(id);
 }
 
 void EntityTree::debugDumpMap() {
     qCDebug(entities) << "EntityTree::debugDumpMap() --------------------------";
-    QReadLocker locker(&_entityToElementLock);
-    QHashIterator<EntityItemID, EntityTreeElementPointer> i(_entityToElementMap);
+    QReadLocker locker(&_entityMapLock);
+    QHashIterator<EntityItemID, EntityItemPointer> i(_entityMap);
     while (i.hasNext()) {
         i.next();
-        qCDebug(entities) << i.key() << ": " << i.value().get();
+        qCDebug(entities) << i.key() << ": " << i.value()->getElement().get();
     }
     qCDebug(entities) << "-----------------------------------------------------";
 }
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -1500,8 +1500,11 @@ int EntityTree::processEraseMessageDetails(const QByteArray& dataByteArray, cons
 }
 
 EntityTreeElementPointer EntityTree::getContainingElement(const EntityItemID& entityItemID)  /*const*/ {
-    QReadLocker locker(&_entityMapLock);
-    EntityItemPointer entity = _entityMap.value(entityItemID);
+    EntityItemPointer entity;
+    {
+        QReadLocker locker(&_entityMapLock);
+        entity = _entityMap.value(entityItemID);
+    }
     if (entity) {
         return entity->getElement();
     }
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -1526,9 +1526,11 @@ void EntityTree::clearEntityMapEntry(const EntityItemID& id) {
 }
 
 void EntityTree::debugDumpMap() {
+    // QHash's are implicitly shared, so we make a shared copy and use that instead.
+    // This way we might be able to avoid both a lock and a true copy.
+    QHash<EntityItemID, EntityItemPointer> localMap(_entityMap);
     qCDebug(entities) << "EntityTree::debugDumpMap() --------------------------";
-    QReadLocker locker(&_entityMapLock);
-    QHashIterator<EntityItemID, EntityItemPointer> i(_entityMap);
+    QHashIterator<EntityItemID, EntityItemPointer> i(localMap);
     while (i.hasNext()) {
         i.next();
         qCDebug(entities) << i.key() << ": " << i.value()->getElement().get();
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -331,7 +331,7 @@ bool EntityTree::updateEntity(EntityItemPointer entity, const EntityItemProperti
 
     // TODO: this final containingElement check should eventually be removed (or wrapped in an #ifdef DEBUG).
     if (!entity->getElement()) {
-        qCDebug(entities) << "UNEXPECTED!!!! after updateEntity() we no longer have a containing element??? entityID="
+        qCWarning(entities) << "EntityTree::updateEntity() we no longer have a containing element for entityID="
                 << entity->getEntityItemID();
         return false;
     }
@@ -364,7 +364,7 @@ EntityItemPointer EntityTree::addEntity(const EntityItemID& entityID, const Enti
     // You should not call this on existing entities that are already part of the tree! Call updateEntity()
     EntityTreeElementPointer containingElement = getContainingElement(entityID);
     if (containingElement) {
-        qCDebug(entities) << "UNEXPECTED!!! ----- don't call addEntity() on existing entity items. entityID=" << entityID
+        qCWarning(entities) << "EntityTree::addEntity() on existing entity item with entityID=" << entityID
                           << "containingElement=" << containingElement.get();
         return result;
     }
@@ -420,7 +420,7 @@ void EntityTree::deleteEntity(const EntityItemID& entityID, bool force, bool ign
     EntityTreeElementPointer containingElement = getContainingElement(entityID);
     if (!containingElement) {
         if (!ignoreWarnings) {
-            qCDebug(entities) << "UNEXPECTED!!!!  EntityTree::deleteEntity() entityID doesn't exist!!! entityID=" << entityID;
+            qCWarning(entities) << "EntityTree::deleteEntity() on non-existent entityID=" << entityID;
         }
         return;
     }
@@ -428,8 +428,7 @@ void EntityTree::deleteEntity(const EntityItemID& entityID, bool force, bool ign
     EntityItemPointer existingEntity = containingElement->getEntityWithEntityItemID(entityID);
     if (!existingEntity) {
         if (!ignoreWarnings) {
-            qCDebug(entities) << "UNEXPECTED!!!! don't call EntityTree::deleteEntity() on entity items that don't exist. "
-                        "entityID=" << entityID;
+            qCWarning(entities) << "EntityTree::deleteEntity() on non-existant entity item with entityID=" << entityID;
         }
         return;
     }
@@ -476,7 +475,7 @@ void EntityTree::deleteEntities(QSet<EntityItemID> entityIDs, bool force, bool i
         EntityTreeElementPointer containingElement = getContainingElement(entityID);
         if (!containingElement) {
             if (!ignoreWarnings) {
-                qCDebug(entities) << "UNEXPECTED!!!!  EntityTree::deleteEntities() entityID doesn't exist!!! entityID=" << entityID;
+                qCWarning(entities) << "EntityTree::deleteEntities() on non-existent entityID=" << entityID;
             }
             continue;
         }
@@ -484,8 +483,7 @@ void EntityTree::deleteEntities(QSet<EntityItemID> entityIDs, bool force, bool i
         EntityItemPointer existingEntity = containingElement->getEntityWithEntityItemID(entityID);
         if (!existingEntity) {
             if (!ignoreWarnings) {
-                qCDebug(entities) << "UNEXPECTED!!!! don't call EntityTree::deleteEntities() on entity items that don't exist. "
-                            "entityID=" << entityID;
+                qCWarning(entities) << "EntityTree::deleteEntities() on non-existent entity item with entityID=" << entityID;
             }
             continue;
         }
@@ -973,7 +971,7 @@ int EntityTree::processEditPacketData(ReceivedMessage& message, const unsigned c
                                      const SharedNodePointer& senderNode) {
 
     if (!getIsServer()) {
-        qCDebug(entities) << "UNEXPECTED!!! processEditPacketData() should only be called on a server tree.";
+        qCWarning(entities) << "EntityTree::processEditPacketData() should only be called on a server tree.";
         return 0;
     }
 
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -90,16 +90,17 @@ void EntityTree::eraseAllOctreeElements(bool createNewRoot) {
     if (_simulation) {
         _simulation->clearEntities();
     }
-    {
-        QWriteLocker locker(&_entityMapLock);
-        foreach(EntityItemPointer entity, _entityMap) {
+    QHash<EntityItemID, EntityItemPointer> localMap;
+    localMap.swap(_entityMap);
+    this->withWriteLock([&] {
+        foreach(EntityItemPointer entity, localMap) {
             EntityTreeElementPointer element = entity->getElement();
             if (element) {
                 element->cleanupEntities();
             }
         }
-        _entityMap.clear();
-    }
+    });
+    localMap.clear();
     Octree::eraseAllOctreeElements(createNewRoot);
 
     resetClientEditStats();
--- a/libraries/entities/src/EntityTree.h
+++ b/libraries/entities/src/EntityTree.h
@@ -119,9 +119,6 @@ public:
     // use this method if you only know the entityID
     bool updateEntity(const EntityItemID& entityID, const EntityItemProperties& properties, const SharedNodePointer& senderNode = SharedNodePointer(nullptr));
 
-    // use this method if you have a pointer to the entity (avoid an extra entity lookup)
-    bool updateEntity(EntityItemPointer entity, const EntityItemProperties& properties, const SharedNodePointer& senderNode = SharedNodePointer(nullptr));
-
     // check if the avatar is a child of this entity, If so set the avatar parentID to null
     void unhookChildAvatar(const EntityItemID entityID);
     void deleteEntity(const EntityItemID& entityID, bool force = false, bool ignoreWarnings = true);
@@ -183,7 +180,8 @@ public:
     int processEraseMessageDetails(const QByteArray& buffer, const SharedNodePointer& sourceNode);
 
     EntityTreeElementPointer getContainingElement(const EntityItemID& entityItemID)  /*const*/;
-    void setContainingElement(const EntityItemID& entityItemID, EntityTreeElementPointer element);
+    void addEntityMapEntry(EntityItemPointer entity);
+    void clearEntityMapEntry(const EntityItemID& id);
     void debugDumpMap();
     virtual void dumpTree() override;
     virtual void pruneTree() override;
@@ -275,9 +273,8 @@ signals:
 protected:
 
     void processRemovedEntities(const DeleteEntityOperator& theOperator);
-    bool updateEntityWithElement(EntityItemPointer entity, const EntityItemProperties& properties,
-                                 EntityTreeElementPointer containingElement,
-                                 const SharedNodePointer& senderNode = SharedNodePointer(nullptr));
+    bool updateEntity(EntityItemPointer entity, const EntityItemProperties& properties,
+            const SharedNodePointer& senderNode = SharedNodePointer(nullptr));
     static bool findNearPointOperation(const OctreeElementPointer& element, void* extraData);
     static bool findInSphereOperation(const OctreeElementPointer& element, void* extraData);
     static bool findInCubeOperation(const OctreeElementPointer& element, void* extraData);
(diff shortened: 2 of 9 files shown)


libraries/entities/src/EntityTree.h
```diff
@@ -309,8 +306,8 @@ public slots:
         _deletedEntityItemIDs << id;
     }
 
-    mutable QReadWriteLock _entityToElementLock;
-    QHash<EntityItemID, EntityTreeElementPointer> _entityToElementMap;
+    mutable QReadWriteLock _entityMapLock;
+    QHash<EntityItemID, EntityItemPointer> _entityMap;
```
AndrewMeadows: EntityTree now maintains a map from EntityItemID directly to EntityItemPointer, instead of a map from EntityItemID to EntityTreeElement.  The size of the map is unchanged but the lookup avoids extra work.

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -1001,7 +1001,6 @@ int EntityTreeElement::readElementDataFromBuffer(const unsigned char* data, int
                             if (currentContainingElement.get() != this) {
                                 currentContainingElement->removeEntityItem(entityItem);
                                 addEntityItem(entityItem);
```
AndrewMeadows: `EntityTreeElement::addEntityItem()` and `removeEntityItem()` already maintain the `EntityItem::_element` backpointer to the element.

libraries/entities/src/EntityTree.cpp
```diff
@@ -136,29 +139,24 @@ void EntityTree::postAddEntity(EntityItemPointer entity) {
...
 
 bool EntityTree::updateEntity(const EntityItemID& entityID, const EntityItemProperties& properties, const SharedNodePointer& senderNode) {
-    EntityTreeElementPointer containingElement = getContainingElement(entityID);
-    if (!containingElement) {
-        return false;
+    EntityItemPointer entity;
+    {
+        QReadLocker locker(&_entityMapLock);
+        entity = _entityMap.value(entityID);
     }
```
AndrewMeadows: The explicit context here minimizes the lock context on `_entityMapLock`.  That is, it avoids keeping the lock during the potentially more expensive `updateEntity()` call below.

AndrewMeadows:
Test Plan

There is no functionality change here.  This PR is pure optimization and cleanup.  That said, the map that this PR introduces is used to lookup the entity by ID whenever update packets hare handled, so if a bug were introduced then it would probably manifest as entity update packets being ignored, and entity updates being broken, either at the server or client.

(0) Run a domain and assignment-clients using the code in this PR.  This PR affects the entity-server so you should be running an entity-server on this PR.

(1) Login with two users: UserA and UserB.

(2) UserA: create and modify several objects while UserB watches.
--> All the changes that UserA makes should be visible to both UserA and UserB.

(3) UserB: create and modify several objects while UserA watches.
--> All the changes that UserB makes should be visible to both UserA and UserB.

(4) Shutdown the domain and load it back up again. Login and verify that the objects load correctly after stop/start cycle.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1181/HighFidelity-Beta-PR10847-5f1fdaf44511ec575282e87e5c47aeee8a270e9c.exe

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1188/HighFidelity-Beta-PR10847-c0581890ca739f69f57b0b2b47ffd57bff133fca.exe

AndrewMeadows:
Jenkins, rebuild this please.

AndrewMeadows:
Jenkins, rebuild this please.

CainFoool:
@AndrewMeadows: "test this please" I believe is what you gotta use

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1197/HighFidelity-Beta-PR10847-c0581890ca739f69f57b0b2b47ffd57bff133fca.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1197/HighFidelity-Beta-PR10847-c0581890ca739f69f57b0b2b47ffd57bff133fca.exe

AndrewMeadows:
No @CainFoool, Jenkins searches for two patterns: "test this please" or "build this please".  We added the second string because the first would occasionally confuse testers.  I try to use the less ambiguous pattern.

libraries/entities/src/EntityTree.cpp
```diff
@@ -1502,27 +1499,38 @@ int EntityTree::processEraseMessageDetails(const QByteArray& dataByteArray, cons
...
-        _entityToElementMap[entityItemID] = element;
-    } else {
-        _entityToElementMap.remove(entityItemID);
+void EntityTree::addEntityMapEntry(EntityItemPointer entity) {
+    EntityItemID id = entity->getEntityItemID();
+    QWriteLocker locker(&_entityMapLock);
+    EntityItemPointer otherEntity = _entityMap.value(id);
+    if (otherEntity) {
+        qCDebug(entities) << "EntityTree::addEntityMapEntry() found pre-existing id " << id;
+        assert(false);
```
sethalves: is there a qCCritical ?

CainFoool:
Testing.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1199/HighFidelity-Beta-PR10847-f4e48857cdac740c2512b22b1f2dcb9dbad4ebbf.exe

libraries/entities/src/EntityTree.cpp
AndrewMeadows: I suppose this should be `qWarning`.  I'll change it.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1199/HighFidelity-Beta-PR10847-f4e48857cdac740c2512b22b1f2dcb9dbad4ebbf.dmg

CainFoool:
# Test Passed

Test | Result
-------- | --------
(0) Run a domain and assignment-clients using the code in this PR. This PR affects the entity-server so you should be running an entity-server on this PR. | Pass
(1) Login with two users: UserA and UserB. | Pass
(2) UserA: create and modify several objects while UserB watches. | Pass
--> All the changes that UserA makes should be visible to both UserA and UserB | Pass
(3) UserB: create and modify several objects while UserA watches. | Pass
--> All the changes that UserB makes should be visible to both UserA and UserB. | Pass
(4) Shutdown the domain and load it back up again. Login and verify that the objects load correctly after stop/start cycle. | Pass
(5) Verify that Developer --> Physics --> Highlight Simulation Ownership menu option works again. | Pass

Let me know if this further needs tested. Everything working all good on my side!

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1204/HighFidelity-Beta-PR10847-1e4e1526dd2fd7fc919a25f3c93d61587706733c.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1204/HighFidelity-Beta-PR10847-1e4e1526dd2fd7fc919a25f3c93d61587706733c.dmg

libraries/entities/src/EntityTree.cpp
```diff
@@ -331,8 +329,7 @@ bool EntityTree::updateEntityWithElement(EntityItemPointer entity, const EntityI
      }
 
     // TODO: this final containingElement check should eventually be removed (or wrapped in an #ifdef DEBUG).
-    containingElement = getContainingElement(entity->getEntityItemID());
-    if (!containingElement) {
+    if (!entity->getElement()) {
         qCDebug(entities) << "UNEXPECTED!!!! after updateEntity() we no longer have a containing element??? entityID="
```
jherico: should also be changed to a warning

libraries/entities/src/EntityTree.cpp
```diff
@@ -1502,27 +1499,38 @@ int EntityTree::processEraseMessageDetails(const QByteArray& dataByteArray, cons
 }
 
 EntityTreeElementPointer EntityTree::getContainingElement(const EntityItemID& entityItemID)  /*const*/ {
-    QReadLocker locker(&_entityToElementLock);
-    EntityTreeElementPointer element = _entityToElementMap.value(entityItemID);
-    return element;
+    QReadLocker locker(&_entityMapLock);
+    EntityItemPointer entity = _entityMap.value(entityItemID);
+    if (entity) {
```
jherico: is there any reason not to scope the lock to just the `_entityMap` access here?

libraries/entities/src/EntityTree.cpp
```diff
@@ -1502,27 +1499,38 @@ int EntityTree::processEraseMessageDetails(const QByteArray& dataByteArray, cons
...
+void EntityTree::clearEntityMapEntry(const EntityItemID& id) {
+    QWriteLocker locker(&_entityMapLock);
+    _entityMap.remove(id);
 }
 
 void EntityTree::debugDumpMap() {
     qCDebug(entities) << "EntityTree::debugDumpMap() --------------------------";
-    QReadLocker locker(&_entityToElementLock);
-    QHashIterator<EntityItemID, EntityTreeElementPointer> i(_entityToElementMap);
+    QReadLocker locker(&_entityMapLock);
```
jherico: you could create a local copy of the `_entityMap` here and then iterate over it instead of keeping the lock during iteration / serialization

libraries/entities/src/EntityTree.cpp
```diff
@@ -91,11 +91,14 @@ void EntityTree::eraseAllOctreeElements(bool createNewRoot) {
         _simulation->clearEntities();
     }
     {
-        QWriteLocker locker(&_entityToElementLock);
-        foreach(EntityTreeElementPointer element, _entityToElementMap) {
-            element->cleanupEntities();
+        QWriteLocker locker(&_entityMapLock);
```
jherico: I would suggest minimizing lock time here with 
```
QHash<EntityItemID, EntityItemPointer> localEntityMap; 
localEntityMap.swap(_entityMap);
```
and then iteratoring over the local map without the lock... however I'm not sure if the lock is intended to protect access both to the map and to the `EntityTreeElementPointer::cleanupEntities` call, so not sure.   There are other places in the code where there is no such protection though

libraries/entities/src/EntityTree.cpp
```diff
@@ -136,29 +139,24 @@ void EntityTree::postAddEntity(EntityItemPointer entity) {
...
-
-    return updateEntityWithElement(existingEntity, properties, containingElement, senderNode);
+    return updateEntity(entity, properties, senderNode);
 }
 
-bool EntityTree::updateEntity(EntityItemPointer entity, const EntityItemProperties& properties, const SharedNodePointer& senderNode) {
-    EntityTreeElementPointer containingElement = getContainingElement(entity->getEntityItemID());
+bool EntityTree::updateEntity(EntityItemPointer entity, const EntityItemProperties& origProperties,
+        const SharedNodePointer& senderNode) {
+    EntityTreeElementPointer containingElement = entity->getElement();
```
jherico: in theory if thread 1 calls `eraseAllOctreeElements` and thread 2 calls `updateEntity`, there could be a race condition between this call and the call to `element->cleanupEntities`.  Not sure if that's a problem or not.

ZappoMan:
The claim is that this is an optimization... shouldn't there be a test plan that demonstrates the performance boost? Should we add a unit test for EntityTree that shows the expected behavior still works?

libraries/entities/src/EntityTree.cpp
AndrewMeadows: This method isn't actually called anywhere, and if it were it would be in a debug context.  In my view it is cruft, but I figured it would be easier to maintain it rather than purge.

Domains of today maybe have 10k entities but as things get more interesting the container will be larger.  In most contexts I would be loathe to make such a copy but in this one I could make an exception.

libraries/entities/src/EntityTree.cpp
AndrewMeadows: Good point.  The code is actually modifying the tree elements which should be locked.  After auditing the code I find that all existing calls to `eraseAllOctreeElements()` are properly locked except one in the `Application` dtor.  In any case, the problem code should be locked to make it more future-proof.  It looks like the right way to do this is via lambda inside a `withWriteLock()` context.  Since we're throwing away the `_entityMap` here the `swap` strategy works well.

libraries/entities/src/EntityTree.cpp
AndrewMeadows: I will remove the `_entityMap` lock in `eraseAllOctreeElements()`.

libraries/entities/src/EntityTree.cpp
AndrewMeadows: I thought about it, but I knew the subsequent code was relatively inexpensive.  I suppose I should use the same pattern I used for the expensive stuff for consistency.

libraries/entities/src/EntityTree.cpp
AndrewMeadows: `updateEntityWithElement()` was removed because we no longer need to lookup the element before getting to the entity itself.  Instead we go directly to the entity and then get the element from it.  The foundation for this has been in place for a couple years:  the `EntityItem` has a backpointer to its containing element.  This is maintained by the `EntityOctreeElement::addEntity()` and peer methods.  I audited the code and as far as I can tell it is water tight.

The performance gain here isn't huge, and I don't have concrete data about a performance problem, but the new solution is faster by inspection and simpler to read.  We're trading "at least one function call and a vector scan" for "a data member lookup" (both methods still require the map lookup).

Part of the motivation for this PR is to increase code quality: http://www.osnews.com/images/comics/wtfm.jpg

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1234/HighFidelity-Beta-PR10847-d5018cf88c0c97d7bf50e9cca604f92d800e2ee9.dmg

AndrewMeadows:
The new code is exactly that: faster by inspection.  We have a code path that is given an `EntityItemID` and needs to get to the `EntityItem` (to update it, for example, potentially also updating the element where it lives).  The old code would first go to the element (via a map lookup) and then would walk its list of entities (doing `EntityItemID` comparisons along the way) until it found the thing it was looking for.  The new code does a map lookup, and if it needs the element it gets it from the `EntityItem`.

AndrewMeadows:
The Windows build failed for some mysterious reason.  Jenkins, rebuild this please.

cozza13:
test this please

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1238/HighFidelity-Beta-PR10847-d5018cf88c0c97d7bf50e9cca604f92d800e2ee9.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1238/HighFidelity-Beta-PR10847-d5018cf88c0c97d7bf50e9cca604f92d800e2ee9.dmg

jherico:
I'm pretty certain that this code is not thread safe.  In fact I'm pretty certain the current code isn't thread safe either.  Consider the following code in `EntityItem.cpp`

    if (_element && _element->getTree()) {
        _element->getTree()->trackIncomingEntityLastEdited(lastEditedFromBufferAdjusted, bytesRead);
    }

and this code in `EntityTreeElement.cpp`

    withWriteLock([&] {
        numEntries = _entityItems.removeAll(entity);
    });
    if (numEntries > 0) {
        assert(entity->_element.get() == this);
        entity->_element = NULL;
        return true;
    }

No lock protects theses access to `EntityItem::_element`  In theory, if one thread is calling the first block, while another thread is calling the second block, `_element` can become null in the first block between the initial test in the conditional and the call to `_element->getTree()->...`

That's the current code base and we might be surviving _only_ by some sort of implied locking, or by order of operations, or just by luck.  This PR could exacerbate the situation by removing some of the implicit locking or changing the order of operations.  That terrifies me.  I hate to be a downer, but I don't feel comfortable with changes to locking in code that interacts with the `_element` member in `EntityItem` until I'm positive this won't trigger more mystery crashes.

ZappoMan:
I agree with Austin on this... but I think to his point the implicit locking is on the tree... basically all the threads that touch entities should be doing read/write locking of the entire tree... which means they can't change the tree, the elements, or the entities without a write lock on all of them. That *presumably* is what protects us here.

That was how it was originally *intended to be* written back in the day.

AndrewMeadows:
Only `EntityTreeElement` ever changes the value of `EntityItem::_element`, so the code inside `EntityTreeElement` (which is changing the value) is thread-safe.  Meanwhile, I changed the code inside `EntityItem` to make a local copy of the smart pointer and to use that for subsequent logic, which should solve the thread-safety problem there.  Most of the time the `EntityItem::_element` is being used to get to the `EntityTree`, which is safe to multi-check because the backpointer to the tree is only ever set to a valid tree and is then never cleared.

libraries/entities/src/EntityTreeElement.cpp
```diff
@@ -918,6 +919,7 @@ bool EntityTreeElement::removeEntityItem(EntityItemPointer entity) {
         numEntries = _entityItems.removeAll(entity);
     });
     if (numEntries > 0) {
+        // NOTE: only EntityTreeElement should ever be changing the value of entity->_element
         assert(entity->_element.get() == this);
         entity->_element = NULL;
```
jherico: `removeEntityItem` and `addEntityItem` still access the `EntityItem::_element` without locks.  There are still potential race conditions if an `EntityItemPointer` as `removeEntityItem` called on one thread and `addEntityItem` called concurrently on another.  Even if we move these into the write locks, it doesn't necessarily help if you're working with two different `EntityTreeElement` instances.  What happens if an entity is moving between elements on one thread while being deleted in another thread?

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1288/HighFidelity-Beta-PR10847-3a2721219a098439ee5875e15d240875bc443ce7.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1288/HighFidelity-Beta-PR10847-3a2721219a098439ee5875e15d240875bc443ce7.exe

AndrewMeadows:
Entities are only Added, Moved, or Deleted by instances of `AddEntityOperator`, `UpdateEntityOperator`, or `DeleteEntityOperator`.  Each of these operators lock the tree before they call `EntityTreeElement::addEntity()` or `EntityTreeElement::removeEntity()`.

AndrewMeadows:
I believe these changes might fix a problem that Philip "rediscovered" on July 3rd when he wrote a script that would create many boxes in regular grid out several hundred meters.  He had boxes lined up, centered on the **xy**, **yz**, and **zx** planes... hence the root octree element had a very large number of entities in its list.  Since each EntityID-to-EntityItemPointer lookup currently requires a linear search through the entities in the element, and a large number of entities were in the root element, he was noticing terrible performance when looking around his test region with HMD.  As I recall: he worked around this problem by shifting the planes of his test objects off of the cardinal planes of the world.

AndrewMeadows:
Ok, I take that back...

None of the EntityOperators lock the tree themselves.  In some cases the tree is obviously locked in a higher context that creates and submits an operator for recursion.  In other cases... dunno.  I tried to poke around some of them to prove to myself that the tree is locked but I failed: either the tree is not locked for those or I got lost trying to figure it out.

I guess I'll hack some debug code into place so I can query whether the tree is properly locked or not in a particular context.

ZappoMan:
Regarding the possibility that Philips large block test was slow because of this issue: it is possible - but I don't think the entity lookup code was happening in that case. Namely in streaming entities from the server, there shouldn't be any need to lookup by ID. 

That being said, I agree this lookup of entities by ID would be negatively effected by large entity counts like this. Let's make a unit test that times this case, as well as tests that validity of lookups. We can use this into test to proof this performance boost and also to prevent regression in future changes.

AndrewMeadows:
I wrote a script to try to measure speed improvements between this PR and hifi/master (see script below).  The results are for RELEASE builds on 2013 Mac laptop (16GB ram, i7) running linux (Xubuntu-16.10).  The script would create several entities such than they all lived in the root tree element, and would then try to repeatedly update a property on the first and last entities.  Each edit attempt would require an entity lookup by entityID.

```
------------------------PR-10847------------------------

numEntities =  12  numEdits = 10000  firstDt = 2918  lastDt = 2883 msec
numEntities =  12  numEdits = 10000  firstDt = 2821  lastDt = 2876
numEntities =  12  numEdits = 10000  firstDt = 2912  lastDt = 2888
numEntities =  12  numEdits = 10000  firstDt = 2875  lastDt = 2927

numEntities = 300  numEdits = 10000  firstDt = 3391  lastDt = 3906
numEntities = 300  numEdits = 10000  firstDt = 3389  lastDt = 3799
numEntities = 300  numEdits = 10000  firstDt = 3189  lastDt = 3808
numEntities = 300  numEdits = 10000  firstDt = 3342  lastDt = 3867

numEntities = 1875  numEdits = 10000  firstDt = 4354  lastDt = 7594
numEntities = 1875  numEdits = 10000  firstDt = 4031  lastDt = 7380
numEntities = 1875  numEdits = 10000  firstDt = 3920  lastDt = 7530
numEntities = 1875  numEdits = 10000  firstDt = 3792  lastDt = 7421

------------------------hifi/master------------------------

numEntities =  12  numEdits = 10000  firstDt = 3017  lastDt = 3078
numEntities =  12  numEdits = 10000  firstDt = 3017  lastDt = 3038
numEntities =  12  numEdits = 10000  firstDt = 2917  lastDt = 3001
numEntities =  12  numEdits = 10000  firstDt = 3031  lastDt = 2992

numEntities = 300  numEdits = 10000  firstDt = 3352  lastDt = 3968
numEntities = 300  numEdits = 10000  firstDt = 3338  lastDt = 3932
numEntities = 300  numEdits = 10000  firstDt = 3220  lastDt = 3985
numEntities = 300  numEdits = 10000  firstDt = 3377  lastDt = 3991

numEntities = 1875  numEdits = 10000  firstDt = 4074  lastDt = 9303
numEntities = 1875  numEdits = 10000  firstDt = 3972  lastDt = 9254
numEntities = 1875  numEdits = 10000  firstDt = 3874  lastDt = 9349
numEntities = 1875  numEdits = 10000  firstDt = 3781  lastDt = 9530
```
A large part of the cost to edit an entity is constant and involves translation of data from JS to C++ however the lookup can get expensive for crowded tree elements.  As expected this PR's lookup scales better (O(log(N))) than hifi/master (O(N)), and removes a small portion of the constant cost (probably mostly through function call overhead).
```
// Creates a large number of entities on the cardinal planes of the octree (all
// objects will live in the root octree element).  Measures how long it takes
// to update the properties of the first and last entity.  The difference
// between the two measurements shows how the cost of lookup changes as a
// function of the number of entities.  For best results run this in an
// otherwise empty domain.

var firstId;
var lastId;
var NUM_ENTITIES_ON_SIDE = 25;

// create the objects
createObjects = function () {
    var STRIDE = 0.75;
    var WIDTH = 0.5;
    var DIMENSIONS = { x: WIDTH, y: WIDTH, z: WIDTH };
    var LIFETIME = 20;

    var properties = {
        name: "",
        type : "Box",
        dimensions : DIMENSIONS,
        position : { x: 0, y: 0, z: 0},
        lifetime : LIFETIME,
        color : { red:255, green: 64, blue: 255 }
    };

    // xy
    var planeName = "xy";
    for (var i = 0; i < NUM_ENTITIES_ON_SIDE; ++i) {
        for (var j = 0; j < NUM_ENTITIES_ON_SIDE; ++j) {
            properties.name = "Box-" + planeName + "-" + i + "." + j;
            properties.position = { x: i * STRIDE, y: j * STRIDE, z: 0 };
            var red = i * 255 / NUM_ENTITIES_ON_SIDE;
            var green = j * 255 / NUM_ENTITIES_ON_SIDE;
            var blue = 0;
            properties.color = { red: red, green: green, blue: blue };
            if (i == 0 && j == 0) {
                firstId = Entities.addEntity(properties);
            } else {
                Entities.addEntity(properties);
            }
        }
    }

    // yz
    var planeName = "yz";
    for (var i = 0; i < NUM_ENTITIES_ON_SIDE; ++i) {
        for (var j = 0; j < NUM_ENTITIES_ON_SIDE; ++j) {
            properties.name = "Box-" + planeName + "-" + i + "." + j;
            properties.position = { x: 0, y: i * STRIDE, z: j * STRIDE };
            var red = 0;
            var green = i * 255 / NUM_ENTITIES_ON_SIDE;
            var blue = j * 255 / NUM_ENTITIES_ON_SIDE;
            properties.color = { red: red, green: green, blue: blue };
            Entities.addEntity(properties);
        }
    }

    // zx
    var planeName = "zx";
    for (var i = 0; i < NUM_ENTITIES_ON_SIDE; ++i) {
        for (var j = 0; j < NUM_ENTITIES_ON_SIDE; ++j) {
            properties.name = "Box-" + planeName + "-" + i + "." + j;
            properties.position = { x: j * STRIDE, y: 0, z: i * STRIDE };
            var red = j * 255 / NUM_ENTITIES_ON_SIDE;
            var green = 0;
            var blue = i * 255 / NUM_ENTITIES_ON_SIDE;
            properties.color = { red: red, green: green, blue: blue };
            lastId = Entities.addEntity(properties);
        }
    }
};

createObjects();

// measure the time it takes to edit the first and last entities many times
// (requires a lookup by entityId each time)
changeProperties = function (id) {
    var newProperties = { color : { red: 255, green: 255, blue: 255 } };
    Entities.editEntity(id, newProperties);
}

// first
var NUM_CHANGES = 10000;
var firstStart = Date.now();
for (var k = 0; k < NUM_CHANGES; ++k) {
    changeProperties(firstId);
}
var firstEnd = Date.now();
var firstDt = firstEnd - firstStart;

// last
var lastStart = Date.now();
for (var k = 0; k < NUM_CHANGES; ++k) {
    changeProperties(lastId);
}
var lastEnd = Date.now();
var lastDt = lastEnd - lastStart;

// print the results
var numEntities = 3 * NUM_ENTITIES_ON_SIDE * NUM_ENTITIES_ON_SIDE;
print("numEntities = " + numEntities + "  numEdits = " + NUM_CHANGES + "  firstDt = " + firstDt + "  lastDt = " + lastDt);
```

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1449/HighFidelity-Beta-PR10847-3450ceacaa0ad02ee6bb762564c385d510880059.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1449/HighFidelity-Beta-PR10847-3450ceacaa0ad02ee6bb762564c385d510880059.exe

libraries/entities/src/EntityTreeElement.cpp
ZappoMan: These are protected with a read/write lock higher up in the EntityTree... which is the public API for changing any entity.

Nex-Pro:
Testing this

Nex-Pro:
### Possible FAIL

(0) Run a domain and assignment-clients using the code in this PR. This PR affects the entity-server so you should be running an entity-server on this PR. **DONE**

(1) Login with two users: UserA and UserB. **DONE**

(2) UserA: create and modify several objects while UserB watches. **DONE**
--> All the changes that UserA makes should be visible to both UserA and UserB.**SUCCESS**

(3) UserB: create and modify several objects while UserA watches. **DONE**
--> All the changes that UserB makes should be visible to both UserA and UserB.**SUCCESS**

(4) Shutdown the domain and load it back up again. Login and verify that the objects load correctly after stop/start cycle.**SUCCESS**

(5) Verify that Developer --> Physics --> Highlight Simulation Ownership menu option works again.**SUCCESS**

phys_ownership



In addition I use an entity server script I had previously created to edit the color property of a sphere over time and both clients saw the changes simultaneously.

**ISSUE FOUND:** 

On interface "A" I set an elongated sphere to be dynamic and moved it with the cursor. 
It moved smoothly as expected with no rotation.
On interface "B" it was observed to be spinning.

On interface "B" I set the elongated sphere to be dynamic and moved it with the cursor. 
It moved smoothly as expected with no rotation.
On interface "A" it was observed to be spinning.

This does not appear to be happening in #6859

This repro'd even after restarting the server.

Nex-Pro:
@AndrewMeadows Any ideas why that rotation is showing up for third parties ?

AndrewMeadows:
No @Nex-Pro, that sounds bad so I'll try to reproduce it myself.  This branch is now a bit out of date and could use a rebase.  I'll first try to reproduce and irregardless of the results (good or bad) I'll rebase and try to reproduce again.

AndrewMeadows:
@nex-pro: I wasn't able to reproduce the bug:

(1) UserA moves asymmetrical dynamic "sphere" entity via the cursor (which uses he default "far-grab" script) and UserB who is watching sees the entity rotating while UserA does not.

I tried in the old version and also after rebase on top of the latest hifi/master.

We need someone in QA to try to test again and try to reproduce.  If it can't be reproduced then I think we should be able to go ahead and merge this.  I don't know of anything changed in this project that would possibly introduce such a discrepancy between local physics.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1686/HighFidelity-Beta-PR10847-de2139106245304cf0703d12df63c8b91d89ec61.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1686/HighFidelity-Beta-PR10847-de2139106245304cf0703d12df63c8b91d89ec61.dmg

Nex-Pro:
Retesting

Nex-Pro:
@AndrewMeadows I just tested for that odd rotation behaviour with the latest build of this pr. I get the same problem. Try running two clients on the same machine as the domain. It doesn't happen every time. But it is definitely happening.

AndrewMeadows:
@Nex-Pro, how does the grabbed object spin?  About the vertical axis?  About the axis of grab?  Randomly?  Does it, perhaps, look like it is bumping against another object (which may be very small and hard to see)?  Are UserB and UserA (observer and owner, respectively) looking at each other from across a distance?  Or are they standing next to each other both looking toward the object?  Is the grabbed object closer to UserA or to UserB?

The reason I ask all of those questions is: again I wasn't able to reproduce, but I have at least one theory about how it might happen.  If UserB (the observer) had an object in its simulation that was not known to UserA... for example, if it were a very small box near where the test object were moving bug small enough, and UserA were far enough away to have it culled from its view by the entity-server... then it is possible that UserB would see the grabbed object colliding with this tiny object, and would simulate it accordingly.  Meanwhile the grabber UserA might not know about the small object and therfore show the sphere not rotating.  If this were the case then UserB would tend to see the object spinning but also popping back to the rotation that UserA sees it... whenever UserA happened to send a recent update -- the spinning object would tend to spin away from UserA's rotation for it and then suddenly snap back to what UserA sees.

How to test this theory?  Here are some ideas:

(1) Make sure the field is very clear of obstacles.  In other words, test in a domain that is basically empty of all other objects except for a floor.  Does it still happen?

(2) Make UserA and UserB stand next to each other facing the same direction such that both are likely to have the same objects culled/visible in their view for LOD.  Does it still happen?

(3) Make the grabbed object small such that it is less likely to hit something nearly invisible.  Does it still happen?

(4) Make the grabbed object collisionless.  Does it still happen?

(5) Set the `angularDamping` of the grabbed object to 1.0.  Does it still happen to the same degree?

If it still reproduces in any of these conditions then send me a copy of your models.json.gz and specify exactly where UserA and UserB are to stand in order to reproduce it.

Nex-Pro:
@AndrewMeadows 

Ok can now rule out this PR as the cause. I got a repro in latest dev build. 

Was using the default content and on the builders platform. 
I first checked for any hidden entities and found nothing with a bounding box that encroached on the location apart from the builders platform.
Disabled physics for the build platform and the issue stopped immediately.
Grabbed a copy of the build platform and checked the model for stray geometry. There is no stray geometry in the model. It consists of two cuboid sub-meshes. The larger constructed of six separate quads each made of two triangles. The smaller consists of six joined quads each of two triangles.
There are exactly 32 vertices in the model. I think we can rule out the model or stray geometry as the cause.

I think I know what might be happening.
When User 'A' gives a push to the object user 'B' sees the object dip down momentarily. This may be causing a collision between the the object and the floor for 'B'. Hence the spin is always in the direction of the push.
The smaller the gap is between the moving entity and the floor the more likely it is that the rogue rotation will begin.
Also after a short time the entity rotation is corrected and both observers see it in the same orientation.

I hope this is of some help.

Nex-Pro:
@AndrewMeadows I think this might be related to https://highfidelity.fogbugz.com/f/cases/2871/Physical-Entities-jitter-around-on-entity-update-for-non-physics-owners

Nex-Pro:
@AndrewMeadows I'm curious as to what would trigger the negative Y movement for third parties when this happens.

AndrewMeadows:
Thanks @Nex-Pro, that makes sense: a slight position discrepancy between UserA (owner) and UserB (observer) could cause significant physics divergence when the object is being grabbed and moved just above a surface: were it to float over for UserA but drag slightly for UserB.  All of my testing was done for a weightless object that was not near the floor at all.

I'm not sure what causes the vertical drop on the observer side.  It might be caused by some extrapolation during the time-of-flight of the position update when the object has a non-zero gravity.  I tried to reduce this problem by including the "measured acceleration" in the update which should be used for time-of-flight extrapolation instead of gravity.  So maybe a bug in that system or perhaps there is some fundamental flaw of the system of which I'm unaware.

PR #10876, opened 2017-07-04
Fix AC crash (buffer overflow)

Atlante45:
This PR fixes a buffer overflow in the `AvatarData::toByteArray` method.
We use to assume the data would fit in an MTU sized packet and just write to the raw storage.
This PR changes the code such that is allocates upfront enough memory for all the data.

[Bug](https://highfidelity.fogbugz.com/f/cases/2927/Avatar-mixer-crashed-by-a-particular-avatar-the-crashatar)

### DIFF
--- a/assignment-client/src/Agent.cpp
+++ b/assignment-client/src/Agent.cpp
@@ -614,11 +614,11 @@ void Agent::processAgentAvatar() {
             if (avatarByteArray.size() > maximumByteArraySize) {
                 qWarning() << " scriptedAvatar->toByteArrayStateful() without facial data resulted in very large buffer:" << avatarByteArray.size() << "... reduce to MinimumData";
                 avatarByteArray = scriptedAvatar->toByteArrayStateful(AvatarData::MinimumData, true);
-            }
 
-            if (avatarByteArray.size() > maximumByteArraySize) {
-                qWarning() << " scriptedAvatar->toByteArrayStateful() MinimumData resulted in very large buffer:" << avatarByteArray.size() << "... FAIL!!";
-                return;
+                if (avatarByteArray.size() > maximumByteArraySize) {
+                    qWarning() << " scriptedAvatar->toByteArrayStateful() MinimumData resulted in very large buffer:" << avatarByteArray.size() << "... FAIL!!";
+                    return;
+                }
             }
         }
 
--- a/assignment-client/src/Agent.cpp
+++ b/assignment-client/src/Agent.cpp
@@ -605,7 +605,7 @@ void Agent::processAgentAvatar() {
         AvatarData::AvatarDataDetail dataDetail = (randFloat() < AVATAR_SEND_FULL_UPDATE_RATIO) ? AvatarData::SendAllData : AvatarData::CullSmallData;
         QByteArray avatarByteArray = scriptedAvatar->toByteArrayStateful(dataDetail);
 
-        auto maximumByteArraySize = NLPacket::maxPayloadSize(PacketType::AvatarData) - sizeof(AvatarDataSequenceNumber);
+        int maximumByteArraySize = NLPacket::maxPayloadSize(PacketType::AvatarData) - sizeof(AvatarDataSequenceNumber);
 
         if (avatarByteArray.size() > maximumByteArraySize) {
             qWarning() << " scriptedAvatar->toByteArrayStateful() resulted in very large buffer:" << avatarByteArray.size() << "... attempt to drop facial data";
--- a/assignment-client/src/Agent.cpp
+++ b/assignment-client/src/Agent.cpp
@@ -604,6 +604,24 @@ void Agent::processAgentAvatar() {
 
         AvatarData::AvatarDataDetail dataDetail = (randFloat() < AVATAR_SEND_FULL_UPDATE_RATIO) ? AvatarData::SendAllData : AvatarData::CullSmallData;
         QByteArray avatarByteArray = scriptedAvatar->toByteArrayStateful(dataDetail);
+
+        auto maximumByteArraySize = NLPacket::maxPayloadSize(PacketType::AvatarData) - sizeof(AvatarDataSequenceNumber);
+
+        if (avatarByteArray.size() > maximumByteArraySize) {
+            qWarning() << " scriptedAvatar->toByteArrayStateful() resulted in very large buffer:" << avatarByteArray.size() << "... attempt to drop facial data";
+            avatarByteArray = scriptedAvatar->toByteArrayStateful(dataDetail, true);
+
+            if (avatarByteArray.size() > maximumByteArraySize) {
+                qWarning() << " scriptedAvatar->toByteArrayStateful() without facial data resulted in very large buffer:" << avatarByteArray.size() << "... reduce to MinimumData";
+                avatarByteArray = scriptedAvatar->toByteArrayStateful(AvatarData::MinimumData, true);
+            }
+
+            if (avatarByteArray.size() > maximumByteArraySize) {
+                qWarning() << " scriptedAvatar->toByteArrayStateful() MinimumData resulted in very large buffer:" << avatarByteArray.size() << "... FAIL!!";
+                return;
+            }
+        }
+
         scriptedAvatar->doneEncoding(true);
 
         static AvatarDataSequenceNumber sequenceNumber = 0;
--- a/libraries/avatars/src/AvatarData.cpp
+++ b/libraries/avatars/src/AvatarData.cpp
@@ -491,11 +491,14 @@ QByteArray AvatarData::toByteArray(AvatarDataDetail dataDetail, quint64 lastSent
         unsigned char* validityPosition = destinationBuffer;
         unsigned char validity = 0;
         int validityBit = 0;
+        int numValidityBytes = (int)std::ceil(numJoints / (float)BITS_IN_BYTE);
 
 #ifdef WANT_DEBUG
         int rotationSentCount = 0;
         unsigned char* beforeRotations = destinationBuffer;
 #endif
+
+        destinationBuffer += numValidityBytes; // Move pointer past the validity bytes
 
         if (sentJointDataOut) {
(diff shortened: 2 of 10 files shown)


ZappoMan:
We need to test the performance of this. I think you could do this in a higher performance way - for the smaller well known size sections in particular. The toByteArray was already one of the more expensive part of the avatar mixer - we don't want to significantly impact performance.

libraries/avatars/src/AvatarData.cpp
```diff
@@ -188,15 +188,18 @@ QByteArray AvatarData::toByteArray(AvatarDataDetail dataDetail, quint64 lastSent
...
 
-    QByteArray avatarDataByteArray(udt::MAX_PACKET_SIZE, 0);
-    unsigned char* destinationBuffer = reinterpret_cast<unsigned char*>(avatarDataByteArray.data());
-    unsigned char* startPosition = destinationBuffer;
+    QByteArray avatarDataByteArray;
+    avatarDataByteArray.reserve(udt::MAX_PACKET_SIZE);
+
+    auto safeCopy = [&](const void* src, int size) {
+        avatarDataByteArray.append(reinterpret_cast<const char*>(src), size);
+    };
```
hyperlogic: I don't see why you need to use a lambda here.  IMHO, local lambdas like this are bad pattern.  This should be a static function or inlined directly because you only invoke it once.

ZappoMan: it's also not entirely clear to me how this is "safe" sure - you're appending, which won't overflow... but it doesn't ensure that the code actually ends up making a less than MTU sized buffer.

What you really need to do is:
1) try to copy all the "desired" parts of the avatar into the buffer.
2) IF you go over the MTU size, you should not copy those parts -- AND switch the flags to indicate that that portion didn't fit
3) depending on the caller, you should inform them that the parts didn't fit, so they can keep track of sending it the next time.

Atlante45:
The actual fix in the first 2 commits. the following 2 commits are small optimizations that are optional and I'm willing to remove.

Fix:
1) Compute the maximum size the byte array can be based on the data (numJoints/numBlendshapes)
    - Use that size to create the array and ensure we never write more than that amount of bytes to it.
2) Add behavior to the Interface and the Agent similar to the one in the Avatar Mixer where it tries to repack the byte array with less data if the one generated does not fit.


Optimisations:
1) We used to check the sizes of all the `AvatarDataPacket` structs at runtime and `abort` if they didn't match.
    - Moved those to static asserts so that we won't even be able to compile if they don't match.
2) We used to use 2 loops back to back to write the validity bit mask and then the data.
    - Changed it to a single loop using 2 pointers to write at different points into the array.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1365/HighFidelity-Beta-PR10876-3552ec6430f72771299df61ee6476b93a897de5f.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1369/HighFidelity-Beta-PR10876-9a1d38384d4693fc34ae4b3bd4875a47ea6266fb.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1369/HighFidelity-Beta-PR10876-9a1d38384d4693fc34ae4b3bd4875a47ea6266fb.dmg

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1371/HighFidelity-Beta-PR10876-4c8dd8e59b7b76e58fddd588980f146af9cf32bb.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1371/HighFidelity-Beta-PR10876-4c8dd8e59b7b76e58fddd588980f146af9cf32bb.exe

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1391/HighFidelity-Beta-PR10876-1d07ab70f0c5d3e2beb5dffe7cc949dfe8fdb8c0.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1391/HighFidelity-Beta-PR10876-1d07ab70f0c5d3e2beb5dffe7cc949dfe8fdb8c0.dmg

Atlante45:
## Repro:

This is a heap corruption, so it is somewhat hard to repro.

On master:
- Start you sandbox
- Start 2 Interfaces and go to your sandbox
- Open DS log (Sandbox tray icon > View Logs > Domain Server tab)
- Change your avatar to `https://ia801500.us.archive.org/28/items/pk_cetanu_hotmail_Zzz_201701/upload5/zzz.fst`
- Move the avatar around
- Eventually, your interface will crash or you will see that in the DS log:

>[07/06 13:34:52] [DEBUG] [hifi.networking] **Killed "Avatar Mixer"** (W) ...
>[07/06 13:34:52] [DEBUG] [default] Reset UUID for assignment - UUID: ...
>[07/06 13:34:52] [DEBUG] [default] Deploying assignment - UUID: ...
>[07/06 13:34:52] [DEBUG] [default] Assignment deployed with ...
>[07/06 13:34:52] [DEBUG] [hifi.networking] **Added "Avatar Mixer"** (W) ...

Which means the Avatar Mixer crashed.

Given the random nature of the crash, this could take a few minutes to occur.

## Test Plan:

With this PR:
- Start Sandbox
- Start 2 Interfaces and go to your sandbox
- Open DS log (Sandbox tray icon > View Logs > Domain Server tab)
- Change one avatar to `https://ia801500.us.archive.org/28/items/pk_cetanu_hotmail_Zzz_201701/upload5/zzz.fst`
- Change the other avatar to any avatar from the marketplace (or keep the default avatar)
- Move both avatars alternatively and make sure they can see each other move.
- Do that for several minutes to make sure neither Interface or the avatar mixer crashes as mentioned above.

YuppyPlays:
Testing.

YuppyPlays:
Will test when built.

YuppyPlays:
Testing complete:

In Master: **Check, Crashed as expected, though my crash was instant with the interface when loading in the avatar script provided.

In this PR: **Pass, No crash occurred, for 2 minutes I moved each avatar back and fourth and had no DS logs regarding avatar mixer or otherwise, no crashes either :)**

Also general smoke tested to ensure basic functions are working in this PR. All looking good other than some small things already mentioned in current master (such as noise reduction not enabling).

Atlante45:
@hyperlogic @ZappoMan The code has been totally re-written using a different approach is ready for review again.

assignment-client/src/Agent.cpp
```diff
@@ -604,6 +604,24 @@ void Agent::processAgentAvatar() {
...
+
+            if (avatarByteArray.size() > maximumByteArraySize) {
+                qWarning() << " scriptedAvatar->toByteArrayStateful() without facial data resulted in very large buffer:" << avatarByteArray.size() << "... reduce to MinimumData";
+                avatarByteArray = scriptedAvatar->toByteArrayStateful(AvatarData::MinimumData, true);
+            }
+
+            if (avatarByteArray.size() > maximumByteArraySize) {
+                qWarning() << " scriptedAvatar->toByteArrayStateful() MinimumData resulted in very large buffer:" << avatarByteArray.size() << "... FAIL!!";
+                return;
+            }
```
AndrewMeadows: This `if` block should be inside the one above.

assignment-client/src/Agent.cpp
```diff
@@ -604,6 +604,24 @@ void Agent::processAgentAvatar() {
...
+        }
```
AndrewMeadows: The only code I can find in this PR that makes me wonder about performance is this chunk here where it is theoretically possible to pack the data three times.  So the question is: What could cause the avatar update packet to overshoot its size limit?

The answer appears to be: (a) number of face tracker coefficients and (b) number of joints.

How many of each? It isn't clear.

This is a potential "grief mode" in that someone could make a particularly data-heavy avatar and use it to load an avatar-mixer with re-packings, but we have worse/easier modes.  We should worry about this when it starts to happen.

Atlante45: I took this behavior, directly from the Avatar Mixer and applied it to the clients too to have a consistent behavior and to make sure we still send some data even if someone has a bad avatar.

The code added to that PR could not be used as a grief mode, unless it is to grief itself, as it only runs on the client of the person wearing the avatar.
However, the existing code in the Mixer could be exploited that way.

I agree that this is an issue that should be fixed, but the goal of that PR is only to fix the client/mixer crash that can occur due to big/bad avatars.

The second step of that fix will consist in figuring out a way to handle an avatar with a lot of joints, possibly by fragmenting the updates (they would still be atomic updates) and reducing the send rate for said avatar.

Designing a new function that does not require rebuilding the byte array 3 times in the worst case scenario could be part of that second step.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1515/HighFidelity-Beta-PR10876-c32f4ab0d3e0d82974c6d6b5f8c787d331e833ce.exe

Atlante45:
test this please

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1524/HighFidelity-Beta-PR10876-c32f4ab0d3e0d82974c6d6b5f8c787d331e833ce.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1524/HighFidelity-Beta-PR10876-c32f4ab0d3e0d82974c6d6b5f8c787d331e833ce.exe

PR #10913, opened 2017-07-07
Introducing the Stages to the render::Scene, Better scene update phase and getting away from the singleton model

samcake:
This PR do a bunch of cleaning and sorting of the Scene / Stages object which were so far held in the DeferredLightingEffect singleton.
Since we introduced the 2nd camera we discovered a bunch of problems with the way the light Stage and Background stage were stored in the DeferredLightingEffect singleton and also with the different rate at which the scene objects need to be  updated vs the rate at which they are used fro rendering (because of the 2nd camera, there are now 2 views in flight).

This pr:
- Introduce  the Stage class to the Scene, the Scene can have several Stages which are representing sub-scene of specialized components. These Stages are available by name from the Scene. For example the light stage representing all the known lights in the scene, the Backround stage and more to come.
 Right now the Stage class interface is very simple and simply provide a way to retrieve it on the Scene.

- Introduce a clear Scene update task where we will do all scene updates job(s) once per render frame.
  This task integrate the flushing of the current queue of transactions to the scene (and not done explicitly before calling engine run)
- This Task also contains a bunch of initialization jobs to install properly the different stages and default lighting environment (work which was done in the DeferredLightingEffect singleton previously)

- IN a bunch of places where the rendering was relying on the DeferredLightingEffect singleton to get to the light stage we have replaced this by accessing it from the scene. this is done by starting from the RenderArg available almost in every render calls: args->scene->getStage<LightStage>()
I had to pass the RenderArg instead of sometime the simple Batch in order to do this which explains the multitude of render() calls touched...

## TEST PLAN
THis should work exactly the same, the lighting should work correctly, the behavior of the zone global lighting and ambient and skybox hsould work the same (can use the ZOne stack view for checking)
ANd the performance should be not worse.


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1439/HighFidelity-Beta-PR10913-4c1bf6af4716f8dfe6b6ea5cecf6865dcd9d1d3d.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1439/HighFidelity-Beta-PR10913-4c1bf6af4716f8dfe6b6ea5cecf6865dcd9d1d3d.exe

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1440/HighFidelity-Beta-PR10913-a8086764dae94e8195d939d69ed76a03e8d99d2f.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1440/HighFidelity-Beta-PR10913-a8086764dae94e8195d939d69ed76a03e8d99d2f.dmg

Nex-Pro:
Will test this when it is ready.

Nex-Pro:
Testing

Nex-Pro:
### PASS

Visited various domains in this build and stable. Did a snapshot comparison of one location in photoshop, against stable, no noticeable difference in rendering. Zone stacks looked the same.

libraries/render-utils/src/BackgroundStage.cpp
```diff
@@ -52,27 +54,22 @@ BackgroundStage::BackgroundPointer BackgroundStage::removeBackground(Index index
...
-  
     const auto& lightingModel = inputs;
     if (!lightingModel->isBackgroundEnabled()) {
         return;
     }
 
-
     // Background rendering decision
-    auto backgroundStage = DependencyManager::get<DeferredLightingEffect>()->getBackgroundStage();
+    auto backgroundStage = renderContext->_scene->getStage<BackgroundStage>();
```
AndrewMeadows: `RenderArgs` has a `_scene`, and `RenderContext` has a `_scene`.  Maybe they are the same `_scene`?  The difference between these two things gets fuzzy in my mind.

libraries/entities-renderer/src/RenderableZoneEntityItem.cpp
```diff
@@ -554,11 +554,11 @@ void RenderableZoneEntityItemMeta::setProceduralUserData(QString userData) {
 
 void RenderableZoneEntityItemMeta::render(RenderArgs* args) {
     if (!_stage) {
-        _stage = DependencyManager::get<DeferredLightingEffect>()->getLightStage();
+        _stage = args->_scene->getStage<LightStage>();
```
AndrewMeadows: Some places check the return value of `_scene->getStage<LightStage>()` but some (like this one) do not.  How do we know when it is safe to not bother validating that the returned stage pointer is valid?

interface/src/Util.cpp
```diff
@@ -34,7 +34,7 @@
 
 using namespace std;
 
-void renderWorldBox(gpu::Batch& batch) {
+void renderWorldBox(RenderArgs* args, gpu::Batch& batch) {
```
samcake: Legacy...
WE have been using RenderArgs* all over the place, so this is how i added it here.
IN an ideal world i would like to pass only  RenderArgs* which contains also the batch in use (as you are suggesting a few comments below).
I didn't do it yet in this pr by fear of breaking something.
i would suggest we move forward with this ugly signature discrepencies so we can get the "Stage" feature in .
I can address the signature and unify on RenderArgs* in a later pr

interface/src/ui/overlays/Text3DOverlay.cpp
```diff
@@ -138,7 +138,7 @@ void Text3DOverlay::render(RenderArgs* args) {
     _textRenderer->draw(batch, 0, 0, getText(), textColor, glm::vec2(-1.0f), getDrawInFront());
     //        so before we continue, we must reset the pipeline
     batch.setPipeline(args->_pipeline->pipeline);
-    args->_pipeline->prepare(batch);
+    args->_pipeline->prepare(batch, args);
```
samcake: About the ugly signature agreed, see my comment above.
I would suggest to move forward with this in this pr to avoid breaking something else.

About the args->pipeline, that's a good point, i can simply rename the member.

libraries/avatars-renderer/src/avatars-renderer/Avatar.cpp
```diff
@@ -610,7 +610,7 @@ void Avatar::render(RenderArgs* renderArgs) {
     if (showCollisionShapes && shouldRenderHead(renderArgs) && _skeletonModel->isRenderable()) {
         PROFILE_RANGE_BATCH(batch, __FUNCTION__":skeletonBoundingCollisionShapes");
         const float BOUNDING_SHAPE_ALPHA = 0.7f;
-        _skeletonModel->renderBoundingCollisionShapes(*renderArgs->_batch, getUniformScale(), BOUNDING_SHAPE_ALPHA);
+        _skeletonModel->renderBoundingCollisionShapes(renderArgs, *renderArgs->_batch, getUniformScale(), BOUNDING_SHAPE_ALPHA);
```
samcake: About the ugly signature agreed, see my comment above.

libraries/render-utils/src/BackgroundStage.cpp
samcake: They are the same yes ,
RenderCOntext conatins an Args and the Scene and is the first parameter of the render engine job's run call.
The RenderArgs is the parameter we pass all over the place to render individual "things", it also contains a Scene
we could at some point decide to converge even more...

libraries/entities-renderer/src/RenderableZoneEntityItem.cpp
samcake: k

samcake:
test this please

libraries/entities-renderer/src/RenderableZoneEntityItem.cpp
samcake: I have added asserts or if everywhere we re grabbing the stage
When we access it, we know it exists so asserts should be good

samcake:
test this please

samcake:
test this please

samcake:
test this please

murillodigital:
test this please

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1495/HighFidelity-Beta-PR10913-0c3755483b99ecc1a88af17ac9bb8bd1ba598e6a.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1495/HighFidelity-Beta-PR10913-0c3755483b99ecc1a88af17ac9bb8bd1ba598e6a.exe

PR #10914, opened 2017-07-07
Dont think during mouse move

sethalves:
- avoid doing a lot of work during mouse-move event hander in grab.js and WebTablet.js; save the processing for a timer callback.


sethalves:
to test:
- use the mouse to move dynamic things (in flat-screen mode)
- in flat-screen mode, switch to tablet mode and move the tablet around with the mouse

CainFoool:
Testing once built.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1447/HighFidelity-Beta-PR10914-e4aa4823f991b4587b129d234751f285b39ce87b.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1447/HighFidelity-Beta-PR10914-e4aa4823f991b4587b129d234751f285b39ce87b.exe

CainFoool:
Testing.

AndrewMeadows:
This solution looks like it would work but... the fundamental problem remains a sharp edge that will continue to cut.  Maybe we should try to fix the scripting API itself?  It would suck if every aspiring Hifi scripter had to examine their wounds, discover the edge, and then apply this band-aid: a custom state machine of javascript timers and callbacks.

jherico:
@AndrewMeadows as we clear rendering off the main thread this should cease to be an issue.

AndrewMeadows:
When we clear rendering off the main thread we would remove the code added in this PR?  Or would we leave it in as harmless legacy workaround cruft?

sethalves:
If, later on, reverting this doesn't harm performance, I'm happy to do so.

ZappoMan:
@sethalves @AndrewMeadows @jherico - regarding whether or not this is a work-around or that we should "fix the scripting API itself" -- Let's be clear... This is a question of event based programming paradigms. There's nothing "broken" about the scripting API that needs to be fixed. The fact of the matter is that when anyone writes code for "event" based systems (think every GUI application ever built in the last 25 years) they need to not write "long running blocking code" in their event handlers. The same exact mistake as this code would slow down a ground-up C++ windows app that chose to "think" on every mouse move event it received.

Yes - the work Austin is doing to move the rendering/processing code around is going to help provide additional headroom for scripts... but it will always be the case that scripts handling "events" should be smart about how they use that event data. The point is... this is not "legacy workaround cruft"... it will always be better to code mouse event handling in this manner.

PR #10875, opened 2017-07-03
Spectator Camera

zfox23:
## Description
By default, your monitor shows a preview of what you're seeing in VR. Using the Spectator Camera app, your monitor can display the view from a virtual hand-held camera - perfect for taking selfies or filming your friends!

The Discovery team slowly started working on the Spectator Camera towards the end of May. There have been some pauses and hiccups along the way - including temporarly moving all effort to a brand-new project - but we are finally ready and proud to share the Spectator Camera and its test plan. Enjoy!

## Spectator Camera Test Plan
[Follow this link for a copy of the Spectator Camera Master Test Plan.](https://docs.google.com/spreadsheets/d/1ohM8hzqHAfT2P9Qy9jtBUtrESSPWV6cTSYTCJOp-L9k/edit?usp=sharing) This copy is meant for 2017-07-13's tests and can be edited directly.

PR #10955, opened 2017-07-12
Make far clip plane distance equal to main camera's

zfox23:
This PR will be code reviewed normally, but tested as part of the overall test plan on #10875.

### DIFF
--- a/interface/src/SecondaryCamera.h
+++ b/interface/src/SecondaryCamera.h
@@ -42,7 +42,7 @@ public:
     glm::quat orientation{};
     float vFoV{ 45.0f };
     float nearClipPlaneDistance{ 0.1f };
-    float farClipPlaneDistance{ 100.0f };
+    float farClipPlaneDistance{ 16513.0f };
     SecondaryCameraJobConfig() : render::Task::Config(false) {}
 signals:
     void dirty();
--- a/interface/src/SecondaryCamera.h
+++ b/interface/src/SecondaryCamera.h
@@ -42,7 +42,7 @@ public:
     glm::quat orientation{};
     float vFoV{ 45.0f };
     float nearClipPlaneDistance{ 0.1f };
-    float farClipPlaneDistance{ 16513.0f };
+    float farClipPlaneDistance{ 16384.0f };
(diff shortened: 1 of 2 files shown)


zfox23:
[Make far clip distance for secondary camera the same as the primary render pass](https://trello.com/c/Sw0GUb2P/250-make-far-clip-distance-for-secondary-camera-the-same-as-the-primary-render-pass)

interface/src/SecondaryCamera.h
```diff
@@ -42,7 +42,7 @@ class SecondaryCameraJobConfig : public render::Task::Config { // Exposes second
     glm::quat orientation{};
     float vFoV{ 45.0f };
     float nearClipPlaneDistance{ 0.1f };
-    float farClipPlaneDistance{ 100.0f };
+    float farClipPlaneDistance{ 16384.0f };
```
ZappoMan: there's a constant for this... you should use it.

scripts/system/spectatorCamera.js
```diff
@@ -63,7 +63,7 @@
     var cameraIsDynamic = false;
     var vFoV = 45.0;
     var nearClipPlaneDistance = 0.1;
-    var farClipPlaneDistance = 100.0;
+    var farClipPlaneDistance = 16384.0;
```
ZappoMan: I think there's also a JS constant for this

zfox23: I can't find it - the `Camera` `Q_OBJECT` only gives access to the camera entity UUID, which exposes properties like position and orientation, but not far clip plane distance.

Howard had the idea of removing these vars from the JS because they're redundant. I added them because they're informative for future script writers. Your comment here makes me lean further towards wanting to remove them all...

zfox23: Haha, nevermind - you can do `Camera.frustum.projection`, but that doesn't give a nice number...I'd have to use those values to back-calculate the far clip plane distance.

I don't even think that's what you're asking about anymore.

zfox23: I'm just going to remove this entirely.

PR #10995, opened 2017-07-18
fix broken debug build

AndrewMeadows:
* stub NullDisplayPlugin::copyTextureToQuickFramebuffer()


jherico:
There's nothing to QA for this.  The NullDisplayPlugin is only instantiated in debug builds.

PR #10998, opened 2017-07-19
Overlay appears only on marketplace items

davidkelly:
Which really means entities you know are marketplace items.  Which right now, means things _you_ got from the marketplace.

### DIFF
--- a/interface/src/ui/overlays/ContextOverlayInterface.cpp
+++ b/interface/src/ui/overlays/ContextOverlayInterface.cpp
@@ -27,6 +27,7 @@ ContextOverlayInterface::ContextOverlayInterface() {
 
     _entityPropertyFlags += PROP_POSITION;
     _entityPropertyFlags += PROP_ROTATION;
+    _entityPropertyFlags += PROP_MARKETPLACE_ID;
 
     auto entityTreeRenderer = DependencyManager::get<EntityTreeRenderer>().data();
     connect(entityTreeRenderer, SIGNAL(mousePressOnEntity(const EntityItemID&, const PointerEvent&)), this, SLOT(createOrDestroyContextOverlay(const EntityItemID&, const PointerEvent&)));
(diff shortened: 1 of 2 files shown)


interface/src/ui/overlays/ContextOverlayInterface.cpp
```diff
@@ -35,27 +36,30 @@ ContextOverlayInterface::ContextOverlayInterface() {
 void ContextOverlayInterface::createOrDestroyContextOverlay(const EntityItemID& entityItemID, const PointerEvent& event) {
     if (event.getButton() == PointerEvent::SecondaryButton) {
         qCDebug(context_overlay) << "Creating Context Overlay on top of entity with ID: " << entityItemID;
```
zfox23: Please move this debug line underneath the next `if()`, so we only get this if we're actually creating the overlay :)

interface/src/ui/overlays/ContextOverlayInterface.h
```diff
@@ -56,6 +56,8 @@ public slots:
 private:
     bool _verboseLogging { true };
     QUuid _currentEntityWithContextOverlay{};
+    QString _marketplaceID;
```
zfox23: Please reset this to empty when we destroy the overlay (inside `ContextOverlayInterface::destroyContextOverlay()`)

zfox23: Also, a little nitpicky, but perhaps name this to something like `_contextualMarketplaceID` to make it easier to search for (and clearer that it isn't some generic Marketplace ID)?

davidkelly: ohhh yes -- doh!

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1781/HighFidelity-Beta-PR10998-920ac17caf0d7b002c5548995e15cb5b2fc9c1f0.exe

PR #11009, opened 2017-07-20
update EntityItem bounding cube when on changed dimensions

AndrewMeadows:
* fixed bug where entity's bounding box was not updated when making dimensions smaller

### DIFF
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1628,6 +1628,8 @@ void EntityItem::updateDimensions(const glm::vec3& value) {
     if (getDimensions() != value) {
         setDimensions(value);
         markDirtyFlags(Simulation::DIRTY_SHAPE | Simulation::DIRTY_MASS);
+        _queryAACubeSet = false;
+        dimensionsChanged();
     }
 }
 
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1423,26 +1423,20 @@ void EntityItem::setDimensions(const glm::vec3& value) {
 ///
 AACube EntityItem::getMaximumAACube(bool& success) const {
     if (_recalcMaxAACube) {
-        // * we know that the position is the center of rotation
         glm::vec3 centerOfRotation = getPosition(success); // also where _registration point is
         if (success) {
             _recalcMaxAACube = false;
-            // * we know that the registration point is the center of rotation
-            // * we can calculate the length of the furthest extent from the registration point
-            //   as the dimensions * max (registrationPoint, (1.0,1.0,1.0) - registrationPoint)
-            glm::vec3 dimensions = getDimensions();
-            glm::vec3 registrationPoint = (dimensions * _registrationPoint);
-            glm::vec3 registrationRemainder = (dimensions * (glm::vec3(1.0f, 1.0f, 1.0f) - _registrationPoint));
-            glm::vec3 furthestExtentFromRegistration = glm::max(registrationPoint, registrationRemainder);
+            // we want to compute the furthestExtent that an entity can extend out from its "position"
+            // to do this we compute the max of these two vec3s: registration and 1-registration
+            // and then scale by dimensions
+            glm::vec3 maxExtents = getDimensions() * glm::max(_registrationPoint, glm::vec3(1.0f) - _registrationPoint);
 
-            // * we know that if you rotate in any direction you would create a sphere
-            //   that has a radius of the length of furthest extent from registration point
-            float radius = glm::length(furthestExtentFromRegistration);
+            // there exists a sphere that contains maxExtents for all rotations
+            float radius = glm::length(maxExtents);
 
-            // * we know that the minimum bounding cube of this maximum possible sphere is
-            //   (center - radius) to (center + radius)
+            // put a cube around the sphere
+            // TODO? replace _maxAACube with _boundingSphereRadius
             glm::vec3 minimumCorner = centerOfRotation - glm::vec3(radius, radius, radius);
-
             _maxAACube = AACube(minimumCorner, radius * 2.0f);
         }
     } else {
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1353,12 +1353,6 @@ bool EntityItem::setProperties(const EntityItemProperties& properties) {
 
     SET_ENTITY_PROPERTY_FROM_PROPERTIES(lastEditedBy, setLastEditedBy);
 
-    AACube saveQueryAACube = _queryAACube;
-    checkAndAdjustQueryAACube();
-    if (saveQueryAACube != _queryAACube) {
-        somethingChanged = true;
-    }
-
     // Now check the sub classes
     somethingChanged |= setSubClassProperties(properties);
 
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1353,6 +1353,11 @@ bool EntityItem::setProperties(const EntityItemProperties& properties) {
 
     SET_ENTITY_PROPERTY_FROM_PROPERTIES(lastEditedBy, setLastEditedBy);
 
+    AACube saveQueryAACube = _queryAACube;
+    if (checkAndMaybeUpdateQueryAACube() && saveQueryAACube != _queryAACube) {
+        somethingChanged = true;
+    }
+
     // Now check the sub classes
     somethingChanged |= setSubClassProperties(properties);
 
--- a/libraries/entities/src/EntityItemProperties.cpp
+++ b/libraries/entities/src/EntityItemProperties.cpp
@@ -2169,5 +2169,5 @@ bool EntityItemProperties::parentRelatedPropertyChanged() const {
 }
 
 bool EntityItemProperties::queryAACubeRelatedPropertyChanged() const {
-	return parentRelatedPropertyChanged() || dimensionsChanged();
+    return parentRelatedPropertyChanged() || dimensionsChanged();
 }
--- a/libraries/entities/src/EntityItemProperties.cpp
+++ b/libraries/entities/src/EntityItemProperties.cpp
@@ -2157,12 +2157,17 @@ QList<QString> EntityItemProperties::listChangedProperties() {
     return out;
 }
 
-bool EntityItemProperties::parentDependentPropertyChanged() const {
-    return localPositionChanged() || positionChanged() ||
-        localRotationChanged() || rotationChanged() ||
-        localVelocityChanged() || localAngularVelocityChanged();
+bool EntityItemProperties::transformChanged() const {
+    return positionChanged() || rotationChanged() ||
+        localPositionChanged() || localRotationChanged();
 }
 
 bool EntityItemProperties::parentRelatedPropertyChanged() const {
-    return parentDependentPropertyChanged() || parentIDChanged() || parentJointIndexChanged();
+    return positionChanged() || rotationChanged() ||
+        localPositionChanged() || localRotationChanged() ||
+        parentIDChanged() || parentJointIndexChanged();
+}
+
+bool EntityItemProperties::queryAACubeRelatedPropertyChanged() const {
+	return parentRelatedPropertyChanged() || dimensionsChanged();
 }
--- a/libraries/entities/src/EntityItemProperties.h
+++ b/libraries/entities/src/EntityItemProperties.h
@@ -86,8 +86,9 @@ public:
 
     EntityPropertyFlags getChangedProperties() const;
 
-    bool parentDependentPropertyChanged() const; // was there a changed in a property that requires parent info to interpret?
-    bool parentRelatedPropertyChanged() const; // parentDependentPropertyChanged or parentID or parentJointIndex
+    bool transformChanged() const;
+    bool parentRelatedPropertyChanged() const;
+    bool queryAACubeRelatedPropertyChanged() const;
 
     AABox getAABox() const;
 
--- a/libraries/entities/src/EntityScriptingInterface.cpp
+++ b/libraries/entities/src/EntityScriptingInterface.cpp
@@ -221,7 +221,7 @@ QUuid EntityScriptingInterface::addEntity(const EntityItemProperties& properties
         _entityTree->withWriteLock([&] {
             EntityItemPointer entity = _entityTree->addEntity(id, propertiesWithSimID);
             if (entity) {
-                if (propertiesWithSimID.parentRelatedPropertyChanged()) {
+                if (propertiesWithSimID.queryAACubeRelatedPropertyChanged()) {
                     // due to parenting, the server may not know where something is in world-space, so include the bounding cube.
                     bool success;
                     AACube queryAACube = entity->getQueryAACube(success);
@@ -435,7 +435,7 @@ QUuid EntityScriptingInterface::editEntity(QUuid id, const EntityItemProperties&
                     entity->rememberHasSimulationOwnershipBid();
                 }
             }
-            if (properties.parentRelatedPropertyChanged() && entity->computePuffedQueryAACube()) {
+            if (properties.queryAACubeRelatedPropertyChanged() && entity->computePuffedQueryAACube()) {
                 properties.setQueryAACube(entity->getQueryAACube());
             }
             entity->setLastBroadcast(usecTimestampNow());
--- a/libraries/entities/src/EntityScriptingInterface.cpp
+++ b/libraries/entities/src/EntityScriptingInterface.cpp
@@ -435,7 +435,7 @@ QUuid EntityScriptingInterface::editEntity(QUuid id, const EntityItemProperties&
                     entity->rememberHasSimulationOwnershipBid();
                 }
             }
-            if (properties.queryAACubeRelatedPropertyChanged() || entity->computePuffedQueryAACube()) {
+            if (properties.queryAACubeRelatedPropertyChanged() && entity->checkAndMaybeUpdateQueryAACube()) {
                 properties.setQueryAACube(entity->getQueryAACube());
             }
             entity->setLastBroadcast(usecTimestampNow());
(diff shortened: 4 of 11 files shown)


AndrewMeadows:
I discovered a bug that was preventing me from sorting entities by size:  some entities had an incorrectly large bounding box.  It turned out: the bounding box would only get updated when it increased, not when it decreased.  The bounding box would remain too big forever.

libraries/entities/src/EntityItem.cpp
```diff
@@ -1423,26 +1423,20 @@ void EntityItem::setDimensions(const glm::vec3& value) {
...
-            //   that has a radius of the length of furthest extent from registration point
-            float radius = glm::length(furthestExtentFromRegistration);
+            // there exists a sphere that contains maxExtents for all rotations
+            float radius = glm::length(maxExtents);
 
-            // * we know that the minimum bounding cube of this maximum possible sphere is
-            //   (center - radius) to (center + radius)
+            // put a cube around the sphere
+            // TODO? replace _maxAACube with _boundingSphereRadius
             glm::vec3 minimumCorner = centerOfRotation - glm::vec3(radius, radius, radius);
```
AndrewMeadows: This is just a faster way to get the same thing done with new commentary.

libraries/entities/src/EntityItem.cpp
```diff
@@ -1634,6 +1628,8 @@ void EntityItem::updateDimensions(const glm::vec3& value) {
     if (getDimensions() != value) {
         setDimensions(value);
         markDirtyFlags(Simulation::DIRTY_SHAPE | Simulation::DIRTY_MASS);
+        _queryAACubeSet = false;
+        dimensionsChanged();
```
AndrewMeadows: When we change dimensions we now set some flags that will help later code recompute `_queryAACube`.

libraries/entities/src/EntityItemProperties.h
```diff
@@ -86,8 +86,9 @@ class EntityItemProperties {
 
     EntityPropertyFlags getChangedProperties() const;
 
-    bool parentDependentPropertyChanged() const; // was there a changed in a property that requires parent info to interpret?
-    bool parentRelatedPropertyChanged() const; // parentDependentPropertyChanged or parentID or parentJointIndex
+    bool transformChanged() const;
+    bool parentRelatedPropertyChanged() const;
+    bool queryAACubeRelatedPropertyChanged() const;
```
AndrewMeadows: The method `parentRelatedPropertyChanged()` was being used for multiple purposes and in some contexts it was too broad about the parameters it was checking.  I removed one method and added two others that were more specific and could be used more appropriately.

libraries/physics/src/EntityMotionState.cpp
```diff
@@ -577,7 +577,7 @@ void EntityMotionState::sendUpdate(OctreeEditPacketSender* packetSender, uint32_
         properties.setActionData(_serverActionData);
     }
 
-    if (properties.parentRelatedPropertyChanged() && _entity->computePuffedQueryAACube()) {
+    if (properties.transformChanged() && _entity->computePuffedQueryAACube()) {
         // due to parenting, the server may not know where something is in world-space, so include the bounding cube.
         properties.setQueryAACube(_entity->getQueryAACube());
     }
```
AndrewMeadows: This was one of those places where the old method was too broad.  This is where we harvest changes from the physics `RigidBody` and update properties on the `EntityItem`.  The physics engine does not change a lot of the properties that were being checked before (e.g. `parentID` and `parentJointIndex`).

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -435,7 +435,7 @@ QUuid EntityScriptingInterface::editEntity(QUuid id, const EntityItemProperties&
                     entity->rememberHasSimulationOwnershipBid();
                 }
             }
-            if (properties.parentRelatedPropertyChanged() && entity->computePuffedQueryAACube()) {
+            if (properties.queryAACubeRelatedPropertyChanged() || entity->computePuffedQueryAACube()) {
                 properties.setQueryAACube(entity->getQueryAACube());
             }
```
AndrewMeadows: This change is interesting.  The behavior of SpatiallyNestable::computePuffedQueryAACube() is as follows:

if the queryAACube is already cached: do nothing and return false
else: compute an artificially large queryAACube whether or not the object actually needs one and return true

The cases where an EntityItem needs an artificially large bounding box are (please correct me if I'm wrong @sethalves):

(a) When the entity is attached to an Avatar joint: its position changes constantly depending on the avatar's skeleton configuration so expand the bounding box so that we don't have to continually update it each frame the avatar moves
(b) When an entity has "children" attached to it we expand the bounding box so it is more likely to overlap a view that also overlaps the children

The problem I was having was that when the "AACube related properties" did change the bounding box had already been updated by the time the code reached this point and relying on a successful update of the expanded box would prevent the property change from reaching the server.

I don't think this code is quite right, but I put it up for review so @sethalves and I can consider it: either the `entity->computePuffedQueryAACube()` is completely unnecessary at this point (already having been computed in all cases) or my new code will sometimes NOT update the bounding box when it should.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1803/HighFidelity-Beta-PR11009-fcd3126b85c2f8687d0ffc4f1148720aa28fa747.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1803/HighFidelity-Beta-PR11009-fcd3126b85c2f8687d0ffc4f1148720aa28fa747.dmg

libraries/shared/src/SpatiallyNestable.cpp
```diff
@@ -999,7 +999,9 @@ bool SpatiallyNestable::computePuffedQueryAACube() {
     bool success;
     AACube currentAACube = getMaximumAACube(success);
     // make an AACube with edges thrice as long and centered on the object
-    _queryAACube = AACube(currentAACube.getCorner() - glm::vec3(currentAACube.getScale()), currentAACube.getScale() * 3.0f);
+	const float PUFF_COEFICIENT = _children.size() > 0 ? 3.0f : 1.0f;
```
sethalves: indent wrong, spel COEFFICIENT

libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -2157,12 +2157,17 @@ QList<QString> EntityItemProperties::listChangedProperties() {
...
 
 bool EntityItemProperties::parentRelatedPropertyChanged() const {
-    return parentDependentPropertyChanged() || parentIDChanged() || parentJointIndexChanged();
+    return positionChanged() || rotationChanged() ||
+        localPositionChanged() || localRotationChanged() ||
+        parentIDChanged() || parentJointIndexChanged();
+}
+
+bool EntityItemProperties::queryAACubeRelatedPropertyChanged() const {
+	return parentRelatedPropertyChanged() || dimensionsChanged();
```
sethalves: indent wrong

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1806/HighFidelity-Beta-PR11009-0fc0cf736f6589b6b0bf7ac41bd451a1d2331f7e.dmg

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1834/HighFidelity-Beta-PR11009-a8dac0cb79b49b2646d34780a43374b6322737e0.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1834/HighFidelity-Beta-PR11009-a8dac0cb79b49b2646d34780a43374b6322737e0.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1835/HighFidelity-Beta-PR11009-e1eb2236169524a9c47dc1abfe36b186a5b788a4.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1835/HighFidelity-Beta-PR11009-e1eb2236169524a9c47dc1abfe36b186a5b788a4.exe

AndrewMeadows:
### Test Plan:

Do this on hifi/master, and then again for this PR:

**(1)** Login with HMD.

**(2)** Run this script, either directly from this URL or copy to local file:
https://raw.githubusercontent.com/AndrewMeadows/hifi/e1eb2236169524a9c47dc1abfe36b186a5b788a4/scripts/developer/debugging/queryAACubeInspector.js

**(3)** The script in **(2)** will render a blue box outline around every nearby object, which represents the object's "bounding box" used for visibility checks on the server.

**(4)** Make a new entity small enough that it is easy to walk up and "grab" it.
--> note the size of its bounding box before you grab it

**(5)** Near-grab (e.g. grab directly with your hand)
--> note that the size of the bounding box will expand (3X size) the moment it is grabbed

**(6)** While grabbing the entity move it around inside its bounding box
--> the bounding box should not move

**(7)** While grabbing the entity move it outside the its bounding box
--> the bounding box should update to be centered on the entity as soon as the entity would move outside of it

**(8)** Let go of the entity
--> note that the bounding box remains big after letting go.
--> everything, including this step, should be the same behavior in both hifi/master and this PR
There is a bug whereby grabbed objects maintain their expanded bounding box when released.  **This PR is not trying to fix that one!**

This is where the behavior starts to diverge...

**(9)** Go into "create" mode, select the object and change its dimensions to be slightly bigger/smaller, but don't make it so big as to extend outside of its current bounding box
--> on hifi/master the bounding box will remain large
--> on this PR the bounding box should reduce in size to be tighter around the object

**(10)** Expand the object's dimensions such that it extends a bit outside of its current bounding box
--> the bounding box should update to enclose the object at its new size

**(11)** Reduce the one of the object's dimension components by half:
--> on hifi/master the bounding box will remain large
--> on this PR the bounding box should shrink to be tighter around the object

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

(1) Login with HMD.
**Worked in both**

(2) Run this script, either directly from this URL or copy to local file:
https://raw.githubusercontent.com/AndrewMeadows/hifi/e1eb2236169524a9c47dc1abfe36b186a5b788a4/scripts/developer/debugging/queryAACubeInspector.js
**Worked in both**

(3) The script in (2) will render a blue box outline around every nearby object, which represents the object's "bounding box" used for visibility checks on the server.
**Worked in both**

(4) Make a new entity small enough that it is easy to walk up and "grab" it.
--> note the size of its bounding box before you grab it
**Worked in Master; but in PR I seem to be getting various glitches different each time; when in Create the lasers go crazy, the bounding box is off the object (Not around it but outside of where the cube is completely) and also experienced grips functioning as lasers? Unsure if PR specific as I've had the lasers flashing in Master earlier tonight, but this is strange behaviour.**

(5) Near-grab (e.g. grab directly with your hand)
--> note that the size of the bounding box will expand (3X size) the moment it is grabbed
**Unable to continue in PR; Master worked as expected.**

(6) While grabbing the entity move it around inside its bounding box
--> the bounding box should not move
**Unable to continue in PR; Master worked as expected.**

(7) While grabbing the entity move it outside the its bounding box
--> the bounding box should update to be centered on the entity as soon as the entity would move outside of it
**Unable to continue in PR; Master worked as expected.**

(8) Let go of the entity
--> note that the bounding box remains big after letting go.
--> everything, including this step, should be the same behavior in both hifi/master and this PR
There is a bug whereby grabbed objects maintain their expanded bounding box when released. This PR is not trying to fix that one!
**Unable to continue in PR; Master worked as expected.**

This is where the behavior starts to diverge...

(9) Go into "create" mode, select the object and change its dimensions to be slightly bigger/smaller, but don't make it so big as to extend outside of its current bounding box
--> on hifi/master the bounding box will remain large
--> on this PR the bounding box should reduce in size to be tighter around the object
**Unable to continue in PR; Master worked as expected.**

(10) Expand the object's dimensions such that it extends a bit outside of its current bounding box
--> the bounding box should update to enclose the object at its new size
**Unable to continue in PR; Master worked as expected.**

(11) Reduce the one of the object's dimension components by half:
--> on hifi/master the bounding box will remain large
--> on this PR the bounding box should shrink to be tighter around the object
**Unable to continue in PR; Master worked as expected.**

Also, verify that joining two objects together should still expand the parent's bounding box:
**Unable to continue in PR; Master worked as expected.**

(12) Create two entities while running the script from (2)
**Unable to continue in PR; Master worked as expected.**
(13) Parent one to the other
--> the bounding box of the parent should expand (3X)
**Unable to continue in PR; Master worked as expected.**
(14) Remove the child from the parent
--> I dunno what will happen to the parent's bounding box at this point. As long as this PR doesn't make the bounding box worse than hifi/master this PR should remain unblocked.
**Master kept parent box after removing the child from the parent; PR**

**Fail** Please see step 4 for main issue.

AndrewMeadows:
Ok I think this is working now.  Note: I will add to the test plan: make sure UserB sees the same changing bounding cube as UserA does.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1857/HighFidelity-Beta-PR11009-29daac83a419a2338ad8ae7687b1d35a5fb35790.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1857/HighFidelity-Beta-PR11009-29daac83a419a2338ad8ae7687b1d35a5fb35790.exe

AndrewMeadows:
@YuppyPlays, thanks for testing.  This is ready again.

YuppyPlays:
Testing complete:

[As instructed, Master no longer needs to be re-tested, just the PR is used and checked/marked for the following steps]

(1) Login with HMD.
 **Check**

(2) Run this script, either directly from this URL or copy to local file:
https://raw.githubusercontent.com/AndrewMeadows/hifi/e1eb2236169524a9c47dc1abfe36b186a5b788a4/scripts/developer/debugging/queryAACubeInspector.js
 **Check**

(3) The script in (2) will render a blue box outline around every nearby object, which represents the object's "bounding cube" used for visibility checks on the server.
 **Check**

(4) Make a new entity small enough that it is easy to walk up and "grab" it.
--> note the size of its bounding cube before you grab it
 **Check, it is slightly larger than the cube itself as I believe is intentional**

(5) Near-grab (e.g. grab directly with your hand)
--> note that the size of the bounding cube will expand (3X size) the moment it is grabbed
 **Check, When grabbing the size does indeed triple in expansion**

(6) While grabbing the entity move it around inside its bounding cube
--> the bounding cube should not move
 **Check**

(7) While grabbing the entity move it outside the its bounding cube
--> the bounding cube should update to be centered on the entity as soon as the entity would move outside of it
--> everything, including this step, should be the same behavior in both hifi/master and this PR
 **Check**


This is where the behavior starts to diverge...

(7.5) When testing this PR (not necessary when testing hifi/master): login UserB who is also running the same script and watching the results introduced by UserA (note: UserB can be using a compatible interface that is not from this PR -- as long as UserB isn't changing stuff they won't affect the bounding cubes).
**Check, logged in a second laptop to spectate and kept an eye with the headset off but both screens visible on their respective monitors; User B saw the same thing as User A when going through all the tests again but with B spectating.**

(8) Let go of the entity
--> in hifi/master the bounding cube remains big after letting go.
--> in this PR the bounding cube will restore to a smaller size
 **Check**

(9) Go into "create" mode, select the object and change its dimensions to be slightly bigger/smaller, but don't make it so big as to extend outside of its current bounding cube
--> on hifi/master the bounding cube will remain large
--> on this PR the bounding cube should reduce in size to be tighter around the object
 **Check**

(10) Expand the object's dimensions such that it extends a bit outside of its current bounding cube
--> the bounding cube should update to enclose the object at its new size
 **Check**


(11) Reduce the one of the object's dimension components by half:
--> on hifi/master the bounding cube will remain large
--> on this PR the bounding cube should shrink to be tighter around the object
 **Check**

Also, verify that joining two objects together should still expand the parent's bounding cube:
 **Check**

(12) Create two entities while running the script from (2)
 **Check**

(13) Parent one to the other
--> the bounding cube of the parent should expand (3X)
 **Check**

(14) Remove the child from the parent
--> I dunno what will happen to the parent's bounding cube at this point. As long as this PR doesn't make the bounding cube worse than hifi/master this PR should remain unblocked.

**Check, the bounding box stays larger as if it's encapsuling both still up until you grab and move, after letting go re-adjusts to be tight-fitting.**

**Pass** I believe; please let me know otherwise or if I can assist any further :)

libraries/shared/src/SpatiallyNestable.cpp
```diff
@@ -946,13 +946,37 @@ AACube SpatiallyNestable::getMaximumAACube(bool& success) const {
...
+                _queryAACube += descendantAACube.getMaximumPoint();
+            }
+        });
+        _queryAACubeSet = true;
+
+        return true;
+    } else {
+        return false;
     }
-    return success;
```
ZappoMan: this is weird 
1) you have a success variable, that you don't appear to be using... it's not initialized, but you include it in a logical test. You're assuming that getMaximumAACube() will always set success to either true or false, but that's not a future proof assumption, in which case the function begins behaving non-deterministicly.
2) your two "returns" happen in a control flow, but the bottom exit from the function doesn't return anything... I realize there's no way to reach that control point... but it feels like it reads cleaner if you set success = true in the if clause and fall out to return success.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2069/HighFidelity-Beta-PR11009-6b386000dbb6d33d4d99b2e277181b100a1118f2.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2069/HighFidelity-Beta-PR11009-6b386000dbb6d33d4d99b2e277181b100a1118f2.exe

PR #11022, opened 2017-07-21
WL21463 - entitySelectionTool: use multiarg findRayIntersection inste…

1P-Cusack:
…ad of editOverlay.

We're seeing the ignoreRayIntersection flag not take effect before findRayIntersection calls.  This may be due to editOverlay and editOverlays becoming non-blocking in 1f7d2b2 .

This altered the flow in mousePressEvent significantly; the first block, intended to handle scale/clone only, started handling rotation (should have been second block) and sometimes selection (should have been third block).

Similarly, in the various rotate grabbers' onMove methods, the pickRay will no longer intersect anything other than rotateOverlayTarget; this avoids some awful behavior when scrubbing over the size and clone grabbers.

This also reverts unnecessary parts of the prior commits to keep the diff for this WL cleaner, and adds a few TODO comments to revisit about redundant statements and incorrect names.

In addition, we've noticed but not fixed herein:

* There is a minor edgecase near 0 and 180, where it's difficult to get within a degree or two of the poles occasionally.
* The scale/clone grabbers don't stay disappeared for rotation in some cases.  This doesn't impact usability anymore, but it'd be nice to determine why they come back when they hide briefly.
* The addGrabbers for yaw/pitch/roll could be deduplicated, and yaw has some multiselect "reposition" enable/disable logic that pitch and roll lack.

Reviewed-by: LaShonda Hopper <lashonda@1stplayable.com>

### DIFF
--- a/scripts/system/libraries/entitySelectionTool.js
+++ b/scripts/system/libraries/entitySelectionTool.js
@@ -1560,7 +1560,6 @@ SelectionDisplay = (function() {
             visible: rotationOverlaysVisible
         });
 
-        // TODO: we have not implemented the rotating handle/controls yet... so for now, these handles are hidden
         Overlays.editOverlay(yawHandle, {
             visible: rotateHandlesVisible,
             position: yawCorner,
@@ -3615,24 +3614,21 @@ SelectionDisplay = (function() {
         onMove: function(event) {
             var pickRay = generalComputePickRay(event.x, event.y);
             Overlays.editOverlay(selectionBox, {
-                ignoreRayIntersection: true,
                 visible: false
             });
             Overlays.editOverlay(baseOfEntityProjectionOverlay, {
-                ignoreRayIntersection: true,
                 visible: false
             });
-            Overlays.editOverlay(rotateOverlayTarget, {
-                ignoreRayIntersection: false
-            });
 
-            var result = Overlays.findRayIntersection(pickRay);
+            var result = Overlays.findRayIntersection(pickRay, true, [rotateOverlayTarget]);
 
             if (result.intersects) {
                 var center = yawCenter;
                 var zero = yawZero;
+                // TODO: these vectors are backwards to their names, doesn't matter for this use case (inverted vectors still give same angle)
                 var centerToZero = Vec3.subtract(center, zero);
                 var centerToIntersect = Vec3.subtract(center, result.intersection);
+                // TODO: orientedAngle wants normalized centerToZero and centerToIntersect
                 var angleFromZero = Vec3.orientedAngle(centerToZero, centerToIntersect, rotationNormal);
                 var distanceFromCenter = Vec3.distance(center, result.intersection);
                 var snapToInner = distanceFromCenter < innerRadius;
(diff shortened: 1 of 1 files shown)


1P-Cusack:
To test, I'd suggest trying all grabber handles: resize, clone, yaw/pitch/roll.  In addition, test selection and movement.  The overlay should show up on the correct axis for rotation.  Scrubbing over any other controls during rotation shouldn't cause jumps.  Moving around the object while rotating etc shouldn't cause problems.

Anything else you can think of re object selection, sizing, rotation, and movement might be worthwhile too.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1851/HighFidelity-Beta-PR11022-26cc8134eba5d1e1a49ef0a3a7778efdea631968.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1851/HighFidelity-Beta-PR11022-26cc8134eba5d1e1a49ef0a3a7778efdea631968.dmg

huffman:
Nice find @1P-Cusack ! I filed a bug for this particular type of issue as it might affect other scripts.

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

Okay so, I recorded 2 videos to get a side by side view of identical actions I did with a basic cube within latest Master (6903) and PR 110022, both with fresh installs, here are my results thus far:

Grabbing:
**Same on both**
Moving object:
**Same on both**
Vertical grabber icon:
**Same on both**
Rotate icon(s):
**I found in Master the angle visual was showing (pic below of comparison) and it seemed to work much better in this PR for rotating; note in master the rotation seemed to be reversed with hand movement? but in PR works as I'd expect it to**
image
Duplicating item (green box grabber):
**I found in Master it didn't seem to register me clicking on it in some instances (seemed random) - but in PR worked with every click and drag**
Size Grabbers:
**Same on both**

Notes: Rotation was definitely better in PR but there was a bit of jumpiness, though I do admit this could entirely just be my hand positioning as it wasn't always present.

Please let me know if there is anything else I can do/provide to assist, :)

1P-Cusack:
@YuppyPlays: that's interesting re the "angle visual" not showing up for you with this PR (this is made up of three `rotate*Overlay*` overlays in the code, I think).  I didn't see that behavior when I was testing, will try to flush it out.  =)  Is that 100% consistent for you?  Does it depend on where you're standing, where the object is, or other?

I'd expect a large amount of inconsistency in master, because the issue was a sort of asynchronous / race condition thing.  Sometimes the overlays would rotate correctly, sometimes they wouldn't, at which point rotation would be very unresponsive (trying to rotate against the wrong plane).  The "reversed with hand movement" is possibly another case of the rotationTargetOverlay not updating before this fix.

I'll see what I can do about the "angle visual" in here; it's probably a similar cause but different symptom.

YuppyPlays:
@1P-Cusack  Sorry I may have not been clear, I was trying to say PR it **Does appear** but in Master it does not.

1P-Cusack:
@YuppyPlays: Ah, okay.  =)  So if I'm reading this right, with this PR -- everything's the same or better, with nothing worse, compared to master?

I suppose we could try to test against a build from before 1f7d2b2 , which seems plausibly the place where it broke, but I'm not sure the best way to go about that, or if it's necessary.

YuppyPlays:
@1P-Cusack 
I can confirm this definitely is working better for me compared to the latest master :)

PR #11023, opened 2017-07-21
Fix for audio deadlock & audio device selection

jherico:
Because of a VS 2013 bug in `std::call_once`, the use of a blocking connections on the main thread to request the `AudioClient` to do something is causing a deadlock.  This is because the main thread is inside a `std::call_once` and the AudioClient _may_ enter into another `std::call_once` before it processes the blocking invocation.  

## Testing

If you're experiencing deadlocks on startup with build 6896, please test with this build and see if they still occur.

To test further...  

  * remove any `Audio/XXX/INPUT` entries from your Interface.json file (located in `~\AppData\Roaming\High Fidelity\` or `~\AppData\Roaming\High Fidelity - dev` for PR builds
 * Open Interface
 * In desktop mode, the audio input and output should match the defaults in your system control panel
 * In Oculus mode the audio inputs and output should automatically switch to the appropriate Rift default devices
 * In either mode, if you change the settings and exit interface, when you restart and go back into that mode (desktop or HMD) the settings should be preserved.

### DIFF
--- a/interface/src/scripting/AudioDevices.cpp
+++ b/interface/src/scripting/AudioDevices.cpp
@@ -63,6 +63,14 @@ QVariant AudioDeviceList::data(const QModelIndex& index, int role) const {
 void AudioDeviceList::resetDevice(bool contextIsHMD, const QString& device) {
     auto client = DependencyManager::get<AudioClient>().data();
     auto deviceName = getSetting(contextIsHMD, _mode).get();
+
+    // FIXME hack to prevent deadlock on startup.  The real fix will be to have the 
+    // audio client emit success and failure messages in response to audio device 
+    // switches made here
+#if 1
+    QMetaObject::invokeMethod(client, "switchAudioDevice", 
+        Q_ARG(QAudio::Mode, _mode), Q_ARG(QString, deviceName));
+#else
     bool switchResult = false;
     QMetaObject::invokeMethod(client, "switchAudioDevice", Qt::BlockingQueuedConnection,
         Q_RETURN_ARG(bool, switchResult),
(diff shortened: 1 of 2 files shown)


Nex-Pro:
Will test when built

Nex-Pro:
@jherico  can I assume oculus mode also applies to Vive mode ?

jherico:
@Nex-Pro the way OpenVR handles audio devices is different from the Oculus approach.  When you activate OpenVR it will optionally change your system default audio device (based on your preferences in the OpenVR control panel).  So the Vive doesn't really expose a preferred audio device in the same way as the Oculus.  I'm not sure what the interaction would be.

Nex-Pro:
@jherico I'll drop this if not needed

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1855/HighFidelity-Beta-PR11023-5b5e94700ecc20e0cfd939bec833e8f14a4e187d.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1855/HighFidelity-Beta-PR11023-5b5e94700ecc20e0cfd939bec833e8f14a4e187d.exe

Triplelexx:
Does fix the deadlock I was encountering.

interface/src/scripting/AudioDevices.cpp
```diff
@@ -85,6 +105,7 @@ void AudioDeviceList::resetDevice(bool contextIsHMD, const QString& device) {
             QMetaObject::invokeMethod(client, "switchAudioDevice", Q_ARG(QAudio::Mode, _mode));
         }
     }
+#endif
```
AndrewMeadows: Is this cruft?  or important reference implementation?

PR #11024, opened 2017-07-21
[Don't Merge] Worklist Item #21389: Representation of collision shapes need updating.

1P-Cusack:
Interim PR (NOT FINAL)
Purpose:  This PR is to allow for feedback and confirming if this approach is reasonable prior to going further.

* This is a work in progress.
* This is has the majority of the foundation of the fix for Worklist Item #21389.
** The current progress is vetting the current approach using <code>entity::Shape:Cylinder</code>
** There are some additional integration points which need to be amended or added which aren't present as of yet.
*** Planned integration points
**** ShapeEntityItem
**** RenderableShapeEntityItem
**** ShapeFactory
* Suggested Review Order:
** 4f22883
** 893ffa1

The approach here is to determine and retain the most appropriate ShapeType to associate with a <code>RenderableShapeEntityItem</code> instance so that information is available when the physics are initialized for the instance.  When the physics are set up for the instance, the correct hash key and bullet shape should be created based on this information.  This will also allow any other systems querying the info from <code>getShapeType()</code> to have the correct data dispersed to it.

The <code>computeShapeInfo</code> function appears to be the appropriate place to determine <code>ShapeType</code> as it's a prerequisite of the physics initialization process.  <code>RenderableShapeEntityItem</code> overrides <code>computeShapeInfo</code>.   This specified version will determine its <code>ShapeType</code> if needed, prior to setting up the <code>ShapeInfo</code>.  The <code>ShapeType</code> will be determined by a helper function added to the <code>ShapeFactory</code> which is responsible for creating the appropriate bullet collision shape objects for the instances of shapes.

### DIFF
--- a/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
@@ -16,6 +16,7 @@
 #include <StencilMaskPass.h>
 #include <GeometryCache.h>
 #include <PerfStat.h>
+#include <ShapeFactory.h>
 
 #include <render-utils/simple_vert.h>
 #include <render-utils/simple_frag.h>
@@ -84,6 +85,23 @@ bool RenderableShapeEntityItem::isTransparent() {
     } else {
         return getLocalRenderAlpha() < 1.0f || EntityItem::isTransparent();
     }
+}
+
+void RenderableShapeEntityItem::computeShapeInfo(ShapeInfo& info) {
+
+	if (_collisionShapeType == ShapeType::SHAPE_TYPE_NONE) {
+		if (_shape == entity::Shape::NUM_SHAPES)
+		{
+			EntityItem::computeShapeInfo(info);
+
+			//--EARLY EXIT--( allow default handling to process )
+			return;
+		}
+
+		_collisionShapeType = ShapeFactory::computeShapeType(getShape(), getDimensions());
+	}
+
+	return EntityItem::computeShapeInfo(info);
 }
 
 void RenderableShapeEntityItem::render(RenderArgs* args) {
--- a/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
@@ -48,18 +48,22 @@ RenderableShapeEntityItem::Pointer RenderableShapeEntityItem::baseFactory(const 
 }
 
 EntityItemPointer RenderableShapeEntityItem::factory(const EntityItemID& entityID, const EntityItemProperties& properties) {
-    return baseFactory(entityID, properties);
+	auto result = baseFactory(entityID, properties);
+
+	qCDebug(entities) << "Creating RenderableShapeEntityItem( " << result->_name << " ): " << result.get() << " ID: " << result->_id;
+
+	return result;
 }
 
 EntityItemPointer RenderableShapeEntityItem::boxFactory(const EntityItemID& entityID, const EntityItemProperties& properties) {
     auto result = baseFactory(entityID, properties);
-    result->setShape(entity::Cube);
+    result->setShape(entity::Shape::Cube);
     return result;
 }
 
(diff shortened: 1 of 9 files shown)


1P-Cusack:
This builds locally; however, I thought that it may run into some issues here.  This branch is based on a older version of master (tags/Release-6850) to avoid some sandbox issues I was having.  

I don't have access to see what the exact build error is, but if this needs to be resolved let me know and I can re-base it on a more recent version of master as opposed to doing it later as I'd planned.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1870/HighFidelity-Beta-PR11024-cbbb484f69a0f6c2d119c89dbd0ce782fc06c052.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1870/HighFidelity-Beta-PR11024-cbbb484f69a0f6c2d119c89dbd0ce782fc06c052.exe

libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
```diff
@@ -82,6 +88,23 @@ bool RenderableShapeEntityItem::isTransparent() {
...
+
+            //--EARLY EXIT--( allow default handling to process )
+            return;
+        }
+
+        _collisionShapeType = ShapeFactory::computeShapeType(getShape(), getDimensions());
+    }
+
+    return EntityItem::computeShapeInfo(info);
+}
```
AndrewMeadows: @1P-Cusack I think most of the work for this PR belongs here in `RenderableShapeEntityItem::computeShapeInfo()`.  What I'm envisioning is: this method configures the `ShapeInfo` argument without invoking any methods on `ShapeFactory`.  When it is done we'll have a `ShapeInfo` that can be submitted to the `ShapeManager` in return for `a btCollisionShape`.  The `ShapeManager` will either already have the correct shape and will just return a pointer to it (and increment an internal refcount) or it will ask the `ShapeFactory` to build one for it (and will the add it to its cache and set the refcount to 1).

I don't think we'll need to change the `ShapeFactory` much, except to be able to handle cases when the best collision shape happens to be some fundamental Bullet shape that the `ShapeFactory` doesn't yet handle: cylinder, cone, maybe  something else.  Most of the time the `ShapeInfo` will be configured (by `RenderableShapeEntityItem::computeShapeInfo()`) to be a `ConvexHull`, which the `ShapeFactory` already knows how to make.

For example, a tetrahedron is just a convex hull with 4 points.  A dodecahedron is a convex hull with 12 points.  When the `RenderableShapeEntityItem` happens to be one of those then this method should update the `ShapeInfo` with those points in a list and flag it to be a convex hull.  An axially symmetric cylinder would configure the `ShapeInfo` to be a cylinder, but an asymmetric cylinder would have to be a convex hull with appropriate points in the list.

libraries/shared/src/ShapeInfo.cpp
```diff
@@ -46,26 +47,35 @@ void ShapeInfo::setParams(ShapeType type, const glm::vec3& halfExtents, QString
 }
 
 void ShapeInfo::setBox(const glm::vec3& halfExtents) {
+    //TODO_CUSACK:  Should this pointlist clearance added in case
+    //              this is a re-purposed instance?
```
AndrewMeadows: I'm 99% sure we don't need to clear the point list here when the `ShapeInfo` is set to a box, but I'm not sure we currently do the right thing when computing the hash of the `ShapeInfo`... which would be: for the `SHAPE_TYPE_BOX` case (and other cases that don't need the point list) the hash algorithm would NOT walk the point list.

I'd have to audit the code to remember/verify.

AndrewMeadows:
@1P-Cusack, this project doesn't need to be delivered in just one PR.  You could make it work only for the cylinder-y case, ignoring the assymetry problem, and deliver that.  Then handle cylinder asymmetry in a second PR, and then handle other cases in a third.

1P-Cusack:
@AndrewMeadows splitting up the PRs sounds like a great idea, we'll put the first PR up today.  Thanks for the help and patience here.

PR #11028, opened 2017-07-23
Script getmeshes for models

sethalves:
- vertex-color support for OBJ writer
- OBJ reader is now resilient against some vertices having colors while others don't
- EntityScriptingInterface::getMeshes support for model entities
- fix a bug in BufferView::getNumElements when the buffer-view has a non-zero offset
- vertex color support for ModelScriptingInterface
- new js calls in ModelScriptingInterface: getVertexCount and getVertex

### DIFF
--- a/libraries/fbx/src/OBJReader.cpp
+++ b/libraries/fbx/src/OBJReader.cpp
@@ -118,7 +118,7 @@ glm::vec3 OBJTokenizer::getVec3() {
     auto z = getFloat();
     auto v = glm::vec3(x, y, z);
     while (isNextTokenFloat()) {
-        // the spec(s) get(s) vague here.  might be w, might be a color... chop it off.
+        // ignore any following floats
         nextToken();
     }
     return v;
@@ -130,6 +130,7 @@ bool OBJTokenizer::getVertex(glm::vec3& vertex, glm::vec3& vertexColor) {
     auto y = getFloat(); // And order of arguments is different on Windows/Linux.
     auto z = getFloat();
     vertex = glm::vec3(x, y, z);
+    vertexColor = glm::vec3(1.0f); // default, in case there is not color information
 
     auto r = 1.0f, g = 1.0f, b = 1.0f;
     bool hasVertexColor = false;
@@ -139,7 +140,7 @@ bool OBJTokenizer::getVertex(glm::vec3& vertex, glm::vec3& vertexColor) {
         // only a single value) that it's a vertex color.
         r = getFloat();
         if (isNextTokenFloat()) {
-            // Safe to assume the following values are the green/blue components. 
+            // Safe to assume the following values are the green/blue components.
             g = getFloat();
             b = getFloat();
 
@@ -351,6 +352,8 @@ bool OBJReader::parseOBJGroup(OBJTokenizer& tokenizer, const QVariantHash& mappi
     bool result = true;
     int originalFaceCountForDebugging = 0;
     QString currentGroup;
+    bool anyVertexColor { false };
+    int vertexCount { 0 };
 
     setMeshPartDefaults(meshPart, QString("dontknow") + QString::number(mesh.parts.count()));
 
@@ -416,10 +419,20 @@ bool OBJReader::parseOBJGroup(OBJTokenizer& tokenizer, const QVariantHash& mappi
 
             bool hasVertexColor = tokenizer.getVertex(vertex, vertexColor);
             vertices.append(vertex);
-            
-            if(hasVertexColor) {
+
+            // if any vertex has color, they all need to.
+            if (hasVertexColor && !anyVertexColor) {
+                // we've had a gap of zero or more vertices without color, followed
+                // by one that has color.  catch up:
+                for (int i = 0; i < vertexCount; i++) {
+                    vertexColors.append(glm::vec3(1.0f));
+                }
+                anyVertexColor = true;
+            }
+            if (anyVertexColor) {
                 vertexColors.append(vertexColor);
             }
+            vertexCount++;
         } else if (token == "vn") {
             normals.append(tokenizer.getVec3());
         } else if (token == "vt") {
--- a/libraries/fbx/src/OBJReader.cpp
+++ b/libraries/fbx/src/OBJReader.cpp
@@ -130,7 +130,6 @@ bool OBJTokenizer::getVertex(glm::vec3& vertex, glm::vec3& vertexColor) {
     auto y = getFloat(); // And order of arguments is different on Windows/Linux.
     auto z = getFloat();
     vertex = glm::vec3(x, y, z);
-    vertexColor = glm::vec3(1.0f); // default, in case there is not color information
 
     auto r = 1.0f, g = 1.0f, b = 1.0f;
     bool hasVertexColor = false;
@@ -415,7 +414,8 @@ bool OBJReader::parseOBJGroup(OBJTokenizer& tokenizer, const QVariantHash& mappi
                 #endif
             }
         } else if (token == "v") {
-            glm::vec3 vertex, vertexColor;
+            glm::vec3 vertex;
+            glm::vec3 vertexColor { glm::vec3(1.0f) };
 
             bool hasVertexColor = tokenizer.getVertex(vertex, vertexColor);
             vertices.append(vertex);
--- a/libraries/gpu/src/gpu/Buffer.h
+++ b/libraries/gpu/src/gpu/Buffer.h
@@ -192,7 +192,7 @@ public:
     BufferView(const BufferPointer& buffer, Size offset, Size size, const Element& element = DEFAULT_ELEMENT);
     BufferView(const BufferPointer& buffer, Size offset, Size size, uint16 stride, const Element& element = DEFAULT_ELEMENT);
 
-    Size getNumElements() const { return (_size - _offset) / _stride; }
+    Size getNumElements() const { return _size / _stride; }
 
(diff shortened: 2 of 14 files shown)


hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1877/HighFidelity-Beta-PR11028-a1107deef19c713d84bf1a46c9234745bb6bf31f.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1877/HighFidelity-Beta-PR11028-a1107deef19c713d84bf1a46c9234745bb6bf31f.dmg

sethalves:
to test:
- rez a model with url http://headache.hungry.com/~seth/hifi/mesh-tests/plant-missing-color.obj
- on master it wont appear, with this PR it will appear, but have some white areas
- use CREATE and alt-drag several copies of the plant around
- press ctrl-j and run-from-url this script: http://headache.hungry.com/~seth/hifi/mesh-tests/test.js
- small spheres should appear at each vertex of the original plants
- a new single entity that looks the same as the original plants should appear 1 meter above the original

cain-kilgore:
Testing.

cain-kilgore:
# Test Passed
Test | Result
-------- | --------
rez a model with url http://headache.hungry.com/~seth/hifi/mesh-tests/plant-missing-color.obj | Pass
on master it wont appear, with this PR it will appear, but have some white areas | Pass
use CREATE and alt-drag several copies of the plant around | Pass
open running-scripts dialog and run-from-url this script: `...` | Pass
small spheres should appear at each vertex of the original plants | Pass
a new single entity that looks the same as the original plants (check size, colors, shape) should appear 1 meter above the originals | Pass
open running-scripts dialog and delete the test script | Pass

All good from here. Let me know if this needs further testing.

libraries/fbx/src/OBJReader.cpp
```diff
@@ -416,10 +419,20 @@ bool OBJReader::parseOBJGroup(OBJTokenizer& tokenizer, const QVariantHash& mappi
...
+                // we've had a gap of zero or more vertices without color, followed
+                // by one that has color.  catch up:
+                for (int i = 0; i < vertexCount; i++) {
+                    vertexColors.append(glm::vec3(1.0f));
+                }
+                anyVertexColor = true;
+            }
+            if (anyVertexColor) {
                 vertexColors.append(vertexColor);
             }
```
AndrewMeadows: `vertexColor` is declared at the top of the current loop but not initialized, but nevertheless it is stuffed into `vertexColors` in the case where there were already some colors... but maybe not for this particular vertex!  It appears that the `tokenizer.getVertex(vertex, vertexColor)` call does **not** have a side effect of updating the color if none was found in the OBJ file... which means it is theoretically possible for some vertices to get uninitialized colors.

Maybe declare `vertexColor` before the `while` loop so that subsequent uncolored vertices inherit the last color?  Or else declare AND initialize it inside the `while` loop.

Meanwhile, maybe should reserve space in `vertexColors` when it happens to be empty?
```
if (vertexColors.empty() && vertexCount > 0) {
    vertexColors.reserve(vertexCount);
}
```

libraries/gpu/src/gpu/Buffer.h
```diff
@@ -192,7 +192,7 @@ class BufferView {
     BufferView(const BufferPointer& buffer, Size offset, Size size, const Element& element = DEFAULT_ELEMENT);
     BufferView(const BufferPointer& buffer, Size offset, Size size, uint16 stride, const Element& element = DEFAULT_ELEMENT);
 
-    Size getNumElements() const { return (_size - _offset) / _stride; }
+    Size getNumElements() const { return _size / _stride; }
```
AndrewMeadows: This doesn't look right.  @samcake, please confirm.

libraries/model/src/model/Geometry.cpp
```diff
@@ -187,6 +202,12 @@ model::MeshPointer Mesh::map(std::function<glm::vec3(glm::vec3)> vertexFunc,
     gpu::BufferView resultVertexBufferView(resultVertexBufferPointer, vertexElement);
     result->setVertexBuffer(resultVertexBufferView);
 
+    gpu::Element colorElement = gpu::Element(gpu::VEC3, gpu::FLOAT, gpu::XYZ);
```
AndrewMeadows: It is a bit wasteful to store colors in a `VEC3` of `FLOAT`.  Could use `gpu::VEC3, gpu::NUINT8, gpu::RGB` but would probably have to scale the data from the OBJ accordingly.  @samcake could probably give some advice on how to do this.

libraries/render-utils/src/Model.h
```diff
@@ -257,6 +257,8 @@ class Model : public QObject, public std::enable_shared_from_this<Model> {
     int getResourceDownloadAttempts() { return _renderWatcher.getResourceDownloadAttempts(); }
     int getResourceDownloadAttemptsRemaining() { return _renderWatcher.getResourceDownloadAttemptsRemaining(); }
 
+    Q_INVOKABLE MeshProxyList getMeshes();
```
AndrewMeadows: This method could probably be `const`.

libraries/script-engine/src/ModelScriptingInterface.cpp
```diff
@@ -132,12 +153,48 @@ QScriptValue ModelScriptingInterface::transformMesh(glm::mat4 transform, MeshPro
...
 }
 
+QScriptValue ModelScriptingInterface::getVertexCount(MeshProxy* meshProxy) {
+    if (!meshProxy) {
+        return QScriptValue(false);
+    }
+    MeshPointer mesh = meshProxy->getMeshPointer();
+    if (!mesh) {
+        return QScriptValue(false);
+    }
```
AndrewMeadows: Interesting that `false` is being returned here.  Should maybe return `0`?

libraries/script-engine/src/ModelScriptingInterface.h
```diff
@@ -29,6 +29,8 @@ class ModelScriptingInterface : public QObject {
     Q_INVOKABLE QScriptValue newMesh(const QVector<glm::vec3>& vertices,
                                      const QVector<glm::vec3>& normals,
                                      const QVector<MeshFace>& faces);
+    Q_INVOKABLE QScriptValue getVertexCount(MeshProxy* meshProxy);
+    Q_INVOKABLE QScriptValue getVertex(MeshProxy* meshProxy, int vertexIndex);
```
AndrewMeadows: The script can only access one vertex at a time?  Would be cool if it could get all of the vertices as an array.  The "JS variable to `QScriptValue` to C++ calculations back to `QScriptValue` to JS variable" round trip is _soo ssllooww_.

libraries/fbx/src/OBJReader.cpp
sethalves: vertexColor is always initialized by getVertex.

libraries/gpu/src/gpu/Buffer.h
sethalves: it was wrong before.  this function isn't used much, and it was previously never used on a buffer with an offset.

libraries/model/src/model/Geometry.cpp
sethalves: I have to match what other parts of the system do.  I can't change the format for just this part of the code.

libraries/script-engine/src/ModelScriptingInterface.cpp
sethalves: well, maybe it's not a good interface, but by returning false the js can tell the difference between an unready mesh and an empty one.

libraries/script-engine/src/ModelScriptingInterface.h
sethalves: Yeah, returning an array of Vec3's might be interesting.  That's contained in the MeshProxy type, but its a black-box to js.

libraries/fbx/src/OBJReader.cpp
sethalves: I've moved the initialization to outside of getVertex so it's easier on the reader.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1914/HighFidelity-Beta-PR11028-a3f4aeb1826e2249a2f9e1332f0e108bcaf2d730.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1914/HighFidelity-Beta-PR11028-a3f4aeb1826e2249a2f9e1332f0e108bcaf2d730.exe

libraries/gpu/src/gpu/Buffer.h
AndrewMeadows: Yeah, I think you're right.  I looked at the `Buffer.h` code more carefully.  I find it a little hard to follow since there is the `Buffer::_buffer` which has a` size()` and then there is the `Buffer` which has a `_size`.  The two are not the same.

libraries/fbx/src/OBJReader.cpp
AndrewMeadows: Hrm... AFAICT `OBJTokenizer::getVertex()` doesn't update `vertexColor` when there is no color data available, but in any case: raising the context of `vertexColor` works for me.

PR #11036, opened 2017-07-24
fix bugs preventing avatar from walking on spherical planet

AndrewMeadows:
* fix direction of walk motor for arbitrarily rotated avatar
* remove unnecessary math injecting NaN's into camera orientation

PR #11039, opened 2017-07-25
WL 21477 Gravity Script

amvmoody:
# Info
- Check with desktop mode.

# QA

## Setup
Install the build from this PR https://github.com/highfidelity/hifi/pull/11036 .

## Steps - Single Person
1. Enter Create mode and create a sphere that's greater in size then 5 meters (you will orbit when you start moving otherwise).
2. Attach the Script https://raw.githubusercontent.com/amvmoody/hifi/1bf2855e9b5a4960333e5c777816a0dfadc8bd58/unpublishedScripts/gravity.js to the Sphere
3. Click on the sphere and you should start moving towards the planet provided you are not flying, otherwise fly in the down direction.
4. Walk around the sphere and make sure you are able to move in every direction [forward, backwards, rotate left, rotate right, strafe left, strafe right, fly up, fly down. ] while you are on various points of the sphere [-z, z, -y, y, -x, x] 
5. Repeat with various sphere sizes [10m, 20m, 100m]

### DIFF
--- a/unpublishedScripts/gravity.js
+++ b/unpublishedScripts/gravity.js
@@ -1,6 +1,5 @@
 //
 //  gravity.js
-//  scripts/system/
 //
 //  Created by Alan-Michael Moody on 7/24/17
 //  Copyright 2016 High Fidelity, Inc.
--- a/unpublishedScripts/gravity.js
+++ b/unpublishedScripts/gravity.js
@@ -1,0 +1,55 @@
+//
+//  gravity.js
+//  scripts/system/
+//
+//  Created by Alan-Michael Moody on 7/24/17
+//  Copyright 2016 High Fidelity, Inc.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+(function () {
+
+    var _entityID;
+
+    this.preload = function (entityID) {
+        _entityID = entityID;
+    };
+
+    function update(deltatime) {
+        var planet = Entities.getEntityProperties(_entityID);
+        var direction = Vec3.normalize(Vec3.subtract(MyAvatar.position, planet.position));
+        var localUp = Quat.getUp(MyAvatar.orientation);
+
+        MyAvatar.orientation = Quat.multiply(Quat.rotationBetween(localUp, direction), MyAvatar.orientation);
+    }
+
+    function init() {
+        Script.update.connect(update);
+    }
+
+    function clean() {
+        Script.update.disconnect(update);
+        MyAvatar.orientation = Quat.fromVec3Degrees({
+            x: 0,
+            y: 0,
+            z: 0
+        });
+    }
+
+    var _switch = true;
+
+    this.clickDownOnEntity = function(uuid, mouseEvent) {
+
+        if (_switch) {
+            init();
+        } else {
+            clean();
+        }
+        _switch = !_switch;
+    };
+
+    Script.scriptEnding.connect(clean);
+
+});
--- a/unpublishedScripts/gravity.js
+++ b/unpublishedScripts/gravity.js
@@ -18,10 +18,12 @@
 
     function update(deltatime) {
         var planet = Entities.getEntityProperties(_entityID);
-        var direction = Vec3.normalize(Vec3.subtract(MyAvatar.position, planet.position));
+
+        //normalization happens in rotationBetween.
+        var direction = Vec3.subtract(MyAvatar.position, planet.position);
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1913/HighFidelity-Beta-PR11039-1bf2855e9b5a4960333e5c777816a0dfadc8bd58.dmg

unpublishedScripts/gravity.js
```diff
@@ -0,0 +1,55 @@
...
+        _entityID = entityID;
+    };
+
+    function update(deltatime) {
+        var planet = Entities.getEntityProperties(_entityID);
+        var direction = Vec3.normalize(Vec3.subtract(MyAvatar.position, planet.position));
+        var localUp = Quat.getUp(MyAvatar.orientation);
+
+        MyAvatar.orientation = Quat.multiply(Quat.rotationBetween(localUp, direction), MyAvatar.orientation);
+    }
```
AndrewMeadows: Not a blocker for this work, but after looking at this logic I decided to audit the code for:

`MyAvatar.orientation()` =   which maps to `SpatiallyNestable::setOrientation()`
`Quat.multiply()`   which maps to `Quat::multiply()`

The first does not normalize incoming quaternions.  The second does not automatically normalize the result.  This means there is a danger that floating point drift could feedback in the `A = B * A` math and eventually cause MyAvatar's orientation to explode into NaN (Not a Number).  I should add some code in the method that handles setting MyAvatar's orientation via script to always normalize the incoming quaternion.

unpublishedScripts/gravity.js
```diff
@@ -0,0 +1,55 @@
...
+
+    var _entityID;
+
+    this.preload = function (entityID) {
+        _entityID = entityID;
+    };
+
+    function update(deltatime) {
+        var planet = Entities.getEntityProperties(_entityID);
+        var direction = Vec3.normalize(Vec3.subtract(MyAvatar.position, planet.position));
```
AndrewMeadows: Normalization of `direction` isn't necessary.  `Quat.rotationBetween(A, B)` will normalize the arguments before computing the result.

amvmoody: Removed normalization here.

unpublishedScripts/gravity.js
amvmoody: Added normalization here.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1920/HighFidelity-Beta-PR11039-34594b914da0a3679878c47fc885769b910edeef.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1920/HighFidelity-Beta-PR11039-34594b914da0a3679878c47fc885769b910edeef.exe

Nex-Pro:
Testing this with the above build

Nex-Pro:
Strafing is broken. Avatar will walk  in a square pattern when holding shift+a or shift + d on a 10m sphere.

The effect isn't very reliable. Sometimes not triggering other times avatar falls off.

On a large sphere 1000m I found my avatar to be strafing left away from the surface while on the near vertical side of the sphere. The  avatar was upright. Avatar orientation snapped back when I attempted to get a screen shot. 
Avatar also became apparently frozen in a falling pose on looking at stats it was falling at more than 400m sec. and 22kms down. HELP ! No visual frame of reference. 

I'm pretty sure @thoys  did this more than a year ago.

AndrewMeadows:
@Nex-Pro, when you say "Testing this with the above build"... do  you mean you're using the windows installer from this PR or from PR-11036 as requested in the test plan?

The strafing problems and also problems when walking near the sphere's equator sound like issues that were fixed in the C++ of PR-11036.

The re-orientation logic in the JS script appears to be mathematically correct.  If there are any glitches in walking around a sphere I doubt they are caused by the script; they would have to be caused by fundamental bugs in the C++.

Nex-Pro:
@AndrewMeadows Tested with both builds And yes the strafing is fixed.

Nex-Pro:
@AndrewMeadows Tried this with a 1km sphere and the avatars are well clear of the surface. I know that isn't related to this PR.  It is probably due to the resolution of the sphere and the collisions being calculated based on the radius. Basically a limitation of the sphere primitive.

Nex-Pro:
Ok the triggering thing is simply that sometimes the click is correctly disconnecting the avatar but you don't notice that you are disconnected then walk off the sphere.

Nex-Pro:
### PASS

PR #11043, opened 2017-07-25
Updates from Threaded Rendering project

jherico:
This is a set of changes from the threaded rendering project that can be merged to master without impacting performance or behavior, in order to prevent too much drift in the project branch.

* Adds support for additional threads emitting warnings if blocking method invocations are made from them. Will be needed because the render thread absolutely cannot make blocking invocations.
* Add support for `moveToNewNamedThread` where a lambda can be provided that takes the target thread as a parameter.  This is needed if an object needs to move additional components to the target thread prior to the thread starting (such as an OpenGL context wrapper)
* Modify `moveToNewNamedThread` to set the thread name so that it's visible in the Visual Studio debugger.  This happens automatically inside Qt for debug builds, but not for release builds.  Since we primarily run on release, this should aid debugging.
* Remove superfluous member from `render::Args`.  This was only used in the `RenderableWebEntityItem` and can be fetched there via the dependency manager without having to pass through `Args`
* Change display plugin presentation notifications from an event to a signal.  
* Remove `DisplayPlugins.h` from `Application.h`, rely on forward declarations.
* Modify `Application.cpp` to account for the change to present notifications
* Clean up a number of C++ files that relied on `Application.h` for `DisplayPlugins.h`
* On systems where the rendering is stressed, tablet interaction still works as well as it does on master

## Testing

Application behavior and performance should be unchanged from the current dev build.  Verify the following

* Switching between display plugins is functional
* Application shuts down smoothly
* Application does not crash when minimized for more than 30-40 seconds
* Animations run properly and at the correct speed

### DIFF
--- a/libraries/shared/src/ThreadHelpers.cpp
+++ b/libraries/shared/src/ThreadHelpers.cpp
@@ -38,11 +38,14 @@ void setThreadName(const std::string& name) {
 
 void moveToNewNamedThread(QObject* object, const QString& name, std::function<void(QThread*)> preStartCallback, std::function<void()> startCallback, QThread::Priority priority) {
     Q_ASSERT(QThread::currentThread() == object->thread());
-    // setup a thread for the NodeList and its PacketReceiver
+
+    // Create the target thread
     QThread* thread = new QThread();
     thread->setObjectName(name);
 
-    // Execute any additional work to do before the thread starts (like moving members to the target thread
+    // Execute any additional work to do before the thread starts like moving members to the target thread.
+    // This is required as QObject::moveToThread isn't virutal, so we can't override it on objects that contain
+    // an OpenGLContext and ensure that the context moves to the target thread as well.  
     preStartCallback(thread);
 
     // Link the in-thread initialization code
@@ -54,8 +57,10 @@ void moveToNewNamedThread(QObject* object, const QString& name, std::function<vo
         startCallback();
     });
 
-    // Make sure the thread will be destroyed and cleaned up
+    // Make sure the thread will be destroyed and cleaned up.  The assumption here is that the incoming object 
+    // will be destroyed and the thread will quit when that occurs.  
     QObject::connect(object, &QObject::destroyed, thread, &QThread::quit);
+    // When the thread itself stops running, it should also be deleted.
     QObject::connect(thread, &QThread::finished, thread, &QThread::deleteLater);
 
     // put the object on the thread
--- a/libraries/shared/src/ThreadHelpers.cpp
+++ b/libraries/shared/src/ThreadHelpers.cpp
@@ -10,29 +10,66 @@
 
 #include <QtCore/QDebug>
 
+// Support for viewing the thread name in the debugger.  
+// Note, Qt actually does this for you but only in debug builds
+// Code from https://msdn.microsoft.com/en-us/library/xcb2z8hs.aspx
+// and matches logic in `qt_set_thread_name` in qthread_win.cpp
+#ifdef Q_OS_WIN
+#include <qt_windows.h>
+#pragma pack(push,8)  
+struct THREADNAME_INFO {
+    DWORD dwType; // Must be 0x1000.  
+    LPCSTR szName; // Pointer to name (in user addr space).  
+    DWORD dwThreadID; // Thread ID (-1=caller thread).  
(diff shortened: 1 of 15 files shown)


cain-kilgore:
Testing once built.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1941/HighFidelity-Beta-PR11043-576ec45384b0b4dfc6ab610f537c4e05eee59e94.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1941/HighFidelity-Beta-PR11043-576ec45384b0b4dfc6ab610f537c4e05eee59e94.dmg

cain-kilgore:
# Test Passed
Test | Result
-------- | --------
Switching between display plugins is functional | Pass
Application shuts down smoothly | Pass
Application does not crash when minimized for more than 30-40 seconds | Pass
Animations run properly and at the correct speed | Pass

Everything seems to be working okay. Animations on Avatars and the Snail appeared to be running at normal speeds.

Let me know if this needs further testing.

libraries/shared/src/ThreadHelpers.cpp
```diff
@@ -10,29 +10,66 @@
...
+    THREADNAME_INFO info{ 0x1000, name.c_str(), (DWORD)-1, 0 };
+    __try {
+        RaiseException(MS_VC_EXCEPTION, 0, sizeof(info) / sizeof(ULONG_PTR), (ULONG_PTR*)&info);
+    } __except (EXCEPTION_EXECUTE_HANDLER) { }
+#endif
+}
+
+void moveToNewNamedThread(QObject* object, const QString& name, std::function<void(QThread*)> preStartCallback, std::function<void()> startCallback, QThread::Priority priority) {
+    Q_ASSERT(QThread::currentThread() == object->thread());
+    // setup a thread for the NodeList and its PacketReceiver
```
ZappoMan: this comment seems out of place

libraries/shared/src/shared/QtHelpers.cpp
```diff
@@ -11,11 +11,24 @@
...
 
+#include "../Profile.h"
 Q_LOGGING_CATEGORY(thread_safety, "hifi.thread_safety")
 
 namespace hifi { namespace qt {
 
+static QHash<QThread*, QString> threadHash;
+static QReadWriteLock threadHashLock;
+
+void addBlockingForbiddenThread(const QString& name, QThread* thread) {
```
ZappoMan: is this "not yet used??"

libraries/shared/src/shared/QtHelpers.cpp
```diff
@@ -30,9 +43,23 @@ bool blockingInvokeMethod(
...
         qCWarning(thread_safety) << "BlockingQueuedConnection invoked on main thread from " << function;
+        return QMetaObject::invokeMethod(obj, member,
+            Qt::BlockingQueuedConnection, ret, val0, val1, val2, val3, val4, val5, val6, val7, val8, val9);
+    } 
+
+    {
+        QReadLocker locker(&threadHashLock);
+        for (const auto& thread : threadHash.keys()) {
+            if (currentThread == thread) {
+                qCWarning(thread_safety) << "BlockingQueuedConnection invoked on forbidden thread " << threadHash[thread];
```
ZappoMan: do we want an assert() or even a forced crash here?

libraries/shared/src/shared/QtHelpers.cpp
jherico: correct.

libraries/shared/src/shared/QtHelpers.cpp
jherico: maybe eventually, but for now it would be more valuable to see whether or not this is occurring in logs.

libraries/shared/src/ThreadHelpers.cpp
jherico: will remove

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2272/HighFidelity-Beta-PR11043-0e508432c4c745d08d890502d7279c062fc7b105.exe

PR #11044, opened 2017-07-25
Handle the case when the camera is inside the bounding box of an entity

zfox23:
Also remove some magic numbers and fix a bug where head orientation was used when it should have been camera orientation.

DO NOT TEST

### DIFF
--- a/interface/src/ui/overlays/ContextOverlayInterface.cpp
+++ b/interface/src/ui/overlays/ContextOverlayInterface.cpp
@@ -53,6 +53,7 @@ ContextOverlayInterface::ContextOverlayInterface() {
 static const xColor BB_OVERLAY_COLOR = {255, 255, 0};
 static const uint32_t LEFT_HAND_HW_ID = 1;
 static const xColor CONTEXT_OVERLAY_COLOR = { 255, 255, 255 };
+static const float CONTEXT_OVERLAY_INSIDE_DISTANCE = 1.0f; // in meters
 static const float CONTEXT_OVERLAY_CLOSE_DISTANCE = 1.5f; // in meters
 static const float CONTEXT_OVERLAY_CLOSE_SIZE = 0.12f; // in meters, same x and y dims
 static const float CONTEXT_OVERLAY_FAR_SIZE = 0.08f; // in meters, same x and y dims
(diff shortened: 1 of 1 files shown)


interface/src/ui/overlays/ContextOverlayInterface.cpp
```diff
@@ -14,6 +14,9 @@
 
 #include <EntityTreeRenderer.h>
 
+static const float CONTEXT_OVERLAY_TABLET_OFFSET = 30.0f; // Degrees
+static const float CONTEXT_OVERLAY_TABLET_ORIENTATION = 210.0f; // Degrees
+static const float CONTEXT_OVERLAY_TABLET_DISTANCE = 0.65F; // Meters
```
davidkelly: 👍

interface/src/ui/overlays/ContextOverlayInterface.cpp
```diff
@@ -105,15 +109,11 @@ bool ContextOverlayInterface::createOrDestroyContextOverlay(const EntityItemID&
...
-                PickRay pickRay(cameraPosition, direction);
-                _bbOverlay->setIgnoreRayIntersection(false);
-                auto result = qApp->getOverlays().findRayIntersection(pickRay);
-                _bbOverlay->setIgnoreRayIntersection(true);
-                contextOverlayPosition = result.intersection - direction * CONTEXT_OVERLAY_FAR_OFFSET;
-                contextOverlayDimensions = glm::vec2(CONTEXT_OVERLAY_FAR_SIZE, CONTEXT_OVERLAY_FAR_SIZE) * glm::distance(contextOverlayPosition, cameraPosition);
-            } else {
+            if (AABox(bbPosition - (dimensions / 2.0f), dimensions * 2.0f).contains(cameraPosition)) {
+                // If the camera is inside the box, position the context overlay 1 meter in front of the camera.
+                contextOverlayPosition = cameraPosition + 1.0f * (qApp->getCamera().getOrientation() * Vectors::FRONT);
```
davidkelly: In the spirit of getting rid of magic numbers (like you did with CONTEXT_OVERLAY_TABLET_OFFSET, etc.. in this PR), maybe we can do the same for the 1.0f?

zfox23: ah yes, great idea.

PR #11048, opened 2017-07-25
WL21389 PR1:  Representation of collision shapes need updating (details below).

1P-Cusack:
Breaking up this ticket into several smaller PRs as suggested by Andrew via: https://github.com/highfidelity/hifi/pull/11024#issuecomment-317592305

This is the foundation PR for WL21389 and has support for Cylinder-Y shape.  

**Expected Behavior:**

- **Symmetrical** cylinders with **major y-axis** should now collide as if they're cylinders as opposed to spheres/ellipsoids.
- Cubes and Spheres which benefit from the updated foundation should behave as they did before.
- Asymmetrical shapes and 'hedrons aren't currently supported.
    - Asymmetrical shapes and 'hedrons aren't expected to work properly just yet.  These will be covered in a later PR.

**Testing:**
- Enter sandbox or a domain where you have rez permissions
- Under the Create menu, choose to create a
    - Cylinder
    - Box
    - Sphere
- Collisions with these shapes should work as expected keeping in mind the aforementioned caveats and notes from Expected Behavior and Pre-Existing Issues sections.

**Pre-existing Issue(s)**
- When creating a box and experimenting with various dimensions, it was observed that at times the collision wouldn't trigger, the solid shape would disappear or the shape would be come a wireframe.  This was observed on the tags/RELEASE-6896 and tags/RELEASE-6902 branches as well.  Though it was harder to trigger the behavior on RELEASE-6902.

Note(s):
- This PR is currently based on tags/RELEASE-6902 as opposed to master@HEAD

### DIFF
--- a/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
@@ -16,6 +16,7 @@
 #include <StencilMaskPass.h>
 #include <GeometryCache.h>
 #include <PerfStat.h>
+#include <ShapeFactory.h>
 
 #include <render-utils/simple_vert.h>
 #include <render-utils/simple_frag.h>
@@ -48,11 +49,12 @@ RenderableShapeEntityItem::Pointer RenderableShapeEntityItem::baseFactory(const 
 }
 
 EntityItemPointer RenderableShapeEntityItem::factory(const EntityItemID& entityID, const EntityItemProperties& properties) {
-	auto result = baseFactory(entityID, properties);
+    auto result = baseFactory(entityID, properties);
 
-	qCDebug(entities) << "Creating RenderableShapeEntityItem( " << result->_name << " ): " << result.get() << " ID: " << result->_id;
+    //TODO_CUSACK: Remove this before final PN
+    qCDebug(entities) << "Creating RenderableShapeEntityItem( " << result->_name << " ): " << result.get() << " ID: " << result->_id;
 
-	return result;
+    return result;
 }
 
 EntityItemPointer RenderableShapeEntityItem::boxFactory(const EntityItemID& entityID, const EntityItemProperties& properties) {
@@ -84,6 +86,23 @@ bool RenderableShapeEntityItem::isTransparent() {
     } else {
         return getLocalRenderAlpha() < 1.0f || EntityItem::isTransparent();
     }
+}
+
+void RenderableShapeEntityItem::computeShapeInfo(ShapeInfo& info) {
+
+    if (_collisionShapeType == ShapeType::SHAPE_TYPE_NONE) {
+        if (_shape == entity::Shape::NUM_SHAPES)
+        {
+            EntityItem::computeShapeInfo(info);
+
+            //--EARLY EXIT--( allow default handling to process )
+            return;
+        }
+
+        _collisionShapeType = ShapeFactory::computeShapeType(getShape(), getDimensions());
+    }
+
+    return EntityItem::computeShapeInfo(info);
 }
 
 void RenderableShapeEntityItem::render(RenderArgs* args) {
--- a/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
@@ -48,18 +48,22 @@ RenderableShapeEntityItem::Pointer RenderableShapeEntityItem::baseFactory(const 
 }
 
 EntityItemPointer RenderableShapeEntityItem::factory(const EntityItemID& entityID, const EntityItemProperties& properties) {
-    return baseFactory(entityID, properties);
+	auto result = baseFactory(entityID, properties);
+
+	qCDebug(entities) << "Creating RenderableShapeEntityItem( " << result->_name << " ): " << result.get() << " ID: " << result->_id;
+
+	return result;
 }
 
 EntityItemPointer RenderableShapeEntityItem::boxFactory(const EntityItemID& entityID, const EntityItemProperties& properties) {
     auto result = baseFactory(entityID, properties);
-    result->setShape(entity::Cube);
+    result->setShape(entity::Shape::Cube);
     return result;
 }
 
 EntityItemPointer RenderableShapeEntityItem::sphereFactory(const EntityItemID& entityID, const EntityItemProperties& properties) {
     auto result = baseFactory(entityID, properties);
-    result->setShape(entity::Sphere);
+    result->setShape(entity::Shape::Sphere);
     return result;
 }
 
--- a/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
@@ -132,7 +132,7 @@ void RenderableShapeEntityItem::computeShapeInfo(ShapeInfo& info) {
         case entity::Shape::Icosahedron:
         case entity::Shape::Cone: {
             //TODO WL21389: SHAPE_TYPE_SIMPLE_HULL and pointCollection (later)
-            //_collisionShapeType = SHAPE_TYPE_SIMPLE_HULL;
+            _collisionShapeType = SHAPE_TYPE_ELLIPSOID;
         }
         break;
         case entity::Shape::Torus:
--- a/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
@@ -16,7 +16,6 @@
 #include <StencilMaskPass.h>
 #include <GeometryCache.h>
 #include <PerfStat.h>
-#include <ShapeFactory.h>
 
 #include <render-utils/simple_vert.h>
 #include <render-utils/simple_frag.h>
@@ -91,65 +90,70 @@ void RenderableShapeEntityItem::computeShapeInfo(ShapeInfo& info) {
     const glm::vec3 entityDimensions = getDimensions();
 
     switch (_shape){
-    case entity::Shape::Quad:
-    case entity::Shape::Cube: {
-        _collisionShapeType = SHAPE_TYPE_BOX;
-    }
-    break;
-    case entity::Shape::Sphere: {
-
-        float diameter = entityDimensions.x;
-        const float MIN_DIAMETER = 0.001f;
-        const float MIN_RELATIVE_SPHERICAL_ERROR = 0.001f;
-        if (diameter > MIN_DIAMETER
-            && fabsf(diameter - entityDimensions.y) / diameter < MIN_RELATIVE_SPHERICAL_ERROR
-            && fabsf(diameter - entityDimensions.z) / diameter < MIN_RELATIVE_SPHERICAL_ERROR) {
-
-            _collisionShapeType = SHAPE_TYPE_SPHERE;
+        case entity::Shape::Quad:
+        case entity::Shape::Cube: {
+            _collisionShapeType = SHAPE_TYPE_BOX;
         }
-        else {
+        break;
+        case entity::Shape::Sphere: {
+
+            float diameter = entityDimensions.x;
+            const float MIN_DIAMETER = 0.001f;
+            const float MIN_RELATIVE_SPHERICAL_ERROR = 0.001f;
+            if (diameter > MIN_DIAMETER
+                && fabsf(diameter - entityDimensions.y) / diameter < MIN_RELATIVE_SPHERICAL_ERROR
+                && fabsf(diameter - entityDimensions.z) / diameter < MIN_RELATIVE_SPHERICAL_ERROR) {
+
+                _collisionShapeType = SHAPE_TYPE_SPHERE;
+            }
+            else {
+                _collisionShapeType = SHAPE_TYPE_ELLIPSOID;
+            }
+        }
+        break;
+        case entity::Shape::Cylinder: {
+            _collisionShapeType = SHAPE_TYPE_CYLINDER_Y;
+            // TODO WL21389: determine if rotation is axis-aligned
+            //const Transform::Quat & rot = _transform.getRotation();
+
+            // TODO WL21389: some way to tell apart SHAPE_TYPE_CYLINDER_Y, _X, _Z based on rotation and
+            //       hull ( or dimensions, need circular cross section)
+            // Should allow for minor variance along axes?
+
+        }
+        break;
+        case entity::Shape::Triangle:
+        case entity::Shape::Hexagon:
+        case entity::Shape::Octagon:
+        case entity::Shape::Circle:
+        case entity::Shape::Tetrahedron:
+        case entity::Shape::Octahedron:
+        case entity::Shape::Dodecahedron:
+        case entity::Shape::Icosahedron:
+        case entity::Shape::Cone: {
+            //TODO WL21389: SHAPE_TYPE_SIMPLE_HULL and pointCollection (later)
+            //_collisionShapeType = SHAPE_TYPE_SIMPLE_HULL;
+        }
+        break;
+        case entity::Shape::Torus:
+        {
+            // Not in GeometryCache::buildShapes, unsupported.
+            _collisionShapeType = SHAPE_TYPE_ELLIPSOID;
+            //TODO WL21389: SHAPE_TYPE_SIMPLE_HULL and pointCollection (later if desired support)
+        }
+        break;
+        default:{
             _collisionShapeType = SHAPE_TYPE_ELLIPSOID;
         }
-    }
-    break;
-    case entity::Shape::Cylinder: {
-        _collisionShapeType = SHAPE_TYPE_CYLINDER_Y;
-        // TODO WL21389: determine if rotation is axis-aligned
-        //const Transform::Quat & rot = _transform.getRotation();
-
-        // TODO WL21389: some way to tell apart SHAPE_TYPE_CYLINDER_Y, _X, _Z based on rotation and
-        //       hull ( or dimensions, need circular cross section)
-        // Should allow for minor variance along axes?
-
-    }
-    break;
-    case entity::Shape::Triangle:
-    case entity::Shape::Hexagon:
-    case entity::Shape::Octagon:
-    case entity::Shape::Circle:
-    case entity::Shape::Tetrahedron:
-    case entity::Shape::Octahedron:
-    case entity::Shape::Dodecahedron:
-    case entity::Shape::Icosahedron:
-    case entity::Shape::Cone: {
-        //TODO WL21389: SHAPE_TYPE_SIMPLE_HULL and pointCollection (later)
-        //_collisionShapeType = SHAPE_TYPE_SIMPLE_HULL;
-    }
-    break;
-    case entity::Shape::Torus:
-    {
-        // Not in GeometryCache::buildShapes, unsupported.
-        _collisionShapeType = SHAPE_TYPE_NONE;
-        //TODO WL21389: SHAPE_TYPE_SIMPLE_HULL and pointCollection (later if desired support)
-    }
-    break;
-    default:{
-        //_collisionShapeType = SHAPE_TYPE_NONE; // Remains SHAPE_TYPE_NONE.
-    }
-    break;
+        break;
     }
 
     EntityItem::computeShapeInfo(info);
+}
+
+// This value specifes how the shape should be treated by physics calculations.
+ShapeType RenderableShapeEntityItem::getShapeType() const {
+    return _collisionShapeType;
 }
 
 void RenderableShapeEntityItem::render(RenderArgs* args) {
--- a/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
@@ -114,26 +114,13 @@ void RenderableShapeEntityItem::computeShapeInfo(ShapeInfo& info) {
     break;
     case entity::Shape::Cylinder: {
         _collisionShapeType = SHAPE_TYPE_CYLINDER_Y;
-        // TODO_CUSACK: determine if rotation is axis-aligned
+        // TODO WL21389: determine if rotation is axis-aligned
         //const Transform::Quat & rot = _transform.getRotation();
 
-#if 0
-        // TODO: some way to tell apart SHAPE_TYPE_CYLINDER_Y
-        // TODO_CUSACK: Should allow for minor variance along axes?
-        if ((entityDimensions.y >= entityDimensions.x) && (entityDimensions.y >= entityDimensions.z)) {
-        }
-        else if (entityDimensions.x >= entityDimensions.z) {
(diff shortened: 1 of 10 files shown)


marko8904:
build this please

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1955/HighFidelity-Beta-PR11048-ef1e426273b3412e63ee511495f8013375250245.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1955/HighFidelity-Beta-PR11048-ef1e426273b3412e63ee511495f8013375250245.dmg

1P-Cusack:
There looks as though there may have been a problem with the build of this PR; however, I don't have access to view the build log.  Could someone take a look and let me know if it's something that I can fix on my end?

sethalves:
it says:  ```

RenderableShapeEntityItem.h:31,           GNU C Compiler 4 (gcc), Priority: Normal
--
‘virtual void RenderableShapeEntityItem::computeShapeInfo(ShapeInfo&)’ can be marked override [-Wsuggest-override]
```

**Comment by hifi-gustavo** [comment]
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/1961/HighFidelity-Beta-PR11048-6b5cabf00af08e4278828da0ac3fceec45a6f717.exe

**Comment by hifi-gustavo** [comment]
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/1961/HighFidelity-Beta-PR11048-6b5cabf00af08e4278828da0ac3fceec45a6f717.dmg

**Comment by Nex-Pro** [comment]
Checking this.

**Comment by Nex-Pro** [comment]
### PASS
Cylinders roll and can stand up.
Also flat cylinders behave like coins. Quite realistic.
Cubes ok at the sizes I tried.
Usual know issues with physics.

@1P-Cusack Nice job.

File: libraries/entities/src/ShapeEntityItem.h [hunk 1]
```diff
@@ -100,6 +100,7 @@ class ShapeEntityItem : public EntityItem {
     float _alpha { 1 };
     rgbColor _color;
     entity::Shape _shape { entity::Shape::Sphere };
+    ShapeType _collisionShapeType { ShapeType::SHAPE_TYPE_NONE };
```
AndrewMeadows: I was looking at these changes thinking, "I don't think we actually need to cache `ShapeEntityItem::_collisionShapeType` as a data member... we should be able to compute it on the fly when we need it."

However, looking at other code outside of the scope of this PR I can see that we're already calling `EntityItem::getShapeType()` in inappropriate places... and we probably don't what to have to constantly recompute this answer.  So I guess this cached value is okay for now.  I'll have to go through the code later and figure out how to remove unnecessary/inappropriate calls to `EntityItem::getShapeType()`.  Once that is done I'll be able to remove this cached value.

libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
```diff
@@ -16,6 +16,7 @@
 #include <StencilMaskPass.h>
 #include <GeometryCache.h>
 #include <PerfStat.h>
+#include <ShapeFactory.h>
```
AndrewMeadows: I would not expect this `#include` to be necessary.  Remove it and verify that it still builds.

How do I know this?  Because by design no `cpp` file needs to know about the `ShapeFactory` except the `ShapeManager` (and certain unit-tests).

libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
```diff
@@ -82,6 +83,75 @@ bool RenderableShapeEntityItem::isTransparent() {
...
+    break;
+    case entity::Shape::Cylinder: {
+        _collisionShapeType = SHAPE_TYPE_CYLINDER_Y;
+        // TODO WL21389: determine if rotation is axis-aligned
+        //const Transform::Quat & rot = _transform.getRotation();
+
+        // TODO WL21389: some way to tell apart SHAPE_TYPE_CYLINDER_Y, _X, _Z based on rotation and
+        //       hull ( or dimensions, need circular cross section)
+        // Should allow for minor variance along axes?
+
```
AndrewMeadows: For all of the as-yet unsupported cases here make `_collisionShapeType = SHAPE_TYPE_ELLIPSOID`.  This will maintain the legacy behavior as was implemented in `ShapeEntityItem::getShapeType()`.  It is theoretically possible that someone somewhere has already created an instance of one of the unfinished shape types and is relying on it to collide like an ellipsoid -- we don't want such an object to suddenly just stop colliding as soon as this PR merges.

libraries/entities/src/ShapeEntityItem.cpp
```diff
@@ -88,12 +88,14 @@ EntityItemProperties ShapeEntityItem::getProperties(EntityPropertyFlags desiredP
...
         case entity::Shape::Cube:
             _type = EntityTypes::Box;
+            _collisionShapeType = ShapeType::SHAPE_TYPE_BOX;
             break;
         case entity::Shape::Sphere:
             _type = EntityTypes::Sphere;
+            _collisionShapeType = ShapeType::SHAPE_TYPE_ELLIPSOID;
             break;
         default:
             _type = EntityTypes::Shape;
```
AndrewMeadows: We will need to recompute the `_collisionShapeType` for this case, otherwise future calls to `getShapeType()` will return incorrect values.

Note, we don't have an `ShapeEntityItem::recomputeCollisionShapeType()` method, but if we did we could call it here.  Tthe only place where `_collisionShapeType` is correctly updated is as a side-effect of `computeShapeInfo()`.  But if you look at the structure of that method... in pseudo code it looks like this:

```
void RenderableShapeEntityItem::computeShapeInfo(ShapeInfo& info) {
    recomputeCollisionShapeType();
    EntityItem::computeShapeInfo(info);
}
```
You get where I'm going right?  Pull the majority of `computeShapeInfo()` into a new method called `recomputeCollisionShapeType()`.  And don't call it inside `computeShapeInfo()`... instead call it here inside `setShape()` -- we will assume that by the time `computeShapeInfo()` is called the `_collisionShapeType` variable has been set correctly.

Finally, this will make it look like the `ShapeEntityItem::computeShapeInfo()` override is no longer necessary, since it just calls the method from the base implementation.  However, my intuition is telling me that ultimately that override WILL be necessary so keep it around and put a comment in there saying something like `// TODO: handle the other shape cases here`.

1P-Cusack: This is a remnant of the prior approach which had <code>ShapeEntityItem</code> handling the shapeType setting and update; the prior approach being the one that snagged on library inclusion issue.  I can see about moving the <code>computeShapeInfo</code> override and its functionality back to <code>ShapeEntityItem</code> from <code>RenderableShapeEntityItem</code> now that the shape class is more explicitly handling the computation rather than trying to keep things contained with <code>ShapeFactory</code>.  I can also likely get rid of the added <code>#include <ShapeFactory.h></code> in <code>RenderableShapeEntityItem</code>  which I believe also still remains at the moment.  The setting of <code>_collisionShapeType</code> should be removed entirely when successfully moving the <code>computeShapeInfo</code> override back to <code>ShapeEntityItem</code>  as it should first be determined when the physics information for the item is set up within <code>PhysicalEntitySimulation</code>.

<code>_collisionShapeType</code> is updated each call to the <code>computeShapeInfo</code> override.  When vetting handling it there, I searched for all calls to the function.  The results seemed to indicate that <code>computeShapeInfo</code> will be triggered when <code>Simulation::DIRTY_SHAPE</code> is set.  <code>Simulation::DIRTY_SHAPE</code> would result in <code>EntityMotionState::computeNewShape</code>'s triggering the call.  It seemed as though when needed that <code>_collisionShapeType</code> would get re-computed taking into account any updated & relevant <code>Entity</code> state.  Is that incorrect?  

I can move the functionality that computes the <code>_collisionShapeType</code> within its own function; however, I elected against that originally given the result of the call reference search for <code>computeShapeInfo</code> and it appeared that if it was encapsulated in its own function it'd always be called in conjunction with that function.  There didn't appear to be a case when they'd be called separately from one another.

libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
1P-Cusack: That's true, thanks for catching that. I'd forgotten to switch that out.

libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
1P-Cusack: I referred to this an a later reply; however, I believe that this is a remnant from the prior approach where some functionality for the <code>ShapeType</code> computation was within <code>ShapeFactory</code>.  I'll remove this.

libraries/shared/src/ShapeInfo.cpp
```diff
@@ -136,7 +144,10 @@ float ShapeInfo::computeVolume() const {
         }
```
1P-Cusack: I'll amend the todo to notate those explicitly.

libraries/entities/src/ShapeEntityItem.h
1P-Cusack: 👍

libraries/entities/src/ShapeEntityItem.cpp
1P-Cusack: Thought of an impediment to moving things back to <code>ShapeEntityItem</code> from <code>RenderableShapeEntityItem</code>, the next phase (PR2) will have the support for the simple hull shapes, and they have a dependency on the <code>GeometryCache</code> for computing their point listings.  <code>GeometryCache</code>, which appears to store the vert info for the shapes, is within the Physics Library so lib inclusion would still be an issue.  This would mean that to avoid the whole library inclusion issues things would need to remain in <code>RenderableShapeEntityItem</code>.

Are there instances of ShapeEntityItem, it's doesn't appear to be said explicitly to be an abstract base class as I've seen others declared to be?  I recall seeing that <code>RenderableShapeEntityItem</code> subverts <code>ShapeEntityItem</code>'s factory function registration.  If that holds and there's no means of explicitly instantiating ShapeEntityItem, then <code>_collisionShapeType</code> can be moved to <code>RenderableShapeEntityItem</code> along with the <code>getShapeType</code> override that's currently within <code>ShapeEntityItem</code>.  Is the vert information needed to generate the point listings for the simple hulls available elsewhere aside from <code>GeometryCache</code>?

AndrewMeadows: Right,` _collisionShapeType` doesn't need to live at the `ShapeEntityItem` level.  Move it to `RenderableShapeEntityItem`.

`ShapeEntityItem` is not pure virtual.  The entity-server will make instances of it, but the server doesn't do physics and doesn't need to know about true collision shape.

Yeah you're right.  The true collision shape type doesn't really need to be known until we need to update it in the physics engine, which is done when `Simulation::DIRTY_SHAPE` is detected in a particular context, and then we call `computeShapeInfo()`.  Ok leave it there.  We're calling `computeShapeInfo()` when we shouldn't but cleaning up that stuff is outside the scope of this project.  I'll probably try to clean that up later.

1P-Cusack: 👍

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2092/HighFidelity-Beta-PR11048-cc4fbc97cdab663701315f823e928e9fd2429646.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2092/HighFidelity-Beta-PR11048-cc4fbc97cdab663701315f823e928e9fd2429646.exe

libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
```diff
@@ -82,6 +82,80 @@ bool RenderableShapeEntityItem::isTransparent() {
...
+        case entity::Shape::Circle:
+        case entity::Shape::Tetrahedron:
+        case entity::Shape::Octahedron:
+        case entity::Shape::Dodecahedron:
+        case entity::Shape::Icosahedron:
+        case entity::Shape::Cone: {
+            //TODO WL21389: SHAPE_TYPE_SIMPLE_HULL and pointCollection (later)
+            //_collisionShapeType = SHAPE_TYPE_SIMPLE_HULL;
+        }
+        break;
```
AndrewMeadows: Although `_collisionShapeType` is initialized to `SHAPE_TYPE_ELLIPSOID` in the ctor, it is possible to set the property to `Box` to get `SHAPE_TYPE_BOX` and then to Tetrahedron and reach the code above... which will incorrectly keep the `SHAPE_TYPE_BOX` setting.

So I think, until they are handled properly, these cases (`Triangle` through `Cone`) should set `_collisionShapeType = SHAPE_TYPE_ELLIPSOID` as does the `Torus` case below.  Maybe just fall through instead of `break`?

Also, not a blocker for this PR but I mention it because I just realized it myself: the `Circle` case should eventually be handled like the `Cylinder`, kinda like the `Quad` case is handled like a `Box`, however in both of these cases we can't honor an arbitrary Y component in the dimensions property.  In other words, what does it mean for a `Quad` shape to have dimensions `{ x: 1, y: 1, z: 1}`?  Isn't one of the dimensions zero by definition of it being a `Quad`?  It's fine that the render pipeline draws it as a flat rectangle, but the code that generates the collision shape in the physics engine will honor that dimension and make the thing collide, unexpectedly, like a cube.

1P-Cusack: Thanks! I missed that one last round 🙂 

Regarding the Quad & Circle, I'll be sure to keep that in mind.  I have seen some odd behavior that I can't get 100% of the time when testing the circle which I noted in my PR2 branch as something to look into.  My first impression was that it was a side effect of being generated as a flat cone within GeometryCache, but it may be that it just needs to be treated as a cylinder rather than a hull.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2156/HighFidelity-Beta-PR11048-0b79809f547c9ec90c3b611a02972feabdaa3584.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2156/HighFidelity-Beta-PR11048-0b79809f547c9ec90c3b611a02972feabdaa3584.exe

AndrewMeadows:
Ok the code changes here are a step forward and look correct.  I added the QA_READY label again.  Some things should be retested:

(1) Make sure cylinders can still be created and that they collide correctly.

(2) Make an `EntityItem` of `type: "Shape"` and `shape: "Quad"` with `dimensions: {x:1, y: 1, z: 1}` and report back how it collides.  Right now it is expected to collide like a 1x1x1 cube even though it looks like a flat 1x0x1 square.  This is ok, we'll fix it later.

(3) Make an `EntityItem` of `type: "Shape"` and `shape: "Octohedron"` and verify that it collides like an ellipsoid, which is not quite correct, but is the same as in hifi/master.  This will be fixed in a future PR.

Nex-Pro:
Looking at this.

Nex-Pro:
1. Cylinder is looking and working well. Very satisfying to watch. **PASS**
We will need to drag @davidkelly onto this one ...  I want coin-shaped cylinders to do that nice sound as they spin down to the surface they are on. :)

2. No quad in the shape type drop-down on tablet or toolbar mode. **FAIL**

I made one in the console and had an invisible but editable entity.
Got this back from it.
{  
   "id":"{8ef9bc5e-5434-4365-a6ff-dcf61501ed80}",
   "type":"Shape",
   "created":"2017-08-01T19:11:42Z",
   "age":221.975341796875,
   "ageAsText":"0 hours 3 minutes 41 seconds",
   "lastEdited":1501614764794626,
   "lastEditedBy":"{36abb9d6-cd28-4200-850f-70f1f9dbbd4e}",
   "position":{  
      "x":18.05433464050293,
      "y":-200.9967498779297,
      "z":16.777950286865234
   },
   "dimensions":{  
      "x":1,
      "y":1,
      "z":1
   },
   "naturalDimensions":{  
      "x":1,
      "y":1,
      "z":1
   },
   "naturalPosition":{  
      "x":0,
      "y":0,
      "z":0
   },
   "rotation":{  
      "x":-0.0000152587890625,
      "y":0.2724498510360718,
      "z":-0.0000152587890625,
      "w":0.9621576070785522
   },
   "velocity":{  
      "x":0,
      "y":0,
      "z":0
   },
   "gravity":{  
      "x":0,
      "y":-10,
      "z":0
   },
   "acceleration":{  
      "x":0,
      "y":0,
      "z":0
   },
   "damping":0.39346998929977417,
   "restitution":0.5,
   "friction":0.5,
   "density":1000,
   "lifetime":-1,
   "script":"",
   "scriptTimestamp":0,
   "serverScripts":"",
   "registrationPoint":{  
      "x":0.5,
      "y":0.5,
      "z":0.5
   },
   "angularVelocity":{  
      "x":0,
      "y":0,
      "z":0
   },
   "angularDamping":0.39346998929977417,
   "visible":1,
   "collisionless":0,
   "ignoreForCollisions":0,
   "collisionMask":31,
   "collidesWith":"static,dynamic,kinematic,myAvatar,otherAvatar,",
   "dynamic":1,
   "collisionsWillMove":1,
   "href":"",
   "description":"",
   "actionData":"",
   "locked":0,
   "userData":"",
   "marketplaceID":"",
   "name":"myQuad",
   "collisionSoundURL":"",
   "color":{  
      "red":20,
      "green":100,
      "blue":200
   },
   "shape":"Quad",
   "boundingBox":{  
      "brn":{  
         "x":17.366413116455078,
         "y":-201.4967803955078,
         "z":16.09002113342285
      },
      "tfl":{  
         "x":18.74225616455078,
         "y":-200.49671936035156,
         "z":17.465879440307617
      },
      "center":{  
         "x":18.05433464050293,
         "y":-200.9967498779297,
         "z":16.777950286865234
      },
      "dimensions":{  
         "x":1.3758430480957031,
         "y":1.00006103515625,
         "z":1.3758583068847656
      }
   },
   "originalTextures":"{\n}\n",
   "parentID":"{00000000-0000-0000-0000-000000000000}",
   "parentJointIndex":65535,
   "queryAACube":{  
      "x":17.188308715820312,
      "y":-201.86277770996094,
      "z":15.911925315856934,
      "scale":1.7320507764816284
   },
   "localPosition":{  
      "x":18.05433464050293,
      "y":-200.9967498779297,
      "z":16.777950286865234
   },
   "localRotation":{  
      "x":-0.0000152587890625,
      "y":0.2724498510360718,
      "z":-0.0000152587890625,
      "w":0.9621576070785522
   },
   "clientOnly":0,
   "owningAvatarID":"{00000000-0000-0000-0000-000000000000}",
   "renderInfo":{  

   }
}

3. Octahedron acts like an ellipsoid. **PASS**

AndrewMeadows:
The question, @Nex-Pro, is whether that behavior is new to this PR or still happens in hifi/master.  I was assuming that people could actually use a "Quad" `ShapeEntityItem`, but it sounds like that is not the case.  I reject the failure of step (2) -- I don't think it blocks this PR.  What about step (3)?

Nex-Pro:
@AndrewMeadows I had already noted step 3 as successful.

So apart from step 2 conflicting with the test notes this is a Pass

Nex-Pro:
### FAIL

Cube, sphere cylinder and circle collide as expected

None of the other entities collide as per their appearance.

Can someone check that edit.js is creating the entities with the correct properties ?

Here's the JSON from a cone.

{  
    "id":"{f0938857-5675-4aa3-a145-ef637fd1b88f}",
    "type":"Shape",
    "created":"2017-09-15T15:54:54Z",
    "age":733.6507568359375,
    "ageAsText":"0 hours 12 minutes 13 seconds",
    "lastEdited":1505491544402902,
    "lastEditedBy":"{87682771-8520-4be5-946e-6818f434eacc}",
    "position":{  
        "x":12.647798538208008,
        "y":-200.14198303222656,
        "z":-29.44781494140625
    },
    "dimensions":{  
        "x":2,
        "y":2,
        "z":2
    },
    "naturalDimensions":{  
        "x":1,
        "y":1,
        "z":1
    },
    "naturalPosition":{  
        "x":0,
        "y":0,
        "z":0
    },
    "rotation":{  
        "x":0.3196917772293091,
        "y":-0.4087129235267639,
        "z":0.15538263320922852,
        "w":0.840604305267334
    },
    "velocity":{  
        "x":0,
        "y":0,
        "z":0
    },
    "gravity":{  
        "x":0,
        "y":-10,
        "z":0
    },
    "acceleration":{  
        "x":0,
        "y":0,
        "z":0
    },
    "damping":0.39346998929977417,
    "restitution":0.5,
    "friction":0.5,
    "density":1000,
    "lifetime":-1,
    "script":"",
    "scriptTimestamp":0,
    "serverScripts":"",
    "registrationPoint":{  
        "x":0.5,
        "y":0.5,
        "z":0.5
    },
    "angularVelocity":{  
        "x":0,
        "y":0,
        "z":0
    },
    "angularDamping":0.39346998929977417,
    "visible":1,
    "collisionless":0,
    "ignoreForCollisions":0,
    "collisionMask":31,
    "collidesWith":"static,dynamic,kinematic,myAvatar,otherAvatar,",
    "dynamic":1,
    "collisionsWillMove":1,
    "href":"",
    "description":"",
    "actionData":"",
    "locked":0,
    "userData":"{\"grabbableKey\":{\"grabbable\":true}}",
    "marketplaceID":"",
    "name":"",
    "collisionSoundURL":"",
    "color":{  
        "red":24,
        "green":125,
        "blue":15
    },
    "shape":"Cone",
    "boundingBox":{  
        "brn":{  
            "x":10.919840812683105,
            "y":-201.55386352539062,
            "z":-31.106252670288086
        },
        "tfl":{  
            "x":14.37575626373291,
            "y":-198.7301025390625,
            "z":-27.789377212524414
        },
        "center":{  
            "x":12.647798538208008,
            "y":-200.14198303222656,
            "z":-29.44781494140625
        },
        "dimensions":{  
            "x":3.4559154510498047,
            "y":2.823760986328125,
            "z":3.316875457763672
        }
    },
    "originalTextures":"{\n}\n",
    "parentID":"{00000000-0000-0000-0000-000000000000}",
    "parentJointIndex":65535,
    "queryAACube":{  
        "x":10.91574764251709,
        "y":-201.87403869628906,
        "z":-31.17986488342285,
        "scale":3.464101552963257
    },
    "localPosition":{  
        "x":12.647798538208008,
        "y":-200.14198303222656,
        "z":-29.44781494140625
    },
    "localRotation":{  
        "x":0.3196917772293091,
        "y":-0.4087129235267639,
        "z":0.15538263320922852,
        "w":0.840604305267334
    },
    "clientOnly":0,
    "owningAvatarID":"{00000000-0000-0000-0000-000000000000}",
    "renderInfo":{  

    }
}

PR #11080, opened 2017-07-29
Fix web touch emulation

jherico:
The touch events have been leading to occasional crashes.  This is particularly easy to replicate by using multiple hands on the tablet at once.  Running in debug mode this rapidly leads to assertion errors inside the QML code.  I believe the problem is that we're not sending touch events the way actual hardware would and as such we're breaking the internal state of the QML classes, which have some fairly complicated internal logic on how to process touch events (examine [qquickwindow.cpp](http://code.qt.io/cgit/qt/qtdeclarative.git/tree/src/quick/items/qquickwindow.cpp?h=5.9) and see all the instances of `touchMouseId` for example).

Fortunately I have a laptop with a multi-touch screen and was able to create a small test app that reported incoming touch events.  Armed with the documentation for touch processing as well as some empirical data, I determined how we should behave... 

The rules for touch events are:

 * When a first touch point is detected, send a `TouchBegin` event.  If this `TouchBegin` event is accepted, further `TouchUpdate` events and eventually a `TouchEnd` event will be sent.  If `TouchBegin` is _not_ accepted, no further events will be sent until the next `TouchBegin`
* If a second touch point is added in between an accepted `TouchBegin` and `TouchEnd`, it arrives as a `TouchUpdate` where the new touch point has the flag `TouchPointPressed`.  
* Similarly, if we release a touch point between `TouchBegin` and `TouchEnd` and it is NOT the last touch point, it arrives as a `TouchUpdate` where the leaving touch point has the flag `TouchPointReleased`, and subsequent `TouchUpdate` and `TouchEnd` events do not include this point unless it is re-added as in the previous bullet point

The old code sent `TouchBegin` and `TouchEnd` events strictly based on the pressed / released state of the incoming pointer event and without regard to whether an existing `TouchBegin` had already been sent.  This meant that you could have got interleaved or nested `TouchBegin`/`TouchEnd` messages from the left and right hands, depending on the order of operations.  Since both hands were reporting the same `TouchDevice` this likely violated the contract for how touch events would be delivered.  Additionally the old code did not respect the promise that a non-accepted `TouchBegin` message would result in no further messages until the next `TouchBegin`.

The new code instead tracks the active touch points, and send a `TouchBegin` when a first point becomes active, and a `TouchEnd` when the last point ceases to be active.  All other messages are sent as `TouchUpdate`.  The code also tracks whether a given `TouchBegin` was accepted and only sends the `TouchUpdate` and `TouchEnd` if it was.  It still tracks the state of all the points, so it knows when to send the next `TouchBegin`

## Testing

This is a well scoped change (it only applies to the Web3DOverlay type) but is a serious change in logic, with significant new amount of internal state tracking in order to try to emulate the expected behavior of touch events.  

Care should be taken to thoroughly test the tablet functionality with hand controllers, as well as with mouse and eye tracking interaction.

### DIFF
--- a/interface/src/ui/overlays/Web3DOverlay.cpp
+++ b/interface/src/ui/overlays/Web3DOverlay.cpp
@@ -248,7 +248,7 @@ void Web3DOverlay::render(RenderArgs* args) {
             if (!self) {
                 return;
             }
-            if (self->_pressed && overlayID == selfOverlayID) {
+            if (overlayID == selfOverlayID && (self->_pressed || (!self->_activeTouchPoints.empty() && self->_touchBeginAccepted))) {
                 PointerEvent endEvent(PointerEvent::Release, event.getID(), event.getPos2D(), event.getPos3D(), event.getNormal(), event.getDirection(),
                                       event.getButton(), event.getButtons(), event.getKeyboardModifiers());
                 forwardPointerEvent(overlayID, event);
@@ -337,19 +337,11 @@ void Web3DOverlay::setProxyWindow(QWindow* proxyWindow) {
 }
 
 void Web3DOverlay::handlePointerEvent(const PointerEvent& event) {
-    // FIXME touch event emulation is broken in some way.  Do NOT enable this code
-    // unless you have done a debug build of the application and verified that 
-    // you are not getting assertion errors on handling the touch events inside
-    // Qt.
-#if 0
     if (_inputMode == Touch) {
         handlePointerEventAsTouch(event);
     } else {
         handlePointerEventAsMouse(event);
     }
-#else
-    handlePointerEventAsMouse(event);
-#endif
 }
 
 void Web3DOverlay::handlePointerEventAsTouch(const PointerEvent& event) {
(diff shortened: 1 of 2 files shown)


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2104/HighFidelity-Beta-PR11080-0dfec816f889b9dfcfc6e99e2d5da8fb8d547ee6.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2104/HighFidelity-Beta-PR11080-0dfec816f889b9dfcfc6e99e2d5da8fb8d547ee6.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2110/HighFidelity-Beta-PR11080-0dfec816f889b9dfcfc6e99e2d5da8fb8d547ee6.dmg

jherico:
test this please

interface/src/ui/overlays/Web3DOverlay.cpp
```diff
@@ -349,106 +349,98 @@ void Web3DOverlay::handlePointerEventAsTouch(const PointerEvent& event) {
...
-    QPointF windowPoint(windowPos.x, windowPos.y);
-
-    if (event.getType() == PointerEvent::Press && event.getButton() == PointerEvent::PrimaryButton) {
-        this->_pressed = true;
-    } else if (event.getType() == PointerEvent::Release && event.getButton() == PointerEvent::PrimaryButton) {
-        this->_pressed = false;
+    //do not send secondary button events to tablet
+    if (event.getButton() == PointerEvent::SecondaryButton ||
+        //do not block composed events
+        event.getButtons() == PointerEvent::SecondaryButton) {
```
ZappoMan: this appears to be saying A || A?
Is this a typo?

jherico: `getButton` vs `getButtons`

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2113/HighFidelity-Beta-PR11080-7d18d2e01788e8c7d07cd204edad031f37493c55.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2113/HighFidelity-Beta-PR11080-7d18d2e01788e8c7d07cd204edad031f37493c55.exe

interface/src/ui/overlays/Web3DOverlay.cpp
ZappoMan: ok

PR #11121, opened 2017-08-05
Adding dummy file showing contents of build directory

dkuokka:
Added dummy file showing contents of build directory


hifi-gustavo:

PR #11124, opened 2017-08-05
Zone local physics

sethalves:
- allow a physics simulation per zone, enabling vehicles, elevators, moving sidewalks, etc


sethalves:
to test:
- create a small zone
- make a box, move it to the floor of the zone, and make the box a child of the zone
- find "localized simulation" checkbox for the zone
- experiment with walking into the zone with your avatar
- better test soon

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2294/HighFidelity-Beta-PR11124-2e5f5b6680c47d4e9ff419e18074458def193e40.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2294/HighFidelity-Beta-PR11124-2e5f5b6680c47d4e9ff419e18074458def193e40.exe

Nex-Pro:
Testing this.

Nex-Pro:
In general this works smoothly from POV of avatar being moved and observers both in HMD and desktop. ( tested with two geographically separated clients. Away from the zone everything checks out ok.

Has some issues with regard to user experience:

An entity created within the zone becomes a child of that zone.

If a child entity of that zone is moved far enough away it becomes orphaned but does not become a child again if returned to the zone. ( volume swept by zone ? )

Noticed that the ghosting and flying options appear twice in the zone props and in each pair the settings are not in agreement.

Messes with the spectator cam in interesting but not very useful ways.

Fingerpaint acts in a bizarre manner while travelling on the elevator. Drawing a stroke side to side while going up moves up faster than the avatar is moving up. Opposite happens when moving down.
Very counter-intuitive.

Dynamic objects moved from another zone and dropped do not interact with the elevator platform.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2337/HighFidelity-Beta-PR11124-15d99e08a639855f8ef1c80b7bb2180e2227ad26.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2337/HighFidelity-Beta-PR11124-15d99e08a639855f8ef1c80b7bb2180e2227ad26.exe

interface/src/Application.cpp
```diff
@@ -4919,81 +4934,145 @@ void Application::update(float deltaTime) {
...
+                foreach (ObjectMotionState* motionState, motionStates) {
+                    EntityMotionState* entityMotionState = dynamic_cast<EntityMotionState*>(motionState);
+                    if (entityMotionState) {
+                        EntityItemPointer entity = entityMotionState->getEntity();
+                        auto oldPhysicsEngine = motionState->getPhysicsEngine();
+                        auto newPhysicsEngine = motionState->getShouldBeInPhysicsEngine();
+                        if (oldPhysicsEngine && newPhysicsEngine) {
+                            oldPhysicsEngine->removeObject(motionState);
+                            newPhysicsEngine->addObject(motionState);
+                        }
```
AndrewMeadows: There a possibility that `oldPhysicsEngine == newPhysicsEngine` is `true` in this context, right?  When I follow the logic back it looks to me like this is the case.  Perhaps we should avoid unnecessary remove+add from/to the identical `PhysicsEngine` instance?

interface/src/avatar/AvatarManager.cpp
```diff
@@ -413,6 +421,7 @@ void AvatarManager::handleCollisionEvents(const CollisionEvents& collisionEvents
             auto collisionSound = myAvatar->getCollisionSound();
             if (collisionSound) {
                 const auto characterController = myAvatar->getCharacterController();
+                // TODO -- should this convert local velocity to world?
```
AndrewMeadows: `PhysicsEngine` has a data member called `_originOffset`.  My plan was to use that to allow a limited physics simulation (only simulates nearby stuff) to remain stable when trying to simulate a collection of objects far from the origin (say, a 20x20x20m simulation out near <1000, 1000, 1000>).  That was the plan, however AFAIK we don't every actually set this to a non-zero offset, so it is largely untested (for a couple years now).  In any case, there is not yet a way to give `PhysicsEngine` an arbitrary transform with rotation (and scale?!), which means the "velocity" of any `RigidBody` in a `PhysicsEngine` simulation will always be in the world-frame.  Dunno if that answers the question in the comment.

libraries/avatars/src/AvatarData.h
```diff
@@ -631,6 +634,8 @@ class AvatarData : public QObject, public SpatiallyNestable {
         return _lastSentJointData;
     }
 
+    virtual void handleZoneChange() {};
```
AndrewMeadows: I don't think this should live in `AvatarData` class.  Should at least be moved up to the `Avatar` class.

libraries/avatars/src/AvatarData.h
```diff
@@ -689,6 +694,7 @@ public slots:
     bool avatarScaleChangedSince(quint64 time) const { return _avatarScaleChanged >= time; }
     bool lookAtPositionChangedSince(quint64 time) const { return _headData->lookAtPositionChangedSince(time); }
     bool sensorToWorldMatrixChangedSince(quint64 time) const { return _sensorToWorldMatrixChanged >= time; }
+    bool sensorToSimulationMatrixChangedSince(quint64 time) const { return _sensorToSimulationMatrixChanged >= time; }
```
AndrewMeadows: `sensorToSimulationMatrixChangedSince()` appears to have been added but is never called AFAICT.  Similarly for the data member that it changes.  If it happens to be useful despite my research please evaluate whether it really belongs down in the `AvatarData` class -- surely it belongs in one of the derived classes.

libraries/entities/src/EntityItem.cpp
```diff
@@ -103,6 +104,15 @@ EntityItem::~EntityItem() {
     assert(!_physicsInfo);
 }
 
+QString EntityItem::toString() const {
+    QString name = getName();
+    if (name == "") {
+        return getID().toString();
+    }
+    return name + ":" + getID().toString();
+}
```
AndrewMeadows: This, and the method it overrides, have been added but are never actually called. Perhaps this is a temporary debug method that should be removed?  Or is it a debug method that is useful enough to keep around?

libraries/physics/src/PhysicsEngine.cpp
```diff
@@ -223,10 +225,18 @@ void PhysicsEngine::addObjects(const VectorOfMotionStates& objects) {
...
+    addObjectToDynamicsWorld(object);
+}
+
 VectorOfMotionStates PhysicsEngine::changeObjects(const VectorOfMotionStates& objects) {
     VectorOfMotionStates stillNeedChange;
     for (auto object : objects) {
         uint32_t flags = object->getIncomingDirtyFlags() & DIRTY_PHYSICS_FLAGS;
+        if (flags & Simulation::DIRTY_HIERARCHY) {
+            object->maybeSwitchPhysicsEngines();
+        }
```
AndrewMeadows: Something is funny here... 

`PhysicsEngine::changeObjects()` is being called in a specific `PhysicsEngine` instance.  To the naive eye it looks like the object may be moving to a completely different `PhysicsEngine` instance.  But then we continue to handle the other flags... in this context?  It doesn't look right to me, but I don't have a good idea of how it **should** work.

libraries/physics/src/ObjectMotionState.cpp
```diff
@@ -38,16 +38,20 @@ const glm::vec3& ObjectMotionState::getWorldOffset() {
...
-// static
-uint32_t worldSimulationStep = 0;
 void ObjectMotionState::setWorldSimulationStep(uint32_t step) {
-    assert(step > worldSimulationStep);
-    worldSimulationStep = step;
+    PhysicsEnginePointer physicsEngine = getPhysicsEngine();
+    if (!physicsEngine) {
+        return;
+    }
+    physicsEngine->setWorldSimulationStep(step);
```
AndrewMeadows: Now that I look at this more carefully... it looks like you are allowing the `simulationStep` to be set on a `MotionState` instance, which under the hood reaches up and sets the `simulationStep` on the `PhysicsEngine` in which it is simulated.  This is unexpected API.

My memory tells me: the original reason I made the `simulationStep` as a static in the `ObjectMotionState` class was because I needed it in a context (I forget exactly where) in which I felt it was too expensive to justify walking all the way up to the `PhysicsEngine` to get it -- too many CPU cycles and complexity.  So I was "caching" the value in a single spot where all `ObjectMotionStates` could get to it.  Now that each `ObjectMotionState` keeps a back pointer to its simulation... we probably don't need this `setWorldSimulationStep()` call.  Comment this method out and let the compiler tell you if is is superfluous or not, and if it is **not** then the code that uses it needs to be scrutinized for sanity.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2360/HighFidelity-Beta-PR11124-4a8c6d5268174986e00499f2a4cc24f84bb4accb.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2360/HighFidelity-Beta-PR11124-4a8c6d5268174986e00499f2a4cc24f84bb4accb.exe

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1516,27 +1559,57 @@ void MyAvatar::updateMotors() {
...
-    glm::vec3 parentVelocity = getParentVelocity(success);
-    if (!success) {
-        qDebug() << "Warning: getParentVelocity failed" << getID();
-        parentVelocity = glm::vec3();
+    // attempt to keep the avatar stuck to its parent by matching velocities
+    QUuid parentID = getParentID();
+    glm::vec3 parentInducedVelocity;
+    glm::vec3 parentInducedAngularVelocity;
+    if (parentID != QUuid() && parentID != physicsEngine->getID()) {
+        bool success;
```
sethalves: this needs to be fixed

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1516,27 +1559,57 @@ void MyAvatar::updateMotors() {
...
+                glm::quat deltaRot = parentRotation * glm::inverse(_previousParentRotation);
+                glm::vec3 deltaRotEu = glm::eulerAngles(deltaRot);
+                deltaRotEu.x = 0.0f;
+                deltaRotEu.z = 0.0f;
+                glm::quat newRot = getOrientation() * glm::quat(deltaRotEu);
+                setOrientation(newRot);
+            }
+            _previousParentRotation = parentRotation;
+            _previousParentLocationSet = true;
+        }
```
hyperlogic: parentInducedVelocity should be computed as followed.

    parentInducedVelocity = getParentVelocity() + glm::cross(getParentAngularVelocity(), childPosition - parentPosition);

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1516,27 +1559,57 @@ void MyAvatar::updateMotors() {
...
+            glm::vec3 childPosition = getPosition();
+            parentInducedVelocity = parentVelocity + qParentAngularVelocity * (childPosition - parentPosition);
+
+            if (_previousParentLocationSet) {
+                glm::quat deltaRot = parentRotation * glm::inverse(_previousParentRotation);
+                glm::vec3 deltaRotEu = glm::eulerAngles(deltaRot);
+                deltaRotEu.x = 0.0f;
+                deltaRotEu.z = 0.0f;
+                glm::quat newRot = getOrientation() * glm::quat(deltaRotEu);
+                setOrientation(newRot);
```
hyperlogic: Is this code here to rotate the avatar as the parent platform rotates?

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2417/HighFidelity-Beta-PR11124-0af0ba892346adc7661555c6e9d6b4b60deae28b.dmg

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2433/HighFidelity-Beta-PR11124-994bd30c3573d20b4c8180b6ee0ef29c237d9321.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2433/HighFidelity-Beta-PR11124-994bd30c3573d20b4c8180b6ee0ef29c237d9321.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2495/HighFidelity-Beta-PR11124-fd4c7d1c7c24d14f9a7568857e6f3a1e693d2c2a.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2495/HighFidelity-Beta-PR11124-fd4c7d1c7c24d14f9a7568857e6f3a1e693d2c2a.exe

ShoaibSmith:
POV on HMD and desktop seem to be working properly.  

The elevator platform has no collision for me. When I step onto it, it passes through me and the world. 

The lighting in the world flickers when the elevator platform goes up. 



Checked to see if the problems nex-pro had were still occurring:

>>An entity created within the zone becomes a child of that zone.

>> If a child entity of that zone is moved far enough away it becomes orphaned but does not become a child again if returned to the zone. ( volume swept by zone ? )

I will look further into these tomorrow.

>>Noticed that the ghosting and flying options appear twice in the zone props and in each pair the settings are not in agreement.

This issue appears to be still broken. Ghosting and flying options appear only once for me but the settings are not saving properly.

>>Fingerpaint acts in a bizarre manner while travelling on the elevator. Drawing a stroke side to side while going up moves up faster than the avatar is moving up. Opposite happens when moving down.

>>Very counter-intuitive.

>>Dynamic objects moved from another zone and dropped do not interact with the elevator platform.

Was unable to test these thoroughly because of the elevator collision issue noted above.

sethalves:
@ShoaibSmith -- which avatar were you wearing when you were testing?

ShoaibSmith:
@sethalves I was using the Default Avatar, Wooden Mannequin.

tapalisa:
Someone who ran this PR does not have their environment variable tester flag set. Please check your settings.

anshuman64:
Added a ticket to continue as future work: https://highfidelity.fogbugz.com/f/cases/3948/Zone-Specific-Physics

PR #11130, opened 2017-08-07
Fix crash in Context Overlays code

zfox23:
Fixes [FB6798](https://highfidelity.fogbugz.com/f/cases/6798). I was able to reproduce this by spawning the xylophone from the marketplace, opening the Create app, clicking the xylophone, deleting it, then moving my mouse.

I also added a readlock on the entity tree inside `enableEntityHighlight()` and `disableEntityHighlight()` because everywhere else we do `findEntityByEntityItemID()` we have a readlock. It makes sense to do that.

Easy test plan here.

**Test Plan**
1. After installing Interface, open `Interface.json` and change `inspectionMode` to `true`.
2. Spawn something from the marketplace in some domain.
3. Hover over the entity with your mouse and verify that the yellow bounding box appears around the entity.
4. Hover off of the entity and verify that the yellow bounding box disappears.
5. Open the Create app. Hover over the entity. Left-click the entity. Press "Delete" on your keyboard. Verify that the entity disappears.
6. Move your mouse. Verify that you don't crash.

### DIFF
--- a/interface/src/ui/overlays/ContextOverlayInterface.cpp
+++ b/interface/src/ui/overlays/ContextOverlayInterface.cpp
@@ -257,15 +257,23 @@ void ContextOverlayInterface::openMarketplace() {
 }
 
 void ContextOverlayInterface::enableEntityHighlight(const EntityItemID& entityItemID) {
-    if (!qApp->getEntities()->getTree()->findEntityByEntityItemID(entityItemID)->getShouldHighlight()) {
-        qCDebug(context_overlay) << "Setting 'shouldHighlight' to 'true' for Entity ID:" << entityItemID;
-        qApp->getEntities()->getTree()->findEntityByEntityItemID(entityItemID)->setShouldHighlight(true);
-    }
+    auto entityTree = qApp->getEntities()->getTree();
+    entityTree->withReadLock([&] {
+        auto entityItem = entityTree->findEntityByEntityItemID(entityItemID);
+        if ((entityItem != NULL) && !entityItem->getShouldHighlight()) {
+            qCDebug(context_overlay) << "Setting 'shouldHighlight' to 'true' for Entity ID:" << entityItemID;
+            entityItem->setShouldHighlight(true);
+        }
+    });
 }
 
 void ContextOverlayInterface::disableEntityHighlight(const EntityItemID& entityItemID) {
-    if (qApp->getEntities()->getTree()->findEntityByEntityItemID(entityItemID)->getShouldHighlight()) {
-        qCDebug(context_overlay) << "Setting 'shouldHighlight' to 'false' for Entity ID:" << entityItemID;
-        qApp->getEntities()->getTree()->findEntityByEntityItemID(entityItemID)->setShouldHighlight(false);
-    }
+    auto entityTree = qApp->getEntities()->getTree();
+    entityTree->withReadLock([&] {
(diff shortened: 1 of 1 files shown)


interface/src/ui/overlays/ContextOverlayInterface.cpp
```diff
@@ -257,15 +257,23 @@ void ContextOverlayInterface::openMarketplace() {
 }
 
 void ContextOverlayInterface::enableEntityHighlight(const EntityItemID& entityItemID) {
-    if (!qApp->getEntities()->getTree()->findEntityByEntityItemID(entityItemID)->getShouldHighlight()) {
-        qCDebug(context_overlay) << "Setting 'shouldHighlight' to 'true' for Entity ID:" << entityItemID;
-        qApp->getEntities()->getTree()->findEntityByEntityItemID(entityItemID)->setShouldHighlight(true);
-    }
+    auto entityTree = qApp->getEntities()->getTree();
+    entityTree->withReadLock([&] {
```
davidkelly: if we are writing to the entity, do we need a write lock?

interface/src/ui/overlays/ContextOverlayInterface.cpp
```diff
@@ -257,15 +257,23 @@ void ContextOverlayInterface::openMarketplace() {
...
+    });
 }
 
 void ContextOverlayInterface::disableEntityHighlight(const EntityItemID& entityItemID) {
-    if (qApp->getEntities()->getTree()->findEntityByEntityItemID(entityItemID)->getShouldHighlight()) {
-        qCDebug(context_overlay) << "Setting 'shouldHighlight' to 'false' for Entity ID:" << entityItemID;
-        qApp->getEntities()->getTree()->findEntityByEntityItemID(entityItemID)->setShouldHighlight(false);
-    }
+    auto entityTree = qApp->getEntities()->getTree();
+    entityTree->withReadLock([&] {
```
davidkelly: same question as above

interface/src/ui/overlays/ContextOverlayInterface.cpp
davidkelly: actually I guess it is ok as-is -- we have appropriate locking on the entity when we call the get/set below.  And we are not deleting or adding to the _tree_, so a read lock seems corect

interface/src/ui/overlays/ContextOverlayInterface.cpp
davidkelly: with the same answer as above

OmegaHeron:
Will test.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2309/HighFidelity-Beta-PR11130-07a4beac73ed8c789148816a52ed7d5ee00a5d02.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2309/HighFidelity-Beta-PR11130-07a4beac73ed8c789148816a52ed7d5ee00a5d02.exe

OmegaHeron:
Tested Windows 10
QA Pass

With master 6979 I reproduced crash 100% this PR aims to fix.
This PR - no crash despite repeated attempts.
General smoke test looks good.

PR #11155, opened 2017-08-10
Fix bad plugins causing instant crash on startup

jherico:
Old installations of Interface that contained the `hifiSixense.dll` plugin and were built with Qt 5.6 will now crash when you try to start them from a new 5.9 based build.  The current installer doesn't properly remove all the previous version's files and so the `hifiSixense.dll` is left behind in the plugins directory.  When Qt tries to load it the the QPlugin loader, it crashes because of the Qt version mismatch.  

## Testing

This can't be tested as a PR because the fix is to remove pre-existing files in the install folder.  The PR installer will install to a new directory.


birarda:
If you manually send a tester the DLL and tell them where to put it they could test:
(1) that Interface crashes once they drop that file in
(1) that it is removed on a second install
(2) that they are able to launch Interface after the second install

marko8904:
build this please

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2394/HighFidelity-Beta-PR11155-0452bcfbd5a8c387d8947d34499305bd7ded9092.exe

ZappoMan:
You can use this version of the sixsense DLL to test this.
[hifiSixense.zip](https://github.com/highfidelity/hifi/files/1215718/hifiSixense.zip)

To test:
1) install a recent dev-download build of HiFi
2) Unzip that file, you'll get a hifiSixense.dll, place that DLL in the "plugins" directory for your installation of hiFi...
3) Run interface --- verify you get an "instant crash" and interface doesn't startup.
4) Run this PR's installer and install in the same place as your install from step 1
5) Run interface --- it *should* run properly

cozza13:
tested the flow above and it does not crash

conklin94122:
Tested the following with zero issues:

1) Clean uninstall
2) Install 6895 from/download, using Express settings
3) Confirm interface works
4) Install 7008 using Express settings
5) Confirm interface works
6) Quit and restart interface
7) Confirm interface works

PR #11167, opened 2017-08-10
Protect against oob _uniformMappings in GLShader::getUniformLocation.

1P-Cusack:
Spawning and rotating a new object in the sandbox, I was occasionally tripping over cases where _uniformMappings was empty.  Since there already is a (probably more expensive) check against missing mappings, added another bounds check here.

Note this doesn't account for possible negative values, could do a (size_t) cast on the lhs of the compare (version) if desired.

(If this is the wrong approach, that's fine; holler if you'd prefer an Issue opened or similar instead.  I can likely refine repro steps and figure out something closer to root cause if desired.)

### DIFF
--- a/libraries/gpu-gl/src/gpu/gl/GLShader.h
+++ b/libraries/gpu-gl/src/gpu/gl/GLShader.h
@@ -48,7 +48,10 @@ public:
     }
 
     GLint getUniformLocation(GLint srcLoc, Version version = Mono) const {
-        // This check protect against potential invalid src location for this shader, if unknown then return -1.
+        // These checks protect against potential invalid src location for this shader, if unknown then return -1.
+        if (version >= _uniformMappings.size()) {
+            return -1;
+        }
         const auto& mapping = _uniformMappings[version];
         auto found = mapping.find(srcLoc);
         if (found == mapping.end()) {


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2425/HighFidelity-Beta-PR11167-61069bd2a0baa405f4dafb899a4872d9dec6d43c.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2425/HighFidelity-Beta-PR11167-61069bd2a0baa405f4dafb899a4872d9dec6d43c.exe

samcake:
HI 1P, i m curious, which object triggered that case ?
In theory, it shouldn't happen
The _uniformMappings should always be the correct size

Let me know, 
THanks

1P-Cusack:
@samcake Having trouble reproing it now, but it was just creating a cube, and attempting to rotate it.  I'll spend a little more time later trying to recreate the exact circumstances (I've rebased since then, can probably go back).  It *might* have been while things were still loading in, before lighting fixed itself, but I'm not sure; wasn't able to get it to happen by trying to do things quickly during load just now.  =(

samcake:
@1P-Cusack  THanks
If you could repro the crash and grab the log we could hopefully find some log of the shader failing to compile (normally....)
Because i double checked and the _uniformMappings must be the correct size if the shader is compiled.
THanks for digging this

Leander1P:
I haven't seen this again or been able to repro it; possibly it was a temporary version mismatch or similar?

At any rate, feel free to close if you so desire, I'll reopen with more details if I ever see it again.  Apologies for the delay.

PR #11197, opened 2017-08-16
Rig: Use a registry to prevent use after free crashes/corruption

hyperlogic:
Create a global registry to hold all the currently active Rig instances.  Use this registry and it's mutex to prevent accessing the rig after it has already been destroyed, or is in the process of being destroyed on the main thread.

### DIFF
--- a/libraries/animation/src/Rig.cpp
+++ b/libraries/animation/src/Rig.cpp
@@ -32,6 +32,10 @@
 #include "AnimUtil.h"
 #include "IKTarget.h"
 
+static int nextRigId = 1;
+static std::map<int, Rig*> rigRegistry;
+static std::mutex rigRegistryMutex;
+
 static bool isEqual(const glm::vec3& u, const glm::vec3& v) {
     const float EPSILON = 0.0001f;
     return glm::length(u - v) / glm::length(u) <= EPSILON;
@@ -48,6 +52,26 @@ static bool isEqual(const glm::quat& p, const glm::quat& q) {
 const glm::vec3 DEFAULT_RIGHT_EYE_POS(-0.3f, 0.9f, 0.0f);
 const glm::vec3 DEFAULT_LEFT_EYE_POS(0.3f, 0.9f, 0.0f);
 const glm::vec3 DEFAULT_HEAD_POS(0.0f, 0.75f, 0.0f);
+
+Rig::Rig() {
+    // Ensure thread-safe access to the rigRegistry.
+    std::lock_guard<std::mutex> guard(rigRegistryMutex);
+
+    // Insert this newly allocated rig into the rig registry
+    _rigId = nextRigId;
+    rigRegistry[_rigId] = this;
+    nextRigId++;
+}
+
+Rig::~Rig() {
+    // Ensure thread-safe access to the rigRegstry, but also prevent the rig from being deleted
+    // while Rig::animationStateHandlerResult is being invoked on a script thread.
+    std::lock_guard<std::mutex> guard(rigRegistryMutex);
+    auto iter = rigRegistry.find(_rigId);
+    if (iter != rigRegistry.end()) {
+        rigRegistry.erase(iter);
+    }
+}
 
 void Rig::overrideAnimation(const QString& url, float fps, bool loop, float firstFrame, float lastFrame) {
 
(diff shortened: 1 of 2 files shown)


hyperlogic:
https://highfidelity.fogbugz.com/f/cases/6789/possible-crash-in-Rig-updateAnimationStateHandlers-related-to-Rig-this-in-invokeMethod

hyperlogic:
QA Test:
  * I don't have a reliable way to reproduce this, but perhaps switching avatars frequently, and restarting scripts, might reproduce it.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2524/HighFidelity-Beta-PR11197-aef19c6f9794da3c7be83553812ee88b3790109e.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2524/HighFidelity-Beta-PR11197-aef19c6f9794da3c7be83553812ee88b3790109e.exe

anshuman64:
I gave this a go and it seems to be fine. Tried switching avatars, loading/unloading/reloading scripts. Also did a general smoke test. Would appreciate if one other person tried it out before marking QA Pass

YuppyPlays:
Will have a look now.

Testing.

YuppyPlays:
Testing complete:

Given the unknown way of triggering the crash, here are the following steps I took to see if there were any crashing, all come back clear but, just to cover my ground:
 - Changed back and fourth from HMD to Desktop (Ctrl+1 / Ctrl+4)
 - Changed domains back and fourth
 - Changed avatars repeatedly with different variations (Matthew, Mannequin, Priscilla, etc)
 - Changed avatar size rapidly repeatedly
 - Loaded/Deleted/Reloaded scripts of various kinds
 - General smoke test

No deadlocks/crashes experienced. I'd call this a **Pass** given the steps taken, though without an exact repro instruction it is of course hard to judge.

libraries/animation/src/Rig.cpp
```diff
@@ -933,8 +957,19 @@ void Rig::updateAnimationStateHandlers() { // called on avatar update thread (wh
         int identifier = data.key();
         StateHandler& value = data.value();
         QScriptValue& function = value.function;
-        auto handleResult = [this, identifier](QScriptValue result) { // called in script thread to get the result back to us.
-            animationStateHandlerResult(identifier, result);
+        int rigId = _rigId;
+        auto handleResult = [rigId, identifier](QScriptValue result) { // called in script thread to get the result back to us.
```
ZappoMan: since you're capturing by value, why not just pass _rigId here?

libraries/animation/src/Rig.cpp
```diff
@@ -49,6 +53,26 @@ const glm::vec3 DEFAULT_RIGHT_EYE_POS(-0.3f, 0.9f, 0.0f);
 const glm::vec3 DEFAULT_LEFT_EYE_POS(0.3f, 0.9f, 0.0f);
 const glm::vec3 DEFAULT_HEAD_POS(0.0f, 0.75f, 0.0f);
 
+Rig::Rig() {
+    // Ensure thread-safe access to the rigRegistry.
+    std::lock_guard<std::mutex> guard(rigRegistryMutex);
+
+    // Insert this newly allocated rig into the rig registry
```
ZappoMan: this whole approach feels sketchy-- why not use smart pointers to rig's instead?

libraries/animation/src/Rig.cpp
hyperlogic: I wanted it to be clear to the reader/maintainer of this code that this lambda does not capture this.  Passing a value starting with an underscore might be misleading.

libraries/animation/src/Rig.cpp
hyperlogic: I agree, I would have preferred to use smart ptrs, but the problem is that the Rig class is a embedded member of the satanic model class, not held onto by pointer.  Also the animation library does not have access to any of the Model libraries for it to have a back pointer.  I could have changed Model, but that would be undoing some of the refactoring work andrew had done, and it would have touched many many files.

PR #11204, opened 2017-08-17
Base requirements change : dev system must have at least 8GB of RAM.

EagleJumpDev:
Issue : systems with <8GB of RAM (in this test case, 6GB) will encounter out-of-memory errors while building the HiFi source.

My solution : maybe systems with >=8GB of RAM would not have this issue?

Images attached showing this issue. Issue detected on a system built using the Windows7+VisualStudio2017 specs and required libraries (all from a fresh install) noted in hifi/BUILD_WIN.md

outofmemory
outofmemory2
outofmemory3


hifi-gustavo:

PR #11210, opened 2017-08-18
split packet construction from packet sending in OctreeSendThread

AndrewMeadows:


libraries/octree/src/Octree.cpp
```diff
@@ -925,8 +925,8 @@ int Octree::encodeTreeBitstream(const OctreeElementPointer& element,
         roomForOctalCode = packetData->startSubTree(newCode);
 
         if (newCode) {
-            delete[] newCode;
             codeLength = numberOfThreeBitSectionsInCode(newCode);
+            delete[] newCode;
```
AndrewMeadows: Delete the data **AFTER** using it.  The reason this wasn't causing problems is because `params.chopLevels` is always false these days.

assignment-client/src/octree/OctreeServer.cpp
```diff
@@ -134,9 +132,8 @@ void OctreeServer::trackEncodeTime(float time) {
     const float MAX_SHORT_TIME = 10.0f;
     const float MAX_LONG_TIME = 100.0f;
 
-    if (time == SKIP_TIME) {
+    if (time == 0.0f) {
         _noEncode++;
-        time = 0.0f;
```
AndrewMeadows: Using `0.0` for this "skip time" functionality makes it easier to use the `OctreeServer` time stats API without losing any ambiguity as to what a "skip" means.

assignment-client/src/octree/OctreeSendThread.cpp
```diff
@@ -458,84 +458,77 @@ int OctreeSendThread::packetDistributor(SharedNodePointer node, OctreeQueryNode*
...
+                nodeData->stats.encodeStopped();
+
+                somethingToSend = true;
+            }
+        });
+
+        OctreeServer::trackEncodeTime((float)(usecTimestampNow() - encodeStart));
+    }
+    return somethingToSend;
+}
```
AndrewMeadows: This is the primary goal of this PR: move the "build the packet payload" work into this method while keeping all of the "coalesce packet payloads and send them" stuff intact.

It will now be possible to override this virtual method and experiment with different payload construction strategies.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2611/HighFidelity-Beta-PR11210-aee35b751ece12436a742ac387eacc687c53b5a8.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2611/HighFidelity-Beta-PR11210-aee35b751ece12436a742ac387eacc687c53b5a8.exe

AndrewMeadows:
### Test plan:

No behavior should have changed.

The code changes relate to the building and sending of Octree data packets from the entity-server (assignment-client), however the only kind of Octree data we're streaming these days are EntityItem's.  

(1) Run a local domain using the executables **from this PR**.
(2) Login to the local domain and verify that all the objects show up.
(3) Enable this menu option: **Developers --> Physics --> Highlight Simulation Ownership**
--> when you spin your view around you should see little red "down" arrows show up over all the objects in view.  Each red icon indicates an update arriving from the entity-server and should show up for all objects when moving view far enough side to side.
(4) Connect with a second avatar and verify that any changes the first avatar makes (changing position, color, shape) to an entity also shows up for the second.

cain-kilgore:
Testing.

cain-kilgore:
# Test Passed
Test | Result
--- | ---
(1) Run a local domain using the executables from this PR. | Pass
(2) Login to the local domain and verify that all the objects show up. | Pass
(3) Enable this menu option: Developers --> Physics --> Highlight Simulation Ownership | Pass
--> when you spin your view around you should see little red "down" arrows show up over all the objects in view. Each red icon indicates an update arriving from the entity-server and should show up for all objects when moving view far enough side to side. | Pass
(4) Connect with a second avatar and verify that any changes the first avatar makes (changing position, color, shape) to an entity also shows up for the second. | Pass

Everything seems to be working and changing as expected. I had Aaron create multiple shapes and entities and it was updating properly on my end.

assignment-client/src/octree/OctreeSendThread.cpp
```diff
@@ -458,84 +458,77 @@ int OctreeSendThread::packetDistributor(SharedNodePointer node, OctreeQueryNode*
...
+    OctreeQueryNode* nodeData = static_cast<OctreeQueryNode*>(params.nodeData);
+    if (!nodeData->elementBag.isEmpty()) {
+        quint64 encodeStart = usecTimestampNow();
+        quint64 lockWaitStart = encodeStart;
+
+        _myServer->getOctree()->withReadLock([&]{
+            OctreeServer::trackTreeWaitTime((float)(usecTimestampNow() - lockWaitStart));
+
+            OctreeElementPointer subTree = nodeData->elementBag.extract();
+            if (subTree) {
```
Atlante45: If `subTree` is false, it's possible that `OctreeServer::trackTreeWaitTime` and `OctreeServer::trackEncodeTime` would be called twice.

Would that be an issue?

assignment-client/src/octree/OctreeSendThread.cpp
```diff
@@ -576,14 +568,14 @@ void OctreeSendThread::traverseTreeAndSendContents(SharedNodePointer node, Octre
             }
             _packetData.changeSettings(true, targetSize); // will do reset - NOTE: Always compressed
         }
-        OctreeServer::trackTreeWaitTime(lockWaitElapsedUsec);
-        OctreeServer::trackEncodeTime(encodeElapsedUsec);
+        if (!somethingToSend) {
+            // these stats were not updated so we record zero to record this fact
+            OctreeServer::trackTreeWaitTime(0.0f);
```
ZappoMan: this doesn't look right, it looks like you still are actually waiting on a lock in traverseTreeAndBuildNextPacketPayload() and sometimes producing a false for somethingToSend, in which case, you're calling trackTreeWaitTime() twice, once with the actual wait time and once with 0.0.

That will have to erroneous effects on the statistics being tracked...

- it will increase `_noTreeWait` count when in fact there was a wait for the tree
- it will lower the "average" wait time, because one loop will count as two waits, one of the actual wait time and one of zero, resulting in the "recorded" wait time being half of the actual wait time.

ZappoMan:
The Test Plan on this PR is not sufficient... 
- This PR changes out statistics are calculated, but has no test plan for verifying that the statistics are still correct.
We need a better test plan.

assignment-client/src/octree/OctreeServer.cpp
ZappoMan: There's a difference between not tracking the time (skipping recording) and tracking a time of 0.0.

I'm pretty sure this change will break statistics.

Three passes of: 1.0, SKIP, 1.0 - result in an average of 1.0
Three passes of: 1.0, 0,0, 1.0 - result in an average of 0.667

AndrewMeadows: Ah I see.  I understand the intended behavior now.  I was taking the old code for its word and was trying to NOT change behavior.  Turns out the old code was not properly skipping measurements: on skip it would set `time` to zero and then roll that into the average.  But I've fixed it now.

I suppose it might make QA comparisons of old and new stats a little funny since the old stats would be expected to have a slightly lower average time for all of those times when skip was incorrectly contributing zero time to the average, however I expect the "error" to be small.

AndrewMeadows:
I need to figure out how to view the stats and interpret them before I can write an updated test plan.  I'll try to get to that soon.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2692/HighFidelity-Beta-PR11210-45f194f6bc18dc9bcae1974721b9ec284878c77f.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2692/HighFidelity-Beta-PR11210-45f194f6bc18dc9bcae1974721b9ec284878c77f.exe

assignment-client/src/octree/OctreeSendThread.cpp
AndrewMeadows: Now fixed: we track the times one way in the `if` and another way in the `else`.

assignment-client/src/octree/OctreeServer.cpp
```diff
@@ -136,64 +136,67 @@ void OctreeServer::trackEncodeTime(float time) {
...
+            _longEncode++;
+            _averageLongEncodeTime.updateAverage(time);
+        } else {
+            _extraLongEncode++;
+            _averageExtraLongEncodeTime.updateAverage(time);
+        }
+        _averageEncodeTime.updateAverage(time);
     }
-    _averageEncodeTime.updateAverage(time);
 }
```
AndrewMeadows: The intended behavior of the original code was for the "skip" to not affect the average, however that is not what was happening: it would set `time` to zero and then would contribute that to the average.

We now only contribute time when it is not to be "skipped".

Note, this changes the behavior of the stats so it will be difficult to compare the before and after in a way that would be informative about a possible problem in the new code, especially since the stats running average window was set so large.

If you look at the top of this file you'll see that the effective "running average window" was set to 1e6 samples.  I've dropped this down to only 1e3 samples.

To understand the problem consider these facts:

The very first sample would get full weight when computing the average.
Subsequent samples would pull the average in their direction with a weight of only 1/sample_window.

Now, under the old conditions consider the case when the very first sample was bad (e.g. a "skip" that was incorrectly added as a zero measurement, or even a spike in a noisy signal): the average starts off at whatever the bad measurement was at full weight.  It would then take a few multiples (about 3 or more) of the sample window at the "correct" measurement before the average would be pulled to a more reasonable value.  When the sample window is 1e6 but the sample rate is 1e3... it would take at least 3000 seconds (50 minutes) before the average could be "used".  In short: when the window is too large it ruins the average's real-time usefulness.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2744/HighFidelity-Beta-PR11210-d6b914283c4a12d8d00526568386830345725b60.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2744/HighFidelity-Beta-PR11210-d6b914283c4a12d8d00526568386830345725b60.exe

AndrewMeadows:
**Test Plan 2**

This is how to examine the entity-server timing stats:

**(1)** Run a local domain and 6 assignment-clients
**(2)** In a web browser navigate to http://127.0.0.1:40100/
**(3)** Follow the UUID link to the entity-server
**(4)** Look for item **3. timing**

Now that you know how to do that, the test plan is as follows:

**(5)** Using a hifi/master installer: run a local domain and its assignment-client servers
**(6)** Make sure the domain has several entities inside: between 100 and 1000
**(7)** Immediately after restarting the entity-server (alternatively restart all assignment-cleints) navigate to the entity-server's page with timing details and note the times.  In particular the avgCompressAndWriteTime stat -- it will probably be rather low (less than 1.0) but it might be high (100 or more).
**(8)** Login to the domain while watching the entity-server stats (which should update every second or so) and spin your avatar around so that lots and lots of entity-server packets are generated.
**(9)** Observe the slow motion of the stats.  They will gradually move toward some average but it will take a long time for them to get there.  You don't need to wait the full 50 minutes, but verify that they continually drift in one direction -- we don't expect the values to go both up and down.
**(10)** Take a screenshot of the values and note the directions they were moving, either up or down.

**(11)** Shutdown the domain and bring it back up on this PR.
**(12)** Navigate to the entity-server stats page, login, and repeat the rotation of the avatar while watching the stats move around.
**(13)** Within a few seconds (assuming you've got enough objects in the domain such that thousands of events are sampled every second) you should see the timing values reach a point where they actually vary up and down around some true average value.
**(14)** Take a screenshot of the timing values and verify the following:
Whatever direction the stats were drifting in **(10)** they should have been moving toward the values here in **(14)**.

AndrewMeadows:
Jenkins, rebuild this please.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2753/HighFidelity-Beta-PR11210-d6b914283c4a12d8d00526568386830345725b60.dmg

marko8904:
build this please

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2756/HighFidelity-Beta-PR11210-d6b914283c4a12d8d00526568386830345725b60.dmg

marko8904:
build this please

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2760/HighFidelity-Beta-PR11210-d6b914283c4a12d8d00526568386830345725b60.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2760/HighFidelity-Beta-PR11210-d6b914283c4a12d8d00526568386830345725b60.exe

Nex-Pro:
testing this

Nex-Pro:
@AndrewMeadows See the timings on these screengrabs. #7070 was constantly creeping up over time.

#11210 remained more or less around 31-32

Had an array of 800 spheres in both builds.

Let me know if you want me to leave 7070 to soak longer.

SamGondelman:
Ok, it's hard to tell here because there's no ground truth to compare to...but at least looking at avgCompressAndWriteTime, it seems like the two tests on 7070 are creeping towards the value from this PR.  Based on that I am comfortable calling this a pass.

Atlante45:
Looks good to me.

assignment-client/src/octree/OctreeServer.cpp
ZappoMan: This needs to be fixed before we merge this.

assignment-client/src/octree/OctreeSendThread.cpp
```diff
@@ -458,84 +458,80 @@ int OctreeSendThread::packetDistributor(SharedNodePointer node, OctreeQueryNode*
     return _truePacketsSent;
 }
 
+bool OctreeSendThread::traverseTreeAndBuildNextPacketPayload(EncodeBitstreamParams& params) {
+    bool somethingToSend = false;
+    OctreeQueryNode* nodeData = static_cast<OctreeQueryNode*>(params.nodeData);
+    if (!nodeData->elementBag.isEmpty()) {
+        quint64 encodeStart = usecTimestampNow();
```
SamGondelman: Shouldn't encodeStart be moved into the lock?

assignment-client/src/octree/OctreeSendThread.cpp
```diff
@@ -458,84 +458,80 @@ int OctreeSendThread::packetDistributor(SharedNodePointer node, OctreeQueryNode*
...
-                // Our trackSend() function is implemented by the server subclass, and will be called back
-                // during the encodeTreeBitstream() as new entities/data elements are sent
-                params.trackSend = [this](const QUuid& dataID, quint64 dataEdited) {
-                    _myServer->trackSend(dataID, dataEdited, _nodeUuid);
-                };
-
-                // TODO: should this include the lock time or not? This stat is sent down to the client,
-                // it seems like it may be a good idea to include the lock time as part of the encode time
-                // are reported to client. Since you can encode without the lock
-                nodeData->stats.encodeStarted();
```
AndrewMeadows: @SamGondelman I put the lock inside the `encodeStart` outside the lock because of this `TODO` comment.  Meanwhile the final part of the comment looks wrong to me: we can't encode without the lock because `encodeTreeBitstream()` takes `subtree` as an argument and we can't risk `subtree` changing while encoding it.

Note: some of the stuff above this line (`params` initialization and copying of `viewFrustum`) has been moved outside of the lock context (and more importantly outside of the `while` loop), so the encode stat measurement should now be even more accurate and correct.

PR #11214, opened 2017-08-18
Fix Image3DOverlays with fractional alpha

SamGondelman:
[FB6912](https://highfidelity.fogbugz.com/f/cases/6912/image3d-overlay-with-alpha-doesn-t-render-properly-unless-drawInFront-is-true) and [FB7077](https://highfidelity.fogbugz.com/f/cases/7077/Low-alpha-portions-of-PNGs-are-rendered-as-fully-transparent-in-image3d-overlays)

Image3DOverlays weren't properly setting their payload keys to be transparent once their images loaded.

Related:
- web3d overlays weren't setting their shapeKey withOwnPipeline, which manifested itself in a strange lighting issue if you had image3d overlays and web3d overlays together.
- line3ds were cheating by setting their alpha to 0.5 when glow was enabled.  This was a hack and could easily leave them with an incorrect payload key if their glow changed.  Now they should set their keys correctly and don't mess with their alpha.
- the image3d texCoords were off by 0.5 pixels for some reason

Test:

- Run [this](https://gist.githubusercontent.com/SamGondelman/ce12a905418eac2cb9e12a09f861b79d/raw/dba63c2902454032dbeb8300206ca854afadd96d/Image3DOverlayTest.js).  On master you would see something messed up, like:
image3dbefore
Now everything should work, like this:
image3dafter3
- While looking at those image3d overlays, toggle the tablet.  The lighting on the image3d overlays shouldn't change.  (This won't happen on master, but that's because everything else was already messed up).
- On master, the two image3d overlays on the right had tex coords that were slightly off (bleeding around the edges of the overlays):
image3dafter2
image3dafter5
- Now they should look properly clamped:
image3dafter4
- The line3d issue isn't very obvious.  On master, if you create a line3d overlay with glow > 0.0, and then set glow = 0.0, it would still have alpha = 0.5.  Now, if you do that, at the end the line will have alpha = 1.0 (or whatever you set it to).  Since the lines are so thin this is difficult to confirm.


hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2610/HighFidelity-Beta-PR11214-f56e95225311528b8786d66da33714ecce9b16d2.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2610/HighFidelity-Beta-PR11214-f56e95225311528b8786d66da33714ecce9b16d2.dmg

Nex-Pro:
Testing this

Nex-Pro:
## PASS ! 

@SamGondelman 

hifi-snap-by--on-2017-08-19_00-02-20
hifi-snap-by--on-2017-08-19_00-06-00

interface/src/ui/overlays/Image3DOverlay.cpp
```diff
@@ -58,12 +59,28 @@ void Image3DOverlay::render(RenderArgs* args) {
...
     if (!_visible || !getParentVisible() || !_texture || !_texture->isLoaded()) {
         return;
     }
 
+    // Once the texture has loaded, check if we need to update the render item because of transparency
+    if (!_textureIsLoaded && _texture && _texture->getGPUTexture()) {
+        _textureIsLoaded = true;
+        if (_texture->getGPUTexture()->getUsage().isAlpha()) {
+            auto itemID = getRenderItemID();
+            setAlpha(0.5f);
```
jherico: it feels like rather than explicitly setting the alpha, we should have an additional boolean member flag like `_alphaTexture`.  Then in the `getShapeKey()` method you can add `if (_alphaTexture || getAlpha() != 1.0f) {` to return a translucent key if required.

interface/src/ui/overlays/Image3DOverlay.cpp
```diff
@@ -19,6 +19,7 @@
 
 #include "GeometryUtil.h"
 
+#include "Application.h"
```
jherico: try to avoid creating an unnecessary dependency on this header.  You can use `AbstractViewStateInterface` to get the scene pointer and create the transaction instead.

ctrlaltdavid:
Testing with a PNG that is black with alpha going from 0.0 to 1.0, though the first part of the low alpha end is no longer fully transparent (which is good) the high alpha end seems not to reach fully opaque (seems to be about 50%).

interface/src/ui/overlays/Image3DOverlay.cpp
SamGondelman: The real issue isn't actually in getShapeKey, but in getKey, which only checks the alpha.  This is also how line3ds do it, presumably because of that.  I guess the real solution is to have getKey() check something like isTransparent(), which line3ds and image3ds override.

SamGondelman:
@ctrlaltdavid That sounds like a very reasonable side effect of always setting the alpha to 0.5f.  I had convinced myself that the alpha wasn't being used but that's obviously false.  Looking in to a better fix.

YuppyPlays:
Testing.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2651/HighFidelity-Beta-PR11214-3a80c6cbe24c2c03c6a8750ba30202c3fd480565.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2651/HighFidelity-Beta-PR11214-3a80c6cbe24c2c03c6a8750ba30202c3fd480565.exe

ctrlaltdavid:
Tested latest build with a PNG that is black with alpha going from 0.0 to 1.0 - it renders nicely.

cain-kilgore:
# Test Passed
Test | Desktop | HMD
--- | --- | ---
Run this. On master you would see something messed up `...` | Pass | Pass
While looking at those image3d overlays, toggle the tablet. The lighting on the image3d overlays shouldn't change. | Pass | Pass
two image3d overlays on the right had tex coords that were slightly off, should now look properly clamped | Pass | Pass

Everything's looking good!

interface/src/ui/overlays/Web3DOverlay.cpp
```diff
@@ -318,8 +318,8 @@ void Web3DOverlay::render(RenderArgs* args) {
 }
 
 const render::ShapeKey Web3DOverlay::getShapeKey() {
-    auto builder = render::ShapeKey::Builder().withoutCullFace().withDepthBias();
-    if (getAlpha() != 1.0f) {
+    auto builder = render::ShapeKey::Builder().withoutCullFace().withDepthBias().withOwnPipeline();
```
samcake: Does that mean that the web3d overlay uses a specific gpu::Pipeline ?
what is the difference ?

samcake: Yes indeed

PR #11229, opened 2017-08-22
make building tests optional

ElderOrb:
... to reduce re-compile time on dev environments

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,6 +247,9 @@ endif()
 
 set_packaging_parameters()
 
+option(BUILD_TESTS "Build tests" ON)
+MESSAGE(STATUS "Build tests: " ${BUILD_TESTS})
+
 # add subdirectories for all targets
 if (NOT ANDROID)
   add_subdirectory(assignment-client)
@@ -259,7 +262,9 @@ if (NOT ANDROID)
   if (NOT SERVER_ONLY)
     add_subdirectory(interface)
     set_target_properties(interface PROPERTIES FOLDER "Apps")
-    add_subdirectory(tests)
+    if (BUILD_TESTS)
+      add_subdirectory(tests)
+    endif()
   endif()
   add_subdirectory(plugins)
   add_subdirectory(tools)


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/2668/HighFidelity-Beta-PR11229-a18410ef079e4c154a3b09a8556b1379c16ae246.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/2668/HighFidelity-Beta-PR11229-a18410ef079e4c154a3b09a8556b1379c16ae246.exe

YuppyPlays:
@ElderOrb Could we please get a test plan for QA :)

ElderOrb:
Hmmm... Well.. this option is more interesting to devs as it has no influence on the functionality but just speedup builds. But if you are asking about how to verify it, then.... 

The following build configurations should result in building tests (like it used to work before changes)
cmake .. -G "Visual Studio 15 Win64"

... and this one should result in excluding building tests from build: 

cmake .. -G "Visual Studio 15 Win64" -DBUILD_TESTS=OFF

ElderOrb:
Is there anything else required from me to make it merged?

AndrewMeadows:
Us hifi devs are split on this one.  Some of us think this is a good idea.  Others wonder why @ElderOrb doesn't just select the right target to build instead of ALL_BUILD.  Meanwhile maybe we should have a new target that only builds the essential targets?

ElderOrb:
> Others wonder why @ElderOrb doesn't just select the right target to build instead of ALL_BUILD. 

Sure, I can select only 'interface', but it is just less convenient. Also, less projects in project tree speedup search in solution and results to lesser memory consumption by VS. I understand this is probably not enough to justify but after all this is just an option, disabled by default for backward-compatibility.

AndrewMeadows:
Ok those are decent reasons.  I cloned this branch and verified that it works on linux: test targets are not available but others are.

PR #11298, opened 2017-09-02
7382 Create Menu does not Work on Tablet

ElderOrb:
https://highfidelity.fogbugz.com/f/cases/7382/Create-Menu-does-not-Work-on-Tablet

### DIFF
--- a/interface/resources/qml/hifi/tablet/TabletButton.qml
+++ b/interface/resources/qml/hifi/tablet/TabletButton.qml
@@ -3,6 +3,8 @@ import QtGraphicalEffects 1.0
 
 Item {
     id: tabletButton
+
+    property string captionColorOverride: ""
     property var uuid;
     property string icon: "icons/tablet-icons/edit-i.svg"
     property string hoverIcon: tabletButton.icon
@@ -102,7 +104,7 @@ Item {
 
     Text {
         id: text
-        color: "#ffffff"
+        color: captionColorOverride !== "" ? captionColorOverride: "#ffffff"
         text: tabletButton.text
         font.bold: true
         font.pixelSize: 18


ElderOrb:
for QA:

1. Switch to VR
2. Go to the domain "Zaru".
3. Ensure that both Create icon and text below becomes grey in tablet menu
4. Go to the domain 'localhost'
5. Ensure that both Create icon and text below are white in tablet menu

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3042/HighFidelity-Beta-PR11298-b5faa8ac567518aac7af91211263f52fb56ec337.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3042/HighFidelity-Beta-PR11298-b5faa8ac567518aac7af91211263f52fb56ec337.exe

interface/resources/qml/hifi/tablet/TabletButton.qml
```diff
@@ -3,6 +3,8 @@ import QtGraphicalEffects 1.0
 
 Item {
     id: tabletButton
+
+    property string captionColorOverride: ""
```
jherico: `color` is a primitive type in QML, so this should be `property color captionColorOverride`, possibly initialized to `null`

ElderOrb: I also thought about this, the issue is that color can't be null. It will lead to the following: 

'Unable to assign null to QColor'

jherico: hmmm....  what happens if you try to assign to a `Qt.rgba()` value to the string?

ElderOrb: Made simple test, the following works just fine as QColor has proper conversion to string:  

    property string a: Qt.rgba(128, 128, 0, 128)

    Rectangle {
        width: 100
        height: 100
        color: a
    }

anshuman64:
@jherico how's this looking?

interface/resources/qml/hifi/tablet/TabletButton.qml
jherico: not for nothing, but it would probably just be have a root level `property color captionColor: "#ffffff"` and then have the text element declare `color: tabletButton.captionColor`

ElderOrb: Agree, but in this case how would I deal with binding inside ToolbarButton.qml ? 

        color: captionColorOverride !== "" ? captionColorOverride: (button.isActive ? "#000000" : "#ffffff")

Current approach allows manipulating 'captionColorOverride' without breaking this binding.

PR #11299, opened 2017-09-02
Cleanup CMake

jherico:
This modifies a number of cmake files in order to ease multi-platform development and make the individual files easier to read and manage.  

- Modify the way Qt is located to ease multi-platform development on a single machine (see below)
- Move a large amount of boilerplate relating to our standard usage of CMake to `cmake/init.cmake`
- Move a large amount of boilerplate relating to compiler options to `cmake/compiler.cmake`
- Add options for toggling individual parts of the build..,  `BUILD_SERVER`, `BUILD_CLIENT` etc
  - The pre-existing SERVER_ONLY cmake flag is still honored and should result in the same build targets as before
- Add all the CMake macros, modules and other files to a new top level `cmake` project in the IDE to make them easier to access and edit.  Visual Studio 2017 actually has a very nice CMake editor that does auto completion.

## Qt Setup

Currently our Qt libraries are located via the environment variable `QT_CMAKE_PREFIX_PATH`.  Unless you're building for Android, in which case `ANDROID_QT_CMAKE_PREFIX_PATH`.  Or if you're building for UWP, then `UWP_QT_CMAKE_PREFIX_PATH`.  This needlessly complicates the setup for doing builds for alternative platforms which require cross-compiling or a different Qt runtime.  

The new system does away with the variants of `QT_CMAKE_PREFIX_PATH` and instead uses knowledge of the well defined locations of various Qt runtimes within a Qt installation directory.  _If_ you have an environment variable `QT_CMAKE_PREFIX_PATH` it will still take precedence over the new mechanism.  However, if you set no variable, CMake will attempt to find the root of the Qt install in it's default directory (`C:/Qt` for Windows, `~/Qt` for all other platforms).  This location can be overridden with the `QT_ROOT` environment variable.  It will then look for the version directory inside Qt root, defaulting to `5.9.1`, but again, this can be overridden by a `QT_VERSION` environment variable.  Finally it will look for the actual runtime folder, the name of which will vary depending on the current platform and build target.  This last can be overridden with a `QT_ARCH` environment variable.  

The resulting `QT_DIR` will be set to `${QT_ROOT}/${QT_VERSION}/${QT_ARCH}` and `QT_CMAKE_PREFIX_PATH` will be set to `${QT_DIR}/lib/cmake`.  

The upshot is that if you don't set any environment variables at all, but you've installed Qt in the default location, everything will just work.  

*This means you'll be able to go from a Mac build directory directly to an Android build directory and run `cmake` without having to worry about what environment variables are set.  It also means that an IDE that triggers a CMake regeneration won't fail because it needs environment variables you've only set in shell window.  

## Testing

No application testing required.  This will only affect the build process, not the application itself.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,254 +4,81 @@ else()
   cmake_minimum_required(VERSION 3.2)
 endif()
 
-if (USE_ANDROID_TOOLCHAIN)
-  set(CMAKE_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/cmake/android/android.toolchain.cmake")
-  set(ANDROID_NATIVE_API_LEVEL 19)
-  set(ANDROID_TOOLCHAIN_NAME arm-linux-androideabi-clang3.5)
-  set(ANDROID_STL c++_shared)
-endif ()
-
-if (WIN32)
-  cmake_policy(SET CMP0020 NEW)
-endif (WIN32)
-
-if (POLICY CMP0043)
-  cmake_policy(SET CMP0043 OLD)
-endif ()
-
-if (POLICY CMP0042)
-  cmake_policy(SET CMP0042 OLD)
-endif ()
-
-set_property(GLOBAL PROPERTY USE_FOLDERS ON)
-set_property(GLOBAL PROPERTY PREDEFINED_TARGETS_FOLDER "CMakeTargets")
+include("cmake/init.cmake")
 
 project(hifi)
-add_definitions(-DGLM_FORCE_RADIANS)
-set(CMAKE_CXX_FLAGS_DEBUG  "${CMAKE_CXX_FLAGS_DEBUG} -DDEBUG")
 
-find_package( Threads )
+include("cmake/compiler.cmake")
 
-if (WIN32)
-  if (NOT "${CMAKE_SIZEOF_VOID_P}" EQUAL "8")
-    message( FATAL_ERROR "Only 64 bit builds supported." )
-  endif()
-
-  add_definitions(-DNOMINMAX -D_CRT_SECURE_NO_WARNINGS)
-
-  if (NOT WINDOW_SDK_PATH)
-    set(DEBUG_DISCOVERED_SDK_PATH TRUE)
-  endif()
-
-  # sets path for Microsoft SDKs
-  # if you get build error about missing 'glu32' this path is likely wrong
-  if (MSVC_VERSION GREATER_EQUAL 1910) # VS 2017
-    set(WINDOW_SDK_PATH "C:/Program Files (x86)/Windows Kits/10/Lib/${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION}/x64" CACHE PATH "Windows SDK PATH")
-  elseif (MSVC_VERSION GREATER_EQUAL 1800) # VS 2013
-    set(WINDOW_SDK_PATH "C:\\Program Files (x86)\\Windows Kits\\8.1\\Lib\\winv6.3\\um\\${WINDOW_SDK_FOLDER}" CACHE PATH "Windows SDK PATH")
-  else()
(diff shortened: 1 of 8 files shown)


hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3043/HighFidelity-Beta-PR11299-9c03fa0369c2a4c8369c9b99290904f76501c099.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3045/HighFidelity-Beta-PR11299-e64227bccdf2b83d171f7297b587e86a7ee0d483.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3045/HighFidelity-Beta-PR11299-e64227bccdf2b83d171f7297b587e86a7ee0d483.exe

OmegaHeron:
Built this on Linux as well - no change to my workflow as I already provided QT path (/opt/qt59 which is where side loaded Qt5.9.x lives on Ubuntu Xenial if installed by PPA).  One ancient build script I had failed where I defined QT Path with -DQT_CMAKE_PREFIX_PATH=/opt/qt59/lib/cmake option to cmake, but was fine if I exported it via bash.  Not really a problem as I doubt many use the option to cmake opting to define env var.

CMakeLists.txt
```diff
@@ -4,285 +4,118 @@ else()
...
-
-if (WIN32)
-    if (NOT EXISTS ${QT_CMAKE_PREFIX_PATH})
-        message(FATAL_ERROR "Could not determine QT_CMAKE_PREFIX_PATH.")
-    endif ()
+if (ANDROID OR UWP)
+    set(MOBILE 1)
+else()
+    set(MOBILE 0)
 endif()
```
AndrewMeadows: `MOBILE` is set but never used AFAICT.   I assume this is a step toward the Glorious Future.

CMakeLists.txt
```diff
@@ -4,285 +4,118 @@ else()
...
-set(CMAKE_AUTOMOC ON)
-# Instruct CMake to run rcc automatically when needed
-set(CMAKE_AUTORCC ON)
+if (ANDROID OR UWP)
+    option(BUILD_SERVER "Build server components" OFF)
+    option(BUILD_TOOLS "Build tools" OFF)
+else()
+    option(BUILD_SERVER "Build server components" ON)
+    option(BUILD_TOOLS "Build tools" ON)
+endif()
```
AndrewMeadows: Maybe should use the new `MOBILE` variable here?
```
if (MOBILE)
    ...
else()
    ...
endif()
```

CMakeLists.txt
```diff
@@ -4,285 +4,118 @@ else()
...
+
+if (BUILD_CLIENT)
     add_subdirectory(interface)
     set_target_properties(interface PROPERTIES FOLDER "Apps")
-    if (BUILD_TESTS)
-      add_subdirectory(tests)
+    if (ANDROID)
+        add_subdirectory(gvr-interface)
+        set_target_properties(gvr-interface PROPERTIES FOLDER "Apps")
```
AndrewMeadows: Just to confirm: this really should be `if (ANDROID)` and not `if (MOBILE)` ?

CMakeLists.txt
jherico: Yes.  There are a number of places where we're going to end up doing something like `if (ANDROID OR UWP)`.  I'd rather have some common value we can check.

CMakeLists.txt
jherico: the gvr-interface is specifically for android.  it contains a bunch of java wrappers for the app.

CMakeLists.txt
jherico: maybe.  honestly it still needs more work, because on some platforms BUILD_SERVER shouldn't even be an option... the cmake should just act like the server components don't exist.

PR #11311, opened 2017-09-06
wallet cleanup

davidkelly:
Seems a couple things that were fixed creeped back into the wallet.  This re-fixes them -- the clearing of your cached passphrase in particular.  Also our caching of the passphrase, and the logic around set vs change passphrase was problematic.   Now, whenever the private key will not decrypt, we clear the passphrase (since it is wrong).  This fixes issues with entering the wrong passphrase, having it cached, and next time you enter it (in passphrase modal), us thinking it is a change to the passphrase.  That guaranteed you couldn't decrypt as we first decrypt with cached passphrase before changing it.  

Finally, I removed createIfNeeded.  We _never_ just create a keypair.  There is a call to generate the keypair, and that is now exposed so it can be called directly when needed.  Since this was called as part of a balance call for instance, you could inadvertently create keys after one of the above failures, creating keys with either a cached or empty passphrase.

### DIFF
--- a/scripts/system/commerce/wallet.js
+++ b/scripts/system/commerce/wallet.js
@@ -71,6 +71,10 @@
             case 'maybeEnableHmdPreview':
                 Menu.setIsOptionChecked("Disable Preview", isHmdPreviewDisabled);
                 break;
+            case 'passphraseReset':
+                onButtonClicked();
+                onButtonClicked();
+                break;
             case 'walletReset':
                 onButtonClicked();
                 onButtonClicked();
--- a/interface/resources/qml/hifi/commerce/wallet/Help.qml
+++ b/interface/resources/qml/hifi/commerce/wallet/Help.qml
@@ -55,6 +55,7 @@ Item {
         text: "DEBUG: Clear Cached Passphrase";
         onClicked: {
             commerce.setPassphrase("");
+            sendSignalToWallet({method: 'passphraseReset'});
         }
     }
     HifiControlsUit.Button {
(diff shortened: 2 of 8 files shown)


scripts/system/commerce/wallet.js
```diff
@@ -71,6 +71,10 @@
             case 'maybeEnableHmdPreview':
                 Menu.setIsOptionChecked("Disable Preview", isHmdPreviewDisabled);
                 break;
+            case 'passphraseReset':
+                onButtonClicked();
```
cain-kilgore: Curious as to why `onButtonClicked()` is in there twice, for case `walletReset` too.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3069/HighFidelity-Beta-PR11311-89cda8d3062b7bcdee55b793f467551f40d4b6e9.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3069/HighFidelity-Beta-PR11311-89cda8d3062b7bcdee55b793f467551f40d4b6e9.exe

scripts/system/commerce/wallet.js
davidkelly: me too -- I followed the existing pattern, which appears strange, but it works to dismiss the app in both tablet and toolbar modes.

howard-stearns: There will come a time when we will audit this, and that double code wont' pass muster.
But that day is not today.

PR #11370, opened 2017-09-15
Fixing the web surface of the tablet

samcake:
This pr should fix the web surface of the tablet not being at the right position with regard to the tablet model itself. In particular when moving the tablet, the web surface appears to be inside or ahead of the tablet movement.

This is the quick fix, there is a broader issue waiting to be solved regarding all the values changing during gameplay that need to be captured per value before being passed to the render  scene through the render  transactions. This is touching many part of the code that will be addressed in future work.

## TEST PLAN
When showing the tablet in HMD:
- grab it and shake it.
- move around and rotate your avatar.
in this PR, the web surface of the tablet moves correctly never drifting away from the table model itself.
it s showing the bug in current master.

When showing the tablet in Desktop:
- grab it with the mouse and shake it.
- move around and rotate your avatar.
in this PR, the web surface of the tablet moves correctly never drifting away from the table model itself.
it s showing the bug in current master.

Cheers

### DIFF
--- a/interface/src/ui/overlays/Base3DOverlay.cpp
+++ b/interface/src/ui/overlays/Base3DOverlay.cpp
@@ -271,11 +271,12 @@ void Base3DOverlay::update(float duration) {
     // TODO: Fix the value to be computed in main thread now and passed by value to the render item.
     //       This is the simplest fix for the web overlay of the tablet for now
     if (_renderTransformDirty) {
+        _renderTransformDirty = false;
         auto itemID = getRenderItemID();
         if (render::Item::isValidID(itemID)) {
             render::ScenePointer scene = qApp->getMain3DScene();
             render::Transaction transaction;
-                transaction.updateItem<Overlay>(itemID, [](Overlay& data) {
+            transaction.updateItem<Overlay>(itemID, [](Overlay& data) {
                 auto overlay3D = dynamic_cast<Base3DOverlay*>(&data);
                 if (overlay3D) {
                     auto latestTransform = overlay3D->evalRenderTransform();
@@ -284,7 +285,6 @@ void Base3DOverlay::update(float duration) {
             });
             scene->enqueueTransaction(transaction);       
         }
-        _renderTransformDirty = false;
     }
 }
 
--- a/interface/src/ui/overlays/Base3DOverlay.cpp
+++ b/interface/src/ui/overlays/Base3DOverlay.cpp
@@ -256,15 +256,74 @@ bool Base3DOverlay::findRayIntersection(const glm::vec3& origin, const glm::vec3
 void Base3DOverlay::locationChanged(bool tellPhysics) {
     SpatiallyNestable::locationChanged(tellPhysics);
 
-    auto itemID = getRenderItemID();
+    // Force the actual update of the render transform now that we notify for the change
+    // so it s captured for the time of rendering
+    notifyRenderTransformChange();
+
+ /*   auto itemID = getRenderItemID();
     if (render::Item::isValidID(itemID)) {
         render::ScenePointer scene = qApp->getMain3DScene();
         render::Transaction transaction;
         transaction.updateItem(itemID);
         scene->enqueueTransaction(transaction);
-    }
+    }*/
 }
 
 void Base3DOverlay::parentDeleted() {
     qApp->getOverlays().deleteOverlay(getOverlayID());
+}
+
+void Base3DOverlay::update(float duration) {
+    if (_renderTransformDirty) {
+        auto self = this;
+        // queue up this work for later processing, at the end of update and just before rendering.
+        // the application will ensure only the last lambda is actually invoked.
+        /*  void* key = (void*)this;
+                std::weak_ptr<SpatiallyNestable> weakSelf = shared_from_this();
+                AbstractViewStateInterface::instance()->pushPostUpdateLambda(key, [weakSelf]() {
+                    // do nothing, if the model has already been destroyed.
+                    auto spatiallyNestableSelf = weakSelf.lock();
+                    if (!spatiallyNestableSelf) {
+                        return;
+                    }
+                    auto self = std::dynamic_pointer_cast<Base3DOverlay>(spatiallyNestableSelf);
+        */
+        #ifdef UpdateInMain
+            self->setRenderTransform(self->evalRenderTransform());
+        #else
+            auto renderTransform = self->evalRenderTransform();
+        #endif
+            auto itemID = self->getRenderItemID();
+         if (render::Item::isValidID(itemID)) {
+            render::ScenePointer scene = qApp->getMain3DScene();
+            render::Transaction transaction;
+            #ifdef UpdateInMain
+                transaction.updateItem(itemID);
+            #else
+                transaction.updateItem<Overlay>(itemID, [renderTransform](Overlay& data) {
+                auto overlay3D = dynamic_cast<Base3DOverlay*>(&data);
+                if (overlay3D) {
+                    auto latestTransform = overlay3D->evalRenderTransform();
+                    overlay3D->setRenderTransform(latestTransform);//  evalRenderTransform();
+                }
+            });
+            #endif
+                 scene->enqueueTransaction(transaction);
+            
+        }
+        //     });
+        _renderTransformDirty = false;
+    }
+}
+
+void Base3DOverlay::notifyRenderTransformChange() const {
+    _renderTransformDirty = true;
+}
+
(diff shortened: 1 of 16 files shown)


hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3344/HighFidelity-Beta-PR11370-30bf80096314871fe772d12be76d6b0a29795b25.exe

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

**When showing the tablet in HMD:**
grab it and shake it. **Check**
move around and rotate your avatar. **Check**
in this PR, the web surface of the tablet moves correctly never drifting away from the table model itself. **Check**
it s showing the bug in current master. **Check**

**When showing the tablet in Desktop:**
grab it with the mouse and shake it. **Check**
move around and rotate your avatar. **Check**
in this PR, the web surface of the tablet moves correctly never drifting away from the table model itself. **Check**
it s showing the bug in current master. **Check, though not nearly as noticeable as HMD**

**Pass**

cain-kilgore:
test this please

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3346/HighFidelity-Beta-PR11370-30bf80096314871fe772d12be76d6b0a29795b25.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3346/HighFidelity-Beta-PR11370-30bf80096314871fe772d12be76d6b0a29795b25.exe

interface/src/ui/overlays/Web3DOverlay.cpp
```diff
@@ -184,6 +184,7 @@ void Web3DOverlay::update(float deltatime) {
         // update globalPosition
         _webSurface->getSurfaceContext()->setContextProperty("globalPosition", vec3toVariant(getPosition()));
     }
+    Billboard3DOverlay::update(deltatime);
```
jherico: lots of the overlays types have an `update` function override, but you've only added the parent call in `Web3DOverlay`.  Don't you need to do this in `Image3DOverlay`, `ModelOverlay` and `Text3DOverlay` as well?

Also I'd suggest adding a _private_  `using Parent = XXX` declaration to each class where you're going to do this, so you can just generically add this code as `Parent::update(deltaTime);`

Shouldn't all of those types also have the same updates to their `render` code to pull the render transform?

interface/src/ui/overlays/Base3DOverlay.cpp
```diff
@@ -256,15 +256,46 @@ bool Base3DOverlay::findRayIntersection(const glm::vec3& origin, const glm::vec3
...
 
-    auto itemID = getRenderItemID();
-    if (render::Item::isValidID(itemID)) {
-        render::ScenePointer scene = qApp->getMain3DScene();
-        render::Transaction transaction;
-        transaction.updateItem(itemID);
-        scene->enqueueTransaction(transaction);
-    }
+    // Force the actual update of the render transform through the notify call
+    notifyRenderTransformChange();
```
jherico: Since you're not using any signal handling here, it looks like you could replace this with `_renderTransformDirty = true` and eliminate the function `notifyRenderTransformChange`

interface/src/ui/overlays/Base3DOverlay.cpp
```diff
@@ -256,15 +256,46 @@ bool Base3DOverlay::findRayIntersection(const glm::vec3& origin, const glm::vec3
...
+    // In Base3DOverlay, if its location or bound changed, the renderTrasnformDirty flag is true.
+    // then the correct transform used for rendering is computed in the update transaction and assigned.
+    // TODO: Fix the value to be computed in main thread now and passed by value to the render item.
+    //       This is the simplest fix for the web overlay of the tablet for now
+    if (_renderTransformDirty) {
+        auto itemID = getRenderItemID();
+        if (render::Item::isValidID(itemID)) {
+            render::ScenePointer scene = qApp->getMain3DScene();
+            render::Transaction transaction;
+                transaction.updateItem<Overlay>(itemID, [](Overlay& data) {
```
jherico: bad indentation.

interface/src/ui/overlays/Base3DOverlay.cpp
```diff
@@ -256,15 +256,46 @@ bool Base3DOverlay::findRayIntersection(const glm::vec3& origin, const glm::vec3
...
+                auto overlay3D = dynamic_cast<Base3DOverlay*>(&data);
+                if (overlay3D) {
+                    auto latestTransform = overlay3D->evalRenderTransform();
+                    overlay3D->setRenderTransform(latestTransform);
+                }
+            });
+            scene->enqueueTransaction(transaction);       
+        }
+        _renderTransformDirty = false;
```
jherico: since `_renderTransformDirty` isn't atomic and can likely be modified on multiple threads (I think), you should move this line to right after `if (_renderTransformDirty)`, i.e. before you read the transform.  that way if the transform changes again between the `if` and setting it to true, you'll just capture it again in the next pass.  

Not sure this matters since the actual transform is read in a transaction.

interface/src/ui/overlays/Base3DOverlay.cpp
```diff
@@ -256,15 +256,46 @@ bool Base3DOverlay::findRayIntersection(const glm::vec3& origin, const glm::vec3
...
+    //       This is the simplest fix for the web overlay of the tablet for now
+    if (_renderTransformDirty) {
+        auto itemID = getRenderItemID();
+        if (render::Item::isValidID(itemID)) {
+            render::ScenePointer scene = qApp->getMain3DScene();
+            render::Transaction transaction;
+                transaction.updateItem<Overlay>(itemID, [](Overlay& data) {
+                auto overlay3D = dynamic_cast<Base3DOverlay*>(&data);
+                if (overlay3D) {
+                    auto latestTransform = overlay3D->evalRenderTransform();
```
jherico: if you call this outside of the transaction and pass `latestTransform` by value into the transaction lambda, does it still fix the problem?

interface/src/ui/overlays/Web3DOverlay.cpp
samcake: agreed, but htis was a lot more code to review and we want that fix in quickly...

interface/src/ui/overlays/Base3DOverlay.cpp
samcake: if i was doing the same with the models yes (capturing their transform in main thread, not in the transaction) but other problems appear then when in desktop mode.
I went for the easiest fix right now

interface/src/ui/overlays/Base3DOverlay.cpp
samcake: i m going to keep that one so we could potentially trigger  from different places, ok ?

interface/src/ui/overlays/Web3DOverlay.h
```diff
@@ -21,6 +21,8 @@ class Web3DOverlay : public Billboard3DOverlay {
     Q_OBJECT
 
 public:
+    using Parent = Billboard3DOverlay;
```
jherico: should be private.  If there were another class derived from Web3DOverlay and you declared `using Parent = Web3DOverlay` in it, you'd have an ambiguous symbol.  `Parent` should only be a visible type from within the class itself, not in subclasses.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3353/HighFidelity-Beta-PR11370-97617c6dbe587d2fe1cff821efaa424bd74a7fc2.exe

samcake:
build this please

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3365/HighFidelity-Beta-PR11370-c687932ca7ec4d8ada3ab5a8b02f156883cfa8fd.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3371/HighFidelity-Beta-PR11370-524c94cdd4e4f1f7a1ec5abb32fc2c8e028eeae9.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3371/HighFidelity-Beta-PR11370-524c94cdd4e4f1f7a1ec5abb32fc2c8e028eeae9.exe

PR #11432, opened 2017-09-21
Fix overlay rendering to use cached Render Transform information

samcake:
This pr touches all the Overlay 3Ds to make sure they capture their _renderTransform (or extra properties representing their position) during the GameLoop update. and use theese captured render tranform during rendering calls.

This PR goes into game-render-interface branch where we consolidate these cahgnges accross the code of Interface.

## CODE REVIEW HINTS
I have changed in every  3DOverlay::render calls the way the transform or transform coordinates are passed to the batch to be coming from the cached variables _renderTransform.
the _renderxxx elements are all computed in Game loop from the evalRenderTransform call

which is itself called in the update of the overlays if needed and then passed as value to the render items.
To actually trigger this to happen, we call "notifyRenderTransformChange" anywhere this needs to be recomputed (when dimension changes, or position or...)

At the end of the day, all the variables needed to draw the overlay elements are computed in Game loop, not in Render loop

## TEST PLAN
- make sure that the overlay elements are rendering correctly at the right position and without latency
- check that the tablet web overlay is moving together correctly with the tablet Model
- using edit mode, check that the overlay / wdgets are behaving correclty.
 - make sure they are all visible as expected
 - make sure they behave correctly when modified or when the inspected object is moved

## KNOWN ISSUE
- When using the tablet in desktop, the tablet is showing position tearing when moving around with the avatar


hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3530/HighFidelity-Beta-PR11432-454e167d93aee556bdcc8c69e6a6c1b65e8bc452.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3530/HighFidelity-Beta-PR11432-454e167d93aee556bdcc8c69e6a6c1b65e8bc452.dmg

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3538/HighFidelity-Beta-PR11432-2d369a458647fd18b716ff474bcc608c75feec76.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3538/HighFidelity-Beta-PR11432-2d369a458647fd18b716ff474bcc608c75feec76.exe

cain-kilgore:
Testing.

cain-kilgore:
# Test Failed
Test | Result
--- | ---
make sure that the overlay elements are rendering correctly at the right position and without latency | Fail
check that the tablet web overlay is moving together correctly with the tablet Model | Fail
using edit mode, check that the overlay / wdgets are behaving correclty. | Pass
make sure they are all visible as expected | Pass
make sure they behave correctly when modified or when the inspected object is moved | Pass

I was having an issue where when I launched Settings > General (to get the Desktop Overlay to appear in HMD), the laser pointer wasn't appearing but I was able to interact.

Additionally, I wasn't able to click on any menu buttons on the Tablet with the Laser Pointer. I had to go up to it and physically tap on the buttons.

samcake:
@cain-kilgore 
about the 2 fails:
- does the ui surface of the table is moving correctly with the tablet model ?
- are you saying that you coudln;t see the laser when interacting with hud ?

interface/src/ui/overlays/Base3DOverlay.cpp
```diff
@@ -268,18 +268,17 @@ void Base3DOverlay::update(float duration) {
...
-    //       This is the simplest fix for the web overlay of the tablet for now
     if (_renderTransformDirty) {
-        _renderTransformDirty = false;
         auto itemID = getRenderItemID();
         if (render::Item::isValidID(itemID)) {
+            _renderTransformDirty = false;
+            // Capture the render transform value in game loop before 
+            auto latestTransform = evalRenderTransform();
             render::ScenePointer scene = qApp->getMain3DScene();
             render::Transaction transaction;
```
jherico: you might want to consider modifying the overlay `update` signature to take a scene and transaction parameter so you don't have to do this.

interface/src/ui/overlays/Circle3DOverlay.cpp
```diff
@@ -438,3 +438,7 @@ bool Circle3DOverlay::findRayIntersection(const glm::vec3& origin, const glm::ve
 Circle3DOverlay* Circle3DOverlay::createClone() const {
     return new Circle3DOverlay(this);
 }
+
+Transform Circle3DOverlay::evalRenderTransform() {
```
jherico: this method looks superfluous... isn't this the same implementation as `Transform Base3DOverlay::evalRenderTransform`?

interface/src/ui/overlays/Base3DOverlay.cpp
```diff
@@ -292,7 +291,7 @@ void Base3DOverlay::notifyRenderTransformChange() const {
     _renderTransformDirty = true;
 }
 
-Transform Base3DOverlay::evalRenderTransform() const {
```
jherico: it looks like the only reason this is no longer const is because it modifies _renderStart and _renderEnd in one subclass.  Would recommend changing those members to be 'mutable'

samcake: it s not const because of the BIllboard case
I agree with  that function should be const otherwise

interface/src/ui/overlays/Circle3DOverlay.cpp
samcake: you are right
and actually i haven;t change the transform used for rendering in that class...
fixing that

samcake:
I need these changes to move forward
this will be staged and reviewed again in the main pr
so we can go over the issues again
merging

interface/src/ui/overlays/Base3DOverlay.cpp
samcake: good idea

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3558/HighFidelity-Beta-PR11432-b424505f6560248fb99c1b358863b0ed7bf75229.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3558/HighFidelity-Beta-PR11432-b424505f6560248fb99c1b358863b0ed7bf75229.dmg

PR #11443, opened 2017-09-22
Laser Pointer fixes and updates

SamGondelman:
- Ray pick results now include the searchRay (origin and direction)
- Fixes a bug with precision picking
- Laser pointer creation happens on the thread you create the laser pointer from instead of being deferred to the main thread.  This fixes an issue where you couldn't edit a laser pointer immediately after creating it.
- Adds LaserPointers.setLaserLength to override the laser pointer length.  Set it <= 0.0 to disable the override.
- Fixes setLockEndUUID to only use the renderState, never the defaultRenderState.

Test plan:
- Laser pointers in default scripts (teleport, hand lasers) should still work.
- Run [this](https://gist.githubusercontent.com/SamGondelman/26abff5da64503df321254c8a846536b/raw/6823a83391936040f76ab7bf4f97d2ff67ecf475/LaserPointerTest5.js).  A sphere should follow your mouse and intersect with everything.  The color of the sphere should change as you rotate (if it's hitting something.  If it doesn't hit anything, it will be pink.).
- Press 1 to toggle precision picking (should be noticeable on models with complex geometry).
- Press 2 to toggle a demo of setLaserLength.  The sphere will change distance from you according to a function in the script.  It will not turn pink when you aren't intersecting anything.
- Hover over an entity and press 3.  The sphere will lock on to the center of that entity.  Pressing 2 will not affect the laser.  While it is locked, point your mouse so it isn't intersecting anything.  The sphere shouldn't turn pink, it should keep changing colors as your direction changes.  Press 3 again to disable locking.

### DIFF
--- a/interface/src/raypick/LaserPointer.cpp
+++ b/interface/src/raypick/LaserPointer.cpp
@@ -184,8 +184,10 @@ void LaserPointer::disableRenderState(const RenderState& renderState) {
 
 void LaserPointer::update() {
     RayPickResult prevRayPickResult = DependencyManager::get<RayPickScriptingInterface>()->getPrevRayPickResult(_rayPickUID);
-    if (_renderingEnabled && !_currentRenderState.empty() && _renderStates.find(_currentRenderState) != _renderStates.end() && prevRayPickResult.type != IntersectionType::NONE) {
-        updateRenderState(_renderStates[_currentRenderState], prevRayPickResult.type, prevRayPickResult.distance, prevRayPickResult.objectID, prevRayPickResult.searchRay, false);
+    if (_renderingEnabled && !_currentRenderState.empty() && _renderStates.find(_currentRenderState) != _renderStates.end() &&
+            (prevRayPickResult.type != IntersectionType::NONE || _laserLength > 0.0f || !_objectLockEnd.first.isNull())) {
+        float distance = _laserLength > 0.0f ? _laserLength : prevRayPickResult.distance;
+        updateRenderState(_renderStates[_currentRenderState], prevRayPickResult.type, distance, prevRayPickResult.objectID, prevRayPickResult.searchRay, false);
         disableRenderState(_defaultRenderStates[_currentRenderState].second);
     } else if (_renderingEnabled && !_currentRenderState.empty() && _defaultRenderStates.find(_currentRenderState) != _defaultRenderStates.end()) {
         disableRenderState(_renderStates[_currentRenderState]);
--- a/interface/src/raypick/LaserPointer.cpp
+++ b/interface/src/raypick/LaserPointer.cpp
@@ -92,8 +92,7 @@ void LaserPointer::updateRenderStateOverlay(const OverlayID& id, const QVariant&
     }
 }
 
-void LaserPointer::updateRenderState(const RenderState& renderState, const IntersectionType type, const float distance, const QUuid& objectID, const bool defaultState) {
-    PickRay pickRay = qApp->getRayPickManager().getPickRay(_rayPickUID);
+void LaserPointer::updateRenderState(const RenderState& renderState, const IntersectionType type, const float distance, const QUuid& objectID, const PickRay& pickRay, const bool defaultState) {
     if (!renderState.getStartID().isNull()) {
         QVariantMap startProps;
         startProps.insert("position", vec3toVariant(pickRay.origin));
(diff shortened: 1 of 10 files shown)


hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3569/HighFidelity-Beta-PR11443-cbe621f2b7a9a832b812d042e6af1d10f69e7734.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3573/HighFidelity-Beta-PR11443-0e3f1c514968ae166aa3ed178b07229f58e9eab2.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3573/HighFidelity-Beta-PR11443-0e3f1c514968ae166aa3ed178b07229f58e9eab2.exe

Nex-Pro:
T E S T I N G this

Nex-Pro:
## PASS

Laser pointers in default scripts (teleport, hand lasers) should still work. **SUCCESS**

Run this. A sphere should follow your mouse and intersect with everything. **SUCCESS**

The color of the sphere should change as you rotate (if it's hitting something. If it doesn't hit anything, it will be pink.). **SUCCESS**

Press 1 to toggle precision picking (should be noticeable on models with complex geometry). **SUCCESS**
Press 2 to toggle a demo of setLaserLength. The sphere will change distance from you according to a function in the script. It will not turn pink when you aren't intersecting anything. **SUCCESS**
Hover over an entity and press 3. The sphere will lock on to the center of that entity. **SUCCESS** Pressing 2 will not affect the laser. **SUCCESS**
While it is locked, point your mouse so it isn't intersecting anything. The sphere shouldn't turn pink, it should keep changing colors as your direction changes. **SUCCESS**
Press 3 again to disable locking.

interface/src/raypick/LaserPointerManager.cpp
```diff
@@ -105,6 +87,14 @@ void LaserPointerManager::setPrecisionPicking(QUuid uid, const bool precisionPic
     }
 }
 
+void LaserPointerManager::setLaserLength(QUuid uid, const float laserLength) {
+    QReadLocker lock(&_containsLock);
+    if (_laserPointers.contains(uid)) {
+        QWriteLocker laserLock(_laserPointerLocks[uid].get());
+        _laserPointers[uid]->setLaserLength(laserLength);
+    }
```
AndrewMeadows: Not  a blocker but...

I noticed that uid is looked up twice in `_laserPointers` (once in the `contains()` call and again to set the `laserLength`.  Normally I'm opposed to multiple lookups in a map when the same thing could be achieved with only one, however the size of the `_laserPointers` map is expected to be small right?  In which case the lookup cost is minimal.  Out of curiosity: what would be a reasonable max size of the container during normal user operation? 10 or less?  100's?

Also, there are two maps that often move in synchrony: `_laserPointers` and `_laserPointerLocks`.  Many operations require dual lookups: one for each map.  It occurs to me that if it were possible for the lock to be part of the "LaserPointer" instance itself then these operations would only require one lookup instead of two.

interface/src/raypick/LaserPointer.cpp
```diff
@@ -92,8 +92,7 @@ void LaserPointer::updateRenderStateOverlay(const OverlayID& id, const QVariant&
     }
 }
 
-void LaserPointer::updateRenderState(const RenderState& renderState, const IntersectionType type, const float distance, const QUuid& objectID, const bool defaultState) {
-    PickRay pickRay = qApp->getRayPickManager().getPickRay(_rayPickUID);
+void LaserPointer::updateRenderState(const RenderState& renderState, const IntersectionType type, const float distance, const QUuid& objectID, const PickRay& pickRay, const bool defaultState) {
```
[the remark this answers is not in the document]
SamGondelman: It was mostly a reminder to me when I was writing it, I left it in as a reminder to anyone looking at this in the future that updateRenderState is just supposed to be modifying the laser overlays using the given information.

interface/src/raypick/LaserPointerManager.cpp
SamGondelman: Since the performance impact of creating many additional laser pointers far outweighs the cost of these lookups, I won't worry about this for now, but I can address it in a separate PR.

PR #11458, opened 2017-09-26
Fix long tablet lag on first load

jherico:
Fix for [FB 7782](https://highfidelity.fogbugz.com/f/cases/7782).  The first time a tablet is loaded it causes the main thread to lock up for multiple seconds.

This was an intersection of several smaller issues. First, the `pickUrl` code (which I've removed) was being used when interacting with the QML surface cache, but then once the surface was acquired, it was re-loaded with the same QML code, but referenced as `_url`.  Since `pickUrl` transformed the `_url` member, that meant that the same QML content was being loaded through two different URLs, completely defeating the purpose of the cache.  

There was a significant amount of additional duplicated functionality between the `loadSourceUrl` method and the `buildWebSurface` method which called it.  I've eliminated this, and the `loadSourceUrl` method entirely.  Instead, when it would previously get called in the context of the URL being changed, we now either simply set the URL property on the surface (if the old and new URLs were both web content) or we completely reload the QML surface, thus properly using the cache.

There was some oddness with changing the base URL for QML surfaces depending on whether we were loading web or QML content, apparently because the `qml` prefix was left in the `tabletUI.js` script.  By eliminating that prefix in the script, I was able to load the tablet without having to mess with different base URLs for the QML content.  

Finally, since the tablet is widely used in HMD mode, it makes sense to cache the tablet QML content at application startup, so I've added code to do this in `Application.cpp` in the same place where we pre-load the web overlay QML.  

Testing with this build show negligible latency when bringing up the tablet for the first time, as compared to the current build.

## Testing

Launch the application and load the tablet.  Compared to the current master build the tablet should appear much faster and not trigger a freeze of the application.  There may be some small stutter, but it should be much less than in master.

WebOverlays should be tested with actual web content, and the tablet should be thoroughly tested as well.  Any existing tablet applications should be re-tested as well to ensure their UI still functions.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -2369,6 +2369,7 @@ void Application::initializeUi() {
 
     // Pre-create a couple of Web3D overlays to speed up tablet UI
     auto offscreenSurfaceCache = DependencyManager::get<OffscreenQmlSurfaceCache>();
+    offscreenSurfaceCache->reserve(TabletScriptingInterface::QML, 1);
     offscreenSurfaceCache->reserve(Web3DOverlay::QML, 2);
 }
 
--- a/interface/src/ui/overlays/Web3DOverlay.cpp
+++ b/interface/src/ui/overlays/Web3DOverlay.cpp
@@ -62,6 +62,7 @@ static const float OPAQUE_ALPHA_THRESHOLD = 0.99f;
 
 const QString Web3DOverlay::TYPE = "web3d";
 const QString Web3DOverlay::QML = "Web3DOverlay.qml";
+
 Web3DOverlay::Web3DOverlay() : _dpi(DPI) {
     _touchDevice.setCapabilities(QTouchDevice::Position);
     _touchDevice.setType(QTouchDevice::TouchScreen);
@@ -97,6 +98,10 @@ Web3DOverlay::~Web3DOverlay() {
     }
 }
 
+void Web3DOverlay::rebuildWebSurface() {
+    destroyWebSurface();
+    buildWebSurface();
+}
 
 void Web3DOverlay::destroyWebSurface() {
     if (!_webSurface) {
@@ -136,17 +141,23 @@ void Web3DOverlay::buildWebSurface() {
         return;
     }
     gl::withSavedContext([&] {
-        _webSurface = DependencyManager::get<OffscreenQmlSurfaceCache>()->acquire(pickURL());
         // FIXME, the max FPS could be better managed by being dynamic (based on the number of current surfaces
         // and the current rendering load)
         if (_currentMaxFPS != _desiredMaxFPS) {
             setMaxFPS(_desiredMaxFPS);
         }
-        loadSourceURL();
-        _webSurface->resume();
+
+        if (isWebContent()) {
+            _webSurface = DependencyManager::get<OffscreenQmlSurfaceCache>()->acquire(QML);
+            _webSurface->getRootItem()->setProperty("url", _url);
+            _webSurface->getRootItem()->setProperty("scriptURL", _scriptURL);
+        } else {
+            _webSurface = DependencyManager::get<OffscreenQmlSurfaceCache>()->acquire(_url);
+            setupQmlSurface();
+        }
+        _webSurface->getSurfaceContext()->setContextProperty("globalPosition", vec3toVariant(getPosition()));
         _webSurface->resize(QSize(_resolution.x, _resolution.y));
-        _webSurface->getRootItem()->setProperty("url", _url);
-        _webSurface->getRootItem()->setProperty("scriptURL", _scriptURL);
+        _webSurface->resume();
     });
 
     auto selfOverlayID = getOverlayID();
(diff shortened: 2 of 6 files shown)


interface/src/Application.cpp
```diff
@@ -2369,6 +2369,7 @@ void Application::initializeUi() {
 
     // Pre-create a couple of Web3D overlays to speed up tablet UI
     auto offscreenSurfaceCache = DependencyManager::get<OffscreenQmlSurfaceCache>();
+    offscreenSurfaceCache->reserve(TabletScriptingInterface::QML, 1);
```
jherico: pre-load the tablet QML at application startup.

interface/src/ui/overlays/Web3DOverlay.cpp
```diff
@@ -97,6 +98,10 @@ Web3DOverlay::~Web3DOverlay() {
     }
 }
 
+void Web3DOverlay::rebuildWebSurface() {
```
jherico: never change the QML root content...  always release and reload the tablet, so that the QML surface cache functions properly.

interface/src/ui/overlays/Web3DOverlay.cpp
```diff
@@ -187,88 +198,61 @@ void Web3DOverlay::update(float deltatime) {
...
-    if (!_webSurface) {
-        return;
-    }
-
-    QUrl sourceUrl(_url);
-    if (sourceUrl.scheme() == "http" || sourceUrl.scheme() == "https" ||
-        _url.toLower().endsWith(".htm") || _url.toLower().endsWith(".html")) {
-
-        _webSurface->setBaseUrl(QUrl::fromLocalFile(PathUtils::resourcesPath() + "/qml/"));
-        _webSurface->load("Web3DOverlay.qml");
```
jherico: surfaces acquired from the QML surface cache should never have `load` called on them.

interface/src/ui/overlays/Web3DOverlay.cpp
```diff
@@ -187,88 +198,61 @@ void Web3DOverlay::update(float deltatime) {
...
-        _webSurface->resume();
-        _webSurface->getRootItem()->setProperty("url", _url);
-        _webSurface->getRootItem()->setProperty("scriptURL", _scriptURL);
-
-    } else {
-        _webSurface->setBaseUrl(QUrl::fromLocalFile(PathUtils::resourcesPath()));
-        _webSurface->load(_url, [&](QQmlContext* context, QObject* obj) {});
-        _webSurface->resume();
+void Web3DOverlay::setupQmlSurface() {
```
jherico: no idea why github can't figure this part of the diff out.

interface/src/ui/overlays/Web3DOverlay.cpp
```diff
@@ -594,11 +578,25 @@ QVariant Web3DOverlay::getProperty(const QString& property) {
...
-        });
+    if (url != _url) {
+        bool wasWebContent = isWebContent();
+        _url = url;
+        if (_webSurface) {
+            if (wasWebContent && isWebContent()) {
+                // If we're just targeting a new web URL, then switch to that without messing around
+                // with the underlying QML
+                AbstractViewStateInterface::instance()->postLambdaEvent([this, url] {
+                    _webSurface->getRootItem()->setProperty("url", _url);
```
jherico: changing one URL to another is fine within a web QML surface...

interface/src/ui/overlays/Web3DOverlay.cpp
```diff
@@ -594,11 +578,25 @@ QVariant Web3DOverlay::getProperty(const QString& property) {
...
+                    _webSurface->getRootItem()->setProperty("scriptURL", _scriptURL);
+                });
+            } else {
+                // If we're switching to or from web content, or between different QML content
+                // we need to destroy and rebuild the entire QML surface
+                AbstractViewStateInterface::instance()->postLambdaEvent([this, url] {
+                    rebuildWebSurface();
```
jherico: but don't try to change the root content for any other condition.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3617/HighFidelity-Beta-PR11458-40c42d35a5ba6560e21187bad60d5255df78d49b.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3617/HighFidelity-Beta-PR11458-40c42d35a5ba6560e21187bad60d5255df78d49b.exe

interface/src/ui/overlays/Web3DOverlay.cpp
```diff
@@ -594,11 +578,25 @@ QVariant Web3DOverlay::getProperty(const QString& property) {
...
-    _url = url;
-    if (_webSurface) {
-        AbstractViewStateInterface::instance()->postLambdaEvent([this, url] {
-            loadSourceURL();
-        });
+    if (url != _url) {
+        bool wasWebContent = isWebContent();
+        _url = url;
+        if (_webSurface) {
+            if (wasWebContent && isWebContent()) {
```
ZappoMan: This looks weird. Does isWebContent() change value in these 3 lines of code?

ZappoMan: Wait. Never mind I see that you change _url between these calls.

cain-kilgore:
Testing.

cain-kilgore:
# Test Failed
Test | Result
--- | ---
Launch the application and load the tablet. Compared to the current master build the tablet should appear much faster and not trigger a freeze of the application. There may be some small stutter, but it should be much less than in master. | Pass
WebOverlays should be tested with actual web content, and the tablet should be thoroughly tested as well. Any existing tablet applications should be re-tested as well to ensure their UI still functions. | Pass

The keyboard in Landscape mode is broken - it will render but it does not fully fill up the bottom half of the tablet. See below image. Confirmed not happening in master.
d728c4d6-70c1-44a6-91a0-dd964613d1d1

jherico:
@cain-kilgore Can you please confirm the exact steps you're using to trigger this, becuase I _AM_ seeing this in build 7180 of master.

cain-kilgore:
@jherico: 

Opening the Tablet, Creating a Box in the Create Mode & attempting to input one of the property TextFields results in the keyboard showing up and causing the visual glitch.

jherico:
@cain-kilgore I just did those steps in dev-welcome with the 7180 build and I saw the exact same glitch in the keyboard.  I'm going to have to say it's not related to this PR.

PR #11469, opened 2017-09-27
Don't constantly reload animGraph if URL didn't change

SamGondelman:
Possible fix for [FB7882](https://highfidelity.fogbugz.com/f/cases/7882/T-Pose-when-Switching-from-HMD-to-Desktop).

`MyAvatar::postUpdate` polls for the initialization of the skeleton model rig animation node and calls `initAnimGraph()` if it isn't loaded.  That calls `_skeletonModel->getRig().initAnimGraph(graphUrl)`, which would constantly reset the _animLoader.  If the _animLoader hadn't finished downloading, the success signal would never be fired, so you'd be stuck in a T-pose and the logs would fill with `Starting request for: file:///C:/Users/[User]/src/hifi/build/interface/Release/resources/avatar/avatar-animation.json`.

Now, the _animLoader is only reset if the url has changed, so the success signal will have time to fire.

Test plan:
- With a Rift, repeatedly put on and take off your headset. maybe 20 or so times.  At no point should your avatar be stuck in a T-Pose.
- Your logs shouldn't fill up with `Starting request for: file:///C:/Users/[User]/src/hifi/build/interface/Release/resources/avatar/avatar-animation.json`.

### DIFF
--- a/libraries/animation/src/Rig.cpp
+++ b/libraries/animation/src/Rig.cpp
@@ -1619,28 +1619,30 @@ void Rig::updateFromControllerParameters(const ControllerParameters& params, flo
 }
 
 void Rig::initAnimGraph(const QUrl& url) {
-    _animGraphURL = url;
+    if (_animGraphURL != url) {
+        _animGraphURL = url;
 
-    _animNode.reset();
+        _animNode.reset();
 
-    // load the anim graph
-    _animLoader.reset(new AnimNodeLoader(url));
-    connect(_animLoader.get(), &AnimNodeLoader::success, [this](AnimNode::Pointer nodeIn) {
-        _animNode = nodeIn;
-        _animNode->setSkeleton(_animSkeleton);
+        // load the anim graph
+        _animLoader.reset(new AnimNodeLoader(url));
+        connect(_animLoader.get(), &AnimNodeLoader::success, [this](AnimNode::Pointer nodeIn) {
+            _animNode = nodeIn;
+            _animNode->setSkeleton(_animSkeleton);
 
-        if (_userAnimState.clipNodeEnum != UserAnimState::None) {
-            // restore the user animation we had before reset.
-            UserAnimState origState = _userAnimState;
-            _userAnimState = { UserAnimState::None, "", 30.0f, false, 0.0f, 0.0f };
-            overrideAnimation(origState.url, origState.fps, origState.loop, origState.firstFrame, origState.lastFrame);
-        }
+            if (_userAnimState.clipNodeEnum != UserAnimState::None) {
+                // restore the user animation we had before reset.
+                UserAnimState origState = _userAnimState;
+                _userAnimState = { UserAnimState::None, "", 30.0f, false, 0.0f, 0.0f };
+                overrideAnimation(origState.url, origState.fps, origState.loop, origState.firstFrame, origState.lastFrame);
+            }
 
-        emit onLoadComplete();
-    });
-    connect(_animLoader.get(), &AnimNodeLoader::error, [url](int error, QString str) {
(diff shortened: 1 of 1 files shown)


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3660/HighFidelity-Beta-PR11469-306cf883feb5851a17c124356cc7c3c37edd8eb5.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3660/HighFidelity-Beta-PR11469-306cf883feb5851a17c124356cc7c3c37edd8eb5.exe

libraries/animation/src/Rig.cpp
```diff
@@ -1619,28 +1619,30 @@ void Rig::updateFromControllerParameters(const ControllerParameters& params, flo
...
-        emit onLoadComplete();
-    });
-    connect(_animLoader.get(), &AnimNodeLoader::error, [url](int error, QString str) {
-        qCCritical(animation) << "Error loading" << url.toDisplayString() << "code = " << error << "str =" << str;
-    });
+            emit onLoadComplete();
+        });
+        connect(_animLoader.get(), &AnimNodeLoader::error, [url](int error, QString str) {
+            qCCritical(animation) << "Error loading" << url.toDisplayString() << "code = " << error << "str =" << str;
+        });
```
AndrewMeadows: Each `connect()` creates an entry somewhere in Qt data structures, right?  And those connections persist until: (a) they are explicitly disconnected, (b) the object instance is deleted (in this case the `Rig`), or (c) they are replaced by a duplicate connection IFF `Qt::ConectionType type = Qt::UniqueConnection` was passed as the final optional argument to `connect()` (not done here).

So, not that this is a new problem being introduced by this PR but... doesn't this `connect()` create a "
memory leak" whenever `initAnimGraph()` is called more than once?  Not to mention a pileup of slots that would all get invoked should the signal ever emit?  Is this a real problem or am I misunderstanding something?

If this is a real problem then I wonder: are there other places in our codebase where we're piling up connections?

SamGondelman: https://stackoverflow.com/questions/10570857/are-signals-in-qt-automatically-disconnected-when-one-of-the-class-is-deleted

AndrewMeadows: Right, which is the case (b) I mentioned above: the connections are cleaned up whenever the Rig instance is deleted, but not until then.  Which means: whenever we call `Rig::initAnimGraph()` more than once on a particular `Rig` instance we introduce... not a true "memory leak" (hence the reason I used quotes above) but a "resource leak": we leave a connection that is not useful and occupies memory.

SamGondelman: I don't think so.  The signal is cleaned up when either side of the connection is deleted, not just the Rig instance.  We always reset _animLoader when we create a new connection.  Hence, the old signal is always cleaned up.

anshuman64:
Hey, I'm getting a crash on load with this PR. Bugsplat sent.

SamGondelman:
@anshuman64 That was probably the build system issue.  Hopefully it'll be fine after rebuild.

build this please

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3683/HighFidelity-Beta-PR11469-306cf883feb5851a17c124356cc7c3c37edd8eb5.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3683/HighFidelity-Beta-PR11469-306cf883feb5851a17c124356cc7c3c37edd8eb5.exe

conklin94122:
QA PASS

PR #11506, opened 2017-10-02
Fix and improve laser pointer locking

SamGondelman:
- Reduce lookups by including the necessary locks in RayPicks and LaserPointers
- Move RayPick locking earlier in RayPickManager::update to avoid a crash

Test plan:
- Laser pointers (teleport, hand lasers) should work as expected.
- Shouldn't crash in RayPickManager::update, although I couldn't actually repro this crash (reported by @sethalves)

### DIFF
--- a/interface/src/raypick/LaserPointerManager.cpp
+++ b/interface/src/raypick/LaserPointerManager.cpp
@@ -31,7 +31,6 @@ void LaserPointerManager::enableLaserPointer(const QUuid uid) {
     QReadLocker lock(&_containsLock);
     auto laserPointer = _laserPointers.find(uid);
     if (laserPointer != _laserPointers.end()) {
-        QWriteLocker laserLock(laserPointer.value()->getLock());
         laserPointer.value()->enable();
     }
 }
@@ -40,7 +39,6 @@ void LaserPointerManager::disableLaserPointer(const QUuid uid) {
     QReadLocker lock(&_containsLock);
     auto laserPointer = _laserPointers.find(uid);
     if (laserPointer != _laserPointers.end()) {
-        QWriteLocker laserLock(laserPointer.value()->getLock());
         laserPointer.value()->disable();
     }
 }
@@ -49,7 +47,6 @@ void LaserPointerManager::setRenderState(QUuid uid, const std::string& renderSta
     QReadLocker lock(&_containsLock);
     auto laserPointer = _laserPointers.find(uid);
     if (laserPointer != _laserPointers.end()) {
-        QWriteLocker laserLock(laserPointer.value()->getLock());
         laserPointer.value()->setRenderState(renderState);
     }
 }
@@ -58,7 +55,6 @@ void LaserPointerManager::editRenderState(QUuid uid, const std::string& state, c
     QReadLocker lock(&_containsLock);
     auto laserPointer = _laserPointers.find(uid);
     if (laserPointer != _laserPointers.end()) {
-        QWriteLocker laserLock(laserPointer.value()->getLock());
         laserPointer.value()->editRenderState(state, startProps, pathProps, endProps);
     }
 }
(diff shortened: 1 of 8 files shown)


interface/src/raypick/RayPickManager.cpp
```diff
@@ -114,7 +115,6 @@ void RayPickManager::update() {
             }
         }
 
-        QWriteLocker lock(_rayPickLocks[uid].get());
```
SamGondelman: This lock should have been higher up to avoid reading parameters on the main thread that are modifiable from script threads.

YuppyPlays:
Awaiting build for QA.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3816/HighFidelity-Beta-PR11506-97b09d680ca7bbff215a7765d082f909deed5d73.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3816/HighFidelity-Beta-PR11506-97b09d680ca7bbff215a7765d082f909deed5d73.exe

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

Laser pointers (teleport, hand lasers) should work as expected. **Check**
Shouldn't crash in RayPickManager::update, although I couldn't actually repro this crash **Did not experience any crashes whilst going through basic functionality of the lasers and tablet as well as messing around with the lasers trying to break them**

**Notes: I am unsure if this is standard behaviour and non-PR related, but I have found if you point the laser in the distance at an object, then very quickly/sharpy move the controller to aim the laser close up, the ball which usually collides at the end of the laser is much larger and then quickly shrinks to the appropriate size.**

SamGondelman:
@YuppyPlays yeah that happens in master too as a side-effect of how laser pointers are updated

interface/src/raypick/LaserPointerManager.cpp
```diff
@@ -26,46 +25,50 @@ QUuid LaserPointerManager::createLaserPointer(const QVariant& rayProps, const La
...
 
 void LaserPointerManager::enableLaserPointer(const QUuid uid) {
     QReadLocker lock(&_containsLock);
-    if (_laserPointers.contains(uid)) {
-        QWriteLocker laserLock(_laserPointerLocks[uid].get());
-        _laserPointers[uid]->enable();
+    auto laserPointer = _laserPointers.find(uid);
+    if (laserPointer != _laserPointers.end()) {
+        QWriteLocker laserLock(laserPointer.value()->getLock());
+        laserPointer.value()->enable();
```
AndrewMeadows: I haven't audited the rest of the code but I'm wondering if it would be possible: rather than lock the LaserPointer in this external context, change the implementation of LaserPointer::enable() (and similar methods) to implicitly lock.  Like this:
```
void LaserPointer::enable() {
    getLock();
    // do enable stuff here
}
```
Such a pattern would simplify the code here in the `LaserPointerManager` context.

SamGondelman: I will do that.

ZappoMan:
> Notes: I am unsure if this is standard behaviour and non-PR related, but I have found if you point the laser in the distance at an object, then very quickly/sharpy move the controller to aim the laser close up, the ball which usually collides at the end of the laser is much larger and then quickly shrinks to the appropriate size.

@YuppyPlays - can you file a fogbugz with a repro case for this bug?

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3925/HighFidelity-Beta-PR11506-6b02cbb9112923e90fd92baa91304f1bbd3ebe93.dmg

PR #11522, opened 2017-10-04
workaround for bad physics polling on login

AndrewMeadows:


AndrewMeadows:
This works around a bug where sometimes on login it takes a rather long time before your avatar will collide with the objects nearby.  The interface will poll for nearby objects in an attempt to figure out if there are enough nearby objects loaded for safe physics simulation (you don't want to simulate the cups on the table if the table hasn't loaded yet).  The list of objects scanned in the poll appear to be incorrect and violates some of the assumptions made in the scan but rather than dig down and fix it correctly now I decided it would be better to provide a quick **workaround** for the short term.

**Test Plan:**

**(1)** Using an interface from hifi/master (NOT from this PR) log into some other domain and then teleport to dev-welcome.
--> notice how long it takes for your avatar to be able to move or walk on the floor
If this takes too long try this workaround: spin your avatar around one full rotation and then wait three seconds.  Hopefully that will unjam the blocked logic.

**(2)** Perform the same experiment as **(1)** using the interface from this PR
--> notice how long it takes for your avatar to be able to move or walk on the floor.  It should be shorter time than in (1).

**(3)** The code change in this PR is rather focused on this "is physics ready?" poll at login so no other behaviors should change.  However the rules require me to say:  This PR changes code around "login".  Please run all existing acceptance tests that relate to "login".

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3876/HighFidelity-Beta-PR11522-850cb3b24d861d1b3df31868d506d79a21150a51.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3876/HighFidelity-Beta-PR11522-850cb3b24d861d1b3df31868d506d79a21150a51.exe

AndrewMeadows:
Don't merge this yet.  I'm now working on a more correct solution that isn't a workaround and it doesn't look too hard.

interface/src/Application.cpp
```diff
@@ -5855,11 +5855,18 @@ bool Application::nearbyEntitiesAreReadyForPhysics() {
         bool result = true;
         foreach (EntityItemPointer entity, entities) {
             if (entity->shouldBePhysical() && !entity->isReadyToComputeShape()) {
-                static QString repeatedMessage =
-                    LogHandler::getInstance().addRepeatedMessageRegex("Physics disabled until entity loads: .*");
-                qCDebug(interfaceapp) << "Physics disabled until entity loads: " << entity->getID() << entity->getName();
-                // don't break here because we want all the relevant entities to start their downloads
-                result = false;
+                // BUG: the findEntities() query above is sometimes returning objects that don't actually overlap
+                // TODO: investigate and fix findQueries() but in the meantime...
```
ZappoMan: `findQueries`?

ZappoMan:
There are several other places in the code where we use `findEntities(AABox...)` if this is your fix... we really need to do the same filter in those places too... you could actually do the filter inside of `EntityTree::findEntities(AABox...)` and then it would still be a hack, but at least would fix all cases.

Regardless... it looks to me like the problem is probably in `EntityTreeElement::getEntitites(AABox...)`

```
void EntityTreeElement::getEntities(const AACube& cube, QVector<EntityItemPointer>& foundEntities) {
    forEachEntity([&](EntityItemPointer entity) {
        bool success;
        AABox entityBox = entity->getAABox(success);

        // If the entities AABox touches the search cube then consider it to be found
        if (!success || entityBox.touches(cube)) {
            foundEntities.push_back(entity);
        }
    });
}
```

In particular if `entity->getAABox(success);` returns false... then the entity will be included in the list... that seems very suspicious to me. @sethalves might have more info on this....

AndrewMeadows:
@zappoman, I came to similar conclusions after looking into it in more detail.  While auditing the code I realized that our editing tools (edit.js and other scripts) use the existing  `EntityTreeElement::getEntities()` implementations to find entities.  Due to fragility and sensitivity of those tools I don't want to mess with the existing `getEntities()` implementations.  In particular, scripted scans will currently find all child entities in an `EntityTreeElement` whose parents cannot be found (because `!success`) which could include objects far away but also truly nearby objects that would otherwise be invisible (and could not be edited or "fixed").

Rather than change the old scans' behaviors I have decided to implement my own correct scan, but in a more general way.  I'll submit that in a different PR.

Closing this one for being the wrong solution.

PR #11533, opened 2017-10-05
Fix model loading for real this time

SamGondelman:
- cherry pick the performance fix that went into RC-56 (which would break model loading on master)
- start moving any work that captures values from gameplay entities out of doRenderUpdateAsynchronousTyped and into doRenderUpdateSynchronousTyped, where it belongs
- move around and add some locking to support ^ that ^

This is part 1 of work that needs to be done to fix doRenderUpdateAsynchronousTyped so that it doesn't capture gameplay values in the render thread.  There is more work to be done, but hopefully this doesn't break anything new.

Test plan:
- Download a content set with a good number of model entities. Make sure there are multiple models with exactly the same fbx file and at least some models that have parent model entities. The dev-welcome content is pretty good, but it would be good to test other content sets too.
- All model entities should load normally. Reload your content (Edit -> Reload Content) while looking at one part of the domain and then turn to look behind you. No models should be missing once all your downloads finish.
- If model A and model B have the same model file and you look at only model A (model B is out of view and at least several meters away), reload content, and wait a while so that everything has loaded, when you turn to look at model B it should appear.
- Create some of each type of entity.  Modify their properties (color, visibility, textures, model url, dimensions, etc.).  For each type, the changes should take effect as expected.  Pay special attention to models, shapes, and web entities.
- Compare performance to identical content with master right now. You should see noticeably better sim rate/game rate in this PR.

### DIFF
--- a/libraries/entities-renderer/src/RenderableEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableEntityItem.cpp
@@ -329,25 +329,27 @@ bool EntityRenderer::needsRenderUpdateFromEntity(const EntityItemPointer& entity
     return false;
 }
 
-void EntityRenderer::doRenderUpdateAsynchronous(const EntityItemPointer& entity) {
-    auto transparent = isTransparent();
-    if (_prevIsTransparent && !transparent) {
-        _isFading = false;
-    }
-    _prevIsTransparent = transparent;
+void EntityRenderer::doRenderUpdateSynchronous(const ScenePointer& scene, Transaction& transaction, const EntityItemPointer& entity) {
+    withWriteLock([&] {
+        auto transparent = isTransparent();
+        if (_prevIsTransparent && !transparent) {
+            _isFading = false;
+        }
+        _prevIsTransparent = transparent;
 
-    bool success = false;
-    auto bound = entity->getAABox(success);
-    if (success) {
-        _bound = bound;
-    }
-    auto newModelTransform = entity->getTransformToCenter(success);
-    if (success) {
-        _modelTransform = newModelTransform;
-    }
+        bool success = false;
+        auto bound = entity->getAABox(success);
+        if (success) {
+            _bound = bound;
+        }
+        auto newModelTransform = entity->getTransformToCenter(success);
+        if (success) {
+            _modelTransform = newModelTransform;
+        }
 
-    _moving = entity->isMovingRelativeToParent();
-    _visible = entity->getVisible();
+        _moving = entity->isMovingRelativeToParent();
+        _visible = entity->getVisible();
+    });
 }
 
 void EntityRenderer::onAddToScene(const EntityItemPointer& entity) {
--- a/libraries/entities-renderer/src/RenderableEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableEntityItem.cpp
@@ -291,18 +291,6 @@ void EntityRenderer::updateInScene(const ScenePointer& scene, Transaction& trans
     });
 }
 
-void EntityRenderer::update(const ScenePointer& scene, Transaction& transaction) {
-    if (!isValidRenderItem()) {
-        return;
-    }
-
-    if (!needsUpdate()) {
-        return;
-    }
-
-    doUpdate(scene, transaction, _entity);
-}
-
 //
 // Internal methods
 //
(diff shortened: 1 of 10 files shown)


Nex-Pro:
Awaiting Win Build

libraries/entities-renderer/src/RenderableEntityItem.cpp
```diff
@@ -346,25 +329,27 @@ bool EntityRenderer::needsRenderUpdateFromEntity(const EntityItemPointer& entity
...
 }
 
-void EntityRenderer::doRenderUpdateAsynchronous(const EntityItemPointer& entity) {
-    auto transparent = isTransparent();
-    if (_prevIsTransparent && !transparent) {
-        _isFading = false;
-    }
-    _prevIsTransparent = transparent;
+void EntityRenderer::doRenderUpdateSynchronous(const ScenePointer& scene, Transaction& transaction, const EntityItemPointer& entity) {
+    withWriteLock([&] {
```
SamGondelman: doRenderUpdateAsynchronous (which happens on the render thread) should never be setting values that can be accessed from the main thread.  This work is moved to doRenderUpdateSynchronous (on the main thread) and surrounded by a write lock.

As a general rule, doRenderUpdateAsynchronous should surround any work it does with a read lock, and doRenderUpdateSynchronous should surround any work it does with a write lock, if they are modifying values that are shared between threads.

libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
```diff
@@ -88,31 +88,33 @@ void ShapeEntityRenderer::doRenderUpdateSynchronousTyped(const ScenePointer& sce
...
-    _dimensions = entity->getDimensions();
-    _orientation = entity->getOrientation();
+        if (_shape == entity::Sphere) {
+            _modelTransform.postScale(SPHERE_ENTITY_SCALE);
+        }
 
-    if (_shape == entity::Sphere) {
-        _modelTransform.postScale(SPHERE_ENTITY_SCALE);
-    }
+        _modelTransform.postScale(_dimensions);
```
SamGondelman: Values like _shape and _modelTransform should not be modified from the render thread, and need to be surrounded by a lock.

libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
```diff
@@ -126,15 +128,16 @@ void ShapeEntityRenderer::doRender(RenderArgs* args) {
...
-    batch.setModelTransform(_modelTransform); // use a transform with scale, rotation, registration point and translation
-
+    GeometryCache::Shape geometryShape;
     bool proceduralRender = false;
-    glm::vec4 outColor = _color;
+    glm::vec4 outColor;
     withReadLock([&] {
+        geometryShape = MAPPING[_shape];
+        batch.setModelTransform(_modelTransform); // use a transform with scale, rotation, registration point and translation
+        outColor = _color;
```
SamGondelman: When shared values are read in doRender, they need to be surrounded by a read lock.

libraries/render-utils/src/Model.cpp
```diff
@@ -267,6 +267,11 @@ void Model::updateRenderItems() {
     });
 }
 
+void Model::setRenderItemsNeedUpdate() {
+    _renderItemsNeedUpdate = true;
+    emit requestRenderUpdate();
+}
```
SamGondelman: Let the Model communicate to the ModelEntityRenderer that we need an update.

Nex-Pro:
Awaiting fresh

Nex-Pro:
Fresher even

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3923/HighFidelity-Beta-PR11533-a6b7578c3c923578aa60133b4a1695320edce081.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3923/HighFidelity-Beta-PR11533-a6b7578c3c923578aa60133b4a1695320edce081.exe

Nex-Pro:
In progress

Nex-Pro:
### PASS

Content is in-view immediately as discussed.

PR #11551, opened 2017-10-09
Model asks for renderUpdate if animating

SamGondelman:
Models need to request constant renderUpdates if they are animating (this was being done in render() before, but we made Models only meta, so they weren't calling that anymore).

Test Plan:
- Models should animate without you having to move your head.
- Performance smoke test with content that has several animating models.

### DIFF
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -1244,6 +1244,7 @@ void ModelEntityRenderer::doRenderUpdateSynchronousTyped(const ScenePointer& sce
             mapJoints(entity, model->getJointNames());
         }
         animate(entity);
+        emit requestRenderUpdate();
     }
 }
 


libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -1244,6 +1244,7 @@ void ModelEntityRenderer::doRenderUpdateSynchronousTyped(const ScenePointer& sce
             mapJoints(entity, model->getJointNames());
         }
         animate(entity);
+        emit requestRenderUpdate();
```
jherico: I would reproduce the logic from `RenderableEntityItem.cpp`, specifically `if (!_renderUpdateQueued && needsRenderUpdate()) {`.  It's possible that in the future if we are smart enough in `needsRenderUpdate` we'll notice that the animation frame won't change by the next render frame and we can skip the update.

cozza13:
looked over the shoulder of Samuel. I am happy to QA approve this as there will be another PR against stable

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/3974/HighFidelity-Beta-PR11551-873ae9b9d650244fd1bb303492759c92912b3ca0.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/3974/HighFidelity-Beta-PR11551-873ae9b9d650244fd1bb303492759c92912b3ca0.exe

PR #11590, opened 2017-10-13
Fix tablet laser ignoreOverlay

SamGondelman:
There was a bug in the RayPick caching mechanism that didn't take into account identical PickRays with different ignore/include combinations.  Also, the tablet doesn't necessarily exist when the controller modules are created, so the laser on the tablet screen wasn't properly ignoring the back of the tablet, and it seems like there was z-fighting with the intersection.  This should maybe help with [FB8076](https://highfidelity.fogbugz.com/f/cases/8076/Laser-scrolling-on-tablet-is-nigh-impossible) and [FB7965](https://highfidelity.fogbugz.com/f/cases/7965/Laser-pressing-or-tablet-buttons-is-unreliable).

Test plan:
- In HMD, use the hand lasers on the tablet.  It should be much easier to scroll and press buttons.  Try with multiple tablet orientations/hand directions, and in domains with lots of content.  Reload scripts a few times, laser interaction should still work.

Known issue:
- When you move your laser off of the tablet screen, it goes through the tablet for a frame.


YuppyPlays:
Awaiting build for QA.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4131/HighFidelity-Beta-PR11590-b0dc2313c51e0feeab2822d9b6a33d8cdd3a0fdc.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4131/HighFidelity-Beta-PR11590-b0dc2313c51e0feeab2822d9b6a33d8cdd3a0fdc.exe

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:


In HMD, use the hand lasers on the tablet. It should be much easier to scroll and press buttons. Try with multiple tablet orientations/hand directions, and in domains with lots of content.

**Notes: A few quirks I noticed whilst testing which may or may not be considered bugs..
 - The lasers; if you hold but are not holding enough to "Click" on the tablet and you move the laser off the tablet and back on, it will create the end-laser-sphere to become huge colliding on the edge of the tablet, unsure if this is already fixed in another fogbugz case.
 - Although I could not reproduce consistently, I noticed if I held the laser button (trigger) down then click "Create" but do not release fully (so the laser is still being rendered) - it at random sometimes starts to not collide with the tablet, but rather going through it.
 - Probably not a bug at all but worth noting, if you're standing in a Zone in "Create" mode and you go to use the laser on the tablet, it will go through tablet and assume you're trying to grab the Zone entity.
 - This one could just be a Vive thing or even just "My Vive" thing (though not noticed in the past nor in any game other than this instance) - when I held the trigger down, but not "Clicking", it would activate lasers... then I noticed it seemingly would not do that and would only register if I clicked the trigger fully, but the other controller then would have this trait and seems to just "Shift" in behaviour. I don't know how to repro other than just keep using the laser to press things on the tablet and notice eventually using the "Not entirely clicked in" method suddenly stops working and only full clicks render.**

Please let me know if any more information is required, happy to explain with video if I can repro and if required.

SamGondelman:
Dante pointed out that this result shouldn't actually be used, so I'm going to investigate more.

SamGondelman:
@YuppyPlays I've got a plan to fix the end sphere sizing problem.  The other things sound like controller module issues, which I'll talk to @druiz17 about, and maybe some hardware problems.

Anyways, this is ready for testing again once it builds.

interface/src/raypick/RayPickManager.h
```diff
@@ -48,11 +84,18 @@ class RayPickManager {
...
+    bool checkAndCompareCachedResults(QPair<glm::vec3, glm::vec3>& ray, RayPickCacheOverlays& cache, RayPickResult& res, const RayCacheKeyOverlays& key);
+    bool checkAndCompareCachedResults(QPair<glm::vec3, glm::vec3>& ray, RayPickCacheEntities& cache, RayPickResult& res, const RayCacheKeyEntities& key);
+    bool checkAndCompareCachedResults(QPair<glm::vec3, glm::vec3>& ray, RayPickCacheHUD& cache, RayPickResult& res, const RayPickFilter::Flags& key);
+
+    void cacheResult(const bool intersects, const RayPickResult& resTemp, const RayCacheKeyOverlays& key, RayPickResult& res, QPair<glm::vec3, glm::vec3>& ray, RayPickCacheOverlays& cache);
+    void cacheResult(const bool intersects, const RayPickResult& resTemp, const RayCacheKeyEntities& key, RayPickResult& res, QPair<glm::vec3, glm::vec3>& ray, RayPickCacheEntities& cache);
+    void cacheResult(const bool intersects, const RayPickResult& resTemp, const RayPickFilter::Flags& key, RayPickResult& res, QPair<glm::vec3, glm::vec3>& ray, RayPickCacheHUD& cache);
```
SamGondelman: Is there a cleaner way to do this?  I don't want to have to repeatedly convert between QVector<EntityItemID>/QVector<OverlayID> and QVector<QUuid>, but there's a lot of repeated code as it is right now.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4175/HighFidelity-Beta-PR11590-991a16a7b6a132b1ade78c6486cdaaccaf650461.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4175/HighFidelity-Beta-PR11590-991a16a7b6a132b1ade78c6486cdaaccaf650461.exe

interface/src/raypick/RayPickManager.h
SamGondelman: austin has a change that will make this simpler

PR #11591, opened 2017-10-13
Add some Quat and Uuid JavaScript helper properties

ctrlaltdavid:


ctrlaltdavid:
**Test Plan**

Check method values:
- Open the script debug window (Developer > Script Log (HMD Friendly)...) and run the attached testMethods.js.
- Each line will show the value from the new method (e.g., Quat.ZERO) or related code followed by the expected value. The two values should be identical.

Test the following scripts which have been updated:
- Grabbing wearables. [attachedEntitiesManager.js]
- DoppleGanger app from the Marketplace. [attachedEntitiesManager.js]
- away.js
- Unparenting entity using Edit > Unparent Entity and Ctrl-Shift-P. [edit.js]
- /system/nameTags.js
- Controller display. [controllerDisplay.js]
- Hand controllers operations. [controller modules]
- Entity camera. [entityCameraTool.js]
- Entity selection. [entitySelectionTool.js]
- Tablet opening and grabbing. [WebTablet.js]
- Tutorial that creates a tetherball. [createTetherBallStick.js]

No test for overlayUtils.js because its function is not currently used in scripts.

ctrlaltdavid:
QA ready

Nex-Pro:
Awaiting build

scripts/system/libraries/entitySelectionTool.js
```diff
@@ -1636,7 +1508,7 @@ SelectionDisplay = (function() {
             position = SelectionManager.localPosition;
             registrationPoint = SelectionManager.localRegistrationPoint;
         } else {
-            rotation = Quat.fromPitchYawRollDegrees(0, 0, 0);
+            rotation = Quat.ZERO;
```
ZappoMan: hmmm - is this really equivalent???

scripts/system/libraries/entitySelectionTool.js
```diff
@@ -2692,7 +2564,7 @@ SelectionDisplay = (function() {
         var onBegin = function(event, pickRay, pickResult) {
             var properties = Entities.getEntityProperties(SelectionManager.selections[0]);
             initialProperties = properties;
-            rotation = (spaceMode === SPACE_LOCAL) ? properties.rotation : Quat.fromPitchYawRollDegrees(0, 0, 0);
+            rotation = (spaceMode === SPACE_LOCAL) ? properties.rotation : Quat.ZERO;
```
ZappoMan: ibid

scripts/system/libraries/entitySelectionTool.js
ctrlaltdavid: 0, 0, 0 degrees is a zero rotation, so yes ... see the attached script, which I forgot to attach but will attach now

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4134/HighFidelity-Beta-PR11591-fc4e440dc46dac74b12397a942ef29ffdb2bdfe5.dmg

ctrlaltdavid:
Both generate an object, { x: 0, y: 0, z: 0, w: 1 }

quat-zero

ctrlaltdavid:
Build this please

Nex-Pro:
@ctrlaltdavid Quat.ZERO nice addition I had previously assumed that existed and used it in one of my scripts !

Nex-Pro:
Come on Jenkins !

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4136/HighFidelity-Beta-PR11591-fc4e440dc46dac74b12397a942ef29ffdb2bdfe5.dmg

Nex-Pro:
J E N K I N S ! gah

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4136/HighFidelity-Beta-PR11591-fc4e440dc46dac74b12397a942ef29ffdb2bdfe5.exe

scripts/system/libraries/entitySelectionTool.js
Atlante45: This should really be called `IDENTITY`.

Atlante45:
@ctrlaltdavid I don't think we should have `Uuid.SELF`. it doesn't really have anything to do with UUIDs beyond the fact that Avatars use UUIDs.
And UUIDs are available in contexts where a self UUID doesn't make sense like in entity server scripts.

Something else I just noticed is that scripters shouldn't even have to know about `AVATAR_SELF_ID` they should exclusively be using `MyAvatar.sessionUUID`.
But I'm now realizing it's necessary in some cases, so I think we need a better solution here.
Scripters should have to know when to use `AVATAR_SELF_ID` and when to use `MyAvatar.sessionUUID`.

Nex-Pro:
Ok this is no longer QA ready so this is on hold.

ctrlaltdavid:
@Atlante45 I think Uuid.SELF is valid because it's a magic number you need to use in some situations instead of a "real" UUID. For example, Overlays.addOverlay()'s "parentID" property. Perhaps there's some other way that the JavaScript API could work but for the present, Uuid.SELF would seem a useful property.
I'll get a third opinion ... @thoys ?

Atlante45:
Yes, but that's a MyAvatar magic value, not a Uuid magic value. The values
in Vec3 are default generic values like ZERO, ONE, HALF, UNIT_X, etc they
are not specific magic values. It should be they same with UUID.

My greater point was that SELF shouldn't even be known to script users,
this should be a backend implementation detail.
I just understand why it's necessary for overlays, but it should be
transparent to the script users.
That's not related to this PR though.

If we do want to keep SELF in this PR it should be part of MyAvatar as
`MyAvatar.SELF_ID`

thoys:
I could see usages for SELF_ID, but I agree to move it out of the UUID API into MyAvatar. Maybe give it a longer name too, to encourage people to only use it when it is really necessary, like `MyAvatar.EXPORT_FRIENDLY_ID`. And where it is used now, try to use `MyAvatar.sessionUUID` instead.

MiladNazeri:
Just to clarify for the sake of future documentation,  is the reason why to use one over the other because  MyAvatar.sessionUuid changes between domains, which is why using AVATAR_SELF_ID(equal to {00000000-0000-0000-0000-000000000001}) is needed so that scripts running between domains can automatically reset the avatar entities ID property?

If that is the case, then is the only time this is needed would be when writing Interface only scripts that work between domains?

Overall, I also see this as being a useful constant to have after seeing multiple js scripts creating a variable for this.  I agree though about it being in MyAvatar, but giving it a safe name to reflect why you would use it.  If it is only used for Interface scripts, maybe that could be a part of the naming?  INTERFACE_AUTO-UPDATE_ID?

ctrlaltdavid:
The UUID magic value {00000000-0000-0000-0000-000000000001} is used in a number of situations:

C++
- Avatar: The magic value is replaced by the sessions value when processing avatar entities' properties.
- EntityEditPacketSender: The avatar session ID is replaced by the magic value when writing JSON and vice versa when processing JSON.
- EntityItem: The magic value is replaced by the sessions value when processing and sending entity properties. The magic value is also checked against as an alternative to the session UUID when doing some physics stuff.
- AvatarManager: The magic value is also checked against as an alternative to the session UUID when finding an avatar for given UUID.
- Application: The magic value is also checked against as an alternative to the session UUID when exporting entities. And  the value is used as the UUID when exporting avatar entities.
- InterfaceParentFinder: The magic value is also checked against an alternative to the session UUID when finding the parent of an item with a given UUID.
- SpatiallyNestable: The magic value is also checked against an alternative to the session UUID in various operations.

JavaScript:
- Some developer test scripts.
- Many /system/controllers/ files, e.g., for anchoring laser pointer to avatar or grabbing an entity.
- The WebTablet.
- away.js to anchor the "paused" image.
- The  Shapes app to attach the UI overlays and lasers to the avatar's hands.
- Some Marketplace scripts.

Note: If MyAvatar.sessionUUID is used in the Shapes app, for example, the UI disappears if the domain stops (which causes the MyAvatar.sessionUUID to change to null) and cannot be reattached. Presumably the WebTablet also needs to use the magic value in order to be able to display when not connected to a domain. And similarly for the other system scripts that currently use it.

From SpaticallyNestable.cpp: "// In practice, the AVATAR_SELF_ID should only be used for local-only objects."


So for this PR I propose to move Self.UUID to MyAvatar.SELF_ID.

@Atlante45 @thoys @MiladNazeri

Atlante45:
Yes, I don't think we should address the presence of that ID in scripts in this PR.
But centralizing its use will make things easier when we do address as it'll be easier to locate the uses to address them.

In the meantime, if we do add this to the docs, we should make it clear that this ID is most likely not the final design and very likely to change in the future.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4204/HighFidelity-Beta-PR11591-8d4e979dc8bc75c7ebafdb31e5ae1a6daf3a9682.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4204/HighFidelity-Beta-PR11591-8d4e979dc8bc75c7ebafdb31e5ae1a6daf3a9682.exe

scripts/system/libraries/entitySelectionTool.js
jherico: concur....  a `ZERO` quat would imply `{ w: 0, x: 0, y: 0, z: 0 }`, which is not the same as an identity quaternion of `{ w: 1, x: 0, y: 0, z: 0 }`

interface/src/avatar/MyAvatar.h
```diff
@@ -826,6 +828,9 @@ public slots:
 
     // height of user in sensor space, when standing erect.
     ThreadSafeValueCache<float> _userHeight { DEFAULT_AVATAR_HEIGHT };
+
+    const QString SELF_UUID() { return SELF_ID; }
```
Atlante45: Coding standard: This should be `getSelfUUID`, and to stay consistent with the naming of Avatar IDs, it should actually be called `getSelfID`.

The `const` on the return value does not do anything here since we return by value.
The function itself should be `const` and the return be either a value or a const ref.

IMO, this function should also be public and the definition should look like this:
```cpp
Q_PROPERTY(QUuid SELF_ID READ getSelfID CONSTANT)
const QUuid& getSelfID() const { return AVATAR_SELF_ID; }
```

The script engine takes care of the Uuid conversion for us, and this we stay consistent with the typing.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4250/HighFidelity-Beta-PR11591-fcc7a3e440bd3f84f7574837437e353a0f93f31e.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4250/HighFidelity-Beta-PR11591-fcc7a3e440bd3f84f7574837437e353a0f93f31e.exe

unpublishedScripts/marketplace/shapes/utilities/utilities.js
```diff
@@ -27,6 +27,7 @@ if (typeof Vec3.abs !== "function") {
 }
 
 if (typeof Quat.ZERO !== "object") {
+    // TODO: Change to Quat.IDENTITY.
```
thoys: Why not now? Incompatibility of marketplace item with current release?

Nex-Pro:
T E S T I N G  win10

Nex-Pro:
[9/18 23:8:58] [testMethods.js] INFO - Script Engine starting:testMethods.js
[9/18 23:8:58] [testMethods.js] Quat.IDENTITY = {"x":0,"y":0,"z":0,"w":1} ; {"x":0,"y":0,"z":0,"w":1}
[9/18 23:8:58] [testMethods.js] Uuid.NULL = {00000000-0000-0000-0000-000000000000} ; {00000000-0000-0000-0000-000000000000}
[9/18 23:8:58] [testMethods.js] Uuid.isNull(Uuid.NULL) = true ; true
[9/18 23:8:58] [testMethods.js] MyAvatar.SELF_ID = {00000000-0000-0000-0000-000000000001} ; {00000000-0000-0000-0000-000000000001}

Mostly looks good. I'd like to discuss a couple of points with you @ctrlaltdavid

Nex-Pro:
### PASS

Grabbing wearables. [attachedEntitiesManager.js] **SUCCESS**
DoppleGanger app from the Marketplace. [attachedEntitiesManager.js] **SUCCESS**
away.js **SUCCESS**
Unparenting entity using Edit > Unparent Entity and Ctrl-Shift-P. [edit.js] **SUCCESS**
/system/nameTags.js **SUCCESS**
Controller display. [controllerDisplay.js] **SUCCESS**
Hand controllers operations. [controller modules] **SUCCESS**
Entity camera. [entityCameraTool.js] **SUCCESS**
Entity selection. [entitySelectionTool.js] **SUCCESS**
Tablet opening and grabbing. [WebTablet.js] **SUCCESS**
Tutorial that creates a tetherball. [createTetherBallStick.js] **SUCCESS**

PR #11593, opened 2017-10-13
Reduce present traffic on main thread event queue.

hyperlogic:
In an effort to diagnose the root cause of several deadlocks, it seems possible that the main thread event queue can grow to a size where it can take many seconds to drain.  This loss in responsiveness leads to a deadlock crash.

This PR attempts to address this in two ways:
  * Change the connection between the DisplayPlugin::presented signal and the Application::onPresent slot to be a DirectConnection.  This should reduce the number of signal events piling up on the main thread.
  * Within Applicaiton::onPresent use an atomic counter to prevent the main thread from filling up with repeated Idle events.

Before this change loading a domain with a slow update rate would sometimes cause the event queue to fill to over 4000 events.   After this change, the same domain would only fill to 400 events, a reduction by a factor of 10.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -207,6 +207,17 @@
 #if defined(Q_OS_WIN)
 #include <VersionHelpers.h>
 
+#ifdef DEBUG_EVENT_QUEUE
+// This is a HACK that uses private headers included with the qt source distrubution.
+// To use this feature you need to add these directores to your include path:
+// E:/Qt/5.9.1/Src/qtbase/include/QtCore/5.9.1/QtCore
+// E:/Qt/5.9.1/Src/qtbase/include/QtCore/5.9.1
+#define QT_BOOTSTRAPPED
+#include <private/qthread_p.h>
+#include <private/qobject_p.h>
+#undef QT_BOOTSTRAPPED
+#endif
+
 extern "C" {
  _declspec(dllexport) DWORD NvOptimusEnablement = 0x00000001;
 }
@@ -264,9 +275,7 @@ private:
         switch ((int)event->type()) {
             case ApplicationEvent::Render:
                 render();
-                // Ensure we never back up the render events.  Each render should be triggered only in response
-                // to the NEXT render event after the last render occured
-                QCoreApplication::removePostedEvents(this, ApplicationEvent::Render);
+                qApp->_pendingRenderEventCount--;
                 return true;
 
             default:
@@ -2712,9 +2721,15 @@ bool Application::importFromZIP(const QString& filePath) {
     return true;
 }
 
+// thread-safe
 void Application::onPresent(quint32 frameCount) {
-    postEvent(this, new QEvent((QEvent::Type)ApplicationEvent::Idle), Qt::HighEventPriority);
-    if (_renderEventHandler && !isAboutToQuit()) {
+    if (_pendingIdleEventCount.load() == 0) {
+        _pendingIdleEventCount++;
+        postEvent(this, new QEvent((QEvent::Type)ApplicationEvent::Idle), Qt::HighEventPriority);
+    }
+
+    if (_renderEventHandler && !isAboutToQuit() && _pendingRenderEventCount.load() == 0) {
+        _pendingRenderEventCount++;
         postEvent(_renderEventHandler, new QEvent((QEvent::Type)ApplicationEvent::Render));
     }
 }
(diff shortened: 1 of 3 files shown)


hyperlogic:
https://highfidelity.fogbugz.com/f/cases/8146/

hyperlogic:
QA test plan:
   * On a variety of video cards and CPU types.  Enter domains such as dev-welcome which has a lot of content.
   * Specifically try a domain where the Game Rate: number on the stats display is lower than the Present Rate: number.   Domains with lots of physics entities can sometimes have this property.
   * Deadlocks should not occur.

hyperlogic:
This stack trace is an example where the root cause might be a very large event queue on the main thread, which is not able to be drained before updating the deadlock heartbeat.

https://app.bugsplat.com/individualCrash/?debugger=1&id=14992&database=interface_alpha

interface/src/Application.cpp
```diff
@@ -2712,9 +2721,15 @@ bool Application::importFromZIP(const QString& filePath) {
     return true;
 }
 
+// thread-safe
 void Application::onPresent(quint32 frameCount) {
-    postEvent(this, new QEvent((QEvent::Type)ApplicationEvent::Idle), Qt::HighEventPriority);
-    if (_renderEventHandler && !isAboutToQuit()) {
+    if (_pendingIdleEventCount.load() == 0) {
```
jherico: No need to separate the increment and the test.

    if (_pendingIdleEventCount.fetch_add(1) == 0) {
     postEvent(this, new QEvent((QEvent::Type)ApplicationEvent::Idle), Qt::HighEventPriority);
    }

interface/src/Application.cpp
```diff
@@ -2712,9 +2721,15 @@ bool Application::importFromZIP(const QString& filePath) {
...
+        _pendingIdleEventCount++;
+        postEvent(this, new QEvent((QEvent::Type)ApplicationEvent::Idle), Qt::HighEventPriority);
+    }
+
+    if (_renderEventHandler && !isAboutToQuit() && _pendingRenderEventCount.load() == 0) {
```
jherico: No need to separate the increment and the test.

    if (_renderEventHandler && !isAboutToQuit() && _pendingRenderEventCount.fetch_add(1) == 0) {
     postEvent(this, new QEvent((QEvent::Type)ApplicationEvent::Idle), Qt::HighEventPriority);
    }

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4141/HighFidelity-Beta-PR11593-a2a6cd0c7a92339d7da6b302f8d4eb367c456ac2.dmg

interface/src/Application.cpp
jherico: oh wait, I see you've made the onPresent call happen on the display plugin thread...  disregard the above comment....

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4141/HighFidelity-Beta-PR11593-a2a6cd0c7a92339d7da6b302f8d4eb367c456ac2.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4149/HighFidelity-Beta-PR11593-67d43b4fac11f90ee26342e1b771e8aea3a02b3e.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4149/HighFidelity-Beta-PR11593-67d43b4fac11f90ee26342e1b771e8aea3a02b3e.exe

Nex-Pro:
T E S T I N G   With i7 and gtx1070

Nex-Pro:
This is a **PASS** on my i7 and gtx1070 but this needs tesing on other CPUs and GFX cards.


QA test plan:


On a variety of video cards and CPU types. Enter domains such as dev-welcome which has a lot of content.

Specifically try a domain where the Game Rate: number on the stats display is lower than the Present Rate: number. Domains with lots of physics entities can sometimes have this property.

Tested using an itel i7 and an nvidia gtx1070.

As my game and present rate at dev-welcome were basically the same at 60 it was necessary to create a situation that would drop the game rate. So I used my local sandbox.
I enclosed the default platform with some walls and created large numbers of physics entities approx 1000 2.8m spheres. The Game rate dropped to approx 20 with presnt rate at 60.
I was able to walk around barging the spheres out of the way without crashing.

I then corralled the spheres into a smaller area this dropped my fame rate further to approx 15fps. My avatar was eventually able to tunnel down to the floor through the spheres unaided.

**1000 2.8m dynamic spheres with gravity of -12.**
clb400-hifi-snap-by--on-2017-10-16_14-22-26

hifi-snap-by--on-2017-10-16_14-23-30


Apart from low frame rates no other issues were noticable.

Deadlocks should not occur. Saw no deadlocks.


***Noticed a strange side-effect.*** 

As I was taking a snapshot I got a notification indicating that I should set up my wallet. This was not expected as commerce was not on ( no wallet icon ) and I was not logged in.
I checked interface.json and there was no entry for commerce.

See logs. 

[11593_Logs.zip](https://github.com/highfidelity/hifi/files/1387606/11593_Logs.zip)


***PS: This test took a lot of balls.***

Nex-Pro:
Testing with HMD

Nex-Pro:
Seems ok in HMD. Interacted with and flew inside a torrent of falling dynamic spheres with no issues. App exited cleanly.

Nex-Pro:
I'm testing this with other collision types. I think spheres may be based simply on radius and euclidian point distance calcs. Looks like hull collisions are way more expensive with much lower entity counts.

Nex-Pro:
Ouch.

Nex-Pro:
Ok This is a **PASS** on i7 with non-sphere collisions. Using simple hulls colliding with static ents.
Doesn't need a lot of ents to slow things down.
I did the second test with the medieval village and set a bunch of the trees to be dynamic with submesh collisions on. Really hammered the frame rate.
Let me know if you want a copy of the models.gz file that I used.

Did not crash.

PR #11604, opened 2017-10-16
Scale LaserPointer end overlay in C++

SamGondelman:
Adds an option to LaserPointers to automatically scale their end overlay by distance and avatar scale.  This way, the controller module sphere scaling, which was being done in the scripts, won't lag behind the rest of the overlays.

Test plan:
- Create an entity near you.  Point your blue hand lasers at it.  Move the lasers off of the object so they are instead intersecting an entity that's far away (a large floor, or a wall, etc.).  While still holding the trigger, move the lasers back onto the nearby entity.  The spheres at the end of the lasers shouldn't flash very large.  They should be the expected dimensions.
- The sphere at the end of the lasers should still be correctly sized when you laser over the HUD sphere, web entities, and the tablet surface.

### DIFF
--- a/interface/src/raypick/LaserPointer.cpp
+++ b/interface/src/raypick/LaserPointer.cpp
@@ -15,13 +15,14 @@
 #include "avatar/AvatarManager.h"
 
 LaserPointer::LaserPointer(const QVariant& rayProps, const RenderStateMap& renderStates, const DefaultRenderStateMap& defaultRenderStates,
-        const bool faceAvatar, const bool centerEndY, const bool lockEnd, const bool enabled) :
+        const bool faceAvatar, const bool centerEndY, const bool lockEnd, const bool distanceScaleEnd, const bool enabled) :
     _renderingEnabled(enabled),
     _renderStates(renderStates),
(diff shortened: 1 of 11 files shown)


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4184/HighFidelity-Beta-PR11604-aee817fbbc4c9031066fc2b2f6bba1f9acea047e.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4184/HighFidelity-Beta-PR11604-aee817fbbc4c9031066fc2b2f6bba1f9acea047e.exe

Nex-Pro:
T E S T I N G

Nex-Pro:
Spheres are tiny on the tablet and almost invisible on web entities.
Sphere size is constant for other entities.
Sphere scales with avatar size.

SamGondelman:
The first and third things are how it behaves on master, too.  If we want to change that we can put up bugs and address them separately, but this PR didn't touch that behavior, so we can call this a pass.

PR #11607, opened 2017-10-16
CMake tweaks

jherico:
This PR modifies the cmake files to do two things...

First, it allows CMake command line options to be used in finding the Qt installation.  Currently in master if you have Qt 5.9.2 installed, and you set the environment variable `QT_VERSION` to `5.9.2` it will find and successfully configure the project.  However, if have no such environment variable and instead call CMake with the command line option `-DQT_VERSION=5.9.2` it will tail, looking instead of Qt 5.9.1.  This is because of a bug in the code that sets these values because it looks only at the environment variable, not at whether the CMake variable value may already have been set from the command line.  

Second, it explicitly marks the autoscribe generated shaders as not needing AUTOMOC.  Failing to do this will produce a ton of warnings in CMake 3.10 because of [CMP0071](https://cmake.org/cmake/help/git-stage/policy/CMP0071.html)

## Testing

No QA pass necessary.  This only impacts developer CMake configuration, not the actual code.


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4188/HighFidelity-Beta-PR11607-17a68476095ad70b086cba94283fed58125b62cb.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4188/HighFidelity-Beta-PR11607-17a68476095ad70b086cba94283fed58125b62cb.exe

PR #11618, opened 2017-10-17
Generalized Pick and Pointer managers and classes

SamGondelman:
- Moves a bunch of work out of RayPickManager and RayPick and into PickManager and Pick, which will be used for additional types of picking, like the Stylus.
- Adds a Pointer interface that LaserPointer implements, and replaces LaserPointerManager with PointerManager.  
- Pick and Pointer scripting interfaces, Picks and Pointers (RayPick and LaserPointers are still functional).  Filter flags and intersection result types have moved from "RayPick." to "Picks." and scripts have been updated accordingly.


interface/src/Application.cpp
```diff
@@ -1819,11 +1819,11 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer, bo
 
     // Setup the mouse ray pick and related operators
     DependencyManager::get<EntityTreeRenderer>()->setMouseRayPickID(_rayPickManager.createRayPick(
-        RayPickFilter(DependencyManager::get<RayPickScriptingInterface>()->PICK_ENTITIES() | DependencyManager::get<RayPickScriptingInterface>()->PICK_INCLUDE_NONCOLLIDABLE()),
+        PickFilter(DependencyManager::get<RayPickScriptingInterface>()->PICK_ENTITIES() | DependencyManager::get<RayPickScriptingInterface>()->PICK_INCLUDE_NONCOLLIDABLE()),
```
SamGondelman: So, the reason this is the scripting interface is because it's mimicking a script to create a RayPick on the C++ side.  Technically it *could* be building the filter itself, but this way you're guaranteed consistency between scripts and here.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4238/HighFidelity-Beta-PR11618-bb370dfae81f0865f919834cc43d98343077990d.exe

interface/src/raypick/JointRayPick.h
```diff
@@ -11,14 +11,14 @@
 #ifndef hifi_JointRayPick_h
 #define hifi_JointRayPick_h
 
-#include <pointers/rays/RayPick.h>
```
jherico: I don't think we should be moving this back.

interface/src/raypick/Pick.h
```diff
@@ -0,0 +1,253 @@
...
+};
+
+template<typename T>
+class Pick : protected ReadWriteLockable {
+
+public:
+    Pick(const PickFilter& filter, const float maxDistance, const bool enabled);
+
+    virtual const T getMathematicalPick() const = 0;
+    virtual RayToEntityIntersectionResult getEntityIntersection(const T& pick, bool precisionPicking,
```
jherico: I think we need to generalize PickResults such that we don't have different kinds of avatar, overlay and entity.  Additionally, we should try to produce a pick result that doesn't depend on the type of pick involved, so a sphere pick and a ray pick can have a common interface.

interface/src/raypick/RayPickScriptingInterface.h
```diff
@@ -46,20 +46,20 @@ public slots:
...
-    unsigned int PICK_COARSE() { return RayPickFilter::getBitMask(RayPickFilter::FlagBit::PICK_COARSE); }
-    unsigned int PICK_INCLUDE_INVISIBLE() { return RayPickFilter::getBitMask(RayPickFilter::FlagBit::PICK_INCLUDE_INVISIBLE); }
-    unsigned int PICK_INCLUDE_NONCOLLIDABLE() { return RayPickFilter::getBitMask(RayPickFilter::FlagBit::PICK_INCLUDE_NONCOLLIDABLE); }
-    unsigned int PICK_ALL_INTERSECTIONS() { return RayPickFilter::getBitMask(RayPickFilter::FlagBit::PICK_ALL_INTERSECTIONS); }
-    unsigned int INTERSECTED_NONE() { return IntersectionType::NONE; }
-    unsigned int INTERSECTED_ENTITY() { return IntersectionType::ENTITY; }
-    unsigned int INTERSECTED_OVERLAY() { return IntersectionType::OVERLAY; }
-    unsigned int INTERSECTED_AVATAR() { return IntersectionType::AVATAR; }
-    unsigned int INTERSECTED_HUD() { return IntersectionType::HUD; }
+    static unsigned int PICK_NOTHING() { return 0; }
```
jherico: these should all be should be `static constexpr`

interface/src/raypick/JointRayPick.h
SamGondelman: Even if we generalize PickResult, we still need to do the actual picks somewhere.  I think that should be RayPick.  But it would allow us to move PickManager into the pointers library, so I'll look into it.

interface/src/raypick/Pick.h
SamGondelman: We could do this...or we could give PickManager another template type for the result that it expects, and let different types of Picks have whatever results they want.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4267/HighFidelity-Beta-PR11618-0dcf593d79d4475c8bdbb446a0f15ee9418160aa.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4273/HighFidelity-Beta-PR11618-72f402365a2ab8658ea8b9dc2fd4b631868aa5c6.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4273/HighFidelity-Beta-PR11618-72f402365a2ab8658ea8b9dc2fd4b631868aa5c6.exe

libraries/pointers/src/pointers/PickManager.h
```diff
@@ -0,0 +1,229 @@
...
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+#ifndef hifi_PickManager_h
+#define hifi_PickManager_h
+
+#include <unordered_map>
+
+#include "scripting/HMDScriptingInterface.h"
```
SamGondelman: this...seems like it shouldn't be building?

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4280/HighFidelity-Beta-PR11618-c2c6d5b4cd61a93bcb9885af2864161d0e544543.dmg

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4283/HighFidelity-Beta-PR11618-3572f7f5db9e989c0100817fabd77d0292eceb61.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4283/HighFidelity-Beta-PR11618-3572f7f5db9e989c0100817fabd77d0292eceb61.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4295/HighFidelity-Beta-PR11618-d7745efc8e1a256f22aa5bf24a14799a2c2f83ef.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4295/HighFidelity-Beta-PR11618-d7745efc8e1a256f22aa5bf24a14799a2c2f83ef.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4315/HighFidelity-Beta-PR11618-24ce09f91413122d6d92570b4abc75cad44a8223.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4315/HighFidelity-Beta-PR11618-24ce09f91413122d6d92570b4abc75cad44a8223.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4326/HighFidelity-Beta-PR11618-5cb2946c3d8915a7d4404a195f10151a64b89d71.dmg

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4328/HighFidelity-Beta-PR11618-7686948daf75789c200ffa96fd30ff51806d8825.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4328/HighFidelity-Beta-PR11618-7686948daf75789c200ffa96fd30ff51806d8825.exe

interface/src/Application.cpp
```diff
@@ -1826,25 +1835,29 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer, bo
...
+                entityResult.intersection = vec3FromVariant(result["intersection"]);
+                entityResult.distance = result["distance"].toFloat();
+                entityResult.surfaceNormal = vec3FromVariant(result["surfaceNormal"]);
+                entityResult.entityID = result["objectID"].toUuid();
+                entityResult.entity = DependencyManager::get<EntityTreeRenderer>()->getTree()->findEntityByID(entityResult.entityID);
+            }
         }
         return entityResult;
     });
     DependencyManager::get<EntityTreeRenderer>()->setSetPrecisionPickingOperator([&](QUuid rayPickID, bool value) {
```
jherico: Is this actually used anywhere? I mean does anything actually call `EntityTree::setPrecisionPicking`?  I guess since it's a slot it's conceivable that some script might call it if the tree is exposed to scripts.  Dunno.

interface/src/Application.cpp
```diff
@@ -5832,6 +5845,8 @@ void Application::registerScriptEngineWithApplicationServices(ScriptEnginePointe
 
     scriptEngine->registerGlobalObject("RayPick", DependencyManager::get<RayPickScriptingInterface>().data());
     scriptEngine->registerGlobalObject("LaserPointers", DependencyManager::get<LaserPointerScriptingInterface>().data());
+    scriptEngine->registerGlobalObject("Picks", DependencyManager::get<PickScriptingInterface>().data());
```
jherico: I'd recommend we use singular form here.  I know `LaserPointers` is plural, but the general convention for JS interfaces is singular:  `Document`, `Window`...  So I would suggest leave `LaserPointers` because it's legacy, but add the new ones as `Pointer` and `Picks`

We should also consider creating a new global `Hifi` object and starting to add the interfaces as properties on it, so `Hifi.Pick`, `Hifi.Pointer`, so that we can migrate away from creating top level script names that might eventually conflict with a JS global

interface/src/raypick/LaserPointer.cpp
```diff
@@ -51,14 +50,14 @@ LaserPointer::~LaserPointer() {
 }
 
 void LaserPointer::enable() {
-    qApp->getRayPickManager().enableRayPick(_rayPickUID);
+    Pointer::enable();
```
jherico: Whenever I create a class, I always declare an alias for the parent class at the very top of the class (so it's private) like so: 
  `using Parent = Pointer;`.  This means in the class I can do things like `Parent::enable` instead of `Pointer::enable`.  That means if I change the class hierarchy, by adding a new intermediate class or otherwise moving things around, I don't break things (as long as I remember to update the class declaration and the `Parent` statement right below it).

interface/src/raypick/LaserPointer.cpp
```diff
@@ -106,10 +105,6 @@ void LaserPointer::updateRenderStateOverlay(const OverlayID& id, const QVariant&
     }
 }
 
-const RayPickResult LaserPointer::getPrevRayPickResult() {
-    return qApp->getRayPickManager().getPrevRayPickResult(_rayPickUID);
-}
-
 void LaserPointer::updateRenderState(const RenderState& renderState, const IntersectionType type, const float distance, const QUuid& objectID, const PickRay& pickRay, const bool defaultState) {
```
jherico: eventually we want to completely divorce the LaserPointer from rendering.  The pointers themselves should just contain data describing as much as is required to know how to render the object.  A separate rendering manager should able to iterate all the active pointers and render appropriately.

libraries/pointers/src/pointers/Pick.h
```diff
@@ -91,36 +101,63 @@ class RayPickFilter {
...
+
+class PickQuery : protected ReadWriteLockable {
+    Q_GADGET
+public:
+    PickQuery(const PickFilter& filter, const float maxDistance, const bool enabled);
 
-    virtual const PickRay getPickRay(bool& valid) const = 0;
+    enum PickType {
+        Ray = 0,
+        Stylus
```
jherico: I'm not sure I get this....  a stylus is just a ray with a specific max length, but rays can already have max distance, right?  Is there some other logic that applies to stylus and not to ray?  If so can we add a comment?

libraries/shared/src/RegisteredMetaTypes.h
```diff
@@ -125,42 +125,51 @@ QVector<QUuid> qVectorQUuidFromScriptValue(const QScriptValue& array);
 QScriptValue aaCubeToScriptValue(QScriptEngine* engine, const AACube& aaCube);
 void aaCubeFromScriptValue(const QScriptValue &object, AACube& aaCube);
 
-class PickRay {
+// MathPicks also have to overide operator== for their type
+class MathPick {
+    virtual operator bool() const = 0;
+    virtual QVariantMap toVariantMap() const = 0;
```
jherico: In general it might be simpler to say `QVariant toVariant()`

libraries/shared/src/RegisteredMetaTypes.h
```diff
@@ -125,42 +125,51 @@ QVector<QUuid> qVectorQUuidFromScriptValue(const QScriptValue& array);
...
+        return (origin == other.origin && direction == other.direction);
+    }
+    QVariantMap toVariantMap() const override {
+        QVariantMap pickRay;
+        pickRay["origin"] = vec3toVariant(origin);
+        pickRay["direction"] = vec3toVariant(direction);
+        return pickRay;
+    }
 };
+namespace std {
```
jherico: There's some cool code [here](https://stackoverflow.com/questions/2590677/how-do-i-combine-hash-values-in-c0x) for a generic `hash_combine` function that we might want to add to the shared library, which would simplify these functions.

scripts/system/controllers/controllerDispatcher.js
```diff
@@ -252,7 +252,7 @@ Script.include("/~/system/libraries/controllerDispatcherUtils.js");
                     length: 1000
                 };
 
-                if (rayPicks[h].type === RayPick.INTERSECTED_ENTITY) {
+                if (rayPicks[h].type === Picks.INTERSECTED_ENTITY) {
```
jherico: Does `RayPick.INTERSECTED_ENTITY` no longer work?  Third party scripts might still be depending on it.

interface/src/Application.cpp
SamGondelman: I never found a place where it was actually called, no.

interface/src/Application.cpp
SamGondelman: I was trying to match `Entities` and `Overlays`, but yeah either works.

libraries/pointers/src/pointers/Pick.h
SamGondelman: If we're mimicking the current JS stylus behavior, the stylus is different than a ray.  It has its own picking logic.  I can add a comment but it will also be evident from the StylusPick class itself, once we make that.

libraries/shared/src/RegisteredMetaTypes.h
SamGondelman: Since we use the result in C++ in a few places, I wanted to avoid the extra conversion to a QVariantMap everywhere.

scripts/system/controllers/controllerDispatcher.js
SamGondelman: I did remove it but I will put it back.

PR #11641, opened 2017-10-20
hand-controller search laser will hit non-collidable entities

sethalves:
- hand-controller search laser will hit non-collidable entities

### DIFF
--- a/scripts/system/controllers/controllerModules/farActionGrabEntity.js
+++ b/scripts/system/controllers/controllerModules/farActionGrabEntity.js
@@ -573,7 +573,7 @@ Script.include("/~/system/libraries/controllers.js");
 
         this.laserPointer = LaserPointers.createLaserPointer({
             joint: (this.hand === RIGHT_HAND) ? "_CAMERA_RELATIVE_CONTROLLER_RIGHTHAND" : "_CAMERA_RELATIVE_CONTROLLER_LEFTHAND",
-            filter: RayPick.PICK_ENTITIES | RayPick.PICK_OVERLAYS,
+            filter: RayPick.PICK_ENTITIES | RayPick.PICK_OVERLAYS | RayPick.PICK_INCLUDE_NONCOLLIDABLE,
             maxDistance: PICK_MAX_DISTANCE,
             posOffset: getGrabPointSphereOffset(this.handToController(), true),
             renderStates: renderStates,


sethalves:
https://highfidelity.fogbugz.com/f/cases/8524/lasers-going-through-avatar-store-glasses

scripts/system/controllers/controllerModules/farActionGrabEntity.js
```diff
@@ -573,7 +573,7 @@ Script.include("/~/system/libraries/controllers.js");
 
         this.laserPointer = LaserPointers.createLaserPointer({
             joint: (this.hand === RIGHT_HAND) ? "_CAMERA_RELATIVE_CONTROLLER_RIGHTHAND" : "_CAMERA_RELATIVE_CONTROLLER_LEFTHAND",
-            filter: RayPick.PICK_ENTITIES | RayPick.PICK_OVERLAYS,
+            filter: RayPick.PICK_ENTITIES | RayPick.PICK_OVERLAYS | RayPick.PICK_INCLUDE_NONCOLLIDABLE,
```
ZappoMan: this isn't exactly what we agreed to...

The correct behavior is:
Ray intersects with:
- Anything that IS collidable
- Anything that is grabbable
- Anything that is triggerable

it should not intersect with non-collidable, unless those are grabbable or triggerable

SamGondelman: We don't currently have the ability to do that.  findRayIntersection would need to be modified to be able to ignore entities based on if they are grabbable/triggerable, and then additional flags would need to be added to RayPickFilter that would allow you to configure that.

sethalves: we should probably make grabbable / triggerable be first class properties, also (rather than userData json).

sethalves: @ZappoMan I know this isn't what we want, but I thought there was a chance this would work better than without this PR until we can fix it.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4324/HighFidelity-Beta-PR11641-a75e5c751148dcd2b0834692887db491d3e2f93a.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4324/HighFidelity-Beta-PR11641-a75e5c751148dcd2b0834692887db491d3e2f93a.exe

scripts/system/controllers/controllerModules/farActionGrabEntity.js
ZappoMan: We don't need to "rush" this in.. there are work arounds that the content team can use (like make things collidable) that seem to meet their needs.

ZappoMan: and yes we should make these first class properties and add this support to the ray pick manager

sethalves:
okay, we won't rush it it.

PR #11663, opened 2017-10-24
fix incorrect collision shape sharing for some ModelEntityItem configurations

AndrewMeadows:
- fix bug where sometimes a ModelEntityItem would have the wrong collision shape
- fix debug rendering of ModelEntityItem collision shapes

### DIFF
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -695,12 +695,8 @@ void RenderableModelEntityItem::computeShapeInfo(ShapeInfo& shapeInfo) {
 void RenderableModelEntityItem::setCollisionShape(const btCollisionShape* shape) {
     const void* key = static_cast<const void*>(shape);
     if (_collisionMeshKey != key) {
-        if (_collisionMeshKey) {
-            collisionMeshCache.releaseMesh(_collisionMeshKey);
-        }
         _collisionMeshKey = key;
-        // toggle _showCollisionGeometry forces re-evaluation later
-        _showCollisionGeometry = !_showCollisionGeometry;
+        emit requestCollisionGeometryUpdate();
     }
 }
 
@@ -1103,6 +1099,10 @@ bool ModelEntityRenderer::needsRenderUpdate() const {
         if (model->getRenderItemsNeedUpdate()) {
             return true;
         }
+
+        if (_needsCollisionGeometryUpdate) {
+            return true;
+        }
     }
     return Parent::needsRenderUpdate();
 }
@@ -1169,6 +1169,15 @@ bool ModelEntityRenderer::needsRenderUpdateFromTypedEntity(const TypedEntityPoin
     return false;
 }
 
+void ModelEntityRenderer::setCollisionMeshKey(const void*key) {
+    if (key != _collisionMeshKey) {
+        if (_collisionMeshKey) {
+            collisionMeshCache.releaseMesh(_collisionMeshKey);
+        }
+        _collisionMeshKey = key;
+    }
+}
+
 void ModelEntityRenderer::doRenderUpdateSynchronousTyped(const ScenePointer& scene, Transaction& transaction, const TypedEntityPointer& entity) {
     DETAILED_PROFILE_RANGE(simulation_physics, __FUNCTION__);
     if (_hasModel != entity->hasModel()) {
@@ -1201,6 +1210,7 @@ void ModelEntityRenderer::doRenderUpdateSynchronousTyped(const ScenePointer& sce
         model = std::make_shared<Model>(nullptr, entity.get());
         connect(model.get(), &Model::setURLFinished, this, &ModelEntityRenderer::requestRenderUpdate);
         connect(model.get(), &Model::requestRenderUpdate, this, &ModelEntityRenderer::requestRenderUpdate);
+        connect(entity.get(), &RenderableModelEntityItem::requestCollisionGeometryUpdate, this, &ModelEntityRenderer::flagForCollisionGeometryUpdate);
         model->setLoadingPriority(EntityTreeRenderer::getEntityLoadingPriority(*entity));
         model->init();
         entity->setModel(model);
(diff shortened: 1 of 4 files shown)


libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -1327,28 +1362,11 @@ void ModelEntityRenderer::doRender(RenderArgs* args) {
...
-            // clear model's collision geometry
-            model::MeshPointer mesh = nullptr;
-            _model->setCollisionMesh(mesh);
-        }
+    bool showCollisionGeometry = (bool)(args->_debugFlags & (int)RenderArgs::RENDER_DEBUG_HULLS);
+    if (showCollisionGeometry != _showCollisionGeometry) {
+        _showCollisionGeometry = showCollisionGeometry;
+        flagForCollisionGeometryUpdate();
     }
-#endif
```
AndrewMeadows: This code that was `ifdef`'ed out was part of the logic cascade that would allow rendering of `ModelEntityItem` collision shapes under certain debug menu configurations.  It was broken during the transition to threaded rendering since some of these operations cannot be done in the thread of this context.

The solution is to set some local flags (in the `ModelEntityRenderer` instance, which is also known as the "meta render item") which is then checked in the `doRenderUpdateSynchronousTyped()` context where the work is done in the correct thread.

libraries/shared/src/ShapeInfo.cpp
```diff
@@ -250,16 +252,26 @@ const DoubleHashKey& ShapeInfo::getHash() const {
...
+        if (_type == SHAPE_TYPE_COMPOUND || _type == SHAPE_TYPE_SIMPLE_COMPOUND) {
+            numHulls = (uint32_t)_pointCollection.size();
+        } else if (_type == SHAPE_TYPE_SIMPLE_HULL) {
+            numHulls = 1;
+        }
+        if (numHulls > 0) {
+            hash = DoubleHashKey::hashFunction(numHulls, primeIndex++);
+            _doubleHashKey.setHash(_doubleHashKey.getHash() ^ hash);
+            hash = DoubleHashKey::hashFunction2(numHulls);
+            _doubleHashKey.setHash2(_doubleHashKey.getHash2() ^ hash);
```
AndrewMeadows: These changes fix the bug where two distinct collision configurations would generate the same "key" to a shape in the `ShapeManager`, which would cause the second object created to get the wrong collision shape: it would share the old shape rather than generate a new one.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4377/HighFidelity-Beta-PR11663-90b82bbb75da8b95017b6376c6defdde2927ec1b.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4377/HighFidelity-Beta-PR11663-90b82bbb75da8b95017b6376c6defdde2927ec1b.exe

AndrewMeadows:
Test plan:

Verify the bug in hifi/master:
(1) Use the two FBX asset links to create two ModelEntityItem's.  Both object's "Collision shape type" should be set to "Good - Sub-meshes":

https://github.com/AndrewMeadows/hifi_tests/raw/ba865edf32f54c95f4df00f88a0f887e3cf8f754/assets/models/collisions/xyzCross-single-mesh.fbx

https://github.com/AndrewMeadows/hifi_tests/raw/ba865edf32f54c95f4df00f88a0f887e3cf8f754/assets/models/collisions/xyzCross-three-submeshes.fbx

The first model is a single mesh and the "Good - Sub-meshes" option will produce a single convex hull collision shape around the entire thing.  Whereas the second model has three submeshes, each with different color (material) and the "Good - Sub-meshes" option will wrap each submesh in its own convex hull.

However the bug will make the second object collide like the first.  The bug is order-dependent so whichever one is created first will store the correct collision shape and the object created second will inherit the same collision shape as the first.

Note: 

(2) Bump your avatar into the objects to examine the object's collision shapes.
--> The second object is expected to collide like the first (this is a BUG).

(3) Make sure the **Developer --> Physics --> Draw Collision Shapes** option is disabled.  This option is not working in hifi/master.

(4) Relog using the interface client from this PR.

(5) Repeat step (2): bump your avatar into the two objects
--> The two objects are expected to collide differently: the single-mesh will collide like one big convex hull while the three-mesh will collide as three.

(6) Enable the **Developer --> Physics --> Draw Collision Shapes** option.
--> The two objects should be drawn at they collide and should agree visibly with how the the avatar bumps into them.

(7) Scale the two objects to non-default dimensions:
--> Verify they render correctly in both cases: with **Developer --> Physics --> Draw Collision Shapes** enabled and with it disabled.
--> Verify they collide as expected in their scaled state.

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

**Verify the bug in hifi/master:**

(1) Use the two FBX asset links to create two ModelEntityItem's. Both object's "Collision shape type" should be set to "Good - Sub-meshes":
**Check** 
https://github.com/AndrewMeadows/hifi_tests/raw/ba865edf32f54c95f4df00f88a0f887e3cf8f754/assets/models/collisions/xyzCross-single-mesh.fbx
**Check** 
https://github.com/AndrewMeadows/hifi_tests/raw/ba865edf32f54c95f4df00f88a0f887e3cf8f754/assets/models/collisions/xyzCross-three-submeshes.fbx
**Check** 
The first model is a single mesh and the "Good - Sub-meshes" option will produce a single convex hull collision shape around the entire thing. Whereas the second model has three submeshes, each with different color (material) and the "Good - Sub-meshes" option will wrap each submesh in its own convex hull.
**Check** 
However the bug will make the second object collide like the first. The bug is order-dependent so whichever one is created first will store the correct collision shape and the object created second will inherit the same collision shape as the first.
**Check** 
(2) Bump your avatar into the objects to examine the object's collision shapes.
--> The second object is expected to collide like the first (this is a BUG).
**Check** 
(3) Make sure the Developer --> Physics --> Draw Collision Shapes option is disabled. This option is not working in hifi/master.
**Check** 

**Verify the bug is fixed in this PR:**

(4) Relog using the interface client from this PR.
**Experienced "Protocol version mismatch - Domain version: 7279", re-setup same testing on PR Sandbox, if this is an issue please let me know**
(5) Repeat step (2): bump your avatar into the two objects
--> The two objects are expected to collide differently: the single-mesh will collide like one big convex hull while the three-mesh will collide as three.

(6) Enable the Developer --> Physics --> Draw Collision Shapes option.
--> The two objects should be drawn at they collide and should agree visibly with how the the avatar bumps into them.
**Check, see image below also. Left is 1st FBX, Right is 2nd FBX. All looks and collides correctly**
image

(7) Scale the two objects to non-default dimensions:
--> Verify they render correctly in both cases: with Developer --> Physics --> Draw Collision Shapes enabled and with it disabled.
**Check**
--> Verify they collide as expected in their scaled state.
**Check, tried much larger and much smaller than when rezzed in, all worked as expected.**


Very clear instructions, much appreciated :) Full **Pass** also!

**Notes: I had to use the PR Sandbox when using PR Interface, due to protocol mismatch**

samcake:
@andrew, can you bring this test into hifi_tests/test rail ?

AndrewMeadows:
I'm currently working on a more general "model shapes" test.  That project is what motivated this PR work.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4390/HighFidelity-Beta-PR11663-27f4d182d97c9c7b74b757c6a8bdab78e7274083.dmg

marko8904:
build this please

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4392/HighFidelity-Beta-PR11663-27f4d182d97c9c7b74b757c6a8bdab78e7274083.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4392/HighFidelity-Beta-PR11663-27f4d182d97c9c7b74b757c6a8bdab78e7274083.exe

PR #11691, opened 2017-10-26
[do not merge] WL 21555: Web3DOverlay refactoring into Window/Web/Tablet 3D Overlays

humbletim:
This PR contains proposed refactoring and cleanup for Web3D Overlays that reworks things into a generalized Window3DOverlay class with separate subclasses for the different concern areas:

- **Window3DOverlay** -- abstract base class where "window like" concepts live (eg: mapping mouse/touch inputs, bootstrapping surface cache, etc.).
- **Web3DOverlay** -- extends *Window3DOverlay* with support for dynamic Web Views.
- **Qml3DOverlay** -- extends *Window3DOverlay* with support for dynamic Qml Views.
- **Tablet3DOverlay** -- extends *Qml3DOverlay* with support for the unique Tablet View.

The proposed next step is to discuss/review in context and examine whether this approach makes sense as a strategy for making window-like 3d overlays easier to manage/debug moving forward.

cc: @themelissabrown


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4458/HighFidelity-Beta-PR11691-4ea5448680841531981cda0345c8f93bdf99260d.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4458/HighFidelity-Beta-PR11691-4ea5448680841531981cda0345c8f93bdf99260d.exe

ZappoMan:
This is gonna be a doozy... if we do like this approach, we're going to need a very thorough testing strategy.

humbletim:
@ZappoMan Yeah... I suppose it is something of a doozy.  Although these changes are meant to be 100% backwards-compatible (at both user experience and scripting level).  With one exception --
 WebTablet.js now uses an overlay type of `'tablet3d'` and `.source` property for its .qml URL.

Note that this factoring also provides headroom for Tablet functionality to expand as rapidly as apparently needed lately, with less risk of breaking existing user projects that depend on having stable Web3D support.  

I can offer a test script(s) that hit Web3D overlays from multiple angles, reliably crash the current beta (and do not seem to crash this PR build).

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4558/HighFidelity-Beta-PR11691-c41858d395883cc3f23ce5e902bf622094695f2e.dmg

libraries/ui/src/ui/OffscreenQmlSurface.cpp
```diff
@@ -1158,7 +1158,7 @@ void OffscreenQmlSurface::emitWebEvent(const QVariant& message) {
 
 void OffscreenQmlSurface::sendToQml(const QVariant& message) {
     if (QThread::currentThread() != thread()) {
-        QMetaObject::invokeMethod(this, "emitQmlEvent", Qt::QueuedConnection, Q_ARG(QVariant, message));
+        QMetaObject::invokeMethod(this, "sendToQml", Qt::QueuedConnection, Q_ARG(QVariant, message));
```
humbletim: note: `"emitQmlEvent"` doesn't seem to be a thing so if this is a bug (which I think it is) then the same issue currently exists in master [libraries/ui/src/ui/OffscreenQmlSurface.cpp](https://github.com/highfidelity/hifi/blob/master/libraries/ui/src/ui/OffscreenQmlSurface.cpp#L1161).

themelissabrown:
Likely to have follow on effects impacting the pointer refactor we're currently doing. Hold on merging until after that is merged and tested.

interface/resources/qml/hifi/overlays/Window3DOverlay.qml
```diff
@@ -0,0 +1,218 @@
...
+            sendToScript(message);
+        }
+        onWebEventReceived: {
+            log('onWebEventReceived', this, currentView);
+            webEventReceived(message);
+        }
+    }
+
+    // -- generalized component loader (.source for qmlView, .sourceComponent for webView -- assigned as part of Bindings above)
+    Loader {
```
jherico: Using the loader for QML content is no longer recommended.  We have code which will expose certain interfaces to QML based on the loaded URL, but for that to work, the built in Loader mechanism isn't supported, as the loading has to go through some custom C++ code.  See pr #11815

PR #11693, opened 2017-10-26
replace 'Circle' and 'Quad' ShapeEntityItem with flat 'Cylinder' and 'Cube', respectively

AndrewMeadows:
- deprecate `Circle` shape type in favor of a flat `Cylinder`
- deprecate `Quad` shape type in favor of a flat `Cube`
- replace `DoubleHashKey` with new and improved `HashKey` (used by `ShapeManager`)
- update `ShapeInfo` unit tests to verify `HashKey` implementation


AndrewMeadows:
The `Circle` shape option for `ShapeEntityItem` had some problems.  The main one being: the `dimensions.y` property component was not enforced to be zero (it could be anything), and this hidden value was ignored for some things but used for others.  It was the only shape option that suffered from this oddity.  In short, this video:

♬ One of these things ♪ is not like the others ♫
https://www.youtube.com/watch?v=6b0ftfKFEJg

libraries/entities/src/ShapeEntityItem.h
```diff
@@ -17,7 +17,7 @@ namespace entity {
         Quad,
         Hexagon,
         Octagon,
-        Circle,
+        Circle, // use Cylinder instead
```
AndrewMeadows: The changes above here, and the removal of `Circle` from some HTML and JS files further down are what removes the shape option.  The rest of the changes below are related to some cleanup around how we generate the "shape key" for the `ShapeManager`, which supplies physics collision shapes.

libraries/physics/src/ShapeManager.cpp
```diff
@@ -32,7 +32,7 @@ const btCollisionShape* ShapeManager::getShape(const ShapeInfo& info) {
     if (info.getType() == SHAPE_TYPE_NONE) {
         return nullptr;
     }
-    DoubleHashKey key = info.getHash();
+    HashKey key = info.getHash();
```
AndrewMeadows: I'm removing `DoubleHashKey` which had a messy and confusing API and replacing it with `HashKey` which fulfills the same purpose (see below) but is cleaner.

libraries/shared/src/HashKey.h
```diff
@@ -0,0 +1,52 @@
...
+
+// HashKey for use with btHashMap which requires a particular API for its keys.  In particular it
+// requires its Key to implement these methods:
+//
+//      bool Key::equals()
+//      int32_t Key::getHash()
+//
+// The important thing about the HashKey implementation is that while getHash() returns 32-bits,
+// internally HashKey stores a 64-bit hash which is used for the equals() comparison.  This allows
+// btHashMap to insert "dupe" 32-bit keys to different "values".
```
AndrewMeadows: This comment explains why we need `HashKey` (and why we needed `DoubleHashKey`).

libraries/shared/src/ShapeInfo.cpp
```diff
@@ -256,119 +256,46 @@ bool ShapeInfo::contains(const glm::vec3& point) const {
...
-            uint32_t hash = DoubleHashKey::hashFunction(numHulls, primeIndex++);
-            _doubleHashKey.setHash(_doubleHashKey.getHash() ^ hash);
-            hash = DoubleHashKey::hashFunction2(numHulls);
-            _doubleHashKey.setHash2(_doubleHashKey.getHash2() ^ hash);
+            _hashKey.hashUint64(1);
         }
     }
-    return _doubleHashKey;
+    return _hashKey;
 }
```
AndrewMeadows: The reason for using `HashKey` was to clean up this portion of the code.  The old version was confusing and hard to maintain because the hashes were managed explicitly here instead of inside `DoubleHashKey`.

tests/physics/src/ShapeInfoTests.cpp
```diff
@@ -23,108 +23,78 @@
 
 QTEST_MAIN(ShapeInfoTests)
 
+// Enable this to manually run testHashCollisions
+// (NOT a regular unit test; takes ~40 secs to run on an i7)
+//#define MANUAL_TEST true
```
AndrewMeadows: I manually ran this unit test to verify that `HashKey` was properly scattering the bits and not creating incorrect collisions for 10M shapes.  My confidence level is high.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4461/HighFidelity-Beta-PR11693-1d297d05ca62383314354bc01ffe3ff59e06f59a.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4461/HighFidelity-Beta-PR11693-1d297d05ca62383314354bc01ffe3ff59e06f59a.exe

AndrewMeadows:
**Test Plan:**

**(1)** Using the latest dev-download Interface client...

**(2)** Create a `Sphere` entity item.

**(3)** Examine the properties of the Sphere and set its **Shape** to be `Circle`.
--> The entity should look like a flat circle.
--> Note: the lighting might be wrong.  This is a known but ignored bug.

**(4)** Logout and log back in using the interface client from this PR
--> the entity should now look like a cylinder
--> its lighting should be correct

**(5)** Examine its properties.  Its **Shape** should now be `Cylinder`.

**(6)** Change the cylinder to be a "circle-like" object by reducing the y-component of the cylinder's dimensions to be very small.
--> Note: you can't just set its dimension.y to be zero, but you can make it very small
--> Its lighting should be correct no matter how small you make it.

**(7)** Give the disk larger diameter: 4m or so
--> Verify the disk's collision shape is correct by bumping your avatar into it.

**(8)** Make the very thin cylinder dynamic, with non-zero gravity, and allow it to drop to the ground:
--> Verify its dynamic collisions appear correct

**(9)** Walk on top of the disk and then walk off backwards
--> Verify your avatar can walk off the disk without the disk following the avatar around

ZappoMan:
There's a lot going on in this PR... more than just removing circle. As discussed, we need to keep content using circle working correctly. What's the plan for this PR?

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4487/HighFidelity-Beta-PR11693-bff40e1782c032b48d0b3b04c96dfecc31a95b58.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4487/HighFidelity-Beta-PR11693-bff40e1782c032b48d0b3b04c96dfecc31a95b58.exe

AndrewMeadows:
Good news!  Backwards compatibility for `Circle` and `Quad` is now in place.  If you create one then it will appear correctly as a flat `ShapeEntityItem` with the correct surface area.

Nex-Pro:
Testing

Nex-Pro:
@AndrewMeadows This is a quick turnaraound. I only posted those idiosyncrasies yesterday.

Nex-Pro:
### FAIL


To repro 
Using the latest dev-download Interface client. Connect to an entity-server NOT running this PR.

(2) Create a Sphere entity item.

(3) Examine the properties of the Sphere and set its Shape to be Circle.
--> The entity should look like a flat circle.**SUCCESS**
--> Note: the lighting might be wrong. This is a known but ignored bug. **Previously observed***

(4) Logout and log back in using the interface client from this PR
--> the entity should still look like a flat circle**SUCCESS**
--> its lighting should be correct (bonus!)**SUCCESS**

(5) Examine its properties.
--> Its Shape should now be Cylinder**SUCCESS**
--> The y component of its dimensions should now be small**SUCCESS** 0.0010

(6) Give the disk larger diameter: 4m or so
--> Verify the disk's collision shape is correct by bumping your avatar into it.**FAIL** on a couople of attempts to do this the collisions failed on the disk.

(7) Make the very thin cylinder dynamic, with non-zero gravity, and allow it to drop to the ground:
--> Verify its dynamic collisions appear correct **SUCCESS** But read notes below.

(8) Walk on top of the disk and then walk off backwards
--> Verify your avatar can walk off the disk without the disk following the avatar around **SUCCESS** I can no longer tow the disk around.

(9) Repeat steps (1) through (8) for a Quad shape **FAIL** Cannot find option for Quad in either build.
--> Using the hifi/master interface the Quad will not render! this is a bug
--> Using the interface from this PR the Quad will be turned into a shallow Cube and will render!





### Following the test plan has revealed a possible bug in edit js.

After setting the ent to be dynamic the first time the avatar interacts with the ent using mouse it reverts to the scale it had before it was changed to be a circle.

To repro:

In dev latest

Create a sphere, scale to be some size using shift and the corner handle.

In properties select circle.

It becomes a thin disk.s

change to this PR 

edit the properties to have negative gravity

set it to be dynamic.

quit 'create' mode.

touch the circle ( cylinder ) with the mouse

observe it is now the size it was before being set to disc.

libraries/entities/src/EntityItem.cpp
```diff
@@ -1431,11 +1431,11 @@ const Transform EntityItem::getTransformToCenter(bool& success) const {
...
 
 void EntityItem::setDimensions(const glm::vec3& value) {
-    if (value.x <= 0.0f || value.y <= 0.0f || value.z <= 0.0f) {
-        return;
-    }
-    if (_dimensions != value) {
-        _dimensions = value;
+    // NOTE: setDimensions() is used by internal calls whereas
+    // updateDimensions() is used for changes from external sources (scripts, network)
+    glm::vec3 newDimensions = glm::max(value, glm::vec3(MIN_ENTITY_DIMENSION));
```
ZappoMan: why are we introducing a new MIN_ENTITY_DIMENSION? how did you pick this number?

Since we now support IPD to properly shrink you down, shouldn't that value be quite small? 1 nano-meter? even smaller?

libraries/entities/src/ShapeEntityItem.cpp
```diff
@@ -249,13 +275,9 @@ void ShapeEntityItem::computeShapeInfo(ShapeInfo& info) {
...
-        case entity::Shape::Quad: {
-            // Not in GeometryCache::buildShapes, unsupported.
-            _collisionShapeType = SHAPE_TYPE_ELLIPSOID;
-            //TODO WL21389: Add a SHAPE_TYPE_QUAD ShapeType and treat 
-            // as a special box (later if desired support)
-        }
-        break;
+        case entity::Shape::Quad:
+            // Quads are deprecated, but just in case we get here
+            // fall through to Cylinder case
```
ZappoMan: you mean cube...

are they really deprecated? or are the supported but implemented using the other shape?

libraries/entities/src/ShapeEntityItem.cpp
```diff
@@ -196,6 +209,19 @@ void ShapeEntityItem::setColor(const QColor& value) {
     setAlpha(value.alpha());
 }
 
+void ShapeEntityItem::setDimensions(const glm::vec3& value) {
+	// HACK for backward compatibility of Circle and Quad shapes (2017.10.26)
```
ZappoMan: is this really a HACK? isn't this just a simple way to implement the original intent of the original feature? Maybe we don't need to have such a negative comment that misleads future developers into thinking this needs to be changed,

libraries/shared/src/HashKey.h
ZappoMan: Couldn't we implement this using Qt's built in hash functions? and wrap them so that they are bt friendly?

scripts/system/html/entityProperties.html
```diff
@@ -58,7 +58,6 @@
                         <option value="Octagon">Octagon</option>
                         <option value="Cylinder">Cylinder</option>
                         <option value="Cone">Cone</option>
-                        <option value="Circle">Circle</option>
```
ZappoMan: why are we removing this from the html? now that we correctly implement this to be a very flat cylinder, isn't it a convenience to the users to support this type properly?

unpublishedScripts/marketplace/shapes/modules/createPalette.js
```diff
@@ -299,21 +299,6 @@ CreatePalette = function (side, leftInputs, rightInputs, uiCommandCallback) {
...
-            },
-            {
-                icon: {
-                    properties: {
-                        url: Script.resolvePath("../assets/create/circle.fbx"),
-                        dimensions: { x: 0.024, y: 0.0005, z: 0.024 },
-                        localRotation: Quat.fromVec3Degrees({ x: 90, y: 0, z: 0 })
-                    }
-                },
-                entity: {
```
ZappoMan: why are we removing this? this was added by the shapes team because it's a feature that other similar tools have... forcing content creators to make cylinders seems like extra work we're putting on them. we should keep this type

libraries/entities/src/EntityItem.h
```diff
@@ -52,6 +52,7 @@ typedef std::shared_ptr<EntityTreeElement> EntityTreeElementPointer;
 using EntityTreeElementExtraEncodeDataPointer = std::shared_ptr<EntityTreeElementExtraEncodeData>;
 
 
+const float MIN_ENTITY_DIMENSION = 0.001f; // 1mm
```
ZappoMan: if my concern is not clear about this, let me clarify. With this code you've added, ALL content that's smaller than 1mm on any dimension will suddenly change size... and quite possibly distort, since you'll only clamp the dimensions smaller than 1mm -- this will definitely appear like broken content to people with small objects. we can't do this. maybe if you picked a much smaller size it would be ok...

unpublishedScripts/marketplace/shapes/modules/createPalette.js
ctrlaltdavid: The Shapes app just provided all (most) of the Shape entity types that happened to be available at the time of development. It already includes a cylinder and it's easy to squash it flat with the Shapes UI.

As part of removing Circle from Shapes app the "../assets/create/circle.fbx" file should also be removed.

AndrewMeadows:
Closing this.

Turns out it will be easier to fix the problems with Circles than to remove them.  Meanwhile, some people might find them useful.

PR #11710, opened 2017-10-28
More Louder #1

kencooke:
This PR increases the global HRTF gain by +6dB. This has two consequences:
1) Before, a spatialized source was approximately the same loudness as a non-spatialized (stereo, non-diegetic) source, up to the onset of distance attenuation (<= 1 meter). After, a nearby spatialized source will be 6dB louder than a non-spatialized source. So the relative mix levels have changed.
2) The peak HRTF gain is now +12dB instead of +6dB (see graph below), so a nearby avatar talking at full scale may result in more dynamic-range compression. Multiple loud nearby sources will hit the limiter harder, with gain reductions of -10dB or more.

ipsilateral_hrtf

### DIFF
--- a/libraries/audio/src/AudioHRTF.h
+++ b/libraries/audio/src/AudioHRTF.h
@@ -21,7 +21,7 @@ static const int HRTF_TABLES = 25;      // number of HRTF subjects
 static const int HRTF_DELAY = 24;       // max ITD in samples (1.0ms at 24KHz)
 static const int HRTF_BLOCK = 240;      // block processing size
 
-static const float HRTF_GAIN = 1.0f;    // HRTF global gain adjustment
+static const float HRTF_GAIN = 2.0f;    // HRTF global gain adjustment = +6dB
 
 class AudioHRTF {
 


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4533/HighFidelity-Beta-PR11710-1caf14640e725346d42d992071c9085f35c44aa9.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4533/HighFidelity-Beta-PR11710-1caf14640e725346d42d992071c9085f35c44aa9.exe

kencooke:
### Test Plan

This PR needs an audio-savvy tester to verify that this change is desirable.
1) Using both client and server from this PR, avatar voices should be uniformly louder.
2) When an avatar is very close (< 1m) and talking loudly, the resulting dynamic-range compression should not be annoying

kencooke:
After more testing, I like this change but I think it can be done better.

Boosting all avatar audio and letting extremely close (or multiple) sources go into compression, seems like a good solution to the "make things louder" problem.

But my opinion is that:
1) It should be adjustable rather than a hardcoded change. It should be easy to disable.
2) It should only apply to avatar voices,  not other sounds (who's level is already controlled by the script). 
This would let you to adjust the relative balance of the mix.

kencooke:
Superseded by #11718

PR #11711, opened 2017-10-28
More Louder #2

kencooke:
This PR changes the distance-attenuation curve for spatialized sounds. Spatialized sources will be unattenuated for the first 2m distance instead of 1m. This has the side-effect that far sounds will be +6dB louder.

### DIFF
--- a/assignment-client/src/audio/AudioMixerSlave.cpp
+++ b/assignment-client/src/audio/AudioMixerSlave.cpp
@@ -523,7 +523,7 @@ float computeGain(const AvatarAudioStream& listeningNodeStream, const Positional
     }
 
     // distance attenuation
-    const float ATTENUATION_START_DISTANCE = 1.0f;
+    const float ATTENUATION_START_DISTANCE = 2.0f;
     float distance = glm::length(relativePosition);
     assert(ATTENUATION_START_DISTANCE > EPSILON);
     if (distance >= ATTENUATION_START_DISTANCE) {
--- a/libraries/audio-client/src/AudioClient.cpp
+++ b/libraries/audio-client/src/AudioClient.cpp
@@ -1815,7 +1815,7 @@ float AudioClient::azimuthForSource(const glm::vec3& relativePosition) {
 
 float AudioClient::gainForSource(float distance, float volume) {
 
-    const float ATTENUATION_BEGINS_AT_DISTANCE = 1.0f;
+    const float ATTENUATION_BEGINS_AT_DISTANCE = 2.0f;
 
     // I'm assuming that the AudioMixer's getting of the stream's attenuation
     // factor is basically same as getting volume


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4536/HighFidelity-Beta-PR11711-4f52c1962a434b363acb38d65b17de22fe0d2dc9.dmg

kencooke:
Jenkins, test this please

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4539/HighFidelity-Beta-PR11711-4f52c1962a434b363acb38d65b17de22fe0d2dc9.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4539/HighFidelity-Beta-PR11711-4f52c1962a434b363acb38d65b17de22fe0d2dc9.exe

kencooke:
### Test Plan
This PR needs an audio-savvy tester to verify that this change is desirable.
Using both client and server from this PR, avatar voices should seem slightly louder, but should still attenuate with distance in a way that seems physically plausible.

ZappoMan:
@kencooke - do you have any resources or research that reference how attenuation "in the real world" would work relative to this? Or is this such an arcane topic that we're best going with "seems physically plausible"? I am sure we just pulled 1m out of thin air (or some darker location)... how do we judge 2m vs 1m?

kencooke:
@ZappoMan The most "physically correct" attenuation (for outdoors) would be to increase at +6dB/log2(distance) until fairly close to your ear, where the approximation of point-source in the far field breaks down. We saturate at 1m for two reasons, one is that people rarely get closer than that (so you don't really notice) and the other is to make good use of the limited dynamic range (max is your system volume, beyond that volume becomes nonlinear as the compressor kicks in. 

2m is far enough that you might notice some loss of immersion, but it rescales the range to be 6dB louder (until you max at system volume). There is a trivial way to get more effective range... turn up your system volume. But I agree that the mismatch between our app and other apps becomes unpleasant, so we have to make a tradeoff here. Same thing with "realistic" vs "its hard to hear someone far away, not facing me, with a quiet mic."  Realistic might not be desirable.

Far sounds can already be made louder by changing the 6dB falloff to something smaller (which we support) or using a different kind of curve. Note that indoors, the "correct" attenuation is way more complicated due to things like reverberation and occlusion.

kencooke:
After more testing, I personally don't like this change.

As an experiment, I scripted different recordings (located at height of 2m) and walked around the scene. It seems unnatural how the level stops getting louder as you get extremely close, and makes it difficult to "locate the invisible source". Sounds have the directionality of a point-source but attenuate like 2m wide area-source. I think this is a step in the wrong direction.

PR #11717, opened 2017-10-31
oblique mirror projection for secondary camera

dback2:
This adds an optional flag obliqueMirrorProjection to the SecondaryCamera that when set to true and with a valid attachedEntityID set will call setObliqueMirrorProjection instead of the existing setPosition/setOrientation/setProjection calls for the view frustum in the run function. In SecondaryCamera::setObliqueMirrorProjection it will use the entity properties from the attachedEntityID for the mirror to set the position and orientation, calculate the standard perspective projection matrix, and then apply the oblique near plane clipping to that projection and set it to the view frustum.

This is for use in the mirror client script and is currently experimental as it has not yet yielded the desired results. You can see the current results using this mirror JSON - https://hifi-content.s3.amazonaws.com/davidback/development/mirror/mirrorsmall2.json - which enables the obliqueMirrorProjection flag and sets the attachedEntityID to the mirror entity ID.

Previously all of the math from SecondaryCamera::setObliqueMirrorProjection was attempted in this mirror client script - https://hifi-content.s3.amazonaws.com/davidback/development/mirror/mClient.js - in the updateSpectatorCamera, getPerspectiveProjection, and obliqueNearPlaneClipping functions and then set as a custom projection matrix in SecondaryCamera.

Test Plan:

As of now it is not currently being used in world so no functionality changes should be noticed.

-Test spectator camera (get Spectator Camera from marketplace, go into HMD mode, and use Spectator app on tablet) and verify it is unaffected by this change and still working as expected.
-Test existing mirrors (such as those in the A&J store in empty domain) and verify their behavior is unaffected by this change and still working as expected.


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4570/HighFidelity-Beta-PR11717-530fa5da6826431590f3d56ea27300cd956751e4.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4570/HighFidelity-Beta-PR11717-530fa5da6826431590f3d56ea27300cd956751e4.exe

interface/src/SecondaryCamera.cpp
```diff
@@ -79,7 +81,10 @@ class SecondaryCameraJob {  // Changes renderContext for our framebuffer and vie
                 srcViewFrustum.setPosition(_position);
                 srcViewFrustum.setOrientation(_orientation);
             }
-            srcViewFrustum.setProjection(glm::perspective(glm::radians(_vFoV), ((float)args->_viewport.z / (float)args->_viewport.w), _nearClipPlaneDistance, _farClipPlaneDistance));
+            if (_useCustomProjection)
+                srcViewFrustum.setProjection(_customProjection);
```
SamGondelman: This would work for testing, but I don't think you're going to want to be setting the projection matrix from a script.  You have the attached entity ID here, so you should be able to calculate the matrix in C++.  Otherwise, since the script updates slower than the rendering, you'll see the mirror rendering lag.  Also, missing curly braces here.

dback2: Good point I did not think about that, thanks for the suggestion. Before trying to do this custom projection matrix we were previously already setting position, orientation, near/far clip distance, and fov from script, but I will try do everything here instead and see what happens. After many attempts doing oblique near plane clipping to our perspective projection in this script https://hifi-content.s3.amazonaws.com/davidback/development/mirror/mClient.js (lines 181, 202-276) and setting it as this custom projection matrix has not yet given the desired results.

SamGondelman: One thing to try would be flipping your rows/columns, I can look into it in more depth later but it looks like you might have them flipped (https://stackoverflow.com/questions/17717600/confusion-between-c-and-opengl-matrix-order-row-major-vs-column-major)

dback2: Did you mean flipped in the obliqueNearPlaneClipping function in the script? We wouldn't want row major in that matrix[2] is row 2 column 0 and matrix[6] is row 2 column 1? Either way I tried flipping them in that function and unfortunately it did not give better results.

SamGondelman: In row major, matrix[0] is row 0, col 0, matrix[1] is row 0, col 2, matrix[2] is row 0, col 2, etc. (moves along one row before going to the next).

Hm...it would be helpful to narrow down the issue.  The problem is either in getPerspectiveProjection or obliqueNearPlaneClipping.  If you were to move the calculations to c++, you could use the result from glm::perspective as your original projection matrix, so that at least you know that is correct.

dback2: Ah yes you're right, read it wrong my bad.

I suspect the problem is in obliqueNearPlaneClipping because at one point I was checking to verify the result from getPerspectiveProjection matched the previously calculated glm::perspective call in code by toggling the useCustomProjection flag. But yes as part of currently trying to move all the calculations to code I would go back to building the non-obliqued projection using glm::perspective like before and then attempt the oblique clip plane calculations on that.

dback2:
Updated to remove the custom projection matrix and instead do all position, orientation, and most importantly projection math that we were doing in the mirror client script in code instead. The results seem to be unfortunately nearly the same as what I had using the custom projection set from script. Would you by chance have any time to maybe take a look at the current behavior using the JSON above and see if you might know what's going wrong if you have a free moment? Much appreciated.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4609/HighFidelity-Beta-PR11717-7e4516f79c95308d62693007e5d2eb683150e022.dmg

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4615/HighFidelity-Beta-PR11717-7e4516f79c95308d62693007e5d2eb683150e022.dmg

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4617/HighFidelity-Beta-PR11717-8c0666415bd14622557fb01945cb619e77c75e90.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4617/HighFidelity-Beta-PR11717-8c0666415bd14622557fb01945cb619e77c75e90.exe

PR #11724, opened 2017-11-01
Fix serverScripts not being able to see entities in parent/child tree

ctrlaltdavid:


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4595/HighFidelity-Beta-PR11724-328fc29bc595faa62ef1ae69604d6fabebca7ece.dmg

ctrlaltdavid:
**Test Plan**

Run the PR Sandbox and Interface.

Check that the Stopwatch from the Marketplace works:
- A stopwatch should rez in front of you.
- The top button should start/stop timing. When running, each second a tick is heard and the second moves.
- The side button resets the stopwatch.

Restart Interface and Sandbox and check that the stopwatch still works.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4595/HighFidelity-Beta-PR11724-328fc29bc595faa62ef1ae69604d6fabebca7ece.exe

ctrlaltdavid:
QA ready

cain-kilgore:
Testing.

cain-kilgore:
# Test Passed
Test | Result
--- | ---
A stopwatch should rez in front of you. | Pass
The top button should start/stop timing. When running, each second a tick is heard and the second moves. | Pass
The side button resets the stopwatch. | Pass

assignment-client/src/entities/EntityTreeSendThread.cpp
```diff
@@ -423,12 +425,14 @@ bool EntityTreeSendThread::traverseTreeAndBuildNextPacketPayload(EncodeBitstream
...
     nodeData->stats.encodeStarted();
+    auto entityNodeData = static_cast<EntityNodeData*>(_node.toStrongRef()->getLinkedData());
     while(!_sendQueue.empty()) {
         PrioritizedEntity queuedItem = _sendQueue.top();
         EntityItemPointer entity = queuedItem.getEntity();
         if (entity) {
             // Only send entities that match the jsonFilters, but keep track of everything we've tried to send so we don't try to send it again
-            if (entity->matchesJSONFilters(jsonFilters)) {
+            if (entity->matchesJSONFilters(jsonFilters) || entityNodeData->isEntityFlaggedAsExtra(entity->getID())) {
+                entityNodeData->insertSentFilteredEntity(entity->getID());
```
AndrewMeadows: @ctrlaltdavid are you able to explain what is "FlaggedAsExtra" and why it is necessary to check about it here?

Looking at the EntityNodeData implementation I see that `EntityNodeData::_sentFilteredEntities` is a `QSet< QHash< QUuid, QSet<QUuid> > >` which means the performance of this check is questionable (why he didn't just us a big map is a mystery to me).

Meanwhile, `EntityItem::matchesJSONFilters()` appears to be a misnomer.  If `jsonFilters` is empty then it returns `true`.  Which means it is considered to be "sentFiltered" when there is no filter.  Do we really need to track all of these sent entities when the filter feature isn't being used (which is probably the case for the vast majority of `EntityNodeData` instances)?  It seems wasteful.

ZappoMan: @birarda @Atlante45 @huffman - Hey server team... these questions seem to relate to your implementation of ESS and JSON queries - can you respond.

ctrlaltdavid: @AndrewMeadows

The JSON filter for entity server scripts is created in `EntityScriptServer::run()` as `{"flags":{"includeAncestors":true,"includeDescendants":true},"serverScripts":"+"}`.
The goal of the JSON filter is to send the entity that the server script resides in plus its ancestors and descendants so that the server script can modify them.
This is, as far as I can see, the only JSON filter currently used.

In the highlighted lines above...
- `entity->matchesJSONFilters(jsonFilters)` lets through all entities if there is no filter (or a custom filter other than that above) or if the above filter is being used then only entities with non-null `serverScripts` property.
- `entityNodeData->isEntityFlaggedAsExtra(entity->getID()` lets through ancestor/descendant entities. These are flagged as "extra" in `EntityTreeSendThread::preDistributionProcessing()` via `addAncestorsToExtraFlaggedEntities()` and `addDescendantsToExtraFlaggedEntities()`.
- `entityNodeData->insertSentFilteredEntity(entity->getID());` enables the flag-as-extra code in `EntityTreeSendThread::preDistributionProcessing()` to run.

The trick is that `addAncestorsToExtraFlaggedEntities()` and `addDescendantsToExtraFlaggedEntities()` are done in the frame after the set of entities has already been processed (in which the entity with the `serverScripts` property is sent but not its ancestors or descendants because they haven't been flagged yet. Hence the use of `isFullScene` to cause a resend. 

The above make things work the same as how it worked before the `EntityPriorityQueue` and related changes of 7218.


I could look at using a more efficient data structure for `EntityNodeData::_sentFilteredEntities` if you'd like.

ctrlaltdavid:
I've had some further thoughts about this which I'm going to look into.

assignment-client/src/entities/EntityTreeSendThread.cpp
```diff
@@ -423,12 +425,14 @@ bool EntityTreeSendThread::traverseTreeAndBuildNextPacketPayload(EncodeBitstream
     uint64_t sendTime = usecTimestampNow();
     auto nodeData = static_cast<OctreeQueryNode*>(params.nodeData);
     nodeData->stats.encodeStarted();
+    auto entityNodeData = static_cast<EntityNodeData*>(_node.toStrongRef()->getLinkedData());
```
Atlante45: You need to pin the shared pointer here otherwise it could get deleted while you're using `entityNodeData`.
```cpp
auto sharedNode = _node.toStrongRef(); // Pin the shared ptr so it doesn't get destructed
auto entityNodeData = static_cast<EntityNodeData*>(sharedNode->getLinkedData());
```

ctrlaltdavid: Thanks. Done!

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4690/HighFidelity-Beta-PR11724-05275afb4fe034461354cf1f8e0ebd3ba70969a6.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4690/HighFidelity-Beta-PR11724-05275afb4fe034461354cf1f8e0ebd3ba70969a6.exe

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

Run the PR Sandbox and Interface. **Check**

Check that the Stopwatch from the Marketplace works:
A stopwatch should rez in front of you. **Check**
The top button should start/stop timing. When running, each second a tick is heard and the second moves. **Check**
The side button resets the stopwatch. **Check**
Restart Interface and Sandbox and check that the stopwatch still works. **Check, though the watch doesn't "keep going", it requires being being pressed to continue/start. Though I would assume this is expected.**

**Pass** - tested in HMD & Desktop.

Atlante45:
@birarda can you check the logic? It looks right to me, but since you designed this, you'll probably have a better eye for issues.

birarda:
LGTM, thanks for investigating and fixing this

PR #11736, opened 2017-11-02
scaleToFit() the model before updating render items

AndrewMeadows:
- fix bug causing ModelEntityItem to not update its scale in a timely fashion

### DIFF
--- a/libraries/render-utils/src/Model.h
+++ b/libraries/render-utils/src/Model.h
@@ -262,6 +262,8 @@ public:
 
     Q_INVOKABLE MeshProxyList getMeshes() const;
 
+    void scaleToFit();
+
 public slots:
     void loadURLFinished(bool success);
 
@@ -320,7 +322,6 @@ protected:
     virtual void initJointStates();
 
     void setScaleInternal(const glm::vec3& scale);
-    void scaleToFit();
     void snapToRegistrationPoint();
 
     void computeMeshPartLocalBounds();
--- a/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
+++ b/libraries/entities-renderer/src/RenderableModelEntityItem.cpp
@@ -215,6 +215,7 @@ void RenderableModelEntityItem::updateModelBounds() {
         model->setScaleToFit(true, getDimensions());
         model->setSnapModelToRegistrationPoint(true, getRegistrationPoint());
         updateRenderItems = true;
+        model->scaleToFit();
     }
 
     bool success;


libraries/render-utils/src/Model.h
```diff
@@ -320,7 +322,6 @@ public slots:
     virtual void initJointStates();
 
     void setScaleInternal(const glm::vec3& scale);
-    void scaleToFit();
```
AndrewMeadows: `Model::scaleToFit()` needs to be a `public` method (not `protected`) so it can be called in `RenderableModelEntityItem` context.

AndrewMeadows:
Test Plan:

(1) Create a ModelEntityItem

(2) In desktop mode use **create.js** to select the object and drag the selection handles around to adjust its dimensions.  This will change the object's dimensions but will also change its position in the world.
--> When using hifi/master interface the object's scale will sometimes not update even though its position changes.  Sometimes you can get it in a situation where its visible scale is outright incorrect... until you deselect it.
--> when using this PR the visible dimensions should always update in a timely manner.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4651/HighFidelity-Beta-PR11736-490483fde889e31ddee4c2697e4716eccc4bb792.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4651/HighFidelity-Beta-PR11736-490483fde889e31ddee4c2697e4716eccc4bb792.exe

starphys:
Testing

starphys:
FAIL
The visible dimensions of the model updated in a timely manner consistently. However, the model always stayed rooted in the same spot on the grid while resizing it, but I expected it to change position in the world. I also expected to be able to move it smoothly, but oving it around was difficult, and it frequently would stick in place or not move in the direction I was moving my mouse in.

Nex-Pro:
@AndrewMeadows  I think this is a **PASS**

The model center moves in a predictable and expected manner when the scale handles are manipulated.

If ( looking at the front of the ent ) I grab the right hand side middle handle and drag to the my right  then drag the left side middle scale handle also to the right the net result is that the ent is moved to the right.
It always updates the scale correctly.

Please confirm this is the intended behaviour.

Nex-Pro:
Same for fore/aft

AndrewMeadows:
Yes, that sounds right: dragging the edit box handles around moves the corners of the box.  The object remains inside this bounding box.

PR #11770, opened 2017-11-08
sort and throttle UpdateRenderables work... when necessary

AndrewMeadows:
- reduce gamerate lag when large number of renderables need updates

### DIFF
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -38,6 +38,22 @@
 
 #include "RenderableWebEntityItem.h"
 
+// implement these methods BEFORE including PrioritySortUtil.h
+namespace PrioritySortUtil {
+    glm::vec3 getObjectPosition(const EntityRendererPointer& object) {
+        return object->getEntity()->getPosition();
+    }
+
+    float getObjectRadius(const EntityRendererPointer& object) {
+        return 0.5f * object->getEntity()->getQueryAACube().getScale();
+    }
+
+    uint64_t getObjectAge(const EntityRendererPointer& object) {
+        return object->getUpdateTime();
+    }
+}
+#include <PrioritySortUtil.h>
+
 size_t std::hash<EntityItemID>::operator()(const EntityItemID& id) const { return qHash(id); }
 std::function<bool()> EntityTreeRenderer::_entitiesShouldFadeFunction;
 
@@ -338,47 +354,16 @@ void EntityTreeRenderer::updateChangedEntities(const render::ScenePointer& scene
         // we expect the cost to updating all renderables to exceed available time budget
         // so we first sort by priority and update in order until out of time
         uint64_t sortStart = usecTimestampNow();
-        std::priority_queue<SortableEntityRenderer> sortedRenderables;
+        using SortableRenderer = PrioritySortUtil::Sortable<EntityRendererPointer>;
+        std::priority_queue< SortableRenderer > sortedRenderables;
         {
             PROFILE_RANGE_EX(simulation_physics, "SortRenderables", 0xffff00ff, (uint64_t)_renderablesToUpdate.size());
+            PrioritySortUtil::PriorityCalculator<EntityRendererPointer> priorityCalculator(view);
             std::unordered_map<EntityItemID, EntityRendererPointer>::iterator itr = _renderablesToUpdate.begin();
-            glm::vec3 viewCenter = view.getPosition();
-            glm::vec3 forward = view.getDirection();
-            const float OUT_OF_VIEW_PENALTY = -10.0f;
             while (itr != _renderablesToUpdate.end()) {
-                // priority = weighted linear combination of:
-                //   (a) apparentSize
-                //   (b) proximity to center of view
-                //   (c) time since last update
-                EntityItemPointer entity = itr->second->getEntity();
-                glm::vec3 entityPosition = entity->getPosition();
-                glm::vec3 offset = entityPosition - viewCenter;
-                float distance = glm::length(offset) + 0.001f; // add 1mm to avoid divide by zero
-
-                float diameter = entity->getQueryAACube().getScale();
-                float apparentSize = diameter / distance;
-                float cosineAngle = glm::dot(offset, forward) / distance;
-                float age = (float)(sortStart - itr->second->getUpdateTime()) / (float)(USECS_PER_SECOND);
-
-                // NOTE: we are adding values of different units to get a single measure of "priority".
-                // Thus we multiply each component by a conversion "weight" that scales its units relative to the others.
-                // These weights are pure magic tuning and should be hard coded in the relation below,
-                // but are currently exposed for anyone who would like to explore fine tuning:
-                const float APPARENT_SIZE_COEFFICIENT = 1.0f;
-                const float CENTER_SORT_COEFFICIENT = 0.5f;
-                const float AGE_SORT_COEFFICIENT = 0.25f;
-                float priority = APPARENT_SIZE_COEFFICIENT * apparentSize
-                    + CENTER_SORT_COEFFICIENT * cosineAngle
-                    + AGE_SORT_COEFFICIENT * age;
-
-                // decrement priority of things outside keyhole
-                if (distance > view.getCenterRadius()) {
-                    if (!view.sphereIntersectsFrustum(entityPosition, 0.5f * diameter)) {
-                        priority += OUT_OF_VIEW_PENALTY;
-                    }
-                }
-
-                sortedRenderables.push(SortableEntityRenderer(itr->second, priority));
+                float priority = priorityCalculator.computePriority(itr->second);
+                SortableRenderer entry(itr->second, priority);
+                sortedRenderables.push(entry);
                 ++itr;
             }
         }
@@ -398,7 +383,7 @@ void EntityTreeRenderer::updateChangedEntities(const render::ScenePointer& scene
             std::unordered_map<EntityItemID, EntityRendererPointer>::iterator itr;
             size_t numSorted = sortedRenderables.size();
             while (!sortedRenderables.empty() && usecTimestampNow() < expiry) {
-                const EntityRendererPointer& renderer = sortedRenderables.top().renderer;
+                const EntityRendererPointer& renderer = sortedRenderables.top().getObject();
                 renderer->updateInScene(scene, transaction);
                 _renderablesToUpdate.erase(renderer->getEntity()->getID());
                 sortedRenderables.pop();
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -12,6 +12,7 @@
 #include "EntityTreeRenderer.h"
 
 #include <glm/gtx/quaternion.hpp>
+#include <queue>
 
 #include <QEventLoop>
 #include <QScriptSyntaxCheckResult>
@@ -272,7 +273,7 @@ void EntityTreeRenderer::addPendingEntities(const render::ScenePointer& scene, r
     }
 }
 
-void EntityTreeRenderer::updateChangedEntities(const render::ScenePointer& scene, render::Transaction& transaction) {
+void EntityTreeRenderer::updateChangedEntities(const render::ScenePointer& scene, const ViewFrustum& view, render::Transaction& transaction) {
     PROFILE_RANGE_EX(simulation_physics, "Change", 0xffff00ff, (uint64_t)_changedEntities.size());
     PerformanceTimer pt("change");
     std::unordered_set<EntityItemID> changedEntities;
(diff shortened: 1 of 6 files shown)


libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -286,21 +287,129 @@ void EntityTreeRenderer::updateChangedEntities(const render::ScenePointer& scene
...
+    {
+        PROFILE_RANGE_EX(simulation_physics, "CopyRenderables", 0xffff00ff, (uint64_t)changedEntities.size());
+        if (_renderablesToUpdate.empty()) {
+            for (const auto& entityId : changedEntities) {
+                auto renderable = renderableForEntityId(entityId);
+                if (!renderable) {
+                    continue;
+                }
+                _renderablesToUpdate.insert({ entityId, renderable });
+            }
```
AndrewMeadows: When `_renderablesToUpdate` starts off empty then the logic to add to it is simple (above).  When not empty the logic to update it is more complicated (below).  In particular, we must watch for old entries whose renderables are no longer valid.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -286,21 +287,129 @@ void EntityTreeRenderer::updateChangedEntities(const render::ScenePointer& scene
...
+                    _renderablesToUpdate.insert({ entityId, renderable });
+                }
+            }
         }
-        _renderablesToUpdate.insert({ entityId, renderable });
     }
 
-    if (!_renderablesToUpdate.empty()) {
+    float expectedUpdateCost = _avgRenderableUpdateCost * _renderablesToUpdate.size();
+    const float MAX_UPDATE_RENDERABLES_TIME_BUDGET = 2 * USECS_PER_MSEC;
```
AndrewMeadows: When we have too many renderables to update the gameloop rate will drop.  To solve this we can sort the updates and timebox their execution, however the sort can be expensive so we try to only do it when necessary.  We measure the average time-cost/update and use that to estimate how long it will take us to process the pending updates.  If we don't have enough time then we'll use the sort strategy, otherwise we just update everything with no time-boxing.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -286,21 +287,129 @@ void EntityTreeRenderer::updateChangedEntities(const render::ScenePointer& scene
...
+        	    // decrement priority of things outside keyhole
+        	    if (distance > view.getCenterRadius()) {
+            	    if (!view.sphereIntersectsFrustum(entityPosition, 0.5f * diameter)) {
+                	    priority += OUT_OF_VIEW_PENALTY;
+            	    }
+        	    }
+
+                sortedRenderables.push(SortableEntityRenderer(itr->second, priority));
+                ++itr;
+            }
```
AndrewMeadows: This logic was copied from `AvatarManager` where we prioritize and sort `AvatarData` updates from avatar-mixer.

AndrewMeadows:
Test plan:

(1) Login to a relatively empty domain using the latest build from hifi/master.

(2) Run this butterflyOverload.js script, which will create 600 butterflies with lifetimes of 180 seconds (3 min).

(3) Look around to make sure you have seen all of the butterflies and then walk out of the swarm and orient your avatar to look away as a few number of objects.

(4) Make stats visible: **Developer --> Timing --> Performance Timer --> Show Timing Details**

(5) Note the value of the **Game Rate** stat, third stat down on leftmost column.  It is expected that the 600 animating butterflies will cause lag in this stat.

(6) Turn around and walk into the middle of the swarm and get up close to one particular butterfly.  Note how well the animation is updating and also note the **Game Rate** stat value.

(7) Repeat the butterflyOverload test using the interface from this PR and note the time in **Game Rate** when looking away from the pile of butterflies: you should see a higher frame rate than in (5).

(8) Turn around and walk up to one of the butterflies: it should animate at a decent rate, but other butterflies further away near the edge of your window may be updating at a lower rate.  Note the **Game Rate** stat value: is should be higher than in (6).

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4758/HighFidelity-Beta-PR11770-0e3173236c71fcfe8dfc54070977603a0c52e94d.dmg

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -286,21 +287,129 @@ void EntityTreeRenderer::updateChangedEntities(const render::ScenePointer& scene
...
+            std::unordered_map<EntityItemID, EntityRendererPointer>::iterator itr = _renderablesToUpdate.begin();
+            glm::vec3 viewCenter = view.getPosition();
+            glm::vec3 forward = view.getDirection();
+            const float OUT_OF_VIEW_PENALTY = -10.0f;
+            while (itr != _renderablesToUpdate.end()) {
+        	    // priority = weighted linear combination of:
+        	    //   (a) apparentSize
+        	    //   (b) proximity to center of view
+        	    //   (c) time since last update
+                EntityItemPointer entity = itr->second->getEntity();
```
ZappoMan: something is weird with your formatting on this

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -286,21 +287,129 @@ void EntityTreeRenderer::updateChangedEntities(const render::ScenePointer& scene
...
+        	    float diameter = entity->getQueryAACube().getScale();
+        	    float apparentSize = diameter / distance;
+        	    float cosineAngle = glm::dot(offset, forward) / distance;
+        	    float age = (float)(sortStart - itr->second->getUpdateTime()) / (float)(USECS_PER_SECOND);
+
+        	    // NOTE: we are adding values of different units to get a single measure of "priority".
+        	    // Thus we multiply each component by a conversion "weight" that scales its units relative to the others.
+        	    // These weights are pure magic tuning and should be hard coded in the relation below,
+        	    // but are currently exposed for anyone who would like to explore fine tuning:
+                const float APPARENT_SIZE_COEFFICIENT = 1.0f;
```
ZappoMan: indent?

libraries/entities-renderer/src/EntityTreeRenderer.cpp
ZappoMan: Any chance this can be generalized into a template so we can DRY up this code and use common implementation. I believe we have similar code in AvartarMixer too

libraries/entities-renderer/src/EntityTreeRenderer.h
```diff
@@ -51,6 +51,14 @@ using ModelWeakPointer = std::weak_ptr<Model>;
 
 using CalculateEntityLoadingPriority = std::function<float(const EntityItem& item)>;
 
+class SortableEntityRenderer {
+public:
+    SortableEntityRenderer(EntityRendererPointer r, float p) : renderer(r), priority(p) {}
```
ZappoMan: coding standard variable names

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4760/HighFidelity-Beta-PR11770-e697379d29ad5fb606bc8c4ef0b21a956b29025f.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4760/HighFidelity-Beta-PR11770-e697379d29ad5fb606bc8c4ef0b21a956b29025f.exe

AndrewMeadows:
I'm adding the DO_NOT_MERGE label.  This is not ready for test yet, since I'll be modifying the code that performs the sorting.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4797/HighFidelity-Beta-PR11770-dbb9746878a399d67730869ede64447192f9c408.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4797/HighFidelity-Beta-PR11770-dbb9746878a399d67730869ede64447192f9c408.dmg

Nex-Pro:
Testing Win10

Nex-Pro:
still in progress.

Nex-Pro:
### PASS

Spent some time iterating over this with @AndrewMeadows and had to hack a little on the script.

All requirements are met.

AndrewMeadows:
I have code ready that uses the `PrioritySortUtil` template for the avatar updates.  I'll create a PR for that after this one merges into master.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4824/HighFidelity-Beta-PR11770-35962a3ca46ae519049733c9ddfb74d19dc3ac71.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4824/HighFidelity-Beta-PR11770-35962a3ca46ae519049733c9ddfb74d19dc3ac71.exe

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -37,6 +38,22 @@
 
 #include "RenderableWebEntityItem.h"
 
+// implement these methods BEFORE including PrioritySortUtil.h
+namespace PrioritySortUtil {
+    glm::vec3 getObjectPosition(const EntityRendererPointer& object) {
```
ZappoMan: this seems odd to me... shouldn't these be part of the template.

libraries/shared/src/PrioritySortUtil.h
```diff
@@ -0,0 +1,127 @@
...
+#define hifi_PrioritySortUtil_h
+
+#include <glm/glm.hpp>
+#include "ViewFrustum.h"
+
+namespace PrioritySortUtil {
+    // PrioritySortUtil is a helper template for sorting 3D objects relative to a ViewFrustum.
+    // To use this utility:
+    //
+    // (1) Declare and implement the following methods for your "object" type T:
```
ZappoMan: why aren't these part of your template definition... this predefining seems weird and would limit the ability to use this template multiple times in the same CPP file... this feels wrong.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4841/HighFidelity-Beta-PR11770-4b7d74b01b931f3ab8ff20145729c4009bff0319.dmg

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4853/HighFidelity-Beta-PR11770-dacfa7e2f76084ebdbe7ff920c3401734dc0897a.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4853/HighFidelity-Beta-PR11770-dacfa7e2f76084ebdbe7ff920c3401734dc0897a.exe

AndrewMeadows:
This needs a retest because the implementation of the fix has been refactored.

Nex-Pro:
@AndrewMeadows Check.

AndrewMeadows:
@Nex-Pro, uh... are you trying to say that you tested it and it passed?  Or something else?

Nex-Pro:
@AndrewMeadows  lol @ ***"Rumor has it on a decent 4GHz i7 you may need 2k or more to get the game loop to noticeably slow down: adjust the script as necessary."***

Nex-Pro:
@AndrewMeadows the "check" meant  I read your comment on needing a retest.

Nex-Pro:
@AndrewMeadows That issue will be addressed in version 0.0.2 of my new project "QALinter" expected to go beta in 2035. :)

Nex-Pro:
@AndrewMeadows 
We have an odd problem. If I run the https://raw.githubusercontent.com/AndrewMeadows/hifi-js/master/test/butterflyOverload.js  script and use the HMD I get slammed to 45 fps which does not recover when looking to the void. 
If I run this modified version http://s3-us-west-1.amazonaws.com/hifi-makers/pioneer-nexial.protagonist/1000Butterflies.js which has more butterflies I get a hit on the game rate but it recovers to 90 when I look to the void. So somehow the script itself is having an effect. Any ideas?

Also on first (clean ) installing the latest build it seemed to work with the linked script for few moments. Then subsequent runs of the scripts had no animations playing. ( possibly the animation itself was unavailable ) so in response I unistalled and re-installed the earlier build which seemed to work fine. 
Reinstalled ( clean ) the latest build and animations were working and stayed working.

AndrewMeadows:
@Nex-Pro, yeah, I changed the script to always update one butterfly per frame.  Apparently the script resets butterflies that get too far away, moving them back to the center.  This is putting load in the main thread that is independent of the particular failure mode that we want to check.  I'll modify the script to just create the butterflies but not move them around.

Nex-Pro:
@AndrewMeadows Cool. But I was surprised at how much the game rate was being hit just by a single script.

Nex-Pro:
@AndrewMeadows Did you inadvertently create a new JS stress test ?

AndrewMeadows:
@Nex-Pro, ok I've pared the new script down.  It now creates 1k non-moving butterflies with 90sec lifespan, with no update loop.  That should stress only the animation/renderable-update context.

I haven't yet profiled the interface when it is running the old script, but yeah: I expect it to reveal a bottleneck and probably some inefficiencies in some other context on the main-loop.

Nex-Pro:
@AndrewMeadows  Retesting.

Nex-Pro:
@AndrewMeadows I am getting a game rate of ~44 in **HMD** when most of the butterflies are in view. Recovers to 90 if I move so that I see less or none of them.  
Close-by butterflies are animating smoothly. 
It is possible to be, say a couple of meters, outside the volume occupied by the butterflies and see none of them animating smoothly. (*** they are all animating to some degree*** ).

AndrewMeadows:
@Nex-Pro, yes.  If you step out of the volume then the **angularSize** and **proximityToCenter** terms in the prioritization formula could be weak enough allow the **timeSinceLastUpdate** term to dominate, which would tend to make even the closest butterflies to have have reduced animation update rates.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -286,21 +287,112 @@ void EntityTreeRenderer::updateChangedEntities(const render::ScenePointer& scene
...
+                    continue;
+                }
+                _renderablesToUpdate.insert({ entityId, renderable });
+            }
+        } else {
+            // we weren't able to update all renderables last frame
+            // so we have to be more careful when processing changed renderables
+            std::unordered_map<EntityItemID, EntityRendererPointer>::iterator itr;
+            for (const auto& entityId : changedEntities) {
+                auto renderable = renderableForEntityId(entityId);
```
ZappoMan: can you add a comment here explaining this logic a little more. I assume that `renderableForEntityId(entityId);` will return a nulptr if the entity no longer exists... and that this logic is intended to help remove deleted entities from the `_renderablesToUpdate` list

libraries/entities-renderer/src/EntityTreeRenderer.cpp
ZappoMan: I wonder if we should have a new header file where we place all these timebox budget constants... right now, they're all in place in the code... why did you pick 2ms? how much do all the other time boxes add up to.

Let's move all these constants in a common header so we can see what the entire budget adds up to.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -286,21 +287,112 @@ void EntityTreeRenderer::updateChangedEntities(const render::ScenePointer& scene
...
             const auto& renderable = entry.second;
             renderable->updateInScene(scene, transaction);
         }
+        size_t numRenderables = _renderablesToUpdate.size() + 1; // add one to avoid divide by zero
         _renderablesToUpdate.clear();
+
+        // compute average per-renderable update cost
+        float cost = (float)(usecTimestampNow() - updateStart) / (float)(numRenderables);
+        const float blend = 0.1f;
+        _avgRenderableUpdateCost = (1.0f - blend) * _avgRenderableUpdateCost + blend * cost;
```
ZappoMan: this looks like the `SimpleMovingAverage` class or the `MovingAverage` template  implementation for 10 samples, you should use one of those instead of hand rolling.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -286,21 +287,112 @@ void EntityTreeRenderer::updateChangedEntities(const render::ScenePointer& scene
...
+            float getRadius() const override { return 0.5f * _renderer->getEntity()->getQueryAACube().getScale(); }
+            uint64_t getTimestamp() const override { return _renderer->getUpdateTime(); }
+        private:
+            const EntityRendererPointer& _renderer;
+        };
+
+        // prioritize and sort the renderables
+        uint64_t sortStart = usecTimestampNow();
+        using SortableRenderer = PrioritySortUtil::Sortable<EntityRendererPointer>;
+        std::priority_queue< SortableRenderer > sortedRenderables;
```
ZappoMan: I think the coding standard says to squish these spaces... `std::priority_queue< SortableRenderer >` should be `std::priority_queue<SortableRenderer>`

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -286,21 +287,112 @@ void EntityTreeRenderer::updateChangedEntities(const render::ScenePointer& scene
...
+        {
+            PROFILE_RANGE_EX(simulation_physics, "SortRenderables", 0xffff00ff, (uint64_t)_renderablesToUpdate.size());
+            PrioritySortUtil::Prioritizer prioritizer(view);
+            std::unordered_map<EntityItemID, EntityRendererPointer>::iterator itr = _renderablesToUpdate.begin();
+            while (itr != _renderablesToUpdate.end()) {
+                float priority = prioritizer.computePriority(PrioritizableRenderer(itr->second));
+                //SortableRenderer entry(itr->second, priority);
```
ZappoMan: what is this comment telling me?

libraries/shared/src/PrioritySortUtil.h
```diff
@@ -0,0 +1,150 @@
...
+        const T& getThing() const { return  _thing; }
+        void setPriority(float priority) { _priority = priority; }
+        bool operator<(const Sortable& other) const { return _priority < other._priority; }
+    private:
+        T _thing;
+        float _priority;
+    };
+
+    // Prioritizable isn't a template because templates can't have pure-virtual methods.
+    class Prioritizable {
```
ZappoMan: I'm still confused by what a `Sortable` is vs a `Prioritizable` the naming doesn't seem intuitive to me...

libraries/shared/src/PrioritySortUtil.h
```diff
@@ -0,0 +1,150 @@
...
+     private:
+        // Yes really: the data member is a const reference!
+        // PrioritizableThing only needs enough scope to compute a priority.
+        const Thing& _thing;
+     }
+
+(2) Loop over your things and insert each into a priority_queue:
+
+    PrioritySortUtil::Prioritizer prioritizer(viewFrustum);
+    std::priority_queue< PrioritySortUtil::Sortable<Thing> > sortedThings;
```
ZappoMan: this is part of my confusion, it feels like "sortable" and "prioritizable" are basically synonyms in this usage... and so which is which is confusing to me.

ZappoMan: even weirder that you're making a priority_queue of sortables?

libraries/shared/src/PrioritySortUtil.h
```diff
@@ -0,0 +1,150 @@
...
+
+    // Prioritizable isn't a template because templates can't have pure-virtual methods.
+    class Prioritizable {
+    public:
+        virtual glm::vec3 getPosition() const = 0;
+        virtual float getRadius() const = 0;
+        virtual uint64_t getTimestamp() const = 0;
+    };
+
+    class Prioritizer {
```
ZappoMan: why not have Prioritizer  be a specialization of `std::priority_queue<>` this would allow you to just insert objects into the `Prioritizer` and have it do the `computePriority` and the push into the actual queue. I like the general direction you're going, but it still feels like the user of the prioritizer has to do a lot of work around the outside of the class that could be done on the inside of the class.

By making the caller have so many steps, it feels like there's a higher risk of them getting one of them wrong.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4940/HighFidelity-Beta-PR11770-500f38182799cc976020f3d461a00817b1ce3533.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4940/HighFidelity-Beta-PR11770-500f38182799cc976020f3d461a00817b1ce3533.exe

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4942/HighFidelity-Beta-PR11770-2f5b7f32c6bd6ed9f6a4e4e80164bb430a6a1ee0.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4942/HighFidelity-Beta-PR11770-2f5b7f32c6bd6ed9f6a4e4e80164bb430a6a1ee0.dmg

AndrewMeadows:
This needs yet another re-test.

libraries/entities-renderer/src/EntityTreeRenderer.cpp
ZappoMan: ping? is there a reason you're hand rolling your own? why aren't you using `SimpleMovingAverage`

YuppyPlays:
Testing.

YuppyPlays:
(1) Login to a relatively empty domain using the latest build from hifi/master.
**Check**
(2) Make stats visible: Developer --> Timing --> Performance Timer --> Show Timing Details so you an watch the Game Rate stat, third down from the top of the leftmost column.
**Check**
(3) Copy this butterflyOverload.js script to local file and run it. Initially it creates 600 animating butterflies with lifespans of 3 minutes, but depending on your test machine that might not be enough to demonstrate the problem. Rumor has it on a decent 4GHz i7 you may need 2k or more to get the game loop to noticeably slow down: adjust the script as necessary.
**Check**
(4) Look around to make sure you have seen all of the butterflies and then walk out of the swarm and orient your avatar to look away from the swarm.
**Check**
Note: why turn around and face away from the butterflies? Because we want to eliminate any render-rate issues from the test. With the animated content out of view the render rate will be high, and if the game rate is still low it proves that the interface is working hard to update the animations of all the things NOT in view.
**Check**
(5) Note the value of the Game Rate stat while looking away from the butterflies. It is expected that the 600 animating butterflies will cause lag in this stat.
**Check**
(6) Turn around and walk into the middle of the swarm and get up close to one particular butterfly. Note how well or poorly the animation is updating and also note the Game Rate stat value. Remember this to compare with step (8).
**Check**
(7) Repeat the butterflyOverload test using the interface from this PR and note the time in Game Rate when looking away from the flock of butterflies you should see a higher frame rate than in (5).
**Check**
(8) Turn around and walk up to one of the butterflies: it should animate at a decent rate, but other butterflies further away near the edge of your window may be updating at a lower rate. Note the Game Rate stat value: is should be higher than in (6).
**Check**

That brought even my mighty hardware to a laggy state! - but the PR fixed this :) Distance seems to effect the framerate/smoothness of the animation per butterfly and seems to massively increase performance within this scenario. **Pass**

PR #11804, opened 2017-11-13
Fixed Entity Animation Property not applying to FBXs

amantley:
Fixed the Animation Properties:  Animation play, Animation loop, Animation hold, First frame, Last frame and Animation frame(start frame).  These were not being applied but now they are working.  

I also created a new test suite for entities so that this wouldn't just be under miscellaneous.

The Fogbugz case is [7613](https://highfidelity.fogbugz.com/f/cases/7613/Animation-Property-changes-are-not-applying-to-animated-FBXs)

The TestRail test case is [C2957](https://highfidelity.testrail.net/index.php?/cases/view/2957).


hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4878/HighFidelity-Beta-PR11804-842cddde88603469e5791ded31682d244010ba9a.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4878/HighFidelity-Beta-PR11804-842cddde88603469e5791ded31682d244010ba9a.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4944/HighFidelity-Beta-PR11804-1a435b10c16b3f9d09acb2a3861fd78afd74729a.dmg

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

Unfortunately ran into some issues, see results below. **Fail**.

TestRail Results: https://highfidelity.testrail.net/index.php?/tests/view/13342&group_by=cases:section_id&group_order=asc&group_id=733

Nex-Pro:
@amantley Welcome aboard.

Nex-Pro:
Same result as Yuppy

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4997/HighFidelity-Beta-PR11804-cc43d8115c3fff02e5aa0d878f0eeabf427bb248.dmg

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -988,6 +989,19 @@ void ModelEntityRenderer::animate(const TypedEntityPointer& entity) {
         return;
     }
 
+    //get entity model anim props
```
ZappoMan: most of the comments in the system have a space between the // and the actual comment, let's keep consistency

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -996,11 +1010,32 @@ void ModelEntityRenderer::animate(const TypedEntityPointer& entity) {
...
     auto interval = now - _lastAnimated;
     _lastAnimated = now;
-    float deltaTime = (float)interval / (float)USECS_PER_SECOND;
-    _currentFrame += (deltaTime * _renderAnimationProperties.getFPS());
+
+
+    //here we implement the looping animation property
+    //if we have played through the animation once then we hold on the last frame
+   
+    if( isLooping || ( _currentFrame < _endAnim ) ){
```
ZappoMan: coding standard space placement... this should be: 

`if(isLooping || (_currentFrame < _endAnim)) {`

libraries/entities-renderer/src/RenderableModelEntityItem.h
```diff
@@ -185,6 +185,7 @@ class ModelEntityRenderer : public TypedEntityRenderer<RenderableModelEntityItem
     bool _animating { false };
     uint64_t _lastAnimated { 0 };
     float _currentFrame { 0 };
+    float _endAnim{ 0 };
```
ZappoMan: coding standard: variable names should be descriptive... is this the ending frame?

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -1322,8 +1357,17 @@ void ModelEntityRenderer::doRenderUpdateSynchronousTyped(const ScenePointer& sce
         auto newAnimationProperties = entity->getAnimationProperties();
         if (newAnimationProperties != _renderAnimationProperties) {
             withWriteLock([&] {
+                if ( (newAnimationProperties.getCurrentFrame() != _renderAnimationProperties.getCurrentFrame()) || (newAnimationProperties.getFirstFrame() != _renderAnimationProperties.getFirstFrame()) || (newAnimationProperties.getLastFrame() != _renderAnimationProperties.getLastFrame()) || (newAnimationProperties.getRunning() && !_renderAnimationProperties.getRunning())) {
```
ZappoMan: coding standard says the lines shouldn't be this long

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -1322,8 +1357,17 @@ void ModelEntityRenderer::doRenderUpdateSynchronousTyped(const ScenePointer& sce
...
+                    if (!(newAnimationProperties.getCurrentFrame() > newAnimationProperties.getLastFrame()) && !(newAnimationProperties.getCurrentFrame() < newAnimationProperties.getFirstFrame())) {
+                        _currentFrame = newAnimationProperties.getCurrentFrame();
+                        _endAnim = _currentFrame + ( newAnimationProperties.getLastFrame() - newAnimationProperties.getFirstFrame() );
+                        _lastAnimated = 0;
+                    }
+                }else if ( _renderAnimationProperties.getLoop() && !newAnimationProperties.getLoop()) {
```
ZappoMan: coding standard spaces... should be 
` } else if (_renderAnimationProperties.getLoop() && !newAnimationProperties.getLoop()) {`

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -1322,8 +1357,17 @@ void ModelEntityRenderer::doRenderUpdateSynchronousTyped(const ScenePointer& sce
...
+                     int currentframe_mod_length = (int)(_currentFrame  - (int)(glm::floor(newAnimationProperties.getCurrentFrame()))) % ((int)(glm::floor(newAnimationProperties.getLastFrame())) - (int)(glm::floor(newAnimationProperties.getFirstFrame())) + 1);
```
ZappoMan: coding standard - use camel case for variable names

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -996,11 +1010,32 @@ void ModelEntityRenderer::animate(const TypedEntityPointer& entity) {
...
+    }
 
     {
-        int animationCurrentFrame = (int)(glm::floor(_currentFrame)) % frameCount;
+        //where are we in the currently defined animation segment?       
+        int animationCurrentFrame = (int)(glm::floor(_currentFrame - _renderAnimationProperties.getFirstFrame())) % updatedFrameCount;
+        //this gives us the absolute frame value to use by adding the first frame value.
+        animationCurrentFrame += _renderAnimationProperties.getFirstFrame();
+        
+
```
ZappoMan: please remove all this white space, one line separation is reasonable

amantley:
@ZappoMan
Thanks I'll make these changes in the morning.

ZappoMan:
Something about this PR has me scratching my head... mainly because I am not sure how this got broken, but also because I'm not sure this fix will "actually fix" the animations.... The reason I say this is because the *original implementation* (we're talking a couple years ago)... the current frame was incremented in the ModelEntityItem base class... not in the renderable. The reason for this is that the animation "should continue forward" even if no one is looking at it, even if no one is connected to the domain.

So - for example here's a test we need to add to test rails and make sure it passes (something like this)...

1) create a looping animation with 120 frames, and set it to loop and 1fps.
2) start a timer when you start the animation.
3) disconnect from the domain wait 1 minute
4) reconnect to the domain and you should observe that the animation is on frame 60.
5) connect a second client to the domain after an additional 30 seconds... you should observe that for both clients the animation is on frame 90.
6) disconnect the original client, wait 60 seconds, reconnect, you should observe: on the client that remained connected the entire time, you should have seen the animation loop from frame 120 to 1 then up to frame 30. on the second client, you should see the animation on frame 30 when you connect.

Angus, can you confirm that you fix does indeed fix the behavior to behave this way?

I would assume you need to implement the handing of _currentFrame in the base class to achieve this.

Note: The base class runs on the server and the client, and the base class instance on the server should be getting an "update" signal allowing it to move the current frame forward.

amantley:
@ZappoMan
Hi ok,  I will work on getting this functioning so that the motion continues.  Originally I made a false assumption that where _currentFrame was incremented would be the right place to fix these properties.  Thanks for all your input on this.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/5080/HighFidelity-Beta-PR11804-013d16cee9e0d30c05458e76685ae8675549efba.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5080/HighFidelity-Beta-PR11804-013d16cee9e0d30c05458e76685ae8675549efba.dmg

Nex-Pro:
Testing

Nex-Pro:
### Fail 
https://highfidelity.testrail.net/index.php?/tests/view/16478&group_by=cases:section_id&group_order=asc&group_id=733

Apart from changing from initial T pose to arms by sides pose nothing works on this.

amantley:
Hi, sorry I'm in the middle of reworking this.  I've taken away the QA ready now.

Nex-Pro:
@amantley Np. it is a tricky one.

Nex-Pro:
@amantley Side note. The input fields for the start frame etc. are not being sanitized against negative numbers. ( in edit js I think )

Atlante45:
@ZappoMan I have just spoken with @amantley to help him figure out some issues he was having with the entity server and he brought this up.
I don't think we should have the server keep track of the current frame. I understand that's what we used to do, but I don't think that's the right solution.

Instead of storing the current frame, the server should store the animation startup timestamp.
Using the startup timestamp, the fps and the first and last frame, anyone can compute the current frame.
And this way, the server doesn't have to keep track of the current frame, which if you have 1000s of animated entities, is not an insignificant amount of processing.
In the spirit of having the Entity Server be a spatial database, it should do as little work as possible.

This will handle the case of a client disconnecting from the server, but on top of that, this will handle the entity server itself restarting/shutting down.

(bare with my terrible example)
Let's say I want to make a clock and move the hands with an animation.
If the server keeps track of the current frame, then the clock will run a bit slower on each restart.
But if we use an animation startup timestamp then the server restarting won't matter. The server could be shut down for a month and still be on time when it's eventually brought back up.

My suggestion is that we change the AnimationPropertyGroup to have a `startTimestamp` instead of a `currentFrame` and do the frame computation on each client.


cc @jherico @samcake @birarda

amantley:
Hi,  I have a property that I have defined already that can do what Clement is talking about.  I will rename it and I will make it be updated in the the model property dialogue.  

Thanks Clement, this clarified for me what needs to be done to finish this up.

birarda:
That approach sounds generally good to me. 

I imagine you'll need to use the measured clock skew between you and the server. Keep in mind that our values for skew are approximate and might not be exact (we rely on 1/2 the round trip time between nodes, but it might be asymmetric). As long as you're okay with that error, then you're good.

Atlante45:
That issue exists right now (or rather used to). Since the server computes the current frame and sends it to you, when you get that value, at least as old as the time it takes a packet to get to you, and that's without counting packets NAKed and resent.

ZappoMan:
@Atlante45 - your understanding of what the server tells the client is incomplete. The client gets a "last simulated" epoch as well as a current simulated state. This information is sufficient to always calculate the rendered simulated state accurately... there is no drift or lag associated with networking roundtrips.

ZappoMan:
we absolutely should not need to add additional time stamp properties to solve this problem. The server and client already have enough information to correctly handle this. As for the concern about the CPU cost of several thousand simulated entities, the design allows for the simulation to scale by adjusting the update frequency with available CPU. Ultimately the simulation can operate at a very low frequency and the clients and servers still can stay in sync with the existing information (last simulated time + last simulated state). In the case of the generic entity simulation, some simulations can cause tree state changes: like ballistic trajectories. We do want to run these simulations "at some reasonable rate" so that the new observers will get quality view frustum queries.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5139/HighFidelity-Beta-PR11804-522dc3dfb156c078e2158006e4e2d7adfa6d564f.dmg

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -402,7 +402,8 @@ void EntityTreeRenderer::update(bool simulate) {
     PerformanceTimer perfTimer("ETRupdate");
     if (_tree && !_shuttingDown) {
         EntityTreePointer tree = std::static_pointer_cast<EntityTree>(_tree);
-        tree->update(simulate);
+        //angus
+        //tree->update(simulate);
```
ZappoMan: you probably need this right?

libraries/entities/src/AnimationPropertyGroup.cpp
```diff
@@ -183,6 +209,8 @@ bool AnimationPropertyGroup::appendToEditPacket(OctreePacketData* packetData,
     APPEND_ENTITY_PROPERTY(PROP_ANIMATION_FIRST_FRAME, getFirstFrame());
     APPEND_ENTITY_PROPERTY(PROP_ANIMATION_LAST_FRAME, getLastFrame());
     APPEND_ENTITY_PROPERTY(PROP_ANIMATION_HOLD, getHold());
+    //angus
+    APPEND_ENTITY_PROPERTY(PROP_ANIMATION_CURRENTLY_PLAYING_FRAME, getCurrentlyPlayingFrame());
```
ZappoMan: you shouldn't need a new property.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5206/HighFidelity-Beta-PR11804-d607ca0914445d26e9d23bbeed17517c0ba3a315.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/5206/HighFidelity-Beta-PR11804-d607ca0914445d26e9d23bbeed17517c0ba3a315.exe

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/5226/HighFidelity-Beta-PR11804-101e67498ad643f4fedb053f589e6075e57c7991.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5226/HighFidelity-Beta-PR11804-101e67498ad643f4fedb053f589e6075e57c7991.dmg

amantley:
@Zappoman

I should have branched to avoid pushing all this stuff to the pr.  Everything seems to work now, so this will be QA ready tomorrow morning.

YuppyPlays:
Awaiting build for QA.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5248/HighFidelity-Beta-PR11804-64d5b2fe653ba9f688095cc31b0627a02c650bf8.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/5248/HighFidelity-Beta-PR11804-64d5b2fe653ba9f688095cc31b0627a02c650bf8.exe

YuppyPlays:
Testing halted.

**Fail** I don't think this is particularly normal and I don't remember the FBX looking like this?..

image

image

image

amantley:
This PR is now ready for QA.  
I changed the test case C2957 in TestRail to include the test suggested by Zappoman to test that the server is setting the current frame of the animation and that all the clients are seeing the same point of the animation.

amantley:
@YuppyPlays 
Have you tried going looking at the entity properties and hitting "reset dimensions"

YuppyPlays:
@amantley 
Just tried it, got this:
image


Looks more normal but seems to be strange transparency and warping in certain areas (Such as shoes warping and the hands being seethrough)

Should I continue with this or does something need changing?

amantley:
@YuppyPlays 
Hi if the animation will play that is all that is important.  The character is non-standard and is just to test an entity animation control.

YuppyPlays:
@amantley 
Ah I see, I will continue now.

Testing.

YuppyPlays:
Testing complete:

**Fail**

Please see TestRail Results: https://highfidelity.testrail.net/index.php?/tests/view/18298

amantley:
@YuppyPlays 
Hi there were two mistakes in the test plan, which I have corrected
1) in step 1, when you restart the animation it should start from First frame not the Animation frame
2) in step 8 it should say to set the FPS to 1, that way 60 seconds is 60 frames.

YuppyPlays:
@amantley 
I see, no problem will test now.

Testing.

YuppyPlays:
Testing complete:

**Fail** on Step 9 & 10 as 2nd client just sees T-pose of FBX.

TsetRail Results: https://highfidelity.testrail.net/index.php?/tests/view/21347

amantley:
@YuppyPlays 
Hi, I just tried 9 and 10 and they worked fine.  At first I had an issue with the motion being frozen in the second client as you described but then I changed the path of the right12.fbx file to be relative to atp: and everything worked.  So in conclusion if you followed the set up instructions and used an absolute path for the motion then that will break the motion.  Therefore you must add right12.fbx to the asset server.

YuppyPlays:
@amantley 

TestRail test plan may need a slight adjustment to reflect this for future testing.

amantley:
@YuppyPlays 
ok I just updated the preconditions to reflect the new path.

YuppyPlays:
@amantley excellent. much appreciated :)

Would this need a further QA on 9 & 10 or is this ok to go approved with the mentioned methods you got it working with?

amantley:
@YuppyPlays 
I'm not sure what is standard procedure.  I'm happy with the current functionality.
It was tricky getting all the domain permissions to view the domain with the animation from the second client, but that is not related to this PR.

amantley:
@YuppyPlays 
Hi this needs further QA on cases 9 and 10 using the updated path instructions for the animation right12.fbx

YuppyPlays:
@amantley 
No problem, will start in 2 secs, just setting up secondary laptop to test with :)

YuppyPlays:
Testing.

YuppyPlays:
Testing complete.

**Solid Pass**

TestRail Results: https://highfidelity.testrail.net/index.php?/tests/view/21927&group_by=cases:section_id&group_order=asc&group_id=733

themelissabrown:
build this please

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5439/HighFidelity-Beta-PR11804-64d5b2fe653ba9f688095cc31b0627a02c650bf8.dmg

libraries/entities/src/ModelEntityItem.cpp
```diff
@@ -186,6 +188,117 @@ void ModelEntityItem::appendSubclassData(OctreePacketData* packetData, EncodeBit
...
+    }
+
+    auto now = usecTimestampNow();
+
+    // update the interval since the last animation.
+    auto interval = now - _lastAnimated;
+    _lastAnimated = now;
+
+    // if fps is negative then increment timestamp and return.
+    if (getAnimationFPS() < 0.0) {
```
AndrewMeadows: There is a new warning here because `0.0` (a `double`) is implicitly being converted to `float`.

`0.0` --> `0.0f`

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -736,7 +738,7 @@ void EntityTreeRenderer::mouseReleaseEvent(QMouseEvent* event) {
     PickRay ray = _viewState->computePickRay(event->x(), event->y());
     RayToEntityIntersectionResult rayPickResult = _getPrevRayPickResultOperator(_mouseRayPickID);
     if (rayPickResult.intersects && rayPickResult.entity) {
-        //qCDebug(entitiesrenderer) << "mouseReleaseEvent over entity:" << rayPickResult.entityID;
+        // qCDebug(entitiesrenderer) << "mouseReleaseEvent over entity:" << rayPickResult.entityID;
```
AndrewMeadows: This is pure nitpick, not a blocker: commented out code does not need (and shouldn't have IMO) a space after the `//`.

libraries/entities/src/ModelEntityItem.cpp
```diff
@@ -186,6 +188,117 @@ void ModelEntityItem::appendSubclassData(OctreePacketData* packetData, EncodeBit
...
+    int updatedFrameCount = getAnimationLastFrame() - getAnimationFirstFrame() + 1;
+    
+    if (!getAnimationHold() && getAnimationIsPlaying()) {
+        float deltaTime = (float)interval / (float)USECS_PER_SECOND;
+        _currentFrame += (deltaTime * getAnimationFPS());
+        if (_currentFrame > getAnimationLastFrame()) {
+            if (getAnimationLoop()) {
+                while ((_currentFrame - getAnimationFirstFrame()) > (updatedFrameCount - 1)) {
+                    _currentFrame -= (updatedFrameCount - 1);
+                }
```
AndrewMeadows: My Spidey sense tells me there is danger in this code.  When I look at it more carefully I come up with these reasons:

(1) `_currentFrame` should never accidentally be some arbitrarily large value, such as `1e9`, but if it were, this `while` loop logic might spin a long time to achieve the result it wants.

(2) if `(updatedfFrameCount - 1)` is zero the `while` loop will spinlock.

(3) if (`updatedFrameCount -1)` is negative the `while` loop will also spinlock (because _currentFrame is `float`).

Maybe these odd cases are impossible according to code elsewhere, but it isn't obvious when I read **this** code.

I believe the intention of this code can be achieved without a `while` loop.  I would suggest the exact solution here but when I try to formulate it I realize I would want to rewrite the code between lines 276 **(1*)** **(2*)** through 299 **(3*)**, and in order to do that I would have to understand all of the various cases being handled.

**(1*)** `updatedFrameCount` is computed high up, but is not used in all cases below.  I would want to only compute it where it is actually used.

**(2*)** There are implicit casts from `int` to `float` throughout this logic.

**(3*)** Unnecessary whitespace at end of function.

amantley:
@AndrewMeadows 
Thanks for the review.  I meant to remove that while loop but I'll get now.

amantley:
@AndrewMeadows 
I addressed the issues you raised except the possible changes to Animation Property Group 

@YuppyPlays 
This needs another QA to make sure nothing got broken.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/5592/HighFidelity-Beta-PR11804-079d9639e477844bdb79f213cafc21334f15f806.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5592/HighFidelity-Beta-PR11804-079d9639e477844bdb79f213cafc21334f15f806.dmg

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

**Pass** but note PR crashed twice during, unsure if PR specific but I haven't had a PR build crash in quite a while.
TestRail Results: https://highfidelity.testrail.net/index.php?/tests/view/24297&group_by=cases:section_id&group_order=asc&group_id=733

amantley:
@YuppyPlays 
Hi, I can't recreate the crash on my machine.  I have tried looking at the animation while it plays and toggling all the animation menu items.

amantley:
@YuppyPlays 
I deleted some debug print statements and an unused else statement.
So it will need another QA

YuppyPlays:
@amantley 

Strange, I'll see if I get it again with this build and will attach logs if necessary.

Awaiting build for QA.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5634/HighFidelity-Beta-PR11804-5bc113c81844d9d7753c8214ab67dd6f41f60d7c.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/5634/HighFidelity-Beta-PR11804-5bc113c81844d9d7753c8214ab67dd6f41f60d7c.exe

YuppyPlays:
Testing.

YuppyPlays:
Testing complete:

Still getting crashes on my main PC and this time had it happen on the laptop also, bugsplats report from both have been sent. Other than the crashing issues I experienced though, it's a **Pass**
However, I must note once more I have not had these crashes in recent PR's, so I am unsure as to what is causing it. Bugsplat hopefully sheds some light on why it crashed.

TestRail Results: https://highfidelity.testrail.net/index.php?/tests/view/24298&group_by=cases:section_id&group_order=asc&group_id=733

amantley:
@YuppyPlays 
@ZappoMan 
I can't still get interface to crash as you describe and I don't see anything in bugsplats that relates.
But... I noticed that in my set up I have an absolute path for the right12.fbx animation.  When I then tried to load the animation into the asset browser, interface crashed.

amantley:
I just restarted my domain-server.exe and now everything works fine again, even with the relative atp path for the animation.

YuppyPlays:
@amantley 

I used the exact method given in TestRails test plan which required 2 of the 3 files to be on the ATP, not sure if this helps at all.

amantley:
I’m not using the fst file so you could try deleting that

PR #11826, opened 2017-11-16
Fix for black screen on HMDs on startup

jherico:
Fix for [FB9107](https://highfidelity.fogbugz.com/f/cases/9107/Logging-in-with-HMD-results-in-black-world).  

I don't quite understand why this change fixes rendering when you start in HMD mode.  I suspect that by making a direct connection the first present and subsequent render happens earlier than it otherwise would, leading to the use of some uninitialized value that remains in that state until you switch display plugins.  

An alternative solution might be to suppress the behavior in `onPresent` until after a certain point in startup, but since this is a P1 that needs an immediate fix and I'm uncertain as to the proper fix I'm submitting this as is.  

## Testing

Set an environment variable of `HIFI_DEBUG_HMD=1` on your system.  

* Start Interface 
* Switch to the `HMD Simulator` display plugin in the Display menu
* Shut down Interface
* Restart interface

On master on the second start you will see a black screen, potentially with the cursor and other UI elements on the HUD view, but no scene, not even a skybox.  In this build you should see the scene of whatever environment your client is in, or the default skybox if you're not connected to any domain.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -1823,6 +1823,9 @@ Application::Application(int& argc, char** argv, QElapsedTimer& startupTimer, bo
     // Preload Tablet sounds
     DependencyManager::get<TabletScriptingInterface>()->preloadSounds();
 
+    _pendingIdleEvent = false;
+    _pendingRenderEvent = false;
+
     qCDebug(interfaceapp) << "Metaverse session ID is" << uuidStringWithoutCurlyBraces(accountManager->getSessionID());
 }
 
--- a/interface/src/Application.h
+++ b/interface/src/Application.h
@@ -708,7 +708,7 @@ private:
 
     friend class RenderEventHandler;
 
-    std::atomic<bool> _pendingIdleEvent { false };
-    std::atomic<bool> _pendingRenderEvent { false };
+    std::atomic<bool> _pendingIdleEvent { true };
+    std::atomic<bool> _pendingRenderEvent { true };
 };
 #endif // hifi_Application_h


hyperlogic:
Reverting this change might have some consequences.  Specifically, the main thread event queue will become congested in situations where the present thread is running at a higher frequency then the update a.k.a main thread.   It was introduced in PR #11593, which was introduced in an effort to increase responsiveness of the tablet.  I'd be concerned that there's a regression in tablet performance in heavy domains such as dev-welcome.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/4988/HighFidelity-Beta-PR11826-abd9488becb15ee7b12867fcfefa2a06b89931fa.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/4988/HighFidelity-Beta-PR11826-abd9488becb15ee7b12867fcfefa2a06b89931fa.dmg

Nex-Pro:
@jherico I just see it listed as "open VR (Vive)" in the display menu.

It works fine. I am no longer getting the black screen that I have to fix using ctrl+1 then ctrl+4

jherico:
@hyperlogic I don't think this will cause much of a regression, if any.  It adds one async event per frame, but it still uses your logic to prevent additional idle and render events from entering the queue while an event is pending.  This will add one additional event per frame presented on the present thread, which should be at most 90 events per second.  If that's going to topple us over, we have bigger problems.

conklin94122:
@ZappoMan Thoughts on the concerns of @hyperlogic, above?

ZappoMan:
This PR looks very risky to me.

ZappoMan:
I would say we need to do extensive performance testing to ensure this doesn't cause regression in performance.

jherico:
@hyperlogic @ZappoMan I've changed the PR.  Instead of modifying the connection type it suppresses the idle and render events until the Application constructor has completed.  This seems to work, which in turn seems to validate the idea that the issue is caused by the present event arriving earlier for HMD devices, before some critical value has been set.

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/5006/HighFidelity-Beta-PR11826-7bbc719d799b0f78a79e5288792e6db2907ceaf1.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5006/HighFidelity-Beta-PR11826-7bbc719d799b0f78a79e5288792e6db2907ceaf1.dmg

ZappoMan:
Looks good to me!

cain-kilgore:
# Test Passed
Test | Result
--- | ---
Start Interface | Pass
Switch to the HMD Simulator display plugin in the Display menu | Pass
Shut down Interface | Pass
Restart interface | Pass

Also a pass on master branch. Looks good!

PR #11853, opened 2017-11-19
8992 Login window is small in tablet

ElderOrb:
**Test plan**

Original steps to reproduce (from FB: https://highfidelity.fogbugz.com/f/cases/8992/Login-window-is-small-in-tablet) 

> This only seemed to be a problem when it automatically popped up when going to a restricted domain. When I logged in through the menu the login screen looked good.

I had no access to restricted domains so I had to do the following to reproduce the issue:

1. Switch to VR
2. Open tablet, go to any domain available
3. Launch 'server' and 'assignment-cilent' 
4. Switch to 'localhost' domain
5. Change permissions for local domain to be like the ones below: 

image

Before the fix: 

6. Open tablet and wait a bit. Tablet should auto-refresh and broken login dialog should appear. 

After the fix: 

6. Open tablet and wait a bit. Tablet should auto-refresh and normally-looking login dialog should appear. 

For devs: I've evaluated a few options besides delayed 'pushOntoStack', but this fix is just minimally intrusive one. I'm opened to any suggestions how to make better. Original issue happens because login dialog appears too fast, before tablet gets its real geometry. I've tried to fix geometry but realized it might lead to a lot of regressions because login dialog calculates not only own geometry but also tries to adjust geometry of its parent.

### DIFF
--- a/libraries/ui/src/ui/TabletScriptingInterface.cpp
+++ b/libraries/ui/src/ui/TabletScriptingInterface.cpp
@@ -418,7 +418,10 @@ void TabletProxy::setQmlTabletRoot(OffscreenQmlSurface* qmlOffscreenSurface) {
 
         if (_initialScreen) {
             if (!_showRunningScripts && _initialPath.second == State::QML) {
-                pushOntoStack(_initialPath.first);
+                auto path = _initialPath.first;
+                QTimer::singleShot(0, [this, path] { // delay pushing a bit to allow tablet to get proper geometry
+                    pushOntoStack(path);
+                });
             } else if (_initialPath.second == State::Web) {
                 QVariant webUrl = _initialPath.first;
                 QVariant scriptUrl = _initialWebPathParams.first;


hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/5083/HighFidelity-Beta-PR11853-15248bd2dde5a7a66157250e21845d4c772c3647.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5083/HighFidelity-Beta-PR11853-15248bd2dde5a7a66157250e21845d4c772c3647.dmg

ZappoMan:
you need to merge upstream from latest master or rebase

ElderOrb:
done

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5088/HighFidelity-Beta-PR11853-79353bb852bf84bce16597e60060f1adcaf5b000.dmg

themelissabrown:
@ElderOrb can you provide a test plan for this?

Nex-Pro:
Testing

Nex-Pro:
### Fail

I get a black screen on the tablet.
c15852-hifi-snap-by--on-2017-11-22_17-44-47

I never get the login screen unless I deliberately click on the login at the top ( ***after clicking the X*** ) or using the menu.
Perms for the domain are set as per the instructions.

RyanDowne:
This fogbugz ticket is tagged with Milestone 60 but this PR is not. Does it need to be?

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5144/HighFidelity-Beta-PR11853-ac36f678212f66387429895d07ebb04dfecfc650.dmg

ElderOrb:
@Nex-Pro please, re-check. I've updated code to resolve merge conflict.

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5146/HighFidelity-Beta-PR11853-b7ccd324695e04e3c945cc6e9a29123596533b00.dmg

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/5173/HighFidelity-Beta-PR11853-b7ccd324695e04e3c945cc6e9a29123596533b00.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5173/HighFidelity-Beta-PR11853-b7ccd324695e04e3c945cc6e9a29123596533b00.dmg

Nex-Pro:
### PASS

Managed to get the login screen to show without any intervention and each time it was correctly sized and usable. 

Aesthetics aren't great.

hifi-snap-by--on-2017-11-27_22-52-52
hifi-snap-by--on-2017-11-27_23-04-11

ElderOrb:
@Nex-Pro, agree, but it probably deserves separate PR

libraries/ui/src/ui/TabletScriptingInterface.cpp
```diff
@@ -418,7 +418,10 @@ void TabletProxy::setQmlTabletRoot(OffscreenQmlSurface* qmlOffscreenSurface) {
 
         if (_initialScreen) {
             if (!_showRunningScripts && _initialPath.second == State::QML) {
-                pushOntoStack(_initialPath.first);
+                auto path = _initialPath.first;
+                QTimer::singleShot(0, [this, path] { // delay pushing a bit to allow tablet to get proper geometry
+                    pushOntoStack(path);
+                });
```
AndrewMeadows: The first argument of `QTimer::singleShot()` is a value of msec.  So here we're supplying a zero msec delay.

This looks like a `HACK`.  It looks fragile and I wonder things like: "Does it work on all platforms?", "Is this solution going to break under a future Qt version upgrade?".

Hacks are ok, especially when they are quick workarounds for critical bugs, or just mysterious bugs, but I prefer that they be labled as `HACK` or `WORKAROUND` in comments.

I'm wondering what other devs think about this.  We should get a second opinion.

ElderOrb: Right, this is I would say *well-known* hack :). The idea is just 'enqueue' pushing to even loop, so that it will happen after resize. As for the concerns about other platforms - unless Qt has some bugs it should theoretically work.

druiz17: Does the issue you are trying to address happen for any other QML file besides the login page?

ElderOrb: No, login only as far as I know. And the reason for this - is complicated geometry calculation which login dialog is executing.

ElderOrb: I'm talking about this function: 

        function resize() {
            var targetWidth = Math.max(titleWidth, mainTextContainer.contentWidth)
            var targetHeight = mainTextContainer.height + 3 * hifi.dimensions.contentSpacing.y + buttons.height

            parent.width = root.width = Math.max(d.minWidth, Math.min(d.maxWidth, targetWidth))
            parent.height = root.height = Math.max(d.minHeight, Math.min(d.maxHeight, targetHeight))
        }

druiz17: from the C++ or QML side?

cain-kilgore:
Testing.

BillWindwalker:
Maybe small do to Tablet screen size is less then 15" 🐴

cain-kilgore:
# Test Passed
image

Appears okay to me.

PR #11904, opened 2017-11-30
Fix adjusting equipped entities

druiz17:
Fix adjusting equipped entities

ticket - https://highfidelity.fogbugz.com/f/cases/9774/Attempting-to-adjust-an-equipped-item-with-the-other-hand-unequips-it

### DIFF
--- a/scripts/system/controllers/controllerModules/nearParentGrabEntity.js
+++ b/scripts/system/controllers/controllerModules/nearParentGrabEntity.js
@@ -81,7 +81,8 @@ Script.include("/~/system/libraries/cloneEntityUtils.js");
         };
 
         this.otherHandIsParent = function(props) {
-            return this.getOtherModule().thisHandIsParent(props);
+            var otherModule = this.getOtherModule();
+            return (otherModule.thisHandIsParent(props) && otherModule.grabbing);
         };
 
         this.startNearParentingGrabEntity = function (controllerData, targetProps) {


druiz17:
Test plan 
make sure the step 9 works in this test case
- https://highfidelity.testrail.net/index.php?/tests/view/17831

Nex-Pro:
Awaiting win build

hifi-gustavo:
Windows build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dwindows/5305/HighFidelity-Beta-PR11904-eaa049863d49bab559845511f8041da529e44205.exe

hifi-gustavo:
Mac OS X build is available at: https://deployment.highfidelity.com/jobs/pr-build/label%3Dmacos/5305/HighFidelity-Beta-PR11904-eaa049863d49bab559845511f8041da529e44205.dmg

YuppyPlays:
Testing w/ Cain.

YuppyPlays:
Testing complete:

[Used 2 Windows computers]

Test Step 9:
User A, with an object equipped in their left hand, uses their right hand to adjust the objects position (gun is held in typical position, other hand takes it, tilts it to be barrel-up, left hand takes gun from right hand, gun is equipped with barrel pointing up) **I don't think this is a pass, if I'm understanding this correctly..**
My result for this:
Left hand equipped weapon - I adjust gun slightly, lets say pointing down.
Change to Right hand, it is in default position when equipped, I adjust it to point the barrel up.
Change back to left hand, barrel is NOT pointing up but pointing down as per it's last position in that hand.

I think that makes this a **Fail** - but would like to know devs thoughts.

Expected Result: 
The object in User A's hand is adjusted according to how they manipulated it **Check?..  Please read previous**

Nex-Pro:
@druiz17 

Step 9:

***"User A, with an object equipped in their left hand, uses their right hand to adjust the objects position (gun is held in typical position, other hand takes it, tilts it to be barrel-up, left hand takes gun from right hand, gun is equipped with barrel pointing up)"***

This unfortunately is not clear. 

There are several issues with that 'Step' and until they are resolved no useful information can be gleaned from attempting to reproduce it.

However, once the item has been positioned with the right hand then dropped. ( You cannot remove an equipped item with the other hand AFAIK ( arms aren't long enough ) If the user then re-equips with the left hand the item is in the orientation and position that was set with the right hand.

If the same actions are carried out with the right hand equipping and the left hand adjusting. When the item is grabbed with the right hand it is equipped in the position and orientation set by the left hand.

( I think that near or far grabbing an equippable item with one hand prevents the other from equipping. )
It must be released by the other hand first.

The positioning is related to the object UUID only so any other pop-guns equip as normal.

druiz17:
@Nex-Pro If you have an entity equipped with one have can you adjust that entity with the other hand?

Nex-Pro:
@druiz17 could you clarify please?

druiz17:
What I am trying to say if you equip an item like the ping-pong gun onto your left hand...can you grab and adjust the ping-pong gun with your right hand. Once you release the item with the right hand the gun should still be equipped to your left hand and have the adjusted position. I am asking if this works.

Nex-Pro:
@druiz17 Yes! it works. Also it remembers the position if the item is dropped and then re-equipped with the same hand. There is an association between the the remembered position/orientation and the UUID of the equipped item. Works correctly for either hand.

